PR #10, opened 2015-03-10
Pipes

jamiees2:
I improved the previous method of returning a tuple from an alias to instead capture output.
This makes for easier alias programming, as there is no worry about somehow holding on to the output to return at the end of the function.

Another thing that this makes possible is to create an alias which perhaps does some sort of long lookup which outputs each step of the way, for example a database lookup.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -288,11 +288,15 @@ def run_subproc(cmds, captured=True):
                 stdin.seek(0)
                 stdin, _ = stdin.read(), stdin.close()
 
-            # Redirect the output streams temporarily
-            new_stdout, new_stderr = StringIO(), StringIO()
-            with redirect_stdout(new_stdout), redirect_stderr(new_stderr):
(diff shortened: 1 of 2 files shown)


scopatz:
Hey @jamiees2 - Could you add the Python Software License to the relevant parts of the code, if you are borrowing from v3.5

scopatz:
Thanks for the PR, btw!

jamiees2:
@scopatz No problem! I added the software license and an explanation for the code.

scopatz:
Thanks @jamiees2! I do like the idea of capturing stdout and stderr naively.  However, I also like the idea of being able to return a tuple of str.  I am going to update it (and the docs) a bit to before merging into master.

jamiees2:
Perhaps the two could be merged at the end? At L296, L299, if the alias returns a tuple, add it to the stdin, stderr output/print it?

scopatz:
Yeah, that is exactly what I was thinking.

PR #11, opened 2015-03-10
cd with no arguments now changes to home dir.

abingham:
This is how cd works in bash, zsh, etc.


abingham:
If there's a more appropriate place to make this change, let me know. But since this is really `cd`-specific behaviour, this seemed reasonable.

scopatz:
Thanks! This was the right place.

PR #15, opened 2015-03-11
changed aliasing to support optional return values, and interactive prompts

jamiees2:
This makes returning values optional, supports tuples of both sizes, along with a return type str, and most importantly, supports interactive output.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -289,16 +289,28 @@ def run_subproc(cmds, captured=True):
                 stdin, _ = stdin.read(), stdin.close()
             # Redirect the output streams temporarily. merge with possible
             # return values from function.
-            new_stdout, new_stderr = StringIO(), StringIO()
-            with redirect_stdout(new_stdout), redirect_stderr(new_stderr):
+            if stdout is PIPE:
+                new_stdout, new_stderr = StringIO(), StringIO()
+                with redirect_stdout(new_stdout), redirect_stderr(new_stderr):
+                    rtn = alias(cmd[1:], stdin=stdin)
+                proxy_stdout, proxy_stderr = new_stdout.getvalue(), new_stderr.getvalue()
+                if rtn:
+                    if len(rtn) >= 1 and rtn[0]:
+                        proxy_stdout += rtn[0]
+                    if len(rtn) >= 2 and rtn[1]:
+                        proxy_stderr += rtn[1]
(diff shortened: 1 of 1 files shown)


jamiees2:
In essence, the logic behind this is that capturing output should not be neccessary if captured=False, so buffering it makes little sense. Supporting multiple types of return values makes it very extendable and easy to use for new developers.

jamiees2:
A good testcase for the issues with constantly buffering the output is the following:
```python
aliases["bla"] = lambda k, stdin=None: [print(i) for i in range(1000000000)]
```

xonsh/built_ins.py
```diff
@@ -289,16 +289,34 @@ def run_subproc(cmds, captured=True):
...
+            if stdout is PIPE:
+                new_stdout, new_stderr = StringIO(), StringIO()
+                rtn = None
+                with redirect_stdout(new_stdout), redirect_stderr(new_stderr):
+                    rtn = alias(cmd[1:], stdin=stdin)
+                proxy_stdout, proxy_stderr = new_stdout.getvalue(), new_stderr.getvalue()
+                if isinstance(rtn, tuple):
+                    if len(rtn) >= 1 and rtn[0]:
+                        proxy_stdout += rtn[0]
+                    if len(rtn) >= 2 and rtn[1]:
```
scopatz: This behaviour with tuple return types seems strange to me.  I think if it returns a tuple then it should be of length-2, nothing else really makes sense...

xonsh/built_ins.py
```diff
@@ -289,16 +289,34 @@ def run_subproc(cmds, captured=True):
...
+                        proxy_stderr += rtn[1]
+                elif isinstance(rtn, str):
```
scopatz: I like this, though.

scopatz:
Thanks for this @jamiees2 - I have refactored it a bit to move the aliases to their own function.  The other one was getting kinda long :).

scopatz:
Also, the refactor doesn't break your example

scopatz:
Hopefully you agree that it is nicer.

jamiees2:
Yes, works great! Thanks for your hard work!

xonsh/built_ins.py
jamiees2: Yeah, i wrote this at first, and then realized i could support a string return type instead :stuck_out_tongue_winking_eye:

Issue #13, opened 2015-03-10
Multi-line input never terminates


mbollmann:
When starting a multi-line input, I can't seem to ever get out of multi-line mode no matter how many times I press 'enter'.

I tried to do some good old print-debugging in `shell.py`, outputting the contents of `col` and `buf` [immediately after line 106](https://github.com/scopatz/xonsh/blob/master/xonsh/shell.py#L106), and `col` is always zero no matter what I entered before. So it seems that `buf.pop()` gets called for every line I enter after the first, which means that the buffer never holds a syntactically valid input.

For reference, the output of my print-debugging:

```
>>> if True:
col is 0, buf is ['if True:\n']
Try executing:
if True:

SyntaxError!
...   print(1)
col is 0, buf is ['if True:\n', '  print(1)\n']
Try executing:
if True:

SyntaxError!
... 
col is 0, buf is ['if True:\n', '  print(1)\n']
Try executing:
if True:

SyntaxError!
...   print(2)
col is 0, buf is ['if True:\n', '  print(2)\n']
Try executing:
if True:

SyntaxError!
... 
```

Let me know if I can help debug this problem in any way; I don't understand enough of the code so far to know how to proceed from here.

scopatz:
Thanks for reporting @mbollmann. Super important question, what does the following command return:

```console
$ python -c "import readline; print(readline.__file__)"
```

shoyer:
Just installed xonsh and came here to report the exact same issue.
```
shoyer@tcc-shoyer.local ~ $ python -c "import readline; print(readline.__file__)"
/Users/shoyer/miniconda/envs/xonsh/lib/python3.4/lib-dynload/readline.so
```

scopatz:
@crbates - I think that this is the same issue you were seeing too.  This seems mac specific. @mbollmann and @crbates, are you on conda too?

mbollmann:
```
bollmann@underberg ~ $ python3 -c "import readline; print(readline.__file__)"
/usr/lib/python3.4/lib-dynload/readline.cpython-34m-x86_64-linux-gnu.so
```

I'm neither on Mac nor on Conda -- using Xubuntu 14.04 and installed xonsh from source via setup.py.

shoyer:
I'm on OS X and use conda environments to manage applications, but I also installed xonsh from source (well, `pip install https://github.com/scopatz/xonsh/archive/master.zip`).

[...]

PR #19, opened 2015-03-11
Fix for issue "Multi-line input never terminates #13"

torgny:
Only tested manually.


scopatz:
I can confirm that this works on two systems, including a mac that it was broken for before.  Thanks a ton @torgny! This is much cleaner than what was there before. I'll go in and remove the other `RL_POINT` refs later.

scopatz:
RL_POINT has been removed.

PR #30, opened 2015-03-14
fix #28

### LINKED ISSUE #28: Path auto complition for vim returns error
Hi, thanks for this new cool shell, I realy like it! The only problem I've got so far is when I type 'vim ' and then press tab button it prints error message "/usr/share/bash-completion/bash_completion: line 1856: compopt: not currently executing completion function" instead of list of files in a directory. If I press tab again, it shows files as usual, but the error message is still printed to command line, so path becomes unreadable. Example:

    pablo@alien /some/path branch $ vim /usr/share/bash-completion/bash_completion: line 1856: compopt: not currently executing completion function
    /usr/share/bash-completion/bash_completion: line 1856: compopt: not currently executing completion function
    
    .git/                 AUTHORS               GPATH                 MinSizeRel/           benchmark/            dependencies/

wrywerytwreywery:
Bash completion still works when auto-completing paths for vim, but writes to stderr.

This worked for me to fix this problem, and hasn't seemed to create any problems anywhere else, though I'm not sure how general of a fix this is.

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -119,7 +119,7 @@ class Completer(object):
         script = BASH_COMPLETE_SCRIPT.format(filename=fnme, line=line, n=n,
                     func=func, cmd=cmd, end=endidx+1, prefix=prefix, prev=prev)
         out = subprocess.check_output(['bash'], input=script, 
-                                      universal_newlines=True)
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks!  This was effectively the fix I was thinking of.  Glad to see it worked :)

PR #31, opened 2015-03-14
Auto-formatting prompt string

mbollmann:
Changing the prompt in a meaningful way usually requires me to write a function, but I found myself duplicating a lot of the code in `default_prompt()` for that. Small changes could also be made by doing

```
from xonsh import environ
environ.default_prompt_template = ...
```

but this feels a bit hacky. I wondered if it wouldn't be more convenient if $PROMPT would automatically format itself if it's a string. Setting the prompt to a static string seems rather unusual to me. What do you think?

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -39,12 +39,27 @@ default_prompt = ('{BOLD_GREEN}{user}@{hostname}{BOLD_BLUE} '
                   '{cwd}{BOLD_RED}{curr_branch} {BOLD_BLUE}${NO_COLOR} ')
 
 def format_prompt(template=default_prompt):
-    """Formats a xonsh prompt template string."""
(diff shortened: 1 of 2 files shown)


scopatz:
Hey @mbollmann - I like this idea a lot.  Could you maybe add some info to the docstring about what variables are in the format string? ie "hostname is the host name of the computer", etc.

scopatz:
Thanks!

mbollmann:
Maybe this functionality should be mentioned in the tutorial as well?

Merging with #37 -- if accepted -- should be easy as it's just renaming variables.

scopatz:
Thanks @mbollmann! Yes, this should definitely be mentioned in the tutorial somewhere. Or maybe we should start a new doc page on xonsh environment variables.  Either way a PR would be most appreciated :)

PR #36, opened 2015-03-14
Slightly modified cd error messages

wrywerytwreywery:
I would suggest adding a newline to the cd error message about the number of arguments
it requires, and adding a new error message to distinguish between "this
doesn't exist" and "this isn't a directory" as bash does.


scopatz:
Looks great! Thanks!

PR #37, opened 2015-03-14
Attempt to fix #34

### LINKED ISSUE #34: Virtual Terminal Prompt Issues
Running with xonsh as my shell (set via `chsh` rather than starting from within bash), some virtual terminals seem to have trouble with the `PROMPT` environment variable being a Python function rather than a string, leading it to display, e.g., the following as its prompt:

`<function default_prompt at 0x7f5ffba65400>`

Tested with lxterminal, terminator, and gnome-terminal on Ubuntu 14.04 64-bit.

wrywerytwreywery:
I was looking for a fix for #34, and the following has worked for me.  I am honestly not quite sure how things are fitting together here as far as the terminal emulators I have tried, but my current hypothesis is that they actually rely on the `PROMPT` environment variable, and of course they don't know how to interpret it as a Python function.  My proposed solution is to move the xonsh prompt to a variable called `XONSH_PROMPT`, and use that to update a string `PROMPT` (again, a bit hackish, but it seems to work).

I am not sure how best to reconcile this with #31, or whether this is even a change worth making.  What do you think?

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -13,9 +13,9 @@ from subprocess import Popen, PIPE
 from contextlib import contextmanager
 from collections import Sequence, MutableMapping, Iterable, namedtuple
 
-from xonsh.tools import string_types, redirect_stdout, redirect_stderr
+from xonsh.tools import string_types, redirect_stdout, redirect_stderr, is_function_string
 from xonsh.inspectors import Inspector
-from xonsh.environ import default_env, locale_env
+from xonsh.environ import BASE_ENV, default_env, locale_env
 from xonsh.aliases import DEFAULT_ALIASES
 
 ENV = None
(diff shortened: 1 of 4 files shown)


scopatz:
This now needs a rebase or merge from master.

wrywerytwreywery:
Okay; I will try to merge in the changes from master.  My most recent commit seems (at least through a quick test)  to have fixed the issue, even if you open xonsh from within, e.g., bash from within xonsh.

xonsh/built_ins.py
```diff
@@ -45,7 +45,13 @@ def __init__(self, *args, **kwargs):
         if len(args) == 0 and len(kwargs) == 0:
             args = (os.environ,)
         for key, val in dict(*args, **kwargs).items():
-            self[key] = val
+            if is_function_string(val):
+                if key in (ENV or {}):
```
scopatz: It is much faster (and more explicit) for this to be `if ENV is None or key in ENV`

scopatz:
So after thinking about this a bit, I think it might be best if we simply don't include callables in the return of the `Env.detype()`.  This is because no callable can truly be exported, which is the point of the `detype()`. Users starting xonsh back up would get the either the default value or the value that is in their xonshrc file.  Would you mind implementing that solution @wrywerytwreywery? Sorry this is taking so long.

wrywerytwreywery:
That's much cleaner, I agree.  Will do.

scopatz:
Thanks :)

wrywerytwreywery:
Because I screwed up this branch, it will probably be easier to send via a fresh pull request.  Expect that in a few.  I'm pretty new to Github; is it possible just to close this PR?

scopatz:
Sounds good to me.  You can close a PR either by hitting the close button on the website or by deleting the branch on github.

PR #38, opened 2015-03-14
change terminal title to show message or current directory

wrywerytwreywery:
This is a proposed change that lets xonsh update the title of a virtual terminal to show a message (specified via a function or a string in the `XONSH_TITLE` environment variable).  Currently defaults to showing the current working directory.  I do a lot of work in tabbed terminals, and this small addition has made a world of difference for me.

The check in lines 172-173 is so that xonsh doesn't try to set the title when running in a tty.  I am not sure if there is actually a robust cross-platform way to detect whether one is running in a tty or not; if so, then this check should definitely be replaced.

### DIFF
--- a/xonsh/shell.py
+++ b/xonsh/shell.py
@@ -4,6 +4,7 @@ import traceback
 from cmd import Cmd
 import builtins
 from argparse import Namespace
+import sys
 
 from xonsh.execer import Execer
 from xonsh.completer import Completer
@@ -147,6 +148,13 @@ class Shell(Cmd):
             self.reset_buffer()
             self.cmdloop(intro=None)
 
+    def settitle(self):
+        env = builtins.__xonsh_env__
+        t = env.get('XONSH_TITLE', env['PWD'].replace(env['HOME'], '~') + ' | xonsh')
+        if callable(t):
+            t = t()
+        sys.stdout.write("\x1b]2;%s\x07" % t)
+
     @property
     def prompt(self):
         """Obtains the current prompt string."""
@@ -161,4 +169,6 @@ class Shell(Cmd):
                 p = p()
         else:
             p = "set '$PROMPT = ...' $ "
+        if env.get('TERM','linux')!='linux':
+            self.settitle()
         return p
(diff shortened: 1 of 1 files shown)


xonsh/shell.py
```diff
@@ -161,4 +169,6 @@ def prompt(self):
                 p = p()
         else:
             p = "set '$PROMPT = ...' $ "
+        if env.get('TERM','linux')!='linux':
```
scopatz: PEP8 on this line.

xonsh/shell.py
```diff
@@ -147,6 +148,13 @@ def cmdloop(self, intro=None):
             self.reset_buffer()
             self.cmdloop(intro=None)
 
+    def settitle(self):
+        env = builtins.__xonsh_env__
+        t = env.get('XONSH_TITLE', env['PWD'].replace(env['HOME'], '~') + ' | xonsh')
+        if callable(t):
+            t = t()
+        sys.stdout.write("\x1b]2;%s\x07" % t)
```
scopatz: I tend to prefer str.format()

scopatz:
Thanks for the PR! I'll try it tomorrow....

scopatz:
Seems to work on my machine.  I'd be nice to know compatibility, though, I agree.

wrywerytwreywery:
Thanks for mergin!  I'll try to find some time to look into compatability (that last check still seems suspect).

scopatz:
Thanks for the awesome feature @wrywerytwreywery! I did look into the compatability and it seems like the same thing *should* work on Macs at least.  I also updated the logic a bit, (test if None, not == 'linux') etc.

wrywerytwreywery:
One small issue with this behavior is that, on a linux TTY (at least on my machine), the `TERM` variable _is_ set; it is just set to `'linux'`.  It might work to test whether it is not set (becomes `None`) _or_ it is set to `'linux'`, and not try to set the title in that case.

scopatz:
Ahh Ok, I understand.  I think that you are right that both need to be tested.  If you want to open a PR, I'd really appreciate it.

PR #39, opened 2015-03-14
Update to #38 (Virtual Terminal Title)

wrywerytwreywery:
This patch tries to make sure that xonsh only tries to set the title when running in a virtual terminal.

### DIFF
--- a/xonsh/shell.py
+++ b/xonsh/shell.py
@@ -165,7 +165,8 @@ class Shell(Cmd):
             if callable(t):
                 t = t()
         else:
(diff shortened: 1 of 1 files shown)


wrywerytwreywery:
Second commit updates the default title to include the user and hostname, and uses the `format_prompt` function to format the title, which makes it much easier to make a custom title.

scopatz:
Thanks @wrywerytwreywery! A mention of this great feature in the tutorial or elsewhere in the docs would be most welcome.

PR #40, opened 2015-03-14
fix for multiline prompt (given the update from #31)

wrywerytwreywery:
The length of the multiline prompt should be based on the length of the prompt _after_ calling `format_prompt`, not before.


scopatz:
Nice catch! Thanks!

PR #41, opened 2015-03-15
Accept -c argument (a la bash)

wrywerytwreywery:
This patch allows xonsh to run a single command with the -c argument, similar to bash (though not quite 100% compatable since here the command to be run must follow -c directly).

This seems to allow running a single command through ssh to a machine running xonsh (which is what I was hoping to accomplish), e.g. `ssh machine_running_xonsh echo "hello"`


scopatz:
This is fantastic. I was meaning to do this eventually. I am glad that you beat me to it :)

PR #42, opened 2015-03-15
Second attempt at fix for #34

wrywerytwreywery:
Attempt to fix the issue with the repr of Python functions being used as a prompt.


scopatz:
Thanks again @wrywerytwreywery. This is much cleaner.

PR #43, opened 2015-03-15
Prompt documentation (for #31)

mbollmann:
I kept this pretty short for now.

I agree that a separate doc page for environment variables would be nice, but I wasn't sure how to best fit that into the current structure of the docs. Split up the tutorial into several files? Make a subsection "User guide" that contains more detailed information than the tutorial? Just make another page on the same level as the others?


scopatz:
Thanks @mbollmann! I think just making another page on the same level for now would be good enough. We can think about a user guide later as it expands.

PR #47, opened 2015-03-15
Add a #coding declaration to setup.py

katrielalex:
Python2 won't even parse Unicode files without a #coding declaration, because it assumes ascii.

This way, at least you get a sane error (`xonsh currently requires Python 3.4+`).


scopatz:
Thanks a ton @katrielalex!

Issue #54, opened 2015-03-16
Whole traceback thrown when subproc'ing to a non-existing command


hfaran:
This is how bash handles it:
```bash
^_^hfaran@_ ~
$ foobar
-bash: foobar: command not found
```

And here is xonsh:
```python
hfaran@_ ~ $ foobar
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/site-packages/xonsh/shell.py", line 109, in default
    self.execer.exec(code, mode='single', glbs=None, locs=self.ctx)
  File "/usr/local/lib/python3.4/site-packages/xonsh/execer.py", line 110, in exec
    return exec(code, glbs, locs)
  File "<xonsh-code>", line 1, in <module>
  File "/usr/local/lib/python3.4/site-packages/xonsh/built_ins.py", line 383, in subproc_uncaptured
    return run_subproc(cmds, captured=False)
  File "/usr/local/lib/python3.4/site-packages/xonsh/built_ins.py", line 355, in run_subproc
    stdin=stdin, stdout=stdout)
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/subprocess.py", line 859, in __init__
    restore_signals, start_new_session)
  File "/usr/local/Cellar/python3/3.4.3/Frameworks/Python.framework/Versions/3.4/lib/python3.4/subprocess.py", line 1457, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: 'foobar'
```

Ideally, I think it would be preferable not to see the raw exception.

scopatz:
It probably would be, I agree. Thanks for pointing this out @hfaran.  Pull requests welcome, of course.

hfaran:
I would be more than happy to submit a pull request for this (whenever I find the time, which may be later rather than sooner). That being said, in your opinion, what would be good behaviour here (or good output)?

On the one hand, there is the argument, that since is xonsh is actually a Python superset, the whole stacktrace should be shown. On the other hand, xonsh is also a shell, and seeing that stacktrace every time you enter a command that cannot be found presents a usability issue.

I am of the opinion that the internals of a shell should never be exposed, i.e., if a stacktrace is thrown, that means something unexpected has happened and is a bug in the shell.

So for a shell like xonsh, while I consider a `FileNotFoundError` propogating to the top in this context unacceptable, I do consider two things acceptable:
[...]

PR #58, opened 2015-03-16
subprocess mode error message

wrywerytwreywery:
This is a simple first go at fixing #54, prettifying error messages in subprocess mode.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -351,8 +351,12 @@ def run_subproc(cmds, captured=True):
             stdin = PIPE
         else:
             stdin = prev_proc.stdout
-        proc = Popen(aliased_cmd, universal_newlines=uninew, env=ENV.detype(),
-                     stdin=stdin, stdout=stdout)
+        try:
+            proc = Popen(aliased_cmd, universal_newlines=uninew, env=ENV.detype(),
+                         stdin=stdin, stdout=stdout)
+        except FileNotFoundError:
+            print('xonsh: subprocess mode: command not found: {0}'.format(aliased_cmd[0]))
+            return
         if prev_is_proxy:
             proc.communicate(input=prev_proc.stdout)
         procs.append(proc)
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -11,9 +11,9 @@ from io import TextIOWrapper, StringIO
 from glob import glob, iglob
 from subprocess import Popen, PIPE
 from contextlib import contextmanager
-from collections import Sequence, MutableMapping, Iterable, namedtuple
+from collections import Sequence, MutableMapping, Iterable, namedtuple, OrderedDict
 
-from xonsh.tools import string_types, redirect_stdout, redirect_stderr
+from xonsh.tools import string_types, redirect_stdout, redirect_stderr, levenshtein, levenshtein_sort_helper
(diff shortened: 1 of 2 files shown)


wrywerytwreywery:
The most recent commit implements suggestions for typos in commands.  Right now, defaults to making suggestions, and showing a max of 5, and it tries to sort suggestions in some kind of intelligent (but not computationally expensive) way.

Environment variables:
`SUGGEST_COMMANDS` is a boolean indicating whether suggestions should be made
`SUGGEST_ERROR_THRESHOLD` is an int indicating how "close" a command must be to consider it as a suggestion.
`SUGGEST_MAX_NUM` is an int indicating the maximum number of suggestions that should be shown.

Happy to make changes, or to roll back to the simpler error message in 4177c6bc068d5db4b8146b66a3c29df9a86aa717 if that is preferred.

xonsh/tools.py
```diff
@@ -210,3 +210,38 @@ class redirect_stdout(_RedirectStream):
 class redirect_stderr(_RedirectStream):
     """Context manager for temporarily redirecting stderr to another file."""
     _stream = "stderr"
+
+
+# Modified from Public Domain code, by Magnus Lie Hetland
+# from http://hetland.org/coding/python/levenshtein.py 
+def levenshtein(a,b,max_dist=float('inf')):
```
scopatz: PEP8 spaces after commas

xonsh/built_ins.py
```diff
@@ -351,8 +351,39 @@ def run_subproc(cmds, captured=True):
...
         else:
             stdin = prev_proc.stdout
-        proc = Popen(aliased_cmd, universal_newlines=uninew, env=ENV.detype(),
-                     stdin=stdin, stdout=stdout)
+        try:
+            proc = Popen(aliased_cmd, universal_newlines=uninew, env=ENV.detype(),
+                         stdin=stdin, stdout=stdout)
+        except FileNotFoundError:
+            cmd = aliased_cmd[0]
+            print('xonsh: subprocess mode: command not found: {0}'.format(cmd))
```
scopatz: Would you mind moving most of this block to its own function and the calling it from here? This function is already getting kinda long :)

scopatz:
Frankly, this is incredible! I am so happy to see this.  Just a couple of minor comments to address.

scopatz:
Thanks a ton @wrywerytwreywery

xonsh/built_ins.py
wrywerytwreywery: No problem (I am assuming you mean mostly the suggestion stuff should move).  Am I correct in assuming that tools.py is a fine place to dump these kinds of things?  Or should I make a new suggester.py file or something like that?

scopatz: Ummhmm.  I think tools.py is fine.  If it gets much beyond a handful of functions, then a new module is worth it. For now, though, I don't think it is needed.

scopatz:
Also @wrywerytwreywery - please ping me (by leaving a message on the issue) when you think this is ready for me to look at. I basically live off of github emailing me :)

wrywerytwreywery:
Oh, sorry about that!  I think this is ready for a look now.

scopatz:
This is *incredible*!  I can't thank you enough @wrywerytwreywery!

PR #65, opened 2015-03-16
Added cleaner error for commands not found

Samathy:


scopatz:
This seems to be a duplicate of #58 - which has more features.

scopatz:
#58 has been merged in @Samathy. Is that good enough?

Samathy:
Looks great! - I'm happy with this.

PR #67, opened 2015-03-16
Propagate SyntaxError message in exec-ing .xonshrc to the user

naufraghi:


scopatz:
Thanks @naufraghi! This might be even more helpful with the full traceback...

naufraghi:
Do you have some ideas about the level the python inside xonsh sould be exposed to the user?

I mean, a syntax error can be presented as souch, and a `--debug` option can be suggested to the user for a full traceback, or we can always be in `--debug` mode.

I think in the long run hiding tracebacks and better explain errors is a features, perhaps not so important now, given the current user base.

scopatz:
Yes, I think that you are right that we should probably just have a debug mode.  Maybe this would be best as an environment variable (`$XONSH_DEBUG`?).  The user can either start up with `--debug`, which sets this variable, or they can set this variable while xonsh is running.

Issue #68, opened 2015-03-16
Handle paths with spaces


jessejoe:
When you tab complete folders with spaces in them, the spaces are not delimited with `\ `, so you have to instead put quotes around them:

```
jesse@rmbp13 ~/xonsh $ mkdir "folder with spaces"
jesse@rmbp13 ~/xonsh $ ls
folder with spaces
jesse@rmbp13 ~/xonsh $ cd folder with spaces/
                             ^ Hit tab to complete
cd takes 0 or 1 arguments, not 3
jesse@rmbp13 ~/xonsh $ ls folder with spaces/
                             ^ Hit tab to complete
ls: folder: No such file or directory
ls: spaces/: No such file or directory
ls: with: No such file or directory
jesse@rmbp13 ~/xonsh $ cd "folder with spaces"
jesse@rmbp13 ~/xonsh/folder with spaces $
```

Adding the the `\ ` instead doesn't work either (maybe why they aren't added during tab completion?):
```
jesse@rmbp13 ~ $ cd folder\ with\ spaces/
................
...
SyntaxError: <xonsh-code>:1:10: Invalid token '\\'
```

This is on OSX with xonsh `0.1.2`.

scopatz:
Ahh super interesting.  Maybe the tab-completion should add the quotes.  How do you feel about that?

scopatz:
Thanks for reporting @jessejoe!

wrywerytwreywery:
To me, having the completer add the quotes feels like the right thing.  I started working on this, but it looks like it is going to be difficult to get something that feels "right."

That is, it's easy to make the completer add the quotes.  But it seems like it will be hard to get it to complete halfway-finished quoted strings (particularly after a space).  So if we have a directory called `name with spaces`, then it will complete `na` or `"na` or `'na`, but not `"name wi`.

This is maybe fine, but the downside is if we do: tab complete, backspace, tab; then it will not re-complete to the right value.

I will submit a pull request with what I have now in case it helps, but it isn't perfect.

scopatz:
Cool, It seems like the right thing to do to me would be to insert quotes on both sides, though.  I'll check out your PR.

jessehudl:
Adding quotes seems *okay*, that's what `cmd.exe` does in Windows, however it is very annoying when it matches something else, you have to backspace, tab again, possibly backspace again, tab again, etc. This would honestly make it a little too annoying for daily use for me.

jessejoe:
[...]

PR #72, opened 2015-03-17
Auto-completer: wrap paths in quotes where necessary

wrywerytwreywery:
This is a first attempt to have the tab-completer add quotes to paths where necessary.

It does not go all the way (see the discussion for #68), but it may be a start.

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -107,8 +107,15 @@ class Completer(object):
         space = ' '  # intern some strings for faster appending
         slash = '/'
         tilde = '~'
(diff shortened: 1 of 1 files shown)


scopatz:
Hmm this doesn't actually seem to work for me for `ls`, but does with `cd` even though I don't notice anything obviously wrong. This is probably because bash will replace with `"\ ".  There are some definite subtleties to work out here. But this is a good start.

scopatz:
Thanks for the PR @wrywerytwreywery

PR #74, opened 2015-03-17
Attempt to fix #73

### LINKED ISSUE #73: curr_branch not showing up
I have curr_branch in my prompt however the branch doesn't seem to be showing up.

Here is my git version:

```bash
glenbot@Glens-Air ~/Code/budget/personal_finance $ git --version
git version 1.9.5 (Apple Git-50.3)
```

My prompt (which is the default prompt):

```bash
$PROMPT = '{BOLD_GREEN}{user}@{hostname}{BOLD_BLUE} {cwd}{BOLD_RED}{curr_branch} {BOLD_BLUE}${NO_COLOR} '
```
any ideas?

wrywerytwreywery:
Use git directly to get `current_branch` for prompts.  This seems to fix an issue where `current_branch` isn't properly set on some machines, at least for me.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -18,18 +18,12 @@ def current_branch(cwd=None):
     """
     cwd = os.getcwd() if cwd is None else cwd
     try:
-        # note that this is about 10x faster than bash -i "__git_ps1"
-        d = subprocess.check_output(['which', 'git'], cwd=cwd, 
-                                    stderr=subprocess.PIPE,
-                                    universal_newlines=True)
-        d = os.path.dirname(os.path.dirname(d))
-        input = ('source ' + d + '/lib/git-core/git-sh-prompt; '
-                 '__git_ps1 "${1:-%s}"')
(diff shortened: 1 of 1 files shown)


scopatz:
I don't think that this is quite right.  The `__git_ps1` bash function does a lot more than `git rev-parse`, though this would be a good fallback in the event that the first call fails.

scopatz:
Thanks for putting this in, though, @wrywerytwreywery

wrywerytwreywery:
It does feel to me like using the git binaries would be the right way to go (at least in terms of compatability), but I'll trust your judgment.  Closing this request.

scopatz:
Well, I think we still can use the git binary, just as a fallback if `__git_ps1` fails.  The good and bad thing about git is that it is half in C and half in Bash.  They have actually ported functionality from C to Bash.  This seems insane to be, but clearly they are doing something right :).  The point, though, is that there is functionality in git's Bash layer that isn't in the binary.

Issue #44, opened 2015-03-15
Line wrap covers up prompt


wrywerytwreywery:
I haven't yet figured out exactly when/why this happens, but sometimes the input line wraps around to cover the prompt, rather than wrapping onto a different line.  Even when making the terminal much larger, the input seems to want to wrap around the original size of ther terminal or something like that, rather than wrapping at the current size.

![xonsh_wraparound](https://cloud.githubusercontent.com/assets/549404/6657203/7f93ef9a-cb1a-11e4-9b68-d4b4c1f26dda.png)

![xonsh_wraparound2](https://cloud.githubusercontent.com/assets/549404/6657204/a4b8974e-cb1a-11e4-86ee-a7031dd057cf.png)

scopatz:
I can't reproduce this locally, but I bet this comes from readline not computing the size of the line properly.  This might be because the title line is not being ignored. I am going to put in a potential fix in a few min.

scopatz:
What terminal are you using? I have tried this in both xterm and gnome-terminal... My potential fix caused other problems.

scopatz:
Just tried with rxvt and rxvt-unicode and can't trigger this.

scopatz:
For reference the fix was to change the settitle() line to the following:

```python
sys.stdout.write("\001\x1b]2;{0}\x07\002".format(t))
```

Maybe you could try this out and see if it works for you.

wrywerytwreywery:
I am not sure whether there are other ways to trigger this, but I have been able to reproduce consistently by doing the following:

* Open new terminal (small, e.g., 80*24)
* Type enough characters to force wrap
* Before hitting return, resize window to be larger
* Hit Return
* Type enough characters to wrap the old terminal's size

I will try your fix in a few.

wrywerytwreywery:
Changing `settitle` as above didn't seem to fix, nor does removing the call to `settitle`.  

It does seem like a readline issue, since once it is in this failure mode, a whole bunch of readline-ish things (up arrow, for example) break down.  Worse, these errors seem to persist when opening a new terminal window.

Not sure when I'll have a chance, but I will try to poke around and gather some more info.  I don't remember it breaking in this way yesterday; not sure whether that was just luck, or whether something we changed yesterday dit this, or whether it's something specific to my machine.

scopatz:
Well we did revert back to using the `PROMPT` variable from `XONSH_PROMPT`...

icarito:
[...]

PR #75, opened 2015-03-17
minimize calls to reset_screen_size

wrywerytwreywery:
This works for me for fixing #44!  Sorry it took so long...

I do think `reset_screen_size` has to be called every time a new prompt is made for this to work.  As it stands now, this still requires forcing a new prompt after resizing the window, but that's not a real issue.


scopatz:
Ok thanks!  I am going to put in a comment and then merge up to master (after some tests). I am also going to put in some safety checks for people without readline. Also, also, I am going to report to upstream Python.

PR #76, opened 2015-03-17
update oldpwd when using cd

admiralobvious:
The cd alias now updates $OLDPWD and also checks if the argument is a
‘-‘ (single dash) and cd to $OLDPWD if it is. Based on bash’s cd
behaviour. http://ss64.com/bash/cd.html

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -25,9 +25,9 @@ def cd(args, stdin=None):
         return '', 'cd: no such file or directory: {0}\n'.format(d)
     if not os.path.isdir(d):
         return '', 'cd: {0} is not a directory\n'.format(d)
-    env.update({'OLDPWD': os.getcwd()})
+    env['OLDPWD'] = os.getcwd()
     os.chdir(d)
-    env.update({'PWD': os.getcwd()})
+    env['PWD'] = os.getcwd()
     return None, None
 
 def exit(args, stdin=None):
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -11,7 +11,8 @@ def cd(args, stdin=None):
     If no directory is specified (i.e. if `args` is None) then this
     changes to the current user's home directory.
     """
-    cur_oldpwd = builtins.__xonsh_env__['OLDPWD']
+    env = builtins.__xonsh_env__
+    cur_oldpwd = env.get('OLDPWD') or os.getcwd()
     if len(args) == 0:
         d = os.path.expanduser('~')
     elif len(args) == 1:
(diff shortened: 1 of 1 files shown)


xonsh/aliases.py
```diff
@@ -11,16 +11,20 @@ def cd(args, stdin=None):
     If no directory is specified (i.e. if `args` is None) then this
     changes to the current user's home directory.
     """
+    cur_oldpwd = builtins.__xonsh_env__['OLDPWD']
```
scopatz: I think that this would be a bug if `$OLDPWD` is not already set - which it isn't on my system :)  You probably use the `get()` method on the environment with some sensible default, like `os.getcwd()` or None.

scopatz:
Thanks for the PR @admiralobvious.

scopatz:
Hi @admiralobvious - is this ready for another look?

admiralobvious:
Hi, yes, you may check cc143e2.

xonsh/aliases.py
```diff
@@ -11,18 +11,23 @@ def cd(args, stdin=None):
...
     else:
         return '', 'cd takes 0 or 1 arguments, not {0}\n'.format(len(args))
     if not os.path.exists(d):
         return '', 'cd: no such file or directory: {0}\n'.format(d)
     if not os.path.isdir(d):
         return '', 'cd: {0} is not a directory\n'.format(d)
+    env.update({'OLDPWD': os.getcwd()})
     os.chdir(d)
-    builtins.__xonsh_env__['PWD'] = os.getcwd()
+    env.update({'PWD': os.getcwd()})
```
scopatz: I don't think that you should use update for a single key.  It is better to just set the value.

xonsh/aliases.py
```diff
@@ -11,18 +11,23 @@ def cd(args, stdin=None):
...
         d = args[0]
+        if d == '-':
+            d = cur_oldpwd
     else:
         return '', 'cd takes 0 or 1 arguments, not {0}\n'.format(len(args))
     if not os.path.exists(d):
         return '', 'cd: no such file or directory: {0}\n'.format(d)
     if not os.path.isdir(d):
         return '', 'cd: {0} is not a directory\n'.format(d)
+    env.update({'OLDPWD': os.getcwd()})
```
scopatz: same here.

scopatz:
Thanks for putting this in.  A couple of small notes.

admiralobvious:
Fixed in 4a6d101. Not sure what I was thinking, hehe.

scopatz:
Works for me! Thanks a ton @admiralobvious

PR #82, opened 2015-03-17
HTML5 Video format for the demo

SavinaRoja:
Tested this on a local sphinx build, seems to do the job.

Producing video from .gif source:

  1. Explode .gif using imagemagick. `convert -coalesce xonsh-demo.gif xonsh-frame%03d.png`
  2. Encode to video with libav. `avconv -r 10 -f image2 -i "xonsh-frame%03d.png" xonsh-demo.webm`

The .gif has 511 frames, each displaying for 100ms (`identify -format "Frame %s: %Tcs\n" xonsh-demo.gif`). There are some points where it might be nice for the transition to be delayed for a second or so. The .gif frames could be duplicated while maintaining framerate (increases .gif filesize, little change in .webm filesize, simple conversion) or their display times could be extended (no change in .gif filesize, little change in .webm filesize, slightly more complicated conversion). If there is interest in adding the pauses, I would advocate the first option.


SavinaRoja:
To clarify about the pause points, it would be nice to have a short pause to view some of the output before the `clear` command resets the screen.

scopatz:
Thank you so much for this @SavinaRoja!

PR #83, opened 2015-03-17
Updated environ.current_branch to bounce if git isnt installed and check...

glenbot:
... for multiple locations for the git bash scripts


scopatz:
Thanks a ton for this fix @glenbot.

Issue #60, opened 2015-03-16
Tab Completion on  OS X


tim-d-blue:
Tab completion does not work. Instead it appears an actual tab character is inserted.

Python 3.4.3
OS X 10.10.2 (14C1510)
14.1.0 Darwin Kernel Version 14.1.0
xonsh master as of now...

drvinceknight:
Same issue.

scopatz:
Crazy! Thanks for reporting @tim-d-blue. Do you (or @drvinceknight) have readline installed?  What does the following give you?

```console
$ python -c 'import readline'
```

drvinceknight:
That imports without a problem for me...

scopatz:
hmmm

drvinceknight:
I wish I knew how to help...

scopatz:
How did you obtain your version of Python?

scopatz:
In particular, do you know if it was built using libedit?

drvinceknight:
I updated python3 from the macinstaller. So I don't know if it was built
with libedit. I'll check first thing tomorrow (anyone know how I check?).

drvinceknight:
As far as I can tell: no.

Here is the output of:

```
import sysconfig
print(sysconfig.get_config_vars())
```

```
[...]

PR #93, opened 2015-03-18
Tab Completion on OS X

corbinbs:
Some changes to help support tab completion on OS X as discussed in issue #60


scopatz:
Thanks for putting this in @corbinbs!

Issue #92, opened 2015-03-18
Error during use of xonsh scripts


SavinaRoja:
Was testing the use of xonsh for scripting and ran into an error. Would love to help fix if this is a current bug or limitation.

With a basic script, `xonsh-script.sh`:

    #!/usr/bin/xonsh
    
    echo "echo"
    print("print()")

when I execute this script I get the following error from argparse:

    usage: xonsh [-h] [-c COMMAND]
    xonsh: error: unrecognized arguments: ./xonsh-script.sh

Is this expected?

scopatz:
Hi @SavinaRoja - this is a current limitation, and your help would be most appreciated!  We don't have a way for xonsh to take a file as in input argument right now.  It should be a relatively simple fix in `xonsh/main.py`

scopatz:
Feel free to close @SavinaRoja if #95 works for you.

SavinaRoja:
If you execute the script wthout shebang a la `xonsh script.sh`, things work just fine.

But with the shebang and execution a la `./script.sh`, I get the following error:

```
Traceback (most recent call last):
  File "/home/pablo/workspace/xonsh/xonsh/execer.py", line 118, in _parse_ctx_free
    mode=mode, debug_level=self.debug_level)
  File "/home/pablo/workspace/xonsh/xonsh/parser.py", line 271, in parse
    debug=debug_level)
  File "/home/pablo/workspace/venvs/xonsh-dev/lib/python3.4/site-packages/ply/yacc.py", line 265, in parse
    return self.parseopt_notrack(input,lexer,debug,tracking,tokenfunc)
  File "/home/pablo/workspace/venvs/xonsh-dev/lib/python3.4/site-packages/ply/yacc.py", line 921, in parseopt_notrack
    lookahead = get_token()     # Get the next token
  File "/home/pablo/workspace/xonsh/xonsh/lexer.py", line 66, in token
    self.last = self.lexer.token()
  File "/home/pablo/workspace/venvs/xonsh-dev/lib/python3.4/site-packages/ply/lex.py", line 384, in token
    newtok = self.lexerrorf(tok)
  File "/home/pablo/workspace/xonsh/xonsh/lexer.py", line 414, in t_error
    self._error(msg, t)
  File "/home/pablo/workspace/xonsh/xonsh/lexer.py", line 76, in _error
    self.errfunc(msg, location[0], location[1])
[...]

PR #95, opened 2015-03-18
main.py takes script file or script on stdin

wrywerytwreywery:
This is an attempt to fix issue #92.

main.py now takes an optional positional argument, the filename of a script to be run.  If present, xonsh will execute that script and exit, rather than entering the command loop (e.g., `xonsh some_xonsh_script`)

If stdin is non-empty and no filename is specified, xonsh will run the contents of stdin as a xonsh script and exit, rather than entering the command loop (e.g., `echo "print(7+3)" | xonsh`).

### DIFF
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -13,6 +13,11 @@ parser.add_argument('-c',
         dest='command', 
         required=False, 
         default=None)
+parser.add_argument('file',
+        metavar='script-file',
+        help='If present, execute the script contained in script-file and exit',
+        nargs='?',
+        default=None)
 
 def main(argv=None):
     """Main entry point for xonsh cli."""
@@ -21,10 +26,23 @@ def main(argv=None):
 
     shell = Shell()
 
-    if args.command is None:
-        shell.cmdloop()
-    else:
+    if args.command is not None:
+        # run a single command and exit
         shell.default(args.command)
+    elif args.file is not None:
+        if os.path.isfile(args.file):
+            with open(args.file) as f:
+                for line in f:
+                    shell.default(line)
+        else:
+            print('xonsh: {0}: No such file or directory.'.format(args.file))
+    elif not sys.stdin.isatty():
+        # run a script given on stdin
+        for line in sys.stdin:
+            shell.default(line)
+    else:
+        # otherwise, enter the shell
+        shell.cmdloop()
 
 if __name__ == '__main__':
     main()
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -33,14 +33,14 @@ def main(argv=None):
         # run a script contained in a file
         if os.path.isfile(args.file):
             with open(args.file) as f:
-                code = shell.execer.compile(f.read(), mode='exec')
-                shell.execer.exec(code, mode='exec')
+                code = shell.execer.compile(f.read(), mode='exec', glbs=shell.ctx)
+                shell.execer.exec(code, mode='exec', glbs=shell.ctx)
         else:
             print('xonsh: {0}: No such file or directory.'.format(args.file))
     elif not sys.stdin.isatty():
         # run a script given on stdin
-        code = shell.execer.compile(sys.stdin.read(), mode='exec')
-        shell.execer.exec(code, mode='exec')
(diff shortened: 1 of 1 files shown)


SavinaRoja:
I have been spending a bit of time working on this issue. I will give your code a test and chime in.

SavinaRoja:
I like your handling of stdin so that you can pipe in. I would observe merely from a cli design point of view that the file option should be placed in a mutually exclusive group with '-c COMMAND'.

However, as I have encountered in my own tests, this approach doesn't appear entirely sufficient. Try running the following script with xonsh:

    #!/usr/bin/env xonsh
    
    echo "echo"
    print("print")
    
    for i, x in enumerate('xonsh'):
        print(i, x)

The `for` loop does not seem to work.

wrywerytwreywery:
Thanks for testing, and for pointing that out, @SavinaRoja.  I'll work on this some more.

scopatz:
So the for-loop probably doesn't work because it is being sent through the `shell.default()` This will exec the code in `'single'` mode (ie single statements).  For files, we need to exec in `'exec'` mode.  (Python has three execution modes, eval, exec, and single, which you can read about in the exec() and eval() docs).

wrywerytwreywery:
Yep.  Pushing new fix right now.

scopatz:
For file input, you probably need to do the following:

1. instantiate an execer
2. use the `xonsh.built_ins.xonsh_builtins()` with the execer in a with-statement
3. use the `execer.exec()` method in the with-block.

scopatz:
haha cool.

wrywerytwreywery:
Just pushed an update.  Since we already initialized a `Shell` by that point, I was just planning on using its execer...but it looks like that doesn't get, e.g., things from the `xonshrc`.  I'll ping when I have something.

wrywerytwreywery:
Okay, I think this is ready for a look now, @scopatz.

One thing is that in order to run a command (alias or otherwise) from a script, it seems like it _must_ be run in subprocess mode explicitly.  So having `ls` on a line by iteself doesn't execute `ls`, but having `$[ls]` does.  Is this the intended behavior?

wrywerytwreywery:
Actually, just pushed another changeset to get variables in from the xonshrc.  Sorry for so many quick changes.  I think it's _actually_ ready now...

xonsh/main.py
```diff
@@ -21,10 +26,24 @@ def main(argv=None):
...
-    else:
+    if args.command is not None:
+        # run a single command and exit
         shell.default(args.command)
+    elif args.file is not None:
+        # run a script contained in a file
+        if os.path.isfile(args.file):
+            with open(args.file) as f:
+                code = shell.execer.compile(f.read(), mode='exec')
+                shell.execer.exec(code, mode='exec')
```
scopatz: You probably need to pass in `glbs=shell.ctx` or `glbs={}` here.  Otherwise, it is going to use the main.py globals.

xonsh/main.py
```diff
@@ -21,10 +26,24 @@ def main(argv=None):
...
-    if args.command is None:
-        shell.cmdloop()
-    else:
+    if args.command is not None:
+        # run a single command and exit
         shell.default(args.command)
+    elif args.file is not None:
+        # run a script contained in a file
+        if os.path.isfile(args.file):
+            with open(args.file) as f:
```
scopatz: Also, I would read in the file to a local variable, then call exec outside of the file block.

scopatz:
This looks good to me.  Having to write `$[ls]`is not the intended behaviour, but this is great.  Just waiting for Travis.  And no worries about the many small changes.

scopatz:
Awesome! Thanks @wrywerytwreywery!  I added some trailing newline safety checks before merging, btw. 

Some docs on these fantastic features from either @wrywerytwreywery  or @SavinaRoja would be very welcome! 

Thanks again.

wrywerytwreywery:
Woo-hoo!  Maybe tonight I'll try to document this, as well as the other environment variables I introduced, before moving forward with the other features I'm working on.

scopatz:
That'd be fantastic!

scopatz:
I am going to try fixing the parser errors with nested lists tonight. :)

PR #99, opened 2015-03-19
fix for comment parsing

wrywerytwreywery:
This is an attempt to fix the issues with running scripts from #92 by changing the way comments are parsed.


scopatz:
Thanks!  This seems to work and also cleans up an old PLY missing token warning.

PR #103, opened 2015-03-19
Documentation and refactoring of window title and suggestions

wrywerytwreywery:
Documentation for the environment variables for setting the window title and offering suggestions for mis-typed commands.  

I renamed some of these variables and shuffled things around a little bit, in an attempt to avoid screwing up the table of environment variables in the docs, and for more consistent (I think) naming of environment variables.


scopatz:
Thanks for putting this in @wrywerytwreywery!

PR #105, opened 2015-03-19
Removed `subproc_line`, fixed typos, small changes

SmartViking:
`subproc_line` wasn't used and probably contained a bug (I think there should be a `not` in the if clause, only way it makes sense), so I just assumed it could be removed. I could be mistaken.

I haven't actually tested the changes as I'm running python 3.2, so I made sure to not to change behavior, but there's a risk of typo-type errors.

### DIFF
--- a/xonsh/ast.py
+++ b/xonsh/ast.py
@@ -91,7 +91,7 @@ class CtxAwareTransformer(NodeTransformer):
 
     def ctxremove(self, value):
         """Removes a value the most recent context."""
-        for ctx in self.contexts[::-1]:
+        for ctx in reversed(self.contexts):
             if value in ctx:
                 ctx.remove(value)
                 break
@@ -121,7 +121,7 @@ class CtxAwareTransformer(NodeTransformer):
         if lname is None:
             return node
         inscope = False
-        for ctx in self.contexts[::-1]:
+        for ctx in reversed(self.contexts):
             if lname in ctx:
                 inscope = True 
                 break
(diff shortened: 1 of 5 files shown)


scopatz:
It should be!  Thanks @SmartViking. Does xonsh actually work on v3.2? If you find any bugs, please report them.  I'd like for xonsh to support multiple Python versions, but I don't have the personal bandwidth to do so, aside from updates going forward (v3.5+).

SmartViking:
The problems that came up was that `Bytes`, `NameConstant`, `arg`, `NonLocal`, `withitem`, `Try`, `Starred` and `YieldFrom` couldn't be imported from `ast` (in ast.py). It's also possible that it's a problem that `ast.py` imports "from itself" so to speak, I don't know if that's a problem when you use this package stuff.

When I tried to mitigate by removing those from the import, it failed like this:

```
$ python as3.py   #  renamed file because of self-import problem
  File "as3.py", line 15, in <module>
    from xonsh.tools import subproc_toks
ImportError: No module named xonsh.tools
```

I don't know if that's got to do with how I tried to run the program or if that's a newer way of importing stuff in python3.4.

This ast node stuff makes it seem more difficult to support more versions, but my understanding is very superficial.

SmartViking:
I'm using Python 3.2.3 btw

xonsh/ast.py
```diff
@@ -91,7 +91,7 @@ def ctxadd(self, value):
 
     def ctxremove(self, value):
         """Removes a value the most recent context."""
-        for ctx in self.contexts[::-1]:
+        for ctx in reversed(self.contexts):
```
scopatz: Ahh great. Thanks.  Too used to thinking in numpy where the original is a view.

scopatz:
> This ast node stuff makes it seem more difficult to support more versions, but my understanding is very superficial.

Yeah, it does make it a little more difficult.  The AST for python does change slightly from version to version.  Normally it just grows by adding nodes or attributes to nodes that already exist.  But the solution is not hard conceptually, though it might be annoying to implement.

Basically, the AST nodes are imported and in some cases the way that they are used has to be dependent on the Python version number.  In most cases these should be just slight variations of the existing source.  Hunting them all down, though....

scopatz:
Thanks for putting this PR in.

Issue #86, opened 2015-03-18
Syntactic Improvements


scopatz:
See this reddit thread: http://www.reddit.com/r/Python/comments/2z8hl1/xonsh_a_pythonish_bashcompatible_shell_language/cpgxelz

Basically the idea is to have non-modal syntax (${} means to look things up in the env in Python-mode and it means to compute something in Python when in subprocess-mode.)  One of these should get a different syntax.

The other idea is to convert the contents of the ${} expression to a string automatically in subprocess mode.  This would enable the following:

```console
echo ${3 + 4}
```

wrywerytwreywery:
I like this idea.  The syntax didn't bother me, since I viewed `${...}` as sort of the "run this in the other mode" syntax, but I can see how it would be confusing.

My suggestion would be to leave `${...}` as environment lookup in Python mode, and come up with a different syntax for executing Python in subprocess mode.  I feel like the new syntax probably shouldn't use `$`, since that is (I think) pretty strongly tied to the shell, and it seems like we want something that is tied to Python instead.  So my feeling would be something like `py(...)`, though that would have its own problems.

Converting those outputs into strings does make good sense, I think.

scopatz:
> I feel like the new syntax probably shouldn't use $,

I completely agree.  

> something like py(...), though that would have its own problems.

I really like the idea of using parentheses here because they 'capture' and then return the expression inside of them.  When I was thinking about it this morning, I came up with `!()` as the syntax.  `!` has a connotation of execution and is definitely different than `$`.  `|()` could mean 'pipe to python', but I could see that getting super confusing if you are using a lot of pipes on the same line.  

I am open to suggestions here, but I kinda like `!()` and if I had to choose today that would be it.

> Converting those outputs into strings does make good sense, I think.

Sounds good!

wrywerytwreywery:
I kind of like the `!(...)` syntax.

```
for i in range(500):
    $[touch !("file%03d" % i)]
```

My only (slight) reservation is that `!` means something different in BASH-land, and exploring the history using `!` might be something that xonsh should have in the longer-term.

Something that bash doesn't use (at least to my knowledge) is `@`.  It is arguably uglier (and loses the execution connotation), but I could swear I have seen this character used in some string templating languages, so it might not be too bad for denoting "compute a string and fill it in here":

```
for i in range(500):
    $[touch @("file%03d" % i)]
```

But I could go either way on that.  

FWIW, I do think the pipe syntax would be too confusing.

[...]

PR #112, opened 2015-03-20
convert result of ${} expression to string

wrywerytwreywery:
This is an attempt to deal with part of #86 by converting the results of `${...}` expressions to strings automatically.  Now, for example, the following works:

```
$ echo ${20+2.}
22.0
```


scopatz:
Thanks @wrywerytwreywery!

PR #113, opened 2015-03-20
Syntax

wrywerytwreywery:
This is a first go at implementing my suggestion for a new syntax, as discussed in #86.  With this update, `@(...)` evaluates Python code, and `${...}` does environment lookup regardless of context.  Happy to take suggestions for improvement.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -175,6 +175,13 @@ PROMPT             xosh.environ.default_prompt The prompt text, may be str or
 MULTILINE_PROMPT   ``'.'``                     Prompt text for 2nd+ lines of
                                                input, may be str or 
                                                function which returns a str.
+TITLE              xonsh.environ.default_title The title text for the window
+                                               in which xonsh is running.  As
+                                               with PROMPT, may be a str or a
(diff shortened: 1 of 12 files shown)


scopatz:
> ${...} does environment lookup regardless of context.

Awesome thanks!

> @(...) evaluates Python code

I still sort of like the `!(...)` syntax. That doesn't block this PR, because it can always be changed again.

@wrywerytwreywery Do you know why travis failed here?

scopatz:
Oh, you probably need to update the tests.

wrywerytwreywery:
I will go ahead and update the tests.

Is it a compromise to support both `!()` and `@()`?  That would be easy, and neither of them collides with any other features...

scopatz:
> Is it a compromise to support both !() and @()? Or is that just bad design? :)

Haha I think that is just bad design.  As long as I think of `@()` as working like a decorator, in that it takes an expression as input and returns something, I am happy with it.  If it really starts to bother us, we can always change it.

scopatz:
Thanks @wrywerytwreywery! This looks great!

PR #114, opened 2015-03-20
Fall back to using git binary to determine branch when necessary

wrywerytwreywery:
Resurrecting #74 (in response to issue #73), this time using the git binary only as a fallback, for when the current method fails.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -53,7 +53,7 @@ def current_branch(cwd=None):
             s = s.strip()
             if s != '':
                 branch = s
-        except:
+        except subprocess.CalledProcessError:
             pass
 
     return branch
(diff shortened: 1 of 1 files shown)


xonsh/environ.py
```diff
@@ -44,6 +44,18 @@ def current_branch(cwd=None):
...
+    # fall back to using the git binary if the above failed
+    if branch is None:
+        try:
+            s = subprocess.check_output(['git', 'rev-parse','--abbrev-ref', 'HEAD'],
+                    stderr=subprocess.PIPE, cwd=cwd,
+                    universal_newlines=True) 
+            s = s.strip()
+            if s != '':
+                branch = s
+        except:
```
scopatz: This should probably just catch `subprocess.CalledProcessError`.

scopatz:
Thanks for putting this in @wrywerytwreywery

wrywerytwreywery:
I updated as you suggested, so this is ready for another look, I think.

scopatz:
Thanks!

Issue #48, opened 2015-03-15
implement dirstack


TheJJ:
`cd -` and `cd -%d` would be awesome.

This allows switching to some previous directory.

scopatz:
I'd happily accept a PR that implemented this kind of feature :)

wrywerytwreywery:
I'm happy to work on this unless somebody else wants dibs.

mbollmann:
I think it would be great if this functioned as a proper stack similar to `pushd` and `popd` by default. I have functions in my .bashrc that basically replace `cd` with `pushd` and provide functionality to view the full dirstack and easily travel to a specific dir in the stack. Just a suggestion.

TheJJ:
jup, the like the zsh cd feature completes the - with the last n directories visited, so e.g. typing `cd -3` visits the third-last one. `cd -<tab>` shows the whole stack and completes the number.

mbollmann:
That sounds even better than my bash aliases. I'm not familiar with zsh, but this behaviour sounds great.

I won't have the time to implement this before the weekend though, so go ahead @wrywerytwreywery if you like. :)

scopatz:
@TheJJ - #116 has been merged.  Does this satisfy the issue here?

TheJJ:
Nice! `cd -` works now, but `cd -$number` is still missing :wink:
Completion for the number with display of the associated stack value would be ideal.

scopatz:
I think that you can do `cd -number`.

TheJJ:
```
cd -2
cd: no such file or directory: -2
```

scopatz:
Ahh, seems like there might be a bug.  @wrywerytwreywery?

wrywerytwreywery:
`pushd`, `popd`, and `dirs` have been implemented, but `cd -n` hasn't been implemented yet, nor has `$AUTO_PUSHD`.  Those are on my queue.

scopatz:
I believe that this is now implemented.  Please re-open if I am in error

PR #116, opened 2015-03-21
Implement pushd, popd, dirs

wrywerytwreywery:
This changeset implements `pushd`, `popd`, and `dirs`.  This implementation is similar to the bash version, but with the meaning of `+` and `-` switched (as though `$PUSHD_MINUS` were set), since I think that's how most people use it anyway.  I think I got most of the features right, but I would be glad for any feedback.

This does not implement all of the features suggested in #48.  In particular, those involving changes to `cd` and tab completion aren't included here, but will come later.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -437,6 +437,7 @@ def unload_builtins():
              '__xonsh_pyexit__', '__xonsh_pyquit__', 
              '__xonsh_subproc_captured__', '__xonsh_subproc_uncaptured__', 
              '__xonsh_execer__', 'evalx', 'execx', 'compilex', 
+             'default_aliases'
              ]
     for name in names:
         if hasattr(builtins, name):
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -301,7 +301,7 @@ def _run_callable_subproc(alias, cmd, captured=True, prev_proc=None,
 
 def run_subproc(cmds, captured=True):
     """Runs a subprocess, in its many forms. This takes a list of 'commands,'
-    which may be a list of command line arguments or a string, represnting
+    which may be a list of command line arguments or a string, representing
     a special connecting character.  For example::
 
         $ ls | grep wakka
@@ -419,7 +419,7 @@ def load_builtins(execer=None):
     BUILTINS_LOADED = True
 
 def unload_builtins():
-    """Removes the xonsh builtins from the Python builins, if the 
+    """Removes the xonsh builtins from the Python builtins, if the 
     BUILTINS_LOADED is True, sets BUILTINS_LOADED to False, and returns.
     """
     global BUILTINS_LOADED, ENV
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -414,7 +414,7 @@ def load_builtins(execer=None):
     builtins.evalx = None if execer is None else execer.eval
     builtins.execx = None if execer is None else execer.exec
     builtins.compilex = None if execer is None else execer.compile
-    builtins.aliases = Aliases(DEFAULT_ALIASES)
+    builtins.default_aliases = builtins.aliases = Aliases(DEFAULT_ALIASES)
     builtins.aliases.update(bash_aliases())
     BUILTINS_LOADED = True
 
--- a/xonsh/dirstack.py
+++ b/xonsh/dirstack.py
@@ -38,7 +38,7 @@ def pushd(args, stdin=None):
     if args.dir is None:
         try:
             new_pwd = DIRSTACK.pop(0)
-        except:
+        except IndexError:
             e = 'pushd: Directory stack is empty\n'
             return None, e
     elif os.path.isdir(args.dir):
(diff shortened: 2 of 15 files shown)


xonsh/built_ins.py
```diff
@@ -414,7 +414,7 @@ def load_builtins(execer=None):
     builtins.evalx = None if execer is None else execer.eval
     builtins.execx = None if execer is None else execer.exec
     builtins.compilex = None if execer is None else execer.compile
-    builtins.aliases = Aliases(DEFAULT_ALIASES)
+    builtins.default_aliases = builtins.aliases = Aliases(DEFAULT_ALIASES)
```
scopatz: `default_aliases` also needs to be removed in the `unload_builtins()` function.

xonsh/dirstack.py
```diff
@@ -0,0 +1,216 @@
...
+
+DIRSTACK = []
+
+
+def pushd(args, stdin=None):
+    global DIRSTACK
+
+    try:
+        args = pushd_parser.parse_args(args)
+    except SystemExit:
```
scopatz: Does this actually happen?

scopatz: I am curious :)

xonsh/dirstack.py
```diff
@@ -0,0 +1,216 @@
+import os
+import builtins
+from argparse import ArgumentParser
+
+DIRSTACK = []
+
+
+def pushd(args, stdin=None):
```
scopatz: docstring needed.

xonsh/dirstack.py
```diff
@@ -0,0 +1,216 @@
...
+        if e is not None:
+            return None, e
+
+    if not builtins.__xonsh_env__.get('QUIET_PUSHD', False):
+        return dirs([], None)
+
+    return None, None
+
+
+def popd(args, stdin=None):
```
scopatz: docstring needed.

xonsh/dirstack.py
```diff
@@ -0,0 +1,216 @@
...
+        if e is not None:
+            return None, e
+
+    if not builtins.__xonsh_env__.get('QUIET_PUSHD', False):
+        return dirs([], None)
+
+    return None, None
+
+
+def dirs(args, stdin=None):
```
scopatz: docstring needed.

xonsh/dirstack.py
```diff
@@ -0,0 +1,216 @@
...
+        try:
+            new_pwd = DIRSTACK.pop(0)
+        except:
+            e = 'popd: Directory stack is empty\n'
+            return None, e
+    else:
+        try:
+            num = int(args.dir[1:])
+            assert num >= 0
+        except:
```
scopatz: It is much better to do `if num >= 0` than an except.  Also, I think that int will only ever throw a value error.  So that error is what should be caught.

scopatz:
Thanks for implementing this @wrywerytwreywery! There are some issues I have commented on inline. Also, a file in the docs that hooks this new module up through autodocs is needed in addtion to what is here.

scopatz:
Probably should document the PUSHD_MINUS like behaviour somewhere too.

xonsh/dirstack.py
wrywerytwreywery: Yes it does, actually!  If there is an error when parsing, `ArgumentParser` prints the help message and then exits using `SystemExit`.

scopatz: That is so silly!

xonsh/dirstack.py
wrywerytwreywery: Hmmm...the reason I kept them together there and used the catch-all was that I wanted the same error message, both in the case that a negative number is supplied, and in the case that what the user put in isn't a number at all.  To me, this seemed cleaner than repeating the error code in a second place...

scopatz: Well, I don't think we should really ever be using catch-all exceptions.  Maybe that is a little puritanical :) And in general, I think if we can test a condition rather than using an exception, we should. Exceptions being for exceptional cases, etc.

wrywerytwreywery:
Okay, I just pushed a whole bunch of things:
* fixed a bug in `pushd`'s handling of the `-n` flag
* unloaded `default_aliases` from `builtins` (I think)
* switched back to the conventional meanings of `+` and `-`
* added support for `$PUSHD_MINUS` to switch the meanings of `+` and `-`
* added support for `$DIRSTACK_SIZE` to set a maximum number of directories to remember (default 20)
* added support for `$PUSHD_SILENT` to prevent `pushd` and `popd` from calling `dirs` after they are done
* added module to api documentation, and added docstrings for the functions
* refactored per your suggestions

I think this will be ready for a look once travis finishes.  

It might also be worth adding an explanation to the docs.  Is the tutorial the right place for this?

scopatz:
> It might also be worth adding an explanation to the docs. Is the tutorial the right place for this?

I think we are now at the point where the aliases deserve their own page.

xonsh/built_ins.py
```diff
@@ -437,6 +437,7 @@ def unload_builtins():
              '__xonsh_pyexit__', '__xonsh_pyquit__', 
              '__xonsh_subproc_captured__', '__xonsh_subproc_uncaptured__', 
              '__xonsh_execer__', 'evalx', 'execx', 'compilex', 
+             'default_aliases'
```
scopatz: yep, this is correct :)

scopatz:
Thanks a ton @wrywerytwreywery!

PR #119, opened 2015-03-21
Complex numbers

wrywerytwreywery:
Add support for complex numbers syntax (e.g., `1+2j`) to xonsh

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -99,6 +99,15 @@ def test_int_literal():
 def test_float_literal():
     yield check_ast, '42.0'
 
+def test_imag_literal():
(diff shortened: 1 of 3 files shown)


scopatz:
It is so tempting to also add `i` as an option...

scopatz:
A test would be nice here too.  Thanks @wrywerytwreywery!

wrywerytwreywery:
As an engineer, I can't bring myself to support `i` as an alternative to `j` :)

I added a couple of simple test cases for complex numbers, so this should be ready for another look.

scopatz:
Awesome! Thanks!

PR #120, opened 2015-03-21
Modify string regex (allow multiline strings)

wrywerytwreywery:
I believe this fixes the issue from #92, allowing xonsh to handle multline strings.

### DIFF
--- a/xonsh/lexer.py
+++ b/xonsh/lexer.py
@@ -5,6 +5,8 @@ import sys
 from ply import lex
 from ply.lex import TOKEN
 
+def anyof(*regexes):
+    return '(' + '|'.join(regexes) + ')'
 
 class Lexer(object):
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks @wrywerytwreywery!  For some reason multiline strings still don't work in xonshrc, but this is great.

wrywerytwreywery:
Can you paste in what you tried, specifically?  I put the following in my xonshrc as a quick test, and it seems to work:

```
print("""hey
""")

print('''

test

'\'

''')
```

scopatz:
Hah! You are right.  It works.  It was that I didn't have a newline at the end of my xonshrc. I thought that I had fixed that bug.... maybe not.  I'll take a look now.

scopatz:
Ok that bug is now fixed in master (c48b255)

PR #121, opened 2015-03-21
Allow running scripts not in the path

wrywerytwreywery:
This changeset allows xonsh to run scripts with shebang lines as in bash, rather than having to invoke other interpreters directly.  Even though `subprocess` can handle shebang lines on its own, falling back to using xonsh as the default interpreter necessitates doing a little bit of checking.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -300,6 +300,38 @@ def _run_callable_subproc(alias, cmd, captured=True, prev_proc=None,
         proc = ProcProxy(rtnout, rtnerr)
     return proc
 
+def is_script(fname):
+    """
+    Checks whether a file should be considered for running as a script
+    """
+    return os.path.isfile(fname) and fname != os.path.basename(fname)
+
+def get_script_subproc_command(fname, args):
+    """
+    Given the name of a script outside the path, returns a list representing
+    an appropriate subprocess command to execute the script.  Raises
+    PermissionError if the script is not executable.
+    """
+    # make sure file is executable
+    if not os.access(fname, os.X_OK):
+        raise PermissionError
+    
+    # find interpreter
+    with open(fname) as f:
+        first_line = f.readline().strip()
+    m = re.match(r'#![ \t]*(.+?)$', first_line)
+    
+    # xonsh is the default interpreter
+    if m is None:
+        interp = ['xonsh']
+    else:
+        interp = m.group(1).strip()
+        if len(interp) > 0:
+            interp = shlex.split(interp)
+        else:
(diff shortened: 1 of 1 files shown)


xonsh/built_ins.py
```diff
@@ -354,10 +377,16 @@ def run_subproc(cmds, captured=True):
...
+        except PermissionError:
+            cmd = aliased_cmd[0]
+            print('xonsh: subprocess mode: permission denied: {0}'.format(cmd))
+            return
         except FileNotFoundError:
             cmd = aliased_cmd[0]
             print('xonsh: subprocess mode: command not found: {0}'.format(cmd))
-            print(suggest_commands(cmd, ENV, builtins.aliases))
+            s = suggest_commands(cmd, ENV, builtins.aliases)
+            if s.strip():
```
scopatz: should be `if len(s.strip()) > 0`

xonsh/built_ins.py
```diff
@@ -335,7 +337,28 @@ def run_subproc(cmds, captured=True):
...
+                return
+            # find interpreter
+            with open(fname) as f:
+                first_line = f.readline().strip()
+            m = re.match(r'#![ \t]*(.+?)$', first_line)
+            if m is None:
+                interp = ['xonsh']
+            else:
+                interp = m.group(1).strip()
+                if interp != '':
```
scopatz: `if len(interp) > 0`

xonsh/built_ins.py
```diff
@@ -335,7 +337,28 @@ def run_subproc(cmds, captured=True):
         stdout = last_stdout if cmd is last_cmd else PIPE
         uninew = cmd is last_cmd
         alias = builtins.aliases.get(cmd[0], None)
-        if alias is None:
+        if os.path.isfile(cmd[0]) and cmd[0] != os.path.basename(cmd[0]):
```
scopatz: Maybe this block could be its own function, too.

xonsh/built_ins.py
```diff
@@ -335,7 +337,28 @@ def run_subproc(cmds, captured=True):
...
+            aliased_cmd = cmd
+            fname = cmd[0]
+            # make sure file is executable
+            if not (os.access(fname, os.X_OK)):
```
scopatz: no need for grouping parentheses here.

scopatz:
Very cool PR @wrywerytwreywery!  Just a few small issues to clean up.

wrywerytwreywery:
Refactored per your suggestions.  Ready for another look.

scopatz:
Thanks again @wrywerytwreywery!

PR #100, opened 2015-03-19
Nested Lists and Tuples

scopatz:
This should fix #61

Issue #85, opened 2015-03-18
Xonsh interprets certain nested structures incorrectly


wrywerytwreywery:
The following do not work as expected in xonsh:

```
$ [[i] for i in range(5)]
[0, 1, 2, 3, 4]
```
```
$ [(i,) for i in range(5)]
[0, 1, 2, 3, 4]
```
```
$ [[1],[2],[3],[4],]
[1, [2], [3], [4]]
```

I am not sure exactly what the pattern is, but I did notice these failure modes and figured it might be helpful to point them out.  I know there had been some issues in the past with lists inside of dictionaries; maybe this is related?

scopatz:
Thanks for reporting @wrywerytwreywery! This is clearly a parser bug.  I think I know where too look for it.  I'll take a stab tomorrow, because I am too tired tonight.

scopatz:
#100 should address this @wrywerytwreywery

wrywerytwreywery:
Fixed with #100.  Closing.

Issue #61, opened 2015-03-16
Crash on parsing list nested in dict


mbollmann:
When parsing a line containing a dict that has a list (or tuple) as a value, xonsh crashes.

```
bollmann@underberg ~/repositories/xonsh master >>> {'foo': ['a', 'b']}
Traceback (most recent call last):
  File "/usr/local/bin/xonsh", line 12, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/home/bollmann/repositories/xonsh/scripts/xonsh", line 3, in <module>
    main()
  File "/home/bollmann/repositories/xonsh/xonsh/main.py", line 25, in main
    shell.cmdloop()
  File "/home/bollmann/repositories/xonsh/xonsh/shell.py", line 152, in cmdloop
    super(Shell, self).cmdloop(intro=intro)
  File "/usr/lib/python3.4/cmd.py", line 138, in cmdloop
    stop = self.onecmd(line)
  File "/usr/lib/python3.4/cmd.py", line 211, in onecmd
    return self.default(line)
  File "/home/bollmann/repositories/xonsh/xonsh/shell.py", line 105, in default
    code = self.push(line)
  File "/home/bollmann/repositories/xonsh/xonsh/shell.py", line 126, in push
    locs=self.ctx)
  File "/home/bollmann/repositories/xonsh/xonsh/execer.py", line 87, in compile
    code = compile(tree, self.filename, mode)
ValueError: Dict doesn't have the same number of keys as values
bollmann@underberg ~/repositories/xonsh (master) $ 
```

Other nested structures seem to work fine. Putting the list in a variable first and then constructing the dict also works fine.

```
bollmann@underberg ~/repositories/xonsh master >>> x = ['a', 'b']
bollmann@underberg ~/repositories/xonsh master >>> {'foo': x}
{'foo': ['a', 'b']}
bollmann@underberg ~/repositories/xonsh master >>> {'a': {'b': 'c', 'd': 'e'}}
{'a': {'b': 'c', 'd': 'e'}}
bollmann@underberg ~/repositories/xonsh master >>> [{'a': 'b', 'c': 'd'}]
[{'a': 'b', 'c': 'd'}]
bollmann@underberg ~/repositories/xonsh master >>> [{'a': 'b'}, {'c': 'd'}]
[{'a': 'b'}, {'c': 'd'}]
bollmann@underberg ~/repositories/xonsh master >>> {'a': {'b': 'c', 'd': 'e'}}
[...]

PR #122, opened 2015-03-21
Nested structures

wrywerytwreywery:
Attemped fix for issues #61 and #85, where xonsh interprets nested structures incorrectly (and an alternative to PR #100).

The issue was being unable to distinguish between an AST list or tuple node that was organic, and one that was created using `ensure_has_elts`.  Hopefully this catches most/all issues.  I'll continue to test, and would appreciate any feedback.

### DIFF
--- a/xonsh/parser.py
+++ b/xonsh/parser.py
@@ -1589,11 +1589,18 @@ class Parser(object):
                          | test_or_star_expr comma_test_or_star_expr_list comma_opt
         """
         p1, p2 = p[1], p[2]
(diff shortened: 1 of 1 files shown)


wrywerytwreywery:
Woah this seems to have failed quite a lot.  I'll take another look...

scopatz:
Yeah, it looks like a lot of ast.Tuple nodes are remaining in....

scopatz:
My only other point is that all of the cases that people have reported should become tests.

Issue #87, opened 2015-03-18
More variables for prompt


dimaqq:
At the very least, most basic bash parity:

* host (an actual hostname as opposed to domain name or fqdn)
* basename of cwd (i.e. last element of cwd)

mbollmann:
I can try to put this in. Do you have any further suggestions of variables that would be nice to have?

With an increasing number of variables, I wonder if we should do `if '{varname}' in prompt: ...` before computing them, instead of just computing them all and passing them to `.format()`.

scopatz:
> I can try to put this in. Do you have any further suggestions of variables that would be nice to have?

Awesome, thanks!  I don't have any suggestions, myself...

> With an increasing number of variables, I wonder if we should do if '{varname}' in prompt: ... before computing them, instead of just computing them all and passing them to .format().

Probably.  Though the colors should probably always be added, rather than checking all the colors.

mbollmann:
@dimaqq, can you clarify what you mean by "an actual hostname as opposed to domain name or fqdn"? The current `{hostname}` variable should return the "actual hostname", shouldn't it?

scopatz:
@mbollmann - that is what I thought!

mbollmann:
Most shells seem to have different variables for "full hostname" and "hostname up to the first dot", but I can't get `socket` to return anything containing a dot on any of my machines (though that might be configuration-dependant!), and I'm not sure if this is what @dimaqq meant.

But I have some other stuff that I added already. Refactoring the formatting so that the variables are only computed when they're used is trickier than I thought, but I hope to have sth. ready soon.

scopatz:
sweet, thanks @mbollmann

abadger:
@mbollmann Regarding making it so that variables are only computed when used, I have something that might provide a framework for that:

https://github.com/abadger/xonsh/commit/22e61899cfd2d13dfa5531f5ef6a739807d751eb

I was waiting to PR that until I heard if it was a desirable direction on the mailing list but I can put it up for review sooner if that helps.

It currently separates "variable only needs to be determined once" and "variable needs to be determined each time the prompt is formatted".  In my implementation the variables that are determined each time the prompt is formatted are only computed if the variable is accessed.  So those do what you want and are not computed if the prompt does not make use of them.

It would be nice to implement the same thing for the static strings although it may not be a huge win: after all, they are only computed once, not every time the prompt is displayed.

dimaqq:
@mbollmann about names

[...]

PR #125, opened 2015-03-22
Prompt formatter refactor

abadger:
This refactor of the prompt formatting makes a few improvements in response to #87 

* Add base_cwd and short_host prompt variables
* Only compute values for variables which are used

Additionally, it:
* Reorganizes the prompt var code so that adding new variables to be substituted can be pretty easily recognized.
* Speeds up the curr_branch var code when the git shell prompt scripts are not found.
* Fixes the curr_branch code to return empty string instead of None in a few corner cases.
* Should lay the foundation for making formatters user customizable should we choose to enable that.


scopatz:
Hi @abadger - thanks for putting this in!  I think it has the right idea, but there are a number of style issues that I am going to comment on in-line.  Once those are fixed up, this should be good to go.  The major non-style request that I have would be for you to add some tests for these classes (maybe not for the functionality you didn't implement, like `curr_branch()`).

xonsh/environ.py
```diff
@@ -7,62 +7,188 @@
...
 from xonsh.tools import TERM_COLORS
 
-def current_branch(cwd=None):
-    """Gets the branch for a current working directory. Returns None
-    if the cwd is not a repository.  This currently only works for git, 
-    bust should be extended in the future.
-    """
-    branch = None
-    cwd = os.getcwd() if cwd is None else cwd
+class PromptFormatter(dict):
```
scopatz: please inherit from `collections.MutableMapping` and not `dict`.

xonsh/environ.py
```diff
@@ -7,62 +7,188 @@
...
+    '''
```
scopatz: PEP8 - use triple double quotes rather than triple single quotes.  This applies elsewhere in the file too.

xonsh/environ.py
```diff
@@ -7,62 +7,188 @@
...
+    Prompt Formatter.  However, you most likely want to inherit from
+    DefaultPromptFormatter if you are writing a custom formatter.
+
+    Custom Prompt Formatters will need to implement their own __init__()
+    method that adds entries to several dicts depending on the needs of the
+    variables they add.  The Formatter may also implement methods to implement
+    those variables (although very simple variables can be written with
+    pre-existing functions.)
+
+    Example:
```
scopatz: Add a `::` at the end of the line and then indent the rest of the block.  We use numpydoc for our docstring format and I don't think that this would render properly as is.

xonsh/environ.py
```diff
@@ -7,62 +7,188 @@
...
+        self['time24'] = functools.partial(time.strftime, '%H:%M:%S')
+        self._run_every.add('time24')
+
+    def login_time(self):
+        who_out = subprocess.check_output(['who', '-m'])
+        return ' '.join(who_out.decode().split()[2:4])
+    '''
+
+    def __init__(self, *args, **kwargs):
+
```
scopatz: PEP8 extra whitespace line.

xonsh/environ.py
```diff
@@ -7,62 +7,188 @@
...
+        super(PromptFormatter, self).__init__(*args, **kwargs)
+        self._run_every = set()
```
scopatz: `run_every` is part of the API and should not be private. This will be easier to implement when you go to a MutableMapping.

xonsh/environ.py
```diff
@@ -7,62 +7,188 @@
...
+        formatters = dict(
+                base_cwd=partial(self.cwd, False),
+                cwd=self.cwd,
+                curr_branch=self.curr_branch,
+                hostname=socket.gethostname,
+                short_host=lambda: socket.gethostname().split('.', 1)[0],
+                time=datetime.now,
+                user=self.user,
+        )
+        self.update(formatters)
```
scopatz: no reason not to stick the `formatters` directly into the update call.

xonsh/environ.py
```diff
@@ -7,62 +7,188 @@
...
+        self._run_every.update(frozenset(('base_cwd', 'cwd', 'curr_branch', 'time')))
```
scopatz: `'user'` should be in this set.

xonsh/environ.py
```diff
@@ -7,62 +7,188 @@
...
+            value = self[key]
+        except KeyError:
+            return default
+        return value
+
+
+class DefaultPromptFormatter(PromptFormatter):
+    def __init__(self, *args, **kwargs):
+        super(DefaultPromptFormatter, self).__init__(*args, **kwargs)
+        self._env = builtins.__xonsh_env__
```
scopatz: env should also be public

xonsh/environ.py
```diff
@@ -7,62 +7,188 @@
...
+                                            universal_newlines=True)
+            except subprocess.CalledProcessError:
+                continue
+            else:
+                # Trust the git-prompt script determined if we're in a git
+                # repo and if so, which branch we're in
+                branch = branch.strip()
+                break
+
+        else: # http://bit.ly/for_else
```
scopatz: PEP8 - two spaces before comment.

xonsh/environ.py
```diff
@@ -80,17 +206,11 @@ def format_prompt(template=default_prompt):
...
-    branch = '' if branch is None else ' ' + branch
-    p = template.format(
-            user=env.get('USER', '<user>'),
-            hostname=socket.gethostname(),
-            cwd=cwd.replace(env['HOME'], '~'),
-            curr_branch=branch,
-            **TERM_COLORS
-            )
+    global prompt_formatter
+    if not prompt_formatter:
```
scopatz: should be `if prompt_formatter is not None:`

scopatz:
I hope that these help!

xonsh/environ.py
abadger: <nod>  Yeah -- about that, I was trying to build the docs so that I could check how all of this affects the docstrings.  In addition to cleaning up the example formatting there also needs to be documentation updates to add the new prompt variables.  And I'm not sure if xonsh has configured sphinx to pull documentation from the class's docstring or the ```__init___``` docstring (When I've used sphinx in the past it's been an either or).

Unfortunately, I couldn't figure out how to build the docs... I got to the point where I had to modify the Makefile to use sphinx-build-3.4 so that it could use autodoc on the source files but then it appeared that cloud_sptheme wasn't ported to python3 yet. (pip3.4 install --user cloud_sptheme fails).  Any clues on how you do it so that I can be sure the documentation is all building correctly?

xonsh/environ.py
abadger: Why is that?  collections.MutableMapping is an abstract base class.  So if we inherit from that we have to implement all of the Mapping methods ourselves.  If we inherit from dict then we only have to implement the methods that we care about.   The one caveat that I know of being that if you implement ```__getitem__()``` you also have to implement ```get()``` because of how dict's get() method is implemented.  (And if we were overriding ```__setitem__()```, similar issues for the other value setting methods).

Inheriting from dict directly is less code to write and maintain, less objects to create (no need for a separate private dict to actually store the items), and somewhat faster methods (since the methods we don't override are implemented in C).  If you know of some problem that ```MutableMapping()`` solves, I can change it, it just doesn't seem to make sense since really we only need a custom ```__getitem__()```  for this to work with python's str.format() method.

xonsh/environ.py
scopatz: Thanks for these questions @abadger 

> And I'm not sure if xonsh has configured sphinx to pull documentation from the class's docstring or the __init___ docstring (When I've used sphinx in the past it's been an either or).

I believe that I have it set up to use both.  It appends the `__init__` docstring to the class docstring.

> I got to the point where I had to modify the Makefile to use sphinx-build-3.4 so that it could use autodoc on the source.  

You should not need to do that.

> but then it appeared that cloud_sptheme wasn't ported to python3 yet. 

It does wok on Python 3. I have used it many times. See https://pythonhosted.org/cloud_sptheme/install.html#requirements

> (pip3.4 install --user cloud_sptheme fails). Any clues on how you do it so that I can be sure the documentation is all building correctly?

I don't know what is going on here, and you might report it to them.  I used conda to get the package, not pip.  If pip is failing though, you can always try from source.  See the previous link.  Also, it is in a lot of package managers, depending on what OS you are on.

xonsh/environ.py
abadger: ```_run-every`` needs to be marked with a single underscore.  In C++ terms, it can be considered a protected attribute.  It's part of the API if you are subclassing but it is not part of the API if you are calling from code outside of the inheritance tree.

Long winded explanation:  You need to have attributes of your class that are public, private, and protected in an OOP design.  python, gives us conventions to achieve privacy but the interpreter (mostly) doesn't enforce it.  So the conventions are guidance to other programmers on what their code should not touch in order to be a good citizen.  The convention for public attributes is that they should not start with an underscore.  Any programmer working on code unrelated to the class knows they should not touch attributes that start with an underscore but may touch any attributes that do not start with an underscore.  A programmer working on a class knows that they can touch the attributes of the class that begin with a leading underscore as long as they do not let too much of the implementation of that attribute leak into the outside world.  Now the question remains, what do python programmers need to do to implement protected attributes?  If you omit the leading underscore, programmers working on unrelated code will think that the attribute is public and also use it.  This is compounded by the fact that these programmers know the least about the code and thus are least able to make an informed decision about whether there's an alternative to that attribute, what the consequences are to using that attribute, and how to code a fix should that attribute change behaviour in the future.  This is definitely not a good situation.  On the other hand, if you mark the attribute with a leading underscore, programmers who are subclassing your code are the ones affected.  They will see the leading underscore and have to figure out whether there's an alternative to using the attribute, what the consequences are to using it, and have to fix their code should that attribute change.  However, since they are writing code that relies intimately on your class anyway understanding the answers to those questions is of benefit to them anyway and something that they are better prepared to do since they are writing code that extends yours rather than simply making use of the services it provides.

xonsh/environ.py
abadger: Yeah, I was wondering about that.  When would users change during the lifetime of a shell?  I tried to think about sudo, su, and ssh.  It seemed like su and ssh both create a new shell.  sudo will run a program as another user but by the time you get your shell prompt back you'll be the same user you started as.  It also seemed to make a sort of conceptual sense in that if you became a new user and were able to keep interacting with your original shell the shell would need to have all sorts of authorization handling to prevent the second account from doing things as the first account.

If you can think of something that I've overlooked, I'm happy to add user in there, though.

xonsh/environ.py
scopatz: > Why is that?

Copying dict subclasses can have strange behaviour (http://code.activestate.com/lists/python-dev/135141/, https://hg.python.org/cpython/file/22a0c925a7c2/Objects/dictobject.c#l2003). MutableMapping is more general.  And in my experience, using ABCs has always been much safer than subclassing the built-in types.  I'd rather us use MutableMapping than ever have to go down this non-obvious debugging rabbit hole again.

Another good reason is consistency.  Other places in the xonsh code base use MutableMapping, and not dict.

> Inheriting from dict directly is less code to write and maintain, less objects to create (no need for a separate private dict to actually store the items), and somewhat faster methods (since the methods we don't override are implemented in C).

These things are not hard and only around 20 lines of code to implement.  Plus it gives us flexibility and safety.

I hope that this helps explain it.

abadger: okee dokee.

xonsh/environ.py
scopatz: I understand.  What I was trying to say was that I don't think that there is a good reason for this to be private/protected.  Rather, this could conceivably be something that a user wants to change and therefore should be public.  

This is less true of env, where there are many other options for users to manipulate the the value.

xonsh/environ.py
scopatz: Oh yeah! You are right.  Sorry, my bad!

>

xonsh/environ.py
abadger: sphinx has to run the code through the python compile step so we cannot use a version of sphinx-build running on python2.  Must run it via python3.  I noticed that your scripts/xonsh also used a /usr/bin/python shebang  instead of /usr/bin/python3 so maybe you have your system setup so that python maps to python3 (perhaps just when running certain applications)?  That would explain why you don't need that but other people do (I had to change the shebang line to run from a checkout as well.  The shebang is properly rewritten when installed if python3 seutp.py install is used).

Looks like python3 support in cloud_sptheme on pypi is broken.  They have it fixed in source control but they haven't made a release since 2013 so they haven't pushed the fix out where it's easily available.  I've locally my copy so that it runs.

autodoc isn't doing a good job of organizing the docs for the new classes.  https://toshio.fedorapeople.org/environ.html  what do you think of this plan:

* Pull the formatters out of environ into their own file.
* Use autodoc of specific methods of the formatter classes instead of autodoc of the whole file.  * Turn off inherited-members for the new formatter documentation.
* Either create a new rst page in api for prompt formatters or add it as a separate section o nteh environ page (sine format_prompt() is there).

Question: Do you still want documentation of the prompt variables in format_prompt?  I think it may be more logical to document that in DefaultPromptFormatter (probably in the class docstring since not all format variables are implemented via a method.)  Will need to point the link from the tutorial section to DefaultPromptFormatter as well.

scopatz: >
> sphinx has to run the code through the python compile step so we cannot
> use a version of sphinx-build running on python2. Must run it via python3.
> I noticed that your scripts/xonsh also used a /usr/bin/python shebang
> instead of /usr/bin/python3 so maybe you have your system setup so that
> python maps to python3 (perhaps just when running certain applications)?
> That would explain why you don't need that but other people do (I had to
> change the shebang line to run from a checkout as well. The shebang is
> properly rewritten when installed if python3 seutp.py install is used).
>

I probably do.  Adding the trailing 3 is probably reasonable for the
moment.


> Looks like python3 support in cloud_sptheme on pypi is broken. They have
> it fixed in source control but they haven't made a release since 2013 so
> they haven't pushed the fix out where it's easily available. I've locally
> my copy so that it runs.
>
Ahh, you should probably complain to them :)


> autodoc isn't doing a good job of organizing the docs for the new classes.
>
What do you mean by this?  In terms of the ordering? I think that in sphinx
you can switch the ordering around.

> https://toshio.fedorapeople.org/environ.html what do you think of this
> plan:
>
>    - Pull the formatters out of environ into their own file.
>
> I am not a huge fan of that, yet. I don't think that there is enough code
to justify that.

>
>    - Use autodoc of specific methods of the formatter classes instead of
>    autodoc of the whole file.
>
> kinda against this because it makes it much more annoying to maintain the
docs.

>
>    - * Turn off inherited-members for the new formatter documentation.
>
> I'd be happy with that.

>
>    - Either create a new rst page in api for prompt formatters or add it
>    as a separate section o nteh environ page (sine format_prompt() is there).
>
> I'd be happier with a separate section.

>
>
> Question: Do you still want documentation of the prompt variables in
> format_prompt? I think it may be more logical to document that in
> DefaultPromptFormatter (probably in the class docstring since not all
> format variables are implemented via a method.)
>
 Yes, I think it makes sense to move most of the docstring contents.

> Will need to point the link from the tutorial section to
> DefaultPromptFormatter as well.
>
Sure.

xonsh/environ.py
abadger: _run_every is not in the purview of the user at all.  It's an implementation detail of ```__getitem__```.

abadger:
Okay, I think everything from the original comments (including tests) has been addressed except for the choice of public/private which I disagree with.  The original comment's documentation issue has been addressed but not my followon comments.

abadger:
Okay, and documentation updated as well.  I can really only work on things on weekends, let me know if there's anything else that needs doing.

scopatz:
Ok thanks @abadger! I'll look at this more tomorrow!

abadger:
Alright... git didn't seem to want to act nice when I tried to rebase on top of the lastest master but we can always do a manual diff of the latest state back onto the HEAD when we're ready... I was going to squash the commits once the review was done anyway.

With the latest commit it's now possible to customize the prompt formatting from the .xonshrc.  It looks like this:

``` python
import subprocess
from xonsh.environ import DefaultPromptFormatter

class MyPromptFormatter(DefaultPromptFormatter):
    def __init__(self):
        DefaultPromptFormatter.__init__(self)
        self['tty'] = self.get_tty()

    def get_tty(self):
        tty = subprocess.check_output('tty').decode().strip()
        segments = tty.split('/')
        return '/'.join(segments[-2:])

$PROMPT_FORMATTER=MyPromptFormatter
$PROMPT='{CYAN}[{tty}@{short_host}{NO_COLOR} {cwd}{RED}{curr_branch}{NO_COLOR}{CYAN}]{NO_COLOR}$ '
```
The sample .xonshrc defines a custom PromptFormatter class that subclasses and extends the DefaultPromptFormater that xonsh provides.  The new class registers a single function, get_tty, that will output the tty that the shell is running on.  (Note: this does not need to be a method of the promptformatter class.  In this particular case, it's just a handy way of organizing the code.  In other cases, you can store common resources on the class it you like.)  As we expect the tty to stay the same for the life of the shell, this is all we have to do in the class to enable the tty.

In order to use the new formatter class we need to tell xonsh that we want to use it instead of the default.  We do that by setting the $PROMPT_FORMATTER environment variable to the new class.  xonsh will take care of instantiatng the class itself.  Then we define our $PROMPT as normal, utilizing the "tty" prompt var that we defined in the formatter.  When we invoke xonsh with this .xonshrc we get the following prompt:

```
[pts/10@roan /srv/git/wishlist/xonsh prompt-formatter]$
```

tests/test_prompt.py
```diff
@@ -0,0 +1,178 @@
...
+import nose
+from nose.tools import eq_, ok_
+
+from .tools import mock_xonsh_env
+from xonsh.tools import TERM_COLORS
+
+from xonsh.environ import PromptFormatter, DefaultPromptFormatter
+
+
+class TestPromptFormatter(TestCase):
```
scopatz: Please use nose for your test interface rather than unittest directly. This is so it is consistent with the other tests.

xonsh/environ.py
```diff
@@ -81,17 +113,117 @@ def format_prompt(template=default_prompt):
...
+        self.update(dict(*args, **kwargs))
+
+    def user(self):
+        """Return the current user's username"""
+        return self._env.get('USER', '<user>')
+
+    def cwd(self, full_path=True):
+        """Return the current working directory
+
+        :kwarg full_path: If True (the default), return the full path of the
```
scopatz: Please use numpydoc format for your docstrings.

tests/test_prompt.py
```diff
@@ -0,0 +1,178 @@
...
+from __future__ import unicode_literals, print_function
+
+import os
+import sys
+import time
+import socket
+from datetime import datetime, timedelta
+from unittest import TestCase, skip, mock
+
+sys.path.insert(0, os.path.abspath('..'))  # FIXME
```
scopatz: Is this a holdover from another file?

scopatz:
Hi @abadger - this is looking pretty good.  Just a couple of medium refactors w.r.t. testing and I'd be happy to merge.  Thanks!

wrywerytwreywery:
I've only been halfway following this, but I do think this is nice functionality.  I have to agree with @scopatz that, given the way `_run_every` is used (almost anyone implementing a subclass has to touch it), I would make it public.  Alternatively, you could have some sort of different way to register a function to be called each time (and hide the fact that this modifies `_run_every` from the user).  Maybe something like `formatter.register_callevery(name, func)`?

Really, my bias is also against OOP; it seems to me like this is complicating the problem more than it needs to be.  Why not just keep around a dictionary holding the formatting keys, where each one maps to a string (if computed at init time) or to a function of no arguments (if it needs to be computed each time the prompt is displayed)?  Then `format_prompt` should just format the string, calling the functions in the dictionary where necessary.  And adding a new variable for the formatting string becomes much easier than creating a subclass of the formatter; rather, it would just be making a string or a function, and adding that into the dictionary.

tests/test_prompt.py
abadger: Yeah, that was copied from one of your existing test files.  near as I could tell it was so that you could import xonsh from the correct directory if you didn't happen to run it from the toplevel checkout.

tests/test_prompt.py
abadger: nose doesn't have a complete suite of test tools.  It only provides ok_() and eq_().  For other tests you either end up without good test messages, haing to rite out the test messages manually, or reinventing the wheel by reimplementing the unittest tests yourself.

scopatz: Hi @abadger - In my experience nose has all of the tests that unittests has and more.  What do you think is missing? I am curious....

abadger: Ah... dir(nose.tools) tells me you're right.  They're there, just not documented well on the nose.tools page.  I can do that then.

scopatz: OK thanks :)!

abadger:
And done.

abadger:
We could put _run_every inside of a method.  But that should be a private method.  The thing is that while there are many things inside of the class that may need to register with run_every there's nothing outside of the class that needs to.  If the method of adding new formatting functions was to call a xonsh function, to do the registration then that function would be public but that's not the design here.

Your idea of keeping a dictionary is messier than encapsulating this inside of a class and doesn't give as much room for sharing resources or extending the functionality.  You still need to have all of the same data and methods but instead of the class keeping track of it (and being overridable if the user is really creative), you have all of that implementation detail tucked into a xonsh function that is much harder to get access to and extend.  I wouldn't have thought about the need for functions that run only once and only if they're in a prompt string unless @mbollmann had mentioned it in #87.  So what would his options be if the implementation hadn't included that capability?  With this class, he would be able to subclass it and then override __getitem__ to add that capability in.  With only the dict to register the values, he'd have had to submit a patch to modify the format_prompt function. 

Without a class, you still have to keep around the data necessary to implement these strategies of mapping the data to the prompt vars.  So you'd still need the dict you mention that maps prompt variables to strings or methods.  You'd still have the logic that decides whether to display the string, call the function and cache the value, or call the function every time a prompt is displayed.  And you still have to store any additional values to implement that (the _run_every set/dict/registration_function).  Without a class to hold these pieces together, you get a bunch of pieces instead of a consistent object to model what's going on.  And if you want to extend one of those pieces (suppose you like the curr_branch functionality but want to color it depending on whether the checkout has modifications or not) rather than write from scratch then your job is much harder.

the last reason for a class that I can think of is that some extensions are better implemented with shared state.  self._env is an example of that sort of shared state.  Data that is more expensive to calculate are even better examples.  A class provides a ready made container for calculating the data once and sharing the results.

wrywerytwreywery:
I am not so against storing things in a class (though we'll have to agree to disagree on how "messier" is defined, it seems).  I do think, though, that requiring users to make a subclass of `DefaultPromptFormatter` in order to add a new variable to the prompt could be the kind of thing that turns someone off.  Having that as an option could be a good thing for some of the reasons you mentioned, but I would personally like to see an easier way (e.g., through defining a string or function and hooking it in to the formatter in some way) in addition.

abadger:
Okay, implemented a function to do that and test cases.  Some things I ran across while doing that:

* format_prompt() takes an argument to set the template but the prompt formatter is being extracted by the environment.  It feels like these should match.  We could pull the prompt formatter out in shell.Shell, instantiate it and save it as an instance variable there.
* I noticed that shell.Shell.settitle() and prompt () actually have the ability to use $PROMPT and $TITLE as callables to set the format string.  This feels like it's a subset of the prompt formatter functionality (the main thing it lacks is the ability to override individual pieces of the creation of a prompt string.  If you set PROMPT to a callable you'd end up having to reimplement everything yourself.)  SHould we remove that?

scopatz:
> format_prompt() takes an argument to set the template but the prompt formatter is being extracted by the environment. It feels like these should match. We could pull the prompt formatter out in shell.Shell, instantiate it and save it as an instance variable there. Then pass that as a parameter to format_prompt()

I'd say it is fine to just look up the prompt formatter in the environment and kill the global version. If a formatter does not exist in the environment, it should set a default one.

> I noticed that shell.Shell.settitle() and prompt () actually have the ability to use $PROMPT and $TITLE as callables to set the format string. This feels like it's a subset of the prompt formatter functionality (the main thing it lacks is the ability to override individual pieces of the creation of a prompt string. If you set PROMPT to a callable you'd end up having to reimplement everything yourself.) SHould we remove that?

Nope, users should have the ability to pass in a simple callable.

abadger:
I don't think you understood the first question (although , I can certainly implement what you're asking for :-)  format_prompt() currently takes the prompt template from a parameter and the formatter (really, the formatter class -- but that's something that we can change) from the environment.  It seems like it should take them both from the same source (either parameters or the environment).  So I'm really asking which of those should be the source conceptually.

Regarding the global variable: the PROMPTFORMATTER environment variable currently takes a class but I can change that to an instance and then we don't need the global variable.  The one drawback of it holding the instance directly is that there's no possibility of moving the added prompt variables over to the new PromptFormatter inside of xonsh's code should the user choose to change it while the shell  is running.  The user would have to manage those themselves.  But that wasn't something I personally was interested in anyway, so I'm happy to change the code to remove the global.

For the callable question, why?  One of python's guiding principles is that there's one obvious way to do it.  Having PROMPT be a callable means that we have three ways of doing the same thing of which one operates in a different manner to achieve the same goals:

``` python
# Callable way of defining the prompt string:
def do_something():
    return 'calculated prompt'

$PROMPT=do_something

# PromptFormatter way:
from environ import add_prompt_vars
def do_something():
    return 'calculated prompt'

add_prompt_vars('prompt', do_something)
PROMPT='{prompt}'
```

scopatz:
> so I'm happy to change the code to remove the global.

Sounds good.

> For the callable question, why? One of python's guiding principles is that there's one obvious way to do it. 

Because a callable that returns a string is the obvious way.  I know for a fact that other users really like this feature (as do I).  I think it would be a mistake to remove it.  Another one of Python's guiding principles (vis-a-vis duck typing) is that interfaces should be able to handle anything reasonable that the user throws at them.  That is why I am happy to support more advanced formatting here, but it shouldn't come at the cost of more basic interfaces, like allowing strings or callables.

abadger:
So, I still don't know whether you want the format_prompt() function to utilize env vars exclusively or parameters exclusively :-)

For callables in prompt, what I'm driving at is that we shouldn't have two methods of doing things.  If you want PROMPT to be able to take a callable we should remove the PROMPTFORMATTER environment variable and instead use PROMPT to invoke the promptformatter.  Maybe something like: if PROMPT is a string, then create a default prompt formatter and run that over the PROMPT.  If prompt is a callable, call that to return the prompt.  if a PromptFormatter is passed in, the PromptFormatter is a callable that knows how to return a prompt string.

For end users things then look like this:

```python

# Simple string with default var substitutions
$PROMPT='prompt string for {user} $'

# Use a simple callable 
def i_control_everything():
    return 'calculated prompt'
$PROMPT=i_control_everything

# Override the defaults
formatter = DefaultPromptFormatter('prompt_string')
formatter.add_prompt_vars('var_name', calculate_prompt_var, run_every=False)
$PROMPT=formatter
```

This makes for a more consistent interface where there's only one way to do things again, by modifying the value of $PROMPT.

There's one major difference from currently -- currently there's two steps, first define a prompt template.  then pass through a formatter.  This combines those two steps into a single one.  Not a problem, just a difference.

There's one problem with this but it's going to exist as long as we're allowing a simple callable so I can't see any way to avoid that given that you want those.  It will be extremely easy for users to produce code that makes for slow prompt string creation because a simple callable doesn't give the user any guidance on caching calculated values.

scopatz:
Exclusively as env vars, please.  about to hop on a plane, I'll look at the rest later.

abadger:
Here's an implementation of the prompt reorganization and customization that uses the $PROMPT callable like I talked about in the earlier comment: https://github.com/scopatz/xonsh/pull/162

I like it better than this pull request but let me know what you think.

scopatz:
@abadger would you mind if I closed this PR?  There has been no movement on this in a very long time and we have moved on with another formatter. Sorry.

scopatz:
Closing due to inactivity.  Please feel free to reopen.

PR #126, opened 2015-03-22
Allow arguments for xonsh shell scripts

wrywerytwreywery:
This change allows scripts to be run from xonsh with additional arguments (e.g., `xonsh test.sh uno zwei trois`), and for these arguments to be referenced within the script in bash style (with `$0`, `$1`, ...) by loading them into the script's environment.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -91,11 +91,11 @@ class Env(MutableMapping):
 
     def __getitem__(self, key):
         m = re.match(r'ARG(\d+)',key)
-        if m is not None and 'ARGS' in self._d.keys():
+        if (m is not None) and (key not in self._d) and ('ARGS' in self._d):
             args = self._d['ARGS']
             ix = int(m.group(1))
             if ix >= len(args):
-                e = "Not enough arguments given to access argument {0}."
+                e = "Not enough arguments given to access ARG{0}."
                 raise IndexError(e.format(ix))
             return self._d['ARGS'][ix]
         return self._d[key]
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -90,6 +90,14 @@ class Env(MutableMapping):
     #
 
     def __getitem__(self, key):
+        m = re.match(r'ARG(\d+)',key)
+        if m is not None and 'ARGS' in self._d.keys():
+            args = self._d['ARGS']
+            ix = int(m.group(1))
+            if ix >= len(args):
+                e = "Not enough arguments given to access argument {0}."
+                raise IndexError(e.format(ix))
+            return self._d['ARGS'][ix]
         return self._d[key]
 
     def __setitem__(self, key, val):
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -40,6 +40,8 @@ class Env(MutableMapping):
     use in a subprocess.
     """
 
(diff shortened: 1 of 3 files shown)


scopatz:
It looks like this now has some merge conflicts with master @wrywerytwreywery.  Would you mind updating them, please? Thanks!

xonsh/main.py
```diff
@@ -5,19 +5,27 @@
 import subprocess
 from argparse import ArgumentParser, Namespace
 
-from xonsh.shell import Shell
+from xonsh.shell import Shell, builtins
```
scopatz: hmmm you should probably just `import builtins` above.

scopatz:
First off, this is a great and needed feature.  Thanks for putting this in @wrywerytwreywery! But...

> for these arguments to be referenced within the script in bash style (with $0, $1, ...) 

I really dislike that bash does this.  It isn't intuitive to me at all.  I don't think that xonsh should support this syntax.  I'd much rather have there be an $ARGS environment variable whose value is a list of strings. 

My understanding is that $INT comes from a time when Bash did not have arrays or really any data structures other than strings.  However, in xonsh, our Env allows for non-string types, so we can do the right thing.

wrywerytwreywery:
I will merge in the changes from master.

I think `$ARGS` makes sense for python mode, but not for subprocess mode.  For example, the following would not work (without modifying the parser): `echo $ARGS[1]`.

I halfway agree about `$<INT>`.  My feeling, though, is that xonsh should support this syntax, both because it is familiar to bash users, and because it is much easier to write, e.g., `$0` than `$ARGS[0]` (and for the reason mentioned above).  For now, I am updating this PR to do both `$<INT>` and `$ARGS`.

If you decide against this, feel free to modify, or let me know and I can do so.

scopatz:
Hi @wrywerytwreywery, you bring up a very good point about `$ARGS[0]`, though I have pretty much decided against the `$<INT>` syntax.  

I think perhaps a cleaner solution - that doesn't involve changing the parser at would be to make the Env lookup more dynamic. For example, if `$ARGS` exists `$ARG<N>` does not exist in the env, then `$ARG<N>` would return the nth index of `$ARGS`.  This could be implemented from within the Env, class which already does special things with the names it sees. Also this would work equally well in Python mode and subprocess mode.

Something like that is what I would prefer to see.  How does that sound?

wrywerytwreywery:
That works for me.  I will go ahead and implement.

scopatz:
Thanks!

wrywerytwreywery:
This should be ready for another look now.

xonsh/built_ins.py
```diff
@@ -90,6 +90,14 @@ def undo_replace_env(self):
     #
 
     def __getitem__(self, key):
+        m = re.match(r'ARG(\d+)',key)
```
scopatz: probably best to compile this regex at class or module scope.

scopatz: PEP8 space after comma

scopatz:
Other than the simple compile issue, this looks great.  Also, something about this behaviour should be added to the docs.

scopatz:
Thanks @wrywerytwreywery!

wrywerytwreywery:
Okay, I think this is ready for another look.  I'll try to find the time to add something to the docs, as well.

scopatz:
Thanks!

PR #129, opened 2015-03-22
Documentation for running scripts

wrywerytwreywery:
It isn't spectacular, but I tried to add some relevant text to the tutorial about running scripts with xonsh (as opposed to using the REPL).  I also fixed a few typos and grammatical issues, and removed trailing whitespace.


scopatz:
Thanks @wrywerytwreywery!

Issue #110, opened 2015-03-20
CLI-commands cannot deal with preceeding whitespace


SmartViking:
```
uglemat@uglemat ~/Public/xonsh/xonsh master $      ls
............................................. 
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/xonsh/shell.py", line 137, in push
    locs=self.ctx)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/execer.py", line 84, in compile
    tree = self.parse(input, ctx, mode=mode)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/execer.py", line 62, in parse
    tree = self._parse_ctx_free(input, mode=mode)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/execer.py", line 118, in _parse_ctx_free
    mode=mode, debug_level=self.debug_level)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/parser.py", line 271, in parse
    debug=debug_level)
  File "/usr/lib/python3/dist-packages/ply/yacc.py", line 269, in parse
    return self.parseopt_notrack(input,lexer,debug,tracking,tokenfunc)
  File "/usr/lib/python3/dist-packages/ply/yacc.py", line 1051, in parseopt_notrack
    tok = self.errorfunc(errtoken)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/parser.py", line 2119, in p_error
    column=self.lexer.token_col(p)))
  File "/usr/local/lib/python3.4/dist-packages/xonsh/parser.py", line 341, in _parse_error
    raise err
  File "<string>", line None
SyntaxError: <xonsh-code>:1:3: ('code:      ',)
uglemat@uglemat ~/Public/xonsh/xonsh master $ 
```

Maybe this can be fixed by putting a str.lstrip in the right place? Should `$[  ls]` work, in that case?

scopatz:
I actually don't think that they should be able to deal with this. I'll explain more after work.

scopatz:
Hi @SmartViking! Sorry for not getting this yesterday.  Basically, here is the situation. In Python mode, we have a situation that is:

* whitespace insensitive inside of an expression
* has blocks defined by initial indentation level

Meanwhile in subprocess mode we have a situation that is:

* whitespace sensitive inside of expressions
* does not allow blocks

and in Bash:

* whitespace sensitive inside of expressions
* has blocks with delimiters (if-fi, for-end, etc)
[...]

PR #130, opened 2015-03-22
`lstrip` the first line of shell input

SmartViking:
Addresses #110


scopatz:
Thanks a ton @SmartViking!

Issue #135, opened 2015-03-23
Commands that start with '/' or './' result in exception


pdavydov108:
Hello, I updated xonsh to a new verson today (via `pip`), I think it's 0.1.3 now. If I type `ls` it works just fine, but `/bin/ls` returns an exception:
````
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/xonsh/shell.py", line 123, in default
    self.execer.exec(code, mode='single', glbs=self.ctx)  # no locals
  File "/usr/local/lib/python3.4/dist-packages/xonsh/execer.py", line 110, in exec
    return exec(code, glbs, locs)
  File "<xonsh-code>", line 1, in <module>
  File "/usr/local/lib/python3.4/dist-packages/xonsh/built_ins.py", line 476, in subproc_uncaptured
    return run_subproc(cmds, captured=False)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/built_ins.py", line 412, in run_subproc
    aliased_cmd = get_script_subproc_command(cmd[0], cmd[1:])
  File "/usr/local/lib/python3.4/dist-packages/xonsh/built_ins.py", line 359, in get_script_subproc_command
    first_line = f.readline().strip()
  File "/usr/lib/python3.4/codecs.py", line 313, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc5 in position 24: invalid continuation byte
````
Same error occurs if I execute `./a.out`, but with different message:
````
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 24: invalid start byte
````
It worked fine in previous version.

Update: it looks like it only happens for compiled executables. Python scripts work normally for example.

scopatz:
Thanks for reporting @pdavydov108

scopatz:
Hi @pdavydov108 - I believe that the current master (bac6bfb) should fix this issue.

scopatz:
Would you please verify and let us know! Thanks.

pdavydov108:
Hi @scopatz ! I've installed the current master (1e76fd076575c3c3d9706e337ef9aa6d41fabe3a) and it works just fine, thanks.

PR #136, opened 2015-03-23
xonsh should be able to run binaries outside the path, as well as scripts

wrywerytwreywery:
This should fix the issue mentioned in #135, where xonsh has trouble handling binary files outside the path.  There might be something more elegant, but the simplest fix is to fall back to running the command exactly as specified if _anything_ goes wrong when trying to figure out how to run the specified file.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -414,6 +414,9 @@ def run_subproc(cmds, captured=True):
                 e = 'xonsh: subprocess mode: permission denied: {0}'
                 print(e.format(cmd[0]))
                 return
+            # fall back to trying to run directly with subproc
+            except:
+                aliased_cmd = cmd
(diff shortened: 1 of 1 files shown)


scopatz:
Hi @wrywerytwreywery - thanks for putting this in.  However, I think the right fix is to read in bytes in `get_script_subproc_command()`, rather than a string.  Also, it is probably a good idea to only read in the first two bytes and check if they are `#!` before reading in the full line.

scopatz:
I can take a stab at this later if you want me to.

wrywerytwreywery:
yes, by all means.

scopatz:
I think that bac6bfb should close this issue.

wrywerytwreywery:
Seems to work for me.  Closing.

PR #140, opened 2015-03-23
Command line switch --no-rc

mbollmann:
Just a minor addition: a command-line switch to prevent parsing the .xonshrc file.

Useful when a recent change broke something in your .xonshrc file, and probably in other scenarios. Bash has it as well.

### DIFF
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -14,6 +14,11 @@ parser.add_argument('-c',
                     dest='command',
                     required=False,
                     default=None)
+parser.add_argument('--no-rc',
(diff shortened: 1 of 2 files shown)


xonsh/main.py
```diff
@@ -33,7 +38,7 @@ def main(argv=None):
 
     args = parser.parse_args()
 
-    shell = Shell()
+    shell = Shell() if not args.norc else Shell(ctx={})
```
scopatz: Double negative logic. Nothing wrong with this per say....

scopatz:
Thanks for the fix @mbollmann!

Issue #132, opened 2015-03-23
there should be an environmental variable $XONSH_VERSION


zaxebo1:
when we are in zsh, there is a $ZSH_VERSION environmental variable
when we are in bash, there is a $BASH_VERSION environmental variable. Rather bash has many many other bash specific envionmental variables, which are probably never needed. :)  

I think there should be a $XONSH_VERSION env variable to immediately check the version. and accordingly act in my script

scopatz:
A good idea.

scopatz:
This should be implemented now

PR #143, opened 2015-03-24
Version

wrywerytwreywery:
As suggested in issue #132, this changeset introduces the xonsh version number as an environment variable `$XONSH_VERSION`.  I also tried to minimize the number of places where the version is hard-coded in the source (though I am not aware of a way to remove it from `meta.yaml`.


scopatz:
Thanks for this, @wrywerytwreywery!  I think it is more common to use `__version__` than `VERSION`, though (see PEP8 https://www.python.org/dev/peps/pep-0008/#version-bookkeeping).  Also, for the `meta.yaml`, you are right.  I think the only half-way decent way to do that is to write it out in setup.py.  

I am happy to merge this in as is, but if you want to change the name, let me know!

wrywerytwreywery:
Woops; wasn't aware of that.  I'll go ahead and change the name quickly.

scopatz:
Thanks!

wrywerytwreywery:
Alright; I went ahead and changed the name.  But it felt to weird to me to use the name `__version__` in the files making use of it, so I imported it as `XONSH_VERSION` in all those places.  Is that reasonable?

scopatz:
Yep totally

scopatz:
Hopefully travis will finish before my plane takes off ;)

scopatz:
Thanks!

GodFatherandSon:
Sent from Yahoo Mail on Android

Issue #133, opened 2015-03-23
In .xonshrc, classes don't seem to bind their methods


abadger:
Current xonsh checkout: f44013b31756ba5491f2a7e1dffb7ad64513b28e
python: 3.4.1
OS: fedora 21

Here's a minimal .xonshrc that defines a class and then attempts to use it:

``` python
class Test:
    def __init__(self):
        self.msg('hello world')
    def msg(self, m):
        print(m)

$PF=Test()
```

Instead of storing an instance of the class in $PF, it throws the following error:

``` 
Traceback (most recent call last):
  File "scripts/xonsh", line 3, in <module>
    main()
  File "/srv/git/wishlist/xonsh/xonsh/main.py", line 36, in main
    shell = Shell()
  File "/srv/git/wishlist/xonsh/xonsh/shell.py", line 94, in __init__
    execer=self.execer)
  File "/srv/git/wishlist/xonsh/xonsh/environ.py", line 168, in xonshrc_context
    execer.exec(rc, glbs={}, locs=env)
  File "/srv/git/wishlist/xonsh/xonsh/execer.py", line 110, in exec
    return exec(code, glbs, locs)
  File "/home/badger/.xonshrc", line 9, in <module>
  File "/home/badger/.xonshrc", line 260, in __init__
AttributeError: 'Test' object has no attribute 'msg'
Exception ignored in: <bound method Shell.__del__ of <xonsh.shell.Shell object at 0x7fb350b844e0>>
Traceback (most recent call last):
  File "/srv/git/wishlist/xonsh/xonsh/shell.py", line 102, in __del__
    teardown_readline()
  File "/srv/git/wishlist/xonsh/xonsh/shell.py", line 65, in teardown_readline
    import readline
  File "<frozen importlib._bootstrap>", line 2237, in _find_and_load
  File "<frozen importlib._bootstrap>", line 2222, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 2164, in _find_spec
  File "<frozen importlib._bootstrap>", line 1940, in find_spec
  File "<frozen importlib._bootstrap>", line 1908, in _get_spec
TypeError: 'NoneType' object is not iterable
```

Not sure if we are supposed to be able to store python code inside of environment variables however the problem is deeper than that; modifying the xonsh source code to try to access the class via the context also leads to the same error.

scopatz:
I am pretty sure this is the same as #131

wrywerytwreywery:
[...]

Issue #134, opened 2015-03-23
Lexer freeze


eldamir:
I have this little beauty in my .xonshrc

```python
def foo():
    ascii_art = [
        "(╯°□°）╯︵ ┻━┻",
        "¯\_(ツ)_/¯",
        "┻━┻︵ \(°□°)/ ︵ ┻━┻",
    ]

    import random
    i = random.randint(0,len(ascii_art)) - 1
    print("    Get to work!")
    print(ascii_art[i])

foo()
```

When I run my terminal, it hangs forever. The prompt is never shown. When I interrupt with ctrl+c, I get the following stack trace

```
Traceback (most recent call last):
  File "/usr/local/bin/xonsh", line 3, in <module>
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/xonsh/shell.py", line 123, in default
    self.execer.exec(code, mode='single', glbs=self.ctx)  # no locals
    main()
  File "/usr/local/lib/python3.4/dist-packages/xonsh/main.py", line 36, in main
  File "/usr/local/lib/python3.4/dist-packages/xonsh/execer.py", line 110, in exec
    return exec(code, glbs, locs)
  File "<xonsh-code>", line 1, in <module>
  File "/usr/local/lib/python3.4/dist-packages/xonsh/built_ins.py", line 476, in subproc_uncaptured
    return run_subproc(cmds, captured=False)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/built_ins.py", line 456, in run_subproc
    prev_proc.communicate()[0]
    shell = Shell()
  File "/usr/local/lib/python3.4/dist-packages/xonsh/shell.py", line 94, in __init__
  File "/usr/lib/python3.4/subprocess.py", line 941, in communicate
    self.wait()
  File "/usr/lib/python3.4/subprocess.py", line 1536, in wait
    (pid, sts) = self._try_wait(0)
  File "/usr/lib/python3.4/subprocess.py", line 1493, in _try_wait
    (pid, sts) = _eintr_retry_call(os.waitpid, self.pid, wait_flags)
  File "/usr/lib/python3.4/subprocess.py", line 487, in _eintr_retry_call
    return func(*args)
    execer=self.execer)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/environ.py", line 168, in xonshrc_context
KeyboardInterrupt
    execer.exec(rc, glbs={}, locs=env)
[...]

PR #144, opened 2015-03-25
Lexer Overhaul

wrywerytwreywery:
This patch includes a rewrite of the lexer to make use of Python's `tokenize` module, and an update of the relevant test cases to reflect these changes.

I also changed the parser slightly (I think the only notable difference from a syntax perspective is that it is now illegal to have a space in between a `$` and a name in an environment lookup).

As far as I can tell, this fixes issues #133 and #134 (I have also added tests to the test suite for those issues), passes all tests, and runs the complicated examples from the tutorial as expected.


scopatz:
Hi @wrywerytwreywery - thanks for putting this in.

> I also changed the parser slightly (I think the only notable difference from a syntax perspective is that it is now illegal to have a space in between a $ and a name in an environment lookup).

That should have been illegal anyway :)

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+    state['last'] = token
+    if token.type == tokenize.DEDENT:
+        state['indents'].pop()
+        yield _new_token('DEDENT', ' '*state['indents'][-1], token.start)
+    elif token.type == tokenize.INDENT:
+        state['indents'].append(level)
+        yield _new_token('INDENT', token.string, token.start)
+
+    try:
+        n = next(stream)
```
scopatz: I think that doing `n = next(stream, None)` is much better here.

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+
+def preprocess_tokens(tokstream):
+    tokstream = clear_NL(tokstream)
+    state = {'indents': [0], 'pymode': [True], 'last': None}
+    for token in tokstream:
+        for i in handle_token(state, token, tokstream):
+            yield i
+
+
+def clear_NL(tokstream):
```
scopatz: PEP8 no uppercase letters in function names.

scopatz:
Hi @wrywerytwreywery - from my first glance, I have to say that I am quite impressed.  I was very sceptical that this would work, or an on-par solution even.  I still think that catching the ERRORTOKENs is a little weird, but practicality beats purity. And thiss does offer a nice and hopefully permanent solution to some of our tokenizing issues.  (Frankly, I'll be happy if I never have to look at another string literal regex ever again ~_~.)

I am going to go over it with a fine tooth comb before I go to sleep right now. I saw some style issues that should be addressed before merging.

scopatz:
(Also, sorry I was travelling today and didn't get to this sooner.)

scopatz:
As a general comment, docstrings are needed on all functions and classes.

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+def handle_indent(state, token, stream):
+    level = len(token.string)
+    state['last'] = token
+    if token.type == tokenize.DEDENT:
+        state['indents'].pop()
+        yield _new_token('DEDENT', ' '*state['indents'][-1], token.start)
+    elif token.type == tokenize.INDENT:
+        state['indents'].append(level)
+        yield _new_token('INDENT', token.string, token.start)
+
```
scopatz: Shouldn't the rest of this function body be in an else-block to prevent the following code from being accessed accidentally after an INDENT or DEDENT.  In fact, if you did this, I think you might be able to get rid of the recursion and simply have:

```python
else:
    yield token
```

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+    try:
+        n = next(stream)
+    except:
+        n = None
+    if n is not None:
+        if n.type != tokenize.ENDMARKER:
+            for i in handle_token(state, n, stream):
```
scopatz: these two if-statement should be in a single if-statement on one line:  ie:

```python
if n is not None and n.type != tokenize.ENDMARKER:
```

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+                yield i
+
+
+def handle_dollar(state, token, stream):
+    try:
+        n = next(stream)
+    except:
```
scopatz: This should be `n = next(stream, None)` followed by an `if n is None:`

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+        yield _new_token('DOLLAR_LBRACE', '${', token.start)
+    else:
+        e = 'expected NAME, (, [, or {{ after $, but got {0}'
+        m = e.format(n)
+        yield _new_token("ERRORTOKEN", m, token.start)
+
+
+def handle_at(state, token, stream):
+    try:
+        n = next(stream)
```
scopatz: Same comment as in `handle_dollar()`

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+    else:
+        yield _new_token('AT', '@', token.start)
+        state['last'] = token
+        for i in handle_token(state, n, stream):
+            yield i
+
+
+def handle_question(state, token, stream):
+    try:
+        n = next(stream)
```
scopatz: `n = next(stream, None)`

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+    else:
+        yield _new_token('QUESTION', '?', token.start)
+        state['last'] = token
+        for i in handle_token(state, n, stream):
+            yield i
+
+
+def handle_backtick(state, token, stream):
+    try:
+        n = next(stream)
```
scopatz: `n = next(stream, None)`

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+        state['last'] = n
+    else:
+        e = "Could not find matching backtick for regex on line {0}"
+        m = e.format(token.start[0])
+        yield _new_token("ERRORTOKEN", m, token.start)
+
+
+def handle_newline(state, token, stream):
+    try:
+        n = next(stream)
```
scopatz: `n = next(stream, None)`

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+    if not state['pymode'][-1]:
+        state['last'] = token
+        yield _new_token('WS', ' ', token.start)
+    else:
+        yield from []
+
+special_handlers = {
+    tokenize.ENCODING: lambda s, t, st: [],
+    tokenize.COMMENT: lambda s, t, st: [],
+    tokenize.ENDMARKER: lambda s, t, st: [],
```
scopatz: Same lambda is redefined three times.  Probably best to define it once (ie `_empty = lambda s, t, st: []`) and then refer to it multiple times here.

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+    if n.type == tokenize.OP and n.string == '(' and \
+            n.start == token.end:
+        state['pymode'].append(True)
+        yield _new_token('AT_LPAREN', '@(', token.start)
+        state['last'] = n
+    else:
+        yield _new_token('AT', '@', token.start)
+        state['last'] = token
+        for i in handle_token(state, n, stream):
+            yield i
```
scopatz: Unless we are trying explicitly for Python 2 compatibility here, you can use `yield from` here instead of the loop:

```python
yield from handle_token(state, n, stream)
```

This comment applies to all other similar instances in the file.

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+
+    found_match = False
+    sofar = '`'
+    while n is not None:
+        sofar += n.string
+        if n.type == tokenize.ERRORTOKEN and n.string == '`':
+            found_match = True
+            break
+        try:
+            n = next(stream)
```
scopatz: `n = next(stream, None)`

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+        except:
+            n = None
+    if found_match:
+        yield _new_token('REGEXPATH', sofar, token.start)
+        state['last'] = n
```
scopatz: Is this line guaranteed to execute? Does it need to come after the yield?

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+    try:
+        n = next(stream)
+    except:
+        n = None
+
+    yield _new_token('NEWLINE', '\n', token.start)
+    state['last'] = token
+
+    if n is not None:
+        if n.type != tokenize.ENDMARKER:
```
scopatz: ifs on same line.

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+        yield _new_token(token_map[(typ, st)], st, token.start)
+    elif typ in token_map:
+        state['last'] = token
+        yield _new_token(token_map[typ], st, token.start)
+    elif (typ, st) in special_handlers:
+        for i in special_handlers[(typ, st)](state, token, stream):
+            yield i
+    elif typ in special_handlers:
+        for i in special_handlers[typ](state, token, stream):
+            yield i
```
scopatz: yield from

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+            yield _new_token('WS', ' '*(cur[1]-old[1]), old)
+    if (typ, st) in token_map:
+        state['last'] = token
+        yield _new_token(token_map[(typ, st)], st, token.start)
+    elif typ in token_map:
+        state['last'] = token
+        yield _new_token(token_map[typ], st, token.start)
+    elif (typ, st) in special_handlers:
+        for i in special_handlers[(typ, st)](state, token, stream):
+            yield i
```
scopatz: yield from

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+
+
+def handle_token(state, token, stream):
+    typ = token.type
+    st = token.string
+    if not state['pymode'][-1]:
+        if state['last'] is not None and state['last'].end != token.start:
+            cur = token.start
+            old = state['last'].end
+            yield _new_token('WS', ' '*(cur[1]-old[1]), old)
```
scopatz: This seems like it would replace tabs with a single space

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+        m = "Unexpected token: {0}".format(token)
+        yield _new_token("ERRORTOKEN", m, token.start)
+
+
+def preprocess_tokens(tokstream):
+    tokstream = clear_NL(tokstream)
+    state = {'indents': [0], 'pymode': [True], 'last': None}
+    for token in tokstream:
+        for i in handle_token(state, token, tokstream):
+            yield i
```
scopatz: yield from

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+
+
+def clear_NL(tokstream):
+    for i in tokstream:
+        if i.type != tokenize.NL:
```
scopatz: Same line.

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+
+
+def single_error(exc):
+    m = "{} (line {}, column {})".format(exc.msg, exc.lineno, exc.offset)
+    yield _new_token("ERRORTOKEN", m, (0, 0))
+
+
+def tok(s):
+    try:
+        return iter(tokenize.tokenize(BytesIO(s.encode('utf-8')).readline))
```
scopatz: I don't think that you need to call `iter()` here do you, because `tokenize()` is already a generator.  I could be wrong....

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+    except Exception as e:
+        return iter(single_error(e))
```
scopatz: What is the thinking behind this? I don't quite understand why it is needed.

xonsh/lexer.py
```diff
@@ -1,13 +1,307 @@
...
+def _new_token(type, value, pos):
+    o = LexToken()
+    o.type = type
+    o.value = value
+    o.lineno, o.lexpos = pos
+    return o
+
 
 def anyof(*regexes):
     return '(' + '|'.join(regexes) + ')'
```
scopatz: Is this function used anymore?

xonsh/lexer.py
```diff
@@ -33,53 +327,26 @@ def __init__(self, errfunc=lambda e, l, c: print(e)):
...
-    @lineno.setter
-    def lineno(self, value):
-        if self.lexer is not None:
-            self.lexer.lineno = value
+        pass
 
     def input(self, s):
         """Calls the lexer on the string s."""
-        self.lexer.input(s)
+        s = re.sub(r'#.*', '', s)
```
scopatz: Can't we just use the functions above to ignore comments?  Also, regexes like this should be precompiled and not compiled on each function call.

xonsh/lexer.py
```diff
@@ -33,53 +327,26 @@ def __init__(self, errfunc=lambda e, l, c: print(e)):
...
-        self.errfunc(msg, location[0], location[1])
-        self.lexer.skip(1)
-
-    def _make_tok_location(self, token):
-        return (token.lineno, self.token_col(token))
+        try:
+            self.last = next(self.token_stream)
+            return self.last
+        except StopIteration:
+            return None
```
scopatz: probably best to rewrite this as 

```python
self.last = next(self.token_stream)
return self.last
```

xonsh/parser.py
```diff
@@ -333,15 +331,18 @@ def expr(self, p):
...
-        return self.lexer.token_col(t)
+        return t.lexpos
 
     @property
     def lineno(self):
-        return self.lexer.lineno
+        try:
+            return self.lexer.last.lineno
+        except:
+            return 0
```
scopatz: I don't know what the exception is here, but it is probably best to use `if hasattr(last, ..)` or `if last is None:` or similar.

xonsh/parser.py
```diff
@@ -333,15 +331,18 @@ def expr(self, p):
...
 
     @lineno.setter
     def lineno(self, value):
-        self.lexer.lineno = value
+        pass
```
scopatz: Probably best to kill this whole method, since it longer works as advertised.

xonsh/parser.py
```diff
@@ -1516,8 +1520,8 @@ def p_atom(self, p):
             bt = '`'
             if isinstance(p1, (ast.Num, ast.Str, ast.Bytes)):
                 pass
-            elif (p1 is True) or (p1 is False) or (p1 is None):
-                p1 = ast.NameConstant(value=p1, lineno=self.lineno,
+            elif (p1 == 'True') or (p1 == 'False') or (p1 == 'None'):
+                p1 = ast.NameConstant(value=eval(p1), lineno=self.lineno,
```
scopatz: It would be nice to split this out into three separate elif-statements and not eval a string that we know to be a singleton.  It is more lines of code, but likely faster overall.

scopatz:
Ok @wrywerytwreywery!  Thanks again, and thanks for dealing with the many in-line comments. Once it is fixed up, I'd love to merge this in.

xonsh/lexer.py
wrywerytwreywery: I think this is actually a holdover from when I was handling `ENDMARKER` differently, and so I can just kill the following code.

scopatz: OK :)

xonsh/lexer.py
wrywerytwreywery: The problem is that some things (conveniently things that are hard to catch) cause `tokenize.tokenize` to throw an exception (like dedent not matching an indent level, or similar).  In those cases, so that we can report the error easily, I just wanted to yield a single `ERRORTOKEN` so that we would get a consistent syntax error out of the parser.

xonsh/lexer.py
wrywerytwreywery: I had a lot of trouble with just using tokens to get rid of comments.  In particular, something like `echo "hello # comment` broke pretty easily, because when wrapped to try in subprocess mode, it was interpreted as `$[echo "hello" # comment ]`, which killed the last bracket and made it fail.

This was a quick hack that seemed to get to the right answer; if there's a more elegant way, I'm all for it!

xonsh/lexer.py
wrywerytwreywery: I did it this way because I always wanted `self.last` to be a valid token once we started going.  Looking at the parser again, though, it seems like that might not be a concern.

xonsh/lexer.py
wrywerytwreywery: I didn't know `iter` took that second argument!  Definitely cleans things up.

xonsh/lexer.py
wrywerytwreywery: My initial thought is that tab characters shouldn't exist in this context (outside of Python).  Is that not valid?  If not, I can look and see how tab characters are trated in this context, and I can try to find a workaround.

xonsh/lexer.py
wrywerytwreywery: Ah, but that doesn't work there because `tokenize` is a generator...I'll rework this.

wrywerytwreywery: I tried a few things, but haven't found a solution yet.  Will revisit under the light of day.

wrywerytwreywery:
Alright, I think this is ready for another look.  I think I took care of all of the issues you brought up (except for the lone inline comment remaining above, the issue with adding `WS` tokens in subprocess mode), as well as a couple others I noticed this morning.

wrywerytwreywery:
Actually, I just noticed an issue with error handling and a couple of other opportunities to clean up.  I'll ping you again when it's ready.

scopatz:
OK! Sounds good.

wrywerytwreywery:
Okay, I think it's _actually_ ready for a look now, once travis finishes.

xonsh/lexer.py
scopatz: I think that this could cause some weird issues when debugging, thought the column count should still be correct.  I don't really want for this issue to block the whole PR, though.  Maybe you could make an issue for it so that we remember to investigate later?

xonsh/lexer.py
```diff
@@ -1,27 +1,333 @@
...
+            return
+        except tokenize.TokenError as e:
+            # this is recoverable in single-line mode (from the shell)
+            # (e.g., EOF while scanning string literal)
+            yield _new_token('ERRORTOKEN', e.args[0], (0, 0))
+            return
+        except IndentationError as e:
+            # this is never recoverable
+            yield _new_token('ERRORTOKEN', e, (0, 0))
+            return
```
scopatz: I didn't think that you could have both returns and yields.  In any event, it is probably better to use `break` here.

scopatz:
This looks great @wrywerytwreywery.  Only the last little `return` to `break` issue.

wrywerytwreywery:
Okay, I replaced those `return`s with `break`s, and I included a change that should fix the whitespace issue from above.

scopatz:
Ok, everything here looks great.  However, I think that this breaks the current behaviour of `xonsh.tools.subproc_toks()`.  For example, the following used to work but now gives an error:

```console
$ ls; ls -l
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/shell.py", line 123, in default
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/execer.py", line 111, in exec
    return exec(code, glbs, locs)
NameError: name 'ls' is not defined
```

scopatz:
Maybe the lexer no longer issues SEMI tokens?

wrywerytwreywery:
This is weird.  The following still works for me, for example: ` echo "hey" ; ls`
But not `ls ; ls`

I will dig in and see what is happening.

wrywerytwreywery:
Okay, I think I've fixed this one.  The issue was with the way semicolons were being handled inside of `subproc_toks`, as you suggested.

But I found another issue with the parser's grammar, so I will try to fix that before I push again.

wrywerytwreywery:
Well, I wasn't able to figure out the other issue yet (I'll submit separate issue since I do think it is separate), but I figured I should push what I have.

scopatz:
What is the other issue, out of curiosity?

wrywerytwreywery:
Just posted it, #147.

scopatz:
Thanks for the heroic effort here!

scopatz:
Did you want to ping the other issues that this addresses or should I?

wrywerytwreywery:
I can go ahead and do that.

scopatz:
Also @wrywerytwreywery - are you on the xonsh mailing list?  Or is there a way I can contact you aside from github issues? :)

scopatz:
Also, if you don't care to give it, that is fine too!

wrywerytwreywery:
Sure thing.  I've just joined the xonsh mailing list.  My e-mail is hartz@mit.edu

PR #145, opened 2015-03-25
Use landscape.io

javValverde:
Hi,

First of all thak you for xonsh, it is awesome!!!
I have added a badge for landscape.io (static code analysis). I think it is a good way of preventing bad code/potential bugs.
You only need to create an account on https://landscape.io and it should work. If you find the rules too strict, feel free to modify ".landscape.yaml"


scopatz:
Thanks @javValverde! This service seems to be super helpful.

PR #150, opened 2015-03-26
xonsh should error on finding a line break inside a regex path (backtick)

wrywerytwreywery:
This change causes xonsh to throw an error when a regex expression started with a backtick doesn't have a matching backtick on the same line (if a newline is found when scanning forward to find the matching backtick).


scopatz:
Thanks @wrywerytwreywery!

PR #151, opened 2015-03-26
Ignore Python keywords in subprocess mode

wrywerytwreywery:
Currently, xonsh gets confused by Python keywords showing up in subprocess mode (since they are tokenized differently from other names).  For example:
`git config --global ...`

This change causes all names (regardless of whether they are Python keywords) to be tokenized as `NAME` tokens when in subprocess mode, so that they can be interpreted properly.

### DIFF
--- a/xonsh/lexer.py
+++ b/xonsh/lexer.py
@@ -265,13 +265,17 @@ def handle_token(state, token, stream):
     """
     typ = token.type
     st = token.string
-    if not state['pymode'][-1]:
+    pymode = state['pymode'][-1]
+    if not pymode:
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks @wrywerytwreywery - Should this have cooresponding changes in the parser that look for python keyword tokens?  I think in `p_subproc_atom()` or similar...

wrywerytwreywery:
By the time they are getting parsed in subprocess mode, they should all be `NAME` tokens, so I don't think there's any need to change the parser.

scopatz:
Yup you are right.  Thanks again @wrywerytwreywery!

Issue #141, opened 2015-03-24
Job Control


wrywerytwreywery:
I think this won't be easy given how xonsh is put together, but it would be great to support some job control commands (ctrl-z, fg, bg, etc) in addition to having the ability to start a subprocess in the background.

scopatz:
I should note that ending with `&` already exists.  So I think bg is a simple alias along the lines of:

```python
from xonsh.built_ins import subproc_uncaptured
def bg(args, stdin=None):
    subproc_uncaptured(args, '&')
    return None
```

We might need to build a jobs set, but that shouldn't be too hard.  Also Ctrl-z would be a feature of the Shell, I feel like, which would then need to send the SIGTSTP to the current process.

SmartViking:
In bash/zsh each job is given a number and you can select which to `fg` with `fg %n`, or kill like `kill %n`, I think xonsh should support something similar to that.

wrywerytwreywery:
bg and fg are a little more complicated, as they can act on running jobs to push them to the background or bring them to the foreground, so we would definitely need a jobs list.  And we would also need the shell to have a notion of which process, if any, is in the foreground.

There is also the question of the nice way to get xonsh to grab the `SIGSTP` and send it to the current foreground process; presumably something in `signal` can help (and we can use `os.kill` to send the command).

I think this is doable.  Unless someone else wants to take a stab, I can put it on my queue.

wrywerytwreywery:
The behavior that I really want, by the way, is at least the following:
* start a process (not backgrounded)
* stop with ctrl-z
* run another command
* bring original process back to foreground

and...
* start a process (not backgrounded)
* realize i wanted it backgrounded
* stop with ctrl-z
* send to background to let it run but keep the shell open for other commands

SmartViking:
One job may contain several processes, like this:

```
$ find . | cat > file &
[1] 14954 14955
$
[1]  + done       find . | cat > file
```

Both `find` and `cat` are considered (here by zsh) as one job.

SmartViking:
You can intercept CTRL-Z with this:

```python
def handler(num, frame):
    pass

signal.signal(signal.SIGTSTP, handler)
```
[...]

PR #152, opened 2015-03-27
Job control

wrywerytwreywery:
This PR is an attempt to implement some amount of job control in xonsh, as mentioned in #141.

This changeset implements the following:
* suspending foreground process with ctrl-z
* resuming suspended process in foreground or background
* listing of all current jobs
* killing foreground process with ctrl-c

It also:
* documents the `jobs`, `fg`, and `bg` commands
* modifies `exit` to kill (maybe too forcefully?) all subprocesses, suspended or otherwise, before exiting
* fixes a bug where ctrl-c would kill all child processes, including those running in the background

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -4,6 +4,7 @@ not to be confused with the special Python builtins module.
 import os
 import re
 import sys
+import time
 import shlex
 import signal
 import locale
@@ -19,6 +20,7 @@ from xonsh.tools import string_types, redirect_stdout, redirect_stderr, suggest_
 from xonsh.inspectors import Inspector
 from xonsh.environ import default_env
 from xonsh.aliases import DEFAULT_ALIASES, bash_aliases
+from xonsh.jobs import _get_pid, _print_one_job, _get_job_number, _wait_for_active_job, ProcProxy 
 
 ENV = None
 BUILTINS_LOADED = False
@@ -292,7 +294,6 @@ def iglobpath(s):
 
 WRITER_MODES = {'>': 'w', '>>': 'a'}
 
-ProcProxy = namedtuple('ProcProxy', ['stdout', 'stderr'])
 
 def _run_callable_subproc(alias, args, captured=True, prev_proc=None, 
                           stdout=None):
@@ -347,6 +348,7 @@ def is_script(fname):
 
 RE_SHEBANG = re.compile(r'#![ \t]*(.+?)$')
 
+
 def get_script_subproc_command(fname, args):
     """
     Given the name of a script outside the path, returns a list representing
@@ -376,6 +378,7 @@ def get_script_subproc_command(fname, args):
             interp = ['xonsh']
 
     return interp + [fname] + args
+
 
 def run_subproc(cmds, captured=True):
     """Runs a subprocess, in its many forms. This takes a list of 'commands,'
@@ -438,7 +441,7 @@ def run_subproc(cmds, captured=True):
             stdin = prev_proc.stdout
         try:
             proc = Popen(aliased_cmd, universal_newlines=uninew, env=ENV.detype(),
-                         stdin=stdin, stdout=stdout)
+                         stdin=stdin, stdout=stdout, preexec_fn=os.setsid)
         except PermissionError:
             cmd = aliased_cmd[0]
             print('xonsh: subprocess mode: permission denied: {0}'.format(cmd))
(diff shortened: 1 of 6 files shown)


xonsh/built_ins.py
```diff
@@ -438,7 +443,7 @@ def run_subproc(cmds, captured=True):
             stdin = prev_proc.stdout
         try:
             proc = Popen(aliased_cmd, universal_newlines=uninew, env=ENV.detype(),
-                         stdin=stdin, stdout=stdout)
+                         stdin=stdin, stdout=stdout, preexec_fn=os.setsid)
```
scopatz: `preexec_fn` and `os.setsid()` seem to be posix only.  Could we make it so that this does not get passed in unless `os.name == 'posix'`  Maybe a dict of kwargs is the best.

wrywerytwreywery:
I can certainly make that change.  I am actually discovering that `os.setsid` has a couple of other issues that I didn't notice yesterday, as well, so I will need to find another fix.

wrywerytwreywery:
Still not quite ready for prime-time, but it is getting closer.

xonsh/built_ins.py
```diff
@@ -436,9 +441,15 @@ def run_subproc(cmds, captured=True):
             stdin = PIPE
         else:
             stdin = prev_proc.stdout
+        def subproc_pre():
```
scopatz: This doesn't seem like it needs to be defined inside of `run_subproc()` though I could be missing something subtle :)

xonsh/jobs.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def _reactivate_job():
+    if len(builtins.__xonsh_all_jobs__) == 0:
+        return
+    builtins.__xonsh_active_job__ = max(builtins.__xonsh_all_jobs__.items(),
+                                        key=lambda x: x[1]['started'])[0]
+
+
+_block_when_giving = [signal.SIGTTOU, signal.SIGTTIN,
```
scopatz: Probably could be tuple.

xonsh/jobs.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+def print_one_job(num):
+    """
+    Print a line describing job number ``num``.
+    """
+    job = builtins.__xonsh_all_jobs__[num]
+    act = '*' if num == builtins.__xonsh_active_job__ else ' '
+    status = job['status']
+    cmd = [' '.join(i) if isinstance(i, list) else i for i in job['cmds']]
```
scopatz: `isinstance(i, Sequence)` with `from collections import Sequence` is more general

scopatz: Though this doesn't block :)

xonsh/jobs.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def handle_sigchld(num, frame):
+        obj.done = True
+        job['bg'] = True
+        job['status'] = 'stopped'
+
+    _give_terminal_to(pgrp)  # give the terminal over to the fg process
+    signal.signal(signal.SIGCHLD, handle_sigchld)
+    while obj.poll() is None and not obj.done:
+        time.sleep(0.01)
```
scopatz: This could probably be `0.05` without anyone noticing.

xonsh/jobs.py
wrywerytwreywery: Is there any chance that `i` would ever be anything other than a list?  I can change it to be safe.

xonsh/jobs.py
scopatz: That is, human response time is 0.1 at a minimum.

wrywerytwreywery:
Sorry to be making changes as you're going through this, but...9b896d1585f0164def86e1502d276a1e9c132322 fixed a huge bug with the way forks were treated in subprocesses, so if you're going to be testing more, it might be good to pull that in :)

xonsh/jobs.py
scopatz: No, not really.  That is why this doesn't block.  `i` is definitely an implementation detail of xonsh that we have complete control over.  If someone wants to go into `builtins.__xonsh_all_jobs__` and change things - and then xonsh breaks - that is their own fault and problem :).

scopatz:
> Sorry to be making changes as you're going through this, but... 

No worries, I mostly going through for style.

xonsh/jobs.py
wrywerytwreywery: Yep.  I think we could probably get away with as much as .2 seconds without it bothering anybody (at least, it didn't bother me, even when I was explicitly looking for it).  I'm not sure how resource-hungry `Popen.poll` is, so I'm not sure how much we need to cut that down.  I'll change to 0.1s for now.

scopatz:
This looks awesome by the way!

wrywerytwreywery:
Alright; phew!  I think with that last push, it's ready for a real look.  I've tested quite a bit, and will continue to do so, but it could probably use another set of eyes on it if you find the time.  Any feedback would be appreciated.

scopatz:
@wrywerytwreywery - This looks great!  Though I did find an issue piping an alias through a command.  Specifically, the grep below hangs forever without a Crtl-c, but works on master (7c1a6a28ee5961268eb575593c9905eeea6fff92):

```console
scopatz@athenaie ~ $ aliases["hello"] = lambda arg, stdin=None: "Hello World!\n"
scopatz@athenaie ~ $ hello | grep o
Hello World!
```

scopatz:
Note that alias to alias piping seems to work.

wrywerytwreywery:
Okay, I think this most recent push should squash that one.  The problem was not closing the later process's `stdin` after writing the output of the `ProcProxy` to it.

scopatz:
Works like a charm.  I am happy to merge this.  Do you have any lingering reservations or should I just go ahead? :)

wrywerytwreywery:
Great!  I think I've squashed all the bugs I know of, so it should be ready.

scopatz:
Thanks!

PR #155, opened 2015-03-28
Error Reporting

wrywerytwreywery:
xonsh sometimes has a tendency to fall into infinite loops of wrapping statements in `$[]` upon encountering a syntax error.  This patch is an attempt to fix this issue.

An example of code that fails in this way is:

```
def foo(x):
    print x
```

### DIFF
--- a/xonsh/execer.py
+++ b/xonsh/execer.py
@@ -142,6 +142,8 @@ class Execer(object):
                     last_error_line = last_error_col = -1
                     input = '\n'.join(lines)
                     continue
+                if line.startswith('$'):
+                    raise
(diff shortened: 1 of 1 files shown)


wrywerytwreywery:
This changeset now also includes the following changes:

* attempts to avoid the infinite `$[]` loop mentioned in the original PR
* provides more detailed syntax error messages in some cases (mismatched parens, brackets, and braces)
* reports the original syntax error encountered if subprocess mode fails, rather than the error message generated after surrounding the statement with `$[]`

scopatz:
Thanks @wrywerytwreywery!

PR #157, opened 2015-03-28
Fix lexer's handling of #

wrywerytwreywery:
You were right, @scopatz, the handling of `#` in the lexer was totally bogus (a trivial case where it breaks down is an octothorpe inside of a string).  After some testing, I think the main code in `lexer.py` should handle comments appropriately, so just removing that silly regex replacement seems to fix this.

Amusingly, I ran into this issue when trying to run a `git commit -m` with an issue number in the commit message.

### DIFF
--- a/xonsh/lexer.py
+++ b/xonsh/lexer.py
@@ -325,9 +325,6 @@ def _new_token(type, value, pos):
     return o
 
 
-COMMENT_REGEX = re.compile(r'#.*')
-
-
 class Lexer(object):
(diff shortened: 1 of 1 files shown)


scopatz:
> Amusingly, I ran into this issue when trying to run a git commit -m with an issue number in the commit message.

Haha that is super funny.

scopatz:
Seems to have the right behaviour, thanks!

PR #158, opened 2015-03-28
Change semantics for running executables from outside the path

wrywerytwreywery:
This patch changes the semantics of handling executables from outside the path to more closely match what bash does with these kinds of files.

* if the file is a binary (defined as in bash), run it
* otherwise:
  * if the first line is a shebang, use that result as the interpreter
  * otherwise, use xonsh as the interpreter

This also includes several PEP8 formatting changes, but the substantive changes are 352-368, and 378-387, and 442 in the updated file.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -27,12 +27,13 @@ from xonsh.jobs import ProcProxy
 ENV = None
 BUILTINS_LOADED = False
 INSPECTOR = Inspector()
-LOCALE_CATS = {'LC_CTYPE': locale.LC_CTYPE, 'LC_COLLATE': locale.LC_COLLATE, 
-    'LC_TIME': locale.LC_TIME, 'LC_MONETARY': locale.LC_MONETARY, 
-    'LC_MESSAGES': locale.LC_MESSAGES, 'LC_NUMERIC': locale.LC_NUMERIC,}
+LOCALE_CAT = {'LC_CTYPE': locale.LC_CTYPE, 'LC_MESSAGES': locale.LC_MESSAGES,
+              'LC_COLLATE': locale.LC_COLLATE, 'LC_NUMERIC': locale.LC_NUMERIC,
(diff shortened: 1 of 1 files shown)


xonsh/built_ins.py
```diff
@@ -342,15 +345,27 @@ def _run_callable_subproc(alias, args, captured=True, prev_proc=None,
...
+
+def _is_binary(fname, limit=80):
+    with open(fname, 'rb') as f:
+        for i in range(limit):
+            char = f.read(1)
+            if char == b'\0':
+                return True
+            if char == b'\n':
+                return False
+            if char == b'':
```
scopatz: These might be better as elif-statements.

xonsh/built_ins.py
```diff
@@ -342,15 +345,27 @@ def _run_callable_subproc(alias, args, captured=True, prev_proc=None,
...
 
 
+def _is_runnable_name(fname):
+    return os.path.isfile(fname) and fname != os.path.basename(fname)
+
+
+def _is_binary(fname, limit=80):
+    with open(fname, 'rb') as f:
+        for i in range(limit):
+            char = f.read(1)
```
scopatz: What happens when the file is less than the limit number of bytes?  Is that what the empty string case is for?

scopatz:
Looks good overall, just a couple of questions on `_is_binary()`

xonsh/built_ins.py
wrywerytwreywery: Yep. `f.read` will return an empty string if it has already hit EOF.  I checked it on a short file to make sure, and it seems to work as expected.

xonsh/built_ins.py
wrywerytwreywery: Semantically, they should be equivalent, since entering one of those blocks means you won't check any of the others, and passing over one means that the others will be checked.  That said, I can make that change if you think that's better from a style perspective.

scopatz: This is non-blocking so don't worry about it.

scopatz:
Thanks @wrywerytwreywery!

PR #160, opened 2015-03-28
Job control bugfix

wrywerytwreywery:
This fixes a bug in #152 (Job Control).

The old method for waiting for child processes to terminate/stop was to wait for a `SIGCHLD` signal; when that signal was received, the active job was considered done.  The problem with this approach is that processes earlier in the pipeline also send `SIGCHLD` to the parent upon terminating/stopping.  It was possible to construct a pipeline such that the last process would be interpreted as terminated before it actually was, and it and xonsh would be left fighting over the input to the terminal.

This change implements a much simpler and more robust check for a change of status in a child process.

### DIFF
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -110,8 +110,11 @@ def wait_for_active_job():
                 obj.done = True
                 job['bg'] = True
                 job['status'] = 'stopped'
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks!

PR #162, opened 2015-03-29
Prompt via prompt

abadger:
Alternate implementation of prompt vars customization.  This makes all customization of prompts happen by giving a callable to the $PROMPT environment variable instead of having separate ways to customize prompts by giving $PROMPT a callable or giving $PROMPT_FORMATTER a prompt formatter.


tests/test_prompt.py
```diff
@@ -0,0 +1,285 @@
...
+from nose.plugins.skip import SkipTest
+
+from .tools import mock_xonsh_env
+from xonsh.tools import TERM_COLORS
+
+from xonsh.environ import (BasePrompt, DefaultPrompt,
+        DEFAULT_PROMPT, DEFAULT_TITLE, add_prompt_var)
+
+
+class Base_TestPrompt:
```
scopatz: Tests should be in a procedural style: http://xonsh.org/devguide.html#python-style-guide

xonsh/environ.py
```diff
@@ -7,72 +7,153 @@
...
 from xonsh.tools import TERM_COLORS
 
-def current_branch(cwd=None):
-    """Gets the branch for a current working directory. Returns None
-    if the cwd is not a repository.  This currently only works for git, 
-    bust should be extended in the future.
+# This is only needed because of the way that we create prompt objects in
+# shell.py.  If we find a way to create prompt objects before
+# environ.add_prompt_var() can be called then we wouldn't need this.
+_prompt_vars_queue = {}
```
scopatz: PEP8 global constants should be in all caps

scopatz: Actually, I don't really understand why this isn't an instance variable on the Prompt.

xonsh/environ.py
```diff
@@ -81,18 +162,127 @@ def format_prompt(template=default_prompt):
               BOLD_INTENSE, BACKGROUND_INTENSE
     + NO_COLOR -- Resets any previously used color codes
     """
+    def __init__(self, prompt_template, *args, **kwargs):
+        super(DefaultPrompt, self).__init__(prompt_template, *args, **kwargs)
+        self._env = builtins.__xonsh_env__
+
+        self.update(TERM_COLORS)
+        self.update(dict(
```
scopatz: I don't think the call to dict is needed here.  Update accepts the same arguments as dict.

scopatz:
Hi @abadger, so I think the major issue here is still the tests.  However, I also think that it is a little weird that you got rid of the `$PROMPT_FORMATTER`.  It is great that the formatters are callable. What would be strange to me as a user though is that if I set my $PROMPT in the rc file as a template string, when I go to retrieve that later, I would expect to see exactly the string I set previously.  However, what I would get here is some rich DefaultFormatter object that I know nothing about.  I just wanted my string back.  Maybe @wrywerytwreywery has some thoughts.

But frankly, if you added back the `$PROMPT_FORMATTER` and made the tests compliant with the existing procedural style, I would probably merge this.

xonsh/environ.py
abadger: It's not a constant.  It's being modified during runtime.

It can't be an instance variable because we may not have an instance when the toplevel add_prompt_vars() function that @wrywerytwreywery wanted to have is called.

abadger:
I thought about that but I had a different solution than $PROMPT_FORMATTER.  I thought of saving the formatter into an instance variable on the Shell class.

scopatz:
Same comment here. @abadger. Would you mind if we closed this PR?

scopatz:
I am closing due to inactivity.  Please feel free to reopen.

PR #163, opened 2015-03-29
Optimization for waiting for active job to stop

wrywerytwreywery:
This patch improves the `wait_for_active_job` method.  Rather than polling, we can just wait for `os.waitpid` to return.  This minimizes the time delay between the process actually ending and the next prompt being displayed, and probably saves some extra CPU churning by eliminating the loop.


scopatz:
Thanks!

Issue #107, opened 2015-03-19
Import Hooks for xonsh source files


scopatz:
We should have import hooks [1] that allow for xonsh modules to be imported using the normal import mechanism.  I don't think this would be super hard, but it would be awesome to have.

1. https://docs.python.org/3/reference/import.html

wrywerytwreywery:
Closed by #164.

PR #164, opened 2015-03-29
Import hooks for xonsh files (*.xsh)

scopatz:
This PR adds the capability for xonsh source files (*.xsh) to be imported using Python's normal import mechanisms, which is frankly pretty great.  In addition to being able to write xonsh scripts, libraries, and packages, it also gives us a way to source (in the bash sense) a xonsh file via:

```python
from myxonsh import *
```

I am not enamoured with the file *.xsh file extension, but this is what the mailing list sort of agreed to.  We can always change it later.  (My personal fav was *.xo).  It'd be nice to have this is v0.1.4 if you are available to review it tonight @wrywerytwreywery.


scopatz:
Also this closes #107

wrywerytwreywery:
I don't think I was in time for that conversation; `.xsh` is fine, but I kind of like `.xpy`, which I think fits better because xonsh seems to be closer to python than to the other `*sh` languages...

Anyway, I'll take a look now.

xonsh/imphooks.py
```diff
@@ -0,0 +1,78 @@
...
+    def find_spec(self, fullname, path, target=None):
+        """Finds the spec for a xonsh module if it exists."""
+        spec = None
+        path = sys.path if path is None else path
+        name = fullname.rsplit('.', 1)[-1]
+        fname = name + '.xsh'
+        for p in path:
+            if not isinstance(p, string_types):
+                continue
+            if fname not in os.listdir(p):
```
wrywerytwreywery: I ran into trouble right away with this trying to list a nonexistent location, and later trying to list a zip archive.  Adding a check before this line worked for me:
```
            if not os.path.isdir(p):
                continue
```

xonsh/imphooks.py
```diff
@@ -0,0 +1,78 @@
...
+        else:
+            execer = self._execer
+        return execer
+    #
+    # MetaPathFinder methods
+    #
+    def find_spec(self, fullname, path, target=None):
+        """Finds the spec for a xonsh module if it exists."""
+        spec = None
+        path = sys.path if path is None else path
```
wrywerytwreywery: I really wanted my PWD always to be in the path so I could always import a `.xsh` file from the current directory when using the shell.  So I might recommend something like the following after this line:
```
        path += [builtins.__xonsh_env__['PWD']]
```

Maybe this doesn't quite work if we're running a script, since, in that case, we want to add not the current working directory, but rather the location of the script we're running.  Do we have a flag somewhere to tell whether we're running interactively, versus running in a script?

wrywerytwreywery: Also, this seems not _quite_ to work after a little more testing.  With a directory structure:

```
* testdir
|---* __init__.py
|---* test_script.py
|---* test_script2.py
```

It took two tries for the import to go through.

```
>>> import testdir
  File "/usr/local/lib/python3.4/site-packages/xonsh/shell.py", line 123, in default
  File "/usr/local/lib/python3.4/site-packages/xonsh/execer.py", line 115, in exec
    return exec(code, glbs, locs)
ImportError: No module named 'testdir'
>>> import testdir
>>> testdir
<module 'testdir' from '/home/testuser/testdir/__init__.py'>
```

Putting `testdir` in my path worked fine.

wrywerytwreywery:
This is pretty great!  A couple of comments/questions above.

scopatz:
> I don't think I was in time for that conversation; .xsh is fine, but I kind of like .xpy, which I think fits better because xonsh seems to be closer to python than to the other *sh languages...

OK - both you and @crbates have said *.xpy, so I think we have a winner.  I'll make the change.

wrywerytwreywery:
> OK - both you and @crbates have said *.xpy, so I think we have a winner. I'll make the change.

Well, in the time between making my comment and seeing your response, I came back around to .xsh (since it is closer to the name xonsh), but whatever you think is best is fine, I'm sure.

scopatz:
Alright, I'll leave it as is for now and we can change later if we want to.

xonsh/imphooks.py
scopatz: good catch!

xonsh/imphooks.py
scopatz: So I think that the way to do what you want is through the traditional Python module loading hook.  Namely:

```python
import sys
sys.path.insert(0, '.')
```

I don't think that this is the sort of decision that xonsh should make for users since it could also affect Python module loading.  It is better for folks who want this behaviour to stick it in their xonshrc, I think.

scopatz: > Do we have a flag somewhere to tell whether we're running interactively, versus running in a script?

No, I don't think so, but we probably should eventually. I'll make another issue for that.

scopatz:
Alright, I think that this is ready for another look.  I just added the extra check you pointed out.

xonsh/imphooks.py
wrywerytwreywery: > I don't think that this is the sort of decision that xonsh should make for users since it could also affect Python module loading. It is better for folks who want this behaviour to stick it in their xonshrc, I think.

Shouldn't we try to mirror Python's mechanism, though?  I absolutely don't think we want to mess with `sys.path`, but if this import hook is using its own path (what is passed in to `find_spec`?), then I do think it makes sense to add a `'.'` to the start of that (I tried to do something like this non-destructively above).  Without this, it also seems that relative imports without `.` won't work (e.g., a xonsh file `A` should be able to import a file `B` in the same module by doing `import B` _or_ `from . import B`, no?)

scopatz: Oh I see.  I didn't realize that implicit relative importing was broken.  Serves me right for not testing it :)

xonsh/imphooks.py
```diff
@@ -0,0 +1,80 @@
...
+import os
+import sys
+import builtins
+from importlib.machinery import ModuleSpec
+from importlib.abc import MetaPathFinder, SourceLoader
+
+from xonsh.tools import string_types
+from xonsh.execer import Execer
+
+class XonshImportHook(MetaPathFinder, SourceLoader):
```
wrywerytwreywery: PEP8: need another newline before line 14

xonsh/imphooks.py
```diff
@@ -0,0 +1,80 @@
...
+        if hasattr(builtins, '__xonsh_execer__'):
+            execer = builtins.__xonsh_execer__
+            if self._execer is not None:
+                self._execer = None
+        elif self._execer is None:
+            self._execer = execer = Execer(unload=False)
+        else:
+            execer = self._execer
+        return execer
+    #
```
wrywerytwreywery: PEP8: not enough newlines; maybe another newline before line 33

xonsh/imphooks.py
```diff
@@ -0,0 +1,80 @@
...
+            raise ImportError(msg)
+        with open(filename, 'r') as f:
+            src = f.read()
+        src = src if src.endswith('\n') else src + '\n'
+        execer = self.execer
+        execer.filename = filename
+        ctx = {}   # dummy for modules
+        code = execer.compile(src, glbs=ctx, locs=ctx)
+        return code
+
```
wrywerytwreywery: PEP8: Whitespace in this line

xonsh/imphooks.py
scopatz: Actually - implicit relative imports were removed in Python 3.  So `import B` will/should never work in a package.  I can go ahead and add '.' for convenience.

scopatz:
Ok @wrywerytwreywery - I think that this is ready for another look!

xonsh/imphooks.py
wrywerytwreywery: I had no idea.  This project is my first time ever writing Python 3, so I'm learning all these little things...

scopatz: > I had no idea. This project is my first time ever writing Python 3, so
I'm learning all these little things...

Ahh cool.  Yeah. There are few little gotchas. I am more used to doing
single codebase Python 2 & 3 projects, so it is nice to *not* have to worry
about some of this stuff.  I had forgotten about the import lookup
mechanism though.

wrywerytwreywery:
Importing from modules in `.` still doesn't seem to work, but importing from scripts located in `.` does.  The only remaining thing, I think, is whitespace on line 85 :)  I'm happy to fix that and merge in unless you have any other reservations, or you can of course go ahead and do so.

scopatz:
> Importing from modules in . still doesn't seem to work, but importing from scripts located in . does. 

What is the module / script distinction here? I got the space.

wrywerytwreywery:
Sorry for not being clear.  Importing from a directory containing an `__init__.py` doesn't work if that directory is in `.`, but importing from a script `*.xsh` in `.` does.

scopatz:
Oh I see. OK.

scopatz:
Alright, I have it so that it now behaves like the regular Python command prompt.  `from . import B` will fail in both Python and xonsh and with the same error message.

wrywerytwreywery:
I'm not sure what the actual intended behavior is in Python3, but here's what I notice from xonsh right now (trying to get my terminology right this time, too):

* both `from . import B` and `import B` seem to work if `B.xsh` is in the same package as the module from which it is being imported
* `from . import B` does not work for two modules in the same directory outside of a package
* importing a module in PWD seems to work
* importing from a package in PWD, or importing a module from such a package, does not work (though adding `'.'` to `sys.path` seems to fix, so maybe our adding `'.'` to the `path` argument earlier wasn't actually helping anything)

scopatz:
Yes. I think that this is the behaviour of Python 3

wrywerytwreywery:
Alright; then that sounds good to me.  Any lingering reservations?

scopatz:
Yeah, let me test sub-packages really quickly.

scopatz:
Ok it seems to work

scopatz:
Now I think it is ready :)

wrywerytwreywery:
This is awesome!  I just ran it through some tests of my own, and everything seems good here, too.  I'll go ahead and merge in.

scopatz:
Woohoo!

PR #165, opened 2015-03-30
Alternative Prompt Format Refactor

wrywerytwreywery:
This is a proof-of-concept for an alternative prompt formatter implementation.

New variables for the format string are added to the environment variable `$FORMAT_DICT`.  They can either be strings (which are inserted into the template verbatim) or callables (which are called each time, and whose result is inserted into the template).

`$PROMPT` can still be a callable (which returns the template string) or a string itself.

Values are only computed/used if they show up in the template string.

I have not written tests yet, but I like this structure a lot.  I would be happy to write tests if this is what we want to go with.

Examples:
```
>>> $FORMAT_DICT['test'] = "hey"
>>> $PROMPT = "{test} {cwd} $ "
hey ~ $
```
```
>>> import random
>>> $FORMAT_DICT['test'] = lambda: random.randint(1,9)
>>> $PROMPT = "{test} {cwd} $ "
9 ~ $ 
2 ~ $ 
1 ~ $ 
2 ~ $ 
9 ~ $ 
```

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -86,20 +86,6 @@ _formatter = string.Formatter()
 
 def format_prompt(template=DEFAULT_PROMPT):
     """Formats a xonsh prompt template string.
-
-    The following keyword arguments are recognized in the template string:
-
-    + user -- Name of current user
-    + hostname -- Name of host computer
-    + cwd -- Current working directory
-    + curr_branch -- Name of current git branch (preceded by a space), if any
-    + (QUALIFIER\_)COLORNAME -- Inserts an ANSI color code
-        - COLORNAME can be any of:
-              BLACK, RED, GREEN, YELLOW, BLUE, PURPLE, CYAN, WHITE
-        - QUALIFIER is optional and can be any of:
-              BOLD, UNDERLINE, BACKGROUND, INTENSE,
-              BOLD_INTENSE, BACKGROUND_INTENSE
-    + NO_COLOR -- Resets any previously used color codes
     """
     env = builtins.__xonsh_env__
     template = template() if callable(template) else template
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -12,8 +12,7 @@ from warnings import warn
 from xonsh import __version__ as XONSH_VERSION
 from xonsh.tools import TERM_COLORS
 
-
-def current_branch(cwd=None):
+def current_branch(cwd=None, pad=True):
     """Gets the branch for a current working directory. Returns None
     if the cwd is not a repository.  This currently only works for git,
     bust should be extended in the future.
@@ -61,15 +60,24 @@ def current_branch(cwd=None):
         except subprocess.CalledProcessError:
             pass
 
+    if pad and branch is not None:
+        branch = ' ' + branch
     return branch
 
 
-default_prompt = ('{BOLD_GREEN}{user}@{hostname}{BOLD_BLUE} '
+DEFAULT_PROMPT = ('{BOLD_GREEN}{user}@{hostname}{BOLD_BLUE} '
                   '{cwd}{BOLD_RED}{curr_branch} {BOLD_BLUE}${NO_COLOR} ')
-default_title = '{user}@{hostname}: {cwd} | xonsh'
+DEFAULT_TITLE = '{user}@{hostname}: {cwd} | xonsh'
+
+FORMAT_DICT = dict(user=os.environ.get('USER', '<user>'),
+                   hostname=socket.gethostname().split('.',1)[0],
+                   fqdn=socket.getfqdn(),
+                   cwd=lambda: builtins.__xonsh_env__['PWD'].replace(builtins.__xonsh_env__['HOME'], '~'),
+                   curr_branch=lambda: current_branch() or '',
+                   **TERM_COLORS)
 
 
-def format_prompt(template=default_prompt):
+def format_prompt(template=DEFAULT_PROMPT):
     """Formats a xonsh prompt template string.
 
     The following keyword arguments are recognized in the template string:
@@ -87,14 +95,10 @@ def format_prompt(template=default_prompt):
     + NO_COLOR -- Resets any previously used color codes
     """
     env = builtins.__xonsh_env__
-    cwd = env['PWD']
-    branch = current_branch(cwd=cwd)
-    branch = '' if branch is None else ' ' + branch
-    p = template.format(user=env.get('USER', '<user>'),
-                        hostname=socket.gethostname(),
-                        cwd=cwd.replace(env['HOME'], '~'),
-                        curr_branch=branch, **TERM_COLORS)
-    return p
+    fmt = env.get('FORMAT_DICT', FORMAT_DICT)
+    fmt = {k:v for (k,v) in fmt.items() if '{{{}}}'.format(k) in template}
+    fmt = {k:(v() if callable(v) else v) for (k,v) in fmt.items()}
+    return template.format(**fmt)
 
 
 RE_HIDDEN = re.compile('\001.*?\002')
@@ -123,8 +127,9 @@ def multiline_prompt():
 BASE_ENV = {
     'XONSH_VERSION': XONSH_VERSION,
     'INDENT': '    ',
-    'PROMPT': default_prompt,
-    'TITLE': default_title,
+    'FORMAT_DICT': FORMAT_DICT,
+    'PROMPT': DEFAULT_PROMPT,
+    'TITLE': DEFAULT_TITLE,
     'MULTILINE_PROMPT': '.',
     'XONSHRC': os.path.expanduser('~/.xonshrc'),
     'XONSH_HISTORY_SIZE': 8128,
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -13,6 +13,7 @@ from warnings import warn
 from xonsh import __version__ as XONSH_VERSION
 from xonsh.tools import TERM_COLORS
 
+
 def current_branch(cwd=None, pad=True):
     """Gets the branch for a current working directory. Returns None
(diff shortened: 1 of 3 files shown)


xonsh/environ.py
```diff
@@ -82,21 +92,15 @@ def format_prompt(template=default_prompt):
...
-            user=env.get('USER', '<user>'),
-            hostname=socket.gethostname(),
-            cwd=cwd.replace(env['HOME'], '~'),
-            curr_branch=branch,
-            **TERM_COLORS
-            )
-    return p
+    fmt = env.get('FORMAT_DICT', FORMAT_DICT)
+    fmt = {k:v for (k,v) in fmt.items() if '{{{}}}'.format(k) in template}
+    fmt = {k:(v() if callable(v) else v) for (k,v) in fmt.items()}
```
scopatz: PEP8 single space after colons

scopatz: and commas :)

xonsh/environ.py
```diff
@@ -82,21 +92,15 @@ def format_prompt(template=default_prompt):
...
-    p = template.format(
-            user=env.get('USER', '<user>'),
-            hostname=socket.gethostname(),
-            cwd=cwd.replace(env['HOME'], '~'),
-            curr_branch=branch,
-            **TERM_COLORS
-            )
-    return p
+    fmt = env.get('FORMAT_DICT', FORMAT_DICT)
+    fmt = {k:v for (k,v) in fmt.items() if '{{{}}}'.format(k) in template}
```
scopatz: I think that this might deserve a regular expression, because what if someone wanted to do something with more advanced string formatting such as `'{test:>10}'`. I think the right regex is something like:

```python
'[{]([^:]*):?.*?[}]'
```

scopatz:
I like this mechanism a lot more than the other options we currently have, if for no other reason that this is a lot less code.

scopatz:
It also feels more intuitive to me. How do you feel about it @abadger?

xonsh/environ.py
wrywerytwreywery: Right you are.  I usually use `%` in my own code, so I didn't even know that anything fancy was possible with `format`.  I think we can't get this right with a simple regex, but we could use `string.Formatter.parse` to do the right thing without any hassle (the names will be the second element in each tuple it returns).

xonsh/environ.py
wrywerytwreywery: I'll get used to this PEP8 stuff eventually...

scopatz: haha me too :)

xonsh/environ.py
scopatz: Ahh even better.

xonsh/environ.py
```diff
@@ -82,21 +97,17 @@ def format_prompt(template=default_prompt):
...
-            cwd=cwd.replace(env['HOME'], '~'),
-            curr_branch=branch,
-            **TERM_COLORS
-            )
-    return p
+    template = template() if callable(template) else template
+    fmt = env.get('FORMAT_DICT', FORMAT_DICT)
+    included_names = set(i[1] for i in FORMATTER.parse(template))
+    fmt = {k: v for (k, v) in fmt.items() if k in included_names}
+    fmt = {k: (v() if callable(v) else v) for (k, v) in fmt.items()}
```
scopatz: This might be faster as a single loop over fmt, though this is certainly the most expressive. Probably doesn't matter too much.

wrywerytwreywery: I originally had those two combined, but that led to a line that was way longer than the 80 character limit, and I couldn't find a nice place to put a linebreak.  I should probably just use a for loop to construct the dict (save some time versus looping twice).

scopatz:
I was gonna do the v0.1.4 release soon here.  I can wait an hour or two though, if you wanted to add tests and see this in the release.

wrywerytwreywery:
While it would be nice to see it in this release, I think I'd really better wait until tomorrow.  No need to delay.

scopatz:
Ok!

scopatz:
Don't kill yourself over it :)

abadger:
I don't like it nearly as much.

* If you want to extend this right, you have to  do:
``` python
import  environ.FORMAT_DICT
import copy
fdict = copy.copy(FORMAT_DICT)
fdict['newvar'] = 'something'
$PROMPT_FORMAT. = fdict
```
That doesn't seem right, especially after @wrywerytwreywery asked for a single easy to use function in my other implementation.
* This leaves out the call-only-once feature that I added to the other implementation as a response to some things that @mbollman pointed out.
* This pushes more logic into format_prompt() which means that anyone needing to extend the types of formatters (for instance, to add call-only-once) will have to duplicate the format_prompt() functionality instead of being able to override just the specific pieces of the logic that are insufficient for their needs. 

All that said, if you guys don't like OOP then you just won't like any implementation I'll come up with so I guess I'll stop working on this and let you all implement this as you like.

wrywerytwreywery:
I'm sorry, but, to be honest, I'm not sure I understand your arguments.

> If you want to extend this right,...

You actually don't need to make a copy of the dictionary, nor do you have to import anything.  You can just do the following, for example, from within your xonshrc: 
`$FORMAT_DICT['newvar'] = 'something'`

> This leaves out the call-only-once feature that I added to the other implementation as a response to some things that @mbollman pointed out.

This doesn't preclude the possibility of computing something only once.  For example:

```
$FORMAT_DICT['something'] = some_function()  # this will only be called once
$FORMAT_DICT['something_else'] = some_function  # this will be called every time
```

wrywerytwreywery:
That said, this PR is just my suggestion for dealing with the issue.  Ultimately, it's up to @scopatz.

abadger:
Actually you're almost right -- you need to do a copy when you put the FORMAT_DICT into the default xonsh environment on line 134.  I wasn't thinking that FORMAT_DICT was put into the default environment.  (Someone has to make a copy because otherwise there's no way to restore the FORMAT_DICT to its default state and [less importantly now but could be of interest later] changing global values isn't nice with any sort of parallel processing code.)

For the computed only once case, your example doesn't address quite the same thing as my case.  Your example computes only once but It is computed even if the value is not used in the prompt.  Getting both of those properties is why I could no longer rely only on whether the format value was callable().  As I wrote in the documentation in my pull request, that comes in handy when you have to make a long computation to determine a value and that value may not be used in everyone's prompt string.  (Retrieving the fqdn is one example of that as problems in DNS can mean the code waits for a timeout before falling back to gethostname().)

wrywerytwreywery:
> Your example computes only once but It is computed even if the value is not used in the prompt.

This is untrue.  The dictionary comprehension on lines 103-105 makes sure only to call these functions when they are needed.  For example:

```
>>> def test_function():
...     print('i am being called')
... 
>>> $FORMAT_DICT['not_in_string'] = test_function
>>> 
>>> #notice, not being printed (function is not called)
>>> 
>>> $PROMPT = '{not_in_string} $ '
i am being called
None $  # now being printed
i am being called
```

> Someone has to make a copy because otherwise there's no way to restore the FORMAT_DICT to its default state.

I don't see that as being a common use case.  You can avoid needing to restore to the original state by simply not overwriting any of the preexisting keys, and because `format_prompt` will not call functions that don't exist in the `$PROMPT` string, it doesn't cost much to store these extra values.

wrywerytwreywery:
Sorry for misunderstanding your earlier comment.  You are right that the "only once" functions are called, regardless of whether they actually show up in the prompt or not.  To me, this is a minor deficiency when weighed against the simplicity we expose to the end user.  I do think, though, that this means that fqdn shouldn't be in the defaults.

@scopatz, do you have an opinion on anything we've been discussing?

abadger:
Well similarly, it doesn't cost much to do the copy when you set it in the default environment on line 134.  Then you avoid any problems for people later.  I mean, if people set the DEFAULT_PROMPT string to their random value in their .xonshrc we'd tell them they were doing it wrong.  This is the same thing except you're dealing with a mutable value here and that's what's giving it the appearance of acceptablitiy.

wrywerytwreywery:
That is fair.  I can make it copy the dictionary.

scopatz:
@wrywerytwreywery let me know when this is ready for another look!

wrywerytwreywery:
I think all of the functionality that I want is there, so this is ready for another look (although it will need a rebase from master).  If we want to move forward with this, I can write some docs and tests.

I also think I want to change `FORMAT_DICT` to `FORMATTER_DICT` so that it can't be confused with a verb (and thus a function).

scopatz:
That all sound good to me.  I think I prefer this method to the other options on the table right now.

wrywerytwreywery:
Alright; I added some docs for this, and I think the code is all set.

Do we need tests for this?  The prompt didn't have any tests before, but I can do so if you think they're necessary.

xonsh/environ.py
```diff
@@ -61,40 +62,39 @@ def current_branch(cwd=None):
...
 
-def format_prompt(template=default_prompt):
-    """Formats a xonsh prompt template string.
+def _replace_home(x):
+    return x.replace(builtins.__xonsh_env__['HOME'], '~')
+
+FORMATTER_DICT = dict(user=os.environ.get('USER', '<user>'),
+                      hostname=socket.gethostname().split('.', 1)[0],
+                      cwd=lambda: _replace_home(builtins.__xonsh_env__['PWD']),
+                      curr_branch=lambda: current_branch() or '',
```
scopatz: It seems like this extra level of indirection isn't great.  Instead maybe `current_branch()` should be altered to to return an empty string when not on a branch.  This is non-blocking.

scopatz:
Thanks @wrywerytwreywery!  I would eventually like to see some tests for this, but it didn't have tests originally and it is the kind of thing where everyone will notice if it fails.  So for now, I think the benefits outweigh the costs. Thanks again!

PR #168, opened 2015-03-31
more robust subproc injection

### LINKED ISSUE #167: Parser error for  "ls file; ls file"
So `ls file; ls file` will now enter multiline mode, and then fail.  Defining a function will similarly fail:

```console
scopatz@athenaie ~ $ def f():
`·.,¸,.·*¯`·.,¸,.·*¯     ls file
`·.,¸,.·*¯`·.,¸,.·*¯     ls file
`·.,¸,.·*¯`·.,¸,.·*¯ 
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/execer.py", line 125, in _parse_ctx_free
    debug_level=self.debug_level)
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/parser.py", line 282, in parse
    debug=debug_level)
  File "/home/scopatz/miniconda3/lib/python3.4/site-packages/ply/yacc.py", line 265, in parse
    return self.parseopt_notrack(input,lexer,debug,tracking,tokenfunc)
  File "/home/scopatz/miniconda3/lib/python3.4/site-packages/ply/yacc.py", line 1047, in parseopt_notrack
    tok = self.errorfunc(errtoken)
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/parser.py", line 2154, in p_error
    column=p.lexpos))
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/parser.py", line 351, in _parse_error
    raise err
  File "<string>", line None
SyntaxError: <xonsh-code>:2:0: ('code: $[',)

  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/shell.py", line 141, in push
    locs=self.ctx)
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/execer.py", line 89, in compile
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/execer.py", line 67, in parse
    tree = self._parse_ctx_free(input, mode=mode)
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/execer.py", line 156, in _parse_ctx_free
    raise original_error
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/execer.py", line 125, in _parse_ctx_free
    debug_level=self.debug_level)
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/parser.py", line 282, in parse
    debug=debug_level)
  File "/home/scopatz/miniconda3/lib/python3.4/site-packages/ply/yacc.py", line 265, in parse
    return self.parseopt_notrack(input,lexer,debug,tracking,tokenfunc)
  File "/home/scopatz/miniconda3/lib/python3.4/site-packages/ply/yacc.py", line 1047, in parseopt_notrack
    tok = self.errorfunc(errtoken)
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/parser.py", line 2154, in p_error
    column=p.lexpos))
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/parser.py", line 351, in _parse_error
    raise err
  File "<string>", line None
SyntaxError: <xonsh-code>:2:7: ('code: file',)
```

Looks like the execer still needs some work :)

scopatz:
This should fix #167, and many other relevant cases.

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -51,7 +51,8 @@ def subproc_toks(line, mincol=-1, maxcol=None, lexer=None, returnline=False):
             break
         if len(toks) == 0 and tok.type in ('WS', 'INDENT'):
             continue  # handle indentation
-        elif len(toks) > 0 and toks[-1].type == 'SEMI':
+        elif len(toks) > 0 and toks[-1].type == 'SEMI' and pos < maxcol and \
+                tok.type not in ('NEWLINE', 'DEDENT', 'WS'):
             toks.clear()
         if pos < mincol:
             continue
@@ -77,6 +78,8 @@ def subproc_toks(line, mincol=-1, maxcol=None, lexer=None, returnline=False):
         else:
             el = line[pos:].split('#')[0].rstrip()
             end_offset = len(el)
+    if toks[-1].type == 'SEMI':
+        toks.pop()
     if len(toks) == 0:
         return  # handle comment lines
     beg, end = toks[0].lexpos, (toks[-1].lexpos + end_offset)
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -32,9 +32,10 @@ else:
 
 DEFAULT_ENCODING = sys.getdefaultencoding()
 
+
 def subproc_toks(line, mincol=-1, maxcol=None, lexer=None, returnline=False):
-    """Excapsulates tokens in a source code line in a uncaptured 
-    subprocess $[] starting at a minimum column. If there are no tokens 
+    """Excapsulates tokens in a source code line in a uncaptured
+    subprocess $[] starting at a minimum column. If there are no tokens
     (ie in a comment line) this returns None.
     """
     if lexer is None:
@@ -88,18 +89,22 @@ def subproc_toks(line, mincol=-1, maxcol=None, lexer=None, returnline=False):
         rtn = line[:beg] + rtn + line[end:]
     return rtn
 
+
 def decode(s, encoding=None):
     encoding = encoding or DEFAULT_ENCODING
     return s.decode(encoding, "replace")
+
 
 def encode(u, encoding=None):
     encoding = encoding or DEFAULT_ENCODING
     return u.encode(encoding, "replace")
 
+
 def cast_unicode(s, encoding=None):
     if isinstance(s, bytes):
         return decode(s, encoding)
     return s
+
 
 def safe_hasattr(obj, attr):
     """In recent versions of Python, hasattr() only catches AttributeError.
@@ -110,6 +115,7 @@ def safe_hasattr(obj, attr):
         return True
     except:
         return False
+
 
 def indent(instr, nspaces=4, ntabs=0, flatten=False):
     """Indent a string a given number of spaces or tabstops.
@@ -216,8 +222,9 @@ TERM_COLORS = {
     'BACKGROUND_INTENSE_WHITE': '\001\033[0;107m\002',   # WHITE
     }
 
-# The following redirect classes were taken directly from Python 3.5's source 
-# code (from the contextlib module). This can be removed when 3.5 is released, 
+
+# The following redirect classes were taken directly from Python 3.5's source
+# code (from the contextlib module). This can be removed when 3.5 is released,
 # although redirect_stdout exists in 3.4, redirect_stderr does not.
 # See the Python software license: https://docs.python.org/3/license.html
 # Copyright (c) Python Software Foundation. All rights reserved.
@@ -251,7 +258,7 @@ class redirect_stdout(_RedirectStream):
             with redirect_stdout(f):
                 help(pow)
 
-    Mostly for backwards compatibility. 
+    Mostly for backwards compatibility.
     """
     _stream = "stdout"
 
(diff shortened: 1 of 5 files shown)


wrywerytwreywery:
This is a huge improvement!

wrywerytwreywery:
I noticed an issue (not sure if there is any way around this).  Single-token subprocess expressions seem not to work inside scripts.  So, for example, the following works from the shell and in a script:

```
for i in range(20):
    cd ~/xonsh
    ls .
    for i in range(5):
        print('howdy')
```

The following works from the shell, but not in a script:
```
for i in range(20):
    cd ~/xonsh
    ls
    for i in range(5):
        print('howdy')
```

It fails with the following error:
```
    main()
  File "/usr/local/lib/python3.4/site-packages/xonsh/main.py", line 53, in main
    code = shell.execer.compile(code, mode='exec', glbs=shell.ctx)
  File "/usr/local/lib/python3.4/site-packages/xonsh/execer.py", line 92, in compile
    code = compile(tree, self.filename, mode)
TypeError: expected some sort of stmt, but got <_ast.Call object at 0x7fe242329cf8>
```

In fact, just something like `ls` on one line by itself in a script seems not to work.

After looking a little more closely, it seems that things that can parse down to valid Python give this error (so `ls -la` fails in the same way).  Interestingly, `ls ;` fails in a different way:
```
    main()
  File "/usr/local/lib/python3.4/site-packages/xonsh/main.py", line 53, in main
    code = shell.execer.compile(code, mode='exec', glbs=shell.ctx)
  File "/usr/local/lib/python3.4/site-packages/xonsh/execer.py", line 89, in compile
  File "/usr/local/lib/python3.4/site-packages/xonsh/execer.py", line 77, in parse
    tree = self.ctxtransformer.ctxvisit(tree, input, ctx, mode=mode)
  File "/usr/local/lib/python3.4/site-packages/xonsh/ast.py", line 80, in ctxvisit
    node = self.visit(node)
  File "/usr/local/lib/python3.4/ast.py", line 245, in visit
    return visitor(node)
  File "/usr/local/lib/python3.4/ast.py", line 301, in generic_visit
    value = self.visit(value)
  File "/usr/local/lib/python3.4/ast.py", line 245, in visit
    return visitor(node)
  File "/usr/local/lib/python3.4/site-packages/xonsh/ast.py", line 188, in visit_For
    self.generic_visit(node)
  File "/usr/local/lib/python3.4/ast.py", line 301, in generic_visit
    value = self.visit(value)
  File "/usr/local/lib/python3.4/ast.py", line 245, in visit
    return visitor(node)
  File "/usr/local/lib/python3.4/site-packages/xonsh/ast.py", line 141, in visit_Expr
    return self.try_subproc_toks(node)
  File "/usr/local/lib/python3.4/site-packages/xonsh/ast.py", line 108, in try_subproc_toks
    newnode = newnode.body
AttributeError: 'NoneType' object has no attribute 'body'
```

scopatz:
Nice catch @wrywerytwreywery.  This was actually a bug in the context sensitive AST transformer.  I have pushed up a fix to this branch.

xonsh/tools.py
```diff
@@ -49,13 +49,22 @@ def subproc_toks(line, mincol=-1, maxcol=None, lexer=None, returnline=False):
         pos = tok.lexpos
         if tok.type != 'SEMI' and pos >= maxcol:
             break
-        if len(toks) > 0 and toks[-1].type == 'SEMI':
+        if len(toks) == 0 and tok.type in ('WS', 'INDENT'):
+            continue  # handle indentation
+        elif len(toks) > 0 and toks[-1].type == 'SEMI':
```
wrywerytwreywery: I think this is the line responsible for the `NoneType` error from having `ls;` inside a block in a script.  What is happening was that there is a `DEDENT` token after the semicolon, so the line gets cleared.  Modifying this condition to be as follows seems to fix:
```
        elif len(toks) > 0 and toks[-1].type == 'SEMI' \
                and pos < maxcol and tok.type not in ('NEWLINE', 'DEDENT', 'WS'):
            toks.clear()
```
I think that's all the conditions we want:
* the last thing we saw is was a semicolon
* this token should still be considered (`pos < maxcol`)
* this token is an actual meaningful token (not `DEDENT` or `NEWLINE` or `WS`)

wrywerytwreywery:
Hi, @scopatz.  Your last push seems to have fixed the problem, except for the `ls;` case.  I've put into an inline comment what I believe to be a fix for that case.

scopatz:
Your fix doesn't quite work for me.  I am going to have to take a look more closely later.

wrywerytwreywery:
Do you have an example of code that my fix fails on?  I haven't been able to break it...

scopatz:
I have just pushed up the fix with a failing test.

scopatz:
Working on this now.

scopatz:
Ok @wrywerytwreywery - I believe that I have the full fix in now.  I also added a bug fix for the XonshConsoleLexer that was causing pygments to guess this lexer for all files.

wrywerytwreywery:
Are we expecting `$[ls;]` or `$[ls; echo]` to work, or no?  That is, is `$[...]` intended to contain a single command only?  Bash allows these constructs, but they seem to break in xonsh as of now.

wrywerytwreywery:
I also made a pass for PEP8 fixes in `tools.py`.  Should I push those changes to this branch?  Or do something else with them?

scopatz:
You can push to this branch.

scopatz:
> Are we expecting $[ls;] or $[ls; echo] to work, or no? That is, is $[...] intended to contain a single command only? Bash allows these constructs (like $(ls; echo "hey")), but they seem to break in xonsh as of now.

Hey!  Sorry. Now that I am back to my computer I can respond with a real keyboard. So I think that the Python semicolon should take precedence over the bash or subprocess one.  Here is how I think your examples should be processed or written by a user:

```python
# orig                   -> mine
$[ls;]                    -> $[ls];
$[ls; echo]           -> $[ls]; $[echo]
$(ls; echo "hey") -> $(ls) + $(echo "hey")
```

Right now the execer, if it is given a line that looks like `ls; echo "hey"` will process it to look like `$[ls]; $[echo "hey"]`.  

I think that there are many reasons for doing it this way, but the important ones are:

* If the first subprocess call fails, the second one won't be executed - which is normal for Python.
* Semi-colons might not always be valid separators in the underlying OS. specifically,  I think something like `$[ls; echo]` could cause problems on Windows
* we don't currently run the subprocess in shell mode, so shell syntax could break things.
* correctly encapsulating an entire line would be difficult in the `CtxAwareTransformer` class since you typically only have the expression statement node that you are on, and not the next one too. (This isn't impossible, but it would be much more confusing)

Overall, I think that this method where "One command, One subprocess" is just a lot cleaner to reason about, for users (and developers?). Sure it might make the `subproc_toks()` a little more sneaky, but the benefits elsewhere are worth it.

wrywerytwreywery:
I think that all sounds right to me; my only thought was whether xonsh can/should do some of those transformations for the user.  As a user, I hoped/expected `$[ANYTHING]` to be the same as `ANYTHING`, assuming that `ANYTHING` is a piece of subprocess mode code (i.e., `$[ls;]` should behave the same as `ls;`).  That's hard from an implementation standpoint, though, and maybe it's unnecessary, particularly now that subprocess mode code seems to be usable just about everywhere without the `$` escapes (including in function defs and loops now).

scopatz:
Ahhh that is an interesting idea. I think that would have to be done in the parser, and is probably beyond the scope here. Something to keep in mind though!

wrywerytwreywery:
Okay, great!  I will go ahead and merge in.

Issue #166, opened 2015-03-30
Add flag for running interactively or not


scopatz:
Probably should have this as an environment variable.

wrywerytwreywery:
Closing since this has been implemented (#169)

PR #169, opened 2015-03-31
Interactive Flag, and Parser Bugfix

wrywerytwreywery:
Two changes here:
* Include a boolean environment variable called `$XONSH_INTERACTIVE` (#166), which is `True` if we are running in interactive mode and `False` otherwise (running a single command or running a script).
* Modify `parser.py` to convert environment variables to strings when they are part of a subprocess expression (so that, e.g., `echo $SOMETHING` can be called, where `$SOMETHING` is neither a string nor a list of strings).


scopatz:
Great! thanks!

PR #170, opened 2015-03-31
Attempt to fix #148

### LINKED ISSUE #148: File-redirect and `&` aren't compatible
```bash
ls > file &
```
The above doesn't create `file`. The reason for the bug is in `built_ins.run_subproc`. When it's a background job it doesn't wait, and consequently doesn't write to the file at the end.

Maybe the fix would be send the object created by `open` as the `stdout` argument to `Popen`? Then, presumably, the process will begin writing to the file before all the output is computed, which is more standard behavior.

wrywerytwreywery:
This is an attempt to fix #148, implementing @SmartViking's suggestion of feeding an appropriate file object directly into the constructor for `Popen`.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -427,7 +427,15 @@ def run_subproc(cmds, captured=True):
         write_target = cmds[-1][0]
         write_mode = WRITER_MODES[cmds[-2]]
         cmds = cmds[:-2]
-        last_stdout = PIPE
+        if write_target is not None:
+            try:
+                last_stdout = open(write_target, write_mode)
+            except FileNotFoundError:
+                e = 'xonsh: {0}: no such file or directory'
+                print(e.format(write_target))
+                return
+        else:
+            last_stdout = PIPE
     last_cmd = cmds[-1]
     prev = None
     procs = []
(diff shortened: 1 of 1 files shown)


xonsh/built_ins.py
```diff
@@ -427,7 +427,15 @@ def run_subproc(cmds, captured=True):
         write_target = cmds[-1][0]
         write_mode = WRITER_MODES[cmds[-2]]
         cmds = cmds[:-2]
-        last_stdout = PIPE
+        if write_target is not None:
+            try:
+                last_stdout = open(write_target, write_mode)
```
scopatz: I am not seeing where this file is closed....

scopatz:
Thanks for putting this in!

xonsh/built_ins.py
wrywerytwreywery: Ah; I had thought that subprocess would close this itself when it was done (since `close_fds` defaults to `True`), but I must have been misunderstanding.  I added a check to the end of `run_subproc` to close the file.

scopatz:
Cool thanks!  Is this ready now?

wrywerytwreywery:
Yep; should be.

scopatz:
Great work, as always @wrywerytwreywery :)

PR #173, opened 2015-03-31
Pygments Hooks

scopatz:
Here are hooks into pygments that add `xonsh` and `xonshcon` (xonsh console, like pycon) lexers if setuptools is available on install.  This is not 100% in that:

* closing braces in switching between pymode and subproc mode can mess up after multiple nestings
* naked subproc lines (`ls -l` rather than `$[ls -l]`) will receive Python mode highlighting. I am not actually sure that there is a way around this.

Even with these deficiencies, this is still *much* better than nothing and makes our docs a ton more readable. Thus, I think that it should be merged in after review.

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -40,3 +40,4 @@ For those of you who want the gritty details.
     tools
     openpy
     main
+    pyghooks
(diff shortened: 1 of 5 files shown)


scopatz:
This also opens the door for doing syntax highlighting as the user types, potentially.

wrywerytwreywery:
Looks great!  Just made a couple small PEP8 fixes and merged in.

scopatz:
Sweet thank!

PR #175, opened 2015-04-01
Un-detypes env for mutable value access

### LINKED ISSUE #174: Updating $PATH error
Updating $PATH or other container variables in-place does not invalidate the detype'd cache.  Should be a quick fix, but I don't want to forget.

scopatz:
This should fix #174 and is built on #168 (because my text editor needs to have syntax highlighting ~_~)


wrywerytwreywery:
Looks good to me!  I'd like to wait to merge until we figure out what to do about #168, though, unless we want to put together a separate version of this without those other changes (or maybe we could just merge in commit f150048aace39fd6323a4f4de5f1bcb0b90ea93b, though I'm not sure how to do that kind of selective merge in git).

scopatz:
We can wait :)

PR #180, opened 2015-04-02
Glob Fix

wrywerytwreywery:
If globbing an expression containing `*` fails, I think the right behavior is to add that string verbatim to the argument list, rather than ignoring it.  For example, consider the following:
`$ scp other_machine:file*.txt .`

Currently, xonsh tries to expand the whole expression `other_machine:file*.txt` on the local machine, finds no matches, and effectively deletes it from the argument list by replacing it with the (empty) results of globbing.

This patch should fix this issue.


scopatz:
Thanks @wrywerytwreywery!

PR #182, opened 2015-04-03
some landscape fixes

scopatz:
Not ready yet.  Just using the PR to trigger landscape.

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -56,7 +56,7 @@ def source_bash(args, stdin=None):
         inp = 'source {0}\nenv >> {1}\n'.format(args, f.name)
         try:
             subprocess.check_output(['bash'],
-                                    input=input,
+                                    input=inp,
                                     env=denv,
                                     stderr=subprocess.PIPE,
                                     universal_newlines=True)
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -5,7 +5,6 @@ import platform
 import builtins
 import subprocess
 import shlex
-import signal
 from warnings import warn
 
 from xonsh.dirstack import dirs, pushd, popd
@@ -38,7 +37,7 @@ def cd(args, stdin=None):
     return None, None
 
 
-def exit(args, stdin=None):
+def exit(args, stdin=None):  # pylint:disable=redefined-builtin,W0622
     """Sends signal to exit shell."""
     builtins.__xonsh_exit__ = True
     kill_all_jobs()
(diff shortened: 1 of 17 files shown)


scopatz:
OK, I think this is ready to merge.  Landscape it being slow, but this definitely improves our score.

xonsh/aliases.py
```diff
@@ -54,7 +53,7 @@ def source_bash(args, stdin=None):
     denv = env.detype()
     with tempfile.NamedTemporaryFile(mode='w+t') as f:
         args = ' '.join(args)
-        input = 'source {0}\nenv >> {1}\n'.format(args, f.name)
+        inp = 'source {0}\nenv >> {1}\n'.format(args, f.name)
         try:
             subprocess.check_output(['bash'],
                                     input=input,
```
wrywerytwreywery: Should now be `input=inp,`

xonsh/completer.py
```diff
@@ -133,7 +131,7 @@ def bash_complete(self, prefix, line, begidx, endidx):
         fnme = self.bash_complete_files.get(cmd, None)
         if func is None or fnme is None:
             return set()
-        idx = 0
+        idx = n = 0
```
wrywerytwreywery: What is the rationale behind this change?  The loop that follows will assign `n` anyway, no?

xonsh/parser.py
```diff
@@ -634,10 +633,10 @@ def p_varargslist(self, p):
         p1, p2 = p[1], p[2]
         p3 = p[3] if lenp > 3 else None
         p4 = p[4] if lenp > 4 else None
-        p5 = p[5] if lenp > 5 else None
+        #p5 = p[5] if lenp > 5 else None
         p6 = p[6] if lenp > 6 else None
         p7 = p[7] if lenp > 7 else None
-        p8 = p[8] if lenp > 8 else None
+        #p8 = p[8] if lenp > 8 else None
```
wrywerytwreywery: If we want to delete these, should we just delete them (rather than commenting out)?

wrywerytwreywery:
I just took a quick look; a few questions/comments in-line.

xonsh/completer.py
scopatz: `n` is used outside of the for-loop below, and so if by some chance `splt` is length-0, then this prevents a name lookup error.  I agree that this is pretty obtuse, but it was one of the things that landscape was yelling about.  It is better and easier to 'fix' it than to suppress the error.

xonsh/parser.py
scopatz: likely :) On it now.

scopatz:
Thanks for the review @wrywerytwreywery. I think this is ready for another look.

xonsh/completer.py
wrywerytwreywery: Ah, alright; I didn't catch that.

wrywerytwreywery:

PR #183, opened 2015-04-03
Many PEP8 Changes

wrywerytwreywery:
With the help of [yapf](https://github.com/google/yapf), I tried to minimize the number of PEP8 violations in the codebase and make things more readable, at least to my eyes.  I think there are no substantive changes here (and the tests I tried suggest that these changes don't break anything).  This may or may not be something we want to try to do now (reformatting like this can always wait, I suppose).


scopatz:
Sooner is better, probably :).

scopatz:
Thanks!

Issue #181, opened 2015-04-02
Add env variable for whether quotes should be added to tab-completions of files with spaces inthe name


scopatz:
See #68

wrywerytwreywery:
I am not sure how I feel about this.  Not adding quotes means that tab completion (as currently implemented) would lead to completions that aren't valid.  To me, it seems like if you want to turn off quotes, the only other valid option is not to offer completions for names with spaces in them.  Alternatively, we could allow escaping spaces with `\`, but that feels weird (to me, the "right" way in xonsh is to use Python strings for this, though I could understand that that might be a turn-off for some).

scopatz:
I am not a huge fan of this either, which is why I won't implement it personally :).  Though I do empathize with people who want tools to work for them.  So if someone plopped down a PR that implemented this, I'd be all for including it.  I think the current behaviour should remain as the default, though.

mfeif:
Hi there. You say that you want current behaviour to remain the default;

I'm having a weird thing wherein the "current behavior" is different on my Mac laptop vs my Ubuntu server. On Ubuntu:

```
touch "my file.txt"
ls my<TAB>
(leads to...)
ls my file.txt
```

On OSX:
```
touch "my file.txt"
ls my<TAB>
(leads to...)
ls 'my file.txt'
```

Interestingly, on Ubuntu, if I preface the `m` with a `"` then xonsh changes it into:
`ls 'my file.txt'`

I just installed xonsh today, after hearing you on the podcast, so the environments should be "pretty" clean. I didn't have any weirdo stuff going on (though trying to set the PROMPT ala the tutorial doesn't work on OSX).

My thought is that "correct" is to make valid commands/filenames, etc. And to conform to how bash/etc work(ed). I like the way it's working on OSX Any idea why these two platforms are different, and how I might be able to address it?

Both shells are being launched directly when logging in (not from an existing bash/sh/tsh/whatever).

gforsyth:
Hey @mfeif -- what are the contents of `$BASH_COMPLETIONS` on your Ubuntu box?  

And how did you install xonsh?

mfeif:
In both cases, I installed via pip from a zip file pulled from a github tag (xonsh-0.2.3.zip). In the case of the server, I installed with sudo.

`$BASH_COMPLETIONS` are: ['/etc/bash_completion', '/usr/share/bash-completion/completions/git']

gforsyth:
[...]

Issue #101, opened 2015-03-19
Tab-completing .. does not append / because Ellipsis exists.


ivh:
When trying to type "cd ../.." or similar, I use TAB to get the forward slash. This does not work in xonsh because the three-dot Ellipsis exists as an option.

I understand that this might be just me, but it's highly annoying. :)

BTW, del(...) causes ValueError: expression which can't be assigned to in Del context

scopatz:
This should be as simple as removing the `... ` case from completer.py

ivh:
You mean from the XONSH_TOKENS list? This seems to remove ..., yes, but there seem to be two .. so the slash still does not get appended.

ivh:
... and when I remove .. from that list as well, it *does* get completed, but with a blank instead of a slash.

(I won't have time to look at this today, just thought I'd get it out there.)

scopatz:
OK @ivh - well whenever you get a chance, unless some one beats you to it. Just wanted to let you know that I am very open to this change.

scopatz:
This change has been implemented at some point.  Feel free to reopen if it hasn't been.

ivh:
I just tried (with master) and .. does not get its slash tab-appended, same as before. Still for the same reasons, I think, there is the ellipsis and a second instance of ..

wrywerytwreywery:
I am planning on working a little bit on tab competion today (to try to deal with names with spaces more appropriately), so I can try to figure out this problem while I am working on that (unless you want to go for it, @ivh).

ivh:
Please do @wrywerytwreywery, I don't have time for this soon. :)

PR #184, opened 2015-04-03
Completer Updates

wrywerytwreywery:
This is an attempt to deal with the issue from #101, as well as some lingering issues from #68 (and now #181).  The completer now tries a little harder to decide whether to complete only paths, only Python, or both.

One downside is that `Cmd` seems to ignore the order in which the completions are given (I tried to provide a more sane ordering, but it was ignored).  Though, looking at the source for `Cmd`, I'm not sure where that is happening.  For example, reversing the order of the returned lists caused the return value of `xonsh.Shell.completedefault` to be reversed, but hitting tab a second time to show all completions leads to the options being presented in the normal (un-reversed) order.  So as of right now, strings come first, then `..`, then `.`, then capitalized names, then lower-case names.  This is far from ideal, but I don't know a way around it as of right now.

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -69,30 +69,47 @@ class Completer(object):
         """
         space = ' '  # intern some strings for faster appending
         slash = '/'
+        ctx = ctx or {}
+        cmd = line.split(' ', 1)[0]
         if begidx == 0:
+            # the first thing we're typing; could be python or subprocess, so
+            # anything goes.
             rtn = self.cmd_complete(prefix)
-        elif line.split(' ', 1)[0] in self.bash_complete_funcs:
+        elif cmd in self.bash_complete_funcs:
             rtn = set()
             for s in self.bash_complete(prefix, line, begidx, endidx):
                 if os.path.isdir(s.rstrip()):
                     s = s.rstrip() + slash
                 rtn.add(s)
-            if len(rtn) == 0:
+            if len(rtn) != 0:
                 rtn = self.path_complete(prefix)
             return sorted(rtn)
+        elif cmd not in ctx and cmd not in XONSH_TOKENS:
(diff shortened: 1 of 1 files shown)


wrywerytwreywery:
I still don't know of a way to solve the sorting problem, but I am still going to go back and try to implement a couple more things for dealing with names with spaces, so this probably isn't quite ready yet.

scopatz:
OK!

wrywerytwreywery:
Ah, never mind.  I think what is here is ready for a look.

wrywerytwreywery:
What I had been hoping to fix was, e.g., `"name with space<TAB>`, which will currently try to complete based only on the last word, rather than on the whole quoted string, but I'm not sure how to do that.

scopatz:
Ahh, that might require tokenizing the string to figure out if the cursor is in a string or not.

scopatz:
Hmm this seems to break Bash completion for me. I have tried it with git, valgrind, etc.

wrywerytwreywery:
Hmm.  Well I have never had bash completion working yet, so maybe this is my excuse to get that working so that I can try to reproduce.

wrywerytwreywery:
Okay; that was a really silly mistake.  Can you try again with what is there now?

> Ahh, that might require tokenizing the string to figure out if the cursor is in a string or not.

Yes, I think so.  The question is what the limitations of `Completer` are; in particular, I'm not sure whether it is possible to have it make suggestions for pieces longer than what it thinks the last token is (the input called `prefix`).  I am happy to play around with it some more, but that might be a separate PR.  This one's main focus is on changing the situations in which xonsh uses `path_complete`.

scopatz:
Works for me! Thanks @wrywerytwreywery!

PR #186, opened 2015-04-04
Improved Command Suggestions

wrywerytwreywery:
This is not a big change, but in my experience, it leads xonsh to give better suggestions to incorrect commands.  The change is essentially to ignore case when deciding what commands to suggest, and also when deciding how to order the suggestions.


scopatz:
Great as always!

PR #189, opened 2015-04-05
More IO Redirection Options

wrywerytwreywery:
This changeset implements additional redirect capabilities (redirect stderr to file with `2>` and `2>>`, redirect both stdout and sterr to file with `&>` and `&>>`, grab stdin from file with `<`).  Not as fully-featured as bash's redirects, but it is something.

I also updated many of the error messages from xonsh to be printed to stderr instead of stdout.

Also fixes an unrelated bug with detyping `FORMATTER_DICT` (which I encountered along the way, and which we should fix even if this update is rejected) in built_ins.py, line 72.

The code is not very pretty right now (and so is not yet ready to be merged, I think, until I make another formatting pass), but it seems to be functional.  I'd be glad for any feedback if you have a chance to try it out, @scopatz.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -69,7 +69,7 @@ class Env(MutableMapping):
             return self._detyped
         ctx = {}
         for key, val in self._d.items():
-            if callable(val):
+            if callable(val) or isinstance(val, (MutableMapping)):
                 continue
             if not isinstance(key, string_types):
                 key = str(key)
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -572,7 +572,8 @@ def run_subproc(cmds, captured=True):
                                               captured=captured,
                                               prev_proc=prev_proc,
                                               default_stdin=stdin,
-                                              stdout=stdout)
+                                              stdout=stdout,
+                                              stderr=stderr)
             continue
         subproc_kwargs = {}
         if os.name == 'posix':
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -422,6 +422,29 @@ def _open(fname, mode):
         print(e.format(fname), file=sys.stderr)
 
 
+def _pipe_check(cmds):
+    error = None
+    hitpipe = False
+    inredir = False
+    outredir = False
+    for i in cmds:
+        if not isinstance(i, str):
+            continue
+        if i == '|':
+            hitpipe = True
+            if inredir or outredir:
+                error = "pipe cannot be used after another redirect"
+                break
+        if i.endswith('>'):
+            outredir = True
+        if i.endswith('<'):
+            inredir = True
+            if hitpipe:
+                error = "cannot receive input both from file and pipe"
+                break
+    return error
+
+
 def run_subproc(cmds, captured=True):
     """Runs a subprocess, in its many forms. This takes a list of 'commands,'
     which may be a list of command line arguments or a string, representing
@@ -436,6 +459,12 @@ def run_subproc(cmds, captured=True):
     Lastly, the captured argument affects only the last real command.
     """
     global ENV
+    # check for misused pipes
+    err = _pipe_check(cmds)
+    if err is not None:
+        print('xonsh: error: {}'.format(err), file=sys.stderr)
+        return
+
     background = False
     if cmds[-1] == '&':
         background = True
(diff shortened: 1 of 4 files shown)


scopatz:
Hi @wrywerytwreywery!  Thanks for putting this in.  Before you go too much further (and before I fall asleep) I wanted to throw some of my thoughts out there, that will hopefully make the tokenizer simpler and the syntax more intuitive.

First off, I really like [fish's use](http://fishshell.com/docs/current/)  of `^` to represent stderr.  I am not such a fan that they kept 0, 1, 2 as file descriptors for stdin, stdout, and stderr. But oh well.  So basically, what I propose is the following:

* stdin redirection with `<` (✓)
* stdout redirection, 'w' mode with `>` (✓)
* stdout redirection, 'a' mode with `>>` (✓)
* stderr redirection, 'w' mode with `^` 
* stderr redirection, 'a' mode with `^^` 
* stderr and stdout redirection, 'w' mode with `^>` 
* stderr and stdout redirection, 'w' mode with `^^>>` 

This seems the most consistent to me. How do you feel about it?

wrywerytwreywery:
I have thought about this a little bit over the last day or so.

My personal preference would be to offer both syntaxes.  I think it is nice to have the above as an alternative syntax, but I think the more general redirections (not yet implemented here) that fish and bash offer are still useful (really, I would like to be able to redirect arbitrary file descriptors to one another, _and_ to redirect from/to open file-like objects in the Python context as well).  So I wouldn't want to give those up, though I think what you've described above is quite nice as an alternative/additional syntax.

scopatz:
So your concern is redirecting stderr to stdout, etc, is that right?

wrywerytwreywery:
> So your concern is redirecting stderr to stdout, etc, is that right?

Yes, but also sending things to arbitrary file descriptors, sending both stderr/stdout to pipes, etc could hold some appeal.  That said, the cases above probably do account for 99% of what people use redirection for, so maybe it's not worth worrying about things beyond those...

scopatz:
I completely agree that there are times where we want to support merged streams, probably beyond just dumping to a file.  For example:

```console
# command with a lot of output that I know fails
$ cmd 2&>1 | grep "error"
```

ie the piping you mention :). I think we should definitely support that!  Most of my objections come from using 0, 1, 2 for file descriptions.  My claim is that if they are so special, then they should get a syntax other than magic numbers. I think that <, ^, and > should also act as the descriptors themselves.  That was all.  

I am open to syntax suggestions for merging file descriptors.  However, here are some options I would be happy with:

```console
^&>  # merge stderr into stdout
>&^  # merge stdout into stderr
^|>  # merge stderr into stdout
>|^  # merge stdout into stderr
^>  # merge stderr into stdout, perhaps this is too confusing with the redirection operator.
>^  # merge stdout into stderr
```

Again open to ideas!

scopatz:
Also, sorry, I totally spaced this use case before!

wrywerytwreywery:
I am still trying to think through the syntax here.  I think my issue is that `>` doesn't _just_ mean stdout in those examples; it means "send stdout somewhere" or "stdout," depending on context.  Likewise with `^`.  This is a linguistic issue in some sense, as the meaning of `>` in the examples above changes depending on the context in which it is used.  

This is why `2>&1`, despite being arcane, works for me.  `>` doesn't mean stdout; it means "send something somewhere."

So I think I like keeping `>` with that meaning makes sense (it's shape also suggests this).  Then the question is, what do we call stdout and stderr?  In some sense, I still think using the file descriptors makes sense.  But that is hard for a novice to approach.  It could be `o` and `e` or `out` and `err` (these are, in some sense, like Python using `or` and `and` instead of `&&` and `||`).

`echo "hey" > log.txt`
`echo "hey" o> log.txt`
`echo "hey" out> log.txt`
`echo "hey" e> log.txt`
`echo "hey" err> log.txt`

`echo "hey" e>o | vim -`
`echo "hey" err>out | vim -`

`echo "hey" err>> error.txt`
`echo "hey" e>> error.txt`

This isn't perfect.  It's a little verbose, and `err>>` blends in with the other arguments in a way that `^^` doesn't (it could be mistaken for a filename).  But I think those issues are preferable to the syntax with `^` and `>` because they seem to have more general rules, as opposed to sequences of characters you have to memorize.

My current proposal would be to support the syntax I described here (`e` or `err` means stderr, and `o` or `out` means stdout, `>` always means write to, `>>` always means append to).  

I might also argue for supporting bash's conventions in addition, to make transitioning easier (and because I do think the generality of using file descriptors could be nice).

scopatz:
Hi @wrywerytwreywery - I like this proposal.  out and err fit my brain in a nice way, (o and e are OK, but probably too short). You are right that my previous statements built up an ambiguity between redirect actions (verb) and location (noun).  Avoiding that is smart.

I also agree that file descriptors are probably too strong of a convention to completely ignore for the time being.  So adding them into the language makes a lot of sense for migration.

scopatz:
I say that we should go with your proposal.

wrywerytwreywery:
Alright, that sounds good to me.  Hopefully I'll have some time in the next day or so to implement!

scopatz:
Awesome!

scopatz:
This looks like it needs a merge from master and some conflicts to be resolved.

wrywerytwreywery:
Yes.  I was thinking about this just today (so much for "some time in the next day or so"...).  Hopefully I'll have some time in the near future to circle back around to this; sorry for being so slow!

scopatz:
No worries!

gthole:
I'd like to weigh in here and say that maintaining BASHwards compatibility is super helpful on stdin/stderr redirection.  Some standard `vim` plugins (like syntastic) fail sadly without traditional UNIX-style redirection syntax, so while `err>` and `out>` are super-cool, keeping the old-style `2>` `1>` around is important too.

That said, I'm really loving `xonsh`!  Finally basic arithmetic in the shell. :smiley:

scopatz:
Thanks for the feedback @gthole! That is a pretty compelling case to make for keeping `2>` and `1>`.  Though you seem to indicate that maybe having them as a compatibility would be good enough. That is, also having `err>` and `out>` around too.

gthole:
@scopatz You got it spot on - I'd probably use `err>` in my own daily shell use over `2>`, for example, but I like the idea that `xonsh` will "just work" for established BASH syntax.  (And maybe `all>` for `&>`?)

wrywerytwreywery:
I'm working on this today.  Given all the big-time changes since I originally put this up, I figured it's easier to start fresh.

scopatz:
Sounds good :)

PR #190, opened 2015-04-05
Attempt to fix #188

### LINKED ISSUE #188: Parser Error
The following runs fine in xonsh:
`$ bash -c "echo 'hey'; ls"`

As does the following:
`$ bash -c "echo 'hey'" > test.out`

But the following fails:
```
$ bash -c "echo 'hey'; ls" > test.out
................ 
  File "/usr/local/lib/python3.4/site-packages/xonsh/execer.py", line 135, in _parse_ctx_free
    debug_level=self.debug_level)
  File "/usr/local/lib/python3.4/site-packages/xonsh/parser.py", line 258, in parse
    tree = self.parser.parse(input=s, lexer=self.lexer, debug=debug_level)
  File "/usr/local/lib/python3.4/site-packages/ply/yacc.py", line 265, in parse
    return self.parseopt_notrack(input,lexer,debug,tracking,tokenfunc)
  File "/usr/local/lib/python3.4/site-packages/ply/yacc.py", line 1047, in parseopt_notrack
    tok = self.errorfunc(errtoken)
  File "/usr/local/lib/python3.4/site-packages/xonsh/parser.py", line 2357, in p_error
    column=p.lexpos))
  File "/usr/local/lib/python3.4/site-packages/xonsh/parser.py", line 329, in _parse_error
    raise err
  File "<string>", line None
SyntaxError: <xonsh-code>:1:11: ('code: "echo \'hey\'; ls"',)

  File "/usr/local/lib/python3.4/site-packages/xonsh/shell.py", line 149, in push
    locs=self.ctx)
  File "/usr/local/lib/python3.4/site-packages/xonsh/execer.py", line 92, in compile
  File "/usr/local/lib/python3.4/site-packages/xonsh/execer.py", line 70, in parse
    tree = self._parse_ctx_free(input, mode=mode)
  File "/usr/local/lib/python3.4/site-packages/xonsh/execer.py", line 148, in _parse_ctx_free
    raise original_error
  File "/usr/local/lib/python3.4/site-packages/xonsh/execer.py", line 135, in _parse_ctx_free
    debug_level=self.debug_level)
  File "/usr/local/lib/python3.4/site-packages/xonsh/parser.py", line 258, in parse
    tree = self.parser.parse(input=s, lexer=self.lexer, debug=debug_level)
  File "/usr/local/lib/python3.4/site-packages/ply/yacc.py", line 265, in parse
    return self.parseopt_notrack(input,lexer,debug,tracking,tokenfunc)
  File "/usr/local/lib/python3.4/site-packages/ply/yacc.py", line 1047, in parseopt_notrack
    tok = self.errorfunc(errtoken)
  File "/usr/local/lib/python3.4/site-packages/xonsh/parser.py", line 2357, in p_error
    column=p.lexpos))
  File "/usr/local/lib/python3.4/site-packages/xonsh/parser.py", line 329, in _parse_error
    raise err
  File "<string>", line None
SyntaxError: <xonsh-code>:1:8: ('code: "echo \'hey\'; ls"',)
```

It seems like the combination of `;` and `>` that confuses the parser (even though `;` is inside a string?!)...

wrywerytwreywery:
This may not be the most elegant solution, but I think it is robust in that it should only consider "standalone" semicolons as terminators for commands (should fix #188).

### DIFF
--- a/xonsh/execer.py
+++ b/xonsh/execer.py
@@ -162,11 +162,9 @@ class Execer(object):
                     last_error_line = last_error_col = -1
                     input = '\n'.join(lines)
                     continue
-                maxcol = line.find(';', last_error_col)
-                maxcol = None if maxcol < 0 else maxcol + 1
                 sbpline = subproc_toks(line,
                                        returnline=True,
-                                       maxcol=maxcol,
+                                       maxcol=None,
                                        lexer=self.parser.lexer)
                 if sbpline is None:
                     # subprocess line had no valid tokens, likely because
--- a/xonsh/execer.py
+++ b/xonsh/execer.py
@@ -162,9 +162,19 @@ class Execer(object):
                     last_error_line = last_error_col = -1
                     input = '\n'.join(lines)
                     continue
+                maxcol = None
+                self.parser.lexer.input(line)
+                for tok in self.parser.lexer:
+                    if tok.lexpos < last_error_col:
+                        continue
+                    if tok.type == 'SEMI':
+                        maxcol = tok.lexpos
+                        break
+                if maxcol is not None:
+                    maxcol += 1
                 sbpline = subproc_toks(line,
                                        returnline=True,
-                                       maxcol=None,
+                                       maxcol=maxcol,
                                        lexer=self.parser.lexer)
                 if sbpline is None:
                     # subprocess line had no valid tokens, likely because
--- a/xonsh/execer.py
+++ b/xonsh/execer.py
@@ -132,10 +132,8 @@ class Execer(object):
         self.parser.lexer.input(line)
         for tok in self.parser.lexer:
             if tok.type == 'SEMI':
-                maxcol = tok.lexpos
+                maxcol = tok.lexpos + mincol + 1
                 break
-        if maxcol is not None:
-            maxcol += mincol + 1
         return maxcol
 
     def _parse_ctx_free(self, input, mode='exec'):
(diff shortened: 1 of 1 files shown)


scopatz:
It seems like there has to be a way to do this in `subproc_toks()` itself, rather than tokenizing twice. I am playing around with it now....

scopatz:
Ok - After fiddling with this some more, there isn't a nice way of doing this that doesn't involve retokenization.  So this approach is probably the correct one.  That said, I have a couple of requests.

The first is that if you could make this a separate (private?) method, maybe `_find_next_break()`(?) that would be great.

The second is that it would be nice if this new method first checked and then returned if there wasn't a semicolon in the line at all. This would short-circut the tokenization.

The third, and perhaps optional, one is that you don't have to tokenize the line from the begining.  You should be able to start from mincol (`last_error_col`) and instead return `tok.lexpos + mincol` rather than just returning `tok.lexpos`.  This should be faster than tokenizing everything.

Thanks @wrywerytwreywery!

wrywerytwreywery:
I agree with all the suggestions above  (not sure that we'll see any noticeable speedup by tokenizing less, but the code is certainly cleaner, and it makes sense to avoid wasting those cycles anyway).  Just pushed an updated version.

>  there isn't a nice way of doing this that doesn't involve retokenization.

Yeah; I tried to work with this in `subproc_toks`, too, but couldn't get anything to go...

scopatz:
Thanks @wrywerytwreywery! This looks great!

PR #191, opened 2015-04-05
prevents detyping mappings as string to env.

scopatz:
I found this with respect to running xonsh in bash in xonsh with the new `$FORMATTER_DICT`


scopatz:
Whoops! didn't realize that this was in #189

wrywerytwreywery:
Makes sense for it to be its own separate issue, though, especially since #189 will take some time.  Merging.

PR #192, opened 2015-04-05
Attribute Completion

scopatz:
Should work on everything that is `dir()`-able in xonsh, including environment variables, subprocs!

I was getting pretty annoyed at not having this :)

### DIFF
--- a/tests/test_builtins.py
+++ b/tests/test_builtins.py
@@ -97,5 +97,6 @@ def test_ensure_list_of_strs():
         obs = ensure_list_of_strs(inp)
         yield assert_equal, exp, obs
 
+
 if __name__ == '__main__':
     nose.runmodule()
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -5,7 +5,7 @@ import nose
 from nose.tools import assert_equal
 
 from xonsh.lexer import Lexer
-from xonsh.tools import subproc_toks
+from xonsh.tools import subproc_toks, subexpr_from_unbalanced
 
 LEXER = Lexer()
 LEXER.build()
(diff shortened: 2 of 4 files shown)


wrywerytwreywery:
This is fantastic!

One small issue (non-blocking) is that this offers completions for integers, such as `7.imag`, which is not valid Python.

Also, I made one small change in my local copy, a change to the condition under which we attempt python completion.  Line 88 of completer.py (from one of my earlier PRs) currently tests:
`elif cmd not in ctx and cmd not in XONSH_TOKENS:`

I modified the test to be:
`elif cmd not in ctx and cmd in allcmds:`

where `allcmds` is a set containing all commands in the path and all aliases.  If we build up that structure once at the start of `complete`, then we could also pass it to `cmd_complete` to avoid doubling up on work.

```
        allcmds = set()
        for d in path:
            if os.path.isdir(d):
                allcmds |= set(os.listdir(d))
        allcmds |= set(builtins.aliases.keys())
```

This allowed completions for, for example:
`x = "cat".<TAB>`
and
`{"cat": "dog".<TAB>}`

Does that change seem reasonable?  If so, should it come here, or in a separate PR?

scopatz:
> One small issue (non-blocking) is that this offers completions for integers, such as 7.imag, which is not valid Python.

Yeah.... Oddly, this is a tab completion that IPython allows (I don't think that there is a nice way around it).  Even more annoying is that IPython will not complete `(7).imag`, which is valid Python.  Luckily, xonsh will complete `(7).imag`.  I figured if `7.imag` was OK for IPython then it would be OK here too.  At least for now.

>  If so, should it come here, or in a separate PR?

Seems OK.  But I think a PR would help :)

wrywerytwreywery:
Sounds good!  Just merged.  I will submit a PR for the other change.

scopatz:
Sounds great.

PR #193, opened 2015-04-06
Completer Bugfix

wrywerytwreywery:
This fixes an issue with #192 by causing invalid commands actually to cause an error to be thrown, including a modification to `Shell` to report those errors in a nicer way.


scopatz:
Thanks @wrywerytwreywery!

PR #194, opened 2015-04-06
Completer Update

wrywerytwreywery:
Modify the completer by changing when it calls `path_complete` as discussed in #192.

This change makes it so that `attr_complete` is called in the following cases, rather than `path_complete`:

`x = "cat".<TAB>`
and
`{"cat": "dog".<TAB>`

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -5,8 +5,8 @@ import re
 import builtins
 import subprocess
 
-from xonsh.tools import subexpr_from_unbalanced
 from xonsh.built_ins import iglobpath
+from xonsh.tools import subexpr_from_unbalanced, all_command_names
 
 RE_DASHF = re.compile(r'-F\s+(\w+)')
 RE_ATTR = re.compile(r'(\S+(\..+)*)\.(\w*)$')
@@ -72,6 +72,7 @@ class Completer(object):
         dot = '.'
         ctx = ctx or {}
         cmd = line.split(' ', 1)[0]
+        allcmds = all_command_names()
         if begidx == 0:
             # the first thing we're typing; could be python or subprocess, so
             # anything goes.
@@ -85,11 +86,11 @@ class Completer(object):
             if len(rtn) == 0:
                 rtn = self.path_complete(prefix)
             return sorted(rtn)
-        elif cmd not in ctx and cmd not in XONSH_TOKENS:
+        elif cmd not in ctx and cmd in allcmds:
             # subproc mode; do path completions
             return sorted(self.path_complete(prefix))
         else:
-            # if we're here, we're definitely python?
+            # if we're here, we're not a command, but could be anything else
             rtn = set()
         rtn |= {s for s in XONSH_TOKENS if s.startswith(prefix)}
         if ctx is not None:
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -271,13 +271,18 @@ class Completer(object):
         al_hash = hash(tuple(builtins.aliases.keys()))
         self._alias_checksum = al_hash
         cache_valid = cache_valid and al_hash == self._alias_checksum
+        pm = self._path_mtime
         if cache_valid:
             for d in filter(os.path.isdir, path):
                 m = os.stat(d).st_mtime
-                if m > self._path_mtime:
-                    self._path_mtime = m
+                if m > pm:
+                    pm = m
                     cache_valid = False
-                    break
+                    # run through everything the first time to get the highest
+                    # mtime for the cache, to avoid rebuilding multiple times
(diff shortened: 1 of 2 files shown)


scopatz:
Hi @wrywerytwreywery - so my only concern with this is that `all_command_names()` is called every time a completion is requested.  This seems expensive since (1) it hits a lot of places on the disk, and (2) it isn't likely to change much from TAB to TAB.  

There are a few of different ways to address this.  Here is what I could think of off the time of my head:

1. call `all_command_names()`, only inside the if- and elif-blocks.  This would mean the `elif cmd not in ctx and cmd in allcmds` condition would have to change somehow.
2. cache the results of the `all_command_names()` and only recompute if the contents of PATH have changed (probably should check via hashing the value).  It might be easiest/nicest to implement this via an `allcmds` property on the Completer.

scopatz:
Let me know if any of that doesn't make sense.

wrywerytwreywery:
Yeah, I usually only start caching/optimizing when I see a need, and I hadn't noticed it being sluggish at all on my machine.  But you're right that this could be slow.

Anyway, here is an updated version that tries to take into account both of the above.  It only calls `_all_commands` inside of the relevant branches, and it implements a cache.  The cache is more complicated than I would have liked, but I think it should keep track of everything necessary.  The cache gets invalidated if:
* `$PATH` has changed
* The keys of `builtins.aliases` have changed
* The contents (in terms of names) of any of the directories in `$PATH` have changed.

This means that in the case where nothing has changed, we are only moderately faster (I am assuming that computing `mtime` is faster than looping over the result of `listdir` almost always), and rebuilding the cache is a little bit slower than before (one pass to get `mtimes`, then the normal pass to build the set).

Maybe the caching isn't necessary given that we're only calling `_all_commands` when necessary, but I think it should still speed things up a bit in that case.

scopatz:
mtime seems much faster :)

```python
In [3]: %timeit os.stat('.').st_mtime
1000000 loops, best of 3: 1.17 µs per loop

In [4]: %timeit os.listdir('.')
10000 loops, best of 3: 73 µs per loop

In [5]: %timeit [d for d in os.listdir('.')]
10000 loops, best of 3: 75.9 µs per loop
```

xonsh/completer.py
```diff
@@ -257,3 +262,31 @@ def attr_complete(self, prefix, ctx):
...
         return attrs
+
+    def _all_commands(self):
+        path = builtins.__xonsh_env__.get('PATH', None) or []
+        # did PATH change?
+        path_hash = hash(tuple(path))
+        cache_valid = path_hash == self._path_checksum
+        self._path_checksum = path_hash
+        # did aliases change?
+        al_hash = hash(tuple(builtins.aliases.keys()))
```
scopatz: I think that it would be more correct to have `hash(tuple(sorted(builtins.aliases.keys())))` here since key ordering is not guaranteed.  However, the worst that is going to happen is that it is going to trigger a cache miss.

xonsh/completer.py
```diff
@@ -257,3 +262,31 @@ def attr_complete(self, prefix, ctx):
...
+        self._alias_checksum = al_hash
+        cache_valid = cache_valid and al_hash == self._alias_checksum
+        pm = self._path_mtime
+        # did the contents of any directory in PATH change?
+        for d in filter(os.path.isdir, path):
+            m = os.stat(d).st_mtime
+            if m > pm:
+                pm = m
+                cache_valid = False
```
scopatz: you can actually break here.

scopatz: Oh nm, that is bad. sorry.

xonsh/completer.py
```diff
@@ -257,3 +262,31 @@ def attr_complete(self, prefix, ctx):
...
+        self._path_mtime = pm
+        if cache_valid:
+            return self._cmds_cache
+        allcmds = set()
+        for d in path:
```
scopatz: you can use filter here too, non-blocking.

scopatz:
A couple of minor suggestions in-line.  Otherwise looks great!

xonsh/completer.py
wrywerytwreywery: Right.  I went back and forth on whether to enter this loop if the cache had already been invalidated, and on whether to break here or not.  I think where I settled on those things was right.  This way, even if the cache is invalidated, we make sure we have up-to-date info in all of the cached things (path hash, alias hash, max path mtime).

xonsh/completer.py
wrywerytwreywery: Good call.  Will change.

wrywerytwreywery:
Alright, should be ready for another look!

scopatz:
Looks great! Thanks!

PR #195, opened 2015-04-06
I see a string and I want to paint it black

scopatz:
It took a lot of staring to find this one.


wrywerytwreywery:
Nice catch!  Just merged in.

Issue #198, opened 2015-04-06
Slow initialization of xonsh


SavinaRoja:
I started out writing this issue rather more clueless than I am now that I am finishing it... Basically it seems that the start-up time for PLY might be pretty punitive on weaker processors as found on the Raspberry Pi when it comes to scripting. This issue may or may not be a bit complicated to address.

I installed xonsh on a little Raspberry Pi 2 Model B I have for personal file serving so I could tinker with writing xonsh shell scripts (and hopefully replace some active scripts at some point) and noticed that the startup time for xonsh was pretty slow.

Here's some time benchmarks from my desktop workstation and from my raspberry pi, collected using `time bash -c true; time xonsh -c true; time python -c "pass"`:

```
Desktop:
  Bash:               Xonsh:              Python:
    real    0m0.005s    real    0m1.271s    real    0m0.025s
    user    0m0.000s    user    0m1.160s    user    0m0.016s
    sys     0m0.004s    sys     0m0.060s    sys     0m0.004s
-------------------------------------------------------------
Raspberry Pi 2 Model B:
  Bash:               Xonsh:               Python:
    real    0m0.025s    real    0m13.894s    real    0m0.231s
    user    0m0.010s    user    0m12.840s    user    0m0.170s
    sys     0m0.010s    sys     0m0.290s     sys     0m0.030s
```

After some quick testing and I observed that `xonsh -h` was much faster (and in line with my general expectations):

```
real    0m1.881s
user    0m1.750s
sys     0m0.110s
```

With this hint and some some fiddling, I observed that the great majority of the time was being spent during the call to initialize `Shell`.

I then did a basic cProfile of `xonsh.shell.Shell()` on both the desktop and the raspberry pi, the results of which are detailed here: https://gist.github.com/SavinaRoja/1f9051193376e5374aa9

This seems to show that the main chunk of time is being spent initializing the PLY Parser. I am not sure what can be done to improve this. Perhaps there might be an optional installation route for Xonsh to utilize the Lex and Yacc tools natively?

scopatz:
Fascinating: here are my timings.   

```
  Bash:               Xonsh:              Python:
    real    0m0.003s    real    0m0.291s    real    0m0.019s
    user    0m0.003s    user    0m0.291s    user    0m0.019s
    sys     0m0.000s    sys     0m0.063s    sys     0m0.000s
```

[...]

PR #199, opened 2015-04-07
wrap develop too

scopatz:
related to #198


wrywerytwreywery:
I'm not exactly sure how this fixes the problem, but given the thread from earlier, it does seem to fix it.  Merging in.

Issue #84, opened 2015-03-17
Support for the exec builtin


vitaut:
It would be useful if xonsh supported the [exec Bourne shell builtin](http://www.gnu.org/software/bash/manual/bashref.html#index-exec) as some programs (such as Qt Creator) rely on it which requires using a different shell for them.

Currently trying to use `exec` fails with `FileNotFoundError`:

```
$ exec bash
Traceback (most recent call last):
  File "/home/viz/Programs/xonsh/lib/python3.4/site-packages/xonsh/shell.py", line 109, in default
    self.execer.exec(code, mode='single', glbs=None, locs=self.ctx)
  File "/home/viz/Programs/xonsh/lib/python3.4/site-packages/xonsh/execer.py", line 110, in exec
    return exec(code, glbs, locs)
  File "<xonsh-code>", line 1, in <module>
  File "/home/viz/Programs/xonsh/lib/python3.4/site-packages/xonsh/built_ins.py", line 383, in subproc_uncaptured
    return run_subproc(cmds, captured=False)
  File "/home/viz/Programs/xonsh/lib/python3.4/site-packages/xonsh/built_ins.py", line 355, in run_subproc
    stdin=stdin, stdout=stdout)
  File "/usr/lib/python3.4/subprocess.py", line 848, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.4/subprocess.py", line 1446, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: 'exec'
```

scopatz:
This wouldn't be significantly different from the source-bash alias that lives in aliases.py. Pull requests welcome!

scopatz:
Hi @vitaut - I believe that this has now been added to master with #200 via the xexec builtin.  Would you mind verifying this?

vitaut:
AFACS the `xexec` builtin works fine. Is there a way to make `xexec` override Python's `exec` for systems that expect a shell with the `exec` builtin?

scopatz:
@vitaut yes, but it shouldn't be enabled by default.  The way to do this is to stick the following in xonshrc or elsewhere:

```python
aliases['exec'] = aliases['xexec']
```

This will override the Python built-in `exec()` function.

vitaut:
Fair enough, I'll add it to `.xonshrc`.

There is one issue with passing absolute paths to `exec` though:

```
$ exec /bin/sh
[...]

PR #200, opened 2015-04-08
Added exec builtin support

aig787:


xonsh/aliases.py
```diff
@@ -73,6 +73,19 @@ def source_bash(args, stdin=None):
...
 
+def exec(args, stdin=None):
+    """
+    Replaces current process with command specified and passes in the
+    current xonsh environment.
+    """
+    env = builtins.__xonsh_env__
+    denv = env.detype()
+    try:
+        os.execvpe(args[0], args[1:len(args)], denv)
```
scopatz: You are doing it this way for the process replacement, rather than using subprocess, right?

xonsh/aliases.py
```diff
@@ -73,6 +73,19 @@ def source_bash(args, stdin=None):
     return
 
 
+def exec(args, stdin=None):
```
scopatz: This name has to be something different because it is overriding the Python `exec()` builtin.  This isn't a bash exec, either.  Note that `execx()` is taken already by the xonsh builtin function.  I am not sure that this is *that* much different than `execx()` in spirit, either.  Maybe this should be a thin wrapper around that.

xonsh/aliases.py
```diff
@@ -108,6 +121,7 @@ def bash_aliases():
     'EOF': exit,
     'exit': exit,
     'quit': exit,
+    'exec': exec,
```
scopatz: Same naming issues here.

scopatz:
Thanks for putting this in @aig787!  There is a fairly serious name collision, but other than that it looks fine.

xonsh/aliases.py
aig787: Correct.

xonsh/aliases.py
aig787: It sounds reasonable to make it a wrapper around execx. I thought about the naming conflict but left it consistent because I wanted to hear your thoughts what it should be called and whether having the exec keyword taken will shadow the Python `exec()` builtin when using the shell as well.

scopatz: If you are just looking for a new name, `xexec` and `exec-xonsh` are both free.  The more that I think about it, the more that a new (alias) name would be needed, even if this was a thin wrapper around execx.

aig787: If a new alias was added would it still solve #84, or would users just need to manage that themselves (e.g. exec aliased to xexec in xonshrc)

scopatz: Yes it would.  It would work the same way that the `source-bash` alias works now.  Officially we support a generic name that doesn't collide with anything.  However, people can override other names (if they want to) in their xonshrc by renaming their aliases.

aig787: Okay, that makes sense. I'll change the name to xexec and make it a wrapper around execx.

aig787:
I went ahead and changed the name, but I hit a wall at making it a wrapper around `execx()`. For it to function like the bash/sh built in we need the running process to be replaced with the new command but a new instance of python won't be able to interpret the code object `execx()` produces.

scopatz:
I think keeping the implementation like you had it before, will be fine

aig787:
Okay, I'll leave it as is then. It should be ready to merge unless there are any other changes you'd make.

scopatz:
Hi @aig787 - sorry it took me so long to get back to this.  PyCon happened :)  In any event, this can cause xonsh to crash (see below) so I think that some safety should be put in.

```console
scopatz@athenaie ~ BA$H xonsh
scopatz@athenaie ~ $ xexec ls
A NULL argv[0] was passed through an exec system call.
Aborted (core dumped)
scopatz@athenaie ~ BA$H 
```

Also, could you add some documentation to about this alias, please?  I think we are long overdue for an aliases.rst file.  If you could start one add just xexec, that would be awesome! Thanks a million.

aig787:
Okay that should be fixed, I misunderstood what it was wanting as arguments to `os.execvpe`. I added a check to avoid crashes like that in the future too.

scopatz:
Hi @aig787, thanks for the docs.  I think that this needs a little more work though, since args can be length-0.

```python
scopatz@athenaie ~ BA$H xonsh
scopatz@athenaie ~ $ xexec
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/shell.py", line 131, in default
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/execer.py", line 124, in exec
    return exec(code, glbs, locs)
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/built_ins.py", line 543, in subproc_uncaptured
    return run_subproc(cmds, captured=False)
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/built_ins.py", line 473, in run_subproc
    stdout=stdout)
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/built_ins.py", line 346, in _run_callable_subproc
    rtn = alias(args, stdin=stdin)
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/aliases.py", line 85, in xexec
    if args[0]:
```

aig787:
@scopatz, yeah that completely flew under my radar. Fixed.

xonsh/aliases.py
```diff
@@ -73,6 +74,20 @@ def source_bash(args, stdin=None):
...
+    Replaces current process with command specified and passes in the
+    current xonsh environment.
+    """
+    env = builtins.__xonsh_env__
+    denv = env.detype()
+    if (len(args) > 0):
+        try:
+            os.execvpe(args[0], args, denv)
+        except FileNotFoundError as e:
+            return "xonsh: " + e.args[1] + ": " + args[0] + "\n"
```
scopatz: It would be nice if there was another string returned that indicated an error if this there were no args.

aig787: Any preference on what that should be? Otherwise I'll just make it `xonsh: exec: no args specified"

scopatz: Sounds good to me :)

aig787:
Okay that should do it :)

scopatz:
Thanks a ton @aig787!

PR #209, opened 2015-04-16
cd Changes

### LINKED ISSUE #207: crash when cd from deleted dir
bash handles this through some magic that we should probably implement too.

wrywerytwreywery:
This changeset includes the following:

* implements `AUTO_PUSHD` as an environment variable, to have `cd` automatically push directories to the directory stack
* implements the behavior described in #48, where `cd -<integer>` changes to a previous directory (though as of right now, this operates on the directory stack rather than on some separate history, so `AUTO_PUSHD` must be set to `True` to get the behavior that most people probably want).
* makes `cd` prefer changing to actual directories where there is ambiguity (e.g., running `cd -` when there is a directory called `-` should change to that directory, rather than to `OLDPWD`)
* moves `cd` to the `dirstack` module
* fixes #207

### DIFF
--- a/xonsh/dirstack.py
+++ b/xonsh/dirstack.py
@@ -140,15 +140,11 @@ def pushd(args, stdin=None):
             e = 'Invalid argument to pushd: {0}\n'
             return None, e.format(args.dir)
     if new_pwd is not None:
-        e = None
         if args.cd:
             DIRSTACK.insert(0, os.path.expanduser(pwd))
-            _, e = builtins.default_aliases['cd']([new_pwd], None)
+            _change_working_directory(os.path.abspath(new_pwd))
         else:
             DIRSTACK.insert(0, os.path.expanduser(os.path.abspath(new_pwd)))
-
-        if e is not None:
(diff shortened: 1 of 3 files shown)


xonsh/dirstack.py
```diff
@@ -10,6 +10,75 @@
...
+    except FileNotFoundError:
+        return None
+
+
+def _change_working_directory(newdir):
+    env = builtins.__xonsh_env__
+    old = _get_cwd()
+    try:
+        os.chdir(newdir)
+    except:
```
scopatz: Can this except an OSError?

scopatz:
Looks great @wrywerytwreywery.  Just the one exception question.

xonsh/dirstack.py
wrywerytwreywery: Just pushed this change (though I think it should try to catch `FileNotFoundError` as well, so I put that in).

scopatz:
Thanks @wrywerytwreywery!

PR #211, opened 2015-04-17
unpack targets in comprehension

### LINKED ISSUE #210: Dictionary Comprehension Crash
xonsh just crashed when running the following:
```
$ x = {k:float(v) for k,v in x.iteritems()}
```

The error is coming from an `assert False` in the parser.  It looks like an issue with the multiple assignment, since wrapping `k,v` in parens works:
```
$ x = {k:float(v) for (k,v) in x.items()}
```

scopatz:
This should fix #210


wrywerytwreywery:
Looks good!  Just merged in.

PR #213, opened 2015-04-23
History object

aig787:
First workings of a history object and history alias to go with it.

Example output of `history` command:

```bash
$ history
2015-04-23 09:13:49:  history
2015-04-23 09:12:58:  history
2015-04-23 09:12:56:  if 1+1:
                        print('hello')
$
```

It's implemented as an OrderedDict and I've been going back and forth on what I think it should be keyed on. Keying  by command would make searching for commands faster, but an exact ordered history would be expensive. Currently it's keyed on timestamp.

Let me know what should be added/changed!


wrywerytwreywery:
Thanks for putting this in, @aig787!  I don't have time to do a thorough review right now, but I will come back with more comments later this afternoon.  Thanks!

wrywerytwreywery:
Thanks again for putting this in!  I have some thoughts below, and I'll comment on a couple of other things inline.  Hopefully @scopatz can offer his opinion, as well.

1. IIRC, many who were asking for changes to history were asking about storing a "richer" history, including strings representing the output of the execution.  I think this would be a nice feature to have (if we ever sort out return codes, it would be nice to store those, too).  This will make the history much larger, of course, but I recommend we switch to sqlite to address that issue (see bullet 5 below).

2. I think it would be nice for some of these things to be configurable (e.g., how many history items to store, whether to store stdout/stderr strings or just the command that was run, etc).  Similarly, it would be nice if the `history` function were configurable (via args or some other way) to: print more/fewer history entries than 10, print in reverse order, etc.

3. For python mode fragments, it makes sense to store the entered command as a plain string.  For subprocess mode, should we store instead the list representing the lexed form of the command?

3. Right now, it seems like the history is never written to disk, and so is only preserved for a single session.

3. I know there has been some discussion about this already, but I am going to argue for using sqlite instead of JSON to store the history, and to have the history itself live primarily on disk rather than in memory, as [iPython does](https://github.com/ipython/ipython/blob/d56cbbe768f24dadaa1b699568745857ebcc166f/IPython/core/history.py).  The particular practical benefit I would see is that we don't have load the entire (potentially large) history into memory all at once; rather, we can grab particular history items we want straight from the database, making it much less expensive to keep the addition information mentioned above.

  This would mean that the `History` class, rather than being a container itself, would be a frontend to the database.  I think we could also be smart about when to write to the database (only write every `n` commands, or something like that) to avoid some of the disk I/O concerns that come along with this approach.  And I think it solves many of these concerns (how to get the nth item easily, how to avoid the history being big in memory, when to write to disk).

  (I am not sure what we lose by choosing sqlite over JSON; is there some benefit to JSON that I am not seeing?)

3. Regardless of the above decision, I think we need a way to index into the `History` to get the most recent command (which is probably the most common use case of the history object).  Right now, it seems like that would be done with something like `ordered_history[ordered_history.keys()[-1]]`.  I think it would be nicer to be able to do `__history__[-1]` to get that value.  Of course, this should work for arbitrary indices, and not just for -1.  This might actually be easier if we implement the history as a list rather than as an `OrderedDict` (the items in the list could be dictionaries containing the relevant info).

8. Blank lines are currently stored in the history as commands; I think these should probably be ignored.

9. Some small PEP8 formatting things, but those can certainly wait.

xonsh/built_ins.py
```diff
@@ -578,12 +579,15 @@ def load_builtins(execer=None):
     builtins.__xonsh_all_jobs__ = {}
     builtins.__xonsh_active_job__ = None
     builtins.__xonsh_ensure_list_of_strs__ = ensure_list_of_strs
+    builtins.__history__ = History()
```
wrywerytwreywery: This and `ordered_history` should probably be unloaded in the `unload_builtins` function in this same file, as well.

xonsh/shell.py
```diff
@@ -161,6 +161,8 @@ def push(self, line):
 
     def reset_buffer(self):
         """Resets the line buffer."""
+        cmd = ''.join(list(filter(('\n').__ne__, self.buffer)))
```
wrywerytwreywery: The cast to list after calling `filter` should be unnecessary (i.e., `str.join` can be called with a `filter` object as its argument).

Actually, what is the reason for the filtering?  I'm not sure I see it.  That is, couldn't we just store `''.join(self.buffer)`?

xonsh/built_ins.py
aig787: Yes, you're right. I forgot to add them.

xonsh/shell.py
aig787: The reason for filtering it to sort out elements that are a single newline. When multiline input was getting added the final item was a newline.

aig787:
1. You're correct that they were wanting a richer history. The issue I saw with doing that was the potential size of stdout and stderr. Having to load and dump that with JSON would be very inefficient. If we went the sqlite route that wouldn't be as much of an issue.

2. Totally agree. Adding number of items to print to the `history` command was next on my list. Would the other options be environment variable set in the xonshrc?

3.  It might be, I honestly don't know enough about the pros and cons to know. I'd be very open to input on this one.

4. It is only saved for the current session for now, I didn't want to get too wedded to a particular way of storing that info without getting input from people more familiar with xonsh than I am.

5. I originally agreed with @scopatz about using JSON over sqlite, but in working out some of the implementation of it I think I've changed my mind. We'd have to make some pretty significant sacrifices (stderr/stdout saving, history size, etc) to continue to use JSON. The advantage I see to having it that way is that it would be stored in plaintext instead of needing a database client to read it.

7. I agree, but missed that that was happening when I was testing. I'll get that fixed.

8. I'd be happy to fix those :)

xonsh/hist.py
```diff
@@ -0,0 +1,42 @@
...
+
+        Returns
+        -------
+            None
+        """
+        try: 
+            with open(hist_file) as data_file:
+                json_history = json.load(data_file)
+        except:
+            pass
```
scopatz: This should check if the file exists using `os.path.isfile()` rather than using exceptions.

xonsh/hist.py
```diff
@@ -0,0 +1,42 @@
...
+        Parameters
+        ----------
+        cmd: str
+            Command that should be added to the ordered history.
+
+        Returns
+        -------
+            None
+        """
+        self.ordered_history[time.time()] = {'cmd': cmd.strip()}
```
scopatz: Is there are reason to strip() cmd that I am not seeing?

xonsh/hist.py
```diff
@@ -0,0 +1,42 @@
...
+        """Loads previous history from ~/.xonsh_history.json if it exists.
+
+        Parameters
+        ----------
+        hist_file : str, optional
+            Path of the history file.
+
+        Returns
+        -------
+            None
```
scopatz: Not that this blocks, but you probably don't need to document that a function or method returns None :)

xonsh/shell.py
scopatz: Also, you should probably use a lambda like `lamda x: x != '\n'` rather than relying on the magic method of str.

scopatz:
Thanks for putting this in @aig787!

aig787:
@scopatz any opinion on the JSON vs SQLite discussion now that we have a little better idea of the pros and cons?

xonsh/hist.py
aig787: Good to know, thanks!

xonsh/hist.py
aig787: It didn't make sense to store the newline that gets passed in if we send the command straight to this method.

scopatz: My concern is that expression statements such as `x+10\n` is actually parsed differently than a pure expression such as `x+10`.  I sort of think that strip might not be the right thing to do.

scopatz:
@aig787 I am strongly in favor of human readable, version control friendly files here. I think JSON is the better choice.

aig787:
@scopatz any suggestions on how to deal with storing stderr and stdout then? The first thing that comes to mind is compressing them with zlib to dimish their impact and having the History object decompress when needed. But even compressed 10,000 lines of output would be tough to manage

xonsh/hist.py
aig787: Hmm.... Okay. Maybe the `history` alias should handle that and the command should be stored as is then?

xonsh/built_ins.py
aig787: Should be fixed now.

xonsh/hist.py
scopatz: Probably.  I haven't really looked at how `history` is implemented yet.

scopatz:
Compression would help in that case :).  SQLite isn't going to do anything different.  I don't think that it would be so tough to manage.  Is the concern that we wouldn't be able to lazily load in the history?

xonsh/built_ins.py
scopatz: Thanks @aig787!

wrywerytwreywery:
> Compression would help in that case :)

Doesn't compression imply that it won't be human readable?  Or version-control-friendly?

> Is the concern that we wouldn't be able to lazily load in the history?

I think is my main concern, at least.  Might not be a huge worry until the history gets big, but I think it's worth thinking about (speaking as someone who often has ~30 terminals open on a machine, if each is loading the entirety of the history including stdout, stderr, that could be a pain).

> I am strongly in favor of human readable, version control friendly files here.

Is anyone realistically going to be reading/writing the history files directly, or version controlling them?  That seems like an odd use case, and one that would almost certainly have to be done programmatically anyway, so why not just do it in xonsh itself by touching the history object?

We might be able to get the best of both worlds, so to speak, though, if human readable is really a concern: we can make our own history format, where each line is a repl'd dictionary (containing keys like timestamp, command, stdout, etc), or something like that.  This would let us load lazily, but still be human-readable (in a sense).  I have written some similar things before for storing logs that seemed to work pretty well, and I'd be happy to share.

wrywerytwreywery:
> Totally agree. Adding number of items to print to the history command was next on my list. Would the other options be environment variable set in the xonshrc?

I think my rule of thumb would be: anything that controls how the history is stored should be an environment variable (whether to store stdout, where the history should be kept, etc).  And anything having to do with how the `history` command displays the history should be controlled by arguments (ascending/descending order, how many to show, etc).

> It might be, I honestly don't know enough about the pros and cons to know. I'd be very open to input on this one.

This was with regard to a different concern, which deals with how aliases are handled, and which can probably wait until we decide on some of these other issues.

> It is only saved for the current session for now, I didn't want to get too wedded to a particular way of storing that info without getting input from people more familiar with xonsh than I am.

I think this is actually a much more interesting question than I was giving it credit for.  My first instinct is that we want to keep history for separate xonsh sessions (multiple terminals open) separate, but to have new terminal sessions inherit history from all previous sessions (and even from those that are still alive when the new session started).  Living in memory makes this easy; if we store the history on disk directly, though, then we may have to store, with each entry, some kind of session ID representing the session from which that entry came (ugh).

> The advantage I see to having it that way is that it would be stored in plaintext instead of needing a database client to read it.

True, and I usually think human-readability is important, but the history is likely to be a complicated-enough beast that no human is actually going to want to read or modify it anyway, except programmatically, and so I don't see that as a big drawback, so long as we can provide a sane interface (so that, to the extent it's possible, the user doesn't know that they are interacting with a database versus some Python structure living in memory).

aig787:
I personally really like the idea of having the best of both world approach. I think you're right @wrywerytwreywery that we're going to have to sacrifice at least some human readability no matter what we do.

scopatz:
Sorry if the following is a bit rambly...

> I think my rule of thumb would be: anything that controls how the history is stored should be an environment variable (whether to store stdout, where the history should be kept, etc). And anything having to do with how the history command displays the history should be controlled by arguments (ascending/descending order, how many to show, etc).

I totally agree. 

>  Living in memory makes this easy; if we store the history on disk directly, though, then we may have to store, with each entry, some kind of session ID representing the session from which that entry came (ugh).

I don't think that there is a good way to avoid session ids, given that multiple xonshes can be open simultaneously.   Nominally, each session would have its own history file and there would be a xonsh history dir where these all lived.  Maybe there could be an index file for convenience? History is sadly not linear. 

> we can make our own history format 

I am wary of this because it just makes it harder for other tools to interact with our format.  Say, for instance, you just wanted to throw the history into a panda's data frame or wanted to create a little javascript analysis widget.  Using a standard format really enables these kinds of things.  Even if a new format is simple, there still aren't 3rd party tools out there that support it.  

> Is anyone realistically going to be reading/writing the history files directly,

Especially if the stdout and stderr is included, I can definitely see myself grepping through to find old bugs weeks or months later :).  I suppose that in a non-plaintext format one could do something like `history | grep word` rather than `grep word .xonsh-hist/*` But still, I would like there to at least be a fully plaintext option, personally.

It might be possible to build a light-weight lazy json reader/writer that is self-describing.  Somewhat like [fenhl/lazyjson](https://github.com/fenhl/lazyjson). I can play around with this, if you'd like.  

Also, JSON might not be the best option, but I really am not a huge fan of SQLite.  I am definitely open to suggestions. One option would just be to support different backends (JSON, SQLite, CSV) depending on an env variable, but that sounds like a fair amount of work...

wrywerytwreywery:
> I don't think that there is a good way to avoid session ids, given that multiple xonshes can be open simultaneously. Nominally, each session would have its own history file and there would be a xonsh history dir where these all lived. Maybe there could be an index file for convenience? History is sadly not linear. 

Right.  One other question there: when a new terminal is opened, which history should it load from?  The most-recently-modified one?  Some interleaving of all of them?  I don't actually know what readline does, but the right thing to do, at least to start, might be to mimic readline, since that is what the arrow keys use (and so it is likely to be somewhat familiar).

>> we can make our own history format

> I am wary of this because it just makes it harder for other tools to interact with our format. Say, for instance, you just wanted to throw the history into a panda's data frame or wanted to create a little javascript analysis widget. Using a standard format really enables these kinds of things. Even if a new format is simple, there still aren't 3rd party tools out there that support it. 

That's fair, though I do think that the actual format we use isn't really what matters; it's the interface (so that, from xonsh, you could easily generate a structure that could be used with, e.g., pandas).  I can see, though, that if you wanted to do analysis from outside xonsh, that using something more standard makes good sense.  It was just a thought.

> Especially if the stdout and stderr is included, I can definitely see myself grepping through to find old bugs weeks or months later :)

Yes, I do definitely agree that searching through those values is the main (only?) reason that stdout and stderr would be worth saving, but I don't think grep is the right way; or, at least, I think it will be hard to format the JSON (or whatever format we use) such that grep gives you something useful.  

Presumably, the goal of grepping in that way would be to find the value of `n` for which `__history__[n]` gives you the full information about that run?  We would need to make sure that the JSON file (or whatever) is formatted in a way that lets grep actually give you that information.  This seems like it might actually be tricky to get right (IIRC, the default `json.dumps` would dump everything onto one line, so grep wouldn't be particularly helpful).

Most of my thoughts above have been based on the assumption that the best way to navigate that history will always be through python mode, rather than through subproc mode (which I still think is true!).

> , but I really am not a huge fan of SQLite

Is there a reason beyond not being particularly human-readable?  Just curious, as I've been quite happy with SQLite when I've used it for other projects.

> One option would just be to support different backends (JSON, SQLite, CSV) depending on an env variable, but that sounds like a fair amount of work...

Yeah, that seems like overkill to me.  I think we should just find one something that works and run with it, at least as a start.  I'm not particularly set on SQLite; I just think the ability to load lazily is likely to be important.  If we can do some sort of lazy loading of JSON like the one you mentioned, then that works for me.

scopatz:
> Is there a reason beyond not being particularly human-readable? Just curious, as I've been quite happy with SQLite when I've used it for other projects.

I think that it comes from my scientific computing background.  SQL doesn't really scale past 1-10 million rows, and SQLite is the worst of them all.  There is little hope for SQL at petabyte or exabyte scales.  And so if SQL is only good for small- to medium-data sizes there are other formats that do just as well, in my opinion. I guess I have just developed an aversion to it :). 

@aig787 just let us know when you want us to look at this again!

aig787:
I've been taking a look at capturing stderr and stdout and the straightforward way to do it seems to be the redirection options in tools, but it seems to me there has to be a better way to do it. Any suggestions there?

scopatz:
I don't know that there is a better way @aig787.  That is what they are there for :)

aig787:
Okay, I have `ordered_history` changed over to use the new lazyjson library and stdout/stderr being saved, but it's only working for aliases and things running in python mode. I have the redirections happening right before the code is executed in shell.py:

```python
133         try:                                                                    
134             with redirect_stdout(self.stdout):                                  
135                 with redirect_stderr(self.stderr):                              
136                     self.execer.exec(code, mode='single', glbs=self.ctx)  # no locals
```

Is there another place I need to add that to capture things in subprocess mode?

xonsh/shell.py
```diff
@@ -128,7 +115,15 @@ def default(self, line):
         if code is None:
             return
         try:
-            self.execer.exec(code, mode='single', glbs=self.ctx)  # no locals
+            # Temporarily redirect stdout and stderr to save results in
+            # history.
+            with redirect_stdout(self.stdout):
+                with redirect_stderr(self.stderr):
```
scopatz: I think that you can actually write 

```python
with redirect_stdout(self.stdout), redirect_stderr(self.stderr):
```

and save a level of nesting, for what it is worth.

scopatz:
Hi @aig787 - awesome!  I think that you might need to look into the `built_ins.py` file, an in particular, the `_run_callable_subproc()` function https://github.com/scopatz/xonsh/blob/master/xonsh/built_ins.py#L313

This is what is responsible for running both normal subprocess commands and for running aliases.

scopatz:
Hi @aig787, this probably needs an merge from master now.

aig787:
@scopatz I ran into another issue. Applications like vim that rely on having stdout stdin come from a terminal will hang when called in subprocess mode if I'm trying to capture their output and add it to the history. I'm not sure how to check if that's going to happen or not, the only solution that comes to mind is not to store sdout and only keep stderr.

scopatz:
@aig787 - is this a problem on the current HEAD? I can take a look.

aig787:
@scopatz it is now. I'm in the middle of some refactoring, so the code isn't the cleanest but the error is there.

scopatz:
I think that you need to implement tee-like functionality and apply it to both stdout and stderr.  See http://stackoverflow.com/a/11325249/2312428 for an implementation that probably works well enough.

aig787:
Yeah that looks like the way to go

aig787:
I haven't had nearly as much time as I'd like to work on this, and I probably won't for a couple of weeks (just coming into dead and finals week at school). I'd be happy to keep working on it when things slow down, but if it's a priority it might be better to have someone else finish it up.

scopatz:
Hi @aig787 - no worries.  I am on vacation personally and so have been pretty slow to respond myself :).  If no one wants to to pick up your baby, it'd be great if you came back to it afterwards.  Good luck on finals!

PR #216, opened 2015-05-04
Lazy JSON File

scopatz:
Here are some tools for representing lazily loadable JSON files. This has a lot of potential use for history objects as discussed in #213.  @aig787, this one is for you, but @wrywerytwreywery will have to merge.  If you could both take a look that'd be great!

I haven't gone through the work of making all of the output as pretty as possible because I don't want this to block the progress on the history object any longer.  Sorry this took me so long to get around to.  

The basic idea is that for any `obj` that is JSON convertible, you can use the module as follows:

```python
from xonsh.lazyjson import dump, LazyJSON

obj = {....}

with open('/path/to/file', 'w') as f:
    dump(obj, f)  # writes the JSON file with an index inserted

# only reads in the index info
lj = LazyJSON('/path/to/file')  

# each indexing op will return a proxy Node instance if op 
# would give dict or list, otherwise reads value from file.
lj[x][y][z]

# full data structures can be read in manually
lj.load()
lj[x].load()
```

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -38,6 +38,7 @@ For those of you who want the gritty details.
     :maxdepth: 1
 
     tools
+    lazyjson
     openpy
     main
     pyghooks
--- a/docs/api/lazyjson.rst
+++ b/docs/api/lazyjson.rst
@@ -1,0 +1,10 @@
+.. _xonsh_lazyjson:
+
+******************************************
+Lazy JSON Files (``xonsh.lazyjson``)
+******************************************
+
+.. automodule:: xonsh.lazyjson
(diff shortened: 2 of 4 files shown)


aig787:
@scopatz wow that looks really cool! I'll get to work getting it implemented as soon as it gets merged.

wrywerytwreywery:
Sorry I didn't have a chance to look until now; looks good!  Just merged in.

scopatz:
Thanks @wrywerytwreywery!  This should be ready for you @aig787

aig787:
@scopatz thanks I'll get started

PR #218, opened 2015-05-05
completing module to import

zajaczajac:
Simple implementation of the feature requested in https://github.com/scopatz/xonsh/issues/206


scopatz:
Thanks @zajaczajac.  This is a clever quick fix. I kinda like it since it doesn't thrash the file system.  Thanks a ton.

PR #221, opened 2015-05-08
timeit alias

scopatz:
This provides a timeit alias that is mostly co-opted from IPython and then xonshified. Pretty useful, I think :)


wrywerytwreywery:
This is really neat!  I think it's quite a nice addition.  I did have some issues, though.

This worked as expected:
```
$ timeit x = 7
100000000 loops, best of 3: 11.9 ns per loop
```

The following failed because parens aren't allowed to be part of a `subproc_arg_part`:
```
$ timeit print(7)
. 
..... (SNIP)
SyntaxError: <xonsh-code>:1:7: ('code: print',)
```

The following failed differently (because it is valid Python syntax), but for the same reason.
```
$ timeit (7+7)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/shell.py", line 132, in default
  File "/usr/local/lib/python3.4/dist-packages/xonsh/execer.py", line 127, in exec
    return exec(code, glbs, locs)
NameError: name 'timeit' is not defined
```

Passing those things as strings (e.g., `"print(7)"`) worked, but that felt like a pain to me.  I submitted a pull request (#222) that I think should resolve this.

PR #222, opened 2015-05-08
allow grouping syntax to be part of subprocess expressions

wrywerytwreywery:
This PR allows for grouping syntax (parens, braces, brackets) to be part of a subprocess expression.  This should solve an issue with the `timeit` alias, and should also open the door for parenthesized subprocess expressions (so that we can eventually implement <tt>and</tt>, <tt>or</tt>, and <tt>not</tt> properly for subprocesses as we've discussed in the past).

@scopatz: this PR, if accepted, is intended to be merged with your `timeit` branch before merging that with `master`.

### DIFF
--- a/xonsh/lexer.py
+++ b/xonsh/lexer.py
@@ -162,7 +162,7 @@ def handle_lparen(state, token, stream):
     """
     Function for handling ``(``
     """
-    state['pymode'].append((True, '(', ')', token.start))
+    state['pymode'].append((state['pymode'][-1][0], '(', ')', token.start))
     state['last'] = token
     yield _new_token('LPAREN', '(', token.start)
 
@@ -171,7 +171,7 @@ def handle_lbrace(state, token, stream):
     """
     Function for handling ``{``
     """
-    state['pymode'].append((True, '{', '}', token.start))
+    state['pymode'].append((state['pymode'][-1][0], '{', '}', token.start))
     state['last'] = token
     yield _new_token('LBRACE', '{', token.start)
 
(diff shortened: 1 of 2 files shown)


scopatz:
> @scopatz: this PR, if accepted, is intended to be merged with your timeit branch before merging that with master.

Sounds good.  Looking at it now.

wrywerytwreywery:
Actually, this might not be quite the right thing, if we eventually want parens actually to have meaning in subprocess expressions (since this just makes them part of whatever token they live in.

So for `timeit print(7)`, it feels kind of right to me, but for `(echo "hey") and (echo "hello" or echo "hi")` it doesn't work.

Maybe it would be better to let names in subprocess mode have parens in them?  So then `print(7)` would just be a name (and thus would be handled properly by `timeit`), but `(7+7)` would be interpreted as a grouping containing `7+7`, to be executed by xonsh.

scopatz:
> Maybe it would be better to let names in subprocess mode have parens in them?

Probably, this has messed me up on filenames a couple of times. In any event, since this is going to be touching the parser an lexer, would you mind adding a couple of tests for this behavior.  Also, `- (echo "hey") and (echo "hello" or echo "hi")` doesn't seem like current proper xonsh syntax right now because of the ands/ors

wrywerytwreywery:
> doesn't seem like current proper xonsh syntax right now because of the ands/ors

Right.  It is the context in which I was thinking about "proper" handling of parens, though, and is still something I would like to see at some point in the future.

This shouldn't block your `timeit` changes, so if it sounds alright, I'll close this PR, merge in your `timeit` branch, and circle back to this (probably as part of a PR dealing with IO redirection, and with test cases).

scopatz:
Ok that sound good to me.  You don't necessarily need to close this PR though if you don't want to.

> Right. It is the context in which I was thinking about "proper" handling of parens, though, and is still something I would like to see at some point in the future.

I agree!

PR #223, opened 2015-05-09
Feature: Windows support

rbrewer123:
This branch gets xonsh working on Windows.  For my use-case of needing
a better shell on Windows, this initial port of xonsh is already far
more usable than cmd.exe.

I'm running it on Windows 8.1 with Python 3.4.3, ply 3.6, and
pyreadline 2.0.  I also have msysGit 1.9.5.20150319 installed which
provides a number of Unix utilities (e.g. bash, sh, git, ls).  The
nosetests all pass and I created an xonsh.bat script to start it.  I
also verified the nosetests still pass on Ubuntu 14.04.

These are some things to be aware of:

* GNU readline doesn't seem to be readily available on Windows.
  "pip install readline" gives an error.  "pip install pyreadline"
  gives a usable pure Python readline that I'm running with.
* bash on Windows changes PATH, PWD, and possibly other
  path-related environment variables to a '/c/Windows/System32'
  style which breaks xonsh's environment.  To solve this, on Windows
  I revert those variables back to the original os.environ settings.
  For compatibility with bash on Windows it might be nice to
  support that syntax in the future.  (Right now the code
  allows paths like 'c:\\foo' or 'c:/foo' but not '/c/foo').
* Windows Python doesn't have the signals or process group support
  required by the existing job control mechanism.
  I changed jobs.py to first attempt to import jobs_posix.py
  and fall back to jobs_not_implemented.py
  if that import fails.  My guess is that emulating job control
  on Windows would be a bit involved since the Python infrastructure
  would first need to be created.

For the following issues, my preference would be to open a separate
issue for each one to track them.  I think it would be nice to roll
out the initial functionality without waiting for them and perhaps
others can jump in to help fix these too:

* Background jobs (via '&') seem to work.  However, if a background
  job is currently running and you issue any of the job-control commands
  (e.g. jobs, bg, fg), the shell hangs until the background job
  completes.  I haven't investigated this much yet.  My current
  workaround is to not start background jobs.
* My initial attempt at retrieving the current git branch was
  too slow to be usable, so I disabled that on Windows.
* I got a shebang script for xonsh to run directly from xonsh
  with the following initial line::

    #!python -m xonsh.main

  msysGit comes with various Unix utilities implemented or wrapped
  with /bin/sh scripts which seem to work with its included bash
  tools.  I haven't looked too far into how to get them to run with
  xonsh yet but this would be a nice capability.  Also getting the
  xonsh shebang line to work with a more-standard::

    #!/usr/bin/env xonsh

  line would be nice too.

The installation process for Windows is roughtly this:

* install Python 3.4.3 (I use chocolatey.org for this)
* ``pip install ply``
* ``pip install pyreadline``
* ``python setup.py install``
* ``scripts\xonsh``

Let me know if you need me to make any additional changes to
support merging this.

### DIFF
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -11,13 +11,71 @@ from subprocess import TimeoutExpired
 
 from xonsh.tools import ON_WINDOWS
 
-ProcProxy = namedtuple('ProcProxy', ['stdout', 'stderr'])
-"""
-A class representing a Python function to be run as a subprocess command.
-"""
+try:
+    _shell_tty = sys.stderr.fileno()
+except OSError:
+    _shell_tty = None
 
 
-if not ON_WINDOWS:
+if ON_WINDOWS:
+    def _continue(obj):
+        pass
+
+
+    def _kill(obj):
+        return obj.kill()
+
+
+    def ignore_sigtstp():
+        pass
+
+
+    def _set_pgrp(info):
+        pass
+
+    def wait_for_active_job():
+        """
+        Wait for the active job to finish, to be killed by SIGINT, or to be
+        suspended by ctrl-z.
+        """
+        _clear_dead_jobs()
+        act = builtins.__xonsh_active_job__
+        if act is None:
+            return
+        job = builtins.__xonsh_all_jobs__[act]
+        obj = job['obj']
+        if job['bg']:
+            return
+        while obj.returncode is None:
+            try:
+                obj.wait(0.01)
+            except TimeoutExpired:
+                pass
+            except KeyboardInterrupt:
+                obj.kill()
+        if obj.poll() is not None:
+            builtins.__xonsh_active_job__ = None
+
+else:
+    def _continue(obj):
+        os.kill(obj.pid, signal.SIGCONT)
+
+
+    def _kill(obj):
+        os.kill(obj.pid, signal.SIGKILL)
+
+
+    def ignore_sigtstp():
+        signal.signal(signal.SIGTSTP, signal.SIG_IGN)
+
+
+    def _set_pgrp(info):
+        try:
+            info['pgrp'] = os.getpgid(info['obj'].pid)
+        except ProcessLookupError:
+            pass
+
+
     _shell_pgrp = os.getpgrp()
 
     _block_when_giving = (signal.SIGTTOU, signal.SIGTTIN, signal.SIGTSTP)
@@ -33,34 +91,42 @@ if not ON_WINDOWS:
             signal.pthread_sigmask(signal.SIG_SETMASK, oldmask)
 
 
-try:
-    _shell_tty = sys.stderr.fileno()
-except OSError:
-    _shell_tty = None
+    def wait_for_active_job():
+        """
+        Wait for the active job to finish, to be killed by SIGINT, or to be
+        suspended by ctrl-z.
+        """
+        _clear_dead_jobs()
+        act = builtins.__xonsh_active_job__
+        if act is None:
+            return
+        job = builtins.__xonsh_all_jobs__[act]
+        obj = job['obj']
+        if job['bg']:
+            return
+        pgrp = job['pgrp']
+        obj.done = False
 
-
-def _continue(obj):
-    if not ON_WINDOWS:
-        os.kill(obj.pid, signal.SIGCONT)
-
-
-def _kill(obj):
-    if ON_WINDOWS:
-        obj.kill()
-    else:
-        os.kill(obj.pid, signal.SIGKILL)
-
-
-def ignore_sigtstp():
-    if not ON_WINDOWS:
-        signal.signal(signal.SIGTSTP, signal.SIG_IGN)
+        _give_terminal_to(pgrp)  # give the terminal over to the fg process
+        _, s = os.waitpid(obj.pid, os.WUNTRACED)
+        if os.WIFSTOPPED(s):
+            obj.done = True
+            job['bg'] = True
+            job['status'] = 'stopped'
+            print()  # get a newline because ^Z will have been printed
+            print_one_job(act)
+        elif os.WIFSIGNALED(s):
+            print()  # get a newline because ^C will have been printed
+        if obj.poll() is not None:
+            builtins.__xonsh_active_job__ = None
+        _give_terminal_to(_shell_pgrp)  # give terminal back to the shell
 
 
 def _clear_dead_jobs():
     to_remove = set()
     for num, job in builtins.__xonsh_all_jobs__.items():
         obj = job['obj']
-        if isinstance(obj, ProcProxy) or obj.poll() is not None:
+        if obj.poll() is not None:
             to_remove.add(num)
     for i in to_remove:
         del builtins.__xonsh_all_jobs__[i]
@@ -109,11 +175,7 @@ def add_job(info):
     """
     info['started'] = time.time()
     info['status'] = 'running'
-    if not ON_WINDOWS:
-        try:
-            info['pgrp'] = os.getpgid(info['obj'].pid)
-        except ProcessLookupError:
-            return
+    _set_pgrp(info)
     num = get_next_job_number()
     builtins.__xonsh_all_jobs__[num] = info
     builtins.__xonsh_active_job__ = num
@@ -123,50 +185,6 @@ def add_job(info):
 
 def _default_sigint_handler(num, frame):
     raise KeyboardInterrupt
-
-
-def wait_for_active_job():
-    """
-    Wait for the active job to finish, to be killed by SIGINT, or to be
-    suspended by ctrl-z.
-    """
-    _clear_dead_jobs()
-    act = builtins.__xonsh_active_job__
-    if act is None:
-        return
-    job = builtins.__xonsh_all_jobs__[act]
-    obj = job['obj']
-    if isinstance(obj, ProcProxy):
-        return
-    if job['bg']:
-        return
-    if ON_WINDOWS:
-        while obj.returncode is None:
-            try:
-                obj.wait(0.01)
-            except TimeoutExpired:
-                pass
-            except KeyboardInterrupt:
-                obj.kill()
-    else:
-        pgrp = job['pgrp']
-        obj.done = False
-
-        _give_terminal_to(pgrp)  # give the terminal over to the fg process
-        _, s = os.waitpid(obj.pid, os.WUNTRACED)
-        if os.WIFSTOPPED(s):
-            obj.done = True
-            job['bg'] = True
-            job['status'] = 'stopped'
-            print()  # get a newline because ^Z will have been printed
-            print_one_job(act)
-        elif os.WIFSIGNALED(s):
-            print()  # get a newline because ^C will have been printed
-    if obj.poll() is not None:
-        builtins.__xonsh_active_job__ = None
-
-    if not ON_WINDOWS:
-        _give_terminal_to(_shell_pgrp)  # give terminal back to the shell
 
 
 def kill_all_jobs():
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -27,6 +27,7 @@ For those of you who want the gritty details.
     aliases
     dirstack
     jobs
+    proc
     inspectors
     completer
     shell
--- a/docs/api/proc.rst
+++ b/docs/api/proc.rst
@@ -1,0 +1,9 @@
+.. _xonsh_proc:
+
+******************************************************
+Python Procedures as Subprocess Commands (``xonsh.proc``)
+******************************************************
+
+.. automodule:: xonsh.proc
+    :members:
+    :undoc-members:
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -7,6 +7,7 @@ import sys
 import shlex
 import signal
 import locale
+import inspect
 import builtins
 import subprocess
 from io import TextIOWrapper, StringIO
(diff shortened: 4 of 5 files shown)


wrywerytwreywery:
Great; thanks for putting this in!  Some comments below, and I may make some more inline.

Does it make sense to have a variable someplace (environment or otherwise) that tells us whether we're in Windows or not, rather than importing `platform` and checking `platform.system()` in multiple places?  Presumably, we will need to make this check in more and more places as more features are implemented (for example, I'm not sure whether the `ProcProxy` rewrite in #224 will work out-of-the box, but I can check).

One interesting issue:
```
user@PC ~\Documents\xonsh $ gitk
xonsh: subprocess mode: command not found: gitk
Did you mean the following?
      gitk:  Command (C:\Program Files (x86)\Git\bin\gitk)
```
  This happens both from `cmd` and from git bash.  This may actually be my bug ([this code](https://github.com/scopatz/xonsh/blob/master/xonsh/built_ins.py#L395-410) is trying to figure out which interpreter to use based on the shebang line, but it's possible that `subprocess` is going to be better than us at interpreting shebang lines, so maybe we should just let it do its thing and only falling back to explicitly using an interpreter if none is given in the shebang line?).

> bash on Windows changes PATH, PWD, and possibly other path-related environment variables to a

I think there is still something of an issue related to this.  From within git bash, I get the following:
```
user@PC c:\Users\user\Documents\xonsh $ cd
user@PC c:\Users\user $ cd Downloads/
user@PC c:\Users\user\Downloads $ cd
cd: no such file or directory: /c/Users/user
```

> Windows Python doesn't have the signals or process group support required by the existing job control mechanism. I changed jobs.py to first attempt to import jobs_posix.py and fall back to jobs_not_implemented.py if that import fails. My guess is that emulating job control on Windows would be a bit involved since the Python infrastructure would first need to be created.

Yes; I wrote the job control mechanism with only *nix in mind.  [psutil](https://github.com/giampaolo/psutil) at least has support for suspending/resuming, so we may be able to use that, or build something up using some of those ideas.  But I think you're right that implementing full-fledged job control in Windows will be a quite a process (pun intended :neutral_face:).  I'd be happy to help with that.

> Background jobs (via '&') seem to work. However, if a background job is currently running and you issue any of the job-control commands (e.g. jobs, bg, fg), the shell hangs until the background job completes. I haven't investigated this much yet. My current workaround is to not start background jobs.

Hmmm that failure mode is surprising to me.  I can try to find some time to look at it.

> The installation process for Windows is roughtly this:

Worked fine for me (modulo changing my path to get `pip` and `python` in there).  Eventually, we will want to document this, as well.

scopatz:
Hi @rbrewer123!  First, let me say that I fully support your effort here and I would love to see first-class windows support in xonsh, and the sooner the better :).  

In terms of process here, I agree that having smaller, targeted PRs would be better than big monolithic ones.  Let's open separate PRs/issues for all of the points that you mentioned.

Also, out of curiosity, have you used Appveyor?

xonsh/environ.py
```diff
@@ -21,6 +21,11 @@ def current_branch(cwd=None, pad=True):
     bust should be extended in the future.
     """
     branch = None
+
+    if platform.system() == 'Windows':
```
scopatz: For consistency and speed, it would be nice to have (a) global variable(s) in tools.py that is something like:

```python
ON_POSIX = (os.name == 'posix')
```

This saves on memory and speed and decreases the likelihood of annoying typos.

xonsh/jobs.py
```diff
@@ -1,215 +1,13 @@
...
-    job = builtins.__xonsh_all_jobs__[act]
-    job['bg'] = True
-    job['status'] = 'running'
-    print_one_job(act)
-    os.kill(job['obj'].pid, signal.SIGCONT)
+
+import platform
+from warnings import warn
+
+try:
```
scopatz: It would be better to use and if-statement here rather than a try-except, if we can.  Is the issue detecting whether we are in msys or cmd?

scopatz:
Also, maybe jobs*.py should be a subpackage / dir where the current jobs.py file becomes the `jobs/__init__.py`.  I think that this is more elegant than having three files in the root xonsh package that all start with 'jobs'

xonsh/jobs_not_implemented.py
```diff
@@ -0,0 +1,49 @@
...
+    return '', msg
+
+def fg(args, stdin=None):
+    return '', msg
+
+def bg(args, stdin=None):
+    return '', msg
+
+
+def ignore_SIGTSTP():
```
scopatz: PEP8 function names should be lowercase.

xonsh/jobs_posix.py
```diff
@@ -0,0 +1,219 @@
...
+
+try:
+    _shell_tty = sys.stderr.fileno()
+except OSError:
+    _shell_tty = None
+
+_shell_pgrp = os.getpgrp()
+
+
+def ignore_SIGTSTP():
```
scopatz: PEP8

scopatz:
@rbrewer123, let me know if you have any complaints / issues with my feedback.

scopatz:
And also when the PR is ready for another look :)

rbrewer123:
@wrywerytwreywery, thanks for the feedback, and the additional pull requests.  I didn't expect the jobs support on Windows to improve that quickly.  :)

@scopatz, your feedback is helpful, especially about making jobs.py into a directory.  @wrywerytwreywery also helped me realize I haven't tested with vanilla cmd.exe, which seems to be less forgiving than the ConEmu terminal I was running xonsh in.

I haven't used Appveyor, after a quick google, is it something like travis?  I haven't really used travis either, but saw that xonsh is setup for it.  I'm new to collaboration via github and am still trying to figure out the etiquette in addition to just using the tools.  Feel free to point out what I need to improve on, it will help me learn.

I'll let you all know when I've got the next round of the PR ready.

wrywerytwreywery:
> especially about making jobs.py into a directory

On that note: it seems like there is a lot of shared code between those modules.  It could be that the right thing would be actually just to combine them all into one file (`jobs.py`) and change the definitions based on OS where appropriate.  I think making `jobs` a subpackage is preferable to the way it is set up right now, but it might be worth considering, also, just combining everything into one file (`subprocess.py` in the standard library, for example, does this a little bit).

scopatz:
> I haven't used Appveyor, after a quick google, is it something like travis? 

Ok don't worry about it then :).  It is like Travis, but for windows.

> I'll let you all know when I've got the next round of the PR ready.

Great!

xonsh/jobs.py
rbrewer123: The idea I was going for was to test for finer-grained features and use them or fall back rather than inferring whether the features are available from a coarse-grained check of platform.system().  I don't think it's a clear win in the this case, because a minor error in jobs_posix.py could cause it to fall back to the non-implemented case and mask other issues (which is why I added a warning).  I'll revisit it when I move jobs to a subpackage, especially since @wrywerytwreywery already has improved the jobs support for Windows.

rbrewer123:
@wrywerytwreywery I'll take a shot at recombining jobs_windows and jobs_posix.

rbrewer123:
@wrywerytwreywery the gitk bug you mention above I can reproduce.  It seems to happen with all shebang scripts in the msysGit package.  I think on Windows we may need to remap absolute paths like '/bin/sh' to just 'sh' and things will work.  And possibly map things like '/usr/bin/env python' to just 'python'.  My preference would be to open a separate issue for this after the initial PR is merged.

I'm not able to reproduce the 'cd Downloads' issue you're seeing.  Can you provide a little more description of your setup when you're seeing it?  Or are you still seeing it at this point?

scopatz:
Hi @rbrewer123, this needs a merge from master now.

wrywerytwreywery:
> I think on Windows we may need to remap absolute paths like '/bin/sh' to just 'sh' and things will work. And possibly map things like '/usr/bin/env python' to just 'python'. My preference would be to open a separate issue for this after the initial PR is merged.

I think that sounds good (hopefully it will work!).  I'm happy to help with that when the time comes.

> I'm not able to reproduce the 'cd Downloads' issue you're seeing. Can you provide a little more description of your setup when you're seeing it? Or are you still seeing it at this point?

It doesn't happen from `cmd` or from starting `xonsh` with the batch file you included, but if I open git bash (right-clicking inside a Git repo and clicking 'Git Bash') and then open xonsh from within there, I get that error.

Not sure whether it matters, but I was on a machine running Windows 7 64-bit, with Python 3.4.3 installed directly from the python.org installer, ply/pyreadline installed from pip, and bash via installing Git for Windows 1.9

rbrewer123:
@wrywerytwreywery Thanks, that was enough for me to reproduce the git bash problem.  I think it's squashed.  The prompt sometimes doesn't compress back to a ~ in the user's home directory.  Again I think I'd rather open a separate issue for that.

rbrewer123:
@scopatz thanks, @wrywerytwreywery has already done most of it for me on the side.  We're working out the kinks.

rbrewer123:
@scopatz and @wrywerytwreywery, while we work out the merge from master, I think this PR is ready for another round of feedback.

Since Windows normally uses 'dir' for directory listings, and that
maps to the Python builtin dir(), if a user doesn't have unix tools such
as 'ls' installed it's not obvious how to get a directory listing in
xonsh.  I added a default 'xdir' alias that calls the cmd.exe 'dir'
builtin on Windows or /bin/dir on other platforms.  We should probably
document this somewhere so that especially new Windows users aren't lost
on their first command. 

@wrywerytwreywery  improved the job control so '&' works correctly and
you can view outstanding background jobs.  Suspending jobs via ctrl-z
is the only job-related feature missing right now.  I refactored
jobs.py back to a single file and removed jobs_not_implemented.py.  I
think it should be easier to maintain in a single file for now.  @wrywerytwreywery 
has some additional cleanups of that in the merge from master which I plan
to accept.

I tightened my changes to reglob() so that backslashes are only
escaped on Windows.  This means that regexes using backslash special characters
should still work on non-Windows platforms, but will not work
on Windows since it's hard to get around the need for backslashes
in paths.  I added a testcase to verify this functionality on
non-Windows platforms.

xonsh/tools.py
```diff
@@ -31,6 +32,10 @@
 
 DEFAULT_ENCODING = sys.getdefaultencoding()
 
+ON_WINDOWS = (platform.system() == 'Windows')
+ON_MAC     = (platform.system() == 'Darwin')
+ON_POSIX   = (os.name == 'posix')
```
scopatz: PEP8, don't align these equal signs.

xonsh/tools.py
```diff
@@ -366,3 +371,13 @@ def suggestion_sort_helper(x, y):
     inx = len([i for i in x if i not in y])
     iny = len([i for i in y if i not in x])
     return lendiff + inx + iny
+
+
+def escape_windows_command_string(s):
+    s = s.replace('%','%%')
```
scopatz: PEP8 after comma

xonsh/tools.py
```diff
@@ -366,3 +371,13 @@ def suggestion_sort_helper(x, y):
...
+    for c in ('^', '&', '<', '>', '|', "'", '`', ',', ';', '=', '(', ')'):
```
scopatz: Can't this tuple just be a str?

scopatz: Also this might be faster as a regex sub() call... but don't worry about it (ie, non-blocking.)

xonsh/tools.py
```diff
@@ -366,3 +371,13 @@ def suggestion_sort_helper(x, y):
     inx = len([i for i in x if i not in y])
     iny = len([i for i in y if i not in x])
     return lendiff + inx + iny
+
+
+def escape_windows_command_string(s):
```
scopatz: A test for this function would be nice, please

scopatz:
Hey @rbrewer123 - thanks again for all of your hard work here. 

> Since Windows normally uses 'dir' for directory listings, and that
maps to the Python builtin dir(), if a user doesn't have unix tools such
as 'ls' installed it's not obvious how to get a directory listing in
xonsh. I added a default 'xdir' alias that calls the cmd.exe 'dir'
builtin on Windows or /bin/dir on other platforms. We should probably
document this somewhere so that especially new Windows users aren't lost
on their first command. 

This seems fine to me.  We do have an aliases.rst file in the docs dir, where this could be described.  Were you actually noticing a problem though?  xonsh should see `dir filename` and know that this should be executed in subprocess mode, rather than in Python mode.  Only if the filename happened to starte with a literal operator (`dir +1`) would this run in Python mode and fail....at least from my limited testing.  Even this could be made to work with quotes, `dir "+1"`

rbrewer123:
@scopatz thanks for the feedback.  I guess it's been too long since I read PEP 8.  :)  You're welcome on the PR, I'm finding xonsh to be the command shell I never knew I always wanted so it's nice to help improve it.

Regarding the xdir alias, I did some more testing on Windows and Linux.  On Windows, the usual dir command seems to be a builtin of cmd.exe and so ``dir`` always fails in xonsh's subprocess mode.  On Linux it behaves as you describe and an alias seems less needed, e.g. I can list the current directory with ``dir .``.  For cross-platform consistency I'd still like to retain the xdir alias on all platforms.

rbrewer123:
I had the mistaken belief that aliasing ``dir`` would cause problems for Python mode.  But it doesn't, so I removed the ``xdir`` alias and now there is only a ``dir`` alias on Windows to achieve parity with Linux, which doesn't need an alias for ``dir``.

scopatz:
Awesome :0!

scopatz:
@rbrewer123 - let me know when this is ready to be looked at again!

rbrewer123:
@scopatz will do; @wrywerytwreywery did most of the ProcProxy merge for me, but something's amiss with the handle and file descriptor code on Windows and I'm having a difficult time getting the tests to pass.

rbrewer123:
@scopatz it's ready for another look.  My smoke test on Ubuntu 14.04 looks good too.  A big thank you to @wrywerytwreywery for a lot of help getting this done too.  Any thoughts on how/where to document the installation procedure?  My procedure is still as listed above, with pip for dependencies and ``python setup.py install`` for xonsh itself.

After this is merged I'll open separate issues for the remaining things I know about (most of which we've mentioned already in this thread).

scopatz:
> Any thoughts on how/where to document the installation procedure?

Probably the readme or make a new install page in the docs.

scopatz:
Thanks a ton @rbrewer123 (and @wrywerytwreywery for the assist!) This is awesome to have!

Issue #137, opened 2015-03-23
Equivalent to bash's "&&"


averageapps:
Great project! It would be very nice to have a keyword in xonsh that is equivalent to `&&` in bash/zsh or `;and` in fish, meaning that it would execute \<B\> if and only if \<A\> is successful in the following example line:
`<A> <keyword> <B>`.
For maximum "bashwards compatibility" the keyword should be exactly like in bash (&&) if possible.

For completeness' sake there should also be a keyword equivalent to `||` (bash/zsh) or `;or` (fish) that only executes something if the preceding command fails.

The current xonsh behavior (v0.1.2) for this syntax seems to be 100% cpu usage and doing nothing until I hit Ctrl+C

wrywerytwreywery:
I agree wholeheartedly.  I would be happy to take this one on.

mbollmann:
I was already thinking whether the standard subprocess mode $[...] shouldn't return the command's exit code by default, however, a naive implementation would cause the exit code to be printed on screen then each time. Also, the "success" code is 0, which means it can't easily be used in standard Python comparisons, which is awkward.

Alternatively, $[...] could return a StatusCode object, which would evaluate to True if the code was 0, False otherwise, and prints as an empty string, though that seems a bit hacky.

However, I also believe that implementing such a behaviour would require quite a few internal changes, because the return code of subprocess commands isn't even currently stored as far as I remember.

Just some food for thought @wrywerytwreywery, because I investigated this for a little while already :)

scopatz:
Hi All, before anyone gets too far down the road on this, I'd like to make a few points:

1. Right now you can do something that is semantically equivalent to bash `&&` by separating with  semicolons `;`. No, it doesn't check the return status, but it does fail approriately.
2. I would like to see Python's logical operators used (`and`, `or`, `not`) rather than Bash's and C's logical operators (`&&`, `||`, `!`). There is no reason to add complexity to the language for this feature.
3. As @mbollmann alluded to, we probably need a rich history class for this to be done well or correctly. There are other issues open right now with respect to that.

I'd also like to note that precedence rules come in to play here. Thus the need for grouping with `()` arises.  On the bright side, the parser side of things is implemented in Python mode. So implementing it in subproc mode would be more about copying than the work.

averageapps:
@scopatz You are right, using standard python operators for this would be more elegant.
Thanks for the hint that the semicolon behaves this way.

PR #224, opened 2015-05-09
ProcProxy Rewrite

wrywerytwreywery:
This is a rewrite of `ProcProxy` to make it look (almost) the same as `Popen`.  Right now, though, I think this should be a "drop-in" replacement for the current `ProcProxy`.  But this should make it easier to implement things like better IO redirection (#189) and `and`, `or`, and `not` (#137), which I can turn my attention to next.

~~P.S. I am not sure how, but this seems to have broken some of the test cases for the `imphooks` module, so it is still a work in progress.  If you wouldn't mind helping me try to figure that one out, @scopatz, I would appreciate it :)~~

P.P.S. No docs yet; wanted to make sure this was something we actually want first, and also wanted to fix the imphooks tests.

### DIFF
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -2,8 +2,8 @@
 
 Code for several helper methods in the `ProcProxy` class have been reproduced
 without modification from `subprocess.py` in the Python 3.4.2 standard library.
-The contents of that file (and, thus, the reproduced methods) is
-Copyright (c) 2003-2005 by Peter Astrand <astrand@lysator.liu.se>, and was
+The contents of `subprocess.py` (and, thus, the reproduced methods) are
+Copyright (c) 2003-2005 by Peter Astrand <astrand@lysator.liu.se> and were
 licensed to the Python Software foundation under a Contributor Agreement.
 """
 import io
@@ -25,7 +25,7 @@ if iswindows:
 
 class ProcProxy(Thread):
     """
-    Class representing a Python function to be run as a subprocess-mode command.
+    Class representing a function to be run as a subprocess-mode command.
     """
     def __init__(self, f, args,
                  stdin=None,
@@ -55,7 +55,8 @@ class ProcProxy(Thread):
         """
         self.f = f
         """
-        The function to be executed.  It should be a function of four arguments, described below.
+        The function to be executed.  It should be a function of four
+        arguments, described below.
 
         Parameters
         ----------
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -1,6 +1,7 @@
 import io
 import os
 import sys
+import platform
 
 from threading import Thread
 from subprocess import Popen
@@ -8,7 +9,13 @@ from collections import Sequence
 
 from xonsh.tools import redirect_stdout, redirect_stderr
 
-class ProcProxy(Thread, Popen):
+iswindows = platform.system() == 'Windows'
+if iswindows:
+    import _winapi
+    import msvcrt
+
+
+class ProcProxy(Thread):
     def __init__(self, f, args,
                  stdin=None,
                  stdout=None,
@@ -18,7 +25,8 @@ class ProcProxy(Thread, Popen):
         self.args = args
         self.pid = None
         self.returncode = None
-       
+        self.wait = self.join
+
         handles = self._get_handles(stdin, stdout, stderr)
         (self.p2cread, self.p2cwrite,
          self.c2pread, self.c2pwrite,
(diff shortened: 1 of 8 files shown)


wrywerytwreywery:
Wait, are those test cases actually the intended behavior?  The particular test case is importing the following script:

```
# I am a test module.
aliases['echo'] = lambda args, stdin=None: print(' '.join(args))

$WAKKA = "jawaka"
x = $(echo "hello mom" $WAKKA)
```

The first line should change it so that `echo` gets the alias instead of the `echo` executable, right?  Then, once we have that, we have an alias that returns `None`, which makes me think that the right value for `x` in that case should be an empty string, as opposed to `hello mom jawaka\n`, which is what the test case is currently expecting (and which is what would be the output of the `echo` program).

Am I missing something?

wrywerytwreywery:
Oh, I see; I didn't realize that the stdout and stderr from the old `ProcProxy`'s call were being redirected.  I will implement that now.

wrywerytwreywery:
More testing suggests that this breaks job control (since `fg`, `bg`, and `jobs` are now jobs themselves, foregrounding a job doesn't currently work as expected).  So this one isn't quite ready yet.

wrywerytwreywery:
Okay, that most recent push should fix the issue.  It makes since that, because `ProcProxy` instances can't be controlled in the same way as other processes (no suspend/resume, etc), they should not be considered jobs themselves.

Moving forward, we could implement a variant of `ProcProxy` that uses `multiprocessing` instead of `threading` (as we've discussed in the past), but I think `threading` as the default makes good sense since a lot of our built-ins rely on sharing variables.

wrywerytwreywery:
Also, just as a note, this doesn't work with #223 as it stands now (because of no job control on Windows), but it is close.  If both PRs are accepted, I think have a short-term fix for until we can get some kind of job control implemented for Windows.

xonsh/proc.py
```diff
@@ -0,0 +1,90 @@
...
+import sys
+
+from threading import Thread
+from subprocess import Popen
+from collections import Sequence
+
+from xonsh.tools import redirect_stdout, redirect_stderr
+
+class ProcProxy(Thread, Popen):
+    def __init__(self, f, args, stdin, stdout, stderr, universal_newlines):
```
scopatz: Many of these should have default args, I think.

xonsh/proc.py
```diff
@@ -0,0 +1,90 @@
...
+        self.f = f
+        self.args = args
+        self.pid = None
+        self.returncode = None
+       
+        handles = self._get_handles(stdin, stdout, stderr)
```
scopatz: Is there anyway to do this without relying on the CPython private interface?

scopatz: Just asking....

xonsh/proc.py
```diff
@@ -0,0 +1,90 @@
...
+        if self.errread != -1:
+            self.stderr = io.open(self.errread, 'rb', -1)
+            if universal_newlines:
+                self.stderr = io.TextIOWrapper(self.stderr)
+
+        Thread.__init__(self)
+        self.start()
+
+    def run(self):
+        if self.f is not None:
```
scopatz: Probably better to say 

```python
if self.f is None:
    return
# need to make file-likes here that work in the "opposite" direction
....
```

That way a level of nesting is saved and the conditional tests a positive.  This is not a blocking issue, but would be nice if you are going back through this part of the code.

xonsh/proc.py
```diff
@@ -0,0 +1,90 @@
...
+            elif r is not None:
+                stdout.write(str(r))
+            return True
+        except:
+            return False
+    return wrapped_simple_command_proxy
+
+class SimpleProcProxy(ProcProxy):
+    def __init__(self, f, args, stdin, stdout, stderr, universal_newlines):
+        ProcProxy.__init__(self, _simple_wrapper(f), args,
```
scopatz: probably should use super() here.

scopatz:
Hi @wrywerytwreywery - thanks a lot for this.  I like it.  It really simplifies and cleans up the logic in `run_subproc()` and is most welcome.  I think that this probably is the way that we want to go, ultimately. 

If threading works for now, let's run with it. 

I think the big thing that is missing right now is the docs, as you pointed out.

My only real technical concern is the `Popen._get_handles()` function. I don't really understand what is going on here per say.  Do you have a link you could post on this?

wrywerytwreywery:
Thanks for the comments!  I can definitely make the changes you mentioned inline.  I usually avoid `super` in favor of being more explicit about the method I want to call, but that's not a big deal.

> I think the big thing that is missing right now is the docs, as you pointed out.

There are also (unfortunately) still some timing issues (sometimes, for example, `cd` doesn't finish quickly enough for the change to be "caught" by the next prompt, resulting in the wrong pwd being reported).  It's easy enough to fix, but I haven't found an elegant way yet (still working on it).  I'll let you know when it is ready, and I'll go ahead and document.

> My only real technical concern is the `Popen._get_handles()` function. I don't really understand what is going on here per say.  Do you have a link you could post on this?

No link that I know of.  I found out about it by looking into the `subprocess` source.  Basically, it just sets up appropriate file descriptors for the IO streams in a subprocess.  There is a helpful comment in `subprocess.py` just above where the method is called:
```
# Input and output objects. The general principle is like
# this:
#
# Parent                   Child
# ------                   -----
# p2cwrite   ---stdin--->  p2cread
# c2pread    <--stdout---  c2pwrite
# errread    <--stderr---  errwrite
#
# On POSIX, the child objects are file descriptors.  On
# Windows, these are Windows file handles.  The parent objects
# are file descriptors on both platforms.  The parent objects
# are -1 when not using PIPEs. The child objects are -1
# when not redirecting.

(p2cread, p2cwrite,
 c2pread, c2pwrite,
 errread, errwrite) = self._get_handles(stdin, stdout, stderr)
```

I didn't want to be handling things like `PIPE` from scratch when `Popen` already implemented everything I want, and it was conveniently encapsulated in that nice method (and cross-platform!).  I think it basically "just works."  Because `Popen` deals with those fd's differently than `ProcProxy` wants to, I thought this was the right place to grab from.

That said, I admit it's a little strange to inherit from `Popen` just to get that method with no other common functionality...

scopatz:
Maybe we could just copy the `_get_handles()` source?  It seems pretty simple (below).  This seems worth it to me to not rely on a private and undocumented interface :)

```python
        def _get_handles(self, stdin, stdout, stderr):
            """Construct and return tuple with IO objects:
            p2cread, p2cwrite, c2pread, c2pwrite, errread, errwrite
            """
            p2cread, p2cwrite = -1, -1
            c2pread, c2pwrite = -1, -1
            errread, errwrite = -1, -1

            if stdin is None:
                pass
            elif stdin == PIPE:
                p2cread, p2cwrite = os.pipe()
            elif stdin == DEVNULL:
                p2cread = self._get_devnull()
            elif isinstance(stdin, int):
                p2cread = stdin
            else:
                # Assuming file-like object
                p2cread = stdin.fileno()

            if stdout is None:
                pass
            elif stdout == PIPE:
                c2pread, c2pwrite = os.pipe()
            elif stdout == DEVNULL:
                c2pwrite = self._get_devnull()
            elif isinstance(stdout, int):
                c2pwrite = stdout
            else:
                # Assuming file-like object
                c2pwrite = stdout.fileno()

            if stderr is None:
                pass
            elif stderr == PIPE:
                errread, errwrite = os.pipe()
            elif stderr == STDOUT:
                errwrite = c2pwrite
            elif stderr == DEVNULL:
                errwrite = self._get_devnull()
            elif isinstance(stderr, int):
                errwrite = stderr
            else:
                # Assuming file-like object
                errwrite = stderr.fileno()

            return (p2cread, p2cwrite,
                    c2pread, c2pwrite,
                    errread, errwrite)

```

wrywerytwreywery:
My preference would be just to use what is already there, but I suppose it's fine just to copy that code over (as well as the Windows equivalent, of course, given what we're working on there).

scopatz:
My concern with relying on a private interfaces is that (1) different python implementations are free to implement this in different ways, and (2) that CPython would be within their rights to change this in a way that breaks the xonsh.  In these cases, it could be difficult or annoying to track down a version that works and come up with a fix.  The better/easier/more maintainable solution (in my opinion) is to just copy this code. I hope that makes sense!

wrywerytwreywery:
Yes, that makes sense.  The reason I had wanted to avoid that, is that these things tend to chain together.  I'm not opposed, but it will require copying over more code than what you have included above, which I had thought would make `proc.py` harder to maintain in the long run (because it stops being a simple 100-line file and starts being a 250-line file with a lot of low-level weirdness in it).

That said, I think your points probably outweigh my concerns, so I'll go ahead and proceed that way.

wrywerytwreywery:
Alright, @scopatz, I think this is ready for another look.  Let me know what you think.  If we're alright with this representation, then IO redirection will be next on my list, along with parentheses and subprocess grouping (with `and`, `or`, and `not`).

Merging with #223 might be a bit involved, but I am happy to help with that.

xonsh/proc.py
```diff
@@ -0,0 +1,307 @@
...
+import sys
+import platform
+
+from threading import Thread
+from subprocess import Popen, PIPE, DEVNULL, STDOUT
+from collections import Sequence
+
+from xonsh.tools import redirect_stdout, redirect_stderr
+
+iswindows = platform.system() == 'Windows'
```
scopatz: PEP8: global variables should be named in all uppercase.  `IS_WINDOWS`.  Also a better place for this might be tools.py, because I can see it being used by many modules.

scopatz:
Hi @wrywerytwreywery.  This looks great! Just the one note on iswindows and I think that this is good to go.

scopatz:
I can fix that up, too, if you'd like.

wrywerytwreywery:
No problem.  I think having it in `tools` is the right place.  I changed things to match what @rbrewer123 was doing in his Windows updates.  So `tools` now has variables: `ON_WINDOWS`, `ON_MAC`, and `ON_POSIX`.

scopatz:
Awesome job @wrywerytwreywery! Thanks a ton!

PR #227, opened 2015-05-13
small fix from ProcProxy changes

wrywerytwreywery:
The first of hopefully not too many :)

We want to wait for the active job regardless of whether the command we have just run is a Python function or not (this fixes an issue where `fg`, for example, would go into the first branch, and we would not end up actually waiting for the job that we had just `fg`-ed to finish).


scopatz:
Thanks!

PR #229, opened 2015-05-14
Prompt toolkit as alternative shell

anula:
Sorry that it took so long, but here is the first version of xonsh using prompt_toolkit.
First of all it introduces a new env variable: $PROMPT_TOOLKIT_SHELL and prompt_toolkit version of xonsh is launched only if this one is set to True and prompt_toolkit can be imported. Otherwise it runs readline version. This variable is by default set to False.
Prompt_toolkit shell also introduces another variable: $LEXER which specifies what lexer should be used to highlight syntax (None by default).
So, for example, following ~/.xonshrc will make xonsh use prompt_toolkit version of the shell with syntax highlighting:
```
$PROMPT_TOOLKIT_SHELL = True
from xonsh.pyghooks import XonshLexer
$LEXER = XonshLexer
```
Completion also works and generally it has most of the functionality of readline shell. 
Among things not yet done there is indentation: in readline shell <Shift-TAB> is used for it, but here this is already binded to iterating the list of possible completions in reversed order.
I can easily rebind it (for scrolling through a list of completions you can also use Ctrl-N and Ctrl-P like in vim for example) or I can use normal <TAB> for indentation - I would detect if situation indicates that indent is needed or rather autocompletion.
Another thing: I can probably change appearance of autocompletion options although I am not yet sure to what extend.
Also, is there a place where all environment variables are declared together with their default values? It seems a bit error prone if each time you want to get a variable you need to provide default by yourself.
Maybe it would be also good idea to change name $PROMPT_TOOLKIT_SHELL to something like $ADVANCED_SHELL or $ENHANCED_SHELL.
Now I am waiting for your opinions ;)

### DIFF
--- a/xonsh/history.py
+++ b/xonsh/history.py
@@ -1,0 +1,89 @@
+"""History object for use with prompt_toolkit."""
+import os
+
+from prompt_toolkit.history import History
+
+
+class LimitedFileHistory(History):
+    """History class that keeps entries in file with limit on number of those.
+
+    It handles only one-line entries.
+    """
+
+    def __init__(self):
+        """Initializes history object."""
+        super().__init__()
+        self.new_entries = []
+        self.old_history = []
+
+    def append(self, entry):
+        """Appends new entry to the history.
+
+        Entry sould be a one-liner.
+        """
+        super().append(entry)
+        self.new_entries.append(entry)
+
+    def read_history_file(self, filename):
+        """Reads history from given file into memory.
+
+        It first discards all history entries that were read by this function
+        before, and then tries to read entries from filename as history of
+        commands that happend before current session.
+        Entries that were appendend in current session are left unharmed.
+
+        Parameters
+        ----------
+        filename : str
+        Path to history file.
+        """
+        self.old_history = []
+        self._load(self.old_history, filename)
+        self.strings = self.old_history[:]
+        self.strings.extend(self.new_entries)
+
+
+    def save_history_to_file(self, filename, limit=-1):
+        """Saves history to file.
+
+        It first reads existing history file again, so nothing is overrided. If
+        combined number of entries from history file and current session
+        exceeds limit old entries are dropped.
+        Not thread safe.
+
+        Parameters
+        ----------
+        filename : str
+        Path to file to save history to.
+        limit : int
+        Limit on number of entries in history file. Negative values imply
+        unlimited history.
+        """
+        def write_list(lst, file):
+            text = ('\n'.join(lst)) + '\n'
+            file.write(text.encode('utf-8'))
+
+        if limit < 0:
+            with open(filename, 'ab') as hf:
+                write_list(new_entries, hf)
+            return
+
+        new_history = []
+        self._load(new_history, filename)
+
+        if len(new_history) + len(self.new_entries) <= limit:
+            with open(filename, 'ab') as hf:
+                write_list(self.new_entries, hf)
+        else:
+            new_history.extend(self.new_entries)
+            with open(filename, 'wb') as hf:
+                write_list(new_history[-limit:], hf)
+
+    def _load(self, store, filename):
+        """Loads content of file filename into list store."""
+        if os.path.exists(filename):
+            with open(filename, 'rb') as hf:
+                for line in hf:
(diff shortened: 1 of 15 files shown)


scopatz:
Hi @anula! First off, super huge thanks for putting this in! This is truly awesome. I'll leave some comments in-line too, but I wanted to respond to some of your questions.

> $LEXER which specifies what lexer should be used to highlight syntax (None by default).

This should probably have a more specific name, like `$HIGHLIGHTING_LEXER`, since there is another lexer for the xonsh language itself.  Also, since there is only one syntax highlighting lexer, it is probably OK to set this equal to `XonshLexer` if pygments is installed.

> Also, is there a place where all environment variables are declared together with their default values?

There is the `BASE_ENV` variable in environ.py that can be used for this https://github.com/scopatz/xonsh/blob/master/xonsh/environ.py#L125

> Maybe it would be also good idea to change name $PROMPT_TOOLKIT_SHELL

I kinda like the name `$PROMPT_TOOLKIT_SHELL`.

xonsh/history.py
```diff
@@ -0,0 +1,82 @@
+import os
```
scopatz: This module could use a simple docstring.

xonsh/prompt_toolkit_shell.py
```diff
@@ -0,0 +1,113 @@
...
+from prompt_toolkit.shortcuts import create_cli, create_eventloop
+from pygments.token import Token
+
+from xonsh.base_shell import BaseShell
+from xonsh.history import LimitedFileHistory
+from xonsh.pyghooks import XonshLexer
+from xonsh.tools import format_prompt_for_prompt_toolkit
+from xonsh.prompt_toolkit_completer import PromptToolkitCompleter
+
+from prompt_toolkit.key_binding.manager import KeyBindingManager
```
scopatz: PEP8 please move this import up to be next to the shortcuts importer above. Thanks!

scopatz:
Also, this code looks really beautiful!  I only had the couple of comments and what I posted before and I think that this should be ready to merge in.  The only other thing I can think of is that it might be nice to be able to set the shell from the command line (in main.py), but I would almost like that to be a separate PR.

scopatz:
Also can you add some autodoc hooks to the docs, please, for the modules you added? Thanks!

scopatz:
Also, note that the most recent merge to master for Windows has caused merge conflicts with this PR. Sorry!

anula:
No problem, I will resolve them.
Thanks for the comments, can you also tell me what key should I bind for indentation? I am kinda used to  using <TAB> for indentation and I think it should be possible (works in ptpython) , although it may be a bit tricky because we are not using prompt_toolkit's multiline mode (yet), so maybe it would be easier to just rebind <Shift-TAB>.
I can also leave this change for another CL because this one is already a bit big.

scopatz:
> Thanks for the comments, can you also tell me what key should I bind for indentation?

For readline we are currently using Shift-Tab for indentation. 

> I can also leave this change for another CL because this one is already a bit big.

That would be fine too!

anula:
Ok, I rebased on top of master, fixed all things you pointed out, added doc hooks (I hope that I've done it correctly), and squashed commits into one.

> Also, since there is only one syntax highlighting lexer, it is probably OK to set this equal to XonshLexer if pygments is installed.

So, the highlighting provided by XonshLexer can sometimes be a bit annoying, so I thought that it may be better to make it None by default, but maybe I am wrong.

I will add a key for indentation in another PR, the same thing for setting the shell from command line.

If all that is ok with you it is ready to merge, otherwise just let me know!

scopatz:
> So, the highlighting provided by XonshLexer can sometimes be a bit annoying, so I thought that it may be better to make it None by default, but maybe I am wrong.

That is fine.

scopatz:
Thanks a ton @anula! This is great.  I have noticed that this seems to have broken git completion in readline mode... would you mind taking a look at that too, please?  Thanks!

anula:
Sure, but can you elaborate a bit? Just autocompletion for commands starting with git seems to be working for me...

scopatz:
Opps your right @anula :)  False alarm! Everything is working normally on my machine.

PR #237, opened 2015-05-16
windows: doc updates

rbrewer123:
I'm having some trouble getting the right sphinx environment to build the docs (I think sphinx is defaulting to python2 instead of python3), so I haven't tested these.  Thought I would put them out there anyway for comments.


scopatz:
Thanks for the docs!

Issue #234, opened 2015-05-16
windows: enable git branch checking


rbrewer123:
Git branch checking is currently disabled on Windows due to it having really poor performance.  My first  implementation (the one that was too slow) attempted to run 'git' to check availability, since 'which' is not available on Windows.  A simple Python implementation of 'which' is probably the first thing to try to see if acceptable performance is achieved.  

If that's not enough, perhaps some very simple caching of the last branch state would be enough, with the cache being invalidated if any command is run.  For example, I will often press 'enter' several times in a row on the command line to create some visual space.  If the cached git branch were displayed in that case, the perception would be that it's very fast.  When I enter any command at all besides a blank line, if the prompt took a little longer it may not be as noticeable as in the case when I press enter several times in a row.

wrywerytwreywery:
Hmmm my initial feeling would have been just to skip the other checks on Windows and do something like the following (from `current branch` in `env.py`):
```
        try:                                                                    
            cmd = ['git', 'rev-parse', '--abbrev-ref', 'HEAD']                  
            s = subprocess.check_output(cmd,                                    
                                        stderr=subprocess.PIPE,                 
                                        cwd=cwd,                                
                                        universal_newlines=True)                
            s = s.strip()                                                       
            if len(s) > 0:                                                      
                branch = s                                                      
        except subprocess.CalledProcessError:                                   
            pass                            
```

But that is slow on Windows?

One kludgy option (not sure how reliable it would be, but it might be faster) is to climb to directory tree looking for `.git/HEAD` and try to read the current branch from that.

rbrewer123:
I hadn't tried your first idea.  My quick test is that although it's still slower than I'd prefer, it's within the realm of sanity and may be acceptable on faster hardware.  I'll work on a patch but it may be a couple days to get back to it.

rbrewer123:
I'm getting faster at the process, so #238 is available.

rbrewer123:
Closing this since branch checking was enabled on Windows (with marginal performance) by #238.  If performance is still an issue, we can open a separate issue later.

PR #238, opened 2015-05-16
windows: enable git branch checking in prompt

rbrewer123:
Initial solution to #234 as proposed by @wrywerytwreywery.  I will probably still remove it from my prompt in my .xonshrc on my Windows tablet, but it may be acceptably fast on better hardware.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -18,10 +18,6 @@ from xonsh.dirstack import _get_cwd
 def ensure_git(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
-        # getting the branch was slow on windows, disabling for now.
-        if ON_WINDOWS:
-            return ''
-
         # Get cwd or bail
         kwargs['cwd'] = kwargs.get('cwd', _get_cwd())
         if kwargs['cwd'] is None:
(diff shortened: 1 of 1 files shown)


xonsh/environ.py
```diff
@@ -38,7 +34,11 @@ def wrapper(*args, **kwargs):
         except subprocess.CalledProcessError:
             return
         return func(*args, **kwargs)
-    return wrapper
+    if ON_WINDOWS:
```
wrywerytwreywery: If we're on Windows, we don't even need to go to the trouble of defining the `wrapper` function if we're just going to throw it away.  I might recommend sliding this check up.  So rather than defining the function first and then doing the Windows check, we could do:
```
if ON_WINDOWS:
    return func

@wraps(func)
def wrapper
...
```

wrywerytwreywery:
Thanks for putting this in!  I ran some tests and everything looks good.  Just one quick thought (commented inline).

xonsh/environ.py
scopatz: Yep, I agree with this.

rbrewer123:
Updated to address the comments and merged from master.  get_dirty_working_directory() also seemed to have a bug that I fixed.

rbrewer123:
@scopatz or @wrywerytwreywery this is ready for another look.

wrywerytwreywery:
Looks great, @rbrewer123; thanks again!

It was painfully slow on some big repos my Windows machine, but removing the check for a repo being "dirty" helped dramatically (might be helpful if speed is still an issue on your setup).

PR #239, opened 2015-05-17
Add prompt output for mercurial repository branches and bookmarks

ngoldbaum:
This expands the existing default prompt to output hg branch and bookmark information.  I've also taken the opportunity to generalize the existing prompt logic that deal with VCS information so it should be easier to add support for e.g. SVN in the future if anyone wants or needs it.

~~I've implemented this using the python bindings to the mercurial command server (python-hglib on pypi):~~

~~http://mercurial.selenic.com/wiki/PythonHglib~~

~~This is an LGPL library so no worries about mercurial being GPL.~~

~~This functionality depends on installing an optional package from pypi. Arguably, it's one that hg users will likely have installed, but it does add a new soft dependency.  I don't know how sensitive xonsh is to adding dependencies from pypi.~~

I'm happy to bikeshed the inclusion of bookmark information or the way I'm displaying it.

By the way, xonsh is awesome! Now that yt works under python3 I can use yt inside of xonsh :+1:

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -14,6 +14,10 @@ from xonsh import __version__ as XONSH_VERSION
 from xonsh.tools import TERM_COLORS, ON_WINDOWS, ON_MAC
 from xonsh.dirstack import _get_cwd
 
+try:
+    import hglib
+except ImportError:
+    hglib = None
 
 def ensure_git(func):
     @wraps(func)
@@ -40,13 +44,32 @@ def ensure_git(func):
         return func(*args, **kwargs)
     return wrapper
 
+def ensure_hg(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        kwargs['cwd'] = kwargs.get('cwd', _get_cwd())
+        if kwargs['cwd'] is None:
+            return
+
+        # bail if hglib is not installed
+        if hglib is None:
+            return
+
+        try:
+            c = hglib.open(kwargs['cwd'])
+            c.tip()
+        except OSError:
+            # hglib is installed but hg isn't, so bail
+            return
+        except hglib.error.CommandError:
+            # cwd is not an hg repository
+            return
+
+        return func(*args, **kwargs)
+    return wrapper
 
 @ensure_git
-def current_branch(cwd=None, pad=True):
-    """Gets the branch for a current working directory. Returns None
-    if the cwd is not a repository.  This currently only works for git,
-    but should be extended in the future.
-    """
+def get_git_branch(cwd=None):
     branch = None
     prompt_scripts = ['/usr/lib/git-core/git-sh-prompt',
                       '/usr/local/etc/bash_completion.d/git-prompt.sh']
@@ -77,17 +100,40 @@ def current_branch(cwd=None, pad=True):
         except subprocess.CalledProcessError:
             pass
 
+    return branch
+
+@ensure_hg
+def get_hg_branch(cwd=None):
+    branch = None
+
+    client = hglib.open(cwd)
+    branch = client.branch().decode('utf8')
+    bookmarks, current_ind = client.bookmarks()
+    if current_ind == -1:
+        active_bookmark = None
+    else:
+        active_bookmark = bookmarks[current_ind][0].decode('utf8')
+
+    if active_bookmark is not None:
+        return "{0}, {1}".format(branch, active_bookmark)
+
+    return branch
+
+def current_branch(pad=True):
+    """Gets the branch for a current working directory. Returns None
+    if the cwd is not a repository.  This currently only works for git and hg
+    and should be extended in the future.
+    """
+
+    branch = get_git_branch() or get_hg_branch()
+
     if pad and branch is not None:
         branch = ' ' + branch
+
     return branch or ''
 
-
 @ensure_git
-def branch_dirty(cwd=None):
-    """Returns a boolean as to whether there are uncommitted files on the
-    current branch of a git repository (if there is one).  Currently only
-    supports git, but could be extended in the future.
-    """
+def git_dirty_working_directory(cwd):
     try:
         cmd = ['git', 'status', '--porcelain']
         s = subprocess.check_output(cmd,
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Let me know if the merge commit needs to be cleaned up as well.  Fixing it requires reading the git documentation, and those pages always look like this to me: http://git-man-page-generator.lokaltog.net/

xonsh/environ.py
```diff
@@ -201,13 +206,32 @@ def wrapper(*args, **kwargs):
...
+        if hglib is None:
+            return
+
+        try:
+            c = hglib.open(kwargs['cwd'])
+            c.tip()
+        except OSError:
+            # hglib is installed but hg isn't, so bail
+            return
+        except hglib.error.CommandError:
```
wrywerytwreywery: I am actually getting an `hglib.error.ServerError` when I am not in a directory that contains a mercurial repo.  So we may need to modify this line to catch both of those types of Exceptions.

wrywerytwreywery: Full traceback in case it is helpful:
```
    main()
  File "/usr/local/lib/python3.4/site-packages/xonsh/main.py", line 78, in main
    shell.cmdloop()
  File "/usr/local/lib/python3.4/site-packages/xonsh/readline_shell.py", line 114, in cmdloop
    super().cmdloop(intro=intro)
  File "/usr/local/lib/python3.4/cmd.py", line 126, in cmdloop
    line = input(self.prompt)
  File "/usr/local/lib/python3.4/site-packages/xonsh/readline_shell.py", line 127, in prompt
    return super().prompt
  File "/usr/local/lib/python3.4/site-packages/xonsh/base_shell.py", line 105, in prompt
    p = format_prompt(p)
  File "/usr/local/lib/python3.4/site-packages/xonsh/environ.py", line 366, in format_prompt
    for (k, v) in fmt.items()
  File "/usr/local/lib/python3.4/site-packages/xonsh/environ.py", line 367, in <dictcomp>
    if k in included_names}
  File "/usr/local/lib/python3.4/site-packages/xonsh/environ.py", line 290, in current_branch
    branch = get_git_branch() or get_hg_branch()
  File "/usr/local/lib/python3.4/site-packages/xonsh/environ.py", line 221, in wrapper
    c = hglib.open(kwargs['cwd'])
  File "/usr/local/lib/python3.4/site-packages/hglib/__init__.py", line 11, in open
    return client.hgclient(path, encoding, configs)
  File "/usr/local/lib/python3.4/site-packages/hglib/client.py", line 63, in __init__
    self.open()
  File "/usr/local/lib/python3.4/site-packages/hglib/client.py", line 193, in open
    self._readhello()
  File "/usr/local/lib/python3.4/site-packages/hglib/client.py", line 75, in _readhello
    ch, msg = self._readchannel()
  File "/usr/local/lib/python3.4/site-packages/hglib/client.py", line 99, in _readchannel
    raise error.ServerError()
hglib.error.ServerError
```

wrywerytwreywery: And also, it seems to happen even if I am inside a repo, any time I am not sitting directly at the root of the repo...

wrywerytwreywery:
Thanks for putting this in!  I tend to use Mercurial more often than Git, so this functionality would be very nice.  I made one comment inline, and I had a few other thoughts, as well.

> I don't know how sensitive xonsh is to adding dependencies from pypi.

Personally, soft dependencies don't bother me at all; as long as not having them won't break xonsh or slow it down, then I think we're fine.

That said, I think my preference in this particular case would be to invoke `hg branch` through subprocess rather than using `hglib`.  For one, this would mirror the `git` interface more closely.  More importantly, though:
* this avoids the dependency issue altogether.
* this would also get around the issue that I commented on inline.
* on my machine, starting up in a directory with no `hg` repo was pretty much instantaneous, but starting up in a directory with an `hg` repo took more than 10 seconds.  That said, running `hg` branch from bash in an `hg` repo took no time.  Not sure whether this is somehow just me (the machine I'm currently on is Debian 7, Mercurial 2.2.2, python-hglib 1.6, Python 3.4.3) or whether it will be wide-spread.

> Let me know if the merge commit needs to be cleaned up as well.

I think the only thing there is that under PEP8, function and class definitions should be preceded by two blank lines (there are a bunch of places in this file that only have one blank line preceding the functions).

> http://git-man-page-generator.lokaltog.net/

That is hilarious!

xonsh/environ.py
ngoldbaum: I didn't see that on mercurial 3.4, I'll try testing it out with a build of mercurial 2.2.

ngoldbaum:
OK, the latest commit should address all the concerns @wrywerytwreywery brought up.  Please let me know if anyone has additional comments.

ngoldbaum:
Invoking mercurial needs to spin up at python2 interpreter, which has some overhead. This can be avoided by using a mercurial daemon, e.g. [chg](https://bitbucket.org/yuja/chg).

I don't think there's much we can do to improve performance here.  But please do let me know if you are still seeing slowdowns with your big repo.

All that said, Mercurial 2.2.2 is a pretty old mercurial release from 2012.  Mercurial does rolling releases with strong backward compatibility and interoperability guarantees so it's usually not a problem to update from debian backports or maintain a custom installation in your home directory. If you update you will likely see significant performance improvements, especially when operating on large repositories.

wrywerytwreywery:
> That said, please do let me know if you are still seeing slowdowns with your big repo.

For me, these changes make a night-and-day difference in terms of speed (much better!), and I'm glad that we can get most of the info we need without even invoking the binary!  That said, I understand that my hg is a little out of date, so my experience might not be typical.

Is this version noticeably slower for you?  If not, I'm ready to merge in.

>  I've also added a short circuit so the slowdown associated with invoking mercurial only happens when the current working directory is inside a mercurial repository.

Very nice!

> I'm happy to bikeshed the inclusion of bookmark information or the way I'm displaying it.

To me, that looks completely fine.

jzuhone:
+1

scopatz:
Thanks for putting this in @ngoldbaum! I'll let @wrywerytwreywery merge when he is happy with it.  Looks good to me though. :)

ngoldbaum:
Yup, this is about as fast as my original version with my setup.  Glad this was easy!

wrywerytwreywery:
All set; thanks again!

scopatz:
Has anyone else noticed that this broke alias lookup?  I am struggling to figure out how, but my `ls` is no longer colorized.

scopatz:
ls

scopatz:
Or maybe this was from the windows shebang...

wrywerytwreywery:
Shoot, yes.  That has to be the Windows shebang.  Should be an easy fix, though, I think.

scopatz:
Yeah, I think it is just a matter of checking if the alias exists first.

PR #240, opened 2015-05-17
IO Redirection

wrywerytwreywery:
This is still a work in progress, but I figured it would be good to have it up here for discussion anyway.  This _almost_ implements the syntax discussed in #189.

What is still not working is redirecting multiple streams to the same location (so `all` and `err>out` are not working yet, though they do parse properly).


scopatz:
Hi @wrywerytwreywery - This looks pretty great as is!  Some tutorial docs are probably needed for this too.  Also, it looks like this now needs a merge from master.  Please let me know when you think it is ready for another look.

wrywerytwreywery:
> This looks pretty great as is!

Glad to hear it.  I think right now it should work basically as a drop-in replacement for what we currently have, and hopefully I'll have the other pieces working before 0.1.6!

> Some tutorial docs are probably needed for this too.

Definitely; will do once I have things working.

I'll let you know when it's ready for another look.

scopatz:
Sounds great!

wrywerytwreywery:
Alright, @scopatz, I think this is ready for another look now.  I fixed the issue with not being able to merge stdout and stderr, and fixed another small bug with error reporting (to make sure a newline follows an error message in the shell).  Some of it is kind of a kludge, but it seems to work for me so far!

I also took a first stab at docs.  They are not great, but they might serve until something better comes along.

Any feedback would be appreciated.

@gthole, you mentioned compatability with some Vim plugins as a concern.  Would you mind testing this with those plugins if you have a chance?  Or letting me know which ones are worth trying to set up/test on my own machine?

xonsh/lexer.py
```diff
@@ -30,7 +30,7 @@
         # basic operators
         '+': 'PLUS', '-': 'MINUS', '*': 'TIMES', '/': 'DIVIDE',
         '//': 'DOUBLEDIV', '%': 'MOD', '**': 'POW', '|': 'PIPE',
-        '&': 'AMPERSAND', '~': 'TILDE', '^': 'XOR', '<<': 'LSHIFT',
+        '~': 'TILDE', '^': 'XOR', '<<': 'LSHIFT',
```
scopatz: Why did you need to get rid of the amersand token?  `&` is a valid Python operator....

xonsh/lexer.py
```diff
@@ -43,12 +43,13 @@
 for (op, type) in _op_map.items():
     token_map[(tokenize.OP, op)] = type
 
-token_map[tokenize.NUMBER] = 'NUMBER'
 token_map[tokenize.STRING] = 'STRING'
 token_map[tokenize.NEWLINE] = 'NEWLINE'
 token_map[tokenize.INDENT] = 'INDENT'
 token_map[tokenize.DEDENT] = 'DEDENT'
 
+_REDIRECT_NAMES = {'out', 'err', 'all', 'o', 'e', 'a'}
```
scopatz: This could probably be a frozenset.

scopatz:
Hi @wrywerytwreywery - This looks pretty great.  Could you add some tests for the parser as well, please? Thanks!

xonsh/lexer.py
wrywerytwreywery: It's not actually gone; it's just being handled in a different way because it can now have some special meanings beyond just ``AMPERSAND`` (that is, it might be a ``IOREDIRECT``).

The ``handle_ampersand`` function is what is actually responsible for handling it, and I believe that is working as expected (``&`` in Python mode still works).

```
>>> 7 & 2
2
>>> 7 &> 2
xonsh: subprocess mode: command not found: 7
```

xonsh/lexer.py
wrywerytwreywery: Yep.  I can change that.  There are several other things in built_ins.py that could also be frozensets.

wrywerytwreywery:
Yep, can do.

gthole:
@wrywerytwreywery The vim plugins that had issues before (syntastic, git gutter) both now work delightfully on this branch.  I also did a little tooling around with the shell redirection and love the `out`, `err`, `all` syntax.

wrywerytwreywery:
@gthole Fantastic; glad to hear it!

wrywerytwreywery:
Alright, @scopatz, I think this is ready for another look!  Any feedback would be appreciated.

scopatz:
Hi @wrywerytwreywery - this looks great.  However, I noticed a problem with `err>out` not working as expected when `e>o` does:

```console
scopatz@athenaie ~ $ time ls err>out > t.txt
Unrecognized redirection command: err>out
scopatz@athenaie ~ $ time ls e>o > t.txt
scopatz@athenaie ~ $ 
```

Any ideas?

wrywerytwreywery:
That was just a silly mistake.  I have a set of all valid `e>o` commands.  In it, I had accidentally reproduced `err>o` twice and not included `err>out`.  I changed it so that that set is created programmatically (the right way to go all along...), so that shouldn't be an issue anymore.

scopatz:
Thanks a ton @wrywerytwreywery!  Sorry for the delay!

PR #241, opened 2015-05-17
Windows Shebang

### LINKED ISSUE #233: windows: shebang (#!) scripts don't work
On Windows, running shebang scripts from xonsh generally doesn't work.  My preliminary investigation shows that _is_runnable_name() may be too restrictive and preventing the shebang-checking code from ever executing.  On Linux this is no big deal since Linux intrinsically can run shebang lines.  On Windows, this needs to be fixed.

I think a good answer here is for _is_runnable_name() to do an effective 'which' command to see if anything on the PATH may be runnable.  

For increased compatibility with msysGit, cygwin, and other unix-compatibility schemes on Windows, I think '#!' handling on Windows should additionally allow unix-style pathnames.  A simple way to do this is to check for a '#!' path starting with '/' and just take the basename of the path, relying on PATH lookup to find the right thing to execute.  Many '#!' scripts use '/usr/bin/env' anyway to perform a similar lookup of the correct file on the PATH.

Does this seem like a reasonable direction to go?

wrywerytwreywery:
This is a hack to get shebang lines working on Windows (should fix #233).  It does change the handling in *nix, too, but I think not in a way that has any ill effects.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -208,7 +208,7 @@ def _get_runnable_name(fname):
     if os.path.isfile(fname) and fname != os.path.basename(fname):
         return fname
     for d in builtins.__xonsh_env__['PATH']:
-        if fname in os.listdir(d):
+        if os.path.isdir(d) and fname in os.listdir(d):
             return os.path.join(d, fname)
     return None
 
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -227,7 +227,9 @@ def _is_binary(fname, limit=80):
 
 
 def _un_shebang(x):
-    if x.startswith('/usr/bin') or x.startswith('/usr/local/bin') or x.startswith('/bin'):
+    if x == '/usr/bin/env':
+        return []
+    elif any(x.startswith(i) for i in ['/usr/bin', '/usr/local/bin', '/bin']):
         x = os.path.basename(x)
     elif x.endswith('python') or x.endswith('python.exe'):
         x = 'python'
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks @wrywerytwreywery!

xonsh/built_ins.py
```diff
@@ -301,9 +324,10 @@ def run_subproc(cmds, captured=True):
         stdout = last_stdout if cmd is last_cmd else PIPE
         uninew = cmd is last_cmd
         alias = builtins.aliases.get(cmd[0], None)
-        if _is_runnable_name(cmd[0]):
+        n = _get_runnable_name(cmd[0])
+        if n is not None:
             try:
-                aliased_cmd = get_script_subproc_command(cmd[0], cmd[1:])
+                aliased_cmd = get_script_subproc_command(n, cmd[1:])
```
scopatz: Ahh this is probably messing up running aliases...

xonsh/built_ins.py
```diff
@@ -301,9 +324,10 @@ def run_subproc(cmds, captured=True):
         stdout = last_stdout if cmd is last_cmd else PIPE
         uninew = cmd is last_cmd
         alias = builtins.aliases.get(cmd[0], None)
-        if _is_runnable_name(cmd[0]):
+        n = _get_runnable_name(cmd[0])
```
scopatz: Or rather this is, because it doesn't check if the alias exists first.

wrywerytwreywery: Yeah, this is the one.

PR #242, opened 2015-05-17
proposed fix for alias bug from #241

wrywerytwreywery:
I think this should do it (as you suggested, @scopatz, just some shuffling of the checks).


scopatz:
Works for me! Thanks for the quick turn around :)

PR #244, opened 2015-05-17
Aliases as Environment Variable

adqm:
This changeset is a reorganization of the code to store aliases in an environment variable (`$ALIASES`) instead of in the global namespace.  I think I caught all the places we would need to make this change.

I am not sure whether there was a reason to have the aliases live where they currently do, but it "feels right" to me to have them in the environment (away from the Python execution).  Non-trivial downside, of course, is that everyone has to rewrite their `.xonshrc`...

Happy to discuss more or to make changes as necessary.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -167,6 +167,9 @@ They can be seen in the table below:
 ================== ============================= ================================
 variable           default                       description
 ================== ============================= ================================
+ALIASES            xonsh.aliases.DEFAULT_ALIASES A dictionary containing aliases.
+                                                 See `Aliases`_ below for more
+                                                 information.
 PROMPT             xonsh.environ.DEFAULT_PROMPT  The prompt text.  May contain
                                                  keyword arguments which are
                                                  auto-formatted (see `Customizing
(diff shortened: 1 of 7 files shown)


rbrewer123:
Syntactically, I don't have a strong opinion either way.  I agree that less pollution of the normal Python namespace is better.

With this patch, would $ALIASES be passed to subprocesses?  I would be a little concerned about data leakage to subprocesses, and also that with a large set of aliases there may be size limitations on passing them all as a single environment variable.  Preventing $ALIASES from being sent would allay those concerns, but then it would seem they aren't really an environment variable at that point.

wrywerytwreywery:
> With this patch, would $ALIASES be passed to subprocesses?

Nope.  When we detype the environment to pass it to subprocesses, callable values and dictionary-like values (such as the aliases mapping) are skipped:
```
...
        for key, val in self._d.items():
            if callable(val) or isinstance(val, MutableMapping):
                continue
...
```

>  but then it would seem they aren't really an environment variable at that point.

That is true enough.  But I think we already have several things in the environment that have that same property, so I think it's still a win to get it out of the Python namespace.

An alternative would be to leave it where it is but make it "private" (e.g., `__aliases__`), but, at least for me, it's a pain to type out those underscores.  Also, I think aliases are closer to the "shell-side" of xonsh than to the "python-side," so referencing them with a shell-like `$ALIASES` feels right to me.

scopatz:
Hi All, Thanks for putting this in @wrywerytwreywery.  While I wouldn't mind adding `$ALIASES`, I am somewhat against removing `aliases`.  The reasons that I made `aliases` as builtin rather than the environment variable in the first place are as follows:

1. It is something that only may be used from Python mode, since it is a dictionary
2. It is part of the xonsh interface, and thus should be public (ie not `__aliases__`)
3. The only subprocesses which xonsh might spawn where aliasing would be valuable are other xonsh instances (and maybe with some tricks, other shells too).  It really isn't part of the env in the same way that say `$PATH` is. And even in the case where a subprocess might use aliases (another xonsh instance), they will reread in the xonshrc file anyway.  

To @rbrewer123 point, I don't think that it really pollutes the normal execution context since it goes into the builtins.  

So I am OK with adding another reference to this data structure.  This basically amounts to doing the following in the xonshrc:

```python
$ALIASES = aliases
```

But I do think that this is a core part of the xonsh language and should fundamentally be a builtin.

PR #245, opened 2015-05-18
windows: use PATHEXT to find runnable files and skip shebang parsing

rbrewer123:
On Windows, xonsh couldn't find some files to run, especially script files ending in .cmd.  In addition, the shebang parsing defaults to using xonsh as the interpreter if it is a text file with no '#!' in the first line.  This caused '.cmd' files to be run by xonsh instead of the Windows cmd.exe.

This patch enables xonsh to use the PATHEXT environment variable to determine which files Windows can run directly.  Also, when a file is in PATHEXT, it skips shebang parsing for that file since Windows will be able to run the file directly.

(The above issues were manifesting when attempting to run 'ack' as installed from chocolatey.org, which provides an 'ack.cmd' file to run it).

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -208,8 +208,16 @@ def _get_runnable_name(fname):
     if os.path.isfile(fname) and fname != os.path.basename(fname):
         return fname
     for d in builtins.__xonsh_env__['PATH']:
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks @rbrewer123!

PR #248, opened 2015-05-20
Completion from man pages

zajaczajac:
This is a pull request concerning issue https://github.com/scopatz/xonsh/issues/247


xonsh/completer.py
```diff
@@ -136,10 +138,19 @@ def cmd_complete(self, cmd):
...
-        """Completes a name of a module to import"""
+        """Completes a name of a module to import."""
         modules = set(sys.modules.keys())
         return {s for s in modules if s.startswith(prefix)}
 
+    def option_complete(self, prefix, cmd):
+        """Completes an option name, basing on content of man page."""
+        manpage = subprocess.Popen(["man", cmd], stdout=subprocess.PIPE)
+        # This is a trick to get rid of reverse line feeds
+        text = subprocess.check_output(["col", "-b"], stdin=manpage.stdout).decode('utf-8')
```
scopatz: Wouldn't `man -7` be better than piping like this?

xonsh/completer.py
```diff
@@ -136,10 +138,19 @@ def cmd_complete(self, cmd):
...
+        pattern = r'^\s*(-\w|--[a-z-]+)[\s,]'
+        matches = re.findall(pattern, text, re.M)
```
scopatz: Please precompile the regex globally.

scopatz:
Hi @zajaczajac - I was thinking that it might be best to have cache of the set of options on disk that gets read in at start-up, so the entire man page need not be read in and parsed each time that tab is pressed.

zajaczajac:
Cool idea - I'd like to implement this

xonsh/completer.py
zajaczajac: Well, actually I was looking for such an option in man manual. But I'm developing under OS X and here I have man version 1.6c (which was preinstalled) and there is no such option.
So I believe this version with pipes is going to work on greater number of machines. Maybe there's some other clean & nice solution?

scopatz: Ahhh. Ok.  I hadn't considered the Mac OSX issue, but you are right.  What I was planning on doing originally was using PLY to write a simple lexer and parser for the troff language.  I have a 'man' branch on my repo that started this if you want to check it out...

zajaczajac:
I've commited next version, including caching

scopatz:
Hi @zajaczajac - this all seems OK, except that it doesn't actually work on my Ubuntu 14.10 machine.  When I try to tab complete an option that has a known man page but no bash completion, such as `firefox -`, I don't get a list of suggestions.  Luckily, it doesn't crash or anything. It just doesn't complete.  Does it actually work on your mac?

zajaczajac:
Now it works also on my Ubuntu 13.10. The problem was using of "DEVNULL" instead of "subprocess.DEVNULL" - it somehow worked on OS X.

scopatz:
Hmmm This still isn't working for me.  There are no errors being thrown though...

scopatz:
Also the cache file is being created...

zajaczajac:
Can you please how do you exactly test this? I tested it on my 2 machines and it works. I write for example "gzip -" and then press TAB (dash is needed).

scopatz:
So please copy over a man page file into the tests directory. You can find the location of a man page with `man -w`:

```console
scopatz@athenaie ~ $ man -w yes
/usr/share/man/man1/yes.1.gz
```

Then inside of a new `test_man.py` file, you should be able to set the `MANPATH` environment variable to the current directory:

```python
import os
os.environ['MANPATH'] = os.path.dirname(__file__)
```

Then you should be able to make an instance of the man page completer class here and test that the completion method returns what is actually in the man page.

zajaczajac:
OK, I've added a test. I had to turn it off for Windows because it doesn't work there (no man?). I also surrounded executing man with try-except.

rbrewer123:
@zajaczajac, I think this is a neat idea and I tried it out on my Ubuntu 14.04.2 box.  I'm getting the following results.  I type `nosetests -` and press tab and I see nice completion results that filter down as I add additional letters.  It's very slick.  I don't see those results in bash so I believe your ManCompleter is doing that.  However, it looks like I'm only seeing completion results for some of the options (e.g. options that take a parameter don't seem to be listed).  

I tried some other programs to see what would happen.  For `xload -` and `pdftex -` I don't see any completion results.  xload and pdftex use long options with a single `-` while nosetests' long options start with `--`.  I'm wondering if the regular expressions need to be a little more flexible to pick up more of the options.  Are you able to reproduce anything similar?  

I'm seeing the cache file get created and populated with results so that looks good.

scopatz:
Hi @zajaczajac - I thanks for all of your effort here.  Like @rbrewer123 has said, I think that this could use some work to make it more robust and capture more cases. But I was able to confirm that it is actually working in some instances.  This is a great start!

PR #249, opened 2015-05-21
Add option to run xonsh in prompt_toolkit mode from command line.

anula:
Simply adding possibility to run xonsh in "prompt toolkit mode" from command line, as @scopatz mentioned in #229 .
One can now simply run xonsh like that:
```
xonsh --prompt_toolkit_shell
```
to run it in prompt_toolkit mode.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -212,6 +212,19 @@ SUGGEST_MAX_NUM    ``5``                         xonsh will show at most this
                                                  an invalid command.  If
                                                  negative, there is no limit to
                                                  how many suggestions are shown.
+SHELL_TYPE         ``'readline'``                Which shell is used.
+                                                 Currently two shell types are
+                                                 supported: ``'readline'`` that
+                                                 is backed by python's readline
+                                                 module and ``'prompt_toolkit'``
+                                                 that uses external library of
+                                                 the same name. For using
+                                                 prompt_toolkit shell you need
+                                                 to have `prompt_toolkit
+                                                 <https://github.com/jonathanslenders/python-prompt-toolkit>`_
+                                                 library installed. To specify
+                                                 which shell should be used, do
+                                                 so in the run control file.
 ================== ============================= ================================
 
 Environment Lookup with ``${}``
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -446,7 +446,7 @@ BASE_ENV = {
     'LC_TIME': locale.setlocale(locale.LC_TIME),
     'LC_MONETARY': locale.setlocale(locale.LC_MONETARY),
     'LC_NUMERIC': locale.setlocale(locale.LC_NUMERIC),
-    'PROMPT_TOOLKIT_SHELL': False,
+    'SHELL_TYPE': 'readline',
     'HIGHLIGHTING_LEXER': None,
 }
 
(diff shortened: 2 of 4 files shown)


scopatz:
Hi @anula - sorry for taking so long to review this.  I envisioned there being a top-level option called `--shell` that can then be set to either `readline` or `prompt-toolkit` or similar.  

```console
$ xonsh --shell=prompt-toolkit
```

I think that this makes it more extensible for the long term.  I could see using this to start up a Jupyter instance, etc.  Does this make sense?

scopatz:
The code looks good, btw.  I am just not sure it is the best code for the long term.

anula:
Yes, it really sounds better. I, probably, did it under assumption that prompt_toolkit is optional, but if it changes in future your solution seems to be better.

scopatz:
Great!

scopatz:
Thanks!

anula:
Maybe I should completely remove $PROMPT_TOOLKIT_SHELL env variable? I is something that matters only during shell start, we probably don't want to change shell type during session of xonsh. Maybe it doesn't really fit for an environment variable.

scopatz:
Well, you could have a `$SHELL_TYPE` variable that is set to the same value as the command line option.

anula:
Yeah, I guess you are right. I had some doubts about having variable just to control something what happens only during the start of shell, but anyway it is good idea to have variable that tells you what kind of shell are you running. Thanks.

scopatz:
No worries.  We can always make it read-only if we really need to :).

anula:
Sorry for delay.
I have deleted $PROMPT_TOOLKIT_SHELL and added $SHELL_TYPE variable.
I also changed the command line option, and now one can run xonsh like that:
```
xonsh --shell_type=prompt_toolkit
```
for prompt_toolkit shell or like that:
```
xonsh --shell_type=readline
```
for readline.
Any thoughts?

scopatz:
Hi @anula - I get the following bug when I try to start xonsh without prompt_toolkit installed:

```python
scopatz@athenaie ~ $ bash
scopatz@athenaie ~ BA$H xonsh
  File "/home/scopatz/.local/bin/xonsh", line 3, in <module>
    main()
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/main.py", line 58, in main
    shell = Shell(**shell_kwargs)
  File "/home/scopatz/.local/lib/python3.4/site-packages/xonsh/shell.py", line 44, in __init__
    env['SHELL_TYPE']))
xonsh.tools.XonshError: None is not recognized as a shell type
scopatz@athenaie ~ BA$H 
```

scopatz:
Also, would you mind adding `SHELL_TYPE` to the docs, please?

anula:
Ups, sorry, I got tangled in a bit.
About adding it to docs, you mean putting it somewhere [here](http://xonsh.org/tutorial.html#environment-variables) or somewhere else? I can probably also add a section there about prompt_toolkit mode in general, but probably it would be better to do in another PR.

scopatz:
> About adding it to docs, you mean putting it somewhere here or somewhere else?

Yes, in table in that section there is great!

> add a section there about prompt_toolkit mode in general, but probably it would be better to do in another PR.

I agree. Another PR would be better.

anula:
Ok, bug fixed, `$SHELL_TYPE` added to docs. I hope that this time I tested it better!

scopatz:
Thanks @anula!

PR #250, opened 2015-05-22
fix issue where prompt_toolkit is always preferred

wrywerytwreywery:
This took me forever to track down.  `PROMPT_TOOLKIT_SHELL` is detyped as `"False"` (a string), whose boolean value is `True`.  So the shell was passing to the `if env['PROMPT_TOOLKIT_SHELL']` under certain conditions, even if `PROMPT_TOOLKIT_SHELL` was set to `False`.  This patch should fix, I believe.

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -8,7 +8,7 @@ from nose.tools import assert_equal, assert_true, assert_false
 from xonsh.lexer import Lexer
 from xonsh.tools import subproc_toks, subexpr_from_unbalanced, is_int, \
     always_true, always_false, ensure_string, is_env_path, str_to_env_path, \
-    env_path_to_str, escape_windows_title_string
+    env_path_to_str, escape_windows_title_string, convert_bool, is_bool
 
 LEXER = Lexer()
 LEXER.build()
@@ -151,7 +151,12 @@ def test_subexpr_from_unbalanced_parens():
 
 def test_is_int():
     yield assert_true, is_int(42)
+    yield assert_true, is_int(True)
     yield assert_false, is_int('42')
+
+def test_is_bool():
+    yield assert_true, is_int(True)
+    yield assert_false, is_int('True')
 
 def test_always_true():
     yield assert_true, always_true(42)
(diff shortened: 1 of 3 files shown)


anula:
Wow, that's surely my fault, I didn't even notice `DEFAULT_ENSURERS`. But could you tell me in what situations the problem arises? I never experienced it.

wrywerytwreywery:
I still don't know quite _why_ it happens, but it has been reproducible for me.  When I log in directly to a TTY (no X, no terminal emulator), it doesn't happen, but when I use a terminal emulator in an X session (tested on gnome-terminal and lxterminal), then it happens.  Not sure why it is detyping in that situation, but I know it is happening...

xonsh/tools.py
```diff
@@ -394,6 +394,16 @@ def is_int(x):
     return isinstance(x, int)
 
 
+def is_bool(x):
+    """Tests if something is a boolean"""
+    return isinstance(x, bool)
+
+
+def convert_bool(s):
```
scopatz: Could you add some tests for these functions please? Thanks!

wrywerytwreywery:
Alright; put in a few simple test cases for these.  Let me know if it needs anything else.

anula:
Maybe it actually won't be needed when I finish #249 ? We decided there that it is probably better to remove `PROMPT_TOOLKIT_SHELL` and replace it by something like `$SHELL_TYPE`

tests/test_tools.py
```diff
@@ -151,8 +151,13 @@ def test_subexpr_from_unbalanced_parens():
 
 def test_is_int():
     yield assert_true, is_int(42)
+    yield assert_true, is_int(True)
     yield assert_false, is_int('42')
 
+def test_is_bool():
+    yield assert_true, is_int(True)
+    yield assert_false, is_int('True')
```
scopatz: Shouldn't these be `is_bool()`?

wrywerytwreywery:
@anula: That sounds right.  I'll just close this PR, then, under the assumption that #249 will take care of it.

PR #251, opened 2015-05-22
Add possibility to insert indent in prompt_toolkit shell.

anula:
I have been playing a bit with this, and decided that when you use this kind of autocompletion that prompt_toolkit provides (that is autocompletion in a box) it is highly unintuitive for Shift-Tab to insert sudden indentation instead of simply going back in autocompletions list. I guess it is common behavior that if some key iterates forward (Tab in this case) then Shift plus the key should iterate backward.
So I decided to make Tab the key for indentation (it is just more intuitive), and now it behaves like this:
 - if there are only whitespaces before cursor position Tab inserts indent
 - otherwise it starts autocompletion.

Surely I can map <Shift-TAB> to behave like this (that it inserts indent only if there are only whitespaces) if you think that would be better, or maybe just reproduce behavior from readline shell, overriding default Shift-TAB completely.
Let me know what works better for you!

### DIFF
--- a/xonsh/prompt_toolkit_key_bindings.py
+++ b/xonsh/prompt_toolkit_key_bindings.py
@@ -1,0 +1,36 @@
+"""Key bindings for prompt_toolkit xonsh shell."""
+import builtins
+
+from prompt_toolkit.filters import Filter
+from prompt_toolkit.keys import Keys
+
+
+class TabShouldInsertIndentFilter(Filter):
+    """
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks a ton @anula (and sorry for taking a while to get the this. I am on vacation ~_~).  I agree that your description of Shift-Tab iterating backwards is better for autocompletion.

PR #252, opened 2015-05-23
Fix prompt_toolkit autocompletion bug

anula:
There was a small bug that caused autocompletion to work properly only when started at the end of a line. 
Here is appropriately small fix.


scopatz:
Thanks @anula!

PR #253, opened 2015-05-24
Subprocess mode: grouping, `and`, and `or` 

adqm:
This changeset allows grouping of subprocess-mode commands with parentheses, `and`, and `or`.  For example: `mkdir test and ls test` and `(mkdir test) and (ls test)` should both work.  This also subsumes PRs #240 and #250 (which are ready for a look but haven't yet been accepted).

More complicated groupings also work:
`(mkdir test and ls test) or ls test`
`mkdir test and (ls test or ls test)`
and so does combining these things with pipes or IO redirects.

~~There is one remaining issue that I know of: this code modifies `subproc_uncaptured` to return `True` or `False`, depending on whether the subprocess call succeeded.  I would like for this value to print _only if_ the user explicitly typed the `$[` and `]` characters, but not to print if the user just typed the command that goes between them (i.e., if xonsh inserted those tokens itself).  @scopatz, do you know whether that is possible?  I have tried a few different things (without any luck), and I will keep hacking away when I have time.~~ I think I got it!

No docs ~~or tests~~ yet, and only a few tests, because I wanted to make sure that this is a direction we actually want to go.

Aside from those issues, I think things are working.  That said, I have only done a little bit of testing myself, but more eyes would be appreciated.

### DIFF
--- a/xonsh/parser.py
+++ b/xonsh/parser.py
@@ -1636,7 +1636,7 @@ class Parser(object):
         if len(p) == 2:
             # plain-old atoms
             bt = '`'
-            if isinstance(p1, (ast.Num, ast.Str, ast.Bytes)):
+            if isinstance(p1, (ast.Num, ast.Str, ast.Bytes, ast.Call)):
                 pass
             elif p1 == 'True':
                 p1 = ast.NameConstant(value=True,
@@ -2119,9 +2119,12 @@ class Parser(object):
                             lineno=lineno,
                             col=col)
         elif p1 == '$[':
-            p0 = xonsh_call('__xonsh_subproc_uncaptured__', p2,
-                            lineno=lineno,
-                            col=col)
+            if isinstance(p2, ast.Call):
+                p0 = p2
+            else:
+                p0 = xonsh_call('__xonsh_subproc_uncaptured__', p2,
+                                lineno=lineno,
+                                col=col)
         else:
             assert False
         return p0
@@ -2195,7 +2198,9 @@ class Parser(object):
         col = self.col
         lenp = len(p)
         p1 = p[1]
-        if lenp == 2:
+        if isinstance(p1, str):
+            p0 = [ast.Str(s=p1, lineno=lineno, col_offset=col)]
+        elif lenp == 2:
             p0 = [self._subproc_cliargs(p1, lineno=lineno, col=col)]
         elif p[2] == '&':
             p0 = p1 + [ast.Str(s=p[2], lineno=lineno, col_offset=col)]
@@ -2227,6 +2232,7 @@ class Parser(object):
                         | REGEXPATH
                         | DOLLAR_NAME
                         | GT
+                        | SUBSHELL_COMMAND
                         | LT
                         | RSHIFT
                         | IOREDIRECT
--- a/xonsh/parser.py
+++ b/xonsh/parser.py
@@ -1628,9 +1628,9 @@ class Parser(object):
                 | FALSE
                 | REGEXPATH
                 | DOLLAR_NAME
-                | DOLLAR_LBRACE test RBRACE
-                | DOLLAR_LPAREN subproc RPAREN
-                | DOLLAR_LBRACKET subproc RBRACKET
+                | DOLLAR_LBRACE test SUBPROC_END_RBRACE
+                | DOLLAR_LPAREN subproc SUBPROC_END_RPAREN
+                | DOLLAR_LBRACKET subproc SUBPROC_END_RBRACKET
         """
         p1 = p[1]
         if len(p) == 2:
@@ -2231,9 +2231,9 @@ class Parser(object):
                         | RSHIFT
                         | IOREDIRECT
                         | AT_LPAREN test RPAREN
-                        | DOLLAR_LBRACE test RBRACE
-                        | DOLLAR_LPAREN subproc RPAREN
-                        | DOLLAR_LBRACKET subproc RBRACKET
+                        | DOLLAR_LBRACE test SUBPROC_END_RBRACE
+                        | DOLLAR_LPAREN subproc SUBPROC_END_RPAREN
+                        | DOLLAR_LBRACKET subproc SUBPROC_END_RBRACKET
         """
         lenp = len(p)
         p1 = p[1]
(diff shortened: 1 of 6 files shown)


wrywerytwreywery:
Noticed some more issues; still working on it.  Will post again when ready.

scopatz:
This seems like it needs a merge from master too...

scopatz:
Could you rebase or merge this off of master, please? That way the diff won't have all of the redirect stuff in it too :)

wrywerytwreywery:
Yep; just rebased.  Will let you know when I have the remaining issues fixed (it will take some gymnastics, I'm afraid...).

xonsh/parser.py
```diff
@@ -2173,43 +2173,80 @@ def _subproc_cliargs(self, args, lineno=None, col=None):
...
         """
         lineno = self.lineno
         col = self.col
         lenp = len(p)
         p1 = p[1]
         if lenp == 2:
-            p0 = [self._subproc_cliargs(p1, lineno=lineno, col=col)]
-        elif p[2] == '&':
-            p0 = p1 + [ast.Str(s=p[2], lineno=lineno, col_offset=col)]
+            p0 = [ast.Tuple(elts=[ast.Str(s='cmd', lineno=lineno, col_offset=col), self._subproc_cliargs(p1, lineno=lineno, col=col)], lineno=lineno, col_offset=col, ctx=ast.Load())]
```
scopatz: Long line.

scopatz:
At a first glance, this looks really good.  I think that the And/OrProc classes are pretty clever and intuitive.

rbrewer123:
I also like the AndProc and OrProc stuff.

wrywerytwreywery:
I thought it was pretty cute, too, but, unfortunately, it makes it hard (impossible?) to pass around the file descriptors the way I would like to, so mixing with pipes doesn't work with that setup.  I think it might be time to start over again, actually...

What I really want, I think, is that something like `$[ls and ls]` turns into  `$[ls] and $[ls]`.  That would mean that I don't need (m)any of the fancy things I put in recently, and all of this can be handled by the Python side of things.  The downside of things is that I'm not sure how piping to/from such a thing should work, but maybe pipes should latch on to their closest subproc (they take precedence over `and` and `or`).

P.S. Sorry I've been absent recently; I should be able to pick this up again in the next couple of days, and hopefully finish it off.  I'll ping when it is ready.

scopatz:
Hi @wrywerytwreywery here are a couple of thoughts - and super sorry for the delay!

> What I really want, I think, is that something like `$[ls and ls]` turns into `$[ls] and $[ls]`.

I think that there is a difference between python's `and` and suproc's `and`.  Currently, `$[]` always returns None and `$()` always returns a string.  This means that expressions like `$[ls] and $[ls]` are perfectly valid. (Though since it returns None, only the first `ls` will be executed.)  For example:

```console
scopatz@athenaie ~ $ x = $[echo]

scopatz@athenaie ~ $ x is None
True
```

For `$[]`, I guess that this could be changed to return some sort of result object where `__bool__()` is overridden to get the correct `and` and `or` and `not` behaviour.  However, you couldn't do the same for captured subprocs since they are supposed to return a string. (Or maybe you could by subclassing str.) 

I had always sort of envisioned that this would work with the operators being in the subprocess, ie `$[ls and ls]`.  This is more consitent with the way piping and redirection work in that we do `$[ls | grep x]` not `$[ls] | $[grep x]`.  Unless I am mistaken, I believe that `built_ins.run_subproc()` takes a `cmds` argument like:

```python
[['ls'], '|', ['grep', 'x']]
```

And so the ops here could be:

```python
[['ls'], 'and', ['ls']]
```

That said I am in favor of whatever works and we can always change it in the future if it needs to be :)

krader1961:
This is really old, as in almost six months, and no longer cleanly merges. Having looked at a few proposed pulls of this nature on this project I can't help but think that the project needs something like the Python PEP mechanism for non-trivial enhancement requests.

@adqm should this pull request be closed or do you believe it is still viable? Would it be better to close this and continue the discussion in a feature request issue?

adqm:
I think the idea is still viable, but it requires a lot of messing with low-level things.  I had two completely separate rewrites both come very close to working, but both had little issues (in particular, with piping and IO redirection of the combined statements) that were very difficult to resolve.  I haven't had time to work on this, and probably won't for a while (at least until mid-December).  It would be fine to close this PR, or to leave it open and I may come back to it when I have a chance; whatever best helps keep things organized.

krader1961:
On Thu, Nov 12, 2015 at 4:07 AM, adam j hartz <notifications@github.com>
wrote:

> I think the idea is still viable, but it requires a lot of messing with
> low-level things. I had two completely separate rewrites both come very
> close to working, but both had little issues (in particular, with piping
> and IO redirection of the combined statements) that were very difficult to
> resolve. I haven't had time to work on this, and probably won't for a while
> (at least until mid-December). It would be fine to close this PR, or to
> leave it open and I may come back to it when I have a chance; whatever best
> helps keep things organized.
>

I can't speak for @scopatz but personally I don't like seeing pull requests
in the queue that have had no activity in more than roughly four weeks
(i.e., a month). Also, non-trivial fixes and all enhancements should have
an associated issue, IMHO. So my recommendation is to close this PR and
open an issue that describes what you intend to implement and why (i.e.,
the rationale for the change). I'd also include a reference to this PR so
people can easily find your initial attempts at implementing the
enhancement.

-- 
Kurtis Rader
Caretaker of the exceptional canines Junior and Hank

scopatz:
While this is fairly stale at this point, I actually prefer for the opener of the PR to close it, unless it is truly no longer applicable. So if @adqm whishes to close this, that is fine.  The relevant issue for this is #137.

adqm:
I am going to start fresh with this.  I think there were some good ideas here, but it's pretty messy and _way_ out of date, and it still had some bugs.

scopatz:
Thanks @adqm, a couple of weeks ago, I had started to try to work on this myself. I got through the parsing bit, I think.  In any event, if you don't want to go from total scratch, check out the `andor` branch.

adqm:
> It may be hard to debug this without seeing your code.  If we can't figure it out here, then please send us your code (in a private note here, or via e-mail) and we can take a closer look.

Sounds good, thanks!

If you had your heart set on this, then by all means, continue; if not, I'm happy to dig back in (kind of looking forward to it as a way to get reacquainted with the codebase).

scopatz:
My heart is not set on this, and I'd really love it if you or someone else were to take it over :)

adqm:
Okay, I'll call dibs.  Will post another PR when I have something working, or close to it.

PR #255, opened 2015-05-27
some minor cli changes to shell type

scopatz:
Just some quick rearrangement, made the option be `--shell-type` for consistency with other options, and turned choices into a tuple.  Probably for @anula to review and merge since this is her code :)


anula:
Looks good to me. Thanks @scopatz !

PR #258, opened 2015-06-03
windows: more cmd.exe builtins as aliases

rbrewer123:
Many commands that I consider part of the Windows command line are in fact builtins of the ``cmd.exe`` shell, e.g. ``dir``, ``copy``, ``move``, ``md``.  This PR creates a default alias for nearly all of them that don't overlap with existing functionality of xonsh (such as ``chdir``, ``pushd``, etc.).

The specific aliases are easily seen in the changeset.


wrywerytwreywery:
Thanks for this, @rbrewer123!

A couple of thoughts:
  * For commands that share names with *nix programs (`mkdir`, `rmdir`, `time`), do we want to shadow those?  Or is there a different naming convention we can use?  Or maybe a switch in `main.py` to set which should be preferred?

  I just know that if I were using xonsh on Windows, I would expect, for example, `time`, to be the *nix variant, and would be disappointed if I got the Windows version.  Someone else might feel the opposite, so I think it might be good to allow both, depending on some setting.

  I could imagine aliasing `mkdir.exe` to `win_mkdir` and `mkdir` to `posix_mkdir`, and then letting xonsh choose which to use based on whether we are trying to emulate Windows or POSIX (again, this could be set via a command line flag, maybe?  Or an environment variable?).

*  `WINDOWS_CMD_ALIASES` might want to be a tuple instead of a list (or maybe even a set or frozen set), since we don't really care about order and we don't need a mutable structure.

rbrewer123:
@wrywerytwreywery, those are good points.  Alas, I'm more comfortable at the Linux command line and am using mSysGit to provide a number of Linux commands under Windows, so I'm probably not the ideal test case either.  

Another option to consider would be a sample ``.xonshrc`` for Windows that sets up these aliases.  Since my motivation is to make it easier for a new xonsh user on Windows to get started, a sample ``.xonshrc`` might be more explicit and encourage them to customize it themselves.  I could add some better comments in the sample ``.xonshrc`` about commenting out ``mkdir`` et al for different effects.

I think part of what is confusing is that there is no ``mkdir.exe`` on Windows by default; it's a builtin of ``cmd.exe``.  mSysGit  provides ``mkdir.exe`` which is the POSIX version.  So I would be a little concerned that aliasing the builtin to ``win_mkdir`` and the exe to ``posix_mkdir`` and then aliasing one of them back to ``mkdir`` would add to the confusion.  I guess I'm more in favor of either aliasing ``mkdir`` to the builtin or not aliasing it so the exe is picked up (and different users will choose differently).

I did some tests to get more data points for discussion:

* ``cmd.exe`` uses its ``mkdir`` builtin instead of the mSysGit ``mkdir.exe`` I have installed
* mSysGit's bash uses the mSysGit ``mkdir.exe``

I'm now leaning toward the sample ``.xonshrc`` but I welcome more discussion.  Is that a path you'd like to go down?

I'll change the type of ``WINDOWS_CMD_ALIASES`` to be more explicit.

wrywerytwreywery:
Sorry it took so long to come back to this.  Thinking about it again, what you have is fine in that regard (I misunderstood what was happening "under the hood" there), so I think that what I had originally suggested for handling the naming issues doesn't actually make sense.

I think the only thing I would recommend would be to change the type of `WINDOWS_DEFAULT_ALIASES`, and then I'm happy to merge this in (or I'm happy just to make that change if you prefer).

I don't think it requires any documentation, since those commands are things that Windows users would probably expect to be built in to the shell, and if someone is worried about those commands, it is easy to override or delete them.

rbrewer123:
Ah, @wrywerytwreywery, I was busy too, but finally back to this.  I switched it to a set so hopefully we are good to merge.  Thanks!

wrywerytwreywery:
Great!  Just merged in!

Issue #220, opened 2015-05-07
Subprocess exit code exception (bash "set -e")


rbrewer123:
I saw some mention of subprocess exit codes under another issue but thought opening a separate issue would be best.   

The upcoming history mechanism may provide a useful way to fetch the exit code of a recently run process.  For scripts, it would also be nice to have a mechanism similar to "set -e" in bash, where any subprocess that exits with a non-zero exit code is considered an error.  For xonsh, perhaps this mode could raise a python exception similar to subprocess.check_call().  I would suggest raising exactly the CalledProcessError exception from the subprocess module. 

I think enabling this mode per-module via a module-level global variable would be convenient ("subprocess_check = True") and/or via a "with" statement ("with subprocess_check(True):").

scopatz:
Hi @rbrewer123, thanks for noting this. I think this is good feature to have, just like traces (`set -x`).  The variable name might need to be special and/or namespaced (`__xonsh_subproc_check__`), but the general idea is good.  A context manager would also be nice.

scopatz:
I believe that this has now been implemented.

PR #259, opened 2015-06-03
subprocess exit code exceptions

rbrewer123:
This is a prototype implementation of #220 that will raise a ``subprocess.CalledProcessError`` for subprocesses that return a non-zero exitcode.  This behavior is only activated when the builtin ``__xonsh_subproc_check__`` is set to True (it defaults to False).

The below test script ``pinger.xsh`` demonstrates the feature.  Running the script like ``xonsh pinger.xsh localhost`` should print ``success``.  Running the script with an invalid host such as ``xonsh pinger.xsh badhost`` should show a traceback.

Unfortunately, line 12 of ``pinger.xsh`` (assigning ``__xonsh_subproc_check__`` to True) is not what I would prefer.  I was imagining that a global like line 13 would work.  Any ideas for getting that behavior?  The idea is that line 13 could appear in any .xsh file that uses process-based exception handling, but other imported .xsh files that don't use that behavior would use the default behavior where the subprocess exit codes are not checked and no exceptions are raised.

I'm interested in feedback and any tips on getting the behavior described above.  I'll probably also try to get a context manager for ``__xonsh_subproc_check__`` before asking this to be merged.

```python
#!/usr/bin/env xonsh
#
# Usage:
#   pinger.xsh <hostname>
#

import builtins
import platform

builtins.__xonsh_subproc_check__ = True
#__xonsh_subproc_check__ = True

if ON_WINDOWS:
    ping -n 1 $ARG1 &> NUL:
else:
    ping -c 1 $ARG1 &> /dev/null

print('success')
```


scopatz:
@rbrewer123 - would you mind merging from master please?  Seems like there are some conflicts...

rbrewer123:
@scopatz, github is saying it could still be merged automatically, has something changed?  I've been away and not following the latest happenings closely; hopefully sometime this week I can get caught up and take another look at this.

rbrewer123:
Weird, as soon as I posted that github is telling me the branch has conflicts...

scopatz:
Hopefully the conflict is easy to resolve!  Please ping when it is ready to take a look.

rbrewer123:
@scopatz, this is now up-to-date with master.  Where it stands is that setting `__xonsh_subproc_check__ = True` in the current context will raise a proper `subprocess.CalledProcessError` exception when a subprocess has a non-zero return code.  The problem is that I don't know how to make that work correctly when other .xsh modules are imported.  I'd like to have `__xonsh_subproc_check__` be a module-scope setting.  So one module can be written to expect those exceptions, a different module could be written to not expect them, and they can still be combined into a larger script without having to alter either one.  Do you have any suggestions for how to do that?  I'm open to you just going ahead and doing it if you think that is easier.

I still need to write some nosetests for it too before this could be merged.

After that, I think having a context manager would be appropriate so different places in a single module could use it differently.  I think a context manager could be a separate PR though.

For me, being able to check for subprocess errors is the only thing really holding me back from starting to write some useful scripts with xonsh, and thus retiring bash scripts whenever I can.

Name this `exc_exitcode.xsh`:
```
#!/usr/bin/env xonsh

import subprocess

__xonsh_subproc_check__ = True


def grep_it(strng, filename):
    try:
	occurs = $(grep @(strng) @(filename))
    except subprocess.CalledProcessError:
        print('caught CalledProcessError exception')
        print('{} NOT in {}'.format(strng, filename))
    else:
        if occurs:
            print('{} in {}'.format(strng, filename))
        else:
            print('{} NOT in {}'.format(strng, filename))


grep_it('foobar1', 'testing.txt')
grep_it('foobar2', 'testing.txt')
```

Name this `pinger.xsh`:

```
#!/usr/bin/env xonsh
#
# Usage:
#   pinger.xsh <hostname>
#

import platform

import exc_exitcode

#__xonsh_subproc_check__ = True


if ON_WINDOWS:
    ping -n 1 $ARG1 &> NUL:
else:
    ping -c 1 $ARG1 &> /dev/null

print('success')
```

When you run `exc_exitcode.xsh` by itself, it correctly raises the exception.  When you run `pinger.xsh`, which imports `exc_exitcode.xsh`, the exception is not raised.

xonsh/built_ins.py
```diff
@@ -568,17 +568,24 @@ def run_subproc(cmds, captured=True):
...
-            return output
+            output = out
     elif last_stdout not in (PIPE, None, sys.stdout):
         last_stdout.close()
 
+    if job and job['obj'].returncode != 0 and builtins.__xonsh_ctx__['__xonsh_subproc_check__']:
+        cmd = [' '.join(i) if isinstance(i, list) else i for i in job['cmds']]
+        cmd = ' '.join(cmd)
+        raise CalledProcessError(job['obj'].returncode, cmd, output)
+    return output
```
scopatz: I think we only want to return output if we are captured.  Otherwise, we should be returning None.

xonsh/built_ins.py
```diff
@@ -568,17 +568,24 @@ def run_subproc(cmds, captured=True):
...
         if prev_proc.stdout not in (None, sys.stdout):
-            output = prev_proc.stdout.read()
+            out = prev_proc.stdout.read()
         if captured:
-            return output
+            output = out
     elif last_stdout not in (PIPE, None, sys.stdout):
         last_stdout.close()
 
+    if job and job['obj'].returncode != 0 and builtins.__xonsh_ctx__['__xonsh_subproc_check__']:
```
scopatz: I am going to comment on this line in the main conversation.

scopatz:
Ok @rbrewer123, I agree that this should be a module-level variable.  I am happy with the context manager, etc, coming later. 

I think basic problem with the PR as it stands is the condition: `builtins.__xonsh_ctx__['__xonsh_subproc_check__']`.  This checks built-in scope, rather than global (module) scope.  Luckily, we have the dynamism of Python on our side. I think that something like the following will work:

First, rewrite the condition block to look like (remember to `import inspect` too):

```python
if job and job['obj'].returncode != 0:
    stx = inspect.stack()
    check = find_subproc_check(stx)
    del stx  # remove as soon as possible
    if check:
        cmd = [' '.join(i) if isinstance(i, list) else i for i in job['cmds']]
        cmd = ' '.join(cmd)
        raise CalledProcessError(job['obj'].returncode, cmd, output)
```

Note that the `inspect.stack()` call has to (or really should) live at the point of use. Then somewhere else, implement the find function as:

```python
def find_subproc_check(stx):
    for frme, name, _, _, _, _ in stx:
        if not name.endswith('.xsh'):
            continue
        check = frme.f_globals.get('__xonsh_subproc_check__', None)
        if check is not None:
            break
    else:
        check = False
    return check
```

The beauty of this is that the module closest to the function in the stack is the one that has precedence.  You might need to fine tune this a bit, but this is the general idea.

rbrewer123:
@scopatz, thanks, that helps a lot.  I'll continue that path and see how it goes.

gforsyth:
Hey @rbrewer123 -- this has been implemented in #641 so I'm going to go ahead and close this PR.  If you would like it reopened just let me know.

Thanks!

PR #261, opened 2015-06-03
Change handling of tilde

### LINKED ISSUE #260: inconsistent tilde (~) handling
On Windows, handling of tilde (``~``) expansion is inconsistent.  xonsh seems to expand tilde correctly (to the user's home directory) for its builtin functions such as ``cd ~``.

Windows commands perform no tilde expansion, so for example, doing ``move foobar.txt ~`` results in renaming the file ``foobar.txt`` to a file named ``~`` in the current directory.

Unix commands ported to Windows in the msysGit suite (e.g. ``cp``, ``mv``) perform their own tilde expansion.  Unfortunately, they expand tilde to the location where msysGit is installed.

To summarize: on windows there are currently 3 different behaviors of tilde expansion.  It seems like this could be cleaned up if on Windows xonsh always performed tilde expansion of all arguments in subprocess mode.  Does this seem like a valid way to proceed?

wrywerytwreywery:
I believe this should fix #260.  Also made some additions to the gitignore file, to ignore temporary files from vim and emacs (I can remove that or move it to a separate PR if necessary).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,13 @@ dist/
 xonsh.egg-info/
 docs/_build/
 
+# temporary files from vim and emacs
+*~
+*#
+.#*
+*.swp
+*.swo
+
 # Virtualenv
 pip-selfcheck.json
 bin/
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks @wrywerytwreywery. Seems to work for me

Issue #262, opened 2015-06-05
source command


scopatz:
See this mailing list post https://groups.google.com/d/topic/xonsh/OUUEwdL90TI/discussion

rbrewer123:
@wrywerytwreywery implemented this in #263.  Thanks!

PR #263, opened 2015-06-05
"source" alias

wrywerytwreywery:
Implementation of the `source` alias as discussed in #262.


rbrewer123:
Works well for me.  Thanks for the feature @wrywerytwreywery!

PR #270, opened 2015-06-18
Favor runnable extensions on Windows

BYK:
`xonsh` favors extensionless files with shebangs on Windows which
is causing certain issues when some packages come with both Linux
and Windows runnables where the Linux version is extensionless.

This patch favors files ending with any extension listed in
`PATHEXT` over extensionless files on Windows.

Fixes the issue mentioned in here:
https://groups.google.com/d/msg/xonsh/nYraaPmewP0/as6d5TrrBU8J

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -207,14 +207,17 @@ def _get_runnable_name(fname):
     for d in builtins.__xonsh_env__['PATH']:
         if os.path.isdir(d):
             files = os.listdir(d)
-            if fname in files:
-                return os.path.join(d, fname)
+
             if ON_WINDOWS:
                 PATHEXT = builtins.__xonsh_env__.get('PATHEXT', [])
                 for dirfile in files:
                     froot, ext = os.path.splitext(dirfile)
                     if fname == froot and ext.upper() in PATHEXT:
                         return os.path.join(d, dirfile)
+
+            if fname in files:
+                return os.path.join(d, fname)
+
     return None
 
 


scopatz:
@rbrewer123 - you want to take a look at this? You know much more about Windows than I do :)

scopatz:
I will say that it looks fine to me though.

rbrewer123:
@scopatz, thanks, I'm on it.  :)

@BYK, thanks for pointing out this problem and submitting this PR.  I tried it out, and I think it's most of the fix we want, but there's a remaining case to catch.  For my testing I have the following directory structure:

```
junk\
        foob  - xonsh shebang script
        foob.exe  - executable
```

When I cd into junk and enter `foob`, I see the desired behavior that `foob.exe` is run.  However, when I enter `.\foob` it runs `foob` instead of the `foob.exe` that I would expect.  Also, when I cd to the parent of junk and run `junk\foob`, again `foob` is executed instead of `foob.exe`. 

I think the first couple lines of `_get_runnable_name()` are causing this remaining issue, but I don't know if that check is needed on Windows or not.  @wrywerytwreywery, what is the initial `if` check in `_get_runnable_name()` supposed to do?

BYK:
@rbrewer123 - It may sound weird but I'd say what you described is the expected behavior since when you start to use an actual path, then it should favor the exact path, if that's there.

I mean I'm using this patched version since I submitted the PR and very happy with that :D If you think it should favor those files at all times, I can definitely look into implementing that.

BYK:
>It may sound weird but I'd say what you described is the expected behavior since when you start to use an actual path, then it should favor the exact path, if that's there.

Eh, I just tested this on PowerShell and it behaves exactly like how you described so I am not sure. The reason for that may simply be PowerShell not supporting extensionless executable files though.

rbrewer123:
@BYK, yeah, there may not be a perfect answer here.  I've thought about it more too, and the problem with my proposal above is that the extensionless file ends up being completely "shadowed" by the `.exe` on Windows with no possibility of calling it.  It's nice that your PR still allows running the extensionless file if that is really needed.  

I think flexibility beats consistency in this case.  Perhaps with some more experience we can decide if further tweaking of this behavior is actually needed.

Thanks again for the PR and the thoughtful discussion.

BYK:
You're welcome and thank you as well for the discussion :)
Glad to see the efforts to support xonsh on Windows since it is a very big deal :)

PR #279, opened 2015-06-26
Update to conda recipe 

melund:
Update the conda recipe to build directly from the git repository, and use the tags in git to create the right version number. Also added some fixes to make build work on windows.


scopatz:
Thanks @melund! Sorry about the delay in merging. Looks great!

PR #282, opened 2015-07-08
Catch cases where the commands launched by subprocess are not found. 

melund:
subprocess.CalledProcessError is only raised if the command is found but returns a non-zero exit code. 

This will allow the code to work on windows where the bash command is not on the path. The reason it hasn't been a problem for users is that most people have MsysGit installed on windows, and that adds the bash command to the path.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -270,8 +270,6 @@ def get_git_branch(cwd=None):
                 branch = s
         except (subprocess.CalledProcessError, FileNotFoundError):
             pass
-        except OSError:
-            pass
 
     return branch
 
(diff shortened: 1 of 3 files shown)


xonsh/environ.py
```diff
@@ -268,7 +268,9 @@ def get_git_branch(cwd=None):
             s = s.strip()
             if len(s) > 0:
                 branch = s
-        except subprocess.CalledProcessError:
+        except (subprocess.CalledProcessError, FileNotFoundError):
+            pass
+        except OSError:
```
scopatz: Can't these all be in the same exception.  ie:

```python
except (subprocess.CalledProcessError, FileNotFoundError, OSError):
            pass
```

scopatz:
This looks good @melund!  Just the one minor issue.

melund:
Thanks for pointing that out @scopatz . They could be one exception, but I never intended to include OSError catch. FileNotFoundError is the subclass of OSError that we wanted to catch.

scopatz:
Thanks @melund!

PR #283, opened 2015-07-10
Work-around a bug in "conda convert"

melund:
This is temporary work-around for a bug in the conda-build packages

Removing (>=) in the version specification should allow conda convert to correctly convert conda packages build for linux can be converted to windows.


melund:
I was a bit hasty here. The work around caused other issues with conda build... Closing

PR #284, opened 2015-07-10
Temporary work-around for bug in conda-build

melund:
Removing (>=) in the version specification should allow conda convert to correctly convert conda packages build for linux can be converted to windows.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -13,10 +13,10 @@ build:
 
 requirements:
   build:
-    - python >=3.4
+    - python 3.4*
     - ply
   run:
-    - python >=3.4
+    - python 3.4*
     - ply
 
 about:


scopatz:
Cool Thanks!

asmeurer:
What is the bug in conda? Can you open a conda-build issue for it?

melund:
@asmeurer ... The bug is in the "conda convert" part of conda-build. I made a pull request to conda-build that fixes the problem. 
https://github.com/conda/conda-build/pull/481

asmeurer:
I see what you are doing now. conda build already handles specifying a Python dependency automatically. You should just list the dependency as `python`, and build it with the Python 3.4 version of conda (or use `conda build --python 3.4`).

asmeurer:
This is what you should be doing https://github.com/scopatz/xonsh/pull/304

Issue #280, opened 2015-06-26
xonsh on ubuntu does not suggest missing packages


eskhool:
In bash presently, giving any command gives us the following helpful completion. Logging the issue here to track such enhancements (somehow bash in ubuntu must be trapping the command not found and using apt to complete..it is a useful function)

The program 'nethogs' is currently not installed. You can install it by typing:
sudo apt-get install nethogs

scopatz:
Thanks for bringing this up! I'll see if I can figure out how to get this to work.

scopatz:
#287 adds this capability.

scopatz:
closed by #287

PR #287, opened 2015-07-11
added command-not-found capability for ubuntu and debian

scopatz:
This adds the feature described in #280

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -346,8 +346,9 @@ def suggest_commands(cmd, env, aliases):
         length = max(len(key) for key, _ in items) + 2
         alternatives = '\n'.join('    {: <{}} {}'.format(key + ":", length, val)
                                  for key, val in items)
-        rtn = '{}\n{}\n\n'.format(tips, alternatives)
-        rtn += command_not_found(cmd)
+        rtn = '{}\n{}'.format(tips, alternatives)
+        c = command_not_found(cmd)
+        rtn += ('\n\n' + c) if len(c) > 0 else ''
     return rtn
 
 
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -21,6 +21,7 @@ import re
 import sys
 import builtins
 import platform
+import subprocess
 from collections import OrderedDict, Sequence
 
 if sys.version_info[0] >= 3:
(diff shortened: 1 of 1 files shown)


adqm:
Hi @scopatz; sorry for being so absent (that probably won't improve much the rest of the summer, unfortunately).

I think this is a good idea, but a few things:

* If `command-not-found` doesn't exist, but the other suggester offers suggestions, I get two blank lines after the entry:
```
hartz@lappy-508 ~ $ gemtail
xonsh: subprocess mode: command not found: gemtail
    getmail:  Alias
    email:    Alias


hartz@lappy-508 ~ $
```

* After installing `command-not-found`, it seems like the version in Debian Jessie (the current stable release) uses a different CLI than the one on Ubuntu.  After installing that, I get:
```
hartz@lappy-508 ~ $ gemtail
xonsh: subprocess mode: command not found: gemtail
    getmail:  Alias
    email:    Alias

Usage: command-not-found [options] <command-name>

command-not-found: error: no such option: --no-failure-msg
hartz@lappy-508 ~ $
```

  If it's helpful, the `command-not-found` version on Debian Jessie is 0.2.26.

* Even if `command-not-found` is installed and uses the right CLI, I am not sure the output will be pretty when concatenating the two outputs (we may want to modify the output of `command-not-found`, and/or of the current suggestions).  I think at the very least, this will report that the command in question was not found multiple times (3?): once from my suggester, once at the start of the `command-not-found` output, and once at the end of the `command-not-found` output).  That's a lot of verbosity for a user to have to navigate to get the appropriate suggestions...

  Oooohh, maybe the `--no-failure-msg` flag is supposed to handle this?

scopatz:
Hi @adqm, thanks for the feedback and testing. I have fixed the issues that you mentioned.  You are correct about the purpose of `--no-failure-msg`, but it was one line to re-implement it manually :). 

> sorry for being so absent (that probably won't improve much the rest of the summer, unfortunately).

No worries!

scopatz:
I think that this is ready for another look.

scopatz:
ping!

adqm:
Sorry for the delay; just merged in.  Looks good!

scopatz:
Thanks!

PR #300, opened 2015-07-14
Fix a bunch of `hg` related errors on Windows

BYK:
xonsh throws an exception and quits when I `cd` into a Mercurial
directory. This patch fixes a few issues causing it to throw exceptions
on Windows when trying to get the `hg` branch name and info.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1,7 +1,6 @@
 """Environment for the xonsh shell."""
 import os
 import re
-import locale
 import socket
 import string
 import locale
@@ -9,8 +8,7 @@ import builtins
 import subprocess
 from warnings import warn
 from functools import wraps
-from collections import MutableMapping, MutableSequence, MutableSet, \
-    defaultdict, namedtuple
+from collections import MutableMapping, MutableSequence, MutableSet, namedtuple
 
 from xonsh import __version__ as XONSH_VERSION
 from xonsh.tools import TERM_COLORS, ON_WINDOWS, ON_MAC, string_types, is_int,\
@@ -314,7 +312,7 @@ def get_hg_branch(cwd=None, root=None):
         return "{0}, {1}".format(
             *(b.strip(os.linesep) for b in (branch, active_bookmark)))
 
-    return branch.strip(os.linesep)
+    return branch.strip(os.linesep) if branch else None
 
 
 def current_branch(pad=True):
@@ -344,10 +342,10 @@ def git_dirty_working_directory(cwd=None):
 
 @ensure_hg
 def hg_dirty_working_directory(cwd=None, root=None):
-    id = call_hg_command(['identify', '--id'], cwd).strip(os.linesep)
+    id = call_hg_command(['identify', '--id'], cwd)
     if id is None:
         return False
-    return id.endswith('+')
+    return id.strip(os.linesep).endswith('+')
 
 
 def dirty_working_directory(cwd=None):


adqm:
Thanks for submitting this, @BYK!  I don't have mercurial installed on my Windows machine, and with this patch, I get an error when starting xonsh (presumably, since the `ensure_hg` wrapper no longer makes sure that mercurial is installed):
```
C:\Users\hartz\Documents\xonsh>scripts\xonsh.bat
  File "C:\Python34\lib\runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Python34\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\hartz\Documents\xonsh\xonsh\main.py", line 92, in <module>
    main()
  File "C:\Users\hartz\Documents\xonsh\xonsh\main.py", line 88, in main
    shell.cmdloop()
  File "C:\Users\hartz\Documents\xonsh\xonsh\readline_shell.py", line 114, in cm
dloop
    super().cmdloop(intro=intro)
  File "C:\Python34\lib\cmd.py", line 126, in cmdloop
    line = input(self.prompt)
  File "C:\Users\hartz\Documents\xonsh\xonsh\readline_shell.py", line 127, in pr
ompt
    return super().prompt
  File "C:\Users\hartz\Documents\xonsh\xonsh\base_shell.py", line 106, in prompt

    p = format_prompt(p)
  File "C:\Users\hartz\Documents\xonsh\xonsh\environ.py", line 410, in format_pr
ompt
    for (k, v) in fmt.items()
  File "C:\Users\hartz\Documents\xonsh\xonsh\environ.py", line 411, in <dictcomp
>
    if k in included_names}
  File "C:\Users\hartz\Documents\xonsh\xonsh\environ.py", line 367, in branch_co
lor
    return (TERM_COLORS['BOLD_RED'] if dirty_working_directory() else
  File "C:\Users\hartz\Documents\xonsh\xonsh\environ.py", line 362, in dirty_wor
king_directory
    return git_dirty_working_directory() or hg_dirty_working_directory()
  File "C:\Users\hartz\Documents\xonsh\xonsh\environ.py", line 352, in hg_dirty_
working_directory
    id = call_hg_command(['identify', '--id'], cwd)
  File "C:\Users\hartz\Documents\xonsh\xonsh\environ.py", line 291, in call_hg_c
ommand
    env=hg_env)
  File "C:\Python34\lib\subprocess.py", line 607, in check_output
    with Popen(*popenargs, stdout=PIPE, **kwargs) as process:
  File "C:\Python34\lib\subprocess.py", line 859, in __init__
    restore_signals, start_new_session)
  File "C:\Python34\lib\subprocess.py", line 1112, in _execute_child
    startupinfo)
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

I think we could get away with leaving out the `ensure_git` wrapper because the way most people have gotten `ls`, etc on Windows is through msysgit, so if someone has a reasonably-sane shell environment on Windows, I think we can expect `git` to be around.  This isn't the case with hg, I think.  But maybe that isn't such a reasonable assumption?

All told, it might be nicer to do checks once at startup for whether `git` and `hg` are installed, so that we don't have to keep doing those checks on the fly (I think one of the main reasons for bypassing the `ensure_git` check was that it was pretty slow on Windows).  I do think the checks themselves are useful, and it would probably be a good idea to have some mechanism for `ensure_hg` and `ensure_git` both to work.

Of course, doing the checks on startup only has disadvantages, too; if someone installs mercurial or git midway through a session, it would be nice for the shell to know that; but maybe we could have some way of explicitly forcing that check again after startup.  It does feel like a waste to be continually checking for whether git/hg is installed...

@scopatz, @rbrewer123, do you have a feeling on this?

scopatz:
I think it is probably more robust to just do the check each time.

scopatz:
Also, we should be careful with adding functionality to startup.

BYK:
@scopatz - lol, sorry about that, very good catch!

I was wondering if I can get `locate_binary` to work on Windows properly which I did. Instead of bundling that up with this patch, I submitted #302. Once that gets onto master, I'll rebase this one and it should work without issues!

BYK:
Updated!

scopatz:
Thanks @BYK!

PR #302, opened 2015-07-15
Enable `locate_binary` on Windows

BYK:
This makes Git and Mercurial integrations more robust on windows.

Prerequisite for #300.


scopatz:
Thanks @BYK!

PR #309, opened 2015-07-20
fix for vim/emacs not coming back from background

### LINKED ISSUE #265: Vim refuses to come back to foreground with fg
https://asciinema.org/a/0qa0ncj2izljo0oyr9wn3pi2u

When inside of vim, CTRL+Z works fine, however running the `fg` command fails, giving me the following output : 

```
*[1] running: vim (15707)

*[1] stopped: vim & (15707)
```

Running on archlinux.
Xonsh version : 0.1.6
Vim version : 7.4.729

adqm:
This is an attempt to fix #265.  It isn't perfect (if you suspend vim, then kill the PID, then `fg` it, you are left with an unusable terminal), but I think it works better than before since that is an edge-case.

### DIFF
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -19,7 +19,7 @@ except OSError:
 
 if ON_WINDOWS:
     def _continue(obj):
-        pass
+        return None
 
 
     def _kill(obj):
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks @adqm. Looks great and works for me.

PR #314, opened 2015-07-24
Added config file for binstar build service

melund:
I thought it would be nice to automate the creation of the conda builds on Anaconda.org (formerly binstar.org)

This PR adds a config file to use the free build server on anaconda.org. 

It builds using the linux-64 worker on Anaconda.org, then converts the conda packages to osx-64, linux-32, win-32, win-64. 

@scopatz : To make it work you need to configure the build server for your account.

With "binstar-build" client installed run this command:
binstar-build save  https://github.com/scopatz/xonsh

Then go to http://anaconda.org/scopatz/xonsh/settings/ci and make it look like this:
image

You may need to enable the webhook to github. 

That should ensure that new conda builds are added to your "dev" channel on Anaconda.org every time you push something to master. You can, of course, configure it as you like.

### DIFF
--- a/.binstar.yml
+++ b/.binstar.yml
@@ -1,0 +1,19 @@
+package: xonsh
+
+platform: 
+ - linux-64
+
+engine:
+ - python=3
+
+script:
+  - conda build recipe
+
+after_success:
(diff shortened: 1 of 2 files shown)


scopatz:
Cool thanks @melund!

Issue #313, opened 2015-07-24
aliases not found when stored in .aliases files


michaelaye:
I source in my aliases via an `.aliases` file, for having them both available in bash and zsh.
xonsh does not pick them up.
My `.bash_profile` part that sources them in looks like the usual bash idiom:
```bash
if [ -f ~/.aliases ]; then
    . ~/.aliases
```

michaelaye:
for further evaluation of xonsh I would need a quick workaround to get my aliases in. is there one? I tried 'source .aliases' but that does not work.

asmeurer:
I think you have to use `source-bash`.

michaelaye:
still doesn't work:
```bash
klay6683@MacL2938 ~ $ source-bash .aliases
klay6683@MacL2938 ~ $ cdsrc
xonsh: subprocess mode: command not found: cdsrc
```

michaelaye:
hm...
```bash
klay6683@MacL2938 ~ $ source-bash?
Traceback (most recent call last):
  File "/Users/klay6683/miniconda3/lib/python3.4/site-packages/xonsh/base_shell.py", line 42, in default
    self.execer.exec(code, mode='single', glbs=self.ctx)  # no locals
  File "/Users/klay6683/miniconda3/lib/python3.4/site-packages/xonsh/execer.py", line 127, in exec
    return exec(code, glbs, locs)
  File "<xonsh-code>", line 1, in <module>
NameError: name 'source' is not defined
```

scopatz:
`source-bash` doesn't grab aliases.  (yes we probably need to fix this.)  You probably want to look at the `xonsh.aliases.bash_aliases()` function.  This could probably be generalize to accept a filename or a list of files. From there it would be super easy to make an alias command that does this. Or you could fix the `xonsh.aliases.source_bash()` function to also pull out aliases.  That might be even better.  Would you be willing to put in a PR to that effect?

michaelaye:
Is it a declared functionality to parse and take on things from the bash config? Then this should be fixed in wherever you do the bash profile parsing. To be able to parse another file after launch of xonsh is sugar on top, I think. So which of above solutions is responsible for parsing the bash profile?

michaelaye:
This works, gladly, to switch conda environments:
```python
source-bash activate py27
```

scopatz:
[...]

PR #319, opened 2015-07-26
added login shell to bash

scopatz:
This might help with #313

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -72,7 +72,7 @@ def xexec(args, stdin=None):
 def bash_aliases():
     """Computes a dictionary of aliases based on Bash's aliases."""
     try:
-        s = subprocess.check_output(['bash', '-i'],
+        s = subprocess.check_output(['bash', '-i', '-l'],
                                     input='alias',
                                     stderr=subprocess.PIPE,
                                     universal_newlines=True)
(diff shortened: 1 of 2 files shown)


michaelaye:
Getting this now:
```python
klay6683@MacL2938 ~ $ aliases['cdsrc']
['cd', '$HOME/Dropbox/src']
klay6683@MacL2938 ~ $ cdsrc
cd: no such file or directory: $HOME/Dropbox/src
```

scopatz:
I think that you could fix this by changing the alias in bash to `cd ~/Dropbox/src`

michaelaye:
That's not a fix, that's a cheat. The user should not have to change her bash setup to make xonsh work.

scopatz:
hey @melund - would you mind taking a look at this PR and possibly merging it?

melund:
Hi @scopatz . I can't test it since I use Windows. However, reading up on bash, it seems reasonable. I can still click merge button if you want :)

The last issue raised by @michaelaye is another. I can't do that on windows either. 
`cd $USERPROFILE` works fine. But `cd $USERPROFILE/Documents` does not. In the shell, it is not a problem. If I press 'tab' the autocompleter expands it to the full path. Should we create an separate issue for it?

scopatz:
> I can still click merge button if you want :)

Please do!

> Should we create an separate issue for it?

I think so.  Thanks!

melund:
I have create a new issue for the problem reported by  @michaelaye 
https://github.com/scopatz/xonsh/issues/338

PR #225, opened 2015-05-10
Add support for CDPATH

naufraghi:
CDPATH works like PATH for `cd`, no CDPATH is like CDPATH=. but more prefixes can be added.

PR #320, opened 2015-07-28
Closes #307, do not use full path for completion, use basename only

### LINKED ISSUE #307: Regression in current folder tab completion (./executable)
In latest master when I type `./some_exe`, I have this turned into `some_executable`. Expected behavior: `./some_executable`

In version 0.1.6 this works well.

naufraghi:
Follow up from the pull request that closed #225 (add CDPATH support)

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -104,7 +104,7 @@ class Completer(object):
                 return sorted(self.module_complete(prefix))
             if cmd in self._all_commands():
                 # subproc mode; do path completions
-                return sorted(self.path_complete(prefix))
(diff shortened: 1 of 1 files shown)


naufraghi:
Not sure it fixes the `./executable` regression, I have the same problem disabling `completer._add_cdpaths` at all.

Is there some way to activate the completer for *directory changing* command only? `cd`, `pushd`, but not `cat`?

scopatz:
Thanks @naufraghi! This doesn't fix that problem, but it doesn't seem to hurt.

PR #322, opened 2015-07-28
ArchLinux installation instructions

### LINKED ISSUE #275: Update Install Documentation to Include Instructions for Homebrew and Pacman
Acoording to [this tweet](https://twitter.com/scopatz/status/613383868851183616) from @scopatz, `xonsh` is actually **in** Homebrew and the AUR. Unfortunately, instructions for installing `xonsh` through either of these PM's is missing from the website, misleading users into using pip/conda when they don't have to.

Updating the docs on the site to reflect that you can use pacman or homebrew to get `xonsh` would be a good move, IMO. If I ran Arch or OSX, I would test these and add myself, but I'm a Debian man.

kseistrup:
Instructions for installing xonsh on ArchLinux from AUR added.

If merged, this closes #275.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -99,6 +99,20 @@ the following from the source directory,
 
     $ python setup.py install
 
+ArchLinux users can install xonsh with e.g. yaourt or aura:
+
+**yaourt:**
+
+.. code-block:: bash
+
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks @kseistrup!

PR #327, opened 2015-07-29
Add -V/--version

kseistrup:
Command line switches `-V`/`--version` added (these are the same switches that the Python interpreter uses):

```sh
$ xonsh --version
xonsh/0.1.6
$ 
```


scopatz:
Thanks @kseistrup!

PR #331, opened 2015-07-30
Hiding traceback in the shell

melund:
Hi 

This one is for discussion. I find it a bit annoying to get full tracebacks when working in the shell. If I just try a few things that fail and my screen is complely full of junk. 

So I thought that it may be an idea to hide the traceback by default, and only show the exception. 

This pull request implements that idea. Let me know what you think.

------------------------------------------
The functionality can be controlled with the environment variable $XONSH_SHOW_TRACEBACK

If the variable is not set, the default is to hide the traceback but with an extra notice that the traceback can be enabled:
mel@home ~ $x = {foo:1, bar}
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK=True
ValueError: Dict doesn't have the same number of keys as values
mel@home ~ $

### DIFF
--- a/xonsh/base_shell.py
+++ b/xonsh/base_shell.py
@@ -43,7 +43,7 @@ class BaseShell(object):
         except XonshError as e:
             print(e.args[0], file=sys.stderr)
         except:
-            traceback.print_exc()
+            _print_exception()
         if builtins.__xonsh_exit__:
             return True
 
@@ -65,9 +65,13 @@ class BaseShell(object):
         except SyntaxError:
             if line == '\n':
                 self.reset_buffer()
-                traceback.print_exc()
+                _print_exception()
                 return None
             self.need_more_lines = True
+        except:
+            self.reset_buffer()
+            _print_exception()
+            return None
         return code
 
     def reset_buffer(self):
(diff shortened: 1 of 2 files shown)


xonsh/base_shell.py
```diff
@@ -108,3 +112,14 @@ def prompt(self):
...
+        
+def _print_exception():
+    """Print exceptions with/without traceback."""
+    if not 'XONSH_SHOW_TRACEBACK' in builtins.__xonsh_env__:
+        sys.stderr.write('xonsh: For full traceback set: '
+                         '$XONSH_SHOW_TRACEBACK=True\n')
+    if builtins.__xonsh_env__.get('XONSH_SHOW_TRACEBACK', False):
+        traceback.print_exc()
+    else:
+        type, value, exc_traceback = sys.exc_info()
```
scopatz: `type` is a python built-in function.

scopatz:
Thanks @melund, I think that this looks good.  Just a couple of points:

1. Can you add `XONSH_SHOW_TRACEBACK` to the docs, it should be in the table in this section: http://xonsh.org/tutorial.html#environment-variables

2. I like the note that tells people how to turn the traceback on.  However, if someone has set `$XONSH_SHOW_TRACEBACK=False`, then they probably shouldn't get the message at all.  So maybe if the variable is undefined or None, should the message be printed.  I can add this in another PR after this one, if you want.

melund:
I am glad you like the idea... 

1. I will add it to the docs as soon as I find some spare time. So wait with merging this PR
2. You are right, but the code already handles that. The notice is not printed if the XONSH_SHOW_TRACEBACK exists.

scopatz:
On 2 - opps! you are right! I misread my terminal.  So just ping this thread when the docs are ready.  Thanks!

melund:
I updated the docs. I needed to expand the table a bit, thus the many changed lines. Sorry for that :(

scopatz:
Thanks @melund!

kseistrup:
This is a wonderful feature, @melund — kudos!

PR #335, opened 2015-07-31
ignore case on completion

### LINKED ISSUE #290: Use readline init file
https://cnswww.cns.cwru.edu/php/chet/readline/readline.html#SEC32

https://cnswww.cns.cwru.edu/php/chet/readline/readline.html#SEC9

scopatz:
As requested by @asmeurer. Should close #290.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -164,80 +164,83 @@ useful to have more sophisticated types, like functions, in the environment.
 There are handful of environment variables that xonsh considers special.
 They can be seen in the table below:
 
-==================== ============================= ================================
-variable             default                       description
-==================== ============================= ================================
-PROMPT               xonsh.environ.DEFAULT_PROMPT  The prompt text.  May contain
-                                                   keyword arguments which are
-                                                   auto-formatted (see `Customizing
-                                                   the Prompt`_ below).
-MULTILINE_PROMPT     ``'.'``                       Prompt text for 2nd+ lines of
-                                                   input, may be str or
-                                                   function which returns a str.
-TITLE                xonsh.environ.DEFAULT_TITLE   The title text for the window
-                                                   in which xonsh is running.
-                                                   Formatted in the same manner
-                                                   as PROMPT (see `Customizing the
-                                                   Prompt`_ below).
-FORMATTER_DICT       xonsh.environ.FORMATTER_DICT  Dictionary containing variables
-                                                   to be used when formatting PROMPT
-                                                   and TITLE (see `Customizing the
-                                                   Prompt`_ below).
-XONSHRC              ``'~/.xonshrc'``              Location of run control file
-XONSH_HISTORY_SIZE   8128                          Number of items to store in the
-                                                   history.
-XONSH_HISTORY_FILE   ``'~/.xonsh_history'``        Location of history file
-XONSH_INTERACTIVE                                  ``True`` if xonsh is running
-                                                   interactively, and ``False``
-                                                   otherwise.
-BASH_COMPLETIONS     ``[] or ['/etc/...']``        This is a list of strings that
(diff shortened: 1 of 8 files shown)


asmeurer:
I think command completion is case insensitive here, but file completion is not.

scopatz:
Do you think that command completion should be case sensitive?

asmeurer:
No, IMHO everything should be case insensitive (although others may disagree, so customization is probably a good thing). It's definitely more important for file completion than command completion, though, since commands with capital letters are rare.

scopatz:
Ok @asmeurer - I have added case insensitivity for file names. This was a little more subtle than I first thought, but should work now.

scopatz:
This is ready for review now!

adqm:
Personally, I really don't want case-insensitive completion at all (for files or for commands).  Is there a way to switch this on/off?

I also think the sane default for that setting might platform-dependent (since Windows is case-insensitive but *nix environments tend to be case-sensitive).

asmeurer:
The OS X filesystem is case insensitive.

asmeurer:
In very basic tests this seems to work.

scopatz:
Ok @adqm, I have added the `$CASE_SENSITIVE_COMPLETIONS` environment variable, which defaults to `xonsh.tools.ON_LINUX`.  This should address your concerns. Sorry I didn't implement this originally.  I was being a little lazy :)

adqm:
@scopatz, I ran this through some quick tests, and I think it is working nicely!  I haven't had a chance to test on Windows yet; should we wait to merge in until we've done some quick Windows tests, or are you reasonably sure that this will work under Windows as well?

scopatz:
I have no reason to think that it won't work on Windows :).  Maybe @rbrewer123 or @melund or someone could test?

scopatz:
Otherwise, I think that it is good to merge.

melund:
It seems to work on windows, although there is some weird behaviour

I have two folders:
```
E:\test $ ls
SUBFOLDER_2/  subfolder_1/
```

```
E:\test $ cd subfol <TAB>
SUBFOLDER_2 subfolder_1
E:\test $ cd SUBFOLDER_
```

I think it should expand to the lowercase version. And there should have been slashes at the for the suggestions. 
SUBFOLDER_2/ subfolder_1/


Another weird thing. When I enter a lower case folder using capital letters, the prompt shows the stuff I wrote instead of the actual path. 

```
E:\test $ cd SUBFOLDER_1
E:\test\SUBFOLDER_1 $

E:\test $ cd SuBFolDER_1
E:\test\SuBFolDER_1 $
```

In my normal shell it looks like this:
```
E:\test $ cd SuBFolDER_1        
E:\test\subfolder_1 $                       
```

There is also another issue with path completion on windows when folders have white spaces. It is broken. I think I have a fix for that. I will push it when I have tested it a little more. 

/Morten

adqm:
> I think it should expand to the lowercase version.

I agree here, but I think that is hard (I couldn't see a way to change the order in which the completer returns its completions, and it seems always to want to put upper-case first), and I don't think this issue is Windows-specific.

> Another weird thing. When I enter a lower case folder using capital letters, the prompt shows the stuff I wrote instead of the actual path.

This almost has to be a quirk caused by Windows's case-insensitivity; that value is coming from `os.getcwd()`!  I don't know that we'll find a way around that one.  Maybe on Windows, the two are actually the same, and so it doesn't matter which is displayed (?).

> There is also another issue with path completion on windows when folders have white spaces. It is broken.

Depending on what type of broken you mean, that might not be Windows-specific.  I think you're right that a separate Issue/PR is the right place.


I will go ahead and test a little bit on Windows myself, but other than that, I think this is pretty much ready to merge; if we find fixes for those issues later, I think they can come in later.

PR #334, opened 2015-07-30
Some fixes for prompt-toolkit integration.

jonathanslenders:

PR #339, opened 2015-08-02
Prompt toolkit fixes

scopatz:
This contains an additional change over and above #334 that allows the tests to pass.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -23,6 +23,7 @@ the xonsh shell
         "It is pronounced <i>conch</i>",
         "It is pronounced <i>quanxh</i>",
         "It is pronounced <i>zonsch</i>",
+        "It is pronounced <i>jeaunsch</i>",
         "The shell, bourne again",
         "Snailed it",
         "Starfish loves you",
@@ -30,6 +31,7 @@ the xonsh shell
         "This is Major Tom to Ground Xonshtrol",
         "Sally sells csh and keeps xonsh to herself",
         "Nice indeed. Everything's accounted for, except your old shell.",
+        "I wanna thank you for putting me back in my snail shell.",
         ];
     document.write(taglines[Math.floor(Math.random() * taglines.length)]);
     </script>
--- a/tests/test_prompt_toolkit_tools.py
+++ b/tests/test_prompt_toolkit_tools.py
@@ -36,7 +36,7 @@ def test_format_prompt_for_prompt_toolkit():
         ),
         ]
     for test, ans in cases:
-        assert_equal(format_prompt_for_prompt_toolkit(test), ans)
+        yield assert_equal, ''.join(ans), format_prompt_for_prompt_toolkit(test)
 
 
 if __name__ == '__main__':


rbrewer123:
Using the latest `python-prompt-toolkit` from github this seems to be mostly working on Windows, and performance seems pretty good too.  The only problem I'm noticing so far is that the prompt itself is showing the raw color escape codes instead of the colors themselves.  I see the same behavior when running xonsh inside `cmd.exe` or ConEmu.  I haven't had a chance to investigate this further but thought I would note it for now.  My guess is that there are a few lines in xonsh's readline code to deal with that that need to be brought into xonsh's prompt-toolkit code.

melund:
I also tested this briefly on windows. It didn't work until I disabled my fancy colored prompt. I think @jonathanslenders mentioned that prompt-toolkit works with pygments tokens to set the color because ANSI color codes are not consistent across consoles. 

Another annoying thing; The prompt jumps up and down when the autocomplete windows popup. This happens when the prompt is close to the button of the windows. I briefly tried [ptpython](https://github.com/jonathanslenders/ptpython/), and it didn't have the problem. It reserves five lines at the button of the window, so there is always room for the auto-completer window to popup. Maybe it is a setting in prompt-toolkit.


/Morten

scopatz:
Ahh Ok.  Is the color issue something we need to take care of for this PR?

scopatz:
Or can it be a separate issue?

melund:
I guess it can be merged, since it not active by default. On windows it is only useable if the default prompt is disabled. So the .xonshrc have to look like this: 

```cmd
$SHELL_TYPE = 'prompt_toolkit'
$PROMPT = '$'
```

Then we can create a new separate issue.

scopatz:
Alright, merge whenever you want!

rbrewer123:
I'm fine merging it if it's working on one platform, and then we can get parity on the rest later.  I should be able to try it on ubuntu tonight, or @melund you're welcome to merge if you get to it first.  The completion popups are pretty snazzy.

rbrewer123:
I tried running it on my Ubuntu 14.04 box.  I'm using prompt-toolkit installed fresh from github.  I'm seeing the escape characters instead of colors in my prompt, same as on Windows.  So I set `$PROMPT = '$ '` in my `.xonshrc`.  But as soon as I type anything that matches a file in my current directory, I get the following traceback:

```
$ aException in thread Thread-1:
  File "/usr/lib/python3.4/threading.py", line 920, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.4/threading.py", line 868, in run
  File "/usr/local/lib/python3.4/dist-packages/prompt_toolkit-0.46-py3.4.egg/prompt_toolkit/interface.py",n
    CompleteEvent(text_inserted=True)))
  File "/usr/local/lib/python3.4/dist-packages/xonsh/prompt_toolkit_completer.py", line 30, in get_complets
    self.ctx)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/completer.py", line 163, in complete
    rtn |= self.path_complete(prefix)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/completer.py", line 234, in path_complete
    return {_normpath(s) for s in paths}
  File "/usr/local/lib/python3.4/dist-packages/xonsh/completer.py", line 234, in <setcomp>
    return {_normpath(s) for s in paths}
  File "/usr/local/lib/python3.4/dist-packages/xonsh/completer.py", line 62, in _normpath
    initial_dotslash |= (os.altsep and p.startswith(os.curdir + os.altsep))
TypeError: unsupported operand type(s) for |=: 'bool' and 'NoneType'
```

So if I have files starting with `a` in the current directory, as soon as I type `a` I get the traceback.  I also get it as soon as I press space after finishing typing a valid command (either an alias or an executable to run).  

I'll have to do some more experimenting on Windows to see if the same thing is happening.  @scopatz, what environment are you running this on, and are you able to reproduce this?

melund:
@rbrewer123 @scopatz ... This actually a error introduced in my last PR. I didn't have oppotunity to test it on linux. Sorry for that. I will fix it right away.

melund:
I have fixed problem in this https://github.com/scopatz/xonsh/pull/359
@rbrewer123 you are wellcome to merge and try  prompt-toolkit-fixes again....

scopatz:
I just updated this with master, which includes the #359 fixes.

rbrewer123:
@scopatz, I just verified this on Ubuntu and it is working well.  The color escape codes for the prompt remains an issue (probably on Windows too) that we can work on separately.  I also noted what @melund mentioned about the command line jumping up and due to the completion window.  I think reserving space at the bottom of the window (somehow) would be a good solution.  It's pretty cool to see this working, thanks for the PR!

@melund, thanks for the prompt (hehe) fix on this, I didn't even suspect we might have let a little bug squeeze through to master.  :)

PR #340, opened 2015-08-02
fixes hanging indent error

### LINKED ISSUE #315: xonsh hangs on incorrectly indented for loop
```
condamaster%=🐚 for i in range(1):
.............. print(i)
..............

```

Xonsh hangs here. The issue is that the print(i) isn't indented (side-issue, it should indent automatically, just like IPython).

scopatz:
This should fix #315.

### DIFF
--- a/tests/test_execer.py
+++ b/tests/test_execer.py
@@ -4,6 +4,8 @@ import os
 import sys
 import ast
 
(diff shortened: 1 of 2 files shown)


adqm:
Looks good to me!

Issue #297, opened 2015-07-12
Prompt formatter functions return None


scopatz:
If they return None, then this should be the same as returning an empty string.

PR #342, opened 2015-08-02
prompt formatting functions may return None

scopatz:
This should address the issue brought up in #297 by @asmeurer


asmeurer:
Awesome. This will make it easier to write prompt functions that just have side effects.

PR #345, opened 2015-08-04
fix pretty printing bug on windows

adqm:
This bug was mentioned by @BYK in a [post on the mailing list](https://groups.google.com/d/msg/xonsh/fRoEQR0QGeA/6mNa83TABAAJ).  It seems not to be something specific to that example, though, but rather something completely general: on Windows only, any time we attempt to pretty-print a Python value, this error occurs.

This patch should fix (tested quickly on Windows 8 and Debian Jessie).

### DIFF
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -57,7 +57,7 @@ parser.add_argument('args',
 
 def _pprint_displayhook(value):
     if value is not None:
-        __builtins__['_'] = value
+        builtins._ = value
         pprint(value)
 
 


scopatz:
Thanks!

melund:
@adqm : Weird I never got this error on windows. Just tested with different older builds. 

The fix doesn't seem to cause problems for me either. So I guess it is ok. 

However, doing: `$(ls).split()` has an other side effect which I first noticed now. It captures all the color codes as well. Is this also Windows related? 

```
~\Documents\xonsh (master)
$ $(ls).split()
['\x1b[0m\x1b[0mCONTRIBUTING\x1b[0m',
 '\x1b[01;34mdocs\x1b[0m/',
 '\x1b[0mlicense\x1b[0m',
 '\x1b[01;34mrecipe\x1b[0m/',
 '\x1b[01;34mscripts\x1b[0m/',
 '\x1b[01;34mtests\x1b[0m/',
 '\x1b[01;34mxonsh\x1b[0m/',
 '\x1b[01;32msetup.py\x1b[0m*',
 '\x1b[0mREADME.rst\x1b[0m',
 '\x1b[0mrequirements.txt\x1b[0m',
 '\x1b[m']

Issue #293, opened 2015-07-12
Add dirname, etc to FORMATTER_DICT


scopatz:
Would be nice .

scopatz:
Hey @asmeurer - what did you mean by this exactly.  They way Python string formatting works seems to forbid the calling of functions in the formatter.  The following could not work

```python
'{dirname({x})}'.format(dirname=os.path.dirname, x='/yo/mama')
```

The following could be made to work, but it is a little weird:

```python
'{dirname[{x}]}'.format(dirname=os.path.dirname, x='/yo/mama')
```

asmeurer:
I think I just wanted a special variable for `basename(getcwd())` so I don't have to write https://github.com/asmeurer/dotfiles/blob/13f21dbbb8ae87dc7b2f544f53a8cf957c5f2639/.xonshrc#L18 (which incidentally I would keep anyway because of the "current working directory not found" bit, but maybe xonsh could do something there as well). It seems useful. In bash it's `\W`, which is a pretty common variable for PS1s.

Also, in bash, `\W` is `~` is the cwd is $HOME.

asmeurer:
"dirname" probably came from me misremembering the dozens of os.path functions btw.

scopatz:
Ok - that is very doable.

PR #346, opened 2015-08-04
Added more cwd formatters

scopatz:
Should address the issues #293 raised by @asmeurer

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -910,6 +910,10 @@ By default, the following variables are available for use:
   * ``user``: The username of the current user
   * ``hostname``: The name of the host computer
   * ``cwd``: The current working directory
+  * ``dir_cwd``: The dirname of the current working directory, e.g. ``/path/to`` in
+    ``/path/to/xonsh``.
+  * ``base_cwd``: The basename of the current working directory, e.g. ``xonsh`` in
+    ``/path/to/xonsh``.
   * ``curr_branch``: The name of the current git branch (preceded by space),
     if any.
   * ``branch_color``: ``{BOLD_GREEN}`` if the current git branch is clean,
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -910,9 +910,9 @@ By default, the following variables are available for use:
   * ``user``: The username of the current user
   * ``hostname``: The name of the host computer
   * ``cwd``: The current working directory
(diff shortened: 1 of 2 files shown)


xonsh/environ.py
```diff
@@ -396,12 +396,15 @@ def _replace_home(x):
...
-                      hostname=socket.gethostname().split('.', 1)[0],
-                      cwd=lambda: _replace_home(builtins.__xonsh_env__['PWD']),
-                      curr_branch=current_branch,
-                      branch_color=branch_color,
-                      **TERM_COLORS)
+FORMATTER_DICT = dict(
+    user=os.environ.get(USER, '<user>'),
+    hostname=socket.gethostname().split('.', 1)[0],
+    cwd=lambda: _replace_home(builtins.__xonsh_env__['PWD']),
+    dir_cwd=lambda: os.path.dirname(_replace_home(builtins.__xonsh_env__['PWD'])),
```
adqm: Should we make a named (private) function that returns ``_replace_home(builtins.__xonsh_env__['PWD'])`` so that we can use that for `cwd`, `dir_cwd`, and `base_cwd`, rather than repeating that code three times?

docs/tutorial.rst
```diff
@@ -910,6 +910,10 @@ By default, the following variables are available for use:
   * ``user``: The username of the current user
   * ``hostname``: The name of the host computer
   * ``cwd``: The current working directory
+  * ``dir_cwd``: The dirname of the current working directory, e.g. ``/path/to`` in
```
adqm: For some reason, these names feel strange to me.  Can we call these `cwd_dir` and `cwd_base`?

Even then, `cwd_dir` feels weird since `cwd` is, itself, a directory...

adqm:
Hi @scopatz, I'm fine to merge this in, I think.  Just a couple of thoughts/questions above.

scopatz:
Good suggestions, thanks @adqm.  I've implemented them now.

adqm:
Alright, looks good!  Merging.

docs/tutorial.rst
asmeurer: Yeah, isn't "cwd_dir" kind of like "ATM machine"?

scopatz: A bit :).  Feel free to PR in a better name.

PR #347, opened 2015-08-05
Added DIRS rule to environ

scopatz:
Appearently, some people - like the [freedesktop standards](http://standards.freedesktop.org/basedir-spec/latest/ar01s02.html) - use `DIRS` instead of `PATH`.  A bit odd, but easy to support.


adqm:
Looks good to me.  Merging now.

scopatz:
Thanks @adqm!

PR #348, opened 2015-08-05
Fix error when checking for indentation

BYK:
I have the following portion in my `.xonshrc` file:

    if not pid_exists($SSH_AGENT_PID):
        # Run SSH Agent + Add SSH Keys
        for l in re.split(r'[\n;]', $('ssh-agent')):
            if '=' not in l: continue
            
            parts = l.split('=', 1)
            os.system("SETX {0} {1}".format(*parts))
            builtins.__xonsh_env__[parts[0]] = parts[1]
    
        ssh-add ~/.ssh/id_rsa

And it was throwing an error at the last line saying index was out of range. Figure it was because `lines[idx-1].rstrip()` was an empty string so `lines[idx-1].rstrip()[-1]` was the culprit.

Not sure if this is the best way to fix it and how I can add tests for this.

### DIFF
--- a/xonsh/execer.py
+++ b/xonsh/execer.py
@@ -178,7 +178,8 @@ class Execer(object):
                     last_error_line = last_error_col = -1
                     input = '\n'.join(lines)
                     continue
-                if last_error_line > 1 and lines[idx-1].rstrip()[-1] == ':':
+
+                if last_error_line > 1 and lines[idx-1].rstrip()[-1:] == ':':
                     # catch non-indented blocks and raise error.
                     prev_indent = len(lines[idx-1]) - len(lines[idx-1].lstrip())
                     curr_indent = len(lines[idx]) - len(lines[idx].lstrip())
--- a/tests/test_execer.py
+++ b/tests/test_execer.py
@@ -79,6 +79,12 @@ def test_bad_indent():
             'x = 1\n')
     assert_raises(SyntaxError, check_parse, code)
 
+def test_indent_with_empty_line():
+    code = ('if True:\n'
+            '\n'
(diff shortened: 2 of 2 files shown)


xonsh/execer.py
```diff
@@ -178,7 +178,9 @@ def _parse_ctx_free(self, input, mode='exec'):
                     last_error_line = last_error_col = -1
                     input = '\n'.join(lines)
                     continue
-                if last_error_line > 1 and lines[idx-1].rstrip()[-1] == ':':
+
+                stripped_line = lines[idx-1].rstrip()
+                if last_error_line > 1 and stripped_line and stripped_line[-1] == ':':
```
scopatz: I think that you could condense this down to:

```python
if last_error_line > 1 and lines[idx-1].rstrip()[-1:] == ':':
```

Nice catch though.

BYK: Isn't that the exact same line as before:

    if last_error_line > 1 and lines[idx-1].rstrip()[-1] == ':':

am I missing something here? :)

BYK: Oh wait, I see the difference now lol.

scopatz: pretty subtle :)

BYK: Done.

Where should I add the tests for this?

scopatz:
There is a test for situations like this in https://github.com/scopatz/xonsh/blob/master/tests/test_execer.py#L77  You can add another if you'd like!

BYK:
Added a small test case

scopatz:
Thanks @BYK!

PR #353, opened 2015-08-12
Rich History

scopatz:
This PR adds a rich history object and picks up where @aig787 left off. 

History is now a set of LazyJSON files that store the following data:

```js
{'env': {},  // Environment that xonsh was started with
 'sessionid': str, // UUID4 for the session
 'ts': [start, stop],  // start and stop timestamps for session [s since epoch]
 'locked': True, // boolean for whether the file is in use or not
 'cmds': [  // array of commands
    {'inp': str, // input command
     'ts': [start, stop], // timestamps for the command
     'rtn': int, // command return code
     'out' str, // stdout and stderr of command, for subproc commands this is only available on Linux.
     }, 
    ...
  ],
}
```

This PR does not address the things that one might want to *do* with such history. It was getting a bit heavyweight, and I figured that those could come in smaller, dedicated PRs.  

Also, capturing subproc output while remaining responsive in the same terminal was **so much more difficult** than I had anticipated.  This is why it is only implemented for Linux right now.  Implementation and testing on other platforms should come in time :).


adqm:
Woo-hoo!  Taking a look now.

Looks like we are still failing on the AppVeyor build.  I don't know anything about that, so I'll leave it to you :)

xonsh/aliases.py
```diff
@@ -65,9 +67,70 @@ def xexec(args, stdin=None):
...
             os.execvpe(args[0], args, denv)
         except FileNotFoundError as e:
-            return "xonsh: " + e.args[1] + ": " + args[0] + "\n"
+            return 'xonsh: ' + e.args[1] + ': ' + args[0] + '\n'
     else:
-        return "xonsh: exec: no args specified\n"
+        return 'xonsh: exec: no args specified\n'
+
+
+def history(args, stdin=None):
```
adqm: Just to be clear, this command and the other new aliases in this file are something you want to wait on?  Looks like they are just kind of stubbed out for now.

aig787:
@scopatz thanks for finishing this off! Its been on my list but things haven't slowed down.

xonsh/aliases.py
scopatz: Yeah, they were written by @aig787 originally and so they should just be considered stubs for the moment.  I imagine they'll come pretty quick once this is in :)

xonsh/base_shell.py
```diff
@@ -38,12 +117,25 @@ def default(self, line):
...
             return
+        hist = builtins.__xonsh_history__
+        ts1 = None
+        tee = Tee() if builtins.__xonsh_env__.get('XONSH_STORE_STDOUT', False) \
+                    else io.StringIO()
         try:
+            ts0 = time.time()
             self.execer.exec(code, mode='single', glbs=self.ctx)  # no locals
+            ts1 = time.time()
+            hist.last_cmd_rtn = 0  # returncode for success
```
adqm: If we start a process and it fails, it looks like this will still come up 0 ("success") since that exception isn't raised up, right?

scopatz:
Thanks gents! I am looking at the windows build now.  It is some weird file system locking issue in the tests.

xonsh/built_ins.py
```diff
@@ -528,17 +531,21 @@ def run_subproc(cmds, captured=True):
                        universal_newlines=uninew)
         else:
             prev_is_proxy = False
+            
+            usetee = (stdout is None) and (not background) and \
+                     ENV.get('XONSH_STORE_STDOUT', False) 
+            cls = TeePTYProc if usetee else Popen
             subproc_kwargs = {}
-            if ON_POSIX:
+            if ON_POSIX and cls is Popen:
```
adqm: I haven't tested this yet, but I am worried that without the `_subproc_pre`, job control will probably not work as expected...anyway, will test.

xonsh/proc.py
```diff
@@ -336,9 +342,67 @@ def wrapped_simple_command(args, stdin, stdout, stderr):
                         stderr.write(r[1])
                 elif r is not None:
                     stdout.write(str(r))
-                return True
+                return 0  # returncode for succees
```
adqm: I was hoping to use `True` and `False` for these values for the `and`/`or` business, but we can discuss that once I've gotten back into gear and finished that off...

xonsh/history.py
```diff
@@ -0,0 +1,192 @@
...
+            hist = lazyjson.LazyJSON(f).load()
+        hist['cmds'].extend(self.buffer)
+        if self.at_exit:
+            hist['ts'][1] = time.time()  # apply end time
+            hist['locked'] = False
+        with open(self.filename, 'w') as f:
+            lazyjson.dump(hist, f, sort_keys=True)
+
+
+class History(object):
```
adqm: Maybe I am just being dense, but is there a clean way to get access to the elements in the history?  From what I could tell, I had to do:
```
from xonsh.lazyjson import LazyJSON
LazyJSON(__xonsh_history__.filename).load
```
to get something that looked like a Python object I could mess with.

Can we add indexing, etc to the ``History`` class directly to make accessing the history easier?

scopatz:
OK!  I figured it out.  The lazyjson stuff was failing on windows because offsets were being miscalculated due to newlines taking up two characters, rather than just one.  This is fixed by passing `newline='\n'` into the file handle.

xonsh/base_shell.py
scopatz: Ahh good point, this should only be set if the last command is none.

xonsh/built_ins.py
scopatz: This seemed to work with backgrounding, which uses the same infrastructure.  Though you are right, it probably needs testing.

xonsh/proc.py
scopatz: Hmm, yeah, I understand. They definitely got this backwards with `0` being True.  On the otherhand, it is such a strong part standard/convention/whatever that it is probably worth it to not break this. For the `and/or` stuff, testing `rtn == 0` probably isn't the worst thing in the world, though it is annoying :)

xonsh/history.py
scopatz: There is nothing like that yet.  All the information exists, it just hasn't been exposed to the user yet.  I can do that as part of this PR, for sure, but I was hoping to do that in one that is coming up next.

scopatz:
Also, note that once some of the future PRs related to history are in, I am planning on writing an "advanced" or "history" tutorial page that goes into these features.

xonsh/environ.py
```diff
@@ -454,8 +457,9 @@ def multiline_prompt():
     'TITLE': DEFAULT_TITLE,
     'MULTILINE_PROMPT': '.',
     'XONSHRC': os.path.expanduser('~/.xonshrc'),
-    'XONSH_HISTORY_SIZE': 8128,
-    'XONSH_HISTORY_FILE': os.path.expanduser('~/.xonsh_history'),
+    'XONSH_HISTORY_SIZE': (8128, 'files'),
+    'XONSH_HISTORY_FILE': os.path.expanduser('~/.xonsh_history.json'),
+    'XONSH_STORE_STDOUT': True,
```
adqm: I might argue that this should default to `False`, since I'm not sure it will be used terribly often and it has the potential to take up a lot of disk space.  What do you think, @scopatz?

scopatz: Good question @adqm.  So I think that it should be on by default. This is because my main motivation for writing it up like this was to enable reproducible workflows. Some things I plan on implementing in near future are a replay functionality as well as a semantic diffing of history files. Not having stdout recording on makes it impossible to recover "OMG what did it just do!" moments.

That said, people will certainly have different limits for what they want to store. This is why I implemented garbage collection based on total file size.  Maybe we should change the default `XONSH_HISTORY_SIZE` from `(8128, 'files')` to be `(16, 'MB')` or some other, similar small number.  

 IMO this is going to be a feature that really sets xonsh apart from other shells.  But it is new.  Maybe we should have a buy-in moment when xonsh is first started. For example, we could have a warning message on startup, until they set a certain config variable.  The user would only have to  do this once and could either turn it on or off, by default, as well as set a default size.  How does that sound?

scopatz:
Added the ability to index into history by field name.  Also started to fix up `!n` and `!!` aliases, but the stdout capturing isn't playing nicely with these aliases, sadly.  I still think that this could go in as is, subject to any other comments.

scopatz:
...that I might not have addressed :)

xonsh/teepty.py
```diff
@@ -0,0 +1,234 @@
...
+import tempfile
+
+# The following escape codes are xterm codes.
+# See http://rtfm.etla.org/xterm/ctlseq.html for more.
+MODE_NUMS = ('1049', '47', '1047')
+START_ALTERNATE_MODE = frozenset('\x1b[?{0}h'.format(i).encode() for i in MODE_NUMS)
+END_ALTERNATE_MODE = frozenset('\x1b[?{0}l'.format(i).encode() for i in MODE_NUMS)
+ALTERNATE_MODE_FLAGS = tuple(START_ALTERNATE_MODE) + tuple(END_ALTERNATE_MODE)
+
+RE_HIDDEN = re.compile(b'(\001.*?\002)')
```
asmeurer: This is entirely pedantic, but if I understand the readline behavior correctly, the regular expression should be `\001[^\001\002]*\002`.

asmeurer: Actually, maybe I have it backwards, and yours is correct. Maybe I should check what readline actually does instead of just trying to read the code.

scopatz: FWIW I think yours would work too.

rbrewer123:
@scopatz, is this PR supposed to be writing to `$XONSH_HISTORY_FILE`?  I'm not seeing that file get created.  The files in `$XONSH_DATA_DIR` look good.  (Currently testing on Ubuntu).

scopatz:
@rbrewer123 - nope, the `$XONSH_HISTORY_FILE` is no longer used.  That was a readline feature that we were using in lieu of this PR.  The `$XONSH_DATA_DIR` is used instead and follows the open desktop standard for where files should live.

scopatz:
OK! I have fixed some lingering (and highly annoying) UI issues with readline.  This should be truly ready to go now.  I'd really appreciate anyone reviewing this!

scopatz:
If there are things people still think I need to do, please let me know. I'll be happy to do them.. I just want to see this get in :)

adqm:
This is working really well!  A couple of things:

I hit an issue with ``!n``:
```
hartz@lappy-511 ~ $ !n 2
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK=True
NameError: name 'cd' is not defined
```

Looks like it was trying to execute a subprocess-mode command in Python mode?  It didn't happen every time, though, so I am not sure what is going on there.  I haven't yet been able to reliably reproduce.

Also, I still can't look up values in the history object:
```
hartz@lappy-511 ~ $ __xonsh_history__['cmds']
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK=True
TypeError: 'History' object is not subscriptable
```

scopatz:
Hi @adqm! It is hard to say what is going on with the `!n` without seeing the full history.  I am able to use this `!n 2` even when this refers a cd command.  Also note that the `!n` is zero indexed and accepts negative numbers.

As per the history lookup, you have to lookup by field. The fields are Sequences that know when and how to lookup the desired index, ie whether it is in the buffer in memory on on disk.  So for example, to get the previous input you would do like:

```python
__xonsh_history__.inps[-1]
```

Right now, the following fields are available:

```python
__xonsh_history__.tss    # timestamps
__xonsh_history__.inps  # input strings
__xonsh_history__.outs  # output strings
__xonsh_history__.rtns  # return values. 
```

I could add a `cmds` field that returns the whole command dictionary too if you'd like.

adqm:
Also, a big issue.  Somehow, after locking the screen and coming back:
```
hartz@lappy-511 ~/myxonsh scopatz-hist $ getmail
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK=True
PermissionError: [Errno 1] Operation not permitted
hartz@lappy-511 ~/myxonsh scopatz-hist $ cd
    main()
  File "/usr/local/lib/python3.4/dist-packages/xonsh/main.py", line 101, in main
    shell.cmdloop()
  File "/usr/local/lib/python3.4/dist-packages/xonsh/readline_shell.py", line 232, in cmdloop
    self._cmdloop(intro=intro)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/readline_shell.py", line 216, in _cmdloop
    self._load_remaining_input_into_queue()
  File "/usr/local/lib/python3.4/dist-packages/xonsh/readline_shell.py", line 124, in _load_remaining_input_into_queue
    buf += os.read(self.stdin.fileno(), 1024)
OSError: [Errno 5] Input/output error
```

In fresh terminals, those commands work fine...not sure what is going on.

adqm:
> As per the history lookup, you have to lookup by field. 

Woops.  That's what I get for just testing before carefully reading the code.

> Also note that the !n is zero indexed and accepts negative numbers.

Yes; I tried that as well.  Worked most of the time.

scopatz:
> In fresh terminals, those commands work fine...not sure what is going on, but it seems like a teepty issue.

Yeah, I have noticed some weirdness in going from master to/from the hist branch in the same terminal.  I have never noticed an issue in a fresh terminal though.

xonsh/environ.py
scopatz: Also, @adqm, any opinion on these issues?

adqm: I agree that this is a very nice feature to have, and is potentially something that could set xonsh apart from competitors, but I guess my feeling is still that the best defaults are conservative ones, so I would still advocate not collecting stdout by default.

If we don't want to go that route, though, then I actually think limiting based on number of lines is better (since with a filesize limit, we could theoretically start garbage collecting (and, thus, losing history) after a very small number of commands).

A warning message might be alright, but I think that probably isn't necessary so long as this behavior is well-documented somewhere.  In particular, I think if this "replay" capability is explained in the tutorial (along with `$XONSH_STORE_STDOUT`), then I think we are probably fine.

adqm:
Also also, piping to vim seems not to work anymore:
```
hartz@lappy-511 ~ $ echo "hello" | vim -
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Mar 31 2015 23:36:07)
Too many edit arguments: "/tmp/tmp8ug2qpf8"
More info with: "vim -h"
```

scopatz:
Ok! I have a fix for the vim issue.  I was cheating with tempfiles before. Thanks for keeping me honest.

scopatz:
Alright, @adqm - I have switched off `$XONSH_STORE_STDOUT` by default, added a 'commands' unit for garbage collection (which is easier to logic about than lines, and much cheaper to gather data on!), and changed the default history size to `(8128, 'commands')`.

scopatz:
I won't worry about the warning message for now then, too, since it doesn't seem to be necessary with these defaults.

scopatz:
I have now updated the `history` command as well. Everything that was there before from @aig787 should be working now.

scopatz:
OK! I have now also added replay capability.

scopatz:
Discovered a bug with piping to less as a result of my fix for piping to `vim -`.

melund:
Hi Anthony @scopatz . I tried to open run with history branch. The first issue I encountered was here: https://github.com/scopatz/xonsh/blob/hist-next/xonsh/readline_shell.py#L277

When pyreadline is used this should be:
```
if RL_LIB:
   RL_LIB.history_set_pos(i)
```

The next error I got when typing a command: 
``` 
[xonsh]~\Documents\xonsh (hist)
$ ls
  File "C:\Users\mel\Anaconda3\envs\xonsh\Scripts\xonsh-script.py", line 6, in <module>
    sys.exit(main())
  File "C:\Users\mel\Documents\xonsh\xonsh\main.py", line 101, in main
    shell.cmdloop()
  File "C:\Users\mel\Documents\xonsh\xonsh\readline_shell.py", line 230, in cmdloop
    self._cmdloop(intro=intro)
  File "C:\Users\mel\Documents\xonsh\xonsh\readline_shell.py", line 214, in _cmdloop
    self._load_remaining_input_into_queue()
  File "C:\Users\mel\Documents\xonsh\xonsh\readline_shell.py", line 119, in _load_remaining_input_into_queue
    r, w, x = select.select([self.stdin], [], [], 1e-6)
OSError: [WinError 10038] An operation was attempted on something that is not a socket
```

scopatz:
Thanks for reporting @melund.  I implemented your fist fix.  However, the second one is interesting.  You shouldn't be seeing TeePTY at all since you are on Windows.  I used the `tools.ON_LINUX` to check for this.  Are you using cygwin or something?

scopatz:
BTW - you can see this condition here: https://github.com/scopatz/xonsh/blob/hist/xonsh/proc.py#L355

scopatz:
OK! I have now also (finally) added a history differ. I think @melund's issue is the only one blocking this now.

melund:
Ohh. I am sorry that I am blocking it :)  I don't use cygwin but the  newest version of git for windows include msys which is similar. When installing git it is possible to add all the UNIX tools to the path. That is very handy and many users will do the same as me ...

scopatz:
Well, you are not blocking it :) But I want to make sure that this doesn't break your system when it gets merged in.  What is the result of 

```python
import platform
print(platform.system())
```

melund:
That prints 'Windows'

I still get the error. No matter what I type in the shell.
``` 
[xonsh]~
$ 1+1
  File "C:\Users\mel\Anaconda3\envs\xonsh\Scripts\xonsh-script.py", line 6, in <module>
    sys.exit(main())
  File "C:\Users\mel\Documents\xonsh\xonsh\main.py", line 101, in main
    shell.cmdloop()
  File "C:\Users\mel\Documents\xonsh\xonsh\readline_shell.py", line 230, in cmdloop
    self._cmdloop(intro=intro)
  File "C:\Users\mel\Documents\xonsh\xonsh\readline_shell.py", line 214, in _cmdloop
    self._load_remaining_input_into_queue()
  File "C:\Users\mel\Documents\xonsh\xonsh\readline_shell.py", line 119, in _load_remaining_input_into_queue
    r, w, x = select.select([self.stdin], [], [], 1e-6)
OSError: [WinError 10038] An operation was attempted on something that is not a socket
``` 

I don't have the problem with prompt_toolkit. However, with prompt_toolkit I get an other error when I exit xonsh:

```
[xonsh]~
$ exit

Exception ignored in: <bound method PromptToolkitShell.__del__ of <xonsh.prompt_toolkit_shell.PromptToolkitShell object at 0x0000000003377828>>
  File "C:\Users\mel\Documents\xonsh\xonsh\prompt_toolkit_shell.py", line 57, in __del__
    teardown_history(self.history)
  File "C:\Users\mel\Documents\xonsh\xonsh\prompt_toolkit_shell.py", line 40, in teardown_history

    history.save_history_to_file(hfile, hsize)
  File "C:\Users\mel\Documents\xonsh\xonsh\prompt_toolkit_history.py", line 83, in save_history_to_file
    if limit < 0:

```

scopatz:
Ok @melund - I put in a fix that should work for readline.  Not sure what is going on with prompt-toolkit. Is there more of the error message?

melund:
I will test it tomorrow. It is getting late here. 
I made a mistake pasting the error message. Here is full trace, 

$ exit
```
Exception ignored in: <bound method PromptToolkitShell.__del__ of <xonsh.prompt_toolkit_shell.PromptToolkitShell object at 0x0000000003647668>>
  File "C:\Users\mel\Documents\xonsh\xonsh\prompt_toolkit_shell.py", line 57, in __del__
    teardown_history(self.history)
  File "C:\Users\mel\Documents\xonsh\xonsh\prompt_toolkit_shell.py", line 40, in teardown_history
    history.save_history_to_file(hfile, hsize)
  File "C:\Users\mel\Documents\xonsh\xonsh\prompt_toolkit_history.py", line 83, in save_history_to_file
    if limit < 0:
TypeError: unorderable types: tuple() < int()
```

scopatz:
Ahh Ok.  I know what the error is.

scopatz:
Alright! I have a fix in for the prompt-toolkit exiting now too.

asmeurer:
Is it possible to tie each history to the tty number? It would be awesome if when I restarted my terminal each tab remembered its own history.

melund:
@scopatz . That fixed both the errors I had.

scopatz:
@asmeurer 

> Is it possible to tie each history to the tty number? It would be awesome if when I restarted my terminal each tab remembered its own history

In theory, yes. That would be a separate PR though, for sure.

@melund, awesome! I have now added a tutorial to the docs, and discovered and fixed a couple of bugs in the process. I am not sure that there is anything left to do for this PR. 

I'd really appreciate another review and/or merge :)

docs/tutorial_hist.rst
```diff
@@ -0,0 +1,402 @@
...
+environment as well as the input/output.  This cycle of do-replay-diff is more meaningful than
+a traditional, "What did I/it/the Universe just do?!" approach.
+
+Of course, nothing has ever stopped anyone from pulling Unix tools like ``env``, ``script``, 
+``diff``, and others together to deliver the same kind of capability. However, in practice, 
+no one does this. With xonsh, rich and useful history come batteries included.
+
+``history`` command
+====================
+All xonsh history inspection and manipulation goes the the top-level ``history`` alias or
```
melund: .... goes *through* the top-level ...

scopatz: fixed, thanks!

docs/tutorial_hist.rst
```diff
@@ -0,0 +1,402 @@
...
+**new history**
+
+.. code-block:: xonshcon
+
+    >>> history replay -o ~/new.json ~/.local/share/xonsh/xonsh-4bc4ecd6-3eba-4f3a-b396-a229ba2b4810.json
+    2  10
+    /home/scopatz/new.json
+
+    ------------------------------------------------------------
+    Just replayed history, new history has following information
```
melund: ... has *the* following ...

scopatz: Thanks, got it.

melund:
I read though tutorial and tried many of the commands my self. Very nice. I think it is time to release this beast into the wild. Then any small changes and improvement can get their own PR.

scopatz:
Whoohoo! Well feel free to merge, whenever you want :).

scopatz:
Oops! you beat me to it!

melund:
He he... I really look forward to try this some more.

scopatz:
Thanks @melund!  I also really want to thank @adqm for hammering on this so much. And @aig787, of course, deserves a ton of credit for getting this started.  I'll send a message out to the mailing list now.

melund:
Yes. Great work. I am off now. I have two kids that usually wake up 5 hours from now.

scopatz:
Haha, well get some rest ;)

PR #355, opened 2015-08-12
allow xonsh to run unreadable binaries

larsks:
this change causes `_is_binary` to return `True` if it receives `EACCES`
when opening a file for reading.  This will allow `xonsh` to run a
binary like `sudo` which is often installed with mode 4111 (s--x--x).

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -4,6 +4,7 @@ not to be confused with the special Python builtins module.
 import os
 import re
 import sys
+import errno
 import shlex
 import signal
 import inspect
@@ -266,15 +267,22 @@ def _get_runnable_name(fname):
 
 
 def _is_binary(fname, limit=80):
-    with open(fname, 'rb') as f:
-        for i in range(limit):
-            char = f.read(1)
-            if char == b'\0':
-                return True
-            if char == b'\n':
-                return False
-            if char == b'':
-                return False
+    try:
+        with open(fname, 'rb') as f:
+            for i in range(limit):
+                char = f.read(1)
+                if char == b'\0':
+                    return True
+                if char == b'\n':
+                    return False
+                if char == b'':
+                    return False
+    except IOError as exc:
+        if exc.errno != errno.EACCES:
+            raise
+
+        return True
+
     return False
 
 


xonsh/built_ins.py
```diff
@@ -266,15 +267,22 @@ def _get_runnable_name(fname):
...
+            for i in range(limit):
+                char = f.read(1)
+                if char == b'\0':
+                    return True
+                if char == b'\n':
+                    return False
+                if char == b'':
+                    return False
+    except IOError as exc:
+        if exc.errno != errno.EACCES:
```
scopatz: Thanks for the PR @larsks! I think that this lets too many files through, though.  I think that this will allow the execution of files with any permision.  Is there any way that you could please refactor this to check for the correct permission bits? Thanks again

esafak:
@scopatz This PR is in response to my observation that sudo doesn't run. Why is that?

scopatz:
Hi @larsks - I understand that.  This is probably because of the rws--x--x permissions rather than.  Right now we are probably just checking for execute permissions (rwx--x--x, or similar). We need to also check for the "s" sticky bit (http://www.linuxquestions.org/questions/linux-newbie-8/what-is-s-instead-of-x-in-the-file-permission-when-i-look-at-usr-bin-chsh-223386/).  

This PR, it seems, would attempt to execute any file, even those that did not have execute bits set, such as (rw-r--r--).

larsks:
@scopatz I'm not sure I follow your comment about letting "too many files through".  If the file isn't executable -- which is really the only permission bit we should care about when trying to launch a command -- then an attempt to run it will fail with "permission denied".

If the file is executable, this will Just Work.

The `s` bit is not the "sticky" bit, this is the SUID bit and is used to run a binary with elevated privileges.  Sudo -- and similar security-sensitive SUID programs -- are often installed such that they are not readable, but only executable. I don't really see a solution other than what I have proposed here.

scopatz:
`os.access()` provides specific functionality for checking SUIDs. See https://docs.python.org/3/library/os.html#os.access.  I think that something like the following would be better.  

```console
scopatz@athenaie ~ $ import os
scopatz@athenaie ~ $ os.access('/usr/bin/sudo', os.X_OK)
True
``

Or in this case, you could just return `os.access(fname, os.X_OK)`

**Comment by larsks** [comment]
I guess I'm not clear on why you're bothering to call `os.access`, rather than just trying to run the binary and let it fail or not.  I'm not really familiar with the `xonsh` code so probably someone else is more qualified to submit a patch that meshes well with the way you're doing things.

**Comment by scopatz** [comment]
Hi @larsks - I'm encouraging you to reopen your PR. I'd really like for xonsh to be a place where people feel comfortable putting in patches, large or small.  Fixing something the addresses your needs is a great way to get involved.  

I think that this is a one or two line change. The reason that I am suggesting this is that I feel that a standard library solution (os.access) is more robust than trying to roll the same functionality ourselves.  If you want to keep the code that you wrote originally, that is fine too.  But in that case I would just request that you add some tests for the functionality that you added (which would be more than a couple of lines).  I hope that makes sense...

As I said, I really do appreciate you putting in this PR in the first place.  I think that it is an important and valuable contribution to address the issue you raised. I have had a long, stressful, and distracting day myself.  I am sorry if any of that came through in my review.

**Comment by adqm** [comment]
@scopatz, another thing to note is that we already check that the file in question is executable in `get_script_subproc_command`; it wouldn't have made it down to the call to `_is_binary` if the file weren't executable.

So I think this is a fine idea.  I might argue that it shouldn't live in `_is_binary`, though (since the name suggests that returning `True` necessarily means that the file in question is a binary, which wouldn't be true anymore after this change).

I think this might want to live directly in `get_script_subproc_command`.  Maybe we could wrap the following with the try/except instead?
```
    # find interpreter
    with open(fname, 'rb') as f:
        first_line = f.readline().decode().strip()
    m = RE_SHEBANG.match(first_line)
```

PR #356, opened 2015-08-14
fixed SUID issues, I think

scopatz:
I would appreciate it if @larsks and @adqm could take a look. Thanks!

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -300,8 +300,13 @@ def get_script_subproc_command(fname, args):
     if not os.access(fname, os.X_OK):
         raise PermissionError
 
-    # if the file is a binary, we should call it directly
-    if _is_binary(fname):
+    if ON_POSIX and not os.access(fname, os.R_OK):
+        # on some systems, sudo will have execute permissions but not read/write 
+        # permisions. This enables things with the SUID set to be run. Needs to come
+        # before _is_binary() is called, because that function tries to read the file.
+        return [fname] + args
+    elif _is_binary(fname):
+        # if the file is a binary, we should call it directly
         return [fname] + args
 
     if ON_WINDOWS:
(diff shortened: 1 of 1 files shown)


larsks:
This completely fails to address the problem I reported.  The problem is in the `_is_binary` method, which blows up when run on a file that is not readable, such as a binary with 4111 (that is `s--x--x--x`, which means "suid and executable but **not readable**").

That is why my patch would return `True` if there was an `EACCES` error trying to open the file in `_is_binary`.  This is the *only* possible course of action in this situation, because since the file is not readable *you cannot make any determination about its content*.

With this patch in place, attempting to run `sudo` when it has mode 4111 will still result in:

    $ sudo ls
    xonsh: subprocess mode: permission denied: sudo

scopatz:
OK, I see what you are saying.  I believe that I have fixed the logic issue with reading files that don't have read permission.

adqm:
This looks like it has the right logic now (the question of checking this permission rather than using try/except is, I think, inconsequential).  What is the reason for the `ON_POSIX` check?  Is it not possible for this to happen on Windows?

scopatz:
I don't know what is / is not possible on Windows, but I assume that their permissions model is different and doesn't support SUID. The `ON_POSIX` check was just to be safe since the only complaints about executable, non-readable files have been from sudo.

larsks:
With the latest commit things seem to work.  I would rewrite the comments to be less specific to `sudo`, because there may be other binaries with similar configurations (e.g., on my system, `ssh-agent` is installed with no read access).

I confess to not understanding the purpose of the `get_script_subproc_command` method in general, unless it's largely to make things work on Windows environments.  In a Unix environment, I feel as if the whole thing could be replaced by `return [fname] + args` and it would be simpler and more robust.

scopatz:
Updated comment to be less sudo specific. Thanks for testing.  I think that this is ready to be merged.

adqm:
> I confess to not understanding the purpose of the get_script_subproc_command method in general, unless it's largely to make things work on Windows environments. In a Unix environment, I feel as if the whole thing could be replaced by return [fname] + args and it would be simpler and more robust.

It is largely to deal with Windows, yes, since I don't believe Windows pays any attention to shebangs by default.  There is also a question of what to do if a script is marked as executable but has no shebang.  We currently default to using xonsh as the interpreter in that case.

I do agree that this piece of code could probably be simplified, but that can wait for a different PR, I think.  For now, I will merge this in.

PR #357, opened 2015-08-14
Update _normpath wrapper to prevent removing a trailing slash

melund:
This is a small update to _normpath() wrapper function. Right now it removes any trailing slashes (at least on Windows), which is annoying when completing directory names. 

I have been thinking a bit about the path completion in general. On windows, it doesn't work well when there are white spaces in the files names. I tried different things to fix the path_complete(). But the real problem is that the completions are only done from the last white space. I haven't any real fixes for this yet.

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -9,6 +9,9 @@ import sys
 
 from xonsh.built_ins import iglobpath
 from xonsh.tools import subexpr_from_unbalanced
+from xonsh.tools import ON_WINDOWS
+
+
 
 RE_DASHF = re.compile(r'-F\s+(\w+)')
 RE_ATTR = re.compile(r'(\S+(\..+)*)\.(\w*)$')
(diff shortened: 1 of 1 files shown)


scopatz:
Hi @melund - Thanks for putting this in.  I agree that path completion needs some love.  It might be time to implement escape characters for space and possibly backslash...

xonsh/completer.py
```diff
@@ -52,11 +55,22 @@ def startswithnorm(x, start, startlow=None):
...
+    p = p.rstrip()
+    trailing_slash = p.endswith(os.sep) 
+    trailing_slash |= (os.altsep and p.endswith(os.altsep))
+    p = os.path.normpath(p)
+    if initial_dotslash:
+        p = os.path.join(os.curdir, p)
+    if trailing_slash:
+        p = os.path.join(p,'')
+#    if ON_WINDOWS:
+#        p = p.replace(os.sep, os.altsep)
```
scopatz: Can you remove these commented out lines please?

melund: Oh yes. Sorry @scopatz ..  I actually added them for a reason. Just forgot to mention it. I was wondering if it would be an idea to just use forward slashes even on Windows. It should off course accept both separators on windows, but only show forward slashes and convert everything to forward slashes when completing. 

There are of course pro/cons to this. Linux natives may find this nicer when they occasionally work on windows. Even someone like me who mostly use windows finds it much more convenient to use '/' when working in Python. The con is that it may not look as windows like... Who else from the development team uses windows sometimes ( maybe @rbrewer123 have an opinion on this)

BYK: I almost exclusively develop on Windows and I say "`/` FTW!". 

That said this behavior should probably be optional.

scopatz: I think that the default behaviour should be the native behaviour of the platform.  (I think @asmeurer agrees with this.) That said, I am all for a top-level switch for the path endings.  That is probably a separate PR, though (nudge-nudge ~_~).

Before this goes in though, @melund, could you add some more commentary for these lines, such as "# uncomment the following if..."

rbrewer123:
I agree that by default xonsh should work with the platform's behavior.  But optional modifications to that behavior are fine.  I don't think '/' works with all Windows commands, and I think it's best if xonsh minimizes the "magic" it is doing to the commands being issued.

melund:
@rbrewer123 . I didn't think the problem with some subprocess commands not accepting forward slashes. I haven't run into that problem on windows, but you are probably right. If they exist, then it is probably a deal breaker.  My thoughts were on using the autocompleter for paths in python mode. Here backward slashes are very annoying since I would need to go back and escape them all.

melund:
@scopatz. I removed the commented lines. It is should be ready for merge

scopatz:
Thanks @melund!

PR #359, opened 2015-08-18
Fix regression in _normpath. 

melund:
It was failing on Linux where os.altsep evaluates to None. Turned out that |= operator does not work quite as I expected. It only operates on boolean values. That was a problem when mixed with the traditional operators (and/or). (None and False) will for example evaluate to None that can't be feed to '|='


scopatz:
`|=` should also do set unions too.

scopatz:
Thanks for the fix @melund!

PR #363, opened 2015-08-19
Fix error where _normpath wrapper converted ./ to ././

melund:
This fixes yet another oversight where _normpath does not handle './' or '.\' correctly. The current code turns './' into '././', because it sees it as having both an initial 'dot-slash' and a trailing 'slash'.


scopatz:
Thanks @melund!

PR #366, opened 2015-08-19
One fix to be compatible with prompt_toolkit==0.47

jonathanslenders:
There was a very small change in the `History` class. `History` is now an abstract base class, which means that it only specifies the interface and doesn't contain any actual implementation.

This pull request makes it compatible again.


scopatz:
seems reasonable to me.  Maybe @rbrewer123 or @melund or @anula what to make sure that it works?

scopatz:
Looks good. thanks!

PR #368, opened 2015-08-20
First attempt to make a hack for prompt-toolkit coloring

melund:
This may not be ready for merging yet.

I made an attempt to implement the suggestions from @jonathanslenders to get coloring working with prompt toolkit. Some part of it is quite a hack, but it works without changing the current way prompt colors are set in xonsh. So it basically strips the color codes from the prompt string and converts them to pygments tokens. It makes a reversed lookup in the TERM_COLORS dictionary (that is the hack-ish part) 

The good thing about this setup is that it only recomputes the prompt once for every input. That seems to speed things up quite a lot. 

I still haven't found out how to reserve the lower lines the terminal. So the prompt still jumps up/down when the completer list pops up (https://github.com/scopatz/xonsh/issues/362).

### DIFF
--- a/xonsh/prompt_toolkit_shell.py
+++ b/xonsh/prompt_toolkit_shell.py
@@ -6,6 +6,8 @@ from warnings import warn
 from prompt_toolkit.shortcuts import get_input, create_eventloop
 from prompt_toolkit.key_binding.manager import KeyBindingManager
 from pygments.token import Token
+from pygments.style import Style
+from pygments.styles.default import DefaultStyle
 
 from xonsh.base_shell import BaseShell
 from xonsh.pyghooks import XonshLexer
@@ -40,23 +42,6 @@ def teardown_history(history):
         warn('do not have write permissions for ' + hfile, RuntimeWarning)
 
 
-def get_user_input(get_prompt_tokens,
-                   history=None,
-                   lexer=None,
-                   key_bindings_registry=None,
-                   completer=None):
-    """Customized function that mostly mimics promp_toolkit's get_input.
-
-    Main difference between this and prompt_toolkit's get_input() is that it
-    allows to pass get_tokens() function instead of text prompt.
-    """
-    return get_input(
-        lexer=lexer,
-        completer=completer,
-        history=history,
-        get_prompt_tokens=get_prompt_tokens,
-        key_bindings_registry=key_bindings_registry)
-
 
 class PromptToolkitShell(BaseShell):
     """The xonsh shell."""
@@ -78,8 +63,10 @@ class PromptToolkitShell(BaseShell):
             print(intro)
         while not builtins.__xonsh_exit__:
             try:
-                line = get_user_input(
-                    get_prompt_tokens=self._get_prompt_tokens(),
+                token_func, Style_cls = self._get_prompt_tokens_and_style()
+                line = get_input(
+                    get_prompt_tokens=token_func,
+                    style = Style_cls,
                     completer=self.pt_completer,
                     history=self.history,
                     key_bindings_registry=self.key_bindings_manager.registry,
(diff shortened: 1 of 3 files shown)


jonathanslenders:
@melund,

About reserving the lines. I just realise that currently, they are only reserved if a toolbar at the bottom is shown. I will look for a way to reserve them always.

scopatz:
Thanks @melund  and @jonathanslenders. Please ping when you think it is ready for another look.

melund:
@scopatz . It is ready for a review :)

xonsh/prompt_toolkit_shell.py
```diff
@@ -78,8 +63,10 @@ def cmdloop(self, intro=None):
             print(intro)
         while not builtins.__xonsh_exit__:
             try:
-                line = get_user_input(
-                    get_prompt_tokens=self._get_prompt_tokens(),
+                token_func, Style_cls = self._get_prompt_tokens_and_style()
```
scopatz: PEP8: `Style_cls` should be `style_cls`

xonsh/tools.py
```diff
@@ -517,37 +519,54 @@ def __iter__(self):
...
-        if not formatted_prompt:
-            formatted_prompt.append(FakeChar(prompt[pos],
-                                             prefix=match.group(0)))
-            pos += 1
-        else:
-            append_escape_seq(formatted_prompt, match.group(0))
-        match = RE_HIDDEN_MAX.search(prompt, pos)
+    parts = RE_HIDDEN_MAX.split(prompt)
+    #ensure that parts is [colorcode, string, colorcode, string,...]
+    if parts and parts[0] is '':
```
scopatz: This relies on short strings being interned, which might not be generally true. It is probably safer to do a length check. Here is how I would write this line.

```python
if len(parts) > 0 and len(parts[0]) == 0:
```

xonsh/tools.py
```diff
@@ -517,37 +519,54 @@ def __iter__(self):
...
-            pos += 1
-        else:
-            append_escape_seq(formatted_prompt, match.group(0))
-        match = RE_HIDDEN_MAX.search(prompt, pos)
+    parts = RE_HIDDEN_MAX.split(prompt)
+    #ensure that parts is [colorcode, string, colorcode, string,...]
+    if parts and parts[0] is '': 
+        parts = parts[1:]
+    else:
+        parts.insert(0,'')
```
scopatz: PEP8: space after comma

xonsh/tools.py
```diff
@@ -517,37 +519,54 @@ def __iter__(self):
...
+    if len(parts)%2 != 0:
+        parts.append()
+    strings = parts[1::2]
+    token_names = [get_xonsh_color_names(c) for c in parts[::2] ]
```
scopatz: PEP8: weird space before last closing `]`

scopatz:
Thanks @melund! This looks great. Just a couple of minor style issues.

xonsh/tools.py
melund: I see why  `parts[0] is ''` is not a good idea. But isn't `if parts` the pythonic way to test if a list is empty.

scopatz: Yeah, the `if parts` section isn't as big of a deal.  Is this ready for
another look?

melund: Yes. BTW. I added the option to override the prompt toolkit styles by
setting $PROMPT_TOOLKIT_STYLES .... I haven't added anything to the docs,
since I am not sure this the way to do it. But it does allow me to test
different styles.

man. 24. aug. 2015 16.03 skrev Anthony Scopatz <notifications@github.com>:

scopatz:
@melund, you can add info to the docs in the docs folder.  For new environment variables you should probably edit the table (which really needs to be its own page, someday) in this section: https://github.com/scopatz/xonsh/blob/master/docs/tutorial.rst#environment-variables

scopatz:
Looks good. Thanks @melund!

PR #369, opened 2015-08-20
Fix for 122361eeca5189e70ff81931f80812fb16a75c3d: history was partly broken.

jonathanslenders:
Hi, in my last "fix", apparently I broke a part of the history. I just noticed that pressing the up-arrow did not always return the last entry.
This should fix it.


scopatz:
Don't know how I missed this @jonathanslenders.  Sorry about the wait!

PR #370, opened 2015-08-24
Allow environment variables in PROMPT

mdboom:
xonsh rocks.

Coming over from zsh, I found it handy to display environment variables in a PROMPT.  For example, to print out the current pyenv by displaying `$PYENV_VERSION`.

This is a simple hack to allow that.  Alternatively, we could treat only `${FOO}` as environment variables, but that would require writing a simple parser for the `PROMPT` language.

Obviously this needs a test, but I thought I'd get some feedback on the concept first.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -950,9 +950,16 @@ For example:
     2 ~ $
     8 ~ $
 
-If a function in ``$FORMATTER_DICT`` returns ``None``, the ``None`` will be 
+If a function in ``$FORMATTER_DICT`` returns ``None``, the ``None`` will be
 interpreted as an empty string.
 
+Environment variables and functions are also available with the ``$``
+prefix.  For example:
+
+.. code-block:: xonshcon
+
+    snail@home ~ $ $PROMPT = "{$LANG} >"
+    en_US.utf8 >
 
 Executing Commands and Scripts
 ==============================
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -421,13 +421,16 @@ def format_prompt(template=DEFAULT_PROMPT, formatter_dict=None):
         fmtter = formatter_dict
     included_names = set(i[1] for i in _FORMATTER.parse(template))
     fmt = {}
-    for d in (builtins.__xonsh_env__, fmtter):
-        for k, v in d.items():
-            if k not in included_names:
-                continue
-            val = v() if callable(v) else v
-            val = '' if val is None else val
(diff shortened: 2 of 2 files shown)


scopatz:
Hi @mdboom! Thanks! 

This is a useful feature that people have been asking for (I think).  This is a nice solution. A more efficient one would be to loop through included_names and check if the name is in either of the namespaces.  Maybe it isn't such a huge deal because it is such a small amount of data...

A test would be nice, but I am not sure we have any prompt tests yet*.  No better time than the present :). 

In short, I think that this is certainly on the right track.

* We've focused much more on testing the language than the UI. But we should get better about that.

adqm:
Hi @mdbloom.  I think this is a very nice idea; thanks for putting it in!

The only thing I would recommend is to keep the two things separate by denoting environment variables with a `$` (as they are in other places) to avoid the possibility of these two things "shadowing" each other.

So my thought is that ``{foo}`` should be looked up in the normal dictionary, and ``{$foo}`` should look up ``foo`` in the environment.

The good news is that ``_FORMATTER`` already handles parsing this:
```
>>> import string
>>> p = string.Formatter()
>>> p.parse('{$hey}')
<formatteriterator object at 0x6ffffc95090>
>>> list(p.parse('{$hey}'))
[('', '$hey', '', None)]
```
so implementing could be just checking for variable names that start with ``$`` and grabbing those from the environment instead of from ``fmttr``  when building ``fmt``, which would involve a slight restructuring of [this loop](https://github.com/scopatz/xonsh/blob/master/xonsh/environ.py#L435L440).

mdboom:
@adqm: That seems like a good idea.  I had originally thought of doing `${foo}`, but then got hung up because that *isn't* compatible with using `string.Formatter()` as a parser and would require a little mini-parser of our own.  `{$foo}`, as you point out, avoids that complication.

mdboom:
I've updated this based on @adqm's suggestion.

scopatz:
Thanks @mdboom! This looks good to me.  Would you mind adding something to the tutorial docs (probably in the Customizing the Prompt section) about using env vars, please?

scopatz:
ping @mdboom! If you want me to add the docs for you, just let me know!

mdboom:
Sorry -- just got waylaid.  Will do soon...

mdboom:
Ok.  Tutorial snippet posted.

scopatz:
Thanks @mdboom! This is awesome!

PR #371, opened 2015-08-26
Add `FORCE_POSIX_PATH` option to force `/` as path separator on WINDOWS

BYK:


xonsh/completer.py
```diff
@@ -7,12 +7,12 @@
 import subprocess
 import sys
 
+from builtins import __xonsh_env__ as ENV
```
scopatz: You probably shouldn't do this at import because `__xonsh_env__` hasn't been added yet.  It better to set 

```python
env = builtins.__xonsh_env__
```

in the function scope where you need it.

BYK: I converted these to a global because this was a recurring pattern. I'm running off of this version and it seems to be working without errors and as expected.

If you feel strongly tho, I can do it as you suggested (and did before).

A mid-way solution would be to have a global `GET_ENV` method as follows:

```py
GET_ENV = lambda name, default=None: builtins.__xonsh_env__.get(name, default)
```

scopatz: The probably with importing globally like this is that it prevents people from swapping out `__xonsh_env__` wholesale and still having everything work.  The other instances of ENV should probably be removed, too, but that is for another PR :)

`GET_ENV` would be fine, but since it is a function it should probably be `get_env` .

scopatz: Also, note that this is why the build is currently failing.

BYK: >Also, note that this is why the build is currently failing.

lol, okay. Going with the `lambda` approach then.

BYK:
@scopatz updated!

xonsh/completer.py
```diff
@@ -165,12 +170,11 @@ def complete(self, prefix, line, begidx, endidx, ctx=None):
...
     def _add_env(self, paths, prefix):
         if prefix.startswith('$'):
-            env = builtins.__xonsh_env__
-            csc = env.get('CASE_SENSITIVE_COMPLETIONS', True)
+            csc = get_env('CASE_SENSITIVE_COMPLETIONS', True)
             startswither = startswithnorm if csc else startswithlow
             key = prefix[1:]
             keylow = key.lower()
-            paths.update({'$' + k for k in env if startswither(k, key, keylow)})
+            paths.update({'$' + k for k in ENV if startswither(k, key, keylow)})
```
scopatz: Hmmm I don't think that ENV is around anymore...

BYK: Ooops :) Fixed ;)

scopatz:
Oh! And sorry to have missed this before, but can you please add documentation about this new variable to the [table here](https://github.com/BYK/xonsh/blob/f40b4db989f3004bdf23ddcb1c054c8d275e557b/docs/tutorial.rst#environment-variables) in the tutorial. Thanks!

BYK:
Oh, docs. Totally :) On my way...

BYK:
@scopatz updated docs too

scopatz:
Thanks @BYK

PR #372, opened 2015-08-26
Fix bug in attribute completer: unresolved reference `oattr`

BYK:
This was introduced in 1a31c82abdca3ad31ef3299d5b91fc3c88f44d97


scopatz:
Thanks @BYK! I had noticed this was an issue, but hadn't gotten around to fixing it yet!

PR #373, opened 2015-08-26
Expand `FORCE_POSIX_PATHS` behavior to `cwd` in `FORMATTER_DICT`

BYK:
Also covers `cwd_dir` and `cwd_base`.


xonsh/environ.py
```diff
@@ -386,7 +386,12 @@ def _replace_home(x):
     if ON_WINDOWS:
```
BYK: May wanna rename `_replace_home` to `_format_cwd` or something.

scopatz:
Looks good, thanks.  The function is private, so I am not too worried about renaming.

PR #381, opened 2015-09-02
Memoize `locate_binary` to make prompt more responsive

BYK:
Realized every time I pressed `CTRL + C`, `where.exe` was called repeatedly. This patch memoizes `locate_binary` to make it faster by avoiding the subprocess call latency.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1,4 +1,5 @@
 """Environment for the xonsh shell."""
+import functools
 import os
 import re
 import socket
@@ -31,6 +32,7 @@ except AttributeError:
 
 def locale_convert(key):
     """Creates a converter for a locale key."""
+
     def lc_converter(val):
         locale.setlocale(LOCALE_CATS[key], val)
         val = locale.setlocale(LOCALE_CATS[key])
@@ -185,7 +187,7 @@ class Env(MutableMapping):
                 p.break_()
                 p.pretty(dict(self))
 
-
+@functools.lru_cache(maxsize=10)
 def locate_binary(name, cwd):
     # StackOverflow for `where` tip: http://stackoverflow.com/a/304447/90297
     locator = 'where' if ON_WINDOWS else 'which'
@@ -249,7 +251,6 @@ def ensure_hg(func):
 @ensure_git
 def get_git_branch(cwd=None):
     branch = None
-
     if not ON_WINDOWS:
         prompt_scripts = ['/usr/lib/git-core/git-sh-prompt',
                           '/usr/local/etc/bash_completion.d/git-prompt.sh']


BYK:
Tests pass on local :S

scopatz:
Hmm this looks like it might be a prompt-toolkit update issue.  Are you using prompt-toolkit locally?

scopatz:
Also, is there any way that this could be made so that the user can set the maxsize of the memo size? The default seems like it should be zero, too.

BYK:
>Hmm this looks like it might be a prompt-toolkit update issue. Are you using prompt-toolkit locally?

Nope. Couldn't get it to working :(

BYK:
>Also, is there any way that this could be made so that the user can set the maxsize of the memo size? The default seems like it should be zero, too.

It looks like there's no good way to make this configurable since at the time of this module's init, the environment variables are not set. Am I wrong? I thought about patching the module after we initialize the environment variables but that sounds a bit hacky. May be we should have 32 as a sane default and see where this goes?

asmeurer:
It's worth noting that path hashing is a source of a lot of confusion in shells like bash and zsh. Capping the cache at 10 elements would lead to even more confusion, along the lines of "this didn't work, but then I did <something> and it worked," where <something> did nothing more than clear the cache. Also at least in bash you can check for hashed commands by using `type` instead of `which`.

BYK:
@asmeurer - good points but I'm not sure what is the best way forward. I do think we should let people play with the items in the cache and we can expose methods to clear/show cache I guess. What do you think?

scopatz:
Yeah, I think that a way of manipulating the any caching that is done needs to be present. I also think that the default should be the safest, which is for this to be off (or the cache value to be zero)

melund:
I agree with @BYK , on windows it seems that where.exe is called all the time. Wouldn't a time dependent cache not solve the problem. It could cache the results a second or two. That should ensure that where is not called repeatable with the same command.

asmeurer:
Why is it not slow on Unix? Is `which` faster than `where`?  What about implementing the PATH search in Python?

scopatz:
The function in question seems to be called for finding git. There is probably a faster way of doing this...

krader1961:
This pull request should be rejected. As @asmeurer noted on September 7 path resolution hashing does in fact cause a lot of confusion among users of shells which do it (e.g., zsh which I'm switching away from). Zsh attempts to minimize the issue by flushing the cache when $PATH changes but that is only a partial solution to the problem (consider what happens when a command that has already been cached is installed in a directory that appears earlier in $PATH). Performance is important but less important than behavior that is easy for humans to model in their brains.

@BYK if you still want to pursue this please open an issue where pros and cons of various approaches can be discussed.

BYK:
I think this will be useful in some form but not necessarily with this implementation. I'm not pushing forward for this PR yet. I'll look into better and more robust implementations when I have the time.

Although I agree with the modeling part, I don't think performance and understandability are necessarily mutually exclusive.

asmeurer:
As I suggested above, I would look into reimplementing `where` in Python if it is too slow.

BYK:
>As I suggested above, I would look into reimplementing where in Python if it is too slow. 

Gonna go with this path. It's not `where` being slow, it is spawning a process on Windows being slow. Anyways, I'll submit a separate PR.

BYK:
Interested parties can have a look at #650.

PR #400, opened 2015-09-30
Conda build w kernelspec

melund:
This will ensure that conda build can create packages that also install the jupyter kernel spec


scopatz:
Thanks @melund!

PR #404, opened 2015-10-05
Fix bash completions failing when using double quotes

BYK:
When using a command with a `"` in it, `bash_complete` was throwing an
uncaught exception. This patch both catches the exception and escapes
the values passed properly to avoid errors.


scopatz:
Thanks @BYK!

PR #405, opened 2015-10-05
Fix invalid lexer type error

BYK:
Anything stored in `env` is always cast to string so remove
`HIGHLIGHTING_LEXER` variable since it is directly passed to
prompt_toolkit and it is always a string when restoring from `env`
causing `prompt_toolkit` to throw an error because it expects a `Lexer`
instance.


BYK:
I should remove this from test `.json` files but I'm not sure how to modify these lazy JSON files properly.

scopatz:
To remove from the test files, you need to do something like:

```python
from xonsh.lazyjson import LazyJSON, dump

fname = 'f.json'
with LazyJSON(fname) as f:
    data = f.load()
del data['env']['HIGHLIGHTING_LEXER']
with open(fname) as f:
    dump(data, f, sort_keys=True)
```

BYK:
@scopatz - Thanks for the tip, updated!

scopatz:
Hmm, looks like CI is broken now...

BYK:
@scopatz - looks like `dump` does not work properly or as expected. Any ideas? It breaks JSON files. (I used `open(fname, 'w')` btw since I couldn't write otherwise ;))

scopatz:
Looking at this now... Sorry about the delay :(

scopatz:
Ahhh I think that this is a newlines issue! I forgot you are on mac. I'll fix it.

scopatz:
Err on windows

scopatz:
Sorry @BYK, to have made that cross platform, you need to do `open(filename, 'w', newline='\n')`.  My bad.  Thanks for putting this in!

BYK:
Hey there. This looks merged is that the case? I updated the JSON files :)

scopatz:
Hey @BYK - yeah I merged this in but I fixed the JSON files first.  Sorry to beat you to it!

BYK:
Happy to be beaten :P

Issue #204, opened 2015-04-12
bash aliases extraction is unreliable


auscompgeek:
The `xonsh.aliases.bash_aliases()` function mostly assumes only the output of the `alias` command will be written. However, many `bashrc`s, including distros' `/etc/bash.bashrc`s set `$PROMPT_COMMAND` when bash is run in interactive mode, which can cause unexpected output.

auscompgeek:
This begins to raise the question of whether this is worth the trouble that it causes...

scopatz:
Thanks for opening this issues @auscompgeek, I think the choice of whether it is worth it or not should be left up the users.  So really, there should be a way to toggle whether this feature is on or off....

PR #410, opened 2015-10-10
aliases: make bash_aliases() more robust

auscompgeek:
* Call bash with `-c alias` rather than passing `alias` through stdin, to avoid `$PROMPT_COMMAND` giving us garbage.
* Check each alias read in is really an alias, to make doubly-sure we don't get garbage in our aliases.

We should have a way of disabling the bash alias extraction altogether though. After all, not everyone uses bash as their main shell, and the idea of a shell calling another shell to automatically pull in aliases scares me.

Fixes: #204

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -1,10 +1,8 @@
 """Aliases for the xonsh shell."""
 import os
-import sys
 import shlex
 import builtins
 import subprocess
-import datetime
 from warnings import warn
 from argparse import ArgumentParser
 
@@ -75,6 +73,7 @@ def xexec(args, stdin=None):
 
 _BANG_N_PARSER = None
 
+
 def bang_n(args, stdin=None):
     """Re-runs the nth command as specified in the argument."""
     global _BANG_N_PARSER
@@ -109,7 +108,8 @@ def bash_aliases():
                                     universal_newlines=True)
     except (subprocess.CalledProcessError, FileNotFoundError):
         s = ''
-    items = [line.split('=', 1) for line in s.splitlines() if line.startswith('alias ') and '=' in line]
+    items = [line.split('=', 1) for line in s.splitlines()
+             if line.startswith('alias ') and '=' in line]
     aliases = {}
     for key, value in items:
         try:
@@ -185,5 +185,3 @@ elif ON_MAC:
 else:
     DEFAULT_ALIASES['grep'] = ['grep', '--color=auto']
     DEFAULT_ALIASES['ls'] = ['ls', '--color=auto', '-v']
-
-
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -104,7 +104,7 @@ def bang_bang(args, stdin=None):
 def bash_aliases():
     """Computes a dictionary of aliases based on Bash's aliases."""
     try:
-        s = subprocess.check_output(['bash', '-i', '-l', '-c', 'alias'],
+        s = subprocess.check_output(['bash', '-i', '-c', 'alias'],
                                     stderr=subprocess.PIPE,
                                     universal_newlines=True)
     except (subprocess.CalledProcessError, FileNotFoundError):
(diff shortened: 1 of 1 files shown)


auscompgeek:
For comparison, here's a before/after on my Arch Linux install, with absolutely no modifications to any `bashrc`s:

```python
seedbox@kairi ~/code/xonsh master $ import xonsh
seedbox@kairi ~/code/xonsh master $ xonsh.aliases.bash_aliases()
{'edbox@kairi:~/code/xonsh\x07\x1b]7;file://kairi/home/seedbox/code/xonsh\x07\x1b]0;seedbox@kairi:~/code/xonsh\x07alias ls': ['ls',
  '--color=auto']}
```

```python
seedbox@kairi ~/code/xonsh fix-bash-aliases $ import xonsh
seedbox@kairi ~/code/xonsh fix-bash-aliases $ xonsh.aliases.bash_aliases()
{'ls': ['ls', '--color=auto']}
```

asmeurer:
+1. This appears to fix the issue I was seeing at https://github.com/scopatz/xonsh/issues/403#issuecomment-145723516 (although not the performance). The same should probably be done for `env`. 

And +1 for being able to disable this. I do use bash, but having xonsh read bash aliases defeats the purpose of putting them in my .xonshrc.

auscompgeek:
> I do use bash, but having xonsh read bash aliases defeats the purpose of putting them in my .xonshrc.

Agreed. An idea I was pondering: we could take a leaf out of vim and enable/disable some things when a .xonshrc exists.

> (although not the performance)

Hold on, I think I can help with the performance aspect slightly (but not much). I've moved the `.startswith` check into the list comp.

auscompgeek:
You know, I've always questioned why this calls bash as a login shell. It'd only slow things down, since any sane user would put aliases in `.bashrc` (or a file sourced from it), not `.bash_profile` or `.profile`. Less code that needs to be executed, the better.

xonsh/aliases.py
```diff
@@ -104,13 +104,12 @@ def bang_bang(args, stdin=None):
...
     try:
-        s = subprocess.check_output(['bash', '-i', '-l'],
-                                    input='alias',
+        s = subprocess.check_output(['bash', '-i', '-c', 'alias'],
                                     stderr=subprocess.PIPE,
                                     universal_newlines=True)
     except (subprocess.CalledProcessError, FileNotFoundError):
         s = ''
-    items = [line.split('=', 1) for line in s.splitlines() if '=' in line]
+    items = [line.split('=', 1) for line in s.splitlines() if line.startswith('alias ') and '=' in line]
```
scopatz: PEP8: line over 80 char

scopatz:
So the login shell was added in #319 for the problem explained in #313. Basically, the problem is that is you source another file in your bashrc that then goes and sets aliases, if you have a non-login shell it won't work.  Which sort of sucks, because the login shells take so much longer to start up.  The `-l` probably needs to go back in :(

scopatz:
I think maybe we should make the login part optional. Also, we should make the running bash or sh at all, optional.

Thanks for your work on this @auscompgeek.  I want to  restructure this stuff for sure.

auscompgeek:
@scopatz Actually, the issue presented in #313 is that the user sourced their ~/.aliases from ~/.bash_profile, when it should be sourced from ~/.bashrc. Personally, I think that spawning a login shell because some users don't know how to use bash is incorrect behaviour.

xonsh/aliases.py
auscompgeek: To be fair, this project would probably cause flake8 to light everything up like a Christmas tree, if I hadn't ignored some things. :stuck_out_tongue_winking_eye:

auscompgeek:
Speaking of linter warnings, shouldn't we have a linter check on Travis or something?

scopatz:
> Edit: Oh, I see we have landscape.io set up here. Seems to be down right now though…

Yeah, it seems to go down relatively frequently

asmeurer:
On OS X the terminal defaults to login shell, so most OS X users use .bash_profile or .profile (myself included).

Issue #403, opened 2015-10-05
xonsh hangs on startup


asmeurer:
This is starting xonsh 0.2.0 from bash. When I keyboard interrupt I get

```
Traceback (most recent call last):
  File "/Users/aaronmeurer/anaconda/bin/xonsh", line 6, in <module>
    sys.exit(main())
  File "/Users/aaronmeurer/anaconda3/lib/python3.4/site-packages/xonsh/main.py", line 71, in main
    shell = builtins.__xonsh_shell__ = Shell(**shell_kwargs)
  File "/Users/aaronmeurer/anaconda3/lib/python3.4/site-packages/xonsh/shell.py", line 27, in __init__
    self._init_environ(ctx)
  File "/Users/aaronmeurer/anaconda3/lib/python3.4/site-packages/xonsh/shell.py", line 56, in _init_environ
    self.execer = Execer()
  File "/Users/aaronmeurer/anaconda3/lib/python3.4/site-packages/xonsh/execer.py", line 41, in __init__
    load_builtins(execer=self)
  File "/Users/aaronmeurer/anaconda3/lib/python3.4/site-packages/xonsh/built_ins.py", line 669, in load_builtins
    builtins.aliases.update(bash_aliases())
  File "/Users/aaronmeurer/anaconda3/lib/python3.4/site-packages/xonsh/aliases.py", line 126, in bash_aliases
    value = shlex.split(value)
  File "/Users/aaronmeurer/anaconda3/lib/python3.4/shlex.py", line 276, in split
    return list(lex)
  File "/Users/aaronmeurer/anaconda3/lib/python3.4/shlex.py", line 266, in __next__
    token = self.get_token()
  File "/Users/aaronmeurer/anaconda3/lib/python3.4/shlex.py", line 93, in get_token
    raw = self.read_token()
  File "/Users/aaronmeurer/anaconda3/lib/python3.4/shlex.py", line 224, in read_token
    self.token = self.token + nextchar
KeyboardInterrupt
```

How do I debug this?

asmeurer:
I left it running and it eventually started with `/Users/aaronmeurer/anaconda3/lib/python3.4/site-packages/xonsh/aliases.py:129: RuntimeWarning: could not parse Bash alias "emacsclient": ValueError('No closing quotation',)
  RuntimeWarning)`. My emacsclient bash alias is 

```
$command -v emacsclient
[...]

PR #412, opened 2015-10-11
Foreign Shells

scopatz:
This PR adds a generalized capability for pulling in information (environment and aliases) from foreign shells, such as Bash.  As a side effect, this also allows you to turn off loading in any data from foreign shells at all.  

This depends on #410 and #411 and was built to mitigate #403.

If @asmeurer and/or @auscompgeek could try this out on their machines and let me know whether it works for them, that'd be great!  I'll merge this in, if I get the OK from either of these people.

Also, note that this disables using login shells by default, though it can easily be reinstated.


scopatz:
Here are some timings on one of my better machines:

```console
# master branch
scopatz@artemis ~ $ time -f "%E real\t%U user\t%S sys" xonsh -c exit
0:00.26 real	0.23 user	0.00 sys

# foreignsh branch, loading bash
scopatz@artemis ~ $ time -f "%E real\t%U user\t%S sys" xonsh -c exit
0:00.18 real	0.16 user	0.00 sys

# foreignsh branch, loading nothing
scopatz@artemis ~ $ time -f "%E real\t%U user\t%S sys" xonsh -c exit
0:00.15 real	0.14 user	0.00 sys

# bash
scopatz@artemis ~ $ time -f "%E real\t%U user\t%S sys" bash -c exit
0:00.00 real	0.00 user	0.00 sys

# python
scopatz@artemis ~ $ time -f "%E real\t%U user\t%S sys" python -c "exit()"
0:00.05 real	0.03 user	0.01 sys
```

xonsh/environ.py
```diff
@@ -73,7 +168,7 @@ class Env(MutableMapping):
       locale via locale.getlocale() and locale.setlocale() functions.
 
     An Env instance may be converted to an untyped version suitable for
-    use in a subprocess.
+    use in a subprocess.
```
auscompgeek: Someone needs to set up their editor to strip trailing whitespace :wink:

xonsh/shell.py
```diff
@@ -26,25 +26,26 @@ class Shell(object):
     def __init__(self, ctx=None, shell_type=None, **kwargs):
         self._init_environ(ctx)
         env = builtins.__xonsh_env__
-
+        # piclk a valid shell
```
auscompgeek: typo

xonsh/foreign_shells.py
```diff
@@ -0,0 +1,243 @@
...
+        currenv = builtins.__xonsh_env__.detype()
+    elif currenv is not None:
+        currenv = dict(currenv)
+    try:
+        s = subprocess.check_output(cmd,stderr=subprocess.PIPE, env=currenv,
+                                    universal_newlines=True)
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        if not safe:
+            raise
+        s = FAILED_COMMAND_STDOUT
```
auscompgeek: Could we just `return {}, {}` here instead?

docs/api/index.rst
```diff
@@ -51,5 +51,6 @@ For those of you who want the gritty details.
     lazyjson
     teepty
     openpy
+    foriegn_shells
```
auscompgeek: typo

xonsh/foreign_shells.py
```diff
@@ -0,0 +1,243 @@
...
+        Path to the static config file. Not compatible with shell not being None.
+        If both shell and config is None, then it will be read from the 
+        $XONSHCONFIG environment variable.
+    issue_warning : bool, optional
+        Issues warnings if config file cannot be found.
+
+    Returns
+    -------
+    env : dict
+        A dictionary of the merged environments.
```
auscompgeek: copy-paste fail?

auscompgeek:
Will `source_bash()` be rewritten to use this?

xonsh/environ.py
scopatz: Yeah, I haven't added that feature to xo yet, soon probably :)

xonsh/foreign_shells.py
scopatz: yep.

scopatz:
Thanks for the review @auscompgeek! I have addressed everything that you brought up in-line.

> Will source_bash() be rewritten to use this?

It probably should be, but I don't want this PR to get too big. Maybe you could open an issue for that?

scopatz:
Also, does this actually work for you?

xonsh/foreign_shells.py
```diff
@@ -0,0 +1,236 @@
...
+        Path to the static config file. Not compatible with shell not being None.
+        If both shell and config is None, then it will be read from the 
+        $XONSHCONFIG environment variable.
+    issue_warning : bool, optional
+        Issues warnings if config file cannot be found.
+
+    Returns
+    -------
+    aliases : dict
+        A dictionary of the merged alaiases.
```
auscompgeek: Maybe you should invest in an editor that does spellcheck...

auscompgeek:
Seems to work well for me on my Arch Linux install. Although, do we have to run both `env` and `alias` when we only want one of the two?

xonsh/foreign_shells.py
scopatz: Again, a feature I have considered adding to xo. PRs welcome.

scopatz:
> Seems to work well for me on my Arch Linux install. Although, do we have to run both env and alias when we only want one of the two?

I can't think of a case where only one is desired. Maybe I am missing something though...  I'd almost certainly accept a PR that added getting one or the other as an optional feature. (Say you just want env, you could set aliascmd to false.) However, getting both simultaneously is fairly advantageous. 

As a historical note, originally env and aliases were obtained together.  Then someone put in a PR to split getting them up, but it nearly doubled the startup time for xonsh. Enough people have complained that startup was too slow to motivate this PR.

Issue #417, opened 2015-10-16
Autocompletion of a class method not always working


davidbrochart:
Say I have this class:
```
class Test:
    def testTest(self):
        print('test')

test = Test()
```
Now when I enter `test.` and press Tab, I have a proposition for testTest, which is correct. But if I enter `test.testT` and press Tab, no proposition at all. I think the issue is with the capital letter.

scopatz:
Thanks for reporting @davidbrochart. What version of xonsh are you on?

scopatz:
I can confirm that this happens on v0.2.2

scopatz:
As a temporary workaround, it seems to work fine with `$CASE_SENSITIVE_COMPLETIONS = False`

davidbrochart:
Thanks for the workaround.

scopatz:
Potential fix in #418

davidbrochart:
I confirm #418 fixes the bug. Thanks a lot!

scopatz:
Great! Thanks!  it has merged into master,

PR #418, opened 2015-10-16
fixed attr case sensitive completion

scopatz:


scopatz:
This should close #417. @davidbrochart would you mind testing this out and letting me know if it works.

PR #428, opened 2015-10-20
Add literal tab insert to prompt tk keybinding

### LINKED ISSUE #408: Can't type literal tab with prompt_toolkit
In bash or zsh I would typically type `Ctl-V TAB` to insert a literal tab, however, when using prompt_tookitas my shell type this simply tries to do tab completion and there doesn't seem to be a way to create a literal tab.

gforsyth:
This adds a key handle to insert a literal tab character when the user hits `Shift+Tab`. 

I just copied the readline keybinding, but it can be changed if folks want a different choice.

Resolves #408

### DIFF
--- a/xonsh/prompt_toolkit_key_bindings.py
+++ b/xonsh/prompt_toolkit_key_bindings.py
@@ -34,9 +34,9 @@ def load_xonsh_bindings(key_bindings_manager):
         event.cli.current_buffer.insert_text(env.get('INDENT'))
 
     @handle(Keys.BackTab)
-    def _(insert_literal_tab):
+    def insert_literal_tab(event):
         """
         Insert literal tab on Shift+Tab instead of autocompleting
         """
-        insert_literal_tab.cli.current_buffer.insert_text(env.get('INDENT'))
+        event.cli.current_buffer.insert_text(env.get('INDENT'))
 
--- a/xonsh/prompt_toolkit_key_bindings.py
+++ b/xonsh/prompt_toolkit_key_bindings.py
@@ -32,3 +32,11 @@ def load_xonsh_bindings(key_bindings_manager):
         indent instead of autocompleting.
         """
         event.cli.current_buffer.insert_text(env.get('INDENT'))
+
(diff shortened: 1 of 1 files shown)


xonsh/prompt_toolkit_key_bindings.py
```diff
@@ -32,3 +32,11 @@ def _(event):
         indent instead of autocompleting.
         """
         event.cli.current_buffer.insert_text(env.get('INDENT'))
+
+    @handle(Keys.BackTab)
+    def _(event):
```
scopatz: Thanks @gforsyth - can you please give this function a more meaningful name, even if it is private? Thanks!

scopatz:
Thanks for putting this in!

gforsyth:
Happy to help!

xonsh/prompt_toolkit_key_bindings.py
```diff
@@ -32,3 +32,11 @@ def _(event):
         indent instead of autocompleting.
         """
         event.cli.current_buffer.insert_text(env.get('INDENT'))
+
+    @handle(Keys.BackTab)
+    def _(insert_literal_tab):
```
scopatz: Hi @gforsyth - I think I might not have been clear enough before. `event` is fine.  It is the method name (`_`) that should be expanded to something more meaningful. Thanks!

gforsyth:
My mistake -- that's fixed up now.  Looking at it now, though, since both of the key bindings end up inserting a literal tab, should I just combine them?  A la:

```Python
    @handle(Keys.Tab, filter=TabShouldInsertIndentFilter())
    @handle(Keys.BackTab)
    def insert_literal_tab(event):
        """
        If 

        i) there are only whitespaces before current cursor position or
        ii) user enters Shift+Tab

        insert literal tab on Shift+Tab instead of autocompleting

        """
        event.cli.current_buffer.insert_text(env.get('INDENT'))
```

or does that seem less readable?

scopatz:
I like it the way it is now rather than the way that you haver proposed. Thanks again!

PR #432, opened 2015-10-21
Updated source-bash to handle multiline values

blarghmatey:
When trying to execute source-bash on a file that contains a value which
has newlines as part of it only the first line is captured. This is due
to using splitlines() naively. In order to overcome this limitation I
added a regex to parse the values.

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -1,5 +1,6 @@
 """Aliases for the xonsh shell."""
 import os
+import re
 import shlex
 import builtins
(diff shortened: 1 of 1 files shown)


scopatz:
Did you mean to close this?

blarghmatey:
Yes, I did some more testing and realized that it doesn't quite work properly. I'm going to dig into this further before I open a new pull request.

scopatz:
Ok thanks!

Issue #277, opened 2015-06-25
Using xonsh with conda environments on windows


melund:
Hi all

Maybe some Anaconda/Windows users will find this useful. 

I had some trouble getting xonsh to play nice with conda environments on Windows. 

On windows, the environments are controlled by two bat-file scripts that add or remove the environments from the path variable and change the prompt. I mange to get the prompt working but calling activate/deactivate from within xonsh didn't seem to change the $PATH variable in xonsh. 

So I ended fixing it by writing Python versions of the deactivate/activate scripts. Just add them to the .xonshrc file. 

``` python 

import os
import builtins

def _deactivate(args, stdin=None):
    # Cal anaconda.bat script to ensure that 'ANACONDA_ENVS' exist as an 
    # environment variable
    $(anaconda.bat)
    if args:
        print("Usage: deactivate\n"
              "\n"
              "Deactivates previously activated Conda\n"
              "environment")
        return
    if 'CONDA_DEFAULT_ENV' not in builtins.__xonsh_env__:
        return
    print("Deactivating environment "+  $CONDA_DEFAULT_ENV +"...")
    $PATH = list(filter( os.path.join($ANACONDA_ENVS,$CONDA_DEFAULT_ENV).__ne__, $PATH))
    $PATH = list(filter( os.path.join($ANACONDA_ENVS,$CONDA_DEFAULT_ENV,"Scripts").__ne__, $PATH))
    del $CONDA_DEFAULT_ENV
    del $CONDA_ENV_PATH

def _activate (args, stdin=None):
    # Cal anaconda.bat script to ensure that 'ANACONDA_ENVS' exist as an 
    # environment variable
    $(anaconda.bat)
    if len(args) != 1:
        print("Usage: activate envname\n"
              "\n"
              "Deactivates previously activated Conda\n"
              "environment, then activates the chosen one.")
        return
    new_environment = args[0]
    if not os.path.isdir(os.path.join( $ANACONDA_ENVS, new_environment)):
        print ("No environment named "
               + new_environment + 
               " exists in " + $ANACONDA_ENVS)
        return
    if 'CONDA_DEFAULT_ENV' in builtins.__xonsh_env__:
       _deactivate([], stdin)
    $CONDA_DEFAULT_ENV = new_environment
    $CONDA_ENV_PATH = os.path.join($ANACONDA_ENVS, $CONDA_DEFAULT_ENV)
    print("Activating environment " + $CONDA_DEFAULT_ENV + "...") 
[...]

PR #433, opened 2015-10-21
Add support for system-wide xonshrc

gforsyth:
As discussed in #277 

Linux and OSX will look for a system-wide config file in  `/etc/xonshrc` in addition to the user's config file.  

Windows system config file is stored in `%ALLUSERSPROFILE%\Application Data\xonsh\xonshrc`

I don't have a windows machine to try this out on, but this is the same way that pip handles global configuration files, so it should work(?).  If a windows user ( @melund ?) can try this out to make sure it's grabbing both configuration files, that would be great.  

The way it's set up now, user `.xonshrc` will override system `xonshrc` if there are conflicting `$env` names.

Also, apologies.  Would've rebased but I wasn't sure if it would re-notify the linked issues and commits in one of the commit messages and opted for less noise.

### DIFF
--- a/docs/envvars.rst
+++ b/docs/envvars.rst
@@ -133,9 +133,8 @@ applicable.
       - The location of the static xonsh configuration file, if it exists. This is
         in JSON format.
     * - XONSHRC
-      - ``~/.xonshrc``
-	``/etc/xonshrc`` (Linux and OSX) 
-	``%ALLUSERSPROFILE%\Application Data\xonsh\xonshrc`` (Windows)
+      - ``('/etc/xonshrc', '~/.xonshrc')`` (Linux and OSX) 
+    	``('%ALLUSERSPROFILE%\xonsh\xonshrc', '~/.xonshrc')`` (Windows)
       - A tuple of the locations of run control files, if they exist.  User defined
 	run control file will supercede values set in system-wide control file if there
 	is a naming collision.
--- a/docs/envvars.rst
+++ b/docs/envvars.rst
@@ -134,7 +134,11 @@ applicable.
         in JSON format.
     * - XONSHRC
       - ``~/.xonshrc``
-      - Location of run control file.
+	``/etc/xonshrc`` (Linux and OSX) 
+	``%ALLUSERSPROFILE%\Application Data\xonsh\xonshrc`` (Windows)
+      - A tuple of the locations of run control files, if they exist.  User defined
+	run control file will supercede values set in system-wide control file if there
+	is a naming collision.
     * - XONSH_CONFIG_DIR
       - ``$XDG_CONFIG_HOME/xonsh``
       - This is location where xonsh configuration information is stored.
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -621,23 +621,22 @@ def xonshrc_context(rcfiles=None, execer=None):
        or sum([os.path.isfile(rcfile) for rcfile in rcfiles]) == 0):
         return {}
     for rcfile in rcfiles:
+        if not os.path.isfile(rcfile):
+            continue
+        with open(rcfile, 'r') as f:
+            rc = f.read()
+        if not rc.endswith('\n'):
+            rc += '\n'
+        fname = execer.filename
+        env = {}
         try:
-            with open(rcfile, 'r') as f:
-                rc = f.read()
-            if not rc.endswith('\n'):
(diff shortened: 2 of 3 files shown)


xonsh/environ.py
```diff
@@ -142,7 +142,10 @@ def xonshconfig(env):
     'XDG_CONFIG_HOME': os.path.expanduser(os.path.join('~', '.config')),
     'XDG_DATA_HOME': os.path.expanduser(os.path.join('~', '.local', 'share')),
     'XONSHCONFIG': xonshconfig,
-    'XONSHRC': os.path.expanduser('~/.xonshrc'),
+    'XONSHRC': ((os.path.join(os.environ['ALLUSERSPROFILE'],
+                              'Application Data', 'xonsh', 'xonshrc'),
+                os.path.expanduser('~/.xonshrc')) if ON_WINDOWS
+               else ('/etc/xonshrc', os.path.expanduser('~/.xonshrc'))),
```
scopatz: It seems like `$XONSHRC` also needs an entry in the `DEFAULT_ENSURERS` dict that looks like:

```python
'XONSHRC': (is_env_path, str_to_env_path, env_path_to_str),
```

This will make this variable be able to be passed into / out of the process environment.

xonsh/environ.py
```diff
@@ -611,24 +614,29 @@ def load_static_config(ctx):
...
+            env = {}
+            try:
+                execer.filename = rcfile
+                execer.exec(rc, glbs=env)
+            except SyntaxError as err:
+                msg = 'syntax error in xonsh run control file {0!r}: {1!s}'
+                warn(msg.format(rcfile, err), RuntimeWarning)
+            finally:
+                execer.filename = fname
+        except:
```
scopatz: Why is this outer try-except block here?  It seems like this loop should check `if not os.path.isfile(rcfile): continue` at the top. Also, there shouldn't be bare exceptions, flake8 complains :)

scopatz:
Thanks a ton for the PR @gforsyth. I have recommended a couple of small changes inline.  Also, please update the relevant part of `docs/envvars.rst` since the type is changing.

gforsyth:
Will do.

xonsh/environ.py
```diff
@@ -611,24 +615,28 @@ def load_static_config(ctx):
...
-        execer.filename = fname
+    for rcfile in rcfiles:
+        if not os.path.isfile(rcfile):
+            continue
+        with open(rcfile, 'r') as f:
+            rc = f.read()
+        if not rc.endswith('\n'):
+            rc += '\n'
+        fname = execer.filename
+        env = {}
```
scopatz: `env` probably shouldn't be redefined each loop iteration.

melund:
@gforsyth  
I can confirm it works on Win10. 

"%ALLUSERSPROFILE%/Application Data" is a hidden hard-link to %ALLUSERSPROFILE%. I think it is a compartibility thing with windows xp.  

Maybe you should just set the path to: 
os.path.join(os.environ['ALLUSERSPROFILE'], 'xonsh', 'xonshrc'))

without the "Application Data" stuff. It will make easier for people to find the xonshrc file. But I don't know how it will play wih XP.

I do get a lot of errors running tests on this branch, but I don't think it is your fault.

/Morten

scopatz:
> I do get a lot of errors running tests on this branch, but I don't think it is your fault.

@melund can you open a new issue with these errors, please? Thanks!

gforsyth:
@melund 
I'll make the change as you suggest -- it should be easier for people to play with their configuration files.  And if Microsoft doesn't support XP, I don't think we need to, either.

docs/envvars.rst
```diff
@@ -134,7 +134,11 @@ applicable.
         in JSON format.
     * - XONSHRC
       - ``~/.xonshrc``
-      - Location of run control file.
+	``/etc/xonshrc`` (Linux and OSX) 
+	``%ALLUSERSPROFILE%\Application Data\xonsh\xonshrc`` (Windows)
```
scopatz: This default seems incorrect. This should be a list/tuple of both the system and user files. Right now they are out of order and not a list, just next to each other.

gforsyth:
Fixed the docs -- also updated the path to the windows system-wide file as suggested by @melund.

scopatz:
Thanks a ton @gforsyth!

Issue #380, opened 2015-09-02
Obscure error with prompt_toolkit on entering git commands


mdraw:
I am using the current git version of xonsh 201484a with prompt_toolkit on Arch Linux.
On entering `git clone '` or `git clone #` (not executing, just entering those characters), xonsh crashes with the following error:

    Exception in thread Thread-184:
    Traceback (most recent call last):      
      File "/usr/lib/python3.4/threading.py", line 920, in _bootstrap_inner
        self.run()                          
      File "/usr/lib/python3.4/threading.py", line 868, in run
        self._target(*self._args, **self._kwargs)
      File "/usr/lib/python3.4/site-packages/prompt_toolkit/interface.py", line 510, in run
        CompleteEvent(text_inserted=True))) 
      File "/usr/lib/python3.4/site-packages/xonsh/prompt_toolkit_completer.py", line 30, in get_completions
    self.ctx)
      File "/usr/lib/python3.4/site-packages/xonsh/completer.py", line 134, in complete
    for s in self.bash_complete(prefix, line, begidx, endidx):
      File "/usr/lib/python3.4/site-packages/xonsh/completer.py", line 265, in bash_complete
    stderr=subprocess.PIPE)
      File "/usr/lib/python3.4/subprocess.py", line 620, in check_output
    raise CalledProcessError(retcode, process.args, output=output)
    subprocess.CalledProcessError: Command '['bash']' returned non-zero exit status 1

I don't know if this is a bug in xonsh or in prompt_toolkit and don't really understand what is happening there, so I am just reporting it here, hoping that someone can explain/fix this.
Strangely, I could only reproduce this bug with git subcommands, while other commands worked fine with these characters. So I guess it could also be a specific problem with git autocompletion.

scopatz:
Thanks for reporting @mdraw. Do you hit TAB after entering the single quote or hash symbol?

scopatz:
This is almost certainly a git completion thing.  Ahhh, also, if you are on prompt-toolkit, pt may be running completion for you automatically.  This probably needs to be more robust.

mdraw:
I did not hit TAB after entering the mentioned commands. The errors occur immediately when typing the single quotes and hashes.
Another detail: I also noticed that the error does not occur if you write something before the `#` sign, like e.g. `git clone git://anongit.kde.org/yakuake.git#branch=frameworks` works fine.
Can you reproduce this?

scopatz:
[...]

Issue #364, opened 2015-08-19
tab-completion launches executables


melund:
I observed this error on windows. If I press tab to autocomplete an executable, then xonsh lauches it. 

[xonsh]~\Anaconda3\Scripts
$ spyder.e \<pres tab\>

That starts the exe file.

adqm:
I have noticed a similar issue on GNU/Linux as well.  For me, it was with the following:

`git commit -m "always import local markdown" _markdown<tab>`

This called `import local` (subprocess mode, so taking a screenshot) and/or `import _markdown` (also subprocess mode), so I ended up with screenshots.

Not sure when this started happening, but I'm pretty sure it wasn't always this way.

scopatz:
This seems bad :(

Carreau:
+1, just experienced with `python3.<tab>`

scopatz:
I think that this is happening because the '.' is getting tab-completed as attribute access, which triggers an evaluation.

scopatz:
@melund et al., Thanks to @adqm can we now close this issue?

denfromufa:
for me tab completion does not work on windows at all. it works like a tab instead.

scopatz:
That sounds like a bug. Can you provide more information, @denfromufa about your setup.

denfromufa:
@scopatz python 2.7.9 in root conda, then conda env. for python 3.4 (xonsh fails on pip, python 2.7, python 3.5). activated python 3.4 - launched xonsh, no luck with tab, although other features work.

```
$ pip list
pip (7.1.2)
ply (3.8)
prompt-toolkit (0.54)
Pygments (2.0.2)
setuptools (18.5)
six (1.10.0)
wcwidth (0.1.5)
wheel (0.26.0)
xonsh (0.2.3)


$ conda list
# packages in environment at C:\Python\Python27\envs\py34:
#
msvc_runtime              1.0.1                    vc10_0  [vc10]
pip                       7.1.2                    py34_0
ply                       3.8                       <pip>
prompt-toolkit            0.54                      <pip>
pygments                  2.0.2                     <pip>
python                    3.4.3                         5
setuptools                18.5                     py34_0
six                       1.10.0                    <pip>
wcwidth                   0.1.5                     <pip>
wheel                     0.26.0                   py34_1
xonsh                     0.2.3                     <pip>
```

scopatz:
Hmm let's wait until we get a new release out in the next couple of days.

denfromufa:
[...]

PR #436, opened 2015-10-25
Make prompt_toolkit's auto_suggest optional

mdraw:
Introducing the new AUTO_SUGGEST environment variable for this,
which can either be True or False. Default is True.

In addition, make mouse_support a class variable of PromptToolkitShell
for consistency's sake.

### DIFF
--- a/xonsh/prompt_toolkit_shell.py
+++ b/xonsh/prompt_toolkit_shell.py
@@ -46,6 +46,11 @@ class PromptToolkitShell(BaseShell):
         super().__init__(**kwargs)
         self.history = setup_history()
         self.pt_completer = PromptToolkitCompleter(self.completer, self.ctx)
+        self.mouse_support = builtins.__xonsh_env__.get('MOUSE_SUPPORT')
+        if builtins.__xonsh_env__.get('AUTO_SUGGEST') is True:
+            self.auto_suggest = AutoSuggestFromHistory()
+        else:
+            self.auto_suggest = None
         self.key_bindings_manager = KeyBindingManager(
             enable_auto_suggest_bindings=True,
             enable_search=True, enable_abort_and_exit_bindings=True)
@@ -59,13 +64,12 @@ class PromptToolkitShell(BaseShell):
         """Enters a loop that reads and execute input from user."""
         if intro:
             print(intro)
-        mouse_support = builtins.__xonsh_env__.get('MOUSE_SUPPORT')
         while not builtins.__xonsh_exit__:
             try:
                 token_func, style_cls = self._get_prompt_tokens_and_style()
                 line = get_input(
-                    mouse_support=mouse_support,
-                    auto_suggest=AutoSuggestFromHistory(),
+                    mouse_support=self.mouse_support,
+                    auto_suggest=self.auto_suggest,
                     get_prompt_tokens=token_func,
                     style=style_cls,
                     completer=self.pt_completer,
--- a/docs/envvars.rst
+++ b/docs/envvars.rst
@@ -19,6 +19,11 @@ applicable.
     * - AUTO_PUSHD
       - ``False``
       - Flag for automatically pushing directorties onto the directory stack.
+    * - AUTO_SUGGEST
+      - ``True``
+      - Enable automatic command suggestions based on history (like in fish shell).
+        Pressing the right arrow key inserts the currently displayed suggestion.
+        (Only usable with SHELL_TYPE=prompt_toolkit)
     * - BASH_COMPLETIONS
       - Normally this is ``('/etc/bash_completion', '/usr/share/bash-completion/completions/git')``
         but on Mac is ``'/usr/local/etc/bash_completion', '/opt/local/etc/profile.d/bash_completion.sh')``.
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -63,6 +63,7 @@ DEFAULT_ENSURERS = {
     'BASH_COMPLETIONS': (is_env_path, str_to_env_path, env_path_to_str),
     'TEEPTY_PIPE_DELAY': (is_float, float, str),
     'MOUSE_SUPPORT': (is_bool, to_bool, bool_to_str),
+    'AUTO_SUGGEST': (is_bool, to_bool, bool_to_str)
 }
 
 #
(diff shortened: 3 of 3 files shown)


scopatz:
Thanks for this @mdraw!

xonsh/prompt_toolkit_shell.py
```diff
@@ -46,6 +46,11 @@ def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self.history = setup_history()
         self.pt_completer = PromptToolkitCompleter(self.completer, self.ctx)
+        self.mouse_support = builtins.__xonsh_env__.get('MOUSE_SUPPORT')
+        if builtins.__xonsh_env__.get('AUTO_SUGGEST') is True:
```
scopatz: This should just be `if builtins.__xonsh_env__.get('AUTO_SUGGEST'):`.  The `is True`, while probably safe here because of how `__xonsh_env__` works, should not generally be used.

xonsh/prompt_toolkit_shell.py
```diff
@@ -46,6 +46,11 @@ def __init__(self, **kwargs):
...
+            self.auto_suggest = AutoSuggestFromHistory()
+        else:
+            self.auto_suggest = None
```
scopatz: Also, I think that these lines need to be moved back down to the command loop so that they can be set in the xonshrc file and dynamically turned on/off.  The shell init happens too early.

scopatz:
Couple small issues to address.

mdraw:
Thanks for the comments.

I deliberately chose to make an explicit comparison to True because it makes it more clear that the value has to be True and cannot be another expression that just *evaluates* to True, like e.g. `[1]`.
But I understand that this may be a bit strange and "un-PEP8-y", so I will remove the `is True`.

Concerning the second comment: I didn't realize making those variables class attributes would make it less flexible. I will move them to the command loop like you suggested.
While we are at it, a little stylistic question: Is it ok to use the same names for the parameter value and the parameter name, like `foo(mouse_support=mouse_support)`?

scopatz:
> While we are at it, a little stylistic question: Is it ok to use the same names for the parameter value and the parameter name, like foo(mouse_support=mouse_support)?

Yup totally OK and preferred.

scopatz:
Thanks @mdraw

mdraw:
Should I move the those environment variable checks into the actual loop (`while not builtins.__xonsh_exit__:`...) inside cmdloop() so they are run in every loop and can always be changed at runtime? (This would probably make xonsh a little bit slower, but I doubt performance is *that* important here).
I am asking because initially I hadn't placed the mouse_support check in the while loop, but directly in cmdloop(), where they are apparently only checked once during the shell startup.

scopatz:
Hmmm Starting up `AUTO_SUGGEST` might be pricey, actually.  I don't know enough about prompt-toolkit to know what the workflow *should* be.  Probably best to leave it where it was put originally by @jonathanslenders et al.  Also, it looks like this branch has conflicts with master.  Would you mind updating and resolving. Thanks!

mdraw:
I believe I can also initialize another variable with AutoSuggest() earlier and only assign its value to the auto_suggest variable during the loop if the environment variable tells it to, so the AutoSuggest object would remain static, the preference dynamic, and the only repeating performance hit in the REPL would be a dictionary lookup and variable assignment (i.e. practically zero).

I will resolve the conflict when I make the next commit.

scopatz:
Sounds good to me.

mdraw:
This is the first time I resolved a merge conflict, so please tell me if something looks off in my last commit. The diff looks larger due to the conflict resolution, but the only actual changes from me are in prompt_toolkit_shell.py.

scopatz:
Looks good to me! Thanks @mdraw

mdraw:
No problem. Thank you very much @scopatz and @jonathanslenders and all the other contributors for making this possible!

I will soon create a similar PR for prompt_toolkit's completion options so you will be able to choose between multi-column completions (which are currently hardcoded), single-column completions and completely deactivating completions to make the prompt less intrusive.
This will require another environment variable that only concerns prompt_toolkit. Since there are now a few variables that only have an effect on the optional prompt_toolkit shell, I am now wondering if would make sense to somehow group them together so it gets obvious that they have no effect on the standard readline shell. Maybe a common prefix (e.g. AUTO_SUGGEST --> PTK_AUTO_SUGGEST and MOUSE_SUPPORT --> PTK_MOUSE_SUPPORT) or do you have other suggestions for grouping them? Hypothetically, if prompt_toolkit would become mandatory and the readline shell type got removed, we would not have this problem of incompatible options, but I don't think you plan on doing this in the long term, @scopatz, or do you?

P.S. xonsh gradually gets closer to becoming my default shell. The only remaining reasons why it isn't already are those: https://github.com/scopatz/xonsh/pull/253, https://github.com/scopatz/xonsh/issues/364 / https://github.com/scopatz/xonsh/issues/380 (and the option to deactivate prompt_toolkit's completions, which I will contribute soon like I mentioned above).

scopatz:
> Maybe a common prefix (e.g. AUTO_SUGGEST --> PTK_AUTO_SUGGEST and MOUSE_SUPPORT --> PTK_MOUSE_SUPPORT) or do you have other suggestions for grouping them? 

meh. I don't think the `PTK_` prefix is needed since only one shell type can be active at a time.  I think it is OK for such non-overlapping environment variables to potentially take on different meanings for different shell types, though in these cases I don't see how they would have different meanings.

> Hypothetically, if prompt_toolkit would become mandatory and the readline shell type got removed, we would not have this problem of incompatible options, but I don't think you plan on doing this in the long term, @scopatz, or do you?

Part of my original design goal was to make xonsh work with the standard library + ply only, and potentially even distribute our own copy of ply eventually.  If we get too dependency heavy, then folks will start complaining about how it isn't as portable as Bash. We have already seen some of this and Python is basically everywhere. Lot's of optional dependencies are great though :)

> xonsh gradually gets closer to becoming my default shell.

Awesome! We'll work towards that!

> The only remaining reasons why it isn't already are those: #253 

#253 seems like a deal breaker for some people. This isn't something I personally would use so much, but if it is preventing people from using xonsh than it jumps up the priority queue.

>  #364 / #380 (and the option to deactivate prompt_toolkit's completions, which I will contribute soon like I mentioned above).

Thanks! I really need someone else to tackle this :)

mdraw:
Alright, thanks for your input.

Issue #421, opened 2015-10-16
Problems using sudo with autocomplete


gforsyth:
I'm stumped.  `sudo` is not playing nice with xonsh.  

I'm using Arch Linux and xonsh v0.2.2 with prompt_toolkit.

Completions work just fine unless I start the line with sudo and then I have no completion support.  I've also noticed that while `!!` will run the previous command, as expected, `sudo !!` returns 

```Bash
sudo: !!: command not found
```

`sudo env` shows that the sudo `SHELL` is set to `xonsh`

*EDIT:
to clarify, if I type `sudo` and then hit TAB it will open up the prompt_toolkit completion list, but shows only the files and folders in that directory.  If I start typing the name of a binary (say, `pacman`) there is no completion.

scopatz:
Thanks for reporting @gforsyth. There are a few different issues here.

The first is that xonsh only does alias expansion on the first token. Since sudo is the first token, it is not expanding !!, which is an alias and not part of xonsh syntax. A simple sudo alias function that does alias expansion would fix this and probably should be in xonsh itself. Pull requests very welcome here.

The second issue is less clear where it is coming from. But xonsh relies on Bash completions. I don't know where sudo bash completions are defined, but it is a fair bet that that location is not present in the `$BASH_COMPLETIONS` variable. That is probably the right place to start looking.

drufat:
I am facing a similar issue. Some autocompletions that work under bash, do not work under xonsh. For example trying the following at the bash command prompt under archlinux with the auto-complete package installed

    $ which gre[TAB]
    grep       gresource 

results in helpful completion tips. Doing the same in xonsh, results in no output. I am not sure what I need to do in xonsh to get the same behavior as in bash.

scopatz:
Hi @drufat, can you check that the files listed in `$BASH_COMPLETIONS` actually exist. If not, we may need to do something different for arch.

gforsyth:
@drufat @scopatz 

arch does use a different setup for completions -- they live in 
`/usr/share/bash-completion/bash_completion` but the attribute completions directory is the same, that is, `/usr/share/bash-completion/completions/git` is valid. 

I also found a bug in the prompt_toolkit_completer that prevented attributes from tab completing for, say, git.  Still trying to iron out a few kinks so that it works as expected, but I'll PR that soon.

scopatz:
[...]

PR #437, opened 2015-10-26
add ON_ARCH env to adjust $BASH_COMPLETIONS per os

gforsyth:
Arch places the default bash_completion scriptlet in
`/usr/share/bash-completion`, not in `/etc/bash_completion` as in most *nix
systems.

This adds an extra flag to tools.py to check if the current OS is
Arch, `ON_ARCH`, (according to `platform.linux_distribution`) and if it is, adjusts
the default location for bash_completion accordingly.

This fixes one of the issues in #421 -- the sudo aliasing isn't handled here, though

### DIFF
--- a/docs/envvars.rst
+++ b/docs/envvars.rst
@@ -21,7 +21,9 @@ applicable.
       - Flag for automatically pushing directorties onto the directory stack.
     * - BASH_COMPLETIONS
       - Normally this is ``('/etc/bash_completion', '/usr/share/bash-completion/completions/git')``
-        but on Mac is ``'/usr/local/etc/bash_completion', '/opt/local/etc/profile.d/bash_completion.sh')``.
+        but on Mac is ``('/usr/local/etc/bash_completion', '/opt/local/etc/profile.d/bash_completion.sh')``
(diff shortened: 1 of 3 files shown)


scopatz:
Thanks @gforsyth - I think you might also need to change the value in the `BASE_ENV` dict too.

gforsyth:
I thought that the line

```Python
'BASH_COMPLETIONS': list(DEFAULT_VALUES['BASH_COMPLETIONS']),
```

would take care of that?  Since `BASH_COMPLETIONS` is already defined depending on `IS_ARCH`, `IS_LINUX`, etc... before `BASE_ENV` is defined.

or I could very well be missing something.

scopatz:
Whoops! You are right. Sorry.

scopatz:
Thanks a ton @gforsyth!

auscompgeek:
Wow. Can't we just check whether `/etc/bash_completion` or `/usr/share/bash_completion` exists instead of performing this unreliable check?

scopatz:
I think that it will skip files that are in this list but not present on the file system.  Feel free to put in a PR improving this in any way. Thanks.

Issue #443, opened 2015-10-29
xonsh.bat script ignoring parameters in windows


educhana:
xonsh.bat files dont pass parameters down

this causes that:

xonsh -c script

ignores the execution of requested script.

should be very simple to fix xonsh.bat simply adding %*
```
@echo off
python -m xonsh.main %*
```

scopatz:
See #444

scopatz:
Thanks for reporting @educhana!

scopatz:
This has been fixed for a while.

PR #444, opened 2015-10-29
Updated the windows script for command line args

scopatz:
As reported in #443. It would be great if @educhana or some other windows user could test this out. Thanks!


auscompgeek:
Looks good. Is there a reason why we aren't just using the normal Python entry points though?

scopatz:
Don't those entry point rely on setuptools and/or conda?

PR #446, opened 2015-10-29
have source_alias search on $PATH if not in cwd

gforsyth:
Similar to bash and zsh behavior, this will allow the `source` command
to be used for files not in the current working directory that are
contained in $PATH.


scopatz:
Cool feature to have. Thanks a million @gforsyth

Issue #447, opened 2015-10-30
$() expects only unicode output


ZyX-I:
Consider the following command:

    echo $(printf '\\x99')

. This will yield

    UnicodeDecodeError: 'utf-8' codec can't decode byte 0x99 in position 0: invalid start byte

The same thing happens when I try to use `$(ls -l)` in my home directory (I have a number of directories with “strange” names for testing my scripts).

$LANG is ru_RU.UTF-8.

Xonsh version 0.2.2.

scopatz:
Hmm these seems bad.  Where are you setting $LANG at? I have found that it needs to be set at the system level.  xonsh relies on Python's `sys.getdefaultenconding()` implicitly - or it should - in all cases. Setting lang at the system level seems to be the only way to do this. I have also found that I need to set `$LC_MESSAGES` equal to the same value as $LANG (and in the same place, `/etc/default/locale` for me.).

ZyX-I:
@scopatz I am running xonsh from zsh. $LANG is set in /etc/env.d/99locale (functionally) or /etc/profile.env (file generated from /etc/env.d and actually sourced by shells), but xonsh is supposed to inherit it from zsh.

ZyX-I:
@scopatz In any case xonsh should not limit itself to only valid UTF-8, regardless of locale: I have seen zip archives which produce weird sequences in file names after unpacking, and unzip with USE=natspec is not always fixing this.

In xonsh `locale` has some strange output:

```Shell
LANG=ru_RU.UTF-8
LC_CTYPE=ru_RU.UTF-8
LC_NUMERIC=C
LC_TIME=C
LC_COLLATE=C
LC_MONETARY=C
LC_MESSAGES=C
LC_PAPER="ru_RU.UTF-8"
LC_NAME="ru_RU.UTF-8"
LC_ADDRESS="ru_RU.UTF-8"
LC_TELEPHONE="ru_RU.UTF-8"
LC_MEASUREMENT="ru_RU.UTF-8"
LC_IDENTIFICATION="ru_RU.UTF-8"
LC_ALL=
```

. In zsh from where I launch xonsh it is

```Shell
LANG=ru_RU.UTF-8
LC_CTYPE="ru_RU.UTF-8"
LC_NUMERIC="ru_RU.UTF-8"
LC_TIME="ru_RU.UTF-8"
LC_COLLATE="ru_RU.UTF-8"
LC_MONETARY="ru_RU.UTF-8"
[...]

PR #450, opened 2015-10-30
locale fixes to rely on the system better

scopatz:
This should address the issues in #447. I would really appreciate it if @ZyX-I or someone else could test this out. Thanks!

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -26,10 +26,8 @@ LOCALE_CATS = {
     'LC_MONETARY': locale.LC_MONETARY,
     'LC_TIME': locale.LC_TIME,
 }
-try:
+if hasattr(locale, 'LC_MESSAGES'):
     LOCALE_CATS['LC_MESSAGES'] = locale.LC_MESSAGES
-except AttributeError:
-    pass
 
 
 def locale_convert(key):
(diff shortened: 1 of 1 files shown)


ZyX-I:
This fixes `env | grep LC` and `locale` output: now first shows only `READNULLCMD` as expected and second shows `ru_RU.UTF-8` everywhere except `LC_ALL`. But this is not fixing UnicodeDecodeError for `echo $(printf '\\x99')`.

scopatz:
OK, I think I see what the deconding issue is.  Basically, a lot of Python builtin functionality such as `str.decode()` assumes UTF8. I thought that this assumed the encoding for the system. I am going to go through and add the capability to set the encoding and error handling in the environment. This will be in another PR.

ZyX-I:
@scopatz Note that my locale encoding *is* UTF-8. I do not need setting *other* encodings, I need it accept *non-UTF-8 input*.

One way to do this is using `surrogateescape` `errors` handling method to make it accept all kinds of input:

    b'\xFF'.decode('utf-8', 'surrogateescape').encode('utf-8', 'surrogateescape') == b'\xFF'

scopatz:
@ZyX-I, see #454. This is basically what I did.

PR #453, opened 2015-10-31
Fixed wrong completion function called if a command is a bash command

semyon-san:
If user ran a command that could be completed by bash completer (e.g. ls, cd) the path completition handling will be run by the bash completer as well, when it should be done by "path_complete".

To fix this I added a check that runs "bash_complete" only if the argument of the command starts with a "-".

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -130,7 +130,7 @@ class Completer(object):
             # the first thing we're typing; could be python or subprocess, so
             # anything goes.
             rtn = self.cmd_complete(prefix)
-        elif cmd in self.bash_complete_funcs and arg.startswith('-'):
+        elif not arg.startswith(('.', os.sep)) and cmd in self.bash_complete_funcs:
             rtn = set()
             for s in self.bash_complete(prefix, line, begidx, endidx):
                 if os.path.isdir(s.rstrip()):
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks for putting in this PR. Unfortunately, I can't merge this as-is because it breaks too much else.  For example `git clo` will not autocomplete to `git clone` under this PR. Bash completion is for all arguments, not just options.  There might be a way to sanely turn off this behavior in certain cases. But in general, I favor returning too many options, rather than too few.

semyon-san:
Hmm, I see. If this is not changed there will be a hard dependency on bash for auto-completing paths, which looks like a hack tbh.

I have an idea: maybe use "path_complete()" if the command argument starts with a dot or an os.sep, otherwise use "bash_complete()"?

scopatz:
HI @semyon-san, can you explain to me the problem that the current situation is causing? Bash completion sometimes completes filenames, depending on the command and the argument. Selectively disabling this seems odd. 

We don't rely on Bash to complete paths, but if it happens to there it is not a problem.

If you just want to disable all bash completions, set the `$BASH_COMPLETIONS` variable to an empty list. You might need to do this in the xonshrc file.

gforsyth:
Hey @semyon-san, thanks for putting this patch in.  I believe at this point that the completion semantics have changed enough that it would be WAY more work to rebase this and get things going than to just work with what's now in place.  
That said, I think that the path completion is much cleaner now and hopefully meets needs but please do take a look at what is now in place and let us know about any deficiencies you come across (or submit a patch!).  

Thanks again!

PR #454, opened 2015-10-31
Encoding fixes for suproc

scopatz:
This should address the remaining issues brought up in #447.  Can you see if this works for you @ZyX-I? Thanks a ton.

### DIFF
--- a/docs/envvars.rst
+++ b/docs/envvars.rst
@@ -151,6 +151,15 @@ applicable.
     * - XONSH_DATA_DIR
       - ``$XDG_DATA_HOME/xonsh``
       - This is the location where xonsh data files are stored, such as history.
+    * - XONSH_ENCODING
+      - ``sys.getdefaultencoding()``
+      - This is the that xonsh should use for subrpocess operations.
+    * - XONSH_ENCODING_ERRORS
+      - ``'surrogateescape'``
+      - The flag for how to handle encoding errors should they happen.
+        Any string flag that has been previously registered with Python
+        is allowed. See the `Python codecs documentation <https://docs.python.org/3/library/codecs.html#error-handlers>`_
+        for more information and available options. 
     * - XONSH_HISTORY_FILE
       - ``'~/.xonsh_history'``
       - Location of history file (deprecated).
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -159,8 +159,8 @@ def test_is_float():
     yield assert_false, is_float('42.0')
 
 def test_is_string():
-    yield assert_true, is_float('42.0')
-    yield assert_false, is_float(42.0)
+    yield assert_true, is_string('42.0')
+    yield assert_false, is_string(42.0)
 
 def test_always_true():
     yield assert_true, always_true(42)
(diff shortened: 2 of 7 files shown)


scopatz:
Also, you may need to set `$XONSH_ENCODING` or `$XONSH_ENCODING_ERRORS`

ZyX-I:
This works for me (including with XONSH_STORE_STDOUT), except that I am not sure why you need to replace `\r\n` with `\n` (currently missing a use-case when line endings needs to be preserved though, but any kind of a change which prevents me from doing `$FOO=$(cat /bin/cat); echo -n $FOO > cat2; chmod +x cat2` looks suspicious even if I do not really do such things). It does not *break* anything though: I have checked that prior to this branch CRLF line endings are also replaced with LF. Not considering a bug without a use-case.

// By the way, just noticed that XONSH_INTERACTIVE is a boolean variable which I did not set and which is set to string `True` in `env` output. But if I set XONSH_STORE_STDOUT to `True` it will be set to string `1` in `env` output and not to `True`. (Same behaviour if I *do* set `$XONSH_INTERACTIVE`, including with `False` which is just empty string with XONSH_STORE_STDOUT and literal `False` with XONSH_INTERACTIVE.) Is this expected or should be reported as a bug?

scopatz:
>  I am not sure why you need to replace \r\n with \n 

This is a Windows compatibility thing, and is not a bug.

>  just noticed that XONSH_INTERACTIVE is a boolean variable which I did not set and which is set to string True in env output.

This is also expected. The xonsh environment has typed values, like you noticed.  However, suprocess mode requires that the environment values be untyped / just strings.  To manage this the xonsh environment has a standard way for handling the types conversion to/from strings for known environment variable names and name patterns.  The `env` command is a subprocess so you'll see the untyped version.  If you want to see the typed environment, evaluate `__xonsh_env__` which is a builtin variable. You can read more about this in the tutorial.

scopatz:
Also, very glad this works for you!

ZyX-I:
> This is also expected. The xonsh environment has typed values, like you noticed. However, suprocess mode requires that the environment values be untyped / just strings. To manage this the xonsh environment has a standard way for handling the types conversion to/from strings for known environment variable names and name patterns. The env command is a subprocess so you'll see the untyped version. If you want to see the typed environment, evaluate __xonsh_env__ which is a builtin variable. You can read more about this in the tutorial.

The problem is not `True` in `env` output. The problem is that

1. XONSH_INTERACTIVE is a boolean variable which is saved as `"True"`/`"False"`.
2. XONSH_STORE_STDOUT is *also* a boolean variable, but it is saved as `"1"`/`""`.

. Looks like an inconsistency.

scopatz:
> Looks like an inconsistency.

It is an inconsistency, but it is not xonsh's fault.  Many other programs such as Bash and others will treat an environment variable as false iff it is not defined or length-0.  Even the string "0" or "false" are intreted as being true-valued.  This is super annoying and frustrating. So xonsh implements the translation to be as robust as possible.  Arguably, you shouldn't need to worry about it that often and xonsh handles it for you anyway.

PR #459, opened 2015-11-01
Foreign Shell Function Aliases

### LINKED ISSUE #420: Sourcing virtualenvwrapper has no effect.
I use ROS and virtualenvwrapper, both of which need to be sourced from a sh file.

Here are the relevant parts of my `.xonshrc`

    # ROS
    source-bash /opt/ros/jade/setup.bash
    source-bash ~/catkin_ws/devel/setup.bash

    # Python Virtualenv
    $WORKON_HOME = "~/.virtualenvs"
    source-bash /usr/local/bin/virtualenvwrapper.sh
    $PIP_VIRTUALENV_BASE = "~/.virtualenvs"

Sourcing ROS allows me to use `roscore`, for instance. However, what is expected for virtualenvwrapper is that, upon opening a new shell, I would be able to use commands such as `workon` or `mkvirtualenv`. Instead, I get `xonsh: subprocess mode: command not found` for all those commands.

scopatz:
This adds the ability to describe functions that are written in foreign shells (such as Bash) and then automatically wrap them as xonsh aliases.  This should happen when you load a foreign shell at startup or when you source a foreign shell file.  

This should fix #420 and builds off of #458.  I'd really appreciate it if someone - and particularly @masasin could try this out. Thanks!

### DIFF
--- a/docs/aliases.rst
+++ b/docs/aliases.rst
@@ -62,8 +62,13 @@ only works on xonsh and Python files.
 
 ``source-bash``
 ====================
-Like the ``source`` command but for Bash files. It implements Bash's source builtin.
+Like the ``source`` command but for Bash files. This is a thin wrapper around
+the ``source-foreign`` alias.
 
+``source-foreign``
+====================
+Like the ``source`` command but for files in foreign (non-xonsh) languages. 
+It will pick up the environment and any aliases.
 
 ``history``
 ====================
--- a/docs/xonshconfig.rst
+++ b/docs/xonshconfig.rst
@@ -28,10 +28,10 @@ variables. For example,
 --------------------
 This is a list (JSON Array) of dicts (JSON objects) that represent the
 foreign shells to inspect for extra start up information, such as environment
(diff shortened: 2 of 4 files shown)


masasin:
I installed the `srcforeign` branch, but it does not seem to have worked. I also tried it by sourcing straight from the shell. 

I did get the `source-foreign` command, so I also used `source-foreign bash /path/to/virtualenvwrapper.sh`, and it did not work either.

scopatz:
Can you try the fsffi branch please, the srcforeign doesn't have what you need. Thanks

masasin:
It works! Thank you!

On Mon, Nov 2, 2015, 16:08 Anthony Scopatz <notifications@github.com> wrote:

> Can you try the fsffi branch please, the srcforeign doesn't have what you
> need. Thanks
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/scopatz/xonsh/pull/459#issuecomment-152938862>.
>
-- 

勝者神優

scopatz:
Woohoo!

PR #461, opened 2015-11-02
Make prompt_toolkit Python completions configurable

mdraw:
Completions provided by prompt_toolkit are now configurable
by setting $COMPLETIONS_DISPLAY.
Changing it at runtime will take immediate effect, so you can quickly
disable and enable completions during shell sessions.

Possible values for $COMPLETIONS_DISPLAY are:
- 'none' for no completions,
- 'single' for single-column completions,
- 'multi' for multi-column completions (default).

### DIFF
--- a/docs/envvars.rst
+++ b/docs/envvars.rst
@@ -39,6 +39,18 @@ applicable.
       - ``[]``
       - A list of paths to be used as roots for a ``cd``, breaking compatibility with 
         bash, xonsh always prefer an existing relative path.
+    * - COMPLETIONS_DISPLAY
+      - ``'multi'``
+      - Configure if and how Python completions are displayed by the prompt_toolkit shell.
+        This option does not affect bash completions, auto-suggestions etc.
+        Changing it at runtime will take immediate effect, so you can quickly
+        disable and enable completions during shell sessions.
+        - If COMPLETIONS_DISPLAY is ``'none'`` or ``'false'``, do not display those completions.
+        - If COMPLETIONS_DISPLAY is ``'single'``, display completions in a single column while typing.
+        - If COMPLETIONS_DISPLAY is ``'multi'`` or ``'true'``, display completions in multiple columns while typing.
+        These option values are not case- or type-sensitive, so e.g.
+        writing ``$COMPLETIONS_DISPLAY = None`` and ``$COMPLETIONS_DISPLAY = 'none'`` is equivalent.
+        (Only usable with SHELL_TYPE=prompt_toolkit)
     * - DIRSTACK_SIZE
       - ``20``
       - Maximum size of the directory stack.
--- a/xonsh/prompt_toolkit_shell.py
+++ b/xonsh/prompt_toolkit_shell.py
@@ -68,15 +68,18 @@ class PromptToolkitShell(BaseShell):
                     auto_suggest = _auto_suggest
                 else:
                     auto_suggest = None
+                completions_display = builtins.__xonsh_env__.get('COMPLETIONS_DISPLAY')
+                multicolumn = True if completions_display == 'multi' else False
+                completer = None if completions_display == 'none' else self.pt_completer
                 line = get_input(
                     mouse_support=mouse_support,
                     auto_suggest=auto_suggest,
                     get_prompt_tokens=token_func,
                     style=style_cls,
-                    completer=self.pt_completer,
+                    completer=completer,
                     history=self.history,
                     key_bindings_registry=self.key_bindings_manager.registry,
-                    display_completions_in_columns=True)
+                    display_completions_in_columns=multicolumn)
                 if not line:
                     self.emptyline()
                 else:
--- a/xonsh/prompt_toolkit_shell.py
+++ b/xonsh/prompt_toolkit_shell.py
@@ -69,7 +69,7 @@ class PromptToolkitShell(BaseShell):
                 else:
                     auto_suggest = None
                 completions_display = builtins.__xonsh_env__.get('COMPLETIONS_DISPLAY')
-                multicolumn = True if completions_display == 'multi' else False
+                multicolumn = (completions_display == 'multi')
                 completer = None if completions_display == 'none' else self.pt_completer
                 line = get_input(
                     mouse_support=mouse_support,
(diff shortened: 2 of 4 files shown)


xonsh/tools.py
```diff
@@ -555,6 +555,21 @@ def ensure_int_or_slice(x):
...
+
+
+def to_completions_display_value(x):
+    if str(x).lower() in {'none', 'false'}:
+        return 'none'
+    if str(x).lower() in {'multi', 'true'}:
+        return 'multi'
+    if str(x).lower() in {'single'}:
+        return 'single'
+    print('Warning: \'{}\' is not a valid value for $COMPLETIONS_DISPLAY. Using \'multi\'.'.format(str(x)))
```
scopatz: Please use a the Python standard library `warn()` function here.  Also you can wrap in double quotes so that you don't have to escape the internal single quotes.  Also this line is too long and you don't need to call `str()` on x, format does that for you.

docs/envvars.rst
```diff
@@ -55,10 +76,11 @@ applicable.
...
       - ``False``
-      - Enable mouse support in the prompt_toolkit shell. This allows clicking
-        for positioning the cursor or selecting a completion. In some terminals
-        however, this disables the ability to scroll back through the history
-        of the terminal.
+      - Enable mouse support in the prompt_toolkit shell.
+        
+        This allows clicking for positioning the cursor or selecting a completion. In some terminals
+        however, this disables the ability to scroll back through the history of the terminal.
+
```
scopatz: I don't think these blank newlines are needed.

xonsh/prompt_toolkit_shell.py
```diff
@@ -68,15 +68,18 @@ def cmdloop(self, intro=None):
                     auto_suggest = _auto_suggest
                 else:
                     auto_suggest = None
+                completions_display = builtins.__xonsh_env__.get('COMPLETIONS_DISPLAY')
+                multicolumn = True if completions_display == 'multi' else False
```
scopatz: This could just be:

```python
multicolumn = (completions_display == 'multi')
```

scopatz:
Thanks for putting this in @mdraw. I think it would be best to allow `True`, `False`, and `None` as their Python values - not as strings here. Does this make sense? This will require a little refactoring from what is here now.

docs/envvars.rst
mdraw: At least on the Github viewer, without those new blank newlines the bullet points and line breaks are completely ignored and everything is shoved into one single, soft-wrapped line. I thought it is much more readable with the newlines.

xonsh/prompt_toolkit_shell.py
mdraw: Good point!

mdraw:
> I think it would be best to allow True, False, and None as their Python values - not as strings here.

They are allowed as their Python values. At least when I was testing it, all of those values were correctly interpreted and automatically translated to the expected lowercase strings by xonsh.tools.to_completions_display_value(), for example: `$COMPLETIONS_DISPLAY = None` is intepreted in the same way as `$COMPLETIONS_DISPLAY = 'none'`
Or do you mean something different with your comment?

xonsh/tools.py
mdraw: Alright, I'll change it accordingly. Thanks for the hint!

xonsh/tools.py
```diff
@@ -555,6 +555,21 @@ def ensure_int_or_slice(x):
...
+def is_completions_display_value(x):
+    return x in {'none', 'single', 'multi'}
+
+
+def to_completions_display_value(x):
+    if str(x).lower() in {'none', 'false'}:
+        return 'none'
+    if str(x).lower() in {'multi', 'true'}:
+        return 'multi'
+    if str(x).lower() in {'single'}:
```
scopatz: Should only call `str(x).lower()` once at the top of the function. Also, this should probably be an if-elif-else statement with a single return statement at the end.

scopatz:
Ahhh OK. I had missed that you were calling `str(x)` in the function. Sorry.

scopatz:
Let me know when this is ready for another look! Thanks again @mdraw

xonsh/tools.py
mdraw: > Should only call str(x).lower() once at the top of the function.

Do you mean this as in `x = str(x).lower() ; if x in {...`?

> Also, this should probably be an if-elif-else statement with a single return statement at the end.

Why would this be better than my current version?

scopatz: > Do you mean this as in x = str(x).lower() ; if x in {...?

yes.

> Why would this be better than my current version?

I find it easier to read and follow. Just a stylistic thing. Not a deal breaker for this PR. Just something I would appreciate.

mdraw: I am still unsure if I understand how you would write the conditions. Do you mean something like this?:

```python
from warnings import warn


def to_completions_display_value(x):
    x = str(x).lower()
    if x in {'none', 'false'}:
        x = 'none'
    elif x in {'multi', 'true'}:
        x = 'multi'
    elif x in {'single'}:
        x = 'single'
    else:
        warn('"{}" is not a valid value for $COMPLETIONS_DISPLAY. '.format(x) +
             'Using "multi".')
        x = 'multi'
    return x
```

(I personally find the multiple returns more clear and concise than this, but I will use the code style that you prefer of course.)

I have a problem with `warnings.warn()`, though:
```python
to_completions_display_value('Invalid value')
```
outputs
```
test.py:15: UserWarning: "invalid value" is not a valid value for $COMPLETIONS_DISPLAY. Using "multi".
  'Using "multi".')
```
`warn()` seems to print a line of the original code there, and I don't know how to turn off this behavior without ugly hacks. Did I do anything wrong there?

scopatz: > Did I do anything wrong there?

Nope that is the expected behaviour. The benefits of warnings is that users can modify how they want to see them globally through Python itself.

scopatz: You might want to change this to a RuntimeWarning though. Otherwise this function looks good. Let me know when you have updated it!

mdraw:
Ok, thanks, for the comments, @scopatz. I have just pushed your suggested changes. Could you please review it again?

scopatz:
Cool beans, thanks!

Issue #439, opened 2015-10-26
Entering a directory name should cd to that directory


kseistrup:
Amiga's `WShell` had a sweet future that I haven't seen implemented in any of the shells I've encountered on POSIX systems: entering a directory name would cd to that directory.

E.g.,
```sh
kseistrup@home ~ $ /usr/bin
kseistrup@home /usr/bin $ 
```
It would be a nice feature to have in `xonsh`.

scopatz:
Cool idea! Pull requests welcome, especially if it is optional.

mdraw:
This behavior is built into the fish shell.
The function in fish that checks if an entered string can be interpreted as an implicit cd can be found here: https://github.com/fish-shell/fish-shell/blob/master/src/path.cpp#L225
Maybe this will help implementing it in xonsh.

scopatz:
Thanks @mdraw, this looks like it checks specific directory prefixes and suffixes. It would probably be easier  to check if `len(args) == 1 and os.path.isdir(args[0]) and not on_path(args[0])`.  The last part is psuedo-code for now, but I think we already have a functions somewhere that does that.

mdraw:
Yes,

    len(args) == 1 and os.path.isdir(os.path.expanduser args[0]))
should do the trick for any possible path if I am not missing something.
Fish's implementation is probably only more complex because it doesn't have access to the awesome Python standard library :)

mdraw:
What do you mean by `on_path(args[0])`, @scopatz?

auscompgeek:
I'll also note that this is also builtin toggleable behaviour in zsh (the autocd option).

kseistrup:
It seems I have been very superficial when trying out other shells than `bash`… :wink:

scopatz:
> What do you mean by on_path(args[0]), @scopatz?

Whether there is a program in `args[0]` that lives in one of the directories of the `$PATH` variable.

gforsyth:
@kseistrup this should be working now (see #463).  You can set `$AUTO_CD=True` in your `xonshrc` and it should work

auscompgeek:
Can we close this issue now?

kseistrup:
I have problems building the `xonsh-git` package on Arch because of Jupyter requirements (which is already installed, so I don't quite understand why it attempts to install it again…).  I'll look into it as soon as I've had some breakfast.

kseistrup:
Just for the record:

[...]

PR #463, opened 2015-11-03
add `AUTO_CD` functionality

gforsyth:
As suggested in #439 

I couldn't find a better place to insert this other than `base_shell.py` -- it might be a bit too hacky, but I'm open to suggestions.

If user enters the name of a directory, e.g.

```
gil@theo ~ $ /usr/bin
```

then Xonsh will first try to run that command.  When it fails and raises
a XonshError, if the env variable `$AUTO_CD=True` then Xonsh will check
to see if the input is a valid directory and that there is no binary in
`$PATH` that matches the input (this last bit is likely redundant).

If all conditions are met, it reruns the "command" but prepends `cd` to it.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -21,7 +21,7 @@ from collections import Sequence, MutableMapping, Iterable, namedtuple, \
 from xonsh.tools import suggest_commands, XonshError, ON_POSIX, ON_WINDOWS, \
     string_types
 from xonsh.inspectors import Inspector
-from xonsh.environ import Env, default_env
+from xonsh.environ import Env, default_env, locate_binary
 from xonsh.aliases import DEFAULT_ALIASES
 from xonsh.jobs import add_job, wait_for_active_job
 from xonsh.proc import ProcProxy, SimpleProcProxy, TeePTYProc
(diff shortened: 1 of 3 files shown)


scopatz:
Hi @gforsyth - thanks a million for putting this in!  I would actually put this in the `xonsh.built_ins.run_subproc()` functions, probably somewhere around [line #525](https://github.com/scopatz/xonsh/blob/master/xonsh/built_ins.py#L525). If autocd is true, if there are no other arguments, if the command is a directory and cannot be found in the aliases or via `locate_binary()`, then replace the command with `cd dir` alias. I think that this is less hacky than sticking it in BaseShell.

gforsyth:
Hey @scopatz , happy to help!  I was sure there must be a less hacky way but couldn't figure it out.  Now it's in `run_subproc()` and working well on my machine.

xonsh/built_ins.py
```diff
@@ -523,6 +523,13 @@ def run_subproc(cmds, captured=True):
             stderr = streams['stderr']
         uninew = (ix == last_cmd) and (not captured)
         alias = builtins.aliases.get(cmd[0], None)
+        if (alias is None
+            and builtins.__xonsh_env__.get('AUTO_CD')
+            and os.path.isdir(cmd[0])
+            and locate_binary(cmd[0], cwd=None) is None
+            and len(cmds)==1):
+            cmd.insert(0, 'cd')
+            return run_subproc(cmds, captured=False)
```
scopatz: Does this need to be recursive? I think you can just set `alias = builtins.aliases.get('cd', None)` here instead of returning.

scopatz:
Is this ready to go @gforsyth?

gforsyth:
Should be good.  Tests passing, works as expected.

scopatz:
Thanks a ton @gforsyth - awesome feature to have. We might consider changing the default to True :)

Issue #457, opened 2015-10-31
doesn't install properly on Windows


lewisl:
Instead of getting xonsh.bat in the scripts directory, you get xonsh--which is appropriate for 'nix.

I manually created a suitable script.

When you run python -m xonsh.main, xonsh comes up, but with characters that are unprintable in the windows console.  You can then change the $PROMPT environment variable and get something clean.  From the documentation it is not clear where the .xonshrc file should be placed as the documentation addresses 'nix style systems.

Guess this isn't targeted to Windows.

lewisl:
.xonshrc in the user's root works as it does on 'nix.  Could improve the windows doc page.  Seems out of date.

scopatz:
Thanks for reporting @lewisl. I know that there are folks out there who use xonsh on windows daily.  I will happily take any PRs that make the user experience better.  Sorry that you are running in to so many issues....

scopatz:
It seems that the install issues have been addressed by #464.  If that was the main purpose of this issue, please close this @lewisl. Thanks!

krader1961:
PR #468 should address this issue in an even more robust solution than #464 did; at least if you use Anaconda for your installs (which I highly recommend).

scopatz:
closing due to age.

PR #464, opened 2015-11-03
bugfix for install missing scripts/xonsh.bat (windows)

selepo:
Windows install is faulty, "xonsh.bat" is not created when installed as reported in issue #457.
Pull request to resolve issue.


scopatz:
Thanks @selepo

PR #465, opened 2015-11-03
Added info to windows.rst

selepo:
Added info about .xonshrc needed to be created and tips about console emulators to use on windows.
Should resolve some of the issues with installing av new user on windows.


scopatz:
Thanks for the doc updates!

Issue #466, opened 2015-11-04
xonsh.bat shouldn't be installed on non-Windows platforms


auscompgeek:

scopatz:
This was done as a convenience for conda convert to be able to generate windows packages on Linux and to hot fix people not being able to use xonsh on windows. A pull request fixing this would be very welcome. I completely agree.

krader1961:
@auscompgeek Please add a "Resolves #466" line to the description. Also, I'm a bit confused by the problem statement since I don't see a xonsh.bat script in my Anaconda tree (it is, of course, in my git repository). Having said that I agree that your proposed change is preferable to handcrafted scripts for invoking xonsh.

auscompgeek:
@krader1961 This *is* #466. It doesn't make sense for an issue to reference itself.

krader1961:
@auscompgeek you're correct. I intended to ask @uranusjr to add that line in his PR #468. The close proximity of the IDs resulted in my getting confused as I bounced between browser pages.

auscompgeek:
1. Make sure you have the latest `HEAD` checked out.
2. `pip install -e .`
3. Look in your `/usr/local/bin` to find a `xonsh.bat`.

I use `pip install --user` and `~/.local/bin`, but the issue is the same, nevertheless.

selepo:
Tried to issue a fix for this with this PR.
The only way I know to handle this is to check for platform in setup.py. 

https://github.com/scopatz/xonsh/pull/510

auscompgeek:
IMO, this was already fixed in #468. I've yet to come across anyone that doesn't have setuptools installed. Closing.

scopatz:
I personally try to avoid setuptools if I can :)

PR #468, opened 2015-11-04
Use console_scripts to generate entry point

uranusjr:
This is better than writing platform-dependant scripts to provide an entry point. The old way is kept as a fallback when setuptools is not available.

Should address #466 (at least when `setuptools` is available).

Reference:

http://python-packaging.readthedocs.org/en/latest/command-line-scripts.html#the-console-scripts-entry-point

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,7 +77,6 @@ def main():
         platforms='Cross Platform',
         classifiers=['Programming Language :: Python :: 3'],
         packages=['xonsh'],
-        scripts=['scripts/xonsh', 'scripts/xonsh.bat'],
         cmdclass={'install': xinstall, 'sdist': xsdist},
         )
     if HAVE_SETUPTOOLS:
@@ -87,8 +86,11 @@ def main():
             'pygments.lexers': ['xonsh = xonsh.pyghooks:XonshLexer',
                                 'xonshcon = xonsh.pyghooks:XonshConsoleLexer',
                                 ],
+            'console_scripts': ['xonsh = xonsh.main:main'],
             }
         skw['cmdclass']['develop'] = xdevelop
+    else:
+        skw['scripts'] = ['scripts/xonsh', 'scripts/xonsh.bat']
     setup(**skw)
 
 logo = """


scopatz:
Thanks for putting this in @Uranusjr. How does this affect the conda convert process, though?

krader1961:
LGTM. I recommend merging.

> How does this affect the conda convert process, though?

I merged this change on my system and executed the conda build commands without error:

```
$ git pull upstream refs/pull/468/head
$ conda build --no-test recipe
$ conda convert -p all -o /Users/krader/conda/conda-bld/ /Users/krader/conda/conda-bld/osx-64/xonsh-0.2.3-py34_11.tar.bz2
```

The resulting `/Users/krader/conda/conda-bld/linux-64/xonsh-0.2.3-py34_11.tar.bz2` file contained a `bin/xonsh` member with what appears to be the correct content:

```
#!/opt/anaconda1anaconda2anaconda3/bin/python
if __name__ == '__main__':
    import sys
    from xonsh.main import main

    sys.exit(main())
```

At least assuming the weird shebang line is normal for Anaconda build output. Similarly the resulting `/Users/krader/conda/conda-bld/win-64/xonsh-0.2.3-py34_11.tar.bz2` file contained a `Scripts/xonsh.bat` member with this content:

```
@echo off
set PYFILE=%~f0
set PYFILE=%PYFILE:~0,-4%-script.py
"%~f0\..\..\python.exe" "%PYFILE%" %*
```

I'm not motivated enough to test that but it appears reasonable.

P.S., Having never run Anaconda build commands before I had to make some educated guesses about the actual commands to run. Which is all the more reason for automating the process via a script. Which should be added as a "make build" target to the makefile I've created in

krader1961:
@uranusjr Please add a "Resolves #466" line to the description.

scopatz:
Thanks for doing the legwork @krader1961

scopatz:
Thanks @uranusjr

uranusjr:
Sorry for the lack of response; was taking a vacation this past weekend. Thanks for the effort!

asmeurer:
Re conda: I recommend using conda's entry_point mechanisms in the meta.yaml for this (that way at least conda users won't have to use these terrible setuptools entry points). I don't remember for sure but I think if you do that it will just work with conda convert (you'll have to check).

melund:
I just checked it. Conda convert does create a bat file when the package is built on linux.  We already have the conda entry_points stuff in meta.yaml:
```
  entry_points:
    - xonsh = xonsh.main:main
```

Building it on windows will make a small exe-file launcher... Maybe that feature haven't made to "conda convert" yet. 

I think there is an issue where conda build script may need the following for the setuptools entrypoints to work. Maybe @asmeurer  can remember something about this. 

```
build:
  preserve_egg_dir: True
```

We may also get rid of the pygments setuptools entry points if we can contribute the xonsh lexers directly to the pygments project. 

/Morten

scopatz:
> if we can contribute the xonsh lexers directly to the pygments project.

That would be awesome :)

PR #470, opened 2015-11-04
set prompt_toolkit completion to trigger on tab

gforsyth:
prompt_toolkit completion wasn't triggering for command arguments unless
the first letter was entered.  now it shows all the options when the
user hits tab, which matches the readline functionality and also the
default behavior in bash, zsh, fish, etc...

previously

```
git <TAB>
```
would return nothing, while

```
git s<TAB>
```

would offer `send-email`, `stage`, `stash`, `status`, etc...

Now 
```
git <TAB>
```

will show all available git arguments so long as `$BASH_COMPLETIONS` is set.


scopatz:
Thabks a ton Gil

PR #473, opened 2015-11-05
fix for history command line argument with negative slice

### LINKED ISSUE #471: "history show" slice notation should allow negative subscripts
You can show the n'th most recent command. For example:

    history show -5

But you can't show the n most recent commands. For example:

    history show -5:-1

or just

    history show -5:

produces this error:

**history: error: unrecognized arguments: -5:-1**

By default when I'm showing the history I typically want just the most recent 20 commands shown. Yes, I can write a function that uses `len(__xonsh_history__)` and some math to construct positive slice indexes but that shouldn't be necessary.

scopatz:
This fixes #471. Would you mind taking a look @krader1961?

### DIFF
--- a/xonsh/history.py
+++ b/xonsh/history.py
@@ -328,6 +328,12 @@ def _create_parser():
     return p
 
 
+def _space_negative_nums(s):
+    """Super janky solution to where argument values cannot start with dash"""
+    if s.startswith('-') and len(s) >= 2 and s[1].isdigit():
+        s = ' ' + s
+    return s
+
 def _show(ns, hist):
     idx = ensure_int_or_slice(ns.n)
     if len(hist) == 0:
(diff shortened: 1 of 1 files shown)


krader1961:
Wow! That is fast turnaround. I'm just about to head out the door to the first meeting of the "All Things Python" MeetUp. I'll review and test your fix when I return home. In a couple of weeks after I'm acquainted with the xonsh code I should be able to provide a patch along with my problem report so you're not having to do all the work :-)

scopatz:
> In a couple of weeks after I'm acquainted with the xonsh code I should be able to provide a patch along with my problem report so you're not having to do all the work :-)

That would be awesome :).  There are now enough users that I alone can't keep up with all of them.  Which is why it is so important and great to have many people contributing!

I totally agree that this is a less than ideal solution for long term extensiblity.  Would you have any objections to me merging this in and then you can put in another PR that fixes this up in the way that you see fit?

krader1961:
On Thu, Nov 5, 2015 at 8:12 AM, Anthony Scopatz <notifications@github.com>
wrote:
>
> That would be awesome :). There are now enough users that I alone can't
> keep up with all of them. Which is why it is so important and great to have
> many people contributing!
>
> I totally agree that this is a less than ideal solution for long term
> extensiblity. Would you have any objections to me merging this in and then
> you can put in another PR that fixes this up in the way that you see fit?
>

I'll code up an alternative this afternoon and we can then decide which one
should be merged or if a hybrid of the two is better.

-- 
Kurtis Rader
Caretaker of the exceptional canines Junior and Hank

scopatz:
Sounds even better!

krader1961:
I have proposed what I believe is a more robust solution that includes unit tests at https://github.com/scopatz/xonsh/pull/476. There are a couple of unrelated changes because my editor removes extraneous trailing whitespace from files I edit.

scopatz:
Superseded by #476

PR #474, opened 2015-11-05
Add equivalent of `\$` in bash as `prompt_end`

BYK:
Adds a new `prompt_end` formatter variable for prompt. The value is `$`
for normal users and `#` for root users.


scopatz:
Thanks @BYK, could you add this to the "Customizing the Prompt" portion of the tutorial docs, please?

BYK:
@scopatz done!

scopatz:
Thanks @BYK!

PR #475, opened 2015-11-05
Add default indent if not env variable not set

henryiii:
This tiny PR adds a default indent if indent is not set. Was trying to fix the error in 0.2.2, but found it was already fixed; this does help though if the user does not have `INDENT` set in `.xonshrc`.

### DIFF
--- a/xonsh/prompt_toolkit_key_bindings.py
+++ b/xonsh/prompt_toolkit_key_bindings.py
@@ -31,12 +31,12 @@ def load_xonsh_bindings(key_bindings_manager):
         If there are only whitespaces before current cursor position insert
         indent instead of autocompleting.
         """
-        event.cli.current_buffer.insert_text(env.get('INDENT'))
+        event.cli.current_buffer.insert_text(env.get('INDENT', '    '))
 
     @handle(Keys.BackTab)
     def insert_literal_tab(event):
         """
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks for putting this in @henryiii, but it is unnecessary because the env object overrides the `get()` method [as you can see here](https://github.com/scopatz/xonsh/blob/master/xonsh/environ.py#L292) to apply its own set of known defaults.  This was done in order to prevent having to repetitively specify the same default in many places across the code base. Now, the default only need to be given once and `get()` with no additional arguments will handle the rest.

henryiii:
But the default doesn't put in any space at all, it that intentional? I noticed that it was a special `.get`.

scopatz:
The default is a four space string as you can see here https://github.com/scopatz/xonsh/blob/master/xonsh/environ.py#L136

scopatz:
If you are on windows using cmd.exe there is a known issue with not indenting

henryiii:
Oh, I see what you are saying. I think I was originally using tab on an empty line, which seems to have a different behavior in master. I realized that, but not after I thought I'd tested the master version. (I'm on Mac atm)

henryiii:
Looks good, I'll close this PR. Thanks!

scopatz:
Thanks!

PR #476, opened 2015-11-06
Correct handling of negative history indexes.

krader1961:
Fix the handling of negative history indexes where a slice specification
is involved. Negative indexes not part of a slice specification worked by
accident because the argparse module (see section "16.4.4.3 Arguments
containing --") treats such values as positional arguments (which is a
misfeature of argparse in my opinion). I've also added unit tests for
"history show" to keep this from being broken in the future.

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -1,12 +1,15 @@
 """Tests the xonsh history."""
 from __future__ import unicode_literals, print_function
+import io
 import os
+import sys
 
 import nose
 from nose.tools import assert_equal, assert_true
 
 from xonsh.lazyjson import LazyJSON
 from xonsh.history import History, CommandField
+from xonsh import history
 
 HIST_TEST_KWARGS = dict(sessionid='SESSIONID', gc=False)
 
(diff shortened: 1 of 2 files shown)


tests/test_history.py
```diff
@@ -68,5 +71,75 @@ def test_cmd_field():
     yield assert_equal, None, hist.outs[-1]
     os.remove(FNAME)
 
+def test_show_cmd():
+    FNAME = 'xonsh-SESSIONID.json'
+    FNAME += '.show_cmd'
+    cmds = ['ls', 'cat hello kitty', 'abc', 'def', 'touch me', 'grep from me']
+
+    def FormatHistLine(idx, cmd):
```
scopatz: PEP8: functions should be named with `lowercase_with_underscores`

tests/test_history.py
```diff
@@ -68,5 +71,75 @@ def test_cmd_field():
...
+        return ' {:d}  {:s}\n'.format(idx, cmd)
+
+    def TestShowCmd(hist_args, commands, base_idx=0, step=1):
```
scopatz: same here.

xonsh/history.py
```diff
@@ -287,16 +288,16 @@ def _create_parser():
...
     subp = p.add_subparsers(title='action', dest='action')
     # show action
     show = subp.add_parser('show', help='displays current history, default action')
     show.add_argument('-r', dest='reverse', default=False, action='store_true',
                       help='reverses the direction')
-    show.add_argument('n', nargs='?', default=None, 
-                      help='displays n current history entries, n may be an int or use '
-                           'Python slice notation')
+    show.add_argument('n', nargs='?', default=None,
+                      help='display n\'th history entry if n is a simple int, '
```
scopatz: is `th` supposed to be `the`?

scopatz:
This looks great to me @krader1961! And many thanks for putting in the unit tests.  There are just a few topical changes to make...

xonsh/history.py
krader1961: Nope, it's supposed to be "n'th" as in "1st", "2nd", "3rd", "4th", etc. The original help implies that "history show 5" shows the five most recent history entries when in fact it shows #5 (i.e., the sixth).

scopatz: Ahhh I usually see this as just `nth`. This is fine then.

scopatz:
Is this ready to go in, now?

krader1961:
> Is this ready to go in, now?

Sadly, no. Changing the signature of history.main (by adding the `hist` kwarg) breaks the history command:

__Expected callable with 2 or 4 arguments, not 3__

So I need to take a look at how the `main` functions are invoked and adapt my change to that API.

scopatz:
Ahh, yeah, I was wondering if that was going to be a problem. I guess that it is.  This main function needs to follow the same rules as other aliases - see http://xonsh.org/tutorial.html#aliases

To get this to work, I would anticipate that you would need a wrapper function that looks like the following.

```python
def main_with_hist(args=None, stdin=None, hist=None):
    old, __xonsh_history__ = __xonsh_history__, hist or __xonsh_history__
    rtn = main(args=args, stdin=stdin)
    __xonsh_history__ = old
    return rtn
```

tests/test_history.py
krader1961: Oops. Sorry about that. My employer uses a modified PEP8 style where function names are in camel-case. Fixed.

krader1961:
Okay, this is ready to merge. Unit tests pass and using it interactively produces the expected results.

scopatz:
Thanks a million @krader1961!

PR #480, opened 2015-11-08
installation of jupyter hook to given destination

Dundee:
Installation on Archlinux using AUR does not work, because destination dir of jupyter hook cannot be changed.


scopatz:
Thanks a ton!

Dundee:
Great, thanks!

PR #482, opened 2015-11-09
Mention where the Arch linux package can be found

migueldvb:
There are many tools to download packages from the Arch User Repository apart from the ones included in the installation instructions.


scopatz:
Thanks!

PR #489, opened 2015-11-10
Make history module mostly lint clean.

krader1961:
This fixes most of the lint errors in the history code. It also fixes two bugs
that were found by pylint due to misnamed variables. This patch does not make
any functional changes. I'll fix the remaining lint errors in a subsequent
patch when I tackle cross-module issues.

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -1,96 +1,103 @@
 """Tests the xonsh history."""
+# pylint: disable=protected-access
 from __future__ import unicode_literals, print_function
 import io
 import os
 import sys
 
 import nose
-from nose.tools import assert_equal, assert_true
+from nose import tools as nose_tools
 
-from xonsh.lazyjson import LazyJSON
-from xonsh.history import History, CommandField
+from xonsh import lazyjson
 from xonsh import history
 
 HIST_TEST_KWARGS = dict(sessionid='SESSIONID', gc=False)
 
-
 def test_hist_init():
+    """Test initialization of the shell history."""
     FNAME = 'xonsh-SESSIONID.json'
     FNAME += '.init'
-    hist = History(filename=FNAME, here='yup', **HIST_TEST_KWARGS)
-    with LazyJSON(FNAME) as lj:
+    history.History(filename=FNAME, here='yup', **HIST_TEST_KWARGS)
+    with lazyjson.LazyJSON(FNAME) as lj:
         obs = lj['here']
-    assert_equal('yup', obs)
+    yield nose_tools.assert_equal, 'yup', obs
     os.remove(FNAME)
 
 
 def test_hist_append():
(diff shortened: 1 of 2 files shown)


tests/test_history.py
```diff
@@ -1,96 +1,103 @@
 """Tests the xonsh history."""
+# pylint: disable=protected-access
 from __future__ import unicode_literals, print_function
 import io
 import os
 import sys
 
 import nose
-from nose.tools import assert_equal, assert_true
+from nose import tools as nose_tools
```
scopatz: I am not sure that I am a fan of this, since it is different that the way I have seen it done in most other projects (numpy, scipy, etc).  What is the reason for renaming the module?

tests/test_history.py
```diff
@@ -1,96 +1,103 @@
...
 
-from xonsh.lazyjson import LazyJSON
-from xonsh.history import History, CommandField
+from xonsh import lazyjson
```
scopatz: Same comments here.

tests/test_history.py
krader1961: Within Google the rule is you never import specific symbols from a module (or package). You only import modules. This has several benefits. First, in the case of something like nose.tools you inevitably find yourself importing more and more symbols. In this specific case I switched several assert_true() calls to assert_is_none() and assert_is_not_none(). Second, by always qualifying references to external symbols with the module name it becomes obvious which symbols are defined in the current module versus symbols that are from a different module. Yes, I know most other projects do it the way you have been doing it. They're wrong :-) Trust me when I say that once the entire project adheres to this convention the readability will be greatly improved.

krader1961: FWIW, If the primary maintainers feel strongly about the Google import style rule that I've employed in this change I'm willing to revert those aspects. I'd much rather see the code be lint clean than fight for a change in the style of the code. On the other hand if I've managed to convince you the Google rule is better then I intend to apply it to all of the other modules as I work my way through the code removing lint errors so that there is a consistent style.

krader1961: Also, I see I may not have answered @scopatz's question:

> What is the reason for renaming the module?

The reason is to avoid namespace conflicts. In this case there is a xonsh.tools module and in general you rename the external module on import rather than an import from within the project.

scopatz: Yeah, I guess I feel pretty strongly that the "not importing symbols" rule breaks convention with basically every other big and small code base I am used to working on.  Also, it would affect pretty much every python file here.  I am ecstatic to have the linter pass, but I think the style changes would be too different for many of us - myself included!

krader1961:
Closing this in order to start with a clean slate for this sequence of patches now that I have a better understanding of the project coding style.

scopatz:
Sounds good!

PR #490, opened 2015-11-11
add open in editor to prompt_toolkit_shell

### LINKED ISSUE #445: Add 'edit and execute command` function
bash and zsh (and I'm sure others) have an edit and execute command that is invoked with Ctrl-e + Ctrl-x which opens the current contents of the command prompt in the system editor.  Edits made to the command are then run on save and quit.  

It could be this functionality is already baked in and I'm missing it?  In any case, it's especially handy for longer commands -- and given xonsh's multi-line input capabilities this seems like a good fit.

Thoughts?

gforsyth:
h/t @jonathanslenders

When using the prompt_toolkit shell, user can press Ctrl-X Ctrl-E to
open the current line in $EDITOR.

Limitations -- it can only open the _current_ line, so if you are
already in a multiline prompt, you can't edit what is already entered in
lines above.

However, you can enter multiline commands in the editor and these are
correctly interpreted by xonsh (assuming correct syntax, etc)

Resolves #445

### DIFF
--- a/xonsh/prompt_toolkit_shell.py
+++ b/xonsh/prompt_toolkit_shell.py
@@ -48,6 +48,7 @@ class PromptToolkitShell(BaseShell):
         self.pt_completer = PromptToolkitCompleter(self.completer, self.ctx)
         self.key_bindings_manager = KeyBindingManager(
             enable_auto_suggest_bindings=True,
+            enable_open_in_editor=True,
             enable_search=True, enable_abort_and_exit_bindings=True)
         load_xonsh_bindings(self.key_bindings_manager)
 


krader1961:
FWIW, While I think this is an important step in the right direction, and therefore gave it a "looks good to me" (LGTM), there really needs to be proper support for multi-line statements. If an issue isn't already open regarding editing multi-line statements please create one.

jonathanslenders:
Multi-line editing has to be approached differently in Xonsh itself. Currently the input is read as several individual single line inputs. Each time enter is pressed, a new line is read.

For the prompt-toolkit shell, we have to pass 'multiline=True' to the get_input function (which has been renamed to 'prompt', a while ago). Currently I'm not entirely sure what the best way is to display the dots for the indentation. (I will have a look into that again.) It should be possible.

Important will be to add a custom key binding for the Enter key. In multiline mode, enter will always insert newlines rather than accepting the input. In our case, we need enter to insert a newline when the input is incomplete (e.g. when a closing parentheses is missing), otherwise we should accept the input. If we don't do this, it will always be required to press Meta+Enter for accepting the input.

scopatz:
Thanks @gforsyth!  

Also note that even though the shell will glue individual lines together for multiline statements, you can also just pass it (`BaseShell.default()`) a single string that is the multiline statement that have been previously assembled.  This should hopefully make multiline input with ptk much easier. Not sure about the dots part though...

PR #492, opened 2015-11-12
Add a makefile defining some common tasks.

### LINKED ISSUE #430: The tests shouldn't use the user's local configuration
The tests currently source my bash profile, which makes them slower than they could be. In general, the tests should avoid using any user configuration to make them reproducible and to avoid false positive test failures.

krader1961:
This makefile makes it easier to run things like unit tests and pylint
against the code. Especially before doing a commit. It also helps ensure
unit tests are not affected by local config files (e.g., ~/.xonshrc).

Resolves #430


scripts/run_tests.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
scopatz: I am not sure that I want to add Bash to this project... Is there any way that this could be in xonsh?

scopatz:
Thanks for putting this in @krader1961. Also, can you add some documentation to the readme or somwhere in `docs/devsguide.rst` that explains this file.

scopatz:
Also, the makefile only does maintenance tasks, so do you think it would be more appropriate to live in the  `tests` directory?

scripts/run_tests.sh
krader1961: It doesn't depend on bash per se. It should work with any POSIX compliant shell installed as /bin/sh. Which is definitely going to be true for any unix like system but probably not for ms-windows. Having said that, sure, I don't see any reason it couldn't be a xonsh script. I'll rework it.

krader1961:
> do you think it would be more appropriate to live in the tests directory?

No, for two reasons. First, it will undoubtedly grow to encompass other actions such as automating the release process. Second, you can't actually run the tests if your cwd is ./tests unless you set PYTHONPATH to the project directory (the xonsh docs are wrong on this point).

krader1961:
> can you add some documentation to the readme or somwhere in docs/devsguide.rst that explains this file.

Yes, will do.

krader1961:
Please take another look. TODO is adding a make target to create HTML documentation in someones local repository so they can easily review changes to the documentation. But that's for another day.

docs/devguide.rst
```diff
@@ -61,22 +61,66 @@ is open to interpretation.
...
-First, install nose: http://nose.readthedocs.org/en/latest/
+First, install nose: http://nose.readthedocs.org/en/latest/. Second, ensure
+your cwd is the root directory of the project (i.e., the one containing the
+.git directory).
 
 To perform all unit tests::
 
-    $ cd tests/
-    $ nosetests
+    $ make test-all
```
scopatz: Using `make test-all` should be an alternative, not the solely recommended way. I don't think that testing should be strictly dependent on make, though if folks have it and want to use it then great.

docs/devguide.rst
```diff
@@ -61,22 +61,66 @@ is open to interpretation.
...
-named test_* and run every function (or method) named test_*.
+    $ make test
 
-Nosetests can also take file(s) as an argument. For example, to run just the
-lexer and parser module tests::
+If you want to run specific tests you can specify the test names. For example
+to run test_parser:
 
-    $ nosetests test_lexer.py test_parser.py
+    $ python3 -c 'import setup; setup.build_tables()'
```
scopatz: Same here, make should be a first-class alternative, though not a requirement.

scopatz:
Cool thanks! Just a couple of small documentation changes.

docs/devguide.rst
krader1961: > Using make test-all should be an alternative, not the solely recommended way. 

That's why I list three ways to do it. If someone wants to invoke nosetests directly I show how to do so in order to avoid picking up local dependencies. And that includes depending on the parser_tables.py module that is somewhere in your PYTHONPATH from having installed xonsh. If you run the tests by typing

    cd tests/
    nosetests

you're not actually testing the code in the ../xonsh directory; assuming you haven't explicitly included that directory at the front of PYTHONPATH, of course. You can see this by running "ls -lutr xonsh" and noticing that the access timestamps haven't changed. If you do the same ls command on the directory where you installed xonsh you'll see that those modules are accessed when you run the tests using the original directions.

docs/devguide.rst
```diff
@@ -61,22 +61,66 @@ is open to interpretation.
...
+    $ make test-all
+
+Or, if you want to do it the hard way:
+
+    $ scripts/run-tests.xsh
+
+Or, if you want to do it the really hard way:
+
+    $ python3 -c 'import setup; setup.build_tables()'
+    $ XONSHRC='' nosetests
```
scopatz: It would be nice if this line was xonsh-compliant, ie

```console
$ env XONSHRC='' nosetests
```

docs/devguide.rst
scopatz: > you're not actually testing the code in the ../xonsh directory; 

Right. I personally don't want to test the code that is in the ../xonsh directory.  I want to test the version of xonsh that is installed locally.  I understand that this is not what everyone wants and some people do want to test what is in the ../xonsh dir. I am happy to support that workflow and encourage people to use whatever workflow works best for them.

docs/devguide.rst
krader1961: Fixed.

docs/devguide.rst
krader1961: > I personally don't want to test the code that is in the ../xonsh directory.
> I want to test the version of xonsh that is installed locally.

That's an interesting approach :-) I've reworded the testing documentation to clarify matters and explicitly show how "cd tests" will result in using the locally installed xonsh source rather than the git repository source.

docs/devguide.rst
krader1961: > Same here, make should be a first-class alternative, though not a requirement.

Make is never a requirement. The contents of `Makefile` can always be executed by hand. Make is simply a well established mechanism in the UNIX community for automating such matters.

scopatz:
Thanks for the PR @krader1961! Sorry about all the back and forth about workflows :)

scopatz:
This is a nice addition that I am sure will help some people.

krader1961:
> Thanks for the PR @krader1961! Sorry about all the back and forth about workflows :)

No need to apologize. As a former Google employee I'm used to reviewers of my code being hard-nosed about requiring changes be perfect with respect to style, functionality, unit tests, etc. And I was similarly hard-nosed when someone asked me to review their change. If anything I would consider you too lenient in accepting suboptimal changes. But because this project is public it obviously requires a bit more flexibility in order to avoid discouraging random contributors. I think you're doing a fine job of balancing those conflicting goals.

scopatz:
Thanks! For what it is worth, I definitely see the project becoming stricter on these sorts of issues as more core developers appear and the project matures. But you are right to infer that currently if something is 'good enough, but not perfect' it is hard to turn it away.

PR #493, opened 2015-11-12
Capture process exit code for 'rtn' and 'last_cmd_rtn' and IGNOREEOF support.

nicfit:
While exploring xonsh I had a failed command and tested out what `echo $?` would do, I found 
that is what not what I expected, which is of course what bash does.  I dug deeper and found 
the rtn and last_cmd_rtm variable in history and found that my failed commands (e.g 'ls missingfille' and 'man missingmanpage') contain 0 for there exits codes, which is wrong as both exit non-zero for fails... this pull request fixes this. 

Also added IGNOREEOF, with bash-like behavior (true/false but not integer values currently).
I can split this into 2 pull-requests if desired.

Thanks for xonsh, the prospect of replacing bash is huge.

### DIFF
--- a/docs/envvars.rst
+++ b/docs/envvars.rst
@@ -74,6 +74,9 @@ applicable.
       - xonsh.environ.FORMATTER_DICT  
       - Dictionary containing variables to be used when formatting PROMPT and TITLE 
         see `Customizing the Prompt <tutorial.html#customizing-the-prompt>`_.
+    * - IGNOREEOF
+      - ``False``
+      - Prevents Ctrl-D from exiting the shell.
     * - INDENT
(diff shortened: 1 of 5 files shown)


krader1961:
Since I gave a LGTM for the IGNOREEOF portion of your pull request but not the exit code portion I think it is worth saying that unrelated changes should be segregated to separate pull requests (i.e., in separate branches).

nicfit:
Okay, I'll close an reopen, as separate pull requests. The IGNOREEOF add and message like "Use exit to close".. ala bash.  Based on your suggestions, I'll mess with the exit status fix for a future pull request.

PR #495, opened 2015-11-12
allow user to enable vi_mode in prompt_toolkit

### LINKED ISSUE #491: make vi mode configurable when using prompt-toolkit
The readline module automatically determines which key-bindings to use by reading `~/.inputrc`. Prompt-toolkit does not appear to have anything analogous. Vi mode can be enabled by passing `vi_mode=True` to the `prompt()` function. But a better option for xonsh is to pass `enable_vi_mode=True` to the `KeyBindingManager()` call.

To be determined is how to tell xonsh to use vi or emacs mode. It could parse `~/.inputrc` for compatibility with readline. But it's probably simpler to use a xonsh specific config option and or CLI argument.

gforsyth:
This adds the envvar `VI_MODE`, (default `False`).  If it is set to `True`,
either in a `.xonshrc` or dynamically, then the prompt is switched to
`vi_mode`.

`vi_mode` defaults to an `--insert--` environment, then user can hit ESC
to enter command mode, etc...'

Resolves #491

### DIFF
--- a/docs/envvars.rst
+++ b/docs/envvars.rst
@@ -152,6 +152,9 @@ applicable.
       - The title text for the window in which xonsh is running. Formatted in the same 
         manner as PROMPT, 
         see `Customizing the Prompt <tutorial.html#customizing-the-prompt>`_.
+    * - VI_MODE
+      - ``False``
+      - Flag to enable ``vi_mode`` in the ``prompt_toolkit`` shell.  
     * - XDG_CONFIG_HOME
       - ``~/.config``
       - Open desktop standard configuration home dir. This is the same default as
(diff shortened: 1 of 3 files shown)


krader1961:
lgtm

krader1961:
FWIW, At some point I'd like to see a discussion about whether the primary, and for the most part only, means of configuring Xonsh specific features should be via environment variables that are exported to every subprocess. I'm dismayed by the huge number of mostly Xonsh specific environment vars listed on http://xonsh.org/envvars.html.

gforsyth:
Might as well open an issue for that.

Are you dismayed regarding performance or regarding style/architecture/etc?

scopatz:
Seems like there is a merge conflict. Would you mind updating @gforsyth? Thanks!

gforsyth:
@scopatz -- merge conflict resolved (I conflicted with my own PR -- that's embarrassing).  I also noticed a bug I introduced that was throwing an exception on exiting `xonsh` but that's cleaned up now.

scopatz:
> I conflicted with my own PR -- that's embarrassing

Happens to the best of us :).  Thanks for the PR!

PR #498, opened 2015-11-14
Added IGNOREEOF env variable to disable CTRL-D shell exit.

nicfit:


krader1961:
LGTM other than the odd '\n' on its own line with an implicit concatenation to the string on the prior line.

scopatz:
Thanks @nicfit!

PR #500, opened 2015-11-15
Make history module mostly lint clean.

krader1961:
This fixes most of the lint errors in the history code. It also fixes
two bugs that were found by pylint due to misnamed variables. This patch
does not make any functional changes. I'll fix the remaining lint errors
in a subsequent patch when I tackle cross-module issues.

### DIFF
--- a/xonsh/history.py
+++ b/xonsh/history.py
@@ -1,4 +1,5 @@
 """Implements the xonsh history object"""
+import argparse
 import os
 import uuid
 import time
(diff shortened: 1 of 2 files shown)


xonsh/history.py
```diff
@@ -274,38 +312,35 @@ def flush(self, at_exit=False):
...
+                            self._cond, at_exit=at_exit)
         self.buffer.clear()
         return hf
 
+
 #
 # Interface to History
 #
-
-_HIST_PARSER = None
```
scopatz: What is the point of removing the cached parser object?

scopatz: I guess you could get the same behaviour with `functools.lru_cache()`, but still...

xonsh/history.py
```diff
@@ -392,13 +431,15 @@ def _main(hist, args):
...
     _MAIN_ACTIONS[ns.action](ns, hist)
 
+
 def main(args=None, stdin=None):
     """This is the history command entry point."""
     _ = stdin
-    _main(builtins.__xonsh_history__, args)
+    _main(builtins.__xonsh_history__, args)  # pylint: disable=no-member
+
+_HIST_PARSER = _create_parser()
```
scopatz: Oh I see. You have moved this to be created when you import the module. I'm not sure that I agree with this. Doing it this way creates the parser when xonsh is started up. Doing it the way that I had originally only creates the parser if and when the `history` command is used. The way implemented here adds to the startup time, which we get a lot of complaints about and so I try to be as careful as possible.

scopatz:
Thanks for putting this in @krader1961. Other than the parser caching issue, I think this looks good.

xonsh/history.py
krader1961: I see your point and concur; although, see the timings below for objective data. I've switched to memoizing the function to avoid the need for the global statement and symbol.

I instrumented the history code as follows:

```
_s = time.time()
_HIST_PARSER = _create_parser()
print('WTF _create_parser {:f}'.format(time.time() - _s))
```

I also captured the starting time of main.py:

```
"""The main xonsh script."""
import time
_s = time.time()
```

Then reported the delta before entering the command loop:

```
ignore_sigtstp()
print('WTF about to shell.cmdloop {:f}'.format(time.time() - _s))
shell.cmdloop()
```

I then invoked xonsh five times. The average time to call _create_parser()
was 0.008793 seconds. The average time to present a shell prompt was
2.057667 seconds. So creating the history parser took 0.4% of the startup
time using a minimal XONSHRC that did little more than set $PATH.

Rerunning with no XONSHRC file (via `XONSHRC='' xonsh`) provided average
times of five runs of 0.010973 seconds for _create_parser() and 0.312528
seconds -- 3.5% of the total startup time.

scopatz:
Thanks again for all of your hard work here @krader1961!

boegel:
#500! :tada:

Issue #486, opened 2015-11-09
Multi-line statements in Windows


johndcook:
Multi-line statements fail on Windows. For example, if I type `if True:` then I get an error message as soon as I hit Return. The same holds for `if`, `for`, or `def` statements.

I've seen this launching xonsh from cmd.exe and from PowerShell, on Windows 8.1 and on Windows 10. Running the latest released version at the time of writing, 0.2.3.

$ if True:
pre_input_hook failed
Readline internal error
Traceback (most recent call last):
  File "C:\bin\Anaconda3\lib\site-packages\pyreadline\modes\basemode.py", line 124, in readline_setup
    self.pre_input_hook()
  File "C:\bin\Anaconda3\lib\site-packages\xonsh\readline_shell.py", line 81, in inserter
    readline.redisplay()
AttributeError: 'module' object has no attribute 'redisplay'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\bin\Anaconda3\lib\site-packages\pyreadline\console\console.py", line 768, in hook_wrapper_23
    res = ensure_str(readline_hook(prompt))
  File "C:\bin\Anaconda3\lib\site-packages\pyreadline\rlmain.py", line 569, in readline
    self.readline_setup(prompt)
  File "C:\bin\Anaconda3\lib\site-packages\pyreadline\rlmain.py", line 564, in readline_setup
    BaseReadline.readline_setup(self, prompt)
  File "C:\bin\Anaconda3\lib\site-packages\pyreadline\rlmain.py", line 253, in readline_setup
    return self.mode.readline_setup(prompt)
  File "C:\bin\Anaconda3\lib\site-packages\pyreadline\modes\basemode.py", line 127, in readline_setup
    traceback.print_exc()
NameError: name 'traceback' is not defined
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK = True
  File "<string>", line None
SyntaxError: None: no further code

gforsyth:
Hi @johndcook , you might try using prompt_toolkit instead of pyreadline as it's better supported in windows.

tritemio:
Same error here on Windows 7. How do you switch to prompt_tookit?

krader1961:
On Mon, Nov 9, 2015 at 4:25 PM, Antonino Ingargiola <
notifications@github.com> wrote:

> Same error here on Windows 7. How do you switch to prompt_tookit?
>
[...]

PR #501, opened 2015-11-15
prompt toolkit windows changes

scopatz:
These were first suggested in #486.


krader1961:
LGTM. My question about simplifying setup.py if we're adding a hard dependency on setuptools shouldn't block merging this change as those simplifications can, and should be, done in a separate PR.

scopatz:
Thanks for the review @krader1961!

PR #502, opened 2015-11-16
Currently .gitignore ignores everything.

nicfit:
The first line removes is the major offender, the second just makes sense to also remove.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,6 @@ docs/_build/
 
 # temporary files from vim and emacs
 *~
-*#
-.#*
 *.swp
 *.swo
 


krader1961:
I don't understand why you think the first line you're removing is "the major offender". That pattern does not cause every file to be ignored. The pattern `*#` causes any filename ending in a hashmark to be ignored. The hashmark in this instance is not treated as introducing a comment. The next pattern you're removing causes any filename that begins with a period to be ignored. Both patterns are perfectly reasonable for this project.

I don't doubt you ran into some problem but it seems unlikely that removing those two patterns is the solution. It would help if you opened an issue that provided details about your environment, the commands you ran, the expected results, and what actually happened.

nicfit:
I use hg-git and in mercurial everything hidden, so I'll leave it on my repo and will file issues more.

krader1961:
> I use hg-git and in mercurial everything hidden,

You have my sympathies as SCM adapters like hg-git and git-svn tend to be rather painful to use except for the most mundane use cases. I would encourage telling the maintainers of hg-git that they are not correctly handling the patterns in .gitignore.

PR #503, opened 2015-11-16
Make ast lint clean.

krader1961:
Note that while I am personally comfortable with the use of functional programming idioms such as the map() function I'm removing them in this change because pylint complains about their use. Which is reasonable given the Python maxim that there should be one, and only one, obvious way to achieve a result. Which means in this instance using a list comprehension.


scopatz:
Thanks @krader1961, looks good to me.

Issue #487, opened 2015-11-10
Installation on Windows with Anaconda


tritemio:
I'm trying to install xonsh on Windows x64 with Anaconda and python 3.4 I get this error:

```
>pip install xonsh
Collecting xonsh
  Using cached xonsh-0.2.3.tar.gz
Requirement already satisfied (use --upgrade to upgrade): ply in c:\users\laser2002j\miniconda3\lib\site-packages (from xonsh)
Building wheels for collected packages: xonsh
  Running setup.py bdist_wheel for xonsh
  Complete output from command C:\Users\laser2002j\Miniconda3\python.exe -c "import setuptools;__file__='C:\\Users\\LASER2~1\\AppData\\Local\\Temp\\pip-build-egozq_by\\xonsh\\setup.py';exec(compile(open(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" bdist_wheel -d C:\Users\LASER2~1\AppData\Local\Temp\tmpd0h8znkypip-wheel-:
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "C:\Users\laser2002j\Miniconda3\lib\encodings\cp1252.py", line 23, in decode
      return codecs.charmap_decode(input,self.errors,decoding_table)[0]
  UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 4434: character maps to <undefined>

  ----------------------------------------
  Failed building wheel for xonsh
Failed to build xonsh
Installing collected packages: xonsh
  Running setup.py install for xonsh
Successfully installed xonsh-0.2.3
```

tritemio:
```
conda install xonsh -c scopatz
```

works as without errors instead.

scopatz:
Thanks for reportung @tritemio. Can you give us more info about your system? Are you sure that this pip you have was coming from conda? What is your default encoding?

tritemio:
This is a miniconda3 installation (user folder) using conda's pip in the root environment. The installation is a few weeks old and is using python 3.4 as default python. The latest Miniconda3 installer may have switched to python 3.5 in the meanwhile however.

For the default encoding, I don't know, I never changed it. I guess is windows 7 default.

jsarver:
I can confirm this issue as well. I think the problem is that beautiful Conch logo in the setup.py is angering windows when trying to use it's default encoding  CP-1252.

[...]

PR #504, opened 2015-11-16
add a uniform character encoding to all files

krader1961:
I wouldn't normally do something like this but issue #487 brought to
my attention the fact that too many of the python modules don't have
an encoding comment and of those that do there is a lot of pointless
inconsistency in the style of the comment.


scopatz:
Thanks @krader1961!

PR #506, opened 2015-11-17
make base_shell lint clean

krader1961:


xonsh/base_shell.py
```diff
@@ -15,11 +13,10 @@
 
 
 class TeeOut(object):
-    """Tees stdout into the original sys.stdout and another buffer instance that is 
-    provided.
-    """
+    """Tees stdout into the original sys.stdout and another buffer."""
 
     def __init__(self, buf, *args, **kwargs):
+        _, _ = args, kwargs
```
scopatz: This seems like the linter is just being silly. `*args, **kwargs` are there to future proof a function or method and this discourages that.  If you really want, I guess the names could be changed to `*_, **__` or `*_args, **_kwargs` but I would argue that is less readable. .

xonsh/base_shell.py
```diff
@@ -43,11 +40,10 @@ def flush(self):
 
 
 class TeeErr(object):
-    """Tees stderr into the original sys.stdout and another buffer instance that is 
-    provided.
-    """
+    """Tees stderr into the original sys.stdout and another buffer."""
 
     def __init__(self, buf, *args, **kwargs):
+        _, _ = args, kwargs
```
scopatz: same.

xonsh/base_shell.py
```diff
@@ -71,9 +67,12 @@ def flush(self):
 
 
 class Tee(io.StringIO):
-    """Class that merges tee'd stdout and stderr into a single buffer, namely itself. 
+    """Class that merges tee'd stdout and stderr into a single buffer.
+
     This represents what a user would actually see on the command line.
     """
+    # pylint is a stupid about counting public methods when using inheritance.
+    # pylint: disable=too-few-public-methods
```
scopatz: Wow, I agree.

xonsh/base_shell.py
```diff
@@ -118,10 +117,10 @@ def default(self, line):
         src, code = self.push(line)
         if code is None:
             return
-        hist = builtins.__xonsh_history__
+        hist = builtins.__xonsh_history__  # pylint: disable=no-member
         ts1 = None
-        tee = Tee() if builtins.__xonsh_env__.get('XONSH_STORE_STDOUT') \
-                    else io.StringIO()
+        store_stdout = builtins.__xonsh_env__.get('XONSH_STORE_STDOUT', False)  # pylint: disable=no-member
```
scopatz: Please remove the `False` here.  The env's `get()` method provides its own defaults so that they can be specified in exactly one place in the code base.

scopatz: Also note that `builtins.__xonsh_env__.get('XONSH_STORE_STDOUT')` will return `False` and not `None`

scopatz:
Thanks again @krader1961. I have added some relatively small comments.

xonsh/base_shell.py
krader1961: It's not really being silly. It's legitimately pointing out that an argument isn't used which frequently implies either a bug (typically due to a misspelled var) or a potential API bogosity. However, as you point out unused args do have legitimate use cases. Within Google the idiom is to assign unused args to `_` to tell both lint and readers of the code that the args are deliberately being ignored.

Having said that I personally think the `*args, **kwargs` presence in this method are problematic. If the interface that TeeOut is expected to implement changes then the signature of its methods should change. Silently discarding unexpected positional and keyword args is a landmine waiting to hurt someone.

xonsh/base_shell.py
krader1961: Okay.

xonsh/base_shell.py
scopatz: > Within Google the idiom is to assign unused args to _ to tell both lint and readers of the code that the args are deliberately being ignored.

Yea, this idiom is pretty common. And I understand what the linter is trying to do, but I think that it is a bit excessive here.

> Having said that I personally think the *args, **kwargs presence in this method are problematic. If the interface that TeeOut is expected to implement changes then the signature of its methods should change. Silently discarding unexpected positional and keyword args is a landmine waiting to hurt someone.

This is meant to duck type file handles like you might find in `io`. These have a lot of arguments, most of which won't be relevant. That is why I think capturing unused arguments is a valid strategy here. Unfortunately, you can't have two arguments with the same name and so doing `*_, **_` is not allowed.

xonsh/base_shell.py
krader1961: Related to this my intent is to eventually replace references to builtins.__xonsh_env__ with something like builtinsx.xonsh.env. That is, rename built_ins.py to builtinsx.py. Then replace the global functions and symbols such as "ENV" with a class that encapsulates that functionality and is exposed as a singleton named builtinsx.xonsh. That will eliminate all of the references to xonsh specific double-underscore references of the builtins module. That will eliminate the need to suppress a few hundred "no-member" lint errors as well as be clearer and more concise.

scopatz: You might want to open a different issue or thread on the mailing list for that proposal.

xonsh/base_shell.py
krader1961: Upon further inspection I see that the TeeOut and TeeErr classes are only instantiated from class Tee and are only called with one positional argument. So why do they need the *args and **kwargs arguments? Their interfaces are only relevant to another class in the same module and thus do not need "future proofing". Too, shouldn't they be private to the module? Only the Tee class is meant to implement the io.StringIO interface.

This type of issue is why I think the code should be lint clean. Most of the errors pylint reports are legitimate. The few that are not can be suppressed via pylint directives.

scopatz:
This works, thanks @krader1961!

PR #508, opened 2015-11-18
make completer mostly lint clean

krader1961:


xonsh/completer.py
```diff
@@ -201,24 +210,29 @@ def find_and_complete(self, line, idx, ctx=None):
             endidx += 1
         endidx = endidx - 1 if line[endidx] in WS else endidx
         prefix = line[begidx:endidx+1]
-        return self.complete(prefix, line, begidx, endidx, ctx=ctx), begidx, endidx
+        return (self.complete(prefix, line, begidx, endidx, ctx=ctx),
+                begidx, endidx)
 
     def _add_env(self, paths, prefix):
+        _ = self
```
scopatz: I am not sure how I feel about this...

xonsh/completer.py
```diff
@@ -201,24 +210,29 @@ def find_and_complete(self, line, idx, ctx=None):
...
             keylow = key.lower()
-            paths.update({'$' + k for k in builtins.__xonsh_env__ if startswither(k, key, keylow)})
+            paths.update('$' + k for k in builtins.__xonsh_env__
+                         if startswither(k, key, keylow))
 
     def _add_dots(self, paths, prefix):
-        if prefix in {'', '.'}:
-            paths.update({'./', '../'})
-        if prefix == '..':
+        _ = self
```
scopatz: same.

xonsh/completer.py
```diff
@@ -201,24 +210,29 @@ def find_and_complete(self, line, idx, ctx=None):
...
+        if prefix in ('', '.'):
+            paths.update('./', '../')
+        elif prefix == '..':
             paths.add('../')
 
     def _add_cdpaths(self, paths, prefix):
         """Completes current prefix using CDPATH"""
+        _ = self
```
scopatz: same.

xonsh/completer.py
```diff
@@ -239,6 +253,7 @@ def cmd_complete(self, cmd):
 
     def module_complete(self, prefix):
         """Completes a name of a module to import."""
+        _ = self
```
scopatz: same.

xonsh/completer.py
```diff
@@ -312,19 +322,28 @@ def bash_complete(self, prefix, line, begidx, endidx):
         return rtn
 
     def _source_completions(self):
+        """Return list of source commands for Bash completion files.
+
+        Note that this silently ignores any file name in the completion list
+        that does not exist.
+        """
+        _ = self
```
scopatz: here too

xonsh/completer.py
```diff
@@ -312,19 +322,28 @@ def bash_complete(self, prefix, line, begidx, endidx):
...
+                        f).replace('\\', '/')
                 srcs.append('source ' + f)
         return srcs
 
     def _load_bash_complete_funcs(self):
+        """Attempt to load any Bash completion functions we can find."""
         self.bash_complete_funcs = bcf = {}
         inp = self._source_completions()
-        if len(inp) == 0:
+        if not inp:  # no completion files could be found
```
scopatz: This is less precise then the original line and uses a negative conditional rather than a positive one.

xonsh/completer.py
```diff
@@ -339,8 +358,9 @@ def _load_bash_complete_funcs(self):
             bcf[cmd] = m.group(1)
 
     def _load_bash_complete_files(self):
+        """Load the Bash completion files we can locate."""
         inp = self._source_completions()
-        if len(inp) == 0:
+        if not inp:  # no completion files could be found
```
scopatz: same

xonsh/completer.py
```diff
@@ -396,30 +417,32 @@ def attr_complete(self, prefix, ctx):
...
-        # did aliases change?
+        # Did aliases change?
         al_hash = hash(tuple(sorted(builtins.aliases.keys())))
         self._alias_checksum = al_hash
         cache_valid = cache_valid and al_hash == self._alias_checksum
         pm = self._path_mtime
-        # did the contents of any directory in PATH change?
-        for d in filter(os.path.isdir, path):
+        # Did the contents of any directory in PATH change?
+        for d in (p for p in path if os.path.isdir(p)):
```
scopatz: I don't see the need for this.  I personally feel that a functional style is legitimate and easier to read than a generator embedded into a loop.

xonsh/completer.py
```diff
@@ -396,30 +417,32 @@ def attr_complete(self, prefix, ctx):
...
-        self._path_mtime = pm
-        if cache_valid:
-            return self._cmds_cache
-        allcmds = set()
-        for d in filter(os.path.isdir, path):
-            allcmds |= set(os.listdir(d))
-        allcmds |= set(builtins.aliases.keys())
-        self._cmds_cache = frozenset(allcmds)
+
+        if not cache_valid:
```
scopatz: This adds a level of nesting and replaces a positive conditional with a negative one.

xonsh/completer.py
```diff
@@ -449,14 +469,14 @@ def option_complete(self, prefix, cmd):
...
-                # This is a trick to get rid of reverse line feeds
+                # This is a trick to get rid of reverse line feeds.
                 text = subprocess.check_output(["col", "-b"],
                                                stdin=manpage.stdout)
                 text = text.decode('utf-8')
                 scraped_text = ' '.join(SCRAPE_RE.findall(text))
                 matches = INNER_OPTIONS_RE.findall(scraped_text)
                 self._options[cmd] = matches
-            except:
+            except Exception:  # pylint: disable=broad-except
```
scopatz: Thanks, a bare except should never be allowed.

xonsh/completer.py
```diff
@@ -467,10 +487,13 @@ def _load_cached_options(self):
...
                 self._options = pickle.load(f)
-        except:
+        except Exception:  # pylint: disable=broad-except
             self._options = {}
 
     def _save_cached_options(self):
         """Save completions to file."""
-        with open(OPTIONS_PATH, 'wb') as f:
-            pickle.dump(self._options, f)
+        try:
```
scopatz: What is the reason for this try block?

xonsh/completer.py
krader1961: The alternative is to either make this method a module level function or wrap it in `@staticmethod`. The rule of thumb within Google is to avoid `@staticmethod` and use the idiom I've introduced here if the function is tightly bound to the class otherwise move it to be a method level function.

xonsh/completer.py
krader1961: You're going to see this a lot in my forthcoming lint cleanup changes since there are a large number of class methods that do not depend on the state of an instance of the class.

xonsh/completer.py
scopatz: If `self` is truly not used, then this can probably be moved to the global scope.

xonsh/completer.py
krader1961: Again, sorry, but this is Google style and even if I weren't a recent Google employee would agree with that style rule.

The predicate is whether or not `inp` has been set to a value that isn't empty (e.g., `[]`) or undefined (e.g., None). If `inp` being an empty sequence versus `None` was important then that would be tested for. The "len(...) == 0" idiom makes sense for other language but not Python.

xonsh/completer.py
scopatz: I think you mean module level, not method level, in which case I agree :)

krader1961: It's a judgement call. If the method is not tightly coupled to this class then, yes, these functions should be moved to module scope. My change to ignore the unused `self` was the simplest change. But I'm happy to promote these methods to functions if that improves readability.

xonsh/completer.py
scopatz: But I argue that what matters here is that it is empty, not that `bool(inp) is False`.  As such, the original expression makes that more explicit. If `inp` where allowed to have a non-Sequence-like interface, than the `bool(inp) is False` mechanism would be e better first test.  But if inp is not Sequence-like, this should die as soon as possible, which the length test does.

xonsh/completer.py
scopatz: I think that is the right thing to do here.  They should probably remain private, though.

xonsh/completer.py
krader1961: The answer is that in Python there should be one obvious way to do something. And the Python community has decided that is list/set/dict comprehensions. Which is why pylint complains about idioms more appropriate for scala, etc. In the Go lang world there aren't any such arguments because an deviation from the output of gofmt is deemed invalid. Pylint is the closest thing we have to that in that in the Python community.

scopatz: To also quote the Zen of Python "Readability counts."  I find the original line much more readable.

xonsh/completer.py
krader1961: The original sequence of statements is suboptimal from both cognitive and performance perspectives. Consider that the assignment to `self._path_mtime` only need to occur if the cached data is not valid. The whole point of this code is to only update the cached data if it isn't valid.

scopatz: > The original sequence of statements is suboptimal from both cognitive and performance perspectives.

I don't think that these two implementations have significantly different performance characteristics, though I could be wrong. In any event, I find this version less readable because of the removal of the filter, the increased average indentation level, and the extra vertical whitespace.

xonsh/completer.py
krader1961: Because I have moved the `try..except...` from the `ManCompleter.__del__` method. There did not appear to be any reason for the asymmetry between the `_load_cached_options` and `_save_cached_options` methods. If there is a reason for the asymmetry then that should be documented.

scopatz: Ahh Ok. that makes a lot more sense.

krader1961: > Ahh Ok. that makes a lot more sense.

Which is why I'm starting to think that future lint cleanups should be more focused. I have one more change in the pipeline that is broad; i.e., tries to address most of the lint problems in a given module. But after that I'm going to focus on a subset of problems in each module to keep the review costs as small as possible. That is, I'm no longer going to try and fix a large and disparate set of lint problems in each module in a given commit.

scopatz: Sounds great to me :)

krader1961:
Somehow while merging the recent commits on master I screwed up this branch (the ID of the commit changed). So I'm going to drop this PR, recreate the branch, and try again.

PR #510, opened 2015-11-18
bugfix for issue 466 with xonsh.bat installed in non-windows

selepo:
Added logic to setup.py to install xonsh.bat non windows and xonsh on non-windows.


scopatz:
Thanks for the PR @selepo. It looks like there are merge conflicts with the current master. Can you please resolve? Thanks!

krader1961:
@selepo The parent commit of your change is over two weeks old. Since then there have been a couple of changes to setup.py. The important one being PR #468 to tell setuptools to construct the appropriate script rather than blindly install the two prepackaged scripts. Your change might still be useful for the non-setuptools case.

selepo:
Sorry about that, I missed that I needed to merge with upstream before making changes. Still learning. Should work now. NOTE: I have only tested this on windows...

asmeurer:
You need to edit the line below (right before the call to `setup`) where `scripts` is currently being set.

scopatz:
Thanks!

PR #512, opened 2015-11-19
fix some completer.py lint errors

krader1961:
This was originally https://github.com/scopatz/xonsh/pull/508 but I screwed
up my branch. Since I also needed to do more restructuring of the code in
light of feedback from @scopatz I decided to save the diff, close that
PR, recreate the branch, then reapply the diff and continue working on it
to reach this state.

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -40,9 +40,91 @@ for ((i=0;i<${{#COMPREPLY[*]}};i++)) do echo ${{COMPREPLY[i]}}; done
 
 WS = set(' \t\r\n')
 
+
+def _add_env(paths, prefix):
+    if prefix.startswith('$'):
+        csc = builtins.__xonsh_env__.get('CASE_SENSITIVE_COMPLETIONS')
+        startswither = startswithnorm if csc else startswithlow
+        key = prefix[1:]
+        keylow = key.lower()
+        paths.update('$' + k for k in builtins.__xonsh_env__
+                     if startswither(k, key, keylow))
+
+
+def _add_dots(paths, prefix):
+    if prefix in ('', '.'):
+        paths.update('./', '../')
+    elif prefix == '..':
+        paths.add('../')
+
+
+def _add_cdpaths(paths, prefix):
+    """Completes current prefix using CDPATH"""
+    env = builtins.__xonsh_env__
+    csc = env.get('CASE_SENSITIVE_COMPLETIONS')
+    for cdp in env.get('CDPATH'):
+        test_glob = os.path.join(cdp, prefix) + '*'
+        for s in iglobpath(test_glob, ignore_case=(not csc)):
+            if os.path.isdir(s):
+                paths.add(os.path.basename(s))
+
+
+def _path_complete(prefix, cdpath=False):
+    """Completes based on a path name."""
+    space = ' '  # intern some strings for faster appending
+    slash = '/'
+    tilde = '~'
+    paths = set()
+    csc = builtins.__xonsh_env__.get('CASE_SENSITIVE_COMPLETIONS')
+    if prefix.startswith("'") or prefix.startswith('"'):
+        prefix = prefix[1:]
+    for s in iglobpath(prefix + '*', ignore_case=(not csc)):
+        if space in s:
+            s = repr(s + (slash if os.path.isdir(s) else ''))
+        else:
+            s = s + (slash if os.path.isdir(s) else space)
+        paths.add(s)
+    if tilde in prefix:
+        home = os.path.expanduser(tilde)
+        paths = {s.replace(home, tilde) for s in paths}
+    _add_env(paths, prefix)
+    _add_dots(paths, prefix)
+    if cdpath:
+        _add_cdpaths(paths, prefix)
+    return {_normpath(s) for s in paths}
+
+
+def _module_complete(prefix):
+    """Completes a name of a module to import."""
+    prefixlow = prefix.lower()
+    modules = set(sys.modules.keys())
+    csc = builtins.__xonsh_env__.get('CASE_SENSITIVE_COMPLETIONS')
+    startswither = startswithnorm if csc else startswithlow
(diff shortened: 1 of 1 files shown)


xonsh/completer.py
```diff
@@ -50,15 +132,19 @@ def startswithlow(x, start, startlow=None):
 
 
 def startswithnorm(x, start, startlow=None):
-    """True if x starts with a string s. Ignores its lowercase version, but
-    matches the API of startswithlow().
+    """True if x starts with a string s.
+
+    Ignores its lowercase version, but matches the API of startswithlow().
     """
+    _ = startlow
```
scopatz: Please remove these kinds of lines.

xonsh/completer.py
```diff
@@ -311,20 +340,11 @@ def bash_complete(self, prefix, line, begidx, endidx):
...
-                srcs.append('source ' + f)
-        return srcs
-
     def _load_bash_complete_funcs(self):
+        """Attempt to load any Bash completion functions we can find."""
         self.bash_complete_funcs = bcf = {}
-        inp = self._source_completions()
-        if len(inp) == 0:
+        inp = _source_bash_completions()
+        if not inp:  # no completion files could be found
```
scopatz: Please return this to the original positive conditional.

xonsh/completer.py
```diff
@@ -339,8 +359,9 @@ def _load_bash_complete_funcs(self):
             bcf[cmd] = m.group(1)
 
     def _load_bash_complete_files(self):
-        inp = self._source_completions()
-        if len(inp) == 0:
+        """Load the Bash completion files we can locate."""
+        inp = _source_bash_completions()
+        if not inp:  # no completion files could be found
```
scopatz: same.

xonsh/completer.py
```diff
@@ -362,6 +383,7 @@ def _load_bash_complete_files(self):
 
     def attr_complete(self, prefix, ctx):
         """Complete attributes of an object."""
+        _ = self
```
scopatz: Please remove this line, make this a static method, or by make this method into a function at global scope.

xonsh/completer.py
```diff
@@ -396,27 +418,30 @@ def attr_complete(self, prefix, ctx):
...
-        # did aliases change?
+        # Did aliases change?
         al_hash = hash(tuple(sorted(builtins.aliases.keys())))
         self._alias_checksum = al_hash
         cache_valid = cache_valid and al_hash == self._alias_checksum
         pm = self._path_mtime
-        # did the contents of any directory in PATH change?
-        for d in filter(os.path.isdir, path):
+        # Did the contents of any directory in PATH change?
+        for d in (p for p in path if os.path.isdir(p)):
```
scopatz: Please return to the `filter()` call to improve readability.

xonsh/completer.py
```diff
@@ -396,27 +418,30 @@ def attr_complete(self, prefix, ctx):
...
             m = os.stat(d).st_mtime
             if m > pm:
                 pm = m
                 cache_valid = False
-        self._path_mtime = pm
+
         if cache_valid:
             return self._cmds_cache
+
+        self._path_mtime = pm
```
scopatz: Why did this move down?

xonsh/completer.py
```diff
@@ -396,27 +418,30 @@ def attr_complete(self, prefix, ctx):
...
         allcmds = set()
-        for d in filter(os.path.isdir, path):
+        for d in (p for p in path if os.path.isdir(p)):
```
scopatz: Please replace with filter()

scopatz:
Thanks for putting this back in @krader1961. Basically the same comments as before.

xonsh/completer.py
krader1961: > Please remove these kinds of lines.

So you really don't care about the code being lint clean?

xonsh/completer.py
krader1961: > Please return this to the original positive conditional. 

No. The original "positive conditional" is both needlessly verbose and antithetical to the Python maxim that there should be one obvious way to do something. In this case we're asking whether or not `inp` contains anything of interest. It doesn't matter if it is a sequence of length zero or None.

xonsh/completer.py
krader1961: The functional idioms such as map and filter are appealing, especially to anyone familiar with a language like Scala or Lisp, but antithetical to the Python way of doing things. Which is why pylint complains about their use.

xonsh/completer.py
scopatz: It is a false conflict to pose readability against clean linting.  It is possible to have both, neither, and any combination in between.  I am happy to have the code lint cleanly, but not at the sacrifice of readability, and especially if that means performing extraneous operations like the assignment statements here. 

In this particular case, the function is meant to match a similar interface.  Thus I believe that the linter is being overly pedantic because it does not associate this function with its partner `startswithlow()`.  But because we know about this association the correct thing is probably to turn this linter error off. This happens in a comment (which is free in terms of execution) and not in an assignment to another unused variable name (which is just a convention and has a run time cost).

xonsh/completer.py
scopatz: Python is a multi-paradigm language, one of which is functional. If map and filter and etc. were antithetical, they would not be builtins.

xonsh/completer.py
scopatz: Also note, from the pylint homepage:

```
Fully customizable
Modify your pylintrc to customize which errors or conventions are important to you. The big advantage with Pylint is that it is highly configurable, customizable, and you can easily write a small plugin to add a personal feature.

More about message control (FAQ)
```

I get the feeling that a lot of the issues that are getting brought up are related to trying to enforce conventions that I don't agree with being followed everywhere all of the time, and not with explicit errors or PEP8 violations or violations to the style addendums listed in the dev's guide (http://xonsh.org/devguide.html#python-style-guide)

I would probably be best to disable such convention enforcement for the whole project, if possible.

PR #513, opened 2015-11-19
Helper for printing TERM_COLORS

nicfit:
Handles the \001 and \002 replacements.

### DIFF
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -204,7 +204,7 @@ class ReadlineShell(BaseShell, Cmd):
                     if inserter is not None:
                         readline.set_pre_input_hook(None)
                 else:
-                    print_color(self.prompt)
+                    print_color(self.prompt, file=self.stdout)
                     if line is not None:
                         os.write(self.stdin.fileno(), line.encode())
                     if not exec_now:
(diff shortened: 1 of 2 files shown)


xonsh/tools.py
```diff
@@ -757,3 +757,7 @@ def format_prompt_for_prompt_toolkit(prompt):
     token_names = [get_xonsh_color_names(c) for c in parts[::2]]
     cstyles = [_make_style(c) for c in token_names]
     return token_names, cstyles, strings
+
+
+def print_color(string):
```
scopatz: This function needs a docstring.

xonsh/readline_shell.py
```diff
@@ -204,9 +204,7 @@ def _cmdloop(self, intro=None):
                     if inserter is not None:
                         readline.set_pre_input_hook(None)
                 else:
-                    self.stdout.write(self.prompt.replace('\001', '')
-                                                 .replace('\002', ''))
-                    self.stdout.flush()
+                    print_color(self.prompt)
```
scopatz: This does not exactly replicate the code that was there originally.  In particular, the original uses `self.stdout`, while the `print_color()` function will go to `sys.stdout`.

scopatz:
Thanks for putting this in @nicfit

nicfit:
It needs work, but it sounds like there is some interest so..I add a docs, and a file kwarg so self.stdout can be passed (or sys.stderr). I'll reopen then.

scopatz:
Is this ready @nicfit?

nicfit:
@scopatz It is.

scopatz:
Thanks @nicfit!

PR #515, opened 2015-11-20
Fixed 'bg' for Popen

nicfit:
To reproduce:
```bash
$ sleep 10
Ctrl-Z
$ bg
```
The job is stopped state instead of runniong (T stat code in 'ps') and stays that way.

### DIFF
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -100,10 +100,9 @@ else:
             return
         job = builtins.__xonsh_all_jobs__[act]
         obj = job['obj']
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks for the fix @nicfit - I'll test it tomorrow.

scopatz:
Thanks a ton @nicfit! I confirmed that this works when `$XONSH_STORE_STDOUT is False` and it doesn't seem to affect things when `$XONSH_STORE_STDOUT is True`. Great work!

PR #518, opened 2015-11-20
enable sudo completion in pt and readline shells

gforsyth:
If the command is 'sudo', this short-circuits the completion and just
cuts sudo out of the logic entirely, allowing the completer to act as if
it weren't there.

This might be a little too hacky, but it works on my machines without issue and enables completion lines beginning with `sudo` for both `prompt_toolkit` and `readline`.  

I'm also submitting this because `sudo` completion doesn't work for me even though it's in my `$BASH_COMPLETIONS`.  If others have `sudo` completion working already, then I'll dig around more to try and determine the root cause of the issue.

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -28,6 +28,8 @@ XONSH_TOKENS = {
     '...'
 }
 
+COMPLETION_SKIP_TOKENS = {'sudo', 'time'}
+
 BASH_COMPLETE_SCRIPT = """source {filename}
 COMP_WORDS=({line})
 COMP_LINE={comp_line}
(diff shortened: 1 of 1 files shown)


xonsh/completer.py
```diff
@@ -138,6 +143,8 @@ def complete(self, prefix, line, begidx, endidx, ctx=None):
             if len(rtn) == 0:
                 rtn = self.path_complete(prefix)
             return sorted(rtn)
+#        elif prefix.startswith('sudo '):
+#            rtn = self.cmd_complete(prefixe[5:], begidx,endidx)
```
scopatz: Can you remove these comments, please?

scopatz:
Thanks @gforsyth! This is a definitely needed feature.  There are a number of commands that run other commands where this kind of thing would be nice (time, etc).  Maybe you could make this a little less hacky by having a set at module scope that contains this commands and then you conditional here can check if the first token is in that set.

gforsyth:
Ok, will do.

gforsyth:
Ok, just added a set of tokens to skip over in the completer.  Right now it's just `time` and `sudo` but we can always add more.

scopatz:
Amazing! Thanks a ton!

PR #520, opened 2015-11-21
Add support for -i and -l options

### LINKED ISSUE #517: -l and -i command line options
If xonsh should be BASHwards compatible, maybe it could be worth introducing -l (login) and -i (interactive) options? At least as fake arguments, because when I set xonsh as my default shell (Linux Mint) and run [Spacemacs](https://github.com/syl20bnr/spacemacs) I get error:

```
Error (use-package): exec-path-from-shell :init: Non-zero exit code from shell /usr/local/bin/xonsh invoked with args ("-l" "-i" "-c" "/usr/bin/printf '__RESULT\\000%s\\000%s' \"${PATH-a421311e83eccdcf118d7f98b41e20d8}\" \"${MANPATH-a421311e83eccdcf118d7f98b41e20d8}\"").  Output was:
"usage: xonsh [-h] [-V] [-c COMMAND] [--no-rc] [-D [ITEM [ITEM ...]]]
             [--shell-type {readline,prompt_toolkit}]
             [script-file] [args [args ...]]
xonsh: error: unrecognized arguments: -l -i
"
```

bpollack:
Support for the login option is very minimal, but the `XONSH_LOGIN` variable is set in this case so that scripts may take an appropriate action.

I know that the guidelines call for all new code to have tests, but I'm unclear what the right way to test this is; we currently have no tests that actually check the `main` function at all, and at any rate, I'd usually prefer using a tool like [cram](https://pypi.python.org/pypi/cram) to test this certain class of feature anyway. I'm happy to amend the PR with guidance.

Fixes #517

### DIFF
--- a/docs/envvars.rst
+++ b/docs/envvars.rst
@@ -205,6 +205,8 @@ applicable.
     * - XONSH_INTERACTIVE
       - 
       - ``True`` if xonsh is running interactively, and ``False`` otherwise.
+    * - XONSH_LOGIN
+      - ``True`` if xonsh is running as a login shell, and ``False`` otherwise.
     * - XONSH_SHOW_TRACEBACK
       - ``False`` but not set
       - Controls if a traceback is shown exceptions occur in the shell. Set ``True`` 
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+"""Tests the xonsh main function."""
+from __future__ import unicode_literals, print_function
+
+import builtins
+from unittest.mock import patch
+
+import nose
+from nose.tools import assert_true, assert_false
+
+import xonsh.main
+
+from tools import mock_xonsh_env
+
+
+def test_login_shell():
+    def Shell(*args, **kwargs):
+        pass
+
+    with patch('xonsh.main.Shell', Shell), mock_xonsh_env({}):
+        xonsh.main.premain([])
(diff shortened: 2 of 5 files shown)


scopatz:
Thanks a ton for putting this in @bpollack! This looks good but some ancillary code needs to be modified since you are adding an environment variable.

First, in environ.py, please add a `XONSH_LOGIN` to the `DEFAULT_ENSURERS` dictionary.  Then add a likewise entry to the `DEFAULT_VALUES`.  Lastly, in the `docs/envvars.rst` file, add approriate documentation for this variable.  With that, this should be good to go.

As per testing, you are right that we don't do a good enough testing the CLI.  Cram looks cool and I was unware of it.  However, for the changes here it is probably sufficient to just test the `premain()` fiunction on its own.  You may need/want to swap out the `builtins.__xonsh_env__` for an empty dictionary.

Thanks again, I really appreciate this!

bpollack:
No problem! Did almost all of this; having a bit of a chicken-and-egg fight with replacing `__xonsh_env__` for `premain`, but I should have a patch for you tomorrow or Monday.

scopatz:
Awesome! Thanks!

bpollack:
@scopatz This is as good as it's getting without a refactor; I'd end up patching half the stdlib to get `main` as-is to be unit-testable.

scopatz:
Looks great! Thanks!

PR #522, opened 2015-11-21
Add support for --config-path option

### LINKED ISSUE #511: Feature-request: Static configuration file as a command line option
I suggest that an option to the commandline --config_file is added that allows the user to specify a file location for a static configuration file. (http://xonsh.org/xonshconfig.html)

This would allow portable usage of xonsh, e.g. install python and xonsh on a usb and bring the shell on a stick.

bpollack:
As with #520, I'm unclear what the best way to test this is, but really happy to amend this if it's heading a promising direction.

Fixes #511

### DIFF
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -2,17 +2,26 @@
 """The main xonsh script."""
 import os
 import sys
-import shlex
-import signal
 import builtins
-import subprocess
-from argparse import ArgumentParser, Namespace
+from argparse import ArgumentParser, ArgumentTypeError
 from contextlib import contextmanager
 
 from xonsh import __version__
 from xonsh.shell import Shell
 from xonsh.pretty import pprint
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks for submitting this @bpollack! I think that there is a little confusion here though.  The config path should affect the static configuration as given by `$XONSHCONFIG`, not the run control as given by `$XONSHRC`. The function it should primarily address is [`environ.load_static_config()`](https://github.com/scopatz/xonsh/blob/master/xonsh/environ.py#L610).

I think that it is sufficient to just modify `$XONSHCONFIG` in main.py. This should be a few lines at most.

bpollack:
@scopatz That makes way more sense; I was not clear why the original was requested anyway. I'll redo this and resend it.

bpollack:
@scopatz How about this?

scopatz:
Looks good to me, thanks!

PR #524, opened 2015-11-21
Fix use-before-define error in "make test"

bpollack:
In the case that a test was edited, "make test" would erroneously
attempt to add `test_fname` to the list of files to test...which doesn't
exist in that branch. Instead, it should be adding `edited_fname` in
that case.


scopatz:
Thanks!

Issue #391, opened 2015-09-16
Tab completion for alias


rgaiacs:
I use

~~~
alias g="git"
~~~

at my `~/.bashrc` and

~~~
[alias]
        ck = checkout
~~~

at my `~/.gitconfig`.

With `zsh`'s tab completion I have

~~~
$ g ck
HEAD    master
~~~

but with `xonsh`'s tab completion I have

~~~
$ g ck 
../  ./   bar
~~~

Could tab completion be add for alias?

scopatz:
Yes, this should probably be something that is added.  I don't think that it would be super hard.  Pull requests welcome :)

scopatz:
I think it is a great idea, btw.

rgaiacs:
> I don't think that it would be super hard. Pull requests welcome :)

I will give a try but will need 3-4 weeks.

scopatz:
That is fine, no huge rush from me :)

gforsyth:
Closing this as it has been implemented.  Please reopen if there are issues remaining.

PR #526, opened 2015-11-23
add alias expansion to completer

gforsyth:
Adds a function to expand aliases in the completer to allow completion
for user-defined aliases.

e.g.
```Bash
aliases['g'] = 'git'
aliases['ck'] = 'checkout'
```
then `'g ck' TAB` will show available branches for checkout.

This is currently restricted to single item aliases.

I think this helps satisfy #391 but it may not work exactly as expected -- I'm not sure if the `aliases` dict pulls in aliases from `~/.gitconfig`

Other questions:

Is there a use case for expanding multiple word aliases?  That would be relatively easy to add in.

Should this be optional?  It does have to run the expansion every time the completer is called though it is quite fast.  A failed search through a 500 key-pair dictionary takes on the order of 1e-8 seconds.

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -78,22 +78,6 @@ def _normpath(p):
 
     return p
 
-def alias_expand(line):
-    """ Expands any aliases present in line if alias does not point to a builtin function and if alias is only a single command.
-    """
-    
-    for word in line.split(' '):
-        if (word in builtins.aliases
-            and type(builtins.aliases.get(word)) is list
-            and len(builtins.aliases.get(word)) == 1):
-            word_idx = line.find(word)
-            line = line[:word_idx] +\
-                    builtins.aliases.get(word)[0] +\
-                    line[word_idx+len(word):]
-
-    return line
-
-
 class Completer(object):
     """This provides a list of optional completions for the xonsh shell."""
 
@@ -139,7 +123,7 @@ class Completer(object):
         dot = '.'
         ctx = ctx or {}
         prefixlow = prefix.lower()
-        line = alias_expand(line)
+        line = builtins.aliases.expand_alias(line)
         cmd = line.split(' ', 1)[0]
         if cmd in COMPLETION_SKIP_TOKENS:
             begidx -= len(cmd)+1
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -78,6 +78,21 @@ def _normpath(p):
 
     return p
 
+def alias_expand(line):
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks for putting this in @gforsyth!  To me the `aliases['ck'] = 'checkout'` should be unnecessary because that is handled by git.  Like in #391, if the following is added to `~/.gitconfig`

```ini
[alias]
        ck = checkout
```

then the master branch of xonsh will currently correctly complete `git ck` to all of the branch names.  So #391 should be about resolving the `g` to `git`, and not `ck` to `checkout`.  I'm gonna look at your code now before I make a fool of myself :)

xonsh/completer.py
```diff
@@ -78,6 +78,21 @@ def _normpath(p):
 
     return p
 
+def alias_expand(line):
+    """ Expands any aliases present in line if alias does not point to a builtin function and if alias is only a single command.
```
scopatz: PEP8 - line too long.

xonsh/completer.py
```diff
@@ -78,6 +78,21 @@ def _normpath(p):
 
     return p
 
+def alias_expand(line):
```
scopatz: This function seems like it would be better suited to being a method on the `built_ins.Aliases` class.  (Also the Aliases class should probably be moved to aliases.py, but that is another story.)

xonsh/completer.py
```diff
@@ -78,6 +78,21 @@ def _normpath(p):
...
+    """
+    
+    for word in line.split(' '):
+        if (word in builtins.aliases
+            and type(builtins.aliases.get(word)) is list
+            and len(builtins.aliases.get(word)) == 1):
+            word_idx = line.find(word)
+            line = line[:word_idx] +\
+                    builtins.aliases.get(word)[0] +\
+                    line[word_idx+len(word):]
```
scopatz: This seems to expand every token in the line.  We probably only want to expand the first token for now.

scopatz:
This looks like the right idea @gforsyth, subject to the changes I mentioned. And I don't think that this should be optional.  I can't think of a situation in which someone wouldn't want this.

xonsh/completer.py
scopatz: The reason for this is that you could have aliases defined in xonsh that end up not being valid because they are not defined elsewhere (like in git).

gforsyth:
Updated per your suggestions -- I didn't realize that the git aliases were handled by git, although it seems obvious in retrospect.  Should be good to go once the tests pass.

scopatz:
Thanks @gforsyth!

PR #528, opened 2015-11-25
Alias function exit status fixes

nicfit:
Currently the only way for an alias function to have non-zero exit status is to throw an exception, and based on XONSH_SHOW_TRACEBACK have a lot of extra output. Alias functions should be able to fail (exit status non-zero) and not ever expect to have a traceback displayed, this PR addresses this.

1. Functions that return stderr output have their exit status set to 1
2. To have the same affect as #1 but in with an Exception instead of `return None, '....'` The exception also allows for setting a return code other than 1, which is necessary in some apps.


nicfit:
And I don't know why all the commits on my other branches show up... it is like 2/3 commits max.

scopatz:
Have you merged from master recently?

nicfit:
@scopatz haha, obsessively.

scopatz:
> @scopatz haha, obsessively.

hah!

scopatz:
Thanks for putting this in @nicfit!

> Functions that return stderr output have their exit status set to 1

Sadly, I think that it is a bit more complicated that that, even though you are right that having the ability to return a return code from aliases would be ideal.  

The problem stems from the fact that stderr is kind of a misnomer. It really should be called something like stdmsg (standard message).   While it is conceived of as a place for error messages to go, the "error" part of this is by convention only and is independent of the return code. The real purpose is to be ["The usual destination is the text terminal which started the program to provide the best chance of being seen even if standard output is redirected (so not readily observed)."](https://en.wikipedia.org/wiki/Standard_streams).  That is, say you write `ls *.txt *.py | sort | grep`, but there are no python files but there are text files, you'll still see the error message because only stdout is subject to the pipe. And similarly for other forms of redirection.  

For example, consider `curl url > file.txt`.  This will download the file, store the contents in file.txt (stdout), and display download statistics (stderr), and have an exit code of zero assuming everything went OK.  

So I think it is not entirely correct to convolve non-empty stderr with non-zero exit codes.  Instead, the return code needs to be given separately.  I think that changing the convention for callable aliases to allow them return a three tuple of (stdout, stderr, rtn) would be the best thing to do. If a return code is not given in this way and there was no exception, assuming zero for the return code seems like the right thing to do.

Does this make sense?

nicfit:
Ha, yea. It makes total sense. A 3-tuple will would work, but I think the Exception is a bit friendlier to use.

scopatz:
Any reason to have closed this @nicfit?

nicfit:
PR #529 is the simple version discussed above, but also with a cleaner commit history. This commit list here is pollution, and likely hg-git cruft/bugs.

PR #530, opened 2015-11-27
Add support for $HISTCONTROL ignoredups and ignoreerr options.

nicfit:


docs/envvars.rst
```diff
@@ -74,6 +74,13 @@ applicable.
       - xonsh.environ.FORMATTER_DICT  
       - Dictionary containing variables to be used when formatting PROMPT and TITLE 
         see `Customizing the Prompt <tutorial.html#customizing-the-prompt>`_.
+    * - HISTCONTROL
+      - ``''``
+      - A comma-separated list of options that determine if commands are saved
```
scopatz: This should probably be a set then, not a string. (I assume that it doesn't make sense for options to be repeated.) You'll most likely want to write your own Environment validators and to- and from-string conversion functions.

scopatz: There are a bunch of examples in environ.py and tools.py, but let me know if you have any questions.

nicfit:
This might be good to go.

xonsh/history.py
```diff
@@ -290,6 +290,15 @@ def append(self, cmd):
         hf : HistoryFlusher or None
             The thread that was spawned to flush history
         """
+        opts = builtins.__xonsh_env__.get('HISTCONTROL', set())
```
scopatz: There is no need to provide a default value here as it will override the default value in `xonsh.environ.DEFAULT_VALUES`

xonsh/history.py
```diff
@@ -290,6 +290,15 @@ def append(self, cmd):
...
+        if ('ignoredups' in opts and len(self.buffer)
```
scopatz: Can you make the comparison explicit here, ie `len(self.buffer) > 0`, please?  Also, why are you testing the length of the buffer rather than the length of history?  So shouldn't this be `len(self) > 0`?

xonsh/history.py
```diff
@@ -290,6 +290,15 @@ def append(self, cmd):
...
+                and cmd['inp'] == self.buffer[-1]['inp']):
```
scopatz: Again, shouldn't this be `self.inps[-1]`

docs/envvars.rst
```diff
@@ -74,6 +74,14 @@ applicable.
       - xonsh.environ.FORMATTER_DICT  
       - Dictionary containing variables to be used when formatting PROMPT and TITLE 
         see `Customizing the Prompt <tutorial.html#customizing-the-prompt>`_.
+    * - HISTCONTROL
+      - ``''``
```
scopatz: I think that this default value should be updated to be an empty set, rather than an empty string.

scopatz:
Thanks a ton @nicfit - just a couple minor things here and there.  The overall structure looks great though!

nicfit:
Thanks for the feedback @scopatz, code updated.

scopatz:
Cool, Thanks a lot!

PR #531, opened 2015-11-27
Highlight (python) code when using prompt_toolkit

Carreau:
rely on Pygments Python Lexer for now,
One need to write a XonshLexer, but lexers based on the current one dont'
work as they don't return whitespace.

screen shot 2015-11-27 at 10 07 51
--- 

Also cleanup unused import in lexer. I've tried to write a simple PygmentsLexer around lexer:get_tokens but the python tokenizer does not yield whitespace, which make it annoying. 

I suppose it would be possible to wrap the tokenizer, and in case of "missing" token, fill in the [literally] blanks.

We can also lookup the first token and see if it's the list of known executable and highlight it differently:

screen shot 2015-11-27 at 10 17 46

But that's more work.

### DIFF
--- a/xonsh/prompt_toolkit_shell.py
+++ b/xonsh/prompt_toolkit_shell.py
@@ -17,8 +17,7 @@ from xonsh.tools import format_prompt_for_prompt_toolkit
 from xonsh.prompt_toolkit_completer import PromptToolkitCompleter
 from xonsh.prompt_toolkit_history import LimitedFileHistory
 from xonsh.prompt_toolkit_key_bindings import load_xonsh_bindings
-
-from .pyghooks import XonshLexer
+from xonsh.pyghooks import XonshLexer
 
 
 def setup_history():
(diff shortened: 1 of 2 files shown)


scopatz:
Awesome work @Carreau.  Did you know that we have xonsh pygments lexer that is installed as part of xonsh?  It lives in pyghooks.py (and could probably use more work).  In any event, it is probably best to use that one rather than the Python lexer.  Would you updating this PR?

Carreau:
> Did you know that we have xonsh pygments lexer that is installed as part of xonsh? It lives in pyghooks.py

Doh , I need to improve my grep-foo.

Yes. PR updated.

xonsh/prompt_toolkit_shell.py
```diff
@@ -17,6 +18,8 @@
 from xonsh.prompt_toolkit_history import LimitedFileHistory
 from xonsh.prompt_toolkit_key_bindings import load_xonsh_bindings
 
+from .pyghooks import XonshLexer
```
scopatz: Can you use xonsh.pyghooks here please? Sorry for being pedantic.

scopatz:
This looks good to go, just the one minor change!

Carreau:
> Can you use xonsh.pyghooks here please?

Sure, habits of relative imports.

> Sorry for being pedantic.

No problem.

Carreau:
(and done)

scopatz:
Woohoo! Thanks @Carreau!

PR #535, opened 2015-11-28
[Proof Of concept] Highlight first token in blue if know command

Carreau:
Mostly playing around, just prototype (as it scan the disk at each keystroke) 

When using prompt-toolkit, if the first command is a known executable, highlight it in blueish.

See `git` and `hub` blue in screenshot, `foo` is not.

screen shot 2015-11-28 at 22 40 41

if it does not match, first token could be highlighted in red (to avoid typo earlier, like fish does) but we would need to inspect python's user namespace also first for assigned variable, and do a bit of look ahead to not paint red on assignment.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -13,16 +13,13 @@ import atexit
 import signal
 import inspect
 import builtins
-import subprocess
-from io import TextIOWrapper, StringIO
 from glob import glob, iglob
 from subprocess import Popen, PIPE, STDOUT
 from contextlib import contextmanager
-from collections import Sequence, MutableMapping, Iterable, namedtuple, \
-    MutableSequence, MutableSet
+from collections import Sequence, MutableMapping, Iterable
 
 from xonsh.tools import suggest_commands, XonshError, ON_POSIX, ON_WINDOWS, \
-    string_types
+    string_types, _compute_known_commands
 from xonsh.inspectors import Inspector
 from xonsh.environ import Env, default_env, locate_binary
 from xonsh.aliases import DEFAULT_ALIASES
@@ -472,6 +469,9 @@ def _redirect_io(streams, r, loc=None):
     else:
         raise XonshError('Unrecognized redirection command: {}'.format(r))
 
+def known_commands():
+    return _compute_known_commands(ENV, builtins.aliases)
+
 
 def run_subproc(cmds, captured=True):
     """Runs a subprocess, in its many forms. This takes a list of 'commands,'
@@ -493,7 +493,6 @@ def run_subproc(cmds, captured=True):
         cmds = cmds[:-1]
     write_target = None
     last_cmd = len(cmds) - 1
-    prev = None
     procs = []
     prev_proc = None
     for ix, cmd in enumerate(cmds):
@@ -501,7 +500,6 @@ def run_subproc(cmds, captured=True):
         stdout = None
         stderr = None
         if isinstance(cmd, string_types):
-            prev = cmd
             continue
         streams = {}
         while True:
@@ -599,7 +597,6 @@ def run_subproc(cmds, captured=True):
                     e += '\n' + suggest_commands(cmd, ENV, builtins.aliases)
                 raise XonshError(e)
         procs.append(proc)
-        prev = None
         prev_proc = proc
     for proc in procs[:-1]:
         try:
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -19,7 +19,7 @@ from contextlib import contextmanager
 from collections import Sequence, MutableMapping, Iterable
 
 from xonsh.tools import suggest_commands, XonshError, ON_POSIX, ON_WINDOWS, \
-    string_types, _compute_known_commands
+    string_types
 from xonsh.inspectors import Inspector
 from xonsh.environ import Env, default_env, locate_binary
 from xonsh.aliases import DEFAULT_ALIASES
@@ -468,9 +468,6 @@ def _redirect_io(streams, r, loc=None):
 
     else:
         raise XonshError('Unrecognized redirection command: {}'.format(r))
-
-def known_commands():
-    return _compute_known_commands(ENV, builtins.aliases)
 
 
 def run_subproc(cmds, captured=True):
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -86,7 +86,9 @@ class KnownCommands(object):
         self._path_mtime = -1
         self._cmds_cache = frozenset()
     
-    def _all_commands(self):
+    def all_commands(self):
+        """find all commands available in PATH
+        """
         path = builtins.__xonsh_env__.get('PATH', [])
         # did PATH change?
         path_hash = hash(tuple(path))
@@ -451,7 +453,7 @@ class Completer(object):
         return attrs
 
     def _all_commands(self):
-        return self._known_commands._all_commands()
+        return self._known_commands.all_commands()
 
 
 
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -78,16 +78,52 @@ def _normpath(p):
 
     return p
 
-class Completer(object):
-    """This provides a list of optional completions for the xonsh shell."""
+class KnownCommands(object):
 
     def __init__(self):
-        # initialize command cache
         self._path_checksum = None
         self._alias_checksum = None
         self._path_mtime = -1
         self._cmds_cache = frozenset()
+    
+    def _all_commands(self):
+        path = builtins.__xonsh_env__.get('PATH', [])
+        # did PATH change?
+        path_hash = hash(tuple(path))
+        cache_valid = path_hash == self._path_checksum
+        self._path_checksum = path_hash
+        # did aliases change?
+        al_hash = hash(tuple(sorted(builtins.aliases.keys())))
+        cache_valid = cache_valid and al_hash == self._alias_checksum
+        self._alias_checksum = al_hash
+        pm = self._path_mtime
+        # did the contents of any directory in PATH change?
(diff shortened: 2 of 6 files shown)


xonsh/pyghooks.py
```diff
@@ -44,6 +44,24 @@ class XonshLexer(PythonLexer):
         'subproc': SUBPROC_TOKENS,
     }
 
+    def get_tokens_unprocessed(self, text):
+        super_iter =  super().get_tokens_unprocessed(text)
+        # import pdb; pdb.set_trace()
+        #if not super_iter:
+        #    return super_iter
+        first=  next(super_iter)
```
scopatz: weird spacing around `=`

xonsh/pyghooks.py
```diff
@@ -44,6 +44,24 @@ class XonshLexer(PythonLexer):
         'subproc': SUBPROC_TOKENS,
     }
 
+    def get_tokens_unprocessed(self, text):
+        super_iter =  super().get_tokens_unprocessed(text)
+        # import pdb; pdb.set_trace()
+        #if not super_iter:
+        #    return super_iter
```
scopatz: please remove comments

xonsh/built_ins.py
```diff
@@ -493,7 +493,6 @@ def run_subproc(cmds, captured=True):
         cmds = cmds[:-1]
     write_target = None
     last_cmd = len(cmds) - 1
-    prev = None
```
scopatz: Are the prev removals just cleanup?

xonsh/built_ins.py
```diff
@@ -472,6 +469,9 @@ def _redirect_io(streams, r, loc=None):
     else:
         raise XonshError('Unrecognized redirection command: {}'.format(r))
 
+def known_commands():
+    return _compute_known_commands(ENV, builtins.aliases)
```
scopatz: Probably better to use `builtns.__xonsh_env__` than `ENV`.  `ENV` should really go away...

scopatz:
Thanks for putting this in @Carreau!  Other than the minor style issues, is there are reason that this is "proof of concept"? I wouldn't mind merging this in.

Carreau:
> is there are reason that this is "proof of concept"

Beyond "it scan the disk at each keystroke" ? it does a scandir on all the directory on PATH every time you change a single char on your line...? That seem like a huge blocker to me :-) I need to add a cache.

And yes my proofs of concepts are usually not cleaned, just to inform of what I am working on. So I'll cleanup, document, fix pep 8, (likely test) ... etc.

adqm:
The `Completer` class has [something](https://github.com/scopatz/xonsh/blob/master/xonsh/completer.py#L417-L442) that might be useful re: caching.

Maybe we should move that out of that class and keep a global cache of available commands, instead of having it live in the completer.  Then both the completer and this (and whatever else we dream up) could rely on it.

xonsh/pyghooks.py
```diff
@@ -44,6 +44,24 @@ class XonshLexer(PythonLexer):
...
+        first=  next(super_iter)
+        from pygments.token import Token
+        from xonsh.built_ins import known_commands
+        if first:
+            if(first[1] == Token.Name) and (first[2] in known_commands()):
```
bpollack: Weird spacing here, too; you'd usually have a space between `if` and the `(`. (And you don't actually need the parentheses here—the operator precedence rules will do what you want—but having them for clarity makes sense.)

xonsh/pyghooks.py
```diff
@@ -44,6 +44,24 @@ class XonshLexer(PythonLexer):
...
+                yield first[0], Token.Name.KnowExecuteble, first[2]
+            else :
```
bpollack: Space before `:`

xonsh/tools.py
```diff
@@ -343,6 +343,16 @@ def command_not_found(cmd):
     return s
 
 
+def _compute_known_commands(env, aliases):
+    list_ = []
```
bpollack: A more common pattern for this would be `list_ = list(aliases)` (Python 2+) or `list_ = [*aliases]` (Python 3.5+).

xonsh/tools.py
```diff
@@ -343,6 +343,16 @@ def command_not_found(cmd):
...
+    list_.extend(aliases)
+    for d in filter(os.path.isdir, env.get('PATH')):
+        for f in os.scandir(d):
+            list_.append(f.name)
+    return list_
+
```
bpollack: Too many blank lines.

bpollack:
One day (not high priority), we should probably pull in something like Watchdog (https://pythonhosted.org/watchdog/) so we can be more intelligent than bash/fish on updating the autocomplete dicts. Alternatively, since autocomplete dicts are most likely going to alter in response to certain commands (`apt-get upgrade`, `brew upgrade`, etc.), maybe we could just preemptively update in response to those.

This looks awesome though. Looking forward to something like it landing. :)

scopatz:
Thanks for the tip @bpollack. Also @Carreau, it looks like there are now merge conflicts :(

Carreau:
No problem, I'l rebase and work on that later, I have a few other things on my plate for the next few days.

scopatz:
Ok no worries!

Carreau:
> Also @Carreau, it looks like there are now merge conflicts :(

No problem, I'm fluent in git. Rebased and cleaned up a bit, if people want to try. 

I'll be happy to get your feedback on **naming** it's not my strong point. 
I also want to make the "known commands cache" a subset of set, to make the checking of an item less verbose and more abstract. 

And yes I now  still not pep-8 compliant, and no tests...

xonsh/completer.py
```diff
@@ -78,16 +78,54 @@ def _normpath(p):
...
+    def all_commands(self):
+        """find all commands available in PATH
+        """
+        path = builtins.__xonsh_env__.get('PATH', [])
+        # did PATH change?
+        path_hash = hash(tuple(path))
+        cache_valid = path_hash == self._path_checksum
+        self._path_checksum = path_hash
+        # did aliases change?
+        al_hash = hash(tuple(sorted(builtins.aliases.keys())))
```
adqm: While we're making changes here, I think we can get a speedup here if, instead of sorting, tuple-izing, and hashing, we hash a frozenset of the keys instead:

That is, if:
```
hash(tuple(sorted(builtins.aliases.keys())))
```
becomes
```
hash(frozenset(builtins.aliases.keys()))
```

Not sure just how significant the speedup will be (quick hacky tests, I get about a factor of 2 on my machine).

xonsh/completer.py
```diff
@@ -78,16 +78,54 @@ def _normpath(p):
...
         self._alias_checksum = None
         self._path_mtime = -1
         self._cmds_cache = frozenset()
+    
+    def all_commands(self):
+        """find all commands available in PATH
+        """
+        path = builtins.__xonsh_env__.get('PATH', [])
+        # did PATH change?
+        path_hash = hash(tuple(path))
```
adqm: Similar to my comment ~4 lines down, I think we are likely to get a speedup here as well (as well as better cache validation) with:

``path_hash = hash(frozenset(path))``

xonsh/completer.py
```diff
@@ -78,16 +78,54 @@ def _normpath(p):
...
+        return self._cmds_cache
+
+class Completer(object):
+    """This provides a list of optional completions for the xonsh shell."""
+
+    def __init__(self, known_commands=None):
+        # initialize command cache
         self._man_completer = ManCompleter()
+
+        if not known_commands:
```
adqm: Is it maybe better to test explicitly whether `known_commands is None` instead of checking its truthiness?

xonsh/completer.py
scopatz: Me culpa here, but Isn't possible for different instances of the frozenset with the same values to hash differently? Or am I being overly paranoid here?

xonsh/completer.py
scopatz: agreed.

xonsh/completer.py
```diff
@@ -78,16 +78,54 @@ def _normpath(p):
 
     return p
 
-class Completer(object):
-    """This provides a list of optional completions for the xonsh shell."""
+class KnownCommands(object):
```
scopatz: This could be `KnownCommandsCache` or `CommandsCache` or similar, but I don't think it matters much.

xonsh/completer.py
adqm: > Me culpa here

Looking at git blame, I think this was me :)  Or maybe we can share the blame.

> Isn't possible for different instances of the frozenset with the same values to hash differently?

Nope. The hash of a frozenset only depends on the objects contained in it (and not on their order or the particular instance we're dealing with).  Source code for cpython's frozenset hashing function is [here](https://hg.python.org/cpython/file/tip/Objects/setobject.c#l761).


We could maybe run into trouble if someone used a mutable type as a key in the aliases dict, but that would cause this to break even if we stuck with hashing the sorted tuple.

xonsh/completer.py
adqm: I agree with adding "cache" to the the name of the class (more descriptive).

adqm: Also, it might make sense to move this class out of `completer.py` and into, e.g., `tools.py`, since it is being used multiple places (shell, completer, pygments lexer all use it).  But that's maybe a smaller issue.

xonsh/completer.py
scopatz: Ahh ok, I think I see.  Because the frozenset is frozen, its hash table is known at init and will be determinitsically created - no potential for later hash collisions causing alternative hash-chains.

xonsh/completer.py
scopatz: Also agree that should live somewhere else.  tools.py would be fine, but it could also live in its own module.

PR #540, opened 2015-11-29
prevent completer from launching executables

adqm:
I believe this should fix the issue in #364.

Since we only want attribute completion for Python objects, I think we can get away with using `eval` instead of `evalx`.

It also kind of bothered me that something like `7.<tab>` gave suggestions that are actually not valid Python, so I hacked in something to take care of that as well.  Maybe it's too hacky?

Really, I think it would be nice to avoid doing the `eval` at all (since, depending on what we are evaling, it might take time).  I had a version that just checked for `expr` as a key in `ctx` and `builtins.__dict__` and did completions based on that.  No `eval` involved, but it doesn't quite get us to the same place (e.g., if `c` is a number, we would lose `c.imag.<tab>`).  Maybe that's better anyway, though...

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -397,7 +397,7 @@ class Completer(object):
                 opts.append(i)
             except:
                 continue
-        if len(attrs) == 0:
+        if len(attr) == 0:
             opts = [o for o in opts if not o.startswith('_')]
         else:
             csc = builtins.__xonsh_env__.get('CASE_SENSITIVE_COMPLETIONS')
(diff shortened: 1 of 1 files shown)


adqm:
With this, we lose completion of things like `$(ls -l).<tab>`, but to me, this is still a net win.

scopatz:
Thanks a ton @adqm!  I can't even tab complete with `$(ls -l).<tab>` right now, so there is no loss :).  Also evaluating to tab complete on that would be terrible.  Also also, if we want that, we know that $() returns a string, so the completer could take advantage of that eventually.

scopatz:
>  No eval involved, but it doesn't quite get us to the same place (e.g., if c is a number, we would lose c.imag.<tab>). Maybe that's better anyway, though...

I use things like `c.imag.<tab>` all the time.  I can see the argument for this being a user-configurable behaviour if it really bothers people to eval objects in this way.

scopatz:
Thanks again!

Issue #536, opened 2015-11-29
dark blue against grey black


stonebig:
On my 'not high quality' desktop screen, dark blue is weakly distinguishible over grey black.

Would it be possible to change to a `better` color ? light blue, light orange,  light yellow, light pink, ... whatever light

![xonsh_blues](https://cloud.githubusercontent.com/assets/4312421/11456060/3e3b6b0e-967b-11e5-9973-d74134666dcc.GIF)

scopatz:
Yes, this is definitely needed! Thanks for reporting @stonebig. I had thought that there was an open issue for this already, but I was wrong...

scopatz:
@stonebig please see if #541 fixes your problem.

melund:
@stonebig Put 
`$PROMPT_TOOLKIT_COLORS = {'BLUE': '#00d8ff'}` 
in your .xonshrc file. Cmd.exe only supports a few different colors. So even though #00d8ff is nice light blue it becomes kind of turquise. But atleast it is readable. 

If you install something like [cmder](http://cmder.net/) then xonsh looks much nicer.

stonebig:
cmder is a little too big to my taste: 4 Mo for a console.

melund:
I am not sure I follow. ConEmu upon which it is based is a little confusing when it comes to configuration. But 4Mb doesn't seem big to me.

stonebig:
1Mo is the size I would have expected, comparing to SciTE.exe.

stonebig:
with @melund suggestion I get this, which is better, just now I feel the green is also too dark.

![xonsh_green](https://cloud.githubusercontent.com/assets/4312421/11484214/167ff98a-97ad-11e5-99f6-8daa1b5afb3b.GIF)

adqm:
What works for me to force `cmd.exe` to use its "light" blue and "light" green colors is:
```
$PROMPT_TOOLKIT_COLORS = {'BLUE': '#1E90FF', 'GREEN': '#00FF18'}
```

For me, that makes both come out readable for me.  Can you give that a shot?

May be that we'll want to adjust the default values based on platform (or just use these as the defaults everywhere).  Or, really, maybe we should have two sets of defaults: one for white backgrounds, and one for "reverse video," since the colors I just posted don't look great on a white background.

scopatz:
> Or, really, maybe we should have two sets of defaults: one for white backgrounds, and one for "reverse video,

[...]

PR #541, opened 2015-11-29
allow use of custom color mappings with prompt_toolkit

adqm:
This is one workaround for the problem mentioned in #536 when using prompt_toolkit.

In particular, this allows specifying a custom mapping from names to colors via a new environment variable `$PROMPT_TOOLKIT_COLORS`.

The following, for example, makes the prompt_toolkit prompt much more readable for me (black background):
```
$PROMPT_TOOLKIT_COLORS = {'BLUE': '#1E90FF', 'GREEN': '#00CD00'}
```


xonsh/tools.py
```diff
@@ -729,9 +729,13 @@ def _make_style(color_name):
     for k, v in _PT_STYLE.items():
         if k in color_name:
             style.append(v)
-    for k, v in _PT_COLORS.items():
+    _custom_colors = builtins.__xonsh_env__.get('PROMPT_TOOLKIT_COLORS', {})
```
scopatz: No need to provide default values here, instead they should be set in the `xonsh.envrion.DEFAULT_VALUES` dict.

scopatz:
Hi @adqm! Thanks for putting this in.  I think that this might overlap slightly in functionality with the `PROMPT_TOOLKIT_STYLES` variable, but there is precious little documentation on that - even in prompt-toolkit - and so I think that this is more intuitive.  

My main comment before merging is that the Env class received an overhaul to allow for default values.  Thus `PROMPT_TOOLKIT_COLORS` should probably use that as well as add an entry to `docs/envvars.rst`.

Also, do you think that it would be valuable to just change the default colors to match the readline shell?  I think that consistency would be nice.  

Thanks again!

adqm:
> I think that this might overlap slightly in functionality with the PROMPT_TOOLKIT_STYLES variable, but there is precious little documentation on that - even in prompt-toolkit - and so I think that this is more intuitive. 

Let me see if there is an easy way to reconcile the two.

> Also, do you think that it would be valuable to just change the default colors to match the readline shell? I think that consistency would be nice. 

I think it would be valuable, but I'm not sure how to do that right now.  IINM, the terminal emulator is usually the one responsible for deciding which actual color to show for, for the different ANSI colors, but I don't think prompt_toolkit has a way to just pass those through.  I can monkey around a bit and see if I can get prompt_toolkit to use the ANSI colors (and let the terminal decide how to display them) rather than using the HTML color codes.

> My main comment before merging is that the Env class received an overhaul to allow for default values. Thus PROMPT_TOOLKIT_COLORS should probably use that as well as add an entry to docs/envvars.rst.

I can certainly make that change, as well.

scopatz:
Thanks. Also, don't worry too much about reconciling the two, because I think if you just want to change the color scheme, the style variable is non-obvious.

adqm:
Okay, assuming this passes tests:
* I updated to use the new `Env` class and moved the default value to the appropriate spot
* I modified the default pygments style to use `PROMPT_TOOLKIT_COLORS` so that users who just want to change the color definitions don't have to go through `PROMPT_TOOLKIT_STYLES`, which can be a pain.  This is enough of a reconciliation for me, at least for now (the prompt and the syntax highlighting use the same color scheme).

I haven't yet found a way to make it use ANSI colors and let the terminal itself do the coloring, and I'm not sure that's easy/possible.

xonsh/prompt_toolkit_shell.py
```diff
@@ -110,21 +111,66 @@ def _get_prompt_tokens_and_style(self):
...
-            if userstyle is not None:
-                styles.update(userstyle)
-
-        return get_tokens, CustomStyle
+        custom_style = _xonsh_style(tokens, cstyles)
+
+        return get_tokens, custom_style
+
+
+def _xonsh_style(tokens=[], cstyles=[]):
```
scopatz: Keyword argument values should not be mutable, unless this is on purpose (which it doesn't seem to be).  Could you replace with tuples, please.

adqm: Yep, that seems sensible.  Done.

scopatz:
Thanks for tackling this @adqm! It was a long time in coming...

adqm:
Happy to help.  I'm hoping to get back into xonsh development after a long hiatus :)

scopatz:
It is great to work with you again! I think you are still the highest contributor (after me) so the hiatus didn't affect you status :)

melund:
Hi @adqm 

I think I originally made `$PROMPT_TOOLKIT_STYLES` for that exact same reason. It just allows the user to change all the pygments styles and not only the colors. 

The TERM_COLORS are automatically created as custom pygments tokens, thus to change a color BOLD_BLUE to something else just requires: 

$PROMPT_TOOLKIT_STYLES = {Token.BOLD_BLUE: 'bg:#00aaaa #000000'}

I am sorry if this was left undocumented....

adqm:
I understand that's why `$PROMPT_TOOLKIT_STYLES` exists, but it felt to me like it was a pain to update styles, and so I wanted an easier way if users _only_ wanted to change colors, while still having `$PROMPT_TOOLKIT_STYLES` for the more complicated changes.

Maybe I am missing something, though.

Are those tokens (e.g., `Token.BOLD_BLUE`) used for everything (i.e., syntax highlighting and the prompt both)?  Or would someone have to change both the `Token.BOLD_BLUE` and, e.g., the `Name.Class` entries to make both the prompt and the syntax highlighting use the same color?

Also, am I correct in assuming that, in order to change `$PROMPT_TOOLKIT_STYLES`, a user has to import `pygments.token.Token`?  That seems like a pain to me.

I realize that this may have added some extra bloat, but I think that's worth the (IMHO) easier interface.  Maybe there is some way to get the "best of both worlds," though?

melund:
I think the custom 'color' tokens are only used for the prompt. It was all I could do without touching the TERM_COLOR implementation.

Don't misundertand me. I really like your changes. The $PROMPT_TOOLKIT_STYLES implementation is too low-level for users. I guess that was also why I never got around to document it. Maybe we should leave it that way. 

However, for most windows users it may not work as expected. Cmd.exe only support handfull of different colors. That is also why the color is very dark blue for those users. The dark blue colors was just the closet match to what xonsh specified. I mostly use [cmder](http://cmder.net/) and it automaticaly adjust the colors to make them visible. So I haven't really had the problem. 

I think we should set the default blue in `_PT_COLORS` to `#0a86f5` That seems to give an acceptable 'turquise' color in Cmd.exe and it looks good in terminals that support all colors. 

CMD.exe
image

cmder
image

scopatz:
> I think we should set the default blue in _PT_COLORS to #00d8ff That seems to give an acceptable 'turquoise' color in Cmd.exe, and it looks good in terminals that support all colors.

Please PR :)

PR #542, opened 2015-11-29
modify twiddle replacement to happen only at the start of the path

adqm:
In the prompt and in the title, `$HOME` is currently replaced everywhere in the PWD, when it should really only be replaced at the very beginning.  This changeset implements this fix.  Not yet tested on Windows.

Examples:

Old:
```
hartz@compy-488 ~ $ cd backups/
hartz@compy-488 ~/backups $ cd home/
hartz@compy-488 ~/backups/home $ cd hartz/
hartz@compy-488 ~/backups~ $ 
```

New:
```
hartz@compy-488 ~ $ cd backups/
hartz@compy-488 ~/backups $ cd home/
hartz@compy-488 ~/backups/home $ cd hartz/
hartz@compy-488 ~/backups/home/hartz $ 
```

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -528,14 +528,18 @@ def _replace_home(x):
     if ON_WINDOWS:
         home = (builtins.__xonsh_env__['HOMEDRIVE'] +
                 builtins.__xonsh_env__['HOMEPATH'][0])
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks for the fix @adqm!

PR #543, opened 2015-11-29
fish-style shortened cwd as short_cwd

adqm:
Not sure whether this should just stay at home in my `.xonshrc` or whether it's something worth including...

I do a lot of work deep in directory trees, and so I found fish's shortening of the CWD in the prompt to be useful.  This changeset implements a shortened CWD as a new entry in the formatter dict, `short_cwd`.

If it's not useful (or too adqm-specific) to have this in the main codebase, I'm happy to close this and just leave it to live on my own machines, but I thought it might be useful.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -57,6 +57,7 @@ DEFAULT_ENSURERS = {
     'CASE_SENSITIVE_COMPLETIONS': (is_bool, to_bool, bool_to_str),
     re.compile('\w*DIRS'): (is_env_path, str_to_env_path, env_path_to_str),
     'COMPLETIONS_DISPLAY': (is_completions_display_value, to_completions_display_value, str),
+    'FORCE_POSIX_PATHS': (is_bool, to_bool, bool_to_str),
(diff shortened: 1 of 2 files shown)


xonsh/environ.py
```diff
@@ -539,6 +539,13 @@ def _replace_home(x):
 
 _replace_home_cwd = lambda: _replace_home(builtins.__xonsh_env__['PWD'])
 
+def _collapsed_pwd():
+    pwd = _replace_home_cwd().split('/')
```
scopatz: I think that `'/'` should be replaced with `os.sep` everywhere in this function.

scopatz:
Thanks @adqm - I think that if you find this useful, other people are bound to as well.  Also, could you add this as an entry to the tutorial docs so that folks know about it.

adqm:
Okay; tested this on GNU/Linux and on Windows (both with and without `$FORCE_POSIX_PATHS` set), so I think it is ready for another look.

scopatz:
Looks great to me, thanks @adqm!

PR #544, opened 2015-11-30
bug fix for cache validation in completer

adqm:
This should fix a silly bug in the cache validation code for the completer (whereby changes in the alias dictionary during runtime would not be noticed).


scopatz:
Whoops! Thanks!

PR #546, opened 2015-12-01
Better default colors for promt tk

melund:
This updates the colors for prompt_toolkit to work better on a black background. On windows, it also changes the default prompt to use CYAN instead of BLUE, which is more readable in cmd.exe. 

On cmd.exe  (windows 10)
image

In cmder (windows 10)
image

I also added a color mapping for colors that work best on light backgrounds. Just haven't made anything that uses it yet.


scopatz:
Looks great to me, thanks @melund

PR #549, opened 2015-12-01
enable string matching on history search

### LINKED ISSUE #548: Prompt toolkit up and down arrows don't match beginning of str
Currently the up and down arrows don't do reverse look up that matches on the start of the string. Rather, it only scrolls through history. This should have the same behavior a readline.

gforsyth:
Resolves #548


scopatz:
Thanks a ton!

PR #550, opened 2015-12-02
Logic bug in source-foreign alias 

scopatz:
This prevented prevcmd from being set implicitly from the command line.


beardedprojamz:
Works for me!

```
$ source-bash $(docker-machine env devbrite)
```

Leaves the docker env vars in my environment.

scopatz:
Thanks!

PR #552, opened 2015-12-02
Fix error when bash DEFAULT_BASH_FUNCSCMD is used on windows.

melund:
On Windows file path can have white spaces, which the bash code did not handle. Moved the logic out of bash and into python.

fixes: https://github.com/scopatz/xonsh/issues/545

### DIFF
--- a/xonsh/foreign_shells.py
+++ b/xonsh/foreign_shells.py
@@ -43,18 +43,7 @@ done
 shopt -s extdebug
 namelocfilestr=$(declare -F $funcnames)
 shopt -u extdebug
-
-# print just name and file
-read -r -a namelocfile <<< $namelocfilestr
-namefile=""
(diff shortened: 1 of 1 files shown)


melund:
I just read funcscmd (http://xonsh.org/xonshconfig.html):
> funcscmd:
> str or None, optional - This is a command or script that can be used to determine the names and locations of any functions that are native to the foreign shell. This command should print only a whitespace separated sequence of pairs function name & filenames where the functions are defined. If this is None (null), then a default script will attempted to be looked up based on the shell name. Callable wrappers for these functions will be returned in the aliases dictionary. default=null

So now I see why my change doesn't cut it. It basically breaks the interface. I can't see how it can easily made to work without doing the magic in bash. It may not be a good a idea that commands are whitespace separated, if it should work with files names that have whitespaces.

PR #555, opened 2015-12-03
Update conda recipes

melund:
Added prompt_toolkit as conda depedency, and removed the app entry in meta.yaml, since this feature is not very mature. Instead added a menu item entry to start xonsh.


recipe/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+$PYTHON setup.py install --conda
+
+cp ${RECIPE_DIR}/xonsh_shortcut.json ${PREFIX}/Menu/xonsh_shortcut.json
+cp ${RECIPE_DIR}/xonsh.ico ${PREFIX}/Menu/xonsh.ico
```
scopatz: The `$PREFIX/Menu` directory does not exist on my system. I'm not sure that this is the right thing to do.  You may need to create such a directory first (`mkdir -p ${PREFIX}/Menu`).  Also, are you sure that it is `Menu` and not `menu`?

melund: Good point. I am not sure how it look on a linux anaconda installation. I got the idea from the spyder packages. I will check how it looks on for the corresponding linux packages.

melund: Maybe it is a windows feature....

scopatz: I think that it is... It is probably safest to remove.

melund: Seems that the linux package for spyder has a diffent folder structure:
share/
   applications/
    - spyder3.desktop
   pixmaps
    - spyder3

Hmm. All of this is probably an undocumented conda feature. Maybe something Continuum uses to put application launchers in the packages they build them self. Maybe @asmeurer  knows something about this. 

Anyway. I will remove the Menu entry. I don't see a way to only apply it on the Windows, since we use conda convert to build the packages.

scopatz: Sounds reasonable. We can always add this capability back in.

scopatz:
Is this ready now?

melund:
Yes :) You can merge. I am heading of to bed now. I couldn't see that the webhook was working. Anyways. We can always fix that.

Issue #557, opened 2015-12-04
cd doesn't complain when it should


kseistrup:
When cd'ing to an inaccessible directory, the builting cd command doesn't complain:

```sh
kseistrup@home ~ $ sudo install -m 0750 -d /tmp/inaccessible  # create a directory accessible to root only
kseistrup@home ~ $ cd /tmp/inaccessible  # here I'd like to see a “cd: /tmp/inaccessible/: Permission denied” or similar
kseistrup@home ~ $ $AUTO_CD=True
kseistrup@gome ~ $ /tmp/inaccessible  # ditto
kseistrup@home ~ $ pwd
/home/kseistrup
kseistrup@home ~ $ aliases['cd']
<function xonsh.dirstack.cd>
```

This is the latest git (AUR xonsh-git-0.2.3.r146.g6e98119-1) running on Python 3.5.

scopatz:
Thanks for reporting @kseistrup

gforsyth:
This is the current behavior on master.  Closing, but please reopen if the issue persists.  Thanks!

PR #558, opened 2015-12-04
error on trying to cd to directory without executable bit

adqm:
This adds a simple check to report an error if a user tries to `cd` into a directory they don't have permissions to enter, per #557.


xonsh/dirstack.py
```diff
@@ -88,6 +88,8 @@ def cd(args, stdin=None):
         return '', 'cd: no such file or directory: {0}\n'.format(d)
     if not os.path.isdir(d):
         return '', 'cd: {0} is not a directory\n'.format(d)
+    if not os.access(d, os.X_OK):
+        return '', 'cd: permission denied: {0}\n'.format(d)
```
scopatz: Just an idea, but now we have the option of returning a returncode as the third element of the tuple. Maybe you could add a `, 1` to the end of this line?

adqm:
Alright, I went in and added return codes to `cd`, as well as the other commands in `dirstack.py`.

Also fixed a bug with `dirs -c` (must not be a very widely-used command since it made it this long without being noticed :) )

scopatz:
Thanks a ton @adqm!  And yeah, I never use dirstack, but I am glad that it is around :)

PR #561, opened 2015-12-04
Bash Completer $XONSH_TOKEN wrapping

gforsyth:
This supercedes #560 

From #560 --
> On Linux, if there's a foldername with a space in it, like "bad folder"
> then tab-completion using `cd` will wrap single quotes around the name,
> e.g.
> 
> ```
> cd bad(TAB)
> ```
> 
> completes to
> 
> ```
> cd 'bad folder/'
> ```
> 
> This behavior is encoded in `path_complete` in `completer.py`.  But if
> you want to remove that poorly named folder, the argument to `rm` is
> completed by `bash_complete` which didn't have this tweak and so the
> result is
> 
> ```
> rm -r bad(TAB)
> ```
> leads to
> 
> ```
> rm -r bad folder
> /usr/bin/rm: cannot remove 'bad': No such file or directory
> /usr/bin/rm: cannot remove 'folder/': No such file or directory
> ```
> 
> This tweaks the `bash_complete` function so that
> 
> ```
> rm -r bad(TAB)
> ```
> 
> returns
> 
> ```
> rm -r 'bad folder/'
> ```
> 
> I don't think this is insane but it might(?) have unintended side-effects?  It would be great if @melund or another Windows user could try this out and make sure that the trailing slash gets treated correctly.

While working on this I went through a couple of very poorly named files, filenames that contained commas, square brackets, etc. and all of these have similar trouble because of the xonsh parser fighting with bash completion.  

This PR instead adds a tokenset that, if a token is found in the completed string, `Completer.bash_complete` will return the `repr` of the string instead and the wrapping single quotes will stop the xonsh parser from having trouble.  

For now, the problem tokens I've added in are 
```
' ',',','[',']','(',')','{','}'
```

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -316,7 +316,12 @@ class Completer(object):
             out = ''
 
         space = ' '
-        rtn = {s + space if s[-1:].isalnum() else s for s in out.splitlines()}
+        slash = '/'
+        rtn = {_normpath(repr(s + (slash if os.path.isdir(s) else '')))
+               if space in s else
+               s + space
+               if s[-1:].isalnum() else
+               s for s in out.splitlines()}
         return rtn
 
     def _source_completions(self):
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -30,6 +30,8 @@ XONSH_TOKENS = {
 
 COMPLETION_SKIP_TOKENS = {'sudo', 'time'}
 
+COMPLETION_WRAP_TOKENS = {' ',',','[',']','(',')','{','}'}
+
 BASH_COMPLETE_SCRIPT = """source {filename}
 COMP_WORDS=({line})
 COMP_LINE={comp_line}
@@ -318,7 +320,7 @@ class Completer(object):
         space = ' '
         slash = '/'
         rtn = {_normpath(repr(s + (slash if os.path.isdir(s) else '')))
-               if space in s else
+               if (COMPLETION_WRAP_TOKENS.intersection(s) != set()) else
                s + space
                if s[-1:].isalnum() else
                s for s in out.splitlines()}
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -9,7 +9,7 @@ import subprocess
 import sys
 
 from xonsh.built_ins import iglobpath
-from xonsh.tools import subexpr_from_unbalanced
+from xonsh.tools import subexpr_from_unbalanced, get_sep
 from xonsh.tools import ON_WINDOWS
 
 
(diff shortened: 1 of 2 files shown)


melund:
Except for https://github.com/scopatz/xonsh/issues/559 then it seems to work. 

But generally, folders with spaces causes all kinds of weird behaviour with the autocompleter. 

For example 
```
cd bad(tab)
```
gives
```
cd 'bad folder/'
```
But then you can't use the autocompleter to go further into the directory structure. 

The only sane way to handle it is how clink does for [cmder](http://cmder.net/) on windows. But it may be incompatible with Xonsh

In cmder: 
```
cd  ba(tab)
```
gives
```
cd "bad folder\
```
Pressing enter here without closing the quotes works fine (in xonsh it enter multi-line mode). Also can continue to explore down the directory tree by pressing tab or typing folder names. 

/Morten

xonsh/completer.py
```diff
@@ -316,7 +318,12 @@ def bash_complete(self, prefix, line, begidx, endidx):
             out = ''
 
         space = ' '
-        rtn = {s + space if s[-1:].isalnum() else s for s in out.splitlines()}
+        slash = '/'
+        rtn = {_normpath(repr(s + (slash if os.path.isdir(s) else '')))
+               if COMPLETION_WRAP_TOKENS.intersection(s) else
+               s + space
+               if s[-1:].isalnum() else
```
scopatz: Maybe this whole expr should just be its own function, an then `rtn = set(map(f, out.splitlines()))`

scopatz:
Is this ready for another look @gforsyth.  Sorry, I don't notice these things unless you ping the conversation.

gforsyth:
Yeah, ready for review @scopatz

xonsh/completer.py
```diff
@@ -78,6 +80,18 @@ def _normpath(p):
 
     return p
 
+def completionwrap(s):
+    """ Returns the repr of input string s if that string contains 
+    a 'problem' token that will confuse the xonsh parser
+    """
+    space = ' '
+    slash = '/'
```
scopatz: Would this be better as `os.sep`? Or better yet, 

```python
slash = os.altsep if ON_WINDOWS and builtins.__xonsh_env__.get('FORCE_POSIX_PATHS') else os.sep
```

adqm: I'm thinking it might be the right time to put that into a function in `tools.py` as a function (`xonsh.tools.get_sep()` or similar?), since it's used a few places now.  We can then go through and replace the hard-coded versions with calls to that function.

Not sure if that belongs in this PR or if it should happen separately.

scopatz: Either way @gforsyth wants it :)

gforsyth:
@scopatz @adqm -- I added `get_sep` to `tools.py` as suggested and use it to define `slash` in `completionwrap` but I haven't replaced it anywhere else in the code.

scopatz:
That is fine @gforsyth, thanks.

scopatz:
Thanks again!

Issue #519, opened 2015-11-21
Ctrl-Z on jobs does not work $XONSH_STORE_STDOUT = True


nicfit:
```bash
$ sleep 10
^Z^Z^Z^Z^Z^Z^Z^Z
````

scopatz:
Thanks for reporting @nicfit

redshodan:
Looks like the True value is coming from the ~/.xonshrc file. I did not create that file and must have been autocreated with $XONSH_STORE_STDOUT=True in it. Once I flip that setting, ^Z works fine.

scopatz:
> I did not create that file and must have been autocreated with $XONSH_STORE_STDOUT=True in it.

Hmm that is strange.  I don't believe that we autocreate that file.  Not sure where that would be coming from.

redshodan:
It's certainly possible I copied the example xonshrc from http://xonsh.org/xonshrc.html and forgot about that. My xonshrc is almost an exact duplicate, minus removing the LD_LIBRARY_PATH and changing the homedir. I've tried a couple of times to break my bash addiction and get into xonsh, with long periods of time inbetween. :) I didn't think I had it set and was trying to figure out why it kept coming up True. Nevermind, sorry to distract from the real problem.

scopatz:
> sorry to distract from the real problem.

No worries!

>  I've tried a couple of times to break my bash addiction and get into xonsh, with long periods of time inbetween. :)

If there are specific areas that you think could use help, please let us know!

PR #562, opened 2015-12-05
include the command being run in the default title

adqm:
This is a little modification (again, another idea borrowed from fish) to make the console title, by default, show the name of the command being run (the current active foreground job).  I often have a bunch of terminals open, and this makes it easier for me to switch between them.

Normal title on my machine is:
`hartz@compy-488: ~ | xonsh`

Opening, e.g., `python` makes it:
`python | hartz@compy-488: ~ | xonsh`

Programs like `vim` that set titles themselves are unaffected.

Also works with pipelines.  So opening `echo "hello" | gedit -` makes it:
`gedit | hartz@compy-488: ~ | xonsh`

Tested to work with `fg`, `bg`, and friends.

### DIFF
--- a/xonsh/base_shell.py
+++ b/xonsh/base_shell.py
@@ -203,6 +203,7 @@ class BaseShell(object):
             os.system('title {}'.format(t))
         else:
             sys.stdout.write("\x1b]2;{0}\x07".format(t))
+            sys.stdout.flush()
 
     @property
     def prompt(self):
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -613,7 +613,8 @@ def run_subproc(cmds, captured=True):
             'obj': prev_proc,
             'bg': background
         })
-    if ENV.get('XONSH_INTERACTIVE'):
+    if ENV.get('XONSH_INTERACTIVE') and not ENV.get('XONSH_STORE_STDOUT'):
+        # set title here to get current command running
         builtins.__xonsh_shell__.settitle()
     if background:
         return
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -613,6 +613,8 @@ def run_subproc(cmds, captured=True):
             'obj': prev_proc,
             'bg': background
         })
+    if ENV.get('XONSH_INTERACTIVE'):
+        builtins.__xonsh_shell__.settitle()
     if background:
         return
     if prev_is_proxy:
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -98,8 +98,8 @@ else:
     DEFAULT_PROMPT = ('{BOLD_GREEN}{user}@{hostname}{BOLD_BLUE} '
                       '{cwd}{branch_color}{curr_branch} '
                       '{BOLD_BLUE}{prompt_end}{NO_COLOR} ')
-                  
-DEFAULT_TITLE = '{user}@{hostname}: {cwd} | xonsh'
+
+DEFAULT_TITLE = '{current_job}{user}@{hostname}: {cwd} | xonsh'
 
 @default_value
 def xonsh_data_dir(env):
@@ -564,6 +564,15 @@ def _collapsed_pwd():
     return leader + sep.join(base)
 
 
+def _current_job():
+    j = builtins.__xonsh_active_job__
+    if j is not None:
+        j = builtins.__xonsh_all_jobs__[j]
+        if not j['bg']:
+            return '{} | '.format(j['cmds'][-1][0])
+    return ''
+
+
 if ON_WINDOWS:
     USER = 'USERNAME'
 else:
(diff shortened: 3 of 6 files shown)


xonsh/base_shell.py
```diff
@@ -203,6 +203,7 @@ def settitle(self):
             os.system('title {}'.format(t))
         else:
             sys.stdout.write("\x1b]2;{0}\x07".format(t))
+            sys.stdout.flush()
```
scopatz: surprised that wasn't there already.

xonsh/built_ins.py
```diff
@@ -613,6 +613,10 @@ def run_subproc(cmds, captured=True):
             'obj': prev_proc,
             'bg': background
         })
+    try:
+        builtins.__xonsh_shell__.settitle()
+    except AttributeError:
+        pass
```
scopatz: This block could be:

```python
if hasattr(builtins.__xonsh_shell__, 'settitle'):
    builtins.__xonsh_shell__.settitle()
```

rather than relying on exceptions.  Or, just remove any conditional and do a straight up `builtins.__xonsh_shell__.settitle()` and rely on duck typing.

xonsh/environ.py
```diff
@@ -580,6 +589,7 @@ def _collapsed_pwd():
     short_cwd=_collapsed_pwd,
     curr_branch=current_branch,
     branch_color=branch_color,
+    current_job=_current_job,
```
scopatz: Can you add this to the relevant `docs/tutorial.rst` section so that people know about it.

scopatz:
Looks great @adqm! Thanks a ton.  Just a few small issues here and there.  The big one is documentation.

xonsh/built_ins.py
adqm: Without the conditional (or the try/except), the `import` tests failed, so I stuck this in.

The problem is actually `builtins` not having `__xonsh_shell__` in that case, not `__xonsh_shell__` not having `settitle`.

I also tried setting a default (`None`) for `__xonsh_shell__` in `builtins.py`, but that didn't work, either (same error).  So maybe imports don't even initialize the builtins?  Not sure what was happening.  Or maybe it's more a quirk of `nosetests3` than anything else?

xonsh/environ.py
adqm: Can do.

xonsh/built_ins.py
scopatz: Ahh ok.  Yeah, that makes more sense.  Some of the tests mock the builtins initialization. I think that code is in `tests/tools.py`. As I recall, the `__xonsh_shell__` came later. So it might not be mocked.

adqm:
Alright, I think it's ready for another look.

I added a dummy shell to the mock environment in `tests/tools.py` so that the tests won't fail anymore, and I replaced the try/except with a check of `$XONSH_INTERACTIVE` (since it doesn't make much sense to set the title if we're not running interactively).

I also added a very short entry for `current_job` to the `$PROMPT` section of the tutorial.

scopatz:
This doesn't seem to work super well `$XONSH_STORE_STDOUT=True`.  The name of the program with flicker in the title at the *end* of the command.  Even something like `sleep 2` can demonstrate this.  Maybe there is another place to in `run_subproc()` to put the call to `settitle()`, or maybe we should be writing to the `prev_proc.stdout` in this case.

I am happy to merge it in as-is, because it does work when this variable is False, if you want to open up an another issue.

adqm:
Hmmm, that's interesting.  Let's hold off on merging, and I'll take another look.

scopatz:
Sounds good.  I am pretty sure that this is because TeePTY forks the process and thus changes stdin, stdout, and stderr.

adqm:
Okay, so `TeePTY` looks like it actually holds on to the main thread, rather than running in the background like `Popen` or `ProcProxy` (hence the wait time and title only flashing briefly).  I'm going to try to "fix up" `TeePTY` so that it behaves a bit more like those other two.  Failing that, we can maybe just merge in as it currently stands (which doesn't try to do that second `settitle` if we're capturing stdout).

scopatz:
> I'm going to try to "fix up" TeePTY so that it behaves a bit more like those other two. 

Because of the process fork, I think that TeePTY needs to run on the main thread.  That said there may be another way to fix it up.  Such a fix would probably also address #519.

scopatz:
Or might address #519.

adqm:
Okay, so since there is at least one other thing that is broken with `TeePTY`, my vote is to merge this in now (if it gets your seal of approval), and to tackle `TeePTY` in another PR (hopefully addressing the remaining issue here, as well as #519).

scopatz:
OK. Sounds good.

scopatz:
Thanks @adqm!

Issue #409, opened 2015-10-08
Can't pipe to grep


blarghmatey:
When I try to pipe the output of a command into grep I get no output even when there should be some. A simple example is when I run `ps aux | grep emacs` while there is an emacs process running. Performing the same operation in bash or zsh yields the expected process information.

I am using Xonsh on CPython 3.5 in Manjaro Linux on a 64 bit Linux 3.19 kernel.

scopatz:
Thanks for reporting @blarghmatey! I can reproduce this.  Sorry about the delay in repsonding, I have been setting up a new machine :).  I'll try to get to this this weekend.

kseistrup:
It works here:

```sh
~ $ ps aux | grep unbound
unbound  15639  0.0  0.3 116008 29612 ?        Ssl  Oct07   2:52 /usr/bin/unbound -d
~ $
```

Cython 3.5.0
xonsh 0.2.0.r32.g652f333
Linux 4.2.2 (x86_64) ArchLinux

scopatz:
@blarghmatey and @kseistrup - do you have `$XONSH_STORE_STDOUT` set to `True`? This seems to be causing the problem on my machine. Probably time to dive back into that TeePTY code...

kseistrup:
Setting `$XONSH_STORE_STDOUT=1` provokes the bug.  Removing the assigment, or setting `$XONSH_STORE_STDOUT=0` makes the expression behave as expected,

blarghmatey:
@scopatz That was the issue. Removing that from my `.xonshrc` made pipes work properly again.

scopatz:
Yeah, piping *should* work here too,  but that is a temporary fix.

scopatz:
Ok - hilariously, the problem is that grep is too fast to start up.

scopatz:
OK @blarghmatey and @kseistrup - I have merged in the PR so I can release it.  Please close this issue if the PR fixes things for you. Thanks!

blarghmatey:
So the TEEPTY_PIPE_DELAY setting allows me to pipe into grep while storing stdout in the history file. Interestingly, I'm now getting a `KeyError` for `FORMATTER_DICT` and `BASH_COMPLETIONS` with the latest release. Printing `__xonsh_env__` shows that neither of those keys are present in the environment.

scopatz:
OK - v0.2.2 should fix this.

PR #563, opened 2015-12-07
[WIP] Tee

adqm:
This is a re-implementation of the "tee" behavior in xonsh, to explicitly use the `tee` command instead of the Python implementation of `teepty`.  I believe it should fix some of the behavior discussed in #562 and #519.

That said, it's not quite ready for prime-time.  A couple issues remain:
* When the title is reset, the new title (including escape chars) shows up in `__xonsh_history__.outs`.  I am not sure where/when/how this is happening, since printing `hist.last_cmd_out` at the end of `run_proc` _doesn't_ include the title (i.e., the title isn't going through `tee`).  @scopatz, maybe you know or can figure out what is going on, as to how that is finding its way into the history?  I'll also keep poking around.
* For some reason, I can't get mutt to work with `$XONSH_STORE_STDOUT` on, though other commands that require terminals (e.g., vim) seem to work fine.  I can fiddle with this some more.


xonsh/built_ins.py
```diff
@@ -569,25 +597,44 @@ def run_subproc(cmds, captured=True):
...
-                proc = cls(aliased_cmd,
-                           universal_newlines=uninew,
-                           env=ENV.detype(),
-                           stdin=stdin,
-                           stdout=stdout,
-                           stderr=stderr,
-                           **subproc_kwargs)
+                if usetee:
+                    _tee_file = tempfile.NamedTemporaryFile()
+                    m, s = pty.openpty()
```
scopatz: I don't understand this.  Doesn't the `tee` command set up its own pty? Or am I missing something about how the processes get hooked up?

xonsh/built_ins.py
```diff
@@ -569,25 +597,44 @@ def run_subproc(cmds, captured=True):
...
+                    _set_pty_size(m)
+                    try:
+                        old_sigwinch_handler = signal.signal(signal.SIGWINCH, _sigwinch_maker(m))
+                    except:
+                        pass
+                    tproc = Popen(['tee', _tee_file.name],
```
scopatz: Or is this a tee-only process that then you PIPE the `aliased_cmd` process through?

scopatz:
Hi @adqm - have you tried this with piping to grep a la #409?

scopatz:
> For some reason, I can't get mutt to work with $XONSH_STORE_STDOUT on, though other commands that require terminals (e.g., vim) seem to work fine. I can fiddle with this some more.

Can you pipe mutt through tee normally?  My guess is that mutt is not issuing the standard "skip me" flags.

scopatz:
> @scopatz, maybe you know or can figure out what is going on, as to how that is finding its way into the history?

This is because you call `pty.openpty()` prior to calling `settitle()`.  `openpty()` will fork the process all code after that happens on *both* the parent and child process.  So the title is being written to the stdout of both the parent (which sets the title in the terminal) and the child (which will write it to what is captured by tee).

Also, and perhaps more importantly, the child is never closed via `os.close()`.  If I am correct, this will keep forking the process everytime you run a command.  The example given in [the python docs here](https://docs.python.org/3.5/library/pty.html#example) is actually pretty bad because the parent and child are both closed automatically when the process ends.

scopatz:
Thanks for putting this in, BTW.  This is a clever idea.  Does it work with backgrounding, etc?

xonsh/built_ins.py
adqm: Right.  That process is just running `tee`, and we just pass the `stdout` of the process we want to it, and pass _its_ stdout to where we wanted stdout to go.

So we do have to set up the pty ourselves, kind of...in fact, without opening the pty, mutt works and vim works, but vim displays a warning about not outputting to a terminal (since it is outputting to a pipe instead).  Opening the pty seems like the right answer.  After some testing, I think you're right that tee just doesn't know how to handle mutt's escapes.  I can find a way around that, I think.

adqm:
> have you tried this with piping to grep a la #409?

Yep.  Seems to work fine.

> Also, and perhaps more importantly, the child is never closed via os.close().

That is true, and it is an issue (also that we open a new pty for _every_ process in the pipeline, instead of just the last).

> Does it work with backgrounding, etc?

Well, kind of.  It allows backgrounding, but I have to be more careful about closing things in that case, as above.  Sometimes the cleanup might not happen properly.

adqm:
> This is because you call pty.openpty() prior to calling settitle(). openpty() will fork the process all code after that happens on both the parent and child process. So the title is being written to the stdout of both the parent (which sets the title in the terminal) and the child (which will write it to what is captured by tee).

That is not actually what is happening.  `openpty` is getting called a lot, and that is a problem (see above), but it does not fork the process.  Also, the title is definitely not getting through to `tee`.  Rather, it looks like it is getting caught by the `Tee` class in `base_shell.py` (which is why I didn't notice it before).

Anyway, I think I have a strategy for fixing.  Will ping when I have something.

adqm:
I think this is actually not going to work.  It fixed some of the problems with the current "tee" implementation, but created others with just about the same severity.

scopatz:
That is too bad. I was kind of hoping for an alternative.  For posterity, what other problems did it cause?

PR #564, opened 2015-12-08
Use get_sep function instead of repeating code

adqm:
tiny change to make use of `tools.get_sep` in one spot, at least, to avoid repeated code.


scopatz:
Thanks @adqm!

PR #565, opened 2015-12-08
Fix execution on Linux, resolve #266

### LINKED ISSUE #266: xonsh startup script fails on some platforms
Hello all, I just found out about xonsh yesterday and have been checking it out.  After installing it (cloned from github, then used pip install . in the cloned directory) my first discovery was that it wasn't put anywhere on my path.  Then I found the 'xonsh' script in the scripts directory.  Upon executing it, I get:

/usr/bin/env: python -u: No such file or directory

The problem is the shebang at the top of the script.  It called /usr/bin/env python -u.  Aside from the fact that 'python' on my system will get you Python 2.7 and Python 3 is called python3 (a common configuration when legacy and current Python are both installed), the real issue is that using parameters like that in a shebang doesn't work on many systems.  After doing some reading, it appears that this would work OK on OSX but fail on most (if not all) Linux systems.  The system processes "python -u" as a single string and looks for an executable with that whole name.

I'm not sure how you'd like this fixed.  Personally I set the PYTHONUNBUFFERED environment variable to 1 before launching the script after changing the shebang to be #!/usr/bin/env python3.  I don't think that's a good general answer, though.  It would fail on systems which have python3 as just 'python' and not guarantee unbuffered access.  Perhaps the disabling of buffering should be handled in code?

lmmx:
Export an environment variable PYTHONUNBUFFERED: equivalent to the `-u` flag, producing unbuffered output from python. Environment variable is an arbitrary string.

### DIFF
--- a/scripts/xonsh
+++ b/scripts/xonsh
@@ -1,3 +1,3 @@
 #!/usr/bin/env sh
-export PYTHONUNBUFFERED="XONSH_SET"
+export PYTHONUNBUFFERED=1
 /usr/bin/env python -c "from xonsh.main import main; main()"
--- a/scripts/xonsh
+++ b/scripts/xonsh
@@ -1,3 +1,3 @@
-#!/usr/bin/env python -u
-from xonsh.main import main
-main()
+#!/usr/bin/env sh
+export PYTHONUNBUFFERED="XONSH_SET"
+/usr/bin/env python -c "from xonsh.main import main; main()"


scripts/xonsh
```diff
@@ -1,3 +1,3 @@
-#!/usr/bin/env python -u
```
scopatz: Can you try quoting the python command, ie

```console
#!/usr/bin/env "python -u"
```

scripts/xonsh
```diff
@@ -1,3 +1,3 @@
...
-from xonsh.main import main
-main()
+#!/usr/bin/env sh
```
scopatz: I am not sure how I feel about this. This line means that to start xonsh requires a posix compatible shell, which is the whole thing we are trying to avoid.  I think this version could make an interesting alternative script (maybe called `xonshell`?) or something for people who really need it.

scopatz: If the quoting thing doesn't work that is....

lmmx:
quoting doesn't work I'm afraid... not sure what you advise in that case

scopatz:
@lmmx - ok, can you try one more thing before we move to having separate scripts, try this file. Note that there is no `-u`

```python
#!/usr/bin/env PYTHONUNBUFFERED=XONSH_SET python
from xonsh.main import main
main()
```

scopatz:
Actually, this is really a boolean so it should be:

```python
#!/usr/bin/env PYTHONUNBUFFERED=1 python
from xonsh.main import main
main()
```

scopatz:
See the docs here: https://docs.python.org/3.5/using/cmdline.html#envvar-PYTHONUNBUFFERED

lmmx:
Freezes when executed - confirmed incompatible with Linux [here](http://unix.stackexchange.com/questions/20880/how-can-i-use-environment-variables-in-my-shebang#comment28504_20926) :pensive:

scopatz:
I am running on ubuntu 14.04 and the following works for me: 

**test.py**
```python
#!/usr/bin/env PYTHONUNBUFFERED=1 python
import os
print(os.environ.get('PYTHONUNBUFFERED'))
```

*then*
```console
$ chmod 755 test.py
$ ./test.py
```

Does this not work for you?

scopatz:
Ahh ok.  I see. I can get test.py to freeze when I start it under bash.  test.py works under xonsh, though :).  So this is not a Linux limitation this is a shell limitation.

scopatz:
See [this thread](http://unix.stackexchange.com/a/20895/129048) and notice "`The shebang line is very limited. Under many unix variants (including Linux), you can have only two words: a command and a single argument. There is also often a length limitation.`"

So I think that the right thing to do might be to have the xonsh script stay as it is.  However, we could add your fix as a `xonsh.sh` script that starts xonsh in a sh-compatible way.  People can alias xonsh.sh to xonsh if they want.

scopatz:
Some documentation about this distinction would be useful too.

scripts/xonsh
```diff
@@ -1,3 +1,3 @@
...
+export PYTHONUNBUFFERED=1
+/usr/bin/env python -c "from xonsh.main import main; main()"
```
adqm: Is there a reason to do this instead of `/usr/bin/env python -m xonsh.main`?

Not sure whether there is any practical difference between the two.

scopatz: The only difference I can think of is that in your version, the `xonsh.main.__name__ == "__main__"`, and in the version here `xonsh.main.__name__ == "/path/to/xonsh/main.py"` while `"__main__" ` is assigned to the `-c` string.  Not a meaningful distinction :)

scopatz:
Also, I now kinda like the name `xon.sh` for this script.

lmmx:
Great! Doubles as the [xon.sh](http://xon.sh/) URL now too -- also v sorry for going unresponsive on this and the `xonsh_kernel` issue, busy month and didn't see any emails coming through, will retry xonsh this weekend

scopatz:
Thanks @lmmx, I really appreciate it. I want to make sure that xonsh works for you.

pothos:
To be usable for scripts the parameters need to be passed:
`/usr/bin/env python3 -c "from xonsh.main import main; main()" $@`

P.S.: `#!/bin/sh` should be enough as first line?

santagada:
I think this was fixed with xon.sh right? we can probably close this pull request then

scopatz:
Yes, I think this can be closed (finally).  Feel free to reopen or open a new issue if desired.

lmmx:
Hurray! Thank you for all your work on it, hope to get it to work properly for me soon (but couldn't stick with it due to too many things on plate at the time) Will get up early and test tomorrow :-)

PR #568, opened 2015-12-08
add `man` to completion skip tokens

gforsyth:
Just a quick addition to allow tab completion when opening man pages


scopatz:
Thanks @gforsyth!

Issue #569, opened 2015-12-11
Startup (compile?) fails on Python 3.5.1


kseistrup:
Having upgraded Python from 3.5.0 to 3.5.1 on Arch Linux I get the following traceback when trying to launch `xonsh`:

```pytb
Traceback (most recent call last):
  File "/usr/bin/xonsh", line 9, in <module>
    load_entry_point('xonsh==0.2.3', 'console_scripts', 'xonsh')()
  File "/usr/lib/python3.5/site-packages/xonsh/main.py", line 111, in main
    args = premain(argv)
  File "/usr/lib/python3.5/site-packages/xonsh/main.py", line 98, in premain
    shell = builtins.__xonsh_shell__ = Shell(**shell_kwargs)
  File "/usr/lib/python3.5/site-packages/xonsh/shell.py", line 29, in __init__
    self._init_environ(ctx)
  File "/usr/lib/python3.5/site-packages/xonsh/shell.py", line 67, in _init_environ
    self.ctx = xonshrc_context(rcfiles=rc, execer=self.execer)
  File "/usr/lib/python3.5/site-packages/xonsh/environ.py", line 685, in xonshrc_context
    execer.exec(rc, glbs=env)
  File "/usr/lib/python3.5/site-packages/xonsh/execer.py", line 124, in exec
    stacklevel=stacklevel)
  File "/usr/lib/python3.5/site-packages/xonsh/execer.py", line 98, in compile
    code = compile(tree, filename, mode)
TypeError: required field "lineno" missing from arg
```

However, I don't quite understand this error message because `compile()` doesn't seem to require a `lineno` arg:

```py
>>> help(compile)
Help on built-in function compile in module builtins:

compile(source, filename, mode, flags=0, dont_inherit=False, optimize=-1)
    Compile source into a code object that can be executed by exec() or eval().
    
    The source code may represent a Python module, statement or expression.
    The filename will be used for run-time error messages.
    The mode must be 'exec' to compile a module, 'single' to compile a
    single (interactive) statement, or 'eval' to compile an expression.
    The flags argument, if present, controls which future statements influence
    the compilation of the code.
    The dont_inherit argument, if true, stops the compilation inheriting
    the effects of any future statements in effect in the code calling
    compile; if absent or false these statements do influence the compilation,
    in addition to any features explicitly specified.
```

adqm:
[...]

PR #570, opened 2015-12-11
ast.arg needs lineno and col_offset in Python 3.5.1

adqm:
I believe this should resolve the issue described in #569, which caused a `compile` error when running xonsh under Python 3.5.1.

Tested on Python 3.5.1 and 3.4 on my machine, but more eyes on it would be appreciated.

### DIFF
--- a/xonsh/parser.py
+++ b/xonsh/parser.py
@@ -709,7 +709,14 @@ class Parser(object):
 
     def p_vfpdef(self, p):
         """vfpdef : NAME"""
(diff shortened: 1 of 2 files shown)


scopatz:
Looks good to me.  Should @kseistrup try this out before we merge?

adqm:
Yep; it's always good to have more than one person run something successfully before assuming everything works :)

scopatz:
Great work @adqm!

PR #571, opened 2015-12-15
Improvements for Completer: Names with Spaces, and Partial Strings

adqm:
This changeset is a long-overdue continuation of #72.  It implements a couple of fixes/changes for the completer, basically trying to mimic the way Windows cmd.exe handles quoted paths:
* implements completion of partial strings (e.g, `"name with spa<tab>` will now complete to include the full name, as well as the proper ending quote)
* implements completion of full strings (e.g. `"name with spaces/"<tab>` will complete to show subdirectories of `name with spaces`, properly quoted [I think])

As it always has (modulo the issue in #181), it still wraps names containing spaces in quotes if they aren't already there.

The main reason I am marking as a work-in-progress is that I haven't yet gotten it to work in readline mode (nor have I tried yet).  Also, I'm not 100% sure I haven't broken something along the way.  Happy for other suggestions as well.

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -17,10 +17,10 @@ RE_DASHF = re.compile(r'-F\s+(\w+)')
 RE_ATTR = re.compile(r'(\S+(\..+)*)\.(\w*)$')
 RE_WIN_DRIVE = re.compile(r'^([a-zA-Z]):\\')
 
-RE_PARTIAL_TRIPLE_DOUBLE_STRING = (r'([bBrRuU]*"""((\\(.|\n))|([^"\\])|("(?!""))|\n)*(?:""")?)', '"""')
-RE_PARTIAL_TRIPLE_SINGLE_STRING = (r"([bBrRuU]*'''((\\(.|\n))|([^'\\])|('(?!''))|\n)*(?:''')?)", "'''")
-RE_PARTIAL_DOUBLE_STRING = (r'([bBrRuU]*"((\\(.|\n))|([^"\\]))*"?)', '"')
-RE_PARTIAL_SINGLE_STRING = (r"([bBrRuU]*'((\\(.|\n))|([^'\\]))*'?)", "'")
+RE_PARTIAL_TRIPLE_DOUBLE_STRING = (r'(([bBrRuU]*""")((\\(.|\n))|([^"\\])|("(?!""))|\n)*(?:""")?)', '"""')
+RE_PARTIAL_TRIPLE_SINGLE_STRING = (r"(([bBrRuU]*''')((\\(.|\n))|([^'\\])|('(?!''))|\n)*(?:''')?)", "'''")
+RE_PARTIAL_DOUBLE_STRING = (r'(([bBrRuU]*")((\\(.|\n))|([^"\\]))*"?)', '"')
+RE_PARTIAL_SINGLE_STRING = (r"(([bBrRuU]*')((\\(.|\n))|([^'\\]))*'?)", "'")
 STRINGS = (RE_PARTIAL_TRIPLE_DOUBLE_STRING, RE_PARTIAL_TRIPLE_SINGLE_STRING,
            RE_PARTIAL_DOUBLE_STRING, RE_PARTIAL_SINGLE_STRING)
 
@@ -35,7 +35,10 @@ def _path_from_partial_string(inp, pos=None):
             _s = x[-1][0]
             if not _s.endswith(end):
                 _s = _s+end
-            return s, eval(_s)
+            val = eval(_s)
+            if isinstance(val, bytes):
+                val = val.decode()
+            return s, val, x[-1][1], end
 
 
 XONSH_TOKENS = {
@@ -160,13 +163,18 @@ class Completer(object):
         prefixlow = prefix.lower()
         _line = line
         line = builtins.aliases.expand_alias(line)
-        rep = None
+        # string stuff for automatic quoting
+        path_str_start = ''
+        path_str_end = ''
         p = _path_from_partial_string(_line, endidx)
+        in_partial_str = False
+        lprefix = len(prefix)
         if p is not None:
-            lpath = len(p[0])
+            lprefix = len(p[0])
             prefix = p[1]
-        else:
-            lpath = len(prefix)
+            path_str_start = p[2]
+            path_str_end = p[3]
+            in_partial_str = True
         cmd = line.split(' ', 1)[0]
         while cmd in COMPLETION_SKIP_TOKENS:
             begidx -= len(cmd)+1
@@ -186,19 +194,19 @@ class Completer(object):
                     s = s.rstrip() + slash
                 rtn.add(s)
             if len(rtn) == 0:
-                rtn = self.path_complete(line, endidx, prefix)
+                rtn = self.path_complete(prefix, path_str_start, path_str_end)
         elif prefix.startswith('${') or prefix.startswith('@('):
             # python mode explicitly
             rtn = set()
         elif prefix.startswith('-'):
-            return lpath, sorted(self._man_completer.option_complete(prefix, cmd))
+            return lprefix, sorted(self._man_completer.option_complete(prefix, cmd))
         elif cmd not in ctx:
             if cmd == 'import' and begidx == len('import '):
                 # completing module to import
-                return lpath, sorted(self.module_complete(prefix))
+                return lprefix, sorted(self.module_complete(prefix))
             if cmd in self._all_commands():
                 # subproc mode; do path completions
-                return lpath, sorted(self.path_complete(line, endidx, prefix, cdpath=True))
+                return lprefix, sorted(self.path_complete(prefix, path_str_start, path_str_end, cdpath=True))
             else:
                 # if we're here, could be anything
                 rtn = set()
@@ -214,8 +222,8 @@ class Completer(object):
         rtn |= {s for s in dir(builtins) if startswither(s, prefix, prefixlow)}
         rtn |= {s + space for s in builtins.aliases
                 if startswither(s, prefix, prefixlow)}
-        rtn |= self.path_complete(line, endidx, prefix)
-        return lpath, sorted(rtn)
+        rtn |= self.path_complete(prefix, path_str_start, path_str_end)
+        return lprefix, sorted(rtn)
 
 
     def find_and_complete(self, line, idx, ctx=None):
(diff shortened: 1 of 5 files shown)


adqm:
Okay, I made some small changes to get it to (almost) work with readline.  It's not perfect there, but I personally think it's still a step up.  I think it's ready for a look.

xonsh/tools.py
```diff
@@ -813,3 +813,49 @@ def print_color(string, file=sys.stdout):
...
+
+RE_STRING_CONT = {
+    '"': r'((\\(.|\n))|([^"\\]))*',
+    "'": r"((\\(.|\n))|([^'\\]))*",
+    '"""': r'((\\(.|\n))|([^"\\])|("(?!""))|\n)*',
+    "'''": r"((\\(.|\n))|([^'\\])|('(?!''))|\n)*",
+}
+
+
+def partial_string_finder(x):
```
scopatz: Public functions should have a docstring

xonsh/tools.py
```diff
@@ -813,3 +813,49 @@ def print_color(string, file=sys.stdout):
...
+            RE_STRING_DOUBLE,
+            RE_STRING_SINGLE]
+RE_BEGIN_STRING = "(" + RE_STRING_START + "|".join(_STRINGS) + ')'
+
+RE_STRING_CONT = {
+    '"': r'((\\(.|\n))|([^"\\]))*',
+    "'": r"((\\(.|\n))|([^'\\]))*",
+    '"""': r'((\\(.|\n))|([^"\\])|("(?!""))|\n)*',
+    "'''": r"((\\(.|\n))|([^'\\])|('(?!''))|\n)*",
+}
```
scopatz: These should all be pre-compiled.  It is waaaay faster.

xonsh/completer.py
```diff
@@ -17,6 +17,35 @@
...
+    if lx == 0:
+        return None
+    elif lx % 2: # odd number: in the middle of a string
+        s = partial[x[-1]:]
+    else:
+        if x[-1] != pos:
+            return None
+        s = partial[x[-2]:x[-1]]
+    b = b[-1]
+    end = re.sub('[rRbBuU]*','',b)
```
scopatz: Should be pre-compiled.

xonsh/completer.py
```diff
@@ -17,6 +17,35 @@
...
+    _s = s
+    if not _s.endswith(end):
+        _s = _s + end
+    try:
+        val = eval(_s)
+    except:
+        return None
+    if isinstance(val, bytes):
+        val = val.decode()
```
scopatz: I think all decode() calls need to get passed `$XONSH_ENCODING` and `$XONSH_ENCONDING_ERRORS` in order to safely handle various unicode issues.  Given the purpose of this function, you might want to pass these values in instead of looking them up here.

xonsh/completer.py
```diff
@@ -17,6 +17,35 @@
...
+        if x[-1] != pos:
+            return None
+        s = partial[x[-2]:x[-1]]
+    b = b[-1]
+    end = re.sub('[rRbBuU]*','',b)
+    _s = s
+    if not _s.endswith(end):
+        _s = _s + end
+    try:
+        val = eval(_s)
```
scopatz: Hmmm I wonder if there isn't a safer way to do this.  It seems like you could trick this into executing any function by passing in something like `"" + f() + ""`.  Or am I wrong about that?  It might be safer to parse an AST and check for a Str or Bytes node....

scopatz: Or maybe I am worrying too much...

xonsh/completer.py
```diff
@@ -137,9 +166,22 @@ def complete(self, prefix, line, begidx, endidx, ctx=None):
...
+        lprefix = len(prefix)
+        if p is not None:
+            lprefix = len(p[0])
+            prefix = p[1]
+            path_str_start = p[2]
+            path_str_end = p[3]
+            in_partial_str = True
         cmd = line.split(' ', 1)[0]
-        if cmd in COMPLETION_SKIP_TOKENS:
+        while cmd in COMPLETION_SKIP_TOKENS:
```
scopatz: Just to be clear, this is to cover cases like `time sudo less ls`

xonsh/completer.py
```diff
@@ -157,20 +199,19 @@ def complete(self, prefix, line, begidx, endidx, ctx=None):
...
             if len(rtn) == 0:
-                rtn = self.path_complete(prefix)
-            return sorted(rtn)
+                rtn = self.path_complete(prefix, path_str_start, path_str_end)
         elif prefix.startswith('${') or prefix.startswith('@('):
             # python mode explicitly
             rtn = set()
         elif prefix.startswith('-'):
-            return sorted(self._man_completer.option_complete(prefix, cmd))
+            return lprefix, sorted(self._man_completer.option_complete(prefix, cmd))
```
scopatz: It seems like the API has changed, but the docstring was not updated.  I'm not clear on what lprefix does yet.  Probably should read more :)

xonsh/tools.py
scopatz: A test here would be nice too.

xonsh/completer.py
scopatz: Ok so it seems that lprefix is just used by prompt-toolkit. Also it seems like lprefix should come after the completions, to be consistent with  prompt-toolkit, but it is your call.

xonsh/completer.py
scopatz: Ok, actually, now that I am looking at it, I think that the following should be safe:

```python
val = eval(_s, globals={'__builtins__': {}})
```

But, even better the standard library provides.  This should probably be `ast.lieteral_eval()`: https://docs.python.org/3/library/ast.html#ast.literal_eval

scopatz:
Much agreed @adqm! I think that this is a significant improvement.  I tried to go over it with a fine tooth comb.  I think it could use some cleanup before merging as it was originally a WIP, but otherwise looks good.

scopatz:
Thanks a ton!

xonsh/completer.py
adqm: I'm not sure that is actually a worry (in your example, the two strings would be handled separately, so the completer would just eval the empty string).  Plus, since we're executing on the user's machine, isn't there a lot of opportunity to run arbitrary functions anyway? :)

But I certainly catch your point, and I can make the change.

xonsh/completer.py
adqm: Yes.  Sorry for making the unrelated change in amidst everything else.

xonsh/completer.py
adqm: Yep.  I had dreams that `lprefix` would be used in both, but unless I'm missing something, readline's completer is much more finicky.

It represents the length of the prefix string (how far back do we have to delete to get to the front of the completion).

adqm:
Thanks for taking a look, and for all the comments.  I'll try to find some time today to get them all taken care of.

xonsh/completer.py
scopatz: Unrelated improvements are great! Just wanted to understand it :)

xonsh/completer.py
scopatz: Thanks!

melund:
This is really great. Thanks @adqm. 

Works like a charm with `FORCE_POSIX_PATHS=True`.... 
But I also tried it without on windows and there are still some issues with backslashes.
```
[xonsh] C:\Users\mel\Documents\xonsh>xonsh
mel@pc ~\Documents\xonsh adqm-completer $ cd ..
mel@pc ~\Documents $ cd scrats\
mel@pc ~\Documents\scrats $ cd 'bad folder\'
cd takes 0 or 1 arguments, not 2
```

adqm:
Glad it (kind of) works!  

Thanks for letting me know about the issue, @melund.  I hadn't yet tried testing on Windows.  I have some ideas that might maybe fix, but it will be a little while before I am able to try it out on a Windows box.  I'll ping you here when I think I have something.

adqm:
Alright @melund, @scopatz, this should be ready for another look.

Since the last check in, I think I hit all of the suggestions from before (let me know if I missed something).  Also, I:

* allowed slightly more complicated completions like `cmd.exe` does, so `"name with spaces/"subdir<tab>` will now complete in an appropriate way
* tried to escape things in strings appropriately, including:
  * fixed the issue on Windows (I think).  It is a little bit ugly (stick another backlash on the end of the string so that it doesn't view the trailing backslash as escaping the quote), but it should avoid the errors
* added a test case.  I wanted to be thorough, but if it's overkill, I can scale it back so that it only tests a couple of simple cases rather than doing the (almost exhaustive) thing it does now
* refactored everything multiple times

melund:
Fantastic. I wanted to do something similar, but just couldn't see how to do it.

It works great for me. You are right that it looks a bit ugly if one haven't got `FORCE_POSIX_PATHS` turned on. Maybe the paths could just be turned into raw strings instead (e.g. `r'my\path\'`). I guess it would be OK on windows. 

The only thing I am missing from cmder and cmd.exe is for the commands to work on something where the end quote is missing. But it may not be related to your PR. 

For example: 
```
mel@PC ~/D/scrats $ cd ..
mel@PC ~/Documents $  cd scrat<tab twice>
mel@PC ~/Documents $  cd 'scrats/a 22.log'
```
I regret and use backspace to remove... Then I press enter without putting in the missing quote. That works on cmd.exe but not here.
```
mel@PCL ~/Documents $ cd 'scrats/
cd: no such file or directory: 'scrats/
``` 

Crazy test. I like it :)

adqm:
> It works great for me. You are right that it looks a bit ugly if one haven't got FORCE_POSIX_PATHS turned on. Maybe the paths could just be turned into raw strings instead (e.g. r'my\path\'). I guess it would be OK on windows. 

That was what I had wanted to do originally, but even in a raw string, the last character being a backslash doesn't work (quirk of Python, not just of xonsh).  It has been marked as "not a bug," so I don't think it's likely ever to change in Python ([see here](http://bugs.python.org/issue1271)), and I think it's important for xonsh to keep Python compatibility (though maybe that's up for debate and we should make our own raw string type that doesn't have that limitation).

Anyway, my thought, for now, at least, is that ugly is better than broken :)

> The only thing I am missing from cmder and cmd.exe is for the commands to work on something where the end quote is missing. But it may not be related to your PR. 

Hmmm, that would be nice.  I suppose we could make the parser check whether something looks like a partial string, and try to interpret it as such.  But I do think that it should live in a separate pull request.

melund:
Yes. I discovered the same with the last backslash. Thought it was a xonsh issue.  That was a shame. Another reason for using FORCE_POSIX_PATH

If @scopatz is OK with it I think it should be merged. I really can't live without it. It may also be time for a new release. I think the conch has out grown its current shell..

scopatz:
Thanks a million @adqm (and @melund  for reveiwing)! This is awesome.

>  It may also be time for a new release. I think the conch has out grown its current shell..

Definitely. I am busy trying to a get a paper out the door in the next few days, but I really welcome anyone who wants to put out a release.  There are [release docs here](http://xonsh.org/devguide.html#branches-and-releases) that are probably more involved than they need to be.  A micro release should just be a tag and creating a pypi and conda package.  I don't need to be the blocker here.  I'd really love it if realeased happened whenever anyone felt like it, myself included :)

Issue #539, opened 2015-11-29
Variable expansion shouldn't work on single quoted string


zn:
Step to reproduce:

1. Install xonsh via `pip`
2. run command `echo 'Hello' | perl -ne 'print "$_"'`

Expected result: display `Hello`

Actual Result: display `/usr/bin/env` (Without newline)

Tested Environment:

- Linux (Fedora 23)
- perl v5.22.0
- Python 3.4.3
- xonsh v0.2.3 (Installed from `pip install https://github.com/scopatz/xonsh/archive/master.tar.gz`)

(got same bug on virtualenv install and root install(`sudo pip install xonsh`)

![image](https://cloud.githubusercontent.com/assets/830515/11456818/96eb70ba-96d9-11e5-955c-890429812e35.png)

scopatz:
Thanks for reporting @zn.  I am not a perl expert, so I am not sure exactly what `$_` is suppossed to signify. Usually this is an environment variable for the command that is currently being run.  (So I guess we are launching something with an env prefix). Perl seems to use this as an alternative to stdin?

zn:
oops. I think I'd reported wrong problem. It seem the real problem is about variable substitution on single-quoted string, not about perl.

scopatz:
There is no difference in singly and doubly quoted strings. But I think I see what you are referring to. There is no way to have $VAR pass through the srting.  Maybe \$VAR will work?  We just use os.path.expandvars()

zn:
`\$VAR` doesn't work.. but I found workaround by assigning `$_` to another variable.
```sh
user $ echo '$_'
/usr/bin/env
user $ echo '\$_'
\/usr/bin/env

# assign $_ to other variable
# work as expected result
user $ $__ = '$_'
user $ echo '$__'
$_
user $ echo 'It work as expected!' | perl -ne 'print "$__"'
It work as expected!
```

adqm:
Another way is to build the strings using Python-mode explicitly (since strings in Python mode don't do the expansion of environment variables):
```
hartz@compy-488 ~ $ echo @("$_")
$_
hartz@compy-488 ~ $ echo "Hello!" | perl -ne @('print "$_"')
Hello!
```

scopatz:
Ahh ok.  I see now. We still should think about implementing \$ as an escape.  This would have to happen in `xonsh.environ.expand_path()`.

adqm:
[...]

Issue #338, opened 2015-07-31
Env variables doesn't expand to paths in sub process mode.


melund:
Environment variables can't be used to create a path in subprocess mode: 

For example on windows: `cd $USERPROFILE` works fine. 

But `cd $USERPROFILE/Documents` does not.  Pressing tab in the shell causes the autocompleter to expand the path. 

So not a big issue in an interactive shell. But typing something like that in a file or having it in an alias doesn't work. 

@michaelaye initially reported this:
https://github.com/scopatz/xonsh/pull/319#issuecomment-125018134
```
klay6683@MacL2938 ~ $ aliases['cdsrc']
['cd', '$HOME/Dropbox/src']
klay6683@MacL2938 ~ $ cdsrc
cd: no such file or directory: $HOME/Dropbox/src
```

scopatz:
I think that we need to start having tests for some of these aliases too.

scopatz:
I remember this used to work.

melund:
This has been fixed

seanfarley:
With the latest xonsh (0.4.5) this is broken again.

scopatz:
I think it does if you put them in quotes for some reason. Of course, it should always...

PR #576, opened 2015-12-19
Environment Variable Expansion

adqm:
This changeset:
* implements the changes discussed at the bottom of #539, including updating documentation
* makes aliases do path expansion at well (as discussed in #338)

I think it is all working, and I hope I didn't break anything else.

Also, separately, I'm not quite sure how the Travis and Appveyor autobuild things work, but I am trying here to make them run in quiet mode to avoid printing a dot for each of the test cases (there are really a lot of them now, after the last PR was merged, and printing the dots really seems to slow things down).


adqm:
Woops this broke things on Windows, it seems.  I'll ping when fixed.

scopatz:
Thanks @adqm. Did a quick review and it seemed good. Didn't see the win error. Just ping.

adqm:
Okay, @scopatz, ready for another look, I think.  Changes from last time:
* included a modified version of `expandvars` that uses xonsh's environment instead of `os.environ` to do replacement in strings
* fixed the regexes that check for which ensurer to use, to avoid `DIRSTACK_SIZE` being interpreted as a `\w*DIRS` variable
* skipped several test cases on Windows, where spoofed environments were necessary
* made the environment variable lookups always fall back to the default values so that just checking, e.g., the value of `$XONSH_STORE_STDOUT` works now (shows the default, rather than raising an IndexError).  Is there a reason things didn't work this way before?

xonsh/tools.py
```diff
@@ -909,3 +909,50 @@ def check_for_partial_string(x):
...
+    end = '}'
+    i = 0
+    while True:
+        m = search(path, i)
+        if not m:
+            break
+        i, j = m.span(0)
+        name = m.group(1)
+        if name.startswith(start) and name.endswith(end):
+            name = eval(name[1:-1]) # change for non-POSIX ${...}
```
scopatz: `ast.literal_eval()` maybe? Or is this a full eval? Also do you need the current locals and globals?  Seems like an empty execution context is desirable here.

xonsh/tools.py
```diff
@@ -909,3 +909,50 @@ def check_for_partial_string(x):
...
         return (string_indices[-2], string_indices[-1], starting_quote[-1])
+
+
+# expandvars is a modified version of os.path.expandvars from the Python 3.5.1
+# source code (root/Lib/posixpath.py, line 266)
+
+_varprog = None
+
+
+def expandvars(path):
```
scopatz: This seems way more complex than using [`Env.replacve_env()`](http://xonsh.org/api/environ.html#xonsh.environ.Env.replace_env) to replaces the contents of os.environ with detyped env, and then call os.expandvars().  Is this doing something more xonsh-specific that I am not seeing? Or is it more performant in some way?

xonsh/built_ins.py
```diff
@@ -180,7 +182,9 @@ def expand_path(s):
     global ENV
     if ENV is not None:
         ENV.replace_env()
-    return os.path.expanduser(os.path.expandvars(s))
+    if ENV.get('EXPAND_ENV_VARS'):
+        s = expandvars(s)
```
scopatz: This still calls `ENV.repleace_env()

scopatz: This still calls `ENV.replace_env()`, which now seems either unnecessary or totally sufficient.

xonsh/tools.py
```diff
@@ -909,3 +909,50 @@ def check_for_partial_string(x):
...
+    """Expand shell variables of form $var and ${var}.  Unknown variables
+    are left unchanged."""
+    global _varprog
+    ENV = builtins.__xonsh_env__
```
scopatz: Should this be lower-case env?

xonsh/tools.py
```diff
@@ -909,3 +909,50 @@ def check_for_partial_string(x):
...
+    if isinstance(path, bytes):
+        path = path.decode(encoding=env.get('XONSH_ENCODING'),
+                           errors=env.get('XONSH_ENCODING_ERRORS'))
```
scopatz: Or should this be upper case env?

xonsh/tools.py
```diff
@@ -909,3 +909,50 @@ def check_for_partial_string(x):
...
+    if '$' not in path:
+        return path
+    if not _varprog:
+        import re
+        _varprog = re.compile(r'\$(\w+|\{[^}]*\})', re.ASCII)
```
scopatz: This doesn't handle `%name%` completion on windows, which someone will probably complain about :) https://docs.python.org/3.5/library/os.path.html#os.path.expandvars  Shouldn't be too hard to modify.

scopatz:
> made the environment variable lookups always fall back to the default values so that just checking, e.g., the value of $XONSH_STORE_STDOUT works now (shows the default, rather than raising an IndexError). Is there a reason things didn't work this way before?

Nope that was an oversight!  Thanks for fixing.

scopatz:
Thanks for all of the hard work here!

xonsh/tools.py
adqm: Well, one thing is that I think it is nice if this works with the `${..}` syntax (none of the `expandvars` in `os` will do that).

More important is the fallback to default values, which wouldn't work with `replace_env`.

xonsh/tools.py
adqm: I think this has to be a full eval, but I think it should happen in the current locals and globals.

My thought is that if we have `x = "HOME"`, both of the following should work:
```
>>> ${x}
>>> echo "${x}"
```

literal eval wouldn't handle that case.

xonsh/tools.py
adqm: you're right that one of them should change :)

xonsh/tools.py
adqm: Ah, yes.  That's what I get for copying from `posixpath.py` and not even looking at the windows equivalent.  I can work that in.

xonsh/tools.py
scopatz: Ahhh OK these are both good enough for me :).

xonsh/tools.py
scopatz: Understood. Got it. That will be super cool to have.

scopatz: But I think you then want to use the xonsh eval or at least the `__xonsh_ctx__` as the execution context.  As written this will use the function's globals and locals.

adqm: Yes, sorry; that was what I meant by the comment above; we will want to eval in the xonsh context.

xonsh/built_ins.py
adqm: Are `replace_env` and `undo_replace_env` used someplace else I am not seeing?  I am planning on just deleting them unless they are useful someplace else.

adqm:
Alright, ready for another look.  This time I:
* modified `expandvars` to do the `%name%` expansion on Windows
* removed `replace_env` and `undo_replace_env`
* modified the eval to evaluate in the right context

xonsh/environ.py
```diff
@@ -242,24 +243,6 @@ def detype(self):
...
-        os.environ.update(self.detype())
-
-    def undo_replace_env(self):
-        """Replaces the contents of os.environ with a detyped version
-        of the xonsh environement.
-        """
-        if self._orig_env is not None:
-            os.environ.clear()
-            os.environ.update(self._orig_env)
-            self._orig_env = None
```
scopatz: I think that these are still potentially useful to users.  Do you mind putting them back in?

scopatz:
Looks great to me, other than the method removal. Basically ready to merge.

scopatz:
I can put them back in if you'd like too.

adqm:
I just put them back in, so we should be good to go, I think (assuming the tests pass).  Let me know if you have any other thoughts/suggestions.

I'll also take a look at your history PR tonight.

scopatz:
Awesome!

PR #577, opened 2015-12-20
add slashes to quoted strings with, e.g., ~ in them

adqm:
Small bugfix for the big completer merge.  Right now, trailing slashes aren't added to paths in quotes if those paths include ~.  For example, `~/Docum<tab>` would not add a trailing slash.  This should fix.


scopatz:
Thanks!

Issue #551, opened 2015-12-02
Prompt-toolkit history seems to be different lookup seems to be different from readline history


scopatz:
That is, it isn't reading from the same history files. This is a little annoying when going back and forth between the two.

scopatz:
Fixed in #578

PR #578, opened 2015-12-20
prompt-toolkit history now reads from xonsh history rather its own custom pickle method.

scopatz:
This addresses #551.

### DIFF
--- a/tests/test_prompt_toolkit_history.py
+++ b/tests/test_prompt_toolkit_history.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
-"""Tests for the LimitedFileHistory class."""
+"""Tests for the PromptToolkitHistory class."""
 import os
-from tempfile import NamedTemporaryFile
 
 import nose
 from nose.tools import assert_equal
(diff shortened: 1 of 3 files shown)


adqm:
I am getting a little bit of weird behavior here.  It looks like the history is being shared fine.  I can run commands in readline, quit out, and get them in prompt_toolkit; but I can't access the history via the up/down arrows from a fresh prompt_toolkit shell until after I have run a command or hit control C.  After I have done that once, though, then it is all there.  Do you know what might be the issue there?

scopatz:
Hmm I noticed that too.  Looking into it now.

scopatz:
Interestingly, it only effects the first command.  I think that this is because we set up a new `prompt()` each time.

scopatz:
So I think that the lack of up-down on the first command is actually an issue with prompt-toolkit itself.  There is no way when using the `prompt()` function to get access to the `Buffer` instance that is created.  Even using dirty tricks like `gc.get_refferers()` doesn't really work because [the buffer makes a copy of the history](https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/buffer.py#L268).  This effectively prevents asynchronous loading of the history. I have tried to load the history synchronously, but it adds too much to the start up time to be useful even for a few thousand history elements.

I vote that we merge this and I open an issue on prompt-toolkit.  Does that sound reasonable?

adqm:
That works for me, assuming the tests pass.  In the slightly longer term, though, we might want to move away from using the `prompt` function, and instead work at a lower level in prompt_toolkit (though I'm not sure exactly what that entails).

scopatz:
Yeah, I agree.  The prompt function is relatively simply but we'd want to make a class out of it (or use the PrompToolkitShell) so that we have access to there variables that it sets up.  I'll make a xonsh issue for that too.

PR #582, opened 2015-12-22
Release script 

### LINKED ISSUE #580: #!/usr/bin/env xonsh in script will eat -h

scopatz:
This adds a top-level xonsh script to help with releases.  Obviously this is mostly untested, so minor fixes should be expected.  However, it is not user facing, so it is not a big deal.  Additionally, this fixes #580, which was more annoying than I thought.

This issue is blocking the release (for me).


xonsh/main.py
```diff
@@ -23,10 +23,16 @@ def path_argument(s):
     return s
 
 
-parser = ArgumentParser(description='xonsh')
+parser = ArgumentParser(description='xonsh', add_help=False)
+parser.add_argument('-h', '--hel[',
```
adqm: Is this (`hel[` vs `help`) intentional?

scopatz: Nope! Thanks for the catch.

scopatz:
Updated. I think that this is ready for another look.

release.xsh
```diff
@@ -0,0 +1,111 @@
...
+    with open(fname, 'w') as f:
+        f.write(upd)
+
+
+def version_update(ver):
+    """Updates version strings in relevant files."""
+    pnfs = [
+        ('__version__\s*=.*', "__version__ = '{0}'".format(ver), 
+         ['xonsh', '__init__.py']),
+        ('version:\s*', 'version: {0}.{{build}}'.format(ver), '.appveyor.yml'),
```
adqm: `'.appveryor.yml'` should be `['.appveryor.yml']` (wrapped in a list or tuple) so that it doesn't get expanded as `./a/p/p/v/e/y/o/r/./y/m/l`.

release.xsh
```diff
@@ -0,0 +1,111 @@
...
+
+def pipify():
+    """Make and upload pip package."""
+    ./setup.py sdist upload
+
+
+def condaify(ver):
+    """Make and upload conda packages."""
+    conda_dir = os.path.dirname(os.path.dirname($(which conda)))
+    conda_bld = os.path.join(conad_dir, 'conda-bld')
```
adqm: typo? `conad_dir` instead of `conda_dir`

release.xsh
```diff
@@ -0,0 +1,111 @@
...
+            if doer == self.dest:
+                setattr(namespace, doer, True)
+            else:
+                setattr(namespace, doer, False)
+
+
+def main(args=None):
+    parser = ArgumentParser('release')
+    parser.add_argument('-d', action='store_true', default=False, 
+                        help='dry run')
```
adqm: I assume dry-run hasn't actually been implemented?  I just tried running this with it turned on, and I was a little bit surprised that it was trying to actually do things.

scopatz: Oh yeah, I should remove this.

scopatz:
Thanks @adqm! I think that this is ready for another look.

adqm:
I'll go ahead and merge in, even though I am still too nervous to actually run the release script :)

The fix for the args seems to work well!  Weird that there isn't a more straightforward fix built in to ArgumentParser, but after some digging of my own, I couldn't find one, either.

I think in some future version of the release script, it might be nice to have the script run through everything that happens locally, force tests to pass, etc, before doing anything that affects the remote things (github, pypi, etc) rather than interleaving everything.  I am a little bit worried that, as written, this could leave us with a "half-finished" release.  But maybe that shouldn't be a real worry.

scopatz:
> I think in some future version of the release script, it might be nice to have the script run through everything that happens locally, force tests to pass, etc, before doing anything that affects the remote things 

Ahh yeah, that is a nice idea.

PR #583, opened 2015-12-24
Fix error when trying to run commands on Windows

BYK:
On Windows, when you both have `command` and `command.bat` files Xonsh
always prefers the `command` version because `'' in PATHEXT` check
always returns `True` (because of the empty string).

This patch fixes this by treating `PATHEXT` as "env path" so Xonsh 
prefers the file with extension first.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -13,16 +13,15 @@ import atexit
 import signal
 import inspect
 import builtins
-import subprocess
-from io import TextIOWrapper, StringIO
 from glob import glob, iglob
 from subprocess import Popen, PIPE, STDOUT
 from contextlib import contextmanager
-from collections import Sequence, MutableMapping, Iterable, namedtuple, \
-    MutableSequence, MutableSet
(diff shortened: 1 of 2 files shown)


xonsh/built_ins.py
```diff
@@ -289,7 +288,7 @@ def _get_runnable_name(fname):
         if os.path.isdir(d):
             files = os.listdir(d)
             if ON_WINDOWS:
-                PATHEXT = builtins.__xonsh_env__.get('PATHEXT')
+                PATHEXT = builtins.__xonsh_env__.get('PATHEXT').split(';')
```
BYK: This line is the actual fix, rest is just clean up. I can separate those changes if you prefer that.

scopatz:
Hi @BYK, I have opened up a PR to your repo with some potential improvements as BYK/xonsh#1.  Thanks for putting this in.

BYK:
@scopatz oh nice, didn't know those existed. Thanks! (and updated the branch)

scopatz:
Thanks @BYK!

PR #585, opened 2015-12-30
Prompt Toolkit subpackage

scopatz:
This moves the prompt-toolkit code into its own `ptk` subpackage. This was done in preparation of some further refactors/customization. 

Also I fixed a small `$AUTO_CD` bug that I discovered.


adqm:
This looks good to me!  A nice change, I think.

PR #587, opened 2015-12-30
temp fix for conda packages

scopatz:
Probably shouln't merge this, but I did want to make people aware of it.  The issue in question is conda/conda-build#713


scopatz:
Conda appears to have been fixed!

PR #592, opened 2015-12-31
Fix for prompt-toolkit reserving too much space.

scopatz:
This addresses the issue with prompt toolkit taking up too much space prematurely.  Instead it allocates new tab-completion space as needed.  

This is also predatory for fixing this history issue for the first command.

Much of this is a fork of ptk itself, with minor but important modifications.


adqm:
I like this change, personally, but this was viewed as a "bug" in older versions of prompt_toolkit, I think, since the prompt "bounces around."  Can we include an option to have it revert back to the normal prompt_toolkit behavior of reserving the lines always (but still keeping the customizable number of lines to reserve)?  Maybe a boolean `$COMPLETIONS_RESERVE_SPACE`?

adqm:
Ah, never mind.  I see that the prompt is much less bouncy than the old ptk one that was mentioned in a previous discussion.  I think this looks fine.

xonsh/ptk/completer.py
```diff
@@ -32,5 +36,25 @@ def get_completions(self, document, complete_event):
                                                      begidx,
                                                      endidx,
                                                      self.ctx)
+            if len(completions) <= 1:
+                pass
+            elif len(os.path.commonprefix(completions)) == len(prefix):
```
adqm: Is there a reason for this check, and not always to do this?  This won't work with the current completer, where not all completions start with the same length substring (some are wrapped in quotes, for example).

docs/envvars.rst
```diff
@@ -63,6 +63,11 @@ applicable.
         writing ``$COMPLETIONS_DISPLAY = None`` and ``$COMPLETIONS_DISPLAY = 'none'`` is equivalent.
         
         (Only usable with SHELL_TYPE=prompt_toolkit)
+    * - COMPLETIONS_MENU_ROWS
+      - ``3``
```
adqm: I might advocate a bigger number than 3, but maybe that's just me.  Maybe we could use 5, as a happy medium between was prompt_toolkit does (7, I think) and this value?

CHANGELOG.rst
```diff
@@ -0,0 +1,70 @@
...
+**Fixed:**
+
+None
+
+**Security:**
+
+None
+
+
+v0.2.1 - v0.2.4
```
adqm: We may actually want to fill this out for these older versions at some point :)

xonsh/ptk/shell.py
```diff
@@ -18,6 +17,7 @@
 from xonsh.ptk.completer import PromptToolkitCompleter
 from xonsh.ptk.history import PromptToolkitHistory
 from xonsh.ptk.key_bindings import load_xonsh_bindings
+from xonsh.ptk.shortcuts import Prompter
 from xonsh.pyghooks import XonshLexer
```
adqm: PEP8: too many blank lines here

xonsh/ptk/shell.py
```diff
@@ -53,17 +54,19 @@ def cmdloop(self, intro=None):
...
+                        mouse_support=mouse_support,
+                        auto_suggest=auto_suggest,
+                        get_prompt_tokens=token_func,
+                        style=style_cls,
+                        completer=completer,
+                        lexer=PygmentsLexer(XonshLexer),
+                        history=self.history,
+                        enable_history_search=True,
+                        reserve_space_for_menu=0,
+                        key_bindings_registry=self.key_bindings_manager.registry,
```
adqm: PEP8: line too long (not sure if we care anymore)

xonsh/ptk/shortcuts.py
```diff
@@ -0,0 +1,451 @@
+"""A prompt-toolkit inspired shortcut collection."""
```
adqm: A whole bunch of PEP8 stuff here, though maybe we don't care if this is mostly from outside xonsh (from forking prompt_toolkit).

CHANGELOG.rst
scopatz: Yeah, for sure.

docs/envvars.rst
scopatz: 5 sounds fine.  I'll bump mine down to 3 :)

xonsh/ptk/completer.py
scopatz: So this is for the case where there is a common beginning to all possible matches that is longer than the typed in prefix.  Pressing tab in these cases will complete the common beginning without bringing up the menu.  Without this check, the space would be reserved without the menu popping up.  I think that if this is changed to a `<=` then this will still work for the cases you bring up.

xonsh/ptk/shell.py
scopatz: I have a change on the `xonfig` branch that fixes this, so maybe we can wait for this one.

xonsh/ptk/shortcuts.py
scopatz: It is mostly from prompt_toolkit stuff that can hopefully be removed relatively soon.  I wouldn't mind passing on this since pretty much the major offenders in the bottom two methods (300+ lines) can all go away on the next prompt-toolkit release.

scopatz:
Thanks for the review @adqm! I think that this is ready for another review.

adqm:
All looks great to me!  Just merged in!

scopatz:
Woohoo! Really happy to see these merged in.

scopatz:
The prompt-toolkit backend is now usable in my opinion :)

Issue #579, opened 2015-12-22
Replace prompt() for ptk with a class


scopatz:
This was first brought up in #578 as a way to asynchronously add to history.

adqm:
Also, I'm not sure we need to make our own wrapper class.  Since `prompt` [wraps `CommandLineInterface`](https://github.com/jonathanslenders/python-prompt-toolkit/blob/41f45c90e37e32bebc1d9e143cb192280a3bc569/prompt_toolkit/shortcuts.py#L463-L466), I'm assuming the thing to do would be to make instances of `CommandLineInterface` and `Application` directly, and presumably that would be give us some more control, access to more variables, etc.

scopatz:
Ahh yeah, using CommandLineInterface directly is probably the best option.

scopatz:
This has been done.

PR #593, opened 2015-12-31
Fix for ptk history scrolling on first prompt

scopatz:
This fixes an issue described in #578 and addresses #579 and jonathanslenders/python-prompt-toolkit#202.  This isn't all that much code, but depends on #592. 

It basically just grabs the buffer and then calls `reset()`.


adqm:
This does seem to fix the history issue.  I think this is fine to merge in alongside #592 when that goes in.

scopatz:
Woohoo Great!

Issue #590, opened 2015-12-30
from module_name import object_name tab-completion


denfromufa:
@scopatz this tab-completion does not work:

from sys import pat<tab here>

Only `import module_name` works.

https://github.com/scopatz/xonsh/issues/364#issuecomment-168070874

scopatz:
We don't currently inspect modules to find their contents.  This should be a relatively easy fix, I would guess.

thomasballinger:
Seems like one could go the IPython route and import the module to find out what names it provides (IPython seems to do this on hitting tab), try to stay safe by only providing completion on modules that have already been imported (this is currently what bpython does for fear of module import side effects - but the user is about to run that code anyway, so this is probably fine). Or do fancy things like AST analysis to guess what objects will be available, which I assume is out of the question.

A separate issue is `from package_name import module_name` which sidesteps these decisions, which would just would leverage existing "knowing what packages exist" infrastructure.

scopatz:
I am pretty sure that we forked some of the IPython code, but don't import the module for the same reasons a bpython.  Fancy AST things are not out of the question if someone wants to implement them. Xonsh is all about the fancy AST things :).  

Thinking out loud, looking for a literal `__all__` would be pretty easy.  If this is not defined, it is pretty easy to accumulate every module-level definition.  This would only apply to pure python and xonsh code, of course. Extension modules would still not be able to be inspected.  

It might be worth looking into the standard library inspect module for tools too. I think that this is what IPython is based on.

asmeurer:
Fixing https://github.com/scopatz/xonsh/issues/344 would also fix this. I would hate for xonsh to end up reimplementing Jedi.

scopatz:
Having jedi as an optional completer would be great if someone wanted to implement it.

denfromufa:
BTW, i used ptpython with autocompletion from jedi and in general it was much smoother experience with installation than xonsh and bpython at least on windows due to python 2 compatibility (yes, it still exists). and it supports jedi already. also built on prompt_toolkit like xonsh, but not bpython.

scopatz:
We have a long-standing issue with Python 2 support, so see the discussion at #66 for more details.  I wil say, though, that the difference between bpython or ptpython and xonsh is pretty significant. Xonsh has its own lexer and parser and targets Python-version specific ASTs. bpython and ptpython can just use whatever interpreter they get. If someone wants to go through the effort of porting to python 2, great!

[...]

PR #597, opened 2015-12-31
Completer Updates

### LINKED ISSUE #575: Duplicate completions provided for subdirectories
Given file: someDocument.txt in Documents/subDir
When I try:
cd Documents/subDir/someDoc<tab>

Instead of completing someDocument.txt, it will display suggestions with and without the parent directory/ies, and not fill in any more of the document name.

(using vanilla xonsh 0.2.3 installed via pip on python3.4 on ubuntu 14.04)

adqm:
A few updates for the completer that should fix some of the issues that are lying around:
* slight efficiency improvement for the known commands cache
* added code to avoid repeated completions (which I think should fix #575)
* implementation of `from module import <tab>` (#590)

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -353,7 +353,11 @@ class Completer(object):
         backslash = '\\'
         double_backslash = '\\\\'
         slash = get_sep()
+        orig_start = start
+        orig_end = end
         for s in paths:
+            start = orig_start
+            end = orig_end
             if (start == '' and
                     (any(i in s for i in CHARACTERS_NEED_QUOTES) or
                      (backslash in s and slash != backslash))):
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -6,6 +6,7 @@ import ast
 import sys
 import shlex
 import pickle
+import inspect
 import builtins
 import subprocess
 
@@ -186,15 +187,25 @@ class Completer(object):
             # anything goes.
             rtn = self.cmd_complete(prefix)
         elif cmd in self.bash_complete_funcs:
+            # bash completions
             rtn = self.bash_complete(prefix, line, begidx, endidx)
             rtn |= self.path_complete(prefix, path_str_start, path_str_end)
             return self._filter_repeats(rtn), lprefix
         elif prefix.startswith('${') or prefix.startswith('@('):
             # python mode explicitly
-            rtn = set()
+            return _python_mode_completions(self, prefix, prefixlow, startswither)
         elif prefix.startswith('-'):
             return sorted(self._man_completer.option_complete(prefix, cmd)), lprefix
         elif cmd not in ctx:
+            ltoks = line.split()
+            if len(ltoks) > 2:
+                if ltoks[0] == 'from' and ltoks[2] == 'import':
+                    # complete thing inside a module
+                    mod = __import__(ltoks[1])
+                    out = sorted(i[0] for i in inspect.getmembers(mod) if i[0].startswith(prefix))
+                    return out, lprefix
+            if len(ltoks) == 2 and ltoks[0] == 'from':
+                return sorted(self.module_complete(prefix)), lprefix
             if cmd == 'import' and begidx == len('import '):
                 # completing module to import
                 return sorted(self.module_complete(prefix)), lprefix
@@ -207,17 +218,21 @@ class Completer(object):
         else:
             # if we're here, we're not a command, but could be anything else
             rtn = set()
-        rtn |= {s for s in XONSH_TOKENS if startswither(s, prefix, prefixlow)}
+        rrn |= _python_mode_completions(self, prefix, prefixlow, startswither)
+        rtn |= {s + space for s in builtins.aliases
+                if startswither(s, prefix, prefixlow)}
+        rtn |= self.path_complete(prefix, path_str_start, path_str_end)
+        return sorted(rtn), lprefix
+
+    def _python_mode_completions(self, prefix, prefixlow, startswither):
+        rtn = {s for s in XONSH_TOKENS if startswither(s, prefix, prefixlow)}
         if ctx is not None:
             if dot in prefix:
                 rtn |= self.attr_complete(prefix, ctx)
             else:
                 rtn |= {s for s in ctx if startswither(s, prefix, prefixlow)}
         rtn |= {s for s in dir(builtins) if startswither(s, prefix, prefixlow)}
-        rtn |= {s + space for s in builtins.aliases
-                if startswither(s, prefix, prefixlow)}
-        rtn |= self.path_complete(prefix, path_str_start, path_str_end)
-        return sorted(rtn), lprefix
+        return rtn
 
     def _canonical_rep(self, x):
         if x.endswith('"') or x.endswith("'"):
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -498,11 +498,11 @@ class Completer(object):
     def _all_commands(self):
         path = builtins.__xonsh_env__.get('PATH', [])
         # did PATH change?
-        path_hash = hash(tuple(path))
+        path_hash = hash(frozenset(path))
         cache_valid = path_hash == self._path_checksum
         self._path_checksum = path_hash
         # did aliases change?
-        al_hash = hash(tuple(sorted(builtins.aliases.keys())))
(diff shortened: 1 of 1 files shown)


xonsh/completer.py
```diff
@@ -200,19 +187,25 @@ def complete(self, prefix, line, begidx, endidx, ctx=None):
...
         elif prefix.startswith('${') or prefix.startswith('@('):
             # python mode explicitly
-            rtn = set()
+            return _python_mode_completions(self, prefix, prefixlow, startswither)
         elif prefix.startswith('-'):
             return sorted(self._man_completer.option_complete(prefix, cmd)), lprefix
         elif cmd not in ctx:
+            ltoks = line.split()
+            if len(ltoks) > 2:
+                if ltoks[0] == 'from' and ltoks[2] == 'import':
```
scopatz: This line can be `and`ed with the line above and save an indentation level :)  Just a suggestion...

xonsh/completer.py
```diff
@@ -200,19 +187,25 @@ def complete(self, prefix, line, begidx, endidx, ctx=None):
...
+                    # complete thing inside a module
+                    mod = __import__(ltoks[1])
```
scopatz: I think that [`__import__()`](https://docs.python.org/3/library/functions.html#__import__) is "discouraged in favor of [`importlib.import_module()`](https://docs.python.org/3/library/importlib.html#importlib.import_module)." It is probably worth following this advice.

xonsh/completer.py
```diff
@@ -200,19 +187,25 @@ def complete(self, prefix, line, begidx, endidx, ctx=None):
...
+                    out = sorted(i[0] for i in inspect.getmembers(mod) if i[0].startswith(prefix))
```
scopatz: `getmembers()` sorts on the name already, so there is no reason to resort.

scopatz: * no need.

xonsh/completer.py
```diff
@@ -225,18 +218,41 @@ def complete(self, prefix, line, begidx, endidx, ctx=None):
...
+        return x
+
+    def _filter_repeats(self, comps):
+        reps = {}
+        for comp in comps:
+            canon = self._canonical_rep(comp)
+            if canon not in reps:
+                reps[canon] = []
+            reps[canon].append(comp)
+        return {max(i, key=len) for i in reps.values()}
```
scopatz: Ahh cool, I didn't know that you could supply a key function to max() and min() to get argmax and argmin behaviour. That is awesome!

xonsh/completer.py
```diff
@@ -317,13 +333,15 @@ def module_complete(self, prefix):
 
     def _quote_paths(self, paths, start, end):
         out = set()
+        quotable = ' (){}-,+*'
```
scopatz: Shouldn't this have `"\t\r\n"` in it too, for truly terrible file names?

scopatz: Also, I think that the `-` is a tad too aggressive.  Many files have dashes in them.  Probably the only characters that are needed in this are whitespace and whatever globbing and expandvars() use as syntax.  Namely, `" \t\r\n${}%*"`

scopatz:
Thanks for all of the fixes here @adqm! In general, I think it looks great. Just a couple, of issues.  The first is about which characters to quote over, which I have already mentioned in-line.  

The other is that this seems to over-quote by quoting non-offenders when there are offenders present.  For example, here are the completions for my home dir.

```console
scopatz@localhost ~ $ ls
Desktop          downloads        pyne            urwid
Downloads        ergs-website     pyne-ci         xe no .sh
allfuncs.sh      fc-bench-method  regolith        xeno.sh
archetype-paper  fc-deploy-opt    sandbox         xo
cycamore         file.txt         staged-recipes  xo no .sh
cyclus           george           temp            xonsh
cyclus-release   miniconda        test.py
cymetric         nltk_data        umdone
scopatz@localhost ~ $ ls xe[TAB]
'xe no .sh'  'xeno.sh'    
scopatz@localhost ~ $ cd pyne[TAB]
'pyne-ci/'  'pyne/'  
```

Note that, `xeno.sh` is quoted because there is a `xe no .sh` (with spaces) present.  Also, the `pyne/` dir is quoted when `pyne-ci/` is present. I expected that only the names with issues would be quoted. Does this make sense?  

Thanks again!

xonsh/completer.py
adqm: Yep.  Works for `sort` and `sorted` as well!

xonsh/completer.py
adqm: Yeah, you're right.  I want to keep parens since I think they lead to syntax errors, but if dashes don't have that problem, then that's fine.  I'll go with ``" `\t\r\n${}%*()"`` for now (including backtick as well since that is used for globbing).  Maybe we should also only quote `%` if we're on Windows, since it isn't used for variable expansion elsewhere.

xonsh/completer.py
adqm: Ah, you learn something new every day.  I had also intended to wrap this in a try/except in case the import fails.

xonsh/completer.py
adqm: Ah, you learn something new every day.  I had also intended to wrap this in a try/except in case the import fails.

adqm:
> Note that, xeno.sh is quoted because there is a xe no .sh (with spaces) present.

Yep, stupid bug.  I know how to fix.

adqm:
Thanks for the suggestions. This should be ready for another look now.

xonsh/completer.py
scopatz: Awesome.

scopatz:
Thanks a ton @adqm!

denfromufa:
some bad things with the master now:

```
[py34] C:\Users\...>xonsh
Exception in thread Thread-2:
  File "c:\python\python27\envs\py34\lib\threading.py", line 920, in _bootstrap_
inner
    self.run()
  File "c:\python\python27\envs\py34\lib\site-packages\xonsh\ptk\history.py", li
ne 71, in run
    buf = self._buf()
  File "c:\python\python27\envs\py34\lib\site-packages\xonsh\ptk\history.py", li
ne 81, in _buf
    buf = builtins.__xonsh_shell__.shell.prompter.cli.application.buffer
AttributeError: 'NoneType' object has no attribute 'application'

-DENIS ~ # import sys
-DENIS ~ # 1+1
2
-DENIS ~ # from math import siException in thread Thread-58
:
  File "c:\python\python27\envs\py34\lib\threading.py", line 920, in _bootstrap_
inner
    self.run()
  File "c:\python\python27\envs\py34\lib\threading.py", line 868, in run
  File "c:\python\python27\envs\py34\lib\site-packages\prompt_toolkit\interface.
py", line 647, in run
    completions = list(buffer.completer.get_completions(document, complete_event
))
  File "c:\python\python27\envs\py34\lib\site-packages\xonsh\ptk\completer.py",
line 39, in get_completions
    if len(completions) <= 1:
TypeError: object of type 'generator' has no len()

-DENIS ~ # from math import si
ImportError: cannot import name 'si'
-DENIS ~ # from math import sin
-DENIS ~ #
```

gforsyth:
@denfromufa -- I have the same error -- i found the culprit.  Line 215 in `completer.py` returns a generator instead of a list of strings.  Fixing now and submitting a PR

scopatz:
Thanks for reporting @denfromufa.

denfromufa:
There is second error as you can see above:

```
AttributeError: 'NoneType' object has no attribute 'application'
```

scopatz:
What version of prompt_toolkit are you on? You probably need to be on the latest (v0.54) for the moment.  Unfortunately prompt-toolkit does not yet have a release with version numbers jonathanslenders/python-prompt-toolkit#205, so this can be a little annoying to look up.  Also, please feel free to move this conversation to a separate issue.

denfromufa:
```
blessings (1.6)
bpython (unknown)
curses (2.2)
curtsies (0.1.21)
docopt (0.6.2)
greenlet (0.4.9)
jedi (0.9.0)
pip (7.1.2)
ply (3.8)
prompt-toolkit (0.54)
ptpython (0.25)
Pygments (2.0.2)
pyreadline (2.1)
requests (2.9.1)
setuptools (18.5)
six (1.10.0)
wcwidth (0.1.5)
wheel (0.26.0)
xonsh (0.2.4)
```

PR #598, opened 2016-01-02
Removal of dollar from completions

scopatz:
In the new completer, anything with a dollar sign will be quoted upon tab-completion. This is breaks Python-mode evaluation of variable lookup (and isn't really needed in subproc mode).  

```console
scopatz@localhost ~ $ $XONSH_[tab]
'$XONSH_INTERACTIVE'     '$XONSH_STORE_STDOUT'
'$XONSH_SHOW_TRACEBACK'  '$XONSH_VERSION'
```

These are all then evaluated as the literal string, rather than the envvar lookup.  This removed the dollar sign from the quotable characters.  I wouldn't mind if it went back in, pending a fix to this issue.

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -65,7 +65,7 @@ XONSH_TOKENS = {
     '...'
 }
 
-CHARACTERS_NEED_QUOTES = ' `\t\r\n${}*()'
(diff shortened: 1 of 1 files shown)


adqm:
I think #599 should take care of this, while still quoting paths that contain dollar signs.

PR #599, opened 2016-01-02
Fix for Dollar Signs In Path Completion

adqm:
This should keep the same quoting semantics for path names, but should avoid quoting environment variables.


scopatz:
Thanks a ton!

Issue #507, opened 2015-11-17
Multiline statements in prompt_toolkit


gforsyth:
This should supercede #467 as multiline code does not indent on prompt_toolkit on any OS, as it isn't yet implemented.  

This is actually two issues, but I believe they're linked closely enough that they need to be dealt with together.  If we want to be able to edit multiline statements in PT, then the prompt should always be a `multiline=True` prompt and will require a custom keybinding to determine when to execute vs. when to create a newline, as mentioned by @jonathanslenders in [here](https://github.com/scopatz/xonsh/pull/490#issuecomment-156034024).  

Since multiline prompts don't do indentation automatically, we also need to use something like `postcmd` found in [`readline_shell.py`](https://github.com/scopatz/xonsh/blob/master/xonsh/readline_shell.py#L129) to handle indenting and dedenting as needed.

In terms of when to execute vs. add a new-line, one criteria for adding a newline is if there's a non-closed parens statement.  Beyond that, what do we need?  We can follow xonsh's current behavior, I think, and always add a newline unless the previous line is only `\n`, i.e. a double-return will execute the statement.

gforsyth:
Ah, one more thing -- as per @scopatz [comment](https://github.com/scopatz/xonsh/pull/490#issuecomment-156713074) -- if the multiline prompt input is properly formatted, we can pass the entire line to `BaseShell.default()` and it should execute without issue.

krader1961:
> In terms of when to execute vs. add a new-line, one criteria for adding a newline is if there's a non-closed parens statement. Beyond that, what do we need? We can follow xonsh's current behavior, I think, and always add a newline unless the previous line is only `\n`, i.e. a double-return will execute the statement.

I definitely want to see multiline statements handled by PT so that they can be edited as a block. But how to do this given the hybrid nature of xonsh is unclear.

In addition to the non-closed parenthesized sequence (by which I assume you mean `$(...)`) we also want to continue accepting input if the previous line ends with a bare backslash since that means the statement is continued whether we're in shell or python mode. Beyond that if we've determined we're in shell syntax mode then a single newline should be sufficient. If we're in python syntax mode then a double-return should be required (and matches the iPython behavior) if the text so far is incomplete or part of a `def`, `class`, or similar block.

Would looking at iPython and similar programs for ideas help?

scopatz:
> Would looking at iPython and similar programs for ideas help?

[...]

PR #601, opened 2016-01-05
"True" multiline in PTK (WIP)

gforsyth:
I'm not sure this is done but I wanted to put it in front of a few more eyes to see if there's anything big that's missing.  This is my crack at a more robust multiline environment in PTK, as per #507.  

It remaps `Enter` to run through a preliminary parser that tries to determine whether to send the current buffer off for evaluation, or to insert a newline instead.  

Current 'triggers' for inserting a newline are:
- Not on first line of buffer and line is non-empty
- Previous character is a colon (covers if, for, etc...) 
- User is in an open paren-block
- Line ends with backslash
- Any text exists below cursor position (relevant when editing previous multiline blocks)

(Anything big missing here?)

If the character before the cursor is a colon, it also indents the following line -- otherwise a newline maintains the current indent level.

I tweaked the `ptk/completer.py` to work with multiline prompts.

Up arrow will move up in history if the cursor is on the first line, otherwise it moves up a line within the current document buffer.  

`xonsh` already saves multiline inputs in its rich history format, but if you open a new terminal window, the up arrow won't bring up the multiline buffer in history.  Not totally sure what to do about this.

### DIFF
--- a/xonsh/ptk/key_bindings.py
+++ b/xonsh/ptk/key_bindings.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 """Key bindings for prompt_toolkit xonsh shell."""
-import re
 import builtins
 
 from prompt_toolkit.filters import Filter, IsMultiline
@@ -38,7 +37,6 @@ def load_xonsh_bindings(key_bindings_manager):
     handle = key_bindings_manager.registry.add_binding
     env = builtins.__xonsh_env__
     indent_ = env.get('INDENT')
-    parens = re.compile('["\'].*[\\(\\)]*.*["\']')
 
     DEDENT_TOKENS = frozenset(['raise', 'return', 'pass', 'break', 'continue'])
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ Current Developments
 * timeit alias will now complete its arguments.
 * $COMPLETIONS_MENU_ROWS environment variable controls the size of the 
   tab-completion menu in prompt-toolkit.
+* Prompt-toolkit shell now supports true multiline input with the ability
+  to scroll up and down in the prompt.
 
 **Changed:**
 


scopatz:

> xonsh already saves multiline inputs in its rich history format, but if you open a new terminal window, the up arrow won't bring up the multiline buffer in history. Not totally sure what to do about this.

I had purposefully split the input [here as you probably know](https://github.com/scopatz/xonsh/blob/master/xonsh/ptk/history.py#L64) since I knew that multi-line editing was not supported.  It is fine to rework and remove this logic given true multiline editing.

xonsh/ptk/key_bindings.py
```diff
@@ -41,3 +41,33 @@ def insert_literal_tab(event):
...
+        - Any text exists below cursor position (relevant when editing previous
+        multiline blocks)
+        """
+        b = event.cli.current_buffer
+        if (not b.document.on_first_line and
+           not b.document.current_line.isspace()):
+            b.newline(copy_margin=True)
+        elif b.document.char_before_cursor == ':':
+            b.newline()
+            b.insert_text(env.get('INDENT'), fire_event=False)
```
scopatz: This doesn't block since it can be added later, but there is no associated dedent behaviour here.  The readline shell does this with the logic that can be found here: https://github.com/scopatz/xonsh/blob/master/xonsh/readline_shell.py#L133

adqm:
It will be fantastic to have this!  Thanks for your work here.

I haven't had a chance to test very thoroughly, but a couple of quick issues:

One issue with the autoindenting:
```
hartz@lappy-513 ~ $ if True:
                        print(8) # autoindents fine
                    else:
                    print(9) # does not autoindent
```
Then pressing Enter a bunch from there just keeps adding newlines (though I'm not sure which of the conditions you mentioned is getting "hit").

Ah, it seems to be a problem with any "dedenting" at all.  For example, the following won't go, either (hitting Enter keeps adding newlines):
```
hartz@lappy-513 ~ $ if True:
                        print(8)
                    print(9)
                    
                    
                    

```

Also, tab complete doesn't seem to work for me on this branch (on any line of the input).

Also, is it possible/feasible to add the `MULTILINE_PROMPT` in?  There is a whole lot of empty space on the left, which messes with my head.

xonsh/ptk/key_bindings.py
```diff
@@ -41,3 +41,33 @@ def insert_literal_tab(event):
...
+        elif b.document.text.count('(') > b.document.text.count(')'):
```
scopatz: I don't think that this is safe on its own. What if someone has a string with a single open parentheses, like `print('(')`.  BaseShell() will set the `self.need_more_lines` boolean attribute in the event of SyntaxError on `push()`.  However, I don't think that the prompt-toolkit method uses the `push()` method.  Also the `push()` method is for gluing together independent lines, so is probably not appropriate here.

Rather, you probably should do what the `push()` method does and see it the current code can be parsed, [like here](https://github.com/scopatz/xonsh/blob/master/xonsh/base_shell.py#L165).  Namely,

```python
mode = 'single' if b.document.text.count('\n') == 0 else 'exec'
try:
      code = self.execer.compile(b.document.text, mode=mode,
                                       glbs=None,
                                       locs=self.ctx)
except SyntaxError:
   ...
```

xonsh/ptk/key_bindings.py
```diff
@@ -41,3 +41,33 @@ def insert_literal_tab(event):
...
+            b.newline()
+        elif b.document.char_before_cursor == '\\':
+            b.newline()
+        elif b.document.find_next_word_beginning() is not None:
+            b.newline(copy_margin=True)
```
scopatz: It would be nice for the `$MULTILINE_PROMPT` to show up, but again, that doesn't block.

scopatz:

xonsh/ptk/key_bindings.py
gforsyth: Ahh, hadn't thought of that case -- I'll try out the `push()` approach.

xonsh/ptk/key_bindings.py
gforsyth: Yeah -- that was one of the things besides history I didn't have working -- I've been digging through the PTK source to figure that bit out but haven't got it yet.  It *may* have to wait until we switch out `prompt` for a more customized `prompt_application` but I haven't given up on that yet.

scopatz: No worries!

gforsyth:
I'll say from the start that this is ugly.  I couldn't come up with a way to be able to use the same ideas as in `push()` because I'm stuck in this keystroke event and the only way I can make it work is to keep things within the same conditional tree.  

I am super open to suggestions because this is hard to read.  That said, I think I've fixed all of the indent behaviors, added `DEDENT_TOKENS` and handled the parentheses edge cases.  

Still need to fix the history and completion.  Should get to that later tonight or early tomorrow (history is easy, completions.. not sure yet).

gforsyth:
ping @adqm @scopatz 
I think this is ready for another look.

xonsh/ptk/key_bindings.py
```diff
@@ -41,3 +45,52 @@ def insert_literal_tab(event):
...
+        # if there are stray parens hanging aroung inside quotes
+        elif (re.search(parens, b.document.text) is not None and
+              ((b.document.text.count('(') -
+               re.search(parens, b.document.text).group().count('(')) >
+              (b.document.text.count(')') -
+               re.search(parens, b.document.text).group().count(')')))):
+            b.newline()
+        # and if there aren't stray parens hanging around inside quotes
+        elif (re.search(parens, b.document.text) is None and
+              b.document.text.count('(') > b.document.text.count(')')):
```
scopatz: Thanks for trying this @gforsyth. I still think that this is not general enough.  Really the best way is to just do a syntax check with the execer itself.  I have opened a PR to your fork, gforsyth/xonsh#1, that implements my vision for how this should work.  Merge it in if it looks good to you!

scopatz:
Thanks @gforsyth! Also, if you want to update the CHANGELOG with these additions that'd be awesome.  I am super excited to see this feature make it in.

xonsh/ptk/key_bindings.py
```diff
@@ -19,12 +20,27 @@ def __call__(self, cli):
...
+
+
 def load_xonsh_bindings(key_bindings_manager):
     """
     Load custom key bindings.
     """
     handle = key_bindings_manager.registry.add_binding
     env = builtins.__xonsh_env__
+    indent_ = env.get('INDENT')
+    parens = re.compile('["\'].*[\\(\\)]*.*["\']')
```
scopatz: the parens variable does not seem to be used.

scopatz:
This looks good to me subject to the couple of minor changes.

gforsyth:
@scopatz -- Removed the regex and updated the changelog.  Let me know if it needs anything else!

scopatz:
Thanks a million @gforsyth!

PR #603, opened 2016-01-05
fix module import to return list instead of generator

gforsyth:
Reported by @denfromufa in #597 -- a quick fix since the `get_completions` methods expect a list of strings (and then themselves return a generator)


scopatz:
Thanks a ton @gforsyth!

PR #605, opened 2016-01-05
made obtaining buffer in history thread safe.

scopatz:
This only affected prompt-toolkit.

### DIFF
--- a/xonsh/ptk/history.py
+++ b/xonsh/ptk/history.py
@@ -77,9 +77,10 @@ class PromptToolkitHistoryAdder(Thread):
                 continue
 
     def _buf(self):
-        if hasattr(builtins, '__xonsh_shell__'):
-            buf = builtins.__xonsh_shell__.shell.prompter.cli.application.buffer
-        else:
-            buf = None
+        # Thread-safe version of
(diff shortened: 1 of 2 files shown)


xonsh/ptk/history.py
```diff
@@ -77,9 +77,10 @@ def run(self):
                 continue
 
     def _buf(self):
-        if hasattr(builtins, '__xonsh_shell__'):
-            buf = builtins.__xonsh_shell__.shell.prompter.cli.application.buffer
-        else:
-            buf = None
+        # Thread-safe version of
+        # buf = builtins.__xonsh_shell__.shell.prompter.cli.application.buffer
+        buf = getattr(getattr(getattr(getattr(getattr(getattr(builtins,
```
adqm: Instead of chaining these together, could we loop over these things?

```
buf = builtins
for i in ['__xonsh_shell__', 'shell', 'prompter', 'cli', 'application', 'buffer']:
    buf = getattr(buf, i, None)
    if buf is None:
        break
return buff
```

It might not be as efficient in the average case (actually, I'm not sure), but it feels cleaner to me.

If you disagree, I'm happy to merge in as-is.

scopatz:
Thanks for the review @adqm!  I have updated the PR.

adqm:

PR #607, opened 2016-01-06
Xonfig

scopatz:
As mentioned by John Lunzer on [the mailing list here](https://groups.google.com/d/msg/xonsh/23MTNWhl4oo/m9ssxGUNCwAJ), xonsh lacks an initial configuration script.  This is important not only for the config file it creates, but it also gives new users a sense of what can be done.  

Xonsh has also lacked an easy way for users to report the basic aspects of their system.  

This pull request accomplishes both of these via a new `xonfig` alias.  System reporting is done through the `xonfig info` or just `xonfig` commands.  For example, my system reports:

```console
scopatz@localhost ~ $ xonfig 
+---------------------+-------+
| xonsh               | 0.2.4 |
| Python              | 3.5.1 |
| PLY                 | 3.7   |
| have readline       | True  |
| have prompt toolkit | True  |
| on posix            | True  |
| on linux            | True  |
| on arch             | False |
| on windows          | False |
| on mac              | False |
| are root user       | False |
| default encoding    | utf-8 |
+---------------------+-------+
```

The configuration script, or `xonfig wizard`, was more in depth to accomplish.  Rather than going with a traditional script, this implementation uses a tree of wizard-relevant nodes.  Two main visitor classes have been implemented which make use of these nodes:

1. a pretty formatter
2. a command-line, prompt-based visitor

The xonfig command then uses these to construct a wizard that can load, modify, and save static configuration files.  

The reason for going to this level of abstraction is that alternative front-end visitors are likely desirable. For example, fish has a web-based configuration (that looks really nice), and on Linux we could have a command line version that uses urwid instead of the shell prompt.  These will presumably come later, at which point custom visitors are all that is needed.

Furthermore, the wizard components (nodes and visitors) should be fully reusable.  This means that if we have a need for sophisticated user prompts in the future, we'll have a mechanism at the ready.

Additionally, to make this work, programmatic access to environment variable documentation was needed. The docs where moved into `xonsh.environ` itself. I think this is a good move overall, as it makes the docs more maintainable.


scopatz:
Is anyone able to take a look at this, please? I know it is kind of a doozy :)

adqm:
I think I should be able to take a detailed look tomorrow.

scopatz:
Thanks @adqm!

gforsyth:
Trying to run the wizard throws a Unicode error -- same behavior with both `readline` and `prompt_toolkit`.

```
Could not find xonsh configuration file.
Would you like to run the xonsh configuration wizard now, yes or no (default)? yes
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/proc.py", line 338, in wrapped_simple_command
    r = f(args, i)
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/aliases.py", line 168, in xonfig
    return main(args)
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/xonfig.py", line 268, in main
    return _MAIN_ACTIONS[ns.action](ns)
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/xonfig.py", line 193, in _wizard
    pv.visit()
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/wizard.py", line 462, in visit
    rtn = super().visit(node)
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/wizard.py", line 286, in visit
    rtn = meth(node)
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/wizard.py", line 514, in visit_question
    self.visit(node.responses[r])
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/wizard.py", line 462, in visit
    rtn = super().visit(node)
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/wizard.py", line 286, in visit
    rtn = meth(node)
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/wizard.py", line 501, in visit_wizard
    self.visit(child)
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/wizard.py", line 462, in visit
    rtn = super().visit(node)
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/wizard.py", line 286, in visit
    rtn = meth(node)
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/wizard.py", line 507, in visit_message
    print_color(node.message)
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/tools.py", line 829, in print_color
    print(format_color(string), file=file)
UnicodeEncodeError: 'ascii' codec can't encode character '\xb7' in position 800: ordinal not in range(128)
```


Typing `no` works fine and then `xonfig` reports


```
+---------------------+-------+
| xonsh               | 0.2.4 |
| Python              | 3.5.1 |
| PLY                 | 3.7   |
| have readline       | True  |
| have prompt toolkit | True  |
| on posix            | True  |
| on linux            | True  |
| on arch             | True  |
| on windows          | False |
| on mac              | False |
| are root user       | False |
| default encoding    | utf-8 |
+---------------------+-------+
```

gforsyth:
Also, on `prompt_toolkit` only, if the term window isn't wide enough, cursor position is wonked.

cursor_pos

scopatz:
Hi @gforsyth, Just curious about what the native encoding of your terminal is.  Is it ASCII?

As for the weird issues with the prompt being too small, I would bet that is a prompt-toolkit bug or that we are using it wrong somehow.  I'll see if I can reproduce.

gforsyth:
Yeah, that's embarrassing.  That's on me -- corruption of my locale somewhere along the line.

As for the ptk bug, I think this may help [patch_stdout_proxy](https://github.com/jonathanslenders/python-prompt-toolkit/blob/d2deee66f364aa5245eacb790153e8637bfa7b59/prompt_toolkit/interface.py#L781)

adqm:
Just starting looking at this now.  For starters, I am seeing the same issue as @gforsyth with the weird text wrap.
virtualbox_debian3_08_01_2016_17_39_02

If the terminal is wide enough, then the wrapping isn't an issue.

adqm:
Also, more text wrapping issues:
virtualbox_debian3_08_01_2016_17_42_36

adqm:
Should we run the ensurers on people's entries to the wizard?  This is perfectly fine, it seems, with letting me set, for example, ``$AUTO_SUGGEST`` to ``7``.

adqm:
It took me a little while to figure out what this meant:
```
Would you like to save this state, yes or no (default)? 
```

Can we change these to actually show the default value?  For example,
```
Would you like to save this state, yes or no (default: no)? 
```

adqm:
Should we also skip automatically asking about the wizard in the case where there is a `.xonshrc`, in addition to the check for there being a `.config/xonsh/...`?

I use a `.xonshrc` exclusively, and I don't really want to be asked about the JSON config every time.

gforsyth:
I think it may also benefit users to have an option to just "Accept" the default for a given envvar.  It's not a big deal to type `True` a couple of times, but `$PATH` and `$BASH_COMPLETIONS` are sort of begging for typos.

xonsh/base_shell.py
```diff
@@ -30,6 +30,7 @@ def close(self):
 
     def write(self, data):
         """Writes data to the original stdout and the buffer."""
+        data = data.replace('\001', '').replace('\002', '')
```
adqm: Out of curiosity, why this addition?  Might there be a better place for this?  Maybe not...

adqm:
@gforsyth, my understanding (and my hope!) is that just hitting enter without entering will accept the default (by virtue of not putting an entry in the config, xonsh should fall back to the default value when looking those things up).

gforsyth:
@adqm Ahh -- that makes sense.

gforsyth:
The line-wrapping problems in PTK go away for me if I disable the multiline environment.  A potential quick and dirty fix could be to set up another application that doesn't have completions, mouse support, multiline, etc... just for grabbing input for the wizard.

scopatz:
Thanks for the feedback @adqm and @gforsyth!  I have tried to address most of the issues that were brought up.

> Can we change these to actually show the default value?

done.

> Should we also skip automatically asking about the wizard in the case where there is a .xonshrc, in addition to the check for there being a .config/xonsh/...?

Yep, done!

> I think it may also benefit users to have an option to just "Accept" the default for a given envvar.

@adqm is right that just hitting enter accepts the default without actually setting anything.  I have edited the text to hopefully make this more clear.

> The line-wrapping problems in PTK go away for me if I disable the multiline environment.

Thanks for the idea for the fix @gforsyth. I implemented it and it seems to work. @adqm, would you mind checking if it solved the resize wrapping issue?  Hopeully it does.

> Should we run the ensurers on people's entries to the wizard? This is perfectly fine, it seems, with letting me set, for example, $AUTO_SUGGEST to 7.

So we actually do one better here.  We run whatever the ensurer's convert method is.  Thus, 7 is a valid entry for autosuggest since `bool(7) is True`.  The JSON that is generated will store the True and not the 7 value.

> So \xb7 is 'middle dot' and I think the culprit is in xonfig.py

I removed the middle dot, just-in-case.  This seemed like it was too easy to hit.

xonsh/base_shell.py
scopatz: This was put in because if `$XONSH_STORE_STDOUT is True` in the readline shell, the literal `\001` and `\002` would be printed.  It is probably worth rethinking all of the color management stuff at some point to only insert these characters for the readline shell when needed.  I think that this is the only place they are relevant and we end up removing them in a lot of other places.  Another PR though :)

scopatz:
I think that this is ready for another look.

gforsyth:
I ran this with `xonsh --no-rc` to trigger the `xonfig` wizard and I got

```
> xonsh --no-rc
  File "/home/gil/miniconda3/envs/xonshtesting/bin/xonsh", line 9, in <module>
    load_entry_point('xonsh==0.2.4', 'console_scripts', 'xonsh')()
  File "/home/gil/miniconda3/envs/xonshtesting/lib/python3.4/site-packages/xonsh/main.py", line 187, in main
    if not any([env['LOADED_CONFIG']] + env['LOADED_RC_FILES']):
TypeError: can only concatenate list (not "tuple") to list
```

Same deal if I remove my `.xonshrc` -- issue seems to be that `env['LOADED_RC_FILES']` is returning an empty tuple.

It works on my end when I do 

```
if not any([env['LOADED_CONFIG']] + list(env['LOADED_RC_FILES'])):
```

scopatz:
Thanks for pointing this out @gforsyth!  I think I finally have a fix for this. In the process I discovered another issue that if you set the `--config-path` at the command line, the run control files would not be loaded. I fixed that too.

adqm:
Another issue just found: if I hit Control-C when in the wizard, I get caught in an inescapable loop of answering the same question I just answered.  This also brings back the glitch where newlines don't seem to be handled properly.

adqm:
> So we actually do one better here. We run whatever the ensurer's convert method is. Thus, 7 is a valid entry for autosuggest since bool(7) is True. The JSON that is generated will store the True and not the 7 value.

That is a good call.  Maybe this is overkill, but can we offer a prompt like "You entered 7, but this value should be True or False.  Assuming True.  Is this okay?"  Or at least a message that lets the user know that what is being stored is not what was actually entered?  For example, "Converting value 7 to True."

docs/envvars.rst
```diff
@@ -4,241 +4,4 @@ The following table displays information about the environment variables that
...
-    * - PROMPT_TOOLKIT_STYLES
-      - ``None``
-      - This is a mapping of user-specified styles for prompt-toolkit. See the 
-        prompt-toolkit documentation for more details. If None, this is skipped.
-    * - PUSHD_MINUS
-      - ``False``
-      - Flag for directory pushing functionality. False is the normal behaviour.
-    * - PUSHD_SILENT
-      - ``False``
-      - Whether or not to supress directory stack manipulation output.
```
gforsyth: Typo here.  supress -> suppress

xonsh/environ.py
```diff
@@ -199,6 +207,239 @@ class DefaultNotGivenType(object):
...
+        "'xonsh.tools._PT_COLORS'.", configurable=False), 
+    'PROMPT_TOOLKIT_STYLES': VarDocs(
+        'This is a mapping of user-specified styles for prompt-toolkit. See '
+        'the prompt-toolkit documentation for more details. If None, this is '
+        'skipped.', configurable=False),
+    'PUSHD_MINUS': VarDocs(
+        'Flag for directory pushing functionality. False is the normal '
+        'behaviour.'),
+    'PUSHD_SILENT': VarDocs(
+        'Whether or not to supress directory stack manipulation output.'),
```
gforsyth: Sorry -- line note in the right place now...

Typo: supress -> suppress

gforsyth:
Ok -- everything seems to work now on my end -- both readline and PTK shells behave re: line-wrapping and the wizard doesn't prompt if my `.xonshrc` is in place.  

While looking over the docs being printed in the wizard I spotted a number of typos.  I started to add line-notes, but I think that's a ridiculous way to point them out and it doesn't belong in this PR.  After this is merged I'll do an editing pass and fix them up and put my liberal arts education to good use.

scopatz:
OK! I think that this is ready for another look.  

The "infinite loop" was actually caused by $PROMPT not being replaced after a ^C and ^D events.  I added some exception handling to manage this. 

I also added conversion messages for when the conversion message is meaningfully changed. (ie 'y' will print a message about conversion to True, but 'True' will not).  I personally feel that asking for confirmation at this point is too chatty. The risk, of course, is that someone would mistype something and have to start to the xonfig wizard over again.  

> After this is merged I'll do an editing pass and fix them up and put my liberal arts education to good use.

@gforsyth haha, all apologies on my end. I really should get around to adding that spell checker to my text editor :).

adqm:
Just took another look.  Some more thoughts:

Ctrl-C working now.

Is it intended that my answers to the wizard show up in the normal xonsh history after exiting the wizard?  When I hit the up arrow after exiting the wizard, but it seems like they should be kept separate.

``$OLDPWD`` probably shouldn't be asked, as it isn't really a configuration option.

The display for the current (default) value of `$PROMPT` is a little bit strange.  It is technically correct, but I find the way it is displayed a little confusing.
```
default value: xonsh.environ.DEFAULT_PROMPT
current value: 
'{BOLD_GREEN}{user}@{hostname}{BOLD_BLUE} {short_cwd} '
'{BOLD_BLUE}{prompt_end}{NO_COLOR} '
```

The display for the current value of `$TITLE` shows the _actual_ current title, not the template.  This could be a bit misleading (could lead one to believe that the title will _always_ be that value unless something is entered here).
```
$TITLE
The title text for the window in which xonsh is running. Formatted in
the same manner as $PROMPT, see 'Customizing the Prompt'
http://xonsh.org/tutorial.html#customizing-the-prompt.
default value: xonsh.environ.DEFAULT_TITLE
current value: 'hartz@virtual1: ~ | xonsh'
```

It is also a little weird that, while the "current" value has quotes around it, entering a value with quotes will produce literal quote characters.

scopatz:
OK! I think that I have addressed the latest issues @adqm, and this is ready for another look.

> Is it intended that my answers to the wizard show up in the normal xonsh history after exiting the wizard? When I hit the up arrow after exiting the wizard, but it seems like they should be kept separate.

Implemented the ability to disable storing an input line in memory.

> $OLDPWD probably shouldn't be asked, as it isn't really a configuration option.

Whoops missed that. fixed.

> The display for the current (default) value of $PROMPT is a little bit strange. It is technically correct, but I find the way it is displayed a little confusing.

So this is just what `pprint.pformat()` will pump out. I have added some explanatory text to this end.  I think that this is best, because otherwise the shell would wrap the lines and this is often *super* hard to read in my opinion, because it splits on whatever character happens to be at the terminal width. With pformat it at least tries to make it human readable. 

> The display for the current value of $TITLE shows the actual current title, not the template.

My bad. Fixed.

> It is also a little weird that, while the "current" value has quotes around it, entering a value with quotes will produce literal quote characters.

This is because it is showing the repr string a la pformat. I think that it has to be this way. If we didn't show it this way things like $INDENT would be invisible and $PROMPT would look like it is missing the trailing whitespace.

scopatz:
Any other thoughts on this @adqm or @gforsyth?

gforsyth:
This looks good from my end.

adqm:
Looks good to me, too.  Merging in! :)

scopatz:
Woohoo!

PR #608, opened 2016-01-07
fix buffer reset in ptk history

gforsyth:
If this is only `buf.reset()`, it's possible to begin entering text in a
new terminal window which is then cleared shortly after startup
(depending, I imagine, on how much history is being loaded and how long
that takes).  Passing in the current document as `initial_document`
fixes this and history is still accessible as before.


scopatz:
Thanks @gforsyth!

adqm:
Fantastic!  This was driving me crazy :)

Issue #610, opened 2016-01-08
`name in __xonsh_env__` always evaluates to True


riccardomurri:
Containment test `name in __xonsh_env__` always evaluates to Python
``True``, even if no key `name` is defined in `__xonsh_env__`.

The reason seems to be that `Mapping.__contains__` (inherited by `Env`
through `MutableMapping`), relies on `self[key]` to throw a `KeyError`
if a key is not defined::

    426     def __contains__(self, key):
    427         try:
    428             self[key]
    429         except KeyError:
    430             return False
    431         else:
    432             return True

However, `Env.__getitem__()` is basically an alias for `Env.get()` and
will return ``None`` when asked for a non-existing key.

Trying to fix this by having `Env.__getitem__()` raise an exception
when a key is not found makes xonsh crash when trying to execute a
command in subprocess mode::

    $ echo $FOO
    Traceback (most recent call last):
      File "/home/rmurri/sw/src/xonsh/xonsh/base_shell.py", line 136, in default
        self.execer.exec(code, mode='single', glbs=self.ctx)  # no locals
      File "/home/rmurri/sw/src/xonsh/xonsh/execer.py", line 127, in exec
        return exec(code, glbs, locs)
      File "<xonsh-code>", line 1, in <module>
      File "/home/rmurri/sw/src/xonsh/xonsh/environ.py", line 293, in __getitem__
        raise KeyError
    KeyError

This shows IMHO that in the code calls to `Env.__getitem__` and `Env.get`
are used interchangeably thought they should not be.  Unfortunately,
the traceback makes it impossible to understand where I should start,
so I'm opening this issue now :-)

scopatz:
Thanks for reporting @riccardomurri! #611 should address this once it is merged in.

scopatz:
@riccardomurri could you please confirm that this works for you and close. Thanks!

riccardomurri:
I've checked branch `env_contains` and it works.

Thanks!

adqm:
Fantastic!  Closing now.  Please feel free to reopen if you feel like there is still something missing.

PR #611, opened 2016-01-08
change behavior of environment get and __getitem__

adqm:
I believe this should fix the issue from #610, by changing the way `__getitem__` and `get` work for `__xonsh_env__`.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -328,6 +328,7 @@ class Env(MutableMapping):
 
     def __iter__(self):
         yield from self._d
+        yield from self.defaults
 
     def __len__(self):
         return len(self._d)
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -301,7 +301,8 @@ class Env(MutableMapping):
             if is_callable_default(val):
                 val = val(self)
         else:
-            raise KeyError()
+            e = "Unknown environment variable: ${}"
+            raise KeyError(e.format(key))
         if isinstance(val, (MutableSet, MutableSequence, MutableMapping)):
             self._detyped = None
         return val
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -328,8 +328,7 @@ class Env(MutableMapping):
             return default
 
     def __iter__(self):
-        yield from self._d
-        yield from self.defaults
+        yield from (set(self._d) | set(self.defaults))
 
     def __len__(self):
         return len(self._d)
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -286,7 +286,25 @@ class Env(MutableMapping):
     #
 
     def __getitem__(self, key):
-        return self.get(key)
+        m = self._arg_regex.match(key)
+        if (m is not None) and (key not in self._d) and ('ARGS' in self._d):
+            args = self._d['ARGS']
+            ix = int(m.group(1))
+            if ix >= len(args):
+                e = "Not enough arguments given to access ARG{0}."
+                raise KeyError(e.format(ix))
+            val = self._d['ARGS'][ix]
+        elif key in self._d:
+            val = self._d[key]
+        elif key in self.defaults:
+            val = self.defaults[key]
+            if is_callable_default(val):
+                val = val(self)
+        else:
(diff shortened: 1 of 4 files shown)


scopatz:
Hi @adqm, Thanks for putting this in.  This doesn't seem to have the right containment behaviour:

```console
scopatz@localhost ~ $ 'AUTO_SUGGEST' in set(__xonsh_env__.keys())
False
scopatz@localhost ~ $ 'AUTO_SUGGEST' in __xonsh_env__
True
```

I think a simpler solution would be to just override the `__contains__()` method.  

Also this PR Causes lookup failures that should probably be None for not existent environment variables. 

```console
scopatz@localhost ~ $ $DOES_NOT_EXIST
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/base_shell.py", line 136, in default
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/execer.py", line 127, in exec
    return exec(code, glbs, locs)
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/environ.py", line 304, in __getitem__
    raise KeyError()
KeyError
```

The fix is to either revert to the previous version or to change the parser to use the `get()` method for lookup rather than `__getitem__()`.  If we change to the `get()` method, the logic for finding the default should probably be moved to `get()`, allowing `__getitem__()` to fail if the key is not present.

scopatz:
I'm up for whatever you think is best ;)

adqm:
> This doesn't seem to have the right containment behaviour:

That should be fixed with this most recent push.  My thought was that I wanted anything in `_d` or in `defaults` to be treated as being `in` the environment (anything with a value).  If we want, I can add a separate check for things that were explicitly set by the user.

That said, I temporarily broke a bunch of other things.  I'll come back to this one.

adqm:
Okay, @scopatz, I think I made all the updates for what felt right to me (though I'm open to other ideas).

What I wanted:
* Keys in `_d` _and_ in `defaults` are both considered to be `in` the environment (feels like anything with a value should be considered to be `in`)
* In Python mode, lookups that fail raise the exception
* In subprocess mode, lookups that fail are replaced with an empty string

For example:
```
hartz@virtual1 ~ $ $FOO
KeyError: 'Unknown environment variable: $FOO'
hartz@virtual1 ~ $ echo $FOO

hartz@virtual1 ~ $ echo "$FOO"
$FOO
hartz@virtual1 ~ $ $FOO is None
KeyError: 'Unknown environment variable: $FOO'
hartz@virtual1 ~ $ $FOO = 7
hartz@virtual1 ~ $ $FOO
7
hartz@virtual1 ~ $ echo $FOO
7
hartz@virtual1 ~ $ echo "$FOO"
7
hartz@virtual1 ~ $ $FOO is None
False
```

xonsh/environ.py
```diff
@@ -286,45 +286,50 @@ def get_ensurer(self, key,
...
+        default is not given here.
+        """
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
     def __iter__(self):
         yield from self._d
+        yield from self.defaults
```
scopatz: Won't this duplicate keys that are in both `_d` and `defaults`?

scopatz: namely, 

```console
scopatz@localhost ~ $ len(list(__xonsh_env__)) == len(set(__xonsh_env__))
False
scopatz@localhost ~ $ len(list(__xonsh_env__))
89
scopatz@localhost ~ $ len(set(__xonsh_env__))
78
```

scopatz:
Thanks @adqm - I think your take on this makes a lot of sense.  Other than the iter issue, could you maybe add a couple of lines to the tutorial docs on this behaviour and then also updated the changelog, please?  With those, I think that this is ready to go in!

adqm:
Okay, I think this should be ready for another look.  Since last time:
* fixed the issue with double-yielding keys
* added changelog entries
* added small bit to tutorial
* made sure `${}` for looking up environment variables had the same semantics (error in Python mode, empty string in subproc mode)

scopatz:
Thanks @adqm!

PR #614, opened 2016-01-10
be less aggressive about adding newlines in multiline mode

adqm:
This change makes it so that hitting enter in the middle of a line does not create a newline, but rather runs the line as normal.

@gforsyth, was there a reason to include this case that I am not seeing?

### DIFF
--- a/xonsh/ptk/key_bindings.py
+++ b/xonsh/ptk/key_bindings.py
@@ -76,14 +76,15 @@ def load_xonsh_bindings(key_bindings_manager):
         b = event.cli.current_buffer
 
         at_end_of_line = b.document.current_line_after_cursor.strip() == ''
+        current_line_blank = (len(b.document.current_line) == 0 or
+                              b.document.current_line.isspace())
 
         # indent after a colon
         if (b.document.char_before_cursor == ':' and at_end_of_line):
             b.newline()
             b.insert_text(indent_, fire_event=False)
         # if current line isn't blank, check dedent tokens
-        elif (not (len(b.document.current_line) == 0 or
-                   b.document.current_line.isspace()) and
+        elif (not current_line_blank and
               b.document.current_line.split(maxsplit=1)[0] in DEDENT_TOKENS):
             b.newline(copy_margin=True)
             _ = b.delete_before_cursor(count=len(indent_))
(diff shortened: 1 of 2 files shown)


gforsyth:
Hey @adqm, it is a bit aggressive -- my thinking was that if you want to edit a multiline input and add something between the lines you had, that you should be able to create newlines in there, but there's almost certainly a more elegant solution to provide that functionality.

adqm:
Sounds good.  Working on some updates for now.  All ready, except for some kind of infinite loop that is also present on master.

adqm:
Alright, I think this is ready for a look now!

xonsh/ptk/key_bindings.py
```diff
@@ -75,25 +73,33 @@ def multiline_carriage_return(event):
...
+                     for i
+                     in b.document.lines_from_current[1:]))):
             b.newline(copy_margin=True)
-        elif not can_compile(b.document.text):
+        elif not current_line_blank and not can_compile(b.document.text):
             b.newline()
         else:
             b.accept_action.validate_and_handle(event.cli, b)
+
+def _is_blank(l):
```
gforsyth: This is much cleaner.

gforsyth:
@adqm, the code is much cleaner and mostly everything works -- only thing I'm noticing is that single lines in Python mode get a newline on Enter instead of executing.  If I do an `import builtins` or similar I think it should execute without the newline insertion.  The code adding a newline to input was my workaround for this, since `compile` has trouble if it doesn't see a newline character at the end of the input.

adqm:
Okay, that's really interesting.  Looking at the infinite loop I was finding, which is actually triggered by adding the newline, and which I was attributing to this, it is actually present in readline mode as well, and seems to be cause by `try_subproc_toks` being a little broken.

I'll add back in the piece that adds a newline, and I'll try to address the infinite loop in a separate PR, since it seems to be unrelated.

To trigger the infinite loop, try running `(7+2<enter>`

> only thing I'm noticing is that single lines in Python mode get a newline on Enter instead of executing

Also weird because some statements (like `print(2)`, which was my "test case") work just fine without adding the newline, so I just assumed everything would be okay without it!

scopatz:
@adqm @gforsyth - do you think this is ready to go in then? It looks good to me.

adqm:
Yes, I think it is set from my end.

scopatz:
Thanks!

Issue #613, opened 2016-01-09
Bug in PTK completion for filenames with single-quotes


gforsyth:
This only effect `prompt_toolkit` as the bug stems from how `lprefix` is processed by `xonsh.Completer.complete`

Take another terrible filename
```
gil@bad_cat ~/test $ ls
gil's.less.terrible.filename
```

Since there's only one file in the directory, tab-completion works if I type a single letter and then hit tab, but if I type `cat gil's<TAB>` it completes to

```
cat gilgil's.less.terrible.filename
```

The issue stems from [_path_from_partial_string in completer.py](https://github.com/scopatz/xonsh/blob/master/xonsh/completer.py#L24) -- it calls `tools.check_for_partial_string` which does find a partial string because of the `'` in the filename.  

Reporting so I remember about this -- not sure of a fix yet but it should hopefully be relatively trivial.  And in any case, `'` in filenames is a terrible idea.

scopatz:
Thanks for reporting @gforsyth!  

> And in any case, ' in filenames is a terrible idea.

Much agreed!

scopatz:
closing due to #615

PR #615, opened 2016-01-10
fix for quotes in filenames when completing

adqm:
I believe this should address #613.  Also adds quotes, comma, question mark, and ampersand to the list of characters that need to be quoted.


gforsyth:
Awesome -- this works on my end.

scopatz:
Thanks @adqm and @gforsyth!

PR #617, opened 2016-01-10
check correct env (denv may not persist some Python objects)

scopatz:
Reported by @frol [here](https://github.com/scopatz/xonsh/issues/108#issuecomment-169996770), this fixes and issue with source-foreign if the envvar is callable, not peristable, or otherwise not detypable.

1 char change.


adqm:
Works for me!  Merged in.

frol:
Thank you! Works fine now.

PR #619, opened 2016-01-10
don't double-wrap things in $[...]

adqm:
This should fix an infinite loop of wrapping certain syntactically-invalid subproc expressions with `$[...]`.

Previously, typing `(7+2<enter>` at the xonsh prompt would enter an infinite loop.


scopatz:
Thanks again, @adqm!

PR #620, opened 2016-01-11
Add a few tests of the PTK multiline parser

gforsyth:
Haven't worked much with mocks before so if this looks crazy, apologies.  

These should hopefully provide a few sanity checks if we want to refactor the multiline parser in the future.

### DIFF
--- a/tests/test_ptk_multiline.py
+++ b/tests/test_ptk_multiline.py
@@ -1,0 +1,33 @@
+# -*- coding: utf-8 -*-
+"""Tests some tools function for prompt_toolkit integration."""
+from __future__ import unicode_literals, print_function
+
+import nose
+from nose.tools import assert_equal
+from unittest.mock import MagicMock
+
+import builtins
+from xonsh.ptk.key_bindings import carriage_return
+from xonsh.environ import Env
+
+from prompt_toolkit.interface import CommandLineInterface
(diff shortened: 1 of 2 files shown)


tests/test_ptk_multiline.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
+"""Tests sample inputs to PTK multiline and checks parser response"""
+import nose
+from nose.tools import assert_equal
+from unittest.mock import MagicMock, patch
+
+import builtins
+from xonsh.environ import Env
+builtins.__xonsh_env__ = Env()
+builtins.__xonsh_env__['INDENT'] = '    '
```
scopatz: These lines could have side effects for other tests.  probably best to stick all of this stuff in a setup() function and then unwind it all in a teardown(). Same goes for below as well.

scopatz:
This looks very nice, @gforsyth. Thanks! Only the setup and teardown issue to worry about.

scopatz:
Hey @gforsyth, is this ready for another look? Please ping the conversation so that I get an email if it is! Thanks.

gforsyth:
Hey @scopatz , this is ready for another check.

scopatz:
Thanks a ton @gforsyth!

Issue #624, opened 2016-01-13
Jupyter kernel times out


esafak:
xonsh times out "waiting for kernel_info reply" and consequently does not work. I've tried deleting $HOME/.local/share/jupyter/kernels/xonsh and reinstalling xonsh (0.2.4). Here's some debug output from jupyter:

```
[D 00:59:45.560 NotebookApp] Found kernel xonsh in /home/user/.local/share/jupyter/kernels
[D 00:59:45.560 NotebookApp] Native kernel (python3) available from /home/user/.local/lib/python3.4/site-packages/ipykernel/resources
[D 00:59:45.561 NotebookApp] Starting kernel: ['/usr/bin/python3', '-m', 'xonsh.jupyter_kernel', '-f', '/run/user/1000/jupyter/kernel-51a0038d-1f3d-43b8-8c1e-d6dfb14e2bf5.json']
[D 00:59:45.602 NotebookApp] Connecting to: tcp://127.0.0.1:52111
[I 00:59:45.603 NotebookApp] Kernel started: 51a0038d-1f3d-43b8-8c1e-d6dfb14e2bf5
[D 00:59:45.603 NotebookApp] Kernel args: {'kernel_name': 'xonsh', 'cwd': 'user/directory'}
[D 00:59:45.603 NotebookApp] 201 POST /api/sessions (127.0.0.1) 47.00ms
[D 00:59:45.610 NotebookApp] Initializing websocket connection /api/kernels/51a0038d-1f3d-43b8-8c1e-d6dfb14e2bf5/channels
[D 00:59:45.613 NotebookApp] Requesting kernel info from 51a0038d-1f3d-43b8-8c1e-d6dfb14e2bf5
[D 00:59:45.613 NotebookApp] Connecting to: tcp://127.0.0.1:54302
[W 00:59:55.614 NotebookApp] Timeout waiting for kernel_info reply from 51a0038d-1f3d-43b8-8c1e-d6dfb14e2bf5
[D 00:59:55.615 NotebookApp] Opening websocket /api/kernels/51a0038d-1f3d-43b8-8c1e-d6dfb14e2bf5/channels
[D 00:59:55.619 NotebookApp] Connecting to: tcp://127.0.0.1:54302
[D 00:59:55.619 NotebookApp] Connecting to: tcp://127.0.0.1:34576
[D 00:59:55.620 NotebookApp] Connecting to: tcp://127.0.0.1:56287
```

[...]

PR #625, opened 2016-01-14
setsid on foreign shell setup to avoid hangs

minrk:
avoids hangs potentially caused by setpgrp

possible or partial fix for #624

### DIFF
--- a/xonsh/foreign_shells.py
+++ b/xonsh/foreign_shells.py
@@ -139,6 +139,8 @@ def foreign_shell_data(shell, interactive=True, login=False, envcmd='env',
         currenv = dict(currenv)
     try:
         s = subprocess.check_output(cmd, stderr=subprocess.PIPE, env=currenv,
+                                    # start new session to avoid hangs
+                                    start_new_session=True,
                                     universal_newlines=True)
     except (subprocess.CalledProcessError, FileNotFoundError):
         if not safe:


xonsh/foreign_shells.py
```diff
@@ -139,6 +139,8 @@ def foreign_shell_data(shell, interactive=True, login=False, envcmd='env',
         currenv = dict(currenv)
     try:
         s = subprocess.check_output(cmd, stderr=subprocess.PIPE, env=currenv,
+                                    # setsid where available to avoid hangs
+                                    preexec_fn=getattr(os, 'setsid', None),
```
scopatz: Hi @minrk, thanks for putting this in.  I am a little hesitant to merge this in as-in due to the warnings in the [Python docs](https://docs.python.org/3/library/subprocess.html#popen-constructor) namely: 

```
Warning The preexec_fn parameter is not safe to use in the presence of threads in your application. 
The child process could deadlock before exec is called. If you must use it, keep it trivial! 
Minimize the number of libraries you call into.

Note If you need to modify the environment for the child use the env parameter rather than doing it in a preexec_fn.
The start_new_session parameter can take the place of a previously common use of preexec_fn to call os.setsid() 
in the child.
```

How do you feel about `start_new_session=True` that @takluyver suggested in #624?  This seems to be the safe version of what is implemented here.

takluyver: The Python docs say that's precisely what start_new_session is for:

> The start_new_session parameter can take the place of a previously common use of preexec_fn to call os.setsid() in the child.

minrk: Done and much better. I was flailing about a bit on this one, but it's nice that you require Python 3 so we can use this.

scopatz: Ummhmm, Python 3 is amazing :)

scopatz:
Thanks again @minrk and @takluyver!

blink1073:
Well done guys!

PR #626, opened 2016-01-14
check that we own _shell_tty before passing ownership

minrk:
It's possible that there is a tty, but we are a detached subprocess and don't control it, in which case `_give_terminal_to` will fail with OSError.

This came up during testing of `setsid` in Jupyter, which seem to cause this to be the case.


scopatz:
Thanks a ton @minrk!

PR #630, opened 2016-01-15
Fix background colors for PTK

BYK:
Background colors in PTK doesn't work at all. This patch makes a weak attempt to make them work. Definitely not a great or permanent solution but should be a good start.

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -53,6 +53,7 @@ VER_FULL = sys.version_info[:3]
 VER_MAJOR_MINOR = sys.version_info[:2]
 V_MAJOR_MINOR = 'v{0}{1}'.format(*sys.version_info[:2])
 
+
 def docstring_by_version(**kwargs):
     """Sets a docstring by the python version."""
     doc = kwargs.get(V_MAJOR_MINOR, None)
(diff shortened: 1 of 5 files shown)


melund:
Be careful when changing the _PT_COLORS mapping. The colors codes corresponding to the names were chosen deliberately to look good on dark barkgrounds on both Linux and Windows. There is an old PR with a discussion about it.

scopatz:
Thanks for putting this in @BYK! Seems like there are some legitimate test failures...

BYK:
@melund - changing the color mapping was not intentional but I see it changed for a few colors. I think this is just an artifact from my continuous rebases for some time. I'll revert them

@scopatz - Gonna look into them. Haven't experienced any issues my self for a long while tho :)

gforsyth:
Hey @BYK -- could you post an image or two of what this looks like on your machine?  I tried it out but I think any changes it's trying to make are being overridden by `.Xresources`.

BYK:
@melund - I tried something better to map the existing dark/light colors to the "intense" modifier.
@gforsyth - here you go (don't judge me, I'm not great with colors and I definitely didn't adjust this properly): colored-xonsh
@scopatz - Tests should be fixed now. Also noticed an issue about `tests/test_execer.py` but I'm guessing that file is not used at all now. May be we should remove it?

BYK:
For reference here's the part of my `~/.xonshrc` file that corresponds to this prompt:

```py
def prompt():
    has_branch_name = bool($FORMATTER_DICT['curr_branch']())
    branch_info = '{branch_bg_color}[{curr_branch_stripped}]' if has_branch_name else ''

    return '{BACKGROUND_YELLOW}{user}@ {BACKGROUND_BLUE}{BLACK}{cwd_styled}%s{NO_COLOR}{BLUE}\n{prompt_end} ' % branch_info

$PROMPT = prompt
```

melund:
@BYK That seems cool. I will try to test it in the next few days, but reviewing the code here on my phone it seems OK. We should make the INTENCE colors the default in the prompt.

BYK:
Hey @melund, thanks for the quick follow up.  Looks like I need to squash and order my commits first :)

>We should make the INTENCE colors the default in the prompt.

I'm not sure if I agree with this since then I don't see a way to make `INTENSE` modifier work. Or did you also implied to reverse its meaning to `INTENSE` would use the other color set? I can simply rename the dicts to make that happen.

BYK:
Rebased properly onto latest master.

scopatz:
Code looks good to me.

BYK:
I've discovered a weird issue. When I try to use `BACKGROUND_GRAY` or `BACKGROUND_INTENSE_GRAY` it throws a key error and then fails in another place. This only happens for `GRAY`, couldn't reproduce anywhere else. Probably not a show stopper but I'd rather et it fixed. Any ideas?

scopatz:
`GRAY` is not a supported color in `TERM_COLORS`

gforsyth:
@BYK -- the `INTENSE` regex isn't working quite right, I don't think.

For example:
```Bash
>>_make_style('BACKGROUND_GREEN')
'bg:#008000'
>>_make_style('BACKGROUND_INTENSE_GREEN')
'bg:#008000'
```

xonsh/tools.py
```diff
@@ -764,71 +765,90 @@ def __iter__(self):
         return iter(self.char)
 
 
-RE_HIDDEN_MAX = re.compile('(\001.*?\002)+')
+COLOR_CODE_SPLIT_RE = re.compile(r'(\001\033\[[\d;m]+\002)')
+TERM_COLORS_REVERSED = {v: k for k, v in TERM_COLORS.items()}
+COLOR_NAME_REGEX = re.compile(r'(?:(\w+)_)?(\w+)')
+INTENSE_QUALIFIER_REGEX = re.compile(r'\bINTENSE$')
```
gforsyth: How about `re.compile(r'\w+_INTENSE$')` ?

BYK:
@gforsyth - Very good catch. Fixed and updated tests to catch that and the background fixes.

melund:
@BYK . I was trying to say that your PR is changing how the default prompt looks. You moved the tweaked colors into the Intense color_mapping. I do think it makes sense, but since the default prompt doesn't use the intense keyword`'{BOLD_GREEN}{user}@{hostname}{BOLD_CYAN} {cwd}{branch_color}{curr_branch} {BOLD_WHITE}{prompt_end}{NO_COLOR} '` the prompt will look like this on windows...

image


I suggest that you also update the default prompt to: 

`'{BOLD_INTENSE_GREEN}{user}@{hostname}{BOLD_INTENSE_CYAN} {cwd}{branch_color}{curr_branch} {BOLD_WHITE}{prompt_end}{NO_COLOR} '`

BYK:
@melund - ah, correct. Fix upcoming.

BYK:
@melund done

melund:
I doesn't work for me. You INTENSE_QUALIFIER_REGEX doesn't detect the INTENSE keyword. If you change the regex to: `INTENSE_QUALIFIER_REGEX = re.compile(r'\w+(?:^|_)INTENSE')` then it should work.

melund:
Btw. The default branch color also have to be changed to INTENSE: https://github.com/scopatz/xonsh/blob/57e7c1f43879e6394342a17638f2a9b6cb80bc23/xonsh/environ.py#L810

BYK:
@melund - this is why I love code reviews :) Refactored the method to fix the issue, updated tests and also updated branch colors. Do you think I should update the new `branch_bg_color` to be intense too?

melund:
@BYK - I don't think backgrounds should be 'INTENSE' but that is just my preference. However, I notized something weird. The following prompt background color is enabled for branch, also colored the background of the stuff before and after the branch.

`$PROMPT= '{BOLD_INTENSE_GREEN}{user}@{hostname}{BOLD_INTENSE_CYAN} {short_cwd}{branch_bg_color}{curr_branch} {BOLD_INTENSE_CYAN}{prompt_end}{NO_COLOR} '`

image

BYK:
>@BYK - I don't think backgrounds should be 'INTENSE' but that is just my preference.

I agree :)

>However, I notized something weird. The following prompt background color is enabled for branch, also colored the background of the stuff before and after the branch.

Before is weird, after is expected since in Bash, until you use `NO_COLOR` rest of the string gets colored. I'm guessing before part is due to the weird tokenization we do. I noticed that happening sometimes too. We are doing very hacky things to make this work so it can use a better structure. That's why I was hesitant to submit this PR.

Anyways, here's what I think: I think this patch adds more value than that bug so we can start going with this and I'll submit one or two more PR's to fix it. Sounds reasonable?

melund:
You are right. I think we can merge without finding the source of the bug. It seems that adding `{NO_COLOR}` after `{curr_branch}` fixes the colors both before and after. 

image

If you add this then I think we can merge it.

BYK:
@melund done!

melund:
@BYK 

I get this error now...

``` 
C:\Users\mel>xonsh
Could not find xonsh configuration or run control files.
  File "c:\users\mel\documents\xonsh\xonsh\main.py", line 191, in main
    shell.cmdloop()
  File "c:\users\mel\documents\xonsh\xonsh\ptk\shell.py", line 103, in cmdloop
  File "c:\users\mel\documents\xonsh\xonsh\ptk\shell.py", line 47, in singleline
    token_func, style_cls = self._get_prompt_tokens_and_style()
  File "c:\users\mel\documents\xonsh\xonsh\ptk\shell.py", line 125, in _get_prompt_tokens_and_style
    custom_style = _xonsh_style(tokens, cstyles)
  File "c:\users\mel\documents\xonsh\xonsh\ptk\shell.py", line 131, in _xonsh_style
    class XonshStyle(Style):
  File "c:\users\mel\documents\xonsh\xonsh\ptk\shell.py", line 172, in XonshStyle
    styles = {k: _make_style(v) for k, v in styles.items()}
  File "c:\users\mel\documents\xonsh\xonsh\ptk\shell.py", line 172, in <dictcomp>
    styles = {k: _make_style(v) for k, v in styles.items()}
  File "c:\users\mel\documents\xonsh\xonsh\tools.py", line 820, in _make_style
    name = 'INTENSE_' + name
TypeError: Can't convert 'NoneType' object to str implicitly

```

BYK:
Weird, gonna try to fix it. I'm guessing `NO_COLOR` somehow getting merged with `INTENSE` which shouldn't really happen.

BYK:
@melund - I can't figure out how you can end up in a state like that but I pushed fix that should prevent it.

BYK:
NOTE: This breaks `PROMPT_TOOLKIT_COLORS` thing. Is it important?

scopatz:
Well, I use PROMPT_TOOLKIT_COLORS, and there were some previous discussipns anout how that is more intuitive.

BYK:
This is why we can't have nice things... :P

Okay, just kidding. Can you share your config for that part so I can see how it is used and factor that into the existing code?

melund:
@BYK - It works for me now. Regarding the `PROMPT_TOOLKIT_COLORS` then I think we ought to keep that functionality. I don't use it but I see why it is nice to have.

BYK:
@melund @scopatz - pushed another change to add `PROMPT_TOOLKIT_COLORS` support back.

scopatz:
Thanks @BYK, seems like you might need to merge from master and fix a merge conflict...

BYK:
@scopatz it was a legitimate test failure and I fixed it. The code is already up to date with `master`. Should be ready for merge now if you and @melund think it is all good.

melund:
Now it works. Nice work @BYK. Small changes easily becomes a long PR ;) I will go ahead and merge it in.

BYK:
Thanks a ton for helping me not break anything for everyone!

PR #634, opened 2016-01-19
Jupyter Fixes

scopatz:
This implements a few small fixes to the Jupyter kernel that makes it usable.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -49,6 +49,9 @@ None
 * Jupyter kernel no longer times out when using foreign shells on startup.
 * Capturing redirections, e.g. ``$(echo hello > f.txt)``, no longer fails
   with a decoding error.
+* Evaluation in a Jupyter cell will return pformatted object.
+* Jupyter with redirect uncaptured subprocs to notebook.
+* Tab completion in Jupyter fixed.
 
 **Security:**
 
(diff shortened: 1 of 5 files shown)


melund:
Wahoo... The kernel works for me now. It has been a while since I tried it. This is really awesome.

scopatz:
Yay!  Do you want to merge this @melund?

blink1073:
I'd recommend setting `os.environ['PAGER'] = 'cat'`, so you can get inline man pages.

scopatz:
@blink1073 is that genreally applicable (i.e. windows). In any event, I think that could easily be another PR if this goes in before then.

blink1073:
there is no "pager" on windows, but you could add an `if os.name != 'nt'` guard.

scopatz:
I think that this requires a bit more thought because of things like cygwin.  It would be nice to just get this merged in and worry about the paging issues later, personally :)

blink1073:
fair enough.

blink1073:
for reference, that guard would work on Cygwin: http://stackoverflow.com/a/7637706

scopatz:
I see, so I guess we'd want to use `xonsh.tools.ON_POSIX` rather than `xonsh.tools.ON_WINDOWS`: https://github.com/scopatz/xonsh/blob/master/xonsh/tools.py#L42

melund:
Great. I think this is good, so I will go ahead and merge.

blink1073:
:tada:  Were you planning a patch release for this?

scopatz:
Yep.  I'll do it now.  It is now relatively easy to do a release because of our release script.

scopatz:
Alright, The release is out.  Have at :)

PR #641, opened 2016-01-22
Raise subproc error

scopatz:
This adds the capability to raise a subprocess error whenever the subprocess returns a greater-than-zero exits status.  This is most useful for scripts and sometimes modules.

This was inspired by Bash's `set -e`.  For xonsh, it turns out, we only needed to add this behaviour for subproc mode, because Python mode will already throw an error.

Hopefully, `set -x` behaviour will come soon too, but that is a little more complicated.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,12 @@ Xonsh Change Log
 Current Developments
 ====================
 
-**Added:** None
+**Added:** 
+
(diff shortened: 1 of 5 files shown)


scopatz:
I have added some more documentation for this now.

gforsyth:
Everything looks good here, merging in.

Issue #644, opened 2016-01-22
Entering multiline mode on windows....


melund:
I just played with xonsh using it's default settings. And I stumbled on a multiline mode issue, which occurs in windows when `$FORCE_POSIX_PATHS=False` (default). 

When autocompleting folders, a backslash is added to the folder name. I think that is reasonable. But it causes xonsh to enter multiline mode when pressing enter. Pressing one extra time then execute the command. 
```
mel@ANYBODY-MEL ~ $ cd Docu<TAB>
mel@ANYBODY-MEL ~ $ cd Documents\
...................
mel@ANYBODY-MEL ~\Documents $
```

I can see a lot of problems here, but maybe we could have 'FORCE_POSIX_PATHS =True' as default. Even on Windows. At least, until it works consistently.

gforsyth:
Urrrgggggg.  I forgot about `\`s being used for paths in windows.  `FORCE_POSIX_PATHS=True` will fix this, as you say, or we can remove the `\` from the multiline triggers if `ON_WINDOWS`

gforsyth:
or something a little more convoluted that checks if the `cmd` at the beginning of the current line of the PTK document is `cd` and if so to not enter multiline mode...

melund:
Fixed by #645

PR #645, opened 2016-01-22
disable multiline for '\' on windows without POSIX

gforsyth:
Partial fix for #644 

If user is on Windows and has `FORCE_POSIX_PATHS` set to False, then
trailing backslash will not enter multiline mode as this messes up
directory navigation

@melund, can you see if this works on your end?  

Note that this doesn't preclude changing the default of `FORCE_POSIX_PATHS` to `True`

### DIFF
--- a/tests/test_ptk_multiline.py
+++ b/tests/test_ptk_multiline.py
@@ -90,5 +90,15 @@ def test_can_compile_and_executes():
         carriage_return(buffer, cli)
         assert bufaccept.mock_calls is not None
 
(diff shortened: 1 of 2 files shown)


scopatz:
Looks like there are some failures

gforsyth:
@scopatz -- forgot that AppVeyor *is* on Windows and probably doesn't need to be mocked up like it isn't.

melund:
Yes. That work nicely. I am merging it in. Thanks.

Issue #643, opened 2016-01-22
Autocomplete issues...


melund:
I get this weird error when using the autocompleter (windows and prompt_toolkit). 

Here is an example. It also happens when I autocomplete other stuff. 

`mel@ANYBODY-MEL ~ $ mv .xonsh<TAB>`

```
mel@ANYBODY-MEL ~ $ mv .xonshException in thread Thread-24:
Traceback (most recent call last):
  File "C:\Users\mel\Anaconda3\lib\threading.py", line 914, in _bootstrap_inner
    self.run()
  File "C:\Users\mel\Anaconda3\lib\threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File "c:\users\mel\anaconda3\lib\site-packages\prompt_toolkit-0.57-py3.5.egg\prompt_toolkit\interface.py", line 673, in run
    completions = list(buffer.completer.get_completions(document, complete_event))
  File "c:\users\mel\documents\xonsh\xonsh\ptk\completer.py", line 38, in get_completions
    self.reserve_space()
  File "c:\users\mel\documents\xonsh\xonsh\ptk\completer.py", line 44, in reserve_space
    window = cli.application.layout.children[1].children[1].content
AttributeError: 'FloatContainer' object has no attribute 'children'
```

gforsyth:
Hey @melund, can you try out #637 ? The default layout in `prompt_toolkit` changed in recent versions which I think is what's causing that error.

scopatz:
Also, can you post the contents of the `xonfig info` command please?

melund:
@gforsyth - I will give it a try... 

Here is the content of my `xonfig info` 

```
mel@ANYBODY-MEL ~ $ xonfig info
+------------------+-------+
| xonsh            | 0.2.5 |
| Python           | 3.5.1 |
| PLY              | 3.7   |
| have readline    | True  |
| prompt toolkit   | 0.57  |
| on posix         | False |
| on linux         | False |
| on arch          | False |
| on windows       | True  |
| on mac           | False |
| are root user    | False |
| default encoding | utf-8 |
+------------------+-------+
```

melund:
@gforsyth - It fixes the issue. Nice.

![image](https://cloud.githubusercontent.com/assets/1038978/12523888/d42b5c60-c159-11e5-808a-c4bba52f5cf2.png)

gforsyth:
[...]

PR #647, opened 2016-01-24
add temporary fix for diverging prompt_toolkit layout

gforsyth:
This is a temporary fix until there's a new release of `prompt_toolkit` to address the differences in the default `prompt` layout between 0.5.7 and current git master.  

This should address #643

### DIFF
--- a/xonsh/ptk/completer.py
+++ b/xonsh/ptk/completer.py
@@ -41,7 +41,11 @@ class PromptToolkitCompleter(Completer):
 
     def reserve_space(self):
         cli = builtins.__xonsh_shell__.shell.prompter.cli
-        window = cli.application.layout.children[1].children[1].content
+        #TODO remove after next prompt_toolkit release
+        try:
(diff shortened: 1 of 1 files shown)


xonsh/ptk/completer.py
```diff
@@ -41,7 +41,11 @@ def get_completions(self, document, complete_event):
 
     def reserve_space(self):
         cli = builtins.__xonsh_shell__.shell.prompter.cli
-        window = cli.application.layout.children[1].children[1].content
+        #TODO remove after next prompt_toolkit release
+        try:
+            window = cli.application.layout.children[1].children[1].content
+        except AttributeError:
+            window = cli.application.layout.children[1].content
```
scopatz: Which one do we want to keep after the next release?

gforsyth: heh.  oops.  we want to keep the second one.  I'll put in a comment to that effect.

scopatz:
Thanks!

PR #649, opened 2016-01-25
Fix case-insensitive match breaking for paths in Windows

BYK:
When using `glob` on Windows, case-insensitive conversion converts a
glob path like `~\D*` to
`[cC]:\[uU][sS][eE][rR][sS]\[bB][yY][kK]\[dD]*`. The `[cC]` part at the
beginning breaks `glob` and fails to yield any results, probably because
it cannot extract the base path.

This patch fixes the issue by skipping the conversion for a sequence
like `\w:` at the beginning.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -185,12 +185,23 @@ def expand_path(s):
     return os.path.expanduser(s)
 
 
+WINDOWS_DRIVE_MATCHER = re.compile(r'^\w:')
+
+
 def expand_case_matching(s):
     """Expands a string to a case insenstive globable string."""
     t = []
     openers = {'[', '{'}
     closers = {']', '}'}
     nesting = 0
+
+    drive_part = WINDOWS_DRIVE_MATCHER.match(s) if ON_WINDOWS else None
+
+    if drive_part:
+        drive_part = drive_part.group(0)
+        t.append(drive_part)
+        s = s[len(drive_part):]
+
     for c in s:
         if c in openers:
             nesting += 1
@@ -211,8 +222,7 @@ def expand_case_matching(s):
                                              c)
                 c = newc
         t.append(c)
-    t = ''.join(t)
-    return t
+    return ''.join(t)
 
 
 def reglob(path, parts=None, i=None):
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -735,36 +735,6 @@ def history_tuple_to_str(x):
 #
 
 
-class FakeChar(str):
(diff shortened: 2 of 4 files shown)


BYK:
Should probably add a test for this.

xonsh/built_ins.py
```diff
@@ -185,12 +185,23 @@ def expand_path(s):
     return os.path.expanduser(s)
 
 
+WINDOWS_DRIVE_MATCHER = re.compile(r'^\w:'
```
gforsyth: need to move that paren up a few lines

xonsh/tools.py
```diff
@@ -870,17 +840,6 @@ def print_color(string, file=sys.stdout):
     print(format_color(string), file=file)
 
 
-def escape_color(string):
```
scopatz: I could have sworn that this function was used somewhere, but I guess not :)

BYK:
Btw. I'm not very happy with this hacky solution. What do you think about implementing `iglob` and `glob` ourselves using `fnmatch` but ising case-insensitive regular expressions?

xonsh/built_ins.py
BYK: You want this at the top? I wanted to keep it close to where it was used.

xonsh/built_ins.py
```diff
@@ -185,12 +185,23 @@ def expand_path(s):
...
+
+                                   )
```
gforsyth: Sorry -- I meant that the closing paren (on line 190) should be moved up so the regex is on one line.

BYK: Hah, sorry. Done :)

scopatz:
Well I am fine with replacing / wrapping these functions.  We already do some amount of this. However, it should the case sensitivity should still be optional.

scopatz:
Thanks for putting this in @BYK!

BYK:
>Well I am fine with replacing / wrapping these functions. We already do some amount of this. However, it should the case sensitivity should still be optional.

Alright. That requires quite a bit more refactoring and tests to go along so let's start with this one and I'll start working on a better fix if that suits you. (and yes case-insensitive will be optional :))

>Thanks for putting this in @BYK!

No problem! :)

scopatz:
This looks good to me, then! Would you mind adding an entry to the CHANGELOG about this :)

BYK:
>This looks good to me, then! Would you mind adding an entry to the CHANGELOG about this :)

There's always one more thing to do, right? :D Sure, update coming :)

BYK:
@scopatz done!

PR #650, opened 2016-01-26
Eliminate `locate_binary` and use a native Python implementation instead

BYK:
We already had a Python implementation for `which`/`where`
functionality but for some reason we were not making enough use of it.
This patch moves ` built_ins._get_runnable_name` to
`environ.locate_binary` and makes everything use it. This improves `git`
and `hg` querying speeds for prompt on Windows.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -8,6 +8,7 @@ import string
 import locale
 import builtins
 import subprocess
+from itertools import chain
 from warnings import warn
 from pprint import pformat
 from functools import wraps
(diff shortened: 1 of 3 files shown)


gforsyth:
You're missing an `import stat` in `xonsh/environ.py`

gforsyth:
Also, as a side note -- since `_is_executable_file` is ripped from `pexpect`, should we note that in the docstring?

BYK:
>You're missing an import stat in xonsh/environ.py
Oh, wow. Kind of surprised how it passed tests and worked locally. Guessing I never hit that `stat.` path.

>Also, as a side note -- since _is_executable_file is ripped from pexpect, should we note that in the docstring? 

Didn't know that. Is it a 1-1 copy? (I just moved this from `built_ins` module)

gforsyth:
It's a 1-1 copy of an older version.  I think we can clean it up a bit and also probably remove a bunch of stuff in there since it's mostly as chunky as it is to support Solaris.  @scopatz, have you promised anyone Solaris support?

BYK:
@gforsyth makes sense. I'm up for the task :)

gforsyth:
@BYK, awesome.  The new version is here: https://github.com/pexpect/pexpect/blob/master/pexpect/utils.py#L6

If we ditch Solaris support we can cut out the entire conditional and also remove the `stat` import since that's the only place it's used.

scopatz:
Anything from pexpect needs to have the following in the docstring or file:

```
This function was forked from pexpect originally:

    Copyright (c) 2013-2014, Pexpect development team
    Copyright (c) 2012, Noah Spurrier <noah@noah.org>

    PERMISSION TO USE, COPY, MODIFY, AND/OR DISTRIBUTE THIS SOFTWARE FOR ANY
    PURPOSE WITH OR WITHOUT FEE IS HEREBY GRANTED, PROVIDED THAT THE ABOVE
    COPYRIGHT NOTICE AND THIS PERMISSION NOTICE APPEAR IN ALL COPIES.
    THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
    WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
    MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
    ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
    WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
    ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
```

@gforsyth I have not personally promised Solaris support to anyone, because I don't know of anyone who uses Solaris :).

BYK:
>Anything from pexpect needs to have the following in the docstring or file:

Yay, custom GPL...

scopatz:
Well this is custom MIT / BSD or something. Just happens to be GPL compatible.  If it was GPL we'd have to rip it out entirely :)

BYK:
Hah. Anyways, update coming. Wish this portion was a separate open-source project so I could just import instead of copying.

BYK:
Did some refactoring as well. Let me know what you think.

asmeurer:
How's the performance of this vs. the old way on non-Windows?

scopatz:
Well, it avoids a subprocess, which is nice.

xonsh/environ.py
```diff
@@ -631,20 +632,60 @@ def _repr_pretty_(self, p, cycle):
...
+        for a_file in os.listdir(directory):
+            base_name, ext = os.path.splitext(a_file)
+            if name == base_name and ext.upper() in extensions:
+                yield os.path.join(directory, a_file)
+else:
+    def yield_executables(directory, name):
+        if name in os.listdir(directory):
+            path = os.path.join(directory, name)
+            if _is_executable_file(path):
+                yield path
```
scopatz: Hmm, this is a little strange to me.  Why don't you have two function: `yield_executables_win()` and `yield_executables_posix()`, and then a top-level wrapper function called `yield_executables()` that has the if-else statement.  I think that is a bit cleaner over all.

scopatz:
Also, nice branch name :)

xonsh/environ.py
BYK: Eh, this is a bit of an optimization. It is based on the assumption that you'll never change platforms, thus the value of these variables at runtime. If you find it confusing, I can write it as you suggested though.

BYK:
>How's the performance of this vs. the old way on non-Windows?

I didn't time it (and honestly not sure if I can properly time it without adding perf code into xonsh internals) but it sure makes the prompt appear faster in a `git` directory. If you really care about it I can time it and share the results.

xonsh/environ.py
scopatz: I'd appreciate it, since I don't  believe that this is something that needs to be optimized.

BYK: I will but it does save one branching per lookup and lookups are done pretty frequently :)

BYK:
@scopatz - well, I think I took best of both worlds with this latest update. What do you think?

>Also, nice branch name :)

Hah, thanks :)

scopatz:
> @scopatz - well, I think I took best of both worlds with this latest update. What do you think?

perfect!

scopatz:
Thanks a ton @BYK and @gforsyth

PR #651, opened 2016-01-26
Have consistent repo detection for `git` and `hg`

BYK:


xonsh/environ.py
```diff
@@ -691,18 +691,36 @@ def locate_binary(name):
...
 
 
+def _get_parent_dir_for(cwd, dir_name):
+    # walk up the directory tree to see if we are inside an hg repo
+    dir_list = [dir_name]
+    path = cwd.split(os.path.sep)
+    while len(path) > 0:
+        if os.path.isdir(os.path.sep.join(path + dir_list)):
+            return os.path.sep.join(path)
+        del path[-1]
```
BYK: Not a fan of this mutation here TBH. I'd prefer `path = path[:-1]`

xonsh/environ.py
```diff
@@ -691,18 +691,36 @@ def locate_binary(name):
         return None
 
 
+def _get_parent_dir_for(cwd, dir_name):
+    # walk up the directory tree to see if we are inside an hg repo
+    dir_list = [dir_name]
+    path = cwd.split(os.path.sep)
```
asmeurer: You should use os.path.split and os.path.join

scopatz: I don't know why most of this looks like it does. I don't think that it was me:)

xonsh/environ.py
scopatz: I think that pop() is probably  the best.

scopatz:
This is probably ready to go in, but we can delay if you want to clean it up.

xonsh/environ.py
BYK: >You should use os.path.split and os.path.join

I thought about using `os.path.join` and because I had to spread the arguments, I decided not to change this part. I'll do it though. For `os.path.split`, I don't think it works as you expect it to work. It just splits the path into two.

BYK:
@scopatz only updated the `os.path.join` part and added a comment about `del path[-1]` part.

scopatz:
Hmm this seem to break the current branch behaviour for me

```console
# before
scopatz@localhost ~ $ cd xonsh/
scopatz@localhost ~/xonsh master $ cd xonsh
scopatz@localhost ~/xonsh/xonsh master $ 

# now
scopatz@localhost ~ $ cd xonsh/
scopatz@localhost ~/xonsh $ cd xonsh
scopatz@localhost ~/xonsh/xonsh $ 
```

gforsyth:
```
(xonshtesting)gil@theo~/git/xonsh$ path = cwd.split(os.path.sep)
(xonshtesting)gil@theo~/git/xonsh$ path
['', 'home', 'gil', 'git', 'xonsh']
(xonshtesting)gil@theo~/git/xonsh$ os.path.join(*(path + ['.git']))
'home/gil/git/xonsh/.git'
```

this is stripping out the leading `/` and so `os.path.isdir` will always return `False`

scopatz:
yep, fix incomming.

scopatz:
See BYK/xonsh#2

BYK:
>this is stripping out the leading / and so os.path.isdir will always return False

Interesting but this is the way it was before. See the old code used in `ensure_hg`.

gforsyth:
I think the difference was in the usage of `os.path.sep.join` vs. `os.path.join`

```
gil@theo ~/git/xonsh pr/651 $ os.path.join(*(path + ['.hg']))
'home/gil/git/xonsh/.hg'
gil@theo ~/git/xonsh pr/651 $ os.path.sep.join(path + ['.hg'])
'/home/gil/git/xonsh/.hg'
```

BYK:
@gforsyth yup. Working on a change that makes use of `os.path.join` and `os.path.split`. `split` is acting a bit weird at root level, that's what I'm trying to figure out.

BYK:
Okay, updated and I think simplified.

xonsh/environ.py
```diff
@@ -691,18 +691,37 @@ def locate_binary(name):
         return None
 
 
+def _get_parent_dir_for(path, dir_name):
+    # walk up the directory tree to see if we are inside an hg repo
+    previous_path = ''
+    while path != previous_path:
+        if os.path.isdir(os.path.join(path, dir_name)):
+            return os.path.sep.join(path)
```
scopatz: I don't think we need to join on this line now :)

BYK: Hah, good catch!

BYK:
@scopatz fixed!

scopatz:
Woohoo! Thanks @BYK!

PR #653, opened 2016-01-27
stop source alias from truncating filename

gforsyth:
the replacement of `locate_binary` in #650 changed the way filenames are returned -- the `[:-1]` truncation in `source_alias` means that files not in the current directory are not found, despite being on `$PATH`


scopatz:
Looks good to me! Thanks.

PR #654, opened 2016-01-27
set process title

scopatz:
This adds `setproctitle` as an optional dependency to enable setting the process (rather than terminal) title. This will make xonsh appear as `"xonsh"` to ps and other process monitors, rather than `/path/to/python /path/to/xonsh` 

Inspired by @asmeurer. Also @asmeurer, note: https://anaconda.org/asmeurer/setproctitle :)

### DIFF
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -130,7 +130,7 @@ def _pprint_displayhook(value):
 def premain(argv=None):
     """Setup for main xonsh entry point, returns parsed arguments."""
     if setproctitle is not None:
-        setproctitle('xonsh')
+        setproctitle(' '.join(['xonsh'] + sys.argv[1:]))
     args, other = parser.parse_known_args(argv)
     if args.file is not None:
         real_argv = (argv or sys.argv)
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -6,6 +6,11 @@ import builtins
 from argparse import ArgumentParser, ArgumentTypeError
 from contextlib import contextmanager
 
+try:
+    from setproctitle import setproctitle
+except ImportError:
+    setproctitle = None
+
 from xonsh import __version__
 from xonsh.shell import Shell
 from xonsh.pretty import pprint
@@ -124,6 +129,8 @@ def _pprint_displayhook(value):
 
 def premain(argv=None):
     """Setup for main xonsh entry point, returns parsed arguments."""
+    if setproctitle is not None:
+        setproctitle('xonsh')
     args, other = parser.parse_known_args(argv)
     if args.file is not None:
         real_argv = (argv or sys.argv)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,8 @@ Current Developments
   and a subprocess command exits with a non-zero return code, a 
   CalledProcessError will be raised. This is useful in scripts that should
   fail at the first error.
+* If the ``setproctitle`` package is installed, the process title will be
+  set to ``'xonsh'`` rather than the path to the Python interpreter.
 
 **Changed:** None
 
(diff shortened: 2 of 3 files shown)


xonsh/main.py
```diff
@@ -124,6 +129,8 @@ def _pprint_displayhook(value):
 
 def premain(argv=None):
     """Setup for main xonsh entry point, returns parsed arguments."""
+    if setproctitle is not None:
+        setproctitle('xonsh')
```
asmeurer: Should maybe be `'xonsh' + ' '.join(sys.argv[1:])`.

asmeurer: Or I guess `' '.join(['xonch'] + sys.argv[1:])` if you don't like trailing whitespace.

asmeurer:
This sadly does not affect the iTerm2 process title in the tab bar. Here are some details on what iTerm2 does https://groups.google.com/forum/#!msg/iterm2-discuss/bG6b1CEoWn0/ndyaGnUtBwAJ. We may also consider checking upstream iTerm2 to see if things can/should be improved on that end.

scopatz:
Hmm so that post says that it should be the same as `ps -aef`. I tested this on Linux and I do see the "xonsh" name.  If you get the right behaviour with ps but not the right behaviour with iTerm2, it is probably worth reporting to them :).  I'm happy to have a special case for iTerm2 too, if there is one.

xonsh/main.py
scopatz: Ahh good point.

scopatz:
Ok, I have added the arguments to the title.  Should be ready for another look!

asmeurer:
Yeah, it does fix `ps` (and that post is wrong, it doesn't match `ps -aef`). Looks OK to merge to me.

asmeurer:
I figured out how to use sysctl with ctypes https://gist.github.com/pudquick/581a71425439f2cf8f09 (`sysctl(b'kern.procname')` gives `b'python3.5', even in master with this PR).

asmeurer:
Well I managed to change the proc name with sysctl:

```
xonshmaster%=🐚 def set_procname(new_name):
..............     size = c_uint(0)
..............     libc.sysctlbyname(b'kern.procname', None, byref(size), None, 0)
..............     buf = create_string_buffer(size.value)
..............     libc.sysctlbyname(b'kern.procname', buf, byref(size), new_name, c_uint(len(new_name)))
..............
..............
xonshmaster%=🐚 set_procname(b'test')
xonshmaster%=🐚 sysctl(b'kern.procname')
b'test'
```

*but*, that didn't change iTerm2. Digging into the iTerm2 code a bit more, it looks like it uses the "command name" (`kp.kp_proc.p_comm`). The search continues...

asmeurer:
I think I could do it with sysctl, but I need to create a [`kinfo_proc`](https://github.com/st3fan/osx-10.9/blob/34e34a6a539b5a822cda4074e56a7ced9b57da71/xnu-2422.1.72/bsd/sys/sysctl.h#L677) structure, and I have no idea how to do that in ctypes.

asmeurer:
Well, I tried writing a C program to change it, but it wasn't working ("Operation not permitted", which either means you can't change it, or I was doing something wrong). I've asked about this [on StackOverflow](https://stackoverflow.com/questions/35194538/is-it-possible-to-set-a-processs-command-name-p-comm).

scopatz:
Ugg crazy! I wouldn't have expected this to be so hard.

asmeurer:
Some updates on [this thread](https://groups.google.com/forum/#!msg/iterm2-discuss/bG6b1CEoWn0/ndyaGnUtBwAJ) (things don't look good).

scopatz:
Yeah, seems like they'll get to eventually was the answer.  I don't really consider this a feature, but rather a bug fix, though.

asmeurer:
It's a feature fix.

PR #658, opened 2016-02-03
Space after close bracket

takluyver:
This will hopefully fix the link in the docs: http://xonsh.org/envvars.html#shell-type


scopatz:
Ahh cool. Thanks @takluyver

PR #663, opened 2016-02-05
make call_hg_command() use the correct environment

### LINKED ISSUE #661: Custom $PATH additions not available in os.environ
I added `$PATH.insert(0, '/usr/local/bin')` to .xonshrc. In xonsh, $PATH is `['/usr/local/bin', '/usr/bin', '/bin']`. However, `import os; os.environ['PATH']` yields `'/usr/bin:/bin'`.

Due to this, [hg_dirty_working_directory](http://xonsh.org/_modules/xonsh/environ.html#hg_dirty_working_directory) doesn't work properly showing false negatives. hg is in /usr/local/bin, so it raises FileNotFoundError when called.

scopatz:
This should fix #661. @ngoldbaum and @moigagoo, would you mind taking a look please?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,7 @@ Current Developments
 **Fixed:**
 
 * Fixed path completion not working for absolute paths or for expanded paths on Windows.
+* Fixed issue with hg dirty branches and $PATH.
 
 **Security:** None
 
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -790,7 +790,7 @@ def get_git_branch(cwd=None):
 
 def call_hg_command(command, cwd):
     # Override user configurations settings and aliases
-    hg_env = os.environ.copy()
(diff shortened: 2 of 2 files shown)


ngoldbaum:
:+1:

moigagoo:
Thanks, hg now works properly!

FYI, `os.environ['PATH']` is still `'/usr/bin:/bin'`, although this is not an issue here.

scopatz:
Great!  I'll go ahead and merge this in then.  As per the `os.environ` thing. Xonsh doesm't currently update the os.environ automatically, as you discovered.  (It would be a lot of extra work on the off chance someone might need it.) Xonsh applications should use the `__xonsh_env__` directly.  However, if you absolutely need os.environ to be updated you can use the `__xonsh_env__.replace_env()` method: http://xonsh.org/api/environ.html#xonsh.environ.Env.replace_env

PR #667, opened 2016-02-07
Workaround bug in ConEmu/cmder (windows terminal emulators)

melund:
Workaround bug in ConEmu/cmder which prevented get_git_branch() from working in these terminal emulators on windows.


scopatz:
Thanks @melund!

PR #675, opened 2016-02-12
Add Vox environment manager to xonsh

moigagoo:
- Registered `vox` as a default alias.
- Added the `{env_name}` prompt variable.
- Added environment name indication to the default prompt.


xonsh/vox.py
```diff
@@ -0,0 +1,197 @@
...
+        """Create a virtual environment in $VIRTUALENV_HOME with ``python3 -m venv``.
+
+        :param name: virtual environment name
+        """
+
+        env_path = join(builtins.__xonsh_env__['VIRTUALENV_HOME'], name)
+
+        print('Creating environment...')
+
+        vbuiltins.__xonsh_env__.create(env_path, with_pip=True)
```
gforsyth: `vbuiltins`?

moigagoo: Thanks a lot for the correction! It's supposed to be `venv`. When I autoreplaced ENV with builtins.__xonsh_env__, I totally overlooked this bit :-(

Fixed in https://github.com/moigagoo/xonsh/commit/b92969ec4b4aa134d7b91ee5f7b4e98d8afe80c2

scopatz:
Looks like there are conflicts now, probably from the new parser being merged in to master. If you could resolve these that would be awesome!

scopatz:
Also, a fair amount of this seems to be reimplementing what argparse does.  We should probably just move to argparse.  There are a lot of examples in xonsh that use it already.

moigagoo:
Conflict resolved, ready to merge.

I didn't quite understand your last comment. Do you mean that I reimplemented argparse functionality in Vox? Or are you talking about the new parser?

I can't see how Vox can be rewritten to utilize argparse since xonsh handles argument parsing already. If it were a standalone app, there would be a way to do it, but I'd still use [a more high-level library](http://cliar.readthedocs.org/).

xonsh/vox.py
```diff
@@ -0,0 +1,197 @@
...
+
+        except IndexError:
+            print('Command "%s" doesn\'t exist.\n' % command_name)
+            self.print_commands()
+
+    @staticmethod
+    def new(name):
+        """Create a virtual environment in $VIRTUALENV_HOME with python3's ``venv``.
+
+        :param name: virtual environment name
```
scopatz: Please use numpydoc style docstrings.

xonsh/vox.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def __init__(self):
+        """Ensure that $VIRTUALENV_HOME is defined and declare the available vox commands"""
+
+        if not builtins.__xonsh_env__.get('VIRTUALENV_HOME'):
+            if os.name == 'nt':
+                home_path = builtins.__xonsh_env__['USERPROFILE']
+
+            elif os.name == 'posix':
+                home_path = builtins.__xonsh_env__['HOME']
```
scopatz: I think that you should use `home_path = os.path.expanduser('~')` here.

xonsh/vox.py
```diff
@@ -0,0 +1,197 @@
...
+
+        while bin_path in builtins.__xonsh_env__['PATH']:
+            builtins.__xonsh_env__['PATH'].remove(bin_path)
+
+        builtins.__xonsh_env__.pop('VIRTUAL_ENV')
+
+        print('Deactivated "%s".\n' % env_name)
+
+    @staticmethod
+    def list():
```
scopatz: methods probably shouldn't have the same name as a builtin type

xonsh/environ.py
```diff
@@ -926,6 +928,15 @@ def _current_job():
     return ''
 
 
+def env_name(pre_chars='(', post_chars=') '):
+    """Extract the current environment name from $VIRTUAL_ENV."""
+    env_path = builtins.__xonsh_env__.get('VIRTUAL_ENV', '')
```
scopatz: I think that `VIRTUAL_ENV` should be properly handled and registered by the envrionment.  See the `xonsh.environ` file. Here you can register defaults, validators, and docstrings.  Then  this would get documentation automatically generated, you could use `__xonsh_env__.get('VIRTUAL_ENV')`, and you could ensure that this is always a string.

xonsh/environ.py
```diff
@@ -101,11 +101,13 @@ def is_callable_default(x):
     """Checks if a value is a callable default."""
     return callable(x) and getattr(x, '_xonsh_callable_default', False)
 if ON_WINDOWS:
-    DEFAULT_PROMPT = ('{BOLD_INTENSE_GREEN}{user}@{hostname}{BOLD_INTENSE_CYAN} '
+    DEFAULT_PROMPT = ('{env_name}'
```
scopatz: Should this go before the user and host name? I don't know how it works elsewhere.

scopatz:
Thanks for resolving the conflicts, @moigagoo!  And thanks for withstanding my many comments.  In addition to those, it would be great if you could add in some autodoc hooks in the `docs/` dir, since vox.py is a new file. Also, it'd be great if you added an entry to the CHANGELOG. People will want to know about this. :)

> I didn't quite understand your last comment. Do you mean that I reimplemented argparse functionality in Vox? 

Yes, this.

> I can't see how Vox can be rewritten to utilize argparse since xonsh handles argument parsing already. If it were a standalone app, there would be a way to do it, 

You could stick an `ArgumentParser` instance in `Vox.__call__()`. Argument parsers let you pass in an arglist manually.  There are many examples of xonsh aliases that do this:

1. https://github.com/scopatz/xonsh/blob/master/xonsh/aliases.py#L76
2. https://github.com/scopatz/xonsh/blob/master/xonsh/aliases.py#L158
3. https://github.com/scopatz/xonsh/blob/master/xonsh/xonfig.py#L283
4. https://github.com/scopatz/xonsh/blob/master/xonsh/replay.py#L119
5. https://github.com/scopatz/xonsh/blob/master/xonsh/history.py#L447

> but I'd still use a more high-level library.

Tools like cliar are great. However, we are trying to maintain maximum protablility by relying only on the Python standard library and PLY as required runtime dependencies. (And I have thought about moving a copy of PLY into xonsh so that it could be basically standalone.)  I consider vox of core value, so it would be nice if it didn't rely on optional dependencies :)

Also, note, that not using argeparse is not blocking this.  It was just a suggestion. We have other many other aliases that don't use argparse, so it is a reasonable choice not to.

xonsh/environ.py
gforsyth: This is the default `conda` format and I *think* what `virtualenv` does, too.

scopatz: OK! Sounds good then :)

xonsh/vox.py
moigagoo: Fixed in https://github.com/moigagoo/xonsh/commit/60daf5e88605d09b57f42633345849e1c063dd0f

xonsh/vox.py
moigagoo: Fixed in https://github.com/moigagoo/xonsh/commit/7488142dbc55d4cb10138394e420fee2a4c069d8

xonsh/vox.py
moigagoo: Fixed in https://github.com/moigagoo/xonsh/commit/89d7b754fd4a5a63a298ba4a6b7401b5acdb3f37

xonsh/environ.py
moigagoo: Fixed in https://github.com/moigagoo/xonsh/commit/926b74ba3f011ef6a264b91d9f49d3802f0d3add

xonsh/environ.py
moigagoo: Yes, this is default format for venv and virtualenv as well.

xonsh/environ.py
```diff
@@ -384,6 +387,8 @@ def xonshconfig(env):
         default='xonsh.environ.DEFAULT_TITLE'),
     'VI_MODE': VarDocs(
         "Flag to enable 'vi_mode' in the 'prompt_toolkit' shell."),
+    'VIRTUAL_ENV': VarDocs(
+        "Path to the currently active Python environment."),
```
scopatz: The should have `configurable=False`, I would think.

xonsh/environ.py
scopatz: Do you want to please add an entry to `DEFAULT_VALUES` and then change this to `__xonsh_env__.get('VIRTUAL_ENV')`

xonsh/vox.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def __init__(self):
+        """Ensure that $VIRTUALENV_HOME is defined and declare the available vox commands"""
+
+        if not builtins.__xonsh_env__.get('VIRTUALENV_HOME'):
+            if os.name == 'nt':
+                home_path = builtins.__xonsh_env__['USERPROFILE']
+
+            elif os.name == 'posix':
+                home_path = os.path.expanduser('~')
```
scopatz: Sorry, not just this line, but this whole conditional can be replaced with the expanduser().  Rather than having the if-elif-else blocks, you can just have the one line.

xonsh/environ.py
moigagoo: This variable should not be set by default since there's no virtual environment activated by default. This is why I didn't put it in `DEFAULT_VALUES`.

xonsh/vox.py
scopatz: Sorry if that wasn't clear.

xonsh/environ.py
scopatz: But you are applying a default value here as an empty string. It is more general to put this empty string in the default values dict, and more consistent with the rest of xonsh.  I'd be happy to make this change, if your don't want to for some reason.

xonsh/vox.py
```diff
@@ -0,0 +1,198 @@
...
+        env_path = join(builtins.__xonsh_env__['VIRTUALENV_HOME'], name)
+
+        if not exists(env_path):
+            print('This environment doesn\'t exist. Create it with "vox new %s".\n' % name)
+            return None
+
+        if os.name == 'nt':
+            bin_dir = 'Scripts'
+
+        elif os.name == 'posix':
```
scopatz: You should probably import `ON_WINDOWS` and `ON_POSIX` from `xonsh.tools` and use them here, for consistency.

xonsh/vox.py
moigagoo: I feel so stupid right now :-)

https://github.com/moigagoo/xonsh/commit/17cb4cbd1708c0f45ce860240cd795f477bcde41

scopatz: I'm sure this was my bad explanation in the early morning :)

On Sat, Feb 13, 2016 at 1:58 PM Konstantin Molchanov <
notifications@github.com> wrote:

> In xonsh/vox.py
> <https://github.com/scopatz/xonsh/pull/675#discussion_r52831995>:
>
> > +
> > +import builtins
> > +
> > +
> > +class Vox:
> > +    """Vox is a virtual environment manager for xonsh."""
> > +
> > +    def __init__(self):
> > +        """Ensure that $VIRTUALENV_HOME is defined and declare the available vox commands"""
> > +
> > +        if not builtins.__xonsh_env__.get('VIRTUALENV_HOME'):
> > +            if os.name == 'nt':
> > +                home_path = builtins.__xonsh_env__['USERPROFILE']
> > +
> > +            elif os.name
>
>  == 'posix':
> > +                home_path = os.path.expanduser('~')
>
> I feel so stupid right now :-)
>
> moigagoo@17cb4cb
> <https://github.com/moigagoo/xonsh/commit/17cb4cbd1708c0f45ce860240cd795f477bcde41>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/scopatz/xonsh/pull/675/files#r52831995>.
>
-- 

Asst. Prof. Anthony Scopatz
Nuclear Engineering Program
Mechanical Engineering Dept.
University of South Carolina
Affiliated Professor of the University of Haifa
scopatz@cec.sc.edu
Office: (803) 777-7629
Cell: (512) 827-8239
Check my calendar
<https://www.google.com/calendar/embed?src=scopatz%40gmail.com>

xonsh/environ.py
moigagoo: This is totally reasonable. Here you go: https://github.com/moigagoo/xonsh/commit/5bb1ea76a33e4f9c7f82849d487caf67623ebdd1

xonsh/vox.py
moigagoo: Fixed: https://github.com/moigagoo/xonsh/commit/1186637e23fe980054dbc0d5ac56047867104bb3

xonsh/environ.py
moigagoo: https://github.com/moigagoo/xonsh/commit/15be794714711a622a94c9ce9b2cc660970d8aee

scopatz:
OK @moigagoo, let us know when you think that this is ready to go in!

moigagoo:
I fixed all the issues you had reported except for the argparse one. Fingers crossed :-)

First, thanks a lot for investing your time in reviewing and commenting on my commits.

Second, I'd like to comment a bit on the argparse issue. I think it's a bit of an overkill in this case. Argparse has such an overcomplicated API, which, I believe, doesn't really pay off in this task. However, if you think it's the right thing to do, I'm ready to do it—xonsh is your project after all :-)

moigagoo:
:fireworks:

scopatz:
> First, thanks a lot for investing your time in reviewing and commenting on my commits.

Thank you for investing your time and effort in making xonsh better!  Hopefully, that will continue :)

> Second, I'd like to comment a bit on the argparse issue. I think it's a bit of an overkill in this case. Argparse has such an overcomplicated API, which, I believe, doesn't really pay off in this task. However, if you think it's the right thing to do, I'm ready to do it—xonsh is your project after all :-)

This is fine. As I said, I don't think argparse should block this since we have many other core utilities in xonsh that don't use it.  Using argpare seems to be about 50%, so I don't think it is a big deal.

If you think it is ready, let me know and I'll merge it in.

moigagoo:
> If you think it is ready, let me know and I'll merge it in.

Sure, go ahead.

scopatz:
Thanks a million, @moigagoo!

PR #678, opened 2016-02-13
best shell type

scopatz:
This adds a 'best' shell type that tries to discover the best available shell from those available on the system. This PR also changes the default to 'best'.  This is meant to address the issue of people needing to manually switch to prompt-toolkit, if it is available.  If and when future shells are implemented, best can be adjusted as needed.

This is a relatively minor change, but I think it is worth it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,9 +4,14 @@ Xonsh Change Log
 
 Current Developments
 ====================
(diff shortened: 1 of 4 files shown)


gforsyth:
Hey @scopatz -- this looks good to me.  Merging it in.

PR #680, opened 2016-02-13
Updated pretty.py with newest commits from the IPython project

melund:
The pretty printer we are using from ipython project has some advanced features. So it is easy to fix if there are any of the built-in structures in xonsh that doesn't display nicely. So if you have any particular issues I can try to fix it.

### DIFF
--- a/xonsh/pretty.py
+++ b/xonsh/pretty.py
@@ -6,6 +6,11 @@ to provide their own pretty print callbacks.
 
 This module is based on ruby's `prettyprint.rb` library by `Tanaka Akira`.
 
+The following implementations were forked from the IPython project:
+* Copyright (c) 2008-2014, IPython Development Team
+* Copyright (C) 2001-2007 Fernando Perez <fperez@colorado.edu>
+* Copyright (c) 2001, Janko Hauser <jhauser@zscout.de>
+* Copyright (c) 2001, Nathaniel Gray <n8gray@caltech.edu>
 
 Example Usage
 -------------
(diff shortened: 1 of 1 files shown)


xonsh/pretty.py
```diff
@@ -1,17 +1,11 @@
...
+replace the old `pprint` python module which does not allow developers
+to provide their own pretty print callbacks.
 
 This module is based on ruby's `prettyprint.rb` library by `Tanaka Akira`.
 
-The following implementations were forked from the IPython project:
-* Copyright (c) 2008-2014, IPython Development Team
-* Copyright (C) 2001-2007 Fernando Perez <fperez@colorado.edu>
-* Copyright (c) 2001, Janko Hauser <jhauser@zscout.de>
-* Copyright (c) 2001, Nathaniel Gray <n8gray@caltech.edu>
```
scopatz: The copyrights need to be maintained due to IPython's license.

xonsh/pretty.py
```diff
@@ -83,6 +77,7 @@ def _repr_pretty_(self, p, cycle):
             Portions (c) 2009 by Robert Kern.
 :license: BSD License.
 """
+from __future__ import print_function
```
scopatz: STrintly speaking, I don't think we need this :)

scopatz:
Thanks for putting this in @melund. I think the copyright is the only big issue.

melund:
Ups. I forgot about the copyright notice. I weakly remember adding it the first time we added pretty.py. It should be up fixed now.

scopatz:
Looks good to me! Thanks @melund!

PR #687, opened 2016-02-18
Color Refactor

scopatz:
First off, I apologize for putting in another giant PR right after the parser refactors. However, like ths parser refactor, this one is totally worth it. At least with this one, most of it has been autogenerated :)

This fundamentally changes how colors are handled though out xonsh. Currently, for historical reasons, colors are all handled via ANSI color codes that were wrapped in hidden escape sequences `\001` and `\002`. The *only* thing that this is convenient for was readline prompts. It a pain for everything else because:

* In the readline shell, you have to strip the `\001` and `\002` off the sides of all color codes when printing colors that are not in a prompt, such as with the history differ, tracer, and the xonfig wizard.
* A genius, if mechanical, system of regexes had evolved to convert a subset of ANSI color codes to a  list of Pygment `(Token, str)` tuples as used by prompt-toolkit.This led to an excessive number of string transformations.

This PR changes all of that by making each Shell class responsible for printing and formatting colors in the way that is most natural to that shell.  This provides some more primitive tools to help reduce that burden. 

This abstraction enables a lot of cool features.  The primary one of which is the ability to dynamically set a color map or style.  This is done through the `$XONSH_COLOR_STYLE`, which simply takes a string style name.  Currently supported string names are those that were present in pygments on my machine. Note that these styles are implemented for ANSI colors even when pygments is not installed! 

These styles had to be autogenerated, which was done via some code that you can find here: https://github.com/scopatz/xolors  I am not convinced that my algorithm there produces the most representative colors for a given style and palette.  However, hopefully this can evolve in time, or someone can come up with something better.

Additionally, because it is easy implement colors, I added coloring back to the `?` and `??` operator output. I even extended this to optionally include syntax highlighting for the source code in ``??``

Furthermore, I had been eyeing the print displayhook for color for some time.  So now, if pygments is installed, the output of Python mode objects will also be syntax highlighted, in addition to being pretty printed.

This, of course, has a cost.  While this won't break people's xonshrc files (I don't believe), it does render prior coloring and style handling basically moot.

I think that this needs to be hammered on.  I also think that along with the and/or subproc operators this makes a nice candidate for v0.3.0. The minor version bump will help imply that something serious is going on :)

In any event, I am looking forward to hearing folks thoughts!


BYK:
So basically, this diff piggy backs on the terminal emulator to display ANSI codes, which is what was happening at the end anyways when we converted color codes to ANSI escape sequences and then them to Pygments tokens and then back to ANSI escape sequences, right?

I think this is a great way forward. The "crazy" idea I had for colors was, with prompt-toolkit, since we had tokenization, was to tokenize the actual prompt parts like `user`, `cwd`, `branch_name` etc so people can color/style them way easier. What do you think?

scopatz:
> The "crazy" idea I had for colors was, with prompt-toolkit, since we had tokenization, was to tokenize the actual prompt parts like user, cwd, branch_name etc so people can color/style them way easier. What do you think?

Hi @BYK, can you spell out what a user would type, please? Would it be just a list of token tuples?

gforsyth:
First off, this looks awesome.  Speaking for the colorblind among us, dynamic colormap changes make it much easier to find a combination that is actually legible.


`AssertionError` from `prompt_toolkit` with this

```
(xonshtesting)gil@theo~/git/xonsh pr/687$ xonsh --no-rc
Could not find xonsh configuration or run control files.
AssertionError
  File "/home/gil/anaconda/envs/xonshtesting/bin/xonsh", line 9, in <module>
    load_entry_point('xonsh==0.2.6', 'console_scripts', 'xonsh')()
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/main.py", line 212, in main
    shell.cmdloop()
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 105, in cmdloop
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 66, in singleline
    style=PygmentsStyle(xonsh_style_proxy(self.styler)),
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/prompt_toolkit-0.57-py3.5.egg/prompt_toolkit/styles/from_pygments.py", line 59, in __new__
  File "/home/gil/anaconda/envs/xonshtesting/lib/python3.5/site-packages/prompt_toolkit-0.57-py3.5.egg/prompt_toolkit/styles/from_dict.py", line 54, in style_from_dict
AssertionError
```

scopatz:
> Speaking for the colorblind among us, 

This was definitely a concern of mine :)

As for the error @gforsyth, what version of prompt-toolkit are you on? I know it says 0.57, but I suspect that you are on master.

gforsyth:
Hey @scopatz -- I was on master, but after checking out 0.57 and installing it I'm still having the same problem.

scopatz:
I see the culprit: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/styles/from_dict.py#L55  

There is no reason for this type check, or at least it is the wrong type check.  I'll put in a fix.

scopatz:
Potential fix in jonathanslenders/python-prompt-toolkit#244

scopatz:
My home computer must not be on the latest ptk!

gforsyth:
Huh... although that seems unlikely.  Oddly, installing `prompt_toolkit` via `pip` instead of checkout out the appropriate release in `git` this works just fine.  

I'm going to do this all again to make sure I didn't make some silly mistake somewhere.  If I did then we don't have to wait for a new ptk release to merge this in

scopatz:
It was probably an install thing as you say.  ptk v0.57 doesn't even have the styles sub-package: https://github.com/jonathanslenders/python-prompt-toolkit/tree/0.57/prompt_toolkit  

So my home computer probably is up-to-date :)

scopatz:
@gforsyth the fix is now on ptk master. Do you mind seeing if there are other latent issues.

gforsyth:
Well this is quite lovely.

2016-02-18-165420_806x533_escrotum

scopatz:
+1000!  That is an awesome image @gforsyth!

gforsyth:
It was a great pleasure to make it, @scopatz!
I did the same in the `readline` shell and no other latent issues popped up.  I think this is good to go.

scopatz:
Feel free to merge :)

gforsyth:
Thanks, @scopatz!

Carreau:
Note that these style won't follow a custom theming of the ansi color of terminals. (like iTerm2 is set to Solarized) . Though now in Pygments master you can use `#ansigreen`/`#ansired`.... and ti will generate `\x1b[31m`... escape sequences, so you can get colors that match your env.

scopatz:
Thanks @gforsyth!

@Carreau, that is a great point. Once the `#ansicolor` stuff hits a pygments release, we should change the `xonsh.pyghooks.DEFAULT_STYLE` to use them.  I had considered this as potentially problematic, but we actually had the same problem even before this was merged. 

Note that if the user is on the readline shell and is using the default style it will directly match the terminal style.

scopatz:
Also, I see that you are responsible for these `#ansicolors` @Carreau! Great work!

BYK:
>Hi @BYK, can you spell out what a user would type, please? Would it be just a list of token tuples? 

As an environment variable, a user would pass a dictionary of token names to styles, like
```py
$XONSH_STYLE = {
    PATH_PORTION_TOKEN: 'bg:#00ff00 #ffffff',
    USERNAME_TOKEN: '#dddddd',
}
```

etc.

By the way this PR completely broke my background coloring :( I resorted back to the behavior before my colors patch so any foreground color overrides any previous background colors whereas in Bash background colors are not affected by foreground color changes. Example:

`'{BACKGROUND_PURPLE}{CYAN}Hello'` should generate a text with purple background and cyan foreground whereas it generates a text with black background (default) and cyan foreground :(

scopatz:
Ahh yeah.  OK.  I am sorry about the background color thing.  I am hoping to have a PR tonight that addresses these issues.

Carreau:
Yeah, Xonsh and IPython where 2 main reason to try to support that.

BYK:
@scopatz no worries and thanks a lot!

scopatz:
@BYK potential fix in #695

PR #695, opened 2016-02-20
More colors: fixes and features

scopatz:
This PR does a few things related to colors. 

First it addresses the issue of background colors not working in PTK as reported by @BYK in #687.

Secondly, it add the ability to pass in hex color codes for both foreground and background colors into *both* ptk and readline shells.  For instance, ``{#fafad2}{bg#fff}Hello{NO_COLOR}`` will give you a light goldenrod hello on a white background (right @FlanFlanagan).  

Thirdly, this adds a matplotlib alias called ``mpl`` to render (highly pixelated) figures using background colors. For example, the following xonsh one-liner will produce the following image:

```console
$ import numpy as np; import matplotlib.pyplot as plt; x = np.linspace(0, 2*np.pi); plt.plot(x, np.sin(x), 'r-'); mpl
```

sin

For some reason, I think that this is hilarious :).

Hilarious and potentially useful...


melund:
Wow... I wish I had come up with that idea. This will be great for the pycon talk

scopatz:
> This will be great for the pycon talk

I agree!

BYK:
@scopatz this doesn't seem to work :( Actually, it seems to be breaking colors even more:

Relevant part of `.xonshrc`:
```py
VISUAL_PATH_SEPARATOR = '{WHITE}\uE0B1{BLACK}'


$FORMATTER_DICT['curr_branch_stripped'] = lambda: $FORMATTER_DICT['curr_branch']().strip()
$FORMATTER_DICT['cwd_styled'] = lambda: VISUAL_PATH_SEPARATOR.join($FORMATTER_DICT['cwd']().split('/'))

def prompt():
    has_branch_name = bool($FORMATTER_DICT['curr_branch']())
    branch_info = '{BACKGROUND_BLACK}{PURPLE}\uE0B0 {branch_color}\uE0A0{curr_branch_stripped}' if has_branch_name else ''

    return '{BACKGROUND_CYAN}{user}@ {BACKGROUND_PURPLE}{CYAN}\uE0B0{BLACK}{cwd_styled}%s{NO_COLOR}\n{INTENSE_CYAN}{prompt_end} ' % branch_info

$PROMPT = prompt
$FORCE_POSIX_PATHS = True
$XONSH_SHOW_TRACEBACK = True
$SHELL_TYPE = 'prompt_toolkit'
$VI_MODE = True
```

Old/expected visual: xonsh-good
Master: xonsh-master
`c` branch: xonsh-c

Rebasing the branch onto master has no effect :(

BYK:
A few other things:

1. With the patch applied, typing into the prompt got extremely slow, like once char per second.
2. The original PR, #687, does break my `.xonshrc` since it removes the color variables from `$FORMATTER_DICT`. This is not a big deal for me since I can fix it easily but worth mentioning somewhere, like the changelog.

scopatz:
> With the patch applied, typing into the prompt got extremely slow, like once char per second.

This is unacceptable.  I think that this is a PTK issue, but I might have a workaround.

> The original PR, #687, does break my .xonshrc since it removes the color variables from $FORMATTER_DICT. This is not a big deal for me since I can fix it easily but worth mentioning somewhere, like the changelog.

Point taken! Will mention it there

scopatz:
PS I had also noticed some "flickering" between typing that was annoying. I wasn't sure what was causing this until you mentioned this @BYK

scopatz:
Hi @BYK - ignoring some unicode codepoints, my version of the coloring seems to do the correct thing with your setup.  I am certainly not seeing the all black and white issue that you are after the username. Can you please try it again?  At least the 1 char per second issue should now be fixed...

scopatz:
Also, on this branch, what does the following look like?

```
from xonsh.tools import print_color
print_color("{BACKGROUND_CYAN}BYK@ {BACKGROUND_PURPLE}{CYAN}>{BLACK}//{WHITE}~{BLACK}//{WHITE}yo{BLACK}{NO_COLOR}")
```

BYK:
@scopatz - Great job! It is working now without any issues, just very bright colors but guessing that's expected :)

BYK:
Actually, I was a bit too quick to be happy. Looks like the colors are not accurate at all. Here is how it looks like with my prompt and the command you provided:

xonsh

melund:
Is this not just a matter of the `$XONSH_COLOR_STYLE` you set?

melund:
Hmm. Seems I do get some weird behaviour. If I only run a single `print_color()` then everything is colored.
image

But if I run a loop over all styles then some coloring is missing:

image

scopatz:
Good news / Bad News.  I think that the logic is right on @BYK's stuff, but the `DEFAULT_STYLE` color map is terrible.  `BLACK` is getting mapped to a green for some insane reason.

scopatz:
@melund - good eye!  I believe that I have now fixed this issue on this branch.

melund:
There is some annoying behaviour on how colors are mapped. I am not sure where the problem lies, though. Is it the limited pallete of the styles, or a error in finding the best color match? 

 On cmder the cyan color (directory and prompt) looks 'cyan like' but 'intense green' (branch) is not very intense at all. 

image

On cmd.exe cyan maps to light grey, and 'intense green' has the same issue :/ 

image

melund:
> @melund - good eye! I believe that I have now fixed this issue on this branch.

It fixed it...
image

BYK:
@scopatz - still broken :(

BYK:
Top one is with this branch, bottom one is the current one (before the color patches)
xonsh

scopatz:
Hi @BYK - I haven't fixed the color map issues yet, only the broken `print_color()` function.

BYK:
@scopatz - ah my bad, sorry

scopatz:
This doesn't yet fix the problem  @BYK, but I have added the ability to dump the current color palette via the `xonfig colors` command.  As you can see, there are some clear problems with the default style (and some others).  Still trying to think up a non-manual fix.

melund:
I think it makes the most sense if we overrule the style when the users ask for a specific color like `{CYAN}` or `{#ff0120}`.  If the user wants to utilize the styles they should use the pygments tokens directly `{Token.Comment}` to get the color of that style...

I guess there is still the problem of mapping that to the colors available in the terminal.

melund:
The matplotlib hook is really fun. However, on win it is a little to wide for the terminal. 

image

Substracting one from the width fixes it. `w -= 1`, but the `h -= 1` doesn't seem to be necessary on win. 

image

adqm:
I have two separate issues with the `mpl` command.  Trying to plot something:

```
hartz@lappy-514 ~ $ import matplotlib.pyplot as p
hartz@lappy-514 ~ $ $XONSH_SHOW_TRACEBACK = True
hartz@lappy-514 ~ $ p.plot([1,3,2,4])
[<matplotlib.lines.Line2D at 0x7fda7afadc50>]
hartz@lappy-514 ~ $ mpl
  File "/usr/local/lib/python3.5/site-packages/xonsh/proc.py", line 335, in wrapped_simple_command
    r = f(args, i)
  File "/usr/local/lib/python3.5/site-packages/xonsh/aliases.py", line 197, in mpl
    show()
  File "/usr/local/lib/python3.5/site-packages/xonsh/mplhooks.py", line 61, in show
    buf = figure_to_rgb_array(fig, w, h)
  File "/usr/local/lib/python3.5/site-packages/xonsh/mplhooks.py", line 17, in figure_to_rgb_array
    fig.set_size_inches(width/dpi, height/dpi, forward=True)
  File "/usr/local/lib/python3.5/site-packages/matplotlib/figure.py", line 707, in set_size_inches
    manager.resize(int(canvasw), int(canvash))
  File "/usr/local/lib/python3.5/site-packages/matplotlib/backends/backend_tkagg.py", line 591, in resize
    self.canvas._tkcanvas.master.geometry("%dx%d" % (width, height))
  File "/usr/local/lib/python3.5/tkinter/__init__.py", line 1685, in wm_geometry
    return self.tk.call('wm', 'geometry', self._w, newGeometry)
RuntimeError: main thread is not in main loop
```

Trying to show the plot multiple times (without actually plotting anything)
```
hartz@lappy-514 ~ $ mpl
  <successfully shows empty plot>                                                                                                          
hartz@lappy-514 ~ $ mpl
  File "/usr/local/lib/python3.5/site-packages/matplotlib/backends/tkagg.py", line 22, in blit
    id(data), colormode, id(bbox_array))
_tkinter.TclError: this isn't a Tk application

  File "/usr/local/lib/python3.5/site-packages/xonsh/proc.py", line 335, in wrapped_simple_command
    r = f(args, i)
  File "/usr/local/lib/python3.5/site-packages/xonsh/aliases.py", line 197, in mpl
    show()
  File "/usr/local/lib/python3.5/site-packages/xonsh/mplhooks.py", line 61, in show
    buf = figure_to_rgb_array(fig, w, h)
  File "/usr/local/lib/python3.5/site-packages/xonsh/mplhooks.py", line 29, in figure_to_rgb_array
    fig.canvas.draw()
  File "/usr/local/lib/python3.5/site-packages/matplotlib/backends/backend_tkagg.py", line 355, in draw
    tkagg.blit(self._tkphoto, self.renderer._renderer, colormode=2)
  File "/usr/local/lib/python3.5/site-packages/matplotlib/backends/tkagg.py", line 30, in blit
    id(data), colormode, id(bbox_array))
_tkinter.TclError: this isn't a Tk application
hartz@lappy-514 ~ $ mpl
^C <ctrl-c fixes this>
hartz@lappy-514 ~ $ mpl
^C^C^C^C^C^C <perma-freeze, can't recover terminal>
```

On a different level: I think `mpl` is cool, but I'm not sure I agree that it belongs in xonsh proper...

scopatz:
@melund - windows width issue should now be fixed.

scopatz:
Hi @BYK, I believe that I have fixed the default color style.  Would you please take a look?

scopatz:
Hi @adqm, I can't seem to reproduce your issues, even when explicitly setting my matplotlib backend to `TkAgg`. (My default is `Qt4Agg`.)  Both of these issues seem to be related to tk not being on the main thread for some reason.  This is odd because the `mpl()` alias is explicitly decorated with `@foreground` , which should force this to be on the main thread.... unless you are spawning xonsh on another Python thread.  This doesn't look like what you are doing so I am somewhat confused as to what is going on here.  

Also, in addition to the TkAgg backend working just fine for me, I am able to render multiple figures one after another with no problems.

> On a different level: I think mpl is cool, but I'm not sure I agree that it belongs in xonsh proper...

I pretty much agree.  I think that this is a prime candidate to be a "user extension" whenever we get around to implementing those.  Until then, I think that we should have a batteries-included model so that we don't miss out on functionality in the intervening time. Also, since mpl is lazily loaded, it doesn't affect the users who don't use it at all other than an extra alias, which is very cheap.  Also, as @melund mentioned, I think this will be really great to show off during the PyCon talk :)

scopatz:
Hey @BYK, could you please test whether the current version fixes your issues? It would be nice to see this go in relatively soon.  I am a little nervous about the state of master right now :)

BYK:
@scopatz - oh, right. Sorry for the delay, I was hoping to test this over the weekend but slipped through my radar.

Anyways, tested it and........ IT WORKS LIKE EXACTLY BEFORE EVERYTHING WAS BROKEN! :tada: 

Great job :)

scopatz:
Wooho!!! I'd appreciate a merge then, or the OK to merge it myself :)

melund:
Colors are much nicer now. But the matplotlib code does not produce a red curve anymore. Let us just handle this in a other PR

image

scopatz:
Ok @melund - feel free to merge then :)

scopatz:
Also @melund - that is weird because the colors on my plots looks correct still.  I agree that this is better served in another PR.

scopatz:
Thanks @melund!

PR #696, opened 2016-02-21
right prompt implementation

scopatz:
As requested by @arkhan, here is the right prompt implementation.

### DIFF
--- a/xonsh/ptk/shortcuts.py
+++ b/xonsh/ptk/shortcuts.py
@@ -4,6 +4,8 @@ from prompt_toolkit.utils import DummyContext
 from prompt_toolkit.shortcuts import (create_prompt_application, 
     create_eventloop, create_asyncio_eventloop, create_output)
 
+from xonsh.shell import prompt_toolkit_version_info
+
 class Prompter(object):
 
     def __init__(self, cli=None, *args, **kwargs):
@@ -18,6 +20,7 @@ class Prompter(object):
             will be created when the prompt() method is called.
         """
         self.cli = cli
+        self.major_minor = prompt_toolkit_version_info()[:2]
 
     def __enter__(self):
         self.reset()
@@ -61,6 +64,8 @@ class Prompter(object):
 
         # Create CommandLineInterface.
         if self.cli is None:
+            if self.major_minor <= (0, 57):
+                kwargs.pop('get_rprompt_tokens', None)
             cli = CommandLineInterface(
                 application=create_prompt_application(message, **kwargs),
                 eventloop=eventloop,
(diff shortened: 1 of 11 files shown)


xonsh/ptk/shortcuts.py
```diff
@@ -61,6 +64,8 @@ def prompt(self, message='', **kwargs):
 
         # Create CommandLineInterface.
         if self.cli is None:
+            if self.major_minor <= (0, 57):
```
gforsyth: Hey @scopatz -- `prompt_toolkit` master still reports itself as version 0.57 so the `get_rprompt_tokens` kwarg never makes it past this conditional.  
I removed it and then `$RIGHT_PROMPT` works as expected.

gforsyth:
As per my comment above, maybe the easiest thing is just to hold this until the next release of `prompt_toolkit`?

scopatz:
> As per my comment above, maybe the easiest thing is just to hold this until the next release of prompt_toolkit?

Sounds good to me.  I think that a release is coming relatively soon.

scopatz:
Note that I have now made this dependent on the `c` branch, due to some conflicts with fixes there.

scopatz:
Hey @gforsyth - ptk v0.58 has been released and the CI failure is from the broken binstar stuff.  Maybe we should just disable this :( @melund, thoughts?

gforsyth:
It would be great if the binstar CI worked, but I don't think we should let it block anything.  Merging this in now.

scopatz:
Thanks @gforsyth! I agree about binstar.  I think I have tracked down the problem to the git version numbering thing.  We should probably just use the release script to fill in version numbers in `meta.yaml`.  It seems more reliable than conda build right now.

melund:
@scopatz . I also think that binstar/anaconda.org is too unreliable. Can you remove it from the CI tests?

I will see what I can do with the git version number stuff. I have also seen some weird behaviour with this. But only on windows. So I thought it was transient bug in conda build on windows. 

I don't want to mix the release script into this. It would just make it difficult to build 'cutting edge'  conda packages from master.

scopatz:
Ok @melund, binstar has been removed.  Let me know if you want me to rollback the change to the recipe that is currently on the `c` branch.  The git version shouldn't get in the way of anything now that the CI is gone.

melund:
Yes. Do that. I suggest you add the following fix on the version line instead. That fixed some git version problem for me: 

```
  version: {{ environ.get('GIT_DESCRIBE_TAG', '0.0.0') }}
```

scopatz:
done!  Nice fix too.

PR #698, opened 2016-02-21
potential fix for indented subproc mode with comments

### LINKED ISSUE #669: Comment on same line as code in script
The lexer / parser seems to hang indefinitely in the event that a comment follows a subprocess command in a script.  Namely, I found that the following failed:

```console
def func():
    ./setup.py install --user  # FIXME this should be made more general
```

scopatz:
This should fix #669

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -125,7 +125,14 @@ def subproc_toks(line, mincol=-1, maxcol=None, lexer=None, returnline=False):
             tok.type = 'NEWLINE'
             tok.value = '\n'
             tok.lineno -= 1
-            tok.lexpos = len(line)
+            if len(toks) >= 2:
+                prev_tok_end = toks[-2].lexpos + len(toks[-2].value)
+            else:
+                prev_tok_end = len(line)
+            if '#' in line[prev_tok_end:]:
+                tok.lexpos = prev_tok_end  # prevents wrapping comments
+            else:
+                tok.lexpos = len(line)
             break
     else:
         if len(toks) == 0:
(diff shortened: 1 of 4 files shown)


adqm:
Still hanging for me with:
```
$ <space>echo "hi" #test
```

scopatz:
@adqm, where? At the command line?

adqm:
Yep, if I just run that command (with the leading space), everything hangs.

xonsh/tools.py
```diff
@@ -125,7 +125,14 @@ def subproc_toks(line, mincol=-1, maxcol=None, lexer=None, returnline=False):
             tok.type = 'NEWLINE'
             tok.value = '\n'
             tok.lineno -= 1
-            tok.lexpos = len(line)
+            if len(toks) >= 2:
+                prev_tok_end = toks[-2].lexpos + len(toks[-2].value)
+            else:
+                prev_tok_end = len(line)
+            if '#' in line[prev_tok_end:]:
```
adqm: Also, is this going to break, for example, on a string that happens to contain `#`?  Or will that make it through alright?

scopatz: It shouldn't because a string will be part of one of the previous tokens.  This is capturing whitespace and comments that are not given their own tokens by the lexer.

scopatz:
Also, the hanging has nothing to do with the comment, because `   echo hello` will also hang.

adqm:
Right, same as we discussed in #669

scopatz:
Right.  OK, I think I have tracked down the hanging (in key bindings, or all places) and addressed it.

adqm:
Two separate issues with similar symptoms!

Looks good to me.  Just merged in.

PR #700, opened 2016-02-24
Workaround bug in ConEmu/cmder (windows terminal emulators) 

melund:
This is similar to #667, but this one fixes git_dirty_working_directory()

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -870,6 +870,12 @@ def git_dirty_working_directory(cwd=None):
                                     stderr=subprocess.PIPE,
                                     cwd=cwd,
                                     universal_newlines=True)
+        if len(s) == 0:
+            # Workaround for a bug in ConEMU/cmder 
+ +          # retry without redirection
+            s = subprocess.check_output(cmd,
+                                        cwd=cwd,
+                                        universal_newlines=True)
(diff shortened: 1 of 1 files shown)


BYK:
@melund which version of Git are you using? Asking because I'm surprised how I never hit these two although I'm on Windows full time and use Git extensively (and these features)

melund:
It is not a problem with git. The raw git command works correctly. It is only a problem with subprocess when running python in cmder/conemu

BYK:
@melund - I understand that. Sorry for not bein clearer: I do use Python + Xonsh inside Conemu at all times and use them with directories covered by `git` and I have never experienced any of these. That's why I'm confused.

melund:
Interesting. I am on windows 10, the newest version of git for windows. It is the cmder "packaged" version of conemu I use....

BYK:
@melund may be that's an old version with a bug (conemu). I'm using the latest Alpha.

melund:
Hmm. I will try with the latest alpha of conemu when I have time. I could be that cmder does something weird...

scopatz:
Thanks @melund - I believe that these test failures are now fixed on the `c` branch, so I will go ahead and merge.

Issue #701, opened 2016-02-24
[Feature request] Make $[cmd] return cmd's returncode instead of None


piotrjurkiewicz:
Now the only way to check command returncode is to access it via `__xonsh_history__.rtns[-1]`. I think that there is a need for a simpler scheme for doing this. `$?` environment variable was proposed, but was rejected as ugly and limited.

I've came up with an idea to use `$[cmd]` for accessing returncodes. 

Now `$[cmd]` construct always returns `None`:

``` python
>>> x = $[/usr/bin/true], $[/usr/bin/false]
>>> x
(None, None)
```

I propose to make it return command returncode instead of `None`:

``` python
>>> x = $[/usr/bin/true], $[/usr/bin/false]
>>> x
(0, 1)
```
This approach isn't more complicated to use than `$?` and, moreover, is fairly more extensible. For example you can assign returncodes of many commands to a tuple, like above.

Or you can use it in loops:

``` python
def wait_until_google_responds():
    while $[ping -c 1 google.com] != 0:
        sleep 1
```

Or you can use it in comprehensions:

``` python
# Ping all neighbors from ARP table and store whether each of them responds in a dict
>>> neigh_responds = {address: $[ping -c 1 @(address)] == 0 for address in $(ip neigh | cut -f1 -d' ').splitlines()}
>>> neigh_responds
{'192.168.48.1': False, '192.168.48.2': False, '192.168.48.3': True}
# Calculate fraction of neighbors which responds
>>> sum(neigh_responds.values()) / len(neigh_responds)
0.3333333333333333
```

adqm:
I like it!  I have actually written the code to do this a long time ago (and to hide the returncode when the command isn't explicitly invoked with `$[...]` so that the command line doesn't always display the return code).  I can try to find it and put in a PR tonight.

scopatz:
Hi @piotrjurkiewicz! Thanks a million for this feature request.  I am in full support for something like this.  I haven't looked at @adqm's PR yet, and maybe that is a better place to have this discussion.  However, my initial impression is that it would be best to have new syntax for this rather than overloading `$[]` to sometimes return the return code and sometimes not.  Maybe `?[]`

scopatz:
@piotrjurkiewicz - does the current master now satisfy your use case?

piotrjurkiewicz:
[...]

PR #672, opened 2016-02-10
and, or, &&, ||

scopatz:
The much requested, long delayed subprocess logical operators have arrived. This should close #137 and #434, finally. This builds on #671.

Though I much dreaded implementing this, it really wasn't so bad once all the pieces were in place.

I'd really appreciate people banging on this to get corner cases I might not have thought of. 

The next release after this is merged should be v0.3.0 as it does modify the language somewhat.

PR #702, opened 2016-02-25
Subproc return values

adqm:
This is a modification/update of some old code I wrote, which implements @piotrjurkiewicz 's suggestion from #701, to make captured subprocesses return a number representing their return code, instead of returning `None`.

This also implements an alternative subproc function that _does_ always return `None`, which is used when a line is forced into subproc mode so that we don't see the extra numbers showing up when running commands interactively.

Some examples:
```
hartz@lappy-514 ~/xonsh $ ls a> /dev/null
hartz@lappy-514 ~/xonsh $ $[ls a> /dev/null]
0
hartz@lappy-514 ~/xonsh $ x = $[/bin/true], $[/bin/false]
hartz@lappy-514 ~/xonsh $ x
(0, 1)
hartz@lappy-514 ~/xonsh $ [$[ls @(i) a> /dev/null] for i in ['xonsh', 'nonexistent', 'tests']]
[0, 2, 0]
hartz@lappy-514 ~/xonsh $ def foo(x):
.........................     return $[ls @(x)]
......................... 
hartz@lappy-514 ~/xonsh $ y = foo('xonsh')
aliases.py       execer.py          main.py          readline_shell.py
ansi_colors.py   execer.py~         mplhooks.py      replay.py
ast.py           foreign_shells.py  openpy.py        shell.py
ast.py~          history.py         parser.py        teepty.py
base_shell.py    imphooks.py        parsers          timings.py
built_ins.py     __init__.py        parser_table.py  tools.py
built_ins.py~    inspectors.py      pretty.py        tracer.py
completer.py     jobs.py            proc.py          vox.py
diff_history.py  jupyter_kernel.py  ptk              wizard.py
dirstack.py      lazyjson.py        __pycache__      xonfig.py
environ.py       lexer.py           pyghooks.py
hartz@lappy-514 ~/xonsh $ y
0
```

Also tries to fix another bug I noticed, involving comparing `None` against `0` when checking whether to raise an exception near the end of `run_subproc`.


adqm:
FWIW, I think this also opens the door to an implementation of `and` and `or` that is more "full-featured" than the one proposed in #672.  Using these return codes, we should be able to make Python's `and` and `or` do the work for us.  By making Python do the "heavy lifting", implementing something that works with grouping, io redirection, and even job control might be more approachable.

I will try to hack something up in the near future as a proof-of-concept.

piotrjurkiewicz:
> we should be able to make Python's and and or do the work for us

This would be tricky. Have in mind that traditional shell `&&` and `||` work inversely than Python's `and` and `or`.

adqm:
> This would be tricky.

That is true, and I am aware of that inversion, but I think I have a plan :)

scopatz:
Thanks a ton for putting this in @adqm!  I also am enticed by the notion of using Python's `and` and `or`.  We'd probably have to have some `int` subclass or some other wrapper object that hold the return value and implements the correct comparison behaviour. 

My main issue with this is that it is potentially confusing to explain that `$[]` does different things if you typed it in manually or not.  This PR also remove the ability of the user to simply have a strictly uncaptured subprocess that can be used manually resolves ambiguous cases. For example, the lines:

```python
ls = 42
$[ls -l]
```

will currently on master just return None, so at the command line this will provide a new prompt. If I understand this PR correctly, this would also print the return code before returning the new prompt.  Also, in the PR the only way to *not* have this print the return code would be to do:

```python
ls = 42
__xonsh_subproc_noreturn__([['ls', '-l']])
```

So I am sort of the opinion that returning the return code is a suitably different notion that deserves its own syntax in order to keep the language clean.  This is especially true if you consider that returning the return code is something that can be applied to both captured and uncaptured subprocessed.  I can think of a few different syntaxes that could be implemented.  Definitely open to suggestions here :)

**new operators**

Some new operators could be established, such as `?[]` and `?()`. Or maybe `$?[]` and `$?()`. Or some other idea.  There are, of course, other options.  I chose `?` because it is already associated with return codes.  This is probably my favorite option.  The user would write `?[ls -l]`

**subprocess annotations:** 

Determine some way to annotate subprocess commands and then use this to specify that the return code should be given.  For example, `$<a0,a1,...>()` and `$<a0,a1,...>[]` would be the generic syntax, and then we might have a return flag, such that the user would write `$<rtn>[ls -l]`

**return redirection**

In addition to the `out`  and `err` that we allow redirection into, we could add a `rtn` option that signals to return the return code.  The user might write $[ls -l >rtn]

**other return token**

Alterntively, like with backgrounding and the final `&` token, we could have some other token that indicates that the return code should be given if it is the last token in a subproc.  The user might write `$[ls -l ?]`

I am sure that there are other options out there.  These are the ones that sprang into my mind that would seem mostly intuitive.  The syntax for the subprocess annotations could almost certainly improved and I have not thought of other uses for such annotations.  It also seems like a lot of typing, but maybe you all see some other uses.

piotrjurkiewicz:
First I was thinking about schema like that:

```
$() -- captures (returns): stdout
       goes to the screen: stderr

?() -- captures (returns): retval
       goes to the screen: stdout and stderr

$[] -- captures (returns): nothing
       goes to the screen: stdout and stderr
```

That way:
- parentheses `()` indicate that this is a *capturing* operation
- and prefix `$`/`?` indicates *what* is being captured (`stdout` or `retval`)

But then I realized that a user might want to capture **both** `stdout` and `retval`. So there is a need for another operator, for example:

```
$?() -- captures (returns): tuple(stdout, retval)
        goes to the screen: stderr
```

But, wait! User might want as well to simultaneously capture `stdout`, `stderr` and `retval`, keeping `stdout` and `stderr` separated. Ugh, it went too far. We need to step back.

For such advanced scenarios, the best way would be to return Python object containing separately captured `stdout`, `stderr` and `retval`. A user can access particular fields of this object separately in his code. So we need a class to represent completed process. It can be named `CompletedProcess` or something like that. But, wait! Such a class already exists in Python 3.5! It is named `subprocess.CompletedProcess`. So we can use it.

The question is what operator should be used for _CompletedProcess capture mode_. As this is a *capturing* operation, it should be enclosed in parentheses to keep consistency. But with what prefix? The character which connotes mostly with _all_ or _everything_ is the star `*`. But `*()` would conflict with Python unpacking. So I propose to use the ampersand `&`.

So, finally, I propose the following schema:

```
$() -- captures (returns): stdout
       goes to the screen: stderr

?() -- captures (returns): retval
       goes to the screen: stdout and stderr

&() -- captures (returns): CompletedProcess (with stdout, stderr and retval inside it)
       goes to the screen: nothing

$[] -- captures (returns): nothing
       goes to the screen: stdout and stderr
```

That way:
- parentheses `()` indicate that this is a *capturing* operation
- and prefix `$`/`?`/`&` indicates *what* is being returned (`stdout`, `retval` or `CompletedProcess`)

scopatz:
Hi @piotrjurkiewicz, Sorry for taking a couple of days to get back to you.

I mostly agree with you. However, my original intention behind the `$[]` and `$()` distinction was that the `()` indicate that nothing is printed to the screen, while `[]` sort of look like `|` so they pass the output through to the screen.

So I think that both of the following cases are reasonable: get the return code and show all output AND get the return code and hide all of the output.  So in my interpretation, 

* `()` and `[]` determines whether or not `stdout` and `stderr` are displayed.
* the prefix `$` or `?` determines the return type, which is based on whether there is something to return or not.

So, 

```
$[] -- uncaptured stdout, so return None
$() -- captured stdout, so return str
?[] -- uncaptured stdout, but has a return code so return int
?() -- capture stdout, return return code
```

The usage of getting a `CompletedProcess` object would thus require another prefix and two more operators.  I think that is getting ahead of things for the original use case.  But I'll note that I like `!` more than `&` because (1) it reminds me of what other shells like Ammonite and IPython do, and (2) `&` is more related to connecting processes together than about just running a process.

I know that all together this would be 6 subprocess operators. However, I think that this would cover all possible cases.  We could even make it so that the CompletedProcess overloads `__bool__()` such that `!(...) and !(...)` and `!(...) or !(...) ` just work.

Alternatively, we could just completely get rid of the notion of `?()` and make people run `!().rtn` to get the return code. This would leave only 4 operators and maybe `?()` on its own is not that useful.

Thoughts?

scopatz:
Or perhaps we could add only the `!()` as this would seem to cover all other use cases. I am now leaning in favor of this....

melund:
I think I am also in favor of keeping it simple and not add overlapping features. If we can achieve everything with `!()` returning a modified version of `CompletedProcess` then I think that would be the best option. 

How does it play with adding and/or in subproc mode? `ls and ls` would not implicitly use `!()` behind the scene.

scopatz:
> How does it play with adding and/or in subproc mode? ls and ls would not implicitly use !() behind the scene.

I think that this would complement #672 or similar, not replace it.  The `ls and ls` is for uncaptured subproc, so at best we would also have to have `![]` in order to translate into `![ls] and ![ls]`.  That is probably doable in most cases, but somewhat tricky in the edge cases.

piotrjurkiewicz:
I agree with you. Indeed, all so far mentioned usecases can be covered with a single operator, which will return subclassed `CompletedProcess`.

`CompletedProcess` can be subclassed like that:

``` python
class XonshCompletedProcess(subprocess.CompletedProcess):
    def __bool__(self):
        return self.returncode == 0
    def __eq__(self, other):
        return self.returncode == other
```

`__bool__` magic method will allow usage of `?(cmd)` in logical expressions:

``` python
def check_file():
    if ?(test -e file):
        if ?(test -f file) or ?(test -d file):
            print("File is a regular file or directory")
        else:
            print("File is not a regular file or directory")
    else:
        print("File does not exist")

def wait_until_google_responds():
    while not ?(ping -c 1 google.com):
        sleep 1
```

And `__eq__` magic method will allow checking for specific return codes (we don't break any existing functionality here, as comparing of original CompletedProcess objects is pointless anyway):

``` python
if ?(curl https://somewebsite.com) in (35, 51, 90):
    print("SSL authorization error")
```

However, as you see above, I would lean towards question mark `?()`, rather than exclamation point `!()` prefix. Exclamation point in logical expressions can mislead someone to thinking that the result of command is negated here. In many languages prefix `!` in logical expressions means negation, including Python to some extent, where it is used in `__ne__` operator (`!=`). Question mark seems to be more safe and natural to use, as it is already associated with return codes in classic shells. And, when used in logical expression, it can be naturally understood as 'asking' the command for something:

``` python
# Can be understood as: 'ask if file exists'
if ?(test -e file):
    do_something()
```
Moreover, I am thinking about another usage of exclamation point: as a command terminator (similar to `&`) indicating that this command should be run in the checked mode. This would allow more fine-grained error raising control than `$RAISE_SUBPROC_ERROR` all-or-nothing variable. But this is a topic for another discussion.

adqm:
Just pushed some changes that bring this close to what is being discussed.  Not sure exactly what info we want to be available from the `CompletedProcess` object, but this handles the cases outlined by @piotrjurkiewicz above.

Happy for more suggestions.

I do still think this opens the doors to a really natural way of thinking about and/or (though the discussion above is right that we would want/need something like an equivalent of this for _uncaptured_ subproc mode calls in order to make Python handle the and/or).  I think it's doable.

piotrjurkiewicz:
Good job!

I've added some comments in specific places of the code.

xonsh/built_ins.py
```diff
@@ -637,15 +637,41 @@ def run_subproc(cmds, captured=True):
...
         raise CalledProcessError(hist.last_cmd_rtn, aliased_cmd, output=output)
-    if captured:
+    if captured=='stdout':
         return output
+    elif captured=='object':
+        return CompletedProcess(prev_proc, output)
 
 
-def subproc_captured(*cmds):
+class CompletedProcess:
```
piotrjurkiewicz: Why do you create `CompletedProcess` class from scratch instead subclassing `subprocess.CompletedProcess`?

If you were subclassed `subprocess.CompletedProcess`, you wouldn't need to write `__init__` and `__getattr__` methods.

All below would be enough:

``` python
class XonshCompletedProcess(subprocess.CompletedProcess):
    def __bool__(self):
        return self.returncode == 0
    def __eq__(self, other):
        return self.returncode == other
    # Optionally, we can add `rtn` as a shorthand alias to `returncode`
    @property
    def rtn(self):
        return self.returncode
```

xonsh/built_ins.py
```diff
@@ -637,15 +637,41 @@ def run_subproc(cmds, captured=True):
             hist.last_cmd_out = output
     if not prev_is_proxy and hist.last_cmd_rtn > 0 and ENV.get('RAISE_SUBPROC_ERROR'):
         raise CalledProcessError(hist.last_cmd_rtn, aliased_cmd, output=output)
-    if captured:
+    if captured=='stdout':
         return output
+    elif captured=='object':
+        return CompletedProcess(prev_proc, output)
```
piotrjurkiewicz: When using subclassed `subprocess.CompletedProcess`, its original constructor can be used here.

So this line would become:

``` python
return CompletedProcess(prev_proc.args, prev_proc.returncode, output, erroroutput)
```

`erroroutput` needs to be read from `stderr` before this call, similarly as `output` is read from `stdout`.

xonsh/built_ins.py
adqm: A couple of reasons: for one, not all of our completed "processes" are actually the result of running a subprocess, and also, we need something that wraps the result of a run (we need a way to create this, and it's not clear from the documentation that `subprocess.CompletedProcess` is intended to be invoked directly, or that it is easy to create such a thing with the proper attributes, other than getting it from `run`, which we don't use).

I think this isn't the right place for `CompletedProcess` to live (maybe `proc.py`), and I think there is probably more information we want to have available from those objects, but I don't think we want to subclass `subprocess.CompletedProcess`, unless I'm missing something.

adqm: One could also, I think, make an argument for not relying on the `CompletedProcess` constructor, which isn't documented, and rather using our own, for the sake of clarity.  An alternative would be to copy over the `CompletedProcess` code...or just use a tuple...

piotrjurkiewicz: `subprocess.CompletedProcess` class is nothing more than a simple wrapper around `args`, `returncode`, `stdout` and `stderr` attributes. See the source code: https://github.com/python/cpython/blob/master/Lib/subprocess.py#L639

I don't see any problem in using it directly. But if you are afraid of that because it is undocumented and for example can change without warning, the whole class code can be copied as well.

To be precise, I don't like the present solution, because `stdout` and `stderr` attributes of your `CompletedProcess` has different meaning than `stdout` and `stderr` attributes of `subprocess.CompletedProcess`. In `subprocess.CompletedProcess` these are strings/bytes containing captured contents of respective streams. In your `CompletedProcess` class these are streams itself, which would need to be `.read()` to access their contents. But this would be pointless anyway, as `stdout` was already read to get `output`, so it is empty. And `stderr` isn't PIPEd on `Popen` creation, so it is empty as well.

scopatz: For what it is worth, I agree with @adqm's reasons.  The standard library is great, but it does have a lot of weird partially documented corner cases (eg readline) that should be avoided if possible.  In addition to the reasons that @adqm gave, I'll also add:

1. [`subprocess.CompletedProcess`](https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess) was added in v3.5. So we'd either need to drop v3.4 support or have a substitute wrapper anyway.
2. Since it is new in the standard library, it is subject to change.
3. This is a relatively simple object, so the xontrol vs benefit of having our own slightly different version vs using an imperfect standard library object tilts in favor of having our own.
4. Duck typing is generally more important in Python than type checking.  I have struggled to think of a use case for this operator where "isinstance(p, subprocess.CompletedProcess)" is strictly more valuable than just having the same interface.  Are there cases that you can think of @piotrjurkiewicz - I am actually interested :)

Also, for what it is worth, I think that this should live in `xonsh.proc` (as @adqm suggested) and should probably have time stamps on it as well.

Anyway, I hope that helps.

piotrjurkiewicz: Well, if the interface is really the same, it does not make a big difference for me. The problem is that now interface isn't the same.

In currently proposed implementation, `CompletedProcess.stdout` and `CompletedProcess.stderr` are (always empty) streams left over `Popen` object, whereas in `subprocess.CompletedProcess` these are strings of read contents from these streams.

I don't care what the underlying implementation is, as long as interface is the same. But having two classes for completed processes, with different interfaces, will lead to confusion.

And one other thing: I have just noticed that stderr is not being captured now when using `?()`. I think it should be (and of course it should be kept separately from stdout, to allow user access them separately).

scopatz: Yeah, I pretty much agree that the interface should be the same, or at least that the interface here should be a superset of the interface of subprocess.

adqm:
Alright, I think this is ready for another look.  I tried to remain true to the interface from `subprocess.CompletedProcess`, but I included some more information that I thought might be relevant.

Happy for feedback/suggestions.

If this looks alright, I can start working on an and/or implementation based on this.

CHANGELOG.rst
```diff
@@ -25,6 +25,9 @@ Current Developments
 * Python mode output is now syntax highlighted if pygments is available.
 * New ``$RIGHT_PROMPT`` environment variable for displaying right-aligned
   text in prompt-toolkit shell.
+* Added ``?(...)`` operator, which returns an object representing the result
+  of running a command.  This object is truthy if the command had a non-zero
```
piotrjurkiewicz: > This object is truthy if the command had a non-zero return code

The object is truthy if the command had a **zero** return code

adqm: Of course.  Stupid typo.  Thanks.

xonsh/built_ins.py
```diff
@@ -555,6 +564,16 @@ def run_subproc(cmds, captured=True):
                 except PermissionError:
                     e = 'xonsh: subprocess mode: permission denied: {0}'
                     raise XonshError(e.format(cmd[0]))
+        _stdin_file = None
+        if captured == 'object' and stdin is not None:
+            _stdin_file = tempfile.NamedTemporaryFile()
+            cproc = Popen(['cat'],
+                          stdin=stdin,
+                          stdout=PIPE)
+            tproc = Popen(['tee', _stdin_file.name],
```
scopatz: I think that will break on non-posix platforms.

xonsh/built_ins.py
```diff
@@ -635,17 +654,40 @@ def run_subproc(cmds, captured=True):
...
+            errout = prev_proc.stderr.read()
+            errout = errout.decode(encoding=ENV.get('XONSH_ENCODING'),
+                                   errors=ENV.get('XONSH_ENCODING_ERRORS'))
+            errout = errout.replace('\r\n', '\n')
+            procinfo['stderr'] = errout
+
     if not prev_is_proxy and hist.last_cmd_rtn > 0 and ENV.get('RAISE_SUBPROC_ERROR'):
         raise CalledProcessError(hist.last_cmd_rtn, aliased_cmd, output=output)
-    if captured:
+    if captured=='stdout':
```
scopatz: PEP8

xonsh/built_ins.py
```diff
@@ -635,17 +654,40 @@ def run_subproc(cmds, captured=True):
...
         return output
+    elif captured=='object':
```
scopatz: PEP8

xonsh/execer.py
```diff
@@ -185,15 +185,15 @@ def _parse_ctx_free(self, input, mode='exec'):
...
-                if line.startswith('$['):
-                    # if we have already wrapped this in subproc tokens
-                    # and it still doesn't work, adding more won't help
-                    # anything
-                    raise original_error
                 sbpline = subproc_toks(line,
                                        returnline=True,
                                        maxcol=maxcol,
                                        lexer=self.parser.lexer)
+                if sbpline.lstrip().startswith('$[$['):
```
scopatz: Maybe you should merge from master.

xonsh/proc.py
```diff
@@ -528,3 +528,33 @@ def _wcode_to_popen(code):
...
+    Represents a completed subprocess-mode command.
+    """
+
+    def __bool__(self):
+        return self.returncode == 0
+
+    def __eq__(self, other):
+        return self.returncode == other
+
+def make_completed_command(**kwargs):
```
scopatz: This is basically an alternate `__init__()`, and so should be the `CompletedCommand.__init__()` or some other class method.  Just my two cents.

scopatz:
Thanks a million @adqm!  Right now I see a couple of blocking issues with the implementation:

1. The `cat` and `tee` implementation probably breaks thing for non-posix systems, unless it is hidden away somehow that wasn't obvious to me.
2. The tutorial needs a section on the new operator.
3. Some cleanup on the completed command itself.

I have got some other stuff to do today, but hopefully I'll be able to take a deeper look later in the day! Thanks again!

xonsh/proc.py
adqm: That was my initial thought, too (perhaps naturally), but `namedtuple` does some weird stuff with `__new__`, and this was the path of least resistance.  I'm happy to look into it a bit more, but this seemed to work with much less pain than I imagine the alternative would take.

xonsh/built_ins.py
adqm: That may be.  I'm not sure.  I guess I was imagining that most Windows folks would have msysgit or similar installed, which would presumably come with these programs.  But maybe that is not a safe assumption.  We could wrap this with a check of `ON_POSIX`, and just give up on capturing stdin (which may not have a whole lot of utility).  Or we could try to make a pure-Python version.  Any preference?

melund: The UNIX tools from "git for windows" are not the path by default. I guess most people don't select the option to add them to the path.

scopatz:
Seems like there are now merge conflicts with the changelog.

xonsh/proc.py
scopatz: I think you can simply add the line:

```python
CompletedCommand.__new__.__defaults__ = (None,) * len(CompletedCommand._fields)
```

This is basically what we do for [`VarDocs`](https://github.com/scopatz/xonsh/blob/9125e0d68dc008e29e300a8ba5d58b4394c37617/xonsh/environ.py#L230). For the reason why this works, see http://stackoverflow.com/a/18348004/2312428

xonsh/built_ins.py
scopatz: I don't think that we capture stdin elsewhere, do we?  I tend to think of capturing stdin as a potential security whole as it could easily capture things like passwords and private keys.  If you want to do a pure python version, that would be fine, but there should probably be an envvar that toggles this behaviour (`$XONSH_STORE_STDIN` perhaps) that is off by default.

scopatz:
Having not written this, I don't feel a huge claim over the syntax, but @piotrjurkiewicz and @adqm,` is there a good reason to use `?()` over `!()`?  Playing around with this, I find `?()` a little awkward to type vs  `!()`. 

This is in addition to the previous discussion that `!` connotes running a subprocess to me (a la IPython syntax) and `?` connotes getting a return value only (a la Bash).

I am happy to put in a PR to change from `?()` to `!()` if you don't mind and don't feel like doing it :)  Otherwise, it would be nice to know what the reason behind `?` is, though I may have missed it before.

melund:
I also prefer the !() syntax.

 I think @piotrjurkiewicz main objection was that !() looked a lot like negation from other languages. So `!(xxx) and !(yyy)` looked weird. But one could argue that it may be a good thing, as we are considering 0 as thruthy :)

xonsh/proc.py
adqm: Cool!  You learn something new every day...

xonsh/execer.py
adqm: I think that line is a result of a merge from master (since it wasn't in the initial parent for this changeset).  I can rebase if you want that not to show up in the diff.

xonsh/proc.py
scopatz: Seriously!

xonsh/execer.py
scopatz: No worries!

adqm:
> is there a good reason to use?()over!()? Playing around with this, I find?()a little awkward to type vs!(). 

I am (perhaps obviously) in favor of `?(...)`, since I think of it as we are asking the command for information about itself.  I also do think `!` sort of implies negation, and even though we're inverting the truthiness of the return code, that certainly doesn't "feel" the same to me as a negation of the subprocess command.

That said, I understand that `!` is related to running commands in some places (vim, among others).  And it may be a little bit easier to type.

I asked a couple of folks not connected to the project, and basically they said they didn't really have any strong feelings one way or the other.  And my feelings aren't particularly strong.  So if the two of you are strongly enough in favor of that change, I don't mind making it.

xonsh/built_ins.py
adqm: I am not afraid of capturing stdin here, since we are never capturing it interactively (only when it comes from a pipe or a file redirect).  I thought it might be nice, since we're capturing stdout and stderr anyway, just to grab all three.  Maybe someone will have a use for it?

At the least, I will wrap it with a check for whether `tee` and `cat` exist.  I can also add the environment variable if you feel strongly enough, but it isn't something that worries me personally.  I guess it could be seen a security vulnerability...

scopatz: I am very much in favor of a top-level envvar.  If implement that along with the `tee` and `cat` availability checks, I think that this should be fine.

scopatz:
I think this case is probably ambiguous enough such that "to the author goes the syntax". It might be nice to do a sanity check with the mailing list and/or gitter.

I am not a huge fan personally of this syntax, but I am not sure that it matters as long as it works and users don't mind. This personally seems to me to be a `get-all-subproc-data` operator.  Something, like `∀(...)` would be awesome, but no one would want to type that :).  Maybe other symbols scream all-ness in a way that I am currently missing.

scopatz:
I almost think `@(...)` would be the most natural, even though we'd have to redo its current Python-mode meaning with something else that I also have a hard time as seeing better than what we have now.

scopatz:
Actually, now that I think about it, and I am not saying this is a good idea, `@(...)` could be used here.  The current `@()` runs Python code in subproc mode.  This usage would run a subprocess in Python mode.

adqm:
Well, while that syntax could make some sense as a "run this in the other mode" command, I'm generally against that kind of context-sensitivity.  That said, I am kind of coming around to `!(...)`.  Will sleep on it.

Just pushed some changes that implement some of the above discussions.  Please ignore `HiddenCompletedCommand` for now; just a placeholder for the real behavior.

Will hold off on documentation until we know for sure what we want to do.

adqm:
Alright, I think this might be ready for a look.  Still no docs, but I think most of the rest is in some kind of shape (also, switched to `!(...)` syntax, so we can give that a try).

xonsh/built_ins.py
```diff
@@ -555,6 +567,20 @@ def run_subproc(cmds, captured=True):
                 except PermissionError:
                     e = 'xonsh: subprocess mode: permission denied: {0}'
                     raise XonshError(e.format(cmd[0]))
+        _stdin_file = None
+        if (captured == 'object' and
+                stdin is not None and
+                __xonsh_env__['XONSH_STORE_STDIN'] and
```
scopatz: This should probably be a `.get()` call

xonsh/built_ins.py
```diff
@@ -635,17 +664,55 @@ def run_subproc(cmds, captured=True):
...
+    Runs a subprocess, capturing the output. Returns an instance of
+    ``CompletedCommand`` representing the completed command.
+    """
+    return run_subproc(cmds, captured='object')
+
+
+def subproc_captured_hiddenobject(*cmds):
+    """
+    Runs a subprocess, capturing the output. Returns an instance of
+    ``COmpletedCOmmand`` representing the completed command.
```
scopatz: captial Os

xonsh/built_ins.py
adqm: With the current implementation of the environment, there won't be any difference (since `__getitem__` falls back to the default value anyway).

xonsh/proc.py
```diff
@@ -528,3 +528,36 @@ def _wcode_to_popen(code):
...
+    """
+
+    def __bool__(self):
+        return self.returncode == 0
+
+    def __eq__(self, other):
+        return self.returncode == other
+
+    def __hash__(self):
+        return hash(self.returncode)
```
scopatz: Won't this have a bunch of hash collisions?

xonsh/parsers/base.py
```diff
@@ -2150,10 +2159,26 @@ def p_subproc_atom_uncaptured(self, p):
...
+        p1 = p[1]
+        p0 = xonsh_call('__xonsh_subproc_captured_object__', args=p[2],
+                        lineno=p1.lineno, col=p1.lexpos)
+        p0._cliarg_action = 'splitlines'
+        p[0] = p0
+    
+    def p_subproc_atom_captured_hiddenobject(self, p):
+        """subproc_atom : bang_lbracket_tok subproc RBRACKET"""
+        p1 = p[1]
+        p0 = xonsh_call('__xonsh_subproc_captured_hiddenobject__', args=p[2],
```
scopatz: These two cases don't work, and I am not wholly sure what their purpose is except as potential setup for and/or behaviour.

```console
scopatz@artemis ~ $ echo !(echo hello)
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/base_shell.py", line 150, in default
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/execer.py", line 126, in exec
    return exec(code, glbs, locs)
AttributeError: 'CompletedCommand' object has no attribute 'splitlines'
scopatz@artemis ~ $ echo ![echo hello]
hello
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/base_shell.py", line 150, in default
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/execer.py", line 126, in exec
    return exec(code, glbs, locs)
AttributeError: 'HiddenCompletedCommand' object has no attribute 'splitlines'
```

scopatz:
I think that there are number of small issues here: the ones I have mentioned inline, lack of tests, some missing docs in environ.py for the new type, pygments not recognizing `![]` and `!()`.  

On the other hand, this does significantly move us forward, this has been open for a long time, it does address the original concern, and I am confident the minor issues can be fixed.  Thus, I think I would prefer it if this was merged and we continued to work on many small PRs rather than a few big ones.

scopatz:
Thanks a million @adqm!

xonsh/proc.py
adqm: Potentially, yes.  But it allows for checking whether the returncode is in a set/dict the same way you can check whether it's in a list/tuple.  Personally, I think the utility outweighs the danger here.

Issue #646, opened 2016-01-23
Install error - cannot import name TemporaryDirectory


QwertyJack:
$ sudo python setup.py 
    Traceback (most recent call last):
      File "setup.py", line 10, in <module>
        from tempfile import TemporaryDirectory
    ImportError: cannot import name TemporaryDirectory

scopatz:
Are you using a python that is before v3.4? TempDir only appeared in 3.2 https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

QwertyJack:
I'm using 2.7. Thank you very much!
Anyway, I hope Xonsh will support python2.

scopatz:
Thanks. As per Python 2, we have an open issue and anyone is free to tackle it :)

tmbdev:
What about mentioning on the home page that it is Python 3 only, and checking for that in the installer and giving a better error message?

gforsyth:
Hey @tmbdev -- I put in a PR to address your comment.  There has been a check in `setup.py` for supported Python version since way back, but it was never reached since we import a submodule that doesn't exist for Python < 3.2.  That should be fixed shortly.

scopatz:
Fir the record @tmbdev, two statements on the version are on the home page. The first is in the second sentence on the home page:

```
Xonsh is a Python-ish, BASHwards-looking shell language and command prompt. The language is a superset of Python 3.4+ with additional shell primitives that you are used to from Bash and IPython.
```

The second is in the dependencies listing: http://xon.sh/#dependencies

If you have ideas on how to increase the visibility of this statement, please let us know!  Happy to make the more clear if we can.

PR #704, opened 2016-02-25
add `tempfile` import to try statement to fail py27 gracefully

gforsyth:
one line change to move `from tempfile import TemporaryDirectory` inside
the try statement so that the install will reach the `sys.version_info`
check and let a user know that Legacy Python is not supported in xonsh.

as per @tmbdev's request in #646

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,8 +7,13 @@ from __future__ import print_function, unicode_literals
 import os
 import sys
 import json
+
 try:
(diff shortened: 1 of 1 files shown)


gforsyth:
No longer a one line change -- I realized that this could mess up installs for people who have Python 3 but not setuptools.

scopatz:
Thanks @gforsyth!

PR #705, opened 2016-02-25
Small addition to index.rst

gforsyth:
It's actually quite hard to figure out the right way to phoneticize `xonsh` in Japanese.  Which is also wonderful and kind of the point.


scopatz:
hahaha amazing!

PR #706, opened 2016-02-26
Ignore untracked files when considering if git workdir is dirty

melund:
This is up for discussion... Most of my other tools don't consider the workdir to be dirty if there are untracked files.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -35,6 +35,8 @@ Current Developments
   capabilities.
 * New ``Token.Color`` token for xonsh color names, e.g. we now use
   ``Token.Color.RED`` rather than ``Token.RED``.
+* Untracked files in git are ignored when determining if a git workdir is 
+  is dirty. This affects the coloring of the branch label. 
 
 **Deprecated:** None
 
--- a/docs/xonshrc.rst
+++ b/docs/xonshrc.rst
@@ -15,5 +15,35 @@ at startup. The following is a real-world example of such a file.
     :code: xonsh
 
 
+Snippets for xonshrc
+=========================
+The following are usefull snippets and code that tweaks and adjust xonsh in various ways.
+If you have any useful tricks, feel free to share them. 
 
-    
+Adjust how git branch label behaves
+---------------------------------
+Xonsh adds a colored branch name to the prompt when working with git or hg repositories. 
+This behavior can be controlled with the ``$PROMPT`` environment variable. See how to `customize the prompt`_ . 
+The branch name changes color if the work dir is dirty or not. This is controlled by the ``{branch_color}`` formatter string. 
+
+
+The following snippet reimplements the formatter also to include untracked files when considering if a git directory is dirty. 
+
+.. code-block:: xonshcon
+
+    >>> from xonsh.environ import git_dirty_working_directory
+    >>> $FORMATTER_DICT['branch_color'] = lambda: ('{BOLD_INTENSE_RED}'
+                                                   if git_dirty_working_directory(include_untracked=True)
+                                                   else '{BOLD_INTENSE_GREEN}')
+
+                                                   
+.. _customize the prompt: http://xon.sh/tutorial.html#customizing-the-prompt
+
+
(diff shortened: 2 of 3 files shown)


scopatz:
I am in favor of this change.  Does anyone feel that this shouldn't be merged?  Also @melund, could you add a note in the CHANGELOG please? Thanks for putting this in!

gforsyth:
I'm not sure. If I'm pedantic about it, if there are untracked files in
your repo then it IS dirty. If you create a new file and don't check it in
right away there's then no cue that it's not versioned.

Maybe this could be optional? Which would mean another ENVVAR but I like
ENVVARS...

I'm fine with @melund's proposal being the default but it should be
changeable without opening the source.

My two cents

scopatz:
We should probably just have two entries in the formatter dict then, one for porcelain and one for not.  I think that is the cleanest solution.

melund:
I am not sure what you mean. It is the `--untracked-files=no` that changes. What about an ENVVAR with the entire git command to test for a dirty folder. Then it is fully customizable ...

melund:
@scopatz .Ohh. Do you mean different `branch_color` entries??  But that is also used for Mercurial as well. Not sure how that works.. Will it not be kind of messy with several `branch_color`entries. 

Of course @gforsyth or anyone else could just add their own `branch_color` function to `$FORMATTER_DICT` in `xonshrc`. Thus, it is possible to implement other rules for what constitutes a dirty working directory without changing the source.

scopatz:
> @scopatz .Ohh. Do you mean different branch_color entries?? But that is also used for Mercurial as well. Not sure how that works.. Will it not be kind of messy with several branch_colorentries.

Yeah, that is what I mean.  I don't think that it would be super messy. I think that there are basically just these two cases: `branch_color` and `branch_color_porcelain`.

As per mercurial, we could tag @ngoldbaum or @MatthewTurk or someone who might know :)

ngoldbaum:
I'm not sure I completely follow the discussion, but the way `hg_dirty_working_directory` is implemented, it will say the working directory is clean if there are untracked files but no changes. I guess the proposed change would make things for git more in line with hg.

melund:
> Yeah, that is what I mean. I don't think that it would be super messy. I think that there are basically  just these two cases: `branch_color` and `branch_color_porcelain` 

I don't like that solution @scopatz . Firstly, we would need to call it them something more saying (the 'porcelain' command toggles machine-readable output and is not related to the untracked files). Could be `branch_color_ignore_untracked_files` and `branch_color_include_untracked_files`, but pretty ugly.

Secondly, we would have to implement the same thing for mercurial. 

I suggest adding an ´untrancked´ to `git_dirty_working_directory` to include untracked files. Then it is easy to make a custom formatter in xonshrc

```
def branch_color_mod():
    return ('{BOLD_INTENSE_RED}' if git_dirty_working_directory(untracked=True) else
            '{BOLD_INTENSE_GREEN}')
$FORMATTER_DICT['branch_color'] = branch_color_mod
```

Would that be OK for the more "pedantic" developers :)  @gforsyth

gforsyth:
Sounds good to me, @melund

ngoldbaum:
FWIW, implementing tracking whether the working directory is dirty for mercurial wouldn't be very hard. We would just need to go from parsing the output of "hg identify" to parsing the output of "hg summary", which contains this information.

scopatz:
Sounds good to me @melund! Thanks for the info @ngoldbaum.  Seems like it would be nice addition.

@melund it looks like this is mostly ready to go.  I think it would be nice to put that snippet that you wrote somewhere in the docs that we could just point people to when they ask about this functionality.  WOuld you mind putting this in? Maybe in xonshrc.rst?

melund:
No problem. I will look at it tomorrow.

melund:
@scopatz .. I tried to put in something like that in the xonshrc.rst. I thought it would be nice to have a section with snippets/tricks for the xonshrc file.

scopatz:
Looks great! Thanks @melund!

PR #709, opened 2016-02-27
fix linefeeds in xonfig wizard on windows

### LINKED ISSUE #707: Environment section in Wizard looks weird
It looks like the environment section of the wizard is formatted wrong. It looks like it is printing the line feed characters. 

image

The closest I could get to the problem is that this section is beeing set by the `YesNo` node. This node sets the entire text into the PROMPT variable. .
https://github.com/scopatz/xonsh/blob/f51227e90ad1144ff842e03c235ea5b0dbbb2ee3/xonsh/wizard.py#L529

scopatz:
Should fix #707.  @melund would you mind taking a look?


melund:
I haven't got time to sit down a review the code. But I seem to work fine. So if you are comfortable then just merge.

scopatz:
No rush from me! Whenever you or someone else gets a chance!

Issue #711, opened 2016-02-29
Regex globbing is misleading


BlaXpirit:
Filename expansion using regular expressions is pinned to the beginning of the filename, but not to the end. Some examples:

```console
$ ls
abac
$ `a.*a`
['abac']
$ `b.*c`
[]
$ `a`
['abac']
$ `b`
[]
$ `bac$`
[]
$ `.*bac$`
['abac']
```

I think that such arbitrary behavior is very inconvenient or even dangerous.

I know that under the hood this is caused by my arch-nemesis, [`re.match`](https://docs.python.org/dev/library/re.html#re.match). Please either pin both the beginning and the end of the string (as if the regex is `^regex$`, and to search anywhere in the filename you'd need to use `.*regex.*`), or, much better, just get rid of the pinning and use [`re.search`](https://docs.python.org/dev/library/re.html#re.search).

gforsyth:
Hey @BlaXpirit , this seems reasonable to me and I've put in a PR to address it.  Behavior is

```
$ ls
abac
$ `a.*a`
['abac']
$ `b.*c`
['abac']
$ `a`
['abac']
$ `b`
['abac']
$ `bac$`
['abac']
$ `.*bac$`
['abac']
```

BlaXpirit:
Now I see that [`re.fullmatch`](https://docs.python.org/dev/library/re.html#re.fullmatch) implements the `^regex$` behavior; I prefer this one.

gforsyth:
Hey @BlaXpirit -- I _think_ that should be ok, but I need to think about it a little more and test to make sure it doesn't screw with things like path completion.

BlaXpirit:
In regards to breaking the API (as mentioned in the PR):

I'm sure that more people "rely" on the behavior to be that of `re.fullmatch` than what it actually is and are already being bitten by it. Nobody would just expect such arbitrary behavior, especially if it is not mentioned in documentation.

scopatz:
Well, my question to you @BlaXpirit is whether you would mind having to set a `$XONSH_REGEX_GLOBBER` or similar environment variable in your xonshrc in order to get whatever behaviour you wanted?  I think this is probably the best solution because I can see use cases for all of these functions.

BlaXpirit:
Well sure, that would be the smallest step towards resolving this. But I don't see why the default would be having to write `regex$` for a full match and `.*regex` for a search, a behavior that isn't used by any major regex-related tool.

[...]

PR #712, opened 2016-02-29
swap re.match for re.search for better globbing

gforsyth:
as mentioned by @BlaXpirit in #711

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -253,12 +253,12 @@ def reglob(path, parts=None, i=None):
     if i1 == len(parts):
         for f in files:
             p = os.path.join(base, f)
-            if regex.match(p) is not None:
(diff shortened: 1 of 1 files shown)


scopatz:
This is an API change that could/would affect the actual result of xonsh code.  If we are going to change it to anything, it is probably best to change it to `fullmatch()`, like was mentioned in the original issue.  Also, it seems like this is the sort of thing that users might just want to configure, and so keeping `match()` as the default seems most sensible.  I don't share the same antipathy for match as @BlaXpirit :)

scopatz:
Thanks for putting in the PR though @gforsyth!

gforsyth:
No problem!  And I hadn't considered that it would change the xonsh API, which is something I don't care to do without a very good reason.

PR #714, opened 2016-03-01
Add optional dependency on the win_unicode_console library

melund:
It enables utf-8 support in windows terminals. ( https://pypi.python.org/pypi/win_unicode_console )

It is only active if win_unicode_console is installed, and it can be disabled by setting `$WIN_UNICODE_CONSOLE` to False. Setting the env-var also enables/disables on the fly.

We better test for a while our self ( @BYK ) to see if there are any unwanted side-effects. 

Install with:
```
conda install -c xonsh win_unicode_console
pip install win_unicode_console
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,9 @@ Current Developments
 * Python mode output is now syntax highlighted if pygments is available.
 * New ``$RIGHT_PROMPT`` environment variable for displaying right-aligned
   text in prompt-toolkit shell.
+* Optional dependency on the win_unicode_console package to enable unicode
+  support in cmd.exe on Windows. This can be disabled/enabled with the
+  ``$WIN_UNICODE_CONSOLE`` environment variable.
 
 **Changed:**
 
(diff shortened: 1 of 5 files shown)


melund:
Hmm. I will look at why it fails tomorrow.

melund:
I like how ['DEFAULT_ENSURERS'](https://github.com/scopatz/xonsh/blob/master/xonsh/environ.py#L58) can be used to implement ´setters´ on environment variables :)

scopatz:
I'm glad :)

scopatz:
Please let us know when you think it is ready!

melund:
Well. It is ready. But I would like to run with it for a while to make sure there are no unexpected side effect. On the other hand it is only active if people have actively installed win_unicode_console ... So maybe it OK to merge.....

scopatz:
Yeah, I think that this is ready to go in.  Could you make a mention in the docs and the CHANGELOG, please?

melund:
Sorry. I forgot about the change log. I also added some stuff to the windows.rst file. I think it is ready now @scopatz

scopatz:
Looks great! Thanks!

Issue #710, opened 2016-02-29
Semicolon-Separated Commands Fail with SyntaxError


moigagoo:
For example:
```
$ hg pull; hg up
           
  File "<string>", line None
SyntaxError: <xonsh-code>:1:3: ('code: pull',)
```

After I type `hg up; hg pull` and press Enter, a blank line is started. Another Enter press leads to the error.

I'm on Mac OS X El Capitan.

BYK:
Same here for some time now.

gforsyth:
Did this ever work?  `hg up; hgpull` will be in subprocess mode where `;` don't have particular meaning (I don't think).  If you force Python mode this works: `$[hg up]; $[hg pull]` which is a little silly.  Do you need semicolon support, or would you be ok with the soon-to-be-merged `and` operator in #672?  

On that PR the following works for me `git fetch and git merge`

scopatz:
I can confirm the failure as well.  I actually expected this to work.  There are even tests for [cases like this](https://github.com/scopatz/xonsh/blob/master/tests/test_tools.py#L147) that pass in the test suite but fail at the command line. I think that this is a legitimate bug.

scopatz:
After a a git bisect, the change that broke this behaviour is 97d8d4c69ec0dd89bef997397d5c5bb54011d390

scopatz:
I had remembered this was working before :)

scopatz:
Though, to be fair, I have no idea why this would rely on double wrapping an @adqm was right to remove it.  Note, that the code I used to test this was the equivalent `echo i; echo j`

moigagoo:
@gforsyth Semicolons should work. I remember somewhere Anthony proposed it as a workaround for not implemented `&&`. `and` is great and I'm going to use it, but I think semicolons should be supported as well to make it easier to switch from bash.

Also, @scopatz confirmed it's a legitimate bug :-)

adqm:
I just put in a PR that I think should fix.  The problem is that here we call the wrapping twice, but on different substrings.

So when we hit the `echo j` in `echo i; echo j`, we have already wrapped the first piece (we have `$[echo i]; echo j` at that point).  The original check would give up there, but this update should work.

scopatz:
Hi @moigagoo, @adqm's fix has been pulled into master. Would you mind testing it and then closing this issue? Thanks!

moigagoo:
[...]

PR #716, opened 2016-03-02
better check for avoiding double-wrapping with $[...]

adqm:
This slight change modifies the way we check for double-wrapping lines in `$[...]`, to avoid the issue in #710 while still preserving the benefits of the original fix.


scopatz:
Thanks a million @adqm! Works great for me.

PR #719, opened 2016-03-04
fix collision with ptk multiline dedent and `pass`

gforsyth:
This is a bit of an edge case, but the password manager `pass` causes a
few headaches with the multiline carriage return logic.  Specifically,
because `pass` is a `DEDENT_TOKEN` when running `pass` as the password
manager the prompt will delete one indent-length's worth of characters
on a `<ENTER>` press.  This is a one-line fix to make sure that people
 can use `pass` in the PTK shell by assuming (correctly) that if the buffer 
document is only one line long, then dedenting is meaningless.
Just running `pass` will be interpreted as a Pythonic `pass`, but
commands like `pass git init` will work.

If someone has a xsh script that uses `pass` it should be wrapped in
`$[]`.

### DIFF
--- a/tests/test_ptk_multiline.py
+++ b/tests/test_ptk_multiline.py
@@ -53,15 +53,32 @@ def test_colon_indent():
     assert_equal(buffer.document.current_line, indent_)
 
 def test_dedent():
-    document = Document(indent_+'pass')
+    document = Document('\n'+indent_+'pass')
     buffer.set_document(document)
     carriage_return(buffer, cli)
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks for putting this PR in @gforsyth!  CI is failing for some reason...

gforsyth:
Hey @scopatz -- ready for another look now.  And a reminder that tests test what you tell them to test.

scopatz:
Thanks @gforsyth!  Looks great!

PR #721, opened 2016-03-07
modify regular expression globbing

adqm:
This patch modifies regular expression globbing as per the discussion in #711 .  For experimentation/discussion.

Also fixes a separate issue I had with regex globbing, whereby returning an empty list would result in that argument disappearing from the args list.


scopatz:
This looks great @adqm. Would you mind adding something to the CHANGELOG about this, please? Also, I am not sure if there are other places in the docs where this needs to be mentioned or clarified.

adqm:
Alrighty; pushed a small update to the changelog.  I didn't think the documentation needed to be updated, but if you disagree, I can try to find a spot to put that in.

scopatz:
Thanks a ton!

Issue #726, opened 2016-03-08
Prompt colors are different between readline and prompt_toolkit 


asmeurer:
The readline shell looks like 

<img width="143" alt="screen shot 2016-03-07 at 9 25 48 pm" src="https://cloud.githubusercontent.com/assets/71486/13590097/2e3ad716-e4ab-11e5-9ff8-4f62888ba4e4.png">

The prompt_toolkit shell looks like

<img width="157" alt="screen shot 2016-03-07 at 9 26 15 pm" src="https://cloud.githubusercontent.com/assets/71486/13590100/38003520-e4ab-11e5-9da3-c07c55d8425b.png">

I am using BOLD_CYAN (see https://github.com/asmeurer/dotfiles/blob/a55ec501eed8e99de0e2f33d0a9031fa78569539/.xonshrc#L156).  My bash prompt looks like the readline one here, and is what I want.

scopatz:
You should simply be getting whatever the terminal sets as the color for [`1;36`](https://github.com/scopatz/xonsh/blob/master/xonsh/ansi_colors.py#L413).  You can pass in a hex color code (eg `#FFFFFF`) instead of `BOLD_CYAN` if you want.  Not sure why this is that different on readline and ptk.  I am assuming that you are on master.

asmeurer:
Yep, that's what the cyan text itself is saying :)

scopatz:
What color does the following give in both Bash and xonsh?

```console
$ echo -e "\x1b[1;36mHello\x1b[0m"
```

asmeurer:
They both give the lighter blue (what I believe is the correct one). FWIW xonsh also prints the `-e` (I guess that flag is only a part of the bash builtin `echo`, not the command `echo`).

scopatz:
Hmmm xonsh does not print the `-e` for me.  You might not have echo installed.

```console
scopatz@localhost ~ $ echo -e "\x1b[1;36mHello\x1b[0m"
Hello
scopatz@localhost ~ $ which echo
/bin/echo
```

scopatz:
Ok @Carreau has implemented an `#ansicolor` fix for the ptk shell once it propagates up to a pygments release.

asmeurer:
`-e` must be a nonstandard option. `/bin/echo -e "\x1b[1;36mHello\x1b[0m"` prints `-e \x1b[1;36mHello\x1b[0m` in bash for me (also my `man echo` doesn't mention any `-e` option).

[...]

PR #728, opened 2016-03-09
Fix colors on recent enough pygments

Carreau:
Step toward #726

xonsh

The mapping names are weird (cause of Pygments historical reason) , and there might be a small issue with the BOLD marker leaking. 

@asmeurer does that fix things for you ? 

(note you need pygments master, put it **should** be possible to avoid that. )

### DIFF
--- a/xonsh/pyghooks.py
+++ b/xonsh/pyghooks.py
@@ -498,25 +498,48 @@ BW_STYLE = {
 }
 _expand_style(BW_STYLE)
 
-DEFAULT_STYLE = {
-    Color.BLACK: '#000000',
-    Color.BLUE: '#0000AA',
-    Color.CYAN: '#00AAAA',
-    Color.GREEN: '#00AA00',
-    Color.INTENSE_BLACK: '#555555',
-    Color.INTENSE_BLUE: '#0000FF',
-    Color.INTENSE_CYAN: '#55FFFF',
-    Color.INTENSE_GREEN: '#00FF00',
-    Color.INTENSE_PURPLE: '#FF00FF',
-    Color.INTENSE_RED: '#FF0000',
-    Color.INTENSE_WHITE: '#aaaaaa',
-    Color.INTENSE_YELLOW: '#FFFF55',
-    Color.NO_COLOR: 'noinherit',
-    Color.PURPLE: '#AA00AA',
-    Color.RED: '#AA0000',
-    Color.WHITE: '#ffffff',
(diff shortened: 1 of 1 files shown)


xonsh/pyghooks.py
```diff
@@ -517,6 +517,28 @@ def _expand_style(cmap):
...
+        Color.INTENSE_PURPLE: '#ansipurple',
+        Color.INTENSE_RED: '#ansidarkred',
+        Color.INTENSE_WHITE: '#ansilightgray',
+        Color.INTENSE_YELLOW: '#ansibrown',
+        Color.NO_COLOR: 'noinherit',
+        Color.PURPLE: '#ansifuchsia',
+        Color.RED: '#ansired',
+        Color.WHITE: '#ansiwhite',
+        Color.YELLOW: '#ansiyellow',
+    }
```
scopatz: Could you move the other definition into an else-block please?

scopatz:
Thanks for putting this in @Carreau!

xonsh/pyghooks.py
Carreau: Done.

scopatz: Thanks!

scopatz:
I am happy with this, but I'll let @asmeurer weigh in.

melund:
It makes the colors better on windows as well, but only with the pygments master. 

@Carreau what did you mean with:
> ...it should be possible to avoid that. ).

Carreau:
the released version of Prompt-Toolkit support the new `#ansiblahblah`, so it should be possible to have this feature without requiring Pygments master.

melund:
Ok. But the current implementation does require pygments master. Right?: ´hasattr(pygments.style, 'ansicolors')´ 

Shouldn't check be made against prompt-toolkit instead?

Carreau:
> Ok. But the current implementation does require pygments master. Right?: ´hasattr(pygments.style, 'ansicolors')´
> 
> Shouldn't check be made against prompt-toolkit instead?

I'm sorry, I don't understand. 

This works with PTK stable, why do you want me to check something on PTK ?

melund:
Ahh. Is PTK adding the the ansicolors style to pygments? It didn't seem to have any effect until I ran pygments directly from master.  Maybe my version of PTK was too old (0.58). I will try with 0.59

scopatz:
I think this is fine and so I am going to go ahead and merge.

scopatz:
Thanks a ton @Carreau!

Issue #733, opened 2016-03-13
globpath seem not that useful


Carreau:
found here, https://github.com/scopatz/xonsh/blob/master/xonsh/built_ins.py#L276-L282

the code is dublicate of `iglobpath`, but using `glob` instead of `iglob`, which seems weird as glob is just convenient wrapper that consume `iglob`:

```python
def glob(pathname, *, recursive=False):
    """Return a list of paths matching a pathname pattern.

    The pattern may contain simple shell-style wildcards a la
    fnmatch. However, unlike fnmatch, filenames starting with a
    dot are special cases that are not matched by '*' and '?'
    patterns.

    If recursive is true, the pattern '**' will match any files and
    zero or more directories and subdirectories.
    """
    return list(iglob(pathname, recursive=recursive))
```

Plus is seem to only be used there: `builtins.__xonsh_glob__ = globpath`.

So it might be useful to deprecate `globpath` or investigate making it only a wrapper around iglobpath if ever the logic in iglobpath is ever extended.

scopatz:
Thanks for bringing this up @Carreau. I think it is best to continue on with this from your PR

scopatz:
@Carreau is this closable now?

Carreau:
sure.

Issue #451, opened 2015-10-30
Double star (**.py) globbing


Carreau:
One of the feature that still makes me go back to fish is the double star wildcard/globing, that automatically expand to subdirectories.

[glob2](https://pypi.python.org/pypi/glob2) could I guess help for that.

scopatz:
Thanks for reporting @Carreau. I thought that this was in xonsh already, but I guess I am wrong.  Pull requests welcome :)  This shouldn't be too hard to implement. We already do some custom glob manipulation. I think it is in tools or `built_ins`

Carreau:
Well, sorry title is imprecise. `**/*.py` works but `**.py` does not, I typed too fast.

Carreau:
and the difference is that `**.py` will also match all py file in cwd :+1: 
see fish difference (`setup.py`):

```
$ ls -1 **/*.py | head
docs/conf.py
tests/test_aliases.py
tests/test_builtins.py
tests/test_dirstack.py
tests/test_environ.py
tests/test_execer.py
tests/test_foreign_shells.py
tests/test_history.py
tests/test_imphooks.py
tests/test_lazyjson.py
~/d/xonsh (master) $ ls -1 **.py | head
docs/conf.py
setup.py   <- this is globbed.
tests/test_aliases.py
tests/test_builtins.py
tests/test_dirstack.py
tests/test_environ.py
tests/test_execer.py
tests/test_foreign_shells.py
tests/test_history.py
tests/test_imphooks.py
```

I know it's a detail, and not really a "bug', but that annoy me :-)

asmeurer:
I'd also like to see support for bash's extended globbing (http://www.linuxjournal.com/content/bash-extended-globbing). The negated wildcards are particularly useful. It doesn't look like glob2 supports that. Maybe there is some other library. Or maybe xonsh should just implement its own globbing.

scopatz:
> Or maybe xonsh should just implement its own globbing.

I am most in favor of this, to keep the number of dependencies down.

scopatz:
Very willing to borrow other people's code, though, given the right license.

krader1961:
[...]

PR #734, opened 2016-03-13
Improve double-star globing.

Carreau:
Not as advance as using a package like glob2, but allow simple use of
double start to match recursively in subdirectory.

e.g:
    $ ls -1 **p.py
    docs/cmdhelp.py
    setup.py

Which match both in CWD and Subdirectory.

See #451,

Also consolidate `globpath` and `iglobpath` (see #733)

---- 

I'm sure there are probably side effects, and It's just one of the annoying behavior that upset me,  so i'm just submitting for people to try it out.

Also the `recursive=True` kwarg-only is only a 3.5+ thing. So it might not be a good thing to get in now.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -96,3 +96,15 @@ manually use the ``$[]`` or ``$()`` operators on your code.
 Yes, context-sensitive parsing is gross. But the point of xonsh is that it uses xontext-sensitive parsing and 
 is ultimately a lot less gross than other shell languages, such as BASH.
 Furthermore, its use is heavily limited here.
+
+
+
+6. Gotchas
+----------
+
+There are a few gotchas when using xonsh across multiple versions of Python,
+where some behavior can differ, as the underlying Python might behave
+differently.
+
+For example double star globbing `**` will only work on Python 3.5+ (ie not on 3.4)
+as recursive globbing is `new in Python 3.5 <https://docs.python.org/3/library/glob.html#glob.glob>`_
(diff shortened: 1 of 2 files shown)


scopatz:
This is great @Carreau! I only have two notes:

1. Could you please add something to the changelog.
2. Could you put in a check or mask or something for v3.5 and `recursion=True`. You probably want to import some version info from [`xonsh.tools`](https://github.com/scopatz/xonsh/blob/master/xonsh/tools.py#L69)

With those, I think it would be ready to go in.

scopatz:
I think this now has a merge conflict with master.

Carreau:
Sorry was travelling, will take care of that again at some point !

scopatz:
No worries!  Whenever you get a chance!

scopatz:
ping, and I think that this now has conflicts with master.

Carreau:
Sorry, will get back to that !

Carreau:
Rebased and py 3.5 conditional added + a tiny bit of doc. 
Beware that `**` will behave differently on 3.4 and 3.5. 
(have you considered dropping 3.4 ?)

scopatz:
> have you considered dropping 3.4?

Nope, not yet.  We'll probably drop v3.4 whenever conda-forge does.

scopatz:
Or at least stop building packages.  There is no harm in keeping v3.4 compatible code around for the time being.

scopatz:
Thanks a million @Carreau!

PR #735, opened 2016-03-13
Fix pronunciation of `conch` in readme

Carreau:
My phonetic is not perfect, and might carry a bit of my French accent,

Would like @scopatz review on that.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -23,6 +23,7 @@ the xonsh shell
         "It is pronounced <i>quanxh</i>",
         "It is pronounced <i>zonsch</i>",
         "It is pronounced <i>jeaunsch</i>",
+        "It is pronounced <i>🐚</i>",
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -17,7 +17,7 @@ xonsh
 
 xonsh is a Python-ish, BASHwards-looking shell language and command prompt.
 The language is a superset of Python 3.4+ with additional shell primitives.
-xonsh (pronounced *conch*) is meant for the daily use of experts and novices 
+xonsh (pronounced *🐚*) is meant for the daily use of experts and novices
```
scopatz: I think this breaks the word play.

scopatz: And this would prefer to keep it the way it is :).

Carreau:
Ok, just added the shell in the docs.

scopatz:
thanks!

PR #736, opened 2016-03-16
minor fixes for !() and ![]

scopatz:
probably for @adqm to review, though other folks are welcome to as well!

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -569,9 +569,9 @@ def run_subproc(cmds, captured=False):
                     e = 'xonsh: subprocess mode: permission denied: {0}'
                     raise XonshError(e.format(cmd[0]))
         _stdin_file = None
-        if (captured == 'object' and
-                stdin is not None and
-                __xonsh_env__['XONSH_STORE_STDIN'] and
+        if (stdin is not None and
+                ENV.get('XONSH_STORE_STDIN') and
+                captured == 'object' and
                 'cat' in __xonsh_commands_cache__ and
                 'tee' in __xonsh_commands_cache__):
             _stdin_file = tempfile.NamedTemporaryFile()
@@ -775,10 +775,9 @@ def load_builtins(execer=None, config=None):
     # would be nice to actually include non-detyped versions.
     builtins.__xonsh_history__ = History(env=ENV.detype(),
                                          ts=[time.time(), None], locked=True)
-    lastflush = _lastflush
-    atexit.register(lastflush)
+    atexit.register(_lastflush)
     for sig in AT_EXIT_SIGNALS:
-        resetting_signal_handle(sig, lastflush)
+        resetting_signal_handle(sig, _lastflush)
     BUILTINS_LOADED = True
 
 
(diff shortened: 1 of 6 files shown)


xonsh/built_ins.py
```diff
@@ -569,9 +569,9 @@ def run_subproc(cmds, captured=False):
                     e = 'xonsh: subprocess mode: permission denied: {0}'
                     raise XonshError(e.format(cmd[0]))
         _stdin_file = None
-        if (captured == 'object' and
-                stdin is not None and
-                __xonsh_env__['XONSH_STORE_STDIN'] and
+        if (stdin is not None and
+                ENV.get('XONSH_STORE_STDIN') and
```
adqm: I don't really understand the insistence to make this a `get` call since it won't change the behavior at all (unless someone removes `XONSH_STORE_STDIN` from the _defaults_ dictionary).  But it doesn't bother me at all.

Is there something I'm missing?

xonsh/proc.py
```diff
@@ -552,8 +550,6 @@ def __bool__(self):
     def __eq__(self, other):
         return self.returncode == other
 
-    def __hash__(self):
-        return hash(self.returncode)
```
adqm: I think we want to keep this around.

For me, the ability to do `!(echo "hi") in {1,2,3}` is worth the possibility of hash collisions if someone wants to use these things as keys in a dictionary or store them in a set.  But I'm open to discussion.

adqm:
Looks good to me.  Just the couple of inline thoughts.

xonsh/built_ins.py
scopatz: It is just a consistency thing. On the `andor` branch I have cleaned up a lot of these thing.  If we wanted to move to indexing everywhere in the code base, that could also be OK

scopatz: Sorry if I am overly obsessive about this :)

xonsh/proc.py
scopatz: Wouldn't `!(echo "hi").returncode in {1,2,3}` be effectively the same thing?

scopatz: I wouldn't mind adding a `rtn` property to `CompletedCommand` to shorten this up.

adqm: I'm not sure I agree, but that's fine.  I thought it was nice to be able just to compare the object directly against the numbers.  But typing `.rtn` isn't too much extra hassle.  If it ends up being too painful, we can always add these things back in later.

In that case, let's add `rtn` as a short-hand for `returncode`.

And if we're getting rid of `__hash__`, I think we should also get rid of `__eq__` (and update the documentation appropriately) for the sake of consistency.

xonsh/built_ins.py
adqm: No, not at all.  I'm all for consistency.  Wasn't aware that `get` was being used everywhere else.

scopatz:
Ok @adqm, I have updated with respect to your requested changes.  I also added some other aliases to CompletedCommand have shortcuts that are the same as the history field names.

adqm:

I may suggest some other small tweaks in a separate PR later.

Issue #741, opened 2016-03-16
Strange Redirection Problems with xonsh in mlterm


xee5ch:
I cannot tell if this is something strange I have done with my mlterm configuration, which is minimal, or my new xonsh configuration. I usually use emacs as a term multiplexer these days, and mlterm was a hold-over for complex script languages I read and sometimes use. I am unable to delete a file when prompted, it immediatetly throws an error when prompted for user input.

```
me@computer ~ $ mlterm -v
mlterm version 3.7.0
me@computer ~ $ import xonsh
me@computer ~ $ xonsh.__version__
'0.2.6'
me@computer ~ $ import prompt_toolkit
me@computer ~ $ SHELL_TYPE=prompt_toolkit
me@computer ~ $ prompt_toolkit.__version__
'0.60'
me@computer ~ $ rm ~/Code/conf.py
/usr/bin/rm: remove regular file '/home/me/Code/conf.py'?
*[1] stopped: rm /home/me/Code/conf.py & (10123)
TypeError: unorderable types: NoneType() > int()

For the last line, if I redirect output with xclip to speed up copy-paste for this ticket, redirection magically is fine and I can actually answer yes or no.


me@computer ~ $ rm ~/Code/conf.py | xclip -sel clip
/usr/bin/rm: remove regular file '/home/me/Code/conf.py'?
```

My dotfiles are in my repo. I just created the most minimal xonsh possible from my minimal needs with `xonfig`.

```
{
 "foreign_shells": [
  {
   "login": true,
   "shell": "bash"
  },
  {
   "shell": "bash"
  }
 ]
}
```

Did I wrap two login shells into each other or something basic I forgot? I have tried running mlterm straight with the default bash shell, but also tried bypassing this and running just xonsh with `mlterm -e xonsh` and I get the same behavior on all attempts.

gforsyth:
Hey @ajstein 
In your `.bashrc` or other `rc` files, do you have an alias for `rm` to `rm -I`?  This might be related to #727

ajstein:
As a matter of fact, yes.  I had not even recalled.

https://github.com/ajstein/dotfiles/blob/personal_linux/.bash_aliases#L3

```
# general aliases
alias ls='ls --color=auto'
alias rm='rm -iv '

```

Does this mean I should proceed with the `andor` branch?

gforsyth:
[...]

PR #742, opened 2016-03-16
[WIP] job control for Python function aliases, more built-in aliases

adqm:
I was a little bothered by this:

```
hartz@lappy-514 ~/test $ timeit echo "hi"
<<<snip>>>
100 loops, best of 3: 6.35 ms per loop
```

so I got the thought that we could/should implement some things within xonsh (as bash does) to avoid the considerable overhead of launching a new process.

With this update, I get roughly 10x speed increase:

```
hartz@lappy-514 ~ $ timeit echo "hi"
<<<snip>>>
1000 loops, best of 3: 486 µs per loop
```

Obviously, I would plan to implement several other commands directly as well.  But I didn't want to tear too far down this path before gathering opinions.


xonsh/coreutils/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .echo import echo
```
scopatz: Can you make this an absolute import please?

scopatz:
Awesome, thanks @adqm!  Just the little import issue.

setup.py
```diff
@@ -142,7 +142,7 @@ def main():
         url='https://github.com/scopatz/xonsh',
         platforms='Cross Platform',
         classifiers=['Programming Language :: Python :: 3'],
-        packages=['xonsh', 'xonsh.ptk', 'xonsh.parsers'],
+        packages=['xonsh', 'xonsh.ptk', 'xonsh.parsers', 'xonsh.xoreutils'],
```
gforsyth: Perfect.

gforsyth:
Hey @adqm -- given the strange behavior of `rm` in #741, what do you think about doing a runaround on it and implementing `rm` in `xoreutils`?  Not foisting that job on you, just curious what you think of it.  Cheers!

adqm:
@gforsyth, yep, definitely part of the plan here! :)

We should still try to find and fix the root cause of that issue, but I think having a faster rm doesn't hurt anyway.

xonsh/xoreutils/cat.py
```diff
@@ -0,0 +1,67 @@
...
+                _r = b'%s$' % _r
+            try:
+                print(_r.decode('unicode_escape'), flush=True, file=out)
+            except:
+                pass
+        return False
+
+
+def cat(args, stdin, stdout, stderr):
+    global line_count
```
scopatz: Does `line_count` really need to be global?  It seems like it could be passed into and returned by `_cat_single_file()`

adqm:
Also, if anyone digs in here, note that this is still very much a work in progress.  I have made what I consider to be some good progress, but to really make this work will require, among other things, modifying the way `ProcProxy` is handled, since we need these to be killable via `^C`.

I suppose many of the commands that cannot run forever (`echo`, `pwd`, `tty`, `rm`, `mkdir`, etc) can be implemented without any real issues.  But I really want `cat` to be here :) and maybe it's time to revisit `ProcProxy` anyway...

xonsh/xoreutils/util.py
```diff
@@ -0,0 +1,15 @@
+def arg_handler(args, out, short, key, val, long=None):
```
scopatz: Maybe call this file `tools.py` to be consistent with the `xonsh/tools.py`. This is not a blocker, though.

scopatz:
Thanks a ton for putting this in @adqm.  It pretty much looks good to me, though I haven't actually tried it out in a few commits.  This definitely needs some more docs though :)

adqm:
Cool.  I think this one needs quite a lot more work, so not ready to merge in yet.  It's fine to test, but when it is ready for a real serious look, I'll ping you and remove the `[WIP]` from the title.

scopatz:
Sounds good.

scopatz:
Also ping!

Will-W:
Is there a plan for enabling/disabling these commands? Something like 'enable' on bash, but Ideally with the option of 'all'? Another approach would be a setting for opt-in/opt-out.

I can see the performance benefits, but there might be some behavioural changes in some cases that people may want to work around (particularly if you start going down the road of things like ping that have wildly differing arguments and behaviour on different platforms).

scopatz:
> Is there a plan for enabling/disabling these commands? Something like 'enable' on bash, but Ideally with the option of 'all'? Another approach would be a setting for opt-in/opt-out.

There are a few ways we could conceive of this.

1. there could be an env var that determined whether these are loaded on startup.
2. there could be a xontrib that loads them if you have it set
3. we could add the ability to unload xontribs
4. people would need to import and call a function to unload them.

funkyfuture:
given that these are all cli applications, i'd suggest to include a framework for these to ease and streamline development of them. [click](http://click.pocoo.org/5/) seems appropriate for that.

since hard dependencies are a no-go, it could be included (which would currently mean to double the code size) or an own module for argument parsing and help text generation could be established.

xonsh/xoreutils/util.py
funkyfuture: well, the most common convention is to call helper modules `utils.py` (mind the plural).

funkyfuture: from a user-perspective end-user tools would be expected in `tools`, imo.

xonsh/xoreutils/__init__.py
```diff
@@ -0,0 +1,6 @@
+from xonsh.xoreutils.echo import echo
+from xonsh.xoreutils.pwd import pwd
+from xonsh.xoreutils.cat import cat
+from xonsh.xoreutils.tty import tty
+from xonsh.xoreutils.tee import tee
+from xonsh.xoreutils.yes import yes
```
funkyfuture: why not relative imports?

melund: From Pep8

Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path ):

import mypkg.sibling
from mypkg import sibling
from mypkg.sibling import example

melund: Also see: http://xon.sh/devguide.html

funkyfuture: ah, i wasn't aware of that. `flake8` is relaxed there.

interesting read: https://www.python.org/dev/peps/pep-0328/

adqm:
> Is there a plan for enabling/disabling these commands?

Definitely, including `enable` is part of the plan.  We can also modify it to have an `--all` option.

> given that these are all cli applications, i'd suggest to include a framework for these to ease and streamline development of them
> or an own module for argument parsing and help text generation could be established.

I see the point here.  I am against bringing in another dependency (whether we include it or not).  We have a little bit of this in `xoreutils.util.py`, but that could be modified to create the help-text as well so we aren't duplicating work.

funkyfuture:
i would put such cli-helpers in a higher namespace. i'll make a review of `click` and check whether it's worth an effort to extract useful functionality.

adqm:
Okay, I haven't forgotten about this, but I have gone down several wrong turns in the process.  Also including a bad merge (apparently), since a whole lot of extra history is showing up here.  I think I have a plan now that should work in all cases.  Will try to implement in the near future (and rebase).  I'll let you know, @scopatz, when this is ready for another serious look.

scopatz:
Great! Thanks! I look forward to it!

adqm:
Okay, _almost_ everything working now.  Just one more bug (that I know of) to squash.

Being able to use job control for Python-function aliases is really nice :)

scopatz:
Sweet! @adqm, I was planning on doing a v0.3.1 release today, but do you think it is worth waiting until tomorrow to get this in?

xonsh/built_ins.py
```diff
@@ -720,6 +724,7 @@ def load_builtins(execer=None, config=None, login=False, ctx=None):
     builtins.__xonsh_execer__ = execer
     builtins.__xonsh_commands_cache__ = CommandsCache()
     builtins.__xonsh_all_jobs__ = {}
+    builtins.__xonsh_enabled_builtin_commands__ = set(all_builtin_commands.keys())
```
scopatz: I think that the word builtin is getting a bit too overloaded here.  Maybe xoreutils instead? Or maybe that is too restrictive for the vision here?

xonsh/built_ins.py
```diff
@@ -466,6 +467,8 @@ def run_subproc(cmds, captured=False):
             stderr = builtins.__xonsh_stderr_uncaptured__
         uninew = (ix == last_cmd) and (not _capture_streams)
         alias = builtins.aliases.get(cmd[0], None)
+        if alias is None and cmd[0] in builtins.__xonsh_enabled_builtin_commands__:
+            alias = all_builtin_commands[cmd[0]]
```
scopatz: I am not sure I follow why just using aliases isn't good enough.  This seems like it creates a secondary alias system...

xonsh/proc.py
```diff
@@ -368,6 +372,163 @@ def __init__(self, f, args, stdin=None, stdout=None, stderr=None,
...
+                        if len(x) != 0:
+                            os.write(self.oerrwrite, x)
+                    except:
+                        pass
+                if p.returncode is not None:
+                    return
+            signal.signal(signal.SIGTSTP, signal.SIG_IGN)
+            return p.returncode
+
+        self.p = Process(target=wrapper, args=(args, self.p2cread, self.c2pwrite, self.errwrite, universal_newlines))
```
scopatz: PEP8 line length

xonsh/proc.py
```diff
@@ -368,6 +372,163 @@ def __init__(self, f, args, stdin=None, stdout=None, stderr=None,
         f = wrap_simple_command(f, args, stdin, stdout, stderr)
         super().__init__(f, args, stdin, stdout, stderr, universal_newlines)
 
+
+class ControllableProcProxy:
```
scopatz: A test for this would be nice :)

xonsh/xoreutils/__init__.py
```diff
@@ -0,0 +1,11 @@
+import importlib
+
+from xonsh.xoreutils.util import all_builtin_commands
+
+cmd_names = {'cat', 'pwd', 'tee', 'tty', 'yes', 'echo', 'which',
```
scopatz: PEP8, shouldn't this be in UPPER since it is a global variable?

xonsh/xoreutils/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+             'enable'}
+
+
+for i in cmd_names:
+    cmd = importlib.__import__('xonsh.xoreutils.%s' % i, globals(), locals(), [i], 0)
```
scopatz: Also, API doc hooks for all of these are needed :)

scopatz: Also I think you should be using [`import_module()`](https://docs.python.org/3/library/importlib.html#importlib.import_module) as per [`__import__()`](https://docs.python.org/3/library/importlib.html#importlib.__import__)'s docstring.

xonsh/dirstack.py
```diff
@@ -15,14 +15,24 @@ def _get_cwd():
         return None
 
 
+def _splitpath(path, sofar=[]):
```
scopatz: The default value shouldn't be mutable, due to Python's side effect issues in this case.

xonsh/xoreutils/cat.py
```diff
@@ -0,0 +1,104 @@
...
+        f = stdin
+    elif os.path.isdir(fname):
+        print("cat: {}: Is a directory.".format(fname), file=err)
+        return True, line_count
+    elif not os.path.exists(fname):
+        print("cat: No such file or directory: {}".format(fname), file=err)
+        return True, line_count
+    else:
+        f = open(fname, 'rb')
+    sep = os.linesep.encode()
```
scopatz: I think all of these encode and decode calls should use `$XONSH_ENCODING` and `$XONSH_ENCODING_ERRORS`

xonsh/xoreutils/echo.py
```diff
@@ -0,0 +1,26 @@
+def echo(args, stdin, stdout, stderr):
+    opts = _parse_args(args)
+    if opts is None:
+        return
+
+    ender = opts['end']
+
+    args = map(str, args)
+    if opts['escapes']:
+        args = map(lambda x: x.encode().decode('unicode_escape'), args)
```
scopatz: Same comments on the encode/decode.  Also, probably best to just define the lambda globally once.

xonsh/built_ins.py
adqm: I think so.  Most (?) other shells have some notion of built-in commands, which is somewhat distinct from coreutils.  That said, I think the vision is actually moving away from just implementing coreutils, toward better handling of built-in commands more generally.

xonsh/built_ins.py
adqm: I think it's more conventional to draw a distinction between the built-in commands and user-defined aliases.  I feel like this distinction is made in most other shells, at least...

Then again, we already kind of abuse that distinction with aliases that are implemented as Python functions.

I would personally like to move a lot of things from `aliases` to built-in commands, as that makes more sense to me.

This structure also makes it easier to implement the `enable` command.

adqm:
> but do you think it is worth waiting until tomorrow to get this in?

Maybe, yes, if that's not too much trouble.  I'd like to see if I can't get these last few things sorted out today.

xonsh/dirstack.py
adqm: I should really be using tuples instead here, as I'm not relying on the mutability anywhere.

xonsh/xoreutils/util.py
scopatz: Users are welcome to import and use these tools too :)

xonsh/built_ins.py
adqm: That said, xoreutils might still be fine.  I think you're right about builtins meaning a whole bunch of different things right now.

scopatz: Yeah, I guess that I am not totally sold on the notion of builtin commands distinct from aliases. Maybe that means that aliases should be renamed :).  Basically my objections are that (1) it takes away from some of the agency of the users since it seems like they shouldn't touch builtins, and (2) it becomes another thing for folks to learn, which makes it harder to teach. I, of course, could be wrong about this :)

I guess the problem that you are trying to solve with this is that you want users to be able to load/unload these commands en masse?  Is that your main concern?

adqm:
This will also require some changes to the documentation, I believe, since it currently draws a bigger distinction between "simple" and "normal" proc proxies (simple ones cannot be job controlled, other ones can).

xonsh/built_ins.py
scopatz: see other comment.

scopatz:
> This will also require some changes to the documentation, I believe, since it currently draws a bigger distinction between "simple" and "normal" proc proxies (simple ones cannot be job controlled, other ones can).

Ahh yeah - that needs noting.

> Maybe, yes, if that's not too much trouble. I'd like to see if I can't get these last few things sorted out today.

OK! Though I think that @melund and @gforsyth should probably also take a look at this PR too.

I think this is shaping up very nicely.  I think that the builtin/alias issue shouldn't be blocking. Though I do think we should have a broader architecture discussion on the topic.

xonsh/built_ins.py
adqm: > I guess the problem that you are trying to solve with this is that you want users to be able to load/unload these commands en masse?

Not necessarily en masse, but on demand (same as `enable` works in bash).

I guess my main concern is that it feels like it would be nice to keep user-defined aliases separate from built-in features of the shell.

> (1) it takes away from some of the agency of the users since it seems like they shouldn't touch builtins

I don't agree with this.  Since aliases are loaded first, it is still just as easy to override the behavior of a built-in (just add an alias, as before!).

> (2) it becomes another thing for folks to learn, which makes it harder to teach.

I think the change would be transparent to most users, since aliases still work.  The only interaction users would have to have with it is via `enable`, if necessary.

I guess part of my disagreement here comes from the name "aliases," which seems (to me) to disagree with how other shells use that term.  But I can't come up with a better name.  commands?

scopatz: Yeah, we are definitely abusing terms all over the place here :)  I kind of just want a single point of entry, and aliases is bad for that.  `commands` is what I thought too, but it isn't as snappy as I would like.  It is worth thinking about.

> I don't agree with this. Since aliases are loaded first, it is still just as easy to override the behavior of a built-in (just add an alias, as before!).

Yeah, I think my objection is on the other side of this. Say one wants to remove a command that is masking an installed binary (eg cat), do you check these builtins or aliases?  I agree that the story on adding masks and mapping is fine as it is.

scopatz: I know that I have been bitten by this removal/disable issue before in bash, and think that it can be done better :)

adqm: The removal would just be:
```
$ enable -n cat
```

No need to know where it lives.

Or maybe you're saying it could be confusing that `enable` wouldn't work on aliases?  To me, that's a reason to make sure `aliases` only contains user-specified aliases.

scopatz: While I am saying that it is somewhat confusing that enable wouldn't on aliases, I believe it to be much more confusing that disable/del wouldn't work on aliases and builtin commands exactly the same. Additionally, that it could be possible to have a builtin and an alias with the same name, is also harder to explain and understand the rules for. 

I guess I am in favor of keeping the core model as simple as possible: one environment, one place where commands go, one interface to color management, etc. I think the benefits to teaching the language for keeping things simple are hard for me to understate.  It is late so the only counter example I can think of is something like if Python had `sys.modules` for most module imports and then another `sys.builtin_modules` for thing which Python itself loaded.  As another, potentially bad example, C++ does something like this with namespaces, where anyone is free to modify any namespace - except `std` which is somehow sacred.

Again, I am happy to put in a PR that changes the name or aliases and related code to something more generic and semantically meaningful to the full suite of uses cases it has been asked to serve. 

I guess another option would be to make aliases, or whatever we call it, a [ChainMap](https://docs.python.org/3/library/collections.html#collections.ChainMap) or similar. This would allow a distinction between command types, and different references to them.  User commands could live in the first map of the chain, these builtins would be the second, etc. But fundamentally, the `run_subproc()` function would only look things up in one mapping, users would only need to know about one mapping, etc.

I hope that some this makes sense :). I'd like for all of us to be happy with the underlying model, too!

scopatz: Also, again, I think that this kinda stuff shouldn't impact the merging of this PR

adqm: Okay, that makes sense to me.  I'll go ahead and make those changes here.

adqm:
@scopatz: unfortunately, I still haven't been able to squash that bug yet (and not for lack of trying).  I'm going to take another look today (and make the changes related to builtins vs aliases), but if it isn't ready, I think it's probably best not to let this hold up the release.

melund:
@adqm. The `ControllableProcProxy` is giving me some problems:

```
snail@sea ~\Documents\xonsh adqm-coreutils $ enable -h
  File "c:\users\mel\documents\xonsh\xonsh\base_shell.py", line 154, in default
  File "c:\users\mel\documents\xonsh\xonsh\codecache.py", line 57, in run_compiled_code
    func(code, glb, loc)
  File "c:\users\mel\documents\xonsh\xonsh\built_ins.py", line 675, in subproc_captured_hiddenobject
  File "c:\users\mel\documents\xonsh\xonsh\built_ins.py", line 523, in run_subproc
    universal_newlines=uninew)
  File "c:\users\mel\documents\xonsh\xonsh\proc.py", line 465, in __init__
    self.p.start()
  File "C:\Users\mel\Anaconda3\envs\x\lib\multiprocessing\process.py", line 105, in start
    self._popen = self._Popen(self)
  File "C:\Users\mel\Anaconda3\envs\x\lib\multiprocessing\context.py", line 212, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
  File "C:\Users\mel\Anaconda3\envs\x\lib\multiprocessing\context.py", line 313, in _Popen
    return Popen(process_obj)
  File "C:\Users\mel\Anaconda3\envs\x\lib\multiprocessing\popen_spawn_win32.py", line 66, in __init__
    reduction.dump(process_obj, to_child)
  File "C:\Users\mel\Anaconda3\envs\x\lib\multiprocessing\reduction.py", line 59, in dump
    ForkingPickler(file, protocol).dump(obj)
AttributeError: Can't pickle local object 'ControllableProcProxy.__init__.<locals>.wrapper'
snail@sea ~\Documents\xonsh adqm-coreutils $ Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "C:\Users\mel\Anaconda3\envs\x\lib\multiprocessing\spawn.py", line 106, in spawn_main
    exitcode = _main(fd)
  File "C:\Users\mel\Anaconda3\envs\x\lib\multiprocessing\spawn.py", line 116, in _main
    self = pickle.load(from_parent)
EOFError: Ran out of input
snail@sea ~\Documents\xonsh adqm-coreutils $

```

adqm:
Gack!

Okay, I guess there are some differences in how processes are forked in GNU/Linux and in Windows that I wasn't expecting.  I will need to rework this substantially, then, since I was relying on that wrapped function being available in both the parent and the child process, and it looks like Windows can't handle that.

Ugh.  This will have to wait until I get near a Windows machine to test, as well...what a disappointment.

scopatz:
Sorry about this @adqm!

xonsh/xoreutils/cat.py
```diff
@@ -0,0 +1,104 @@
...
+    if fname == '-':
+        f = stdin
+    elif os.path.isdir(fname):
+        print("cat: {}: Is a directory.".format(fname), file=err)
+        return True, line_count
+    elif not os.path.exists(fname):
+        print("cat: No such file or directory: {}".format(fname), file=err)
+        return True, line_count
+    else:
+        f = open(fname, 'rb')
```
ScoreUnder: Don't use `os.path.*`. Instead catch exceptions here. Between your "exists" call and your "open" call, the file may have been removed (or a symlink could have beene swung to point to a directory, etc etc). These error messages should come as a result of an exception trying to open the file.

Compare the behaviour of this:
```
try:
    open(fname, 'rb')
except IOError as err:
    print("cat: {}: {}".format(err.strerror, fname), file=err)
```

xonsh/xoreutils/yes.py
```diff
@@ -0,0 +1,25 @@
+import os
+import sys
+
+
+def yes(args, stdin, stdout, stderr):
+    if '--help' in args:
```
ScoreUnder: Compare with GNU's `yes`:

```
/bin/yes --help | head  # Shows help
/bin/yes blah --help | head  # Normal yes behaviour
/bin/yes --blah --help | head  # Raises error
/bin/yes -- blah blah | head  # Normal yes behaviour (but removes "--")
/bin/yes -- --help | head  # Normal yes behaviour (but removes "--")
/bin/yes -- --blah --help | head  # Normal yes behaviour (but removes "--")
/bin/yes -- -- -- | head  # Normal yes behaviour (final string printed is "-- --")
```
Essentially:

* Switches are parsed only up until the first non-switch
* `--` marks the end of switches (only if they haven't ended already)
* Invalid switches raise an error.

These behaviours are incompatible with an approach relying on `in`, which will always be brittle and will result in the builtin `yes` being incapable of printing some strings.

xonsh/xoreutils/yes.py
```diff
@@ -0,0 +1,25 @@
...
+        print(HELP_STR, file=stdout)
+        return 0
+
+    to_print = ["y"] if len(args) == 0 else [str(i) for i in args]
+
+    while True:
+        print(*to_print, file=stdout)
+
+    return 0
```
ScoreUnder: Unreachable

xonsh/xoreutils/tty.py
```diff
@@ -0,0 +1,42 @@
...
+        fd = stdin.fileno()
+    except:
+        fd = sys.stdin.fileno()
+    if not os.isatty(fd):
+        if not silent:
+            print('not a tty', file=stdout)
+        return 1
+    if not silent:
+        try:
+            print(os.ttyname(fd), file=stdout)
```
ScoreUnder: As with `cat`, I think this should handle exceptions rather than using guards. As I've often heard, "it's easier to ask forgiveness than to ask permission".

With that in mind, here's the "ERRORS" section of the man page for `ttyname(3p)`:

```
       The ttyname() function may fail if:

       EBADF  The fildes argument is not a valid file descriptor.

       ENOTTY The file associated with the fildes argument is not a terminal.

       The ttyname_r() function may fail if:

       EBADF  The fildes argument is not a valid file descriptor.

       ENOTTY The file associated with the fildes argument is not a terminal.

       ERANGE The  value of namesize is smaller than the length of the string to be returned including
              the terminating null character.
```

Therefore, the standard states that if you get "ENOTTY", it's not a TTY. If you check that, you won't need `os.isatty`.

As with cat too, it's probably worth printing the error's `strerror` if not silent.

xonsh/xoreutils/tee.py
```diff
@@ -0,0 +1,57 @@
...
+    mode = 'w'
+    if '-a' in args:
+        args.remove('-a')
+        mode = 'a'
+    if '--append' in args:
+        args.remove('--append')
+        mode = 'a'
+    if '--help' in args:
+        print(HELP_STR, file=stdout)
+        return 0
```
ScoreUnder: The above argument parsing is buggy and does not support `--`.

xonsh/xoreutils/tee.py
```diff
@@ -0,0 +1,57 @@
...
+    errors = False
+    files = []
+    for i in args:
+        if i == '-':
+            files.append(stdout)
+        else:
+            try:
+                files.append(open(i, mode))
+            except:
+                print('tee: failed to open {}'.format(i), file=stderr)
```
ScoreUnder: If it's an `IOError`, tell them the `strerror`

xonsh/xoreutils/tee.py
```diff
@@ -0,0 +1,57 @@
...
+                errors = True
+    files.append(stdout)
+
+    while True:
+        r = stdin.read(1024)
+        if r == '':
+            break
+        for i in files:
+            i.write(r)
```
ScoreUnder: Needs to handle the case where a pipe is closed or one of the files is on a full disk

PR #743, opened 2016-03-17
fix type order error as seen in ab781a9

scopatz:
This should fix the int and None ordering issue as also fixed in ab781a96fdd95501f0fd497d322dc470601d8115


adqm:
Yep; almost put in exactly this same change this afternoon.  Merging now.

scopatz:
Great minds and such :)

PR #744, opened 2016-03-17
Add numpy slicing to the parser

gforsyth:
This adds support for 2d and 3d slicing to the xonsh parser.  This is my first outing with `yacc` and there's probably a cleaner way to do it, but this works.

### DIFF
--- a/xonsh/parsers/base.py
+++ b/xonsh/parsers/base.py
@@ -1799,8 +1799,8 @@ class BaseParser(object):
         """subscriptlist : subscript comma_subscript_list_opt comma_opt"""
         p1, p2 = p[1], p[2]
         if isinstance(p1, ast.Slice):
-            if p2 is not None:
-                p1 = ast.ExtSlice(dims=[p1, *p2])
+            if p2 is not None and True in [isinstance(x, ast.Slice) for x in p2]:
+                p1 = ast.ExtSlice(dims=[p1]+p2)
         elif p2 is not None:
             p1.value = ast.Tuple(elts=[p1.value] + [x.value for x in p2],
                                  ctx=ast.Load(), lineno=p1.lineno,
--- a/xonsh/parsers/base.py
+++ b/xonsh/parsers/base.py
@@ -1798,7 +1798,10 @@ class BaseParser(object):
     def p_subscriptlist(self, p):
         """subscriptlist : subscript comma_subscript_list_opt comma_opt"""
         p1, p2 = p[1], p[2]
-        if p2 is not None:
+        if isinstance(p1, ast.Slice):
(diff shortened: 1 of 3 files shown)


scopatz:
Thanks a million for putting this in @gforsyth!  There is a cleaner refactor to do this that also enable N-dim slicing.  I can explain in-line or in the discussion here?

gforsyth:
Discussion here works.  I was sure there was, but I was running into errors with `yacc`

scopatz:
So basically, we want to rely on the recursion of the parser and grammar to generate lists of values, rather than unrolling the list explicitly.  This involves setting up an append rule and then making a list version of this append rule.

I would remove the `p_subscript_2slice()` and `p_subscript_3slice()` methods.  Most of what you need is supplied already by [`p_subscriptlist()`](https://github.com/scopatz/xonsh/blob/master/xonsh/parsers/base.py#L1798).

What is missing from this is the case where you need to create an ExtSlice node rather than a Tuple.  However, you should be able to loop over the values in p1 and p2 here to determine in what you want is an ExtSlice.  The expression `[p1.value] + [x.value for x in p2]` or similar is what becomes the dims arg.

Let me know if this makes sense or if you run into errors.

gforsyth:
Hey @scopatz, thanks for the help!  I think this is ready for another look.

gforsyth:
whoops, forgot to test on py 3.4.  I'll fix that up.

xonsh/parsers/base.py
```diff
@@ -1798,7 +1798,10 @@ def p_trailer_quest(self, p):
     def p_subscriptlist(self, p):
         """subscriptlist : subscript comma_subscript_list_opt comma_opt"""
         p1, p2 = p[1], p[2]
-        if p2 is not None:
+        if isinstance(p1, ast.Slice):
+            if p2 is not None:
+                p1 = ast.ExtSlice(dims=[p1, *p2])
```
scopatz: p2 may still be none in this case, and this not able to be unpacked.

scopatz: I think ExtSlice should only be used if p2 is not None and any of the elements of p2 or p1 is a Slice, otherwise it should just be a Tuple.

scopatz:
> Hey @scopatz, thanks for the help! I think this is ready for another look.

No problem! There definitely is a learning curve to this language design stuff :).  The more people we have who are comfortable modifying the parser, the better!

scopatz:
Is this ready for another look?

gforsyth:
Hey @scopatz -- should be ready for another look.

scopatz:
Thanks a ton @gforsyth!

PR #747, opened 2016-03-18
some logic fixes to ndslicing

scopatz:
Also enables slicing like `[True,0:42]` where booleans and masks are mixed in with slices.  Probably for @gforsyth to take a look at.


gforsyth:
Looks good!  Merging in.

Issue #740, opened 2016-03-16
Add time stamps to !() and ![]


scopatz:
Probably should be a 2-tuple of epoch floats to match the history object.

Issue #738, opened 2016-03-16
Remove or Implement correct behaviour for !() and ![] in subproc mode


scopatz:
I am currently not entirely clear what this is supposed to do, but right now it throws the following errors:

```console
scopatz@artemis ~ $ echo !(echo hello)
Traceback (most recent call last):
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/base_shell.py", line 150, in default
    self.execer.exec(code, mode='single', glbs=self.ctx)  # no locals
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/execer.py", line 126, in exec
    return exec(code, glbs, locs)
  File "<xonsh-code>", line 1, in <module>
AttributeError: 'CompletedCommand' object has no attribute 'splitlines'
scopatz@artemis ~ $ echo ![echo hello]
hello
Traceback (most recent call last):
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/base_shell.py", line 150, in default
    self.execer.exec(code, mode='single', glbs=self.ctx)  # no locals
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/execer.py", line 126, in exec
    return exec(code, glbs, locs)
  File "<xonsh-code>", line 1, in <module>
AttributeError: 'HiddenCompletedCommand' object has no attribute 'splitlines'
```

Maybe this is meant to act on stdout?  But really, then the above could best be handled by `echo @(!(echo hello).stdout)`.  Thoughts?

adqm:
Woops.  Yeah, I guess to me it feels like this shouldn't be valid at all.  I like getting rid of this, and using `@(!(echo hello).stdout)` instead.

Or, since that object is equivalent to its return code for a number of other things, it could be the returncode...but I don't imagine that's useful at all.  So I'm in favor of just pulling it.  I may be able to put that in this afternoon.

scopatz:
I am happy if they are removed :)

scopatz:
Hey @adqm - would you mind putting this in or if I put this in?  I'd like to do a v0.2.7 release soon but I'd also like it if this was cleaned up first.  Just let me know how you want to proceed!

adqm:
Yep.  On it right now (and I'll try to clean up the other couple of issues around this in the same PR).

scopatz:
Great! Thanks!

Issue #737, opened 2016-03-16
Add parser tests for !() and ![]


scopatz:
This is needed for QC.

PR #748, opened 2016-03-19
Fixes for !() and ![]

adqm:
This changeset includes a couple of fixes:

* It addresses #738 by removing erroneous cases from the parser.
* It addresses #737 by adding some test cases related to the new syntax (mostly copied and modified from the $() and $[] tests).
* It addresses #740 by adding timestamps to the `CompletedCommand` object.


scopatz:
Thanks a million @adqm!

PR #749, opened 2016-03-20
Fix colors for cmd exe

melund:
This improves the usefullnes of xonsh in cmd.exe.

It eliminates some of the dark blue and dark red colors when using cmd.exe and the default style. 

I reused one of the old environment variables from the previous colors system ( $PROMPT_TOOLKIT_STYLES)  So it now does the same as it did before. We should consider changing the name though, to something more saying. Any good ideas?

### DIFF
--- a/xonsh/pyghooks.py
+++ b/xonsh/pyghooks.py
@@ -283,7 +283,7 @@ class XonshStyle(Style):
             The style name to initialize with.
         """
         self.trap = {} # for traping custom colors set by user
-        self._mod = {} # for internal modification of the style
+        self._smap = {}
         self._style_name = ''
         self.style_name = style_name
         super().__init__()
@@ -296,7 +296,6 @@ class XonshStyle(Style):
     def style_name(self, value):
         if self._style_name == value:
             return
-        self._mod.clear()
         if value in STYLES:
             cmap = STYLES[value]
         else:
@@ -304,11 +303,11 @@ class XonshStyle(Style):
                  RuntimeWarning)
             cmap = DEFAULT_STYLE
         try:
-            smap = get_style_by_name(value)().styles
+            self._smap = get_style_by_name(value)().styles.copy()
         except (ImportError, pygments.util.ClassNotFound):
-            smap = XONSH_BASE_STYLE
-        compound = CompoundColorMap(ChainMap(self.trap, self._mod, cmap, PTK_STYLE, smap))
-        self.styles = ChainMap(self.trap, self._mod, cmap, PTK_STYLE, smap, compound)
+            self._smap = XONSH_BASE_STYLE.copy()
+        compound = CompoundColorMap(ChainMap(self.trap, cmap, PTK_STYLE, self._smap))
+        self.styles = ChainMap(self.trap, cmap, PTK_STYLE, self._smap, compound)
         self._style_name = value
         if ON_WINDOWS:
             self.enhance_colors_for_cmd_exe()
@@ -326,15 +325,12 @@ class XonshStyle(Style):
         env = builtins.__xonsh_env__
         # Ensure we are not using ConEmu
         if 'CONEMUANSI' not in env:
-            self._mod.update({Token.AutoSuggestion:'#444444'})
+            # Auto suggest needs to be a darker shade to be distinguishable
+            # from the default color
+            self.styles[Token.AutoSuggestion] = '#444444'
             if env.get('INTENSIFY_COLORS_ON_WIN', False):
-                # Only modify the pygments part of the style
-                try:
-                    smap = get_style_by_name(self.style_name)().styles
-                except (ImportError, pygments.util.ClassNotFound):
-                    smap = XONSH_BASE_STYLE
-                s = intensify_colors_for_cmd_exe(smap)
-                self._mod.update(s)
+                s = intensify_colors_for_cmd_exe(self._smap)
+                self._smap.update(s)
 
 def xonsh_style_proxy(styler):
     """Factory for a proxy class to a xonsh style."""
--- a/xonsh/pyghooks.py
+++ b/xonsh/pyghooks.py
@@ -282,8 +282,6 @@ class XonshStyle(Style):
         style_name : str, optional
             The style name to initialize with.
         """
-        if ON_WINDOWS and style_name == 'default':
-            enhance_colors_for_cmd_exe()
         self.trap = builtins.__xonsh_env__.get('PROMPT_TOOLKIT_STYLES') or {}
         self._style_name = ''
         self.style_name = style_name
@@ -310,10 +308,37 @@ class XonshStyle(Style):
         compound = CompoundColorMap(ChainMap(self.trap, cmap, PTK_STYLE, smap))
         self.styles = ChainMap(self.trap, cmap, PTK_STYLE, smap, compound)
         self._style_name = value
+        if ON_WINDOWS:
+            self.enhance_colors_for_cmd_exe()
+            
 
     @style_name.deleter
     def style_name(self):
         self._style_name = ''
+
+    def enhance_colors_for_cmd_exe(self):
+        """ Enhance colors when using cmd.exe on windows.
+            When using the default style all blue and dark red colors
+            are changed to CYAN and intence red.
+        """
+        env = builtins.__xonsh_env__
+        # Ensure we are not using ConEmu
+        if 'CONEMUANSI' not in env:
+            self.styles.update({Token.AutoSuggestion:'#444444'})
+            if self._style_name == 'default' and env.get('PROMPT_TOOLKIT_STYLES') is None: 
+                s = {Token.Name.Variable:'#44ffff',
+                    Token.Generic.Prompt:'#44ffff',
+                    Token.Name.Namespace:'#00aaaa',
+                    Token.Name.Function:'#00aaaa',
+                    Token.Name.Class:'#00aaaa',
+                    Token.Generic.Heading:'#00aaaa',
+                    Token.Literal.String.Symbol:'#00aaaa',
+                    Token.Literal.String:'#ff4444',
+                    Token.Name.Constant:'#ff4444',
+                    Token.Keyword.Type:'#ff4444',
+                    Token.Generic.Error:'#ff4444'}   
+            env['PROMPT_TOOLKIT_STYLES'] = s
+            self.styles.update(s)
 
 
 def xonsh_style_proxy(styler):
@@ -329,27 +354,6 @@ def xonsh_style_proxy(styler):
 
     return XonshStyleProxy
 
-
-def enhance_colors_for_cmd_exe():
-    # Test if we are on WINDOWS with CMD.exe 
-    # and PROMPT_TOOLKIT_STYLES is unset
-    env = builtins.__xonsh_env__
-    if 'CONEMUANSI' not in env and env.get('PROMPT_TOOLKIT_STYLES') is None:
-        # Only change the style if $PROMPT_TOOLKIT_STYLES is not touched and
-        # and we are not using ConEmu. 
-        smap = {Token.Name.Variable:'#44ffff',
-                Token.Generic.Prompt:'#44ffff',
-                Token.Name.Namespace:'#00aaaa',
-                Token.Name.Function:'#00aaaa',
-                Token.Name.Class:'#00aaaa',
-                Token.Generic.Heading:'#00aaaa',
-                Token.Literal.String.Symbol:'#00aaaa',
-                Token.Literal.String:'#ff4444',
-                Token.Name.Constant:'#ff4444',
-                Token.Keyword.Type:'#ff4444',
-                Token.Generic.Error:'#ff4444',
-                Token.AutoSuggestion:'#ansidarkgray'}         
-        env['PROMPT_TOOLKIT_STYLES'] = smap
     
     
 PTK_STYLE = {
--- a/xonsh/pyghooks.py
+++ b/xonsh/pyghooks.py
@@ -334,7 +334,7 @@ def enhance_colors_for_cmd_exe():
     # Test if we are on WINDOWS with CMD.exe 
     # and PROMPT_TOOLKIT_STYLES is unset
     env = builtins.__xonsh_env__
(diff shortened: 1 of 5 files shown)


melund:
@BYK Can you test this on Windows? Any improvements to the style is welcome...

BYK:
@melund - with `cmd.exe` or `conemu`?

melund:
It is only supposed to do something for cmd.exe and the default style. I.e. what new users would see first. Try it with and without this fix to see the difference. Also try to run xonfig colors

BYK:
Got the following error:

>    assert False, "wrong color format %r" % text
>AssertionError: wrong color format '#ansidarkgray'

BYK:
@melund I actually _just_ updated to latest prompt-toolkit and pygments before trying this out. It may have something to do with pygments?

melund:
What does your xonfig report?

melund:
Also can I see the full traceback?

BYK:
Xonfig (from latest master):
```
$ xonfig
+------------------+-------+
| xonsh            | 0.2.7 |
| Python           | 3.5.1 |
| PLY              | 3.6   |
| have readline    | True  |
| prompt toolkit   | 0.60  |
| pygments         | 2.1.3 |
| on posix         | False |
| on linux         | False |
| on arch          | False |
| on windows       | True  |
| on mac           | False |
| are root user    | False |
| default encoding | utf-8 |
+------------------+-------+
```

Full traceback:
```
C:\Users\BYK>python -u -m xonsh.main
  File "C:\Program Files\Python\lib\runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Program Files\Python\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Program Files\Python\lib\site-packages\xonsh\main.py", line 238, in <module>
    main()
  File "C:\Program Files\Python\lib\site-packages\xonsh\main.py", line 214, in main
    shell.cmdloop()
  File "C:\Program Files\Python\lib\site-packages\xonsh\ptk\shell.py", line 112, in cmdloop
  File "C:\Program Files\Python\lib\site-packages\xonsh\ptk\shell.py", line 72, in singleline
    style=PygmentsStyle(xonsh_style_proxy(self.styler)),
  File "C:\Program Files\Python\lib\site-packages\xonsh\pyghooks.py", line 321, in xonsh_style_proxy
    class XonshStyleProxy(Style):
  File "C:\Program Files\Python\lib\site-packages\pygments\style.py", line 76, in __new__
    ndef[0] = colorformat(styledef)
  File "C:\Program Files\Python\lib\site-packages\pygments\style.py", line 33, in colorformat
    assert False, "wrong color format %r" % text
AssertionError: wrong color format '#ansidarkgray'
```

scopatz:
@Carreau, is #ansidarkgray one of the supported colors?

Carreau:
Should be:

```
In [3]: from pygments.console import light_colors

In [4]: light_colors
Out[4]: ['darkgray', 'red', 'green', 'yellow', 'blue', 'fuchsia', 'turquoise', 'white']
```

But only on pygments master.

xonsh/pyghooks.py
```diff
@@ -327,6 +330,28 @@ def __new__(cls, *args, **kwargs):
...
+                Token.Name.Namespace:'#00aaaa',
+                Token.Name.Function:'#00aaaa',
+                Token.Name.Class:'#00aaaa',
+                Token.Generic.Heading:'#00aaaa',
+                Token.Literal.String.Symbol:'#00aaaa',
+                Token.Literal.String:'#ff4444',
+                Token.Name.Constant:'#ff4444',
+                Token.Keyword.Type:'#ff4444',
+                Token.Generic.Error:'#ff4444',
+                Token.AutoSuggestion:'#ansidarkgray'}
```
Carreau: You can't do that. You have to use a conditional to check for `hasattr(pygments.style, 'ansicolors')` as the version of pygments that support `#ansidarkgray` has not been released yet.

scopatz: Nice catch @Carreau!

melund:
Yes. Good catch. I thought I was already in the released version. But it turns out that I am still running pygments from master. 

I didn't want to use the #ansi color specification, but I couldn't get the dark gray color otherwise. I can't find a color value which prompt_toolkit matches to dark gray. I think the following line

`(0x44, 0x44, 0x44, FG.BLACK | FG.INTENSITY, BG.BLACK | BG.INTENSITY),`

is missing from the color table here: 
https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/terminal/win32_output.py#L459

No matter what color I try, I either get black or normal gray.... 

@Carreau is there any other way to get dark gray?

melund:
Ok. Adding the line actually did the trick. Now I can get dark gray colors. I have opened a pull request to prompt_toolkit https://github.com/jonathanslenders/python-prompt-toolkit/pull/268

melund:
@BYK ... I have removed the #ansidarkgray keyword. So please try again. 

However, it won't give dark gray colors due to the bug in prompt_toolkit. My little fix to prompt_toolkit has been merged (jonathanslenders/python-prompt-toolkit#268) so the autocompletion coloring should look like this with the next version of prompt_toolkit. 
image

The rest of the stuff should work as expected.

xonsh/pyghooks.py
```diff
@@ -281,7 +282,7 @@ def __init__(self, style_name='default'):
         style_name : str, optional
             The style name to initialize with.
         """
-        self.trap = {}  # for custom colors
+        self.trap = builtins.__xonsh_env__.get('PROMPT_TOOLKIT_STYLES') or {}
```
scopatz: I am not such a huge fan of this.  I think that the trap serves basically the same purpose as setting the user colors. Also, this has less to do really with prompt-toolkit and more to do with pygments.  I would be in favor of totally deprecating the `$PROMPT_TOOLKIT_STYLES` variable and replacing it with a `$XONSH_USER_COLORS` or `$XONSH_COLOR_MASK` or similar.  This could then be used across both pygments and ansicolors as needed. 

The rest of the code looks ok.

Does this sound reasonable?

xonsh/pyghooks.py
```diff
@@ -307,11 +308,38 @@ def style_name(self, value):
...
+                    Token.Name.Namespace:'#00aaaa',
+                    Token.Name.Function:'#00aaaa',
+                    Token.Name.Class:'#00aaaa',
+                    Token.Generic.Heading:'#00aaaa',
+                    Token.Literal.String.Symbol:'#00aaaa',
+                    Token.Literal.String:'#ff4444',
+                    Token.Name.Constant:'#ff4444',
+                    Token.Keyword.Type:'#ff4444',
+                    Token.Generic.Error:'#ff4444'}   
+            env['PROMPT_TOOLKIT_STYLES'] = s
```
scopatz: Should probably modify this variable in-place, rather than overwriting it.

xonsh/xonfig.py
```diff
@@ -324,6 +324,8 @@ def _tok_colors(cmap, cols):
 
 def _colors(ns):
     cols, _ = shutil.get_terminal_size()
+    if tools.ON_WINDOWS:
+        cols -= 1
```
scopatz: Is this really a bug with `shutil.get_terminal_size()`?  Just curious...

xonsh/pyghooks.py
melund: You are right. The name is stupid. I just recreated the functionality from the old color system.

I think we could even completely remove the environment variable. The shell.styler hold state of the styles anyway. So the same thing can be accomplished without the environment variable. I will make some changes. Hope you like the more :)

xonsh/xonfig.py
melund: Yes. I don't know why that is...

melund:
@scopatz .... I removed the PROMPT_TOOLKIT_STYLES variable completely. My only concern is that we need to find a way to disable this disable this functionality. The obvious is an other env-var, but I don't like that very much.

It could also be nice if this could be more generic, to substitute unreadable colors in any style.

scopatz:
I am happy to merge this in now if you'd like @melund.  We can always add another envvar later.  It is probably better that way.

melund:
I will just add an entry to the change log. Then it should be Ok

melund:
@BYK @scopatz ... I reworked it completely. Now it  does what I originally wanted. It changes the colors to something which is more readable on the windows terminal, and it works with any of the pygments styles. The whole thing is controlled with a environment variable `$INTENSIFY_COLORS_ON_WIN`

I had to make some small modifications to the XonshStyle. I added a extra layer in ChainMap which is cleared together with the style. This ensures that any style specific changes are cleared when the style changes or the `$INTENSIFY_COLORS_ON_WIN` envvar is toggled. 

@BYK try to toggle the env-var and see the effect with `xonfig colors` command.

BYK:
@melund - gonna look at this soon. I was away from my Windows PC for some time, sorry.

melund:
Just rebased on master.

xonsh/pyghooks.py
```diff
@@ -281,7 +282,8 @@ def __init__(self, style_name='default'):
         style_name : str, optional
             The style name to initialize with.
         """
-        self.trap = {}  # for custom colors
+        self.trap = {} # for traping custom colors set by user
+        self._mod = {} # for internal modification of the style
```
scopatz: Is there a problem with just using trap? This was exactly the use case for it.  I don't think there is a need to two traps...

melund: No. But I need to reset it every time the style changes or the env-var is toggled. If the user (e.g. in xonshrc file) or others have added something to the style that would be cleared. That would be kind of rude. I would mind removing it again if it is OK to clear the trap dict.

xonsh/tools.py
```diff
@@ -55,6 +55,12 @@ def pygments_version():
 except ImportError:
     win_unicode_console = None
 
+try:
+    import prompt_toolkit
+except ImportError:
+    prompt_toolkit = None
```
scopatz: This import should just be moved to the function where it is needed.

xonsh/pyghooks.py
scopatz: Well, doesn't this only need to happen if the style is `'default'`?

melund: No. I need to clear whatever modification I made when the style changes. Since those modifications depend on the style. But I could maybe just modify the pygments part of the chainmap directly if the class kept a reference to it. Then I could remove the double trap thing...I will try and ping you again...

melund:
@scopatz . I have updated the code. XonshStyle now keeps a reference to the pygments part of the ChainMap (`smap`). I then just modify that part directly. This eliminates the need for an extra layer in the Chainmap since `smap` is always recreated when the style changes.

melund:
Here the effect it has on the default style:

image

BYK:
With the new variable set to `True` I noticed the text was quite a lot easier to distinguish and read since it had higher contrast. :+1:

melund:
@scopatz and @BYK ... I think it is ready for merge now. 

I updated it again. Now it also expands the range of the gray colors. They all mapped to the same default gray color index. So cmd.exe didn't have the same range of gray shades as in ConEmu. This fixes it. 

Here is the default style without the fix: 
image

And here with: 

image

scopatz:
Hi @melund, I guess I am fine to merge this, but I still don't understand why

1. this doesn't only apply to the default map?
2. Why don't you just change the actual colors in the default color map for the case of cmd.exe? 

This seems like the simplest solution to me...  For example, I'd be interested to see how the monokai style now behaves differently in these two cases.

scopatz:
Or maybe this doesn't really effect other styles because none of the other non-default styles use color names like red, blue, etc.

scopatz:
Nevermind, I think I see what is going on here. I'll go ahead and merge this once appveyor finishes.

melund:
Ok. I hope you got what I was trying to accomplish. Else just keep asking. I don't think I could do easily in any other ways. It only affects the styles from pygments, and it seemed like a big task to recreate them manually. 

It doesn't affect the user colors. So if a user asks for `GREEN` under the default style the still get #00AA00 (https://github.com/scopatz/xonsh/blob/master/xonsh/pyghooks.py#L527 ) and if they shift to monokai it still turns into #2ee22e (https://github.com/scopatz/xonsh/blob/master/xonsh/pyghooks.py#L549) .

scopatz:
Yeah, sorry I was being a little dense here.  It was hard for me to see what the outcome of the code would be without being to play around with it.  I finally sat down and walked through the flow of it and it made more sense after than :)

melund:
>I finally sat down and walked through the flow of it and it made more sense after than :)

Ha ha.. I often tried to do that with yours and @adqm's PRs.. with limited success ;)

melund:
Rebased and fixed changelog conflict.

scopatz:
@BYK, do the tests for this run on your machine? Until appveyor comes back up, you might be our new windows CI :)

BYK:
Hey, sorry for the delay. I'll do my best but I don't have access to my Windows machine sometimes (I'm mostly on Windows though). Here's my test run output:

```
BYK@~DocumentsProjectsxonsh fix_colors_for_cmd_exe
$ nosetests -q
WARNING: Token 'AT' multiply defined
Generating LALR tables
WARNING: 92 shift/reduce conflicts
WARNING: 9 reduce/reduce conflicts
WARNING: reduce/reduce conflict in state 79 resolved using rule (test_or_star_expr -> test)
WARNING: rejected rule (testlist -> test) in state 79
WARNING: reduce/reduce conflict in state 228 resolved using rule (string_tok -> STRING)
WARNING: rejected rule (subproc_arg_part -> STRING) in state 228
WARNING: reduce/reduce conflict in state 322 resolved using rule (test_or_star_expr -> test)
WARNING: rejected rule (testlist -> test) in state 322
C:\Users\BYK\Documents\Projects\xonsh\xonsh\environ.py:52: RuntimeWarning: Failed to set locale 'LC_MESSAGES' to 'C'
  warn('Failed to set locale {0!r} to {1!r}'.format(key, val), RuntimeWarning)
and the bunnymen
and the bunnymen
Ran 132550 tests in 18.394s

OK (SKIP=8)
```

So seems like it is okay :)

scopatz:
Thanks @melund! Sorry about all the delays.  Thanks for the CI @BYK :)

Issue #724, opened 2016-03-07
Headless Shell


scopatz:
I think it is a good idea to have a shell class that is strictly for running code and does not accept input from the user. This could avoid startup time from loading readline, ptk, the completer, history, etc.

scopatz:
Naturally, this would then be used as the implementation of `xonsh -c`

adqm:
I'm happy to give this a go.  I think all the pieces are probably in place, so should just be a matter of rearranging things, I expect...

scopatz:
Sure thing! Have at :)

adqm:
Interesting thing is that (at least on my machine), removing the import for readline doesn't speed things up too much.  There are some other things going on here that are affecting startup times.  As of right now, running `$ xonsh -c echo "hi"`:

* headless takes about .9 seconds
* readline takes about .9 seconds
* prompt_toolkit takes about 1.4 seconds

I'm looking into profiling now, to try to see where the biggest timesinks are, and then we can see whether there isn't a way to avoid those when we aren't going to be running interactively.

scopatz:
Interesting... Thanks for investigating!

PR #750, opened 2016-03-20
Headless Shell and Other Small Speedups

adqm:
This PR implements a headless shell as discussed in #724.  This wasn't as big of a timesaver as I expected, so I implemented a couple of other little things, too:

* only execute the rc files, set up a completer, and import things from foreign shells when we need to (when in login mode)
* cache a compiled version of rc files, to avoid re-parsing them every time

xonsh still doesn't really fly on startup, but this may make some kind of difference.  I saw about a 2x speedup when running a single command, all things told:

On master:
```
hartz@lappy-514:~$ time xonsh -c "echo 'hi'"
hi

real	0m0.934s
user	0m0.808s
sys	0m0.160s
```

On this branch:
```
hartz@lappy-514:~$ time xonsh -c "echo 'hi'"
hi

real	0m0.453s
user	0m0.405s
sys	0m0.060s
```

And using the compiled rc file will help a little bit, even when starting an interactive shell:
```
hartz@lappy-514:~$ xonsh
not using cached marshaled RC
time spent on rc: 0.5360443592071533
hartz@lappy-514 ~ $
```
```
hartz@lappy-514:~$ xonsh
found cache; using cached marshaled RC
time spent on rc: 0.35840320587158203
hartz@lappy-514 ~ $ 
```


adqm:
I think it might also be a good idea to cache compiled versions of scripts as they are run, as parsing seems to take quite a lot of time.  Or, at least, to provide the option to do so (maybe another envvar?).  I'm happy to write that in here, or in a separate PR.

xonsh/environ.py
```diff
@@ -1135,9 +1137,36 @@ def load_static_config(ctx, config=None):
...
+def _splitpath(path, sofar=[]):
+    folder, path = os.path.split(path)
+    if path == "":
+        return sofar[::-1]
+    elif folder == "":
+        return (sofar + [path])[::-1]
+    else:
+        return _splitpath(folder, sofar + [path])
+
+_character_map = {chr(o): '_%s' % chr(o+32) for o in range(65, 91)}
```
scopatz: PEP8: I think this should be `UPPERCASE_WITH_UNDERSCORES`

xonsh/main.py
```diff
@@ -138,6 +138,10 @@ def _pprint_displayhook(value):
             pprint(value)  # black & white case
         builtins._ = value
 
+MODE_SINGLE_COMMAND = 0
+MODE_SCRIPT_FROM_FILE = 1
+MODE_SCRIPT_FROM_STDIN = 2
+MODE_INTERACTIVE = 3
```
scopatz: Maybe this should be an Enum?

scopatz:
Thanks a million for putting this in @adqm! This looks good to go from my end. It just needs to have a changelog entry....

adqm:
Alright, @scopatz; I think this is ready for another look.

xonsh/environ.py
```diff
@@ -1135,9 +1137,36 @@ def load_static_config(ctx, config=None):
...
+def _make_if_not_exists(dirname):
+    if not os.path.isdir(dirname):
+        os.makedirs(dirname)
+
+
 def xonshrc_context(rcfiles=None, execer=None):
     """Attempts to read in xonshrc file, and return the contents."""
     loaded = builtins.__xonsh_env__['LOADED_RC_FILES'] = []
+    homedir = builtins.__xonsh_env__['HOME']
+    cachedir = os.path.join(homedir, '.xonshrc_cache')
```
scopatz: I think that `$XONSH_DATA_DIR/xonshrc_cache` (see [here](http://xon.sh/envvars.html#xonsh-data-dir)) is probably a better location that in the home directory.

scopatz:
This looks great @adqm! Happy to merge after an update to the cache location piece...

adqm:
Alright; just pushed a change for that.

I think eventually, that caching code might want to move elsewhere (tools.py?), and we might want to use it when running/importing other scripts.  Shouldn't be hard to modify it for that purpose...well, I'm not sure about the imports, but I know it should be easy to use it for running scripts.

scopatz:
> I think eventually, that caching code might want to move elsewhere (tools.py?), and we might want to use it when running/importing other scripts. Shouldn't be hard to modify it for that purpose...well, I'm not sure about the imports, but I know it should be easy to use it for running scripts.

Yeah I was thinking that it could be generalized as well.  Tools seems fine.  Maybe it could even be moved into the execer but only used when a certain kwarg is given?  In any event, it isn't a huge concern until we actually want to do that....

scopatz:
Thanks a million @adqm!

PR #751, opened 2016-03-21
New: Add `sudo` functionality on Windows

BYK:
Not perfect since it doesn't (cannot) capture the `stdin` and `stdout` for the launched process but it works and is a good first step.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,7 @@ Current Developments
 
 * Added a new shell type ``'none'``, used to avoid importing ``readline`` or
   ``prompt_toolkit`` when running scripts or running a single command.
+* New: `sudo` functionality on Windows through an alias
 
 **Changed:**
 
--- a/docs/aliases.rst
+++ b/docs/aliases.rst
@@ -178,6 +178,14 @@ The following aliases on Windows are expanded to ``['cmd', '/c', alias]``:
 On Windows, ``which`` is aliased to ``['where']``.
 
 
+``sudo`` on Windows
+====================
+On Windows, if no executables named ``sudo`` are found, Xonsh adds a ``sudo`` alias
+that poly fills the "run as Admin" behavior with the help of ``ShellExecuteEx`` and
+``ctypes``. It doesn't support any actual ``sudo`` parameters and just takes the 
+command to run.
+
+
 ``ls``
 ====================
 The ``ls`` command is aliased to ``['ls', '--color=auto', '-v']`` normally.  On Mac OSX
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -687,10 +687,15 @@ def _is_executable_file(path):
 
 
 def yield_executables_windows(directory, name):
+    normalized_name = os.path.normcase(name)
     extensions = builtins.__xonsh_env__.get('PATHEXT')
     for a_file in os.listdir(directory):
-        base_name, ext = os.path.splitext(a_file)
-        if name == base_name and ext.upper() in extensions:
+        normalized_file_name = os.path.normcase(a_file)
+        base_name, ext = os.path.splitext(normalized_file_name)
+
+        if (
+            normalized_name == base_name or normalized_name == normalized_file_name
+        ) and ext.upper() in extensions:
             yield os.path.join(directory, a_file)
 
 
@@ -708,9 +713,15 @@ def locate_binary(name):
     if os.path.isfile(name) and name != os.path.basename(name):
         return name
 
+    directories = builtins.__xonsh_env__.get('PATH')
+
+    # Windows users expect t obe able to execute files in the same directory without `./`
+    if ON_WINDOWS:
+        directories = [_get_cwd()] + directories
+
     try:
         return next(chain.from_iterable(yield_executables(directory, name) for
-                    directory in builtins.__xonsh_env__.get('PATH') if os.path.isdir(directory)))
+                    directory in directories if os.path.isdir(directory)))
     except StopIteration:
         return None
 
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -1,13 +1,11 @@
 # -*- coding: utf-8 -*-
 """Aliases for the xonsh shell."""
-import os
-import shlex
+
 import builtins
-import subprocess
-from warnings import warn
+import os
 from argparse import ArgumentParser
 
-from xonsh.dirstack import cd, pushd, popd, dirs
+from xonsh.dirstack import cd, pushd, popd, dirs, _get_cwd
 from xonsh.jobs import jobs, fg, bg, kill_all_jobs
 from xonsh.proc import foreground
 from xonsh.timings import timeit_alias
@@ -252,6 +250,25 @@ if ON_WINDOWS:
         DEFAULT_ALIASES[alias] = ['cmd', '/c', alias]
 
     DEFAULT_ALIASES['which'] = ['where']
+
+    if not locate_binary('sudo'):
+        import xonsh.winutils as winutils
+
+        def sudo(args, sdin=None):
+            if len(args) < 1:
+                print('You need to provide an executable to run as Administrator.')
+                return
+
+            cmd = args[0]
+
+            if locate_binary(cmd):
+                return winutils.sudo(cmd, args[1:])
+            elif cmd.lower() in WINDOWS_CMD_ALIASES:
+                return winutils.sudo('cmd', ['/D', '/C', 'CD', _get_cwd(), '&&'] + args)
+            else:
+                print('Cannot find the path for executable "{0}".'.format(cmd))
+
+        DEFAULT_ALIASES['sudo'] = sudo
 
 elif ON_MAC:
     DEFAULT_ALIASES['ls'] = ['ls', '-G']
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -25,7 +25,6 @@ from xonsh.tools import (
 )
 from xonsh.inspectors import Inspector
 from xonsh.environ import Env, default_env, locate_binary
-from xonsh.aliases import DEFAULT_ALIASES
 from xonsh.jobs import add_job, wait_for_active_job
 from xonsh.proc import (ProcProxy, SimpleProcProxy, ForegroundProcProxy,
                         SimpleForegroundProcProxy, TeePTYProc,
@@ -772,6 +771,9 @@ def load_builtins(execer=None, config=None, login=False):
     builtins.evalx = None if execer is None else execer.eval
     builtins.execx = None if execer is None else execer.exec
     builtins.compilex = None if execer is None else execer.compile
+
+    # Need this inline/lazy import here since we use locate_binary that relies on __xonsh_env__ in default aliases
+    from xonsh.aliases import DEFAULT_ALIASES
     builtins.default_aliases = builtins.aliases = Aliases(DEFAULT_ALIASES)
     if login:
         builtins.aliases.update(load_foreign_aliases(issue_warning=False))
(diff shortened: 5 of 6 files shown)


scopatz:
This looks good to me @BYK! Can you add something to the changelog please, as well as also something to `docs/aliases.rst`

BYK:
@scopatz - Done!

BYK:
@scopatz - that said I added it under `v0.2.8` out of habit. Can change to v0.3.0 if you want :)

melund:
image

melund:
Also in general. When we use these aliased commands without input, we should provide a better error message: 
``` 
mel@ANYBODY-MEL ~/Documents/scrats
$ sudo
```

BYK:
@melund I think `mklink` is a `cmd` command (not an executable). If not this is weird, ShellExecuteEx should not need a full path. To alleviate this, we may have a "shell" mode that uses CMD.exe like subproc does. We can trigger this mode automatically if we cannot find the executable path ourselves, what do you think?

For the no arguments case, I agree. This was a very quick mash up since you wanted it on master sooner than later :)

melund:
You are right. ´mklink´ is a ´cmd´ command....

CHANGELOG.rst
```diff
@@ -16,6 +16,12 @@ Current Developments
 
 **Security:** None
 
+v0.2.8
+====================
```
scopatz: Actually, you should leave this under the `Current Developments` since it isn't part of a release yet.  The version number is filled in automatically here by the `release.xsh` script.  Would you mind fixing this up, please?

docs/aliases.rst
```diff
@@ -178,6 +178,14 @@ The following aliases on Windows are expanded to ``['cmd', '/c', alias]``:
 On Windows, ``which`` is aliased to ``['where']``.
 
 
+``sudo`` on Windows
+====================
+On Windows, if no executables named ``sudo`` are found, Xonsh adds a ``sudo`` alias
+that poly fills the "run as Admin" behavior with the help of ``ShellExecuteEx` and
```
scopatz: Looks like `ShellExecuteEx` needs another trailing backtick

CHANGELOG.rst
BYK: Oh wow, I missed that :D

docs/aliases.rst
BYK: Good catch.

BYK:
@melund - should I address the CMD proxy thing in this patch or shall we defer that to a follow up?

melund:
It is fine to do it in another PR. But calling cmd commands as admin is essential so I hope you can find a good fix

scopatz:
@melund - you should be the one who merges this whenever you think it is ready :)

melund:
I will merge it when it is done...

BYK:
@melund, @scopatz - updated! Sorry for the delay.

melund:
Now. I get an error launching applications: 

image

BYK:
Ah, because I do not pass the CWD to `locate_binary`. Sigh. Will look.

melund:
Yes. I just came to the same conclusion :)

BYK:
@melund - I've fixed it but that changed the behavior of `locate_binary` a bit. @scopatz can you look at those changes and give your approval? (or rejection? :))

xonsh/environ.py
```diff
@@ -690,7 +690,8 @@ def yield_executables_windows(directory, name):
     extensions = builtins.__xonsh_env__.get('PATHEXT')
     for a_file in os.listdir(directory):
         base_name, ext = os.path.splitext(a_file)
-        if name == base_name and ext.upper() in extensions:
+
+        if (name == base_name or a_file == name) and ext.upper() in extensions:
```
BYK: I think we should have this change regardless of this patch. This essentially fails to find executables if you provide their full name (with the extension).

xonsh/environ.py
```diff
@@ -710,7 +711,7 @@ def locate_binary(name):
 
     try:
         return next(chain.from_iterable(yield_executables(directory, name) for
-                    directory in builtins.__xonsh_env__.get('PATH') if os.path.isdir(directory)))
+                    directory in ([_get_cwd()] + builtins.__xonsh_env__.get('PATH')) if os.path.isdir(directory)))
```
scopatz: I don't think that this is right, because if users want to run executables from the CWD they should add an empty string or `'.'` to the PATH.

scopatz: Though looking at this, the empty string case will probably fail:

```console
scopatz@localhost ~/faculty/teach uic-ds2016 $ os.path.isdir('')
False
```

This is almost certainly a bug.

melund: This is a case where there is a difference between POSIX and Windows. Win users would probably expect to be able to run executables from the cwd.

gforsyth: @scopatz - I'm not sure if that's a bug -- at least on *NIX `os.path` will use `posixpath.py` which explicity defines the current directory as `.`

scopatz: Points well taken.  I happy if the CWD change is only applied to Win if that is what people expect.

BYK:
@scopatz @melund - updated to limit the change to Windows.

@scopatz - I'll deal with the empty path in `PATH` issue in a separate PR.

scopatz:
Looks good to me!

scopatz:
Also I don't think you need to do anything about the empty PATH issue, at least not for Linux :)

BYK:
>Also I don't think you need to do anything about the empty PATH issue, at least not for Linux :)

Ah, good to know!

scopatz:
And the way this is written you probably don't need to do anything for Windows either since the CWD will always be searched first.

BYK:
>And the way this is written you probably don't need to do anything for Windows either since the CWD will always be searched first.

Yup. This was intentional ;)

melund:
@BYK .. It is still case sensitive. I don't think it should be on windows. 

```
mel@ANYBODY-MEL C:/Program Files/7-Zip
$ ls
Lang       7-zip.dll  7z.exe    7zG.exe       7z.sfx     History.txt  readme.txt
7-zip.chm  7z.dll     7zFM.exe  descript.ion  7zCon.sfx  License.txt
mel@ANYBODY-MEL C:/Program Files/7-Zip
$ sudo 7zfm.exe
Cannot find the path for executable '7zfm.exe'.
```

BYK:
Oh god, I'll never finish this PR...

Will update soon.

BYK:
Updated.

@scopatz - heads up: this new update again changes the behavior of `locate_binary` and I realize I should exclude Linux from this since its file system is case-sensitive. What do you think?

BYK:
@scopatz - oh wait, I did this only for Windows anyways. Ignore :)

xonsh/environ.py
```diff
@@ -687,10 +687,14 @@ def _is_executable_file(path):
...
 def yield_executables_windows(directory, name):
+    casefold_name = name.casefold()
     extensions = builtins.__xonsh_env__.get('PATHEXT')
     for a_file in os.listdir(directory):
         base_name, ext = os.path.splitext(a_file)
-        if name == base_name and ext.upper() in extensions:
+
+        if (
+            casefold_name == base_name.casefold() or casefold_name == a_file.casefold()
+        ) and ext.upper() in extensions:
```
melund: Any specific reasons to use casefold() and not lower() here? Just wondering...

scopatz: Casefold is more generic, and handles situations like eszett to `'ss'` conversion.  I think it is probably the right thing to use here.

melund: Yes. I hadn't used it before, but the python docs told me what it did. I was wondering if you had any specific reasons to use it here. I just tested this on my machine: "ß.txt" and "ss.txt" are two distinct files. Is there something I have misunderstood?

scopatz: Ahh yeah, maybe it should just be lower then.  I am assuming that this is on Windows, @melund?

scopatz: Basically this test should do the same thing that the Windows file system does.  Anything else will cause problems.

melund: Yes. It was on windows. But I must admit that I don't have any experience with such files names. It seems like windows interpret them as different files. 
image

BYK: Yup, I just wanted to do the "right thing" here which clearly is not true for file names. Gonna change.

BYK:
Updated to use `os.path.normcase` which is the "right" way to do this.

melund:
@BYK ... I think it is ready for merge now  :) I am hope you are still ready to implement support for cmd .exe commands.  @scopatz did you figure out what up with Appveyor?

BYK:
@melund - yup, after this goes in :)

BYK:
>@melund - yup, after this goes in :) 

Scratch that. Amended to this commit. Enjoy!

melund:
Cool. The majority of built-in commands seem to work fine. However, the mklink command causes me some headache. 

```
mel@ANYBODY-MEL ~/Documents/scrats
$ ls
folder1  ss.txt  ß.txt
mel@ANYBODY-MEL ~/Documents/scrats
$ sudo mklink /d test folder1
mel@ANYBODY-MEL ~/Documents/scrats
$ ls
folder1  ss.txt  ß.txt
```
So it doesn't create the link. 

It does launch the UAC dialog, but the command doesn't seem to do anything. The same thing works fine without sudo if xonsh is launched in admin mode. I can't see any errors it reports since the external window only opens for a fraction of a second. Do you have any idea what is causing it? Can you make symlinks with the sudo command?

BYK:
@melund for some reason CMD.exe is not running in the current working directory, that's why. Trying to find a solution.

BYK:
@melund - Updated. Unfortunately I had to add a ``cd `_get_cwd()` &&`` in front of all commands since there's no way to run CMD elevated from a specified directory. The internet is full of people suffering from this.

melund:
@BYK .. Well, on windows is easily gets ugly. Anyways, it seems to work nicely now.

I think we can merge when appveyor is up again, or if you can confirm that all tests run correctly. I tried to run them my self, and got some errors. But I don't think it is related to this PR....

BYK:
@melund - tests pass for me:

```
WARNING: Token 'AT' multiply defined
Generating LALR tables
WARNING: 92 shift/reduce conflicts
WARNING: 9 reduce/reduce conflicts
WARNING: reduce/reduce conflict in state 43 resolved using rule (test_or_star_expr -> test)
WARNING: rejected rule (testlist -> test) in state 43
WARNING: reduce/reduce conflict in state 193 resolved using rule (string_tok -> STRING)
WARNING: rejected rule (subproc_arg_part -> STRING) in state 193
WARNING: reduce/reduce conflict in state 367 resolved using rule (test_or_star_expr -> test)
WARNING: rejected rule (testlist -> test) in state 367
C:\Users\BYK\Documents\Projects\xonsh\xonsh\environ.py:51: RuntimeWarning: Failed to set locale 'LC_MESSAGES' to 'C'
  warn('Failed to set locale {0!r} to {1!r}'.format(key, val), RuntimeWarning)
and the bunnymen
and the bunnymen
Ran 132550 tests in 26.068s

OK (SKIP=8)
```

melund:
@BYK .. Since you could run all the tests successfully, I will just go ahead and merge this in. Let us see if we can come up with more ideas to use the elevation for. 

I was imagining something that could be put into .xsh scripts, to trigger a relaunch in elevated mode,    or maybe throw an exception otherwise. Do you think something like this is plausible? 

```
try:
     xonsh.winutils.require_admin()
     # My script that must be run as admin
     # ....
except OSError:
    print('Script must be run with admin right')

```

BYK:
@melund - sounds like something we can look into. That said I'd prefer if we can make that generic so we can use it on POSIX too. Like `require_elevation` or something.

PR #757, opened 2016-03-27
Cache

adqm:
This moves the script cache from #750 into `tools.py`, and it updates code throughout `xonsh` to make use of the cache.

We can discuss whether these options (and defaults) make sense, but right now, I have:

* By default, `xonsh` caches the compiled version of all scripts that are run, and uses the cached version unless the source file has changed since the cached version was created (tested using `mtime`).  This behavior can be turned off by supplying `--no-script-cache` or by setting the `$XONSH_CACHE_SCRIPTS` environment variable to `False`.
* `xonsh` does not do this by default, but it can be made to cache _everything_ (including statements passed to the shell interactively) by supplying `--cache-everything` or setting `$XONSH_CACHE_EVERYTHING` to `True`.  This is probably overkill... and it's a little dangerous because of the context-sensitive nature of `xonsh`...but I thought it was neat, and it wasn't hard to implement.  I can remove that bit if it's not wanted.

I'm not sure how to hook this up so that it works with imports as well.  @scopatz, do you know whether that would be easy?

As always, happy for comments/suggestions.


scopatz:
Very Cool @adqm!

> I'm not sure how to hook this up so that it works with imports as well. @scopatz, do you know whether that would be easy?

I believe that imports (ie when you import a *.xsh file as a module) you don't need to do anything extra.  Python takes care of this for us automatically since we use the standard mechanism for implementing our import hooks (in `xonsh.imphooks`). This can be verified by looking in the various `__pycache__/` directories that are created.

xonsh/environ.py
```diff
@@ -202,6 +204,8 @@ def xonshconfig(env):
                               'xonsh', 'xonshrc'),
                 os.path.expanduser('~/.xonshrc')) if ON_WINDOWS
                else ('/etc/xonshrc', os.path.expanduser('~/.xonshrc'))),
+    'XONSH_CACHE_SCRIPTS': True,
+    'XONSH_CACHE_EVERYTHING': False,
```
scopatz: Can you please add corresponding VarDoc entries below this please?

xonsh/main.py
```diff
@@ -71,6 +72,16 @@ def path_argument(s):
...
+parser.add_argument('--no-script-cache',
+                    help="Do not cache scripts as they are run",
+                    dest='scriptcache',
+                    action='store_false',
+                    default=True)
+parser.add_argument('--cache-everything',
+                    help="Use a cache, even for interactive commands",
+                    dest='cacheall',
+                    action='store_true',
+                    default=False)
```
scopatz: Can you please add respective `arg_undoer` entries below, please?

scopatz:
Just a couple of completeness things to do here. Also, an CHANGELOG entry would be nice too :)  Otherwise this looks good to go!  Assuming appveyor ever works again...

xonsh/base_shell.py
```diff
@@ -148,7 +148,7 @@ def default(self, line):
         tee = Tee() if store_stdout else io.StringIO()
         try:
             ts0 = time.time()
-            self.execer.exec(code, mode='single', glbs=self.ctx)  # no locals
+            run_code_with_cache(src, self.execer, self.ctx, None, 'single')
```
melund: Does `self.push(line)` above not always compile the code? It looks to me that it will now always be compiled at least once. It kind of ruins the point of a cache. Sorry if there is something that I completely miss. I am not familiar with this part of the code..

adqm: D'oh; yes, are right!  Good catch.  I will rework this a bit.

adqm:
Alright, after @melund brought up a glaring flaw in the existing logic, I reworked this so that the `$XONSH_CACHE_EVERYTHING` option actually does something.

For something like my `.xonshrc` file, caching saved something like .3 seconds (.2 seconds with caching, .5 without).  For simple interactive commands, the gains are a little more modest:

```
$ ls -la xonsh
trying cache
checking cache took 0.0002777576446533203 seconds
compiling took 0.009450197219848633 seconds
updating cache took 0.00010585784912109375 seconds
whole command took 0.009852170944213867 seconds
<<<snip output>>>
```
```
$ ls -la xonsh
trying cache
using cache
whole command took 0.0003323554992675781 seconds
<<<snip output>>>
```

I also reorganized things a bit and included some documentation.

scopatz:
It depends what you mean by modest :).  The xonshrc is a factor of 2.5x while the ls example is a factor of 30x.

scopatz:
Hi @adqm - I am not seeing the code cache, even if I set the `$XONSH_CACHE_EVERYTHING = True` or if I startup via `xonsh --cache-everything`.  Is there something I am missing here? Is this an in-memory cache only? Should it be?

The xonsh scripts seem to be cached just fine.

scopatz:
Also this looks like it needs a merge from master now too.

scopatz:
ping!

adqm:
Okay, back in action for a brief while.  I just rebased this onto master.  Hoping I didn't make any mistakes, as it wasn't a trivial process.  We'll see what Travis says, and then this should be ready for another look, I think.

xonsh/shell.py
```diff
@@ -115,19 +117,18 @@ def __getattr__(self, attr):
...
-            if self.login or self.stype != 'none':
-                # load xontrib files listed in the config file
-                names = builtins.__xonsh_config__.get('xontribs', ())
-                for name in names:
-                    xontribs.update_context(name, ctx=context)
-                # load run contol files
-                env = builtins.__xonsh_env__
+            if self.stype == 'none' and not self.login:
+                self.ctx = {}
+            else:
```
scopatz: It looks like this got rid of all of the `xontrib` loading accidentally on merge

scopatz:
This looks great @adqm! Just the one xontribs removal issue.

scopatz:
Also, really looking forward to having this feature :)

adqm:
Okay, I just tried to reproduce the `xontribs` logic.  Not sure how to test, as I wasn't around for that development (I'll need to read up on it).  Can you give it a whirl?

scopatz:
Thanks a ton @adqm! This is awesome!  Everything seemed to check out with the xontrib logic.

PR #762, opened 2016-03-30
quick fix for empty xonshrc files

gforsyth:
If a user has an empty `xonshrc` file lying around, the `exec(ccode, env, None)` prevents `xonsh` from starting


scopatz:
Appveyor clearly hates us...  Thanks!

PR #764, opened 2016-03-31
Update ipynb alias to run jupyter notebook instead

melund:


gforsyth:
Good call -- especially with the 5 second punishment delay when you type `ipython notebook`

PR #766, opened 2016-03-31
remove (actually) unused imports

gforsyth:
I should never have clicked on the Landscape icon...

But there are a bunch of unused imports lying around (and also a bunch that pylint misidentifies as unused) and it's good to keep the house in order.


scopatz:
> I should never have clicked on the Landscape icon...

hahah

scopatz:
Thanks @gforsyth!

PR #767, opened 2016-03-31
remove a few missed unused imports

gforsyth:
Missed a couple of unused imports on the first go-round


scopatz:
Thanks @gforsyth!

PR #772, opened 2016-04-01
ignore Komodo editor project files

rsyring:
will make my life a tad bit easier when contributing, feel free to reject if you want, it can just stay in my fork if you don't want this upstream.


scopatz:
Thanks!

Issue #769, opened 2016-04-01
__xonsh_env__.swap() should support keyword args


rsyring:
Feature request:

```python
with __xonsh_env__.swap(VAR1='foo', VAR2='bar'):
    pass
```
Thanks.

PR #773, opened 2016-04-01
add kwarg support to Env.swap()

rsyring:
fixes gh-769

This code in this PR seems pretty sound to me.  However, I'm not sure the docs are right.  I feel like this feature will be used more in scripting and not interactive mode, but my example is in interactive mode.  Thoughts?

Also, this is the first code I've submitted.  If I've missed any formatting/style issues, I'll be happy to fix them.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -134,9 +134,31 @@ variable in Python.  The same is true for deleting them too.
     Become the Lord of the Files
     >>> del $GOAL
 
-Very nice. All environment variables live in the built-in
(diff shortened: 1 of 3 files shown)


scopatz:
Looks great! Thanks!

melund:
Nice. @rsyring ... With the new ${...} we can do: 
```
$ with ${...}.swap(USERNAME='test'):
.     print($USERNAME)
.
test
```

It doesn't look to bad.

scopatz:
Nice! I like it!

PR #775, opened 2016-04-01
Add option to update os.environ when __xonsh_env__ changes.

melund:
This adds a new option `$UPDATE_OS_ENVIRON`. If set, it will update the os.environ mapping whenever the __xonsh_env__ mapping changes.

This could be useful if you are running scripts that use subprocess calls, and you don't have the option of modifying the code to parse in the `__xonsh_env__`

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -85,6 +85,7 @@ DEFAULT_ENSURERS = {
     'RAISE_SUBPROC_ERROR': (is_bool, to_bool, bool_to_str),
     'RIGHT_PROMPT': (is_string, ensure_string, ensure_string),
     'TEEPTY_PIPE_DELAY': (is_float, float, str),
+    'UPDATE_OS_ENVIRON': (is_bool, to_bool, bool_to_str),
     'XONSHRC': (is_env_path, str_to_env_path, env_path_to_str),
     'XONSH_COLOR_STYLE': (is_string, ensure_string, ensure_string),
     'XONSH_ENCODING': (is_string, ensure_string, ensure_string),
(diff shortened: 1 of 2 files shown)


melund:
There are some issue with this PR, when removing keys from the environment. I have a update underway..

xonsh/environ.py
```diff
@@ -625,10 +639,14 @@ def __setitem__(self, key, val):
             val = ensurer.convert(val)
         self._d[key] = val
         self._detyped = None
+        if self.get('UPDATE_OS_ENVIRON'):
+            self.update_env()
```
scopatz: I think there is a more efficient way to do this.  Everytime a single key is set, the whole of `os.environ` is updated, however it should probably just replace that key.  I think the following would be better

```python
def __setitem__(self, key, val):
    ensurer = self.get_ensurer(key)
     if not ensurer.validate(val):
        val = ensurer.convert(val)
    self._d[key] = val
    self._detyped = None
    if self.get('UPDATE_OS_ENVIRON'):
        dval = ensurer.detype(val)
        os.environ[key] = dval
```

xonsh/environ.py
```diff
@@ -625,10 +639,14 @@ def __setitem__(self, key, val):
...
 
     def __delitem__(self, key):
         del self._d[key]
         self._detyped = None
+        if self.get('UPDATE_OS_ENVIRON'):
+            self.update_env()
```
scopatz: and this line could just be `del os.environ[key]`

melund:
@scopatz ... I updated it so it only changes a single key

melund:
@moigagoo Does this solve your problem in any way?

moigagoo:
@melund This is **exactly** the thing I need, thanks a ton!

scopatz:
This looks great @melund! Thanks a ton! Sorry about the delay in merging.

PR #780, opened 2016-04-01
Add line wrapping to PTK multiline using arrows

gforsyth:
In a multiline statement, the left- and right-arrow keys will now wrap
to the previous or next line, respectively.  If you are at the beginning
of the document, the left-arrow key does nothing.  Same for the
right-arrow at the end of the document.


melund:
This looks good to me. Should it have an entry in the change log? Else it is good to be merged.

scopatz:
Sweet thanks!

PR #781, opened 2016-04-01
fix ptk completer to allow completion earlier in line

gforsyth:
As reported by @asmeurer, the completer had incorrect index values for
completion when the cursor was located before a space, that is

`~/Do<C>emacsclient -nw`

would not offer file path completions (where the cursor is marked by
`<C>`). This was because of (my) naive searching for indices, now
slightly less naive.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,6 +31,7 @@ Current Developments
 **Fixed:**
 
 * Fixed bug with loading prompt-toolkit shell < v0.57.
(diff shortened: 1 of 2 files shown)


asmeurer:
Seems to fix the issue.

melund:
Should it also have a change log?

gforsyth:
Thanks for the reminder, @melund!  I'll get into the habit eventually.

PR #783, opened 2016-04-01
Fix regexpath

melund:
Return an empty list instead of the regex expression when regexpath doesn't find any matches

### DIFF
--- a/xonsh/parsers/base.py
+++ b/xonsh/parsers/base.py
@@ -126,10 +126,10 @@ def xonsh_superhelp(x, lineno=None, col=None):
     return xonsh_call('__xonsh_superhelp__', [x], lineno=lineno, col=col)
 
 
-def xonsh_regexpath(x, lineno=None, col=None):
+def xonsh_regexpath(x, pymode, lineno=None, col=None):
     """Creates the AST node for calling the __xonsh_regexpath__() function."""
-    return xonsh_call('__xonsh_regexpath__', [x], lineno=lineno, col=col)
-
+    return xonsh_call('__xonsh_regexpath__', [x, pymode], lineno=lineno,
+                      col=col)
 
 def load_ctx(x):
     """Recursively sets ctx to ast.Load()"""
@@ -1720,7 +1720,9 @@ class BaseParser(object):
         """atom : REGEXPATH"""
         p1 = ast.Str(s=p[1].strip('`'), lineno=self.lineno,
                      col_offset=self.col)
-        p[0] = xonsh_regexpath(p1, lineno=self.lineno, col=self.col)
+        pymode = True
+        p[0] = xonsh_regexpath(p1, pymode, lineno=self.lineno,
+                               col=self.col)
 
     def p_atom_dname(self, p):
         """atom : DOLLAR_NAME"""
@@ -2214,7 +2216,9 @@ class BaseParser(object):
         """subproc_atom : REGEXPATH"""
         p1 = ast.Str(s=p[1].strip('`'), lineno=self.lineno,
                      col_offset=self.col)
-        p0 = xonsh_regexpath(p1, lineno=self.lineno, col=self.col)
+        pymode = False
+        p0 = xonsh_regexpath(p1, pymode, lineno=self.lineno,
+                             col=self.col)
         p0._cliarg_action = 'extend'
         p[0] = p0
 
--- a/xonsh/parsers/base.py
+++ b/xonsh/parsers/base.py
@@ -126,10 +126,11 @@ def xonsh_superhelp(x, lineno=None, col=None):
     return xonsh_call('__xonsh_superhelp__', [x], lineno=lineno, col=col)
 
 
-def xonsh_regexpath(x, pymode, lineno=None, col=None):
+def xonsh_regexpath(x, pymode=False, lineno=None, col=None):
     """Creates the AST node for calling the __xonsh_regexpath__() function."""
-    return xonsh_call('__xonsh_regexpath__', [x, pymode], lineno=lineno,
+    return xonsh_call('__xonsh_regexpath__', args=[x, pymode], lineno=lineno,
                       col=col)
+
 
 def load_ctx(x):
     """Recursively sets ctx to ast.Load()"""
@@ -1720,8 +1721,7 @@ class BaseParser(object):
         """atom : REGEXPATH"""
         p1 = ast.Str(s=p[1].strip('`'), lineno=self.lineno,
                      col_offset=self.col)
-        pymode = True
-        p[0] = xonsh_regexpath(p1, pymode, lineno=self.lineno,
+        p[0] = xonsh_regexpath(p1, pymode=True, lineno=self.lineno,
                                col=self.col)
 
     def p_atom_dname(self, p):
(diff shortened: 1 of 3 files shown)


gforsyth:
Hey @melund, this looks good but has conflicts with master

melund:
@gforsyth  I have rebased and fixed the changelog conflict

adqm:
I don't actually agree with this change.  It introduces an inconsistency between regex globbing and normal globbing, and it makes error reporting harder.

There was a reason I modified regex globbing to work this way (in #721).

In a normal glob, both in xonsh and in bash, the glob itself is kept around if there is no match.  This is what allows, for example, passing globs through to scp; and it's also what allows error reporting of the form:
```
hartz@lappy-514:~$ cat aaaaaa*
cat: aaaaaa*: No such file or directory
```

If we returned an empty list instead, it is as though that argument hadn't been passed in at all.  In the case of the example above, it will cause `cat` just to hang waiting for `EOF`.  Making this an empty list has the potential, I think, to lead to a lot of extra confusion.

I think it is important to keep out regex globbing consistent with normal globbing.

melund:
Yes. I see there is an inherent problem between subprocess mode and python-mode here. 

In pymode the current behaviour doesn't make sense. It completely ruins this use case.
```
for file in `.*\.py`: 
    some_command(file)
```
It would require a lot of extra code to check the result. 

Since this a language construct it should be pythonic and behave after the principle of least surprise. Imagine `os.listdir()` returning the current working directory instead if the directory was empty. That would surely upset a lot of people :)

So maybe this should be restricted to either subproces or python mode. That would surely be a shame. 

It must be possible to come up with a solution that would make it work differently between the two modes. 

I am not familiar how that part of the code base works. So maybe my ideas are way off. It could return a special list object with extra info and methods. Something which could generate the representation used in subproc mode. It could also be useful in pymode where the regex could be retreived later.

melund:
If it should be restricted to only one mode, then it should be pymode. It is way more versatile that way. What is your take on this @scopatz

scopatz:
Hi all, sorry for taking a few days to respond to this - I had a much needed weekend off :). Thanks a ton for identifying that the 'correct' thing to do is different in python and subproc modes. 

I am personally in favor of keeping the backticks in both and changing it so that in Python mode it returns an empty list and maintaining the current behaviour for subproc mode.  

Luckily I think the implementation is relatively easy since the parser (necessarily) implements this in two different locations: [pymode](https://github.com/scopatz/xonsh/blob/master/xonsh/parsers/base.py#L1719) and [subproc mode](https://github.com/scopatz/xonsh/blob/master/xonsh/parsers/base.py#L2213).  Only the pymode case needs to change.

I think the correct way to do this would be to add a keyword argument to `__xonsh_regexpath__()`, which is actually just an alias to [`xonsh.builtins.regexpath()`](https://github.com/scopatz/xonsh/blob/master/xonsh/built_ins.py#L269), that flags whether an empty list should be returned or not.  The parser can be then easily made to pass in True or False depending on the logic.

Does this make sense?  Would anyone mind taking a stab at this? I think it is pretty easy to implement.

scopatz:
Happy to do it myself as well, if no one else wants to, but it is pretty low hanging fruit for modifying the parser if anyone wants a nice chance to get into that.

melund:
Thanks. I can easily fix this. I just didn't clue where in parser it was implemented.

scopatz:
Thanks a million @melund!

melund:
I will fix the tests later

melund:
@scopatz ... I still haven't wrapped my head around the internals of the parser. I thought my current change would do the trick but it gives my the error below. 

Can you help me figure out what is wrong?

```
$ `.*.py`
  File "c:\users\mel\documents\xonsh\xonsh\main.py", line 236, in main
    shell.cmdloop()
  File "c:\users\mel\documents\xonsh\xonsh\ptk\shell.py", line 111, in cmdloop
  File "c:\users\mel\documents\xonsh\xonsh\ptk\shell.py", line 80, in singleline
    display_completions_in_columns=multicolumn)
  File "c:\users\mel\documents\xonsh\xonsh\ptk\shortcuts.py", line 104, in prompt
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\interface.py", line 370, in run
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\eventloop\win32.py", line 78, in run
    callbacks.feed_key(k)
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\interface.py", line 899, in feed_key
    self._active_cli.input_processor.feed_key(key_press)
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\key_binding\input_processor.py", line 172, in feed_key
    self._process_coroutine.send(key_press)
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\key_binding\input_processor.py", line 144, in _process
    self._call_handler(matches[-1], key_sequence=buffer)
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\key_binding\input_processor.py", line 187, in _call_handler
    handler.call(event)
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\key_binding\registry.py", line 33, in call
    return self.handler(event)
  File "c:\users\mel\documents\xonsh\xonsh\ptk\key_bindings.py", line 132, in multiline_carriage_return
    carriage_return(b, event.cli)
  File "c:\users\mel\documents\xonsh\xonsh\ptk\key_bindings.py", line 55, in carriage_return
    elif not current_line_blank and not can_compile(b.document.text):
  File "c:\users\mel\documents\xonsh\xonsh\ptk\key_bindings.py", line 102, in can_compile
    locs=builtins.__xonsh_ctx__)
  File "c:\users\mel\documents\xonsh\xonsh\execer.py", line 95, in compile
    code = compile(tree, filename, mode)
TypeError: required field "lineno" missing from expr
```

xonsh/parsers/base.py
```diff
@@ -126,9 +126,10 @@ def xonsh_superhelp(x, lineno=None, col=None):
     return xonsh_call('__xonsh_superhelp__', [x], lineno=lineno, col=col)
 
 
-def xonsh_regexpath(x, lineno=None, col=None):
+def xonsh_regexpath(x, pymode=False, lineno=None, col=None):
     """Creates the AST node for calling the __xonsh_regexpath__() function."""
-    return xonsh_call('__xonsh_regexpath__', [x], lineno=lineno, col=col)
+    return xonsh_call('__xonsh_regexpath__', args=[x, pymode], lineno=lineno,
```
scopatz: The problem here is that pymode, needs to be an AST node, not a normal Python value. I think if you add the following line above this one, everything will workout:

```python
pymode = ast.NameConstant(value=pymode, lineno=lineno, col_offset=col)
```

scopatz:
This looks great @melund, just the one line is all that is needed.  Basically, literal values need to be wrapped as the appropriate node type.  You can see how the parser handles None, True, and False here: https://github.com/scopatz/xonsh/blob/master/xonsh/parsers/base.py#L1701

melund:
@scopatz It is ready now. 

@adqm. I hope you agree with this fix, where only the behaviour in python mode is affected

scopatz:
Thanks!

PR #785, opened 2016-04-04
Fix bug on completion on ${...}

melund:


scopatz:
I'm getting this error in the ptk shell (which seems unrelated) and nothing on the readline shell.

```console
scopatz@artemis ~ $ x = ${...}.Exception in thread Thread-39:
  File "/home/scopatz/miniconda3/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/home/scopatz/miniconda3/lib/python3.5/threading.py", line 862, in run
  File "/home/scopatz/miniconda3/lib/python3.5/site-packages/prompt_toolkit/interface.py", line 742, in run
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/ptk/completer.py", line 34, in get_completions
    self.ctx)
ValueError: not enough values to unpack (expected 2, got 0)
```

scopatz:

melund:
Ahh. I saw that as well, but when I tried later I couldn't produce it. So I thought this was related to something else entirely. It doesn't happen if `${...}` is first on the line.

scopatz:
Hmm weird.  What is the test case that you are using?

melund:
If I have it at the beginning of the line then I don't get the error:
```
$ ${...}.<tab> ` 
```
I do get the same error as you if I do `x = ${...}.<tab>`

scopatz:
Sure, but hitting TAB doesn't actually give me any options at that point either...  Does it work for you?

melund:
No. True. It doesn't complete on `${...}`. @gforsyth promised to think about how this could be done.

scopatz:
Thanks @melund!

PR #789, opened 2016-04-09
Even more and/or

scopatz:
This is a branch off of `andor` that has been merged with `master` and updated to enable the `and`, `or`, `&&`, and `||` operators on subprocess.  

Note that this does not allow these operators in subproc mode. Rather everything gets wrapped in the new `![...]` hidden command operator, rather than the uncaptured subproc operator `$[...]`.  Basically, `echo a and echo b` becomes `![echo a] and ![echo b]` rather than `![echo a and echo b]`.  The reasons for doing this are that:

1. It ensures that `and` and `or` for subprocs follow precisely the same logic as for other Python expressions.
2. The parser was crazy difficult to write and involved a lot of redundancy with the existing Python portion trying to make these operators work in subproc mode and pymode alike.

This is the cleanest solution I could come up with that involves relying on `![...]` rather than modifications to how `xonsh.built_ins.run_subproc()` works, which was my earlier approach.

Please let me know what you think!


melund:
Sounds reasonable to me. I don't see much reason to have this in subproc mode. I will try it out later tonight or tomorrow.

melund:
It seems to work as expected. I am not strong enough in the parser to give it a reliable code review. 

Are all subproc now wrapped in `![...]`? Are there any way to group the commands or negate the result? Not that I that think is necessary, but curious users like me may just try it. It caused my xonsh to freeze, no traceback.

$ echo 1 and (echo 2 and echo 3)

scopatz:
> Are all subproc now wrapped in ![...]? 

Yes, that is the trade off I made in order to not add any more craziness to the parser.

> Are there any way to group the commands or negate the result? Not that I that think is necessary, but curious users like me may just try it. It caused my xonsh to freeze, no traceback.

Now there is :).  I also adding grouping with parentheses, in case people want that too. I think @adqm opened an issue for this at some point.

> I am not strong enough in the parser to give it a reliable code review.

Only one way to learn :).  Also, in the end, this didn't change the parser at all, though are some updates to the lexer.

melund:
Works well for me now. If someone else can confirm, then I think it is ready for merge

gforsyth:
Mostly works on my end -- there are some things that don't make complete sense to me.

The examples in the docs work fine:

```console
(xonshtest)gil@theo~/tmp$ ls
bar  exists  foo
(xonshtest)gil@theo~/tmp$ ls exists and ls doesnt
exists
/usr/bin/ls: cannot access 'doesnt': No such file or directory
(xonshtest)gil@theo~/tmp$ ls doesnt and ls exists
/usr/bin/ls: cannot access 'doesnt': No such file or directory
(xonshtest)gil@theo~/tmp$ ls doesnt or ls exists
/usr/bin/ls: cannot access 'doesnt': No such file or directory
exists
```

but some others don't work the way I expect, e.g.

```console
(xonshtest)gil@theo~/tmp$ ls and ls
bar  exists  foo
(xonshtest)gil@theo~/tmp$ ls -l and ls
bar  exists  foo
(xonshtest)gil@theo~/tmp$ ls and ls -l
total 0
-rw-r--r-- 1 gil users 0 Apr 11 11:05 bar
-rw-r--r-- 1 gil users 0 Apr 11 11:02 exists
-rw-r--r-- 1 gil users 0 Mar 10 16:00 foo
(xonshtest)gil@theo~/tmp$ ls . and ls .
bar  exists  foo
bar  exists  foo
(xonshtest)gil@theo~/tmp$ ls . and ls -l
bar  exists  foo
NameError: name 'ls' is not defined
(xonshtest)gil@theo~/tmp$ ls . and ls
bar  exists  foo
NameError: name 'ls' is not defined
```

scopatz:
Good eye @gforsyth!  This is a left-most-name issue. Or rather a bunch of them.

scopatz:
How about now @melund and @gforsyth?

melund:
@scopatz ... 
The first works fine: 
```
$ ls and ls -l
bar  foo
total 0
-rw-r--r-- 1 mel 1049089 0 Apr 12 09:43 bar
-rw-r--r-- 1 mel 1049089 0 Apr 12 09:43 foo
```
But this doesn't 
```
$ ls . and ls -l
bar  foo
xonsh: subprocess mode: command not found: -l
    ll:  Alias
    gl:  Alias
    g:   Alias
    ls:  Alias
    d:   Alias
```

Lastly, this fails with a traceback: 
```
$ ls . and ls -l
bar  foo
xonsh: subprocess mode: command not found: -l
    ll:  Alias
    gl:  Alias
    g:   Alias
    ls:  Alias
    d:   Alias
mel@ANYBODY-MEL ~/Documents/scrats/test
$ ls . and ls
  File "c:\users\mel\documents\xonsh\xonsh\main.py", line 238, in main
    shell.cmdloop()
  File "c:\users\mel\documents\xonsh\xonsh\ptk\shell.py", line 111, in cmdloop
  File "c:\users\mel\documents\xonsh\xonsh\ptk\shell.py", line 80, in singleline
    display_completions_in_columns=multicolumn)
  File "c:\users\mel\documents\xonsh\xonsh\ptk\shortcuts.py", line 104, in prompt
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\interface.py", line 370, in run
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\eventloop\win32.py", line 78, in run
    callbacks.feed_key(k)
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\interface.py", line 899, in feed_key
    self._active_cli.input_processor.feed_key(key_press)
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\key_binding\input_processor.py", line 172, in feed_key
    self._process_coroutine.send(key_press)
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\key_binding\input_processor.py", line 144, in _process
    self._call_handler(matches[-1], key_sequence=buffer)
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\key_binding\input_processor.py", line 187, in _call_handler
    handler.call(event)
  File "c:\users\mel\documents\github\python-prompt-toolkit\prompt_toolkit\key_binding\registry.py", line 33, in call
    return self.handler(event)
  File "c:\users\mel\documents\xonsh\xonsh\ptk\key_bindings.py", line 132, in multiline_carriage_return
    carriage_return(b, event.cli)
  File "c:\users\mel\documents\xonsh\xonsh\ptk\key_bindings.py", line 55, in carriage_return
    elif not current_line_blank and not can_compile(b.document.text):
  File "c:\users\mel\documents\xonsh\xonsh\ptk\key_bindings.py", line 102, in can_compile
    locs=builtins.__xonsh_ctx__)
  File "c:\users\mel\documents\xonsh\xonsh\execer.py", line 95, in compile
  File "c:\users\mel\documents\xonsh\xonsh\execer.py", line 80, in parse
    tree = self.ctxtransformer.ctxvisit(tree, input, ctx, mode=mode)
  File "c:\users\mel\documents\xonsh\xonsh\ast.py", line 117, in ctxvisit
    node = self.visit(node)
  File "C:\Users\mel\Anaconda3\lib\ast.py", line 245, in visit
    return visitor(node)
  File "C:\Users\mel\Anaconda3\lib\ast.py", line 300, in generic_visit
    value = self.visit(value)
  File "C:\Users\mel\Anaconda3\lib\ast.py", line 245, in visit
    return visitor(node)
  File "c:\users\mel\documents\xonsh\xonsh\ast.py", line 192, in visit_Expr
    node.value = self.visit(node.value)  # this allows diving into BoolOps
  File "C:\Users\mel\Anaconda3\lib\ast.py", line 245, in visit
    return visitor(node)
  File "c:\users\mel\documents\xonsh\xonsh\ast.py", line 222, in visit_BoolOp
    node.values[i] = self.try_subproc_toks(val, strip_expr=True)
  File "c:\users\mel\documents\xonsh\xonsh\ast.py", line 153, in try_subproc_toks
    newnode = newnode.body
AttributeError: 'NoneType' object has no attribute 'body'
```

scopatz:
Yep, I can confirm these.

zommerfelds:
Automatically adding `![...]` in some cases seems a bit inconsistent to me. For example, now you can write:
`str(ls or "foo")`
But you can't write just:
`str(ls)`
This is confusing.

To me it feels like there has to be a decision about whether you should be able to mix in subproc mode inside of Python mode anywhere in the expression. I would love to see a language where every single variable that can't be resolved will be tried as a command and if no such command exist, it tries to interpret something like the innermost parenthesis. For example I could imagine this working: `int(wc -l myfile | cut -f1 -d ' ') + 5 + int(someothercommand)`. This behaves like an implicit `$()`.

If that is not the goal (which is fine too), I would still personally think that mixing both styles implicitly is not the best idea. If I want to write something in bash style I would just go with `&&`. If I know I'm in Python mode I would prefer to write `![...] and ![...]` explicitly.

Just my 2 cents. What do you think?

melund:
@zommerfelds I don't believe that it is the intention to mix python and subproc mode the way you show. It should only add the logical operators to subproc mode. So I think you found an unintended side effect, but @scopatz can maybe confirm it.

zommerfelds:
Just tried this:
```
> echo hello && test
AttributeError: 'NoneType' object has no attribute 'body'
```

scopatz:
> @zommerfelds I don't believe that it is the intention to mix python and subproc mode the way you show. It should only add the logical operators to subproc mode. So I think you found an unintended side effect, but @scopatz can maybe confirm it.

Yes, @melund, you are correct.  That is an unintended side effect. This clearly needs more work on the AST transformer.

zommerfelds:
@scopatz Thanks for the clarification.
Another issue with the `and` is that it adds an implicit delimiter which is not present in Bash. For example:
`echo I am happy and sad`

scopatz:
> @scopatz Thanks for the clarification.
Another issue with the and is that it adds an implicit delimiter which is not present in Bash. For example:
echo I am happy and sad

Yeah, this is a change to the language by making and/or keywords in subproc mode.  I am mostly OK with this because, 

1. and/or are already keywords in python mode and so it is a bit inconsistent that they are not in subproc mode (same with for, in, if else, etc.) 
2. I believe that and/or is more readable than `&&/||`
3. One can always recover a literal and/or with quotes, ie `echo "I am happy and sad"`, which is arguably what you should write in Bash anyway.
4. xonsh is not trying to be sh-language compliant. Differences are to be expected.

scopatz:
Ok all!  I believe that I have addressed all of the remaining issues. Please feel free to check it out, report errors, or merge.  Sorry it took me so long to get back to this.

gforsyth:
Awesome!  Everything I can think to try works as expected.  Merging!

melund:
All I have tried works as I excepted. Impressive :) LGTM

scopatz:
Thanks @melund and @gforsyth! This one was a long time in coming :)

Issue #782, opened 2016-04-01
Running a script with tracer fails


melund:
I tried to run a script with the tracer set, but I get the following error: 

```
C:\Users\mel\Documents\scrats>xonsh test.xsh
Traceback (most recent call last):
  File "C:\Users\mel\Anaconda3\Scripts\xonsh-script.py", line 9, in <module>
    load_entry_point('xonsh', 'console_scripts', 'xonsh')()
  File "c:\users\mel\documents\xonsh\xonsh\main.py", line 218, in main
    shell.execer.exec(code, mode='exec', glbs=shell.ctx)
  File "c:\users\mel\documents\xonsh\xonsh\execer.py", line 124, in exec
    return exec(code, glbs, locs)
  File "test.xsh", line 2, in <module>
    `.*.py`
  File "test.xsh", line 2, in <module>
    `.*.py`
  File "c:\users\mel\documents\xonsh\xonsh\tracer.py", line 80, in trace
    lexer=self.lexer, formatter=self.formatter)
  File "c:\users\mel\documents\xonsh\xonsh\tracer.py", line 96, in format_line
    fname = min(fname, replace_home(fname), os.path.relpath(fname), key=len)
  File "c:\users\mel\documents\xonsh\xonsh\environ.py", line 947, in _replace_home
    home = (builtins.__xonsh_env__['HOMEDRIVE'] +
  File "c:\users\mel\documents\xonsh\xonsh\environ.py", line 617, in __getitem__
    raise KeyError(e.format(key))
KeyError: 'Unknown environment variable: $HOMEDRIVE'
```

It seems like all the environment variables (not xonsh' own) are not set when running scripts.

adqm:
Ah, right.  I probably went too far with this.  As a temporary workaround, does it work if you run xonsh with the `--login` option?

melund:
@adqm ...  No. It doesn't seem to be a workaround: 

```
C:\Users\mel\Documents\scrats>xonsh --login test.xsh
Traceback (most recent call last):
  File "C:\Users\mel\Anaconda3\Scripts\xonsh-script.py", line 9, in <module>
    load_entry_point('xonsh', 'console_scripts', 'xonsh')()
  File "c:\users\mel\documents\xonsh\xonsh\main.py", line 218, in main
    shell.execer.exec(code, mode='exec', glbs=shell.ctx)
  File "c:\users\mel\documents\xonsh\xonsh\execer.py", line 124, in exec
[...]

PR #792, opened 2016-04-12
Load os.environ env-vars when running scripts

melund:
@scopatz .. This fix ensures that environment variables from os.environ are always loaded and that $XONSH_VERSION is always set. Even when running scripts.
#782 #


scopatz:
Looks great, thanks!

PR #794, opened 2016-04-12
Expand tildas when using $CDPATH

airdrik:
When $CDPATH contains user directory shortcuts '~', xonsh should expand those so that it can find and return the appropriate directory path.

### DIFF
--- a/xonsh/dirstack.py
+++ b/xonsh/dirstack.py
@@ -41,7 +41,7 @@ def _try_cdpath(apath):
     env = builtins.__xonsh_env__
     cdpaths = env.get('CDPATH')
     for cdp in cdpaths:
-        for cdpath_prefixed_path in iglob(os.path.join(cdp, apath)):
+        for cdpath_prefixed_path in iglob(os.path.expanduser(os.path.join(cdp, apath))):
             return cdpath_prefixed_path
     return apath
 


airdrik:
I looked at the tests for dirstack.py and wasn't sure how to write a test to cover this.

xonsh/dirstack.py
```diff
@@ -41,7 +41,7 @@ def _try_cdpath(apath):
     env = builtins.__xonsh_env__
     cdpaths = env.get('CDPATH')
     for cdp in cdpaths:
-        for cdpath_prefixed_path in iglob(os.path.join(cdp, apath)):
+        for cdpath_prefixed_path in iglob(os.path.expanduser(os.path.join(cdp, apath))):
```
scopatz: You might want to use `builtins.__xonsh_expand_path__()` instead of `os.path.expanduser()` here.

scopatz:
Thanks for putting this in @airdrik!

scopatz:
You can add or modify tests to [`xonsh/tests/test_dirstack.py`](https://github.com/scopatz/xonsh/blob/master/tests/test_dirstack.py), which would be most appreciated.  

Also, you might want to have a look at the developers guide: http://xon.sh/devguide.html

Also also, if you could please add an entry to the CHANGELOG, that'd be great!

airdrik:
In trying to put together some tests, I went back and tried the current code, and it turns out the expected behavior is already there.  I looked to see if I still had the failed attempt that prompted this change, but it was from a previous session and I don't have the details anymore.

scopatz:
Thanks for putting this in anyway @airdrik :)

PR #797, opened 2016-04-16
Aligns the colors with prompt_toolkit specification of colors.

melund:
This aligns the colors with the specification used by prompt_toolkit. 
https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/terminal/win32_output.py#L442

Without this, some colors were simply mapped incorrectly, at least for me on windows. For example 'INTENSE_GREEN' is not just normal green. 
 
This will not be a problem as soon as pygments 2.2 is released because prompt_toolkit will then used the `#ansicolors` specification instead.

### DIFF
--- a/xonsh/pyghooks.py
+++ b/xonsh/pyghooks.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """Hooks for pygments syntax highlighting."""
+import os
 import re
 import string
 import builtins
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks for putting this in @melund! This however, changes the colors away from the Standard VGA colors that are implemented now: https://en.wikipedia.org/wiki/ANSI_escape_code#Colors

Maybe you could add an `elif` block for `ON_WIN` to set the colors like you have them here?

melund:
Yes. I also found the table, and I did consider to put in an `elif`for windows. But I also noted that while the colors are indeed closest to the VGA specification, but it doesn't follow it for a number of colors. yellow (vga: AA5500  xonsh: FFFF00), Intense Red (vga: FF5555, xonsh: FF0000), Intense blue (vga: 5555FF, xonsh: 0000FF), intense purple (vga:FF55FF, xonsh: FF00FF)

So I thought I could just suggest the change without. I think it is really, `prompt_toolkit` that should maybe be changed. Anyway, I will add `elif`clause. It is only really relevant until next pygments is released.

melund:
@scopatz I updated with the new colors only for windows.

melund:
@BYK... You can also merge this if you can accept the changes. It only affects colors cmd.exe and not in ConEmu

scopatz:
This looks great to me! I am going to go ahead and merge.  Thanks a ton @melund!

PR #798, opened 2016-04-18
Fix deadlock on windows

melund:
@BYK and @scopatz  This is a fix for deadlock on windows when running subproc that generates enough data to fill the os pipe. 

This may be a hacky way to solve it, but it is a minimal change and it seems to work.

### DIFF
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -6,6 +6,7 @@ import time
 import signal
 import builtins
 from subprocess import TimeoutExpired
+from io import BytesIO
 
 from xonsh.tools import ON_WINDOWS
 
@@ -41,13 +42,16 @@ if ON_WINDOWS:
             return
         while obj.returncode is None:
             try:
-                obj.wait(0.01)
+                outs, errs = obj.communicate(timeout=0.01)
             except TimeoutExpired:
                 pass
             except KeyboardInterrupt:
                 obj.kill()
+                outs, errs = obj.communicate()
         if obj.poll() is not None:
             builtins.__xonsh_active_job__ = None
+            obj.stdout = BytesIO(outs)
+            obj.stderr = BytesIO(errs)
 
 else:
     def _continue(obj):
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -56,6 +56,8 @@ Current Developments
 * Fixed environment variables from os.environ not beeing loaded when a running 
   a script
 * Fixed bug that prevented `source-alias` from working. 
+* Fixed deadlock on Windows when runing subprocess that generates enough output
+  to fill the OS pipe buffer 
 
 **Security:** None
 


melund:
I realize that this may only be a fix when popen is the class used to run the subproc. Or am I wrong here? I could add a check that the process object is popen and only call `obj.communicate()` in that case. Otherwise maybe @scopatz or @adqm can come up with generic fix....

xonsh/jobs.py
```diff
@@ -41,13 +42,16 @@ def wait_for_active_job(signal_to_send=None):
...
+                outs, errs = obj.communicate(timeout=0.01)
             except TimeoutExpired:
                 pass
             except KeyboardInterrupt:
                 obj.kill()
+                outs, errs = obj.communicate()
         if obj.poll() is not None:
             builtins.__xonsh_active_job__ = None
+            obj.stdout = BytesIO(outs)
+            obj.stderr = BytesIO(errs)
```
scopatz: Doesn't this need to append, rather than overwrite?

scopatz:
Hi @melund, first off, sorry for taking a couple of days to review. My family is visiting, so I have had less time than usual :).

Also, I don't think that this is hacky. It is probably the right thing to do.

scopatz:
Or at least the right approach :)

scopatz:
Also, what was the test case that you were using to fill up the I/O buffer?

melund:
@scopatz 
I used the example by @BYK. 
```
out = $(cmd /c help)
```
Reading up on the io module, I am little puzzled why it would block. The default buffersize is:

```
$ io.DEFAULT_BUFFER_SIZE
8192
```
while the len of the output is less.
```
$ out = $(cmd /c help)
$ len(out)
5642
```

xonsh/jobs.py
melund: @scopatz ... I don't think I can append to the streams. 
```
$ cmd = ['C:\\Windows\\system32\\cmd.exe', '/C', 'help']
$ proc = Popen(cmd, universal_newlines=False, stdin=0, stdout=PIPE, stderr=PIPE)
$ proc.stdout.writable()
False
```
Is it important? We know the `stderr` and `stdout` streams have been emptied.

scopatz:
Sorry about the delay in looking at this again!  Looks like there is now a merge conflict :(

BYK:
I'm back with my Windows PC and can help with this if any needed.

BYK:
>and can help with this if any needed.

This includes reviving the branch by a proper rebase onto master :)

melund:
@scopatz @BYK ... I have rebased...

melund:
It is just really a prof of concept or illustration of where the bug was ... If any of you have different idea for a fix you are well come to take over...

scopatz:
Overall, I think this looks fine.  Maybe @BYK should merge it though since I cannot test it myself.

melund:
Great. But all of you can test that it doesn't have any adverse effect on Linux and mac.

BYK:
Seems to work pretty great on Windows. Gonna check it out my OS X and then merge.

BYK:
Confirmed no problems on OS X.

Issue #800, opened 2016-04-20
source-bat alias for source-cmd


scopatz:
Might be nice to have.

scopatz:
Thanks @melund

PR #801, opened 2016-04-20
Added `source-bat` as an alias for `source-cmd`

melund:
Maybe not the most important PR. We already had `call` as an alias. Anyway, since @scopatz suggested in #800 I had to implement it.


scopatz:
Thanks @melund!

PR #802, opened 2016-04-20
Bugfix for source-cmd

melund:
This fixes a bug which I added in one of the last commits on source_cmd PR. 
The bug means that arguments to the bat files were skipped. Sorry :)

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -254,7 +254,8 @@ def source_cmd(args, stdin=None):
     args[0] = fpath if fpath else args[0]
     if not os.path.isfile(args[0]):
         raise FileNotFoundError(args[0])
-    prevcmd = 'call "{}"\necho off'.format(args[0])
+    prevcmd = 'call "{}"'.format('" "'.join(args))
+    prevcmd += '\necho off'
     args.append('--prevcmd={}'.format(prevcmd))
     args.insert(0, 'cmd')
     args.append('--interactive=0')


melund:
@BYK Maybe you can exercise your Push rights to get this in :)

scopatz:
LGTM

melund:
@scopatz You are also welcome to merge :) I guess there will be another chance for @BYK. I am not too happy with this bug beeing in master

scopatz:
Yeah. Happy to merge it.

xonsh/aliases.py
```diff
@@ -254,7 +254,8 @@ def source_cmd(args, stdin=None):
     args[0] = fpath if fpath else args[0]
     if not os.path.isfile(args[0]):
         raise FileNotFoundError(args[0])
-    prevcmd = 'call "{}"\necho off'.format(args[0])
+    prevcmd = 'call "{}"'.format('" "'.join(args))
```
BYK: This is probably a very bad way of escaping. I've had to deal with CMD escaping and it encourages people to use `^` for all meta characters instead of quoting them.

BYK: Source: https://blogs.msdn.microsoft.com/twistylittlepassagesallalike/2011/04/23/everyone-quotes-command-line-arguments-the-wrong-way/

melund: I knew that there could be a problem, but didn't know this solution. This may actually be a general problem when xonsh launches subproc on windows. Is this something we have to consider?

 I think don't think we need to think about escaping anything here though, but I do agree that we shouldn't wrap anything in quotes either. The command here is written to a batch file which is then parsed to 'cmd.exe' So the batch file should essentially just contain the same thing that the user wrote in the xonsh shell. We shouldn't try to wrap anything in quotes.

melund: @BYK I was thinking it was up to the user to escape their commands if they were using `source-cmd`.

On the other hand it may not be what people expect. Also Popen normally takes care of escaping the individual arguments, but we can't use that here here since the whole command is written to a batch file. Thus we have to use the whole procedure from your link not only the cmd.exe specific ^escapes.

BYK: Interesting but you are probably right. Btw. I believe `Popen` only takes care of the ARGV part, not the CMD part but I may be wrong.

melund: Yes I also think Popen handles the argv part. But it can't rely on that here since everything is written to a batch file. I will make a fix.

PR #803, opened 2016-04-20
Fixes a bug in source_alias

melund:
@scopatz I stumbled on this while looking at source_cmd stuff.


scopatz:
Looks good to me. Thanks again @melund!

PR #804, opened 2016-04-20
Re-enabled the app section in the meta.yaml (conda build script)

melund:
This makes conda packages create a xonsh menu entry when installing.


scopatz:
Looks great! Thanks @melund.  We should probably do this over at https://github.com/conda-forge/xonsh-feedstock too.

melund:
True. We also need to come up with a way to copy the packages over from conda-forge. I just can't see a way to do this without exposing the Token to the xonsh organisation on Anaconda.org

scopatz:
Ahh yeah. I mean, alternatively, we can tell people to just install from the conda-forge channel...

PR #805, opened 2016-04-22
quote arguments for cmd.exe

melund:
This adds code to quote arguments for cmd.exe, which was needed for source_cmd. It implements the suggestion from @BYK's link on how to quote arguments (what I except Popen also does). It also adds ^escapes requred by cmd.exe. Here it reuses the code from `escape_windows_title_string()` which already did similar stuff. I just renamed the function to  `escape_windows_cmd_string()`

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -7,11 +7,11 @@ import nose
 from nose.tools import assert_equal, assert_true, assert_false
 
 from xonsh.lexer import Lexer
-from xonsh.tools import ( subproc_toks, subexpr_from_unbalanced, is_int, 
+from xonsh.tools import (subproc_toks, subexpr_from_unbalanced, is_int, 
     always_true, always_false, ensure_string, is_env_path, str_to_env_path, 
     env_path_to_str, escape_windows_cmd_string, is_bool, to_bool, bool_to_str, 
     ensure_int_or_slice, is_float, is_string, check_for_partial_string,
-    argvquote )
+    argvquote)
 
 LEXER = Lexer()
 LEXER.build()
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -7,10 +7,11 @@ import nose
 from nose.tools import assert_equal, assert_true, assert_false
 
 from xonsh.lexer import Lexer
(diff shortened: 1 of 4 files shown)


tests/test_tools.py
```diff
@@ -7,10 +7,11 @@
 from nose.tools import assert_equal, assert_true, assert_false
 
 from xonsh.lexer import Lexer
-from xonsh.tools import subproc_toks, subexpr_from_unbalanced, is_int, \
-    always_true, always_false, ensure_string, is_env_path, str_to_env_path, \
-    env_path_to_str, escape_windows_title_string, is_bool, to_bool, bool_to_str, \
-    ensure_int_or_slice, is_float, is_string, check_for_partial_string
+from xonsh.tools import ( subproc_toks, subexpr_from_unbalanced, is_int,
```
scopatz: Can you remove the space from the inside of the parentheses, please?

xonsh/aliases.py
```diff
@@ -254,8 +255,10 @@ def source_cmd(args, stdin=None):
     args[0] = fpath if fpath else args[0]
     if not os.path.isfile(args[0]):
         raise FileNotFoundError(args[0])
-    prevcmd = 'call "{}"'.format('" "'.join(args))
-    prevcmd += '\necho off'
+    prevcmd = 'call '
+    prevcmd += ' '.join([argvquote(arg,force=True) for arg in args])
```
scopatz: PEP8 space after comma

scopatz:
Generally looks great @melund! Just a couple of minor style issues. Thanks for putting this in!

melund:
@scopatz

scopatz:
LGTM - Thanks!

BYK:
Nice!

melund:
@BYK. I think some of this should be added to the sudo Windows stuff or what do you think. It skips Popen and some of the commands call cmd.exe

BYK:
@melund hmm, possibly. I think it is only necessary for the CMD command proxies. Hmm, actually I should probably use ARGV escaping for the other stuff. I'll look into this when I'm back home, next to my Windows machine.

Should we create a separate issue to track this?

scopatz:
@BYK, I think another issue would be nice :)

Issue #809, opened 2016-04-28
alias 'which' to also look up aliases.


scopatz:
This would be nice.

gforsyth:
Hey @scopatz, 

Are you envisioning it working something like the following?

```console
$ which source
<function.xonsh.aliases.source_alias>
$ which ipynb
['jupyter', 'notebook', '--no-browser']
$ which cat
/usr/bin/cat
```

scopatz:
Yeah, something like that. That would be fine with me :)

PR #810, opened 2016-04-29
Alias for `which` to support alias lookups

gforsyth:
Caveats:  
1. I think Windows will just ignore this entirely
2. Some, but not all, `which` flags are supported

As requested by @scopatz in #809 

```console
(xonshtest)gil@theo~/git/myxonsh alias_which$ which ls
ls -> ['ls', '--color=auto', '-v']
(xonshtest)gil@theo~/git/myxonsh alias_which$ which -a ls
ls -> ['ls', '--color=auto', '-v']
/usr/bin/ls
(xonshtest)gil@theo~/git/myxonsh alias_which$ which timeit
timeit -> <function timeit_alias at 0x7f6558ec3d08>
(xonshtest)gil@theo~/git/myxonsh alias_which$ which -a timeit
timeit -> <function timeit_alias at 0x7f6558ec3d08>
(xonshtest)gil@theo~/git/myxonsh alias_which$ which DERP
xonsh.tools.XonshError: DERP not in /home/gil/anaconda/bin:/usr/bin:/usr/local/bin:/usr/bin/vendor_perl/:/usr/bin/core_perl/:/opt/cuda/bin or xonsh.builtins.aliases
(xonshtest)gil@theo~/git/myxonsh alias_which$ which -v ls
GNU which v2.21, Copyright (C) 1999 - 2015 Carlo Wood.
GNU which comes with ABSOLUTELY NO WARRANTY;
This program is free software; your freedom to use, change
and distribute this program is protected by the GPL.
(xonshtest)gil@theo~/git/myxonsh alias_which$ which --skip-alias ls
/usr/bin/ls
```

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -453,7 +453,6 @@ def make_default_aliases():
                                            'activate.bat']
             default_aliases['deactivate'] = ['source-cmd-keep-promt',
                                              'deactivate.bat']
-        default_aliases['which'] = ['where']
         if not locate_binary('sudo'):
             import xonsh.winutils as winutils
 
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -329,7 +329,7 @@ def which(args, stdin=None):
     """
     desc = "Parses arguments to which wrapper"
     parser = ArgumentParser('which', description=desc)
-    parser.add_argument('arg', type=str, default='-h',
+    parser.add_argument('arg', type=str,
                         help='The executable or alias to search for')
     parser.add_argument('-a', action='store_true', dest='all',
                         help='Show all matches in $PATH and xonsh.aliases')
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -5,6 +5,7 @@ import os
 import shlex
 import builtins
 import sys
+import subprocess
 from argparse import ArgumentParser
 from collections.abc import MutableMapping, Iterable, Sequence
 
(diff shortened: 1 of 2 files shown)


melund:
Cool. But why ignore Windows :( 

If people (like me) have git installed, they may also have all the msys unix tools on their path. In that case this is also usefull. It is not working now, since `which` is aliased to `where` on windows. That could be skipped if `which` was on the path. 

An other option is to implement the functionality directly in python. We do have [`locate_binary`](https://github.com/scopatz/xonsh/blob/c3b94e52656f96914849e091cf92fbd6d197d45e/xonsh/environ.py#L739) which is kind of a python implementation of `which`. Maybe that could be used instead.

gforsyth:
Hey @melund, I was only ignoring Windows since `where` has very different functionality than `which`, but we can certainly short-circuit that if msys stuff is present.  

And you're right, we can probably do a pure Python implementation that could be included in @adqm's `xoreutils`

melund:
Yes, and I am not too crazy about the current `aliases['which']='where'`part. That is just bound to make people confused.

gforsyth:
Actually @melund, just curious about behavior here -- if you load `xonsh` and then manually set `aliases['which'] = xonsh.aliases.which` does something like that work?

melund:
I already tried your branch with the `aliases['which']='where'` commented out. It worked fine :) But using `aliases['which'] = xonsh.aliases.which` works fine as well.

melund:
Well, the only downside to the MSYS `which` is that it returns these weird paths: 
```
$ which xonsh
/c/Users/mel/Anaconda3/Scripts/xonsh
```

scopatz:
First off, this is amazing @gforsyth!

xonsh/aliases.py
```diff
@@ -313,6 +314,37 @@ def bang_bang(args, stdin=None):
...
+    If '-a' flag is passed, run both to return both `xonsh` match and
+    `which` match
+    """
+    match = set(args).intersection(builtins.aliases)
+    #skip alias check if user asks a skip or the version number
+    if (match and
+      not '--skip-alias' in args and
+      not {'--version','-v','-V'}.intersection(set(args))):
+        match = match.pop()
+        print('{} -> {}'.format(match,builtins.aliases[match]))
```
scopatz: PEP8 space after comma.

xonsh/aliases.py
```diff
@@ -313,6 +314,37 @@ def bang_bang(args, stdin=None):
...
+    Checks if argument is a xonsh alias, then if it's an executable, then
+    finally throw an error.
+    If '-a' flag is passed, run both to return both `xonsh` match and
+    `which` match
+    """
+    match = set(args).intersection(builtins.aliases)
+    #skip alias check if user asks a skip or the version number
+    if (match and
+      not '--skip-alias' in args and
+      not {'--version','-v','-V'}.intersection(set(args))):
```
scopatz: Is there any reason not to use argparse? This isn't blocking, I am just curious...

scopatz:
@melund, should we remove the `aliases['which']='where'` line in favor of this new function? I think that would be best, but I think it makes the most sense if this is your call :)

melund:
@scopatz @gforsyth I agree that the current alias`aliases['which']='where'` should go .....

gforsyth:
@melund @scopatz 
If we remove the `which->where` alias, do we just assume that anyone who thinks to type `which` will have msys stuff available?  (That's fine with me, but just clarifying)

Also switched the argument parsing over to `argparse` which makes it a fair bit simpler to read.  Although passing through the `-v` flag was trickier than I was expecting.

melund:
Well. If someone on windows types `which` and they don't have the binary on path, they should just get an error.

gforsyth:
@melund -- that's a remarkably clear way of looking at it.  I'll strip out the alias.

scopatz:
Thanks a million @gforsyth!  Sorry about the delay in merging.  Family leaves tomorrow :)

PR #812, opened 2016-05-03
Prevented hg branch call for default branch

Siecje:
When ```.hg/branch``` doesn't exist, ```hg branch``` is used to determine the branch.

If the file does not exist it will be the ```default``` branch.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -887,7 +887,7 @@ def get_hg_branch(cwd=None, root=None):
             with open(branch_path, 'r') as branch_file:
                 branch = branch_file.read()
         else:
-            branch = call_hg_command(['branch'], cwd)
+            branch = 'default'
 
         if os.path.exists(bookmark_path):
             with open(bookmark_path, 'r') as bookmark_file:


melund:
Not too family with mercurial, but it seems reasonable. I just gave it a test on my `pygments` which is checkout it works well. I will merge it in. Thanks a lot @Siecje

scopatz:
@Siecje do you have any documentation that this is valid? I'd hate to be reporting the wrong branch.... Or, maybe @ngoldbaum or @MathewTurk knows off the top of their heads.

ngoldbaum:
I think @Siecje is right, according to "hg help glossary":

```
    Branch, default
        The branch assigned to a changeset when no name has previously been
        assigned.
```

scopatz:
Thanks @ngoldbaum!

Siecje:
As far as I know the only time you won't have ```.hg/branch``` is when you have just created a repo.

If you specify a branch it will create ```.hg/branch```.

When you clone a repo it will have ```.hg/branch```.

The reason I found this was because I had just created a new hg repo and the prompt was slow from calling ```hg branch```.

scopatz:
@Siecje, yeah I grok the use-case.  I was just curious if you had any documentation to back up the change of assuming a default name of "default".  @ngoldbaum provided this through `hg help glossary`

scopatz:
As a non-hg-user I figured you all would be much faster at finding the docs than myself.

PR #816, opened 2016-05-05
Set the prompt ending to red if the previous command failed.

Siecje:


melund:
I am not sure this is something that would please everyone. It really a candidate for the xonsh extension (xontrib) we have discussed before. 

If there was a way we could show the fail/success status on the previous prompts then I could see the real benefit. I don't know if something like that is possible with prompt-toolkit. Can we change the prompt just before the user gets a new one?

scopatz:
Hi @Siecje - I agree with @melund that this is more of xontrib feature.  Maybe you could add a new package name and stick this code in there?

Siecje:
@melund I'm not sure I understand what you mean by previous prompts. 
If the previous command didn't have a return code of 0 then the prompt ending (eg, $) will be red. 
It is possible to display the return code.
What do you mean by getting a new prompt? Each time you press enter is that a new prompt? Then yes it is changed. I'm pretty sure I'm missing something.

@scopatz Are you suggesting creating a PyPI package for this? Why wouldn't you want this feature?

melund:
What I was trying to say was that I don't like that one prompt reflects what happened at the previous line. When looking at the output in the terminal it is not easy to see what is gooing on: 

image

If we could somehow change the prompt just before the new prompt was generated we could have something like this: 

image

That would be a something which I see being a part of the default behaviour

Siecje:
That makes sense. To me that seems like YAGNI because it seems like a lot
of work.

scopatz:
Hi @Siecje, let me clarify a bit.  First, I think that the idea behind the `end_color()` function is a good one and something that should probably be in xonsh.  

The objection (at least from me and it sounds like @melund too) is actually adding this behavior to the default prompt. I know people who have this kind of setup in other shells, and it works for them.  Personally, I find it distracting and don't want it as part of the *default*.  That said, we should absolutely give people the ability to set up their prompt to do this.

So I think that you should keep the `end_color()` stuff, but its name should probably be changed to `last_cmd_color()` or something more descriptive.  Folks could use this key at any place in their prompt, not just at the end. You should also add a description of this key to the docs (http://xon.sh/tutorial.html#customizing-the-prompt) so people know it exists. Then return the default prompt to what it was.  

The point about the `xontrib` thing is that we have been discussing a way to have 3rd party extensions, like sphinx has.  Having a xontribtion that changes the default prompt to one that uses the ending color in this way seems reasonable.  Then people would only need to enable this `xontrib` to get your ideal prompt.  There are a few extensions that we are planning on having distributed along with xonsh, and this alternative prompt could/should be one of them.

So it isn't that we don't want this; we do! We just want to sort it out a bit differently than currently implemented.  I hope this makes sense.

Siecje:
If there is a lot of output and you changed the prompt created before the command is run. Then you would have to scroll up to notice the different colour.

scopatz:
Hi @Siecje - now that the xontribs PR has been merged in, you can add your prompt as a very simple  xontrib.  Would you mind doing it this way?

Siecje:
Yeah I will

scopatz:
Thanks!  No need to close this PR, unless you want to start fresh :)

PR #818, opened 2016-05-05
Add return code to source_foreign functions

### LINKED ISSUE #817: source-bash return code will always be 0
Even if the file doesn't exist.

melund:
This fixes #817 by adding a return code to source_foreign

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -218,6 +218,8 @@ def source_foreign(args, stdin=None):
                                           funcscmd=ns.funcscmd,
                                           sourcer=ns.sourcer,
                                           use_tmpfile=ns.use_tmpfile)
+    if fsenv is None:
+        return (None, "xonsh: error: Failed to source: {}".format(ns.prevcmd), 1)
     # apply results
     env = builtins.__xonsh_env__
     denv = env.detype()
@@ -227,10 +229,9 @@ def source_foreign(args, stdin=None):
         env[k] = v
     # If run in un-safe mode we are sure the command completed correctly,
     # thus we can remove any env-vars that were unset by the script.
-    if not ns.safe:
-        for k in denv:
-            if k not in fsenv:
-                env.pop(k, None)
+    for k in denv:
+        if k not in fsenv:
+            env.pop(k, None)
     baliases = builtins.aliases
     for k, v in fsaliases.items():
         if k in baliases and v == baliases[k]:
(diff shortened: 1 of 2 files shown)


melund:
I also added a non-zero return value to the xexec alias if the commands fails

gforsyth:
Hey @melund -- I'm still getting a `0` returncode when I `source-bash` a nonexistent file.

```console
(xonshtest2)gil@theo~/tmp$ ls
foo
(xonshtest2)gil@theo~/tmp$ source-bash bar
(xonshtest2)gil@theo~/tmp$ __xonsh_history__.rtns[-1]
0
```

xonsh/aliases.py
```diff
@@ -282,7 +284,7 @@ def xexec(args, stdin=None):
         except FileNotFoundError as e:
             return 'xonsh: ' + e.args[1] + ': ' + args[0] + '\n'
```
scopatz: This line could also use a non-zero return code while we are at it :)

scopatz:
@melund, First off, thanks a million for putting this in!

@gforsyth @melund - my guess is that this is because we are not running `set -e` at the top of the Bash script that get created in `foreign_shell_data()`.  Thus, the non-existent file is name is being passed to a script that looks like:

```bash
doesntexist  # fails
echo __XONSH_ENV_BEG__  #runs anyway
env
echo __XONSH_ENV_END__
...
```

Rather than 

```bash
set -e
doesntexist  # fails and exits
echo __XONSH_ENV_BEG__
env
echo __XONSH_ENV_END__
...
```

Thus the function actually does succeed.  We should probably implement this fail-on-error semantic.  Not sure that this PR is the right place for it.  Happy to merge this as-is, even though it doesn't solve the original problem yet.

xonsh/aliases.py
melund: oh yes... Maybe we shouldn't add non-zero return codes to this function. It is the equivalent to python exec so it should only use Exceptions. What is you thought on that?

melund: Or should it do both?

melund: Well,  I guess it should just return a non-zero return code. I will add that.

melund:
Sorry for beeing a too quick at pushing this. This actually doesn't work on Win either. The bat file need a post command with `if errorlevel 1 exit 1`.. I have added that now.

I think someone else should make a PR with the `set -e` stuff. I don't have the linux box to test it on.

melund:
@scopatz .. I think this is ready for merge. Maybe you can add and test the `set -e`stuff.

scopatz:
Thanks a ton @melund!  I'll try adding the `set -e` stuff.

PR #819, opened 2016-05-05
Changes for better compatibility with prompt-toolkit 1.0.0

jonathanslenders:
Without this change, the shell would stay in Emacs mode.
Changelog: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/CHANGELOG
Most noticable are the improved Vi bindings.

### DIFF
--- a/xonsh/ptk/shell.py
+++ b/xonsh/ptk/shell.py
@@ -2,6 +2,7 @@
 """The prompt_toolkit based xonsh shell."""
 import builtins
 
+from prompt_toolkit.enums import EditingMode
 from prompt_toolkit.key_binding.manager import KeyBindingManager
 from prompt_toolkit.auto_suggest import AutoSuggestFromHistory
 from prompt_toolkit.layout.lexers import PygmentsLexer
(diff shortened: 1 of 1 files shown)


gforsyth:
Hey @jonathanslenders -- 

First, congratulations on the milestone release!  
Second, thanks for the heads-up on the interface changes.  

I'm going to leave this open for now and also push up similar changes that preserve compatibility with versions of `prompt_toolkit < 1.0.0` and see what @scopatz would like to do.  

Thanks again!

scopatz:
Thanks for putting this in @jonathanslenders!

scopatz:
Out of curiosity, is it safe to assume that everything less than v1.0 is no longer supported?

jonathanslenders:
If required for some reason, I can always push a 0.61 release, but probably not. I'll try to send out pull requests for all projects that depend on prompt-toolkit to upgrade to 1.x.x.

melund:
I think this was handled by @gforsyth PR #820 ... So I will just close.

scopatz:
Thanks @jonathanslenders - I was asking so we would know what we can deprecate :)

PR #820, opened 2016-05-05
add support for prompt_toolkit-1.0.0 and preserve support for previous versions

gforsyth:
hat tip and congratulations to @jonathanslenders in #819 

This makes exactly the same changes but doesn't break `xonsh` for people using `prompt_toolkit` versions `0.57` or `0.58`.  

Is it time that we start thinking about a deprecation schedule for those?

### DIFF
--- a/xonsh/ptk/shortcuts.py
+++ b/xonsh/ptk/shortcuts.py
@@ -6,6 +6,8 @@ from prompt_toolkit.shortcuts import (create_prompt_application,
 
 from xonsh.shell import prompt_toolkit_version_info
 
+import builtins
+
 class Prompter(object):
 
     def __init__(self, cli=None, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


xonsh/ptk/shortcuts.py
```diff
@@ -69,6 +71,15 @@ def prompt(self, message='', **kwargs):
             if self.major_minor <= (0, 57):
                 kwargs.pop('get_rprompt_tokens', None)
                 kwargs.pop('get_continuation_tokens', None)
+            #VI_Mode handling changed in prompt_toolkit v1.0
```
scopatz: PEP8 space after `#`

scopatz:
Thanks for putting this in @gforsyth! And yes, many congrats to @jonathanslenders!

I think it is probably time to deprecate v0.57 and maybe even v0.58

scopatz:
Thanks a ton @gforsyth!

PR #822, opened 2016-05-07
Bundle PLY with xonsh

melund:
I haven't added code to use an installed version instead of the bundled. Do we want to do that? In any case, we would have to check the version, and PLY 3.8 doesn't report its correct version number yet (https://github.com/dabeaz/ply/issues/94) ...

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,8 +1,8 @@
 version: 0.2.7.{build}
 os: Windows Server 2012 R2
 install:
-- C:\Python34\Scripts\pip install ply pyreadline nose pygments prompt_toolkit
-- C:\Python35\Scripts\pip install ply pyreadline nose pygments prompt_toolkit
+- C:\Python34\Scripts\pip install pyreadline nose pygments prompt_toolkit
+- C:\Python35\Scripts\pip install pyreadline nose pygments prompt_toolkit
 build_script:
 - C:\Python34\python setup.py install
 - C:\Python35\python setup.py install
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,8 +1,8 @@
 version: 0.2.7.{build}
 os: Windows Server 2012 R2
 install:
-- C:\Python34\Scripts\pip install pyreadline nose pygments prompt_toolkit
-- C:\Python35\Scripts\pip install pyreadline nose pygments prompt_toolkit
+- C:\Python34\Scripts\pip install ply pyreadline nose pygments prompt_toolkit
+- C:\Python35\Scripts\pip install ply pyreadline nose pygments prompt_toolkit
 build_script:
 - C:\Python34\python setup.py install
 - C:\Python35\python setup.py install
--- a/docs/dependencies.rst
+++ b/docs/dependencies.rst
@@ -5,7 +5,6 @@ Xonsh currently has the following external dependencies,
 *Run Time:*
 
     #. Python v3.4+
-    #. PLY
     #. prompt-toolkit (optional)
     #. Jupyter (optional)
     #. setproctitle (optional)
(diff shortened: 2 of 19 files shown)


.appveyor.yml
```diff
@@ -1,8 +1,8 @@
 version: 0.2.7.{build}
 os: Windows Server 2012 R2
 install:
-- C:\Python34\Scripts\pip install ply pyreadline nose pygments prompt_toolkit
-- C:\Python35\Scripts\pip install ply pyreadline nose pygments prompt_toolkit
+- C:\Python34\Scripts\pip install pyreadline nose pygments prompt_toolkit
+- C:\Python35\Scripts\pip install pyreadline nose pygments prompt_toolkit
```
scopatz: We probably should still install ply so that we are testing the latest version on pypi

docs/dependencies.rst
```diff
@@ -5,7 +5,6 @@ Xonsh currently has the following external dependencies,
 *Run Time:*
 
     #. Python v3.4+
-    #. PLY
```
scopatz: I would change this to `#. PLY (optional, included with xonsh)` or similar

docs/index.rst
```diff
@@ -221,7 +221,6 @@ Xonsh currently has the following external dependencies,
 *Run Time:*
 
     #. Python v3.4+
-    #. PLY
```
scopatz: Same here

recipe/meta.yaml
```diff
@@ -15,12 +15,10 @@ build:
 requirements:
   build:
     - python
-    - ply
```
scopatz: Also, I would keep these around so that the dependency is managed elsewhere.

tests/test_lexer.py
```diff
@@ -9,7 +9,7 @@
 
 import nose
 
-from ply.lex import LexToken
+from xonsh.ply.lex import LexToken
```
scopatz: I would change this to 

```python
try:
    from ply.lex import LexToken
except ImportError:
    from xonsh.ply.lex import LexToken
```

xonsh/lexer.py
```diff
@@ -8,7 +8,7 @@
 from io import BytesIO
 from keyword import kwlist
 
-from ply.lex import LexToken
+from xonsh.ply.lex import LexToken
```
scopatz: Same

xonsh/parsers/base.py
```diff
@@ -1,10 +1,9 @@
 # -*- coding: utf-8 -*-
 """Implements the base xonsh parser."""
 from collections import Iterable, Sequence, Mapping
-
-from ply import yacc
```
scopatz: Same try-except idea here.

xonsh/xonfig.py
```diff
@@ -9,8 +9,7 @@
 from pprint import pformat
 from argparse import ArgumentParser
 
-import ply
```
scopatz: Same try-except idea here.

scopatz:
Thanks a ton for putting this in @melund! I haven't gotten around to it myself :)

melund:
I also did the try-import blocks first. But then I was wondering if needed to add test so we don't use older installed versions PLY. Then I discovered that 3.8 was reporting it self as 3.7 (a bug) and decided to just do the simple thing first.

@scopatz if you think we don't need to check the version then I will add `try` blocks back.

scopatz:
@melund - yeah, I don't think we need to test the version.  It shouldn't matter if the user wants to have an older version of PLY installed.

scopatz:
At least IMHO

scopatz:
Is this ready for another look @melund, or should we hold off?

melund:
@scopatz I think is good to merge. I will think about the subtree stuff, but I think it can wait.

scopatz:
@melund woohoo! Thanks!

PR #824, opened 2016-05-07
set error on exit for foriegn shells

scopatz:
This should address #817.  Not sure what the value of this should be for `cmd.exe`.  @melund or @BYK do you have any idea?

### DIFF
--- a/docs/xonshconfig.rst
+++ b/docs/xonshconfig.rst
@@ -66,9 +66,14 @@ dictionaries have the following structure:
 :runcmd : *str or None, optional* - Command line switches to use when
     running the script, such as ``-c`` for Bash and ``/C`` for cmd.exe.
     ``default=null``
-:seterrcmd : *str or None, optional* - Command that enables exit-on-error
-    for the shell. For example, this is "set -e" in Bash. To disable
-    exit-on-error behavior, simply pass in an empty string. ``default=null``
+:seterrprevcmd : *str or None, optional* - Command that enables exit-on-error
+    for the shell before all other commands. For example, this is "set -e" in Bash.
+    To disable this exit-on-error behavior, simply pass in an empty string.
+    ``default=null``
+:seterrpostcmd : *str or None, optional* - Command that enables exit-on-error
+    for the shell after all other commands. For example, this is "set -e" in Bash.
+    To disable this exit-on-error behavior, simply pass in an empty string.
+    ``default=null``
 
 
 Some examples can be seen below:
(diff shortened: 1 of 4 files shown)


melund:
In cmd.exe you could add a `if errorlevel 1 exit 1`. Unfortunately, it has to be added after all the other commands :(   

This is why I added it as a post-cmd here: https://github.com/scopatz/xonsh/blob/master/xonsh/aliases.py#L270

scopatz:
Gotcha, fix coming in.

scopatz:
OK, updated! This is something I was considering doing anyway so thanks for keeping me honest @melund :)

docs/xonshconfig.rst
```diff
@@ -42,27 +42,39 @@ dictionaries have the following structure:
...
+:runcmd : *str or None, optional* - Command line switches to use when
+    running the script, such as ``-c`` for Bash and ``/C`` for cmd.exe.
+    ``default=null``
+:seterrprevcmd : *str or None, optional* - Command that enables exit-on-error
+    for the shell before all other commands. For example, this is "set -e" in Bash.
+    To disable this exit-on-error behavior, simply pass in an empty string.
+    ``default=null``
+:seterrpostcmd : *str or None, optional* - Command that enables exit-on-error
+    for the shell after all other commands. For example, this is "set -e" in Bash.
+    To disable this exit-on-error behavior, simply pass in an empty string.
```
melund: Does it make sence to use `set -e` for seterrpostcmd ? If not just remove the 'For example, ...'

xonsh/foreign_shells.py
```diff
@@ -80,62 +82,65 @@
...
-    '/bin/bash': '-c',
     'zsh': '-c',
-    '/bin/zsh': '-c',
-    '/usr/bin/zsh': '-c',
     'cmd': '/C',
 }
+DEFAULT_SETERRPREVCMD = {
+    'bash': 'set -e',
+    'zsh': 'set -e',
+    'cmd': '',
```
melund: On windows this could default to `@echo off`. If `echo off` is not set all the  that is not set all the `__XONSH_XXX_BEG__` tokens are repeated twice in the output

scopatz:
@melund thanks! I have updated this pr to reflect your suggestions.

melund:
Great. It works as expected on Windows. There are some merge conflicts, but otherwise it is good to merge

scopatz:
Ahh thanks.  I have fixed the conflict.

melund:
@scopatz. Sorry. I wasn't fast enough to merge. Yet another set of merge conflicts. I think you can just merge it yourself once they are fixed.

scopatz:
Sounds good, thanks @melund

PR #839, opened 2016-05-11
add warning if user has pt-specific options using rlshell

gforsyth:
If user has set any `prompt_toolkit` specific options (of those that are
False or empty strings by default) set in their `xonshrc`, a warning
will be printed if they open a `readline` shell.

Right now this triggers if any of the following are changed from their
defaults:
* `VI_MODE`
* `MOUSE_SUPPORT`
* `RIGHT_PROMPT`

Caveat: This might be _really_ annoying if you switch back and forth between the two.

```console
(xonshtest)gil@theo~/git/myxonsh master$ xonsh --shell-type readline
VI_MODE is set to True but has no effect in the readline shell.
MOUSE_SUPPORT is set to True but has no effect in the readline shell.
RIGHT_PROMPT is set to hi! but has no effect in the readline shell.
To make use of these options, install `prompt_toolkit`.  See https://xon.sh for more information.
```

### DIFF
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -23,6 +23,8 @@ RL_COMPLETION_SUPPRESS_APPEND = RL_LIB = None
 RL_CAN_RESIZE = False
 RL_DONE = None
 
+PT_ONLY_OPTIONS = ['VI_MODE', 'MOUSE_SUPPORT', 'RIGHT_PROMPT']
+
 
 def setup_readline():
     """Sets up the readline module and completion suppression, if available."""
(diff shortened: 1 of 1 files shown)


melund:
Hmm. Yes that is pretty annoying. Could we make a shorter warning like:
```The following options have no effect when using the readline shell: VI_MODE, MOUSE_SUPPORT. .:
To make use of these options, install `prompt_toolkit`.  See https://xon.sh for more information.
```
The last line could be differnt if prompt toolkit was already installed: 
```
To make use of these options, enable `prompt_toolkit` by ... 
```

gforsyth:
Hey @melund, I changed things to match your suggestions (which were spot on!).  
It might still be annoying.  What if there were a `$SUPPRESS_RL_WARNINGS` envvar?  Overkill?

scopatz:
Hi @gforsyth, first off, thanks for putting this in! I switch between shells all the time (I use `$SHELL_TYPE='random'`) and would find this annoying.  Maybe we should have an `$XONSH_VERBOSITY` or something that determines whether or not to print these kinds of messages.  I think this could come in useful elsewhere too.

melund:
At least the warning should not appear if `$SHELL_TYPE='random'`....

gforsyth:
This is too annoying.  I'll think about it some more and see if anything more workable comes to mind.

PR #859, opened 2016-05-12
Fix overriding of user environment with windows fixes

### LINKED ISSUE #857: Cannot set PROMPT in config file on Windows
Steps to repeat:
Be on Windows platform
Generate a config.json, e.g. through xonfig wizard. Set the PROMPT to be anything other than default.
Close and relaunch xonch

Expectation:
That the new prompt takes effect

Actual result:
Default prompt always shows.

This turns out to be the result of a little function called "windows_env_fixes()" in environ.py.
This is it:

```
def windows_env_fixes(ctx):
    """Environment fixes for Windows. Operates in-place."""
    # Windows default prompt doesn't work.
    ctx['PROMPT'] = DEFAULT_PROMPT
    # remove these bash variables which only cause problems.
    for ev in ['HOME', 'OLDPWD']:
        if ev in ctx:
            del ctx[ev]
    # Override path-related bash variables; on Windows bash uses
    # /c/Windows/System32 syntax instead of C:\\Windows\\System32
    # which messes up these environment variables for xonsh.
    for ev in ['PATH', 'TEMP', 'TMP']:
        if ev in os.environ:
            ctx[ev] = os.environ[ev]
        elif ev in ctx:
            del ctx[ev]
    ctx['PWD'] = _get_cwd()
```

This also explained why my 'HOME' environment variable vanished - this is a useful variable to make some Linux/Cygwin derived applications play ball (when running through normal command prompt not bash). I'm not quite sure what problem removing it fixes.

My suggestion:
Re-order default_env() so that windows_env_fixes() comes before load_static_config(). That way anything in config.json will take effect on the now fixed imported environment.

The rest of the changes all appear to relate to bash. Would it not be more appropriate to filter out those imports from load_foreign_env() (if on Windows), rather than unilaterally removing them from the inherited environment?

If I have some time later today I'll submit a pull request.

Will-W:
Fixes #857

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1284,8 +1284,8 @@ def default_env(env=None, config=None, login=True):
     ctx = dict(BASE_ENV)
     ctx.update(os.environ)
     if ON_WINDOWS:
-        # Windows default prompt doesn't work.
-        ctx['PROMPT'] = DEFAULT_PROMPT
+        # Windows style PROMPT definitions don't work in XONSH:
+        del ctx['PROMPT']
 
     if login:
         conf = load_static_config(ctx, config=config)
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1261,10 +1261,8 @@ def xonshrc_context(rcfiles=None, execer=None, ctx=None):
     return ctx
 
 
-def windows_env_fixes(ctx):
+def windows_foreign_env_fixes(ctx):
     """Environment fixes for Windows. Operates in-place."""
-    # Windows default prompt doesn't work.
-    ctx['PROMPT'] = DEFAULT_PROMPT
     # remove these bash variables which only cause problems.
     for ev in ['HOME', 'OLDPWD']:
         if ev in ctx:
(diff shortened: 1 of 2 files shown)


scopatz:
LGTM, but I think @melund or @BYK should handle it, since it is a windows thing. Thanks for putting this in @Will-W!

BYK:
LGTM. @melund?

xonsh/environ.py
```diff
@@ -1261,10 +1261,8 @@ def xonshrc_context(rcfiles=None, execer=None, ctx=None):
     return ctx
 
 
-def windows_env_fixes(ctx):
+def windows_foreign_env_fixes(ctx):
     """Environment fixes for Windows. Operates in-place."""
-    # Windows default prompt doesn't work.
-    ctx['PROMPT'] = DEFAULT_PROMPT
     # remove these bash variables which only cause problems.
     for ev in ['HOME', 'OLDPWD']:
```
melund: @Will-W Didn't you write that `HOME` was still usefull somehow?

xonsh/environ.py
```diff
@@ -1285,13 +1283,24 @@ def default_env(env=None, config=None, login=True):
     # in order of increasing precedence
     ctx = dict(BASE_ENV)
     ctx.update(os.environ)
+    if ON_WINDOWS:
+        # Windows default prompt doesn't work.
+        ctx['PROMPT'] = DEFAULT_PROMPT
```
melund: It should be enough to just delete it `del ctx['PROMPT']`

melund:
Otherwise, it looks good.

melund:
Nice to have some more people on board that use windows -  or are forced to ;)

melund:
Maybe also add an entry in under fixed section: https://github.com/scopatz/xonsh/blob/master/CHANGELOG.rst

xonsh/environ.py
Will-W: HOME is still useful - from Windows. The way I've tweaked it, HOME can still be inherited from the Windows environment, but on Windows it won't be pulled in from bash, because apparently that causes problems (which I can well believe).

melund: Ahh. I understand now ;)

Will-W:
> Nice to have some more people on board that use windows - or are forced to ;)

Yes - embedded development sadly pretty much mandates Windows. Have been looking for a shell with first class Windows support for a while. Happy to help polish it!

The two suggestions were both very sensible, and have pushed a couple of commits.

xonsh/environ.py
Will-W: There's an argument on Windows that you actually want foreign shell import to default to off. They may have a cygwin/bash or git/bash installation on the PATH that isn't much used and importing stuff from it may have some surprising results. It's different than on Linux where your bash setup is what you've probably been living with.

This is a whole separate discussion though, and I haven't been living with the foreign shell stuff enough to have a strong opinion. Just thought I'd note it as a 'first impressions of a Windows user'.

melund:
@Will-W .. Thanks for fixing this :)

PR #860, opened 2016-05-12
lower-case xonsh history string specs before parsing

ngoldbaum:
This makes it possible to have:

```
$XONSH_HISTORY_SIZE = '1 GB'
```

Right now this will error out because the history tuple parsing only knows about lower-case units names. This behavior is supposed to work, I think, since upper-case `GB` is documented here: http://xon.sh/envvars.html#xonsh-history-size


scopatz:
Thanks @ngoldbaum!

PR #861, opened 2016-05-12
fix #846 (compatibility with ptk 1.0)

### LINKED ISSUE #846: COMPLETIONS_MENU_ROWS not working
Hello

My value for `COMPLETIONS_MENU_ROWS` is 5 (default). But it does not seem to work. My terminal grow until the bottom of the window and when it has reached the last line then I don't get completions anymore.

I use gnome-terminal on Ubuntu 14.04

Thank you for your help !

EDIT: I installed HEAD from the git repo. I still have the problem

EDIT2+3: I also have the problem with xtrem and Terminator (It seems Terminal independant)

GuillaumeLeclerc:
I'm not sure how I could write unit tests for this.

But I tried it with single and multiple line inputs with ptk 0.57 and 1.0 (only the 1.0 was breaking before this fix).

### DIFF
--- a/xonsh/ptk/completer.py
+++ b/xonsh/ptk/completer.py
@@ -2,7 +2,7 @@
 """Completer implementation to use with prompt_toolkit."""
 import os
 import builtins
-import xonsh
+import xonsh.shell
 
 from prompt_toolkit.layout.dimension import LayoutDimension
 from prompt_toolkit.completion import Completer, Completion
--- a/xonsh/ptk/completer.py
+++ b/xonsh/ptk/completer.py
@@ -46,8 +46,13 @@ class PromptToolkitCompleter(Completer):
             #old layout to be removed at next ptk release
             window = cli.application.layout.children[1].children[1].content
         except AttributeError:
-            #new layout to become default
-            window = cli.application.layout.children[1].content
+            try:
+                # This is the layout for ptk 1.0
+                window = cli.application.layout.children[0].content.children[1]
+            except AttributeError:
+                #new layout to become default
+                window = cli.application.layout.children[1].content
+            
         if window and window.render_info:
             h = window.render_info.content_height
             r = builtins.__xonsh_env__.get('COMPLETIONS_MENU_ROWS')
@@ -60,3 +65,4 @@ class PromptToolkitCompleter(Completer):
                 else:
                     return LayoutDimension()
             window._height = comp_height
+
--- a/xonsh/ptk/completer.py
+++ b/xonsh/ptk/completer.py
@@ -2,6 +2,7 @@
 """Completer implementation to use with prompt_toolkit."""
 import os
 import builtins
+import prompt_toolkit
 
 from prompt_toolkit.layout.dimension import LayoutDimension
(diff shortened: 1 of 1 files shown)


xonsh/ptk/completer.py
```diff
@@ -46,8 +46,13 @@ def reserve_space(self):
             #old layout to be removed at next ptk release
             window = cli.application.layout.children[1].children[1].content
         except AttributeError:
-            #new layout to become default
-            window = cli.application.layout.children[1].content
+            try:
+                # This is the layout for ptk 1.0
+                window = cli.application.layout.children[0].content.children[1]
```
scopatz: Rather than doing a try-except, could you please do an `if ptk version >= 1` here? The reason for the other try except block is because PTK didn't used to report its version numbers :)

scopatz:
Thanks for putting this in @GuillaumeLeclerc

GuillaumeLeclerc:
I'll do it right now :) Glad I helped. With the crazy amount of issues I reported today it was the least I could do.

GuillaumeLeclerc:
Is it good now ?

xonsh/ptk/completer.py
```diff
@@ -48,6 +49,11 @@ def reserve_space(self):
         except AttributeError:
             #new layout to become default
             window = cli.application.layout.children[1].content
+
+        # This is the layout for ptk 1.0
+        if prompt_toolkit.__version__.startswith("1."):
```
scopatz: You should probably use the `xonsh.shell.promt_tookit_version()` function https://github.com/scopatz/xonsh/blob/master/xonsh/shell.py#L31 as a safety. Also maybe you could put the above tr-except in an else block  after the if here.  Just to make the logic clean.

scopatz:
> I'll do it right now :) Glad I helped. With the crazy amount of issues I reported today it was the least I could do.

Thanks! It is really appreciated!  Sorry about the one other little edit I forgot to mention originally :)

GuillaumeLeclerc:
If we do it in a if then we should check if the user is using ptk or not right ? 

If yes how can I check this ?

scopatz:
Nope, we don't need to check if the user is using PTK or not.  By this point, we know that they are using it, but we just need to check the version.

GuillaumeLeclerc:
is it fine ?

xonsh/ptk/completer.py
```diff
@@ -2,6 +2,7 @@
 """Completer implementation to use with prompt_toolkit."""
 import os
 import builtins
+import xonsh
```
scopatz: Doesn't this need to be `import xonsh.shell`?

GuillaumeLeclerc: It's working this way but maybe it's a bad practice so I'll change to `xonsh.shell` sorry for waisting your time

scopatz: No worries, not wasting anyone's time.  We are always happy to have more xonsh developers :)

scopatz:
Looks great to me, thanks @GuillaumeLeclerc!

PR #863, opened 2016-05-12
An commit `which` fixes #833

### LINKED ISSUE #833: `which` takes not `-v` on OS X
Only, `-a` or `s`.  Though `which_version` (the `xonsh/alias.py` function) does invoke `which` with `-v`

This is untrue if you `brew install gnu-which --with-default-names`, which I guess not many people do... and we **could** determine that by looking as wether the `gnu`string is in the output of `which` without any arguments...

Woosh, with wish which was not [a witch](https://www.youtube.com/watch?v=zrzMhU_4m-g).

Carreau:
CROWD:  A `which`!  A `which`!  A `which`!  We've got a `which`!  A `which`!
GFORSYTH:  We have found a `which`, might we burn it?
CROWD:  Burn it!  Burn!
SCOPATZ:  How do you know it is a `which`?
CARREAU:  It looks like one.
SCOPATZ:  Bring it forward.
WITCH:  I'm not a `gnu-which`.  I'm not a `gnu-which`.
SCOPATZ:  But you have a `-v` flag.
WITCH:  They monkey patched me.
CROWD:  No, we didn't... no.
WITCH:  And this isn't my source code, it's a false one.
SCOPATZ:  Well?
GFORSYTH:  Well, we did do the `-v` flag.
SCOPATZ:  The source code?
GFORSYTH:  And the function signature -- but it is a `which`!
CROWD:  Burn it!  `which`!  `which`!  Burn it!
SCOPATZ:  Did you dress it up like this?
CROWD:  No, no... no ... yes.  Yes, yes, a bit, a bit.

---- 
Note, above text is the commit text, not only the issue one.

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -328,6 +328,10 @@ def bang_bang(args, stdin=None):
 
 def which_version():
     """Returns output from system `which -v`"""
+    if sys.platform == 'darwin':
+        if ('gnu' not in subprocess.run(['which'], stderr=subprocess.PIPE).stderr.decode('utf-8')):
+            return '<no version number available on OS X>'
+
     _ver = subprocess.run(['which','-v'], stdout=subprocess.PIPE)
     return(_ver.stdout.decode('utf-8'))
 


Carreau:
Note II: Did consider having the version number be `<Ni!>`, or `<Blue ! No... Red ! AHHHhhhhh>`.

scopatz:
SCOPATZ: What makes you think it is a `which`?
GFORSYTH: Oh, it turned me into a GNUwt!
SCOPATZ: A GNUwt?
[Silence]
GFORSYTH: Well, I got better.
CROWD: Burn it anyway!

Issue #853, opened 2016-05-12
Xonsh hang in some weird cases


GuillaumeLeclerc:
Hello,

In some cases xonsh just stop responding. I have been facing it in my Ubuntu and Arch Linux setup.

But not on a Mac OSx.

I can reproduce it by doing this (I know it does not really make sense but it's only way I found to reproduce the bug consistently):

```
cat $(which xonch)
which xonch
```

After this command is executed nothing is executed, xonsh seems ignoring my inputs

GuillaumeLeclerc:
I happened to me again just when I was typing `cd ..`
I don't get what's causing this

scopatz:
Do you think that this is caused by the branch checking issue?

GuillaumeLeclerc:
@scopatz I don't know about this issue sorry

GuillaumeLeclerc:
Are we using `communicate` when communicating with subprocesses ? 

Because I saw this on the python docs:

```
Use communicate() rather than .stdin.write, .stdout.read or .stderr.read to avoid deadlocks due to any of the other OS pipe buffers filling up and blocking the child process.
```

Could it come from this ?

melund:
This was/is actually a problem, but one that may soon be fixed https://github.com/scopatz/xonsh/pull/875

GuillaumeLeclerc:
since #875 was merge I will assume it's fixed and reopen if I ever see it again

GuillaumeLeclerc:
It just happened to me while doing a `cd ..` again

melund:
Are you using prompt_toolkit?

GuillaumeLeclerc:
yes

adqm:
I get these random hangs occasionally, too, and I only use the readline shell.

That said, I can't reproduce with the example from the original post.  AFAICT, no noticeable pattern as to when the hanging happens.

I might put a bunch of print statements throughout my local install to see if I can catch what is going on when this happens, and where xonsh is getting stuck...

scopatz:
For what it is worth, I have yet to see this.

GuillaumeLeclerc:
The frequency of these hangs has significantly decreased since I started being up to date with master. it went from 2 times an hour to one time every 3 days or so (which is way less annoying)

laerus:
Hey @GuillaumeLeclerc do you still have this issue with the latest master?

laerus:
this seems fixed on current master, feel free to reopen if the problem still persists!

[...]

PR #867, opened 2016-05-12
fix #855

### LINKED ISSUE #855: Ctrl-Z only working once
Hello,

This workflow make xonsh hang

```
$ vim
Ctrl-z
$fg
Ctrl-z
```

the first Ctrl-z is working (I go back to the shell). But after the second one vim is paused but what I type is ignored

This might be related to #853 since the final state is the same (I type in the terminal and only display what I type)

GuillaumeLeclerc:
This patch fix #855. But there are some other bugs I found with background tasks. 

I'll make some more issues about them. But at least this one is fixed, I really needed it to work properly

I hope my python style is improving. I never used it in a large scale project

### DIFF
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -123,7 +123,6 @@ else:
         if signal_to_send is not None:
             if signal_to_send == signal.SIGCONT:
                 job['status'] = 'running'
-
             os.kill(obj.pid, signal_to_send)
             job['signal'] = None
 
@@ -135,6 +134,8 @@ else:
         if os.WIFSTOPPED(wcode):
             job['bg'] = True
             job['status'] = 'stopped'
+            # The task was stopped therefore we don't have any active task
+            builtins.__xonsh_active_job__ = None
             print()  # get a newline because ^Z will have been printed
             print_one_job(act)
         elif os.WIFSIGNALED(wcode):
@@ -161,16 +162,17 @@ def _clear_dead_jobs():
         del builtins.__xonsh_all_jobs__[i]
         if builtins.__xonsh_active_job__ == i:
             builtins.__xonsh_active_job__ = None
-    if builtins.__xonsh_active_job__ is None:
-        _reactivate_job()
 
 
 def _reactivate_job():
     if len(builtins.__xonsh_all_jobs__) == 0:
         return
-    builtins.__xonsh_active_job__ = max(builtins.__xonsh_all_jobs__.items(),
-                                        key=lambda x: x[1]['started'])[0]
 
+    all_jobs = builtins.__xonsh_all_jobs__.items()
+    reactivable_jobs = [job for job in all_jobs if job[1]['status'] == 'stopped']
+
+    builtins.__xonsh_active_job__ = max(reactivable_jobs,
+                                        key=lambda x: x[1]['started'])[0]
 
 
 def print_one_job(num):
@@ -247,6 +249,7 @@ def fg(args, stdin=None):
     _clear_dead_jobs()
     if len(args) == 0:
         # start active job in foreground
+        _reactivate_job()
         act = builtins.__xonsh_active_job__
         if act is None:
             return '', 'Cannot bring nonexistent job to foreground.\n'
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -137,6 +137,8 @@ else:
         # give the terminal over to the fg process
         _give_terminal_to(pgrp)
 
+        _continue(active_task)
+
         _, wcode = os.waitpid(obj.pid, os.WUNTRACED)
         if os.WIFSTOPPED(wcode):
             print()  # get a newline because ^Z will have been printed
@@ -258,7 +260,6 @@ def fg(args, stdin=None):
     job = get_task(act)
     job['bg'] = False
     job['status'] = RUNNING
-    _continue(job)
     print_one_job(act)
 
 
@@ -271,6 +272,8 @@ def bg(args, stdin=None):
     """
     res = fg(args, stdin)
     if res == None:
-        get_task(tasks[0])['bg'] = True
+        curTask = get_task(tasks[0])
+        curTask['bg'] = True
+        _continue(curTask)
     else:
         return res
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -7,9 +7,14 @@ import signal
 import builtins
 from subprocess import TimeoutExpired
 from io import BytesIO
+from collections import deque
 
 from xonsh.tools import ON_WINDOWS
 
+RUNNING = 1
+STOPPED = 0
+
+tasks = deque()
 
 if ON_WINDOWS:
     def _continue(job):
@@ -55,8 +60,7 @@ if ON_WINDOWS:
 
 else:
     def _continue(job):
-        job['signal'] = signal.SIGCONT
-        job['status'] = 'running'
+        os.kill(job['obj'].pid, signal.SIGCONT)
 
 
     def _kill(obj):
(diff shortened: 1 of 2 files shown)


scopatz:
This looks reasonable to me.  @adqm, do you mind taking a look?  Also, can anyone think up a good way to test this kind of thing?  The more I think about it, the more I believe that we need to have these kinds of top-level integration tests.

GuillaumeLeclerc:
Maybe we could start a xonsh process and write some commands to it. what I fear is that it will not be considered as a tty and will behave strangely

scopatz:
Not saying that tests are blocking here, because it fixes the problem.

scopatz:
Well, that would be one way... Python does have a PTY interface that we could use: https://docs.python.org/3/library/pty.html#module-pty

GuillaumeLeclerc:
`bg` is not working anymore. Does anyone know if it's handled differently than `fg` ?

GuillaumeLeclerc:
`jobs` is still buggy (i.e. never stops)

adqm:
I can't take a look right now, but I should be able to later tonight.

It looks like job control got seriously messed up at some point in the past.  I'm going to assume/hope it was during one of my breaks, and not in the original code I put in :)

GuillaumeLeclerc:
I fixed `jobs`. I also saw there is a difference in the behavior with other shells. `fg` always pop out the latest started task. In other shells it is usually the latest started OR resumed task.

If we want to implement this we will need one more field where we store the last time we resumed a task. Or use a FILO queue.

GuillaumeLeclerc:
Status:

#### `Ctrl-z`

Works on commands, but not on python code

#### `fg` 

works on some processes:
- `vim`
- `node`
- `sleep`

but fails on others:
- `sh -c "sleep 5"`

when you fg with no task number then the behavior is not the same as other shells

#### `bg`

Not working at all. I haven't found out why yet

###### `jobs`

Seems fine now

adqm:
Awesome, thanks!  I just ran this though a couple of paces.  Looks like the fix for #855 is working great!

`bg` and `fg` are still a little bit messed up when they aren't given args.  Sorry for the long wall of text, but here are some of the tests I ran:
```
hartz@thinky-515 ~ $ jobs
 [1] stopped: vim s.cpp & (25361)
hartz@thinky-515 ~ $ fg
*[1] running: vim s.cpp (25361)

 [1] stopped: vim s.cpp & (25361)
hartz@thinky-515 ~ $ gitk &
^Z
 [2] stopped: gitk & (25375)
hartz@thinky-515 ~ $ jobs
 [1] stopped: vim s.cpp & (25361)
 [2] stopped: gitk & (25375)
hartz@thinky-515 ~ $ fg 1
*[1] running: vim s.cpp (25361)
hartz@thinky-515 ~ $ jobs
 [2] stopped: gitk & (25375)
hartz@thinky-515 ~ $ bg
Cannot send nonexistent job to background.
hartz@thinky-515 ~ $ jobs
 [2] stopped: gitk & (25375)
hartz@thinky-515 ~ $ bg 2
*[2] running: gitk & (25375)
hartz@thinky-515 ~ $ jobs
*[2] running: gitk & (25375)
hartz@thinky-515 ~ $ fg
ValueError: max() arg is an empty sequence
```

I am happy to go in and make those changes, but if you want to since you're in the thick of it, that would be great, too!  I would vote for making all these related changes part of this same PR.

If you don't mind, we should also add this to the CHANGELOG as part of this PR (other docs shouldn't need updating since this is all just bugfixes).

adqm:
> bg
> Not working at all. I haven't found out why yet

I had some success with `bg` just now.  Are there specific commands you've been trying that aren't working?

GuillaumeLeclerc:
I'm going to sleep now, and I have a lot to do tomorrow so I'm not sure if I will have time. If you do then of course you can do it.

I think we may want to consider redoing this whole thing in a clean way with queues (It's a common practice in schedulers/job managers) and using the obj's to send the signal instead of bypassing it with raw OS calls. It might interfere at some point and causing #853. 

I think this is the only part left to make this shell robust :)

GuillaumeLeclerc:
Would anyone have the courage to review a potential new job manager ? (I don't guarantee I will do it but if it won't be reviewed then I don't need to motivate myself)

GuillaumeLeclerc:
Hey,

I did a massive refactor of the job manager.

It's working great from my point of view. Could you test it. `bg` should be working :) 

The PR is not done yet for two reasons:

- I'm not filling `__xonsh_active_job__` because I don't know what it should be, and what is it for
- I don't know windows so it is probably broken. Maybe someone could help me on this

scopatz:
> Would anyone have the courage to review a potential new job manager ? (I don't guarantee I will do it but if it won't be reviewed then I don't need to motivate myself)

@GuillaumeLeclerc, in general, don't be worried about this; if you put in a PR, it will be reviewed, if not by any of the other great people with push rights, than I will get to it.  I prioritize looking at PRs more than other issues.

scopatz:
@GuillaumeLeclerc - just skimming this very briefly, I think that this could really use some tests. We have been having a lot of problems with the job management stuff lately, and I think that part of the reason for that was that we have been fixing things as they go - but some of these fixes have ended up breaking other things.  

What we need is a good way to establish that the capabilities that we want are going to be maintained or improved by any changes that are made. I think this will also help the windows folks a lot (and maybe we should re-enable Appveyor?) because the test suite will tell us where to look if something breaks.

All that said, I *really* appreciate you taking this on.  The jobs management definitely needs a refactor to get everything working as expected. This refactor has been coupled with better testing in my mind, though.  Let us know if you have any questions about what this would entail.

Thanks again!

GuillaumeLeclerc:
I'll try to write tests but I'll probably not have the time to do it today. 

I'll take a look at fish's repository and learn how they tested this part (and if they did so)

adqm:
At a first glance, this looks really nice!  I'm happy to go through it with a fine-toothed comb when I have a chance, but that probably won't be until this coming weekend.

xonsh/jobs.py
```diff
@@ -7,13 +7,18 @@
 import builtins
 from subprocess import TimeoutExpired
 from io import BytesIO
+from collections import deque
 
 from xonsh.tools import ON_WINDOWS
 
+RUNNING = 1
+STOPPED = 0
```
scopatz: This is basically an enum.

xonsh/jobs.py
```diff
@@ -97,43 +102,47 @@ def _give_terminal_to(pgid):
...
-                return
+        _continue(active_task)
 
         _, wcode = os.waitpid(obj.pid, os.WUNTRACED)
         if os.WIFSTOPPED(wcode):
-            job['bg'] = True
-            job['status'] = 'stopped'
             print()  # get a newline because ^Z will have been printed
-            print_one_job(act)
+            active_task['status'] = STOPPED
```
scopatz: I think the string flags are more readable than integer flags.

scopatz: Unless there is a reason for the integer ones I am not seeing...

scopatz:
If @adqm is happy I am happy :).  Thanks for offering to put in some tests!

adqm:
> I'm not filling \__xonsh_active_job__ because I don't know what it should be, and what is it for

It is very likely that `__xonsh_active_job__` can be completely replaced with `_get_next_task`.

> I don't know windows so it is probably broken. Maybe someone could help me on this

I can run some Windows tests this weekend.  Windows is kind of iffy with job control anyway (given that there isn't a clean way to suspend a process), but we can at least replicate some minimal piece of functionality.  Eventually, we'll want to dig deeper and figure out how to actually emulate job control there in full.

> I'm filling the start time for tasks but not end. Should it be done here ?

Personally, I think this is fine as is.

xonsh/jobs.py
GuillaumeLeclerc: It was only thinking about performance but I guess it's not relevant. I will change it back

scopatz: Yeah, I think the performance hit is minimal enough for the gains in user readability.

GuillaumeLeclerc:
__xonsh_active_jobs__ seems to be used outside `jobs.py` How is it used ?

GuillaumeLeclerc:
fg and bg are not working with some processes, eg: eclipse. Maybe because we are sending the signal to a process instead of a process group.

What do you think ?

scopatz:
Hi @GuillaumeLeclerc 

> xonsh_active_jobs seems to be used outside jobs.py How is it used ?

`__xonsh_active_jobs__` is used to set $PROMPT and $TITLE strings, which are pretty important, so it is probably not the best idea to remove all references to them.  See  https://github.com/scopatz/xonsh/blob/master/xonsh/environ.py#L1011

> fg and bg are not working with some processes, eg: eclipse. Maybe because we are sending the signal to a process instead of a process group.

Well they weren't working at all before, so this is probably an improvement :). I do think the active job thing needs to be cleaned up, though.

adqm:
> \__xonsh_active_jobs__ is used to set $PROMPT and $TITLE strings, which are pretty important, so it is probably not the best idea to remove all references to them

That said, I think it should be possible to replace that with a call to `_get_next_task`, so long as we ensure that that task is actually running and not stopped.

I think this is definitely a big-time improvement already.  I'm happy to help out with this, but I'm pretty much booked until the weekend.

scopatz:
> That said, I think it should be possible to replace that with a call to _get_next_task, so long as we ensure that that task is actually running and not stopped

I think that is better than `__xonsh_active_job__`. It is better to have an API call where we can, rather than using builtins.

GuillaumeLeclerc:
The problem is the only time a job is active is when it's running. Therefore the shell does not have the terminal at this moment and we are not able to ask for the active job. As soon as the shell get the terminal back that means there are no more active tasks and it will always read None as active task.

Other thing (bad news). I look at the `fish-shell` tests and they is not a single test containing any of these built-ins:

- `fg`
- `bg`
- `jobs`

I'll try do investigate on other shells to check how they do this

adqm:
Maybe we should just rename that variable then, but still keep it around?  We use it to set the title of the terminal, which I find nice, but we don't have to actually use it for anything more meaningful than that.  I'm happy to help work out the details if need be.

Testing these things (any interactive command, really) is going to be hard.  I know it would be desirable, but I'm not sure it's feasible.

GuillaumeLeclerc:
I'll try to fix the title then.

Tcsh does not test this either. The only one I found who has real tests about job control is bash. It seems they are just running scripts and comparing the output with a reference.

GuillaumeLeclerc:
I found out the problem with job control and complex tasks !!! :D 

We were sending signals to the the main process only, not the process and all its children. that's why we were not able to return to complex processes (firefox, ecplise, `sh some_script.sh`). This was because their children were still in the interrupted state.

So now we are sending the required signals to the whole process group and it working perfectly fine. I tested bash scripts and the Scala IDE.

adqm:
Ah, yes; that makes good sense!  Is this ready for a careful look, then?  If so, I'll go through and do a thorough review tonight.

Very glad to have this; thanks!

GuillaumeLeclerc:
Not yet, I still have to fix the "start in background" feature. It's not working for me at the moment

GuillaumeLeclerc:
I dug into this problem this evening and it is a problem with the parser. `built_in.py` receive the commands without their trailing `&`. I tried looking at the parser but I could not understand anything. I was expecting a formal grammar or something like this but I got completely lost.

I think the fix of the parser should not be part of this PR in my opinion.

The only thing left is reviewing the whole PR and someone to take care of the windows part. I don't have a single machine running it so It will be complicated for me. If nobody can do it then I'll try to get the motivation and install a virtual machine.

adqm:
I should be able to look into those things tonight or tomorrow.

@scopatz, is there a hard deadline for the release?

scopatz:
Sorry I missed this @adqm. I anticipate that we might need to do a v0.3.1 release next week, just to clean things up a bit.

adqm:
Ah, no worries.  I'll still try to test this thoroughly tonight.

scopatz:
Thanks!  I think we are almost certainly going to want a v0.3.1 release now.  That 'best' thing is pretty annoying.

xonsh/jobs.py
```diff
@@ -7,13 +7,18 @@
...
+RUNNING = 1
+STOPPED = 0
+
+tasks = deque()
 
 if ON_WINDOWS:
-    def _continue(obj):
-        return None
+    def _continue(job):
+        job['status'] = 'running'
```
adqm: should this be `RUNNING`, for the sake of consistency with the other pieces?

adqm:
I think this all looks great, @GuillaumeLeclerc!  Thanks again for putting it in.

I just put in #962, which builds on this and (hopefully) resolves some of the remaining issues.  Hoping to get this and those new changes merged in at the same time.

scopatz:
Hi @GuillaumeLeclerc - I believe that these changes have been included by @adqm  in #962 (though he rebased so this didn't autoclose.) Please feel free to close this.

scopatz:
Thanks again for putting it in.

Issue #868, opened 2016-05-12
which failing all the time


GuillaumeLeclerc:
Any idea why I would get this ? 

```
$ which
AttributeError: 'module' object has no attribute 'run'
```

scopatz:
This is because the subprocess.run() function wasn't added until Python 3.5 and you are probably on v3.4. We should update the which alias to use `subprocess.check_output()` or find a way to backport run()

scopatz:
Thanks to @gforsyth this should be fixed on master.  Would you mind testing this out please @GuillaumeLeclerc and then closing this issue? Thanks!

gforsyth:
@scopatz @Carreau can either of you tell me which `which` command you ran that threw the error?

scopatz:
`which -v`, `which ls`, etc. I get the following:

```console
scopatz@localhost ~ $ which -v
Illegal option -v
Traceback (most recent call last):
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/proc.py", line 335, in wrapped_simple_command
    r = f(args, i)
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/aliases.py", line 369, in which
    version='{}'.format(which_version()))
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/aliases.py", line 336, in which_version
    _ver = subprocess.check_output(['which','-v'])
  File "/home/scopatz/miniconda/lib/python3.5/subprocess.py", line 629, in check_output
    **kwargs).stdout
  File "/home/scopatz/miniconda/lib/python3.5/subprocess.py", line 711, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['which', '-v']' returned non-zero exit status 2
scopatz@localhost ~ $ which ls
Illegal option -v
Traceback (most recent call last):
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/proc.py", line 335, in wrapped_simple_command
    r = f(args, i)
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/aliases.py", line 369, in which
    version='{}'.format(which_version()))
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/aliases.py", line 336, in which_version
    _ver = subprocess.check_output(['which','-v'])
  File "/home/scopatz/miniconda/lib/python3.5/subprocess.py", line 629, in check_output
    **kwargs).stdout
[...]

PR #869, opened 2016-05-12
update which to be compatible with python3.4

gforsyth:
and one other, small change

this should address #868

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -6,7 +6,7 @@ import shlex
 import builtins
 import sys
 import subprocess
-from argparse import ArgumentParser
+from argparse import ArgumentParser, Action
 from collections.abc import MutableMapping, Iterable, Sequence
 
 from xonsh.dirstack import cd, pushd, popd, dirs, _get_cwd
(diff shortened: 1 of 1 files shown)


xonsh/aliases.py
```diff
@@ -329,11 +329,25 @@ def bang_bang(args, stdin=None):
...
+                                                 stderr=subprocess.PIPE)):
             return '<no version number available on OS X>'
 
-    _ver = subprocess.run(['which','-v'], stdout=subprocess.PIPE)
-    return(_ver.stdout.decode('utf-8'))
+    _ver = subprocess.check_output(['which','-v'])
+    return _ver.decode('UTF-8')
+
+
+class AWitchAWitch(Action):
```
scopatz: <3

xonsh/aliases.py
```diff
@@ -361,16 +376,20 @@ def which(args, stdin=None):
...
         if pargs.all:
             try:
-                subprocess.run(['which'] + args,
-                               stderr=subprocess.PIPE,
-                               check=True)
+                subprocess.check_call(['which'] + args,
+                                      stdout=subprocess.PIPE,
+                                      stderr=subprocess.PIPE)
+                subprocess.call(['which'] + args,
+                               stderr=subprocess.PIPE)
```
scopatz: I don't grok why you have these two calls.  Would you mind adding a comment explaining it? Thanks!

gforsyth: The reason you don't grok it is because they're redundant. Fix incoming.

gforsyth:
Ready for another look @scopatz

scopatz:
Thanks a ton @gforsyth!

PR #871, opened 2016-05-12
add potential fix for which ails

gforsyth:
a fix for those things `which` ails you?  I hope?  Maybe?

ping @scopatz @Carreau

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -328,13 +328,14 @@ def bang_bang(args, stdin=None):
 
 def which_version():
     """Returns output from system `which -v`"""
-    if sys.platform == 'darwin':
-        if ('gnu' not in subprocess.check_output(['which'],
-                                                 stderr=subprocess.PIPE)):
-            return '<no version number available on OS X>'
+    try:
+        subprocess.check_output(['which', '-v'],
+                                stderr=subprocess.PIPE).decode('UTF-8')
+    except subprocess.CalledProcessError:
+        return '<no version number available on your OS>'
 
-    _ver = subprocess.check_output(['which','-v'])
-    return _ver.decode('UTF-8')
+    _ver = subprocess.check_output(['which','-v']).decode('UTF-8')
+    return _ver
 
 
 class AWitchAWitch(Action):


xonsh/aliases.py
```diff
@@ -329,12 +329,15 @@ def bang_bang(args, stdin=None):
 def which_version():
     """Returns output from system `which -v`"""
     if sys.platform == 'darwin':
-        if ('gnu' not in subprocess.check_output(['which'],
-                                                 stderr=subprocess.PIPE)):
+        try:
+            subprocess.check_output(['which', '-v'],
+                                    stdout=subprocess.PIPE,
+                                    stderr=subprocess.PIPE).decode('UTF-8')
+        except subprocess.CalledProcessError
```
Carreau: end with `:`

Carreau: ValueError: stdout argument not allowed, it will be overridden.

gforsyth:
Thanks @Carreau -- I'm not at my finest at the moment.  

Does this work for you?  I'd prefer your solution but we'd have to give up Python 3.4 support.

xonsh/aliases.py
```diff
@@ -329,12 +329,14 @@ def bang_bang(args, stdin=None):
 def which_version():
     """Returns output from system `which -v`"""
     if sys.platform == 'darwin':
-        if ('gnu' not in subprocess.check_output(['which'],
-                                                 stderr=subprocess.PIPE)):
+        try:
```
scopatz: I think that this try-except block should be bumped up to the top level. I am on linux (Ubuntu) but I also don't have a `-v` option:

```console
scopatz@artemis ~ $ /usr/bin/which --version
Illegal option --
Usage: /usr/bin/which [-a] args
```

Carreau:
> I'd prefer your solution but we'd have to give up Python 3.4 support.

Ah... sorry, 

> I'd prefer your solution but we'd have to give up Python 3.4 support.

Why ? Subprocess run was already used before my patch.

scopatz:
Also note that I think you can pass `stderr=subprocess.STDOUT` to pipe everything to stdout.

xonsh/aliases.py
gforsyth: THE WHOLE WORLD IS UPSIDE-DOWN.  On it.

Carreau:
> Also note that I think you can pass stderr=subprocess.STDOUT to pipe everything to stdout.

`which` seem too look like the dark knight.

xonsh/aliases.py
```diff
@@ -329,12 +329,14 @@ def bang_bang(args, stdin=None):
...
+            subprocess.check_output(['which', '-v'],
+                                    stderr=subprocess.PIPE).decode('UTF-8')
+        except subprocess.CalledProcessError:
             return '<no version number available on OS X>'
```
Carreau: remove `on OS X`

Carreau:
Otherwise works for me.

gforsyth:
I may have cut off both of `which` arms and legs now?

Carreau:
> I may have cut off both of which arms and legs now?

I guess so.

scopatz:
Thanks @gforsyth!

PR #872, opened 2016-05-12
'fix which'

Carreau:
@gforsyth @scopatz , check_output can only check on stdout, one need to check on stderr.


scopatz:
Feel free to close this @Carreau

PR #875, opened 2016-05-13
Fix For Deadlock On Captured stdout/stderr

adqm:
It turns out the issue described in #798 is not specific to Windows.  If the PIPE used by a captured subproc gets filled, everything blocks.

This implements a general fix, using a `NamedTemporaryFile` instead of a `PIPE` as the process's `stdout`.  It is likely to be slow as it uses a file on disk, but nothing else I tried got around this quite so nicely, and the speed wasn't a noticeable issue for me.

I ran it through some quick tests that had been locking things up both on GNU/Linux and on Windows, and it seems to pass those cases now.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -428,7 +428,9 @@ def run_subproc(cmds, captured=False):
                 raise XonshError('Multiple redirects for stdout')
             stdout = streams['stdout'][-1]
             procinfo['stdout_redirect'] = streams['stdout'][:-1]
-        elif _capture_streams or ix != last_cmd:
+        elif ix != last_cmd:
+            stdout = PIPE
+        elif _capture_streams:
             _nstdout = stdout = tempfile.NamedTemporaryFile(delete=False)
             _stdout_name = stdout.name
         elif builtins.__xonsh_stdout_uncaptured__ is not None:
@@ -439,7 +441,7 @@ def run_subproc(cmds, captured=False):
         if 'stderr' in streams:
             stderr = streams['stderr'][-1]
             procinfo['stderr_redirect'] = streams['stderr'][:-1]
-        elif captured == 'object':
+        elif captured == 'object' and ix != last_cmd:
             _nstderr = stderr = tempfile.NamedTemporaryFile(delete=False)
             _stderr_name = stderr.name
         elif builtins.__xonsh_stderr_uncaptured__ is not None:
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -429,7 +429,7 @@ def run_subproc(cmds, captured=False):
             stdout = streams['stdout'][-1]
             procinfo['stdout_redirect'] = streams['stdout'][:-1]
         elif _capture_streams or ix != last_cmd:
-            stdout = tempfile.NamedTemporaryFile(delete=False)
+            _nstdout = stdout = tempfile.NamedTemporaryFile(delete=False)
             _stdout_name = stdout.name
         elif builtins.__xonsh_stdout_uncaptured__ is not None:
             stdout = builtins.__xonsh_stdout_uncaptured__
@@ -440,7 +440,7 @@ def run_subproc(cmds, captured=False):
             stderr = streams['stderr'][-1]
             procinfo['stderr_redirect'] = streams['stderr'][:-1]
         elif captured == 'object':
-            stderr = tempfile.NamedTemporaryFile(delete=False)
+            _nstderr = stderr = tempfile.NamedTemporaryFile(delete=False)
             _stderr_name = stderr.name
         elif builtins.__xonsh_stderr_uncaptured__ is not None:
             stderr = builtins.__xonsh_stderr_uncaptured__
@@ -557,9 +557,12 @@ def run_subproc(cmds, captured=False):
     output = b''
     if write_target is None:
         if _stdout_name is not None:
-            stdout = open(_stdout_name, 'rb')
-            output = stdout.read()
-            stdout.close()
+            with open(_stdout_name, 'rb') as stdoutfile:
+                output = stdoutfile.read()
+            try:
+                _nstdout.close()
+            except:
+                pass
             os.unlink(_stdout_name)
         elif prev_proc.stdout not in (None, sys.stdout):
             output = prev_proc.stdout.read()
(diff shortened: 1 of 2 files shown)


adqm:
@scopatz, I'm confused as to why those test cases are failing.  Everything seems to work fine for me under normal operations.  Maybe something with the `stdout` replacement in the test cases?

adqm:
Alright, I think this is ready for a look.

scopatz:
@adqm - what were some of the tests that you used?

scopatz:
Also thanks for putting this in!

adqm:
I often do things like the following for working with TSV files:
`[i.strip().split('\t') for i in $(cat some_big_file).splitlines()]`

But that wasn't working recently.

It seems like just `cat`ing any sufficiently-large file would cause xonsh to hang if the output was being captured.  I can't share the files themselves since they contain sensitive information, but one file that caused the problem (on all my machines) was a little over 100 kB.  Things would just freeze up.

Interestingly, `^C` still worked to kill the process, and I would get partial output after having done that.  I imagine that some buffer was being filled up.  Using a file instead, we have a buffer whose size is only limited by free space on disk.

I haven't tested with `$XONSH_STORE_STDOUT` and the `TeePTY` stuff, but everything else seemed to be working fine (and it now passes all the test cases without cheating just to make that happen).

scopatz:
Maybe we could have a big `cat`  test :).  Good to know that this is easily reproducible. I think my main concern with this, which you mention in the original post, is that it could thrash the file system unnecessarily.  For those of us with SSDs it probably isn't a huge deal, but I have heard of people running xonsh on a raspberrypi, where this could kill things.  Maybe that is not a huge concern...

I am going to play around with it, but it seems like this happens for all procs that are left of a pipe `|`.  Is that correct? I don't think going to a temp file if a big deal for captured processes, per say. 

Also, does having an in-memory structure like `io.StringIO` not work for some reason?

To be clear, I think having something working is better than having it broken and on those grounds, it is certainly mergable.

scopatz:
A couple of things:

1. temp files on the left of the pipe aren't cleaned up, eg `echo yo | grep y` will create a file in `/tmp` (on linux) that contains yo.

2. Furthermore, `echo yo | grep y` doesn't work, since it seems like None is getting passed in to grep as stdin:

```console
scopatz@artemis ~ $ echo yo | grep y    
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/base_shell.py", line 154, in default
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/codecache.py", line 55, in run_compiled_code
    func(code, glb, loc)
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/built_ins.py", line 642, in subproc_captured_hiddenobject
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/built_ins.py", line 535, in run_subproc
    proc.stdout.close()
AttributeError: 'NoneType' object has no attribute 'close'
```

GuillaumeLeclerc:
For those using a ssd it is a huge deal. Ssd's have a limited amount of byte they can write.

It would be way better to use streams instead of memorizing the whole content. I mean an not able to handle infinite streams (/dev/urandom) with xonsh for that particular reason.

adqm:
>A couple of things:
>    temp files on the left of the pipe aren't cleaned up, eg echo yo | grep y will create a file in /tmp (on linux) that contains yo.
>    Furthermore, echo yo | grep y doesn't work, since it seems like None is getting passed in to grep as stdin:

Yes, those were both results of a boneheaded mistake.  The intention was still to use `PIPE` for pipes in the pipeline (because those are generally handled in a streaming fashion, they should be more safe from overruns, though the deadlock will still be a possibility), and use the temp file only for the end when storing stdout/stderr.  I can go in and clean that up.

> For those using a ssd it is a huge deal. Ssd's have a limited amount of byte they can write.

I don't know that I agree with "huge" deal.  I use SSD's in all my machines, and I wouldn't be worried about this.  While it is true that SSD's have a limited number of write actions per sector, the mean time to failure on a modern SSD is very reasonable.

Plus, again, the intention isn't to use to replace _every pipe_, but rather just in those cases when we are storing the output (with `$XONSH_STORE_STDOUT` on, or with the `![...]`, `!(...)`, or `$(...)` operators).  So I can't imagine that putting a real strain on things, as I don't expect most commands will be captured, and if someone is using `$XONSH_STORE_STDOUT`, they are doing a whole lot of writing to disk anyway.

I suppose with `$XONSH_STORE_STDOUT`, we could try to stream the data directly to the history object without using the tempfile as a go-between.  @scopatz, do you think that is possible?

adqm:
> Also, does having an in-memory structure like io.StringIO not work for some reason?

`Popen` needs not just a file-like object, but something with an actual file descriptor.  So pipes work fine (except when they fill up), or actual files, but I don't think it will be possible to something like StringIO to work directly.

In theory, we could try to wrap a pipe in something that flushes to a StringIO or similar if the pipe is full.  But I expect that getting that to work might require modifying `Popen`.

Maybe I'll give that a shot in a separate branch, if that seems preferable.

GuillaumeLeclerc:
@adqm I thought it was for every pipe. It's fine then but would also think it might be better so use a StringIO but it's not a big deal in my opinion.

scopatz:
> Maybe I'll give that a shot in a separate branch, if that seems preferable.

I agree, this is something that can come later.

> I suppose with $XONSH_STORE_STDOUT, we could try to stream the data directly to the history object without using the tempfile as a go-between. @scopatz, do you think that is possible?

My suspicion is that this is a bad idea from a code cleanliness perspective. It might make refactoring very difficult.  But I have been proven wrong about this stuff by you before :)

scopatz:
> Yes, those were both results of a boneheaded mistake. The intention was still to use PIPE for pipes in the pipeline (because those are generally handled in a streaming fashion, they should be more safe from overruns, though the deadlock will still be a possibility), and use the temp file only for the end when storing stdout/stderr. I can go in and clean that up.

No worries!

scopatz:
This looks good to me.  Again, I think that the tempfile thing is perhaps a temporary solution :). Maybe we could make another issue or if you have another branch you could just open up another PR, @adqm....

It would be nice if @melund  or @BYK could test this out on windows to make sure we don't reintroduce the issue they just reported and fixed. I am +1 to merge once they confirm.

melund:
This works well for me :)

PR #876, opened 2016-05-13
reprompt in wizard for invalid input for envvars

### LINKED ISSUE #847: Error in env setting wizard
Hello,

Sorry for posting that many issue in a day but it's my first day using it so I may find a lot of things.

When running the environment wizard. If you mispell the value for $XONSH_HISTORY_SIZE. the wizard just crash with a `KeyError`. And you loose everything so far.

scopatz:
@GuillaumeLeclerc would you mind giving this a try.  This should fix #847


scopatz:
Merging because @GuillaumeLeclerc's approval.

PR #877, opened 2016-05-13
fix for key error when completing __xonsh_shell__.shell.<TAB>

### LINKED ISSUE #826: Tab completion on __xonsh_shell__.shell. fails
[See this mailing list discussion.](https://groups.google.com/d/msgid/xonsh/eb9b3b50-33c0-4652-949b-f042ec160f78%40googlegroups.com?utm_medium=email&utm_source=footer) I think that this is caused by the coloring code being updated and something in the completions not taking this into account.

scopatz:
This gets triggered in the weirdest way (completer accessing a property), but it is a legit bug. This shoudl fix #826.


melund:
Fixes it for me :)

Issue #880, opened 2016-05-13
'which' alias doesn't work on Windows


Will-W:
The reason for this is pretty obvious - Windows doesn't have 'which'. Note though, that the command fails before aliases can even be searched, due to the unconditional evaluation of which_version().

Windows does have 'where' which could be used instead. Alternatively, a pure python xonsh builtin implementation of 'which' could be written and used cross-platform.

Given it appears this isn't the first time there's been a problem, maybe the latter is the way to go?
https://github.com/scopatz/xonsh/issues/833
https://github.com/scopatz/xonsh/issues/866

Will-W:
Addendum: If you do have a cygwin/msys implementation of 'which' on the path, then it will run and work. Except that you'll get back Linux style paths. So, in my default config if I run "which ls" I get back "/usr/bin/ls" - which is a touch surprising. Particularly surprising as the aliases seem to indicate that 'which' is a xonsh builtin (rather than just a wrapper), and xonsh producing Linux paths on Windows isn't what you'd expect.

melund:
If people don't have which from The MSYS/Cygwin I think it is OK that the get and error trying to run it. That it outputs Linux-like paths, is not really xonsh fault. A pure python implementation of which would be nice though.

melund:
Oh... The reason `which` is wrapped in xonsh is to extend it to also report on xonsh aliases. I.e. you can write:
```
[xonshdev] snail@sea ~\Documents\xonsh $ which activate
activate -> ['source-cmd-keep-promt', 'activate.bat']
```

I could imagine wrapping the windows `where` command in the same way.

adqm:
Maybe in the end we should just write it in Python (e.g., as part of #742, which I'll get back to soon).  We already keep a cache of valid command names, which may help.

Will-W:
I think this issue could have been phrased better as 'needs better handling of errors when which binary not present. This is some sample output on my machine:

```
CAMWW1@CAMWW1-W1 ~ $ which
FileNotFoundError: [WinError 2] The system cannot find the file specified
CAMWW1@CAMWW1-W1 ~ $ which -v
FileNotFoundError: [WinError 2] The system cannot find the file specified
CAMWW1@CAMWW1-W1 ~ $ which which
FileNotFoundError: [WinError 2] The system cannot find the file specified
CAMWW1@CAMWW1-W1 ~ $ which fred
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

[...]

PR #884, opened 2016-05-13
Improve handling of which being absent

Will-W:
Error handling for Issue #880

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -54,6 +54,8 @@ Current Developments
 * PLY is no longer a external dependency but is bundled in xonsh/ply. Xonsh can
   therefore run without any external dependencies, although having prompt-toolkit
   recommended.
+* Provide better user feedback when running ``which`` in a platform that doesn't
+  provide it (e.g. Windows).
 
 **Deprecated:** None
 
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -331,14 +331,14 @@ def bang_bang(args, stdin=None):
 def which_version():
     """Returns output from system `which -v`"""
     try:
-        subprocess.check_output(['which', '-v'],
-                                stderr=subprocess.PIPE).decode('UTF-8')
+        _ver = subprocess.check_output(['which', '-v'],
+                                       stderr=subprocess.PIPE).decode('UTF-8')
     except subprocess.CalledProcessError:
-        return '<no version number available on your OS>'
-
-    _ver = subprocess.check_output(['which','-v']).decode('UTF-8')
+        _ver = '<no version number available on your OS>'
+    except subprocess.FileNotFoundError:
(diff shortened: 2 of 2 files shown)


melund:
He he... I hadn't see this when I also made something similar. My suggestion was to fall back to using where: https://github.com/scopatz/xonsh/pull/885

scopatz:
They both look good to me :)  Which functionality do you both want? Or can they both be merged?

melund:
Personally, I don't care. I have 'which' on my PATH. It is just to ensure that Windows users who don't have 'which' can still get the functionality of looking up aliases. Doesn't make sense to merge both

Will-W:
So there's a philosophical question here about whether the shell should be trying to paper over platform differences. On Windows, it's reasonable to have 'which' not function, if it isn't present. Magically falling back to where means it isn't clear without code inspection what is being run (and installing cygwin would then change the behaviour).

One approach would be to have a **where** wrapper for the underlying Windows 'where', that does the same alias lookup. Then make the error for the 'which' wrapper be something like 'which not found -  try where'.

melund:
It is also nice if xonsh code is relatively portable between platforms. But falling back to where is not the nicest thing I agree. 

 Maybe just a fallback to a simple pure python solution instead? Would you be able to add something like that in your PR? 

It would be dead simple since there is already a `locate_binary()` function that searches the PATH here: https://github.com/scopatz/xonsh/blob/281b29f8d77d2f5e17ad500170f81862f7a9a8e7/xonsh/environ.py#L757

It is a little step towards pure python version...

Will-W:
xonsh scripts are only going to be as portable as the availability of the programs that theyrun. If they want to run one built for Linux, they're going to need cygwin (or msys or whatever) anyway to provide the binaries. I don't see that `which` is any different to `ls` in this regard.

The only difference between `which` and `ls` is that we wrap `which` to add some alias looking up functionality. To me that doesn't imply that we should then be making `which` work cross-platform.

That's my way of thinking though, and if the consensus is for a portable version then I'll happily do it.

melund:
@Will-W Implementing some of the core-utils in python is definitely something we want to do. See https://github.com/scopatz/xonsh/pull/742

I may speed up many things since don't have to since wrapping simple commands in subprocesses calls adds a lot of overhead. Especially on windows. 

I thought that having a python fall back if `which` could not be found would be a first step towards the goal. But it can also wait until we implement the entire thing in pure python. 

I am happy to merge your suggestion in if you think it is the best solution for now.

Will-W:
Implementing which fully in Python would be more appropriate as part of that xoreutils task. In the meantime I think that this PR gives much better user feedback on Windows.

I've updated the CHANGELOG, so I think this is ready to merge.

melund:
Thanks a lot @Will-W . Looks great

PR #885, opened 2016-05-13
Fall back to 'where' if  'which' is not available on windows

melund:
Everyone wants to play with the which...

If which is called on windows and the which binary is not available this falls back to using 'where' instead.

Also removes the long error message that contained the entire path. Was that necessary?


melund:
I am closing in favor of #884 ...

PR #886, opened 2016-05-13
Fix 'git status' option for include_untracked=True

TobalJackson:
the '--untracked-files=yes' option for 'git status' returns an error, which causes this function to return False even when there are untracked files present. I referred to this post for the correct argument of 'normal' instead of 'yes' ('no' still seems to work):
https://stackoverflow.com/questions/31708427/cant-see-untracked-files

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -76,6 +76,7 @@ Current Developments
   source operation failed for some reason.
 * Fixed PermissionError when running commands in directories without read permissions
 * Prevent Windows fixups from overriding environment vars in static config
+* Fixed Optional Github project status to reflect added/removed files via git_dirty_working_directory()
 
 **Security:** None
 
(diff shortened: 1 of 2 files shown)


melund:
Thanks a lot for the PR. It looks good to me. Could you add an entry to the change log https://github.com/scopatz/xonsh/blob/master/CHANGELOG.rst ?

melund:
Ahh. I just saw you other PR https://github.com/scopatz/xonsh/pull/888 . There is no need to make a separate PR. Just make the change to your existing branch here: https://github.com/TobalJackson/xonsh/tree/patch-1

 Then the PR should update...

TobalJackson:
Thanks, this is the first patch/pull request I've ever made, so thank you
for your patience.

melund:
Hope it won't be the last :) Thanks again

PR #887, opened 2016-05-13
Add a docker image for easier devlopment

GuillaumeLeclerc:
Hello,

I'm not sure if it's a good idea nor if the doc is in the correct place. But I found it easier than using conda or stuff like this to work on xonsh, so I share it.

Tell me what you think


scopatz:
Hi @GuillaumeLeclerc - Thanks for this! Can you merge from master, please?

GuillaumeLeclerc:
Do you think the doc I added is where it should be ?

Dockerfile
```diff
@@ -0,0 +1,11 @@
+from python:3
```
funkyfuture: should be `python:3.4` or even `python:3.4-alpine`

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+RUN pip install --upgrade pip && pip install \
+  ply \
+  prompt-toolkit \
+  pygments
+RUN mkdir /xonsh
+WORKDIR /xonsh
+ADD ./ ./
+RUN python setup.py install
```
funkyfuture: maybe `setup.py develop`, so one can hack around

Dockerfile
funkyfuture: actually it would be cool to have a wrapper script that takes the desired Python version as argument. or rather embed the `Dockerfile` content in a script.

GuillaumeLeclerc:
That is a great idea. I will add the version of python toolkit as an argument too

GuillaumeLeclerc:
And here it is a simple python script (compatible with python 2 for the ones who want to try xonsh without installing python3 on their machine) that build and run xonsh in a isolated container

Is it fine ?

docker.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+import sys 
+import subprocess
+import os
+
+pythonVersion = '3.5'
+ptkVersion = '1.00'
+
+if len(sys.argv) > 1:
+    pythonVersion = sys.argv[1]
```
gforsyth: Can you use argparse here? This seems prone to errors if someone flip flops the arguments

docker.py
```diff
@@ -0,0 +1,37 @@
...
+    if len(sys.argv) > 2:
+        ptkVersion = sys.argv[2]
+
+print('Building and runing Xonsh')
```
gforsyth: *running

docker.py
```diff
@@ -0,0 +1,37 @@
...
+dockerFile = 'from python:'+pythonVersion + '\n'
+dockerFile += 'RUN pip install --upgrade pip && pip install \\\n'
+dockerFile += '  ply \\\n'
+dockerFile += '  prompt-toolkit=='+ptkVersion+ ' \\\n'
+dockerFile += '  pygments\n'
+dockerFile += 'RUN mkdir /xonsh\n'
+dockerFile += 'WORKDIR /xonsh\n'
+dockerFile += 'CMD /usr/bin/env xonsh\n'
+dockerFile += 'ENV XONSH_COLOR_STYLE "paraiso-dark\n'
+dockerFile += 'ADD ./ ./\n'
```
gforsyth: I am not very docker literate. What is this doing?

GuillaumeLeclerc: It copies the content of the current directory to the container so I can run setup.py

docker.py
GuillaumeLeclerc: I will fix it sorry :)

docker.py
GuillaumeLeclerc: Sure I'll take a look how it works

docker.py
```diff
@@ -0,0 +1,37 @@
...
+if len(sys.argv) > 1:
+    pythonVersion = sys.argv[1]
+    if len(sys.argv) > 2:
+        ptkVersion = sys.argv[2]
+
+print('Building and runing Xonsh')
+print('Using python ', pythonVersion)
+print('Using prompt-toolkit ', ptkVersion)
+
+dockerFile = 'from python:'+pythonVersion + '\n'
```
funkyfuture: a multiline-string would be better readable

docker.py
```diff
@@ -0,0 +1,37 @@
...
+dockerFile += '  prompt-toolkit=='+ptkVersion+ ' \\\n'
+dockerFile += '  pygments\n'
+dockerFile += 'RUN mkdir /xonsh\n'
+dockerFile += 'WORKDIR /xonsh\n'
+dockerFile += 'CMD /usr/bin/env xonsh\n'
+dockerFile += 'ENV XONSH_COLOR_STYLE "paraiso-dark\n'
+dockerFile += 'ADD ./ ./\n'
+dockerFile += 'RUN python setup.py install\n'
+
+with open('./Dockerfile', 'w+') as f:
```
funkyfuture: you can stream the content of the Dockerfile into docker's stdin. (and if a file was necessary that should come from `tempfile`)

docker.py
```diff
@@ -0,0 +1,37 @@
...
+    f.write(dockerFile)
+
+subprocess.call(['docker', 'build', '-t' , 'xonsh', '.'])
+os.remove('./Dockerfile')
+subprocess.call(['docker', 'run', '-ti' , 'xonsh'])
```
funkyfuture: what amout `--rm` or arbritrary argument passing?

funkyfuture:
i think the script's name should be more obvious.

also, it should be considered that three agents could be addressed here with arguments:

- the script itself with the image tag
- the docker run command
- the command executed in the container

what about seperating them with `--` and reverse assignment, so if only one group of arguments are given, they get passed to xonsh.

docker.py
GuillaumeLeclerc: I don't know how to do string interpolation in a multiline string in python

docker.py
GuillaumeLeclerc: Good catch thanks

GuillaumeLeclerc:
What do you propose as a script name ?

I was thinking about passing ENV variables too. It can be handy when you test them. I was needing this when I worked on #895

docker.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python
+import sys 
+import subprocess
+import os
+
+pythonVersion = '3.5'
+ptkVersion = '1.00'
```
scopatz: also , please be pep8 compliant :)

Dockerfile
GuillaumeLeclerc: What is the difference between the two ?

docker.py
GuillaumeLeclerc: From the docker docs:
***
If you build by passing a Dockerfile through STDIN (docker build - < somefile), there is no build context, so the Dockerfile can only contain a URL based ADD instruction. You can also pass a compressed archive through STDIN
***

Since in need to add the content of the current directory this is not possible

I will use tempfile then

GuillaumeLeclerc: I don't know how to have a temp file in the directory I want (ie. current working directory). Do you know how to do this ?

docker.py
funkyfuture: https://docs.python.org/3.4/library/stdtypes.html#text-sequence-type-str

docker.py
funkyfuture: ah, i could bet that was possible once. please also put a `.dockerignore` into context to speed things up.

funkyfuture: https://docs.python.org/3.4/library/tempfile.html#tempfile.TemporaryDirectory

Dockerfile
funkyfuture: https://pythonhosted.org/setuptools/setuptools.html#development-mode

funkyfuture:
> What do you propose as a script name ?

i don't have a good idea. `xonsh-in-docker.py` would somehow be describing.

> I was thinking about passing ENV variables too. It can be handy when you test them.

well that's the `-e` option for the `docker run` command.

docker.py
```diff
@@ -0,0 +1,51 @@
...
+ADD ./ ./
+RUN python setup.py install
+""".format(
+        python_version = args.python,
+        ptk_version = args.ptk,
+        command = args.command,
+        env = ' '.join(args.env))
+        
+
+print(docker_script)
```
funkyfuture: i'd say this isn't necessary as the directives are also prinbted during the build.

GuillaumeLeclerc: It was a debug print sorry I forgot to remove it

GuillaumeLeclerc:
@funkyfuture I pushed a version that use the `-e` flag from docker but now we are forced to use the following syntax for env variables:

`./xonsh-in-docker.py --env "TEST= 1 2"`
`./xonsh-in-docker.py --env TEST=" 1 2"` does not work anymore.

funkyfuture:
what about this?

```
def pop_args(args):
    try:
        return args.pop().strip()
    except IndexError:
        return ''

args = args.split('--')
xonsh_args = pop_args(args)
docker_run_args = pop_args(args)
wrapper_script_args = pop_args(args)
```

so, setting an environment variable:
`./xonsh-in-docker.py -e TEST=foo --`

funkyfuture:
but it doesn't handle `--` between quotes..

GuillaumeLeclerc:
And I'm not sure how to do this with the `argparse` module

funkyfuture:
you then just feed `wrapper_script_args` to argparse and use the other strings for `subrocess.call`.

funkyfuture:
oh, the above example was intedned as pseudo code. so, the args must be casted into a string first. i'll update it.

GuillaumeLeclerc:
@funkyfuture I get it now but it does not solve the "having -- inside problem"

funkyfuture:
you have a Python in your hand. go for it! ;-)

GuillaumeLeclerc:
This is equivalent to writing a parser for a given grammar, maybe overkill in my opinion. Just having the ENV vars is the Docker file instead of passing it with `-e` is better (it triggers a rebuild when you change vars, which is not ideal)

funkyfuture:
``` python


def splitargs_by_double_dashes(args):
    line = ' '.join(args)
    buffer, result, quoted = '', [], False

    while line:
        if not quoted and len(line) > 3 and line[:4] == ' -- ':
            result.append(buffer.strip().split(' '))
            buffer = ''
            line = line[4:]
            continue

        char = line[0]
        if char in ('"', "'"):
            quoted = not quoted
        buffer += char
        line = line[1:]

    result.append(buffer.strip().split(' '))
    while len(result) < 3:
        result.insert(0, '')

    return result
```

xonsh-in-docker.py
```diff
@@ -0,0 +1,53 @@
...
+import os
+import argparse
+
+program_description = """Build and run Xonsh in a fresh, controlled 
+    environment using docker """
+
+parser = argparse.ArgumentParser(description=program_description)
+
+parser.add_argument('env', nargs='*', default=[], metavar='ENV=value')
+parser.add_argument('--python', '-p', default='3.4', metavar='python_version')
```
funkyfuture: should be called `tag` or `python-tag` or `python-image-tag`

GuillaumeLeclerc: True ! I'll change this

GuillaumeLeclerc:
@splitargs_by_double_dashes(['"tot\' -- \' hoho"'])

is wrong, the result split even if the "--" are quoted by double quotes. It can't work without:

- Checking which type of quote is used and not considering the same way single and double quotes
- Ignoring escaped strings

It's very easy to make mistake implementing such parsers. That's what I thought it may be considered over-engineering if the parser is longer than the whole script.

funkyfuture:
i got your point, yet enjoyed the challenge.

let's take two steps back and let us consider this:

- the script builds the image
- it has a `--build` option that only builds the image and leaves the rest to the user.

i think `docker run`'s `--rm` option should be enabled by default to not leave unused containers around. as this aims at testing purposes that is certainly wanted.

GuillaumeLeclerc:
I'll add a --build options and make rm default. Thank you for the advise

scopatz:
Looks like this has some merge conflicts with master to be addressed. I'd rather have this in than worry about the build option, if you just want to clean it up.

GuillaumeLeclerc:
I merged with master, changed the --rm option to --keep so by default the container is not kept. and add a --build option that does not run the docker image, only builds it

GuillaumeLeclerc:
Is everything fine ?

scopatz:
Looks good to me @GuillaumeLeclerc!  We can address other issues as they arise.  Thanks a million!

Issue #879, opened 2016-05-13
Spaces in Python installation path causes xonsh.exe launcher to break on Windows


Will-W:
Steps to repeat:
Have Python installed on Windows, to a directory with a space in the name. This can be Program Files if installed for all users, or it can be your user directory if the username has a space in it (NEVER DO THIS, except that people do).
Install xonsh in the normal way using `python setup.py install` (or via pip).
Run scripts/xonsh.exe.

Expected:
xonsh to run

Actual:
Message "failed to create process."

This is clearly a known issue with pip/setuptools:
https://github.com/pypa/setuptools/issues/398
https://github.com/pypa/pip/issues/2783

I may try and pursue getting it fixed at source if I have some time to understand what's actually going on (setuptools has always been a black hole for me). In the meantime, the second of those two issue discussions contains some workaround code that might be appropriate, to get this issue killed off. Happy to submit a PR.

PR #890, opened 2016-05-13
Fix launcher for Windows install dirs with a space

Will-W:
Feeling productive and useful today. Have another fix, this time for #879.

I've tested this on Windows, both as 'setup.py install', and via a wheel (which should be equivalent to pip).

I haven't tested for regressions on Linux, because I won't be near a Linux machine until Monday. There should be absolutely no reason why this does break something on Linux, but better safe than sorry.


scopatz:
looks great to me! thanks a million!

PR #891, opened 2016-05-14
Report only Syntax error for semi-colon.

scopatz:
This is for correct error reporting for the ";" alone case.


melund:
This works fine for me.

gforsyth:
Here, too.

PR #892, opened 2016-05-14
To join the `which` hunting mob...

scopatz:
I added passing in the env so that it sees the correct $PATH.  Probably for fellow which-hunters @gforsyth and @melund

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -330,14 +330,15 @@ def bang_bang(args, stdin=None):
 @lru_cache(1)
 def which_version():
     """Returns output from system `which -v`"""
+    denv = __xonsh_env__.detype()
     try:
-        _ver = subprocess.check_output(['which', '-v'],
+        _ver = subprocess.check_output(['which', '-v'], env=denv,
                                        stderr=subprocess.PIPE).decode('UTF-8')
     except subprocess.CalledProcessError:
         _ver = '<no version number available on your OS>'
     except FileNotFoundError:
         _ver = "<'which' binary not found>"
-        
+
     return _ver
 
 class AWitchAWitch(Action):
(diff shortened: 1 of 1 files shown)


melund:
Works fine. But I kind of hate the fact that it prints the entire PATH when it can't find the binary. 

Could we change:
```
raise XonshError('{} not in {} or xonsh.builtins.aliases'
                             .format(args[0], ':'.join(__xonsh_env__['PATH'])))
```
to
```
raise XonshError('{} not in $PATH or xonsh.builtins.aliases'.format(args[0]))
```

gforsyth:
Hey @melund -- that was me, just copying the behavior of `gnu-which` but I'm open to removing it.  I'm going to merge in this fix and we can alter the $PATH message in a separate PR.

melund:
@gforsyth .. Oh. If it matches the behaviour of the real which, then I am more positive. I am still considering a pure Python fall back when the which binary is not available.

 It bugs me that it works for aliases but reports 'which binary not found' if it doesn't find an alias and tries to call 'which'. Kind if like half useful

gforsyth:
That does sound super annoying @melund.   Maybe with the new bash on windows thing everyone will get `which`?  But let's not wait around for others to fix it -- I think a pure python `which` would be great.  

That reminds me.. I forgot until now, but I think that's already been done.  Does this work for you on Windows? https://github.com/trentm/which

melund:
@gforsyth.   I will try it. There is also a which in shutil. https://docs.python.org/3/library/shutil.html#shutil.which

We can't use it directly thus since on windows it gets 'PATHEXT' from `os.environ` directly. Also the 'which' you found seems to look stuff from the registry as well.

melund:
@gforsyth. Which.py works with a few updates to python 3. It has some really usefull Windows only functionality. I will give it a try and implement it.Then it can be moved to the xoreutils later.

melund:
#901

Issue #895, opened 2016-05-14
Path compression in prompt


GuillaumeLeclerc:
Hello,

When working with deep file structure. The path displayed in the prompt can be huge. 

I was thinking about a feature to solve this, we could either:

- Only display the last directory but only the first letter of the previous ones: `/d/a/a/b/test'
- add a length in the config and design an algorithm that make the path fit into the number of chars. Example
 `/ho/gui/.c/xonsh` for `/home/guillaumeleclerc/.config/xonsh`

Tell me what you think about this

( Of course it would be mandatory, we could fill a new variable and it's up to the user to add it into his prompt or not)

melund:
It is a good idea, and one we already have. Maybe it is not well documented. Just replace `{cwd}` in your prompt with `{short_cwd}`

melund:
It is here: http://xon.sh/tutorial.html?highlight=short_cwd#customizing-the-prompt

GuillaumeLeclerc:
Cool ! 

What do you think about the second option. I think it would be cool to always display the full path until there are not enough space.

melund:
Yes. I think it is a great idea It is actually one of the reasons I don't use the short prompt. A PR would be awesome

GuillaumeLeclerc:
I will try to implement it then

gforsyth:
Hey @GuillaumeLeclerc, I also think this is a good idea, just a thought to add on to this (that maybe you've already had!), rather than having a static max length set in a config file, you could also have it be a percentage of the current terminal width, so it wouldn't be over-zealous in contracting the working directory if you were, say, working outside of an xserver?

Some info on grabbing terminal widths here: http://rosettacode.org/wiki/Terminal_control/Dimensions#Python

melund:
Shutil have a function for it `shutil.get_terminal_size()`

GuillaumeLeclerc:
@gforsyth That's a great idea. I think the option will support either chars or options as a unit.

GuillaumeLeclerc:
Where do I add a new xonsh ENV var ?

gforsyth:
Hey @GuillaumeLeclerc, they go in `environ.py` in both the the [default ensurers](https://github.com/scopatz/xonsh/blob/master/xonsh/environ.py#L61) and [default values](https://github.com/scopatz/xonsh/blob/master/xonsh/environ.py#L155)

GuillaumeLeclerc:
So I have implemented a first draft that have the following policy:

[...]

PR #899, opened 2016-05-14
Implement a dynamic width CWD

GuillaumeLeclerc:
Hello,

This is my first try to implement the feature described in #895.
To determine how much characters it should allocate to each part of the path it uses a greedy O(n) algorithm.

I have two others algorithms in O(n²) and O(n * max(map(len, parts)).

I think the greedy one, even if it's not perfect (it might lead to output smaller than the target length), is quite good. Tell me if I should an optimal algorithm and which complexity I should choose.

From now there is a new env var: `DYNAMIC_CWD_WIDTH` of type string. two modes are supported:
- The max number of chars to display in the path (`"30"`)
- The percentage of the width of the screen (`"20%"`)

The default value is `20`. Should we change this ?

There are no tests at the moment. 

- Should I add some ? 
- Where should they be ? 
- What should they test ? 

I hope you like this feature as much as I do (It's already running on my computers :) )


xonsh/environ.py
```diff
@@ -1007,6 +1013,36 @@ def _collapsed_pwd():
...
+
+    last = pwd.pop()
+    remaining_space = targetWidth - len(last) 
+    # Reserve space for separators
+    remaining_space_for_text = remaining_space - len(pwd) 
+
+    parts = []
+    for i  in range(len(pwd)):
+        part = pwd[i]
+        part_len = min(len(part), max(1, math.floor(remaining_space_for_text / (len(pwd) - i))))
```
gforsyth: You can use `//` for integer division and avoid importing the `math` library.

gforsyth:
Hey @GuillaumeLeclerc -- this looks awesome!  I'm going to take a closer look later tonight (EST) or tomorrow morning but this looks like a really nice feature to have.

xonsh/environ.py
```diff
@@ -84,6 +86,7 @@ def lc_converter(val):
     'PATHEXT': (is_env_path, str_to_env_path, env_path_to_str),
     'RAISE_SUBPROC_ERROR': (is_bool, to_bool, bool_to_str),
     'RIGHT_PROMPT': (is_string, ensure_string, ensure_string),
+    'DYNAMIC_CWD_WIDTH': (is_string, ensure_string, ensure_string),
```
gforsyth: should this be a float and we can avoid the casting below?

gforsyth: whoops.  never mind -- i see now that you're checking whether it's a percentage or not

gforsyth:
Hey @GuillaumeLeclerc -- I'm liking this!  A few small issues I've run into:

1.  There's a leading `/` in the `PATH`
```console
gil@bad_cat /~ $ cd
gil@bad_cat /~ $ pwd
/home/gil
```

2.  With long directory names, behavior is not what I expect (maybe a minimum required length is required for each directory?)
```console
gil@bad_cat /~ $ cd tmp/
gil@bad_cat ~/tmp $ ls
a_really_rather_long_dir_name  bar  exists  foo
gil@bad_cat ~/tmp $ cd a_really_rather_long_dir_name/
gil@bad_cat rather_long_dir_name $ 
```

Maybe something like

```console
gil@bad_cat ~/t/...long_dir_name $
```

xonsh/environ.py
```diff
@@ -191,6 +194,7 @@ def xonshconfig(env):
     'PUSHD_SILENT': False,
     'RAISE_SUBPROC_ERROR': False,
     'RIGHT_PROMPT': '',
+    'DYNAMIC_CWD_WIDTH': '20',
```
scopatz: could you make the default infinite, please

melund: Do you mean that if defaulted to 'infinite' it could just replace the standard {cwd}?

xonsh/environ.py
GuillaumeLeclerc: Thank you for the tip, it's really appreciated since I'm quite new to python. (I never used it in large scale project only in algorithmic challenges)

xonsh/environ.py
GuillaumeLeclerc: What should be the string representation of infinite ? +INF 

Maybe it would not be good to replace the default cwd because this one if more complex to compute maybe not everyone is willing to pay this price

adqm: > What should be the string representation of infinite ? +INF 

That works.  There are a number of things that generate Python's internal representation for infinity when passed as an argument to `float`.  In fact, it looks like an capitalization of INF works, with or without the "+" sign in front (so `"inf"`, `"Inf"`, `"inF"`, `"INf"`, `"INF"` all work).

Because Python's repr for infinity is lower-case `"inf"`, I might vote for using that.

```
>>> float('inf')
inf
```

> Maybe it would not be good to replace the default cwd because this one if more complex to compute maybe not everyone is willing to pay this price

Do you have any data on how long it takes to compute?  I imagine it would end up not being _that_ different/noticeable for most prompts.

In particular, if the default length is infinity, we could "short-circuit" all the computation and just return the prompt as-is.

adqm:
> Maybe something like
> gil@bad_cat ~/t/...long_dir_name $

I had thought the plan was always to leave the last directory name visible in its entirety?  The ellipsis does help, though.

gforsyth:
>> Maybe something like
>>     gil@bad_cat ~/t/...long_dir_name $
> 
> I had thought the plan was always to leave the last directory name visible in its entirety? The ellipsis does help, though.

Ahh, well, then scratch my point 2 above.  Or rather, the second point 1 (thanks markdown!)

If the last directory name is longer than the allowed width, does it make sense to truncate the end of it, rather than the beginning?

xonsh/environ.py
GuillaumeLeclerc: Ok I will use `inf` then. But having float means I won't be able to use integer division as @melund proposed. (As far as I know python does do allow representation of infinity with the integer type)

On the other side I do like the "short-circuit" thing

I don't think it long to compute it's just an overhead of `len(pwd)` integer division + substring exraction.

GuillaumeLeclerc:
@gforsyth I will fix point 1.

I think we need to discuss a little more about the second point though. I'm not sure what's the best

xonsh/environ.py
gforsyth: Actually, integer division with `//` works with floats.  

```console
> 5 // 2 
2
>5.0 // 2
2.0
```

GuillaumeLeclerc: thanks ! Learning every day. My intuition was that __integer__ division would not work on floats. I was wrong :p

gforsyth: It's not the most intuitive name, is it?  Perhaps we should call `//` floor division instead? ;)

GuillaumeLeclerc:
@gforsyth. Leading `/` is fixed :)

GuillaumeLeclerc:
I think the only remaining thing to think about is:

- What to do for really long directories
- What to do for really deep path

I'm not sure if the current behavior is the best in this situation

xonsh/environ.py
```diff
@@ -361,6 +364,8 @@ def xonshconfig(env):
         'at the prompt. This may be parameterized in the same way as '
         'the $PROMPT variable. Currently, this is only available in the '
         'prompt-toolkit shell.'),
+    'DYNAMIC_CWD_WIDTH': VarDocs('Target length in number of characters '
+        'for the {dynamic_cwd} prompt var'),
```
scopatz: This should explain how the parsing works with `%`, `inf`, etc.  Also, have you tried running `xonfig` with this branch? I am concerned that the braces `{}` might break it.

scopatz:
Thanks for putting this in @GuillaumeLeclerc! Seems like a neat feature.  @gforsyth and @adqm  have a good handle on this one. Just wanted to voice my approval :)

gforsyth:
Ok, this is looking great.  I propose that for directories that are long enough to be truncated that there is a `...` prepended before a leading `/` and then as much of the directory name as can fit (read left-to-right) and then (possibly?) another `...` at the end.  So it might look like:

```console
~ $ cd kind_of_long
~/kind_of_long $ cd really_a_bit_too_long
~/k/really_a_bit_too_long $ cd this_has_really_gotten_out_of_hand
.../this_has_really_go... $
```

or without the trailing Ellipsis


```console
~ $ cd kind_of_long
~/kind_of_long $ cd really_a_bit_too_long
~/k/really_a_bit_too_long $ cd this_has_really_gotten_out_of_hand
.../this_has_really_gotte $
```

I think I prefer the second version.  I also will put forward a strong preference for truncating the _end_ of the current directory name as that best fits my internal language parser.

Any thoughts from @GuillaumeLeclerc , @adqm, @scopatz, @melund and others welcome!

xonsh/environ.py
```diff
@@ -1007,6 +1012,38 @@ def _collapsed_pwd():
     base = [i[0] if ix != l-1 else i for ix,i in enumerate(pwd) if len(i) > 0]
     return leader + sep.join(base)
 
+def _dynamically_collapsed_pwd():
+    sep = get_sep()
+    originial_path = _replace_home_cwd()
+    pwd = originial_path.split(sep)
+    cols, _ = shutil.get_terminal_size()
+    targetWidthRaw = builtins.__xonsh_env__['DYNAMIC_CWD_WIDTH']
```
scopatz: PEP8 variable name please

xonsh/environ.py
```diff
@@ -1007,6 +1012,38 @@ def _collapsed_pwd():
...
+
+    if (targetWidthRaw[-1] == '%'):
+        targetWidth = (cols * float(targetWidthRaw[:-1])) // 100
```
scopatz: PEP8 variable name please.

scopatz:
Hi @GuillaumeLeclerc, I'd also suggest that you implement custom ensurers and completer functions for the envvar, rather than than relying on just a string representation.  Doing something like [`$XONSH_HISTORY_SIZE`](https://github.com/scopatz/xonsh/blob/master/xonsh/environ.py#L95) is my advice, since both of these are numbers with units.

scopatz:
Also I second what @gforsyth said!

GuillaumeLeclerc:
I'll do all this when I have little time, I do prefer the version without the second ellipsis

scopatz:
Thanks @GuillaumeLeclerc, I think that this would be great to have for v0.3 if you get the opportunity.

GuillaumeLeclerc:
I should have the time to do this on Monday, I'm glad you like it.

melund:
Ahh. If we release 0.3 on Friday it won't make it in. That would be a shame. I there anything could be postponed to a separate PR. For example the ensures. Or something we could help with

scopatz:
Well, we can always do a quick v0.3.1 if we want.  I am just getting nervous a bit that I won't have v0.3.0 to make my PyCon slides off of.

melund:
Yes. I agree. We really need to get a 0.3 out.

GuillaumeLeclerc:
I'll to do it before then. I didn't there was a deadline

GuillaumeLeclerc:
Is the behavior correct @gforsyth ? 

I'm adding the ensurer now

GuillaumeLeclerc:
I implemented the ensurers but I'm not sure if it ok:

- Only the first one
- Even if the first one return false, no error is displayed about the format and therefore the function to compute the reduce path throws an exception.

gforsyth:
Hey @GuillaumeLeclerc -- this is great!  I really like this.  Seem to be some conflicts with master now

GuillaumeLeclerc:
I'll fix it right away but what about the two points I raised in my last comment ?

GuillaumeLeclerc:
I fixed the conflicts.

Should I add unit tests ?

xonsh/tools.py
```diff
@@ -737,6 +737,27 @@ def is_history_tuple(x):
     return False
 
 
+def is_dynamic_cwd_width(x):
```
scopatz: Please add doc strings on these functions too.

scopatz:
Tests would be highly appreciated!

GuillaumeLeclerc:
Anything about the changelog ?

scopatz:
Oh yeah, please add an update to the changelog too.

GuillaumeLeclerc:
and about these two problem, Did I do something wrong ? 

- Only the first of the three functions in the tuple is actually called at some point
- Even if the first one return false, no error is displayed about the format and therefore the function to compute the reduce path throws an exception.

scopatz:
> Only the first of the three functions in the tuple is actually called at some point

The other two are called  converting to/from strings.

> Even if the first one return false, no error is displayed about the format and therefore the function to compute the reduce path throws an exception.

Can you provide the traceback please?

GuillaumeLeclerc:
```
<user>@504907e24a68 /xonsh myxonsh # $DYNAMIC_CWD_WIDTH="20m"
ValueError: could not convert string to float: '20m'
  File "/usr/local/bin/xonsh", line 9, in <module>
    load_entry_point('xonsh==0.2.7', 'console_scripts', 'xonsh')()
  File "/usr/local/lib/python3.4/site-packages/xonsh/main.py", line 249, in main
    shell.cmdloop()
  File "/usr/local/lib/python3.4/site-packages/xonsh/ptk/shell.py", line 124, in cmdloop
  File "/usr/local/lib/python3.4/site-packages/xonsh/ptk/shell.py", line 70, in singleline
    prompt_tokens = self.prompt_tokens(None)
  File "/usr/local/lib/python3.4/site-packages/xonsh/ptk/shell.py", line 146, in prompt_tokens
    self.settitle()
  File "/usr/local/lib/python3.4/site-packages/xonsh/base_shell.py", line 228, in settitle
    t = format_prompt(t)
  File "/usr/local/lib/python3.4/site-packages/xonsh/environ.py", line 1147, in format_prompt
    val = v() if callable(v) else v
  File "/usr/local/lib/python3.4/site-packages/xonsh/environ.py", line 1036, in _dynamically_collapsed_pwd
    targetWidth = float(targetWidthRaw)
ValueError: could not convert string to float: '20m'
```

GuillaumeLeclerc:
How can I test these two functions then (except with unit tests)

scopatz:
Hi @GuillaumeLeclerc - yeah, you'll need to add unit tests.  I think @gforsyth had a PR that has now been merged that should let you run the tests now.  And this traceback looks like it is bug in the code that you wrote as part of this PR.  Adding tests for the cases you expect to be used would help here :)

GuillaumeLeclerc:
@scopatz I know the error is comming from my code but since I wrote an ensurer that would return false for this value I don't think why the function is called at all. If they are called with values that does not validate then ensurers then I don't understand what they are for.

I will try to write the tests tomorrow.

scopatz:
> Basically that means I need to validate the input both in the ensurer and every time I use the value ?

So this is a subtle issue I had not considered.  Basically, you had set the PROMPT before you set `DYNAMIC_CWD_WIDTH`.  Could you merge from master and resolve the conflicts, please, so that I can play around with this.

GuillaumeLeclerc:
I have merged it you can take a look when you have time

xonsh/environ.py
scopatz: Also, this is no longer true.  This just affects `{cwd}`

scopatz:
OK, now that I have played around with this, I believe that the Env is preforming correctly.  When you try to set, `$DYNAMIC_CWD_WIDTH="20m"`, it fails (since "20m" is invalid input), throws an error, and does not actually set the `$DYNAMIC_CWD_WIDTH`.  If valid input is provided, this variable is set.  Note that it seems that only `"20%"` and `"20c"` would be allowed.

So I think that this is basically ready to be merged.  It might be nice to add a mention of the new envvar in [this section of the docs](http://xon.sh/tutorial.html#customizing-the-prompt).  Also some unit tests for teh converters would be *highly* valuable.

melund:
This is awesome. Can't wait until it goes into 0.3.1

scopatz:
I went ahead and cleaned this up and merged it.

GuillaumeLeclerc:
I noticed a bug today when running a xonsh script. I need to dig further. But finals are around the corner letting me almost no time.

scopatz:
Understood @GuillaumeLeclerc. No worries, report it when you get a chance, and good luck on your finals!

PR #900, opened 2016-05-15
Load autocompletions from directory

funkyfuture:
this adds the capability to define directories to load autocompletions from.

i started to move platform-specifics to a new module. this isn't finished yet, though working. i will finish it if there is approving feedback.

unfortunatley i can't rebase onto master as git won't stop to autoconvert `scripts/xonsh.bat` what i also try to tackle in `.gitattributes`. without success.

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -5,6 +5,7 @@ import re
 import ast
 import sys
 import shlex
+from pathlib import Path
 import pickle
 import inspect
 import builtins
@@ -40,7 +41,7 @@ def _path_from_partial_string(inp, pos=None):
             else:
                 return None
         string = partial[startix:endix]
-    end = re.sub(RE_STRING_START,'',quote)
+    end = re.sub(RE_STRING_START, '', quote)
     _string = string
     if not _string.endswith(end):
         _string = _string + end
@@ -84,6 +85,7 @@ for ((i=0;i<${{#COMPREPLY[*]}};i++)) do echo ${{COMPREPLY[i]}}; done
 
 WS = set(' \t\r\n')
 
+
 def startswithlow(x, start, startlow=None):
     """True if x starts with a string or its lowercase version. The lowercase
     version may be optionally be provided.
@@ -119,6 +121,7 @@ def _normpath(p):
         p = p.replace(os.sep, os.altsep)
 
     return p
+
 
 class Completer(object):
     """This provides a list of optional completions for the xonsh shell."""
@@ -268,8 +271,9 @@ class Completer(object):
         return {max(i, key=len) for i in reps.values()}
 
     def find_and_complete(self, line, idx, ctx=None):
-        """Finds the completions given only the full code line and a current cursor
-        position. This represents an easier alternative to the complete() method.
+        """Finds the completions given only the full code line and a current
+        cursor position. This represents an easier alternative to the
+        complete() method.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


gforsyth:
Hey @funkyfuture, thanks for putting this in.  

I haven't had the issues you describe with `xonsh.bat` but hopefully someone else can pipe up with a solution.  

Could you split this up into several PRs?  I see the autocompletion loading, the creation of `xonsh.platform` (which you mention) and the removal of some old Py2 compatibility stuff.  I think those are all good ideas but they should be treated separately.

GuillaumeLeclerc:
I'm having the same problem with `xonsh.bat`, but I only do merge no rebase and it seems fine

funkyfuture:
@gforsyth yeah, it doesn't seem to introduce conflicts. will do this europe's evening.

funkyfuture:
updated the commits.

@GuillaumeLeclerc i have git 2.8.2 on Manjalo Linux

funkyfuture:
sorry i can't squash in the missing `from pathlib import Path`. would be great if someone did this locally.

gforsyth:
@funkyfuture -- it's fine to just add another commit to fix the missing import.  Rebasing is great but don't let the great be the enemy of the good.

scopatz:
Yeah @funkyfuture, to second @gforsyth, feel free to merge.

melund:
There are some tests failing now

funkyfuture:
fixed it.

xonsh/completer.py
```diff
@@ -425,38 +427,40 @@ def bash_complete(self, prefix, line, begidx, endidx):
...
+                for _file in (x for x in path.glob('*') if x.is_file()):
+                    sources.append('source ' + str(_file))
+        return sources
 
     def _load_bash_complete_funcs(self):
         self.bash_complete_funcs = bcf = {}
-        inp = self._source_completions()
-        if len(inp) == 0:
+        inp = self._collect_completions_sources()
+        if not inp:
```
scopatz: Please revert this to `if len(inp) == 0`. This is the more precise test, and ambiguity isn't needed here. Thanks!

xonsh/completer.py
```diff
@@ -467,17 +471,16 @@ def _load_bash_complete_funcs(self):
             bcf[cmd] = m.group(1)
 
     def _load_bash_complete_files(self):
-        inp = self._source_completions()
-        if len(inp) == 0:
+        inp = self._collect_completions_sources()
+        if not inp:
```
scopatz: Same.

gforsyth:
I can confirm that this is working on Arch.  There are weird completion behaviors with certain things (most notably `pacman`) but those were there already.  This looks good to merge to me once @scopatz changes have been made.

scopatz:
Hey @funkyfuture - just a couple of small changes but otherwise lgtm

xonsh/completer.py
funkyfuture: my rationale is that both expressions are equal as inp is always a list, so there is no ambiguity. my motivation is a better readability.

scopatz: I personally find the original way much more readable :) But I guess this way is much faster,

```console
scopatz@localhost ~ $ x = []
scopatz@localhost ~ $ timeit "len(x) > 0"
10000000 loops, best of 3: 75.8 ns per loop
scopatz@localhost ~ $ timeit "not x"
10000000 loops, best of 3: 28.4 ns per loop
```

funkyfuture: thanks for the research. never thought about that aspect. what results shows `not len(x)`? ( i currently bugged my xonsh)

scopatz: I am getting, 

```console
scopatz@localhost ~ $ timeit "not len(x)"
10000000 loops, best of 3: 64 ns per loop
```

scopatz:
Ok @funkyfuture - I think this looks good.  Any other changes you want to make, or is it good to merge?

funkyfuture:
gtm

scopatz:
Thanks a million @funkyfuture! This is awesome to have.

PR #901, opened 2016-05-15
Update which to be a pure python implementation.

melund:
@gforsyth I added the pure python 'which' which you suggested :)
I include the `which.py` from https://github.com/trentm/which but updated for Python 3. The interface is is still in `aliases.py`, but I imagine that @adqm could move the whole thing to `xoreutils` when that is ready.

This new which have some differences:
- It can handle multiple inputs. 
- it understands PATHEXT and "App Paths" registration on Windows (i.e. it
    will find everything that `start` does from the command shell); 
- it can print all matches on the PATH; 
- it can note "near misses" on the PATH (e.g. files that match but may not,
    say, have execute permissions)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,12 +28,14 @@ Current Developments
   running the Anaconda python distribution on Windows.
 * Added a menu entry to launch xonsh when installing xonsh from a conda package
 * Added a new ``which`` alias that supports both regular ``which`` and also searches
-  through xonsh aliases
+  through xonsh aliases. A pure python implementation of ``which`` is used. Thanks
+  to Trent Mick. https://github.com/trentm/which/
 * Added support for prompt toolkit v1.0.0.
 * Added ``$XONSH_CACHE_SCRIPTS`` and ``$XONSH_CACHE_EVERYTHING`` environment
   variables to control caching of scripts and interactive commands.  These can
   also be controlled by command line options ``--no-script-cache`` and
   ``--cache-everything`` when starting xonsh.
+  
 
 **Changed:**
 
--- a/setup.py
+++ b/setup.py
@@ -160,7 +160,7 @@ def main():
         url='https://github.com/scopatz/xonsh',
         platforms='Cross Platform',
         classifiers=['Programming Language :: Python :: 3'],
-        packages=['xonsh', 'xonsh.ptk', 'xonsh.parsers', 'xontrib'],
+        packages=['xonsh', 'xonsh.ptk', 'xonsh.parsers', 'xonsh.xoreutils', 'xontrib'],
         package_dir={'xonsh': 'xonsh', 'xontrib': 'xontrib'},
         package_data={'xonsh': ['*.json'], 'xontrib': ['*.xsh']},
         cmdclass=cmdclass
--- a/tests/test_aliases.py
+++ b/tests/test_aliases.py
@@ -3,12 +3,15 @@
 from __future__ import unicode_literals, print_function
 
 import os
+import tempfile
+
 import nose
 from nose.plugins.skip import SkipTest
 from nose.tools import assert_equal
 
 import xonsh.built_ins as built_ins
 from xonsh.aliases import Aliases
+from xonsh.aliases import _which
 from xonsh.environ import Env
 from xonsh.tools import ON_WINDOWS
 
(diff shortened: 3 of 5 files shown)


gforsyth:
Looks awesome, @melund !  Seems to be having trouble on Linux, though:

```console
gil@bad_cat ~/git/xonsh pr/901 $ which python
Exception in thread Thread-38:
  File "/home/gil/miniconda3/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/home/gil/git/xonsh/xonsh/proc.py", line 173, in run
    r = self.f(self.args, sp_stdin, sp_stdout, sp_stderr)
  File "/home/gil/git/xonsh/xonsh/aliases.py", line 389, in which
    exts=exts, verbose=pargs.verbose):
  File "/home/gil/git/xonsh/xonsh/xoreutils/_which.py", line 209, in whichgen
    "platform '%s'" % sys.platform)
xonsh.xoreutils._which.WhichError: 'exts' argument is not supported on platform 'linux'
```

melund:
@gforsyth ... Stupid oversight. I better look at the Docker thing, so I can also test stuff on Linux. Try now.

scopatz:
Hey @melund! Thanks for putting this in.  Also, doesn't it need to be `xoreutils/__init__.py`, rather than `xoreutils/init.py`?  Also don't you need to add some lines to the setup.py to let it know about the subpackage?

melund:
Yes... I finished it late last night :)

melund:
@scopatz ... Fixed now

scopatz:
Hey @melund, this looks good to me! Would you mind adding a simple test, though please? I have noticed that there is now at least one test failure on master locally for me, and so we should probably get better at asking for tests on the interface side of things too.  Also out of curiosity, did the which module that you got from ActiveState have any tests with it that we could use?

melund:
You are right. We ought to better at tests - Just hate nosetest. The module actually had some tests I will port them as well.

scopatz:
> You are right. We ought to better at tests - Just hate nosetest.

Not to derail the thread too much, but is there something that you prefer? Happy to consider other options.

melund:
If you ask I much favor pytest. You can ask @asmeurer about it. I think it was also his favorite. 

But I also think would be a foolish task to convert to another framework at this point. 

Do we have any examples on how to test aliases commands?  I want to run the xonsh code line and record the  stdout, stdin and return code?

scopatz:
> If you ask I much favor pytest. You can ask @asmeurer about it. I think it was also his favorite.

Ahh Ok, that was my guess.

> But I also think would be a foolish task to convert to another framework at this point.

Well I think it might be worth it, especially if it encourages test writing. I've been told that pytest is nose-compaible, though I don't know what that means exactly.

> Do we have any examples on how to test aliases commands? I want to run the xonsh code line and record the stdout, stdin and return code?

No not yet. We really don't have a great way to test xonsh output based on what a user would see when typing in code.  That is definitely beyond this. As long as the _which.py has tests you can port, I am happy :)

melund:
Rebased on master

melund:
>Well I think it might be worth it, especially if it encourages test writing. I've been told that pytest is nose-compatible, though I don't know what that means exactly.

I think it is nose compatible for the simple stuff. But some our tests use some more fancy stuff.

A firs step could be to simplify this test. https://github.com/scopatz/xonsh/blob/master/tests/test_tools.py#L410 
It is responsible for generating the vast majority of the test. 

The test code for `which.py` uses a unittest setup with subprocess calls to test its own parser interface. We don't use that part. If I can easily port the whole thing I will. Otherwise, I will replicate the test for the parts we do use.

scopatz:
> A firs step could be to simplify this test. https://github.com/scopatz/xonsh/blob/master/tests/test_tools.py#L410 
It is responsible for generating the vast majority of the test.

Yeah, that thing has got to be cut down to something more reasonable.

> The test code for which.py uses a unittest setup with subprocess calls to test its own parser interface. We don't use that part. If I can easily port the whole thing I will. Otherwise, I will replicate the test for the parts we do use.

Sounds great! I am sure an easier solution is better and enough.

gforsyth:
@melund -- this is awesome!  

```console
(xonshtest)gil@theo~/git/xonsh pr/901$ which -a ls mkdir
ls -> ['ls', '--color=auto', '-v']
/usr/bin/ls
mkdir -> ['mkdir', '-p']
/usr/bin/mkdir
```

:smiley:

melund:
Yes. The https://github.com/trentm/which is a nice packages. The tests are not easily portable, but I can just write them anew

melund:
@scopatz @gforsyth . I have written some tests. It tests `whichgen()` that is the function from `_which.py` we are using. It doesn't test the actual `which` alias. That is a shame since it also contains a lot of logic as well as the arg parser. It is also the place we are likely to edit in the future :(

BYK:
I have two questions:

1. Is there a way to use this package from PyPI and list it as a dependency instead of embedding its code here?
2. Do you think we can replace all `locate_binary` calls with this and remove our custom implementation from the code base since this is essentially duplicated functionality (and code)

gforsyth:
Hey @BYK -- 

1. I don't think so.  I know @melund made a number of modifications to both update it to Python 3 and I believe also remove some things that weren't working with `xonsh`
2. Yes, I think so.

BYK:
@gforsyth - so shall we fork the original project and publish it ourselves then? This would be a good contribution to the community. We can probably do the same thing for the Windows `sudo` extension. That said this implies having a `xonsh` organization (or we can do these under our own personal accounts but that's not my preference).

//cc @scopatz

gforsyth:
With the incorporation of PLY into the code base we are now at zero external (required) dependencies outside of a base Python 3.4+ install and I'll bet @scopatz would like to keep it that way.  
That said, it's always good to give back to the community -- I think there's a case to be made for starting a `xonsh` organization, splitting of a few packages like `which` and `sudo` and then using something like `subtree` to pull them back into `xonsh` proper.  
But please, please, no submodules.

scopatz:
I think it is a great idea to put the forked versions back out there into the world.  Perhaps the first thing to do would be to contact the original developers and see if they would be interested in the changes?  Maybe that is more applicable to which than sudo? Even if we do this, though, I agree with @gforsyth that we should keep the code in xonsh itself and port it out as desired. This is in order to keep xonsh with minimal dependencies.

In any event, I have often said I am happy to move stuff over to a xonsh org. I even snagged the xonsh org a while ago to make sure that we had it :). I have added all of you as owners. 

That said, I think *right now* is a pretty bad tine to move the mainline xonsh repo.  I'd like to wait until after PyCon if it is all the same to you.  However, moving the docs now is probably fine, if someone wants to do that and make the relevant PR updating xonsh itself too.

scopatz:
Thanks a million @melund!

BYK:
>Even if we do this, though, I agree with @gforsyth that we should keep the code in xonsh itself and port it out as desired. This is in order to keep xonsh with minimal dependencies.

I kind of disagree with you two on this. I totally understand the value of fewer dependencies but we already need prompt-toolkit and pygments for the "Best" experience of Xonsh. Them being optional is actually a bad end-user experince for the first install but that's a separate issue. I think it should be okay to depend on modules, especially we are the ones maintaining it.

>That said, I think right now is a pretty bad tine to move the mainline xonsh repo. I'd like to wait until after PyCon if it is all the same to you. However, moving the docs now is probably fine, if someone wants to do that and make the relevant PR updating xonsh itself too.

Totally agree. We should probably start a "after v0.3" milestone (1.0?) or "after PyCon" milestone to put these tickets into.

adqm:
> need prompt-toolkit and pygments for the "Best" experience of Xonsh. 

This is debatable :)

I always use readline.  Maybe I'm just old-fashioned...

melund:
If we made an independent package, we should still do what we did for PLY, to keep it without dependencies. 

> I kind of disagree with you two on this. I totally understand the value of fewer dependencies but we already need prompt-toolkit and pygments for the "Best" experience of Xonsh. Them being optional is actually a bad end-user experince for the first install but that's a separate issue. I think it should be okay to depend on modules, especially we are the ones maintaining it.

I am with @scopatz on this one. We can have dependecies with package managers that handles it well. Conda is a good example. The default pyhon installers is not the place to add depencies for a tool like xonsh. 

I agree that forking the `which`project and updating to Python3 would be a nice thing to do. I think we could fork the `which` project without anyone being upset. It hasn't been maintained for 7 years. It supports python 2.3 ;)  Changes to support Python3 have been proposed, without a reaction.

I don't know how it would work, I guess we can't really push it to Pypi, since the `which` name is already reserved. 

Also keep in mind it would mean we would have to support the part of the code which we don't use in xonsh.

scopatz:
It could be called which3 or xonsh-which or something, @melund.  Not super worried about that :).  In any event, if someone wants to do that feel free! It is a pretty low priority for me personally :(

scopatz:
Though I think it would be nice to do.

adqm:
Piggybacking on this thread since I didn't know where else to go (and this is where some of the discussion of the xonsh organization lives).

Just a thought: should we transfer the main xonsh repo to be owned by that group?

scopatz:
> Just a thought: should we transfer the main xonsh repo to be owned by that group?

Yes, I think that there is consensus that this is probably the right thing to do and I am happy to do it. I'd prefer if we held off until after PyCon, though, because I think there are a lot of little places to cover and it would be beneficial to me to not have to worry about it before a big talk.

melund:
I really didn't mind if it continued to live under "scopatz/xonsh". What are the benefits of putting it under "xonsh/xonsh" ?

scopatz:
Well, I think there are some practical advantages in the sense that folks other than me can change repo permissions, set up CI hooks, and those kinds of things.

PR #902, opened 2016-05-15
Adds a xonsh.platform module

funkyfuture:
for Linux it adds `distro` as dependency as
https://docs.python.org/3.5/library/platform.html#platform.linux_distribution
is deprecated.

also comes with occasional code cleanup and micro optimizations


funkyfuture:
okay, i can't make a rebase onto master without risking a nervous breakdown due to my issue with git and `xonsh.bat`. the tests are passing. it's not refactored fully anyway, so i leave it for review.

xonsh/platform.py
```diff
@@ -0,0 +1,77 @@
...
+    globals()['ON_' + name.upper()] = platform.system() == name
+
+
+#
+# Linux distro
+#
+
+if ON_LINUX:
+    import distro
+    ON_LINUX = distro.id()
```
funkyfuture: ... `or True`

melund:
Are there other ways around adding a new dependency?

As a side note. `Pretty.py` is copied from the IPython project, so there is no need to change that file. It will just be overridden next time we pull in updates.

funkyfuture:
the according issue on python.org points to this module. i see no sense in reimplementing it. it could be added as a git submodule.

gforsyth:
Many of us are not fans of `git-submodule`.  
As of now, the only distro we need to know about is Arch (or manjaro) since they use a different location for their bash completion files.  Rather than add in a dependency, we can either:

1. Take out the distro-check and just offer a list of possible configurations in the setup wizard for a user to select from
2. `platform.platform()` returns the kernel and usually has the name of the distro included.  For instance, Arch gives
```python
>>> platform.platform()
'Linux-4.5.3-1-ARCH-x86_64-with-arch'
```
and Ubuntu offers
```python
>>> platform.platform()
'Linux-3.19.0-58-generic-x86_64-with-Ubuntu-14.04-trusty'
```

so we can just check with an `'arch' in platform.platform()`

funkyfuture:
> Arch (or manjaro)

Manjaro has no completions in `/etc`. it's fine with the Linux defaults.

if it's only about the bash completion's locations, these can be located heuristically by testing common paths.

xonsh/aliases.py
```diff
@@ -1,28 +1,27 @@
 # -*- coding: utf-8 -*-
 """Aliases for the xonsh shell."""
 
+from argparse import Action, ArgumentParser
+import builtins
+from collections.abc import MutableMapping, Iterable, Sequence
+from functools import lru_cache
```
scopatz: Any reason for this reordering?

funkyfuture: it's easier to find items in sorted lists.

scopatz: I think it is nicer to have them in length order, like before.  I guess this doesn't block the PR, but I'll probably change it back eventually :)

xonsh/tools.py
```diff
@@ -43,34 +49,20 @@ def pygments_version():
...
-    string_types = (str, unicode)
-    unicode_type = unicode
 
-try:
-    import win_unicode_console
-except ImportError:
-    win_unicode_console = None
+IS_ROOT = ctypes.windll.shell32.IsUserAnAdmin() != 0 if ON_WINDOWS else os.getuid() == 0
 
+# TODO move the constants to .platform
```
scopatz: Please remove TODO

xonsh/tools.py
```diff
@@ -17,19 +17,25 @@
...
 import threading
-import subprocess
-from contextlib import contextmanager
-from collections import OrderedDict, Sequence, Set
+import traceback
 from warnings import warn
 
+from xonsh.platform import win_unicode_console
+
+# compatibility imports, TODO remove eventually
```
scopatz: Please remove TODO

xonsh/tools.py
```diff
@@ -633,17 +617,17 @@ def bool_seq_to_csv(x):
 
 
 def is_completions_display_value(x):
-    return x in {'none', 'single', 'multi'}
+    return x in ('none', 'single', 'multi')
```
scopatz: Please keep this as a set, since the uniqueness guarantee is  important.

xonsh/tools.py
```diff
@@ -633,17 +617,17 @@ def bool_seq_to_csv(x):
...
 
 
 def to_completions_display_value(x):
     x = str(x).lower()
-    if x in {'none', 'false'}:
+    if x in ('none', 'false'):
```
scopatz: Same here

xonsh/tools.py
```diff
@@ -633,17 +617,17 @@ def bool_seq_to_csv(x):
...
         x = 'none'
-    elif x in {'multi', 'true'}:
+    elif x in ('multi', 'true'):
```
scopatz: Same here.

xonsh/tools.py
```diff
@@ -595,19 +584,16 @@ def ensure_int_or_slice(x):
...
+            all(isinstance(a, str) for a in x))
 
 
 def csv_to_set(x):
     """Convert a comma-separated list of strings to a set of strings."""
     if not x:
         return set()
     else:
-        return set(x.split(','))
+        return set(y.strip() for y in x.split(','))
```
scopatz: I think this makes it invalid CSV.  Isn't whitespace preserved in CSV?

xonsh/platform.py
```diff
@@ -0,0 +1,79 @@
...
+ON_DARWIN = ON_LINUX = ON_WINDOWS = None
+for name in ('Darwin', 'Linux', 'Windows'):
+    globals()['ON_' + name.upper()] = platform.system() == name
+
+
+#
+# Linux distro
+#
+
+if ON_LINUX and '-ARCH-' in platform.platform():
```
scopatz: I would prefer if all of the `ON_*` variables were bools and then if arch or version info can be stored in other variable names.

xonsh/platform.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+ON_ANACONDA = any(s in sys.version for s in ('Anaconda', 'Continuum'))
+
+
+#
+# OS
+#
+
+ON_DARWIN = ON_LINUX = ON_WINDOWS = None
```
scopatz: Maybe set to `False` rather than `None`

xonsh/platform.py
```diff
@@ -0,0 +1,79 @@
...
+for name in ('Darwin', 'Linux', 'Windows'):
+    globals()['ON_' + name.upper()] = platform.system() == name
```
scopatz: Maybe just set these normally, rather than going through globals() while at global scope :)

xonsh/environ.py
```diff
@@ -558,17 +555,14 @@ def get_ensurer(self, key,
...
         if key in self.ensurers:
             return self.ensurers[key]
-        for k, ensurer in self.ensurers.items():
-            if isinstance(k, string_types):
-                continue
-            m = k.match(key)
-            if m is not None:
-                ens = ensurer
+        for k, ensurer in ((k, v) for k, v in self.ensurers.items()
+                           if not isinstance(k, str)):
```
scopatz: No reason to double loop, please revert. Thanks!

xonsh/environ.py
```diff
@@ -649,16 +643,14 @@ def __setitem__(self, key, val):
...
+                    os.environ[key] = ensurer.detype(val)
 
     def __delitem__(self, key):
         val = self._d.pop(key)
         if self.detypeable(val):
             self._detyped = None
-            if self.get('UPDATE_OS_ENVIRON'):
-                if key in os.environ:
-                    del os.environ[key]
+            if self.get('UPDATE_OS_ENVIRON') and key in os.environ:
```
scopatz: Thanks!

xonsh/environ.py
```diff
@@ -1116,7 +1108,7 @@ def partial_format_prompt(template=DEFAULT_PROMPT, formatter_dict=None):
         if field is None:
             continue
         elif field.startswith('$'):
-            v = builtins.__xonsh_env__[name[1:]]
+            v = builtins.__xonsh_env__[name[1:]]  # FIXME `name` is an unresolved ref
```
scopatz: Ahh Thanks.  It would be great it if you could make an issue for this please!

xonsh/environ.py
```diff
@@ -1201,6 +1193,8 @@ def load_static_config(ctx, config=None):
 def xonshrc_context(rcfiles=None, execer=None):
     """Attempts to read in xonshrc file, and return the contents."""
     loaded = builtins.__xonsh_env__['LOADED_RC_FILES'] = []
+    datadir = builtins.__xonsh_env__['XONSH_DATA_DIR']
+    cachedir = os.path.join(datadir, 'xonshrc_cache')
```
scopatz: It doesn't look like either of these two are actually used...

xonsh/environ.py
```diff
@@ -156,15 +158,8 @@ def xonshconfig(env):
...
-                             '/opt/local/etc/profile.d/bash_completion.sh')
-                        if ON_MAC else
-                        ('/usr/share/bash-completion/bash_completion',
-                             '/usr/share/bash-completion/completions/git')
-                        if ON_ARCH else
-                        ('/etc/bash_completion',
-                             '/usr/share/bash-completion/completions/git')),
-    'CASE_SENSITIVE_COMPLETIONS': ON_LINUX,
+    'BASH_COMPLETIONS': BASH_COMPLETIONS_DEFAULT,
+    'CASE_SENSITIVE_COMPLETIONS': bool(ON_LINUX),
```
scopatz: It would be great if `ON_LINUX` were just a bool

xonsh/tools.py
```diff
@@ -521,19 +513,16 @@ def always_false(x):
 
 def ensure_string(x):
     """Returns a string if x is not a string, and x if it already is."""
-    if isinstance(x, string_types):
-        return x
-    else:
-        return str(x)
+    return str(x)
```
scopatz: Please return to the original.  No need to copy the string if it is already a string.  You should get the same object back.

xonsh/tools.py
```diff
@@ -583,7 +572,7 @@ def ensure_int_or_slice(x):
     """Makes sure that x is list-indexable."""
     if x is None:
         return slice(None)
-    elif is_int(x):
+    elif isinstance(x, int):
```
scopatz: Please use `is_int()` since it gives us more control.

xonsh/tools.py
```diff
@@ -595,19 +584,16 @@ def ensure_int_or_slice(x):
 
 def is_string_set(x):
     """Tests if something is a set"""
-    if isinstance(x, string_types):
-        return False
-    else:
-        return (isinstance(x, Set) and
-                all([isinstance(a, string_types) for a in x]))
+    return (isinstance(x, Set) and
+            all(isinstance(a, str) for a in x))
```
scopatz: Thanks!

scopatz:
Hey @funkyfuture! Thanks a million for putting this in.  I have tried to go through and point out places where I think things could be improved or reverted. Let us know when it is updated.

I think the only major change that I would like to see is that the `ON_` variables are all bools, while other variables would store info about the actual platform. This makes it easier (for me at least) to reason about things.

Other than that, thanks again! This will be great to have!

xonsh/aliases.py
funkyfuture: interesting aesthetics. :-) imho it's messy.

scopatz: Yeah, we all have our preferences :)

xonsh/environ.py
funkyfuture: imo `ON_ARCH` is polluting the namespace. what about a `LINUX_DISTRO`?

xonsh/platform.py
funkyfuture: this statement is actually there for inspections like from IDEs, the values are set anyway.

xonsh/environ.py
scopatz: `LINUX_DISTRO` sounds great to me.

xonsh/platform.py
scopatz: But can you change to False rather than None, please? The IDEs don't need a None, do they?

xonsh/platform.py
funkyfuture: but with four items we had a deal? ;-)

xonsh/tools.py
funkyfuture: not until i did it.

xonsh/platform.py
scopatz: Hmm? I am not sure I understand

xonsh/tools.py
funkyfuture:
``` python
>>> id(a)
139923335929960
>>> id(str(a))
139923335929960
```

it does.

xonsh/tools.py
scopatz: Well, please make an issue and remove the TODO.

xonsh/tools.py
funkyfuture: actually this was the first time i encountered python code that casts to strings like this.

xonsh/tools.py
funkyfuture: about what exactly? i'm currently working on a patch that allows more versatile to define ensurer functions and constraints.

xonsh/tools.py
funkyfuture: thanks for spotting. though i'd say that config values have hardly leading or trailing whitespaces.

xonsh/tools.py
scopatz: Interesting. I can't find anything in the Python docs that say this is guaranteed. Do you know of any?

xonsh/tools.py
funkyfuture: uniqueness is guaranteed. if one wrote `('none', 'single', 'none')` here that'd be an obvious bug. an immutable saves some resources.

xonsh/tools.py
scopatz: Also I know of other weird string cases due to lack of interning:

```python
scopatz@localhost ~ $ x = "x" * 10000
scopatz@localhost ~ $ y = "x" * 10000
scopatz@localhost ~ $ x is y
False
scopatz@localhost ~ $ id(x) == id(y)
False
scopatz@localhost ~ $ x == y
True
```

xonsh/platform.py
funkyfuture: i mean, what you refer to as 'normal' assignment is one code line more than this. it's three lines with a repeated pattern. which isn't reasonable with a higher amount of items.
discussion postponed until Commodore 64s are supported.

xonsh/tools.py
scopatz: Then I would suggest a frozenset :)

xonsh/platform.py
funkyfuture: the thinking is that `None` means they haven't been detected yet. but it doesn't matter. one could also assign a cosmonauts name here.

xonsh/tools.py
scopatz: Well, if we ever need to change `is_int()` for whatever reason, that change will be propogated here. That was my only concern. 

>  i'm currently working on a patch that allows more versatile to define ensurer functions and constraints.

Cool! I look forward to it.

xonsh/platform.py
scopatz: Maybe a comment to that effect would be helpful :)

xonsh/platform.py
scopatz: Ahh, Well, I am not sure line count is something that we should be concerned with. I think the cleverness of this pattern isn't worth the readability hit, personally.

scopatz: Though I do appreciate its cleverness :)

xonsh/tools.py
funkyfuture: i'd say that originates in Python's way to store values in hashing tables. a long time ago i read an article discussing this, but can't remember what it was.

```
>>> id('x' *10000)
28875296
>>> id(str('x' *10000))
28883376
```

why would a `y` matter here?

xonsh/tools.py
scopatz: Sure, but it would be incorrect to call it CSV then.  I think that there are only a couple of places in the code base that use this. If you want to change the names and the docs to reflect the strip, that is probably fine.

xonsh/tools.py
scopatz: Sure, but strings are not interned by default, so I believe that `str(x) is x` happens to be an optimization in the CPython interpreter, rather than a key feature of the str constructor. So potentially running xonsh on other interpreters could change the behaviour.

xonsh/xonfig.py
```diff
@@ -315,15 +316,14 @@ def _format_json(data):
...
-        ('Python', '.'.join(map(str, tools.VER_FULL))),
+        ('Python', '.'.join(map(str, PYTHON_VERSION))),
         ('PLY', ply.__version__),
         ('have readline', is_readline_available()),
         ('prompt toolkit', prompt_toolkit_version() if \
                            is_prompt_toolkit_available() else False),
         ('pygments', tools.pygments_version()),
         ('on posix', tools.ON_POSIX),
         ('on linux', tools.ON_LINUX),
-        ('on arch', tools.ON_ARCH),
```
scopatz: Probably should add the `LINUX_DISTRO` here.

xonsh/tools.py
funkyfuture:
```
$ timeit "1 in (1,2,3)"
10000000 loops, best of 3: 43.3 ns per loop
$ timeit "1 in {1,2,3}"
10000000 loops, best of 3: 70.7 ns per loop
$ timeit "1 in frozenset((1,2,3))"
1000000 loops, best of 3: 497 ns per loop
$ timeit "1 in frozenset({1,2,3})"
1000000 loops, best of 3: 523 ns per loop
```

xonsh/platform.py
funkyfuture: actually, when i encountered such the first time was embarressed and fascinated at the same time. but it feels differntly writing it oneself.

xonsh/tools.py
funkyfuture: it's used for user-entered data who tend to put an extra space after a comma.

oh, and are sets used here with the rational for uniqueness too? i would porpose another approach to this with the mentioned ensurer enhancement.

xonsh/tools.py
funkyfuture: though [here](https://wiki.python.org/moin/PythonSpeed#Use_the_best_algorithms_and_fastest_tools) it says that sets are faster for membership lookups. maybe that rather apllies to large data.

melund: Yes. In your test you are including the time it takes to construct the object. It much faster to look stuff up in hash-tables than iterables.

funkyfuture:
okay, it's updated. certainly worth a review to catch what i missed.

xonsh/platform.py
```diff
@@ -0,0 +1,141 @@
+""" Module for platform-specific constants and implementations """
+
+import os
+import platform
+import sys
+
+try:
+    import distro
```
scopatz: Please list distro as an optional dependency in the docs / readme. Namely here: https://github.com/scopatz/xonsh/blob/master/docs/dependencies.rst

xonsh/platform.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+#
+# Windows
+#
+
+try:
+    import win_unicode_console
+except ImportError:
+    win_unicode_console = None
```
scopatz: Please put this try-except in an `if ON_WINDOWS` block.  It will make it faster to startup on non-Windows platforms.

xonsh/platform.py
```diff
@@ -0,0 +1,141 @@
...
+try:
+    import readline
+except ImportError:
+    HAS_READLINE = False
+else:
+    HAS_READLINE = False
+
+
+try:
+    import prompt_toolkit
```
scopatz: Hmm, I think that trying to import both readline and ptk, when normally only one will be used is pretty inefficient, especially if you happen to have both.  Please modify this to only compute these as needed.  Maybe these need to be descriptors, or be returned to functions.

xonsh/platform.py
```diff
@@ -0,0 +1,141 @@
...
+    BASH_COMPLETIONS_DEFAULT = \
+        ('/usr/share/bash-completion',
+         '/usr/share/bash-completion/completions')
+elif ON_DARWIN:
+    BASH_COMPLETIONS_DEFAULT = \
+        ('/usr/local/etc/bash_completion',
+         '/opt/local/etc/profile.d/bash_completion.sh')
+else:
+    BASH_COMPLETIONS_DEFAULT = \
+        ()
```
scopatz: This can be on the previous line. Also there is no reason to use backslashes here since you can just use and opening paren.

xonsh/tools.py
scopatz: I'd feel a lot better if this was back to what it was.

xonsh/tools.py
scopatz: Same here, please revert.

xonsh/tools.py
scopatz: Yes, please revert these to sets or frozensets

xonsh/tools.py
scopatz: > it's used for user-entered data who tend to put an extra space after a comma.

sure, which is fine, if you want to give it a new name and change the ensurer to be something more useful. My problem is only that it is named CSV but you are proposing to change it to something that is not CSV.  So either the name can change or it should remain CSV compliant.  I don't personally mind too much of either of these :).

> oh, and are sets used here with the rational for uniqueness too? i would porpose another approach to this with the mentioned ensurer enhancement.

IIRC the sets were for uniqueness here. Please feel free to revert this change and address any improvements in another PR.  No reason this all has to be done in this one. I think it slows this one down :)

scopatz:
Thanks again @funkyfuture - I have gone through the code base again.  I think that it would be much better / faster to revert the ensurer changes and just proceed with the core part of this PR, which is for the platform changes.  In general I find smaller PRs better, and I don't mind having lots of little PRs.  These big ones are more prone to delays simply because they are large.

NOTE: sorry this message was delayed because of the bad wifi on the plane.

xonsh/platform.py
funkyfuture: well, it's more performant than the functions. the `HAS_READLINE` is only used with `xonfig info`, is that really necessary? in which context is the `readline` module not availabe?  the Python docs don't mention that it isn't available on any platform.

xonsh/tools.py
funkyfuture: PyPy also behaves like this. if there was a platform that actually didn't, it could be provided by the `platform` module.
my concern is the overhead of function calls and instance checks that effectively result the same like a `str(x)`. the function was used quiet often.

btw, when @GuillaumeLeclerc's Docker script is merged i will submit one to test against different interpreters (it has proven while developing `cerberus`). do you have any othe on mind than CPython 3.4 & 3.5? PyPy3 implements only 3.0 oder 3.2, so that would mean backporting.

scopatz:
Looks like there is now a merge conflict here @funkyfuture :)

xonsh/platform.py
scopatz: The Python readline situation is really weird. [This note](https://docs.python.org/3/library/readline.html#module-readline) in the docs only starts to get at that.  But basically, on  Mac OSX readline is secretly replaced by libedit. On Windows, I am not sure it exists at all unless you pip install gnureadline.  Then I think sometimes the readline module from Python tries to spoof the input() behaviour with stdin, stdout, and stderr that you would get with readline manually.  

Basically, it is impossible to be sure that you actually have a true readline without importing it, which is why all of our `import readline` statements are in try-except blocks.

In any event, I think that it is important to only import the shell interface that will get used.  The exception to this rule is, as you noted, when `xonfig` is run and we are trying to figure out what a user actually has on their system.

funkyfuture:
rebased. 

while reviewing i discovered a bug in the `tools.CommandCache` (looking up dirs instead of files). i noticed that the tests involving completions discovery behave differently locally (passed) than on travis (failed).

adqm:
I am not seeing any glaring bugs in `CommandsCache`.  It looks like it is doing the right thing to me?

Though I guess line 1132 should probably be: `allcmds |= set(i for i in os.listdir(d) if not os.path.isdir(i))`
and maybe it should have more checking than that, even, to make sure the files are actually executable.  I'm happy to put that in.

Is that what you're referring to?  Or am I missing something else?

funkyfuture:
that is what i changed [here](https://github.com/scopatz/xonsh/pull/902/files#diff-ece06b292b3c59cc7bfc962fe8a6b856L1132).

adqm:
Oh, perfect.  Sorry I didn't see that in the diff.

xonsh/platform.py
```diff
@@ -0,0 +1,147 @@
...
+try:
+    import pygments
+except ImportError:
+    HAS_PYGMENTS, PYGMENTS_VERSION = False, None
+else:
+    HAS_PYGMENTS, PYGMENTS_VERSION = True, pygments.__version__
+
+
+try:
+    import prompt_toolkit
```
scopatz: This also should be a function.  We shouldn't import ptk if we know we are going to use readline.

xonsh/tools.py
```diff
@@ -17,69 +17,32 @@
...
-HAVE_PYGMENTS = bool(pygments_version())
-
-VER_3_4 = (3, 4)
-VER_3_5 = (3, 5)
-VER_3_5_1 = (3, 5, 1)
-VER_FULL = sys.version_info[:3]
-VER_MAJOR_MINOR = sys.version_info[:2]
-V_MAJOR_MINOR = 'v{0}{1}'.format(*sys.version_info[:2])
+IS_SUPERUSER = ctypes.windll.shell32.IsUserAnAdmin() != 0 if ON_WINDOWS else os.getuid() == 0
+FRMT_VERSION_MAJOR_MINOR = 'v{0}{1}'.format(*PYTHON_VERSION_INFO[:2])
```
scopatz: Maybe call this `PY_VERSION_MAJOR_MINOR`?

xonsh/tools.py
scopatz: OK, well PyPy was my main concern.  So I suppose this is reasonable.

scopatz: But a code comment to this effect would be very nice, please.

scopatz:
This is looking great @funkyfuture! Just a few small issues. I think the only thing really bothering me right now is the prompt-toolkit import that I think should be lazy.

scopatz:
PS, it would be great to get this in for the release tomorrow, if you have the time today.

funkyfuture:
i'd like to abandon the `ensure_string` with the refactoring of the ensurers.

scopatz:
> i'd like to abandon the ensure_string with the refactoring of the ensurers.

Understood, but that can be done in another PR :).  Is this ready for another look?

funkyfuture:
> Is this ready for another look?

nope, anticipating the conflicts with this PR i'm drinking tea on this.

i can find no use of  `tools.docstring_by_version`. is that obsolete?

any idea why the import is necessary again in `platform.ptk_version`? by that time the module should be availabe in the globals.

scopatz:
> nope, anticipating the conflicts with this PR i'm drinking tea on this.

And there are the conflicts.... :)

> i can find no use of tools.docstring_by_version. is that obsolete?

Yeah, that was from a previous iteration of the parser. Feel free to remove it.

> any idea why the import is necessary again in platform.ptk_version? by that time the module should be availabe in the globals.

Now that it is a function, the ptk isn't available in the globals

funkyfuture:
> And there are the conflicts.... :)

of course. as this pr affects many files, chances are high when something else gets merged.

> Now that it is a function, the ptk isn't available in the globals

but it calls `has_prompt_toolkit` which imports it.

melund:
> but it calls has_prompt_toolkit which imports it.

I think it is because you are inside a local namespace in the function. After you exit the function the `prompt_toolkit` name will not be available to the caller. prompt_toolkit.

funkyfuture:
yep, you're right.

i have no clue what's going on with the 3.5 tests on travis. locally tests are passing and it's usable.

it'd be great if it was merged soon as repeated rebasing becomes more and more of a pita.

scopatz:
I'm looking into this now

scopatz:
I can confirm that the tests work locally.

scopatz:
I am also seeing a simple bug on this branch: 

```console
scopatz@localhost ~/xonsh funkyfuture-platform $ xonfig
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/proc.py", line 335, in wrapped_simple_command
    r = f(args, i)
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/aliases.py", line 449, in xonfig
    def trace(args, stdin=None):
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/xonfig.py", line 455, in main
    return _MAIN_ACTIONS[ns.action](ns)
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/xonfig.py", line 336, in _info
    ('is superuser', tools.IS_SUPERUSER),
AttributeError: module 'xonsh.tools' has no attribute 'IS_SUPERUSER'
```

scopatz:
This seems like it could be a legitimate issue since locally I am running v3.5.1, but the CI is running v3.5.0 and there are some changes here w.r.t the micro version number.

xonsh/parser.py
```diff
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
 """Implements the xonsh parser."""
-from xonsh.tools import (VER_3_4, VER_3_5, VER_MAJOR_MINOR)
+from xonsh.platform import PYTHON_VERSION_INFO
 
-if VER_MAJOR_MINOR <= VER_3_4:
+if PYTHON_VERSION_INFO < (3, 5, 0):
```
scopatz: This is the problem.  This will use the Python v3.4 parser on v3.5.0

xonsh/platform.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+
+#
+# Python & packages
+#
+
+PYTHON_VERSION_INFO = sys.version_info[:3]
+
+if PYTHON_VERSION_INFO < (3, 5, 0):
```
scopatz: This is probably also broken.

xonsh/inspectors.py
```diff
@@ -300,7 +295,7 @@ def find_source_lines(obj):
     return lineno
 
 
-if VER_MAJOR_MINOR <= VER_3_4:
+if PYTHON_VERSION_INFO < (3, 5, 0):
```
scopatz: Also broken

xonsh/tokenize/__init__.py
```diff
@@ -1,6 +1,6 @@
-from xonsh.tools import VER_3_5, VER_FULL
+from xonsh.platform import PYTHON_VERSION_INFO
 
-if VER_FULL >= VER_3_5:
+if PYTHON_VERSION_INFO > (3, 5, 0):
```
scopatz: probably broken.

funkyfuture: this one was it. thanks.

scopatz:
I am seeing noticeably longer startup times on the branch, about an extra 75% - 100%

xonsh/tools.py
```diff
@@ -17,73 +17,31 @@
...
-        v = pygments.__version__
-    except ImportError:
-        v = False
-    return v
-
-if sys.version_info[0] >= 3:
-    string_types = (str, bytes)
-    unicode_type = str
+if has_prompt_toolkit():
+    import prompt_toolkit
```
scopatz: This should not be called here.

scopatz: Though, this might have been there before, now that I look at it.

scopatz:
Ok I have determined that most of the slow down is not this branches fault per se. My initial estimates were wrong.

scopatz:
Ok @funkyfuture - I have gone ahead and merged this.  There were several bugs on this branch that I fixed along the way.  Thanks a ton for the monumental effort! However, please put in much smaller PRs if you can.

funkyfuture:
for smaller commits the compatibility imports in tools would have been useful.

can you point me to what was wrong with `scandir`? patching `pathutils.Path` wasn't certainly the best idea. however it's highly advisable to use Python 3.5's `os.scandir` as it performs about 6x faster than `os.listdir`. i could do a proper port of the `os.DirEntry` api for Python 3.4.

scopatz:
Yeah, the problem was that the Path and DirEntry APIs are not actually all that similar.  The main issue was that Path does not implement a `.path` attribute/property on Python 3.5. This was causing the completer to fail locally for me (even in tests).  

While I think that backporting scandir() is a good idea (because of the performance reasons), we can't cheat and not also backport DirEntry :).  Doing a proper backport is what I suggest.  I removed it so that I could merge and we could proceed with the release.  This was an emergency measure and I hope you don't mind.  I figured it would be better to get this PR merged in some state, rather than not having it at all.

funkyfuture:
don't worry, i agree with the measures.

do you mean to include [scandir](https://github.com/benhoyt/scandir)? on that occasion: what's the problem with git submodules?

my proposal is to only backport `DirEntry` fully api-compatible, that suffices imo.

scopatz:
> do you mean to include scandir? on that occasion: what's the problem with git submodules?

Nope, I didn't know about that project.

> my proposal is to only backport DirEntry fully api-compatible, that suffices imo.

Yep, this is what I was suggesting :)

funkyfuture:
> Nope, I didn't know about that project.

it's the implementation that was included in 3.5. so, with 3.4 one would have the same rocket.

scopatz:
Gotcha, cool

PR #903, opened 2016-05-15
fish-like (Fuzzy) completion

GuillaumeLeclerc:
For the moment I applied the fuzzy search to the files only.

Applying it to python completions was leading to too many matches.

But maybe we could apply it too git completions or commands.

Tell me what you think.

_Complexity is `O(n)` for each possible completion_

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -316,10 +316,11 @@ class Completer(object):
     def _add_env(self, paths, prefix):
         if prefix.startswith('$'):
             csc = builtins.__xonsh_env__.get('CASE_SENSITIVE_COMPLETIONS')
-            startswither = startswithnorm if csc else startswithlow
             key = prefix[1:]
             keylow = key.lower()
-            paths.update({'$' + k for k in builtins.__xonsh_env__ if startswither(k, key, keylow)})
+            paths.update({'$' + k for k in builtins.__xonsh_env__ if 
+                fuzzyMatch(k, key) or 
+                (csc and fuzzyMatch(k.lower(), keylow))})
 
     def _add_dots(self, paths, prefix):
         if prefix in {'', '.'}:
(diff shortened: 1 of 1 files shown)


xonsh/completer.py
```diff
@@ -84,13 +84,23 @@ def _path_from_partial_string(inp, pos=None):
 
 WS = set(' \t\r\n')
 
+def fuzzyMatch(ref, typed):
```
scopatz: PEP8 function names

xonsh/completer.py
```diff
@@ -391,14 +402,19 @@ def path_complete(self, prefix, start, end, cdpath=False):
         """Completes based on a path name."""
         tilde = '~'
         paths = set()
+        sep = get_sep()
+        prefix_parts = prefix.split(sep)
+        prefix_parts[-1] = ""
+        search_pattern = sep.join(prefix_parts)
```
scopatz: I think this could be better achieved with an rpartition than a split.

scopatz:
Thanks a million for putting this in, @GuillaumeLecler! Could you please put in an entry to the CHANGELOG and some tests. Also making this PEP8 compliant is important too.  Thanks!

GuillaumeLeclerc:
I know PEP8 is important but I come (and do on a daily basis) languages that have very different coding conventions. It's hard for me to do the switch multiple times a day. I will change them and try to be more careful.

xonsh/completer.py
GuillaumeLeclerc: I don't know about this. I will investigate and use it.

scopatz: Strings have an rpartition method, see https://docs.python.org/3/library/stdtypes.html?highlight=rpartition#str.rpartition

funkyfuture:
it's very handy to review and fix code styling with the atom editor and [a linter](https://atom.io/packages/linter-flake8) and its enabled error panel.

GuillaumeLeclerc:
I don't want to work outside of vim.

But it might be cool to have a linter integrated in the build process. I usually do it in my other projects.

I even make it part of the CI tests to be sure the code base keep the same code style event if contributors use a different editor.

santagada:
I think this pull request might have become stale (because @adqm already implemented it and merged) but on the good news is that we have a linter in the build process now :).

I think this can be closed.

scopatz:
@GuillaumeLeclerc please let us know your thoughts on this.  Otherwise we'll close it in a few days. Thanks!

GuillaumeLeclerc:
Hello,

I was very busy but I'm back. I will install the latest version of xonsh and tell you if it adds more features than @adqm PR.

GuillaumeLeclerc:
The implementation on master branch is correct (and works better than mine on dotfiles). It's safe to close.

PR #904, opened 2016-05-15
Adds scripts/xonsh.bat to .gitattribues

funkyfuture:
also fixes its order

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
+tests/histories/*.json text eol=lf
+
 # Set the default behavior, in case people don't have core.autocrlf set.
 * text=auto
-
-tests/histories/*.json text eol=lf


GuillaumeLeclerc:
Great Idea :)

funkyfuture:
though it doesn't help ...

funkyfuture:
@melund mentioned in the chat that it could have an lf-ending. then should i just leave to the order so the second statement has an effect?

melund:
It appears that this file was ones added with `crlf` line ending. So that is screwing thing up for you. I don't now why other don't have the issue. 

I think it is safe to change it back to `lf` in git. When it is checked out on windows it will still be changed back to `crlf`. See https://github.com/scopatz/xonsh/pull/906

GuillaumeLeclerc:
I always see this file as "modified" but even when I `git add` it does not get staged

melund:
What is your output of `git config --global core.eol`

melund:
This probably what bites us: 
> When text=auto normalization is enabled in an existing repository, any text files containing CRLFs should be normalized. If they are not they will be normalized the next time someone tries to change them, causing unfortunate misattribution. From a clean working directory:

from https://git-scm.com/docs/gitattributes

melund:
So I think if https://github.com/scopatz/xonsh/pull/906 is merged, you should stop having the issues.

melund:
I may be necessary to clear cache..
```
git rm --cached -r .
git reset --hard
```

funkyfuture:
> What is your output of git config --global core.eol

none.

melund:
I think we should wait with this one and see if #906 fixes it.

scopatz:
@melund it seems like this is also a reasonable change.  Any reason not to merge it?

melund:
Just want to see what effect the other PR had. Let wait a day until we hear how it works for @funkyfuture and @GuillaumeLeclerc

scopatz:
Ok!

funkyfuture:
yes, it helped. thanks! i repushed an updated commit.

melund:
Great. The error was not normalizing all ´crlf`after `* text=auto` was added. I will you merge this now

PR #906, opened 2016-05-15
Remove crlf line endings in xonsh.bat

melund:
This removes the `crlf`lineending from scripts/xonsh.bat. It was causing a lot trouble. It should not have any side effect as I believe the file will still be checkout on windows with `crlf`


scopatz:
Thanks @melund!

PR #916, opened 2016-05-17
Fixed dirstack._try_cdpath to work with CDPATH values prefixed with ~/

sebastien:
The following script

```
mkdir ~/TEST
cd ~/TEST
mkdir A
ln -s A LINK
cd ~
$CDPATH=("~/TEST")
cd LINK
```

yields

```cd: no such file or directory: LINK```

### DIFF
--- a/xonsh/dirstack.py
+++ b/xonsh/dirstack.py
@@ -41,7 +41,7 @@ def _try_cdpath(apath):
     env = builtins.__xonsh_env__
     cdpaths = env.get('CDPATH')
     for cdp in cdpaths:
-        for cdpath_prefixed_path in iglob(os.path.join(cdp, apath)):
+        for cdpath_prefixed_path in iglob(os.path.expanduser(os.path.join(cdp, apath))):
             return cdpath_prefixed_path
     return apath
 
--- a/xonsh/dirstack.py
+++ b/xonsh/dirstack.py
@@ -41,7 +41,7 @@ def _try_cdpath(apath):
     env = builtins.__xonsh_env__
     cdpaths = env.get('CDPATH')
     for cdp in cdpaths:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: xonsh/dirstack.py [hunk 1]
```diff
@@ -41,7 +41,7 @@ def _try_cdpath(apath):
     env = builtins.__xonsh_env__
     cdpaths = env.get('CDPATH')
     for cdp in cdpaths:
-        for cdpath_prefixed_path in iglob(os.path.join(cdp, apath)):
+        for cdpath_prefixed_path in iglob(os.path.expanduser(os.path.join(cdp, apath))):
```
scopatz: Perhaps you could change `os.path.expanduser()` to `builtins.__xonsh_expand_path__()` (which you can also see as `xonsh.built_ins.expand_path()`) so that this works for environment variable names too, like `${HOME}/mydir`.

scopatz:
Thanks for hunting down this issue @sebastien! I really appreciate this.  Just the one minor change.  Also, would you be willing to put in a test for this so that it doesn't happen again. Also, you should totally add an entry to the CHANGELOG about fixing this.

scopatz:
Also, please leave a message in this issue so that we all get an email about it!

sebastien:
Here you are. Not 100% about the style, but hopefully this fits in.

scopatz:
Thanks a ton @sebastien! Looks fine to me.

sebastien:
I found another edge case, here it is https://github.com/sebastien/xonsh/commit/84e61badc8a31cbce021f76acc7dccaf38427a42

I somehow managed to the PR# wrong, but can't get to amend this commit.

scopatz:
That is fine.  Just feel free to open up another pr @sebastien. Or you can git diff, make a new branch, apply the diff to the branch, and have a new commit message.  But really, whatever you think is easiest.

Issue #78, opened 2015-03-17
reverse search not interrupted by ^C


eldamir:
When i press ctrl+r to search my command history, I sometimes want to cancel that search. In Bash, I do so by hitting ctrl+c (^C).

In xonsh, the ctrl+c stroke returns my to a fresh prompt, but as I type, it continues the search from before:

```
ruben@krogan:~ 
(failed reverse-i-search)`entr': cd workspace/contract_management/
ruben@krogan:~ 
(failed reverse-i-search)`entrs': 

```

scopatz:
Thanks for reporting @eldamir.  I am not able to reproduce this. Crtl-c seems to do the right thing in Crtl-r mode.

scopatz:
What kind of system are you on? Where did you get Python from, etc?  What does the following do?

```python
$ python -c 'import readline'
```

wrywerytwreywery:
I have this problem as well.  As @eldamir described, if I hit ctrl-c in ctrl-r mode, it shows a fresh prompt, but typing at that prompt continues the search from earlier.  Hitting Return breaks out of the incremental search, but also runs the command it found, so that's not ideal. 

Ubuntu 14.04
Python 3.4.0 from APT

scopatz:
I have posted a suggested solution in #91

wrywerytwreywery:
I tried out the suggestion in #91 (calling `reset_line_state` on receiving a `KeyboardInterrupt`), but it did not work for me.  I also tried a few other functions within libreadline that looked promising (`free_line_state`, for one), but nothing seemed to do the trick.  I also tried updating Python to 3.4.3 from Python 3.4.0, but that didn't fix, either.

Looking at the docs, it seems like readline should already be trying to call those functions when catching `SIGINT` (and it should be handling `SIGWINCH` to do the terminal resizing, which was an issue for me earlier), so these hooks shouldn't be necessary.  That said, after a little bit of digging, it looks like readline changed the way it handles signals in 6.3 (see [here](https://bugs.python.org/issue23735)).  But I have the same issue on a Debian machine with libreadline 6.2, so that seems unlikely as an explanation.

@scopatz, since you haven't seen any of these issues, can you post what version of libreadline and Python you're using?  

I will also try to set up a VM with more up-to-date software for testing.

wrywerytwreywery:
FWIW, though, Python's REPL also has this bug (or maybe it's somehow intentional?), so I don't think there is going to be an easy fix.

[...]

PR #920, opened 2016-05-18
Allow Ctrl-C to Exit Reverse-Incremental Search in Readline Shell

adqm:
This little patch should allow Ctrl-C to exit reverse-i-search mode in the readline shell (see #78).

Should work on Windows and *nix, I believe.


scopatz:
Works like a dream, thanks @adqm!

Issue #919, opened 2016-05-18
Error parsing some nested expressions


adqm:
I'm not sure exactly what is going on here, but I noticed some interesting parsing issues.  For example:

```
hartz@thinky-515 ~ $ @($(which screens).strip())
.................... 
  File "<string>", line None
SyntaxError: <xonsh-code>:1:0: ('code: @(',)
```

```
hartz@thinky-515 ~ $ @($(which screens))
<ERROR STEMMING FROM "\n" ON THE END OF THE COMMAND NAME>
```

```
hartz@thinky-515 ~ $ x = $(which screens)
hartz@thinky-515 ~ $ @(x.strip())
<RUNS OKAY!>
```

I think I have an idea of what is going wrong and how to fix, but figured I'd report first.  Of course, if anyone else wants to jump in and take this, that'd be fine, too :)

scopatz:
This is a new one on me, but looks suspiciously similar to #913

scopatz:
Though I think they might actually be different

scopatz:
Hey @adqm - is this closable?

adqm:
Ah, yes indeed.  Sorry I forgot!

PR #921, opened 2016-05-18
Add argument to which, to allow printing only the name of an alias

adqm:
Once #919 is fixed, I think it would be nice if `which` could be made always to produce a result that could be executed directly by xonsh.  This patch adds the `-p` (`--plain`) option to `which`, which should allow running, e.g., `@($(which -p my_alias).strip())`.

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -360,6 +360,8 @@ def which(args, stdin=None, stdout=None, stderr=None):
     parser.add_argument('-V', '--version', action='version',
                         version='{}'.format(_which.__version__))
     parser.add_argument('-v', '--verbose', action='store_true', dest='verbose')
+    parser.add_argument('-p', '--plain', action='store_true', dest='plain',
+                        help='Do not display alias expansions')
     parser.add_argument('--very-small-rocks', action=AWitchAWitch)
     if ON_WINDOWS:
         parser.add_argument('-e', '--exts', nargs='*', type=str,
@@ -389,7 +391,10 @@ def which(args, stdin=None, stdout=None, stderr=None):
         nmatches = 0
         # skip alias check if user asks to skip
         if (arg in builtins.aliases and not pargs.skip):
-            print('{} -> {}'.format(arg, builtins.aliases[arg]), file=stdout)
+            if pargs.plain:
+                print(arg, file=stdout)
+            else:
+                print('{} -> {}'.format(arg, builtins.aliases[arg]), file=stdout)
             nmatches += 1
             if not pargs.all:
                 continue


scopatz:
I also thought that this would be useful. Thanks @adqm!

melund:
@scopatz @adqm 
Instead of adding a new '-p' option should we not make this the default? 
```
$ which ls
ls
$which 2to3
```

We could then show the full origin of the alias if user provided the '--verbose/-v' option. That would be consistent with the behaviour when it is not an alias. In that case the 'which' module also shows additional information wit the -v falg. @gforsyth What is your thought on this?
```
$ which ls -v
ls -> ['ls', '--show-control-chars', '--color', '-X']
$ which 2to3 -v
C:\Users\mel\Anaconda3\envs\xonshdev\Scripts\2to3.exe (from given path element 2)
$ $PATH[2]
'C:\\Users\\mel\\Anaconda3\\envs\\xonshdev\\Scripts'
```

scopatz:
I am happy either way, personally, as long as there can be the option to display aliases, etc.

melund:
I don't know what would be best. If it was possible to run subproc lines with comments, i.e.`$(ls  # comment)`

Then we could output the following:
``` 
$ which ls
ls  # -> ['ls', '--show-control-chars', '--color', '-X']
``` 
But I can see why that would introduce a lot of other issues. 

If we decide on a `-p/--plain` option, should we not make sure the output is without the final newline?

adqm:
> If we decide on a -p/--plain option, should we not make sure the output is without the final newline?

I think this is an issue more generally (beyond just the `which` command), and I have been thinking about this a little bit recently.  In my mind, I had settled on either:

1. a new operator that automatically invokes `strip` and/or `split` on captured subproc output.
1. an environment variable to make this the default.

I guess I would lean toward option 2 for this.  I wouldn't necessarily want to make this behavior the default, but it would be nice if `$(...)` could be made to produce output that could generally be handled directly in the command line.

That said, no matter what, it still feels a little painful to me to have to call _two_ functions to inject output back into the command line: `@($(which python))`, not to mention the need for `strip`.  I'm not sure what that syntax would be, but it would be nice if this could be done in a single operation.

Really, I would vote for `!(...)` having this property, and `?(...)` being introduced to replace the super-captured subproc calls we have now.  To me, `!` says "do this" in a way that feels like it matches `@($(...))`; and `?` says, "tell me some information about this."

The behavior of these operators could still depend on an env var for whether to `strip`/`split` the output first.

Ugh, there isn't a clear win here in my mind.

adqm:
Or maybe whatever we come up with for the single-operation `@($(...))` should just _always_ do `strip`/`split` (which, I believe, is equivalent to what backticks or `$(...)` do in bash), and then we could leave `$(...)` alone and not bother with env vars.

I think that's my vote, in the end:

* `$(cmd)` is unchanged
* `?(cmd)` does the super-captured subproc (with return code, etc)
* `!(cmd)` does, effectively: `@([i.strip() for i in $(cmd).split()])`

I could go back/forth on the syntax, but I think those are the operators I would want.

gforsyth:
> Instead of adding a new '-p' option should we not make this the default?

I think this would be fine although if someone uses the `-a` flag I would like to see the alias printed as it currently is, and also if there is a `verbose` flag.  Given the... bumpy... history `which` has had of late we need to think about what a good `verbose` flag would be.  (Those used to `gnu-which` will expect the `-v` flag to return a version number)



> !(cmd) does, effectively: @([i.strip() for i in $(cmd).split()])

I like this.  It seems intuitive to me and roughly in line with how one would execute a shell command from within `less` (or `IPython`).

melund:
I am all for the change. But we must be fast if we should change the language interface. 

We don't want @scopatz making a presentation at pycon with the old syntax. A video like that would very confusing for new users.

melund:
I guess the most important thing is change the current !()/![] to ?()/?[] . So `!` is free for the new syntax.

adqm:
I'm working on it right now so we can at least have a proof-of-concept and make an informed decision :)

Will submit a PR soon, and we can continue the discussion there.

PR #923, opened 2016-05-18
A few readline updates

scopatz:
This PR adds a few readline related fixes.

1. The readline shell will now load user config files, which was requested by @asmeurer way back in #290
2. Adds some more docs on the $TERM variable.
3. Adds the ability to dump out the current readline configuration. These are mostky meant as a developer/debuging tools. Users shouldn't have to use them ever, but I was getting annoyed at having to reimplement them every time I did readline stuff :)


adqm:

PR #924, opened 2016-05-18
don't crash on empty .xonshrc file

adqm:
I think this might have been a merge failure on my part (with #757), but an old issue has crept back in, whereby if a user has an empty `.xonshrc` file (rather than no `.xonshrc` at all), xonsh will crash immediately on startup.  This small patch should avoid that issue.


scopatz:
Thanks @adqm!

PR #925, opened 2016-05-18
Add option to prevent config wizard starting every time xonsh is run

adqm:
This patch changes the question that is asked when a user starts xonsh with no configuration files to allow a third option (specifically, the option not to run the wizard now _or ever_).

When I was doing some testing after removing my `.xonshrc` file, I found it a little annoying that I was asked whether I wanted to run the wizard every time I opened a new terminal.  I'm not sure if I'm alone in this (if so, or if there is a better way to implement this, we can close this PR).

I suppose it is a little bit dangerous as written, as it will blow away an existing config file if one exists _and_ for some reason the wizard is run with `confirm=True` _and_ the user chooses option 3.  But I imagine it would be hard to do that acidentally, and xonsh backs up the old file anyway, so I don't think that's a concern.

### DIFF
--- a/xonsh/wizard.py
+++ b/xonsh/wizard.py
@@ -214,7 +214,7 @@ class StateFile(Input):
 
     attrs = ('default_file', 'check')
 
-    def __init__(self, default_file=None, check=True):
+    def __init__(self, default_file=None, check=True, ask=True):
         """
         Parameters
         ----------
@@ -224,10 +224,14 @@ class StateFile(Input):
             Whether to print the current state and ask if it should be
             saved/loaded prior to asking for the file name and saving the
             file, default=True.
+        ask : bool, optional
+            Whether to ask for the filename (if ``False``, always use the
+            default filename)
         """
         self._df = None
         super().__init__(prompt='filename: ', converter=None,
                          confirm=False, path=None)
+        self.ask_filename = ask
         self.default_file = default_file
         self.check = check
 
@@ -604,7 +608,9 @@ class PromptVisitor(StateVisitor):
             do_save = self.visit(asker)
             if not do_save:
                 return Unstorable
-        fname = self.visit_input(node)
+        fname = None
+        if node.ask_filename:
+            fname = self.visit_input(node)
         if fname is None or len(fname) == 0:
             fname = node.default_file
         if os.path.isfile(fname):
--- a/xonsh/wizard.py
+++ b/xonsh/wizard.py
@@ -224,7 +224,7 @@ class StateFile(Input):
             Whether to print the current state and ask if it should be
             saved/loaded prior to asking for the file name and saving the
             file, default=True.
-        ask : bool, optional
+        ask_filename : bool, optional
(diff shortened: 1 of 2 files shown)


xonsh/xonfig.py
```diff
@@ -274,8 +276,13 @@ def make_wizard(default_file=None, confirm=False):
...
     if confirm:
-        q = 'Would you like to run the xonsh configuration wizard now, ' + YN
-        wiz = YesNo(question=q, yes=wiz, no=Pass())
+        q = ("Would you like to run the xonsh configuration wizard now?\n\n"
+             "1. Yes\n2. No, but ask me later.\n3. No, and don't ask me again."
+             "\n\n1, 2, or 3 [default: 2]? ")
+        passer = Pass()
+        saver = Save(check=False, ask=False, default_file=default_file)
+        wiz = Question(q, {1: wiz, 2: passer, 3: saver},
+                       converter=lambda x: int(x.strip()) if x !='' else 2)
```
scopatz: Just as a side note, `int()` will strip automatically.

scopatz: also PEP8 single space after `!=`

xonsh/wizard.py
```diff
@@ -224,10 +224,14 @@ def __init__(self, default_file=None, check=True):
...
             saved/loaded prior to asking for the file name and saving the
             file, default=True.
+        ask : bool, optional
+            Whether to ask for the filename (if ``False``, always use the
+            default filename)
         """
         self._df = None
         super().__init__(prompt='filename: ', converter=None,
                          confirm=False, path=None)
+        self.ask_filename = ask
```
scopatz: probably should just be called `ask`, to be consistent.

xonsh/wizard.py
```diff
@@ -214,7 +214,7 @@ class StateFile(Input):
 
     attrs = ('default_file', 'check')
```
scopatz: Please add `ask` to this tuple as well

xonsh/wizard.py
scopatz: Also please include `ask` on the PrettyFormatter on line 391

scopatz:
Thanks @adqm! This is a nice idea and basically what I have done myself by hand.  Happy to merge this after the few inline comments have been addressed.

adqm:
Okay, I think this should be ready for another look now.

xonsh/wizard.py
```diff
@@ -224,10 +224,14 @@ def __init__(self, default_file=None, check=True):
             Whether to print the current state and ask if it should be
             saved/loaded prior to asking for the file name and saving the
             file, default=True.
+        ask : bool, optional
```
scopatz: Shouldn't this be `ask_filename`

scopatz:
One more little thing :)

adqm:
D'oh!  I think this should do it.

scopatz:
Looks good to me!

PR #926, opened 2016-05-18
Proposal: Syntax Changes (from #921)

adqm:
Per the discussion in  #921, this changeset introduces a couple of changes to the syntax of the xonsh language, relating to the handling of subprocess args:

* `$(cmd)` ~~is unchanged~~ now _always_ gives the whole string as a single element, even in subproc mode
* `?(cmd)` does the super-captured subproc (with return code, etc)
* `!(cmd)` does, effectively: `@([i.strip() for i in $(cmd).split()])`

Quoting a relevant piece of that discussion:

> That said, no matter what, it still feels a little painful to me to have to call two functions to inject output back into the command line: @($(which python)), not to mention the need for strip. I'm not sure what that syntax would be, but it would be nice if this could be done in a single operation.

>Really, I would vote for !(...) having this property, and ?(...) being introduced to replace the super-captured subproc calls we have now. To me, ! says "do this" in a way that feels like it matches @($(...)); and ? says, "tell me some information about this."

Of course, this is a big and backward-incompatible change, so this is open for discussion.

I updated the docs and existing tests, but I have not yet added documentation or tests for the new syntax.  The changes that are actually meaningful are in `execer.py`, `lexer.py`, `parsers/base.py`, `built_ins.py`, and `tools.py`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,8 @@ Current Developments
 
 * When a subprocess exits with a signal (e.g. SIGSEGV), a message is printed,
   similar to Bash.
+* ``@$(cmd)`` has been added as a subprocess-mode operator, which replaces in
+  the subprocess command itself with the result of running ``cmd``.
 
 **Changed:** None
 
--- a/docs/bash_to_xsh.rst
+++ b/docs/bash_to_xsh.rst
@@ -19,11 +19,10 @@ will help you put a finger on how to do the equivelent task in xonsh.
       - Look up an environment variable via another variable name. In xonsh,
         this may be any valid expression.
     * - ``$(cmd args)`` or ```cmd args```
-      - ``$(cmd args)``
-      - Use the ``$()`` operator to capture subprocesses as strings. Bash's
-        version of (now-deprecated) backticks is not supported. Note that
-        Bash will automatically tokenize the string, while xonsh just returns
-        a str of stdout.
+      - ``@$(cmd args)``
+      - Command substitution (allow the output of a command to replace the
+        command itself).  Tokenizes and executes the output of a subprocess
+        command as another subprocess.
     * - ``set -e``
       - ``$RAISE_SUBPROC_ERROR = True``
       - Cause a failure after a non-zero return code. Xonsh will raise a
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -459,6 +459,10 @@ feed them to a subprocess as needed.  For example:
     for i in range(20):
         $[touch @('file%02d' % i)]
 
+.. note:: A common use of the ``@()`` operator is allowing the output of a
+          command to replace the command itself (command substitution):
+          ``@([i.strip() for i in $(cmd).split()])``.  Xonsh offers a
+          short-hand syntax for this operation: ``@$(cmd)``.
 
 Nesting Subprocesses
 =====================================
--- a/tests/test_lexer.py
+++ b/tests/test_lexer.py
@@ -134,11 +134,23 @@ def test_multiline():
            ('NAME', 'y', 0),]
     yield check_tokens, inp, exp
 
+def test_atdollar_expression():
+    inp = '@$(which python)'
+    exp = [('ATDOLLAR_LPAREN', '@$(', 0),
+           ('NAME', 'which', 3),
+           ('WS', ' ', 8),
+           ('NAME', 'python', 9),
+           ('RPAREN', ')', 15)]
+    yield check_tokens, inp, exp
+
 def test_and():
     yield check_token, 'and', ['AND', 'and', 0]
 
 def test_ampersand():
     yield check_token, '&', ['AMPERSAND', '&', 0]
+
+def test_atdollar():
+    yield check_token, '@$', ['ATDOLLAR', '@$', 0]
 
 def test_doubleamp():
     yield check_token, '&&', ['AND', 'and', 0]
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1622,6 +1622,12 @@ def test_backtick():
 def test_ls_regex_octothorpe():
     yield check_xonsh_ast, {}, '$(ls `#[Ff]+i*LE` -l)', False
 
+def test_injection():
+    yield check_xonsh_ast, {}, '$[@$(which python)]', False
+
+def test_rhs_nested_injection():
+    yield check_xonsh_ast, {}, '$[ls @$(dirname @$(which python))]', False
+
 def test_backtick_octothorpe():
     yield check_xonsh_ast, {}, 'print(`#.*`)', False
 
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -649,6 +649,12 @@ def subproc_captured_stdout(*cmds):
     return run_subproc(cmds, captured='stdout')
 
 
+def subproc_captured_inject(*cmds):
+    """Runs a subprocess, capturing the output. Returns a list of
+    whitespace-separated strings in the stdout that was produced."""
+    return [i.strip() for i in run_subproc(cmds, captured='stdout').split()]
+
+
 def subproc_captured_object(*cmds):
     """
     Runs a subprocess, capturing the output. Returns an instance of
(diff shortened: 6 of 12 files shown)


adqm:
There are also some issues this is bringing up with the parser.  Specifically, it seems like it is not possible to specify arguments after one of the `$()` or `!()` arguments on the command line.  So the following succeed:

```
$ !(which python)
$ echo $(ls)
$ echo !(ls)
```

But the following fail (syntax error?):
```
$ !(which python) test.py
$ echo $(ls) hey
$ echo !(ls) hello
```

It's not immediately apparent to me why this is happening, but it's an issue we should resolve anyway (regardless of this change), so I'll try to track it down and fix it here, and if this is rejected, we can still put in that change.  First guess would be that it involves `subproc_toks` not wrapping the whole expression (because of the ending paren)?

scopatz:
Hi @adqm, thanks a million for putting this in!  

Personally, this seems like it is a sub-case or another instance of [the filtering discussion that we had last year on the mailing list](https://groups.google.com/d/topic/xonsh/fRoEQR0QGeA/discussion).  I think a general solution to would be much preferred to a single operator that does split-and-strip. I am still very much in favor of having something like this in xonsh, even though we decided against doing anything like this at that time. (I think we are in better situation now, though, for sure.)

That said, I really do like the way the current `!()` and `$()` operators work and would not like to see them changed.  I think that this filtering stuff would be a new operator  - or a new use of an old operator somewhere. I wouldn't be opposed if this filtering stuff used the `?`.  

Or perhaps, this could be applied to all subproc modes through something like a trailing `?`, like with `&` and backgrounding.  By default, `![ls ?]` would do a split and strip, but maybe you could add a filter function after this too: `![ls ? myfunc]`.  This would probably be a little harder to implement in the parser, but not impossible.

melund:
Does it have to be a special character. Could it be a normal character like capital X( )? That would of course prevent the of `X` as a function name. 

Any way, what about having that as a user defined operator in xonsh? That way it could do what what @adqm we want or any other for that matter. It could be defined as function and saved in an environment var `$XONSH_OP`

Maybe a little far fetched, but what do you think?

scopatz:
Hey @melund, I don't think variable name characters are a good idea.  This basically makes that variable a keyword of the language. 

I am also not a huge fan of user defined syntax (at least not in a Python context). This would move xonsh closer to perl or forth.  I think this would have a couple of undesirable side effects:

1. It could mean that people using one xonsh operator would not be able to use code from folks using a different operator.
2. It would make xonsh harder to teach and to explain.
3. It would make syntax highlighting and other static analysis tools more difficult to implement.

Again, I am a fan of the core ideas here; I was kind of sad that we delayed this issue originally. I think that there is a way for us to come to a consensus on what this feature should look like :)

melund:
Hey @scopatz ...  I have no experience with Perl, but I can see how all of your points would undesirable.  

What other ideas can we think of? I do like the concept that frequently repeated patterns are simple to write and the syntax is aesthetically pleasing.  

Could we add methods the completed_command  object so we could write something like: `!(cmd).strip_args()` instead of `@([i.strip() for i in $(cmd).split()])` or have a `!(cmd).filter(my_func)` to make abitrary filters.

adqm:
>  I think a general solution to would be much preferred to a single operator that does split-and-strip.

My main reason for wanting this is to have something analogous to backticks in bash, which I used to use all the time.  The current semantics of `$(...)` make that hard (or, at least, verbose; it is a pain to type out all those characters!).

I quite liked the idea of filters, but the more general those things get, the more verbose (necessarily, I think).  I still think we should revisit that discussion because I could see those being quite useful, but it would be also nice to have a shorthand for something that I think is a very common use case.

Unless I'm mistaken, it's not even really possible to make a xonsh function that does this, so it involves rewriting that same piece of code every time one wants this behavior.

> I don't think variable name characters are a good idea.
> I am also not a huge fan of user defined syntax

I agree with these points.

> ...or forth

That's a great idea; we should refactor xonsh into a Forth-based shell! /s

scopatz:
> My main reason for wanting this is to have something analogous to backticks in bash, which I used to use all the time.

Ahh! I didn't grok this until now.  The conciseness is itself the valuable piece of syntax.  My next question then is, would this bit of syntax only apply to subproc mode?  If we don't need this operator to work in Python, this opens up a lot of options... Though this may not be possible.

> That's a great idea; we should refactor xonsh into a Forth-based shell! /s

I am literally dying right now.

adqm:
> My next question then is, would this bit of syntax only apply to subproc mode?

That's the way it's implemented now, yes, and that's the intention.

If it is allowed to operate in Python mode as well, then just `!(which python)` by itself will just return the list, rather than injecting it into subproc mode.  Plus, in Python mode, writing `.split().strip().whatever()` doesn't feel weird (to me, at least); that feeling is unique to subproc mode.

adqm:
Just rebased this onto master, and updated some test cases.  Still happy to discuss more.

melund:
Nice. I really like this. 
```
snail@sea ~\Documents\scratch $ which ls
snail@sea ~\Documents\scratch $ # Running the output before was not nice
snail@sea ~\Documents\scratch $ @([i.strip() for i in $(which ls).split()]) 
lobatus_gigas  pleuroploca_gigantea  strombidae  tubinella
snail@sea ~\Documents\scratch $ # And now it is a breeze
snail@sea ~\Documents\scratch $ !(which ls)
lobatus_gigas  pleuroploca_gigantea  strombidae  tubinella
```

Besides, I don't use the CompletedCommand stuff directly nearly as much as I thought. So having `!` for the new stuff may not be bad. 

> If it is allowed to operate in Python mode as well, then just !(which python) by itself will just return the list, rather than injecting it into subproc mode.

@adqm.  I do get some weird behaviour when using it in python mode.
```
snail@sea ~\Documents\scratch $ d = !(which ls)
 Volume in drive C has no label.
 Volume Serial Number is FA73-929B

 Directory of C:\Users\mel\Documents\scratch


 Directory of C:\Users\mel\Documents\scratch


 Directory of C:\Users\mel\Documents\scratch


 Directory of C:\Users\mel\Documents\scratch

File Not Found
```

scopatz:
> That's the way it's implemented now, yes, and that's the intention.

Ok, cool.  Then we could use things like `%()`, `:()`, `+()` or similar.

melund:
Then I like `:()` the most. I was about to ask if `*()` could have been used. But I guess it clashes with pythons `*args` and `**kwargs` use.  But what about `!!()`?

adqm:
For now, I am actually thinking of putting it in as `@$(...)`.  Nothing new for people to learn; just a nice short-hand.  So instead of `@([i.strip() for i in $(cmd).split()])`, one would write `@$(cmd)`

Does that work for you guys?

As part of these changes, I'll switch `?(` and `?[` back to the ones that use bang.

scopatz:
That works for me!  I think `@$()` is nice in its way.  If we wanted to do filters this way, we could do `@f$()` for a single filter, and `@f@g$()` for multiple.  This harkens back to decoration and function composition.  I like it.

adqm:
Alright, I think this is ready for another look now.

scopatz:
This looks good to me @adqm! @melund?

melund:
@adqm .. It doesn't quite work for me. It opens multiline mode with prompt_toolkit. And if I press enter again, I get this error. 
```
snail@sea ~\Documents\xonsh adqm-syntax $ @$(which ls)
.........................................
  File "<string>", line None
SyntaxError: <xonsh-code>:1:0: ('code: @$(',)
```

adqm:
>   It doesn't quite work for me. It opens multiline mode with prompt_toolkit. And if I press enter again, I get this error. 

Ugh.  I can reproduce this, unfortunately. I know this was working earlier, so I'll have to dig back in and find out what went wrong along the way.

adqm:
I am very confused right now.  Re-built, with no meaningful changes to the code (just a couple of PEP8 things), and everything seems to be working now.  @melund, can you try updating, re-installing, and trying again?

adqm:
Actually, just rebased on to master again.  Maybe best to make a start from master, pull these changes (some of which may conflict with the older versions) and try again.

melund:
Was it as simple as me forgetting to rebuilding the parser tables :S

Then sorry for the confusion. Now it works perfectly :)

image

melund:
This is really a neat feature :)

scopatz:
Thanks! I think this could use some tutorial docs too ;)

melund:
It has :)

scopatz:
oh it does! but that is just a note. it seems like this would deserve its own section

Issue #208, opened 2015-04-15
Bash error on start


aig787:
On a base install of Fedora 20 Workstation I get 
```bash
/bin/sh: scl: line 1: syntax error: unexpected end of file
/bin/sh: error importing function definition for `BASH_FUNC_scl'
/bin/sh: module: line 1: syntax error: unexpected end of file
/bin/sh: error importing function definition for `BASH_FUNC_module'
```
when I try to launch xonsh. The link here https://kb.brightcomputing.com/faq/index.php?action=artikel&cat=2&id=239&artlang=en seems to suggest it has something to do with how bash handles functions post shellshock.

aig787:
The issue resolved itself. Couldn't tell you how or why, but it's very likely not anything xonsh was doing.

scopatz:
Weird, thanks for reporting, though.

janpipek:
I observe a similar behaviour when using `conda` command (or a few others):

```
> conda list
sh: scl: line 1: syntax error: unexpected end of file
sh: error importing function definition for `BASH_FUNC_scl'
sh: module: line 1: syntax error: unexpected end of file
sh: error importing function definition for `BASH_FUNC_module'
sh: scl: line 1: syntax error: unexpected end of file
sh: error importing function definition for `BASH_FUNC_scl'
sh: module: line 1: syntax error: unexpected end of file
sh: error importing function definition for `BASH_FUNC_module'
# packages in environment at /home/honza/anaconda/envs/py34:
#
/bin/sh: scl: line 1: syntax error: unexpected end of file
/bin/sh: error importing function definition for `BASH_FUNC_scl'
/bin/sh: module: line 1: syntax error: unexpected end of file
/bin/sh: error importing function definition for `BASH_FUNC_module'
alabaster                 0.7.3                    py34_0  
appdirs                   1.4.0                     <pip>
babel                     1.3                      py34_0
```

I wasn't able to find any scl use in either xonsh or conda sources, so it should come either from Python binary, bash or OS (and yes, it is also Fedora, v. 22)

scopatz:
Huh - weird @janpipek. `conda list` looks fine on my machine.  What version of conda are you on, what OS, etc.

janpipek:
* Fedora 22 - likely the culprit (as scl is Redhat stuff)
* conda 3.13.0 
* python 3.4.3
* GNU bash, version 4.3.39(1)-release (x86_64-redhat-linux-gnu)

[...]

PR #929, opened 2016-05-18
Document Bash config conflicts, according to issue #208.

terrycloth:
I don't use ReStructured Text very often, so hopefully I didn't mess up the markup. Or the actual information :P


scopatz:
Thanks a ton @terrycloth!

PR #931, opened 2016-05-19
Adding xontrib-prompt-ret-code extension

Siecje:
What is the difference between ```name``` and ```package```?

Currently package is the Python package name to install with pip.
The name is the name of the new module that will be added to ```xontrib```.

Ref: https://github.com/scopatz/xonsh/pull/816

### DIFF
--- a/xonsh/xontribs.json
+++ b/xonsh/xontribs.json
@@ -4,7 +4,13 @@
   "url": "http://xon.sh",
   "description": ["Matplotlib hooks for xonsh, including the new 'mpl' alias ",
                   "that displays the current figure on the screen."]
+  },
+ {"name": "prompt_ret_code",
+  "package": "xontrib-prompt-ret-code",
+  "url": "https://github.com/Siecje/xontrib-prompt-ret-code",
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks a million for putting this in @Siecje and being the first Guinea pig ;)

The difference between name and package is that 

* name is the xontrib name, the name that users will use to type in to load the xontrib.
* the package name is a key into the packages dictionary and should be related to the package that users install.

The reason for this distinction is that a single package may have many xontribs, each with their own names and functionality.

scopatz:
Super excited to see this in!

scopatz:
Also, please feel free to improve the tutorial docs with more info on xontrib and package naming. It could probably use it...

Siecje:
Okay so I used them correctly :)

One thing I have not figured out is how to load it from ```.xonshrc```.

scopatz:
Hi @Siecje, I have opened up #933 to fix these issues. If you could try this out and let me know if it works, that'd be great. When I use this with your xontrib, everything works as expected.  Sorry about the confusion!

PR #933, opened 2016-05-19
Fix many logic issues with xontribs

scopatz:
Thanks to @Siecje, I found many logic issues with the xontribs module:

1. Inability to import *.xsh files before the Shell was initialized (ie in the config or rc files)
2. Import warnings were silently suppressed by default
3. xontrib context would be overwritten as `__xonsh_ctx__` by the time the actual prompt was reached.

This should clear all of these up.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -662,7 +662,7 @@ def ensure_list_of_strs(x):
     return rtn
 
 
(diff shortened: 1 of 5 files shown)


Siecje:
Yes this fixes loading from ```.xonshrc``` automatically.

PR #934, opened 2016-05-19
Which updates

melund:
This updates the `which` alias so both '-v/-V/--version' triggers the version output. Similar to gnu-which. 

It also makes it the default to show where binaries are found, similar to how we by default show where aliases come from: 
``` bash
snail@sea ~ $ which ipconfig
C:\Windows\system32\ipconfig.exe -> (from given path element 6)
snail@sea ~ $ which ls
ls -> ['ls', '--show-control-chars', '--color', '-X']
```
 The `-p/--pain` can still be used to hide this. 

```bash
snail@sea ~ $ which ipconfig -p
C:\Windows\system32\ipconfig.exe
snail@sea ~ $ which ls -p
ls
```

The `-a/--all` option is unchanged. 
```bash
snail@sea ~ $ which ls -a
ls -> ['ls', '--show-control-chars', '--color', '-X']
C:\Program Files\Git\usr\bin\ls.exe -> (from given path element 17)
```

The `--verbose` has a much more limited functionality now. It will report on duplicates and near misses (files that are not executable on posix). 

The PR also updates `whichgen()` from the `_which.py` module to always return tuples of (<path to command>, <where path found>). This is just an internal thing, that made it possible to change the functionality of the `--verbose`

### DIFF
--- a/xonsh/xoreutils/_which.py
+++ b/xonsh/xoreutils/_which.py
@@ -177,8 +177,7 @@ def whichgen(command, path=None, verbose=0, exts=None):
         not a VisualBasic script but ".vbs" is on PATHEXT. This option
         is only supported on Windows.
 
-    This method returns a generator which yields either full paths to
-    the given command or, if verbose, tuples of the form (<path to
+    This method returns a generator which yields tuples of the form (<path to
     command>, <where path found>).
     """
     matches = []
@@ -214,10 +213,7 @@ def whichgen(command, path=None, verbose=0, exts=None):
     if os.sep in command or os.altsep and os.altsep in command:
         if os.path.exists(command):
             match = _cull((command, "explicit path given"), matches, verbose)
-            if verbose:
-                yield match
-            else:
-                yield match[0]
+            yield match
     else:
         for i in range(len(path)):
             dirName = path[i]
@@ -239,18 +235,12 @@ def whichgen(command, path=None, verbose=0, exts=None):
                         fromWhere = "from PATH element %d" % (i-1)
                     match = _cull((absName, fromWhere), matches, verbose)
                     if match:
-                        if verbose:
-                            yield match
-                        else:
-                            yield match[0]
+                        yield match
         match = _getRegisteredExecutable(command)
         if match is not None:
             match = _cull(match, matches, verbose)
             if match:
-                if verbose:
-                    yield match
-                else:
-                    yield match[0]
+                yield match
 
 
 def which(command, path=None, verbose=0, exts=None):
@@ -272,10 +262,13 @@ def which(command, path=None, verbose=0, exts=None):
     If no match is found for the command, a WhichError is raised.
     """
     try:
-        match = whichgen(command, path, verbose, exts).next()
+        absName, fromWhere = whichgen(command, path, verbose, exts).next()
     except StopIteration:
         raise WhichError("Could not find '%s' on the path." % command)
-    return match
+    if verbose:
+        return absName, fromWhere
+    else:
+        return absName
 
 
 def whichall(command, path=None, verbose=0, exts=None):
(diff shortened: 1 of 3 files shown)


xonsh/xoreutils/_which.py
```diff
@@ -345,11 +340,11 @@ def main(argv):
     for arg in args:
         #print "debug: search for %r" % arg
         nmatches = 0
-        for match in whichgen(arg, path=altpath, verbose=verbose, exts=exts):
+        for absName, fromWhere in whichgen(arg, path=altpath, verbose=verbose, exts=exts):
```
adqm: picking nits, but these should be PEP8 variable names (snakes instead of camels)

xonsh/xoreutils/_which.py
```diff
@@ -295,7 +287,10 @@ def whichall(command, path=None, verbose=0, exts=None):
         not a VisualBasic script but ".vbs" is on PATHEXT. This option
         is only supported on Windows.
     """
-    return list( whichgen(command, path, verbose, exts) )
+    if verbose:
+        return list( whichgen(command, path, verbose, exts) )
+    else:
+        return list( absName for absName, _ in whichgen(command, path, verbose, exts))
```
adqm: PEP8: variable name

xonsh/xoreutils/_which.py
```diff
@@ -272,10 +261,13 @@ def which(command, path=None, verbose=0, exts=None):
     If no match is found for the command, a WhichError is raised.
     """
     try:
-        match = whichgen(command, path, verbose, exts).next()
+        absName, fromWhere = whichgen(command, path, verbose, exts).next()
```
adqm: PEP8: variable names

adqm:
The other thing, from a high level, is that it's not clear to me how helpful it is to know that something was found in the nth index into the path.  I would say just show the literal path element itself, but that tends to be encoded in the filename that is returned.  Is there some (potentially) more helpful piece of text we can show?  Or am I missing some helpful use case for that index?

xonsh/xoreutils/_which.py
melund: True, but am just following the existing naming convention of this old module. The same variables are used other places with these names

adqm: Ah, I see.  Ignore those comments, then.

melund:
The actual path element is definitely not useful in it self. On Windows it is useful to know if the command comes from the path because it could also come from other places like the app registry. In general I could see the benefit of knowing which path element it comes from. If you have multiple binaries with the same name on the path you can easily exclude it with: `$PATH.append($PATH.pop(number)` or similar...

adqm:
Ah, then that sounds good to me!  Ran it through a quick test on my machine, and all looks good.  Merging in now.

adqm:
That said, looking at this and thinking about #926, we may indeed want to reverse the logic (people do tend to expect that `which` will give only the location of the command and not these extra characters).

So now I'm swinging back around to: `which`, by default, should return only the location, and with some verbose flag, it can tells us whether it's an alias or command, and more details about how it was found.

melund:
@adqm and @gforsyth 

Yes. Having the `-p/--plain` as default was also what I suggested. However, if '--verbose' is going to be used more, we need to have a short option for it. So I suggest we change the `-v` option back to mean verbose. I.e. `-v/--verbose` and then the version should only be `-V/--version`. This was what the `which` module originally had.

Here is my suggestion for how it could work:

```
$ which ls

$ which 2to3
```

```
$ which ls -v
ls -> ['ls', '--show-control-chars', '--color', '-X']

$ which 2to3 -v
C:\Users\mel\Anaconda3\envs\xonshdev\Scripts\2to3.exe -> (from given path element 2)
```

If the `-p/-plain` is added with the `-v/--verbose`, then plain takes priority. This is to ensure that people can override verbose if they have an alias like this (`which -> ['which', '--verbose']`)
```
$ which ls -v -p
```

The `-a/--all` option would automatically be verbose:
``` 
$ which ls -a
ls -> ['ls', '--show-control-chars', '--color', '-X']
C:\Program Files\Git\usr\bin\ls.exe -> (from given path element 17)
```
But again can be overridden with `-p/--plain`:
``` 
$ which ls -a -p
C:\Program Files\Git\usr\bin\ls.exe
```

We can also consider `-q/--quite` instead of `--plain`. That is what the `which` module did originally. I probably like `-p/--plain` better.

Issue #268, opened 2015-06-15
Files with non-standard characters


mankoff:
Just an FYI that I had a file named `#foo` in `/tmp/`. I tried to delete it and xonsh supported auto-completion of `/tmp/#<TAB>`, but then interpreted the "#" as a comment and deleted the `/tmp` folder. Human error, but perhaps xonsh could escape file characters when TAB-completing the way bash does.

scopatz:
Thanks for reporting.  It is a good idea that we are warming up to. Until then, you can always put the file name in quotes for it to be passed in literally as typed.

ned2:
A limitation of the quoting workaround is that you can't use globbing to say delete all your `#` prefixed files, so having an explicit means of escaping `#` would indeed be most welcome.

scopatz:
For what it is worth, I would just use regex globbing in this case.

ned2:
Ah yes of course. Still absorbing xonsh's capabilities. Thanks @scopatz!

ned2:
Actually, it seems like xonsh is interpreting `#` as a comment even inside a regex. Unless I'm missing something?

adqm:
> Actually, it seems like xonsh is interpreting # as a comment even inside a regex. Unless I'm missing something?

Yep, confirmed here.  This is definitely a parsing bug (well, actually, a lexing bug).  I think I know how to fix.

adqm:
@ned2, I just put in #943, which I believe should fix this..  Do you mind giving that a whirl?

scopatz:
Also, you can use the super-verbose `ls @(__xonsh_glob__("#*"))` if you just need to get something working :)

ned2:
@adqm, looks good! Was able to conveniently delete a bunch of octothorpe-riddled stale emacs autosave files.

scopatz:
Great! @mankoff feel free to close this.  In your original case, you could use strings around the file name, though you could also use regex globbing.

PR #943, opened 2016-05-19
Tokenizer Fix

adqm:
The built-in `tokenize` module was getting confused with "comments" between backticks (anything with an octothorpe).  See the end of #268.  The first of many parser fixes over the next few days, I hope.

This changeset should fix that problem, though it may not yet fix the original issue.  The fix is a little more involved than I would have liked (including a modified version of `tokenize.py` that knows that regex globbing is special), but I could see no other way.  In fact, it may make sense to move much more of the logic into `xonsh.tokenize` in time.

Also includes a slight refactor to the rest of the lexer (part of a first failed attempt, but still not a bad change).

This is a big diff, so it may be easier to see the changes by looking at individual commits.  The commits for the refactor are separate from the commits for adding the copies of the built-in `tokenize` module, which are separate from the commits for actually updating the tokenize module, which are separate from the commits for modifying the lexer to use the new tokenizer.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -59,6 +59,9 @@ Current Developments
   recommended.
 * Provide better user feedback when running ``which`` in a platform that doesn't
   provide it (e.g. Windows).
+* The lexer now uses a custom tokenizer that handles regex globs in the proper
+  way.
+
 
 **Deprecated:** None
 
--- a/setup.py
+++ b/setup.py
@@ -160,7 +160,7 @@ def main():
         url='https://github.com/scopatz/xonsh',
         platforms='Cross Platform',
         classifiers=['Programming Language :: Python :: 3'],
-        packages=['xonsh', 'xonsh.ptk', 'xonsh.parsers', 'xonsh.xoreutils', 'xontrib'],
+        packages=['xonsh', 'xonsh.ptk', 'xonsh.parsers', 'xonsh.xoreutils', 'xontrib', 'xonsh.tokenize'],
         package_dir={'xonsh': 'xonsh', 'xontrib': 'xontrib'},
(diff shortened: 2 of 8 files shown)


scopatz:
Hey @adqm, this looks great! I think this could use a couple of updates prior to merging:

1. a CHANGELOG entry
2. maybe add the tokenize modules to the API docs?  
3. A simple lexer, parser, or tokenizer test that at least covers the `#.*` case.  I don't think we need anything super complex here.  Just enough to prevent the original error from creeping back in.

In the longer term, it would be interesting to consider if most parts of the forked tokenizers are the same and we only need to have version-specific changes for a small subset, similar to how the parsers work. However, this is not a pressing issue.

Also, I agree that it would be nice to push more of the errortoken things from the lexer back to the tokenizer, eventually :)

adqm:
I just pushed changes to the CHANGELOG and a couple of simple test cases.  I'm not quite sure how best to document the tokenizer.

> In the longer term, it would be interesting to consider if most parts of the forked tokenizers are the same and we only need to have version-specific changes for a small subset, similar to how the parsers work. However, this is not a pressing issue.

Yes, I can do that.  I do think large portions of them are the same (or similar)  Do you think that change should go in here?  Or in a separate PR?

scopatz:
> I'm not quite sure how best to document the tokenizer.

No worries about that. This is an implementation detail.

> Yes, I can do that. I do think large portions of them are the same (or similar) Do you think that change should go in here? Or in a separate PR?

I think a separate PR is better so that we can get this fix in.

adqm:
Okay, great.  In that case, I think this is ready to go.

scopatz:
Great!

scopatz:
Thanks @adqm!

PR #944, opened 2016-05-19
fix broken imports in certain tests

gforsyth:
these were introduced in a few places but there is no __init__.py file
in the tests directory so the import breaks (at least on my machine)


scopatz:
Thanks @gforsyth! I have been meaning to get to this for a while. It has been annoying me for months :)

Issue #942, opened 2016-05-19
Can't make the test suite crashing


GuillaumeLeclerc:
Hello,

I was trying to tests. To try it I made a simple test to make it fail and see if it works but the problem is `nosetests` says it's all fine.

I tried adding a function:

```python
def test_tests(): 
    assert_equal(True, False)
    print("HOHOHOHOH")
```

and no tests are failing. Moreover nothing is printed

I tried adding a print on top of the file and it's not getting printed. Then I tried adding a false assertion in other tests and none of them is failing. 

I'm sure I must be something wrong but I can't figure out what.

scopatz:
What if you call this `test_x()` rather than `test_tests()`?

GuillaumeLeclerc:
The same thing happen. The strange thing is nothing is printed, even if I add it on top of the file

It seems the code I'm typing is not taken into account. It looks like tests are cached in some way.

melund:
Nose captures stdout. To print use `nosetest -q -s` 
If you also add a filename to the command  you don't have to run the entire year suite

laerus:
solved with pytest!

scopatz:
woohoo!

Issue #940, opened 2016-05-19
Garbage when running nosetests -q


GuillaumeLeclerc:
I get very strange output when I run `nosetests -q`

```
AssertionError
Could not load xonsh config.
ValueError: Expecting ',' delimiter: line 1 column 17 (char 16)
Xonsh config file is not valid JSON.
total 88
lrwxrwxrwx 1 guillaume guillaume    13 mai   15 21:54 appveyor.yml -> .appveyor.yml
lrwxrwxrwx 1 guillaume guillaume    12 mai   15 21:54 binstar.yml -> .binstar.yml
drwxr-xr-x 3 guillaume guillaume  4096 mai   15 21:54 build
-rw-rw-r-- 1 guillaume guillaume 12155 mai   19 18:34 CHANGELOG.rst
lrwxrwxrwx 1 guillaume guillaume    17 mai   15 21:54 CONTRIBUTING -> docs/devguide.rst
drwxrwxr-x 5 guillaume guillaume  4096 mai   19 18:34 docs
-rw-rw-r-- 1 guillaume guillaume  1573 mai   15 21:54 license
-rw-rw-r-- 1 guillaume guillaume  3114 mai   15 21:54 logo.txt
-rw-rw-r-- 1 guillaume guillaume    17 mai   15 21:54 MANIFEST.in
-rw-rw-r-- 1 guillaume guillaume  1149 mai   19 18:34 README.rst
drwxrwxr-x 2 guillaume guillaume  4096 mai   15 21:54 recipe
-rwxrwxr-x 1 guillaume guillaume  3947 mai   15 21:54 release.xsh
-rw-rw-r-- 1 guillaume guillaume    35 mai   19 18:34 requirements-docs.txt
-rw-rw-r-- 1 guillaume guillaume    33 mai   15 21:54 requirements-tests.txt
-rw-rw-r-- 1 guillaume guillaume    13 mai   15 21:54 requirements.txt
drwxrwxr-x 2 guillaume guillaume  4096 mai   19 18:34 scripts
-rwxrwxr-x 1 guillaume guillaume  6415 mai   19 18:34 setup.py
drwxrwxr-x 8 guillaume guillaume  4096 mai   19 19:05 tests
lrwxrwxrwx 1 guillaume guillaume    11 mai   15 21:54 travis.yml -> .travis.yml
drwxrwxr-x 7 guillaume guillaume  4096 mai   19 18:59 xonsh
[...]

Issue #918, opened 2016-05-17
Switch tests to pytests


scopatz:
B y popular demand...

melund:
I really vote for this, although I dread the work of porting the tests. Not all our `nose` stuff is really pytest compatible. 

It is much more fun to write tests using pytest. Less code and more flexibility. It is also insanely easy to write extension to pytest. The flexibility and extensibility could allow us to create pytest extension that collected `test_*.xsh` files and run them. That shuold be simple.

Maybe we could even make pytest extension that run top level integration tests. So we can test the actual user interface. I have no clue how we approach this, so any help is appreciated. What can we learn from other projects? 

Here is a good introduction to pytest if you haven't used it before:
http://pythontesting.net/framework/pytest/pytest-introduction/

scopatz:
Thanks for adding the details here @melund!

laerus:
we rly need to make this happen before we reach 99% coverage :laughing:

@gforsyth if you could give any hints on where to start on the pytest_port branch i can help with this, or i can just jump in and start brawling with nose.

gforsyth:
Hey @laerus -- that's great!  I've been meaning to dig into that again but I haven't had much time to get in really deep on it (I'm also relatively new to pytest which isn't helping).  I'll merge in recent changes to the branch since a bunch of new tests were added, which may introduce further incompatibilities but the two main pain points were in `test_main.py` and `test_imphooks.py` if you want to take a look at those.  

I'm pretty sure it's due to differences in mocking and patching but I could be wrong on that.

laerus:
we can also use pytest-mock which is a neat plugin

melund:
Let us see how far we can get with plain pytest. But you are right. There are many really neat plug-ins.

laerus:
Mission Accomplished!

PR #945, opened 2016-05-19
[WIP] Pytest port

gforsyth:
This shouldn't go in yet and others are welcome to work on it if they want.  
Regarding several issues:
#942 
#940 
#918 
and others, this is a first stab at changing the testing framework from nose to pytest.  

It's actually not too bad.  The following tests work without any changes needed:
```
test_aliases
test_builtins
test_dirstack
test_foreign_shells
test_inspectors
test_lazy_json
test_lexer
test_ptk_history
test_ptk_multiline
test_replay
test_wizard
```

A few need only minimal changes (some already applied in #944)

```
environ
 change ~from test.tools import~ to ~from tools import~
execer
 change ~setup~ to ~setup_module~
parser
 change ~setup~ to ~setup_module~
history
 change ~from test.tools import~ to ~from tools import~
replay
 works but weird error about command not found: the
```

The only major problems are the following (which I have not addressed):

1. `test_imphooks` -- fails, i think due to differences in mock/patch support
2. `test_main` -- again, patch problem (I think)
3. `test_tools` -- Ok, this works, but on my machine now runs 131680 tests, which is crazy.

Anyway, I'm putting this here so others don't duplicate the effort (which again, wasn't that much).  I'm surprised that there isn't TOO much to do.  That said, if we want to support both nose and pytest, things might get hairy.


melund:
Fantastic. I was just going to open a WIP branch as well :)

I certainly don't think we should support both nose and pytest, and I think we should implement some of our existing test tools (in tests/tools.py) as pytest fixtures. That should really make testing fun and easy.

melund:
Also I would suggest we make a pytest extension collect and run `test_*.xsh` files.

scopatz:
Thanks a million for starting this @gforsyth! I agree with @melund 100%.  I also think it would be nice if we installed the pytest extension so that other projects can use it too.  Similar to how we install the pygments extension or how numpy and pandas add testing functions for their users.

melund:
> I also think it would be nice if we installed the pytest extension so that other projects can use it too. Similar to how we install the pygments extension or how numpy and pandas add testing functions for their users. 

I agree. I have done something similar for one of my own projects. It is pretty easy.. https://github.com/AnyBody-Research-Group/AnyPyTools/blob/master/setup.py#L27

We just have to make the plug-in. If we can run directly from source we could make a fixture to run xonsh code.

gforsyth:
Cool.  I'm probably not going to have time to dig in too deep on this for a while so if you want to tackle part of it, @melund, feel free.  Otherwise I'll come back around to it once I'm done with SciPy prep

gforsyth:
Ok -- so some of the recent test refactoring seems to have made this even easier.  nice job, someone!

On linux I have only two test failures, both related to patching/mocking.  The two "problem" files are `test_main.py` and `test_environ.py`

The `test_environ.py` error is familiar to me because it was popping up in various PRs earlier.  @mitnk might have thoughts on this

```console
test_environ.py:106: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../xonsh/__amalgam__.py:11753: in partial_format_prompt
    return _failover_template_format(template)
../xonsh/__amalgam__.py:11716: in _failover_template_format
    print_exception()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

msg = None

    def print_exception(msg=None):
        """Print exceptions with/without traceback."""
        env = getattr(builtins, '__xonsh_env__', os.environ)
        # flags indicating whether the traceback options have been manually set
>       manually_set_trace = env.is_manually_set('XONSH_SHOW_TRACEBACK')
E       AttributeError: 'dict' object has no attribute 'is_manually_set'

../xonsh/__amalgam__.py:5531: AttributeError
```

Obviously we want to make changes/remove things that are more nose specific, but let's follow Knuth's advice here and just get it working first, yeah?

laerus:
@gforsyth  thank you for updating this branch.
want me to start replacing nose with pytest?

gforsyth:
@laerus -- go for it!

hugo-aa:
@gforsyth I think it's an issue in `tests/tools.py`, you could fix it like this:

```
diff --git a/tests/tools.py b/tests/tools.py
index 5659e28..78f1a81 100644
--- a/tests/tools.py
+++ b/tests/tools.py
@@ -12,7 +12,8 @@ from contextlib import contextmanager
 from nose.plugins.skip import SkipTest

 from xonsh.built_ins import ensure_list_of_strs
-builtins.__xonsh_env__ = {}
+builtins.__xonsh_env__ = Env()
 from xonsh.base_shell import BaseShell
 from xonsh.execer import Execer
 from xonsh.tools import XonshBlockError
```

mitnk:
change to use my right account :)

@gforsyth I think it's an issue in tests/tools.py, you could fix it like this:

```diff
diff --git a/tests/tools.py b/tests/tools.py
index 5659e28..78f1a81 100644
--- a/tests/tools.py
+++ b/tests/tools.py
@@ -12,7 +12,8 @@ from contextlib import contextmanager
 from nose.plugins.skip import SkipTest

 from xonsh.built_ins import ensure_list_of_strs
-builtins.__xonsh_env__ = {}
+from xonsh.environ import Env
+builtins.__xonsh_env__ = Env()
 from xonsh.base_shell import BaseShell
 from xonsh.execer import Execer
 from xonsh.tools import XonshBlockError
```

gforsyth:
That did it for `test_environ.py`.  Thanks, @mitnk!

melund:
@gforsyth 

Maybe update the CI scripts to run pytest: 

AppVeyor.yaml
```
version: 0.3.4.{build}
os: Windows Server 2012 R2
install:
- C:\Python35\Scripts\pip install ply pyreadline pytest pygments prompt_toolkit
build_script:
- C:\Python35\python setup.py install
test_script:
- C:\Python35\Scripts\py.test
```

laerus:
i'm seeing exceptions at `base_shell.py` and `test_main.py`

laerus:
this passes the test `pytest -s test_main.py` , `-s` is the quiet mode, very strange

gforsyth:
I'm working on more changes -- stripping out remaining nose functionality.  We'll see if that makes any difference

laerus:
mistake* the `-s` option is the no capture option

laerus:
i think its only the first case in the `test_main.py` that calls `premain` with an empty list

laerus:
the rest pass the test, with pytest syntax, same mocking

laerus:
```
with patch('xonsh.main.Shell', Shell), mock_xonsh_env({}): 
    xonsh.main.premain([])                                                  
    assert not builtins.__xonsh_env__.get('XONSH_LOGIN') 
```
if you put a *not* this passes the test, so its a premain problem?

laerus:
`premain` is a little buggy at the current master, #1277 fixes some bugs with arguments, dunno if thats relevant, but i've written more tests in this PR including these 4 cases.

laerus:
`test_imphooks.py` fails at this:

```
        if currenv is None and hasattr(builtins, '__xonsh_env__'):
>           currenv = builtins.__xonsh_env__.detype()
E           AttributeError: 'dict' object has no attribute 'detype'
```

we may need to add this detype to the xonsh env mock

laerus:
checkout #1317 , also poor travis i can't find the way to cancel jobs

scopatz:
@laerus If you go to the "Pull Requests" tab you can look at all the current and past jobs.  If you click on one, you can then cancel it with the X in a circle in the upper right, just to the right of the PR number.

gforsyth:
This is ready for a look and also I need eyes on the Python 3.4 failures.  `test_replay.py` is failing with complains about missing attributes.  not sure why it's fine on python 3.5 and I can't look at it right now.

```
    def intensify_colors_on_win_setter(enable):
        """Resets the style when setting the INTENSIFY_COLORS_ON_WIN
        environment variable.
        """
        enable = to_bool(enable)
>       delattr(builtins.__xonsh_shell__.shell.styler, 'style_name')
E       AttributeError: 'NoneType' object has no attribute 'style_name'
```

laerus:
the first test in `test_main.py` fails because we need to mock `sys.stdin.istty()` , we assume we get into `XonshMode.interactive` but we are actually getting into `XonshMode.script_from_file` when we run the tests and so XONSH_LOGIN is False

laerus:
i can't reproduce the fail in `test_replay.py` in python3.4

laerus:
i'm getting an error in `test_parser.py` in python3.4 :

```
E           x(*[("a", 2)], *[("v", 3)], **{"c": 5})
E                          ^
E       SyntaxError: invalid syntax
```

Russell-Jones-OxPhys:
> (why do we want to support 3.4 btw?)

Well, Ubuntu 14.04 uses 3.4, and EOL for that is 2019. Would that be sufficient? Even if we could start migration now (which we can't, though that's mainly because of an infrastructure change) it would be quite a while before we could move all our desktops over to 16.04.

laerus:
@Russell-Jones-OxPhys yes you're are right, it was more of a rhetorical question, debian-stable uses 3.4, even thought they may update to 3.5 at some point, debian-testing is using 3.5.2 and its probably going to get into 3.6 or maybe higher until next debian stable (~3years into the future) release. So maybe at some point around 3.6-3.7 ppl will start building stuff for >3.5 only. (good ol' python)

xonsh/tools.py
```diff
@@ -118,7 +118,7 @@ def decode_bytes(path):
     env = getattr(builtins, '__xonsh_env__', os.environ)
     enc = env.get('XONSH_ENCODING', DEFAULT_ENCODING)
     return path.decode(encoding=enc,
-                       errors=env.get('XONSH_ENCODING_ERRORS'))
+                       errors=env.get('XONSH_ENCODING_ERRORS') or 'strict')
```
laerus: i'm not really sure about this one, if this is not a bug we'll have to mock `XONSH_ENCODING_ERRORS`

gforsyth:
> Could the test be made conditional on the version being run?

Supposedly.  We were doing this in `nose` quite easily but `pytest` seems to think it should try to parse code, even if it's been told to skip it.

gforsyth:
Ok -- we'll see what CI says, but this should be good except for the EnvPath bug (that we think is a bug, right?)

gforsyth:
It's apparently Christmas in June, all.  The AppVeyor build actually succeeded.

laerus:
`test_parser.py` has still many nose remnants, will check again what's up with EnvPath

laerus:
this test complains about EnvPath
```
    def test_env_path():
        # lambda to expand the expected paths
        expand = lambda path: os.path.expanduser(os.path.expandvars(path))
        getitem_cases = [
            ('xonsh_dir', 'xonsh_dir'),
            ('.', '.'),
            ('../', '../'),
            ('~/', '~/'),
            (b'~/../', '~/../'),
        ]
        for inp, exp in getitem_cases:
            obs = EnvPath(inp)[0]  # call to __getitem__
>           assert expand(exp) == obs
E           assert '/home/laerus/' == '~/'
E             - /home/laerus/
E             + ~/
```

gforsyth:
fix incoming

gforsyth:
EnvPath's `__getitem__` method calls `expandvars` which only performs
expansion if the ENV_VAR `EXPAND_ENV_VARS` is `True`.  And since our
environments aren't bleeding over into the tests anymore, this wasn't
being set.  So a quick mock env added in to the expansion tests and
we're good to go.

laerus:
appveyor fails on this:

```
def _executables_in_windows(path):                                              
    if not os.path.isdir(path):                                                 
        return                                                                  
    extensions = builtins.__xonsh_env__['PATHEXT']                              
    ...
```
with:
```
>       extensions = builtins.__xonsh_env__['PATHEXT'] 
E       AttributeError: module 'builtins' has no attribute '__xonsh_env__' 
```

we may need another mock

gforsyth:
I think this is ready for another look

tests/test_contexts.py
```diff
@@ -35,29 +33,29 @@ def block_checks_glb(name, glbs, body, obs=None):
...
-        yield assert_equal, v, glbs[k]
+        assert v == glbs[k]
     if isinstance(body, str):
         body = body.splitlines()
-    yield assert_equal, body, block.lines
-    yield assert_is, glbs, block.glbs
-    yield assert_is, None, block.locs
+    assert body == block.lines
+    assert glbs is block.glbs
+    assert block.locs is None
```
scopatz: Do each of these asserts become their own dot?  Sorry I don't know much about pytest yet...

gforsyth: No, I'm pretty sure it's one dot per "test", but pytest will tell you which one failed (if one does)

gforsyth: Where "test" is a function (or class, etc)

scopatz: It seems like pytest does support yield: https://pytest.org/latest/nose.html  Also, will Pytest keep executing the following asserts if one higher up fails?

laerus: nope it will stop at the first `assert False` in each individual test, pytest favors the style of writing  tests with few or even one assert.

xonsh/tools.py
scopatz: I don't think that this is correct.

tests/test_contexts.py
melund: It is easy to get pytest to run all asserts and report a dot for each. We should just parameterize the tests, by moving the cases outside the functions. https://pytest.org/latest/parametrize.html

scopatz: Well, many of our tests assume independence of yielded asserts. This doesn't block this PR, but I would recommend that we return to that mode.

scopatz:
I more or less think that this is good. However, the CONTRIBUTING guide probably needs to be updated.

tests/test_contexts.py
melund: There is a real potential to make the tests a lot simpler.

scopatz: > It is easy to get pytest to run all asserts and report a dot for each. We should just parameterize the tests, by moving the cases outside the functions. https://pytest.org/latest/parametrize.html

@melund I think that works for most cases.  However, for cases like this won't it create a whole new block object for each case?  The way this works on current master is that it checks consistency for a single block instance, rather than recreating many (hypothetically) identical block instances.

tests/test_tools.py
```diff
@@ -310,7 +315,7 @@ def test_subexpr_from_unbalanced_parens():
         ]
     for expr, exp in cases:
         obs = subexpr_from_unbalanced(expr, '(', ')')
-        yield assert_equal, exp, obs
+        assert exp == obs
```
scopatz: Tests like these should be parameterizations, since all cases are independent of one another and a failure in one should not prevent the other tests from running.

melund: Yes. And when they fail it will be much more clear what is wrong.

tests/test_contexts.py
melund: I guess. But can you give an example where it parameterization would not work, and where we still want all the asserts to execute.

tests/test_tools.py
laerus: i can start migrating to pytest, but i think this doesn't prevents us from merging into master, cause each new thing that gets added will have tests in pytest and we wont need to rewrite these aswell.

gforsyth: @laerus -- I'm working on parametrizing test_tools right now, so I'll finish that and then revise the contribution guide

laerus:
along with CONTRIBUTING `scripts/run_tests.xsh` needs a change aswell

tests/test_contexts.py
scopatz: I think this is exactly that case. The asserts are all different from one another (and thus not parameterizable) but it would still be advantageous to have them all run.

scopatz: Again, I don't want this to block, because we can change this in future PRs and this PR is big enough :)

scopatz:
Thanks @gforsyth! Also, we can probably remove the `run_tests.xsh` script. Does anyone use it?

laerus:
proposal: we can also use pytest-mock which is a way cleaner way to write mocks.

scopatz:
> proposal: we can also use pytest-mock which is a way cleaner way to write mocks.

Let's hold off on too much more with this PR.

melund:
I agree. Let us wait. Also I think we can accomplish what we need with fixtures and monkeypatch which is built-in to pytest.

melund:
I can just make the changes to CONTRIBUTING

gforsyth:
> Let's hold off on too much more with this PR.

Yeah.  I just stopped with the parametrization because this is already huge.  I'm going to update contributing and remove runtests and that's it

laerus:
i meant after this gets merged into master, when we start migrating the tests, except if we want to migrate everything before we merge it

melund:
Go ahead @gforsyth. I think we can take further stuff in separate PRs.

melund:
@laerus. I could imagine something like this in a `conftest.py`:
```python
@pytest.fixture()
def xonsh_env(monkeypatch):
    monkeypatch.setattr(builtins, '__xonsh_env__', Env())
    monkeypatch.setattr(builtins, '__xonsh_ctx__', {})
    ....
    ....
    return builtins.__xonsh_env__
```

And then in any tests modules we could just do the following to run in the default mocked env:

```python
def test_myfunc(xonsh_env):
    xonsh_env['HOME'] = 'some/path'
    assert True
```

If we needed something special for many test functions:

 ```python
@pytest.fixture
def tools_env(xonsh_env):
    xonsh_env.update({'EXPAND_ENV_VARS': True, 
                                       'XONSH_ENCODING_ERRORS':'strict'})
    return xonsh_env

def test_myfunc(tools_env):
    assert True
```

melund:
So the CI tests completed. I think this is ready to go in.

scopatz:
Congrats all!

laerus:
@melund there is also the pytest.yield_fixture for easier setup, teardown

melund:
@laerus . It is a cool feature. I have used it for other projects as well. In this case it will just be handled by `monkeypatch`, so in my simple example it is not needed.

laerus:
@melund ah yea. monkeypatch does teardown aswell. very cool

PR #947, opened 2016-05-19
Which updates

melund:
Further updates to make '-p/--plain' the default output of which. See the discussion with @adqm #934 

This is how the update works now:

```
$ which ls

$ which 2to3

$ which jobs
jobs
```

```
$ which ls -v
aliases['ls'] = ['ls', '--show-control-chars', '--color', '-X']

$ which 2to3 -v
C:\Users\mel\Anaconda3\envs\xonshdev\Scripts\2to3.exe (from $PATH element 2)

$ which jobs -v
aliases['jobs'] = <function jobs at 0x0000026E4AB54400>
```
I have no idea why it writes the memory location of the functions. If I manually do `aliases['jobs']` I get `<function xonsh.jobs.jobs>`

If the `-p/-plain` is added with the `-v/--verbose`, then plain takes priority. This is to ensure that people can override verbose if they have an alias like this (`which -> ['which', '--verbose']`)
```
$ which ls -v -p
```

The `-a/--all` option would automatically be verbose:
``` 
$ which ls -a
aliases['ls'] = ['ls', '--show-control-chars', '--color', '-X']
C:\Program Files\Git\usr\bin\ls.exe (from $PATH element 17)
```
But again can be overridden with `-p/--plain`:
``` 
$ which ls -a -p
C:\Program Files\Git\usr\bin\ls.exe
```

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -362,12 +362,13 @@ def which(args, stdin=None, stdout=None, stderr=None):
                         help='Display the version of the python which module '
                         'used by xonsh')
     parser.add_argument('-v', '--verbose', action='store_true', dest='verbose',
-                        help='Show extra information on for example near misses')
+                        help='Print out how matches were located and show '
+                        'near misses on stderr')
     parser.add_argument('-p', '--plain', action='store_true', dest='plain',
                         help='Do not display alias expansions or location of '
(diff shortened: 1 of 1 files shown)


adqm:
I am feeling good about this.  Merging in now.  Sorry for all the back-and-forth!

PR #948, opened 2016-05-19
Fixed echo @("foo") > bar

scopatz:
This should probably go in for the release tomorrow, if possible


adqm:
Woo-hoo; looks good to me.  Does this need a changelog entry?

scopatz:
I don't think so because it fixes a bug that wasn't in the last release :)

adqm:
Okay, great!  In that case, I'll merge these both in.

PR #949, opened 2016-05-20
str().split() | grep exit now throws a SyntaxError

### LINKED ISSUE #927: unresponsive after: time.time() | grep "exit"
If I run:
```python
import time
time.time() | grep "exit"
```
I fully understand that this is a misguided series of commands (trying to mix python mode and subprocess mode). However the fact that it makes my terminal unresponsive and unrecoverable (Ctrl+C and Ctrl+Z do nothing) seems undesirable. 

I'm using prompt-toolkit.

scopatz:
This was a subtle bug caused by the new and/or stuff.  Should close #927

### DIFF
--- a/tests/test_execer.py
+++ b/tests/test_execer.py
@@ -83,10 +83,6 @@ def test_bad_indent():
             'x = 1\n')
     assert_raises(SyntaxError, check_parse, code)
 
-def test_bad_rhs_subproc():
-    code = 'str().split() | grep exit\n'
-    assert_raises(SyntaxError, check_parse, code)
-
 def test_good_rhs_subproc():
     # nonsense but parsebale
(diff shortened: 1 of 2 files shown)


adqm:
Looks good to me!

I'm getting `NameError` now instead of `SyntaxError` with the example you described (presumably because it is valid python).  Is that expected / does that matter?

Also, does this need a changelong entry?

scopatz:
> I'm getting NameError now instead of SyntaxError with the example you described (presumably because it is valid python). Is that expected / does that matter?

Yeah, probably. :)

> Also, does this need a changelong entry?

I don't think so again, because it fixes an error that wouldn't have made sense or have been relevant in the last release.

PR #958, opened 2016-05-20
Print signal messages

### LINKED ISSUE #955: Segmentation faults and other special return codes should print corresponding messages
If an application like this:

```c
int main() {
    char* p = 0;
    return *p; // Uh-oh!
}
```

segfaults, Bash shows the following:

```
ryan@DevPC-LX:~/langtest/xonsh$ ./seg
Segmentation fault (core dumped)
ryan@DevPC-LX:~/langtest/xonsh$ 
```

But Xonsh shows:

```
ryan@DevPC-LX ~/langtest/xonsh $ ./seg

ryan@DevPC-LX ~/langtest/xonsh $
```

See:

http://tldp.org/LDP/abs/html/exitcodes.html

and:

https://en.wikipedia.org/wiki/Unix_signal#POSIX_signals

Would you be interested in a pull request for this?

kirbyfan64:
Closes #955.

Note that you should probably add a comment somewhere about how `wait_for_active_job` assigns to `obj.signal` if there was a signal sent. Took me ~30 minutes of debugging to figure out why `returncode` was always 0...

Also, I didn't see a place to add tests for this...


scopatz:
Thanks for putting this in @kirbyfan64! It looks like it fails CI, though...

kirbyfan64:
@scopatz Fixed!

xonsh/built_ins.py
```diff
@@ -599,6 +610,16 @@ def run_subproc(cmds, captured=False):
                 errout = errout.replace('\r\n', '\n')
                 procinfo['stderr'] = errout
 
+    if getattr(prev_proc, 'signal', None):
+        sig, core = prev_proc.signal
+        try:
+            sig_str = SIGNAL_STRINGS[sig]
```
scopatz: Probably better to do a `SIGNAL_STRINGS.get()` here rather than waiting on an exception.

xonsh/built_ins.py
```diff
@@ -41,6 +41,17 @@
 if ON_POSIX:
     AT_EXIT_SIGNALS += (signal.SIGTSTP, signal.SIGQUIT, signal.SIGHUP)
 
+SIGNAL_STRINGS = {
```
scopatz: Would you mind changing this to `SIGNAL_MESSAGES` or similar?

xonsh/built_ins.py
```diff
@@ -599,6 +610,16 @@ def run_subproc(cmds, captured=False):
...
+    if getattr(prev_proc, 'signal', None):
+        sig, core = prev_proc.signal
+        try:
+            sig_str = SIGNAL_STRINGS[sig]
+        except KeyError:
+            pass
+        else:
+            if core:
+                sig_str += ' (core dumped)'
+            print(sig_str)
```
scopatz: This should probably be printed to stderr, rather than stdout.  Just do `print(sig_str, file=sys.stderr)`

scopatz:
Thanks again @kirbyfan64, testing this might be kind of a pain... But you should at least put in a changelog entry :)

xonsh/built_ins.py
kirbyfan64: Well, I was following the Python "easier to ask forgiveness than permission", but I'll use `get` if you prefer.

kirbyfan64:
@scopatz Updated! Side note: I really like Xonsh's Git integration that shows the current branch!

scopatz:
> @scopatz Updated! 

Thanks!  Could you add a changelog entry, though too?

> Side note: I really like Xonsh's Git integration that shows the current branch!

Awesome, thanks!  It works with hg too, for what that is worth :).

kirbyfan64:
@scopatz Whoops, forgot that part! Done.

CHANGELOG.rst
```diff
@@ -37,6 +37,8 @@ Current Developments
   ``--cache-everything`` when starting xonsh.
 * Added a workaround to allow ctrl-c to interrupt reverse incremental search in
   the readline shell
+* When a subprocess exits with a signal (e.g. SIGSEGV), a message is printed,
+  similar to Bash.
```
scopatz: Haha whoops, can you merge from master and then move this bit to the new Current Developments section, please? Sorry for the run around, we did the v0.3.0 release in between when you last pulled or forked and now.

kirbyfan64:
@scopatz Gah...done!

scopatz:
Thanks!  Looks great to me!

melund:
@scopatz @kirbyfan64 ... This PR broke master on Windows. I pushed an urgent fix #958

PR #962, opened 2016-05-21
Job Control Fixes (Built on #867)

adqm:
This builds on @GuillaumeLeclerc's great work in #867 refactoring xonsh's job control.  I think all of those changes are ready to be merged in.

Beyond the changes made there, this patch:

* fixes a few PEP8 issues
* brings back the "current job" text in the title bar
* makes windows job control kind of work again (as well as it ever did)
* should fix the issue with launching commands in the background (by bringing back code that looks like it got lost as part of #672)

@scopatz, et al, do you mind taking a look?

### DIFF
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -20,14 +20,11 @@ if ON_WINDOWS:
     def _continue(job):
         job['status'] = 'running'
 
-
     def _kill(obj):
         return obj.kill()
 
-
     def ignore_sigtstp():
         pass
-
 
     def _set_pgrp(info):
         pass
@@ -57,19 +54,16 @@ if ON_WINDOWS:
 
 else:
     def _continue(job):
-        _sendSignal(job, signal.SIGCONT)
-
+        _send_signal(job, signal.SIGCONT)
 
     def _kill(job):
-        _sendSignal(job, signal.SIGKILL)
+        _send_signal(job, signal.SIGKILL)
 
-    def _sendSignal(job, signal):
+    def _send_signal(job, signal):
         os.killpg(job['pgrp'], signal)
-
 
     def ignore_sigtstp():
         signal.signal(signal.SIGTSTP, signal.SIG_IGN)
-
 
     def _set_pgrp(info):
         try:
@@ -77,21 +71,19 @@ else:
         except ProcessLookupError:
             pass
 
-
     _shell_pgrp = os.getpgrp()
 
     _block_when_giving = (signal.SIGTTOU, signal.SIGTTIN, signal.SIGTSTP)
-
 
     def _give_terminal_to(pgid):
         # over-simplified version of:
         #    give_terminal_to from bash 4.3 source, jobs.c, line 4030
         # this will give the terminal to the process group pgid
         if _shell_tty is not None and os.isatty(_shell_tty):
-            oldmask = signal.pthread_sigmask(signal.SIG_BLOCK, _block_when_giving)
+            oldmask = signal.pthread_sigmask(signal.SIG_BLOCK,
+                                             _block_when_giving)
             os.tcsetpgrp(_shell_tty, pgid)
             signal.pthread_sigmask(signal.SIG_SETMASK, oldmask)
-
 
     # check for shell tty
     try:
@@ -110,12 +102,11 @@ else:
             if not task['bg'] and task['status'] == RUNNING:
                 selected_task = tid
                 break
-        if selected_task == None:
+        if selected_task is None:
             return
         tasks.remove(selected_task)
         tasks.appendleft(selected_task)
         return get_task(selected_task)
-
 
     def wait_for_active_job():
         """
@@ -242,7 +233,7 @@ def fg(args, stdin=None):
         return '', 'Cannot bring nonexistent job to foreground.\n'
 
     if len(args) == 0:
-        act = tasks[0] # take the last manipulated task by default
+        act = tasks[0]  # take the last manipulated task by default
     elif len(args) == 1:
         try:
             act = int(args[0])
@@ -271,7 +262,7 @@ def bg(args, stdin=None):
     single number is given as an argument, resume that job in the background.
     """
     res = fg(args, stdin)
-    if res == None:
+    if res is None:
         curTask = get_task(tasks[0])
         curTask['bg'] = True
         _continue(curTask)
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -29,7 +29,6 @@ if ON_WINDOWS:
     def _set_pgrp(info):
         pass
 
-
     def wait_for_active_job(signal_to_send=None):
         """
         Wait for the active job to finish, to be killed by SIGINT, or to be
@@ -43,10 +42,7 @@ if ON_WINDOWS:
         if active_task is None:
             return
 
-        pgrp = active_task['pgrp']
         obj = active_task['obj']
-        # give the terminal over to the fg process
-        _give_terminal_to(pgrp)
 
         _continue(active_task)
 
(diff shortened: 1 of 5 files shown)


xonsh/jobs.py
```diff
@@ -5,25 +5,27 @@
...
 import builtins
-from subprocess import TimeoutExpired
+from subprocess import TimeoutExpired, check_output
 from io import BytesIO
+from collections import deque
 
 from xonsh.tools import ON_WINDOWS
 
+RUNNING = 1
+STOPPED = 0
```
scopatz: Do we want these as integers, rather than string flags?  And if ints, shouldn't it be an enum?

scopatz:
Thanks a million @adqm!  This looks quite good.  The only major thing I would add would be some tests for the ampersand lexer, since this looks like it increased in complexity significantly.  It may be true that most of these cases are covered by the parser tests, but some new lexer ones would be nice too.  Other than that, just a changelog entry :)

melund:
Yes. Too bad this doesn't work perfectly on windows. Starting in the background with a trailing `&` works. But there is no way to suspend a foreground process. @BYK do you know any way this could be done?

adqm:
> The only major thing I would add would be some tests for the ampersand lexer, since this looks like it increased in complexity significantly.

@scoptaz: The interesting thing is that it didn't actually increase in complexity at all. Before #672, this is what it was doing (though it did it in a slightly different way).  So I think we already have test cases for everything (no actual new features here at all).

>Starting in the background with a trailing & works. But there is no way to suspend a foreground process.

@melund, @BYK: AFAIK, that isn't possible/easy.  In #223, when we first introduced the rudimentary job control, I looked around a bit (since this isn't really doable otherwise) and found [https://github.com/giampaolo/psutil](psutil), which somehow hacks in the ability to suspend/resume processes on Windows.  We could consider adding that as a Windows dependency (optional or otherwise).  If we did that, then I think we could make something that kind of works (though it still might not be full-featured), and we could fall back to what we have now if `psutil` isn't installed.

melund:
A change log, and then this could make it for 3.1:)

scopatz:
> So I think we already have test cases for everything (no actual new features here at all).

Great! I suspected this may have been the case.

>  could fall back to what we have now if psutil isn't installed. 

+1, I think that this could certainly work and make things better.  I worry a bit about having two interfaces to job/control. But it doesn't seem like there would be a way around this.  It would be great if psutils way just in the standard library :)

adqm:
Okay.  Just pushed two changelog entries.  I think this is ready to go!

> I worry a bit about having two interfaces to job/control

I will work on improving Windows job control in a new branch and submit a PR once I think I have something reasonable.  Then we can see what we think.

In some sense, we already have many platform dependent things, including two different versions of many pieces of `jobs.py`.  In this case, we would just be adding a third.  I haven't looked into `psutil` in any depth yet, but it could be that it will have some pieces that are useful regardless of platform, in which case we might want to rely on it if it exists, even if a user isn't running Windows.

scopatz:
> In some sense, we already have many platform dependent things, including two different versions of many pieces of jobs.py

Point taken :)  I guess I was saying that it might be worth it to have a small interface layer so that we can handle as many job control mechanisms as we want.  Thinking ahead, this could be useful for running jobs via SSH or via some other remote protocol.  Rather than continuing to hack on the existing jobs, some abstraction might be super useful.

> even if a user isn't running Windows.

Exactly, looking at psutils it seems like we might even want to use this on posix systems if it is available.  We may want to put this in conda-forge :)

scopatz:
Also @adqm, would you mind merging this from master, @adqm so that it doesn't have all of the new syntax changes in the diff. Thanks!

adqm:
okay, just rebased on to master and reverted the string -> integer change.

scopatz:
Thanks @adqm and @GuillaumeLeclerc! Things seem to work again on linux! We'll definitely want to do a v0.3.1 release before PyCon.

BYK:
@adqm - sorry for being late to the game. PSUtil looks awesome but it also looks quite huge. What specific feature do we need from there here?

If we can break up some Windows-needs I can start working on them soon.

adqm:
> sorry for being late to the game. PSUtil looks awesome but it also looks quite huge. What specific feature do we need from there here?

The hope was to use the "suspend" feature from psutil to allow a more "real" job control in Windows.  I was thinking that we couldn't include `psutil`, but rather have it as an optional dependency.  It's also possible we could use something like `pywin32` instead to mimic what `psutil` does with `suspend`.

I looked into it a bit, and I found that the hard part was actually getting the `^Z` to be handled properly (since the `signal` module doesn't really work on Windows, and `^Z` is treated as EOF).  I never actually figured out how to make that work.

It's possible we could do a custom binding (for both readline and ptk) on Windows to handle that.  Then, I would probably think we could:

* If the user has `psutil` installed, we can make that binding, wrap everything in `psutil.Process`, and call `suspend` for that process (and also handle its entry in the tasks queue) when `^Z` is pressed.  We could still treat `^C` the same way.
* If `psutil` isn't installed, we can fall back to what we have now.

Though it seems like `cmd.exe` always interprets `^Z` as EOF, so I'm not sure even the custom keybindings would help there (since I'm not sure the literal `^Z` is reaching the process).

Anyway, does that sound reasonable?  Do you have other ideas?

melund:
Not sure that it is possible without resorting to hacks. I stumbled on this http://www.codeproject.com/Articles/2964/Win-process-suspend-resume-tool

 I guess we could implement the same system calls. It is not totally safe since it is based on pausing threads. But it seems simple.

BYK:
Okay so I don't care about `^Z` for now since I think that's a secondary concern. What @melund has dug up looks pretty interesting but also hacky. I was thinking may be just assigning the lowest priority to a task would be the "Windows-native" equivalent of suspension?

melund:
Maybe I am mixing things here. On linux you can use Ctrl-z to get back focus to xonsh by suspending the application. Right? That is not possible in windows. 

On windows, I can start something in the background with a trailing `&`. If I later foreground that process with `fg` then I can't get the terminal back until I close the process.Ctrl-z to suspend/background jobs doesn't work on windows.

scopatz:
Yeah, suspension and backgrounding are a little different:

* backgrounding a process has it continue running, but not block i/o
* suspending will actually pause the process until it is brought back.

[This page](http://www.linusakesson.net/programming/tty/) really explained it all to me.  (80% of the page is comments so don't be scared.)  My understanding is that Windows only has backgrounding, and so suspension is obtained by backgrounding and setting the priority as low as possible, which is a reasonable solution.

melund:
Ahh. Backgrounding is probably a reasonable solution. Backgrounding works when starting a new process (appending`&`). 

But there is no way to background a process if it is started without `&`, or if it is later foregrounded with `fg`...

melund:
@adqm. Can we hook up `Ctrl-Z` to background processes on windows, instead of suspending which doens't work.

adqm:
@melund, I think the hard part is actually getting Windows terminals (or at least cmd.exe) to interpret the ctrl-z as a ctrl-z, and not as EOF.

maybe with other terminals, though?

adqm:
Or, it might be possible to bind ctrl-z inside of ptk and/or readline...

If we can make that keybinding happen, then setting that up to background the process would be fine.  Then again, if we can do that, I think it would make sense to try to emulate what happens on *nix (including suspending) if that's possible (which I think it is, via the thread suspension hack).  But if that turns out not to be feasible, then just backgrounding would be good.

gforsyth:
Hey @adqm -- it's definitely possible to do that in PTK, at least.

Putting this in `xonsh/ptk/key_bindings.py` works:

```python
    @handle(Keys.ControlZ)
    def _(event):
        print('hi!')
```

gforsyth:
Although I guess @melund should try that out and make sure that PTK gets to it and not the terminal emulator

melund:
@adqm @gforsyth ... Ctrl-Z keybinding works fine with cmd.exe. If we could use that for backgrounding  (maybe with low priority) that would be awesome. There really should be a way to get foreground processes out of the way.

PR #963, opened 2016-05-21
Fix broken master on Windows introduced by #958.

melund:
#958 broke master for windows. This commit ensures that POSIX only signals are not used on Windows


scopatz:
whoops! thanks for catching that!

PR #970, opened 2016-05-21
Allow python -m xonsh

Carreau:
It seems more pythonistic than `python -m xonsh.main`

--- 

Thoughts?

### DIFF
--- a/xonsh/__main__.py
+++ b/xonsh/__main__.py
@@ -1,0 +1,2 @@
+from xonsh.main import main
+main()


scopatz:
Thanks @Carreau - It is a great idea.  Probably better to implement this in `__main__.py` or move rename the current `__main__.py`: https://docs.python.org/3/library/__main__.html

Carreau:
I'm confused, this **does** introduce a `__main__.py`, and adding an `if __name__ == '__main__':` in `__init__.py` does not have the same effect.

Or are you asking me to **move** `xonsh/main.py` (without underscore) to `xonsh/__main__.py` ?
That would be a (painful?) change of API, and it's hard/weird to import `__main__` if you need it.

Second possibility: I'm tired and don't understand a thing.

melund:
It should be enough to rename `main.py`to `__main__.py`. I don't think it would be a big issue.

Btw. can we run xonsh directly from source with this command? `python -m xonsh -c "xonsh command"`. I was thinking about this for the pytest framework. Some top level things could be tested this way. But what about the parser tables?

scopatz:
Ahhh I am so sorry @Carreau!  I thought this was in init for some reason.  I was clearly the tired one, as I responded before breakfast. 

In this case, I think this looks good.  Only thing I would suggest is an absolute import. `from xonsh.main ...` for consistency.

scopatz:
I sort of agree with @Carreau, @melund.  I think it is nice to keep the `__x__` files as simple as possible.

Carreau:
> In this case, I think this looks good. Only thing I would suggest is an absolute import. from xonsh.main ... for consistency.

Done.

scopatz:
Sweet! Thanks @Carreau. You are the best and least sleepy!

PR #974, opened 2016-05-21
added store_as_str to vardocs

### LINKED ISSUE #930: TypeError
Running `xonfig wizard` and trying to set `$HISTCONTROL` raises an exception.

```
<snip>
$HISTCONTROL
A set of strings (comma-separated list in string form) of options
that determine what commands are saved to the history list. By
default all commands are saved. The option 'ignoredups' will not save
the command if it matches the previous command. The option
'ignoreerr' will cause any commands that fail (i.e. return non-zero
exit status) to not be added to the history list.
default value: set()
current value: {'ignoreboth'}
>>> ignoredups
Converted input 'ignoredups' to {'ignoredups'}.
<snip>
TypeError: {'ignoredups'} is not JSON serializable
```
$ xonsh --version
('xonsh/0.2.7',)

Installed from pip.

scopatz:
This should fix #930

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -262,6 +262,7 @@ def make_envvars():
            '{docstr}\n\n'
            '**configurable:** {configurable}\n\n'
            '**default:** {default}\n\n'
+           '**store_as_str:** {store_as_str}\n\n'
            '-------\n\n')
     for var in vars:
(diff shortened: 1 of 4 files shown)


melund:
Fixed it for me.

PR #975, opened 2016-05-21
adds scandir to platform

funkyfuture:
as discussed.

i also added docs for the `platform` module. however they render also the values of constants from the build environment. there seems to be no way to disable this on a `automodule` level. i'd rather add them manually to the `rst` file. any other idea?

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -1,32 +1,38 @@
 # -*- coding: utf-8 -*-
 """Tests the xonsh lexer."""
-from __future__ import unicode_literals, print_function
 import os
+import random
+from tempfile import TemporaryDirectory
+import stat
 
 import nose
 from nose.tools import assert_equal, assert_true, assert_false
 
 from xonsh.lexer import Lexer
-from xonsh.tools import (subproc_toks, subexpr_from_unbalanced, is_int,
-    always_true, always_false, ensure_string, is_env_path, str_to_env_path,
-    env_path_to_str, escape_windows_cmd_string, is_bool, to_bool, bool_to_str,
+from xonsh.tools import (
+    subproc_toks, subexpr_from_unbalanced, is_int, always_true, always_false,
+    ensure_string, is_env_path, str_to_env_path, env_path_to_str,
+    escape_windows_cmd_string, is_bool, to_bool, bool_to_str,
     ensure_int_or_slice, is_float, is_string, check_for_partial_string,
-    argvquote)
+    argvquote, executables_in)
 
 LEXER = Lexer()
 LEXER.build()
 
 INDENT = '    '
 
+
 def test_subproc_toks_x():
     exp = '![x]'
     obs = subproc_toks('x', lexer=LEXER, returnline=True)
     assert_equal(exp, obs)
 
+
 def test_subproc_toks_ls_l():
     exp = '![ls -l]'
     obs = subproc_toks('ls -l', lexer=LEXER, returnline=True)
     assert_equal(exp, obs)
+
 
 def test_subproc_toks_git():
     s = 'git commit -am "hello doc"'
@@ -34,17 +40,20 @@ def test_subproc_toks_git():
     obs = subproc_toks(s, lexer=LEXER, returnline=True)
     assert_equal(exp, obs)
 
+
 def test_subproc_toks_git_semi():
     s = 'git commit -am "hello doc"'
     exp = '![{0}];'.format(s)
     obs = subproc_toks(s + ';', lexer=LEXER, returnline=True)
     assert_equal(exp, obs)
 
+
 def test_subproc_toks_git_nl():
     s = 'git commit -am "hello doc"'
     exp = '![{0}]\n'.format(s)
     obs = subproc_toks(s + '\n', lexer=LEXER, returnline=True)
     assert_equal(exp, obs)
+
 
 def test_subproc_toks_indent_ls():
     s = 'ls -l'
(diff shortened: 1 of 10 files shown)


scopatz:
Great! thanks for putting this in!

xonsh/platform.py
```diff
@@ -96,12 +115,68 @@ def is_readline_available():
...
+""" Default string encoding. """
+
+
+def scandir(path):
+    """ This is either `os.scandir` on Python 3.5+ or a function providing a
+    compatibility layer for it.
+    This function is recommended for iterations over directory entries at a
+    significantly higher speed than `os.listdir` on Python 3.5+.
+    """
+    pass
```
scopatz: This seems unnecessary as it will just be overwritten.

xonsh/tools.py
```diff
@@ -319,11 +318,10 @@ def suggest_commands(cmd, env, aliases):
                 suggested[a] = 'Alias'
 
     for d in filter(os.path.isdir, env.get('PATH')):
-        for f in os.listdir(d):
-            if f not in suggested:
-                if levenshtein(f.lower(), cmd, thresh) < thresh:
-                    fname = os.path.join(d, f)
-                    suggested[f] = 'Command ({0})'.format(fname)
+        for name, path in ((x.name, x.path) for x in scandir(d)
+                           if x.name not in suggested):
```
scopatz: No reason to double loop here.  Please just make a single loop.  Loops in Python are super slow.

xonsh/imphooks.py
```diff
@@ -49,7 +50,7 @@ def find_spec(self, fullname, path, target=None):
                 continue
             if not os.path.isdir(p):
                 continue
-            if fname not in os.listdir(p):
+            if fname not in (x.name for x in scandir(p)):
```
scopatz: Wow, didn't know that would work.  Must still be stuck in Python2 :(

scopatz: Thanks for teaching me something new :)

xonsh/aliases.py
```diff
@@ -412,9 +412,9 @@ def which(args, stdin=None, stdout=None, stderr=None):
         for abs_name, from_where in matches:
             if ON_WINDOWS:
                 # Use list dir to get correct case for the filename
-                # i.e. windows is case insesitive but case preserving
+                # i.e. windows is case insensitive but case preserving
                 p, f = os.path.split(abs_name)
-                f = next(s for s in os.listdir(p) if s.lower() == f.lower())
+                f = next(s.name for s in scandir(p) if s.name.lower() == f.lower())
```
scopatz: It would be nice to store an `flower = f.lower()` variable once here, if you feel like making that change.

xonsh/platform.py
```diff
@@ -96,12 +115,68 @@ def is_readline_available():
...
+
+
+if PYTHON_VERSION_INFO < (3, 5, 0):
+    from pathlib import Path
+
+    class DirEntry:
```
scopatz: Did you copy this directly from somewhere? If so it needs a comment with the original copyright.

scopatz:
Thanks @funkyfuture some hopefully relatively minor changes...

xonsh/platform.py
funkyfuture: nope

xonsh/tools.py
funkyfuture: interestingly the old implementation is more than 1 ns faster than the new with Python 3.5, while the new performs a fraction of a ns better on 3.4.

funkyfuture:
i uodated the pr.

xonsh/tools.py
```diff
@@ -285,6 +285,12 @@ class redirect_stderr(_RedirectStream):
     _stream = "stderr"
 
 
+def executables_in(path):
```
scopatz: Would you mind adding a test for this function please? It could be pretty simple.

scopatz:
This looks great to me! Please just add a test and an entry to the CHANGELOG about this change.

scopatz:
Hey @funkyfuture - sorry to bother again, but you'll want to note the timing results of https://github.com/scopatz/xonsh/pull/991#issuecomment-220877422

funkyfuture:
ai'ght, i updated the pr with tests.

i also did @adqm's measuremts, however came to a result with ratios ~1.1 and  ~1.18.

adqm:
Are those results on Python 3.5 or 3.4?  There was a drastic difference between the timings on those two versions.  I think the conclusion @scopatz and I came to in #991 was that 3.4 and 3.5 should handle this differently (with 3.4 using what is currently on master, and 3.5 using this version), rather than trying to use the compatibility layer.  Of course, we could make these things have the same interface, but I think we want a different underlying implementation for each version.

funkyfuture:
in my case the results were very much the same with both implementations.

funkyfuture:
argh, i checked. there's a 3.5 interpreter in the virtual environment that is supposed to use 3.4. ?:-/

tests/test_tools.py
```diff
@@ -441,5 +490,27 @@ def test_partial_string():
...
+            path = os.path.join(test_path, str(i))
+            if _type == 'file':
+                open(path, 'w').close()
+            elif _type == 'directory':
+                os.mkdir(path)
+            if executable:
+                os.chmod(path, stat.S_IXUSR | stat.S_IRUSR | stat.S_IWUSR)
+
+        result = set(executables_in(test_path))
+        assert_equal(expected, result)
```
scopatz: Thanks for adding this!

scopatz:
Thanks a ton for putting this in @funkyfuture!  I think it will be easier to modify the `executables_in()` function for performance than to modify its usage in place, which is why I went ahead and merged this.

Issue #961, opened 2016-05-20
Recent xonsh forcing $SHELL_TYPE = best?


adqm:
Just as of pulling a couple of minutes ago, xonsh seems to be forcing `$SHELL_TYPE='best'` (which I definitely do not want because I disagree about the definition of best), despite my having `$SHELL_TYPE='readline'` in my `.xonshrc`.

Can confirm my `.xonshrc` is being run.  But my `$SHELL_TYPE` setting is not being honored.

scopatz:
I noticed something like this too, with 'random' always selecting ptk. Definitely a bug.

PR #979, opened 2016-05-22
Respect decisions made about SHELL_TYPE in rc files

adqm:
I believe this should fix the issue described in #961 where `$SHELL_TYPE` was always being set to `"best"` if the `--shell-type` flag was not used, _after_ the rc files were executed (thus overriding values set there).


scopatz:
Looks right to me, thanks!

PR #982, opened 2016-05-22
Slightly more substantial tutorial entry for @$

adqm:
As requested in #926, this adds a slightly more substantial Tutorial entry about the new ``@$()`` operator.  I'm happy to modify if this doesn't satisfy.


scopatz:
Thanks @adqm! I really  appreciate it!

PR #984, opened 2016-05-22
Make .bat files 'text' in .gitattributes

mwiebe:
The file scripts/xonsh.bat was showing up as changed, and making the repo basically unusable. Making .bat files explicitly text in the .gitattributes file fixes it for me.

The commands at the bottom of http://www.marten-online.com/source-versioning/git-dealing-with-line-endings-solution-2.html were necessary to reflect the adjusted .gitattributes file in the working copy (this will throw away any changes in your working copy, don't just run these commands arbitrarily):

```
$ git rm --cached -r .
$ git reset --hard
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,6 @@
 # Set the default behavior, in case people don't have core.autocrlf set.
 * text=auto
 
+*.bat text
+
 tests/histories/*.json text eol=lf


mwiebe:
Hmm, I guess this was an issue of of an older repo, and my git update failed, then I confused myself. After deleting the xonsh repo (A slight adventure - I ran 'rm -rf xonsh' from the xonsh shell. This deleted it as desired, but left the directory there in a state where I had no permissions to access it. Resorted to deleting it from the Windows GUI to fix things), and then cloning from github again, it seems fine now. Git's EOL handling is unfortunate though, in many projects I'm finding the auto mode does terrible things and it's better to explicitly list all the file types in .gitattributes.

melund:
Ah.. So git's auto detection of text files is to blame for the issue. More people have been complaining about this. But it didn't happen to everyone. Original this was even more of a problem since it was committed with `crlf` endings. I think this should just go in

melund:
OK. So we will wait. Let me know if you still think this is necessary...

mwiebe:
I believe the main context this shows up as a real problem is when some contributors use text editors which produce mixed line endings under some circumstances. The error I was seeing here wasn't that, though.

If the autocrlf is working for everyone, I don't see a reason to add explicit extensions in the .gitattributes, but doing so is a reliable way to avoid such problems.

scopatz:
If it makes you feel better @mwiebe, this issue borked my whole repo once too, just checking out a PR locally. It was a nightmare.  Things seem to be more or less working now, and I am in favor making it work better - though I am not 100% sure what that is.

Issue #980, opened 2016-05-22
Drop ptk <1.0 support


scopatz:
It is no longer officially supported and it can help us remove a lot of special cased code.

GuillaumeLeclerc:
I could revise #861 and only keep the code I added ?

scopatz:
It seems like @gforsyth has this one :)

PR #986, opened 2016-05-22
remove prompt_toolkit < v1.0.0 support

gforsyth:
As per #980 this removes all of the cased code for dealing with various
versions of prompt_toolkit in favor of supporting only v1.0.0+


scopatz:
Looks great to me! Thanks @gforsyth!

Issue #983, opened 2016-05-22
On Windows, would be nice for file autocompletions to use raw strings


mwiebe:
Here's how it looks presently when auto-completing quoted strings as file paths:

![image](https://cloud.githubusercontent.com/assets/399551/15451909/6ddfcbbe-1f8f-11e6-99b5-af3b8d8b8b96.png)

If I start with `r`, it inserts double `\\` in the autocomplete, though it's showing the right thing in the autocomplete list and selecting something from that list works well:

![image](https://cloud.githubusercontent.com/assets/399551/15451914/ab1e4dca-1f8f-11e6-9a4f-5184ca3fa3f8.png)

adqm:
Confirmed.  Thanks for reporting, @mwiebe!  I think I'm the one responsible for this bug, but what you describe was certainly the intended behavior (only showing a single backslash when quoting with a raw string).

I am going to have to call it a night soon, but if no one beats me to it, I should be able to take a look at this tomorrow.

Interestingly, it does seem to be Windows-specific (just tested on GNU/Linux with a filename with backslashes in it, and that is working as expected).

adqm:
@mwiebe: Actually, now that I am digging back through this, there is a good reason for the double-backslash on the end of the raw string: specifically, because of how Python parses (even raw) strings, if you want the last character to be a backslash, you need some other bogus means of escaping.  I've always hated it and thought it to be a "bug," but this was actually a [design decision](https://docs.python.org/3/faq/design.html#why-can-t-raw-strings-r-strings-end-with-a-backslash) in Python.

That said, I've just put in #987 which makes progress toward this by forcing paths with backslashes _in the middle_ to be completed as raw strings.  Progress, maybe.

The way I see it, for a real fix, we have a couple of options:

1. tack on a trailing "space" character (so that the trailing backslash can actually show up as a single backslash), or
1. modify raw strings in xonsh to accept a single backslash as the last character.

2\. would mean diverging from Python a bit, so maybe that's undesirable (though I think it would be an improvement!).
1\. might look a little strange with the trailing space inside the quotes (e.g., `r"some\path\ "`), but that's still actually preferable to the trailing `\\` IMHO.  Also, it might take some work to make this work with the completer, though.

@scopatz, any thoughts on the string replacement?

@mwiebe and other windows people, thoughts on trailing `"\ "` vs trailing `"\\"`?

scopatz:
[...]

PR #987, opened 2016-05-22
(slightly) better string quoting for windows completion

adqm:
This does not completely fix the issue described in #983 (see continued discussion there), but it makes progress by forcing paths with backslashes in the middle of them to be completed using raw strings.


scopatz:
This looks good to me, thanks!

PR #989, opened 2016-05-22
Fix bash foreign_shell_data on Windows

melund:
@scopatz .. I realized that bash foreign_shell_data always failed for me. I experimented a bit and this PR is what solved it for me.

I am not completely sure what the change to `DEFAULT_BASH_FUNCSCMD` does. But it was necessary to get the subprocess calling bash to succeed. I have no idea if messes things up on other platforms.

The second change in `parse_func()` was necessary to get `json.loads()` to work.

### DIFF
--- a/xonsh/foreign_shells.py
+++ b/xonsh/foreign_shells.py
@@ -12,6 +12,7 @@ from tempfile import NamedTemporaryFile
 from collections import MutableMapping, Mapping, Sequence
 
 from xonsh.tools import to_bool, ensure_string
+from xonsh.platform import ON_WINDOWS
 
 
 COMMAND = """
--- a/xonsh/foreign_shells.py
+++ b/xonsh/foreign_shells.py
@@ -30,7 +30,7 @@ echo __XONSH_FUNCS_END__
 {seterrpostcmd}
 """.strip()
 
-DEFAULT_BASH_FUNCSCMD = """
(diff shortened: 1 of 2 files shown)


gforsyth:
Hey @melund, doesn't seem to have messed anything up on my end, so it at least works on Arch

xonsh/foreign_shells.py
```diff
@@ -316,6 +317,8 @@ def parse_funcs(s, shell, sourcer=None):
     if m is None:
         return {}
     g1 = m.group(1)
+    if ON_WINDOWS:
+        g1 = g1.replace(os.sep,os.altsep)
```
scopatz: PEP8 space after comma

scopatz:
Thanks a gazillion @melund! I think the first change is certainly cleaner, though I am struggling to understand why the second change is needed at all. But if it works, great!  I think a changelog entry and the minor pep8 issue is all that is needed here.

melund:
I don't get why this was necessary. Especially what improved with the first fix and why it was not a problem on Linux. Anyway, I have updated the PEP8 thing. 

@gforsyth I hope you have ensured that it also loads your bash functions. It is a part of the code that would fail silently.

gforsyth:
Hey @melund, no worries.  I can still `source-bash` without any issues.
Looks like this now has conflicts with master

melund:
I rebased

PR #991, opened 2016-05-23
fix for command cache

adqm:
#902 fixed an issue with `xonsh.tools.CommandsCache`, but it also introduced a bug which prevented the cache from noticing most executables in the path.  This patch fixes that issue.

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -1077,7 +1077,9 @@ class CommandsCache(Set):
             this_one = set()
             for i in os.listdir(path):
                 name  = os.path.join(path, i)
-                if os.path.exists(name) and os.access(name, os.X_OK):
+                if (os.path.exists(name) and
+                        os.access(name, os.X_OK) and
+                        (not os.path.isdir(name))):
                     this_one.add(i)
             allcmds |= this_one
         allcmds |= set(builtins.aliases)
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -1074,8 +1074,12 @@ class CommandsCache(Set):
             return self._cmds_cache
         allcmds = set()
         for path in paths:
-            allcmds |= set(x for x in os.listdir(path)
-                           if os.path.isfile(x) and os.access(x, os.X_OK))
-            allcmds |= set(builtins.aliases)
+            this_one = set()
+            for i in os.listdir(path):
+                name  = os.path.join(path, i)
+                if os.path.exists(name) and os.access(name, os.X_OK):
+                    this_one.add(i)
+            allcmds |= this_one
+        allcmds |= set(builtins.aliases)
         self._cmds_cache = frozenset(allcmds)
         return self._cmds_cache


scopatz:
Hi @adqm! Thanks for putting this in.  This seems like it is basically the same as what is in #975 [here](https://github.com/scopatz/xonsh/pull/975/files#diff-ece06b292b3c59cc7bfc962fe8a6b856R1082).  Can you confirm? Maybe this should remain open until the other one is merged in or closed.

adqm:
Functionally similar, yes, but the solution in #975 seems to be very slow.  And I can't say I'm a fan of the object-oriented style there (in part because I imagine it contributes seriously to the speed).

Off to run some timing tests now.  Please hold off on merging either until I have something conclusive.

adqm:
Some timing results:

Test 1: 100 times, invoke `xonsh.tools.suggest_commands("pytho", __xonsh_env__, aliases)`
Tests suggestion of executables
Average time this branch: 0.1291 seconds
Average time #975: 0.2863 seconds
Ratio: 2.217

Test 2: 100 times, invoke `xonsh.tools.suggest_commands("gemtail", __xonsh_env__, aliases)`
Tests suggestion of aliases
Average time this branch: 0.1299 seconds
Average time #975: 0.2867 seconds
Ratio: 2.207

Test 3: 100 times, invoke `xonsh.completer.Completer().complete("python", "python", 0, 5, __xonsh_env__)`
Tests completion of names
Average time this branch: 0.0488 seconds
Average time #975: 0.1428 seconds
Ratio: 2.926

scopatz:
I think you win this round, Mr. @adqm :)

scopatz:
Thanks!

scopatz:
Though out of curiosity, do you use python v3.4?

adqm:
Ah, yes I do.  Would this explain a big difference in speed between the two?

adqm:
I can try to run on Python 3.5 and see if things are drastically different.

scopatz:
I think it could.  The point of #975 is that scandir is much faster than listdir(), and #975 implements a compatibility layer for scandir for v3.4

adqm:
Okay, let me compile 3.5 and re-run these tests.

scopatz:
That would be great! Thanks!

adqm:
Repeated tests on Python 3.5:

Test 1: 100 times, invoke `xonsh.tools.suggest_commands("pytho", __xonsh_env__, aliases)`
Tests suggestion of executables
Average time this branch: 0.097 seconds
Average time #975: 0.124 seconds
Ratio: 1.278

Test 2: 100 times, invoke `xonsh.tools.suggest_commands("gemtail", __xonsh_env__, aliases)`
Tests suggestion of aliases (I have an alias called `getmail` and one called `email`)
Average time this branch: 0.160 seconds
Average time #975: 0.187 seconds
Ratio: 1.168

Test 3: 100 times, invoke `xonsh.completer.Completer().complete("python", "python", 0, 5, __xonsh_env__)`
Tests completion of names
Average time this branch: 0.062 seconds
Average time #975: 0.024 seconds
Ratio: 0.387


It seems that on py35, the completer did get faster by a substantial margin (though I can't perceive a difference since the real-time difference is something like 4/100 seconds for this test).  If we really want the scandir stuff, my vote would be not to use it for py34, but rather to fall back to this method.

scopatz:
Yeah, I agree, I think that this is  a place where doing the less-good-option on either system makes it so much worse, that we should actually specialize the code based on Python version.

funkyfuture:
@adqm can you please elaborate what the bug was, so i can follow it in tests. 

i can't see more than an extra existence check that is implied by `is_file` and an extra set.

adqm:
No problem @funkyfuture.  I think the issue is resolved in #975, so maybe no need to worry.

The main difference is constructing the full filename of each file, and then running the checks (existence and execution) on the full name.  The code before was checking just the basename of each file (relative to cwd, I imagine), and so almost all files were failing that check.

Issue #758, opened 2016-03-28
Clean up tests (again)


melund:
I must admit that I have kind of ignored running the xonsh test suite for a while. Trying it again now, made me remember why. 

Firstly, the test script doesn't work on Windows. That may be less of a problem, but we @BYK and I ought to fix it at some point. 

The partial string test https://github.com/scopatz/xonsh/blob/master/tests/test_tools.py#L304 generate too many test cases (many thousands) . I didn't think it was a big deal when @adqm did the PR, but it scares new contributors away. I don't think it is necessary to test all combinations. 

Lastly, I get some errors when running tests which I can't fix. Maybe @scopatz can figure out what is going on. 

```
C:\Users\mel\Documents\xonsh>nosetests -q
C:\Users\mel\Anaconda3\lib\site-packages\win_unicode_console\__init__.py:27: RuntimeWarning: sys.stdin.encoding == 'utf-8', whereas sys.stdout.encoding == None, readline hook consumer may assume they are the same
  readline_hook.enable(use_pyreadline=use_pyreadline)
WARNING: Token 'AT' multiply defined
Generating LALR tables
WARNING: 92 shift/reduce conflicts
WARNING: 9 reduce/reduce conflicts
WARNING: reduce/reduce conflict in state 110 resolved using rule (test_or_star_expr -> test)
WARNING: rejected rule (testlist -> test) in state 110
WARNING: reduce/reduce conflict in state 185 resolved using rule (string_tok -> STRING)
WARNING: rejected rule (subproc_arg_part -> STRING) in state 185
WARNING: reduce/reduce conflict in state 273 resolved using rule (test_or_star_expr -> test)
WARNING: rejected rule (testlist -> test) in state 273
======================================================================
ERROR: Failure: JSONDecodeError (Expecting ',' delimiter: line 1 column 3708 (char 3707))
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\mel\Anaconda3\lib\site-packages\nose\failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "C:\Users\mel\Anaconda3\lib\site-packages\nose\loader.py", line 252, in generate
    for test in g():
  File "C:\Users\mel\Documents\xonsh\tests\test_replay.py", line 44, in test_echo
    with a_replay(f) as hist:
  File "C:\Users\mel\Anaconda3\lib\contextlib.py", line 59, in __enter__
    return next(self.gen)
[...]

PR #992, opened 2016-05-23
Add explicit extension ->text/binary mapping in .gitattributes

mwiebe:
Follow-on from discussion in https://github.com/scopatz/xonsh/pull/984, recreating since I can't reopen it.

This makes all the existing extensions in the repo explicitly text or binary. I changed the existing *.json rule that was forcing LF for text to just being text, is there a reason it should have unix line endings? Alternatively, could keep that, and force .bat files to have CRLF.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,29 @@
-tests/histories/*.json text eol=lf
-
 # Set the default behavior, in case people don't have core.autocrlf set.
+# This should be first, because the git documentation says "When more
+# than one pattern matches the path, a later line overrides an earlier line."
 * text=auto
+
+# Text files
+*.bat text
+*.css_t text
+*.in text
+*.json text
+*.py text
+*.rst text
+*.sh text
+*.txt text
+*.xsh text
+*.yaml text
+*.yml text
+CONTRIBUTING text
+license text
+LICENSE text
+Makefile text
+README text
+
+# Binary files
+*.ico binary
+*.gif binary
+*.gz binary
+*.png binary
+*.webm binary
--- a/.gitattributes
+++ b/.gitattributes
@@ -21,6 +21,9 @@ LICENSE text
 Makefile text
 README text
 
+# Files in the lazyjson format require LF line endings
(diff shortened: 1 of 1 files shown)


scopatz:
> I changed the existing *.json rule that was forcing LF for text to just being text, is there a reason it should have unix line endings? 

I believe that this was a testing issue on Windows with the LazyJson / History reading stuff.  Some of the index values were getting messed up if newlines were two characters instead of one.  Did you trying running the tests?

scopatz:
See #758 and #760

mwiebe:
I did not run the tests, and I do get some JSON errors (and a bunch of others) when I run the tests. There seem to also be problems handling temp files, and something about wanting a 'detype' attribute on 'dict' objects.

mwiebe:
I see the tests pass on the PR though, would be nice to get appveyor in the mix running the tests too.

mwiebe:
I see, the data offsets embedded in the lazyjson format requires the data have LF line endings.

mwiebe:
I've put the specific JSON line back in, with a comment. This fixes the JSON errors, so I only get 8 test failures.

mwiebe:
Have you thought of adjusting the lazyjson format so it's compatible with either line-ending pattern? I guess it complicates it somewhat to have to store both # of EOLs and non-EOLs separately where currently there's just a number of characters, so maybe not worth it.

scopatz:
> This fixes the JSON errors, so I only get 8 test failures.

Ahhhhhh Ok, would you mind opening an issue for those.  There shouldn't be any.

> Have you thought of adjusting the lazyjson format so it's compatible with either line-ending pattern? I guess it complicates it somewhat to have to store both # of EOLs and non-EOLs separately where currently there's just a number of characters, so maybe not worth it.

My original idea was that it just wouldn't be worth it, lazyjson being a json subset anyway. Maybe it wouldn't be so hard to implement that though...

scopatz:
Thanks @mwiebe! Hopefully this doesn't break anything else :)

PR #994, opened 2016-05-23
Fix tests on windows.

melund:


scopatz:
Thanks again @melund!

PR #996, opened 2016-05-23
Small alias updates

melund:
Adds 'clear' as an alias to 'cls' on windows. Even if 'clear' is available from MSYS it doesn't have any effect in cmd.exe or ConEmu.

Also sneaks in a verbose options '--all' to the which alias.


adqm:

PR #997, opened 2016-05-23
Bash completions fixes

melund:
This fixes bash completions so path-completions are no longer added to the suggestions. It also sorts the result so the resembles the suggestions in bash

See: https://github.com/scopatz/xonsh/issues/981#issuecomment-220867257

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,7 +24,6 @@ Current Developments
 **Fixed:**
 
 * Show sorted bash completions suggestions.
-* When doing bash completions path completions are no longer added to the suggestions
 * Fix bash completions (e.g git etc.) on windows when completions files have
   spaces in their path names
 * Fixed a bug preventing ``source-bash`` from working on Windows
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,8 @@ Current Developments
 
 **Fixed:**
 
+* Show sorted bash completions suggestions.
+* When doing bash completions path completions are no longer added to the suggestions
 * Fix bash completions (e.g git etc.) on windows when completions files have
   spaces in their path names
 * Fixed a bug preventing ``source-bash`` from working on Windows
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -193,7 +193,7 @@ class Completer(object):
         elif cmd in self.bash_complete_funcs:
             # bash completions
             rtn = self.bash_complete(prefix, line, begidx, endidx)
-            #rtn |= self.path_complete(prefix, path_str_start, path_str_end)
+            rtn |= self.path_complete(prefix, path_str_start, path_str_end)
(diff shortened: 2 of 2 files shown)


adqm:
There was a reason why path completions were added originally: specifically, so that filepaths would be quoted for xonsh rather than for bash if necessary.  That said, there is a real right way to do that, I think: once we get the suggestions back from xonsh, we should probably feed all the ones representing path elements through the `_quote_paths` function.

melund:
> There was a reason why path completions were added originally: specifically, so that filepaths would be quoted for xonsh rather than for bash if necessary

Good catch. Didn't know there was a reason for it. How can we check if it is a path we get back? Does bash ever give us stuff that is already quoted?

adqm:
Well, it shouldn't give back things that are quoted, but it might give things back that are _escaped_ for bash (sorry for misspeaking), i.e., with backslashes before spaces.  I can't remember how exactly we deal with that currently, but I think including the path completions was intended as a workaround for that, IIRC.

I mean, I guess in theory we could just check `os.path.exists` on the completions and use that as a first guess for whether to quote them or not.

But what I really want to do (already started work on it) is to replace bash completion entirely with completions made for xonsh specifically, by porting a lot of the bash_completions code to Python.  Even though it will be a pain to port things over, I think this may be the only real way to make sure we get what we want from the completers (plus I expect it should dramatically improve startup times since we won't have to invoke bash (plus it should be a fun project)).

melund:
That  sounds neat. Bash completions is a little slow on windows. Should I just add the path completions back in, but leave the `sorted(rtn)`?

melund:
I took a look at the `_filter_repeats()` function https://github.com/scopatz/xonsh/blob/master/xonsh/completer.py#L264

I can't see how that should remove the bash versions of any paths. To me it seems like it just removes the shortest versions of the path variables. That would not  necessarily be the bash version, especially if there are a lot of spaces in the word that are all escaped.

scopatz:
Just to interject here, while I am totally in favor of more Python based completions - and is long overdue! - I don't think we should loose the Bash completions entirely. For the foreeable future, I think arbitrary 3rd parties will be writing completion code for Bash *and not* xonsh.  It would be a shame for xonsh to loose the ability to complete against such programs.  That said, if there are Python- and Bash-based completers available for a given command, the Python one should be run and the Bash one should not. Does this make sense?  In some sense, what I am saying is that I would like to keep our interop level high.

adqm:
Yes, indeed.  My current plan was to introduce the new Python-based completers, one of which would be the generic bash completer (so that we don't lose that functionality).

scopatz:
Ahh sounds great!

melund:
@adqm... I made a small fix that may be usable until the new python completer becomes ready. 

Now it only adds entries from the xonsh paths completions if there is a canonical identical entry in the set of bash completions.

adqm:
Yep, looks good to me!

PR #998, opened 2016-05-23
Defaults for bash completions file on Windows

melund:
Add defaults for $BASH_COMPLETIONS on windows. The default values are where GitForWindows stores the completions files in a standard installation. 

This should hopefully ensure that more people get the full xonsh experience by default.


scopatz:
Great! Thanks @melund!

PR #1006, opened 2016-05-24
add completions to ${...}

gforsyth:
Might be a little hacky but it seemed like the best option.  I am open to suggestions.  @melund -- want to try this out and see if it works for you?

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -197,9 +197,9 @@ class Completer(object):
             return sorted(self._replace_canonical_rep(comps, pathcomp)), lprefix
         elif prefix.startswith('${') or prefix.startswith('@('):
             # python mode explicitly
-            rtn = self._python_mode_completions(prefix, ctx,
+            return self._python_mode_completions(prefix, ctx,
                                                  prefixlow,
-                                                 startswither)
+                                                 startswither), lprefix
         elif prefix.startswith('-'):
             comps = self._man_completer.option_complete(prefix, cmd)
             return sorted(comps), lprefix
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -525,6 +525,8 @@ class Completer(object):
         expr = subexpr_from_unbalanced(expr, '[', ']')
         expr = subexpr_from_unbalanced(expr, '{', '}')
         _ctx = None
+        if expr.endswith('()'):
+            return attrs
         try:
             val = builtins.__xonsh_execer__.eval(expr, ctx)
             _ctx = ctx
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -19,7 +19,7 @@ from xonsh.tools import (subexpr_from_unbalanced, get_sep,
 
 
 RE_DASHF = re.compile(r'-F\s+(\w+)')
-RE_ATTR = re.compile(r'([^\s\(\)]+(\..+)*)\.(\w*)$')
+RE_ATTR = re.compile(r'([^\s\(\)]+(\.[^\s\(\)]+)*)\.(\w*)$')
 RE_WIN_DRIVE = re.compile(r'^([a-zA-Z]):\\')
 
 
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -526,11 +526,11 @@ class Completer(object):
         expr = subexpr_from_unbalanced(expr, '{', '}')
         _ctx = None
         try:
-            val = eval(expr, ctx)
+            val = builtins.__xonsh_execer__.eval(expr, ctx)
             _ctx = ctx
         except:  # pylint:disable=bare-except
             try:
-                val = eval(expr, builtins.__dict__)
+                val = builtins.__xonsh_execer__.eval(expr, builtins.__dict__)
                 _ctx = builtins.__dict__
             except:  # pylint:disable=bare-except
                 return attrs  # anything could have gone wrong!
@@ -539,7 +539,7 @@ class Completer(object):
         opts = []
         for i in _opts:
             try:
-                eval('{0}.{1}'.format(expr, i), _ctx)
+                builtins.__xonsh_execer__.eval('{0}.{1}'.format(expr, i), _ctx)
             except:  # pylint:disable=bare-except
                 continue
             else:
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -197,7 +197,7 @@ class Completer(object):
             return sorted(self._replace_canonical_rep(comps, pathcomp)), lprefix
         elif prefix.startswith('${') or prefix.startswith('@('):
             # python mode explicitly
-            return self._python_mode_completions(prefix, ctx,
+            rtn = self._python_mode_completions(prefix, ctx,
                                                  prefixlow,
                                                  startswither)
         elif prefix.startswith('-'):
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -19,7 +19,7 @@ from xonsh.tools import (subexpr_from_unbalanced, get_sep,
 
 
 RE_DASHF = re.compile(r'-F\s+(\w+)')
-RE_ATTR = re.compile(r'(\S+(\..+)*)\.(\w*)$')
+RE_ATTR = re.compile(r'([^\s\(\)]+(\..+)*)\.(\w*)$')
 RE_WIN_DRIVE = re.compile(r'^([a-zA-Z]):\\')
 
 
@@ -525,8 +525,6 @@ class Completer(object):
         expr = subexpr_from_unbalanced(expr, '[', ']')
         expr = subexpr_from_unbalanced(expr, '{', '}')
         _ctx = None
-        if expr.endswith('()'):
-            return attrs
         try:
             val = builtins.__xonsh_execer__.eval(expr, ctx)
             _ctx = ctx
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,8 @@ Current Developments
 * Added comma literals to subproc mode.
 * ``@$(cmd)`` has been added as a subprocess-mode operator, which replaces in
   the subprocess command itself with the result of running ``cmd``.
+* The ``${...}`` shortcut for ``__xonsh_env__`` now returns appropriate 
+  completion options
 
 **Changed:**
 


melund:
@gforsyth Works great if `${...}` is first on the line. However, if it is not I get an exception:

```
snail@sea ~\Documents\xonsh gforsyth-add_env_completion $ d = ${...}.Exception in thread Thread-71:
    self.run()
    self.ctx)
ValueError: too many values to unpack (expected 2)
```

gforsyth:
Oops.  Let me see if I can sort that out

gforsyth:
Well, I fixed it.  There are a few unexpected consequences from this (not necessarily bad but I'd like some input).

The problem was that when `${...}.` begins a line, it is run via `rtn = self.cmd_complete(prefix)`, which then hits `self._python_mode_completions` further down the way.  

When `${...}` isn't the first token, it hits (or used to) `return self._python_mode_completions` which just gave back a set and the PTK completer wasn't happy about this.  Readline doesn't show any results in this case but doesn't crash.  

I removed the immediate return statement, which fixes the issue @melund noticed, so you can do:

2016-05-24-155048_699x159_escrotum

which makes sense to me.  
This also allows for completion options when using things like `$PATH`, a la: 

2016-05-24-155028_699x159_escrotum

What might get a little weird is showing completion options for `xonsh` envvars that are `bool`, although it's not _wrong_, it's maybe just a little odd.  

2016-05-24-155100_699x159_escrotum


I can't find anything that this change has broken, but if a few folks could try out some things they usually expect to work and confirm, that would be pretty great.

melund:
I will try to run with it for a while. The completions for booleans feels a bit weird, but on the other hand this is python. And it is great to have completions for more complex env-var types.

scopatz:
For what it is worth, this looks like the correct behaviour to me.

scopatz:
Also, I tried testing this on ![], $() and other subproc modes and it *does* execute, which is also correct :+1:

melund:
@scopatz ... I was also wondering if it would execute those statements. Good that it doesn't.

But what if it is a python function. 
```
snail@sea ~\Documents\xonsh gforsyth-add_env_completion $ def func():print('hallo')
snail@sea ~\Documents\xonsh gforsyth-add_env_completion $ func().<tab>
```

See the result below. I find it odd that it prints that many times from a single press on <tab>.

image

melund:
Ahh. The function returns `None`. So it is probably called for every entry in `$dir(None)`
```
snail@sea ~\Documents\xonsh gforsyth-add_env_completion $ dir(None)
['__bool__',
 '__class__',
 '__delattr__',
 '__dir__',
 '__doc__',
 '__eq__',
 '__format__',
 '__ge__',
 '__getattribute__',
 '__gt__',
 '__hash__',
 '__init__',
 '__le__',
 '__lt__',
 '__ne__',
 '__new__',
 '__reduce__',
 '__reduce_ex__',
 '__repr__',
 '__setattr__',
 '__sizeof__',
 '__str__',
 '__subclasshook__']
```

I guess the completer shouldn't call python functions at all.

gforsyth:
Hey @melund -- I did a runaround on your Python function case.  
The `expr` variable is a string, so I can't check `NoneType` without running it, which triggers the print statement (or whatever else is inside) but for now I'm restricting the `attr_complete` method to `expr` that don't end in `()` which seems to cover it (I also tried to break it using your example plus an `aliases['hallo'] = 'hallo()'` and that is also not triggering the `print` statement)

scopatz:
Seems to work for me!

melund:
I don't think that is enough. We should never execute unknown functions in the completer. Firstly because we don't know what side effects it can have and secondly because the may not return immediately. 

How is handled in IPython or systems like PtPython?  @scopatz @adqm what are your views on this.

I think we would need something like Jedi to really make this work.

melund:
And if we leave the responsibility to user not to press <tab>  we  should find a way not to execute it several times. One for each attribute it has.

gforsyth:
Hey @melund 
PtPython uses Jedi
IPython uses a similar system to ours.  In fact, the docstring for their `attr_matches` has this line in it: `WARNING: this can still invoke arbitrary C code, if an object
        with a __getattr__ hook is evaluated.`

gforsyth:
We can put a timeout on the return certainly, to obviate your second concern

xonsh/completer.py
```diff
@@ -270,6 +270,20 @@ def _filter_repeats(self, comps):
             reps[canon].append(comp)
         return {max(i, key=len) for i in reps.values()}
 
+    def _replace_canonical_rep(self, comps, replacements):
+        """ Replaces any entries in comps with its cononical duplicate
```
scopatz: cononical -> canonical

scopatz:
@melund, @gforsyth is correct.  I forked our system from IPython's originally.  So it is true that this will call `__getattr__()`, which due to properties and C extensions can be any arbitrary code. Probably because I have been using a system like this for ~8 years or so, I am pretty comfortable with it.  In practice, side effects come up extremely rarely and I don't that I that I have ever had anything catastrophic happen because of it. 

So while I agree that using Jedi or some other tool might help us prevent this (I don't actually know how Jedi works), it would almost certainly be an extension.  

Alternatively, maybe we could have an environment variable that flags whether gettattr() should be used or not.  I think that would be nice to have, with my preference for being default on.  I think that would be best done in another PR, though.

melund:
I don't think the IPython completer calls functions, with the exception of `___getattr__()` hook. But if we do decide to do it we shouldn't call it multiple times. 

But again I still think we should never do it.

melund:
OK. I could be wrong about IPython, it is just not a behaviour I ever noticed.

melund:
We should at least not call them multiple times

scopatz:
> We should at least not call them multiple times

I agree!

scopatz:
And here is the IPython completer: https://github.com/ipython/ipython/blob/master/IPython/core/completer.py  It has a lot of getattrs throughout

melund:
It is not the getattrs() that is the problem. I am OK with them. It is the fact that xonsh evaluates functions in the line that is completed. I just checked with IPython and it certainly doesn't have that behaviour. 

Here is an example:
```
In [2]: count = 0

In [3]: def func():
   ...:     global count
   ...:     count += 1
   ...:     print(count)
   ...:

In [4]: out = func().<TAB then Enter>
  File "<ipython-input-4-f320b5e820ac>", line 1
    out = func().
                 ^
SyntaxError: invalid syntax


In [5]: count
Out[5]: 0
```

In xonsh it looks like this:

```
snail@sea ~ $ count = 0
snail@sea ~ $ def func():
.............     global count
.............     count+=1
.............     print(count)
.............
snail@sea ~ $ out = func().<TAB then Enter>
2
3
4
5
6
7
8
9
10
11
12
13
14
15
16
17
18
19
20
21
22
23
24
snail@sea ~ $ out = func().
snail@sea ~ $ count
24
```

scopatz:
Huh, that is interesting.  I tried something like this early this morning and it did not behave that way.  That is probably where I was getting confused :).  All apologies!  

I agree - the completer should not call arbitrary functions  like this.  I have to run to a meeting, but will look at this later today.

gforsyth:
@melund -- thanks for pointing out that issue.  I think I may have fixed it if you want to take another look.

melund:
I think that IPython uses this regexpression 
https://github.com/ipython/ipython/blob/master/IPython/core/completer.py#L371
to find `expr.attr` while ensuring that it cant be `expr().atrr`

melund:
No problem. I will try later to night.

melund:
There is a difference in the regexes IPython and xonsh use to find object to perform attr completions on: 
IPython: `r"(\S+(\.\w+)*)\.(\w*)$"` 
https://github.com/ipython/ipython/blob/master/IPython/core/completer.py#L371

xonsh: `r'(\S+(\..+)*)\.(\w*)$'`
https://github.com/scopatz/xonsh/blob/master/xonsh/completer.py#L22

To me it seems that makes the difference. The first bracket in the IPython one `(\.\w+)` will not match something with brackets in.

gforsyth:
The regexes are different but neither does the job

```console
(xonshtest)gil@theo~$ RE_ATTR = re.compile(r"(\S+(\.\w+)*)\.(\w*)$")
(xonshtest)gil@theo~$ match = RE_ATTR.match('func().')
(xonshtest)gil@theo~$ match.group(1,3)
('func()', '')
```

melund:
Ahh you are right. I misread it

gforsyth:
I think the point of regexes is that they are misread.  Sigh... 
But I'm trying to figure out a small addition to remove the trailing parens match

melund:
Yes. It still does it if I add an argument to the function: 

```
snail@sea ~\Documents\xonsh master $ count = 0
snail@sea ~\Documents\xonsh master $ def func(arg):
....................................     global count
....................................     count+=1
....................................     print(count)
....................................
snail@sea ~\Documents\xonsh master $ func(1).1
2
3
4
5
6
7
8
9
10
11
12
13
14
15
16
17
18
19
20
21
22
23
24

```

gforsyth:
@melund, I'm almost nervous to say this, but I _might_ have a solution now.  
If history is any judge, you will break it within 5 minutes, but here's hoping.

melund:
I will test. My consern is that it could match and run class methods. E.g. `myobj.func().<tab>`

melund:
What about: `([^\s\(\)]+(\.[^\s\(\)]+)*)\.(\w*)$`

gforsyth:
I'm not sure it's strictly necessary, since the match area can also cover `.`s, so the first match group gets longer but still works as a completion feeder, e.g. for `numpy.` and `numpy.linalg`, the capture groups are, respectively, `numpy` and `numpy.linalg` which then feeds the completer.

melund:
Maybe am a little slow. It is getting late here.
```
snail@sea ~\Documents\xonsh master $ RE_ATTR = re.compile(r"(\S+(\..+)*)\.(\w*)$")
snail@sea ~\Documents\xonsh master $ match = RE_ATTR.match('myobj.func().')
snail@sea ~\Documents\xonsh master $ match.group(1,3)
('myobj.func()', '')
```

melund:
I think it is a problem. Try with the regex I suggested.
```
snail@sea ~\Documents\xonsh gforsyth-add_env_completion$ counter = 0
snail@sea ~\Documents\xonsh gforsyth-add_env_completion$ class C():
.........................................................     def func(self,arg):
.........................................................         global counter
.........................................................         counter+=1
.........................................................         print(counter)
.........................................................
snail@sea ~\Documents\xonsh gforsyth-add_env_completion $ myobj = C()
snail@sea ~\Documents\xonsh gforsyth-add_env_completion$ myobj.func(1).1
2
3
4
5
6
7
8
9
10
11
12
13
14
15
16
17
18
19
20
21
22
23
24
```

gforsyth:
Well @melund, it appears that you're faster than me even when slow.  Updated as per your suggestion.

melund:
Ha ha ... I doubt it :)

melund:
I just can't see how they avoid the problem in IPython.

melund:
So now I think it works well, and I can't find a way to break it. We better let @scopatz take a look also. He can merge it in.

scopatz:
Well, I couldn't break it!  Congrats to both!

PR #1011, opened 2016-05-25
xonsh will read history from locked files at startup to be included in up-arrow history

### LINKED ISSUE #935: Up arrow history search stops accepting new history if __xonsh_history__.buffersize = 0
I changed it yesterday morning and then my "Up arrow history" froze there. When I removed the ```__xonsh_history__.buffersize = 0``` line from my .xonshrc my "Up arrow history" started saving commands again.

As I've been working on my getHistory() command I've been able to monitor the history in XONSH_DATA_DIR and everything is fine there, it's saving everything exactly as I would expect it. 

So these must be two different mechanisms and when __xonsh_history_.flush() occurs after every command it's not additionally appending the command to the "Up arrow history", which I can only assume is only taking place when xonsh exits successfully.

JohnLunzer:
This should resolve #935.

### DIFF
--- a/xonsh/history.py
+++ b/xonsh/history.py
@@ -90,7 +90,7 @@ class HistoryGC(Thread):
             hsize, units = env.get('XONSH_HISTORY_SIZE')
         else:
             hsize, units = to_history_tuple(self.size)
-        files = self.unlocked_files()
+        files = self.files(only_unlocked=True)
         rmfiles_fn = self.gc_units_to_rmfiles.get(units)
         if rmfiles_fn is None:
             raise ValueError('Units type {0!r} not understood'.format(units))
@@ -101,7 +101,7 @@ class HistoryGC(Thread):
             except OSError:
                 pass
 
-    def unlocked_files(self):
+    def files(self, only_unlocked=False):
         """Find and return the history files that are unlocked.
 
         This is sorted by the last closed time. Returns a list of (timestamp,
@@ -115,14 +115,14 @@ class HistoryGC(Thread):
         for f in fs:
             try:
                 lj = lazyjson.LazyJSON(f, reopen=False)
-                if lj['locked']:
-                    continue
+                if only_unlocked:
+                    if lj['locked']:
+                        continue
                 # info: closing timestamp, number of commands, filename
                 files.append((lj['ts'][1], len(lj.sizes['cmds']) - 1, f))
                 lj.close()
             except (IOError, OSError, ValueError):
                 continue
-        files.sort()
         return files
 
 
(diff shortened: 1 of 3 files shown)


xonsh/history.py
```diff
@@ -101,7 +101,7 @@ def run(self):
             except OSError:
                 pass
 
-    def unlocked_files(self):
+    def files(self, only_unlocked=False):
         """Find and return the history files that are unlocked.
```
melund: Maybe update the doc string to match the new behaviour

xonsh/history.py
```diff
@@ -115,14 +115,14 @@ def unlocked_files(self):
         for f in fs:
             try:
                 lj = lazyjson.LazyJSON(f, reopen=False)
-                if lj['locked']:
-                    continue
+                if only_unlocked:
+                    if lj['locked']:
```
scopatz: also maybe glue these two conditions together with an and

xonsh/history.py
```diff
@@ -115,14 +115,14 @@ def unlocked_files(self):
...
+                        continue
                 # info: closing timestamp, number of commands, filename
                 files.append((lj['ts'][1], len(lj.sizes['cmds']) - 1, f))
                 lj.close()
             except (IOError, OSError, ValueError):
                 continue
-        files.sort()
```
scopatz: also, can we keep the sort please

scopatz:
thanks a ton for putting this in!

xonsh/history.py
JohnLunzer: Ahh, I forgot about this. When you take into account locked files the "closing timestamp" is sometimes `None`, then the list can no longer be sorted, get an exception. For garbage collect the order of the files does not matter, does it?

I meant to sort the individual commands by their timestamps in both history.py files for PTK and readline, but I forgot. I'll push this with the other recommended changes sometime soon.

If you have objections please let me know so I can take them into account.

scopatz: > For garbage collect the order of the files does not matter, does it?

Given how the gc is currently implemented, it looks like the sorting does matter.  I think it would be much more complex to reimplement the same behaviour without sorting.  To make this work, I would do something like:

```python
files.append((lj['ts'][1] or time.time(), len(lj.sizes['cmds']) - 1, f))
```

to set the sort time to the current time.

JohnLunzer: Thanks for this. I went this this. At this time I can't see the difference in sorting cmds by their timestamp in history.py for both ptk/readline. So I think for now this should be an improvement.

scopatz:
Thanks a million @JohnLunzer!

PR #1013, opened 2016-05-25
Speed up in attr completer

melund:
Speed up attr completions a tiny bit by only calling __xonsh_execer__.eval on the final candidates. Before it was called on all output from `dir(val)`


scopatz:
Even tiny speedups are important :)

scopatz:
Thanks a ton @melund!

Issue #1016, opened 2016-05-26
Error on exit with running bg jobs


melund:
If I exit xonsh while I have background jobs running I get the following error: 

```
snail@sea ~ $ $XONSH_SHOW_TRACEBACK = True
snail@sea ~ $ jupyter notebook a> null &
[1] running: jupyter notebook a> null & (23628)
snail@sea ~ $ exit
Traceback (most recent call last):
  File "c:\users\mel\documents\xonsh\xonsh\proc.py", line 335, in wrapped_simple_command
    r = f(args, i)
  File "c:\users\mel\documents\xonsh\xonsh\aliases.py", line 146, in exit
    kill_all_jobs()
  File "c:\users\mel\documents\xonsh\xonsh\jobs.py", line 208, in kill_all_jobs
    _kill(job)
  File "c:\users\mel\documents\xonsh\xonsh\jobs.py", line 21, in _kill
    check_output(['taskkill', '/F', '/T', '/PID', str(obj.pid)])
AttributeError: 'dict' object has no attribute 'pid'
```

This is on Windows and current master

adqm:
Confirmed.  The way we were using the `_kill` function was not consistent between when we were killing a single process (e.g., by ctrl-c) and when we were killing everything (when closing xonsh with jobs running in the background).  I believe this should fix; can you give it a look, @melund?

PR #1017, opened 2016-05-26
fix bug with killing jobs on windows

adqm:
I believe this should fix a bug with inconsistent use of `_kill` on Windows.  Symptom described in #1016.


melund:
Fantastic. That fixes it :)

PR #1020, opened 2016-05-26
Multine prompt colors

scopatz:
This change allows multiline prompts to accept colors.  More annoying to implement that you might think :)


adqm:

Issue #298, opened 2015-07-12
Write tab completion functionality in Python


scopatz:
Right now, you have to write them in Bash, and that is no good.

BYK:
Any pointers about where to start? I'm assuming we'd need a compatibility layer for existing bash completers.

scopatz:
Hi @BYK. Sure thing.  There is an existing Completer class in completer.py. This implements bash-completion, man page completion (though this needs more testing), and python object introspection completion.  

What we don't have right now is a way for users or ourselves to write and register external functions with this completion class.  Such functions should probably be keyed on the command name (ls, cd, etc).  Completion functions should have to follow a well-defined API that has arguments for the line, current position, current token, etc. and should return an iterable (though a set is recommended).

Does this make sense?  Let me know if you have any more questions!

BYK:
@scopatz it does make sense, yeah! :)

>Completion functions should have to follow a well-defined API that has arguments for the line, current position, current token, etc. and should return an iterable (though a set is recommended).

I completely agree with this. I also think we need a converter for bash-completion and potentially other strong and established completion syntax such as `zsh`. What do you think?

My main complaint right now is not being able to use any completions under Windows (except for file names).

scopatz:
> My main complaint right now is not being able to use any completions under Windows (except for file names).

Sure that makes sense. The more ways we can get completions the better!

gforsyth:
Closed via #1076

PR #1027, opened 2016-05-28
Framework for Python-based, Programmable Completers

adqm:
This PR is a work in progress, that will eventually include several Python-based completers for tab-completion.  It can also serve as a hub for discussion about this feature, the underlying infrastructure, etc.

This splits the functionality from the `Completer.complete` method into several separate functions, which are stored in an ordered dictionary.  Right now, this code loops over all of these completers until we find one that returns a non-empty set of results.

The current bash completer will be included as one of these so that we do not lose those extra completions, but the plan is to try to load bash completion functions on demand, rather than loading them all at startup time.

I will also include a means of adding new completion functions, and specifying easily the order in which they should be checked.  I would also like to include several common completers so that we don't have to rely on bash completion that often.

Right now, the completers are implemented as a single function that takes largely the same arguments as the `Completer.complete` function:

* `prefix`: the string to be matched
* `line`: a string representing the whole current line, for context
* `begidx`: the index at which `prefix` starts in `line`
* `endidx`: the index at which `prefix` ends in `line`
* `ctx`: the current Python environment

If there is a better interface, we should discuss.

There is also a question of whether the "extra" completers (beyond maybe bash, python, and path completions) should live in a `xontrib` rather than in xonsh proper.


xonsh/completer.py
```diff
@@ -165,80 +162,17 @@ def complete(self, prefix, line, begidx, endidx, ctx=None):
...
-                return sorted(self.module_complete(prefix)), lprefix
-            if cmd in self._cmds_cache:
-                # subproc mode; do path completions
-                comps = self.path_complete(prefix, path_str_start,
-                                           path_str_end, cdpath=True)
-                return sorted(comps), lprefix
+        empty_set = set()
+        for i in completers_enabled:
+            o = completers[i](prefix, line, begidx, endidx, ctx)
+            if isinstance(o, tuple):
```
scopatz: Is it important that this a tuple, rather than a more generic `Sequence`?

adqm: Sequence is fine; I can make that change.  For some reason, I had thought that a `set` would be considered a `Sequence`...

adqm:
Okay, I think this should now be a drop-in replacement for what we currently have, plus a few commands for adding new completers, disabling old completers, and listing those currently in use.

I may not have gotten the `CDPATH` stuff quite right.  Maybe someone can check on that.

I have not yet gone through and tried to speed up bash-based completions, nor have I added many custom completers.  Someday if someone have the time or the energy, it might be worth porting some bash completions to this system.  And it is certainly worth trying to speed up the bash completer (maybe by loading the bash completers on demand), but I'm starting to think that should live in another PR.

As always, happy for feedback.

gforsyth:
Hey @adqm -- this looks awesome and actually feels a bit snappier, although that may be in my head.  

A hearty +1 to loading completions on demand, but yeah, probably a separate PR. 

One problem so far when trying to tab-complete to a non-existent directory (pretty sure this is a `ptk` specific issue)

I have a directory `tmp/` that is empty, but if I try to complete something that doesn't exist I get an error

```console
gil@bad_cat ~ $ cd tmp/doesnException in thread Thread-120:
  File "/home/gil/miniconda3/lib/python3.5/threading.py", line 914, in _r
    self.run()
  File "/home/gil/miniconda3/lib/python3.5/threading.py", line 862, in rn
  File "/home/gil/miniconda3/lib/python3.5/site-packages/prompt_toolkit/n
    completions = list(buffer.completer.get_completions(document, comple)
  File "/home/gil/git/xonsh/xonsh/ptk/completer.py", line 36, in get_coms
    self.ctx)
ValueError: not enough values to unpack (expected 2, got 0)
```

gforsyth:
This also breaks `${...}` completion

adqm:
> This also breaks ${...} completion

Oh, right.  We're certainly missing [this bit](https://github.com/scopatz/xonsh/blob/master/xonsh/completer.py#L198).  I can work that in.

That said, there was a reason we switched to using Python's `eval` instead of the `xonsh` `eval` in the past.  I think that the code in the current master could launch subprocesses upon tab completion...see #364 and #540.  Does the current code not still have that problem?

> One problem so far when trying to tab-complete to a non-existent directory (pretty sure this is a ptk specific issue)

Hmmm...that is a mystery to me.  I only use readline, and you're right that I can't reproduce this there.  Will look into it now in ptk.

adqm:
> Hmmm...that is a mystery to me. I only use readline, and you're right that I can't reproduce this there. Will look into it now in ptk.

This one should be fixed, I think.  Silly mistake.

adqm:
> This also breaks ${...} completion

> Oh, right. We're certainly missing this bit. I can work that in.

> That said, there was a reason we switched to using Python's eval instead of the xonsh eval in the past. I think that the code in the current master could launch subprocesses upon tab completion...see #364 and #540. Does the current code not still have that problem?

Indeed, I can break xonsh by doing `python3.<tab>` on my machine (this launches Python).

I will include the bit of code to force Python mode when using `@()` and `${}`, but I think it's more better to use `eval` instead of xonsh's evaluation function (unless we could prevent xonsh from doing its subproc wrapping, but I don't think that's exactly easy given how things are put together now).

gforsyth:
Yeah, I just noticed that this morning.  I think forcing for `@()` and `${}` is a good solution.  I'll also dig into the PTK stuff

adqm:
>  I'll also dig into the PTK stuff

I think I killed it already.  Was returning the wrong thing in the case when all completers failed.

gforsyth:
> I think I killed it already. Was returning the wrong thing in the case when all completers failed.

Nice.  That fixed it.

adqm:
Alright @gforsyth, I think the most recent commit should fix the issue with `${}` and `@()` completion.  Do you mind taking another look?

adqm:
Though even if everything does work out, we probably want to wait to hear from @scopatz and/or @melund and/or @BYK as well before merging, since this is a pretty big change.

gforsyth:
Hey @adqm -- still no love for `${...}`, I'm afraid.  I'm not sure right now how to go about doing that, since using xonsh eval causes trouble.  But I don't have completion on `${...}` or `$PATH`, etc...

This won't work for the `${...}` case, but for things like `$PATH` we could check if they're defined in `__xonsh_env__` and then use xonsh eval on those? a set intersection shouldn't be too expensive, I don't think?  

I'm open to ideas.  Also, do we want python attributes to be sorted?

adqm:
> But I don't have completion on ${...} or $PATH, etc...

Wait, really?  Those are both working for me.  `$P<tab>` completes to all my env variables that start with `P`, and `${x<tab>` completes to all the variables in the Python context that start with `x`.

Are you not seeing those?  Or were you expecting something else?

If it's broken on your system, I can add in some debugging print statements.

> Also, do we want python attributes to be sorted?

Ack!  They are automatically sorted in readline, so I didn't notice that.  I guess we do want them to be sorted here.  But then my use of `Sequence` as a differentiation fails.

Maybe instead of a set of completions or a tuple of `(completions, prefixlength)`, we could have completers always return a dictionary `{'completions': ..., 'prefixlength': ...}`, where `prefixlength` is optional?  That's kind of ugly, but that would let completions be any iterable.

Or we could just _always_ sort the result before passing it on to the shell.  Maybe that works...

gforsyth:
Sorry, I was unclear.  The two examples you mention work well.  When we were using xonsh eval (with the definitely unacceptable side effects (which are live on master)), you could get attribute completion on xonsh objects, e.g.

2016-05-31-191204_708x152_escrotum
2016-05-31-191219_708x152_escrotum

gforsyth:
Sorting works now, though!

adqm:
> you could get attribute completion on xonsh objects

Oooooh.  I see.  Yes, that would be really nice.  I'll be AFK for awhile, but I'll mull on this in a background process.

It seems like, worst case, we could just special case for those things, check whether they're in the environment, and go from there without having to actually eval the entire expression (though we would still have to eval the bit in `${}`, which could involve launching processes...).  Hmmm....

adqm:
Alright; I think I've found a compromise.  It's not really pretty, but...

I've introduced a new flag to the execer, which causes it to error instead of wrapping with `![]` if it finds an invalid Python expression with a dot in it. This way, we can still get the completion without risking launching subprocesses.

Along these lines, I think we may also want to infer that `$()` gives a string, that `!()` and `![]` give `CompletedCommand` objects, etc, for the purposes of completion.  I'm happy to put that in here, or it could wait until later.

BYK:
I'm guessing this addresses #298 then, right? :)

melund:
> Along these lines, I think we may also want to infer that $() gives a string, that !() and ![] give CompletedCommand objects, etc, for the purposes of completion. I'm happy to put that in here, or it could wait until later.

That is a good idea. If it is straight forward for you, then just do it.

I tried it out and I do get an exception by pressing <tab> on an empty line:
```
snail@sea ~ $ <TAB>Exception in thread Thread-8:                                                                                                             x:0.3.2, py:3.5.1 b:adqm-completers
    self.run()
    self.ctx)
  File "c:\users\mel\documents\xonsh\xonsh\completer.py", line 55, in complete
    o = func(prefix, line, begidx, endidx, ctx)
  File "c:\users\mel\documents\xonsh\xonsh\completers\base.py", line 11, in complete_base
    paths = complete_path(prefix, line, start, end, ctx, False)
  File "c:\users\mel\documents\xonsh\xonsh\completers\path.py", line 168, in complete_path
    for s in iglobpath(prefix + '*', ignore_case=(not csc)):
  File "c:\users\mel\documents\xonsh\xonsh\tools.py", line 1205, in iglobpath
    return _iglobpath(s, ignore_case)[0]
  File "c:\users\mel\documents\xonsh\xonsh\tools.py", line 1194, in _iglobpath
    s = expand_case_matching(s)
NameError: name 'expand_case_matching' is not defined
```

melund:
@adqm It seems you just forgot to add `from xonsh.built_ins import expand_case_matching` to `tools.py` when you moved `_iglobpath()`

adqm:
@melund, thanks for the tip.  I just did a little bit of testing with `$CASE_SENSITIVE_COMPLETIONS=False` and I think things seem to working now with the most recent push.  Do you mind giving it another go?

adqm:
> I'm guessing this addresses #298 then, right? :)

@BYK, indeed; just under a year later!  haha

gforsyth:
@adqm -- woohoo!  Everything is working nicely on my end now.  I'll hold off on merging until we hear from @melund and/or @BYK about windows behavior and also let @scopatz have a look

xonsh/completers/base.py
```diff
@@ -0,0 +1,16 @@
+from collections import Sequence
+
+from xonsh.completers.path import complete_path
+from xonsh.completers.python import complete_python
+from xonsh.completers.commands import complete_command
+
```
gforsyth: PEP8 two blank lines after imports

xonsh/completers/tools.py
```diff
@@ -0,0 +1,17 @@
+import builtins
+
```
gforsyth: PEP8 two blank lines

adqm:
Great!  I'll also want to add some documentation, but glad to hear it all seems to be working!

melund:
@adqm 

Next error :)
``` 
snail@sea ~\Documents\xonsh adqm-completers $<TAB> Exception in thread Thread-10:     
    self.run()
    self.ctx)
  File "c:\users\mel\documents\xonsh\xonsh\completer.py", line 55, in complete
    o = func(prefix, line, begidx, endidx, ctx)
  File "c:\users\mel\documents\xonsh\xonsh\completers\base.py", line 11, in complete_base
    paths = complete_path(prefix, line, start, end, ctx, False)
  File "c:\users\mel\documents\xonsh\xonsh\completers\path.py", line 168, in complete_path
    for s in iglobpath(prefix + '*', ignore_case=(not csc)):
  File "c:\users\mel\documents\xonsh\xonsh\tools.py", line 1266, in iglobpath
    return _iglobpath(s, ignore_case)[0]
  File "c:\users\mel\documents\xonsh\xonsh\tools.py", line 1255, in _iglobpath
    s = expand_case_matching(s)
  File "c:\users\mel\documents\xonsh\xonsh\tools.py", line 1215, in expand_case_matching
    drive_part = WINDOWS_DRIVE_MATCHER.match(s) if ON_WINDOWS else None
NameError: name 'WINDOWS_DRIVE_MATCHER' is not defined
```

adqm:
Thanks!  I should give up on hoping that Windows will "just work"...  Okay, time to commence the Windows testing.

adqm:
Alright, I think (hope) that that was the last one.  Just pushed an update.  Mind taking another look, @melund?

melund:
It works really well. Do you want to add completions for  $() and  !()/![] ? Otherwise I will be happy to merge it in.

adqm:
> Do you want to add completions for $() and !()/![] ?

I think I'm going to wait on those, and put them in through a separate PR.  They aren't as straightforward as I had originally thought, and I think it's fine to separate the move to the framework from the implementation of a few new completers.

That said, let me put in some docs and a changelog entry.  It would also be good to get @scopatz's blessing before merging if he's not totally burnt out from pycon.

adqm:
Okay, updated the docs and also had to make a modification to the start-up procedure to make sure that `register-completer` would work as part of a `.xonshrc` file.  I think it's ready for a last look.

docs/tutorial.rst
```diff
@@ -1022,12 +1022,12 @@ Up, Down, Tab
...
-based on the variable names in the current builtins, globals, and locals,
-as well as xonsh languages keywords & operator, files & directories, and
-environment variable names. In subprocess-mode, you additionally complete
-on any file names on your ``$PATH``, alias keys, and full BASH completion
-for the commands themselves.
+Tab completion is present as well. By default, in Python-mode you are able to
+complete based on the variable names in the current builtins, globals, and
+locals, as well as xonsh languages keywords & operator, files & directories,
+and environment variable names. In subprocess-mode, you additionally complete
+on any file names on your ``$PATH``, alias keys, and full BASH completion for
```
melund: Isn't it only executables files on the path?

adqm: Yes, that's right.  Just updated.

melund:
Otherwise, everything works fine. I will second @gforsyth .. It feels snappier.

scopatz:
I take it that dev on this has moved to the other PR?

adqm:
> I take it that dev on this has moved to the other PR?

Yes; I hadn't really intended that, but since that's where all your comments are, I'll close this one and we can focus on the other.

PR #1029, opened 2016-05-28
Debug Subprocs, And/Or

### LINKED ISSUE #1024: && and || operators in subprocess mode
Hello,

Yet && is being interpreted as two backgrounding characters, and || as two contiguous pipes, which of course doesn't work.

It would be great if things like ```true && echo Owiii``` or ```false || echo Onooo``` would work!

Thanks in advance!

scopatz:
This refactors the `find_next_break()` functionality to make it more accessible.  This allows the context sensitive parser to use it as needed.  This should close #1024.


adqm:

Issue #1014, opened 2016-05-25
Test are failing on windows again


melund:
```
======================================================================
FAIL: test_tools.test_executables_in
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\mel\Anaconda3\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Users\mel\Documents\xonsh\tests\test_tools.py", line 512, in test_executables_in
    assert_equal(expected, result)
AssertionError: Items in the second set but not the first:
'120'
'26'
'158'
'84'
'30'
'126'
'55'
'115'
'69'
'82'
'3'
'44'
'90'
'155'
'161'
'35'
'134'
'163'
'138'
'157'
'106'
'6'
'149'
'110'
'13'
'53'
'62'
'50'
'81'

----------------------------------------------------------------------
Ran 132617 tests in 29.080s

FAILED (SKIP=8, failures=1)

[x] C:\Users\mel\Documents\xonsh>
```

scopatz:
I saw this error earlier today. I thought I fixed it.

scopatz:
Also - I wised Appveyor didn't hate us :(  Maybe we should email them.

scopatz:
Yeah, I thought I fixed this in #1010

melund:
Oh. This is not a way to write tests. Can we please come up with something simpler and more straight forward. It is breaking on Windows, and my mind is breaking just reading it. 

```
def test_executables_in():
    import pdb;pdb.set_trace()
    expected = set()
    with TemporaryDirectory() as test_path:
        for i in range(random.randint(100, 200)):
            _type = random.choice(('none', 'file', 'file', 'directory'))
            if _type == 'none':
                continue
            executable = random.choice((True, True, False))
            if _type == 'file' and executable:
                expected.add(str(i))
            path = os.path.join(test_path, str(i))
            if _type == 'file':
                open(path, 'w').close()
            elif _type == 'directory':
                os.mkdir(path)
            if executable:
                os.chmod(path, stat.S_IXUSR | stat.S_IRUSR | stat.S_IWUSR)

        result = set(executables_in(test_path))
        assert_equal(expected, result)
```

PEP20 clause 3: "Simple is better than complex". 

Also, I don't think tests should use `random` without a very good reason, and then always with a fixed seed.

scopatz:
[...]

PR #1030, opened 2016-05-28
Deterministic tests

scopatz:
This is an attempt to make the tests deterministic and address #1014.  @melund, if you get a chance would you mind seeing if this works on Windows? Thanks!

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -503,14 +503,14 @@ def test_executables_in():
                 expected.add(str(i))
             path = os.path.join(test_path, str(i))
             if _type == 'file':
-                open(path, 'w').close()
+                with open(path, 'w') as f:
+                    f.write(str(i))
             elif _type == 'directory':
                 os.mkdir(path)
             if executable:
                 os.chmod(path, stat.S_IXUSR | stat.S_IRUSR | stat.S_IWUSR)
-
         result = set(executables_in(test_path))
-        assert_equal(expected, result)
+    assert_equal(expected, result)
 
 
 if __name__ == '__main__':
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -7,6 +7,7 @@ import stat
 import nose
 from nose.tools import assert_equal, assert_true, assert_false
 
+from xonsh.platform import ON_WINDOWS
 from xonsh.lexer import Lexer
 from xonsh.tools import (
     subproc_toks, subexpr_from_unbalanced, is_int, always_true, always_false,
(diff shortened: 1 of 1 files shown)


melund:
I get this error. 
```
======================================================================
FAIL: test_tools.test_executables_in
  File "C:\Users\mel\Anaconda3\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\Users\mel\Documents\xonsh\tests\test_tools.py", line 513, in test_executables_in
    assert_equal(expected, result)
AssertionError: Items in the second set but not the first:
'29'
'14'
'53'
'26'
'50'
'5'
'62'
'2'
'38'
'17'
'41'

Ran 132617 tests in 34.972s

FAILED (SKIP=8, failures=1)
```

melund:
But hey. Now it is the same error every time :)

scopatz:
Progress!

scopatz:
How about now?

melund:
Same error. Did some research and I am thinking it may have something to do with the limited functionality of `os.chmod()` on windows. 

>Note Although Windows supports chmod(), you can only set the file’s read-only flag with it (via the stat.S_IWRITE and stat.S_IREAD constants or a corresponding integer value). All other bits are ignored.

https://docs.python.org/3/library/os.html#os.chmod

So on windows you just need to change the file extension to '.exe' to make it executable. That should be picked up by `os.access(filename,os.X_OK)` which is used by `execultables_in()`

scopatz:
Ahh, I see.  Thanks!  Hopefully this latest change will work then...

melund:
O

melund:
Still fails. Seems like I was partly wrong. Not about the extension part, but it seems that `os.access(filename, os.X_OK)` is always true. 

```
(Pdb) p list(executables_in(test_path))
['1.exe', '10.exe', '13.exe', '14', '17', '18.exe', '2', '21.exe', '22.exe', '25.exe', '26', '29', '30.exe', '33.exe', '34.exe', '37.exe', '38', '41', '42.exe', '45.exe', '46.exe', '49.exe', '5', '50', '53', '54.exe', '57.exe', '58.exe', '6.exe', '61.exe', '62', '9.exe']
(Pdb)
```

I guess that the `executables_in()` function actually doesn't work on windows. Maybe that is also why the `which.py` module relies on the extension and only check `os.access` when it is not running on windows: https://github.com/scopatz/xonsh/blob/master/xonsh/xoreutils/_which.py#L151

I am off to bed now. I hope you can manage a release if that is still the plan :)

melund:
The `executables_in()` is doubled by the `yield_executables()` in `envrionment.py` 

https://github.com/scopatz/xonsh/blob/d69c09f4708a9efe946ef780205ffb9ca87ee79b/xonsh/environ.py#L768

I guess `yield_executables()` is not optimized with the `scandir()` stuff, but it understands executables on windows. 

I guess they should really be merged and moved to the new `platforms.py` module.

melund:
I have a fix incoming for this.

Issue #514, opened 2015-11-19
Does it work under Windows/Cygwin?


tkossak:
Windows 7 x64
Cygwin x86
Python3 3.4.3

I installed xonsh with ```pip3 install xonsh``` and when I try to run it with ```xonsh``` I get the error:
```
Exception in thread Thread-2:
Traceback (most recent call last):
  File "/usr/lib/python3.4/threading.py", line 920, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.4/site-packages/xonsh/readline_shell.py", line 281, in run
    RL_LIB.history_set_pos(i)
  File "/usr/lib/python3.4/ctypes/__init__.py", line 364, in __getattr__
    func = self.__getitem__(name)
  File "/usr/lib/python3.4/ctypes/__init__.py", line 369, in __getitem__
    func = self._FuncPtr((name_or_ordinal, self))
AttributeError: function 'history_set_pos' not found (ďż˝[ďż˝ďż˝ďż˝%ďż˝ďż˝Pďż˝ďż˝ďż˝)
```

scopatz:
You may need to install gnureadline or use prompt-toolkit instead.

scopatz:
Thanks for reporting @tkossak

tkossak:
```pip3 install gnureadline``` gave me errors (gcc can't find lncurses) but I
managed to run xonsh with ```prompt-toolkit```. The shell starts with no errors, but
when I run any shell command (ie: ls, vim, echo, etc.) I get the error:

```
Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/xonsh/base_shell.py", line 126, in default
    self.execer.exec(code, mode='single', glbs=self.ctx)  # no locals
  File "/usr/lib/python3.4/site-packages/xonsh/execer.py", line 126, in exec
    return exec(code, glbs, locs)
  File "<xonsh-code>", line 1, in <module>
  File "/usr/lib/python3.4/site-packages/xonsh/built_ins.py", line 636, in subproc_uncaptured
    return run_subproc(cmds, captured=False)
  File "/usr/lib/python3.4/site-packages/xonsh/built_ins.py", line 606, in run_subproc
    wait_for_active_job()
  File "/usr/lib/python3.4/site-packages/xonsh/jobs.py", line 110, in wait_for_active_job
    _give_terminal_to(pgrp)
[...]

PR #1031, opened 2016-05-29
Cygwin: don't use pthread_sigmask

OllieTerrance:
Follow-up from #514.  The signal library doesn't provide `pthread_sigmask()` on Windows or Cygwin.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,8 @@ Current Developments
   in the prompt_toolkit shell would cause xonsh to crash.
 * Fixed broken behavior when using ``cd ..`` to move into a nonexistent
   directory.
+* Partial workaround for Cygwin where ``pthread_sigmask`` appears to be missing
+  from the ``signal`` module.
 
 **Security:** None
 
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -8,7 +8,7 @@ import builtins
 from subprocess import TimeoutExpired, check_output
 from collections import deque
 
-from xonsh.platform import ON_DARWIN, ON_WINDOWS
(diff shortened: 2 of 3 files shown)


adqm:
Thanks for putting this in, @OllieTerrance!

Does this still have the problem with xonsh backgrounding itself?  That would make sense with the signals not being blocked anymore, but is something we probably want to fix.

I won't be able to take a look at this until I am back near a Windows machine, which will be a couple of days.  Maybe someone else can take a look in the meantime.

OllieTerrance:
Indeed, in its current state xonsh will background when the current process finishes or is itself backgrounded, assuming xonsh is started from another shell providing job control -- `exec xonsh` from bash is sufficient to avoid it.  Of course ideally we need another way to block bash from intercepting.

scopatz:
It is doubly weird since the the cygwin docs claim that it is supported: http://pipeline.lbl.gov/code/3rd_party/licenses.win/cygwin-doc-1.4/cygwin-api.html#id4665466

adqm:
@OllieTerrance, sorry for the delay.  I'm just now setting up a Cygwin environment for myself so I can test.  It looks like this also needs a merge from master, and an entry in CHANGELOG.rst.

OllieTerrance:
@adqm: Rebased onto master, changelog note added.  :smile:

adqm:
Okay, great; thanks!  I think this is ready to merge in.  We'll still have to work through some of the other concerns discussed in #514 eventually (the killpg issue and trying to bring back the sigmask behavior), but I think this is a big step in the right direction!

I'll add a quick note to the Windows docs about using `exec bash` when I merge this in.

PR #1035, opened 2016-05-29
Fix Permission error when tab-completing

melund:
This is a quick fix for https://github.com/scopatz/xonsh/issues/1033 . Where `executables_in` fails if the directory is read-protected.  It doesn't fix the problem that `executables_in` return all files on windows. 

@asmeurer . Can you test?


asmeurer:
Fixes the problem for me.

scopatz:
thanks @melund

PR #1037, opened 2016-05-30
xonsh/history.py: apply os.path.expanduser to XONSH_DATA_DIR in history

arielf:
- Fixes https://github.com/scopatz/xonsh/issues/1000
    - Prior unrelated fix: remove 'print's from lambdas (fixes syntax error) was reverted since it didn't work

### DIFF
--- a/xonsh/history.py
+++ b/xonsh/history.py
@@ -110,7 +110,8 @@ class HistoryGC(Thread):
         """
         _ = self  # this could be a function but is intimate to this class
         # pylint: disable=no-member
-        xdd = os.path.abspath(builtins.__xonsh_env__.get('XONSH_DATA_DIR'))
+        xdd = os.path.expanduser(builtins.__xonsh_env__.get('XONSH_DATA_DIR'))
+        xdd = os.path.abspath(xdd)
         fs = [f for f in iglob(os.path.join(xdd, 'xonsh-*.json'))]
         files = []
         for f in fs:
@@ -256,6 +257,7 @@ class History(object):
         if filename is None:
             # pylint: disable=no-member
             data_dir = builtins.__xonsh_env__.get('XONSH_DATA_DIR')
+            data_dir = os.path.expanduser(data_dir)
             self.filename = os.path.join(data_dir, 'xonsh-{0}.json'.format(sid))
         else:
             self.filename = filename


xonsh/history.py
```diff
@@ -110,7 +110,8 @@ def files(self, only_unlocked=False):
         """
         _ = self  # this could be a function but is intimate to this class
         # pylint: disable=no-member
-        xdd = os.path.abspath(builtins.__xonsh_env__.get('XONSH_DATA_DIR'))
+        xdd = os.path.expanduser(builtins.__xonsh_env__.get('XONSH_DATA_DIR'))
+        xdd = os.path.abspath(xdd)
```
scopatz: Seems like this should use [`xonsh.built_ins.expand_path()`](http://xon.sh/api/built_ins.html#xonsh.built_ins.expand_path)

xonsh/history.py
```diff
@@ -256,6 +257,7 @@ def __init__(self, filename=None, sessionid=None, buffersize=100, gc=True,
         if filename is None:
             # pylint: disable=no-member
             data_dir = builtins.__xonsh_env__.get('XONSH_DATA_DIR')
+            data_dir = os.path.expanduser(data_dir)
```
scopatz: Same

xonsh/history.py
```diff
@@ -431,8 +433,8 @@ def _gc(ns, hist):
 
 _MAIN_ACTIONS = {
     'show': _show,
-    'id': lambda ns, hist: print(hist.sessionid),
-    'file': lambda ns, hist: print(hist.filename),
+    'id': lambda ns, hist: hist.sessionid,
+    'file': lambda ns, hist: hist.filename,
```
scopatz: Does this work?

scopatz:
Thanks for putting this in! Just a couple of small changes @arielf!

xonsh/history.py
arielf: Thanks. I'm not very familiar with the code and wasn't able to find the correct object to use the built_in on (`xonsh` isn't defined in this scope, `self` doesn't have a `built_ins` member). More hints appreciated.

xonsh/history.py
arielf: @scopatz
Oops, so sorry.  This indeed doesn't work.  `pylint` gave me a syntax error which led to this misguided change.   I just reverted it, sorry again.

xonsh/history.py
melund: Just add an import at the top of the file: `from xonsh import built_ins`

arielf: @melund thanks for the assist! Appreciated.
This is now fixed in the latest (additional) push.

arielf:
Travis succeeded with python 3.5 but failed with
` ImportError: cannot import name 'built_ins'` for the python 3.4 build.

Looks like the original patch (before switching from `os.path.expanduser` to `xonsh.built_ins.expandpath`) was a bit safer.

Please feel free to comment or accept the prior version (1st commit) of the PR (before the last commit)
Thanks!

adqm:
Probably a cyclic import causing that error.

It should also be possible (though it's maybe not pretty) to remove the import of the `xonsh.built_ins` module, and to reference that function as `builtins.__xonsh_expand_path`.  I think that should avoid the error.

arielf:
@adqm thank you! appreciate these hints.

Unfortunately, removing the import and using  `builtins.__xonsh_expand_path` just causes a different exception (in python3.5)
```
    ...
    data_dir = builtins.__xonsh_expand_path(builtins.__xonsh_env__.get('XONSH_DATA_DIR'))
AttributeError: module 'builtins' has no attribute '_History__xonsh_expand_path'
```
Not sure why the `_History` prefix gets added to what's actually called.
I feel just accepting the 1st commit (using `os.path`) would be the most portable, and the least of evils.

xonsh/history.py
```diff
@@ -110,7 +110,8 @@ def files(self, only_unlocked=False):
         """
         _ = self  # this could be a function but is intimate to this class
         # pylint: disable=no-member
-        xdd = os.path.abspath(builtins.__xonsh_env__.get('XONSH_DATA_DIR'))
+        xdd = built_ins.expand_path(
```
scopatz: To avoid the circular imports, can't you import the module here? Same below, too.

arielf: @scopatz I could but not sure how will this solve the (unrelated?) travis build failure with python 3.4.

Seems like the least of evils, simplest and most portable fix is to avoid all imports and apply https://github.com/scopatz/xonsh/commit/16d2096e2b7958f9f8aab3882c385899e1389f07
It has no issues in travis and fixes the original problem.

scopatz: Sure, that make sense.  Sorry about the red herring.  Please revert this PR back to that state, and I'll go ahead and we can review again and merge.

arielf:
@scopatz thanks!
Reverted last commit.  Travis tests all pass.  Diff vs scopatz/master is small and simple.

scopatz:
Thanks @arielf! Sorry about the run around, again!

arielf:
@scopatz no worries at all.  I'm always learning new things.
Thanks for working on this wonderful software!

scopatz:
No problemo :)  It is a team effort, so please keep contributing!  Most small PRs are much easier, I promise.

PR #1040, opened 2016-05-30
Verbose which

melund:
This fixes a bug on Windows where `which` did not include the current directory. It also gives better verbose output on windows. 

The bug was caused by us explicitly giving `which.whichgen` the `__xonsh_env__['PATH']`  That changes the behaviour of the module a bit.


adqm:
Looks good to me, as well.  Merging now.

Issue #1028, opened 2016-05-28
Warn about suspended and background jobs before exit


willsALMANJ:
Now that #867 and #962 have made job control work, I noticed that xonsh exits immediately when given the `exit` command. Other shells like bash and zsh will print a message about suspended or running jobs and refuse to exit in this case. I think this would be a nice feature to have in xonsh as well.

scopatz:
Thanks for letting us know!  This likely wouldn't be too hard to implement. Happy to point people in the right direction here.

scopatz:
I believe that only this function would need to change: https://github.com/scopatz/xonsh/blob/master/xonsh/jobs.py#L202

PR #1042, opened 2016-05-30
Check no stopped or background jobs exist before exiting interactive shell

willsALMANJ:
This addresses #1028. Please let me know what you think. I wasn't sure what was most readable regarding clean/kill/etc.

### DIFF
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -280,8 +280,8 @@ def clean_jobs():
                 print('-'*5, file=sys.stderr)
                 jobs([], stdout=sys.stderr)
                 print('-'*5, file=sys.stderr)
-                print(('Attempt to exit again immediately to kill all jobs '
-                       'and exit'), file=sys.stderr)
+                print('Type "exit" or press "ctrl-d" again to force quit.',
+                      file=sys.stderr)
                 jobs_clean = False
                 _last_exit_time = time.time()
     else:
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -261,20 +261,21 @@ def clean_jobs():
                 type_str = job_types.pop()
 
             if len(builtins.__xonsh_all_jobs__) > 1:
-                article = ''
+                num = 'are'
                 job_str = 'jobs'
             else:
-                article = 'a '
+                num = 'is a'
                 job_str = 'job'
 
-            print('xonsh: you have {art}{typ} {job}'.format(art=article,
-                                                            typ=type_str,
-                                                            job=job_str))
+            print()
+            print('xonsh: there {num} {typ} {job}'.format(num=num,
+                                                          typ=type_str,
+                                                          job=job_str))
             print('-'*5)
             jobs([])
             print('-'*5)
-            resp = input('Kill running jobs [y/n]? ')
-            if 'yes'.startswith(resp.lower()):
+            resp = input('Enter "exit" or "e" to kill all jobs and exit ')
+            if resp in ('e', 'exit'):
                 kill_all_jobs()
             else:
                 jobs_clean = False
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -251,14 +251,8 @@ def clean_jobs():
     jobs_clean = True
     if builtins.__xonsh_env__['XONSH_INTERACTIVE']:
         _clear_dead_jobs()
-        job_types = set()
-        for job in builtins.__xonsh_all_jobs__.values():
-            if job['bg']:
-                job_types.add('background')
-            elif job['status'] == 'stopped':
-                job_types.add('stopped')
 
-        if job_types:
+        if builtins.__xonsh_all_jobs__:
             global _last_exit_time
             if builtins.__xonsh_history__.buffer:
                 last_cmd_start = builtins.__xonsh_history__.buffer[-1]['ts'][0]
@@ -273,23 +267,13 @@ def clean_jobs():
                 # unfinished jobs in this case.
                 kill_all_jobs()
             else:
-                if 'background' in job_types and 'stopped' in job_types:
-                    type_str = 'background and stopped'
-                else:
-                    type_str = job_types.pop()
-
                 if len(builtins.__xonsh_all_jobs__) > 1:
-                    num = 'are'
-                    job_str = 'jobs'
+                    msg = 'there are unfinished jobs'
                 else:
-                    num = 'is a'
-                    job_str = 'job'
+                    msg = 'there is an unfinished job'
 
                 print()
-                print('xonsh: there {num} {typ} {job}'.format(num=num,
-                                                              typ=type_str,
-                                                              job=job_str),
-                      file=sys.stderr)
+                print('xonsh: {}'.format(msg), file=sys.stderr)
                 print('-'*5, file=sys.stderr)
                 jobs([], outfile=sys.stderr)
                 print('-'*5, file=sys.stderr)
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -255,7 +255,6 @@ def clean_jobs():
                 job_types.add('stopped')
 
         if job_types:
-            jobs_clean = False
             if 'background' in job_types and 'stopped' in job_types:
                 type_str = 'background and stopped'
             else:
@@ -271,6 +270,14 @@ def clean_jobs():
             print('xonsh: you have {art}{typ} {job}'.format(art=article,
                                                             typ=type_str,
                                                             job=job_str))
+            print('-'*5)
+            jobs([])
+            print('-'*5)
+            resp = input('Kill running jobs [y/n]? ')
+            if 'yes'.startswith(resp.lower()):
+                kill_all_jobs()
+            else:
+                jobs_clean = False
     else:
         kill_all_jobs()
 
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -11,6 +11,9 @@ from collections import deque
 from xonsh.platform import ON_DARWIN, ON_WINDOWS, ON_CYGWIN
 
 tasks = deque()
+# Track time stamp of last exit command, so that two consecutive attempts to
+# exit can kill all jobs and exit.
+_last_exit_time = None
 
 
 if ON_DARWIN:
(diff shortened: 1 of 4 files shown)


adqm:
Thanks for putting this in, @willsALMANJ!  I think this is a nice feature to have.  A few thoughts:

* Can we add a newline before the message?  It weirds me out a little bit having that message on the same line as the prompt :)

* Bash has a couple of features associated with this that might be nice:

  * If you try to quit, it shows a listing of all the jobs (this could be accomplished by adding a call to the [`jobs` builtin](https://github.com/scopatz/xonsh/blob/master/xonsh/jobs.py#L211-L220)).
  * In bash, IIRC, if you try to quit two times in a row, the second will cause the jobs to be killed and the shell closed...or an "are you sure you want to quit? y/N" prompt would do the same trick.  My only concern is that, if someone has a lot of background jobs running, it can be a pain to go through and kill them all manually.  It would be nice to have a way for the user to effectively say "yes, i know i have those things running, but i just want to kill them all and quit."

Along those same lines, it would be nice to have a quick way to kill a particular job (either an alias that takes a job number and kills the associated process, or an easy way to reference a job like bash's `%n` syntax).  But that may want to live in a separate PR.

willsALMANJ:
We can add a newline before the message.

I had wondered about a y/n prompt. I didn't realize bash would kill all jobs on the second exit request. The double exit version seems nice, but given that I didn't realize it existed maybe the prompt version is better. Is there a clean way to track exit being called twice in a row (maybe just referencing the shell history?)?

When I try to exit in bash, I don't see the list of jobs, though we could still add it to xonsh.

It would be nice to have the `%n` syntax. I would also like to reference job by command name as can be done in bash.

willsALMANJ:
I added a prompt for killing all jobs and a call to the jobs builtin.

On my system, I already see the text appearing on a new line below the prompt, so I didn't add that yet. Could this newline be setup dependent? I am using prompt_toolkit.

xonsh/jobs.py
```diff
@@ -199,6 +199,54 @@ def add_job(info):
...
+                job_str = 'job'
+
+            print('xonsh: you have {art}{typ} {job}'.format(art=article,
+                                                            typ=type_str,
+                                                            job=job_str))
+            print('-'*5)
+            jobs([])
+            print('-'*5)
+            resp = input('Kill running jobs [y/n]? ')
+            if 'yes'.startswith(resp.lower()):
```
adqm: Is the intention that an empty string (just hitting enter) will kill the jobs and quit?  Right now, it does.  If not, we'll need an explicit test for the empty string.  Or we could put this in a loop (while the input isn't either `'y'` or `'n'`, keep asking), but that feels a little painful.

Either way, we should probably indicate which of the options is the default by capitalizing it.

Maybe `Kill running jobs and exit [Y/n]?`

adqm:
Thanks again @willsALMANJ; this looks really nice!  A few things:

* The one inline comment
* This could use an entry in `CHANGELOG.rst`
* I don't know what would cause this, but when I run with prompt_toolkit, I do not see the message or the prompt, even with background/stopped jobs (xonsh just quits).  Seems to be working fine with readline for me.

willsALMANJ:
I wanted hitting enter to do something just for convenience, but it would be safer for it not to exit, so I changed it so that is the case now. Also, I changed the wording to `Enter "exit" or "e" to kill all jobs and exit` which I think is a little clearer about what the default will be (not exiting). This is also closer to other shells where entering exit twice will kill all jobs and exit.

I was able to reproduce what you were seeing. I had been testing with `exit` while you were testing with `Ctrl+D`. A couple things regarding this:

* Entering `exit`, `quit`, or `EOF` will insert a newline whereas `Ctrl+D` does not. If the exit prompt starts off by inserting a newline, then a blank newline is printed between the prompt and the exit text. This is preferrable to omitting the newline and having the text start on the line with the prompt for `Ctrl+D`, but is there any way to differentiate and only print the newline for `Ctrl+D`.

* In prompt_toolkit, it seems that `Ctrl+D` does not use the `exit` alias (`exit`, `quit`, and `EOF` do). I haven't figured out what `Ctrl+D` does in prompt_toolkit so far.

After I pushed the most recent commit, I realized that my issue with the way bash handles this was not the fact that it wanted you to exit twice in a row but rather the fact that the prompt about stopped jobs didn't let you know you could exit again to kill them. I have an idea for how to do the jobs check the bash way. I will try it out soon.

xonsh/jobs.py
```diff
@@ -199,6 +199,55 @@ def add_job(info):
...
+                job_str = 'job'
+
+            print()
+            print('xonsh: there {num} {typ} {job}'.format(num=num,
+                                                          typ=type_str,
+                                                          job=job_str))
+            print('-'*5)
+            jobs([])
+            print('-'*5)
+            resp = input('Enter "exit" or "e" to kill all jobs and exit ')
```
scopatz: Let's not use a raw input.  We have a whole suite of wizard tools to handle this that uses the same input tools (readline or ptk) that the user is already using.

xonsh/jobs.py
```diff
@@ -199,6 +199,55 @@ def add_job(info):
...
+                num = 'is a'
+                job_str = 'job'
+
+            print()
+            print('xonsh: there {num} {typ} {job}'.format(num=num,
+                                                          typ=type_str,
+                                                          job=job_str))
+            print('-'*5)
+            jobs([])
+            print('-'*5)
```
scopatz: Please print to `file=sys.stderr` for all of these messages.

xonsh/jobs.py
scopatz: Also it would be nice if the user could simply hit whatever key that they were already using to exit.  IPython does this with ^C and ^D.  Users shouldn't need to type an extra word.

scopatz: This is what IPython does with a ^D:

```console
scopatz@artemis ~ $ ipython
Python 3.5.1 |Continuum Analytics, Inc.| (default, Dec  7 2015, 11:16:01) 
Type "copyright", "credits" or "license" for more information.

IPython 4.2.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: 
Do you really want to exit ([y]/n)? 
scopatz@artemis ~ $ 
```

Hitting Ctrl+D again will be interpreted as a yes.

scopatz:
Thanks for putting this in @willsALMANJ!

willsALMANJ:
Okay, here is a new version that has all the functionality I wanted and addresses the issues you both had. However, there are some things that seem questionable which I enumerate below. I want to see what you think at this stage.

Now `clean_jobs()` stores the current time `_last_exit_time` at the end of its execution. At the beginning, it compares `_last_exit_time` to the first time stamp of the last command in history. If it is later, it concludes that `clean_jobs()` was run as a result of the last command, so it has been called two consecutive times and should kill all unfinished jobs.

1. At first, I tried testing that `_last_exit_time` was in between the two time stamps of the last command in history but sometimes the time was coming out later than the second time stamp (maybe because of some multi-threading?), so this wasn't reliable. I think just checking the start time should be okay.

2. Checking the history like this is susceptible to the user editing the history. I think this is okay since the command editing the history should end up at the end of the history buffer with a later time than `_last_exit_time` unless the user does something somewhat crazy.

3. Doing something like `exit; exit` (or something crazier like `exit; ls; cd; exit`) will kill all jobs and exit but this seems okay.

4. `Ctrl+D` in prompt-toolkit calls an exit function that does not go through the exit alias. I rebound `Ctrl+D` to `EOF` so that it calls the same exit function as every other form of exiting. Right now this has the side-effect of printing "EOF" after the prompt. In readline, `Ctrl+D` results in `EOF` being entered as input (it shows up in history) but nothing is printed after the prompt. If this binding is okay, we can try to figure how to avoid printing "EOF" to the terminal.

5. Check that thhe formatting of the message is okay. I just arbitrarily chose `'-'*5` to separate the messages from the `jobs()` output.

6. While working with the history timestamps, I was wishing there were a way to set attributes of the command that would be saved as last command attributes and then overwritten when the next command completed. Then `clean_jobs()` could just check for an `exit` attribute instead of looking at timestamps, but I don't know if this would be useful for anything else.

7. Minor thing I noticed: the `exit` alias prints an empty newline so that at least one newline is printed after the last xonsh prompt on exit if you use `Ctrl+D`. If you exit by entering `exit`, this results in an extra blank newline being printed. Maybe the shell should check if the cursor is at the beginning of a blank line and only print a newline if not.

melund:
Works well for me on Windows. I think the last statement:
`Attempt to exit again immediately to kill all jobs and exit`
is a little difficult to read.

Consider something simpler:

`Type 'exit' or press 'ctrl-d' again to force quit.`

Maybe in a colored version without quotes:
`Type {INTENSE_GREEN}exit{NO_COLOR} or press {INTENSE_GREEN}ctrl-d{NO_COLOR} again to force quit.`

scopatz:
This looks great @willsALMANJ.  Just let us know when the phrasing issue that @melund brought up is addressed.

willsALMANJ:
Okay, I changed it to the suggested text. I was trying to avoid saying something like `Exit again to exit`, but "force quit" is nice. It could possibly be simplified down to just `Exit again to force quit.` I tried the color version, but I liked the monocolor version better (I don't have a strong opinion though).

Also, I changed the `Ctrl+D` binding in prompt-toolkit so that it now behaves the same as readline. The only difference is readline inserts an `EOF` command into the shell history whiereas prompt-toolkit does not (no exit command appears in the history).

I rebased the commits on master and now nose is hanging after 919 tests. It hangs for master too though, so I think this test behavior is unrelated to my changes.

melund:
Cool. If you think the mono-chrome version looks best that is fine. I will wait and let the other have a final look

adqm:
This looks good to me!  Needs a (potentially complicated) merge from master.  I'm happy to help with that.  @melund, I'll let you have the final say here.

Thanks again for putting this in, @willsALMANJ!

melund:
Besides the Change log there is a small merge conflict in `jobs.py`. It doesn't seem complicated and I resolved it locally to test it. It seems to work as expected, but you better check that some of the other changes in `jobs.py` doesn't interfere with your additions.

willsALMANJ:
I resolved the conflict and the changes look okay to me. Should I squash my branch to one commit?

melund:
No. Don't do that. It is nice to see the history. Anyway. What happens in feature branch stays in a feature branch :)

melund:
Looks good. I am merging it now. Thanks a lot for all the work.

PR #1060, opened 2016-06-01
Don't tell users to invoke setup.py directly.

Carreau:
It's better to use `pip install .`, because yif you **really** like to
do things by hand you would have to pass options like
`--single-version-externally-managed`.

Editable install are also available with `pip install -e .`

And with this we will get all the benefits form pep518 (once we can use
it)

https://www.python.org/dev/peps/pep-0518/

And its future siblings, that will tell pip how to deals with
non-setup.py-based installation/wheel building.

Note that if you want a **specific** pip from a **specific** python
version you can use :

`python -m pip install .`

But that starts to be a bit verbose.

--- 

@scopatz Fix you slides :-), nice talk BTW. You forgot basic things like OMG it syntax highlight while I type !

### DIFF
--- a/docs/linux.rst
+++ b/docs/linux.rst
@@ -29,7 +29,7 @@ the following from the source directory,
 
 .. code-block:: bash
 
-    $ python setup.py install
+    $ pip install .
 
 
 Arch Linux users can install xonsh from the Arch User Repository with e.g.
--- a/docs/osx.rst
+++ b/docs/osx.rst
@@ -29,7 +29,7 @@ the following from the source directory,
 
 .. code-block:: bash
 
-    $ python setup.py install
+    $ pip install .
 
 
 .. include:: add_to_shell.rst
--- a/docs/windows.rst
+++ b/docs/windows.rst
@@ -56,7 +56,7 @@ Now install xonsh:
 .. code-block:: bat
 
    > cd xonsh-master
-   > python setup.py install
+   > pip install .
 
 Next, run xonsh:
 


scopatz:
Thanks for putting this in @Carreau - I am not sure I really understand what the advantage of pip is in this case.  Is there any extra documentation you can point me to that says that running setup.py is no longer a best practice?  I just tried this `pip install .` locally and it didn't seem to do anything because I have xonsh already installed:

```console
scopatz@localhost ~/xonsh master $ pip install .
Processing /home/scopatz/xonsh
  Requirement already satisfied (use --upgrade to upgrade): xonsh==0.3.2 from file:///home/scopatz/xonsh in /home/scopatz/.local/lib/python3.5/site-packages
```

This seems like it is the opposite of the behaviour that a user would want if they are trying to install from master to get a bug fix that hasn't been release yet.  It seems like we would want to tell people to do `pip install --force-reinstall .`, which doesn't seem like that much of a win.  Or am I wrong about this. 

I'll admit that I don't follow the pip development stuff that closely because I try to never use pip :)

Carreau:
> just tried this pip install . locally and it didn't seem to do anything because I have xonsh already installed:

That because `__version__` is init should be `0.3.2.dev`, and if possible `3.2.devX` with X an increasing number (from git --describe ?)

if a version number end on `.devX`, `pip install --upgrade` is smart enough to consider it should reinstall .

If you want a development version/editable install, then you can do a [`pip install -e .`](https://pip.pypa.io/en/latest/reference/pip_install/#editable-installs), and you "just" need to do a git pull to update your xonsh (it symlink stuffs).

Advice from [distutils sig](https://mail.python.org/pipermail/distutils-sig/2014-April/024110.html) prefer `pip install .` vs `setup.py install` vs 

PyPA docs : http://python-packaging-user-guide.readthedocs.io/en/latest/distributing/#working-in-development-mode

Also from [setuptools docs](https://pythonhosted.org/setuptools/setuptools.html):

> However, packaging tools that build binary distributions by running setup.py install on the command line or as a subprocess will require modification to work with setuptools. They should use the --single-version-externally-managed option to the install command, combined with the standard --root or --record options.

So technically `setup.py install` should be an internal detail of how pip works [but everybody knows that Python packaging is broken]. Pep [518](https://www.python.org/dev/peps/pep-0518/) is trying to fix that, slowly, and prepare for build tools that are non-setup.py based. And in the end, I think that telling users the right things to do now would save us quite some work later. 

Using `pip install [-e] .` also allows pip to track some extra metadata as to which package are actually installed, where, and remove them correctly, to typically avoid messages like:

```
Detected a distutils installed project ('blah') which we cannot uninstall. The metadata provided by distutils does not contain a list of files which have been installed, so pip does not know which files to uninstall.
```

Just to be devil advocate though, it might be true that as your setup.py install a jupyter_hook, it might not completly work with `-e`, we woudl have to  also extend the `setup.py develop` command. 


Does that shade **some** light on your questions ?

scopatz:
This seems to fail, though, in a couple of important ways. The first is that it doesn't display the stdout of the install unless the user provides the `-v` option, or better `-vvv`.  These are actually very useful for debugging, since we do a number of important things (like build parser tables) when setup.py is run.  This PR should add this option.

Additionally, it seems to totally fail at installing the xontribs, as it is looking for a `__init__.py` file, which does not exist since it is an implicit namespace package.

```console
scopatz@artemis ~/xonsh master $ pip install --force-reinstall -vvv .
Processing /home/scopatz/xonsh
  Running setup.py (path:/tmp/pip-_13oo5zu-build/setup.py) egg_info for package from file:///home/scopatz/xonsh
    Running command python setup.py egg_info
                               ╓██▄
                              ╙██▀██╕
                             ▐██4Φ█▀█▌
                           ²██▄███▀██^██
                         -███╩▀ " ╒▄█████▀█
                          ║██▀▀W╤▄▀ ▐║█╘ ╝█
                     ▄m▀%Φ▀▀  ╝*"    ,α█████▓▄,▄▀Γ"▀╕
                     "▀██¼"     ▄═╦█╟║█▀ ╓ `^`   ,▄ ╢╕
                      ,▀╫M█▐j╓╟▀ ╔▓▄█▀  '║ ╔    ╣║▌  ▀▄
                   ▄m▀▀███╬█╝▀  █▀^      "ÜM  j▐╟╫╨▒   ╙▀≡═╤═m▀╗
                   █æsæ╓  ╕, ,▄Ä   ▐'╕H   LU  ║║╠╫Å^2=⌐         █
                ▄æ%Å███╠█ª╙▄█▀      $1╙       ║║╟╫╩*T▄           ▌
               ╙╗%▄,╦██▌█▌█╢M         ╕      M║║║║█═⌐ⁿ"^         ╫
                 ╙╣▀████@█░█    ▌╕╕   `      ▌║▐▐║█D═≈⌐¬ⁿ      s ║⌐
                   ╙╬███▓║█`     ▌╚     ╕   ╕▌║▐▐╣▌⌐*▒▒Dù`       ▐▌
                    ╙╬██╨U█      ╟      $ ▌ ▌▌▐▐▐M█▄═≤⌐%       ╓⌐ ▌
                     ║║█▄▌║             ╟ ▌ ▌M▐▐▐M█▀▒▒▒22,       ▐▌
                      ███╙^▌            ║ ▌ ⌐M▐▐▐M█≤⌐⌐¬──        ▐M
                      ║██ ▌╙   ╓       H║ ▌╒ M║▐▐M█"^^^^^"ⁿ      ║
                       ██╕╙@▓   ╕      ▌║ H'  ║▐▐▐█══=.,,,       █
                       ╙█▓╔╚╚█     ╠   ▌└╒ ▌▐ ╚║║║▀****ⁿ -      ╓▌
                        ╙█▌¼V╚▌   ▌  ╕ ▌ ║╒ ║ ▌▒╠█▀≤≤≤≤≤⌐       █
                         ╙█▌╔█╚▌     ┘ M ▌║ ╫ UUM██J^^"        ▐▌
                          ╙██╙█╙▌  ╕$j  ▐⌐▌ ▌║╝╟█Å%%%≈═        █
                           ╙╣█╣█^▌ ╠║▐  ║ ▌▐.DU██^[""ⁿ       -╒▌
                             ▀█▄█`▌ ░M▀ ▌▐ Å£╝╝█╜%≈═╓""w   ⁿ⌐ █
                              `▀▄▀`▌ ▌█▐⌐║▐UW╖██%≤═░*─    =z ▄Γ
                                ╙██╙▄▌█  ▌Å╛╣██╨%╤ƒⁿ=    -` ▄┘
                                  █▌╢▓▌▌ W £6█╤,"ⁿ `   ▄≡▀▀▀
                                   █"█▌▌╟Å╓█╓█▀%`    ▄▀
                                   ╙▌██`▒U▓U█%╗*     █
                                    ▌╫║ ▌ÅÅ║▀╛¬`      `"█
                                    ▌╫  ╫╟ █▄     ~╦%▒╥4^
                                    ▌▌  "M█ `▀╕ X╕"╗▄▀^
                                    █▌   ╓M   ╙▀e▀▀^
                                    ╙██▄▄▀
                                      ^^

    running egg_info
    creating pip-egg-info/xonsh.egg-info
    writing pip-egg-info/xonsh.egg-info/PKG-INFO
    writing top-level names to pip-egg-info/xonsh.egg-info/top_level.txt
    writing dependency_links to pip-egg-info/xonsh.egg-info/dependency_links.txt
    writing entry points to pip-egg-info/xonsh.egg-info/entry_points.txt
    writing manifest file 'pip-egg-info/xonsh.egg-info/SOURCES.txt'
    warning: manifest_maker: standard file '-c' not found

    package init file 'xontrib/__init__.py' not found (or not a regular file)
    reading manifest file 'pip-egg-info/xonsh.egg-info/SOURCES.txt'
    reading manifest template 'MANIFEST.in'
    writing manifest file 'pip-egg-info/xonsh.egg-info/SOURCES.txt'
  Source in /tmp/pip-_13oo5zu-build has version 0.3.2, which satisfies requirement xonsh==0.3.2 from file:///home/scopatz/xonsh
  Requirement already satisfied (use --upgrade to upgrade): xonsh==0.3.2 from file:///home/scopatz/xonsh in /home/scopatz/.local/lib/python3.5/site-packages
Could not parse version from link: file:///home/scopatz/xonsh
Cleaning up...
  Removing source in /tmp/pip-_13oo5zu-build
You are using pip version 8.1.1, however version 8.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

Any ideas on how to address this?

scopatz:
Also, thanks for the very detailed response @Carreau

Carreau:
Hum, ok, I'm closing for now and will investigate.

scopatz:
Well, feel free to keep it open if you want. I do want us to progress forward ;)

Carreau:
No worries. I think in the end it is the same problem as figuring out how to distribute xonsh as a wheel, where the build step should not be done on user computer. We'll figure things out 1 step at a time.

scopatz:
See #1126 for further discussion.

PR #1069, opened 2016-06-02
Add some intro taglines

asmeurer:
One is just an improvement over one that's already there, so maybe it ought to
just replace it. The others are Monty Python, from
https://twitter.com/asmeurer/status/701509800530579456.


scopatz:
Thanks @asmeurer!

PR #1076, opened 2016-06-03
Framework for Programmable Tab Completion

adqm:
This builds on #1027 and drastically improves start-up times (except on the first run) by caching information about bash completion functions.


docs/tutorial_completers.rst
```diff
@@ -0,0 +1,130 @@
...
+xonsh actually uses, in the ``xonsh.completers`` module.
+
+.. code-block:: python
+
+    def dummy_completer(prefix, line, begidx, endidx, ctx):
+        '''
+        Completes everything with options "lou" and "carcolh",
+        regardless of the value of prefix.
+        '''
+        return {"lou", "carcolh"}
```
gforsyth: :grin:

xonsh/completers/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+from xonsh.completers.python import (complete_python, complete_import,
+                                     complete_python_mode)
+from xonsh.completers.commands import complete_skipper
+
+completers = OrderedDict()
+completers['python_mode'] = complete_python_mode
+completers['base'] = complete_base
+completers['skip'] = complete_skipper
+completers['cd'] = complete_cd
+completers['rmdir'] = complete_cd
```
gforsyth: this should be `complete_rmdir`, no?

adqm: Oops...

Yes, indeed.

docs/tutorial_completers.rst
```diff
@@ -0,0 +1,130 @@
...
+================================
+
+As with many other shells, xonsh ships with the ability to complete
+partially-specified arguments upon hitting the "tab" key.
+
+In Python-mode, pressing the "tab" key will complete based on the variable
+names in the current builtins, globals, and locals, as well as xonsh language
+keywords & operators, files & directories, and environment variable names. In
+subprocess-mode, xonsh additionally completes based on the names of any
+executable files on your $PATH, alias keys, and full BASH completion for the
```
scopatz: `Bash` is the appropriate spelling, like `Python`

docs/tutorial_completers.rst
```diff
@@ -0,0 +1,130 @@
...
+* ``"<KEY"``, where ``KEY`` is a pre-existing name, indicates that this should be added before the completer named ``KEY``
+
+.. note:: It is also possible to manipulate ``__xonsh_completers__`` directly,
+          but this is the preferred method.
+
+Removing a Completer
+====================
+
+To remove a completer from the list of active completers, run
+``remove-completer NAME``, where ``NAME`` is the unique identifier associated
```
scopatz: I guess I would prefer an interface that was `completer add` and `completer rm` or similar than two different commands.  I'd be happy to modify this later, though, and should not be blocking here, if you are amicable.

xonsh/ast.py
```diff
@@ -92,7 +92,7 @@ class CtxAwareTransformer(NodeTransformer):
     a subprocess.
     """
 
-    def __init__(self, parser):
+    def __init__(self, parser, wrap_subprocs=True):
```
scopatz: I don't know if this makes sense.  This flag basically disables the class from performing any work. Rather it seems that whatever call this (the execer) should simply made to skip this, if desired. Even if this is desired here, I believe that this would be better as part of the kwargs passed to the methods, and not part of the instance state.

xonsh/completer.py
```diff
@@ -163,448 +36,15 @@ def complete(self, prefix, line, begidx, endidx, ctx=None):
...
-        except:
-            self._options = {}
-
-    def _save_cached_options(self):
-        """Save completions to file."""
-        with open(OPTIONS_PATH, 'wb') as f:
-            pickle.dump(self._options, f)
+                res = out
+                lprefix = len(prefix)
+            if res is not None and res != empty_set:
```
scopatz: Isn't it better to test if `len(res) > 0`, rather than creating an extra set instance?

xonsh/completers/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+completers['python_mode'] = complete_python_mode
+completers['base'] = complete_base
+completers['skip'] = complete_skipper
+completers['cd'] = complete_cd
+completers['rmdir'] = complete_rmdir
+completers['bash'] = complete_from_bash
+completers['man'] = complete_from_man
+completers['import'] = complete_import
+completers['python'] = complete_python
+completers['path'] = complete_path
```
scopatz: I believe that you can pass in `OrderedDict([('python_mode', complete_python_mode), ('base', complete_base), ...])` instead.

xonsh/completers/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+
+builtins.__xonsh_completers__ = completers
```
scopatz: Also this should happen in `xonsh.built_ins` and also be unloaded.

xonsh/completers/_aliases.py
```diff
@@ -0,0 +1,125 @@
...
+    if loc == 'start':
+        new[name] = func
+        for (k, v) in builtins.__xonsh_completers__.items():
+            new[k] = v
+    elif loc == 'end':
+        for (k, v) in builtins.__xonsh_completers__.items():
+            new[k] = v
+        new[name] = func
+    else:
+        dir, rel = loc[0], loc[1:]
```
scopatz: Maybe use a name other than the builtin `dir` :)

xonsh/completers/_aliases.py
```diff
@@ -0,0 +1,125 @@
...
+            if rel == k and dir == '<':
+                new[name] = func
+                found = True
+            new[k] = v
+            if rel == k and dir == '>':
+                new[name] = func
+                found = True
+        if not found:
+            new[name] = func
+    builtins.__xonsh_completers__ = new
```
scopatz: Maybe a clear() and an update() would be better than complete replacement.  That way, folks that have a reference to the original object would not loose it.  Namely,

```python
builtins.__xonsh_completers__.clear()
builtins.__xonsh_completers__.update(new)
```

You could also change new to just be a list of tuples.

xonsh/completers/commands.py
```diff
@@ -0,0 +1,53 @@
...
+
+from xonsh.tools import executables_in
+from xonsh.platform import ON_WINDOWS
+from xonsh.completers.tools import get_filter_function
+
+SKIP_TOKENS = {'sudo', 'time', 'timeit', 'which', 'showcmd', 'man'}
+
+
+def complete_command(cmd, line, start, end, ctx):
+    space = ' '
```
scopatz: Should have a docstring.

xonsh/completers/dirs.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+from xonsh.completers.man import complete_from_man
+from xonsh.completers.path import complete_dir
+
+PREVENT_OTHERS = ['path']
```
scopatz: What is this used for?

xonsh/completers/man.py
```diff
@@ -0,0 +1,47 @@
+import os
+import re
+import pickle
+import subprocess
+
+from xonsh.completers.tools import get_filter_function
+
+OPTIONS_PATH = os.path.join(os.path.expanduser('~'), ".xonsh_man_completions")
```
scopatz: I think we should change this location to the `$XONSH_DATA_DIR`

xonsh/completers/python.py
```diff
@@ -0,0 +1,126 @@
...
+
+XONSH_TOKENS = {
+    'and ', 'as ', 'assert ', 'break', 'class ', 'continue', 'def ', 'del ',
+    'elif ', 'else', 'except ', 'finally:', 'for ', 'from ', 'global ',
+    'import ', 'if ', 'in ', 'is ', 'lambda ', 'nonlocal ', 'not ', 'or ',
+    'pass', 'raise ', 'return ', 'try:', 'while ', 'with ', 'yield ', '+', '-',
+    '/', '//', '%', '**', '|', '&', '~', '^', '>>', '<<', '<', '<=', '>', '>=',
+    '==', '!=', '->', '=', '+=', '-=', '*=', '/=', '%=', '**=', '>>=', '<<=',
+    '&=', '^=', '|=', '//=', ',', ';', ':', '?', '??', '$(', '${', '$[', '..',
+    '...', '![', '!(',
```
scopatz: `@` and `@(` should probably be added here.

xonsh/completers/tools.py
```diff
@@ -0,0 +1,33 @@
+import builtins
+
+
+def get_filter_function():
```
scopatz: This name doesn't make a lot of sense to me.  Also, wouldn't it be faster to have the two other functions defined globally and then simply return one or the other.  These get redefined each time this function is called.

xonsh/completers/tools.py
```diff
@@ -0,0 +1,33 @@
...
+    else:
+        def filt(s, x): return s.lower().startswith(x.lower())
+    return filt
+
+
+def is_iterable(x):
+    try:
+        _ = iter(x)
+        return True
+    except:
```
scopatz: This should probably yield a type error.  Also,  I think the prefered version of this function is simply `isinstance(x, collections.abs.Iterable)`

xonsh/completers/tools.py
```diff
@@ -0,0 +1,33 @@
...
+        return False
+
+
+def justify(s, max_length, left_pad=0):
```
scopatz: Please provide a docstring.  Also, this looks suspiciously similar to tools that are already available in `textwrap` in the standard library.

adqm: Oh, cool.  I had no idea that module existed (I've used this piece of code before...).  You're right I can probably just use that.

xonsh/completers/tools.py
adqm: Again, no idea that one existed.  I'll replace throughout.

xonsh/completers/tools.py
adqm: It's the function that is used to `filter` the list, was my reason for calling it that.  Agree about defining them globally.  Will make that change.

xonsh/completers/python.py
adqm: Agreed.  Will do.

xonsh/execer.py
```diff
@@ -68,7 +68,8 @@ def parse(self, input, ctx, mode='exec'):
         # tokens for all of the Python rules. The lazy way implemented here
         # is to parse a line a second time with a $() wrapper if it fails
         # the first time. This is a context-free phase.
-        tree, input = self._parse_ctx_free(input, mode=mode)
+        tree, input = self._parse_ctx_free(input, mode=mode,
+                                           wrap_subprocs=wrap_subprocs)
```
scopatz: So here is where I think that if `wrap_subprocs` is False, it should just 

```
return self.parser.parse(input, filename=self.filename, mode=mode, debug_level=(self.debug_level > 1))
```

Rather than adding arguments to this and the following function.

xonsh/completers/man.py
adqm: Yep, agreed.

xonsh/completers/dirs.py
adqm: Nothing.  A holdover from an older version.  I'll kill it.

xonsh/completers/commands.py
adqm: Yep, I'll add one.

xonsh/execer.py
```diff
@@ -122,12 +127,13 @@ def exec(self, input, mode='exec', glbs=None, locs=None, stacklevel=2):
...
                                 mode=mode,
-                                stacklevel=stacklevel)
+                                stacklevel=stacklevel,
+                                wrap_subprocs=wrap_subprocs)
         if code is None:
             return None  # handles comment only input
         return exec(code, glbs, locs)
 
-    def _parse_ctx_free(self, input, mode='exec'):
+    def _parse_ctx_free(self, input, mode='exec', wrap_subprocs=True):
```
scopatz: Yeah, I don' think that `wrap_subprocs` is needed here.

xonsh/execer.py
```diff
@@ -144,6 +150,8 @@ def _parse_ctx_free(self, input, mode='exec'):
                 else:
                     raise original_error
             except SyntaxError as e:
+                if not wrap_subprocs:
+                    raise e
```
scopatz: ... which will allow these two line to be removed.

xonsh/completers/_aliases.py
adqm: Oh, yes.  Good call.

xonsh/completers/_aliases.py
adqm: Of course, can do.

xonsh/completers/__init__.py
adqm: That was what I had done originally, but, IIRC, I had a problem with circular imports.  I think I still had the code to unload it on exit, but initializing it here was easier.  I can try to see if there's a better way...

xonsh/main.py
```diff
@@ -190,7 +191,8 @@ def premain(argv=None):
                     'completer': False,
                     'login': False,
                     'scriptcache': args.scriptcache,
-                    'cacheall': args.cacheall}
+                    'cacheall': args.cacheall,
+                    'ctx': builtins.__xonsh_ctx__}
```
scopatz: I am not sure why this change is needed...

xonsh/completers/__init__.py
adqm: Yes, indeed.

xonsh/completer.py
adqm: Yes, that would make some sense.

docs/tutorial_completers.rst
adqm: You're right; that's probably a nicer interface.  I can hack that in here.

scopatz:
This is a huge feat @adqm! Thanks a million for putting this in!  Also, thanks in adavance for all of my questions.

docs/tutorial_completers.rst
scopatz: Thanks!

xonsh/completers/__init__.py
scopatz: Yeah, I think we need to streamline and simplify the startup process. Things are kind of a tangled mess right now. (My bad)

xonsh/completers/dirs.py
scopatz: Ahh great!  I thought I was missing something

xonsh/execer.py
adqm: Oooooooh.  Okay.  I was confused by your comments about this initially, but this cleared it up.  I can make that change.

xonsh/completers/tools.py
scopatz: Thanks!  And, I meant `collections.abc.Iterable`.  Sorry about the typo

xonsh/main.py
adqm: This was needed to make sure that, if `register-completer` is invoked inside the xonshrc, that it actually has access to the functions that are defined there (since `register-completer` looks in `builtins.__xonsh_ctx__` for them).

xonsh/completers/tools.py
scopatz: Yeah, it is worth looking into :).  Anytime we can push code burden into the standard library, we probably should!

adqm:
Thanks for the detailed review!  I agree with all your comments.  I'll go through this in detail, hopefully tonight or tomorrow.

xonsh/completers/tools.py
adqm: I had a hard time making `textwrap` do what I wanted it to, so I would prefer to stick with this for now if that works for you.

adqm:
Alright @scopatz, I think I hit all your comments, and fixed another bug in the process.  Should be ready for another look!

xonsh/main.py
scopatz: OK, thanks for the explanation.

scopatz:
Thanks @adqm! The code looks good to me. Playing around with this now.  Would you mind adding API autodoc hooks? Or we can make another issue for that....

scopatz:
Also, thanks to @gforsyth who has snuck in some commits :)

scopatz:
OK I am going to make an executive decision and merge this. It is freaking amazing! I'll add the API docs directly.

scopatz:
Thanks to everyone who helped with this one!

adqm:
Woo-hoo!  We finally made it!  Thanks to all for help with testing :)

PR #1080, opened 2016-06-03
Fix win prompt

melund:
I think this should fix the prompt problem (reported by @phobson)  on windows when no `$PROMPT` was not set in `.xonshrc`

The problem was caused by #989 which basically enabled bash as a foreign shell for windows. This also meant that xonsh would load the `$PROMPT` variable from bash if it were not set in `.xonshrc`. 

@adqm / @scopatz could this also be a problem on other platforms?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,8 @@ Current Developments
 **Fixed:**
 
  * Fixed bug on Windows where tab-completion for executables would return all files.
+ * Fixed bug on Windows which caused the bash $PROMPT variable to be used when no 
+   no $PROMPT variable was set in .xonshrc 
 
 **Security:** None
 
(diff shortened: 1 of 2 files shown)


adqm:
Great; thanks!  I'll take a look tonight when I have access to a Windows machine.

> could this also be a problem on other platforms?

In theory, yes.  But we haven't had anyone report this problem anywhere but on Windows, so my vote would be not to worry about that until we know it's an issue.

scopatz:
Seems good to me too!  I'll let @adqm test and merge, though :)

adqm:
Fixes the problem for me!  Merging in.

melund:
Great. @adqm..  But if bash's prompt override the default prompt on Windows I think it does the same on other platforms when no .xonshrc is set. Can anyone test

scopatz:
@melund for me on Linux running without a xonshrc, there bash does not override the $PROMPT string, even though bash is executed.  This is likely because bash uses `$PS1`, and not `$PROMPT`

melund:
Ahh. Maybe returning a $PROMPT variable is specific to MSYS.

scopatz:
Could be!

PR #1088, opened 2016-06-03
Clear up sentence about the @ operator in the tutorial

asmeurer:
The code in @ is not run in subprocess mode, but rather, the operator itself
is used in subprocess mode.


adqm:
Yes, that's better.  Thanks!

PR #1093, opened 2016-06-04
add current shell_type to xonfig output

gforsyth:
I think this will be useful for troubleshooting as it's something we
often end up asking when users open issues.


scopatz:
Long overdue, thanks @gforsyth!

Issue #1072, opened 2016-06-03
--shell-type readline doesn't work


asmeurer:
If I start xonsh with `--shell-type readline`, it still uses prompt_toolkit. I have to uninstall prompt_toolkit to make it use readline.

adqm:
Confirmed.  It looks like the value that comes in from the command line arguments is being overridden by the value in the `.xonshrc`.

adqm:
This might make a good first patch for someone.  The issue is with [line 47 of `xonsh/shell.py`](https://github.com/scopatz/xonsh/blob/master/xonsh/shell.py#L47), which totally ignores the value of `shell_type` that was passed in (from the command line arguments).

What we should really do is first check the `shell_type` that was passed in, and only check the value from the environment if that is `None`

adqm:
Fixed in #1100; thanks @VHarisop!  @asmeurer, feel free to re-open if this doesn't work for you.

PR #1100, opened 2016-06-05
Fixes shell_type override in #1072

VHarisop:
`$SHELL_TYPE` is now only used if `--shell-type` is `None`.


scopatz:
Thanks a million @VHarisop!  This works for me.

PR #1106, opened 2016-06-05
Fix killpg crash on osx

mitnk:
This PR trying to fix issue https://github.com/scopatz/xonsh/issues/1012

@scopatz @Carreau please help to review it, thanks.


scopatz:
Thanks a ton for putting this in @mitnk!

xonsh/jobs.py
```diff
@@ -61,7 +61,14 @@ def _kill(job):
...
     def _send_signal(job, signal):
-        os.killpg(job['pgrp'], signal)
+        if ON_DARWIN:
+            # This is kind of workaround for OSX, since os.killpg() may
+            # cause PermissionError, we still need to figure out why.
+            # see Github issue #1012
+            for pid in job['pids']:
+                os.kill(pid, signal)
+        else:
+            os.killpg(job['pgrp'], signal)
```
scopatz: Can you maybe change this around a bit so that there are two `_send_signal()` functions defined.  One if we are one, darwin and another one, otherwise.  Maybe:

```python

if ON_DARWIN:
    def _send_signal(...):
        # This is kind of workaround for OSX, since os.killpg() may
        # cause PermissionError, we still need to figure out why.
        # see Github issue #1012
        for pid in job['pids']:
           os.kill(pid, signal)
else:
    def _send_signal(...)
        os.killpg(job['pgrp'], signal)
```

I think this provides a little better isolation of the problem.

scopatz: I'd really appreciate it!

scopatz:
Also, please comment on the issue when you think it is ready for another look, so that we get and email notification. Thanks!

mitnk:
@scopatz thanks for the detailed information :)

updated a bit, please take a look again.

scopatz:
This looks great to me!  @Carreau, should we merge this temporarily? Or does anyone else with a mac wish to confirm that this works?

mitnk:
I updated the comment a little, please see https://github.com/scopatz/xonsh/issues/1012#issuecomment-223802502 for informations.

xonsh/jobs.py
```diff
@@ -6,13 +6,26 @@
...
+if ON_DARWIN:
+    def _send_signal(job, signal):
+        # On OS X, os.killpg() may cause PermissionError when there are
+        # any zombie processes in the process group.
+        # See github issue #1012 for details
+        for pid in job['pids']:
+            os.kill(pid, signal)
+
+else:
+    def _send_signal(job, signal):
```
gforsyth: Defining this in the `else` block will define `_send_signal` on Windows where previously it wasn't.  I don't know if that's a problem or not, but it would be good to know.  @melund -- does this look like something that might cause problems?

melund: Good catch. I have no idea if this is a problem. I think @BYK has a better grasp on these things. I will try to run with it tomorrow to see if it matters.

adqm: `_send_signal` shouldn't ever be called if we are running under Windows, so I don't think this will be a problem.  That said, it might be nice to have this inside the "else" block of the `ON_WINDOWS` conditional to make that clearer.

mitnk: When I wrote the code, I've already noticed it (defining it in Windows branch), but same thoughts as @adqm - 1) this is cleaner 2) Windows logics would never call it (this is Unix only thing).

We can change else to `elif not Windows`, which is make less readable but accurate.

Thoughts on change or not?

scopatz: I think that an `elif not ON_WINDOWS: pass` block is a good idea.

mitnk:
Updated a bit per guys' suggestions, please help to review it, thanks everyone.

scopatz:
This looks good to me! @mitnk Would you mind adding an entry to the CHANGELOG please about this partial fix?  After that, I think this is ready to be merged.

mitnk:
@scopatz updated, please have a check. Thanks!

scopatz:
Thanks a ton @mitnk! Hopefully this will help these annoying mac issues.  I think the best solution might ultimately be to use psutils on Mac for job control.  But we would need this as a default implementation anyway so it is still very useful.  Thanks again!

CHANGELOG.rst
```diff
@@ -40,6 +40,8 @@ Current Developments
   functions
 * The --shell-type CLI flag now takes precedence over $SHELL_TYPE specified in
   .xonshrc
+* Fixed an issue about ``os.killpg()`` on OS X which caused Xonsh crash with
+  occasionlity
```
BYK: *Did you mean __occasionally__?*

xonsh/jobs.py
```diff
@@ -6,13 +6,29 @@
...
+if ON_DARWIN:
+    def _send_signal(job, signal):
+        # On OS X, os.killpg() may cause PermissionError when there are
+        # any zombie processes in the process group.
+        # See github issue #1012 for details
+        for pid in job['pids']:
+            os.kill(pid, signal)
+
+elif ON_WINDOWS:
+    pass
```
BYK: Do we need something like this on Windows: http://stackoverflow.com/questions/1173342/terminate-a-process-tree-c-for-windows

If so I may be able to create this through `ctypes` as I did with `sudo`.

CHANGELOG.rst
mitnk: Thanks for catching it! I meant "occasionality" :)

xonsh/jobs.py
mitnk: Why we need it on Windows? This is only a private helper function for Unix I guess.

aterrel:
Ya'll moved too fast for me, but just wanted to say this fix the crashes on my Mac OS X system! Thanks @mitnk

Carreau:
Thanks ! Sorry I try to be off my computer on week-end. I'll test that !

Issue #1108, opened 2016-06-05
: commands not working anymore?


mburo:
Not sure what happened, but after pulling commit 
0dcd929392eb31a7136ea399bcc59cd14692e410
typing any of the following commands

if True:
for i in range(0,4):
try:

produce error message 
AttributeError: 'NoneType' object has no attribute 'split'

Other commands still work, like print(3+5)
for loops in functions called by aliases also still work.

xonfig
+------------------+----------+
| xonsh            | 0.3.2    |
| Python           | 3.4.3    |
| PLY              | 3.7      |
| have readline    | True     |
| prompt toolkit   | None     |
| shell type       | readline |
| pygments         | None     |
| on posix         | True     |
| on linux         | True     |
| distro           | Fedora   |
| on darwin        | False    |
| on windows       | False    |
| is superuser     | False    |
| default encoding | utf-8    |
+------------------+----------+

scopatz:
Thanks for reporting @mburo!  I doubt that it was that commit specifically, but I think I see a solution.  Fix incoming.

scopatz:
See #1109

scopatz:
Closing due to fix in #1109. Please reopen, if needed!

PR #1109, opened 2016-06-05
hotfix for #1108

scopatz:
@mburo or @gforsyth, can you take a look, please? Thanks in advance!


mburo:
Alright, so this is a hotfix. How do I apply it - I am not that familiar with git ...

scopatz:
you can do something like:

```console
git checkout -b ns
git pull https://github.com/scopatz/xonsh.git ns
```

mburo:
seems to work

scopatz:
Great thanks!

PR #1111, opened 2016-06-05
Eradicate evil

aterrel:
Consider `python setup.py install` to be harmful, the preferred
way to do this is `pip install <dir>`. Please put on a hair
shirt and see: https://youtu.be/5BqAeN-F9Qs?t=10m1s

### DIFF
--- a/docs/linux.rst
+++ b/docs/linux.rst
@@ -29,7 +29,7 @@ the following from the source directory,
 
 .. code-block:: bash
 
-    $ python setup.py install
+    $ pip install .
 
 
 Arch Linux users can install xonsh from the Arch User Repository with e.g.
--- a/docs/osx.rst
+++ b/docs/osx.rst
@@ -29,7 +29,7 @@ the following from the source directory,
 
 .. code-block:: bash
 
-    $ python setup.py install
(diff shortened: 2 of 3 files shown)


adqm:
Thanks for putting this in, @aterrel!  I wasn't really following it too much, but there was some discussion about this in #1060, which has now been closed.  It seems that there were some issues with using this approach.  Maybe you have some ideas of how to fix them?

aterrel:
haha that's what I get for being a smart a$$. I'll close this and make notes over on #1060

scopatz:
Thanks for putting this in @aterrel!  I would have prefered if #1060 or this had remained open so that it was more obvious.  I'll create a plain old issue instead of a PR.

scopatz:
See #1126

aterrel:
well I could reopen it. I just closed it so that there isn't a zombie PR while I learn enough to figure out what the problem really is here.  I have to admit not packaging a library for the last year has eradicated all the dark arts of PyPA from my brain.

scopatz:
I wish I could forget all of the packaging I have done in just the past 3 months :)

PR #1115, opened 2016-06-05
make CONTRIBUTING a hard link

gforsyth:
GitHub is weird about symlinks and I think this better represents the
desired behavior (if you click CONTRIBUTING, you should see the dev
guide).

@melund -- do hard links work on Windows?

### DIFF
--- a/CONTRIBUTING
+++ b/CONTRIBUTING
@@ -1,1 +1,371 @@
-docs/devguide.rst
+.. _devguide:
+
+=================
+Developer's Guide
+=================
+Welcome to the xonsh developer's guide!  This is a place for developers to
+place information that does not belong in the user's guide or the library
+reference but is useful or necessary for the next people that come along to
+develop xonsh.
+
+.. note:: All code changes must go through the pull request review procedure.
+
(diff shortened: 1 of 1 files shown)


melund:
Yes, they work fine.

scopatz:
Crazy, I had no idea hard links worked on windows :)

melund:
Yes. I think it is symbolic links that are implemented slightly different on Windows.

asmeurer:
Hard links work fine in Windows, but they don't work in git. These are now two separate files.

melund:
Yes. Of course. Git uses hard links internally to link files in the worktree. So hard links can't really go into a git repo.

gforsyth:
My bad.  Right now my inclination is to just leave it.  At least people see the right document when they click CONTRIBUTING (for the moment).  

Or, I suppose, we could make the symlink live in the docs folder instead?

scopatz:
> Or, I suppose, we could make the symlink live in the docs folder instead?

Yeah that is probably the best option.... assuming sphinx will pick it up!

PR #1120, opened 2016-06-05
support functions in $RIGHT_PROMPT

nicolasavru:
Functions and empty strings are already handled properly for PROMPT, so
no need to do something different for RIGHT_PROMPT.


scopatz:
Hi @nicolasavru, thanks a million for putting this in. I am not really sure what this is getting at, but neither of the changes seem like they are needed... Could you explain them please? Thanks!

nicolasavru:
As per http://xon.sh/tutorial.html#customizing-the-prompt: "The PROMPT variable can be a string, or it can be a function (of no arguments) that returns a string." However, setting RIGHT_PROMPT to a function did not work because environ.py:109 would cast the function to a string, resulting in (for example) "<function right_prompt at 0x7f8277f2b840>" being printed instead of the function being called and its result printed.

After removing the string cast, the length check needs to be removed because we can't compute len of a function. partial_format_prompt, which calls the function and does the FORMATTER_DICT formatting, handles empty strings properly (it is used that way for PROMPT and works fine), so that length check was unnecessary to begin with.

nicolasavru:
The "" in the first paragraph should be: ```<function right_prompt at 0x7f8277f2b840>```

scopatz:
Hi @nicolasavru - thanks for the explanation. I understand the problem now, but I think that the solution could use some improvement.  I'll make my comments in-line.

xonsh/ptk/shell.py
```diff
@@ -142,8 +142,6 @@ def rprompt_tokens(self, cli):
         prompt.
         """
         p = builtins.__xonsh_env__.get('RIGHT_PROMPT')
-        if len(p) == 0:
-            return []
```
scopatz: These lines should be added back, but you should make it `if isinstance(p, str) and len(p) == 0:`

xonsh/environ.py
```diff
@@ -106,7 +106,6 @@ def to_debug(x):
     re.compile('\w*PATH$'): (is_env_path, str_to_env_path, env_path_to_str),
     'PATHEXT': (is_env_path, str_to_env_path, env_path_to_str),
     'RAISE_SUBPROC_ERROR': (is_bool, to_bool, bool_to_str),
-    'RIGHT_PROMPT': (is_string, ensure_string, ensure_string),
```
scopatz: So, I generally think it is a bad idea to remove the validators.  That said, these validators are clearly wrong.  What we need are `string_or_callable()` validators and associated tests.  Is this something that you would be willing to implement.  The other validation functions live in `xonsh.tools`.  If you add these:

1. I'd *really* appreciate it, and
2. We'd be able to use them for $PROMPT and `$MULTILINE_PROMPT` too!

nicolasavru: I implemented string_or_callable and applied it to $PROMPT and $MULTILINE_PROMPT, please take a look.

I considered adding another validator to check the return type of a callable, but I'm not sure there is much value in that since the function the user assigns could, depending on global state[1], return a string or something else. I can add that validator if you think there is value in that after all.

[1] My particular function is this:
```
def _retcode():
    if __xonsh_history__.rtns:
        ret_code = __xonsh_history__.rtns[-1]
        if ret_code:
            return str(ret_code)
    return ''

$FORMATTER_DICT['retcode'] = _retcode

def right_prompt():
    retcode = _retcode()
    if retcode:
        return '{YELLOW}{BACKGROUND_RED}{retcode}{NO_COLOR}'
    else:
        return '{branch_color}{curr_branch}{NO_COLOR}'
```
so I think depending on global state will be a common use case.

xonsh/ptk/shell.py
nicolasavru: I don't think that this check has value because the user function could just as easily return an empty string, in which case we'd call partial_format_prompt on an empty string anyway. For that reason, partial_format_prompt should properly handle empty strings anyway (and I verified that it does: ```list(string.Formatter().parse('')) == []```).

scopatz: This is as an optimization to avoid a whole, relatively expensive branch.  You are correct that this won't effect the result of the execution, but it will make the default case of not having a right prompt faster, which is what I was after originally.  I hope that this makes some sense :)

xonsh/environ.py
```diff
@@ -104,11 +105,13 @@ def to_debug(x):
...
     'LOADED_RC_FILES': (is_bool_seq, csv_to_bool_seq, bool_seq_to_csv),
     'MOUSE_SUPPORT': (is_bool, to_bool, bool_to_str),
+    'MULTILINE_PROMPT': (is_string_or_callable, ensure_string, ensure_string),
     re.compile('\w*PATH$'): (is_env_path, str_to_env_path, env_path_to_str),
     'PATHEXT': (is_env_path, str_to_env_path, env_path_to_str),
     'PRETTY_PRINT_RESULTS': (is_bool, to_bool, bool_to_str),
+    'PROMPT': (is_string_or_callable, ensure_string, ensure_string),
     'RAISE_SUBPROC_ERROR': (is_bool, to_bool, bool_to_str),
-    'RIGHT_PROMPT': (is_string, ensure_string, ensure_string),
+    'RIGHT_PROMPT': (is_string_or_callable, ensure_string, ensure_string),
```
scopatz: Thanks for adding these! I don't think that this will properly convert functions, but this is probably not a big deal for the moment.

xonsh/environ.py
scopatz: Yeah, I don't think that checking the return type of a callable is good idea, since they can literally do anything :)

xonsh/ptk/shell.py
nicolasavru: Done. Also added a comment to explain that this is an optimization.

scopatz:
Thanks again @nicolasavru! This looks great to me.  Just waiting on CI.

xonsh/environ.py
nicolasavru: Not sure what we can do about that here without checking the return type of the function, which we agree is a bad idea.

If we really want to, we could do ```str(template())``` on environ.py:1149 and ```str(dots())``` on environ.py:1197, but it would be nicer to do that in centralized location.

scopatz: I have some ideas that involve replacing https://github.com/scopatz/xonsh/blob/master/xonsh/environ.py#L579 with None or some other singleton, or perhaps optionally using cloudpickle to handle more complex types...but I think these are well beyond the scope here.

PR #1121, opened 2016-06-05
Add flags to enable/disable pretty printing and syntax highlighting

adqm:
Sometimes the syntax highlighting is distracting to me, and it is sometimes "incorrect" (or at least not what I would expect), particularly for subprocess-mode commands.  I think we should work on this moving forward, but this changeset provides a workaround (and might be desired by some even if we do have a better highlighter in the future).

Also, the pretty printing can take a while, depending on the size of the object in question.

This changeset adds a few new boolean environment variables:
* `$COLOR_INPUT`: controls whether syntax highlighting should be applied to input in the ptk shell
* `$COLOR_RESULTS`: controls whether syntax highlighting should be applied to _output_ in either shell
* `$PRETTY_PRINT_RESULTS`: controls whether output should be "pretty printed"

### DIFF
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -148,17 +148,21 @@ def undo_args(args):
                     au[k](args)
 
 
+LEXER = pyghooks.XonshLexer()
 def _pprint_displayhook(value):
     if value is None or isinstance(value, HiddenCompletedCommand):
         return
     builtins._ = None  # Set '_' to None to avoid recursion
-    if HAS_PYGMENTS:
-        s = pretty(value)  # color case
-        lexer = pyghooks.XonshLexer()
-        tokens = list(pygments.lex(s, lexer=lexer))
+    env = builtins.__xonsh_env__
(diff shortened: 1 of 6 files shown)


xonsh/main.py
```diff
@@ -148,17 +148,21 @@ def undo_args(args):
                     au[k](args)
 
 
+LEXER = pyghooks.XonshLexer()
```
scopatz: This might not be safe if pygments is not installed.  Maybe pyghooks itself would be a better place for the default lexer instance to live?

scopatz:
Thanks @adqm!  Just the one issue with the stock lexer instance and the merge conflict. I like these.

adqm:
Just merged from master.  I just reverted the lexer to where it was being computed before (rather than storing the global, just grab it each time).

scopatz:
Thanks a million billion @adqm!  Works great for me!

PR #1122, opened 2016-06-05
Store result of wrapped subprocess call in _

adqm:
This modifies the semantics of `_` in interactive mode so that it now stores the `CompletedCommand` instance associated with a wrapped subprocess call.

This allows easier access to the previous command's return code (one of the common complains on the slashdot thread), via `_.rtn`.

Not sure whether this is something everyone wants, but I figured I'd put it in, and we can discuss from here.


scopatz:
Seems legit to me, @adqm! Thanks a ton!

Issue #1105, opened 2016-06-05
empty commands 100% CPU ?


mburo:
Just hitting [ret] repeatedly in the terminal shows that something
is consuming lots of CPU cycles (top shows 100% CPU) when
executing empty commands. It looks like at ~6 commands can
be executed each second. Scrolling may take some cycles, but I
doubt that many.

xonfig
+------------------+----------+
| xonsh | 0.3.2 |
| Python | 3.4.3 |
| PLY | 3.7 |
| have readline | True |
| prompt toolkit | None |
| shell type | readline |
| pygments | None |
| on posix | True |
| on linux | True |
| distro | Fedora |
| on darwin | False |
| on windows | False |
| is superuser | False |
| default encoding | utf-8 |
+------------------+----------+

adqm:
>  It looks like at ~6 commands can be executed each second.

Are you saying that when you hit return, it takes ~1/6 seconds for the prompt to come back?  If so, I think that is unusual.  What is the value of your `$PROMPT` environment variable?  In theory, depending on what values are there, it could be computing something heavy each time...

Also, are you operating inside a Git or Mercurial repo?

mburo:
echo $PROMPT
{env_name}{BOLD_GREEN}{user}@{hostname}{BOLD_BLUE} {cwd}{branch_color}{curr_branch}{NO_COLOR} {BOLD_BLUE}{prompt_end}{NO_COLOR}

I am using git cloned xonsh and cd'ed to ~
It looks as if there is a delay after each comman is executed.
I just tried ls: printing is instant, but then there is a noticeable delay
right after before the next prompt is printed.

adqm:
Can you try, just as a test, setting `$PROMPT = "> "` and seeing whether that makes a difference?

(you may first want to store the original prompt somewhere so you can easily set it back)

mburo:
That did it. 4% CPU.

adqm:
I do know that the computation of `branch_color` and `curr_branch` can take a while, depending on your setup.  I typically remove that bit from my prompt.  I use:

`$PROMPT = '{BOLD_GREEN}{user}@{hostname}{BOLD_BLUE} {short_cwd} {BOLD_BLUE}{prompt_end}{NO_COLOR} '`

Can you try that and see how that one performs?

scopatz:
yeah, we really need to set timeout on those.

mburo:
[...]

Issue #849, opened 2016-05-12
Performance issues with default prompt on Windows


Will-W:
I've just set up Xonsh (from master) on Windows. 
On first load up, the first thing I did was hit enter a couple of times, which had a very poor performance loading the next line of prompt (several hundred milliseconds, I'd estimate). It made a poor first impression.

The default prompt is:
'{env_name}{BOLD_INTENSE_GREEN}{user}@{hostname}{BOLD_INTENSE_CYAN} {cwd}{branch_color}{curr_branch}{NO_COLOR} {BOLD_INTENSE_CYAN}{prompt_end}{NO_COLOR} '

After a little digging, it turns out to be the branch_color function that's causing the slowdown (curr_branch is fine). I have git and mercurial installed, and the performance issue is the same whether the working dir is in a repo or not.

Will-W:
For reference - this is a relatively fast machine (Quadcore 2.7Ghz i7, SSD).

melund:
I don't have a feeling of it beeing slow on windows. What is the ouptut of running `xonfig`?

melund:
Can you try to run the following lines?
```
mel@xonsh_dev ~
$ timeit.repeat('current_branch()',setup='from xonsh.environ import current_branch', number=1)
[0.04200691358022368, 0.04197688888888251, 0.03884483950616868]
mel@xonsh_dev  ~
$ import timeit
mel@xonsh_dev  ~
$ timeit.repeat('current_branch()',setup='from xonsh.environ import current_branch', number=1)
[0.044014617283949065, 0.0399265185184845, 0.051671308641971336]
mel@xonsh_dev  ~
$ timeit.repeat('hg_dirty_working_directory()',setup='from xonsh.environ import hg_dirty_working_directory', number=1)
[0.02125866666665388, 0.02629491358021596, 0.022464790123422063]
mel@xonsh_dev  ~
$ timeit.repeat('git_dirty_working_directory()',setup='from xonsh.environ import git_dirty_working_directory', number=1)
[0.04103940740736789, 0.03377461728399567, 0.03384335802468286]
```

Will-W:
+------------------+-------+
| xonsh            | 0.2.7 |
| Python           | 3.5.1 |
| PLY              | 3.7   |
| have readline    | False |
| prompt toolkit   | 1.0.0 |
| pygments         | 2.1.3 |
| on posix         | False |
| on linux         | False |
| on arch          | False |
| on windows       | True  |
| on mac           | False |
[...]

PR #1123, opened 2016-06-06
Remove version control info from default prompt

adqm:
As mentioned in #1105, the computing the VCS branch information can take a really long time, depending on one's hardware configuration.  I think it is in our interest to remove this from the default prompt (though this is, of course, up for discussion).

I think this will make things more palatable for new users, and those who want the branch information can still make it happen (it's described on the tutorial).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,7 @@ Current Developments
   should be wrapped in ``![]`` automatically
 * ``prompt_toolkit`` is now loaded lazily, decreasing load times when using
   the ``readline`` shell.
+* Removed version control information from the default prompt.
 * RC files are now executed directly in the appropriate context.
 
 **Deprecated:** None
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -138,14 +138,12 @@ def is_callable_default(x):
     return callable(x) and getattr(x, '_xonsh_callable_default', False)
 if ON_WINDOWS:
     DEFAULT_PROMPT = ('{env_name}'
-                      '{BOLD_INTENSE_GREEN}{user}@{hostname}{BOLD_INTENSE_CYAN} '
-                      '{cwd}{branch_color}{curr_branch}{NO_COLOR} '
-                      '{BOLD_INTENSE_CYAN}{prompt_end}{NO_COLOR} ')
+                      '{BOLD_INTENSE_GREEN}{user}@{hostname} '
+                      '{BOLD_INTENSE_CYAN}{cwd} {prompt_end}{NO_COLOR} ')
 else:
     DEFAULT_PROMPT = ('{env_name}'
-                      '{BOLD_GREEN}{user}@{hostname}{BOLD_BLUE} '
-                      '{cwd}{branch_color}{curr_branch}{NO_COLOR} '
-                      '{BOLD_BLUE}{prompt_end}{NO_COLOR} ')
+                      '{BOLD_GREEN}{user}@{hostname} '
+                      '{BOLD_BLUE}{cwd} {prompt_end}{NO_COLOR} ')
 
 DEFAULT_TITLE = '{current_job}{user}@{hostname}: {cwd} | xonsh'
 


scopatz:
Thanks for putting this in @adqm, but I think I am a -1 on this. I like the fact that the branch stuff just works (tm). I believe that new users will find it more annoying to go and learn how to customize their prompt.

I'd rather have a timeout added to the branch functions (maybe 0.05?) so that this can't take too long. I think I mentioned something like this on #849 - which to be fair was like a million years ago :)

I can play around with this over the next couple of days.  I think a timeout is a reasonable middle ground between either having this or not having it. 

Of course, I am always interested to hear other opinions and could certainly be swayed :)

adqm:
I think we're going to continue to disagree here :)  I think the branch info is a non-essential feature, and I think opt-in makes sense.

>  I like the fact that the branch stuff just works (tm). I believe that new users will find it more annoying to go and learn how to customize their prompt.

I disagree pretty strongly.  For one, I think not everyone wants (or at least expects) branch information in the prompt (it's just a `git status` away whenever you need it!).  I also think it can put a bad taste in a new user's mouth if their first impression is that the shell is _sloooooow_, and the annoyance of a slow shell for everyone outweighs the inconvenience of adding a few things to the prompt for those who do want that information.

> I think a timeout is a reasonable middle ground between either having this or not having it.

I disagree here, too.  It might be nice as an adjustable option, but for those who do want the branch information in their prompt, I would imagine most would want it even if it takes time.

> Of course, I am always interested to hear other opinions and could certainly be swayed :) 

Obviously the final say is yours, but I do agree that rather than ~~arguing~~ discussing :smile: amongst ourselves, we should gather more opinions.  Certainly from @melund, @BYK, @gforsyth; and ideally from some of the new users as well.  And we do have two votes in favor of removing it from new users in #1105 :)

Another (different) option if we do end up offering an "express" wizard, would be to ask during the wizard whether people want branch info in the prompt, and choose between the two prompts accordingly.

scopatz:
The pain to new users argument is quite persuasive :)

Maybe we should consider some way to change whole modalities.  So there could be a set of defaults for 'use-everything', 'use-nothing', 'default' (somewhere in the middle), and a whole spectrum of other presets.  Maybe this is too weird and gets away from the notion of defaults... But this could help avoid the matplotlib phenomenon of some people hating the defaults forever.  In this world, the mode would be the major (only?) thing that is selected and described in the express wizard.

Definitely thinking out loud here ;)

melund:
I think I am with @scopatz on this one. It shouldn't be slow of course, but this is a really persuasive feature for new-bees and novice users. And it is something which they wouldn't necessary figure out how to enable immediately. It is also something that Cmder and git-bash does by default on the Windows. So I think Xonsh should do so as well. 

People with more shell/python skills will easily find how to disable it.

scopatz:
after some profiling, it looks like the slow part might be in finding the commands. I will investigate more after i sleep ;)

BYK:
I think if we can make this fast enough, it would be great to have it baked-in. Otherwise speed beats this. The presets idea is a nice one though.

@scopatz what commands are you talking about? Locating the executables?

adqm:
> after some profiling, it looks like the slow part might be in finding the commands

In a big repo for someone not using an SSD, though, even running `git status` can take forever (at least the first time moving into said repo).  So finding the executables isn't the only slow part...

asmeurer:
I'm someone who will always customize my own prompt, but for new users, I agree with Anthony. When I was new to git I made several stupid mistakes that could have need avoided if I had prompt integration enabled (I had no idea such a thing was even possible at the time). Timeouts sound like a good idea for all prompt functions, actually.

mitnk:
> In a big repo ... git status can take forever ...

Exactly! Especially for people who work on a big repo without an IDE - they need to type hundreds of commands in that directory to get work done everyday. Slow is unacceptable.

And when the branch name is long, the prompt can take half width of the typing row. This is also one reason that I don't need branch info in the prompt.

gforsyth:
I'm for keeping it in and adding a timeout.  
People who work in big repos where `git status` will drag are likely to be able to customize their prompt and remove the offending item.  
Whereas new users won't think to look up in the docs how to add a feature that, as @asmeurer points out, they don't even know exists.

As for very long branch names, that might be something we can add to the dynamic_cwd_width

scopatz:
> what commands are you talking about? Locating the executables?

Yeah, locating the executables looks like it can take ~ 1 million years, and that is even before running the command and is currently totally unavoidable.  Furthermore it looks like it doesn't even bother to check if git/hg are in the command cache.  Also, it does this twice once for the color and once for the name.

It seems like these just got totally out of hand.

adqm:
> Furthermore it looks like it doesn't even bother to check if git/hg are in the command cache. Also, it does this twice once for the color and once for the name.

:scream:

adqm:
Yeah, replacing `locate_binary` with the following does make a big difference:
```
def locate_binary(name):
    return name in builtins.__xonsh_commands_cache__
```

Of course, that doesn't _quite_ work, since an alias of `git` or `hg` will confuse things here (maybe we could hook into `which`?  Is that likely to be faster than what we had before?

Also, having just played with this, I have to say again: I really can't stand how long the "clean" check takes for some of the repos I work with.  Figuring out the branch is relatively fast; maybe we could compromise by taking the "branch color" out of the default prompt, but leaving the "branch name" bit?

Carreau:
We can also stat the time to get that info, and if it's generally above a threshold prompt user at startup if they want to **deactivate** showing the branch name.

Also, we can do the same if : 
  - after ~200 shell commands git/hg executable was never found.
  - after ~200 shell commands the user was never in a git/hg directory.

The point is to ask **at startup** (and if a tty right)

The longer killer thing, would be a configuration tool like `fish_config` which make it **easy** for new user to change configuration.


I'm slightly in favor of including that by default, as it's often easy to copy default and remove, than to add. 
Adding is also prone to mistakes, and computers/hardware tend to go faster and faster.

adqm:
One more thing, and then I'll shut up.

I just `cd`ed into a repo of mine that is ~5GB big and has ~2000 subdirectories, all told.  I timed it at around 4 seconds to bring up the prompt.  That's fine if you _know_ to expect it, but totally unacceptable to throw on a new user, in my opinion.

I understand that most people don't work with repos that size (and no one _should_), but it can be pretty slow even for smaller repos, and if I were a complete newcomer and didn't know where that time was being spent, I _know_ my conclusion would be that xonsh itself is unusably slow and not worth learning, not "oh it must be the prompt that is causing it to run slowly and there must be a way for me to change the prompt; let me just go do that".

I think it's a mistake to leave it in (at least the coloring bit, since that seems to be where almost all of the time is spent, as long as we speed up the check for whether `hg` and `git` are present).

Okay, I've said my piece.  I'll bow out for now.

mitnk:
It's fine to keep if most of people here want to. One can always remove the version info by hand.

Then I suggest to move the `Customizing the Prompt` in the tutorial up to front (maybe after section `Starting xonsh`). When a new user started xonsh, one may always want to modify the prompt a bit base on his taste. And note users about the possible slowness of version related variables there.

scopatz:
Ok, I think I would like the opportunity to make it blazing fast.  If it isn't fast enough, we can start removing things like color, or the branch name.

@adqm - is this big repo of yours public, or something that I could have access to for testing?

Also on that note, @ngoldbaum, is there an hg repo that yours truly can provide that is potentially large and slow?

BYK:
@scopatz how about using http://www.pygit2.org/index.html for Git commands. That shaves the overhead of binary discovery + subproc call. This will also make it work even when Git is not installed.

Pretty sure we can do something very similar with `hg` since it is pure Python. Heck, for `hg` we may just try to import `hg` library and see if it is there thus avoiding a dependency.

I'm willing to work on this.

adqm:
> is this big repo of yours public, or something that I could have access to for testing?

unfortunately, it's a repository for a class i run, with a good deal of sensitive student information.  that said, i can break out some of the (still large) pieces of it.

ngoldbaum:
Maybe [pypy](https://bitbucket.org/pypy/) or [mozilla-central](https://hg.mozilla.org/mozilla-central/)?

ngoldbaum:
@byk mercurial is GPL, any project that imports the mercurial library is therefore GPL as well. Even if you don't buy the licensing reason, mercurial is python2 only for now. When I originally implemented the mercurial features I used the mercurial command server library (python-hglib), which is available for python3, but that ended up not being so great, as it was still slow, and it required an extra library just to get hg support on the prompt.

ngoldbaum:
@adqm just in case you're using hg, you might be interested in the [fsmonitor extension](https://www.mercurial-scm.org/wiki/FsMonitorExtension). I believe the plan is to add similar functionality to git eventually, using facebook's watchman utility.

asmeurer:
@adqm I'm curious if https://github.com/git/git/blob/master/contrib/completion/git-prompt.sh has been optimized any. How slow is it in bash for your large repo?

adqm:
@asmeurer: I haven't tried git-prompt, but even a bare `git status` takes several seconds, so I'm not hopeful that anything will speed that up...

Again, if we could get rid of the thing that does the coloring, that would satisfy me, since that's where really a lot of time is spent.

scopatz:
Thanks for all of the feedback, everyone!  I have opened up #1166 as an alternative solution.

scopatz:
@BYK thanks for offering! Feel free to try, if you want!  However, my suspicion is that it won't be much faster.  We basically have a pure python implementation for hg (thanks @ngoldbaum!), and it is consistently slower than the git subprocess one.  I think that this is because it walks up the directory tree and does a bunch of `isdir()` tests until it can't anymore.  However, `__git_ps1` does this in bash ends up being much faster for me (on Linux).  Like 10 - 100 times faster.

adqm:
It seems I've lost this one.  Discussion and the newer implementation in #1166.

PR #1129, opened 2016-06-06
Fix which with extended path

mitnk:
Fixing issue: https://github.com/scopatz/xonsh/issues/1128 - which cannot indicate program inside extend PATH directory


melund:
Good catch @mitnk . Thanks a lot.

PR #1130, opened 2016-06-06
environ.py: mention `xonfig styles` to see color styles + typo's

rahiel:
This mentions `xonfig styles` in the docs, removes a double import and fixes some typo's.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -13,7 +13,6 @@ import socket
 import string
 import subprocess
 import shutil
-from itertools import chain
 import sys
 from warnings import warn
 from collections import (Mapping, MutableMapping, MutableSequence, MutableSet,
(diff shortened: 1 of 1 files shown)


xonsh/environ.py
```diff
@@ -13,7 +13,6 @@
 import string
 import subprocess
 import shutil
-from itertools import chain
```
melund: Any reason you remove chain here?

rahiel: Because it is already imported at line 6 [here](https://github.com/rahiel/xonsh/blob/5f4874e793e4df51495791d11ed7817f9bfb1da5/xonsh/environ.py#L6).

melund: Sorry. Didn't see :)

Issue #1127, opened 2016-06-06
Deleting a literal in ptk shell breaks xonsh


adqm:
As mentioned in #1124, running, e.g., `del "hello.txt"` breaks xonsh completely when using the prompt toolkit shell.  Seems to be true of `del`-ing any literal value (rather than a variable).

Traceback from @klonuo copied below:
```
klonuo@home C:\temp $ del "test.txt"
Traceback (most recent call last):
  File "C:\Python35\Scripts\xonsh-script.py", line 9, in <module>
    load_entry_point('xonsh==0.3.2', 'console_scripts', 'xonsh')()
  File "c:\python35\lib\site-packages\xonsh\main.py", line 254, in main
    shell.cmdloop()
  File "c:\python35\lib\site-packages\xonsh\ptk\shell.py", line 115, in cmdloop
    line = self.singleline(auto_suggest=auto_suggest)
  File "c:\python35\lib\site-packages\xonsh\ptk\shell.py", line 84, in singleline
    line = self.prompter.prompt(**prompt_args)
  File "c:\python35\lib\site-packages\xonsh\ptk\shortcuts.py", line 109, in prompt
    document = cli.run(reset_current_buffer=False)
  File "c:\python35\lib\site-packages\prompt_toolkit\interface.py", line 387, in run
    self.eventloop.run(self.input, self.create_eventloop_callbacks())
  File "c:\python35\lib\site-packages\prompt_toolkit\eventloop\win32.py", line 78, in run
    callbacks.feed_key(k)
  File "c:\python35\lib\site-packages\prompt_toolkit\interface.py", line 916, in feed_key
    self._active_cli.input_processor.feed_key(key_press)
  File "c:\python35\lib\site-packages\prompt_toolkit\key_binding\input_processor.py", line 174, in feed_key
    self._process_coroutine.send(key_press)
  File "c:\python35\lib\site-packages\prompt_toolkit\key_binding\input_processor.py", line 146, in _process
    self._call_handler(matches[-1], key_sequence=buffer)
  File "c:\python35\lib\site-packages\prompt_toolkit\key_binding\input_processor.py", line 190, in _call_handler
    handler.call(event)
  File "c:\python35\lib\site-packages\prompt_toolkit\key_binding\registry.py", line 32, in call
    return self.handler(event)
  File "c:\python35\lib\site-packages\xonsh\ptk\key_bindings.py", line 132, in multiline_carriage_return
    carriage_return(b, event.cli)
[...]

Issue #1124, opened 2016-06-06
dir, set, slash


klonuo:
So I'm on Windows, and thought to try xonsh, as any helper compared to bare Windows console is godsend

```shell
klonuo@home ~ $ dir
<function dir>
klonuo@home ~ $ dir /p
NameError: name 'p' is not defined
klonuo@home ~ $ set
set
klonuo@home ~ $ set p
xonsh: subprocess mode: command not found: set
Did you mean one of the following?
    ren:  Alias
    del:  Alias
klonuo@home ~ $ echo klo
xonsh: subprocess mode: command not found: echo
```

`dir` should list folder, like `ls` on *nix. I know that `dir` is Python keyword, but this is shell application and listing a folder should work.

`dir /p` should list files paged, but it seems that no command can be executed with slash as argument prefix, and guess that all Windows commands have arguments which a prefixed with slash (just like `-` or `--` on *nix)

`set` is pandan for `export` on *nix, and it is great Python keyword. Further `set p` should list all environment variables that start on "p".

`echo` cant be found?

I guess there are other issues as I just tried this app as mentioned, but although it can run on Windows, it does not seem like it provides friendlier environment.

ps. I was surprised that `type` (*nix `cat`) works. go figure ;)

klonuo:
As a comment, reproducible crash:

```pytb
C:\temp>echo klo > test.txt

C:\temp>xonsh
klonuo@home C:\temp $ del "test.txt"
Traceback (most recent call last):
  File "C:\Python35\Scripts\xonsh-script.py", line 9, in <module>
    load_entry_point('xonsh==0.3.2', 'console_scripts', 'xonsh')()
  File "c:\python35\lib\site-packages\xonsh\main.py", line 254, in main
    shell.cmdloop()
  File "c:\python35\lib\site-packages\xonsh\ptk\shell.py", line 115, in cmdloop
    line = self.singleline(auto_suggest=auto_suggest)
  File "c:\python35\lib\site-packages\xonsh\ptk\shell.py", line 84, in singleline
    line = self.prompter.prompt(**prompt_args)
  File "c:\python35\lib\site-packages\xonsh\ptk\shortcuts.py", line 109, in prompt
    document = cli.run(reset_current_buffer=False)
  File "c:\python35\lib\site-packages\prompt_toolkit\interface.py", line 387, in run
    self.eventloop.run(self.input, self.create_eventloop_callbacks())
[...]

PR #1139, opened 2016-06-06
fix issue with trying to delete a literal value

adqm:
I believe this should resolve the issue from #1127 and #1124 where attempting to delete a literal value in prompt toolkit shell would cause xonsh to crash.


scopatz:
Looks great to me.

gforsyth:
Looks good to me, @adqm and works on my end.  I'll merge once tests past

Issue #1083, opened 2016-06-03
Aliases should take precedence over Python variables


asmeurer:
If an alias and a Python variable exist of the same name, the Python variable takes precedence. This is a problem because the pattern of defining an alias that is a Python function is 

```py
def alias_name(args, stdin=None):
    ...

aliases['alias_name'] = alias_name
```

But this is wrong. It has to be

```py
def alias_name(args, stdin=None):
    ...

aliases['alias_name'] = alias_name
del alias_name
```

or

```py
def _alias_name(args, stdin=None):
    ...

aliases['alias_name'] = _alias_name
```

The tutorial makes no indication that this is necessary, by the way.

melund:
I think the Python variable should take precedence. Weird bugs otherwise

adqm:
I agree with @melund here.  But I do think we should at least mention this in the docs.

scopatz:
Yeah, I agree. Python should always win, but we can keep this open until we have a mention in the docs.  Maybe in the FAQ or Gotcha section.

asmeurer:
So what's the "right" way to write aliases. I think just using lambda functions for aliases is a copout, since only simple functions can be written as lambdas.

scopatz:
I typically write the alias function with a leading underscore, since varaibles that follow this naming convention do not get brought into the xonsh execution context when you source an `*.xsh` file or load an rc file (which is effectively the same thing as sourcing it).  That is:

```python
def _myalias(...):
    pass

aliases['myalias'] = _myalias
```

asmeurer:
I pushed a documentation PR to https://github.com/scopatz/xonsh/pull/1142

scopatz:
@asmeurer do you feel that this is closable now?

asmeurer:
Sure

PR #1142, opened 2016-06-06
Add note to docs that alias functions should start with underscores

asmeurer:
See #1083.


melund:
Ahh. Thanks a lot @asmeurer .  Keep the fixes coming :)

Issue #1107, opened 2016-06-05
`cd ..` doesn't work well when directory which is deleted


yangchenyun:
```bash
cd /tmp
mkdir -p foo/bar and cd foo/bar
rm -rf /tmp/foo
cd ..
FileNotFoundError: [Errno 2] No such file or directory
```

Possible solution:
in `fish-shell`, when `rm` is executed, it clears the directory stack.
in `bash`, `cd ..` still moves up but raise an error until a valid parent directory is found. 

<img width="783" alt="screen shot 2016-06-04 at 10 31 03 pm" src="https://cloud.githubusercontent.com/assets/1043403/15803788/0d234fb0-2aa4-11e6-8957-0f69bf4cd102.png">

adqm:
Thanks for reporting, @yangchenyun!  It looks like what we are doing is consistent with Bash, but that fish might be doing something nicer (though I'm not sure clearing the directory stack is necessarily the ultimate "right" answer).  I'll investigate...

adqm:
Oh wait, no you're right.  Ours is totally broken.  I kind of like the bash approach.

scopatz:
Thanks again for reporting @yangchenyun! I have noticed something very similar when using version control and removing directories that I am currently in.

scopatz:
@yangchenyun, @adqm has implemented a fix for this, would you mind trying it out.  It has been merged to master

adqm:
I believe this was fixed in #1144, which has been merged into master.  I'm going to close this issue for now.  If this persists, please feel free to re-open!

PR #1144, opened 2016-06-07
Attempted fix for #1107

adqm:
I believe this should fix the broken behavior when an attempted `cd ..` would move the user into a nonexistent directory, and generally restructures a few things about `cd`.


xonsh/dirstack.py
```diff
@@ -5,28 +5,45 @@
...
 
+
 def _get_cwd():
     try:
         return os.getcwd()
     except (OSError, FileNotFoundError):
         return None
 
 
+def _splitpath(path, sofar=tuple()):
```
scopatz: I think you can just do `sofar=()` for an empty tuple.

scopatz: And is in fact much faster, though it probably doesn't matter:

```console
scopatz@artemis ~ $ timeit "()"
100000000 loops, best of 3: 10.1 ns per loop
scopatz@artemis ~ $ timeit "tuple()"
10000000 loops, best of 3: 54.4 ns per loop
```

adqm: better yet; i can remove this function.  i wrote it for a previous version, but it's not being used in this code.

xonsh/dirstack.py
```diff
@@ -5,9 +5,12 @@
 from glob import iglob
 from argparse import ArgumentParser
 
+from xonsh.tools import get_sep
+
 DIRSTACK = []
 """A list containing the currently remembered directories."""
 
+
 def _get_cwd():
```
scopatz: It seems like this function is not used anymore.  Mind killing it completely?

scopatz:
This looks good to me!

xonsh/dirstack.py
adqm: It looks like it is still used a few other places via imports, so I don't think we should kill it.

scopatz: OK

Issue #1025, opened 2016-05-27
XONSH_SHOW_TRACEBACK to a log file.


Carreau:
Would it be possible to set XONSH_SHOW_TRACEBACK to a filepath, and if so have the traceback dump to this file ?

melund:
Yes. I think that would be possible, and not that hard to do. We would just have to add a little code here: https://github.com/scopatz/xonsh/blob/e5217a8816ca3b48c3f30e885a5d09dee612742d/xonsh/tools.py#L360

Maybe also add XONSH_SHOW_TRACEBACK to the DEFAULT_ENSURES with some functionality to validate its type.  https://github.com/scopatz/xonsh/blob/5d32a74a54e87e778b2f5b3aea19c353caa9f379/xonsh/environ.py#L64 

A PR would very welcome :)

VHarisop:
Hey there, is anyone working on this one?

From what I understand from the above discussion, @Carreau proposed that instead of just `True/False`, a string containing a filepath would also be acceptable and force traceback info to be dumped there.

If that is indeed the case, this would be somewhat confusing with respect to both the programmer and the end-user as well (consider the following .xonshrc file):

```
$ cat .xonshrc
XONSH_SHOW_TRACEBACK = 'True'
```

This can be interpreted in at least two ways:
- create a log file called `True` and output traceback info there (i.e. treated like a string / filepath) or
- output traceback info to stderr (i.e. treated like a boolean, after converting it to `str` and matching it against `'True'` and `'False'`)

I propose that, instead, an extra parameter called something like `XONSH_TRACEBACK_FILE` be introduced, and traceback info to be output there only if `XONSH_SHOW_TRACEBACK` is set to `True`.

I could a submit a PR in the following days, if you agree on the above :)

Carreau:
Yeah, I was unsure about the exact semantics, but a general debug log file, not only for traceback would make sens. 

`XONSH_SHOW_TRACEBACK=True|False` could also be dropped in favor of 

`XONSH_TRACEBACK_FILE` and with specific value like `<stdout>|<stderr>|<devnull>`. 

In general (I know I'm becoming old and grumpy) I'm starting to prefer not to duplicate settings.

scopatz:
Hi @VHarisop - no one is working on this right now to the best of my knowledge and a PR would be most welcome!  I also prefer not duplicating envvars, but I also prefer the `$XONSH_SHOW_TRACEBACK` name.  My vision for how this would work is:

```console
[...]

PR #1145, opened 2016-06-07
Added file option for XONSH_SHOW_TRACEBACK

VHarisop:
In response to #1025 , `$XONSH_SHOW_TRACEBACK` may now be either a `bool` or a `str`. In the former case, the traceback is logged to standard error only. In the latter case, traceback is additionally logged to a file whose path is given by the value of `$XONSH_SHOW_TRACEBACK`.


xonsh/tools.py
```diff
@@ -427,9 +427,20 @@ def print_exception(msg=None):
...
+    show_trace, out_file = to_tb_tuple(traceback_opt)
+    if show_trace:
+        if not out_file:
+            # if the user has not specified a file, output to stderr
+            traceback.print_exc()
+        else:
+            # if the user has specified a file, open it in append mode
+            # and output traceback info there, as well as to stderr
+            traceback.print_exc()
+            traceback.print_exc(file = open(out_file, 'a'))
```
adqm: We probably need a check here in case the user has given an invalid file name (e.g., in a directory that doesn't exist).

Might also be good to put this in a `with open(out_file, 'a') as f` block or similar, just to play it safer about closing this file.

xonsh/tools.py
```diff
@@ -427,9 +427,20 @@ def print_exception(msg=None):
...
-                         '$XONSH_SHOW_TRACEBACK = True\n')
-    if env.get('XONSH_SHOW_TRACEBACK', False):
-        traceback.print_exc()
+                         '$XONSH_SHOW_TRACEBACK = True or '
+                         '$XONSH_SHOW_TRACEBACK = \'<filename>\'\n')
+    # get env option for traceback and convert it to a tuple
+    traceback_opt = env.get('XONSH_SHOW_TRACEBACK', False)
+    show_trace, out_file = to_tb_tuple(traceback_opt)
+    if show_trace:
+        if not out_file:
```
adqm: Maybe better to have an explicit comparison against `None`?  Right now, `None` and `''` will yield the same result.  Maybe they should, but if it's the "`None`" case we're checking for, it might be good to check for that explicitly.

xonsh/tools.py
```diff
@@ -589,6 +600,38 @@ def is_bool(x):
...
+    else:
+        raise ValueError('$XONSH_SHOW_TRACEBACK must either be a boolean '
+                         'or a string containing a file path.')
+
+def tb_tuple_to_str(x):
+    """Handles throwback opt detyping to a string."""
+    if isinstance(x, bool):
+        return str(x)
+    elif isinstance(x, str):
+        return '\'' + str(x) + '\''
```
adqm: This should probably just return `repr(x)` instead of wrapping with quotes manually.

xonsh/environ.py
```diff
@@ -120,6 +121,7 @@ def to_debug(x):
     'XONSH_ENCODING_ERRORS': (is_string, ensure_string, ensure_string),
     'XONSH_HISTORY_SIZE': (is_history_tuple, to_history_tuple, history_tuple_to_str),
     'XONSH_LOGIN': (is_bool, to_bool, bool_to_str),
+    'XONSH_SHOW_TRACEBACK': (is_tb_opt, to_tb_tuple, tb_tuple_to_str),
```
adqm: This feels a little strange to me (the fact that we are changing the form of the environment variable).  Can't we just store the repr of the option to detype, and use ast.literal_eval to retype?  That way, the option as the user set it stays intact.

I think the checks that are currently in `to_tb_tuple` could just as easily live directly in print_exception.

adqm:
Thanks for putting this in @VHarisop!  I think this will be a nice feature.

A couple questions and thoughts inline.

VHarisop:
Thanks for the input, @adqm! I'll get back to you soon.

xonsh/tools.py
VHarisop: Fair point. However, the empty string is also treated as a false (see `_FALSES` in  line 638), while being the only string in `_FALSES` that is not a valid filename. I wouldn't mind an explicit check, but I don't think we're being inconsistent here.

xonsh/tools.py
VHarisop: You are right, especially about the `with ... as` block, I'm already on it.

xonsh/environ.py
VHarisop: If I understand correctly, if those checks are put directly inside `print_exception` and `XONSH_SHOW_TRACEBACK` is removed from `DEFAULT_ENSURES`, a user won't know that `XONSH_SHOW_TRACEBACK` is invalid until an exception is raised (please excuse any possible design misunderstandings from my side, as I've only recently gotten acquainted with the codebase).

On the other hand, If you were making a point strictly about the fact that the representation is changed, I too think we can manage with storing just the `repr` of the option.

adqm: That was my main point, yes.  I would be surprised if I set `$XONSH_SHOW_TRACEBACK = True` and then I checked its value later and it was `(True, None)` or similar.

I was not suggesting removing it entirely from the ensurers, but rather modifying how the detyping and retyping works.  If we're worried about the error reporting, we can have the ensurer print an error message if it is given an invalid value.

xonsh/tools.py
```diff
@@ -427,9 +427,20 @@ def print_exception(msg=None):
...
+    traceback_opt = env.get('XONSH_SHOW_TRACEBACK', False)
+    show_trace, out_file = to_tb_tuple(traceback_opt)
+    if show_trace:
+        if not out_file:
+            # if the user has not specified a file, output to stderr
+            traceback.print_exc()
+        else:
+            # if the user has specified a file, open it in append mode
+            # and output traceback info there, as well as to stderr
+            traceback.print_exc()
```
adqm: I also think that if a user has specified a file as the argument, it probably means they don't want to see the whole traceback in the shell.  We could try to find a way to toggle between these behaviors, but I think it would make sense to display only the error message here, and write the full traceback to the specified file.

scopatz: I think that @VHarisop implemented it this way because that is what I asked for.  I am probably the one with the weird ideas here :).

xonsh/environ.py
scopatz: I think the tuple idea is a reasonable one.  This is basically what we do for history.  With out something like this we would need to have two envvars, which could be annoying too.  I think that is the real trade off here.

scopatz: Frankly, I am happy though, either way.  As long as there is a way to set the output file and keep it displayed on stderr.

tests/test_tools.py
```diff
@@ -491,6 +492,41 @@ def test_is_dynamic_cwd_width():
         obs = is_dynamic_cwd_width(inp)
         yield assert_equal, exp, obs
 
+def test_is_tb_opt():
```
melund: Just give a more descriptive name. No need to save the character: `test_is_traceback_optional`

melund: Ahh. Sorry. Just saw it was tests of the individual functions with the same name. My bad :)

melund: But maybe we should consider tb -> traceback. Characters are free, and "Readability counts"

xonsh/tools.py
VHarisop: @adqm, what do you think would be best? Should I  a) check filename validity manually and raise an exception if something goes wrong (something in the lines of `ValueError: Traceback log file is in a non-existing directory`)   or b) put the `with ... as` block in a `try/except` clause and propagate any caught exception "upwards" with a more informative message?

xonsh/tools.py
VHarisop: Yeap, I tried to stick to what was proposed at the discussion for the relevant issue. I too think that displaying only the error message to `stderr` if a filename has been provided makes more sense, so if @scopatz is ok with that I can fix it in my next commit.

xonsh/tools.py
scopatz: I think that a with-block is all that is needed.

xonsh/tools.py
scopatz: I think that removes all utility for me personally, but if that is what people want... Though frankly, I would prefer two environment variables.

xonsh/environ.py
scopatz: Sorry to give you the run around here.  Seems like there wasn't a lot of consensus about how this should work.

xonsh/tools.py
VHarisop: @scopatz Ithink I'll go with what @adqm proposed for now. I would also prefer an extra environment variable, as I already mentioned in the issue discussion. :)

tests/test_tools.py
VHarisop: @melund Point taken, I'll fix the names too.

xonsh/environ.py
VHarisop: @adqm  Well, it turns out there are a few corner cases.

If a user has specified:
```
$XONSH_SHOW_TRACEBACK = True
```

After detyping, it becomes `'True'`, which is also a perfectly valid filename, and it is not clear whether retyping `'True'` should produce the boolean value or the filename itself.

To avoid such ambiguities, I think it is necessary to either use a structured representation like a tuple or introduce another env. variable that refers exclusively to the log output file.

melund: > To avoid such ambiguities, I think it is necessary to either use a structured representation like a tuple or introduce another env. variable that refers exclusively to the log output file.
 
I am definitely for a separate env-var instead.

adqm: Yeah, I'm coming around to that opinion.  +1 for a separate env-var.  Sorry for the trouble, @VHarisop!  And thanks again for your work here.

VHarisop:
Hey everyone, I just pushed the changes. I tried to stay as close to what we discussed as possible.

In a nutshell:
- `$XONSH_SHOW_TRACEBACK` is restored to its original state. 
- Another option, `$XONSH_TRACEBACK_LOGFILE`, is added, which can be either a filepath, the empty string, or None. This option plays a part only if `$XONSH_SHOW_TRACEBACK` is `True`.
- Both options have been added to `DEFAULT_ENSURES` in `xonsh/environ.py`.

scopatz:
Thanks a million for this refactor @VHarisop! My only comment is that I think the `$XONSH_TRACEBACK_LOGFILE` should work anytime that this isn't an empty string.  That is, it should not rely on `$XONSH_SHOW_TRACEBACK`. This should be a minor change, I would hope.

VHarisop:
@scopatz Not trouble at all, expect another commit soon. :)

xonsh/tools.py
```diff
@@ -428,16 +428,59 @@ def print_exception(msg=None):
     if 'XONSH_SHOW_TRACEBACK' not in env:
         sys.stderr.write('xonsh: For full traceback set: '
                          '$XONSH_SHOW_TRACEBACK = True\n')
```
adqm: Should we also expand this message to mention `$XONSH_TRACEBACK_LOGFILE`, or no?

VHarisop: @adqm I agree, I was a bit hasty and didn't see this note until now.

By the way, it seems that this check is rarely (if ever) triggered, because `XONSH_SHOW_TRACEBACK` is set to a default value of `False` (see `DEFAULT_VALUES` in `xonsh/environ.py`). Therefore, user with an empty `~/.xonshrc` will not see this tip.

adqm: Interestingly, it _used_ to show up.  The idea was: if someone had explicitly set that value, we wouldn't show the message; but if it was just on the default `False`, we would show it.  I think we should keep that logic.  It's a little bit ugly, but we can check `if "XONSH_SHOW_TRACEBACK" no in env._d`.  Or we could be a little cleaner and add a new method to `xonsh.environ.Env` that does that check (by looking in `_d`) so at least we don't have to access the private attribute from outside.  Maybe `xonsh.environ.Env.is_manually_set(varname)`?

VHarisop: The latest commit should fix that.

xonsh/environ.py
```diff
@@ -120,8 +120,10 @@ def to_debug(x):
     'XONSH_ENCODING_ERRORS': (is_string, ensure_string, ensure_string),
     'XONSH_HISTORY_SIZE': (is_history_tuple, to_history_tuple, history_tuple_to_str),
     'XONSH_LOGIN': (is_bool, to_bool, bool_to_str),
+    'XONSH_SHOW_TRACEBACK': (is_bool, to_bool, bool_to_str),
     'XONSH_STORE_STDOUT': (is_bool, to_bool, bool_to_str),
     'XONSH_STORE_STDIN': (is_bool, to_bool, bool_to_str),
+    'XONSH_TRACEBACK_LOGFILE': (is_logfile_opt, to_logfile_opt, logfile_opt_to_str),
```
adqm: Can we also add this variable to `DEFAULT_VALUES`, with a value of `None`?  This will allow the tab completer to pick up this variable name.

adqm: Also, this line is too long for PEP8.  Can you please break onto a couple lines?

xonsh/tools.py
```diff
@@ -425,19 +425,63 @@ def suggest_commands(cmd, env, aliases):
...
+
+    # if the trace option has been set, print all traceback
+    # info to stderr.
+    if show_trace:
         traceback.print_exc()
-    else:
-        exc_type, exc_value, exc_traceback = sys.exc_info()
-        exception_only = traceback.format_exception_only(exc_type, exc_value)
-        sys.stderr.write(''.join(exception_only))
+
```
adqm: PEP8: blank line contains whitespace.  Please remove.

xonsh/tools.py
```diff
@@ -425,19 +425,63 @@ def suggest_commands(cmd, env, aliases):
...
+    if not show_trace:
+        # if traceback output is disabled, print the exception's
+        # error message on stderr.
+        display_error_message()
     if msg:
         msg = msg if msg.endswith('\n') else msg + '\n'
         sys.stderr.write(msg)
 
+def display_error_message():
```
adqm: PEP8: should have one additional blank line before this function.

xonsh/tools.py
```diff
@@ -425,19 +425,63 @@ def suggest_commands(cmd, env, aliases):
...
+    """
+    Prints the error message of the current exception on stderr.
+    """
+    exc_type, exc_value, exc_traceback = sys.exc_info()
+    exception_only = traceback.format_exception_only(exc_type, exc_value)
+    sys.stderr.write(''.join(exception_only))
+
+def is_writable_file(filepath):
```
adqm: PEP8: need 2 blank lines before function.

xonsh/tools.py
```diff
@@ -589,6 +633,42 @@ def is_bool(x):
     """Tests if something is a boolean."""
     return isinstance(x, bool)
 
+def is_logfile_opt(x):
```
adqm: PEP8: 2 blank lines

xonsh/tools.py
```diff
@@ -589,6 +633,42 @@ def is_bool(x):
...
+    """
+    Checks if x is a valid $XONSH_TRACEBACK_LOGFILE option. Returns False
+    if x is not a writable/creatable file or an empty string or None.
+    """
+    if x is None:
+        return True
+    return False if not isinstance(x, str) else \
+           (is_writable_file(x) or x == '')
+
+def to_logfile_opt(x):
```
adqm: PEP8: 2 blank lines

xonsh/tools.py
```diff
@@ -589,6 +633,42 @@ def is_bool(x):
...
+        return x
+    else:
+        # if option is not valid, return a proper
+        # option and inform the user on stderr
+        sys.stderr.write('xonsh: $XONSH_TRACEBACK_LOGFILE must be a '
+                         'filepath pointing to a file that either exists '
+                         'and is writable or that can be created.\n')
+        return None
+
+def logfile_opt_to_str(x):
```
adqm: PEP8: 2 blank lines

adqm:
Okay, this looks great to me @VHarisop; thanks again!

Just one more actual thought, and several small formatting (PEP8) issues, all commented inline.  Once those are taken care of, I am +1 to merge in!

melund:
I think that previously `$XONSH_SHOW_TRACEBACK` defaulted to None. If that was the case it would still show the traceback but with a notice on how to disable/enable. If it was True or False it would never show the notice.

VHarisop:
This should be ok now; if there is anything else left to do, let me know.

melund:
This works very nicely overall. 

My only complaint is showing the extra information about dumping to a file if `$XONSH_SHOW_TRACEBACK` have not been manually set. 
```
snail@sea ~\Documents\xonsh $ raise Exception      
Exception
```
This not something most users would be interested in. Maybe we could show the add the information at the top of the traceback if users actually set `$XONSH_SHOW_TRACEBACK=True`. Then there is a chance that the information is relevant for the user.

adqm:
For me, I think the right thing to do is:

* show the error message as the _first_ line (get right to it)
* show the message about `$XONSH_SHOW_TRACEBACK` if it hasn't manually been set.
* show the message about `$XONSH_TRACEBACK_LOGFILE` if it has not been manually set.

I think it is important to have the option of displaying only the error message, but writing the full traceback to file.

melund:
I usually read these things from the bottom up. So I probably want the error just above the new prompt. I guess that is a matter of habit.  

More importantly: The beautiful thing about xonsh is that you don't get all those error lines every time you make a typo. That is why I don't want to show the notice about `$XONSH_TRACEBACK_LOGFILE` per default. 

Let say a new user sees a bug they find interesting. They will then see that they can set `$XONSH_SHOW_TRACEBACK =True`, and they will trigger the error again.  Now if `$XONSH_TRACEBACK_LOGFILE` has not been manually configured we will show the notice about the log file.

I firmly believe we should minimize the notices unless the users need to see them.

scopatz:
> I firmly believe we should minimize the notices unless the users need to see them.

+1 we don't want to annoy users.

VHarisop:
I tend to agree with @scopatz / @melund in this (albeit not strongly - I wouldn't consider it rare for a user to expect that a logfile option, just less common).
Imo, it should go like this:

-- if `XONSH_TRACEBACK_LOGFILE` has been manually set, do not notify about either of the two options (assume the user is aware of both options if he/she has set the logfile option)
-- if `XONSH_SHOW_TRACEBACK` is `True`, notify about the logfile option if it has not already been set manually
-- if `XONSH_SHOW_TRACEBACK` is `False`, notify about the traceback option only.

mitnk:
@VHarisop Great to see `XONSH_TRACEBACK_LOGFILE`!

I think we better not to make dependencies between ENVs - If user set both `XONSH_TRACEBACK_LOGFILE` and `XONSH_SHOW_TRACEBACK`, xonsh should do both as they suggested, which I think is more natural.

VHarisop:
@mitnk, `XONSH_TRACEBACK_LOGFILE` will work if it set, regardless of whether `XONSH_SHOW_TRACEBACK` is `True` or `False`. The env vars are not dependent as far as functionality is concerned.

The only thing that has changed in my last commit is that the user will only be notified about the availability of the `XONSH_TRACEBACK_LOGFILE` option if he/she has set `XONSH_SHOW_TRACEBACK` (see discussion above).

mitnk:
@VHarisop my bad.. please ignore my comment above. I tried your PR, it works perfect!

xonsh/tools.py
```diff
@@ -647,8 +699,49 @@ def is_bool(x):
...
+    if x is None:
+        return True
+    return False if not isinstance(x, str) else \
+           (is_writable_file(x) or x == '')
+
+
+def to_logfile_opt(x):
+    """
+    Converts a $XONSH_TRACEBACK_LOGFILE option to either a str containing
+    the filepath if it is a writable file or '' if the filepath is not
```
mitnk: here the `''`, did you mean `None`?

xonsh/environ.py
```diff
@@ -575,6 +579,12 @@ def xonshconfig(env):
     'XONSH_STORE_STDOUT': VarDocs(
         'Whether or not to store the stdout and stderr streams in the '
         'history files.'),
+    'XONSH_TRACEBACK_LOGFILE': VarDocs(
+        'Specifies a file to store the traceback log to, regardless of whether '
+        'XONSH_SHOW_TRACEBACK has been set. Its value must be a writable file '
+        'or None / the empty string if traceback logging is not desired. '
+        'Logging to a file is not enabled by default.'
+    )
```
mitnk: to keep code consistent, we can put this `(` at the end of last line, also, put and extra `,` after it would be great (so that we don't need to touch this line when adding the next ENV).

xonsh/environ.py
```diff
@@ -575,6 +579,12 @@ def xonshconfig(env):
...
     }
```
mitnk: Could you help here to remove the four spaces before `}` to make it more PEP8 :)

VHarisop: @mitnk Hope it is ok now. Interestingly, pylint did not complain about the line with the leading spaces.

mitnk:
@VHarisop looks good! Thanks!

melund:
Looks good to me as well. Thanks a lot for putting this in.

adqm:
Woo-hoo!  We made it!  Thanks for all the work here, @VHarisop!

PR #1147, opened 2016-06-07
Enable (some) coverage.

Carreau:
Not sure exactly how xonsh is tested completely, so I expect most of the
coverage process to be wrong (for now)

Codecov have [browser
extensions](https://github.com/codecov/browser-extension) that allow you
to see the coverage directly in the pull requests in GitHub.

We might have to give codecov some limited access to the repo (webhook),
and tell the codecov bot to shut its mouth once it's enabled. But it's
super non intrusive, and kind-of useful when you refactor to see that
your code is tested.

Also set some sensible where coverage would fail only if it drops by 10%
in 1 PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,6 @@ python:
 install:
   - pip install -r requirements-tests.txt
 script:
-  - nosetests -q
+  - nosetests -q --with-coverage --cover-package=xonsh
+after_success:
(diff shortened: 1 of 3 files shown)


scopatz:
Great! Happy to have this @Carreau!

Carreau:
Great !

PR #1149, opened 2016-06-07
moved prompt customize section to front of tutorial

mitnk:
Update tutorial for reasons from here: https://github.com/scopatz/xonsh/pull/1123#issuecomment-224157128

Does this change make sense to you?

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -114,107 +114,6 @@ details, but this *is* pretty cool:
 For easier indentation, Shift+Tab will enter 4 spaces.
 And that about wraps it up for the basics section.  It is just like Python.
 
-Customizing the Prompt
-======================
-Customizing the prompt is probably the most common reason for altering an
-environment variable.  The ``PROMPT`` variable can be a string, or it can be a
-function (of no arguments) that returns a string.  The result can contain
-keyword arguments, which will be replaced automatically:
-
-.. code-block:: xonshcon
-
-    >>> $PROMPT = '{user}@{hostname}:{cwd} > '
-    snail@home:~ > # it works!
-    snail@home:~ > $PROMPT = lambda: '{user}@{hostname}:{cwd} >> '
-    snail@home:~ >> # so does that!
-
-By default, the following variables are available for use:
-
-  * ``user``: The username of the current user
-  * ``hostname``: The name of the host computer
-  * ``cwd``: The current working directory, you may use ``$DYNAMIC_CWD_WIDTH`` to
-    set a maximum width for this variable.
-  * ``short_cwd``: A shortened form of the current working directory; e.g.,
-    ``/path/to/xonsh`` becomes ``/p/t/xonsh``
-  * ``cwd_dir``: The dirname of the current working directory, e.g. ``/path/to`` in
-    ``/path/to/xonsh``.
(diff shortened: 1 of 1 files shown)


mitnk:
This change raises an issue - `environment variable` mentioned in this section which is introduced in next section. But I guess this is okay?

docs/tutorial.rst
```diff
@@ -114,6 +114,107 @@ details, but this *is* pretty cool:
 For easier indentation, Shift+Tab will enter 4 spaces.
 And that about wraps it up for the basics section.  It is just like Python.
 
+Customizing the Prompt
+======================
+Customizing the prompt is probably the most common reason for altering an
+environment variable.  The ``PROMPT`` variable can be a string, or it can be a
+function (of no arguments) that returns a string.  The result can contain
+keyword arguments, which will be replaced automatically:
```
Carreau: > The result can contain keyword arguments, which will be replaced automatically

This sentence make little sens – I see what you mean, the string will be formatted using `.format` and keyword arguments, – but that sounds off. 

I though have no better phrasing.

docs/tutorial.rst
```diff
@@ -114,6 +114,107 @@ details, but this *is* pretty cool:
...
+* ``BACKGROUND_`` may be added to the begining of a color name or hex
+  color to set a background color.  For example, ``BACKGROUND_INTENSE_RED``
+  and ``BACKGROUND_#123456`` can both be used.
+* ``bg#HEX`` or ``BG#HEX`` are shortcuts for setting a background hex color.
+  Thus you can set ``bg#0012ab`` or the uppercase version.
+* ``BOLD_`` is a prefix qualifier that may be used with any foreground color.
+  For example, ``BOLD_RED`` and ``BOLD_#112233`` are OK!
+* ``UNDERLINE_`` is a prefix qualifier that also may be used with any
+  foreground color. For example, ``UNDERLINE_GREEN``.
+* Or any other combination of qualifiers, such as
```
Carreau: we might want to link to pygments somewhere around here not to maintain a full duplicate documentation. At some point when pygments 2.2 is release it might support `#ansi...`names.

Carreau:
Make sens to me. Suggestion inline, but you just moved stuff so I would be +1, and if one day someone want to tackle the comments that would be great !

docs/tutorial.rst
mitnk: I'm not the original author, but how about we change `The result` to `It`?

mitnk:
Since there's some context mentioning `environment variable`, e.g. the first sentence. I'll move this section after section of `environment variable`, that will make the reading smoother.

scopatz:
Hi @mitnk - Thanks for putting this in.  I don't think that the move is good idea. I think that the primary thing that people want to do with their shell is run commands, not fiddle with their prompt.  This PR moves the prompt stuff even before running commands or learning how to use the language. 

I think think a better option would be to just add in a link to the customizing the prompt section in the part that you moved it to.  Does this make sense?  This should be a couple of line change.  Would you mind doing that instead please?

mitnk:
@scopatz  I understand the points you put. Let's close this PR for now. We can add links later if we got any decision in #1123. Thanks for review it!

Issue #1137, opened 2016-06-06
Malformed $PROMPT crashes xonsh


adqm:
From [this comment in the recent Reddit thread](https://www.reddit.com/r/Python/comments/4mng7l/xonsh_using_python_as_your_shell/d3xsxs7).

`$PROMPT` variables with one too many open squigglies break xonsh completely.  I tested a few things, and it doesn't look like there is an easy way to escape the squigglies to prevent this.  Example:

```
hartz@tabby-3 ~ $ $PROMPT = "{{hello"
Traceback (most recent call last):
  File "/usr/local/bin/xonsh", line 9, in <module>
    load_entry_point('xonsh==0.3.2', 'console_scripts', 'xonsh')()
  File "/usr/local/lib/python3.5/site-packages/xonsh/main.py", line 260, in main
    shell.cmdloop()
  File "/usr/local/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 116, in cmdloop
    line = self.singleline(auto_suggest=auto_suggest)
  File "/usr/local/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 63, in singleline
    prompt_tokens = self.prompt_tokens(None)
  File "/usr/local/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 137, in prompt_tokens
    toks = partial_color_tokenize(p)
  File "/usr/local/lib/python3.5/site-packages/xonsh/pyghooks.py", line 206, in partial_color_tokenize
    for literal, field, spec, conv in formatter.parse(template):
ValueError: expected '}' before end of string
```

xonsh then quits.

adqm:
Worse, from the conversation on Reddit, it seems like this is happening to some users _before_ having set anything.  I wonder if this isn't related to the `$PROMPT` issue we had on Windows (#1080).

nicolasavru:
I hit a similar issue when trying to start xonsh for the first time from zsh. I had $PROMPT set to ```[%{\x1b[0;35m%}%*%{\x1b[39;49;m%}][%(!.%{\x1b[91m%}%n%{\x1b[39;49;m%}.%{\x1b[0;36m%}%n%{\x1b[39;49;m%})@%{\x1b[0;37m%}%m%{\x1b[39;49;m%}:%{\x1b[0;33m%}%~%{\x1b[39;49;m%}]%(!.#.$) ```,  (sadness like that is one of the reasons I tried xonsh) which threw this:
```
Traceback (most recent call last):
[...]

PR #1150, opened 2016-06-07
Prompt Fixes

adqm:
I believe (hope) this should fix the issues discussed in #1137.

* the `PROMPT` environment variable from foreign shells is always ignored, regardless of OS.
* xonsh doesn't fail quite so horribly when the prompt is malformed (just displays the template with no color).


xonsh/ansi_colors.py
```diff
@@ -43,36 +43,39 @@ def partial_color_format(template, style='default', cmap=None, hide=False):
...
-            toks.append(field)
-            if conv is not None and len(conv) > 0:
-                toks.append(expl)
-                toks.append(conv)
-            if spec is not None and len(spec) > 0:
-                toks.append(colon)
-                toks.append(spec)
-            toks.append(bclose)
-    return ''.join(toks)
+    try:
```
mitnk: maybe we better move this big `try-catch` to where calling `partial_color_format()` to make code more readable, though we need to `try` every future call of this function?

The other two functions change is the same situation.

adqm: Yeah, I understand that this is a little ugly.  I can refactor a bit, since I think the only piece we probably really need to wrap is the initial call to the parser.  But I do think it is worth avoiding having to wrap every call in try/except.

mitnk: yeah, avoiding try/except on every call is make sense on my second thought :)

adqm: Actually, it turns out that the whole loop does have to be wrapped, since the the parser gives us an iterator.  So I think, although it isn't pretty, this might be the best way.  Maybe some others will have offered their opinions when by the time I wake up :)

scopatz: Might I suggest making the loop a private function that then you stick in a try block here?  I think that is better because it will yield two shorter and more readable functions.

xonsh/environ.py
```diff
@@ -1325,17 +1328,22 @@ def windows_foreign_env_fixes(ctx):
...
     # in order of increasing precedence
     ctx = dict(BASE_ENV)
     ctx.update(os.environ)
-    if ON_WINDOWS:
-        # Windows style PROMPT definitions don't work in XONSH:
-        try:
-            del ctx['PROMPT']
-        except KeyError:
-            pass
+    # other shells' PROMPT definitions generally don't work in XONSH:
```
scopatz: I think this is worth a mention in the PROMPT envvar docs because this will also prevent inheriting the prompt variable even if it was perfectly valid because we launched xonsh from within xonsh.  I definitely think the robustness is worth the trade off, but we should probably let people know.

adqm:
Alright @scopatz, I think this is worth another look now!

xonsh/environ.py
```diff
@@ -388,7 +388,8 @@ def xonshconfig(env):
     'PROMPT': VarDocs(
         'The prompt text. May contain keyword arguments which are '
         "auto-formatted, see 'Customizing the Prompt' at "
-        'http://xon.sh/tutorial.html#customizing-the-prompt.',
+        'http://xon.sh/tutorial.html#customizing-the-prompt. '
+        'This value is never inherited from parent processes.'),
```
gforsyth: Test failure here, @adqm -- I think you want to remove the paren at the end of this line.

adqm: Ugh, this is why we shouldn't document anything...

adqm: okay, just pushed an update.

scopatz:
Thanks @adqm! This fails for me locally, though:  I see the following when I start up a new shell:

```console
{BOLD_GREEN}scopatz@localhost{BOLD_BLUE} ~{BOLD_INTENSE_GREEN}{NO_COLOR} {BOLD_
```

scopatz:
Ahh it seems to work with readline but not ptk.

scopatz:
I can't see what is wrong here

adqm:
Okay, I'll dig in some more.

adqm:
Alright @scopatz, I just pushed another commit that should fix that.  Was careless with shuffling things around...

scopatz:
Thanks a ton @adqm!

PR #1158, opened 2016-06-07
Fix (windows) bug in  locate_binary()

melund:
This fixes a regression in `locate_binary()` which prevented it from finding executables when their exetension was also provided. 

The bug broke the `source-cmd` command and thus  the  `activate`/`deactivate` conda scripts. I am sorry for not noticing.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -768,9 +768,10 @@ def locate_binary(name):
 
     directories = builtins.__xonsh_env__.get('PATH')
 
-    # Windows users expect t obe able to execute files in the same directory without `./`
-    #if ON_WINDOWS:
-    #    directories = [_get_cwd()] + directories
+    # Windows users expect t obe able to execute files in the same directory
+    # without `./`
+    if ON_WINDOWS:
+        directories = [_get_cwd()] + directories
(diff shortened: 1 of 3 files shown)


xonsh/environ.py
```diff
@@ -768,8 +769,8 @@ def locate_binary(name):
     directories = builtins.__xonsh_env__.get('PATH')
 
     # Windows users expect t obe able to execute files in the same directory without `./`
-    if ON_WINDOWS:
-        directories = [_get_cwd()] + directories
+    #if ON_WINDOWS:
```
gforsyth: Should these be uncommented or removed, @melund ?

gforsyth:
Awesome, @melund!  Also, i think your apology should read: "I am sorry for not noticing.  But then I did notice.  And then I fixed it.  And then I added more tests so it would never happen again." :microphone: :droplet:

xonsh/environ.py
melund: Ouch.. Those shouldn't have been removed :(   I have added them back in.

adqm:
This one looks totally good to me.  Merging in now.

PR #1163, opened 2016-06-08
Run Control Fixes

adqm:
This patch makes two changes:

* Prevents xonsh from crashing if there is an error other than a syntax error in a run control file
* Modifies the behavior of the xonfig wizard so that it only under the following conditions:
    * we are running interactively
    * no config.json was loaded
    * none of the rc files exists

This way, if there are errors in all the rc files, but they do indeed exist, we do not show the wizard.


scopatz:
Thanks a million @adqm!

PR #1164, opened 2016-06-08
fix issue in docs

mitnk:
I bet there shouldn't be an `@`.


adqm:
Actually, the `@` was intentional because this is normally run in subprocess mode, and to do this, we have to drop in to Python mode (hence the `@`).

mitnk:
@adqm ahh, right! it's running a command, not work out a string. my fault, sorry!

PR #1165, opened 2016-06-08
Tab Completion for `completer` Alias

adqm:
This changeset introduces tab completion for the `completer` alias (_so meta..._).

It also modifies the tab completion engine slightly, so that a completer may raise `StopIteration` to prevent the system from considering the remaining completers (forcing it to return an empty set of completions).


scopatz:
Looks great to me @adqm!

scopatz:
Sorry about the delay getting to this one.

adqm:
No worries!  Glad it made its way in in the end :)

PR #1166, opened 2016-06-08
Version control branch info speedups

scopatz:
This cleans up the version control branch information handling code to make sure that it is snappy. 

As such this does a few things:

1. Doesn't try to make sure that we have the VC system installed, unless we have already generated cache entries.
2. Adds timeouts to everything, and makes sure that the time out is small (0.01 sec)
3. Timeout yield noticeable differences.  Branch names become `<branch-timeout>` and branch colors become yellow.

All in all, in my testing with large repos in both mercurial and git, I think that this is fast enough.  On my desktop, it drops startup times by about 20-25%.  Since my desktop is relatively fast, I believe that this will have an even great benefit on most other boxes.


adqm:
It looks like when I first open a terminal, I see "branch timeout" in yellow.  Is that to be expected?  I guess it makes sense that it could be timing out even if I'm not in a repo, but it's a little surprising.

adqm:
I also always see the branch in yellow.  Maybe because the coloring check is always timing out?

adqm:
Actually, that's more or less what I was expecting (the branch lookup is fast, and the coloring is slow), which was why I had advocated for just removing the coloring piece from the default.  Though this is roughly equivalent (I've yet to see green or red).

adqm:
Is red supposed to be gone as an option?  Even when changing things in a small enough repo that I normally see green, I can't make it go red by changing things.

adqm:
This is wayyyy faster; nice work!

Just couple of things (mentioned above) related to color that I wasn't expecting, and the initial call (on startup?) seems to time out even when I am just in my home, which causes the yellow `<branch-timeout>`.

A quick glance through the code says it all looks good!  Just those couple of questions.

I still think the `<branch-timeout>` could potentially be confusing to new users, but I'm done fighting that fight for now :)

scopatz:
> This is wayyyy faster; nice work!

Thanks!

> Is red supposed to be gone as an option? Even when changing things in a small enough repo that I normally see green, I can't make it go red by changing things.

Ahh whoops, yeah. Thanks for pointing out these logic issues.  I think everything is working now.

> and the initial call (on startup?) seems to time out even when I am just in my home, which causes the yellow <branch-timeout>

Yeah, I think that this is machine dependent.  I think users should be welcome to adjust this - or their prompt - as they see fit. But the default should be very conservative. Maybe we could double it? For me, I never get this just on startup. I guess i just don't want to eat into the 0.1 sec target for startup that I have in my mind :)

> I still think the <branch-timeout> could potentially be confusing to new users, but I'm done fighting that fight for now :)

Yeah, I agree that this could be odd, but it least it is in their face while not getting in their way.  If you can think of a better phrase I am happy to stick it in :)

melund:
It doesn't quite work on windows yet. 

First, with `$VC_BRANCH_TIMEOUT=0.01` I get the yellow branch stuff in all directories. 
```
snail@sea ~\Documents <branch-timeout> $
```
If I increase it to `0.04` it starts to resolve the branch name, but not the status (it is still yellow) 
```
[x]  snail@sea ~\Documents\xonsh vc $ cd ..                                                                                                      
[x]  snail@sea ~\Documents $
```
I have to set it to 0.09 before that works consistently. 

I realize that we need to use the git binary on windows, but we can use a single call to git to get both status and branch name:
```
$ git status --short --branch --untracked-files=no --porcelain
## vc...origin/vc
 M xonsh/ptk/key_bindings.py
```

The branch name is on the first line. 

If we are not in a git repo it will error:
```
snail@sea ~\Documents $ git status --short --branch --untracked-files=no --porcelain          
fatal: Not a git repository (or any of the parent directories): .git
snail@sea ~\Documents $ _.rtn 
128
snail@sea ~\Documents $       
```

This could probably speed things up. But I still think we need to set `$VC_BRANCH_TIMEOUT` to a higher value on windows.

melund:
Well. It may only complicate things if we combine the git 'branch' 'dirty' commands. 

Are there any good ways to time what is faster. Walking up the directory three in Python looking for `.git` file/folder or starting a subprocess with ['git', 'rev-parse', '--abbrev-ref', 'HEAD']?

ngoldbaum:
Unless I'm missing something, it looks like you're not calling hg with any specific environment variables. This is bad, since people can arbitrarily customize hg. Instead, you should set HGRCPATH to the empty string in the subprocess that invokes hg, as `call_hg_command` used to do. That will ensure the output of running hg doesn't depend on the user's configuration.

ngoldbaum:
Oh oops, looks like you are setting HGRCPATH, sorry for the noise.

scopatz:
No worries @ngoldbaum!  @melund - So I don't think it is a great idea to have a function that does both the branch and color in the current formulation because the branch and color are separate entries in the formatter.  Maybe we could have another 'optimized' formatter key (`color_and_branch`?) that computes both at the same time?  But this would be a separate set of functions.  We could even add that to the default prompt.  I thought about this while writing this pull request and decided it was more of a feature request than a bug, and out of scope :)

As per the default value of VC_BRANCH_TIMEOUT, what do you think is reasonable?  It seems like your system needs about 0.1 to get everything.  Would 0.05 be OK.  This would still be able to get the branch name.  We might be able to bump this down once the `color_and_branch` is implemented ;)

melund:
I agree that a combined formatter key is a separate issue. 

Anyway. It behaves a little bit weird. 

```
snail@sea ~\Documents\xonsh <branch-timeout> $ cd ..
snail@sea ~\Documents <branch-timeout> $ cd ..
snail@sea ~ <branch-timeout> $ cd Documents\
snail@sea ~\Documents $
```
Now suddently it seems to be working. It looks like `Documents` is correctly detected as a non git folder.  However, if I continue it doesn't show the timeout again:

```
snail@sea ~\Documents $ cd xonsh\
snail@sea ~\Documents\xonsh $ 
```
If I then change the timeout back into something large, it doesn't start to work again. 

```
snail@sea ~\Documents\xonsh $ 
$VC_BRANCH_TIMEOUT  = 1
snail@sea ~\Documents\xonsh $ cd ..
snail@sea ~\Documents $ cd xonsh\
snail@sea ~\Documents\xonsh $
```

melund:
If I from the beginning set `$VC_BRANCH_TIMEOUT  = 0.1` it seems to work as expected without going into the non-working state.

adqm:
Even 0.1 would be fine.  Actually, that would probably be better.  That would avoid a lot of the problems I was having with it timing out here, but it would still prevent it from "spinning" on the huge repos.

melund:
Yes, A timeout of 0.1 is OK, but my laptop is also relatively fast (i7 6500U). 

It may need to be higher to also work on slower machines, especially on windows. We just need to find a good balance that still provides a reasonable experience.  The value could also be different between platforms. I better check this on slower machines as well.

adqm:
> I better check this on slower machines as well. 

My main machine is pretty old (and the Windows VM on it is _really_ slow), so I can help test here, too.

melund:
Also I think Windows users will tolerate much higher timeout values. Basically, we got nothing to compare against, and everything else is also slow :)

scopatz:
Ok, I have changed it to be 0.2 on windows and 0.1 on Linux.  After playing around a bit more with this today, I agree that 0.1 is probably the safer bet overall. Does this need anything else?  Thanks again!

adqm:
0.1 is okay.  It is still a noticeable delay, but not unbearable.

One thing that is interesting: in the olden days, on my big repo, after the first time cding in (which took ~5 seconds), the subsequent checks were very fast (presumably because of some caching on Git's part).

With this patch, because none of the calls ever finish, the delay associated with each call is about the same (again, noticeable but not painful).

Though it's not the solution I would have picked, this looks fine to me.  I'll let @melund take a look and have the final say.

melund:
I think that 0.2 is probably good. I stil have the weird behavior that it some times stops working. I.e. it stops showing me the branch names. And no matter what I set the `$VC_BRANCH_TIMEOUT` to doesn't start working again. It seems a little random when it happens.

scopatz:
@adqm - do you think it should be dropped down further on posix?  Happy to take any value between 0.1 and 0.01. 

@melund - I have no idea what is causing that... This hasn't happened to me.

adqm:
> @adqm - do you think it should be dropped down further on posix? Happy to take any value between 0.1 and 0.01. 

It's really hard to say, I think, as what number is reasonable depends an awful lot on people's individual setups.  I think if we are going to keep the branch info (including coloring) in the prompt, we want to give it enough time such that it will succeed for most people.  I think 0.1 is reasonable.

I'm still hung up on how first-time users are likely to respond if things are slow...Can we make a big prominent FAQ somewhere that lets people know that if they think it is slow, they can remove the branch information from the prompt?  I want that to be shoved in people's faces, lest they come to the conclusion that xonsh itself is just unbearably slow.  Or maybe the first time it times out, we could give a warning in the shell itself and a suggestion that removing the branch info from the prompt is a way to speed things up?

Also, having spent a day in it, I have to say that on Cygwin, xonsh is pretty much unusable with branch information on.  Can we please at least remove it from the prompt there?  I don't think the timeout will help anything since it is so slow there that we will almost certainly _always_ be hitting the timeout (I was on a beefy machine and I couldn't handle the wait).  Then again, if we have a prominent warning of some kind in the interactive shell, that takes care of itself on Cygwin.

scopatz:
> Can we please at least remove it from the prompt there?

Sure, I think that is reasonable.  It is fair to assume that cygwin users are going to be at least one rung above entry level.  And the added slowness is killer. 

I'll add some communicating information.

scopatz:
OK, I have updated to include @adqm's recommendations.

scopatz:
Hopefully this is good enough :)

adqm:
Looks good to me!  Just two more thoughts.  I'll comment inline.

xonsh/environ.py
```diff
@@ -143,7 +145,13 @@ def default_value(f):
 def is_callable_default(x):
     """Checks if a value is a callable default."""
     return callable(x) and getattr(x, '_xonsh_callable_default', False)
-if ON_WINDOWS:
+
+if ON_CYGWIN:
+    DEFAULT_PROMPT = ('{env_name}'
+                      '{BOLD_INTENSE_GREEN}{user}@{hostname}'
+                      '{BOLD_INTENSE_CYAN} {cwd}{NO_COLOR}'
```
adqm: At least for me, the `BLUE` color on Cygwin looks just fine.  So I might recommend borrowing from the POSIX prompt instead of the Windows prompt.

Also, can remove the center `{NO_COLOR}` at the end of the this line and `{BOLD_INTENSE_CYAN}` from the start of the next line and still be equivalent, I think.

xonsh/environ.py
```diff
@@ -786,228 +798,227 @@ def _yield_executables(directory, name):
...
+    global _FIRST_BRANCH_TIMEOUT
+    if not _FIRST_BRANCH_TIMEOUT:
+        return
+    _FIRST_BRANCH_TIMEOUT = False
+    print('xonsh: branch timeout: computing the branch name, color, or both '
+          'timed out while formatting the prompt. You may avoid this by '
+          'increaing the value of $VC_BRANCH_TIMEOUT or by removing branch '
+          'fields, like {curr_branch}, from your $PROMPT. See the FAQ '
+          'for more details. This message will be suppressed in the future.',
+          file=sys.stderr)
```
adqm: Right now, it looks like this will be suppressed for the remainder of this session, but not globally.  Can we maybe add an option `$SUPPRESS_BRANCH_TIMEOUT_WARNING` (default `False`) and change the words a little, to say something like the following, maybe?

> ... This message will be suppressed for the remainder of this session.  To suppress this message permanently, set `$SUPPRESS_BRANCH_TIMEOUT_MESSAGE = True` in your `.xonshrc` file

and then only show it if `_FIRST_BRANCH_TIMEOUT` AND `$SUPPRESS_BRANCH_TIMEOUT_MESSAGE`?

adqm:
Alright.  Just added a couple of quick comments.  I think this is just about ready to go in.  Sorry for being such a pain about it.

xonsh/environ.py
scopatz: I think you mean or :)

scopatz:
OK, the remaining changes have been implemented.

xonsh/environ.py
adqm: "and not", actually.  woops. :)

adqm: But of course, you've implemented it as the negation, so "or" is right there.

xonsh/environ.py
```diff
@@ -891,6 +912,7 @@ def current_branch(pad=True):
         branch = get_hg_branch()
     if isinstance(branch, subprocess.TimeoutExpired):
         branch = '<branch-timeout>'
+        _first_branch_timeout_message()
```
adqm: Should we have one of these in the spot where we compute the color, too?

scopatz: I tried that initially, I  believe that the answer is no.  This is because the colors don't actually appear as anything, so it is easy to trigger the warning message because the color times out, without triggering the branch name timeout. It is very strange to see the warning message when you are just in your home directory and the branch name doesn't show up at all, because you are not in a repo.

adqm: Is there a way to have it time out on the color _if we found a branch_?  Sorry to make more complications for you, but that seems to more likely place to timeout, so I would like to have the error message when the coloring fails, if we can.

scopatz: Not without doing something wacky. Branch color and branch name are currently fully independent and I think that is how they should remain.  That said, I do think we should have a combined `{color_and_branch}` and migrate the defaults to that. That function would then also call the message function on timeout.

adqm: Okay, sounds good.  Yeah, if we had that, then I would feel a little bit better.  So maybe I'll go ahead and merge this in, and then I can try to get the `{color_and_branch}` in in the near future (before the next release).

scopatz: Sounds great to me!

Issue #1046, opened 2016-05-31
Consider adding zsh-like directory completion


stefanv:
In zshell, I can type e.g. `cd ~/D/S`, which upon TAB will expand to `cd ~/Dropbox/Shared`.

Any hints on how I can add this to Xonsh?

adqm:
Wow, didn't know that about zsh; that's pretty neat!  Would definitely be good to have in xonsh.

I think it is likely that there will be a change to how tab completion is handled in the near future, via #1027.  If that comes through, this change would probably involve modifying `xonsh.completers.path.complete_path`, to make it include those new elements in the set of elements it returns (rather than adding a new completer to the list).

If that doesn't come through, the right place to change is the analogous [`xonsh.completer.path_complete`](https://github.com/scopatz/xonsh/blob/master/xonsh/completer.py#L408-L423) in the current master.

It is possible that the function that does that expansion is something that could be used elsewhere, so it might make some sense to include it in `xonsh.tools` and to import it for the completer.

scopatz:
Sounds like a nice idea! Probably best to wait until the new completion stuff is in before tackling this, though.

gforsyth:
This has been available for quite some time now, just missed closing this out.

PR #1167, opened 2016-06-08
Fuzzy Path Completion

adqm:
This is a first attempt at implementing fuzzy path completion, as described in #1046.  Can be toggled via `$FUZZY_PATH_COMPLETION`, but defaults to being on.


xonsh/completers/path.py
```diff
@@ -151,6 +151,59 @@ def _quote_paths(paths, start, end):
...
+    folder, path = os.path.split(path)
+    if path == "":
+        return sofar[::-1]
+    elif folder == "":
+        return (sofar + (path, ))[::-1]
+    else:
+        return _splitpath(folder, sofar + (path, ))
+
+
+def expanded_match(ref, typed):
```
Carreau: Would be nice to have some docstrings to get a higher overview of how this is working. 

I'm also unsure, would (below) `glob`, and `*` only fuzzy-match if the only the beginning strings matches ?

Would something like https://docs.python.org/3/library/difflib.html#difflib.get_close_matches on scandir be "better" ?

gforsyth: It seems to match anywhere in the string (pretty well, too!).  

if i do `cd g/ons/xo/pt<TAB>` I get both:
```console
git/xonsh/xonsh/completers/
git/xonsh/xonsh/ptk
```

adqm: I can certainly add some docs!  But what @gforsyth said is correct.  This implements _subsequence_ matching, so `pt` will match anything that has `p` and `t` in it (in that order), regardless of where they show up in the string.

I haven't actually played much with fish/zsh, but I think this is what they do (the `expanded_match` function I nabbed from @GuillaumeLeclerc (#903) and built from there).

xonsh/completers/path.py
```diff
@@ -151,6 +151,59 @@ def _quote_paths(paths, start, end):
     return out
 
 
+def joinpath(path):
+    if path is None:
+        return ''
+    if len(path) == 0:
```
gforsyth: Is there a reason these aren't `elif`?  Not that I imagine it makes too much difference, just curious.

xonsh/completers/path.py
Carreau: > It seems to match anywhere in the string (pretty well, too!).

Neeeat ! what if you switch letters ?

like `cd dev/xonhs` ?

adqm: > what if you switch letters ?

I had something like that in the first commit, but I couldn't quite get it to work as I expected and it led to a lot of matches that I think wouldn't make sense to a lot of people.  Maybe that's because I was using Levenshtein distance, and maybe `difflib` would give some better results.

xonsh/completers/path.py
```diff
@@ -151,6 +151,59 @@ def _quote_paths(paths, start, end):
...
+    elif ref[0] == typed[0]:
+        return expanded_match(ref[1:], typed[1:])
+    else:
+        return expanded_match(ref[1:], typed)
+
+
+def _expand_one(sofar, nextone):
+    out = set()
+    for i in sofar:
+        for j in iglobpath(os.path.join(joinpath(i), '*') if i is not None else '*'):
```
gforsyth: PEP8 long line

xonsh/completers/path.py
adqm: I think in the bytecode, they are likely to be identical in this case, because `return` cuts of all execution (we can only get down to the second `if` if the first condition was false).

gforsyth: I agree the bytecode will be identical (ish).  But then why use the `elif` below?  And do you even need the `else` or just a `return`?

xonsh/completers/path.py
Carreau: > and maybe difflib would give some better results.

Yeah... No:

```
(rootenv)~/dev/ipython master$ SequenceMatcher(None, '-aaa','aaaa').ratio()
0.75
(rootenv)~/dev/ipython master$ SequenceMatcher(None, 'a-aa','aaaa').ratio()
0.5
(rootenv)~/dev/ipython master$ SequenceMatcher(None, 'aa-a','aaaa').ratio()
0.75
(rootenv)~/dev/ipython master$ SequenceMatcher(None, 'aaa-','aaaa').ratio()
0.75
```

extend with more `a`s and move the dash:

dsffi6roes0aaaaa 0asuvork5cyii

xonsh/completers/path.py
adqm: > But then why use the elif below?

:no_mouth:

that's a good question... i think what happened is that i wrote the bottom branches first (if/elif/else), and then found i needed the first one later and didn't bother to change the second if to elif.

i should probably be consistent....

adqm:
Also, looks like this doesn't handle `~` yet.  Should be an easy fix, but this probably isn't _quite_ ready to go yet.

xonsh/completers/path.py
Carreau: plus difflib is often behaving in `O(n*3)` with the length of the sequence you give it, while the optimal IIRC is `O(n.log(n))` IIRC. So you would be better reimplementing your LCS or levenstein correctly.

xonsh/completers/path.py
```diff
@@ -151,6 +151,71 @@ def _quote_paths(paths, start, end):
...
+        return get_sep() + _normpath(os.path.join(*path))
+    else:
+        return _normpath(os.path.join(*path))
+
+
+def _splitpath(path):
+    # convert a path into an intermediate tuple representation
+    # if this tuple starts with '', it means that the path was an absolute path
+    path = _normpath(path)
+    if path.startswith(get_sep()):
```
melund: Absolute path on Windows are kind of different. Don't know if this is really something we need to think about?

adqm: I did a little bit of testing on Windows and it seemed to work.  But you're definitely more expert there than I am.  Is there something that isn't working?

melund:
Yes. I will test some more. But right now I get this error: 

```
[x]  snail@sea ~ $ cd doc/x<TAB>Exception in thread Thread-61:
    self.run()
    self.ctx)
  File "c:\users\mel\documents\xonsh\xonsh\completer.py", line 41, in complete
  File "c:\users\mel\documents\xonsh\xonsh\completers\dirs.py", line 12, in complete_cd
    return complete_dir(prefix, line, start, end, ctx, True)
  File "c:\users\mel\documents\xonsh\xonsh\completers\path.py", line 267, in complete_dir
    o, lp = complete_path(prefix, line, start, end, cdpath)
  File "c:\users\mel\documents\xonsh\xonsh\completers\path.py", line 242, in complete_path
    p = splitpath(os.path.expanduser(prefix))
NameError: name 'splitpath' is not defined

```

adqm:
Ah yes.  My fault.  Didn't change all refs to names when renaming.

adqm:
Should be fixed now.

melund:
For me it seems to be case sensitive, so maybe it is not respecting `$CASE_SENSITIVE_COMPLETIONS`

melund:
This is very neat. The only drawback is that I rarely write something like `cd xoh/xo/comp<tab>` but i often write `cd xonhs<tab>`

adqm:
> so maybe it is not respecting $CASE_SENSITIVE_COMPLETIONS

True.  Working on that now.

>  but i often write cd xonhs<tab>

I agree here.  But I've had enough fish and zsh users tell me that this feature is really useful to them, that I am inclined to believe them.

I had thought about calling this "expanded matching" or "subsequence matching," and separately reserving the word "fuzzy" for the typographical errors, which seems more appropriate to me.

The typo matching is hard to get right without adding a bunch of spurious entries.  I can try that again, though.  Maybe the right thing to do there is to fall back to that _only if_ neither the regular path completion or the "expanded matching" found anything.  Then it might not be too bad if we accidentally return a false positive.

melund:
> I had thought about calling this "expanded matching" or "subsequence matching," and separately reserving the word "fuzzy" for the typographical errors, which seems more appropriate to me.

I totally agree. 

> The typo matching is hard to get right without adding a bunch of spurious entries.

Fuzzy completion on just a single level would still be useful.

adqm:
Alright; I just pushed an update that adds "fuzzy" matching and renames the old thing to be "subsequence" matching.

@melund, @gforsyth, @Carreau, @scopatz, It's ready for another look, I think!  Happy for feedback.

scopatz:
Looks like there is a conflict here now

adqm:
It's always the darned changelog...

Just pushed after merging.

gforsyth:
Hey @adqm -- impressive stuff as always, :grinning: 

I'm not sure if this is an issue or not, but (without ever having used fuzzy matching before) a few things don't work the way I expected them to.  

So I have this directory tree:

```console
git
├ xonsh
├ myxonsh
├ xonshhelpers
```

@melund mentioned that he often transposes the last two letters of a directory, so I tried that: 

```console
cd git/xonhs<TAB>
```
which brings up `xonshhelpers` as an option but not `myxonsh` or `xonsh`.  Is this the kind of thing fuzzy matching should catch?  the `xonshhelpers` match seems to be a subsequence match.

On the other hand, if I have this tree:
```console
tmp
├ Aur
├ Drip
├ Drop
├ aur
├ drip
├ drop
```

and I'm in `tmp` and type `cd Dor<TAB>` I'm offered `Aur/`, `Drop/`, and `aur/` so that seems to be working.  But if I'm one directory up and try `cd tmp/Dor<TAB>` I get nothing.

adqm:
@gforsyth,

 > Is this the kind of thing fuzzy matching should catch?

Yeah, It's still a little wonky.  Right now, the "typo" completer (that does the transpositions, etc) only gets called if the other completers don't return anything.  The problem is that if I don't do that, we get a lot of matches that just don't make sense.  I guess I could change the logic so that the fuzzy matcher works if the _original_ matcher doesn't work (even if we get subsequence matches).

The problem is that with the fuzzy matcher, any option is going to lead to weirdness in _some_ case.  The question is which is the least weird....

gforsyth:
Yeah, that makes sense.  And, in fact, I would be really annoyed if there were a bunch of random completions listed.  That's a hard line to walk, there.

scopatz:
This looks great to me! There is a merge conflict from master.  I think that @gforsyth and @melund have thought about this more than myself, so one of them should probably merge :)

adqm:
Alright, just merged from master (it's always the changelog; can we do something about that?).

@gforsyth, @melund, if there are other suggestions, I'm happy to work on them.  If not, I'd also be happy to see this merged in :)

gforsyth:
This is awesome @adqm. Seriously slick.

Carreau:
> It's always the darned changelog...

in IPython we have a `changelog` folder, where we dump new files, and close to release we aggregate, so there is never any conflict.

adqm:
> in IPython we have a changelog folder, where we dump new files, and close to release we aggregate, so there is never any conflict. 

+1000!  @scopatz and I had a little bit of discussion about this in the chat earlier today, and I think we're likely to switch to something like that.

Issue #1169, opened 2016-06-08
Provide file location as a variable when sourcing a file


willsALMANJ:
As far as I can tell, there is no way for a `.xsh` file to tell where it is located (did I miss something?). It would be nice to have the equivalent of `$BASH_SOURCE` in bash or `__file__` in Python. This is useful if you want to store data in other files in the same directory as the `.xsh` file to be sourced without having to hard code all the paths inside the `.xsh` file.

adqm:
Yes, this is a really good idea.

This might be a good first/early patch for someone.  In `xonsh.main.main`, there is a mode `XonshMode.script_from_file`.  When we're running in that mode, we should be able to grab the filename from the arguments that were passed in and store it in `$XONSH_SOURCE` (maybe after calling `os.path.normpath` and/or `os.path.expanduser` to normalize in some way).

adqm:
This has been implemented in #1179, so I'm going to close this issue.  If you notice any other related issues, please feel free to re-open.

willsALMANJ:
#1179 is great, but it is only half of what I wanted. It defines `$XONSH_SOURCE` within a script when you do `xonsh script.xsh`. I also wanted `$XONSH_SOURCE` to defined when one does `source script.xsh` inside of `xonsh` (`$BASH_SOURCE` behaves this way).

For context, the specific reason I want this is for this autojump script I made for `xonsh`: https://github.com/willsALMANJ/autojump-xonsh.  Right now, you need to add

```
import os
$AUTOJUMP_DIR = os.path.join($HOME, '.config/autojump')
source @(os.path.join($AUTOJUMP_DIR, 'autojump.xsh'))
```

to `.xonshrc` for it to work. `$AUTOJUMP_DIR` is needed right now so that `autojump.xsh` knows where to find `autojump-completion.bash`, so it can add it to `$BASH_COMPLETIONS`. If `autojump.xsh` could determine its own location and assume `autojump-completion.bash` was in the same directory, this could be simplified to just

```
source ~/.config/autojump/autojump.xsh
```

scopatz:
@willsALMANJ - I believe that if you import an xsh file as a module, it will have a `__file__` global on it, just like any other python module you import. If it doesn't this is a bug and should be fixed.

scopatz:
Adding `$XONSH_SOURCE` to the source alias is also not hard. I am reopening this for now.

adqm:
[...]

PR #1179, opened 2016-06-08
Implement #1169

cryzed:
I create the path before branching, so the same work (absolutizing and normalizing the path) isn't done twice if the branch is taken. [os.path.abspath](https://docs.python.org/3/library/os.path.html#os.path.abspath) normalizes _and_ absolutizes the path.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,9 +27,8 @@ v0.3.4
   attempted to load.
 * Only show the prompt for the wizard if we did not attempt to load any run
   control files (as opposed to if none were successfully loaded).
-
-
-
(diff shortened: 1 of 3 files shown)


adqm:
Thanks for putting this one in, too, @cryzed!

I think the only thing that is missing is a bit of documentation for the environment variable (which will then be loaded on to the web site page about environment variables).  This should go in `xonsh/environ.py`, ideally in alphabetical order.  See [the entry for `XONSH_INTERACTIVE`](https://github.com/xonsh/xonsh/blob/master/xonsh/environ.py#L530-L532) for an example.

PR #1181, opened 2016-06-08
Fix #1173

### LINKED ISSUE #1173: Configuration fails on Windows due to colon in filename
`xonfig wizard` fails on Windows because the temporary configuration file created has a colon in its filename.

The relevant output is:

    Would you like to save this state, yes or no [default: no]? yes
    filename [default='C:\\Users\\alowe\\.config\\xonsh\\config.json']:
      File "c:\users\alowe\appdata\local\continuum\anaconda3\lib\shutil.py", line 538, in move
        os.rename(src, real_dst)
    OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'C:\\Users\\alowe\\.config\\xonsh\\config.json' -> 'C:\\Users\\alowe\\.config\\xonsh\\config.2016-06-08T11:18:52.170226.json'
    
      File "c:\users\alowe\appdata\local\continuum\anaconda3\lib\site-packages\xonsh\xonfig.py", line 307, in _wizard
        pv.visit()
      File "c:\users\alowe\appdata\local\continuum\anaconda3\lib\site-packages\xonsh\wizard.py", line 481, in visit
        rtn = super().visit(node)
      File "c:\users\alowe\appdata\local\continuum\anaconda3\lib\site-packages\xonsh\wizard.py", line 302, in visit
        rtn = meth(node)
      File "c:\users\alowe\appdata\local\continuum\anaconda3\lib\site-packages\xonsh\wizard.py", line 538, in visit_wizard
        self.visit(child)
      File "c:\users\alowe\appdata\local\continuum\anaconda3\lib\site-packages\xonsh\wizard.py", line 481, in visit
        rtn = super().visit(node)
      File "c:\users\alowe\appdata\local\continuum\anaconda3\lib\site-packages\xonsh\wizard.py", line 302, in visit
        rtn = meth(node)
      File "c:\users\alowe\appdata\local\continuum\anaconda3\lib\site-packages\xonsh\wizard.py", line 623, in visit_save
        backup_file(fname)
      File "c:\users\alowe\appdata\local\continuum\anaconda3\lib\site-packages\xonsh\tools.py", line 1165, in backup_file
        shutil.move(fname, newfname)
      File "c:\users\alowe\appdata\local\continuum\anaconda3\lib\shutil.py", line 552, in move
        copy_function(src, real_dst)
      File "c:\users\alowe\appdata\local\continuum\anaconda3\lib\shutil.py", line 251, in copy2
        copyfile(src, dst, follow_symlinks=follow_symlinks)
      File "c:\users\alowe\appdata\local\continuum\anaconda3\lib\shutil.py", line 115, in copyfile
        with open(dst, 'wb') as fdst:
    OSError: [Errno 22] Invalid argument: 'C:\\Users\\alowe\\.config\\xonsh\\config.2016-06-08T11:18:52.170226.json'

cryzed:
String concatenations using `__add__` are inefficient after a certain number of additions in a row, so I am using string formatting instead. I made the decision to simply replace ":" with "-" because using `datetime`'s `strftime` to recreate `isoformat`'s behavior would involve replicating quite a bit of [isoformat's logic](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat).

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -1171,7 +1171,7 @@ def backup_file(fname):
     import shutil
     from datetime import datetime
     base, ext = os.path.splitext(fname)
-    newfname = base + '.' + datetime.now().isoformat() + ext
+    newfname = '%s.%s%s' % (base, datetime.now().isoformat().replace(':', '-'), ext)
     shutil.move(fname, newfname)
 
 
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -1171,7 +1171,8 @@ def backup_file(fname):
     import shutil
     from datetime import datetime
     base, ext = os.path.splitext(fname)
-    newfname = '%s.%s%s' % (base, datetime.now().isoformat().replace(':', '-'), ext)
+    timestamp = datetime.now().strftime('%Y-%m-%d-%H-%M-%S-%f')
+    newfname = '%s.%s%s' % (base, timestamp, ext)
     shutil.move(fname, newfname)
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -46,7 +46,7 @@ v0.3.4
   ``activate``/``deactivate`` aliases for the conda environements. 
 * Fixed crash resulting from errors other than syntax errors in run control
   file.
-
+* Fixed xonfig wizard failing on Windows due to colon in created filename.
 
 
 v0.3.3
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,9 @@ Current Developments
 
 **Removed:** None
 
-**Fixed:** None
+**Fixed:**
+
+* Fixed xonfig wizard failing on Windows due to colon in created filename.
 
 **Security:** None
 
@@ -46,7 +48,6 @@ v0.3.4
   ``activate``/``deactivate`` aliases for the conda environements. 
 * Fixed crash resulting from errors other than syntax errors in run control
   file.
-* Fixed xonfig wizard failing on Windows due to colon in created filename.
 
 
 v0.3.3


xonsh/tools.py
```diff
@@ -1171,7 +1171,7 @@ def backup_file(fname):
     import shutil
     from datetime import datetime
     base, ext = os.path.splitext(fname)
-    newfname = base + '.' + datetime.now().isoformat() + ext
+    newfname = '%s.%s%s' % (base, datetime.now().isoformat().replace(':', '-'), ext)
```
adqm: To me, I think it is more readable just to use our own formatting string.  Maybe something like the following?
`time.strftime('%Y-%m-%d-%H-%M-%S-%f')`

adqm: Also, it looks like this change kicked this line over the PEP8 limit of 80 characters per line.  Maybe we could assign that timestamp to a variable on the line before, and then reference that here?

adqm:
Thanks for putting this in, @cryzed!  Just a couple quick thoughts inline.

xonsh/tools.py
cryzed: I wanted to avoid using `strftime` because according to `isoformat`'s docs it implements some special behavior depending on `datetime`'s global state:

> If utcoffset() does not return None, a 6-character string is appended, giving the UTC offset in (signed) hours and minutes: YYYY-MM-DDTHH:MM:SS.mmmmmm+HH:MM or, if microsecond is 0 YYYY-MM-DDTHH:MM:SS+HH:MM

But you are right, preserving that behaviour is irrelevant for this. I'll make the suggested fixes.

adqm:
Alright!  This looks good to me.  Tested on Windows, and this does indeed seem to fix the problem.  Merging in now.

PR #1182, opened 2016-06-09
Implement '+' and '-' for fg, and related changes

jeremydonahue:
This implements '+' and '-' for fg for BASHwards compatibility. I use job control a ton and switch between jobs frequently, so these are important to me. I also modified print_one_job() to include a job's position (indicated with a '+' and '-' after the job number). 

This PR also includes a fix to the developer's guide, fixing a reference to docker.py which was renamed to xonsh-in-docker.py.

### DIFF
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -253,9 +253,9 @@ def fg(args, stdin=None):
         act = tasks[0]  # take the last manipulated task by default
     elif len(args) == 1:
         try:
-            if args[0] == '+': # take the last manipulated task
-                 act = tasks[0]
-            elif args[0] == '-': # take the second to last manipulated task
+            if args[0] == '+':  # take the last manipulated task
+                act = tasks[0]
+            elif args[0] == '-':  # take the second to last manipulated task
                 act = tasks[1]
             else:
                 act = int(args[0])
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -182,12 +182,13 @@ def print_one_job(num):
         job = builtins.__xonsh_all_jobs__[num]
     except KeyError:
         return
+    pos = '+' if tasks[0] == num else '-' if tasks[1] == num else ' '
     status = job['status']
     cmd = [' '.join(i) if isinstance(i, list) else i for i in job['cmds']]
     cmd = ' '.join(cmd)
     pid = job['pids'][-1]
     bg = ' &' if job['bg'] else ''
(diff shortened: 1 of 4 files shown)


adqm:
This is great; thanks for putting it in @jeremydonahue!

Just a couple small pedantic, style-related things:
* PEP8: line 256 and 258 need another space before the comment (at least two spaces before inline comment)
* PEP8: line 257 has one too many spaces on its indent (indentation is not a multiple of four)

Also, this could use an entry in CHANGELOG.rst, under "current developments."

xonsh/jobs.py
```diff
@@ -251,9 +253,15 @@ def fg(args, stdin=None):
...
-            act = int(args[0])
-        except ValueError:
+            if args[0] == '+': # take the last manipulated task
+                 act = tasks[0]
+            elif args[0] == '-': # take the second to last manipulated task
+                act = tasks[1]
+            else:
+                act = int(args[0])
+        except (ValueError, IndexError):
             return '', 'Invalid job: {}\n'.format(args[0])
```
scopatz: Shouldn't this be all of the args, rather than args[0]?

scopatz:
Thanks for putting this in @jeremydonahue!

xonsh/jobs.py
adqm: This only happens if `len(args) == 1`, though, right?  So all this does is get rid of the square brackets?

scopatz: I think this could be triggered in the elif, statement though too.

scopatz: Which access index-1

jeremydonahue: @adqm that's right, fg should only ever take one or no arguments so it's fine to just grab the first one, and catching IndexError provides proper error handling in this case.

adqm: Wait, I don't understand.  This whole try/except is inside a branch that has `elif len(args) == 1`.  So we should be guaranteed to have exactly one element in `args` here...

Am I missing something?  I'm not used to being up this late...

jeremydonahue: I don't think you're missing anything. Maybe I should re-write it in a cleaner way? That method could easily get out of hand.

adqm:
This is great, @jeremydonahue; thanks again!  I'm going to go ahead and merge in.

jeremydonahue:
Awesome! No problem, @adqm. I'm really liking Xonsh so far, great job guys!

Issue #1175, opened 2016-06-08
!(command) could be usefully iterable


eode:
So, this is nicely doable with regular file objects:

```python
for line in open('somefile.txt'):
    # do something with each line
```

..this is *very* succinct when dealing with something you only care about the textual output of, as is the case with a lot of system calls.

It might be nice to do something like this:

```python
for line in !(somecommand):
    # do something with each line of stdout
```

..and even more awesome (IMO) if one could do something like this:

```python
try:
    for line in !(somecommand):
        # do something with each line of stdout
except SomeException as error:
        # If the return value is nonzero, and after iterating through all lines of stdout,
        # the iterator raises an appropriate exception instead of StopIteration.
        # Error probably contains the object returned by !(somecommand)
```

..just a thought.. :-)

adqm:
I agree that this would be a cool feature; thanks @eode!  I've marked this with the `newcomer` tag because I think it might be a nice place for someone to jump into the code.

The `!()` operator creates an instance of `xonsh.proc.CompletedCommand`, which contains (among other things) the string representing the process's standard output.  I think an initial solution would just involve adding an `__iter__` method to that class, which loops over the lines in the string representing `stdout`.

eode:
Yep, makes sense to me.  I'll dig in with a fork.

eode:
As to the error situation -- what error would be appropriate?

eode:
I considered creating a subclass of CompletedCommand that's also an exception, but I dunno if that's a good idea.

adqm:
I think `subprocess.CalledProcessError` is a reasonable candidate, since that is what is normally used to indicate those kinds of errors.  

I wonder if that exception should only be raised if [`$RAISE_SUBPROC_ERROR`](http://xon.sh/envvars.html#raise-subproc-error) is `True`.  Though then again, if we have that, calling the subprocess would already have raised an error.  Are you talking about raising that exception specifically within `__iter__`?  What do you think the benefit of that would be, beyond what `$RAISE_SUBPROC_ERROR` does?

eode:
Well, the idea was mainly that you'd be able to process all of the input, regardless of whether or not the process returned an error -- so, for example, if you had something that returned *some* valid input that you *do* want, but borked on an error, this would allow you to process both appropriately.

adqm:
Oh, I see; raising the exception _after_ the last line of stdout.  Gotcha; that's a great idea.  Sorry I misread!

eode:
[...]

PR #1183, opened 2016-06-09
Implementation of #1175, "!(command) could be usefully iterable"

eode:
Implementation of #1175.

I had to add an additional member to _CCTuple, as the actually-executed command couldn't be reassembled by a user without knowledge of Xonsh internals.  I'm pretty sure this is ready to go.

Example code probably explains it best -- here's usage.

```python
# Setup
from subprocess import CalledProcessError

mkdir bleh
cd bleh
touch foo.txt bar.txt


# Simple usage example
for filename in !(ls -1 --color=no):
    print(filename)
# Output:
# foo.txt
# bar.txt


# Example with error handling
try:
    for filename in !(ls -1 --color=no bar.txt foo.txt baz.txt bing.txt):
        print(filename)
except CalledProcessError as error:
        print("Error {} while executing {}".format(error.returncode, error.args[1]))
        print("stderr output:\n" + error.completed_command.stderr)
# Output:
#bar.txt
#foo.txt
#Error 2 while executing ['/bin/ls', '--color', '-1', '--color=no', 'bar.txt', 'foo.txt', 'baz.txt', 'bing.txt']
#stderr output:
#/bin/ls: cannot access 'baz.txt': No such file or directory
#/bin/ls: cannot access 'bing.txt': No such file or directory
```

### DIFF
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -553,25 +553,24 @@ class CompletedCommand(_CCTuple):
         return self.returncode == 0
 
     def __iter__(self):
-        stdout = self.stdout
-        start = 0
-        end = 0
-        while end != -1:
-            end = stdout.find('\n', start)
-            if end == -1:                # no newlines, but possibly more text
-                snippet = stdout[start:]
-            elif stdout[end-1] == '\r':  # newline, check for CR
-                snippet = stdout[start:end-1]
-            else:                        # newline, no CR
-                snippet = stdout[start:end]
-            if snippet or not (end == -1):
-                yield snippet
-            start = end + 1    # to the other side of \n
+        if not self.stdout:
+            raise StopIteration()
+
+        pre = self.stdout
+        post = None
+
+        while post != '':
+            pre, sep, post = pre.partition('\n')
+            # this line may be optional since we use universal newlines.
+            pre = pre[:-1] if pre and pre[-1] == '\r' else pre
+            yield pre
+            pre = post
+
         if self.returncode:
             # I included self, as providing access to stderr and other details
             # useful when instance isn't assigned to a variable in the shell.
             raise XonshCalledProcessError(self.returncode, self.executed_cmd,
-                                          stdout, self.stderr, self)
+                                          self.stdout, self.stderr, self)
 
     @property
     def inp(self):
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -541,7 +541,8 @@ _CCTuple = namedtuple("_CCTuple", ["stdin",
                                    "stdin_redirect",
                                    "stdout_redirect",
                                    "stderr_redirect",
-                                   "timestamp"])
+                                   "timestamp",
+                                   "executed_cmd"])
 
 class CompletedCommand(_CCTuple):
     """Represents a completed subprocess-mode command."""
@@ -564,15 +565,11 @@ class CompletedCommand(_CCTuple):
             if snippet or not (end == -1):
                 yield snippet
             start = end + 1    # to the other side of \n
-        # No \n was found.  ..if a \r is present, it's without \n -- so it
-        # doesn't count as a line ending. ..return all text on the tail.
-        #yield repr(stdout[start:])
         if self.returncode:
-            error = CalledProcessError(
-                self.returncode, 
-                self.args,
-                stdout,
-                self)
+            # I included self, as providing access to stderr and other details
+            # useful when instance isn't assigned to a variable in the shell.
+            cmd = self.executed_cmd
+            error = CalledProcessError(self.returncode, cmd, stdout, self)
             error.completed_command = self
             raise error
             
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -550,18 +550,28 @@ class CompletedCommand(_CCTuple):
         return self.returncode == 0
 
     def __iter__(self):
+        stdout = self.stdout
         start = 0;
-        end = self.stdout.find('\n')
+        end = 0
         while end != -1:
-            yield self.stdout[start:end + 1]
-            start = end + 1
-            end = self.stdout.find('\n', start)
-        yield self.stdout[start:end]
-        if self.rtn:
+            end = stdout.find('\n', start)
+            if end == -1:                # no newlines, but possibly more text
+                snippet = stdout[start:]
+            elif stdout[end-1] == '\r':  # newline, check for CR
+                snippet = stdout[start:end-1]
+            else:                        # newline, no CR
+                snippet = stdout[start:end]
+            if snippet or not (end == -1):
+                yield snippet
+            start = end + 1    # to the other side of \n
+        # No \n was found.  ..if a \r is present, it's without \n -- so it
+        # doesn't count as a line ending. ..return all text on the tail.
+        #yield repr(stdout[start:])
+        if self.returncode:
             error = CalledProcessError(
                 self.returncode, 
-                self.alias, 
-                self.out, 
+                self.args,
+                stdout,
                 self)
             error.completed_command = self
             raise error
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -19,7 +19,7 @@ from subprocess import (Popen, PIPE, DEVNULL, STDOUT, TimeoutExpired,
                         CalledProcessError)
 
 from xonsh.tools import (redirect_stdout, redirect_stderr, ON_WINDOWS, ON_LINUX,
-                         fallback, print_exception)
+                         fallback, print_exception, XonshCalledProcessError)
 
 if ON_LINUX:
     from xonsh.teepty import TeePTY
@@ -545,6 +545,7 @@ _CCTuple = namedtuple("_CCTuple", ["stdin",
                                    "timestamp",
                                    "executed_cmd"])
 
+
 class CompletedCommand(_CCTuple):
     """Represents a completed subprocess-mode command."""
 
@@ -553,7 +554,7 @@ class CompletedCommand(_CCTuple):
 
     def __iter__(self):
         stdout = self.stdout
-        start = 0;
+        start = 0
         end = 0
         while end != -1:
             end = stdout.find('\n', start)
@@ -569,10 +570,8 @@ class CompletedCommand(_CCTuple):
         if self.returncode:
             # I included self, as providing access to stderr and other details
             # useful when instance isn't assigned to a variable in the shell.
-            cmd = self.executed_cmd
-            error = CalledProcessError(self.returncode, cmd, stdout, self)
-            error.completed_command = self
-            raise error
+            raise XonshCalledProcessError(self.returncode, self.executed_cmd,
+                                          stdout, self.stderr, self)
 
     @property
     def inp(self):
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -15,7 +15,7 @@ import builtins
 from functools import wraps
 from threading import Thread
 from collections import Sequence, namedtuple
-from subprocess import Popen, PIPE, DEVNULL, STDOUT, TimeoutExpired
+from subprocess import Popen, PIPE, DEVNULL, STDOUT, TimeoutExpired, CalledProcessError
 
 from xonsh.tools import (redirect_stdout, redirect_stderr, ON_WINDOWS, ON_LINUX,
                          fallback, print_exception)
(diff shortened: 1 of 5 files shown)


xonsh/proc.py
```diff
@@ -15,7 +15,7 @@
 from functools import wraps
 from threading import Thread
 from collections import Sequence, namedtuple
-from subprocess import Popen, PIPE, DEVNULL, STDOUT, TimeoutExpired
+from subprocess import Popen, PIPE, DEVNULL, STDOUT, TimeoutExpired, CalledProcessError
```
adqm: It looks like this change pushed this over the PEP8 limit of 80 characters per line.  This can be refactored by using `from subprocess import (...)` and making the "tuple" span multiple lines.

xonsh/proc.py
```diff
@@ -541,14 +541,38 @@ def _wcode_to_popen(code):
...
 
 class CompletedCommand(_CCTuple):
     """Represents a completed subprocess-mode command."""
 
     def __bool__(self):
         return self.returncode == 0
 
+    def __iter__(self):
+        stdout = self.stdout
+        start = 0;
```
adqm: PEP8: no semicolon

xonsh/proc.py
```diff
@@ -541,14 +541,38 @@ def _wcode_to_popen(code):
...
+                yield snippet
+            start = end + 1    # to the other side of \n
+        if self.returncode:
+            # I included self, as providing access to stderr and other details
+            # useful when instance isn't assigned to a variable in the shell.
+            cmd = self.executed_cmd
+            error = CalledProcessError(self.returncode, cmd, stdout, self)
+            error.completed_command = self
+            raise error
+
```
adqm: PEP8: whitespace in blank line

adqm:
This is fantastic, @eode; thanks for putting it in!

A couple of small style-related comments inline.

This could also use an entry in CHANGELOG.rst, under "current developments."

Also also, on Python 3.4, I get the following error:
```
hartz@virtual1 ~ $ for i in !(ls fdsjklfadsfds):
..................     print(i)
..................     
.................. 
  File "/usr/local/lib/python3.4/dist-packages/xonsh/base_shell.py", line 154, in default
  File "/usr/local/lib/python3.4/dist-packages/xonsh/codecache.py", line 57, in run_compiled_code
    func(code, glb, loc)
  File "/usr/local/lib/python3.4/dist-packages/xonsh/proc.py", line 572, in __iter__
    error = CalledProcessError(self.returncode, cmd, stdout, self)
TypeError: __init__() takes from 3 to 4 positional arguments but 5 were given
```

It looks like `CalledProcessError` was changed between Python 3.4 and Python 3.5 to accept two new inputs.  The easiest workaround (though maybe not the most informative) would be only to give the first two arguments (the return code and the command).  That way, it would be accepted by both the [3.4](https://docs.python.org/3.4/library/subprocess.html#subprocess.CalledProcessError) and [3.5](https://docs.python.org/3.5/library/subprocess.html#subprocess.CalledProcessError) versions of `CalledProcessError`.  They both take `output` as well, so we could give it that, but it's not crucial, as that's an optional argument.

eode:
Hrm, I see..  I was thinking that CalledProcessError handled extra args generically like most of the builtins, but I guess not.   

Think it's reasonable to subclass it?  ..then anyone catching a CalledProcessError would succeed, but they'd have all the info, because they'd actually receive a XonchCalledProcessError (or some such).  It seems like a decent way to remain fairly consistent with the existing code.

If not, I could also just drop it as a parameter, and leave the post-instantiation assignment (line 571) which right now just serves as a shortcut.

Also -- should I PEP8 the whole file, or just my new lines?  ..it's no bother..

adqm:
> Think it's reasonable to subclass it? ..then anyone catching a CalledProcessError would succeed, but they'd have all the info, because they'd actually receive a XonchCalledProcessError (or some such). It seems like a decent way to remain fairly consistent with the existing code.

That seems reasonable to me, I think.  Maybe it would be good to have that also inherit from `XonshError` (from `xonsh.tools`).

I think we may also want to add `XonshError` and that new subclass to the builtins (by adding them to the [`load_builtins`](https://github.com/xonsh/xonsh/blob/master/xonsh/built_ins.py#L628) and [`unload_builtins`](https://github.com/xonsh/xonsh/blob/master/xonsh/built_ins.py#L682) functions) so that they are easier to access from within xonsh (no need to import `CalledProcessError`), including within scripts.

> Also -- should I PEP8 the whole file, or just my new lines? ..it's no bother..

I think we should probably do a big PEP8 sweep of the whole codebase at some point.  But for the sake of making it easy to look through the diff as it relates to this PR, I think it would be good if the rest of the PEP8 changes came in a separate PR.

eode:
Check.  I'll push my next changeset once the new Subclass is done, and the changelog is updated.  ..at some point, we should probably update the tutorial, too, I suppose.  Is that done at the end of each cycle, or as new features are added?

eode:
I guess I'll put the exception in tools.py, unless you have  a preference on where it should be.

adqm:
>  Is that done at the end of each cycle, or as new features are added?

We typically just update the tutorial as we go (that way, making the new releases is really easy; just push the docs to the live copy on the web).  So if you're feeling up to it, that would be great!

> I guess I'll put the exception in tools.py, unless you have a preference on where it should be.

That sounds reasonable to me.  That way we can use it other places in xonsh if we want to.

Thanks again for your work on this!

eode:
No prob.  :-)  I've been looking for a decent python system shell for a long time.

xonsh/proc.py
```diff
@@ -541,14 +542,38 @@ def _wcode_to_popen(code):
...
+        end = 0
+        while end != -1:
+            end = stdout.find('\n', start)
+            if end == -1:                # no newlines, but possibly more text
+                snippet = stdout[start:]
+            elif stdout[end-1] == '\r':  # newline, check for CR
+                snippet = stdout[start:end-1]
+            else:                        # newline, no CR
+                snippet = stdout[start:end]
+            if snippet or not (end == -1):
```
scopatz: I think that you could simplify this implementation a lot by using str.partition():

```python
pre = self.stdout
post = None
while post != '':
    pre, sep, post = stdout.partition('\n')
    pre = pre.rstrip('\r')  # this line may be optional since we use universal newlines.
    yield pre
    pre = post
```

scopatz:
Thanks a million for putting this in @eode! This is an awesome idea!

xonsh/proc.py
adqm: Ah, that's quite nice.  Also retains the ability to check whether `\n` occurs at all, which is an important piece here.

How did I not know about this function (`parition`) until just now?!

scopatz: Yeah, partition is great. One of the most under used str methods.  It really should be used everywhere, so until then is underused :)

adqm:
@eode, I get a little bit of weird behavior now:
```
hartz@virtual1 ~ $ for i in !(ls nonexistent):
..................     print(8)
..................     
.................. 
2
```

No traceback or indication that an error occurred, just a 2 (presumably, from the returncode of that process).  Any ideas what would be causing that?

Also, since we've been merging some substantial changes into master, there are some conflicts.  I'm happy to take care of those at merge time if you want to focus on this branch rather than on the merge.

eode:
It occurred in 397e6cb40c.  I'm looking into what changed there.  I've noticed the behavior myself, and have been puzzling about it for a bit.

eode:
..somehow, inheriting from XonshError is causing it, at the very least indirectly.

eode:
Ok, I found it, and I think i can work around it.  Looks like rather than being printed with print_exception(), since it's a XonshError, it's being caught in a different spot, and only the error code is printed -- in a XonshError,  args[0] is a message.  ..in XonshCalledProcessError, it's an error code.

I think the simplest course is to not make XonshCalledProcessError a subclass of XonshError.

eode:
Ready for you to check / potentially merge.

eode:
scratch that, need to fix the merge.   The area didn't even appear in the merge tool, but changes on both sides were removed (adding references for errors in builtins).

xonsh/proc.py
eode: ..a few things, if I might:

This loops indefinitely -- stdout.partition should be be pre.partition, yes?

And, as you said, this probably doesn't matter, but -- rstrip is greedy.  If someone had:
"\r\r\r\n", the string yielded should, in theory, be "\r\r".  ..that's why I did the check for and removal of just one char.  I know it's finicky beyond relevance, as it's not likely anyone needs to retain extra \r's.  It just bugs a part of my brain.  OK if I satisfy it?

Lastly, and more likely to be a usefully valid point -- the code I had doesn't yield anything if there's no text (including no \n's).  To me, that seems the right way to do it, as it distinguishes between no output and a single newline.  I'm looking at, as the result:

```python
pre = self.stdout
post = None

if not pre:
    raise StopIteration()

while post != '':
    pre, sep, post = pre.partition('\n')
    pre = pre[:-1] if pre[-1] == '\r' else pre    # this line may be optional since we use universal newlines.
    yield pre
    pre = post
```

..you're right, partition is way more straightforward than what I had.  I'll have to tuck that away and use it.  :-)

xonsh/tools.py
```diff
@@ -1191,8 +1218,7 @@ def backup_file(fname):
     import shutil
     from datetime import datetime
     base, ext = os.path.splitext(fname)
-    timestamp = datetime.now().strftime('%Y-%m-%d-%H-%M-%S-%f')
-    newfname = '%s.%s%s' % (base, timestamp, ext)
```
eode: The above two should be *retained* not lost, right?

eode:
..hokay, fixed, modified code based on @scopatz' nice usage of partition, a couple other minor things fixed.

xonsh/proc.py
```diff
@@ -541,14 +542,36 @@ def _wcode_to_popen(code):
...
+            # this line may be optional since we use universal newlines.
+            pre = pre[:-1] if pre and pre[-1] == '\r' else pre
+            yield pre
+            pre = post
+
+        if self.returncode:
+            # I included self, as providing access to stderr and other details
+            # useful when instance isn't assigned to a variable in the shell.
+            raise XonshCalledProcessError(self.returncode, self.executed_cmd,
+                                          self.stdout, self.stderr, self)
```
scopatz: Also, based on some of the gitter/irc chat I think that the raising part of this should be moved to another method, something like:

```python
def itercheck(self):
    yield from self
    if self.returncode:
        rasise ...
```

eode:
..can I know the reasoning behind that?

adqm:
I wasn't around for that conversation, but I could guess:  in the current setup, if a process has a nonzero return code and `$RAISE_SUBPROC_ERROR` is not set, we just get the partial output in the `stdout` field, and no complaints (no exception raised).

This raising an exception could be seen as an inconsistency with that mode of operation.  I'm not sure I agree.  But if there is a really strong feeling amongst the other devs, maybe we should go with that...

I would say that the thing that feels most right to me would probably be to raise the exception only if `$RAISE_SUBPROC_ERROR` is true, but that can't work given how things are currently set up (the exception would be raised before we could enter this iterator).

eode:
Ah, ok, you're concerned about exceptions being raised when $RAISE_SUBPROC_ERROR isn't on -- and if it is on, then the CompletedCommand doesn't even exist, and an exception was raised prior.

Perhaps this is all just a bad idea, and I'm expecting too succinct of wording from the shell, which isn't really a net gain, anyways.  Perhaps we should just encourage use of splitlines:

```python
# not this (six lines)
try:
    for line in !(somecommand):
        # do things here
except CalledProcessError as error:
    for line in error.stdout.splitlines():
        # handle error lines here

# but this (six lines):
cmd = !(somecommand)
for line in cmd.stdout.splitlines():
    # do things here
if not cmd:
    for line in cmd.stderr.splitlines():
        # handle error lines here
```

eode:
Yeah, it doesn't seem to me like an inconsistency -- The command succeeded (at least, it completed).  You're now just iterating over the results.  I understand the desire to have a clean iteration, though, as some code, you just don't care what the return value is, you just want the output.  I'm starting to doubt the usefulness of this overall, though.

eode:
Hrm.  Well, I have no opinion about this anymore.  I've made the requested change (moving the 'raise' to a separate itercheck() method), and if someone feels like merging it, feel free.  If it's not merged after a while, I'll just close it.

adqm:
I think it's still useful!  I'll take one last detailed look tonight.

scopatz:
Hi @eode! I think this looks great.  I think the main issue as proposed originally was that it mixed concerns of checking if the process ran cleanly with iteration.  As you said, clean iteration is probably more expected. I also think that this implemention would also allow a `checkiter()` method, where the code isn't iterated at all if the process failed.  People might want both :).

I am very happy with this as it is.  I'll let @adqm merge it, but it seems like this might have merge conflicts from master. If you could clean those up, that would be awesome! Thanks again, this is a wonderful addition.

adqm:
Just merged in (with some small changes to the tutorial to reflect the addition of `itercheck`).  Thanks again @eode!

eode:
Good call on updating the tutorial, and thanks for handling the new conflicts.  I'm really looking forward to the next time I need to write a sizeable script -- great work with Xonsh in general, guys.

PR #1185, opened 2016-06-09
Fix crash caused by malformed prompt or title

mitnk:
Following @adqm 's fixing prompt in #1150 , this PR try to fix bad `$TITLE` and `$PROMPT` from callables.

The following prompt in xonshrc was causing xonsh crash:

```
def render_prompt():
    return '{HELLO'
$PROMPT = render_prompt
```

@adqm @scopatz please help to review it, thanks.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1136,6 +1136,13 @@ def _get_fmtter(formatter_dict=None):
 
 
 def format_prompt(template=DEFAULT_PROMPT, formatter_dict=None):
+    try:
+        return _format_prompt_main(template, formatter_dict)
+    except:
+        return template() if callable(template) else template
+
+
+def _format_prompt_main(template, formatter_dict):
     """Formats a xonsh prompt template string."""
     template = template() if callable(template) else template
     fmtter = _get_fmtter(formatter_dict)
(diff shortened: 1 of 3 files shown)


xonsh/environ.py
```diff
@@ -1135,7 +1135,26 @@ def _get_fmtter(formatter_dict=None):
...
 
+def _failover_template_format(template):
+    if callable(template):
+        try:
+            # Exceptions raises from function of producing $PROMPT
+            # in user's xonshrc should not crash xonsh
+            return template()
+        except:
+            print_exception()
+            return '$ '
```
mitnk: any suggestions here?

xonsh/environ.py
```diff
@@ -1135,7 +1135,26 @@ def _get_fmtter(formatter_dict=None):
     return fmtter
 
 
+def _failover_template_format(template):
+    if callable(template):
+        try:
+            # Exceptions raises from function of producing $PROMPT
+            # in user's xonshrc should not crash xonsh
+            return template()
+        except:
```
scopatz: This should be `except Exception:` so it doesn't catch keyboard interrupts

xonsh/environ.py
scopatz: This seems fine to me!

scopatz:
This looks good to me! Thanks for putting this in @mitnk! Just a couple of small issue.  Also thanks a ton for putting in the tests!

mitnk:
updated except, thanks

scopatz:
Thanks!

PR #1186, opened 2016-06-09
Proposal: More Ways to Search

adqm:
There was some call in the Gitter/IRC for the ability to do something like the regex globs, but for normal globs, so...

This changeset expands upon the "regex globbing" syntax (<code>``</code>), adding support for additional types of searches with backticks.

The behavior of <code>``</code> is unchanged, but this patch includes the ability to use:
* <code>r``</code> for regex globbing
*  <code>regex``</code> for regex globbing
* <code>g``</code> for normal globbing
* <code>glob``</code> for normal globbing

Just in case those aren't enough, this also allows users to specify their own search functions, and what prefixes should be associated with them.  A dumb example:
```
hartz@virtual1 ~ $ $PATH_SEARCH_FUNCS
{'': <function xonsh.built_ins.regexsearch>,
 'g': <function xonsh.tools.globpath>,
 'glob': <function xonsh.tools.globpath>,
 'r': <function xonsh.built_ins.regexsearch>,
 'regex': <function xonsh.built_ins.regexsearch>}
hartz@virtual1 ~ $ $PATH_SEARCH_FUNCS['b'] = lambda s: ['hey', 'there']
hartz@virtual1 ~ $ 
hartz@virtual1 ~ $ b`test`
['hey', 'there']
```

When trying to use a nonexistent search function, an error is generated, both in Python mode and in subproc mode:
```
hartz@virtual1 ~ $ for i in notdefined`test`:
..................     print(7)
..................     
.................. 
'notdefined' is not a known path search function
hartz@virtual1 ~ $ 
hartz@virtual1 ~ $ ls notdefined`test`
'notdefined' is not a known path search function
```


melund:
This looks cool, and lot of users have requested a glob in python. 

I not too crazy about short version of `glob` and `regex` keywords. 
```
r`......`    and g`.....`
```
They look to me very much like the different string types in python `r'...'`, `f'...'`, `b'....'`. And I think that `glob` and `regex` are more readable anyway. 

Quoting PEP20: _"There should be one-- and preferably only one --obvious way to do it."_ and _"Readability counts."_

Anyway. With this setup users can always add the shortcuts them self to `$PATH_SEARCH_FUNCS`

adqm:
> They look to me very much like the different string types in python r'...', f'...', b'....'. And I think that glob and regex are more readable anyway. 

Interestingly, that had been my motivation (I saw it as a plus that it mirrored Python's different types of strings)...

That said, I'm not wedded to that (though I would definitely be one to add those right back in via `$PATH_SEARCH_FUNCS`).

Happy to continue the conversation.

adqm:
Also, just merged from master to avoid the conflict.

melund:
Ok. I am not picky about this so let us just hear what @scopatz thinks. He weighs many of these things by how easy it is to teach to people.

scopatz:
> Ok. I am not picky about this so let us just hear what @scopatz thinks. He weighs many of these things by how easy it is to teach to people.

Oh snap! I am so predictable :).  I actually have an inner Greg Wilson who yells at me for adding syntax that make it harder to learn, intuit, or reason about.  

I am going to sleep on this, but my initial imporession is that,

1. I like ```r`...```` and ```g`...```` - I like that it mirrors Python strs
2. I agree with @melund that there should only be one way of doing it, but prefer the short version
3. I don't think the ability for run-time level look up of the matching function is...wait for it... easy to teach/learn.  It also opens the door to a bunch of weird cases where the function that is called can deregister itself, and so the same ```x`...```` syntax could be run twice in a row with dramatically different results.  Also how would you *know* that ```g`...```` was a glob unless you checked beforehand each time? It seems like this has the potential to reduce the portability and usefulness of this syntax, rather than increase it. I guess the counter argument to this is that you can go in and swap out builtins like `__xonsh_glob__()`, but at least that is 'private'. If a user messes with that, they are expected to know what they are doing.

Also for the long names and the lookups, I don't see a huge advantage to users importing and calling their own functions. And simple transformations could be handled by loops.

All of that being said, I think that this is a really neat and useful idea.  I think the r and g forms are solid and would be great to have.  It is these other aspects that are more worrisome to me.

If the concern is that r and g are not extensible, I think that we can cross that bridge when we get to it.  For example, if we wanted support for an alternative regex library, we could add another prefix for that. If you have specific use cases, in mind, I 'd love to hear them!  My sense is that the number of use cases that are not supported by the following are pretty small.

```python
import mymatcher
mymatcher(g`*`)
```

On *now* to go sleep on it :)

adqm:
My thoughts were:

* I prefer the short syntax (`r` and `g`), but if someone is writing a script, they may want something more explicit (hence `regex` and `glob`).  We've done similar things in `CompletedCommand`, where there exist both a long and a short name for referring to things.  Having the short names is a nice convenience for those who remember them, but for those who don't (_what did "g" mean, again?_ which gets worse if we add more), I think it's nice to have the full names.
* We don't _know_ what kinds of things people might want to search for (hence the desire for extensibility).  For example, someone might want to hook into the tab completer (or the fuzzy/sequential lookups in #1167, or an external command, or their own weird way or searching, or...).  I guess I feel the same way about this that I did about tab completers, and for largely the same reason.  There is still the question of which ones we want _baked in_, but I still think there's a benefit to extensibility.  Those who don't want to mess with it still have a clean interface, but for those who _do_ want to, they have that available to them as well.  And yes, these could be done via a Python function, but I think it's the same story as globbing; it's really nice to have a short-hand for that.  In theory, I think it could also lower the barrier to entry for a new contributor; if they've already extended their own shell with some cool new search feature, then it becomes easier to test, and to submit a patch.

melund:
Good idea to sleep on it. Adding new syntax should not be done lightly. I think there should only be one syntax, and I am OK with the short ones if you prefer them.
```
r`...` and g`...`
```

I agree with @scopatz that having a `$PATH_SEARCH_FUNCS` variable that changed syntax could be confusing. It reminds me too much of the `IFS` from bash. If we really needed this functionality, we could use the `@` in same way we considered doing for the `@func$(command)`, 
```
@func`....`
```
That would have a taste of decorator from python, and if we used the idea in other places it could a consistent pattern for modifying syntax.

scopatz:
So, I think  the consensus thus far is that `r` and `g` are good.  I think that could be put in pretty much as is.  These should be static parts of the tokenizer / lexer / parser

For the dynamic lookup matching, I actually do prefer the syntax @melund suggested.  I think that ```@foo`...` ``` is a great option.  It keeps that decorator-ish idea and lets people know that they are going into a Python lookup to call a function.  Also, I think that this syntax would be wholly compatible with both Python and subprocess mode.

adqm:
okay, that works for me.

* <code>``</code> unchanged
* <code>r``</code> explicit regex
* <code>g``</code> explicit globs
* <code>@foo``</code> look up a function and run it

scopatz:
Yep, that looks right to me!

adqm:
Alright @scopatz, I think this is ready for another look when you have a chance.  I made the updates you requested, and I modified the search function lookup so that it happens in the current scope rather than inside a special dictionary (not sure if that's a controversial change).

docs/tutorial.rst
```diff
@@ -866,15 +873,60 @@ Let's see a demonstration with some simple filenames:
...
+be used with backticks with the following syntax: ``@<name>`test```
+
+A silly example shows the form of these functions:
+
+.. code-block:: xonshcon
+
+    >>> def foo(s):
+    ...     return [s, 'snail', 'shell']
+    >>> @foo`test`
+    ['test', 'snail', 'shell']
```
scopatz: Maybe a more illustrative example would be a `startswith` or something?

adqm: Probably so, yes.  I can make that change.

scopatz:
> I made the updates you requested

Thanks!

> I modified the search function lookup so that it happens in the current scope rather than inside a special dictionary (not sure if that's a controversial change).

I think a special dictionary would be much better.  Maybe that could be shoved into builtins, like aliases too?  The local context seems simultaneously pretty restrictive (if users want to write libraries or xontribs that register such functions) and too broad (since most objects in the local context won't be usable as matchers).

docs/tutorial.rst
scopatz: Thanks!

adqm:
> I think a special dictionary would be much better.

I was going for the consistency argument again (i.e., everywhere else, `@` does a lookup in the current Python environment).  Although I can understand the special dictionary approach, it doesn't seem to fit with what `@` does everywhere else...

scopatz:
So I think of `@` as "go into Python mode", which is what this is doing.  Like what @melund said, I think this will jive even better with the subproc filters, a la `@foo@bar$(cmd args)`. I would think that these would also look up in a special dictionary for `foo` and `bar` transformation functions.  I guess that they could also look up in the local context.  

I think that it is critical to do the same basic thing for both ```@foo`...` ``` and `@foo$()`. And I think that registration makes more sense for both of these overall.

At least that is the way I have approached this stuff so far :)

adqm:
Okay, that works for me (though it still feels weird to me that `@` behaves differently in different situations).  I will change that back.

scopatz:
Ok great!  Maybe @melund should weigh in too?

melund:
I am a little confused to what you ended up agreeing upon. Did you mean that if xonsh sees:
```
@foo`...`
```
It should look for `foo` in a specialized dictionary instead of jus the local scope? Then I think I like the local scope idea better. What is the benefit of registering `foo` somewhere special?

adqm:
I am with @melund on this one.  That said, @scopatz has a point, too.  I don't think either solution is perfect.  To me, looking up from a special dictionary kind of screws with my mental model of `@` (since all other `@` invocations do the lookup in the normal scope.  That said, looking up in the regular scope could lead to namespace collisions (or at least namespace "pollution").

I could kind of go either way, though I think I still lean toward using the regular namespace.

scopatz:
well, the way it is implemented now, it only looks things up in the ctx. it should at least do a proper Python locals, globals, builtins look up if it is going to be like a normal Python variable. 

I guess what you are saying is that xontrib authors would not need to do anything special.

I suppose I am OK with that as long as it does a real lookup and we are consistent with the filter funcs. I guess my concern with filters is that we'd probably want to provide a bunch of stock str transformations but not want to pollute the usual namespace. maybe people can use str methods directly and maybe I am too worried about this

scopatz:
sorry for

scopatz:
sorry for the accidental close!

scopatz:
Ok, after more thinking about it, I think you have me convinced that this should do normal variable lookup. We can torpedo my wacky ideas about registration :)  I would be nice, though too if we could have any Python expression in between `@` and the first back tick.  I think this would handle the lookup issues automatically.

adqm:
I think the arbitrary Python expression is actually hard to get right from a parsing standpoint.  For now, I'll stick with just names, but I'll restructure things so that we do an actual lookup (probably move most of the logic into the parser rather than the special function in `built_ins`).

adqm:
But if I am inspired along the way by some nice way to handle arbitrary expressions, I'll work that in.

scopatz:
Sounds great to me @adqm! Having just names for now doesn't preclude us from adding more expr types in the future.

adqm:
Alright.  I implemented the changes we discussed and rebased onto master.  I think this is ready for another look.

scopatz:
Thanks a million @adqm! This is great to have!

Issue #1070, opened 2016-06-02
AttributeError: 'PromptToolkitShell' object has no attribute 'completer'


cryzed:
When attempting to execute the following xonsh script:
```
cryzed@arch ~ $ cat update-aur-gits.xsh 
#!/usr/bin/env xonsh


def main():
    packages = []
    for line in $(pacaur -Qsq git).splitlines():
        if not line.endswith('-git'):
            continue

        packages.append(line)

    ![pacaur -S @(' '.join(packages)) --noconfirm --noedit]



if __name__ == '__main__':
    main()
```

This happens:

```
cryzed@arch ~ $ xonsh update-aur-gits.xsh 
Traceback (most recent call last):
  File "/usr/bin/xonsh", line 9, in <module>
    load_entry_point('xonsh==0.3.2', 'console_scripts', 'xonsh')()
  File "/usr/lib/python3.5/site-packages/xonsh/main.py", line 228, in main
    args = premain(argv)
  File "/usr/lib/python3.5/site-packages/xonsh/main.py", line 215, in premain
    shell = builtins.__xonsh_shell__ = Shell(**shell_kwargs)
  File "/usr/lib/python3.5/site-packages/xonsh/shell.py", line 72, in __init__
    ctx=self.ctx, **kwargs)
  File "/usr/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 33, in __init__
    self.pt_completer = PromptToolkitCompleter(self.completer, self.ctx)
AttributeError: 'PromptToolkitShell' object has no attribute 'completer'
```

When changing `$SHELL_TYPE` to `readline` the scripts runs as expected.

```
cryzed@arch ~ $ pacman -Qi xonsh-git python-prompt_toolkit cryzed-python-pygments-hg | grep Version -B 1
Name            : xonsh-git
Version         : 0.3.2.r27.g5bb8c56-1
--
Name            : python-prompt_toolkit
Version         : 1.0.0-1
--
Name            : cryzed-python-pygments-hg
Version         : r4465.4459f971dd43-1
```

gforsyth:
Hey @cryzed, thanks for reporting!
I can't reproduce this on Arch.  Can you post the output of `xonfig`?

cryzed:
Of course!
```
cryzed@arch ~ $ xonfig
+------------------+-------+
| xonsh            | 0.3.2 |
| Python           | 3.5.1 |
| PLY              | 3.7   |
| have readline    | True  |
[...]

PR #1187, opened 2016-06-09
No completer in script

adqm:
This is a workaround (not a real fix) for #1070.  I still haven't been able to track down the actual issue, but this should serve as a Band-aid in the meantime.


melund:
LGTM

PR #1192, opened 2016-06-09
Cygwin Fix

adqm:
This should fix one of the issues discussed in #514, where xonsh would background itself after every command when running under Cygwin.

There is still one more issue outstanding (discussed in #514, which I have not been able to reproduce), but this should take us most of the way toward being completely functional in Cygwin with no hacks required!

### DIFF
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -87,18 +87,39 @@ else:
 
     _shell_pgrp = os.getpgrp()
 
-    _block_when_giving = (signal.SIGTTOU, signal.SIGTTIN, signal.SIGTSTP)
+    _block_when_giving = (signal.SIGTTOU, signal.SIGTTIN,
+                          signal.SIGTSTP, signal.SIGCHLD)
 
-    def _give_terminal_to(pgid):
-        # over-simplified version of:
-        #    give_terminal_to from bash 4.3 source, jobs.c, line 4030
-        # this will give the terminal to the process group pgid
-        if _shell_tty is not None and os.isatty(_shell_tty):
-            if not ON_CYGWIN:
+    if ON_CYGWIN:
+        import ctypes
+        _libc = ctypes.CDLL('cygwin1.dll')
+
+        def _give_terminal_to(pgid):
+            # over-simplified version of:
+            #    give_terminal_to from bash 4.3 source, jobs.c, line 4030
+            # this will give the terminal to the process group pgid
+            if _shell_tty is not None and os.isatty(_shell_tty):
+                omask = ctypes.c_ulong()
+                mask = ctypes.c_ulong()
+                _libc.sigemptyset(ctypes.byref(mask))
+                for i in _block_when_giving:
+                    _libc.sigaddset(ctypes.byref(mask), ctypes.c_int(i))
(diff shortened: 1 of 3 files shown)


xonsh/jobs.py
```diff
@@ -87,18 +87,39 @@ def _set_pgrp(info):
...
-            if not ON_CYGWIN:
+    _block_when_giving = (signal.SIGTTOU, signal.SIGTTIN,
+                          signal.SIGTSTP, signal.SIGCHLD)
+
+    # _give_terminal_to is a simplified version of:
+    #    give_terminal_to from bash 4.3 source, jobs.c, line 4030
+    # this will give the terminal to the process group pgid
+    if ON_CYGWIN:
+        import ctypes
+        _libc = ctypes.CDLL('cygwin1.dll')
```
scopatz: maybe call this _libcygwin, since libc is something else :)

scopatz:
Thanks a million @adqm!  This looks good to me other than the one name issue. Maybe @jivanyatra would be willing to try this out before we merge it?

jivanyatra:
Happy to! What's the best way for me to patch this?

scopatz:
It depends on how you installed it :)

xonsh/jobs.py
adqm: >  since libc is something else

Personally, I don't think it is.  I think this is _one implementation_ of libc.  From the libc man page:

> The term "libc" is commonly used as a shorthand for the "standard C library", a library of standard functions that can be used by all C programs (and sometimes by programs in other languages).

This just happens to be Cygwin's implementation of libc (as opposed to `glibc`, for example).  It has all the same stuff as if we had done `ctypes.CDLL('libc.so.6')` under GNU/Linux).  So I would argue that it's just as much libc as the one on GNU/Linux would be (in fact, one could imagine an implementation that used `ctypes` for both Cygwin and normal operation, but just had:

```
if ON_CYGWIN:
    _libc = ctypes.CDLL('cygwin1.dll')
elif ON_DARWIN:
    _libc = ctypes.CDLL('libSystem.B.dylib')
else:
    _libc = ctypes.CDLL('libc.so.6')
```

That said, if it speeds this one on its way, I'm happy to change the name.  Just let me know.

adqm:
@jivanyatra, the easiest is probably to make a fresh clone of this branch of the repo:
```
git clone -b sigmask https://github.com/adqm/xonsh newxonsh
```

and then to `python3 setup.py install` from within there.

adqm:
Also @scopatz, in https://github.com/xonsh/xonsh/issues/514#issuecomment-225069761, @ellocogato said it works for them.

xonsh/jobs.py
scopatz: *Touche!*  I don't know enough about cygwin to know whether this will always be the libc implementation that is used.  If so, it is fine as it is.  I don't consider this blocking :)  Just a suggestion

scopatz:
Ahh great! I'll go ahead and merge then.

scopatz:
Thanks bunches!

jivanyatra:
Confirmed, works for me as well. Thanks!

PR #1193, opened 2016-06-10
Attempted fix for PermissionError and ProcessLookupError on Cygwin

adqm:
This is the first attempt at a fix for the `ProcessLookupError` and `PermissionError` exceptions under Cygwin (discussed in #541).

@OllieTerrance, @ellocogato, @jivanyatra: would you mind testing this out?  As I mentioned, I am unable to reproduce the issue.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -46,6 +46,9 @@ Current Developments
 * Ensured that the prompt_toolkit shell functions, even without a ``completer``
   attribute.
 * Fixed crash resulting from malformed ``$PROMPT`` or ``$TITLE``.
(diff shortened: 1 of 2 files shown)


scopatz:
Code looks good to me, thanks for really pushing on cygwin @adqm!

OllieTerrance:
This looks good -- I can still get a `ProcessLookupError` on master, but fixed with this patch.  :+1:

adqm:
Okay, I think this is ready for another look, too, @scopatz!  Some comments in #514, and seems to be confirmed working.

scopatz:
Looks great thanks!

PR #1196, opened 2016-06-10
Proposal: Bare and Anonymous Aliases

adqm:
This grew out of [some conversation in the IRC room](https://gitter.im/xonsh/xonsh?at=5759ba4b1cf76dd64535a84e) about the nature of Python-function aliases.

This changeset allows for invoking functions of the proper form as aliases even if they have not been explicitly added to the `aliases` mapping.  It also allows for escaping such things with `@()`, which enables the use of _anonymous_ aliases.  Three examples:

```
hartz@virtual1 ~ $ def foo(a, s=None):
..................     return a[1]+'\n'
.................. 
hartz@virtual1 ~ $ foo cat dog ferret tomato
dog
hartz@virtual1 ~ $ @(lambda a, stdin=None: a[0]+'\n') laptop
laptop
hartz@virtual1 ~ $ echo "potato" | @(lambda a, s=None: 'mashed '+s.strip()+'es\n')
mashed potatoes
```

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -460,10 +460,12 @@ Python Evaluation with ``@()``
 ===============================
 
 The ``@(<expr>)`` operator form works in subprocess mode, and will evaluate
-arbitrary Python code. The result is appended to the subprocess command
-list. If the result is a string, it is appended to the argument list. If the
-result is a list or other non-string sequence, the contents are converted to
-strings and appended to the argument list in order. Otherwise, the result is
+arbitrary Python code. The result is appended to the subprocess command list.
+If the result is a string, it is appended to the argument list. If the result
+is a list or other non-string sequence, the contents are converted to strings
+and appended to the argument list in order. If the result is a function, it is
+treated as an alias (see the section on `Aliases`_ below), even if it was not
+explicitly added to the ``aliases`` mapping.  Otherwise, the result is
 automatically converted to a string. For example,
 
 .. code-block:: xonshcon
@@ -476,6 +478,8 @@ automatically converted to a string. For example,
     4
     >>> echo @([42, 'yo'])
     42 yo
+    >>> echo "potato" | @(lambda a, s=None: a[0]+' '+s.strip()+'es\n') mashed
+    mashed potatoes
 
 This syntax can be used inside of a captured or uncaptured subprocess, and can
 be used to generate any of the tokens in the subprocess command list.
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -463,10 +463,10 @@ The ``@(<expr>)`` operator form works in subprocess mode, and will evaluate
 arbitrary Python code. The result is appended to the subprocess command list.
 If the result is a string, it is appended to the argument list. If the result
 is a list or other non-string sequence, the contents are converted to strings
-and appended to the argument list in order. If the result is a function, it is
-treated as an alias (see the section on `Aliases`_ below), even if it was not
-explicitly added to the ``aliases`` mapping.  Otherwise, the result is
-automatically converted to a string. For example,
+and appended to the argument list in order. If the result in the first position
+is a function, it is treated as an alias (see the section on `Aliases`_ below),
+even if it was not explicitly added to the ``aliases`` mapping.  Otherwise, the
+result is automatically converted to a string. For example,
 
 .. code-block:: xonshcon
 
@@ -478,8 +478,8 @@ automatically converted to a string. For example,
     4
     >>> echo @([42, 'yo'])
     42 yo
-    >>> echo "potato" | @(lambda a, s=None: a[0]+' '+s.strip()+'es\n') mashed
-    mashed potatoes
+    >>> echo "hello" | @(lambda a, s=None: s.strip + " world")
+    hello world
 
 This syntax can be used inside of a captured or uncaptured subprocess, and can
 be used to generate any of the tokens in the subprocess command list.
@@ -992,9 +992,12 @@ If you were to run ``gco feature-fabulous`` with the above aliases in effect,
 the command would reduce to ``['git', 'checkout', 'feature-fabulous']`` before
 being executed.
 
+
+Callable Aliases
+----------------
 Lastly, if an alias value is a function (or other callable), then this
 function is called *instead* of going to a subprocess command. Such functions
-must have the following signature:
+must have one of the following two signatures
 
 .. code-block:: python
 
@@ -1033,6 +1036,28 @@ must have the following signature:
         # examples the return code would be 0/success.
         return (None, "I failed", 2)
 
+
+.. code-block:: python
+
+    def _mycmd2(args, stdin, stdout, stderr):
+        """args will be a list of strings representing the arguments to this
+        command.  stdin is a read-only file-like object, and stdout and stderr
+        are write-only file-like objects
+        """
+        # This form allows "streaming" data to stdout and stderr
+        import time
+        for i in range(5):
+            time.sleep(i)
+            print(i, file=stdout)
+
+        # In this form, the return value should be a single integer
+        # representing the "return code" of the alias (zero if successful,
+        # non-zero otherwise)
+        return 0
+
+
+Adding and Removing Aliases
+---------------------------
 We can dynamically alter the aliases present simply by modifying the
 built-in mapping.  Here is an example using a function value:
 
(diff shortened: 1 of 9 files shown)


scopatz:
Thanks for putting this in @adqm! My initial thought is that this potentially a good idea, but I think it needs a distinct syntax. `@(...)` should be for pure Python mode in my opinion.  I also am inclined to say that whatever this new syntax is should be used by the onymous functions too.

scopatz:
Or maybe it could be a different syntax.

scopatz:
Perhaps something like the following:

```console
hartz@virtual1 ~ $ @foo cat dog ferret tomato
dog
hartz@virtual1 ~ $ @[lambda a, stdin=None: a[0]+'\n'] laptop
laptop
hartz@virtual1 ~ $ echo "potato" | @[lambda a, s=None: 'mashed '+s.strip()+'es\n']
mashed potatoes
```

Also, I find it weird that the arg offset is 0 in the laptop example, should't that be 1?

scopatz:
Or maybe:

```
hartz@virtual1 ~ $ @lambda(a, stdin=None: a[0]+'\n') laptop
laptop
hartz@virtual1 ~ $ echo "potato" | @lambda(a, s=None: 'mashed '+s.strip()+'es\n')
mashed potatoes
```

adqm:
> @(...) should be for pure Python mode in my opinion

I think it still is.  In my mind, this is _identical_ to `@()` as it currently exists: evaluate the contents inside `@()` in Python mode, and then use the result in the subprocess arguments.  So I think they are _actually_ the same operation.  This could also be used to escape the named functions, then, with no problem.  Plus, less syntax to learn :)

That said, if we want a different syntax (i.e., if we want to draw a distinction between Python _functions_ to use as aliases and Python _strings_ to use as arguments), I would vote for making it a drastically different syntax so there is less chance to confuse the two.  One that was mentioned in the chat as a possibility was `%()`.

Personally, I would vote for using `@()` because I think that what this is doing (at least as written) is exactly the definition of `@()`.

> Also, I find it weird that the arg offset is 0 in the laptop example, should't that be 1?

Nope.  The arguments that are passed to an alias don't include the alias itself.  So `laptop` is the first element in the `args` list that goes to that function.

scopatz:
> Nope. The arguments that are passed to an alias don't include the alias itself. So laptop is the first element in the args list that goes to that function.

Oh! Sorry I misread the foo example

scopatz:
Ok, I think I am convinced about calling callables via `@(...)` when they are in the first place "command" position :).  That position is special, and so even other pieces of subproc syntax may act differently in that location.  And, really, what else is going to be done with a callable in that location/   

However, I don't think that we need the bare execution, then. `@(foo)` is basically the same exact thing.  I would be OK with shortening this to `@foo` because the parentheses are extraneous since there is no possibility for whitespace in an NAME.  Also this provides a suitably distinct way of doing things from aliases.  If someone really wants to get rid of the @, they can stick the callable in aliases.

Does this sound reasonable?

adqm:
Yeah, I can get behind that!  Even though it's a few more characters to type, that's much more consistent with everything else we do (having to wrap in `@()` to access the Pythonic things).

scopatz:
Awsome! yay for consistency!

adqm:
Alright @scopatz, this is ready for another look.  I removed the "bare aliases" concept (now they can only be called by escaping with `@()`).

xonsh/built_ins.py
```diff
@@ -401,13 +401,19 @@ def run_subproc(cmds, captured=False):
         elif builtins.__xonsh_stderr_uncaptured__ is not None:
             stderr = builtins.__xonsh_stderr_uncaptured__
         uninew = (ix == last_cmd) and (not _capture_streams)
-        alias = builtins.aliases.get(cmd[0], None)
+
+        if callable(cmd[0]):
+            alias = cmd[0]
+        else:
+            alias = builtins.aliases.get(cmd[0], None)
+            n = locate_binary(cmd[0])
```
scopatz: Maybe there is a variable name that is more descriptive than `n` here.  This function is already a little out of hand :)

xonsh/built_ins.py
```diff
@@ -657,6 +673,7 @@ def load_builtins(execer=None, config=None, login=False, ctx=None):
     builtins.__xonsh_commands_cache__ = CommandsCache()
     builtins.__xonsh_all_jobs__ = {}
     builtins.__xonsh_ensure_list_of_strs__ = ensure_list_of_strs
+    builtins.__xonsh_list_of_strs_or_callables__ = list_of_strs_or_callables
```
scopatz: Do we even need `ensure_list_of_strs()` anymore?

xonsh/built_ins.py
```diff
@@ -626,6 +631,17 @@ def ensure_list_of_strs(x):
     return rtn
 
 
+def list_of_strs_or_callables(x):
```
scopatz: Some tests for this would be nice.

scopatz:
Hi @adqm, thanks!  I think that the implementation here looks great other than the `n` thing. 

Also, I would like some more tests for this.  Maybe a parser test for the lambda case, and a couple of execer tests for lambdas and functions.

Also, this defintely needs a mention in the tutorial somewhere, since the semantics of this are unique.

adqm:
okay, i can probably get those in tonight.

scopatz:
Amazing!

adqm:
Alright, this should be ready for another look.  I replaced `n` with a different name and added a couple of tests.  I think we have to keep `__ensure_list_of_strs__` because it is used elsewhere.

docs/tutorial.rst
```diff
@@ -476,6 +478,8 @@ automatically converted to a string. For example,
     4
     >>> echo @([42, 'yo'])
     42 yo
+    >>> echo "potato" | @(lambda a, s=None: a[0]+' '+s.strip()+'es\n') mashed
```
scopatz: Maybe also add a simple, direct example, like,

```
>>> @(lambda a, stdin=None: a[0]+'s\n') laptop
laptops
```

scopatz:
Hi @adqm, thanks for this!  I think we do need some parser and execer tests too before I would feel comfortable merging this. Characteristics of the parser ans AST transformer seem particularly easily to lose to me without knowing it.

adqm:
Ah, yes.  That is a good call.  I should probably also add a note to the "Aliases" section of the tutorial, as well, as a reminder that `@()` can be used to escape aliases not explicitly in the mapping.  I'll make those changes.

adqm:
@scopatz, now that i think about it, though, the parser and execer are unchanged here (all that changed was `ensure_list_or_strs` -> `list_of_strs_or_callables` and a little bit of logic in `run_subproc`.  I'm not sure where it would be best to put these tests (or what exactly it is that you would like tested.

I'll go ahead and update the tutorial entry, and I'll convert the relevant changelog entries to the new format, but I'll wait to hear back from you about the tests.

scopatz:
@adqm - well the parser isn't entirely unchanged. I guess I would like a parser test for the  `@(lambda , s=None: "x")` case.  

I would also like to see a couple of execer tests for the same lambda case one for the above lambda case and one for a lookup case that may look like:

```
code = (
    'def foo(a,  s=None):\n'
    '    return "bar"\n'
    '@(foo)\n')
yield assert_...
```

Does this make sense?

adqm:
The grammar of the language is unchanged, though.  What you mention has always been valid syntax anyway (it was just interpreted differently), right?

The execer tests I understand.  I can add those.

scopatz:
> The grammar of the language is unchanged, though. What you mention has always been valid syntax anyway (it was just interpreted differently), right? I can still add the test since we aren't yet verifying that's the case.

Yeah, that is right.  The grammar is unchanged.  My concern was that the parser tests on their own should cover all the expected pathways. I would rather us be over covered, than undercovered :).  So I would like the tests to be a little future proof.  This line is currently covered by other tests for py mode, but  we could potentially change the implementation, leaving this feature uncovered by the current tests.  

Execer tests, of course, would also pick up such an error if it were to occur. But a parser test and an execer test would help us narrow down where the problem stemmed from.

That was my thinking.

adqm:
Alright @scopatz, I added in some tests, and I tried to make the tutorial a little more readable (also added a small section about the "other form" of callable aliases [with 4 arguments]).

scopatz:
Looks great! Thanks a million @adqm!

scopatz:
Just waiting on CI.

scopatz:
This is an awesome feature, BTW

Russell-Jones-OxPhys:
(can't do it just now, but have been meaning to file a report about the [@ matrix operator](https://docs.python.org/3/whatsnew/3.5.html#whatsnew-pep-465) in 3.5+)

adqm:
> But scanning through there's another issue, with it not returning errors in older Python versions (e.g. 1@1 on 3.4.3)

Yes, that does seem possible.  I think the tokenizer is always using Python 3.5 syntax, though I would have thought that would be caught as a syntax error via tha parser (on py3.4).

Issue #1200, opened 2016-06-10
pip install error on Windows


klonuo:
Trying to update xonsh raised this error:

```shell
>pip install -U xonsh
Collecting xonsh
  Using cached xonsh-0.3.4.tar.gz
Installing collected packages: xonsh
  Found existing installation: xonsh 0.3.2
    Uninstalling xonsh-0.3.2:
      Successfully uninstalled xonsh-0.3.2
  Running setup.py install for xonsh ... error
    Complete output from command c:\python35\python.exe -u -c "import setuptools, tokenize;__file__='C:\\Temp\\pip-build-57pi877i\\xonsh\\setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record C:\Temp\pip-s3ldhy49-record\install-record.txt --single-version-externally-managed --compile:
    running install
    Building lexer and parser tables.
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "C:\Temp\pip-build-57pi877i\xonsh\setup.py", line 194, in <module>
        main()
      File "C:\Temp\pip-build-57pi877i\xonsh\setup.py", line 190, in main
        setup(**skw)
      File "c:\python35\lib\distutils\core.py", line 148, in setup
        dist.run_commands()
      File "c:\python35\lib\distutils\dist.py", line 955, in run_commands
        self.run_command(cmd)
      File "c:\python35\lib\distutils\dist.py", line 974, in run_command
        cmd_obj.run()
      File "C:\Temp\pip-build-57pi877i\xonsh\setup.py", line 95, in run
        build_tables()
      File "C:\Temp\pip-build-57pi877i\xonsh\setup.py", line 53, in build_tables
        from xonsh.parser import Parser
      File "c:\temp\pip-build-57pi877i\xonsh\xonsh\parser.py", line 3, in <module>
        from xonsh.platform import PYTHON_VERSION_INFO
      File "c:\temp\pip-build-57pi877i\xonsh\xonsh\platform.py", line 207, in <module>
        stderr=subprocess.PIPE)
      File "c:\python35\lib\subprocess.py", line 584, in check_call
        raise CalledProcessError(retcode, cmd)
    subprocess.CalledProcessError: Command '['bash', '--version']' returned non-zero exit status 4294967295
    ----------------------------------------
```

[...]

PR #1202, opened 2016-06-10
Fix for bash wsl

melund:
This a potential fix for the Error when Bash from Windows Subsystem for Linux is on the path. See #1200

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -61,6 +61,7 @@ Current Developments
 * Ensured that the prompt_toolkit shell functions, even without a ``completer``
   attribute.
 * Fixed crash resulting from malformed ``$PROMPT`` or ``$TITLE``.
+* Fix crash on startup when Bash Windows Subsystem for Linux is on the Path. 
 
 **Security:** None
 
--- a/xonsh/platform.py
+++ b/xonsh/platform.py
@@ -205,12 +205,11 @@ if ON_WINDOWS:
         subprocess.check_call([WINDOWS_BASH_COMMAND, '--version'],
                               stdout=subprocess.PIPE,
                               stderr=subprocess.PIPE)
-    except FileNotFoundError:
+    except (FileNotFoundError, CalledProcessError):
         if GIT_FOR_WINDOWS_PATH:
             bashcmd = os.path.join(GIT_FOR_WINDOWS_PATH, 'bin\\bash.exe')
             if os.path.isfile(bashcmd):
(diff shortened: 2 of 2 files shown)


adqm:
Is there any way to test this?  I don't have access to WSL.

melund:
Me neither. Windows insider updates are broken on my computer. @klonuo tested it here: https://github.com/xonsh/xonsh/issues/1200
As I understand, the fix worked for him. I don't think that xonsh would be able to use that version of bash since it runs in its own subsystem. 

But this should at least prevent it from crashing xonsh.

adqm:
Okay, that works for me.  Merging in.

Actually, I don't know much/anything about Windows.  Is there a way to tell whether I have WSL?  Or a way to get it if I don't?

gforsyth:
@adqm I think it's exclusive to Windows 10

adqm:
I am on Win10 on one of my machines.  If I have it up to date, does that mean i have WSL?

gforsyth:
I think so?  What happens if you run `bash` from the command prompt?

melund:
No. It is only available to Windows Insiders Slow or Fast ring. That is MS's beta program.

melund:
I guess you could install xonsh on WSL similar to what you would do on Ubuntu using apt-get. It uses the exact same binaries.

klonuo:
> I am on Win10 on one of my machines. If I have it up to date, does that mean i have WSL?

@adqm, WSL is disabled by default. You'll need to enable it first in Windows Features:

untitled

and also you need to have build >= 14316.

melund:
Just a question out of curiosity. Are there any significant improvement to the console. I heard it should be able to support full VT100 colors :)

klonuo:
I heard about that too, for the anniversary update, together with support for many ANSI codes, but it seems that ANSI sequences wont be available to other terminal host, except internally interpreted on console.
Console is improved a lot compared to previous Windows version, but Linux terminal is far reach IMHO ;)

klonuo:
> I guess you could install xonsh on WSL similar to what you would do on Ubuntu using apt-get. It uses the exact same binaries. 

@melund just to let you know, that latest update on fast ring (14366), added `listxattr` syscall, so it's possible to pip install xonsh in bash for Windows. 

Problem was described for example here: https://github.com/Microsoft/BashOnWindows/issues/385

melund:
Sounds interesting. Which I could upgrade

PR #1205, opened 2016-06-11
fixed piped jobs including alias on OS X

mitnk:
On mac, xonsh crashes with the following command (where `which` is aliased by Python func):

```
which -a ls | grep ls
```

@adqm @scopatz Could you please have a review. Thanks.


scopatz:
Thanks @mitnk!

PR #1210, opened 2016-06-11
Clean up argument parsing in premain by removing undo_args hack

cryzed:
I noticed the strange `undo_args` thing in `premain` while trying to better familiarize myself with xonsh's source code and saw an opportunity for a slight clean up. I tested a few times if the parsing behaviour is identical to that before the changes, and it seems to be the case -- but please test it yourself a bit more extensively!

Something I noticed is, that currently all arguments passed before the positional file argument are silently ignored -- is the current set of switches really always completely irrelevant to the xonsh script-file that is to be run? If that ever changes one could easily implement a set of "whitelisted" arguments that are to be copied over from the old set of parsed arguments to the new one.

### DIFF
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -144,20 +144,17 @@ def premain(argv=None):
     """Setup for main xonsh entry point, returns parsed arguments."""
     if setproctitle is not None:
         setproctitle(' '.join(['xonsh'] + sys.argv[1:]))
-
     builtins.__xonsh_ctx__ = {}
     args, other = parser.parse_known_args(argv)
     if args.file is not None:
         arguments = (argv or sys.argv)
         file_index = arguments.index(args.file)
-
         # A script-file was passed and is to be executed. The argument parser
         # might have parsed switches intended for the script, so reset the
         # parsed switches to their default values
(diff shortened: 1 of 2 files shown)


xonsh/main.py
```diff
@@ -218,15 +197,19 @@ def premain(argv=None):
...
     env = builtins.__xonsh_env__
     env['XONSH_LOGIN'] = shell_kwargs['login']
+
     if args.defines is not None:
         env.update([x.split('=', 1) for x in args.defines])
+
     env['XONSH_INTERACTIVE'] = False
     if ON_WINDOWS:
         setup_win_unicode_console(env.get('WIN_UNICODE_CONSOLE', True))
+
```
scopatz: Please remove this extra vertical whitespace.

scopatz:
Thanks a ton for putting this in @cryzed! I think this is a lot cleaner. I tested it out and it basically seems to work as before.  

Would you mind merging from master and adding a news file about this, please, too? We have added some instructions in the CONTRIBUTING file

scopatz:
Thanks again @cryzed!

Issue #572, opened 2015-12-15
history all command


JohnLunzer:
It would be nice to have a "history all" or "history show --all" command that shows the history from all past and (optionally) concurrent sessions. This behavior would be similar to the way zsh behaves. 

As requested the concurrent updating behavior should not be turned on by default. 

For additional details see the mailing list discussion: https://groups.google.com/forum/#!topic/xonsh/VVIPFQSXitI

josefson:
Zsh shell is really good at working with history, this ability to see other shell instance's history and use it is really handy. But it goes further than that, many zsh users use a fish-like function that perform a substring search in all past history and present from all sessions.
In fact, we are so in use with history functions that it would be really cool if we could somehow import all of our zsh hsitory to xonsh, the same could be said from bash i think. We would not start from point 0. And i think this would improve xonsh adoption once we would have our neat history completion in hands.

scopatz:
Pull requests definitely welcome for that @josefson!

JohnLunzer:
Ha, so hilarious that I opened up two issues for this like 6 months apart. I think I'll close the newer one #952.

I've been working on the "history all", but job work and life has been getting in the way.

I have a python only function that performs history all and I'm pretty happy with it. Part of what I like about it is that it has the option to return a python list with a tuple of (command, command_start_time, history_index).

But when I was converting this function to a "history all" alias I ran into the issue that I think xonsh aliases cannot return a python object? Is this the case? And if it is the case I was thinking about adding a method to the __xonsh__.history object to return the list of all history instead. Does this sound reasonable?

To what @josefson said, I also have made a function to list all Zsh history as well, it would not be difficult to make a "history zsh" or "history bash" alias.

scopatz:
> But when I was converting this function to a "history all" alias I ran into the issue that I think xonsh aliases cannot return a python object? Is this the case?

That is true.  xonsh aliases are like subprocesses.  They can only communicate through stdin, stdout, stderr, and the return code.  We have no notion of a subprocess returning a Python object.  We could potentially consider something like this, but:

1. there are potential security concerns
2. why not just use a Python / xonsh function in python mode?

This is probably off into the weeds a bit ;)

> And if it is the case I was thinking about adding a method to the __xonsh__.history object to return the list of all history instead. Does this sound reasonable?

[...]

PR #1211, opened 2016-06-12
History all

JohnLunzer:
This PR should fully address #572. Some pretty good stuff here!

There are a couple of changes which I'd like to address as they may be seen as controversial. I've removed `history show`. This was redundant as the simpler `history` did the exact same thing. I highly doubt anyone was using `history show` in scripts and so this should not be seen as a breaking change. Though the documentation will need to be updated. 

The documentation will also need to take into account the new functionality for `__xonsh_history__.show()`, which probably needs its docstring to be a little more descriptive as well. 

Frankly there are probably a couple of commits left to make but I wanted to get this out into the wild for people to test so I could get some feedback.


JohnLunzer:
Well it looks like I missed updating some of the tests.

xonsh/history.py
```diff
@@ -334,76 +488,81 @@ def flush(self, at_exit=False):
...
-    if ns.reverse:
-        indices = reversed(indices)
-        inps = reversed(inps)
-    for i, inp in zip(indices, inps):
-        lines = inp.splitlines()
-        lines[0] = ' {0:>{1}}  {2}'.format(i, ndigits, lines[0])
-        lines[1:] = [indent + x for x in lines[1:]]
-        print('\n'.join(lines))
+    @staticmethod
+    def show(ns = None, hist = None, start_index = None, end_index = None,
```
gforsyth: PEP8 no spaces between `=` for default arguments

xonsh/history.py
```diff
@@ -432,26 +591,30 @@ def _gc(ns, hist):
...
+    'session': History.show,
+    'all': History.show,
+    'zsh': History.show,    
+    'bash': History.show,    
     'id': lambda ns, hist: print(hist.sessionid),
     'file': lambda ns, hist: print(hist.filename),
     'info': _info,
     'diff': diff_history._main_action,
-    'gc': _gc,
+    'gc': _gc
```
gforsyth: why remove the trailing comma?  trailing commas are your friend

xonsh/history.py
```diff
@@ -229,6 +232,157 @@ def i_am_at_the_front(self):
         return self is self.hist._queue[0]
 
 
+def _all_xonsh_formatter(*args):
+    """
+    Returns all history as found in XONSH_DATA_DIR.
+
```
gforsyth: PEP8 leading whitespace in blank line

xonsh/history.py
```diff
@@ -229,6 +232,157 @@ def i_am_at_the_front(self):
...
+def _all_xonsh_formatter(*args):
+    """
+    Returns all history as found in XONSH_DATA_DIR.
+    
+    return format: (name, start_time, index)
+    """
+    data_dir = builtins.__xonsh_env__.get('XONSH_DATA_DIR')
+    data_dir = os.path.expanduser(data_dir)
+
+    files = [os.path.join(data_dir,f) for f in listdir(data_dir)
```
gforsyth: PEP8 trailing whitespace

xonsh/history.py
```diff
@@ -229,6 +232,157 @@ def i_am_at_the_front(self):
...
+    Returns all history as found in XONSH_DATA_DIR.
+    
+    return format: (name, start_time, index)
+    """
+    data_dir = builtins.__xonsh_env__.get('XONSH_DATA_DIR')
+    data_dir = os.path.expanduser(data_dir)
+
+    files = [os.path.join(data_dir,f) for f in listdir(data_dir) 
+             if f.startswith('xonsh-') and f.endswith('.json')]
+    file_hist = [lazyjson.LazyJSON(f, reopen=False).load()['cmds']
```
gforsyth: PEP8 trailing whitespace

xonsh/history.py
gforsyth: PEP8 space after comma

xonsh/history.py
```diff
@@ -229,6 +232,157 @@ def i_am_at_the_front(self):
...
+    data_dir = os.path.expanduser(data_dir)
+
+    files = [os.path.join(data_dir,f) for f in listdir(data_dir) 
+             if f.startswith('xonsh-') and f.endswith('.json')]
+    file_hist = [lazyjson.LazyJSON(f, reopen=False).load()['cmds'] 
+                for f in files]
+    commands = [(c['inp'].replace('\n', ''), c['ts'][0])
+                for commands in file_hist for c in commands if c]
+    commands.sort(key=itemgetter(1))
+    return [(c,t,ind) for ind,(c,t) in enumerate(commands)]
```
gforsyth: PEP8 spaces after commas

xonsh/history.py
```diff
@@ -229,6 +232,157 @@ def i_am_at_the_front(self):
...
+
```
gforsyth: PEP8 two blank lines between function definitions

xonsh/history.py
```diff
@@ -229,6 +232,157 @@ def i_am_at_the_front(self):
...
+    Take in History object and return command list tuple with
+    format: (name, start_time, index)
+    """
+    if not hist:
+        hist = builtins.__xonsh_history__
+    start_times= [start for start,end in hist.tss]
+    names = [name[:-1] for name in hist.inps]
+    commands = enumerate(zip(names, start_times))
+    return [(c,t,ind) for ind,(c,t) in commands]
+def _zsh_hist_formatter(location = None):
```
gforsyth: PEP8 two blank lines

xonsh/history.py
```diff
@@ -229,6 +232,157 @@ def i_am_at_the_front(self):
...
+    if not location:
```
gforsyth: PEP8 trailing whitespace

xonsh/history.py
```diff
@@ -229,6 +232,157 @@ def i_am_at_the_front(self):
...
+    """
+    if not hist:
+        hist = builtins.__xonsh_history__
+    start_times= [start for start,end in hist.tss]
+    names = [name[:-1] for name in hist.inps]
+    commands = enumerate(zip(names, start_times))
+    return [(c,t,ind) for ind,(c,t) in commands]
+def _zsh_hist_formatter(location = None):
+    if not location: 
+        location = os.path.join('~','.zsh_history')
```
gforsyth: PEP8 space after comma

xonsh/history.py
```diff
@@ -229,6 +232,157 @@ def i_am_at_the_front(self):
...
+                        try:
+                            start_time = float(start_time.split(':')[1])
+                        except ValueError:
+                            start_time = -1
+                        z_hist_formatted.append((command, start_time, ind))
+                    return z_hist_formatted
+            except:
+                print("There was a problem parsing {}.".format(bash_path))
+                return None
+
```
gforsyth: PEP8 whitespace in blank line

xonsh/history.py
```diff
@@ -229,6 +232,157 @@ def i_am_at_the_front(self):
...
+                        z_hist_formatted.append((command, start_time, ind))
+                    return z_hist_formatted
+            except:
+                print("There was a problem parsing {}.".format(bash_path))
+                return None
+                    
+    else:
+        print("No zsh history file found at: {}".format(z_path))
+        return None
+
```
gforsyth: PEP8 whitespace in blank line

xonsh/history.py
```diff
@@ -229,6 +232,157 @@ def i_am_at_the_front(self):
...
+                    return z_hist_formatted
+            except:
+                print("There was a problem parsing {}.".format(bash_path))
+                return None
+                    
+    else:
+        print("No zsh history file found at: {}".format(z_path))
+        return None
+    
+def _bash_hist_formatter(location = None):
```
gforsyth: PEP8 no spaces around keyword equals

xonsh/history.py
```diff
@@ -229,6 +232,157 @@ def i_am_at_the_front(self):
...
+    if not location: 
+        location = os.path.join('~','.bash_history')
```
gforsyth: space after comma

JohnLunzer:
@gforsyth, I will run autopep8 on it, should have done that before committing. I was more looking for feedback on functionality and the algorithms.

gforsyth:
Hey @JohnLunzer, thanks for putting this in!  You should run PEP8 (or whatever it's called now) on the code and fix the style issues -- I went a little overboard on the inline comments and then stopped myself (apologies).  

`history bash` works, but `history zsh` is throwing this error:
```console
$ history zsh
  File "/home/gil/git/xonsh/xonsh/proc.py", line 336, in wrapped_simple_command
    r = f(args, i)
  File "/home/gil/git/xonsh/xonsh/history.py", line 624, in main
    _main(builtins.__xonsh_history__, args)  # pylint: disable=no-member
  File "/home/gil/git/xonsh/xonsh/history.py", line 618, in _main
    _MAIN_ACTIONS[ns.action](ns, hist)
  File "/home/gil/git/xonsh/xonsh/history.py", line 517, in show
    commands = valid_formats[ns.action]()
  File "/home/gil/git/xonsh/xonsh/history.py", line 271, in _zsh_hist_formatter
    z_hist = z_file.read().splitlines()
  File "/home/gil/miniconda3/lib/python3.5/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa5 in position 22575: invalid start byte
```

`history all` is also throwing an error:

```console
$ history all
  File "/home/gil/git/xonsh/xonsh/proc.py", line 336, in wrapped_simple_command
    r = f(args, i)
  File "/home/gil/git/xonsh/xonsh/history.py", line 624, in main
    _main(builtins.__xonsh_history__, args)  # pylint: disable=no-member
  File "/home/gil/git/xonsh/xonsh/history.py", line 618, in _main
    _MAIN_ACTIONS[ns.action](ns, hist)
  File "/home/gil/git/xonsh/xonsh/history.py", line 517, in show
    commands = valid_formats[ns.action]()
  File "/home/gil/git/xonsh/xonsh/history.py", line 247, in _all_xonsh_formatter
    for f in files]
  File "/home/gil/git/xonsh/xonsh/history.py", line 247, in <listcomp>
    for f in files]
  File "/home/gil/git/xonsh/xonsh/lazyjson.py", line 207, in __init__
    self._load_index()
  File "/home/gil/git/xonsh/xonsh/lazyjson.py", line 235, in _load_index
    locs = json.loads(locs)
  File "/home/gil/miniconda3/lib/python3.5/json/__init__.py", line 319, in loads
    return _default_decoder.decode(s)
  File "/home/gil/miniconda3/lib/python3.5/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/home/gil/miniconda3/lib/python3.5/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

JohnLunzer:
@gforsyth should be fully pep8 compliant now. 

Additionally I fixed both of the bugs you were having. The first I have experienced myself and I had it fixed in the .xonrc alias version of this but mistakenly left it out of this PR. 

The second issue is because you likely have non json files or erroneous json files in your XONSH_DATA_DIR, but I'm handling them now. 

And hopefully all the tests pass too now.

JohnLunzer:
Ahh. My tests are passing locally but obviously because Travis doesn't have my updated code it's still failing the CI tests. Perhaps others who test can confirm the tests pass.

gforsyth:
Hey @JohnLunzer -- Travis runs the updated code (and tests, if applicable).  It's failing because `JSONDecodeError` is new in Python 3.5, so the Python 3.4 tests are failing with an import error.

gforsyth:
Since the `JSONDecodeError` is just there to catch bad files, I think you can replace it with a `ValueError` and it will work the same.

JohnLunzer:
@gforsyth Ahh, I didn't know. I'll try to make that change and see what happens.

xonsh/history.py
```diff
@@ -229,6 +232,177 @@ def i_am_at_the_front(self):
...
+    session = subp.add_parser('session',
+                              help='displays session history, default action')
+    session.add_argument('-r', dest='reverse', default=False,
+                         action='store_true',
+                         help='reverses the direction')
+    session.add_argument('n', nargs='?', default=None,
+                         help='display n\'th history entry if n is a simple '
+                              'int, or range of entries if it is Python '
+                              'slice notation')
+    # 'id' subcommand
```
gforsyth: Hey @JohnLunzer -- shoudl this be down on line 370?

JohnLunzer: I use _create_parser() in History.show() and so I put it above the class definition for History. Not sure if that was necessary.

gforsyth: Sorry, I just meant the comment `# 'id' subcommand`

JohnLunzer: Ahh yes, it looks like that migrated up accidentally!

scopatz:
Looking at this now.

xonsh/history.py
```diff
@@ -432,26 +621,30 @@ def _gc(ns, hist):
 
 
 _MAIN_ACTIONS = {
-    'show': _show,
+    'session': History.show,
```
scopatz: Please add back show.

scopatz: I actually do use this.

tests/test_history.py
```diff
@@ -114,47 +114,51 @@ def run_show_cmd(hist_args, commands, base_idx=0, step=1):
...
 
         # Verify that showing a history range relative to the end of the
         # history works.
-        for x in run_show_cmd(['show', '-2:'], cmds[-2:], len(cmds) - 2):
+        for x in run_session_cmd(['session', '-2:'], 
+                                 cmds[-2:], len(cmds) - 2):
             yield x
-        for x in run_show_cmd(['show', '-4:-2'], cmds[-4:-2], len(cmds) - 4):
+        for x in run_session_cmd(['session', '-4:-2'], 
+                                 cmds[-4:-2], len(cmds) - 4):
```
scopatz: All of these show tests should be restored and potenially new session tests added.

xonsh/tools.py
```diff
@@ -794,11 +794,15 @@ def ensure_int_or_slice(x):
...
-        return int(x)
+    try:
+        if ':' in x:
+            x = x.strip('[]()')
+            return slice(*(int(x) if len(x) > 0 else None for x in x.split(':')))
+        else:
+            return int(x)
+    except ValueError:
+        print("Could not convert index or slice to int.")
+        return False
```
scopatz: This should simply fail if incorrect input is passed to it.  It is the job of an `is_int_or_slice()` function to makes sure that x is valid.

xonsh/history.py
scopatz: Maybe session should just be renamed show, or one should be an alias for the other.

xonsh/history.py
```diff
@@ -432,26 +621,30 @@ def _gc(ns, hist):
...
+}
 
 
 def _main(hist, args):
     """This implements the history CLI."""
     if not args or (args[0] not in _MAIN_ACTIONS and
                     args[0] not in {'-h', '--help'}):
-        args.insert(0, 'show')
-    if (args[0] == 'show' and len(args) > 1 and args[-1].startswith('-') and
+        args.insert(0, 'session')
```
scopatz: The default should be show, please.

xonsh/history.py
```diff
@@ -229,6 +232,180 @@ def i_am_at_the_front(self):
...
+
+    else:
+        print("No zsh history file found at: {}".format(z_path))
+        return None
+
+
+def _bash_hist_formatter(location=None):
+    if not location:
+        location = os.path.join('~', '.bash_history')
+    bash_hist_formatted = list()
```
scopatz: Please use literals (`[]`) instead of type construction `list()`

xonsh/history.py
```diff
@@ -229,6 +232,180 @@ def i_am_at_the_front(self):
...
+                return None
+
+    else:
+        print("No zsh history file found at: {}".format(z_path))
+        return None
+
+
+def _bash_hist_formatter(location=None):
+    if not location:
+        location = os.path.join('~', '.bash_history')
```
scopatz: From the Bash man page:

```
       HISTFILE
              The name of the file in which command history is saved (see HISTORY below).  The default value is ~/.bash_history.  If unset, the command history  is  not
              saved when a shell exits.
```

So we should be looking up `$HISTFILE` in the environment and only falling back to `~/.bash_history` if it is not there.

xonsh/history.py
```diff
@@ -229,6 +232,180 @@ def i_am_at_the_front(self):
...
+                bash_hist = b_txt.splitlines()
+                if bash_hist:
+                    for ind, command in enumerate(bash_hist):
+                        bash_hist_formatted.append((command, 0.0, ind))
+                    return bash_hist_formatted
+        except:
+            print("There was a problem parsing {}.".format(b_path))
+            return None
+    else:
+        print("No bash history file found at: {}".format(b_path))
```
scopatz: These prints should go to `sys.stderr` rather than stdout. They are messages to the user, not input to another program :)

xonsh/history.py
```diff
@@ -229,6 +232,180 @@ def i_am_at_the_front(self):
...
+        return None
```
scopatz: I think you can remove the `return None` lines and the flow will work itself out

xonsh/history.py
```diff
@@ -229,6 +232,180 @@ def i_am_at_the_front(self):
...
+def _bash_hist_formatter(location=None):
+    if not location:
+        location = os.path.join('~', '.bash_history')
+    bash_hist_formatted = list()
+    b_path = os.path.expanduser(location)
+    if os.path.isfile(b_path):
+        try:
+            with open(b_path, 'r') as bash_file:
+                b_txt = bash_file.read()
+                b_txt = b_txt.encode('utf-8', 'replace').decode('utf-8')
```
scopatz: reading in already happens in 'UTF-8' since you are in text mode by default.

xonsh/history.py
```diff
@@ -229,6 +232,180 @@ def i_am_at_the_front(self):
...
+
+
+def _bash_hist_formatter(location=None):
+    if not location:
+        location = os.path.join('~', '.bash_history')
+    bash_hist_formatted = list()
+    b_path = os.path.expanduser(location)
+    if os.path.isfile(b_path):
+        try:
+            with open(b_path, 'r') as bash_file:
```
scopatz: I think you should move the with open outside of the try-block.  and the with only should contain the read() call. It should be one line long.

xonsh/history.py
```diff
@@ -229,6 +232,180 @@ def i_am_at_the_front(self):
...
+        return None
+
+
+def _bash_hist_formatter(location=None):
+    if not location:
+        location = os.path.join('~', '.bash_history')
+    bash_hist_formatted = list()
+    b_path = os.path.expanduser(location)
+    if os.path.isfile(b_path):
+        try:
```
scopatz: The try block should be as small as possible, preferably only one line.

xonsh/history.py
JohnLunzer: @scopatz I will add it back in on popular request, but I encourage you to consider its utility (backwards compatibility) versus its downsides. I would like others to weigh in as well.

As I stated it is redundant as `history` will perform the same functionality. This should be evidence enough for it removal. In addition, with the expanded functionality `show` is ambiguous. Show what? `session` and `all` clarify that ambiguity. For these reasons I am strongly against restoring it, in the name of logic and cleanliness.

In what way are you using `show` that would justify leaving in ambiguous redundancy? If you need the default `session` behavior (which mostly mimics the old `show`) changed to be more acceptable I am open to discussion.

xonsh/history.py
```diff
@@ -229,6 +232,180 @@ def i_am_at_the_front(self):
...
+        return None
+    start_times = [start for start, end in hist.tss]
+    names = [name[:-1] if name.endswith('\n') else name
+             for name in hist.inps]
+    commands = enumerate(zip(names, start_times))
+    return [(c, t, ind) for ind, (c, t) in commands]
+
+
+def _zsh_hist_formatter(location=None):
+    if not location:
```
scopatz: This should be `if location is None`, I believe it is much faster.

xonsh/history.py
```diff
@@ -229,6 +232,180 @@ def i_am_at_the_front(self):
...
+                        z_hist_formatted.append((command, start_time, ind))
+                    return z_hist_formatted
+            except Exception as e:
+                print("There was a problem parsing {}.".format(z_path))
+                raise e
+                return None
+
+    else:
+        print("No zsh history file found at: {}".format(z_path))
+        return None
```
scopatz: Same comments here as with the bash function.

xonsh/history.py
```diff
@@ -229,6 +232,180 @@ def i_am_at_the_front(self):
...
+        except ValueError:
+            # Invalid json file
+            pass
+    commands = [(c['inp'].replace('\n', ''), c['ts'][0])
+                for commands in file_hist for c in commands if c]
+    commands.sort(key=itemgetter(1))
+    return [(c, t, ind) for ind, (c, t) in enumerate(commands)]
+
+
+def _curr_session_formatter(hist):
```
scopatz: maybe make hist a kwarg?

scopatz:
Thanks a million for putting this in @JohnLunzer! Overall, I think this is basically the right idea. I have commented in-line where I think that the implementation could be cleaned up. My top-level concerns are that:

1. the `*_formatter()` funcs aren't really formatters because they don't format strings.  They are more like parsers or collectors, and should be renamed.
2. The `all` history, is only for all xonsh sessions, not all history that it could possibly know about (unless I am mistaken). I would change this to xonsh, and if you want to you can implement an all that grabs all known history.
3. The show -> session rename should be reverted. I think it is fine to have session as an option if you feel that is more natural, but it should alias show.

Thanks again!

xonsh/history.py
scopatz: Hi @JohnLunzer - I guess all that I want is for there to be a show command that displays only this session and for that to be the default behaviour.  Lots of CLI have redundant commands/aliases, git checkout vs git co for example, so I don't really see this being a big deal.  I think if you called your session 'show' from the cli, I would be happy. I am not saying you need to add the old function back in, just preserve the interface.

xonsh/history.py
```diff
@@ -333,77 +511,88 @@ def flush(self, at_exit=False):
         self.buffer.clear()
         return hf
 
+    @staticmethod
+    def show(ns=None, hist=None, start_index=None, end_index=None,
```
scopatz: This seems like it should be a function in the global namespace, not a staticmethod.  It seems deeply tied to the CLI and not the history class itself.

JohnLunzer:
@scopatz , I agree with your first two points, but can you please provide reasoning for this:
> The show -> session rename should be reverted. I think it is fine to have session as an option if you feel that is more natural, but it should alias show.

I am looking for the logic of why `show` is superior to `session`. Just as you pointed out that `all` remains ambiguous, `show` is even more ambiguous. As I said before, `show` what?

This is a huge overhaul of xonsh history, this is an opportunity to clean things up. Before it did not matter to have `show` because of the limited functionality. Please compare these two lists of sub-commands and tell me which makes more sense.
```
session
xonsh
zsh
bash
```
vs 
```
show
xonsh
zsh
bash
```
To me the choice is clear. One list has an ambiguous option and one doesn't. Xonsh is not a 1.0 product, we should not be afraid to clean things up when it makes sense.

scopatz:
> To me the choice is clear. 

And to me it is clear too in the other direction :).  To me show is both shorter and more intuitive.  A user doesn't need to know anything about the xonsh history model to run a `show`  option. Many different kinds of CLIs have a show command, or similar. It provides an initial way to dump out some information about the current state.  I feel that it is well understood that show won't make any modifications, and thus is safe to run if a user is just trying things out.

Then if a user wonders why the history is limitied with show, they can then learn a little more about xonsh's history model.  Show provides a pathway for learning by using the tool.

I feel that session, on the other hand, requires the user to know at least a little about the history model before they feel comfortable using it.  This doesn't matter to old hats, but can be a big deal for new users of xonsh, or new users of shells, or new users of CLIs.

I'd rather not bikeshed this too much :).  But this is why I think it is fine if show and session alias each other, and if there is to be only one it should be show.  I wrote it like I did originally with these "new users" in mind. One of the primary goals for me for xonsh is for it to be approachable in a way that other shells aren't.

I hope that this makes sense, and justifies why this is a blocking issue for me.

JohnLunzer:
@scopatz , those are all perfectly reasonable and exactly what I was looking for. "Add show back, I use it" didn't seem reasonable , hence my pushing the issue.

In the name of cleanliness I will forgo having both `show` and `session` and just stick with `show`.

xonsh/history.py
JohnLunzer: Seems reasonable.

xonsh/history.py
JohnLunzer: Sounds okay.

xonsh/history.py
JohnLunzer: I didn't realize, thanks.

xonsh/history.py
JohnLunzer: This fixed a bug @gforsyth was having. I learned solution from [here](http://stackoverflow.com/questions/22216076/unicodedecodeerror-utf8-codec-cant-decode-byte-0xa5-in-position-0-invalid-s).

If you have another solution to his bug I am open to suggestions.

xonsh/history.py
JohnLunzer: I think you're right, should be fine.

xonsh/history.py
JohnLunzer: Agreed.

JohnLunzer:
Also @scopatz , sorry I didn't not see your shorter answer higher up. Had I seen it that would have been enough of an explanation for me. Not trying to be difficult :)

xonsh/history.py
JohnLunzer: The problem with this is that both zsh and bash use $HISTFILE for their custom locations. I was thinking instead of looking in the environment to look in the .zprofile/.zshrc and .bash_profile/.bashrc respectively for if HISTFILE is set. I am unfamiliar with xonsh's ability to find out these things, maybe it can already do what I'm suggesting?

xonsh/history.py
JohnLunzer: I have no problem replacing this? Why are trailing commas our friend?

gforsyth: Generally it makes editing easier -- if you rearrange a list you don't accidentally introduce syntax errors when you transpose lines.  Also it makes for cleaner diffs when you add items to a list.

JohnLunzer: Thanks for the explanation.

scopatz:
Sounds fine to me :)

xonsh/history.py
JohnLunzer: So I had a good look at this and reading doesn't automatically happen in 'UTF-8' mode. See the [page on built-ins](https://docs.python.org/3/library/functions.html#open) for details. Encoding is determined by whatever `locale.getpreferredencoding()` returns which in the case of my windows machine is `cp1252`. By leaving the encoding alone and doing error handling as an argument to `open` I was able to make this compatible across platforms.

JohnLunzer:
Last major change I need to make is parsing the zshrc/zprofile bashrc/bash_profile files to check for setting HISTFILE. Please let me know if anyone sees any other issues.

JohnLunzer:
This PR should be feature complete and all items of concern addressed. @gforsyth, @scopatz please give this another look over at your convenience.

xonsh/tools.py
```diff
@@ -629,14 +629,23 @@ def swap(namespace, name, value, default=NotImplemented):
...
-# Validators and contervers
+# Validators and converters
 #
 
 
 def is_int(x):
     """Tests if something is an integer"""
     return isinstance(x, int)
 
+def is_str_as_int(x):
```
scopatz: Wouldn't this be named, `is_int_as_str()`?  str as int would be bytes to me :)

scopatz: Also this function could use some unit tests please.

xonsh/tools.py
```diff
@@ -629,14 +629,23 @@ def swap(namespace, name, value, default=NotImplemented):
...
+    """Tests if something is an integer"""
+    if isinstance(x, str):
+        try:
+            int(x)
+        except ValueError:
+            return False
+    return True
```
scopatz: Also I am not sure that this is the correct return value, if I pass in a list, this will return True.

xonsh/tools.py
```diff
@@ -647,6 +656,23 @@ def is_string(x):
     """Tests if something is a string"""
     return isinstance(x, str)
 
+def is_slice(x):
+    """Tests if something is a slice"""
+    return isinstance(x, slice)
```
scopatz: Again, unit tests would be desirable, please.

xonsh/tools.py
```diff
@@ -647,6 +656,23 @@ def is_string(x):
...
+        
+def is_str_as_slice(x):
```
scopatz: Same issues as with int function above.

scopatz:
Thanks @JohnLunzer! I think this is shaping up nicely.  Just a couple of semantic issues with the validators and then more tests. Then I think that this is good to go.

xonsh/tools.py
JohnLunzer: Agreed.

JohnLunzer:
@scopatz, you're right since these are tests it's important that they themselves have tests. I'll write them up.

xonsh/history.py
```diff
@@ -515,20 +514,21 @@ def _show(ns=None, hist=None, start_index=None, end_index=None,
...
-            for line_ind, line in enumerate(c.split('\n')):
-                if line_ind == 0:
-                    print('{:>{width}}: {}'.format(i, line,
-                                                   width=digits + 1))
-                else:
-                    print(' {:>>{width}} {}'.format('', line,
-                                                    width=digits + 1))
-    else:
-        return commands
+        commands = list(reversed(commands))
```
JohnLunzer: Test was failing because reversed returns iterator and was being exhausted by the newly fixed digits length test.

JohnLunzer:
@scopatz, let me know what you think of the tests. 

On another note, test_history.py has nine pep8 violations, test_tools.py has 18 pep8 violations, and tools.py has 45 pep8 violations. Just thought I'd mention it since there was concern shown for pep8 adherence. Didn't feel like it was the place of this PR to correct them.

melund:
Pep8 in test are not very important compared to the main code base. But we can probably fix some of them as part of the move to pytest.

JohnLunzer:
@melund, somebody should probably still run `autopep8 --in-place --aggressive` on `xonsh/tools.py`

tests/test_tools.py
```diff
@@ -10,15 +10,41 @@
...
+    is_string,
+    is_string_or_callable,
+    logfile_opt_to_str,
+    str_to_env_path,
+    subexpr_from_unbalanced,
+    subproc_toks,
+    to_bool,
+    to_bool_or_int,
+    to_dynamic_cwd_tuple,
+    to_logfile_opt)
```
scopatz: This doesn't block, but please revert this to the way it was.  Vertical whitespace is precious ;)

tests/test_tools.py
```diff
@@ -321,13 +347,75 @@ def test_find_next_break():
...
+        ('42.0', False),
+        ([42], False),
+        ([], False),
+        (None, False),
+        ('', False)
+        ]
+    for inp, exp in cases:
+        obs = is_int(inp)
+        yield assert_equal, exp, obs
+def test_is_int_as_str():
```
scopatz: PEP8 two blank lines after functions.

tests/test_tools.py
```diff
@@ -321,13 +347,75 @@ def test_find_next_break():
...
+        ([], False),
+        (None, False),
+        ('', False),
+        (False, False),
+        (True, False),
+        ]
+    for inp, exp in cases:
+        obs = is_int_as_str(inp)
+        yield assert_equal, exp, obs
+def test_is_float():
```
scopatz: PEP8 here too

scopatz:
Ok @JohnLunzer - other than a couple of minor test style issues, this looks basically good to go.  Pleas just also add a news file for this branch.  You can see the instructions here: https://github.com/xonsh/xonsh/blob/master/CONTRIBUTING.rst#changelog

Thanks again for all of your efforts here.  They are very much appreciated!

tests/test_tools.py
JohnLunzer: For the record autopep8 made that change. After working with it for a while, I don't disagree with it. 

My whole point in doing this was because in the _unalphabetized wall of text_ it was not obvious to me which functions were being imported from tools.

With each import is on its own line and in alphabetical order then it is easy to determine what has been imported. It is also easy to keep it alphabetized. These seem like worthwhile trade-offs, and pep8 compliant to boot.

I guess this doesn't make sense to me: "Vertical whitespace is precious". This doesn't seem like a Pythonic statement. Pep8 recommends using vertical whitespace in many places.

gforsyth: I think one concern with this rather tall block of text is that it impacts the ability of others to look through everything that's being imported without scrolling in their editor.  Personally, I also scan text faster left-to-right rather than top-to-bottom

JohnLunzer: I at least recommend that we try to keep this list alphabetized.

gforsyth: I think that's a good idea.

JohnLunzer:
Added in the news file for this branch.

gforsyth:
Hey @JohnLunzer -- this looks really nice -- one change still, I think, that needs to be made.  
The news file should be RST formatted as bullet points and will be rolled into the changelog for the next release, thus the contents should be more along the lines of "new history all command will ..."

That said, what you've written there now is good documentation and would fit in well in the main docs.  And given the added functionality, it probably belongs in the docs anyway.  So maybe you can move that writeup into the history docs and then replace the news file contents with something a little more terse to describe the changes.  

Almost there!

news/history-all.rst
```diff
@@ -0,0 +1,50 @@
+**Added:** None
```
gforsyth: Ahh, also you should remove the `None` here or none of the contents below will get rolled in to the changelog

scopatz: I'll go ahead and do that  I'll merge in a couple of minutes after playing around with it.

scopatz: Ahh nevermind, I agree with your doc comments.

xonsh/history.py
```diff
@@ -1,17 +1,22 @@
...
 import argparse
-import functools
 import os
+import sys
+from functools import lru_cache, partial
+from os import listdir
+from operator import itemgetter
 import uuid
 import time
+from datetime import datetime
```
scopatz: Also, since #1220 we merged this morning, these non-package imports on the above lines should be changed to direct imports. See https://github.com/xonsh/xonsh/blob/master/CONTRIBUTING.rst#imports  If you don't want to take care of it, let me know and I will happily do this.  I am going to go ahead and comment what the imports should be, though.

This one should just be `import datetime`

xonsh/history.py
```diff
@@ -1,17 +1,22 @@
 # -*- coding: utf-8 -*-
 """Implements the xonsh history object."""
 import argparse
-import functools
 import os
+import sys
+from functools import lru_cache, partial
+from os import listdir
```
scopatz: This one should be removed since os is already imported.

xonsh/history.py
```diff
@@ -1,17 +1,22 @@
...
+from operator import itemgetter
```
scopatz: This should just be `import operator`

xonsh/history.py
```diff
@@ -1,17 +1,22 @@
 # -*- coding: utf-8 -*-
 """Implements the xonsh history object."""
 import argparse
-import functools
 import os
+import sys
+from functools import lru_cache, partial
```
scopatz: This should return to just `import functools`

docs/tutorial_hist.rst
```diff
@@ -1,13 +1,8 @@
 .. _tutorial_hist:
 
-************************************
-Tutorial: History
-************************************
```
scopatz: Why are all the headings removed in this file?

news/history-all.rst
```diff
@@ -0,0 +1,21 @@
...
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
+
+.. Warning: this node is ignored when writing this file.
+.. However, @ @rst-options are recognized in this node.
```
scopatz: Not sure that these comments ar needed either...

docs/tutorial_hist.rst
```diff
@@ -400,3 +413,6 @@ implemented:
 * and many more!
 
 Let us know if you'd be interested in working on any of these, inside or outside of xonsh.
+
+.. Warning: this node is ignored when writing this file.
+.. However, @ @rst-options are recognized in this node.
```
scopatz: Same here

JohnLunzer:
Please tell me I'm done :/

Btw, after I merged with master xonsh is wicked slow on my windows machine. Doesn't seem to matter what command/action I use.

I think the branch might be due to the alphabetized imports in test_tools.py. I was just trying to clean things up but I don't want to have to keep manually merging...

scopatz:
> Btw, after I merged with master xonsh is wicked slow on my windows machine. Doesn't seem to matter what command/action I use.

Did you do a clean build?

scopatz:
> Please tell me I'm done :/

Yeah, you can be done if you want.  I think that there is still some cleanup that needs to happen but I can take care of that.

JohnLunzer:
@scopatz I think if you're 95% happy but there are some minor issues it would be more efficient and less exhausting for all if you topped this off rather than to continue back and forth. I'd be extremely grateful for the help.

docs/tutorial_hist.rst
JohnLunzer: Just an accident.

JohnLunzer: I use a tool called Leo to edit source code and it auto imports different file types and cuts them up. Apparently it removed the headings...

scopatz:
No problem, thanks for putting this in @JohnLunzer! It is a great feature to have.

scopatz:
Sorry about all the back and forth. We have been moving pretty quickly recently.  The slow startup is a big concern, thoguh, since many of the recent changes have been in an effort to speed things up.

JohnLunzer:
@scopatz thanks again for all the help and patience. Sorry about the tutorial_hist.rst headings! Haven't used rst files in Leo before... maybe never again ;)

docs/tutorial_hist.rst
scopatz: Haha no worries :)

PR #1213, opened 2016-06-12
added top level xonsh enterpoint for dev

mitnk:
When developing xonsh inside the repo and If we want to test something without installing the code to system, we would run following command:

```
the-repo $ python3 xonsh/main.py -c 'echo hi'
```

But we will get `AttributeError: module 'platform' has no attribute 'system'` since Python will add `the-repo/xonsh` to the beginning of `sys.path`, and we have our own `platform` module inside this dir which causes this issue.

This PR is trying to add a helper enterpoint for developers only, which allow you to test xonsh like this:

```
the-repo $ python3 main.py -c 'echo hi'
```

Does this PR make sense to you?

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,0 +1,13 @@
+"""
+A helper xonsh enterpoint for developers.
+
+With this file, we can run/test xonsh inside the xonsh's repo like following:
+
+$ python3 main.py -c 'echo hi'
+"""
+
+from xonsh.main import main
+
+
+if __name__ == '__main__':
+    main()


scopatz:
Thanks for putting this in @mitnk! Isn't this basically the same as `scripts/xonsh`, though?

mitnk:
Oops! Didn't know that! Thanks for pointing out :)

mitnk:
@scopatz Actually, `scripts/xonsh` seems not work as it was expected. when you run `./scripts/xonsh`, error `ImportError: No module named xonsh.main`; and if you run `python3 scripts/xonsh`, then which actually be run is the one installed on system, not the code in current repo directory.

Or I misunderstand the usage of this script?

mitnk:
maybe we replace `scripts/xonsh` with `main.py`?

adqm:
I am +1 to merge this in, @mitnk.  Thanks for putting it in!  Nice to have a way to run what is currently checked out without having to install it.

@scopatz, do you have any lingering issues with this?

scopatz:
Well, I guess my concern is that this is still redundant. If I run `python -m xonsh` in the root repo directory, it will load the `xonsh/__main__.py` file from the repo, and startup xonsh.  Worst case scenario, I would need to run `env PYTHONPATH=. python -m xonsh`. Does that not work for you two for some reason?

adqm:
right.  but if you just run `python main.py` from the repo, that should load the local copies.

scopatz:
So it is no different from `python -m xonsh`, then?

scopatz:
Why not just softlink `__main__.py`?

adqm:
Ah, yes.  You're right.  Sorry I misread.  In that case, there is no need for the new `main.py`.

scopatz:
Maybe we could add a note about this to the developer docs instead?

mitnk:
Yes, `python -m xonsh` in the repo works for me :) Thanks @scopatz, @adqm ! Closing this PR then.

scopatz:
Great! Feel free to put something in the docs about this then :)

Issue #1176, opened 2016-06-08
~ should be interpreted inside env variables


Cheaterman:
Hello,

Today I tried $PATH.insert(0, '~/somedir') - it fails for reasons that are now obvious, but scopatz said the following on IRC:

`<scopatz> Cheaterman: Well the PATH should just be a dumb list of strings.  However, things that use $PATH should do user and envvar expansion when they use them.  It looks like this is not happening.`

So I suppose this should be considered a bug - xonsh will be even more comfy this way too ^__^

Thanks in advance!

scopatz:
Thanks for reporting this @Cheaterman

laerus:
this issue is fixed!

PR #1214, opened 2016-06-12
Replaced PATH with a custom class that expands paths

VHarisop:
In response to #1176: A custom class called `EnvPath` that subclasses `list` is used for `$PATH` in order to avoid boilerplate code that checks and expands env.vars in lots of different places.

I'm almost certain a much more simple solution exists, and I'm not even sure this is a proper way to subclass `list` :P, but that's all I could come up with for now.


xonsh/tools.py
```diff
@@ -92,6 +92,44 @@ def __init__(self, returncode, command, output=None, stderr=None,
         self.completed_command = completed_command
 
 
+class EnvPath(list):
```
scopatz: instewd of list, this should be collections.MutableSequence

xonsh/tools.py
```diff
@@ -92,6 +92,44 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+                return os.path.expanduser(expandvars(path))
+            else:
+                # if the flag has not been set, do not perform any expansion
+                return path
+        else:
+            # use os.path.expandvars instead if env is not yet available
+            return os.path.expanduser(os.path.expandvars(path))
+
+    def __init__(self, items):
+        items = map(EnvPath._expandpath, items)
```
scopatz: so, I don't think that this should expand on init, but only on get.

xonsh/tools.py
```diff
@@ -92,6 +92,50 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+        super(EnvPath, self).insert(index, EnvPath._expandpath(item))
+
+    def extend(self, iterable):
+        super(EnvPath, self).extend(map(EnvPath._expandpath, iterable))
+
+    def __getitem__(self, item):
+        super(EnvPath, self).__getitem__(EnvPath._expandpath(item))
+
+    def __setitem__(self, index, item):
+        super(EnvPath, self).__setitem__(index, EnvPath._expandpath(item))
```
scopatz: sane here, this should not expand

xonsh/tools.py
```diff
@@ -92,6 +92,50 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+        super(EnvPath, self).__init__(items)
+
+    def append(self, item):
+        super(EnvPath, self).append(EnvPath._expandpath(item))
+
+    def insert(self, index, item):
+        super(EnvPath, self).insert(index, EnvPath._expandpath(item))
+
+    def extend(self, iterable):
+        super(EnvPath, self).extend(map(EnvPath._expandpath, iterable))
```
scopatz: I think you can remove the above three with mutable sequence

scopatz:
thanks for putting this in! it would be good to knoe when bash does variable expansion. If it is on setting or getting and then we should copy that.

VHarisop:
Thanks for the feedback, @scopatz ! I *think* that bash performs the expansion when setting, but I'm not sure about what happens behind the scenes:

```shell
$ export PATH = $PATH:~/Documents
$ echo $PATH
# [...]:/home/vharisop/Documents
```

adqm:
I like this a lot!  Seems like it will solve multiple problems.  Any way we can add some tests to make sure it is working?

adqm:
There was also a suggestion in the IRC room to accept `pathlib.Path` objects in `$PATH`, and to auto-convert them to strings.  I like this idea.  There are a couple of ways we could handle it:

* When adding an item, we check to make sure it is a `str`, `bytes`, or `pathlib.Path`, and when getting items out, convert to string and expandpath
* When getting an item out, convert it to a string no matter what.

Thoughts?

VHarisop:
Hey there @adqm!

I think I could extend this to work with `pathlib.Path` and `byte` objects without a lot of extra effort.
I would add the items as they are (after type checking), and convert them to strings using `bytes.decode()` and `str(pathlib.Path)` during the call to `__getitem__`, expanding them afterwards.

I think I would also have to change the `is_env_path` validator to accept bytes and `pathlib.Path` as well, if I'm not mistaken.

As for adding tests, that is fine by me. I was planning on adding some later anyway if we decided that it's ok to go with this approach. It would be great if someone could suggest testcases using Windows paths, since I'm only able to experiment in a GNU/Linux environment.

adqm:
@VHarisop, thanks for adding that.  I think we should always have `__xonsh_env__` when any user code is running, no?  We have it available when  the RC file is run, at least, and we should have it at all points after that, too, I think.  Though I could be missing something?

VHarisop:
@adqm I think you are right. In my previous commit, there was a call to `_expandpath` in `__setitem__` when the environment was created. `expandvars` would be called next and try to retrieve `__xonsh_env__`, which was being initialized at that point and not yet available, that's why I added that check.

Now that `_expandpath` is only called in `__getitem__` I too think that we can safely assume that `__xonsh_env__` will be available.

xonsh/tools.py
```diff
@@ -92,6 +93,85 @@ def __init__(self, returncode, command, output=None, stderr=None,
         self.completed_command = completed_command
 
 
+class EnvPath(MutableSequence):
```
scopatz: Shouldn't this be in `environ.py` since it is strictly an environment thing.

VHarisop: This would create a cycle of dependencies, since `str_to_env_path` (which is in xonsh.tools) should return an `EnvPath` object.

VHarisop:
Glad to help, @scopatz ! I don't understand your last question, though. Do you mean that I should add testcases that involve `bytes` objects, or that you can't see any reason for allowing `bytes` objects to be added in `$PATH`?

xonsh/tools.py
scopatz: Ahh I see. OK.

scopatz:
>  that you can't see any reason for allowing bytes objects to be added in $PATH?

This one :)

adqm:
I guess my view was: I don't see any reason _not_ to allow bytes objects to be added.  If someone is adding the result of some other function call, for example, that might come back as a `bytes`, and I think it's better for us to do the conversion ourselves rather than expecting them to call `.decode()`

xonsh/tools.py
```diff
@@ -92,6 +93,85 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+
+    @staticmethod
+    def _decode_bytes(path):
+        """
+        Tries to decode a path in bytes using XONSH_ENCODING if available,
+        otherwise using sys.getdefaultencoding().
+        """
+        env = getattr(builtins, '__xonsh_env__', os.environ)
+        enc = env.get('XONSH_ENCODING', sys.getdefaultencoding())
+        return path.decode(encoding=enc)
```
scopatz: You should also pass in `XONSH_ENCODING_ERRORS` as `errors`

xonsh/tools.py
```diff
@@ -92,6 +93,85 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+        """
+        env = getattr(builtins, '__xonsh_env__', os.environ)
+        if env.get('EXPAND_ENV_VARS', False):
+            # expand variables and use os.path.realpath to handle cases
+            # with relative paths like ../ or ./
+            path = os.path.realpath(os.path.expanduser(expandvars(path)))
+        return path
+
+    @staticmethod
+    def _decode_bytes(path):
```
scopatz: These two methods probably should be at global scope, rather than being staticmethods. They don't deal with the class at all.

xonsh/tools.py
```diff
@@ -92,6 +93,85 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+        if len(args) > 0:
+            if isinstance(args[0], str):
+                self._d = args[0].split(os.pathsep)
+            elif isinstance(args[0], pathlib.Path):
+                self._d = [args[0]]
+            elif isinstance(args[0], bytes):
+                # decode bytes to a string and then split based on
+                # the default path separator
+                self._d = EnvPath._decode_bytes(args[0]).split(os.pathsep)
+            elif isinstance(args[0], list):
```
scopatz: This should be Sequence, not list.  What if the user passes in a tuple?

xonsh/tools.py
```diff
@@ -92,6 +93,85 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+        """
+        Tries to decode a path in bytes using XONSH_ENCODING if available,
+        otherwise using sys.getdefaultencoding().
+        """
+        env = getattr(builtins, '__xonsh_env__', os.environ)
+        enc = env.get('XONSH_ENCODING', sys.getdefaultencoding())
+        return path.decode(encoding=enc)
+
+    def __init__(self, *args):
+        self._d = []
```
scopatz: Why `_d`?  This is a list not a dict....

scopatz:
> I guess my view was: I don't see any reason not to allow bytes objects to be added. If someone is adding the result of some other function call, for example, that might come back as a bytes, and I think it's better for us to do the conversion ourselves rather than expecting them to call .decode()

Fair enough :)

xonsh/tools.py
```diff
@@ -92,6 +93,85 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+    def __len__(self):
+        return len(self._d)
+
+    def __delitem__(self, key):
+        self._d.__delitem__(key)
+
+    def insert(self, index, value):
+        self._d.insert(index, value)
+
+    def get_paths(self):
```
scopatz: Maybe this should be a property that is just called `paths`

xonsh/tools.py
```diff
@@ -92,6 +93,85 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+    def __eq__(self, other):
+        # Compare using sets to account for different
+        # order in items as well as unorderable lists
+        # (e.g. containing Path() and str())
+        if isinstance(other, list):
+            return set(self._d) == set(other)
+        elif isinstance(other, EnvPath):
+            return set(self) == set(other)
+        else:
+            return False
```
scopatz: With the MutableMapping interface this doesn't need to be defined.

xonsh/tools.py
```diff
@@ -92,6 +93,85 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+            if isinstance(args[0], str):
+                self._d = args[0].split(os.pathsep)
+            elif isinstance(args[0], pathlib.Path):
+                self._d = [args[0]]
+            elif isinstance(args[0], bytes):
+                # decode bytes to a string and then split based on
+                # the default path separator
+                self._d = EnvPath._decode_bytes(args[0]).split(os.pathsep)
+            elif isinstance(args[0], list):
+                self._d = list(args[0])
```
scopatz: There should be an else-block that raises an error.

VHarisop: @scopatz What's the best type of exception to raise in a constructor? I'm thinking either `ValueError` or `TypeError`.

scopatz: This would probably be a TypeError, since it failed to type check correctly

xonsh/tools.py
VHarisop: @scopatz But this is a MutableSequence, not a MutableMapping (and I don't think it should be).

Removing the `__eq__` method causes some tests to fail (`tests.test_environ.test_env_path` and `test.test_tools.test_str_to_env_path`) as an `EnvPath` object is compared with a list for equality. I think that this method should stay (and be extended to work with any `Sequence`, not just lists).

scopatz: Yes, MutableSequence sorry.  Also, though, you should not be testing set equality for sequences.  The failing tests were probably legitimate with the added functionality. My intuition tells me you shouldn't have to implement this.

VHarisop: That would result in two `EnvPath` objects that contain the same entries in different order to be considered unequal. Do we want that?
Now that I think of it, would we have to compare two `EnvPath` objects at any case? I cannot seem to think of such a scenario, so you might be right after all.

scopatz: Yes, two paths that have different ordering are definitely different.

xonsh/tools.py
```diff
@@ -92,6 +93,85 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+                                      for i in args[0]):
+                    # make TypeError's message as informative as possible
+                    # when given an invalid initialization sequence
+                    raise TypeError(
+                            "EnvPath's initialization sequence should only "
+                            "contain str, bytes and pathlib.Path entries")
+                self._l = list(args[0])
+            else:
+                raise TypeError('EnvPath cannot be initialized with items '
+                                'of type %s' % type(args[0]))
```
scopatz: Shouldn't these type checks happen for every element, not just the fist one?

xonsh/tools.py
```diff
@@ -92,6 +93,85 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+    return os.path.realpath(path)
+
+
+def decode_bytes(path):
+    """
+    Tries to decode a path in bytes using XONSH_ENCODING if available,
+    otherwise using sys.getdefaultencoding().
+    """
+    env = getattr(builtins, '__xonsh_env__', os.environ)
+    enc = env.get('XONSH_ENCODING', sys.getdefaultencoding())
```
scopatz: No need to call `sys.getdefaultencoding()` this is what `xonsh.platform.DEFAULT_ENCODING` is for: https://github.com/xonsh/xonsh/blob/master/xonsh/platform.py#L116

xonsh/tools.py
```diff
@@ -92,6 +93,85 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+    """
+    Performs environment variable / user expansion on a given path
+    if the relevant flag has been set.
+    """
+    env = getattr(builtins, '__xonsh_env__', os.environ)
+    if env.get('EXPAND_ENV_VARS', False):
+        # expand variables and use os.path.realpath to handle cases
+        # with relative paths like ../ or ./
+        path = os.path.expanduser(expandvars(path))
+    return os.path.realpath(path)
```
scopatz: I don't think we should call realpath() here.  If there is a symbolic link on the path it is OK.

xonsh/tools.py
VHarisop: Whoops! You're right, I totally forgot about the case where `EnvPath` is called with more than one arguments. I suppose if a call like

```
EnvPath([path1, path2], [path3, path4])
```

occurs at somepoint this should produce an `EnvPath` containing `[path1, path2, path3, path4]`, right?

scopatz: Well, I think that this should probably work like list does and only take a single iterable rather than `*args`. That will simplify this a bunch :)

xonsh/tools.py
```diff
@@ -92,6 +93,85 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+                    raise TypeError(
+                            "EnvPath's initialization sequence should only "
+                            "contain str, bytes and pathlib.Path entries")
+                self._l = list(args[0])
+            else:
+                raise TypeError('EnvPath cannot be initialized with items '
+                                'of type %s' % type(args[0]))
+
+    def __getitem__(self, item):
+        return expandpath(self._l[item])
```
scopatz: Just double checked, bash seems to apply the expansion on getting (rather than setting) so I think that this is correct.

xonsh/tools.py
VHarisop: Alright. I think we should at least call `abspath()`, which does not resolve symlinks, in order to handle relative paths.

scopatz: Sure, that seems reasonable.

VHarisop:
Thanks for the feedback (and patience) @scopatz!

You might notice I changed `is_env_path` to not accept simple string lists, in order to avoid initializing path variables with lists of strings that were not previously expanded. I also refactored some of the tests in order to reflect the current state of how `$PATH` stuff works.

xonsh/tools.py
```diff
@@ -675,18 +757,15 @@ def ensure_string(x):
 
 def is_env_path(x):
     """This tests if something is an environment path, ie a list of strings."""
-    if isinstance(x, str):
-        return False
-    else:
-        return (isinstance(x, Sequence) and
-                all(isinstance(a, str) for a in x))
+    return False if isinstance(x, (str, bytes)) else isinstance(x, EnvPath)
```
scopatz: I think this should just be `return isinstance(x, EnvPath)`

xonsh/tools.py
```diff
@@ -92,6 +93,87 @@ def __init__(self, returncode, command, output=None, stderr=None,
...
+        else:
+            if isinstance(args, str):
+                self._l = args.split(os.pathsep)
+            elif isinstance(args, pathlib.Path):
+                self._l = [args]
+            elif isinstance(args, bytes):
+                # decode bytes to a string and then split based on
+                # the default path separator
+                self._l = decode_bytes(args).split(os.pathsep)
+            elif isinstance(args, Sequence):
```
scopatz: This should probably be Iterable instead of Sequence.

scopatz:
> Thanks for your feedback (and patience) @scopatz!

No problem! Thanks for bearing with us

VHarisop:
@scopatz This is probably ready for another review, not much else to be done here (or at least I think so :) ).

scopatz:
This looks good to me @VHarisop!  There seem to be some conflicts with master. Could you fix those up please? Also, would you mind please adding a news file for this branch?   See the instructions here: https://github.com/xonsh/xonsh/blob/master/CONTRIBUTING.rst#changelog

scopatz:
Thanks again!

VHarisop:
I fixed the conflicts and added the `news/` entry, so hopefully this is ok now. Thanks for your time!

scopatz:
Thanks a million @VHarisop! This is awesome

melund:
Help @VHarisop @scopatz 

This broke something on Windows. `which`  now suddenly finds stuff that is not executable.
We should have the windows tests back :(

After a lot of debuggin I found that the `PATHEXT` variable is suddenly wrapped with the PathEnv class even though it is not really a path like env-variable. It contains extensions which should be considered executable.

`['.COM', '.EXE', '.BAT']`
Suddenly becomes
`['C:\\Users\\mel\\Documents\\xonsh\\.COM', 'C:\\Users\\mel\\Documents\\xonsh\\.EXE', 'C:\\Users\\mel\\Documents\\xonsh\\.BAT']`

How do I ensure that `PATHEXT` is not beeing wrapped by this?

VHarisop:
@melund you are right, `PATHEXT` is wrapped because it uses `is_env_path` and `to_env_path` for validation and conversion (see `DEFAULT_ENSURERS` in `environ.py`).

I have practically no experience with Windows, but I guess that a temporary fix would be changing the `Ensurer` validation and conversion functions for `PATHEXT` to what they were before the `path_expand` patch.

I'm really sorry for the trouble :(. I'll try and look into it the next few days.

melund:
@VHarishop. I see. We kind of need the old functionality back for things that are list but not path-like.

`is_list, str_to_list, list_to_str`

Then these ensures should be used on `PATHEXT`

PR #1215, opened 2016-06-12
Fix locate_binary assuming path is a list

melund:
This fixes crash when path is tuple and not list.


gforsyth:
LGTM.  Thanks, @melund!

PR #1216, opened 2016-06-12
Do not throw an exception on 'completer list' if a completer has no docstring

pgoelz:
Solves a current bug:

> $ test = lambda a, b, c, d, e: set()

> $ completer add testname test

> $ completer list
>Traceback (most recent call last):
>  File ".../xonsh/0.3.4/libexec/lib/python3.5/site-packages/xonsh/proc.py", line 335, in wrapped_simple_command
>    r = f(args, i)
>  File ".../xonsh/0.3.4/libexec/lib/python3.5/site-packages/xonsh/completers/_aliases.py", line 114, in completer_alias
>    return func(args[1:], stdin=stdin)
>  File ".../xonsh/0.3.4/libexec/lib/python3.5/site-packages/xonsh/completers/_aliases.py", line 42, in _list_completers
>    doc = ' '.join(_comp[c].__doc__.split()) or 'No description provided'
AttributeError: 'NoneType' object has no attribute 'split'

The problem was that the original code called split on the completer function's docstring, which might be None.
Previously, an all-whitespace docstring would have triggered the 'No description provided' label, this is no longer the case. If this was intentional, we can easily bring it back.

### DIFF
--- a/news/completer_list_bugs.rst
+++ b/news/completer_list_bugs.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+* Fixed ``_list_completers`` such that it does not throw a ValueError if no completer is registered.
+* Fixed ``_list_completers`` such that it does not throw an AttributeError if a completer has no docstring.
+
+**Security:** None
--- a/xonsh/completers/_aliases.py
+++ b/xonsh/completers/_aliases.py
@@ -6,6 +6,7 @@ from xonsh.completers.tools import justify
 
 VALID_ACTIONS = frozenset({'add', 'remove', 'list'})
 
+
 def _add_one_completer(name, func, loc='end'):
     new = OrderedDict()
     if loc == 'start':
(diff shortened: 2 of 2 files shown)


gforsyth:
Thanks for putting this in @pgoelz!  This seems like a reasonable change to me, but I'll defer to @adqm who wrote the completion framework to make sure this doesn't break some assumptions I'm not aware of.  

Could you also make an addition to the CHANGELOG for this?  in `xonsh/news` you'll find a file `TEMPLATE.rst` that you can make a copy of, rename and then fill in the appropriate section (the changelog gets compiled from these aggregate files and helps to avoid a bunch of merge conflicts)

pgoelz:
Thanks, I've tried to follow your suggestion.

By the way, is it expected that `max(i for i in [], default=0)` works in xonsh, but in python 3.5 throws `SyntaxError: Generator expression must be parenthesized if not sole argument`?

adqm:
Good catch, @pgoelz, and thanks for the fix!  I'll merge in now.

> By the way, is it expected that max(i for i in [], default=0) works in xonsh, but in python 3.5 throws SyntaxError: Generator expression must be parenthesized if not sole argument?

I think that is actually a bug in xonsh's parser.  Do you mind submitting an issue for that?

pgoelz:
I'll do, thanks.

PR #1220, opened 2016-06-13
Amalgamate Source

scopatz:
So this is an idea that I wanted to get some feedback on that has the possibility of speeding up startup at low / no overhead at run time.

The idea here is to glue most of the source files in a package or subpackage together into a single module `__amalgam__.py`.  Combined with some hooks in the `__init__.py`, the should dramatically reduce the number of files that are being searched for inside of the package.

The top-level amalgamate.py script automatically creates the `__amalgamate__.py` file for a package. This will create and solve a dependency graph of all modules in the package. It then will go through each module and glue it onto the `__amalgamate__.py`, removing duplicate imports as it goes.

This has some funny side effects such as, 

1. All of the modules that are amalgamated share the same globals
2. debugging most things looks like code comes from `__amalgam__`.

I can add the ability to turn off using `__amalgam__` using an envvar.

Sadly, this actually *slows down* my startup times by about 0.1 - 0.12 sec. I think that this is for a couple of reasons.  The first is that xonsh imports ~750 modules, and < 40 of them are in the base `xonsh` package.  Amalgamating just these would have been a 5% effect anyway.  Secondly, none of these other imports are lazy.  Thus, when we glue everything together we end up importing more modules overall.  Thirdly, most of xonsh isn't written to know that it will share a common globals, and so all global operations are performed no matter what.

In any event, I would be interested to know if this speeds things up for anyone else.  Also, before doing too much more, I want to know if this is an avenue worth continuing on or if people think it is too weird. (Also, yes, I know the tests are failing ~_~)

I think that this could potentially be much faster. The amalgamate script could also rewrite all (remaining) imports to be lazy, which would mean that we wouldn't have to worry about it day-to-day.

Feedback seriously appreciated!


gforsyth:
Hey @scopatz -- this is a cool idea.  But yeah, it definitely slows things down on my end.  I lose an extra .25s to startup.

scopatz:
@gforsyth - dang!  I am going to have to try the auto-lazy-loader idea.

scopatz:
Ok, I think this is ready for another look.  Not only do all of the tests pass, but login-interactive shells are consistently as fast as headless shells on this branch. And I know it is nothing to write home about, but it also seems to be 0.02 - 0.03 seconds faster on my desktop.  Hopefully other people will see larger improvements.

Note that I have made most imports lazy and added lazy-and-self-destructive objects for lazily loading data.  The canonical example here are regular expressions, which are expensive to compile - so you compile them globally, rather than per function call, which then makes the module expensive to import. Now we have a solution.

There is still about 0.26 sec. that I can't track down. But, I think that this is a big enough change that the earlier it goes in the better.

scopatz:
Also, more laziness can continue to be added as we go.

santagada:
my only problem is that all this might make using a tool like freeze or any other forms of preloading much more difficult. can this wait a bit until other options are evaluated? (I tried to use freeze but after multiple problems I just stopped)

xonsh/platform.py
```diff
@@ -157,106 +176,117 @@ def scandir(path):
...
-        os.path.join(GIT_FOR_WINDOWS_PATH,
-                     'usr\\share\\bash-completion\\completions'),
-        os.path.join(GIT_FOR_WINDOWS_PATH,
+def _bcd():
+    """A possibly empty tuple with default paths to Bash completions known for
+    the current platform.
+    """
+    if ON_LINUX or ON_CYGWIN:
+        if linux_distro() == 'arch':
+            BASH_COMPLETIONS_DEFAULT = (
```
gforsyth: if user is on arch then `bcd` is undefined when we get to the return statement

gforsyth:
Hey @scopatz -- this is looking pretty neat!  There's the one error for Arch users that's a simple fix.  I also get errors on startup that I've narrowed down to having `${...}` in my `xonshrc`.  That said, if I switch it to `builtins.__xonsh_env__` then it works fine.

scopatz:
OMG I figured it out!  It was freaking setuptools and its console script shenanigans!  Setuptools entrypoints [validates its dependencies](https://www.reddit.com/r/Python/comments/4auozx/setuptools_deployment_being_very_slow_compared_to/) prior to execution.  [Of course they don't tell you this in the docs](http://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point) and since we have no hard dependencies it is pointless to validate.  Also they don't tell you that this add about 150 ms of unavoidable execution time to the start of the command - every time. This seems like the last thing that a command line utility would ever want....

In any event, swapping this out for the old way, my headless startup times are now between 0.09 and 0.11 sec. Which is great!  I also added a `xon.sh` script to close #266 and #565.  I am confident more speed ups are possible with increased laziness.

Thanks for pointing out the arch error @gforsyth. I have fixed that.

> my only problem is that all this might make using a tool like freeze or any other forms of preloading much more difficult. can this wait a bit until other options are evaluated? (I tried to use freeze but after multiple problems I just stopped)

@santagada This shouldn't affect freeze at all because freeze operates on bytecode, which this produces in the same way as any other Python module.  Also, freeze in my opinion isn't a general solutions (for developers and users) because it creates a whole new Python distributions. So it can be a great thing for users who only want the CLI.  At this point, I think it is the package managers choice in how to bundle the application.

As per freeze not working, we are happy to make changes to support it if any specific ones come up.

scopatz:
Also @santagada - this is entirely opt-out via an environment variable.  Users don't have to use the amalgamation.  It is just an option.

scopatz:
Added some docs. I think this is good to go!

gforsyth:
@scopatz -- for some reason I can't run this anymore.  I get 

```console
xonsh
  File "/home/gil/anaconda/bin/xonsh", line 9, in <module>
    load_entry_point('xonsh==0.3.4', 'console_scripts', 'xonsh')()
  File "/home/gil/anaconda/lib/python3.5/site-packages/pkg_resources/__init__.py", line 542, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/home/gil/anaconda/lib/python3.5/site-packages/pkg_resources/__init__.py", line 2568, in load_entry_point
    raise ImportError("Entry point %r not found" % ((group, name),))
ImportError: Entry point ('console_scripts', 'xonsh') not found
```

gforsyth:
@scopatz -- scratch that, just fixed it.  had to nuke my `build` directory for some reason after switching between branches a whole bunch

amalgamate.py
```diff
@@ -0,0 +1,384 @@
...
+import pprint
+from itertools import repeat
+from collections import namedtuple
+from collections.abc import Mapping
+from ast import parse, walk, literal_eval, Import, ImportFrom
+
+ModNode = namedtuple('ModNode', ['name', 'pkgdeps', 'extdeps'])
+ModNode.__doc__ = """Module node for dependency graph.
+
+Attrributes
```
gforsyth: *Attributes

amalgamate.py
```diff
@@ -0,0 +1,384 @@
...
+-----------
+name : str
+    Module name.
+pkgdeps : frozenset of str
+    Module dependencies in the same package.
+extdeps : frozenset of str
+    External module dependencies from outside of the package.
+"""
+
```
gforsyth: PEP8 two lines

amalgamate.py
```diff
@@ -0,0 +1,384 @@
...
+                pkgdeps.update(n.name for n in a.names if n.name in allowed)
+            elif a.module.startswith(pkgdot):
+                p, dot, m = a.module.rpartition('.')
+                if p == pkg and m in allowed:
+                    pkgdeps.add(m)
+                else:
+                    extdeps.add(a.module)
+    return ModNode(name, frozenset(pkgdeps), frozenset(extdeps))
+
+
```
gforsyth: pep8 too many lines

amalgamate.py
```diff
@@ -0,0 +1,384 @@
...
+                    msg = ('Cannot amalgamate almagate import of '
+                           'amalgamated module:\n\n  import {0}.{1}\n'
+                           '\nin {0}/{2}.py').format(pkg, n.name, name)
+                    raise RuntimeError(msg)
+                imp = (Import, n.name, n.asname)
+                if imp not in imps:
+                    imps.add(imp)
+                    keep.append(imp)
+            if len(keep) == len(a.names):
+                #continue  # all new imports
```
gforsyth: these should be cleaned up, yeah?

amalgamate.py
```diff
@@ -0,0 +1,384 @@
...
+                # all new imports
+                s = format_lazy_import(keep)
+            elif len(keep) == 0:
+                s = ', '.join(n.name for n in  a.names)
+                s = '# amalgamated ' + s + '\n'
+            else:
+                #s = format_import(keep)
```
gforsyth: cleanup comment

amalgamate.py
```diff
@@ -0,0 +1,384 @@
...
+                keep = []
+                for n in a.names:
+                    imp = (ImportFrom, a.module, n.name, n.asname)
+                    if imp not in imps:
+                        imps.add(imp)
+                        keep.append(imp)
+                if len(keep) == len(a.names):
+                    continue  # all new imports
+                elif len(keep) == 0:
+                    s = ', '.join(n.name for n in  a.names)
```
gforsyth: pep8 2 spaces after `in`

amalgamate.py
```diff
@@ -0,0 +1,384 @@
...
+                replacements.append((start, stop, s))
+    # apply replacements in reverse
+    lines = raw.splitlines(keepends=True)
+    for start, stop, s in replacements[::-1]:
+        lines[start] = s
+        for i in range(stop - start - 1):
+            del lines[start+1]
+    return ''.join(lines)
+
+
```
gforsyth: pep8 too many lines

scopatz:
@gforsyth - yeah I have had to destroy my build dir a few times :)

xonsh/execer.py
```diff
@@ -8,7 +8,7 @@
 import warnings
 from collections import Mapping
 
-from xonsh import ast
+from xonsh.ast import CtxAwareTransformer
```
gforsyth: In most places you've changed the import to be more explicit -- any reason why that's flipped here?

amalgamate.py
scopatz: yeah, whoops :)

xonsh/jobs.py
```diff
@@ -105,21 +106,36 @@ def _set_pgrp(info):
...
-        _libc = ctypes.CDLL('cygwin1.dll')
+        _libc = LazyObject(lambda: ctypes.CDLL('cygwin1.dll'),
+                           globals(), '_libc')
 
         # on cygwin, signal.pthread_sigmask does not exist in Python, even
         # though pthread_sigmask is defined in the kernel.  thus, we use
         # ctypes to mimic the calls in the "normal" version below.
         def _give_terminal_to(pgid):
-            if _shell_tty is not None and os.isatty(_shell_tty):
+            shtty = _shell_tty()
```
gforsyth: I will not say this phonetically.  I will not say this phonetically.  I will not say this phonetica....

xonsh/main.py
```diff
@@ -1,18 +1,33 @@
 # -*- coding: utf-8 -*-
 """The main xonsh script."""
+import time
+import builtins
+
+#_oi = __import__
```
gforsyth: I imagine we can also remove these?

xonsh/platform.py
```diff
@@ -91,23 +104,26 @@ def ptk_version_info():
...
     """Checks if readline is available to import."""
-    try:
-        import readline
-    except:  # pyreadline will sometimes fail in strange ways
-        return False
-    else:
-        return True
+    spec = importlib.util.find_spec('readline')
+    return (spec is not None)
+    # might need to return to this at some point.
```
gforsyth: more commented code here

xonsh/jobs.py
scopatz: hahah - I know now that Freud was sometime right; there were occasional slips.

xonsh/tracer.py
```diff
@@ -83,7 +88,12 @@ def trace(self, frame, event, arg):
         return self.trace
 
 
-tracer = TracerType()
+_tracer = None #TracerType()
```
gforsyth: old code

xonsh/tracer.py
```diff
@@ -188,12 +198,12 @@ def _create_parser():
...
+def tracermain(args=None):
     """Main function for tracer command-line interface."""
     parser = _create_parser()
     ns = parser.parse_args(args)
     return _MAIN_ACTIONS[ns.action](ns, args)
 
 
-if __name__ == '__main__':
-    main()
+#if __name__ == '__main__':
```
gforsyth: tracing seems to be nonfunctional -- I imagine this is the culprit?

xonsh/execer.py
scopatz: Yes, so this is actually correct.  The way the code amalgamater works is the other moduled that are in the *same* package (and amalgamated) should be `from pkg.x import a, c, d`. This is because the amalgamater puts all such modules in the same globals(), which is effectively what the from-imports do.

So since `xonsh.ast` and `xonsh.execer` are both in the same poackage (`xonsh`) they should have from-imports. 

Alternatively, for modules outside of the current package (or modules that are not amalgamated) the import for be either `import pkg.x` or `import pkg.x as name`. This is because these are the only cases where automatically inserting lazy imports in guarantted to be safe. This is because `from pkg.x import name` may import a variable that cannot be lazily constructed.

scopatz: So *most* of the modifications were to the second case.  We already happened to be coding in the style for the same package imports :)

scopatz: This should probably get added to the dev docs.

melund:
It is very much faster on load for me :) 

I do get an odd warning, which I don't get on master. 
```
C:\Users\mel>xonsh
C:\Users\mel\Anaconda3\lib\site-packages\xonsh\__amalgam__.py:11801: RuntimeWarning: error running xonsh run control file 'C:\\Users\\mel/.xonshrc': expected string or bytes-like object
  warn(msg.format(rcfile, err), RuntimeWarning)
snail@sea ~ $
```

gforsyth:
@melund -- do you have a `${...}` in your `xonshrc`?  I had that same error and the `${...}` was the culprit

melund:
Ahh. That fixed it. Thanks. 

When I tried to debug I got another error: 

```
C:\Users\mel>set XONSH_DEBUG=1

C:\Users\mel>xonsh
  File "C:\Users\mel\Anaconda3\Scripts\xonsh-script.py", line 5, in <module>
    sys.exit(xonsh.main.main())
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\main.py", line 226, in main
    args = premain(argv)
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\main.py", line 213, in premain
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\shell.py", line 44, in __init__
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\shell.py", line 84, in _init_environ
    self.execer = Execer(config=config, login=self.login, xonsh_ctx=self.ctx)
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\execer.py", line 43, in __init__
    load_builtins(execer=self, config=config, login=login, ctx=xonsh_ctx)
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\built_ins.py", line 659, in load_builtins
    builtins.__xonsh_env__ = ENV = Env(default_env(config=config, login=login))
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\environ.py", line 620, in __init__
    self[key] = val
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\environ.py", line 753, in __setitem__
    val = ensurer.convert(val)
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\environ.py", line 73, in to_debug
    builtins.__xonsh_execer__.debug_level = val
AttributeError: module 'builtins' has no attribute '__xonsh_execer__'
```

gforsyth:
Yep.   I can confirm that, too

melund:
It is down to 0.4 seconds on when timing a command. 
Seems to be a little longer when it actually starts up. 

```
C:\Users\mel>timethis xonsh -i -c print(42)

TimeThis :  Command Line :  xonsh -i -c print(42)
TimeThis :    Start Time :  Thu Jun 16 21:56:17 2016

42

TimeThis :  Command Line :  xonsh -i -c print(42)
TimeThis :    Start Time :  Thu Jun 16 21:56:17 2016
TimeThis :      End Time :  Thu Jun 16 21:56:17 2016
TimeThis :  Elapsed Time :  00:00:00.408
```

melund:
This is really dark wizardry to me :)   Nice jobs Scopatz. Where did you get the idea for this?

scopatz:
Ok, I have fixed these issues that you brought up, added some dev doc guidance on the imports, and otherwise cleaned up the code base.

> This is really dark wizardry to me :) Nice jobs @scopatz. Where did you get the idea for this?

Oh man, @melund, you *do not* want to know :).  But basically I read everything out there on how to speed up Python imports, obsessed over it for a week, and then decided to implement it all as a syntax tree transformation so we could write mostly normal Python without having to worry about it.  That accounts for the wizardry.

The darkness comes from a project called [JsonCpp](https://github.com/open-source-parsers/jsoncpp), which I may or may not have [written Python bindings for](https://github.com/scopatz/pyjsoncpp) (which remains my only project with 3 programming language names ~_~).  You'll note that vanilla JsonCpp has an [amalgamate script](https://github.com/open-source-parsers/jsoncpp/blob/master/amalgamate.py), that glues the whole project into a single header and single source file.  This is an amazing idea.  The kicker is that JsonCpp's amalgamate is written in Python :).

And this is why we can have nice things :)

melund:
It solved the problems I had. I will go ahead and merge it in.

Especially, the 'readline'-mode starts really fast now. Prompt-toolkit feels a little slower. Maybe rendering the prompt that is slow.

melund:
Woohoo ;)   I think once we have fixed some of the glaring current issues, we should really make a new release

scopatz:
I agree, I think that this should be v0.4 even.  There is enough that is new.

melund:
Yes. Would it make sense to suggest amalgamation for prompt-toolkit as well? Could it speed things up further?

scopatz:
Yeah, I think we should start amalgamating the subpackages too.  `parsers` might be a little annoying with how it works now, it should still be done :).

scopatz:
It also might be worth it to encourage @jonathanslenders to take our amalgamate script and use it on PTK itself to speed up the PTK imports.

scopatz:
I think a big win might be in the keybindings, which aren't needed until they are used.

jonathanslenders:
It's an interesting approach that you have! Where do I find the amalgamate script?
I'd like to try. Not sure yet I wil use it, but I like the idea of doing an ast transformation to optimize. (That's what it does, right?)

scopatz:
The amalgamate script is here: https://github.com/xonsh/xonsh/blob/master/amalgamate.py

And yes, it does an AST transformation to optimize imports, optionally, but on by deafult.  We also have some lazy data structures that you should fee free to use too! https://github.com/xonsh/xonsh/blob/master/xonsh/lazyasd.py

We have the amalgamater run when [setup.py is called](https://github.com/xonsh/xonsh/blob/master/setup.py#L59). It also needs the following lines in the [`__init__.py`](https://github.com/xonsh/xonsh/blob/master/xonsh/__init__.py) to hook into the package:

```
# amalgamate exclude
# amalgamate end
```

Happy to help, if you want to explore this option!

scopatz:
We could also break out the amalgamate.py script into its own repo / project and then simply have a copy in our repo if you think that would be useful too

gforsyth:
> We could also break out the amalgamate.py script into its own repo / project and then simply have a copy in our repo if you think that would be useful too

I think this is a good idea.  A bunch of project might find it useful

scopatz:
Hey @jonathanslenders @gforsyth - I have now created a standalone package for amalgamate, with hopefully enough docs for people to use independently https://pypi.python.org/pypi/amalgamate

Please let me know if you need any help with this.  PTK is now our slowest import at over 200 ms, when we use it. I imagine that this affects other projects as well.

scopatz:
Note that ~150 of this comes from Pygments, which if you could make a lazy import, would be awesome

scopatz:
with amalgam.

PR #1222, opened 2016-06-13
possible fix for incorrect alias loading

adqm:
There was some discussion in the IRC channel recently about aliases loading incorrectly (see https://gitter.im/xonsh/xonsh?at=575eef6a6092456f6633c518).  I believe this should fix that issue.


gforsyth:
That fixed it, @adqm.  Once Travis finishes up this is good to go.

Issue #1141, opened 2016-06-06
classic shell like stream handling + sub shells


mburo:
In a recent thread there was a question how to mimic bash code like this:

`(echo "abc"; sleep 1; echo "def") | grep abc`

@adqm answered:

"Unfortunately, we don't support the "subshell" idea yet, and so it is hard to group those commands together. It is also not yet possible to use IO redirection to one of those Python objects (only to files). But something along these lines would be really nice in the future. We can think on this for a bit; there has to be a nice way."

I tried it the Python way:

```
def xxx(args='', stdin, stdout, stderr):
  print('abc', file=stdout, flush=True)
  sleep 1
  print('def', file=stdout, flush=True)
```

but

`xxx | grep abc`

just hangs and

`xxx > aaa`

works.

However, with

```
def xxx(args='', stdin, stdout, stderr):
  print('abc', file=stdout, flush=True)
  time.sleep(1)
  print('def', file=stdout, flush=True)
```

`xxx | grep abc` works. So sprinkling in shell commands interferes,
which perhaps is to be expected, unless we somehow inform them about context stdin,stdout,stderr.

It would be nice to do this implicitly like in classic shells. I.e., xonsh could adjust file descriptors 0,1,2 as needed for redirections and pipes. Then we won't need to pass on stdin,stdout,stderr.

adqm:
I definitely agree that this is something we should have.

scopatz:
This seems to be the kind of thing that is being discussed [on the mailing list here](https://groups.google.com/d/msgid/xonsh/CAAKrkh8MqqGix0Z7zFiBN1tj7VaOMLbURgLpWhOPjYHw%3Do7Fag%40mail.gmail.com?utm_medium=email&utm_source=footer). I guess I would want to see a proposed syntax before doing anything.

LarryRiedel:
The kind of thing being discussed on the mailing list [there](https://groups.google.com/forum/#!msg/xonsh/1rnysYG1Uqk/CG0S0n-tBQAJ) is adding support for alternative delimiters around Python compound statement suites, in other words an alternative to using `NEWLINE INDENT` and `DEDENT` around statement lists, as in `p_suite()` in parsers/base.py.

To me using `!{` instead of `NEWLINE INDENT` with either `}` or `!}` instead of `DEDENT` seems to match things like `!()` and `![]` which xonsh already uses, but I don't know what all tokens are already used.

scopatz:
[...]

PR #1223, opened 2016-06-13
Proposal: Inline "suite"

adqm:
There was some discussion in the IRC/Gitter.im rooms a while back about potentially adding a way to add "inline" blocks to xonsh.

This adds a new syntax `!{...}` to wrap "suites," in addition to the normal newline- and indent-based scheme,  to allow for writing more complicated statements on one line.  For example:

```
hartz@tabby ~ $ for i in range(3): !{print(7); print(8)}
7
8
7
8
7
8

hartz@tabby ~ $ for i in range(5): !{if i%2 == 0: !{print('YAY');print(i)} else: !{print('OH NO'); print(i+0)};print(i+20)}
YAY
0
20
OH NO
1
21
YAY
2
22
OH NO
3
23
YAY
4
24
```

This PR is a partial implementation of this idea.  Still a few things missing:
- documentation
- tests
- changes to `subproc_toks` to allow proper automatic wrapping of subproc mode commands (only manually-wrapped subproc commands work as of now)

but I wanted to put this in to gather opinions from core devs before going too far down this road.

### DIFF
--- a/xonsh/lexer.py
+++ b/xonsh/lexer.py
@@ -395,14 +395,12 @@ class Lexer(object):
 
     def input(self, s):
         """Calls the lexer on the string s."""
-        print('HEY', repr(s))
         self.token_stream = get_tokens(s, self)
 
     def token(self):
         """Retrieves the next token."""
         self.beforelast = self.last
         self.last = next(self.token_stream, None)
-        print(self.last)
         return self.last
 
     def __iter__(self):
--- a/xonsh/lexer.py
+++ b/xonsh/lexer.py
@@ -66,7 +66,46 @@ it up here and generate a single PLY token of the given type.  Otherwise, it
 will fall back to handling that token using one of the handlers in
 ``special_handlers``.
 """
+<<<<<<< HEAD
 del _token_map
+=======
+
+# operators
+_op_map = {
+    # punctuation
+    ',': 'COMMA', '.': 'PERIOD', ':': 'COLON',
+    '...': 'ELLIPSIS',
+    # basic operators
+    '+': 'PLUS', '-': 'MINUS', '*': 'TIMES', '@': 'AT', '/': 'DIVIDE',
+    '//': 'DOUBLEDIV', '%': 'MOD', '**': 'POW', '|': 'PIPE',
+    '~': 'TILDE', '^': 'XOR', '<<': 'LSHIFT', '>>': 'RSHIFT',
+    '<': 'LT', '<=': 'LE', '>': 'GT', '>=': 'GE', '==': 'EQ',
+    '!=': 'NE', '->': 'RARROW',
+    # assignment operators
+    '=': 'EQUALS', '+=': 'PLUSEQUAL', '-=': 'MINUSEQUAL',
+    '*=': 'TIMESEQUAL', '@=': 'ATEQUAL', '/=': 'DIVEQUAL', '%=': 'MODEQUAL',
+    '**=': 'POWEQUAL', '<<=': 'LSHIFTEQUAL', '>>=': 'RSHIFTEQUAL',
+    '&=': 'AMPERSANDEQUAL', '^=': 'XOREQUAL', '|=': 'PIPEEQUAL',
+    '//=': 'DOUBLEDIVEQUAL',
+    # extra xonsh operators
+    '?': 'QUESTION', '??': 'DOUBLE_QUESTION', '@$': 'ATDOLLAR',
+    '&': 'AMPERSAND',
+}
+for (op, type) in _op_map.items():
+    token_map[(OP, op)] = type
+
+token_map[IOREDIRECT] = 'IOREDIRECT'
+token_map[STRING] = 'STRING'
+token_map[DOLLARNAME] = 'DOLLAR_NAME'
+token_map[NUMBER] = 'NUMBER'
+token_map[REGEXPATH] = 'REGEXPATH'
+token_map[NEWLINE] = 'NEWLINE'
+token_map[INDENT] = 'INDENT'
+token_map[DEDENT] = 'DEDENT'
+if PYTHON_VERSION_INFO >= (3, 5, 0):
+    token_map[ASYNC] = 'ASYNC'
+    token_map[AWAIT] = 'AWAIT'
+>>>>>>> new handling of semicolons inside of inline block
 
 
 def _make_matcher_handler(tok, typ, pymode, ender):
@@ -188,16 +227,30 @@ def handle_double_pipe(state, token):
 
 
 def handle_banglbrace(state, token):
+    state['inline_suite_level'] += 1
     sl, sc = token.start
     yield _new_token('NEWLINE', '\n', token.start)
     yield _new_token('INDENT', ' ', (sl, sc+1))
 
 
 def handle_rbracebang(state, token):
+    state['inline_suite_level'] -= 1
+    if state['inline_suite_level'] < 0:
+        e = "}! used outside of inline suite."
+        yield _new_token('ERRORTOKEN', e, token.start)
+        return
     sl, sc = token.start
     if state['lexer'].last.type != 'DEDENT':
         yield _new_token('NEWLINE', '\n', token.start)
     yield _new_token('DEDENT', '', (sl, sc+1))
+
+
+def handle_semicolon(state, token):
+    if state['inline_suite_level'] > 0:
+        yield _new_token('NEWLINE', '\n', token.start)
+        return
+    yield _new_token('SEMI', ';', token.start)
+
 
 special_handlers = {
     NL: handle_ignore,
@@ -206,6 +259,7 @@ special_handlers = {
     ENDMARKER: handle_ignore,
     NAME: handle_name,
     ERRORTOKEN: handle_error_token,
+    (OP, ';'): handle_semicolon,
     (OP, ')'): handle_rparen,
     (OP, '}'): handle_rbrace,
     (OP, ']'): handle_rbracket,
@@ -279,6 +333,7 @@ def get_tokens(s, lexer):
     """
     state = {'indents': [0], 'last': None,
              'pymode': [(True, '', '', (0, 0))],
+             'inline_suite_level': 0,
              'stream': tokenize(BytesIO(s.encode('utf-8')).readline),
              'lexer': lexer}
     while True:
@@ -362,6 +417,7 @@ class Lexer(object):
     tokens = tuple(token_map.values()) + (
         'NAME',                  # name tokens
         'WS',                    # whitespace in subprocess mode
+        'SEMI',                  # ;
         'LPAREN', 'RPAREN',      # ( )
         'LBRACKET', 'RBRACKET',  # [ ]
         'LBRACE', 'RBRACE',      # { }
--- a/xonsh/lexer.py
+++ b/xonsh/lexer.py
@@ -66,47 +66,7 @@ it up here and generate a single PLY token of the given type.  Otherwise, it
 will fall back to handling that token using one of the handlers in
 ``special_handlers``.
 """
-<<<<<<< HEAD
 del _token_map
-=======
-
-# operators
-_op_map = {
-    # punctuation
-    ',': 'COMMA', '.': 'PERIOD', ':': 'COLON',
-    '...': 'ELLIPSIS',
-    # basic operators
-    '+': 'PLUS', '-': 'MINUS', '*': 'TIMES', '@': 'AT', '/': 'DIVIDE',
-    '//': 'DOUBLEDIV', '%': 'MOD', '**': 'POW', '|': 'PIPE',
-    '~': 'TILDE', '^': 'XOR', '<<': 'LSHIFT', '>>': 'RSHIFT',
-    '<': 'LT', '<=': 'LE', '>': 'GT', '>=': 'GE', '==': 'EQ',
-    '!=': 'NE', '->': 'RARROW',
-    # assignment operators
-    '=': 'EQUALS', '+=': 'PLUSEQUAL', '-=': 'MINUSEQUAL',
-    '*=': 'TIMESEQUAL', '@=': 'ATEQUAL', '/=': 'DIVEQUAL', '%=': 'MODEQUAL',
-    '**=': 'POWEQUAL', '<<=': 'LSHIFTEQUAL', '>>=': 'RSHIFTEQUAL',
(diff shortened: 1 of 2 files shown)


adqm:
Looks like there are also some other unforeseen issues because things can parse multiple ways now.  I can clean those up, but I wanted to see whether people think this is an idea worth pursuing before I go head with it.

Personally, I think it would be a nice addition.

gforsyth:
Hey @adqm -- I'm not sure I see the utility?  What's the benefit of this over newline and indent?  Just keystrokes?

adqm:
I think so.  I can't find the old conversation (it was several days ago at least), but I think the thought was that it is nice to be able to write things on one line, which I can agree with.  This might be particularly true in readline, which doesn't have the multi-line editing capability that ptk has.

That said, I can see the utility in not having to scroll back through multiple lines if you want to repeat and older command.  For bigger, more complicated things, I think it is nice to have the normal Python way.  But for shorter things, it might be nice to be able to write them on one line.

scopatz:
Something like this was brought up in the sub-shell discussion (#1141) and [on the mailing list here](https://groups.google.com/d/msgid/xonsh/CAAKrkh8MqqGix0Z7zFiBN1tj7VaOMLbURgLpWhOPjYHw%3Do7Fag%40mail.gmail.com?utm_medium=email&utm_source=footer).

> But for shorter things, it might be nice to be able to write them on one line.

Sure but is there a specific example?  I can't think of a case that can't be done on one line - using semicolons and the interfaces that Python provides.  Yes, it may look different than other shells or when using INDENT and DEDENT, but it probably should.

scopatz:
Also, again, I am not sure that this makes xonsh easier to learn or use.  For example, reading a file on this branch could be done with either:

```
with open(fname) as f: !{s = f.read()}
```

or 

```
f = open(fname); s = f.read(); f.close()
```

whereas currently only the second is allowed.

adqm:
> Sure but is there a specific example? I can't think of a case that can't be done on one line - using semicolons and the interfaces that Python provides.

Not the most _meaningful_ example, but one from my original message:
`for i in range(some_number): !{print(7); print(8)}`

I don't think that can be on-lined using normal Python syntax.

scopatz:
> for i in range(some_number): !{print(7); print(8)}

Isn't this basically the same as either of the following:

```
_ = [print(7), print(8) for i in range(some_number)]
_ = [print(7) or print(8) for i in range(some_number)]

# of if you don't want a big list of nones
_ = list(filter(None, print(7) or print(8) for i in range(2)))
```

melund:
I think I lean towards @scopatz here. I think we should be careful not to add so many things that make xonsh look very different from Python. 

I a side note I think we could make a xontrib with Chelsea's oneliner-izer from pycon  https://us.pycon.org/2016/schedule/presentation/2219/   Then we could just point people who request this kind of feature in that direction. I think that should convince even the most dedicated  single line purists that multiline input is a good thing:)

scopatz:
I'll have to watch Chelsea's talk now @melund :)  I think an official xontrib for this would be good.

LarryRiedel:
For me a primary value proposition and differentiator for xonsh could have been it's use as an interactive shell to replace bash, but requiring multiline blocks or arcane Python constructs for what would have been trivial one-liners in bash makes it a non-starter.

Everything xonsh does for script files can be achieved using pure Python, and support for using pure Python instead of xonsh can be justified using the same kind of logic which would coerce ordinary xonsh users who have never used a comprehension or a lambda into to interrupting their interactive workflow to find a way to turn what would have been a trivial compound statement suite like

`for i in range(some_number): !{print(7); print(8)}`

Into some sort of Python wizard incantation like

`list(filter(None, print(7) or print(8) for i in range(some_number)))`

I may have been mistaken though, in that I thought one of xonsh's primary mission statements was to be an alterative to replace bash for interactive use for users who know use bash and have some Python knowledge.  I don't see xonsh filling that role when it requires users to learn Python arcana when what they want to do would have been trivial in bash... and could easily be trivial in xonsh.

adqm:
I have to agree.  It would have taken me a long time to come up with the list comprehension solution.  There are some things that are natural to write as list comprehensions, and some that are not.  I would say that this would make the xonsh _shell_ more useful, even if it isn't used (probably shouldn't be used!) in xonsh scripts.

scopatz:
If this is only for CLI usage, I would recomend that maybe this could be a xontrib with (like @melund suggested) that adds a command/alias for rewriting the line.  This could even be configured to accept user-specified indent and dedent tokens (as envvars?).  This could then look something like:

```
# defaults?
$ONELINER_INDENT = '!{'
$ONELINER_DEDENT = '}!'

# usage
oneliner for i in range(some_number): !{print(7); print(8)}!
```

The oneliner (1line, or whatever you want to call it) command would then replace the indents and dedents with the appropriate number of spaces (probably using the $INDENT envvar) and run this through the execer.  You could even go only with curly braces alone if you wanted.

I don't think this would be a huge burden to people who wanted to use it, and it wouldn't affect anyone who didn't have the xontrib loaded. And perhaps most importantly to me, it wouldn't be duplicate syntax that new users would potentially have to learn, and then learn not to use. (Can I get a `from x import *` ^_^?!)  Rather, it would be a tool that new users would know that they had to learn about.

Also, doing this as an alias fits my brain a little better because subprocs are inherently one line, where as py-mode is not.

Alternatively, perhaps we could consider adding "raw" tokens to the lexer.  I am not suggesting this specific syntax, but suppose any name wrapped in `%NAME%` was tokenized as that token, with a default value, this would allow you to write:

```
for i in range(some_number): %INDENT%print(7); print(8)%DEDENT%
```

among many other "raw" options. There are probably characters other than `%` that we would want to use, but you get the idea :).

Also, I think there is a certain beauty to using `->` (push in) and `<-` (push out) as the indent and dedent:

```
for i in range(some_number): ->print(7); print(8)<-
```

Though perhaps it would be most pythonic to use whitespace escape characters (outside of strings):

```
for i in range(some_number): \t print(7); print(8) \n
```

I think I am out of ideas for the night, but hopefully some of this helps. Right now, I think that a xontrib that folks can use and customize is probably better than changes to the language - though I understand that is probably what you want.  If we were going to make language changes, I'd be most in favor of raw tokens, since I can see those being generally useful.

Russell-Jones:
I'd point out that "for i in range(10): print(7); print(8); print(9)" works in vanilla python3.5 with readline.

scopatz:
right, but there ia no way to exit the block :)

adqm:
I still disagree here.  I think it's much easier to teach someone they can use `!{}` (or `->` and `<-`, which I actually vastly prefer) than to tell them the right way to do a one-line for-loop is with
`_ = list(filter(None, print(7) or print(8) for i in range(some_number)))`

I think it adds a nice feature that can be ignored by those who don't want to use it.  A xontrib could still be ignored by those who don't want to use it, but it would require strictly more work from those who want to use it.  Thus, I don't think a xontrib is the right answer.  Also, doing this _after_ parse time feels far too bizarre to me...

And having to write an additional command name at the start feels like overkill to me, too.  I think there is a beauty to the simplicity of Python, and the best way to preserve that while adding the ability to do more complicated one-liners would be, in my opinion, to use `->` and `<-`.

adqm:
@Russell-Jones, yes, that is true, and thanks for bringing it up.  For more complicated structures, though, (if/else, nested things, etc), that kind of syntax won't work.  And like @scopatz said, there is not way to make one of those pieces _not_ be a part of the block.

scopatz:
Well, since @adqm and I agree on a syntax we should probably just go with that😃

LarryRiedel:
Would the use of -> and <- be to make sure no editor (including readline) has any clue how to find the other end of the block, and to confuse everyone who has ever used bash, C, Java, JavaScript, Ruby, Perl, PHP, Scala, LISP, etc?

adqm:
Yeah, and it also collides with other syntax in Python (`<-` is valid in several contexts).

I've was planning on moving ahead with the similar (but not valid Python) `>>>` and `<<<`.

adqm:
AFAIK, that doesn't have any meaning anywhere (at least not in any language I know)

LarryRiedel:
Would readline or any other editor have any idea how to find the matching <<< for a >>> ?

melund:
I still think that the answer should be the same as in python :)
```
snail@sea ~ $ from __future__ import braces
  File "<xonsh-code>", line 1
SyntaxError: not a chance
```

melund:
Jokes aside: The fact that no other editors will understand, indicate to me that this is better left as a xontrib for those who really think they can't live without. I also think it should be a cli option only.

LarryRiedel:
@melund non-sequitur.  The CPython shell has no trouble finding the matching braces in

`for i in range(some_number): !{print(7); print(8)}`

scopatz:
@LarryRiedel readline and other editors don't care about the tokens they edit, with the exception of the escape sequences that they are programmed to know about (like '\n').

@adqm - I considered the `>>>` and `<<<` but internally rejected it because:

1. It looks too similar to redirection 
2. which in turn looks too similar the French guillemets
3. is one character longer per token than the `->` operators.

Also, in the context that `->` and `<-` would be used, there would be no way for these to overlap with other Python syntax.  `->` must come after a colon.  Furthermore, `<-` must come between statements. We could enforce that this this comes after a semicolon, and it is nonsensical to take a unary `(-)` of the LHS of a statement.  

That all being said, I am still with @melund. I think this would be a thoroughly unpythonic thing to add. I'll go one step further, though, and say that I think that "a cli option only" should be taken off the table.  We really don't want two languages.

I also feel like we have spent a lot of tools to address this issue through other generic avenues. We have great multiline editing, xontribs, function aliases, etc. I understand that people want this because it is what they are used to from other shells, but other shells have a lot of terrible syntax that we have sought to avoid.  And the biggest issue I have with those other options *is* their duplicated syntax. This is explicitly a request to duplicate syntax.  Hopefully this explains my hesitation.

The only option I am really excited and interested in seeing implemented are the "raw" tokens. I think this would be a unique, general purpose language feature which happens to allow for the one-liner behaviour.  Adding specialized duplicate syntax for one part of the language seems like a slippery slope.  I'd be happy to add the raw tokens myself if people would be fine with that option and no one else wants to work on it :).

melund:
@scopatz . You got my vote: <xonsh 2016>

Russell-Jones:
The tokens idea is v. interesting, though seems to me like it would be too verbose in this case. FWIW, the backtick character is unused in python3 (in 2 they meant repr() ).
Another approach would to allow stuff like 
`for x in range(5):;.if (x%2):;..print('foo');.else:;..print('bar');.print('always')` i.e. ';'+('.'*n) is treated like a newline and n spaces.

scopatz:
> The tokens idea is v. interesting, though seems to me like it would be too verbose in this case. 

If it isn't supposed to be used normally, I don't see the verbosity as an issue. In fact, I see it as an advantage because it would really stick out that something special is happening.

> FWIW, the backtick character is unused in python3 (in 2 they meant repr() ).

We use backticks in xonsh elsewhere though :)

scopatz:
Also, maybe we could use `^` for raw tokens, ie `^INDENT^`

adqm:
> If it isn't supposed to be used normally

i don't agree with "it isn't supposed to be used normally."  if that's the approach, i'm not going to bother with finishing this out.

scopatz:
> i don't agree with "it isn't supposed to be used normally." if that's the approach, i'm not going to bother with finishing this out.

I thought that this is what you meant by, 

> I would say that this would make the xonsh shell more useful, even if it isn't used (probably shouldn't be used!) in xonsh scripts.

adqm:
in scripts, sure, i think whitespace should be preferred.  but on the command line, i don't think we should take the approach of _we don't want people to use this so let's not worry about making it usable_.

then again, i don't necessarily like forcing my idea of style on people, even in scripts.  if someone wants to use inline blocks for pieces of scripts, i think that's fine.

so i think the question we should actually be trying to answer (assuming we want this as a language feature) is: _how do we make this as usable as possible while making sure those who don't want to use it aren't inconvenienced?_  at least, that's my general approach to thinking about new features.

fwiw, i don't particularly like the "raw tokens" idea, as it requires understanding the names we're giving to all the tokens, assuming we want to generalize it.  if we don't want to generalize it and it exists only for indent/dedent, then i think we should have a cleaner syntax.

scopatz:
@adqm that all seems reasonable.  I think we have reached the end of what we can meaningfully discuss asynchronously.  Maybe we could have a google hangout about it or something?  We have a lot of proposals on the table and two modes of thought.  My suspicion is that if we do nothing, this will keep coming up. 

I think this is, like you said, a matter of figuring out what people who don't want this won't be super bothered by, but also balancing this with what people who do want this also won't be super bothered by.

As I said, I wouldn't mind `->` and `<-` that tokenize/lex to INDENT and DENENT.  We do something like this for `&&` and `||`.

adqm:
>  I think we have reached the end of what we can meaningfully discuss asynchronously.

I think that may be right.  I don't have a google account, so I can't do hangouts.  But I could do something like https://apprtc.appspot.com/.  Or maybe just a private conversation on Gitter or something like that?  Though text-only does make it harder to communicate effectively.  Just let me know whatever works best.

scopatz:
Happy to have an ApptRTC conversation at some point.

scopatz:
Just wanted to recap the developer conversation here. We have decided to use `!{...}!` where `!{` is tokenized/lexed as an INDENT token and `}!` is tokenized/lexed as a DEDENT. This functions analogously to how `&&` and `||` are substituted for AND and OR tokens, and will have no other side effects other than defining the limits of a block.  Also note that Python itself has several way defining INDENT and DEDENT (tabs and spaces in various amounts).  This simply adds another alias to something that may already be done in many ways in Python, even if only one way is considered the appropriate style. 

We strongly will recomend that this not be used outside of the CLI environment. That is it should not be used in *.xsh modules or scripts.

Furthermore, it is important to point out that - like with everything else in pre-v1.0 xonsh - this is subject to change or removal. If this doesn't work out for some reason, we will change it.  So if you not a fan of this, there will be oppurtunities to change it in the future. However, we do need to give it a shot and the only way we can do that is by trying it out.

This is the best compormise of practicality and idealism that @adqm @gforsyth @Carreau and myself could come up with after considering many many examples, pros and cons. For the time being, let's consider the bike shed painted and its xolor is `!{...}!`

scopatz:
I am going to suggest that we close this if there is no interest in updating it. Thoughts?

scopatz:
I am closing this since there doesn't seem any interest in keeping this alive. Please feel free to reopen or put in a new PR if there is continued interest in this feature.

Issue #1198, opened 2016-06-10
Odd completer behaviour


scopatz:
If I go into the test directory and type `xo test_as<TAB>`, the completer will now eat my `as` rather than complete to the only valid file (`test_ast.py`).

```console
scopatz@artemis ~ $ cd xonsh
scopatz@artemis ~/xonsh bfix $ cd tests/
scopatz@artemis ~/xonsh/tests bfix $ xo test_as<TAB>
scopatz@artemis ~/xonsh/tests bfix $ xo test_
```

This seems like it is unexpected. This was on the readline shell.

```
+------------------+----------+
| xonsh            | 0.3.4    |
| Python           | 3.5.0    |
| PLY              | 3.7      |
| have readline    | True     |
| prompt toolkit   | 1.0.0    |
| shell type       | readline |
| pygments         | 2.1.3    |
| on posix         | True     |
| on linux         | True     |
| distro           | debian   |
| on darwin        | False    |
| on windows       | False    |
| on cygwin        | False    |
| is superuser     | False    |
| default encoding | utf-8    |
+------------------+----------+
```

hugo-aa:
I also found this bug with this - the `o` in `xo<TAB>` will be eaten:

```
$ ls -F
xnotx/ xonsh/
$ cd xo<TAB>
```

This issue can be fixed by set `$SUBSEQUENCE_PATH_COMPLETION = False` in xonshrc file. Related code is here: xonsh.completers/path.py

adqm:
> I also found this bug with this - the o in xo<TAB> will be eaten:

Ah, okay.  My understanding of how this was working in other shells wasn't quite right.  It looks like (on zsh, at least), it only falls back to the subsequence matching if no actual matches are found.  That's probably the "right thing".  So I'll modify it to do:

1. Look for actual completions
2. If nothing found, try subsequence matching
3. If still nothing found, try the fuzzy matching

hugo-aa:
> Look for actual completions
> If nothing found, try subsequence matching
> If still nothing found, try the fuzzy matching

Yeah, I think this is the correct way :)

mitnk:
The issue I mentioned here has been fixed, thanks @adqm

scopatz:
Thanks for the reminder @mitnk!

PR #1224, opened 2016-06-13
only do subsequence and fuzzy matching if no normal matches found

adqm:
I believe this should fix the issue from #1198.  @scopatz, do you mind testing?


scopatz:
Relaunched failed travis job.

scopatz:
Thanks a ton @adqm! This seems to work much better in my simple tests.

Issue #1102, opened 2016-06-05
cat, pipes, and ctrl-d


mburo:
cat | grep aaa
and typing some lines such as

aaa
bbb
ctrl-d

doesn't work as expected

bash prints aaa right after lines that contain aaa;
xonsh doesn't print anything, and execution doesn't stop after ctrl-d.

cat > xxx
aaa
bbb
ctrl-d

works!

xonfig
+------------------+----------+
| xonsh | 0.3.2 |
| Python | 3.4.3 |
| PLY | 3.7 |
| have readline | True |
| prompt toolkit | None |
| shell type | readline |
| pygments | None |
| on posix | True |
| on linux | True |
| distro | Fedora |
| on darwin | False |
| on windows | False |
| is superuser | False |
| default encoding | utf-8 |
+------------------+----------+

adqm:
I can confirm this.  I imagine it is because we are giving the terminal over to the last process in the pipeline, so stdin is trying to go to `grep` in your example, instead of to `cat`.  I can look into fixing this.

adqm:
Thanks for reporting, @mburo!

mburo:
just checked fix #1226 - seems to work for the test cases above.

scopatz:
Closing due to fix, feel free to reopen.

PR #1226, opened 2016-06-14
Fix for pipeline issues

adqm:
This small change should fix issues we've had with pipelines, in particular #1102, by giving the terminal to the _first_ process in a pipeline, rather than the last.


scopatz:
Perhaps @mburo would be willing to test this for us?

scopatz:
Thanks again @adqm!

scopatz:
Thanks for testing @mburo

PR #1227, opened 2016-06-14
Color stderr in red

adqm:
I got to talking about shells with a colleague recently, and we thought it might be nice to differentiate between stderr and stdout in regular output on the terminal.  This patch adds this idea to xonsh.  By default, if stderr is not being redirected elsewhere (if it is going to the terminal), this patch will cause it to be displayed in red (stdout will be displayed in the normal color).

@scopatz, I'm not sure how to hook this in with the coloring framework built in to xonsh (i've mostly avoided that stuff), so it just colors red using an ANSI color escape for now.  Is it necessary to hook into that system?


adqm:
Also not able to see right now why this is screwing up the tests...

Seems to be something with the import hooks, but I'm not seeing how this change could have affected those (and I am passing all tests on my machine).

scopatz:
Awesome!  Some one asked for this super early on, though now I can't find this issue.  This will be great to have.

> @scopatz, I'm not sure how to hook this in with the coloring framework built in to xonsh (i've mostly avoided that stuff), so it just colors red using an ANSI color escape for now. Is it necessary to hook into that system?

Yes, it is pretty much necessary.  Not all systems support ANSI sequences and our color management abstracts that via either raw ANSI or pygments.  Luckily, it is pretty easy to use.  The `xonsh.tools.format_color()` and `xonsh.tools.print_color()` functions should be the only two entry points you need.

Also, as a not eon the implementation, I think that it will almost certainly be the case that people will want to customize this.  I think we should instead of a boolean `$COLOR_STDERR`, we should have a `$STDERR_COLOR` that contains a color string (or None?).  @melund and @BYK should probably tell us what the default color should be on Windows.  My guess is that it will be `INTENSE_RED` rather than just `RED`.

We should also decide if this should be just a color name and have a separate envvar for background colors or whether it should just be a str that has the color codes in it: ie `'RED'` and `'BACKGROUND_BLACK'` vs `'{RED}{BACKGROUND_BLACK}'`.

BYK:
Guessing it needs to be `INTENSE_RED` but @melund is better with colors than I am :)

melund:
Well. `RED` is very unreadable in cmd.exe, so `INTENSE_RED` is better. 

I suggest that `$STDERR_COLOR` could be a string with `INTENSE_RED`  or a list of strings ['INTENSE_RED', 'BACKGROUND_BLACK']

gforsyth:
@adqm -- I have the same test failures as Travis locally, both on your branch but also on master...  which is a little weird.  Going to dig in a little more and see if I can figure it out

PaulReiber:
While I can see the utility of colorizing stderr, I can also see utility of
prepending and/or appending a prefix or suffix (or both) to each line of
stderr.  I recommend that this be designed so as to accommodate a variety
of use cases:

-> end user desires to prepend each line of stderr with   "STDERR: "
-> end user desires to append " - ERROR!" to each line of stderr
-> end user desires to colorize the line red

Just my $0.02 - hope it helps.

Kind regards,
-Paul Reiber
Phone: (650)430-7926
Email: paul@reiber.org
Web: http://bit.ly/reiber

adqm:
I'm find with that.  So maybe just a more general approach for modifying lines of stderr, which defaults to coloring red.

adqm:
Okay, several updates here (much more customizable now), and everything passes local tests for me,  so hopefully travis will cooperate.  I think it's ready for another look.

I haven't tested on Windows, but @BYK and @melund, can you help with that?

adqm:
oh, i see.  looks like ptk doesn't give me a string from `format_color`.  i'll switch to use `print_color` directly.

adqm:
also this will probably cause issues with job control.  looking into that now.

adqm:
_almost working_.  everything good on readline for me, but broken on ptk (some kind of file descriptor overlap?).  i'll come back to this later.

scopatz:
I believe this has now been superceded on master. Please feel free to reopen.

PR #1230, opened 2016-06-14
Source cmd prompt

melund:
This is similar to #1219 . It resets the PROMPT variable before  source'ing `*.bat` files. It also ensures that the xonsh `$PROMPT` variable is restored if it should be afterwards.


scopatz:
Thanks a million @melund!

PR #1231, opened 2016-06-14
Simplify Env.__repr__

melund:
This simplifies  `__xonsh_env__._repr__()` so it does not print the whole environment dictionary.

 It does not change the string representation of the object or the pretty print method. So it does not change what happens if we type: `${...}` at the prompt. 

Before this , calling help on env methods would print the entire enviroment: 
```
snail@sea ~ $ ${...}.swap?   
Type:         method
String form:  <bound method Env.swap of xonsh.environ.Env({'GIT_LFS_PATH': ['C:\\Program Files\\Git LFS'], 'COM <...> 'ANYBODY', 'CONDA_DEFAULT_ENV': 'C:\\Users\\mel\\Anaconda3\\envs\\x', 'SESSIONNAME': 'Console'})>
File:         c:\users\mel\documents\xonsh\xonsh\environ.py
Definition:   (*args, **kwds)
Docstring:
Provides a context manager for temporarily swapping out certain
environment variables with other values. On exit from the context
manager, the original values are restored.

<bound method Env.swap of xonsh.environ.Env({'GIT_LFS_PATH': ['C:\\Program Files\\Git LFS'], 
.....
<VERY LONG SNIPPET>
....
'C:\\Users\\mel\\Anaconda3\\envs\\x', 'SESSIONNAME': 'Console'})>
```
After this change it renders more nicely:

```
snail@sea ~ $ ${...}.swap?  
Type:         method
String form:  <bound method Env.swap of xonsh.environ.Env(...)>
        c:\users\mel\documents\xonsh\xonsh\environ.py
Definition:   (*args, **kwds)
Docstring:
Provides a context manager for temporarily swapping out certain
environment variables with other values. On exit from the context
manager, the original values are restored.

<bound method Env.swap of xonsh.environ.Env(...)>
````

### DIFF
--- a/news/simplify_env_repr.rst
+++ b/news/simplify_env_repr.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** 
+
+* ``__repr__`` on the environment only shows a short representation of the 
+  object instead of printing the whole environment dictionary
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -610,9 +610,9 @@ class Env(MutableMapping):
         """If no initial environment is given, os.environ is used."""
         self._d = {}
         self._orig_env = None
-        self.ensurers = {k: Ensurer(*v) for k, v in DEFAULT_ENSURERS.items()}
-        self.defaults = DEFAULT_VALUES
-        self.docs = DEFAULT_DOCS
+        self._ensurers = {k: Ensurer(*v) for k, v in DEFAULT_ENSURERS.items()}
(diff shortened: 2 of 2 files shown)


gforsyth:
Hey @melund -- this is a great change.  Those attributes which are lists are still printing out their entire contents when the `?` is used.  `defaults`, `docs`, and `ensurers`.  Do we want to hide those, too?

melund:
Ahh. Yes. That is the nature of `?`. But before the whole environment would also get printed. So it is an improvement. 

I don't think we can change that unless we make the attributes classes them self.  I guess `Env.defaults`, `Env.docs` ,`Env.ensure` should just be private attributes.

gforsyth:
That seems like a good solution, and this looks good to me!  Merging now

PR #1232, opened 2016-06-14
add mock xonsh env to test

gforsyth:
`test_format_prompt_with_invalid_func` raises an exception
(intentionally).  when this is checked against `partial_format_prompt`
it eventually makes its way to `print_exception` in `tools.py` which
expects `env` to have an attribute `is_manually_set`.

This just adds a mock xonsh env to the test so that `env` is properly
defined.

I think this was the test failing in #1223 

And no, I don't like that it's failing sporadically.

### DIFF
--- a/tests/test_environ.py
+++ b/tests/test_environ.py
@@ -85,11 +85,13 @@ def test_format_prompt_with_broken_template_in_func():
         assert_in('user', format_prompt(p))
 
 def test_format_prompt_with_invalid_func():
-    def p():
-        foo = bar  # raises exception
-        return '{user}'
-    assert_is_instance(partial_format_prompt(p), str)
-    assert_is_instance(format_prompt(p), str)
+    env = Env({'XONSH_SHOW_TRACEBACK': False})
+    with mock_xonsh_env(env):
+        def p():
+            foo = bar  # raises exception
+            return '{user}'
+        assert_is_instance(partial_format_prompt(p), str)
+        assert_is_instance(format_prompt(p), str)
 
 def test_HISTCONTROL():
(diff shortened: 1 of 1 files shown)


mitnk:
@gforsyth Thanks for catching it!

I'm still checking why `env` inside `print_exception()` got `{}` there. If `{}` is rational value (i.e. not only in test environments) to it, then adding a checker to handle this in `print_exception()` would be better. Adding mock to this test case seems a bit odd.

mitnk:
In `print_exception()`, `env = getattr(os.environ, '__xonsh_env__', os.environ)` can get `os.environ` when `__xonsh_env__` not in os.environ. Then `env` will not have a `is_manually_set` method. We need to improve the code here.

mitnk:
Just find why `getattr(os.environ, '__xonsh_env__', os.environ)` got `{}`, it should be set here:

```
./tests/tools.py:15:builtins.__xonsh_env__ = {}
```

gforsyth:
I'm bothered that if I run `test_environ.py` on its own, that test fails for the reason stated, but if I run the whole suite, it passes.  Which means somewhere we're defining the ENV and not deleting it afterwards.

gforsyth:
This seems like it's no longer necessary.  Don't really know why, but I've learned to accept that.

scopatz:
> This seems like it's no longer necessary. Don't really know why, but I've learned to accept that.

Take a deep breathe and visualize letting go :)

gforsyth:
> Take a deep breath and visualize letting go :)

There is only one rule for washing dishes in this house: When we wash dishes, we wash the dishes to wash the dishes.

PR #1239, opened 2016-06-15
enable character selection in vi_mode

gforsyth:
When `open_in_editor` is enabled, the `VI_MODE` binding to open the
editor is `v`, which is unexpected, as `v` should be character
selection.

The `open_in_editor` option has been disabled, but the `Ctrl-X Ctrl-E`
shortcut to open the current buffer in an editor has been explicitly
enabled, so there shouldn't be any change of behavior for anyone except
those who were using `v` to open an editor in `VI_MODE` (which I think
is no one).

### DIFF
--- a/xonsh/ptk/key_bindings.py
+++ b/xonsh/ptk/key_bindings.py
@@ -136,7 +136,7 @@ def load_xonsh_bindings(key_bindings_manager):
         """
         event.cli.current_buffer.insert_text(env.get('INDENT'))
 
-    @handle(Keys.ControlX, Keys.ControlE, filter= ~has_selection)
+    @handle(Keys.ControlX, Keys.ControlE, filter=~has_selection)
     def open_editor(event):
         """ Open current buffer in editor """
         event.current_buffer.open_in_editor(event.cli)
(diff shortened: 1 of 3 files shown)


laerus:
truth is i've been using 'v' in vi mode to open the editor cause readline does this by default, but readline has no visual-mode (character selection) and ptk has and its cool so i don't mind changing this.

xonsh/ptk/key_bindings.py
```diff
@@ -134,6 +136,11 @@ def _(event):
         """
         event.cli.current_buffer.insert_text(env.get('INDENT'))
 
+    @handle(Keys.ControlX, Keys.ControlE, filter= ~has_selection)
```
scopatz: space after `=`

scopatz:
Thanks for putting this in @gforsyth! Just the one small change.

scopatz:
Looks good to me @gforsyth! This ready to be merged?

gforsyth:
aye, aye, captain

scopatz:
Hoist the master sail!

PR #1242, opened 2016-06-15
point xonfig wizard to hidden docs attribute

gforsyth:
#1231 made `docs` a hidden attribute of `builtins.__xonsh_env__`, but `xonfig wizard` still looks for a visible attribute and throws an error on first-time startup.


melund:
Good catch @gforsyth . Thanks for cleaning up my mess :)

Issue #1201, opened 2016-06-10
Wrong path completions when filenames have spaces


melund:
I have a problem with path completion when folders have spaces. 

If I have the following folders: 
'Program Files (x86)'
'Program Files'
ProgramData

if I type: 
```
snail@sea C:\ $ cd Prog<Tab>
``` 
It only gives me `ProgramData`

If I instead start with a quote. It gives me all completions.
```
snail@sea C:\ $ cd "Prog<Tab>
```

laerus:
can you please add the output of `xonfig` ?

melund:
Sorry about that. I am the current master. 

```
+------------------+----------------+
| xonsh            | 0.3.4          |
| Python           | 3.5.1          |
| PLY              | 3.7            |
| have readline    | True           |
| prompt toolkit   | 1.0.0          |
| shell type       | prompt_toolkit |
| pygments         | 2.1.3          |
| on posix         | False          |
| on linux         | False          |
| on darwin        | False          |
| on windows       | True           |
| on cygwin        | False          |
| is superuser     | False          |
| default encoding | utf-8          |
+------------------+----------------+

$ completer list
Registered Completer Functions:
python_mode : Python-mode completions for @( and ${
       base : If the line is empty, complete based on valid commands, python names, and paths.
              If we are completing the first argument, complete based on valid
              commands and python names.
       skip : Skip over several tokens (e.g., sudo) and complete based on the rest of the
              line.
         cd : Completion for "cd", includes only valid directory names.
      rmdir : Completion for "rmdir", includes only valid directory names.
       bash : Completes based on results from BASH completion.
        man : Completes an option name, based on the contents of the associated man page.
     import : Completes module names and contents for "import ..." and "from ... import ..."
     python : Completes based on the contents of the current Python environment, the Python
              built-ins, and xonsh operators.
       path : Completes based on a path name.
```

melund:
@laerus : Were you not able to reproduce?

laerus:
yes, i can confirm this.

adqm:
Yeah, I wonder if this is a result of #1167.  I didn't implement the quoting of paths there.  If you turn off `$SUBSEQUENCE_PATH_COMPLETION` and `$FUZZY_PATH_COMPLETION`, do you still see this problem?

laerus:
yes it still needs the quote in order to complete pathnames containing space

adqm:
okay, i will try to take a look later today

adqm:
[...]

PR #1244, opened 2016-06-15
Fix missing completions

melund:
This fixes the missing completions in https://github.com/xonsh/xonsh/issues/1201

If any of the completions needs to be quoted then all completions gets quoted.

### DIFF
--- a/news/quoted_path_completions.rst
+++ b/news/quoted_path_completions.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fixed missing completions for ``cd`` and ```rmdir`` when directories had spaces 
+  in their names.
+
+**Security:** None
--- a/xonsh/completers/path.py
+++ b/xonsh/completers/path.py
@@ -277,7 +277,9 @@ def complete_path(prefix, line, start, end, ctx, cdpath=True):
     return paths, lprefix
 
 
-RE_UNQUOTE = re.compile("""(?:[rub](?P<q1>["'])|(?P<q2>["']?))(.+)(?:(?P=q1)|(?P=q2))$""")
+RE_UNQUOTE = re.compile("""(?:[r](?P<raw>(?P<quotes>"[^"]|'[^']|\"""|'''))|"""
+                        """(?P<normal>(?P=quotes)?))"""
+                        """(?P<unquoted>.+)(?:(?P=raw)|(?P=normal))$""")
 
 
 def complete_dir(prefix, line, start, end, ctx, cdpath=False):
--- a/xonsh/completers/path.py
+++ b/xonsh/completers/path.py
@@ -287,6 +287,6 @@ def complete_dir(prefix, line, start, end, ctx, cdpath=False):
     dirs = set()
     for path in paths:
         m = RE_UNQUOTE.match(path)
-        if m and os.path.isdir(m.group(2)):
+        if m and os.path.isdir(m.group(3)):
             dirs.add(path)
     return dirs, lp
--- a/xonsh/completers/path.py
+++ b/xonsh/completers/path.py
@@ -225,7 +225,7 @@ def _expand_one(sofar, nextone, csc):
     return out
 
 
-def complete_path(prefix, line, start, end, ctx, cdpath=True):
+def complete_path(prefix, line, start, end, ctx, cdpath=True, filtfunc=None):
     """Completes based on a path name."""
     # string stuff for automatic quoting
     path_str_start = ''
@@ -269,24 +269,16 @@ def complete_path(prefix, line, start, end, ctx, cdpath=True):
         paths = {s.replace(home, tilde) for s in paths}
     if cdpath:
         _add_cdpaths(paths, prefix)
+    _add_env(paths, prefix)
+    _add_dots(paths, prefix)
+    if filtfunc is not None:
+        paths = {p for p in paths if filtfunc(p)}
     paths = _quote_paths({_normpath(s) for s in paths},
                          path_str_start,
                          path_str_end)
-    _add_env(paths, prefix)
-    _add_dots(paths, prefix)
     return paths, lprefix
 
 
-RE_UNQUOTE = re.compile("""(?:[r](?P<raw>(?P<quotes>"[^"]|'[^']|\"""|'''))|"""
-                        """(?P<normal>(?P=quotes)?))"""
-                        """(?P<unquoted>.+)(?:(?P=raw)|(?P=normal))$""")
-
-
 def complete_dir(prefix, line, start, end, ctx, cdpath=False):
-    paths, lp = complete_path(prefix, line, start, end, cdpath)
-    dirs = set()
-    for path in paths:
-        m = RE_UNQUOTE.match(path)
-        if m and os.path.isdir(m.group(3)):
-            dirs.add(path)
-    return dirs, lp
+    return complete_path(prefix, line, start, end, cdpath,
+                         filtfunc=os.path.isdir)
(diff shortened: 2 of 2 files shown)


gforsyth:
Hey @melund -- this breaks a bunch of stuff on my system.  If I do `cd <TAB>` the only two options are `.` and `..` and it also seems to interfere with both fuzzy matching and subsequence matching

melund:
Weird. I don't see that effect. Maybe we better have @adqm to take a look at my fix. Maybe something more is needed.

However, I did find another instance of env.defaults... So a fix is on it way for that.

gforsyth:
Super weird:  I just noticed that the behavior I described only occurs in my home directory...

gforsyth:
And again, I was wrong about it messing with subsequence matching -- that was just some weird fluke

gforsyth:
Ahh, ok, it's not only my home directory, it's any directory that contains a folder that has a space in the name

melund:
Ahh. I see it too. I guess, `.` and `..` are not getting quoted like the rest of the completions and prompt_toolkit then only presents them.

melund:
I guess we should wait with merging this until we hear from @jonathanslenders that this is actually the intended behaviour. Must the completions be quoted in the same way.

melund:
Well. @jonathanslenders just confirmed that this shouldn't really happen. https://github.com/jonathanslenders/python-prompt-toolkit/issues/320#issuecomment-226318126

So I guess this PR should just be closed.

adqm:
This branch seems to fix things (at least for me).

@jonathanslenders released a new version that had a supposed fix, but even with that update, I am not seeing the completions unless I am running from this branch.  @melund, are you having similar experience?

gforsyth:
Still not working for me on this branch.  In fact now, in a directory that contains a folder with a space in the name, doing `cd <TAB>` completes to a single quote `'` and then offers no further options.

melund:
Ahh. Now I spotted the real error. The quoted directories are stripped by https://github.com/xonsh/xonsh/blob/master/xonsh/completers/path.py#L280

I completely rebased with this new fix.

@adqm and @gforsyth Can you see this makes a difference for you?

gforsyth:
woohoo!  That fixed it on my end, @melund!

adqm:
Ah, yes!  That would make sense.

I'm not sure that is quite the right fix (we could have `r"` strings, for example).  I think the real right thing might be either to try a regex match to see if something is a string (happy to provide that regex if it's helpful) or to try to `ast.literal_eval` everything before the `os.path.isdir` check.

melund:
Well. `ast.literal_eval` doesn't work well on windows when filenames ends with backslash. 

```
snail@sea C:\ $ ast.literal_eval('"Windows\\"')                                                                                                    env:x py:3.5.1 b:quoted_path_completions
  File "<unknown>", line 1
    "Windows\"
             ^
SyntaxError: EOL while scanning string literal
```

You guys are welcome to finish it. Otherwise, it will have to wait until tomorrow night.

melund:
@adqm Does this seem right to you. It should ensure that the first and last quotes are the same and allow for a `r"` `u"` and `b"` strings.

adqm:
I think we actually need something a bit more complicated (I think this one will let certain things through in a way such that the groups don't quite capture what we need, mainly because `q2` can be empty).

Actually, there may be an easier way.  Although it's currently "private" and not well documented (both my fault), the `_path_from_partial_string` function in `xonsh.completers.path` can actually do this.  It will return `None` in the case that something isn't quoted, and if all it contains is the repr for a string, it should return a tuple that has the contents of that string in position 1.

So I think a little check like the following might work:

```
from_partial_string = _path_from_partial_string(original)
if from_partial_string is None:
    real_name = original
else:
    real_name = from_partial_string[1]
```

and then you can do the `isdir` check on `real_name`.

This may or may not be easier than tweaking the regex to get it right.

melund:
I think that is much nicer. But can you give me an example where the regex would fail?

melund:
If I for example have folder called `'bad` which is then quoted as `"'bad"` then `_path_from_partial_string()[1]` returns `bad` where as the regex gives me: `'bad`

adqm:
The one that jumps out is that it misses triple-quoted strings.

I had thought that it might fail for other kinds of strings, too, but it looks like it is catching all of those (at least the cases I am throwing at it).  I think it wouldn't work well if we were looking at strings more generally or using it to tokenize strings that might be invalid (it allows a bare `'` inside of a string wrapped in `''`, for example), but for the purpose we're using it here (where there is just a single validly-formatted string and nothing else), it seems to hit all the important cases.

It looks like it also can't catch an empty string (though I think that is fine because an empty string can't be a directory's name).

melund:
I can probably add triple quoted strings

adqm:
> If I for example have folder called 'bad which is then quoted as "'bad" then _path_from_partial_string()[1] returns bad where as the regex gives me: 'bad 

Ah yes, that is true.  I guess working with `_path_from_partial_string` might be harder than I had thought...

adqm:
In a separate project I had used:
```
SINGLE_STRING = r"'((\\(.|\n))|([^'\\]))*'"
DOUBLE_STRING = r'"((\\(.|\n))|([^"\\]))*"'
TRIPLE_SINGLE_STRING = r"'''((\\(.|\n))|([^'\\])|('(?!''))|\n)*'''"
TRIPLE_DOUBLE_STRING = r'"""((\\(.|\n))|([^"\\])|("(?!""))|\n)*"""'
```

and then combined them with something like:

```
strings = (SINGLE_STRING, DOUBLE_STRING, TRIPLE_SINGLE_STRING, TRIPLE_DOUBLE_STRING)
STRING_MATCH = "[bBrRuU]*" + '(' + '|'.join(strings) + ')'
```

In that case, I hadn't needed to grab the _insides_ of the strings, so this could be modified to use named groups for the innards in `SINGLE_STRING`, `DOUBLE_STRING`, et al.

Then again, if your approach can be modified to account for triple-quoted strings, that might be better.


-------------

That said, there _is_ actually a function in `xonsh.tools` that can do this: `check_for_partial_string`.  There you can check not only whether a string was started, but also whether it was ended.  By looking at the first two elements in the return tuple there, it should be pretty easy to figure out whether we had a fully-formed string.  To me, that beats regex hacking :)

adqm:
(not only because it avoids some painful regex hacking, but also because it avoids duplicating a bunch of code).

adqm:
Also, I'm boarding a plane now, so I'll be AFK for a bit.  But happy to talk more once I have an Internet connection again.

melund:
Ok. It was easy to add triple quotes to the regex. I will have to look at `check_for_partial_string` later this week.

scopatz:

melund:
I think I found a bug in this. So wait until I got this sorted.

melund:
So I finally came up with a regex that worked `(?:[r]?(?P<q>"|'|"""|'''))(?:(?!(?P=q))(.+)|)(?:(?P=q))$` just to realize that it could be done in a much simpler way. 

I added a filter function to `complete_path` which is called before quoting. @adqm Is this a reasonable solution?

gforsyth:
Hey @melund -- I'm horrified of and impressed with your regex.  
All dirs with spaces are showing up in the completion list -- do we want them to be sorted so they appear within the other folders?  As it  is, they sit at the top of the list of possible completions (before `../` and `./`)

melund:
> All dirs with spaces are showing up in the completion list -- do we want them to be sorted so they appear within the other folders? As it is, they sit at the top of the list of possible completions (before ../ and ./)

Definitely :)  Already a separate PR #1249

gforsyth:
No wonder I was so confused. I could've sworn you had already added in the sorting but I thought it was in this pr and couldn't find it.  OK, merging in now

PR #1245, opened 2016-06-15
Use env._defaults 

melund:
I did a searh of the source code. I think is should be the last.

@gforsyth Do you mind taking a look


gforsyth:
I can't find any others, either.  I'll merge this in once Travis finishes up.

scopatz:
LGTM!

PR #1249, opened 2016-06-16
Better sorting of completions. 

melund:
Ignore case and any leading quotes when sorting the completions. This makes path completions look much nicer. 

Right now sorting is done by the top level completer, but we could consider letting the individual completers sort their own results.


gforsyth:
Woohoo!  This works great, @melund.  Merging in now.

Issue #1118, opened 2016-06-05
document what the optional dependencies do


rahiel:
It would be nice to have a few words in the [docs](http://xon.sh/dependencies.html) explaining why I'd want to install them.

adqm:
Agreed!

laerus:
Solved!

rahiel:
Thanks @laerus.

PR #1255, opened 2016-06-17
more information about library dependencies

laerus:
in request of #1118 on why would someone want to install the extra dependencies

### DIFF
--- a/docs/dependencies.rst
+++ b/docs/dependencies.rst
@@ -5,11 +5,11 @@ Xonsh currently has the following external dependencies,
 *Run Time:*
 
     #. Python v3.4+
-    #. PLY (optional, included with xonsh)
-    #. prompt-toolkit (optional)
-    #. Jupyter (optional)
-    #. setproctitle (optional)
-    #. distro (optional)
+    #. PLY (optional, included with xonsh, alternative parsing tool)
+    #. prompt-toolkit (optional, advanced readline library)
+    #. Jupyter (optional, in-browser REPL)
+    #. setproctitle (optional, allow a process to change its title)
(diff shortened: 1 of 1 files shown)


adqm:
Thanks for putting this in, @laerus.  Rather than just adding a description of what each of the optional packages is, can we also add a brief description of what, specifically, will change within xonsh if that is enabled (e.g., prompt_toolkit allows syntax highlighting of input and more interactive tab completion, etc).

laerus:
should i add it inline, inside the parenthesis, or as a new pargraph?

docs/dependencies.rst
```diff
@@ -5,11 +5,11 @@ Xonsh currently has the following external dependencies,
 *Run Time:*
 
     #. Python v3.4+
-    #. PLY (optional, included with xonsh)
-    #. prompt-toolkit (optional)
-    #. Jupyter (optional)
-    #. setproctitle (optional)
-    #. distro (optional)
+    #. PLY (optional, included with xonsh, alternative parsing tool)
```
gforsyth: PLY isn't an alternative parsing tool, it's just that we vend it ourselves in case it isn't installed locally.

laerus: so should i remote the optional part aswell?

gforsyth:
This looks good, @laerus -- additionally, would you mind adding a "Note" under `prompt_toolkit` that states that if you use `prompt_toolkit` then `pygments` is also a dependency?  I feel like we've been bitten by that one a few too many times now

gforsyth:
Awesome, thanks @laerus ! This is good to go once Travis finishes up.

PR #1270, opened 2016-06-17
Fix `xonsh` not being include in setup.py scripts

youtux:
If you use macOS, the expression `'win' in sys.platform` is `True`, since `sys.platform == 'darwin'`. This would prevent `xonsh` to be added in the scripts.

### DIFF
--- a/news/xonsh_darwin.rst
+++ b/news/xonsh_darwin.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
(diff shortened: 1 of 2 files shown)


melund:
Good catch @youtux. Do you mind adding a change log entry as well? Just make a new copy of `Template.rst` file in the `news/` directory and add an change description to it.

youtux:
@melund done!

melund:
Thanks a lot for fixing this.

PR #1273, opened 2016-06-17
fix bug in main: args.config_path not being used

youtux:


scopatz:
Thanks a million @youtux!  This is great!

PR #1274, opened 2016-06-17
Be more explicit in foreign shell wizard

### LINKED ISSUE #1267: xonsh wizard misleading
I think that the "Foreign shell setup" part of the configuration wizard is a little misleading. I think that a paste will explain what I mean:

```
[...]
                      Foreign Shell Setup
                      -------------------
The xonsh shell has the ability to interface with foreign shells such
as Bash, zsh, or fish.

For configuration, this means that xonsh can load the environment,
aliases, and functions specified in the config files of these shells.
Naturally, these shells must be available on the system to work.
Being able to share configuration (and source) from foreign shells
makes it easier to transition to and from xonsh.

Add a foreign shell, yes or no [default: no]? yes
shell name (e.g. bash): zsh
interactive shell [bool, default=True]:
login shell [bool, default=False]:
env command [str, default='env']:
alias command [str, default='alias']:
extra command line arguments [list of str, default=[]]:
current environment [dict, default=None]:
safely handle exceptions [bool, default=True]:
pre-command [str, default='']:
post-command [str, default='']:
foreign function command [str, default=None]:
source command [str, default=None]:

Add a foreign shell, yes or no [default: no]? yes
shell name (e.g. bash):
interactive shell [bool, default=True]:
login shell [bool, default=False]: KeyboardInterrupt
```
Initially I insert all the configuration for my external shell, and then I get the question "Add a foreign shell, yes or no?". In this moment I think that xonsh means "Do you want to save the data you entered?", but instead it means "Add __another__ foreign shell?".

What do you think about it? Second question be "Add another foreign shell?" and maybe give a little "ok, config saved" message before that?

youtux:
Fix #1267

### DIFF
--- a/news/wizard_foreign_shell_prompt.rst
+++ b/news/wizard_foreign_shell_prompt.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+* More informative prompt when configuring foreign shells in the wizard.
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/xonsh/xonfig.py
+++ b/xonsh/xonfig.py
@@ -111,7 +111,7 @@ Thanks for using the xonsh configuration wizard!"""
 
 def make_fs_wiz():
     """Makes the foreign shell part of the wizard."""
-    cond = wiz.create_truefalse_cond(prompt='Add a foreign shell, ' + wiz.YN)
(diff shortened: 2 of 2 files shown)


gforsyth:
Thanks for putting this in @youtux -- _this_ one should definitely have a news entry

youtux:
Done! You may want to squash the commits while merging, in order to avoid too many commits in master.

gforsyth:
Awesome!  Thanks, @youtux !  And no worries about the commits -- we're fans of rich history.

PR #1276, opened 2016-06-18
fix undefined PWD in default environment

gforsyth:
Somewhere between `0.3.4` and now, `PWD` stopped being defined in the default xonsh environment, which breaks a bunch of things (but only at new logins when `xonsh` is the default shell).  

Simple fix to ensure that it exists, so that `$PROMPT` can be resolved and useful things like `cd` work

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1425,6 +1425,7 @@ def default_env(env=None, config=None, login=True):
     # in order of increasing precedence
     ctx = dict(BASE_ENV)
     ctx.update(os.environ)
+    ctx['PWD'] = _get_cwd()
     # other shells' PROMPT definitions generally don't work in XONSH:
     try:
         del ctx['PROMPT']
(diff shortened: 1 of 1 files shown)


xonsh/environ.py
```diff
@@ -1425,6 +1425,7 @@ def default_env(env=None, config=None, login=True):
     # in order of increasing precedence
     ctx = dict(BASE_ENV)
     ctx.update(os.environ)
+    ctx['PWD'] = _get_cwd()
```
scopatz: This should probably be `_get_cwd() or ''` since that function can return None

scopatz:
Thanks @gforsyth!

scopatz:
Is this ready @gforsyth?

gforsyth:
Should be good to go, @scopatz

scopatz:
Thanks!

PR #1277, opened 2016-06-18
complain on invalid cli arguments

### LINKED ISSUE #1073: xonsh --hep doesn't error
`xonsh --hep` doesn't error. I expected it to error and show the command help. 

Possibly related to https://github.com/scopatz/xonsh/issues/1072?

laerus:
this fixes  #1073 and more invalid cases


xonsh/main.py
```diff
@@ -150,6 +150,10 @@ def premain(argv=None):
         setproctitle(' '.join(['xonsh'] + sys.argv[1:]))
     builtins.__xonsh_ctx__ = {}
     args, other = parser.parse_known_args(argv)
+    if not args.args and other and other[0].startswith('-'):
+        print('xonsh: error: invalid argument {!r}'.format(other[0]))
```
scopatz: This print should go to stderr

xonsh/main.py
```diff
@@ -150,6 +150,10 @@ def premain(argv=None):
         setproctitle(' '.join(['xonsh'] + sys.argv[1:]))
     builtins.__xonsh_ctx__ = {}
     args, other = parser.parse_known_args(argv)
+    if not args.args and other and other[0].startswith('-'):
```
scopatz: This condition doesn't seem to work for me.  Try running a script and you'll see that all the script arguments are still treated as invalid.  For example, in the xonsh dir, 

```
$ xonsh release.xsh --hep
```

will return a error, but it the error from xonsh and not the error from release.xsh.  I bet this is a parse known args problem.

scopatz:
Thanks a million for putting this in @laerus!

scopatz:
Also if you could add a news entry that would be awesome!

laerus:
i'm not rushing to a merge yet, i want more tests for both valid and invalid arguments and more feedback, thank you!

xonsh/main.py
laerus: there are two more `print` in main that print errors, should i change them too? @scopatz

laerus:
there is still a buggy case 
`xonsh TTT.xsh -hASDF`
shows usage and:
`error: argument -h/--help: ignored explicit argument 'ASDF'`
same thing happens with with all short options like
`xonsh TTT.xsh -iasdas`
...
`error: argument -i/--interactive: ignored explicit argument 'asdas'`

xonsh/main.py
scopatz: That would nice, but you don't have to if you think it is out of scope

laerus:
do we want to count arguments after *file* as arguments to both xonsh and script?
maybe its less ambiguous if all arguments after *file* are ignored by xonsh

scopatz:
I agree. I think that all arguments after the file should be ignored by xonsh itself and passed down to the script.

laerus:
there is still the strange case of '--' and '-' and dashes in general
also `xonsh -i` *fname* , doesn't seem to get into interactive mode

xonsh/main.py
```diff
@@ -163,6 +163,11 @@ def premain(argv=None):
         # and positional arguments passed before the path to the script-file are
         # ignored.
         args.args = arguments[file_index+1:]
+    elif not args.args and other and other[0].startswith('-'):
```
scopatz: I think that the problem here is that on line 160, you need to reassign `other`. We are using the old other, and we need to use the new other.

laerus: i have removed my if clause at line 166 in my latest commits. trying for a more robust way to fix this.

scopatz: OK!

laerus:
travis will complain, there is no branch that passes these tests (yet) :smiling_imp:

xonsh/main.py
```diff
@@ -113,7 +113,7 @@ def path_argument(s):
                     metavar='args',
                     help='Additional arguments to the script specified '
                          'by script-file',
-                    nargs='*',
+                    nargs=argparse.REMAINDER,
```
scopatz: Ahh yes! I think this is what we want.

xonsh/main.py
```diff
@@ -150,19 +150,19 @@ def premain(argv=None):
         setproctitle(' '.join(['xonsh'] + sys.argv[1:]))
     builtins.__xonsh_ctx__ = {}
     args, other = parser.parse_known_args(argv)
-    if args.file is not None:
+    file = args.file
+    if file is not None:
```
scopatz: please don't use the name 'file', since it is a Python builtin. Thanks!

xonsh/main.py
```diff
@@ -243,6 +243,10 @@ def main(argv=None):
             from xonsh import xonfig  # lazy import
             xonfig.xonfig_main(['wizard', '--confirm'])
         shell.cmdloop()
+    else:
+        print('xonsh: error: invalid argument {!r}'.format(sys.argv[1]))
```
scopatz: Please print to stderr

scopatz:
Thanks @laerus! I think we are back on the right track.

xonsh/main.py
laerus: oops sry, poor python2

xonsh/main.py
laerus: was thinking if i could decouple `premain` from error printing, will get back to this

laerus:
@scopatz atm master can't start in interactive mode after executing a script. Should i try and fix this too? or i could remove the one test that is failling cause of this and see if its ready for merge.

xonsh/main.py
scopatz: It looks like this is the source of the error because you need to `import argparse` at the top of this file.

laerus: this strangely didn't produce any errors, the only test that is failing is the one that i added to check if XONSH_INTERACTIVE is true if the `-i` is used along with a `filename`. Xonsh atm doesn't seem to be able to start in interactive mode when a file is passed @scopatz

scopatz: > this strangely didn't produce any errors

That is probably because of the amalgamation and another file importing argparse. This should still be fixed though.

>  Xonsh atm doesn't seem to be able to start in interactive mode when a file is passed @scopatz

What do mean by start?  Does it run the file?  Xonsh should still be able to run files in interactive mode, even though it won't pop up with a command line prompt.  This behaviour seems to work just fine for me on master right now.

tests/test_main.py
```diff
@@ -33,5 +36,40 @@ def Shell(*args, **kwargs):
...
+            xonsh.main.premain(['--hep'])
+            assert False
+        except SystemExit:
+            pass
+    with patch('xonsh.main.Shell', Shell), mock_xonsh_env({}):
+        try:
+            xonsh.main.premain(['-TT'])
+            assert False
+        except SystemExit:
+           pass
```
scopatz: Awesome tests by the way!

xonsh/main.py
laerus: why does the `-i` option exists? isn't it the same as in ipython and python that runs the script and then gets into interactive mode

scopatz: This says it better than I can in a PR :) http://unix.stackexchange.com/a/46856/129048

tests/test_main.py
laerus: ah rly? i don't like this so much, i just abuse the fact that argparse throws a SystemExit when it doesn't like something. i tried checking out for error messages on stderr but i think nose messes with that, it will be easier with pytest

scopatz: Well, yeah, but it better than not having any of these tests at all ;).

xonsh/main.py
laerus: excuse me if i seem silly but i still don't get the function of `--interactive` 
when i run in **bash**:
 `xonsh --interactive tests/sample.xsh` 
i expect it to run the script and then get into 'interactive' mode with the xonsh prompt and repl
isnt that the use of `--interactive` ?

scopatz: Sadly, no.  The `--interactive` flag just means to start up the shell *like* you were in an interactive terminal.  This means to load the config.json and xonshrc files.  Then if a file is provided, the file is executed and the shell executes.  If no file is provided, the shell goes to an interactive terminal.  The reason this could theoretically be useful is if a user has some configuration that a particular script relies on.  It isn't meant for portability, but can be essential in some cases.

laerus: i'm prolly totally missing the point here, i thought the interactive flag in shells is the same with python and ipython REPL.

scopatz: It is more like the interactive flag for Bash, zsh, and fish

laerus: ooh this clarifies things, ty! one stackoverflow post is wrong about this and i got confused, it just says:
> Interactive: As the term implies: Interactive means that the commands are run with user-interaction from keyboard. E.g. the shell can prompt the user to enter input. 

so the failling test is actually invalid because i was expecting something different

scopatz: Ahh great!  Then we can have a different test.  Our `-i`, `-l` and `-c` flags all come from Bash, etc. for future reference.

laerus:
@scopatz i still feel there is a problem, `premain` sets `XONSH_INTERACTIVE` to False and returns. Then `main` only sets `XONSH_INTERACTIVE` to True if we get the actual prompt at the last elif clause. the elif clause that manages the case of XonshMode.script_from_file doesn't set `XONSH_INTERACTIVE` at all

xonsh/main.py
```diff
@@ -233,7 +230,7 @@ def main(argv=None):
         code = sys.stdin.read()
         run_code_with_cache(code, shell.execer, glb=shell.ctx, loc=None,
                             mode='exec')
-    else:
+    elif len(sys.argv) == 1:
```
scopatz: Shouldn't this just be else?

laerus: yes, that is just for precaution allthough this clause could actually go on the top of main with this condition cause the check is easy and this case is the most frequently used i suppose.

laerus:
another buggy cause

this works fine:
```
> python3 -m xonsh -DONE=100 -c 'print($ONE)'
100
```

but:
```
> XONSH_DEBUG=1 python3 -m xonsh -DONE=100 -DTWO=666 -c 'print($ONE, $TWO)'
  File "/usr/lib/python3.5/runpy.py", line 184, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/laerus/projects/xonsh/xonsh/__main__.py", line 2, in <module>
    main()
  File "/home/laerus/projects/xonsh/xonsh/main.py", line 219, in main
    run_code_with_cache(args.command.lstrip(), shell.execer, mode='single')
  File "/home/laerus/projects/xonsh/xonsh/codecache.py", line 191, in run_code_with_cache
  File "/home/laerus/projects/xonsh/xonsh/codecache.py", line 66, in run_compiled_code
    func(code, glb, loc)
  File "c4a4139eda522b534f58aac1258deae4", line 1, in <module>
  File "/home/laerus/projects/xonsh/xonsh/environ.py", line 746, in __getitem__
    raise KeyError(e.format(key))
KeyError: 'Unknown environment variable: $ONE'

```

`-D` *flag doesn't work as advertised*

laerus:
@scopatz do you think its worth trying to make premain a pure function? adding more and more tests point to that direction, premain needs decoupling from side-effects like creating the Shell and setting the enviroment. Or an alternative would be to remove all argparse logic from premain and into a new function, but thats about the same thing. also i still think $XONSH_INTERACTIVE is a little ambiguous, from prev comment: 

>  i still feel there is a problem, premain sets XONSH_INTERACTIVE to False and returns. Then main only sets XONSH_INTERACTIVE to True if we get the actual prompt at the last elif clause. the elif clause that manages the case of XonshMode.script_from_file doesn't set XONSH_INTERACTIVE at all

i'm talking here about xonsh from master not my branch

scopatz:
> do you think its worth trying to make premain a pure function?

This might be impossible overall because of the need to set things like `__xonsh_ctx__` and creating the environment, etc.  But if you think a higher level of purity would help, go for it. I actually think that the main parser could be a LazyObject.

scopatz:
But you don't need to worry about that right now.

scopatz:
I'd rather have something working first than worry about if it is being efficiently constructed.

laerus:
afaik the only case this doesn't fix atm is the multiple `-D` flags, i'll leave this branch as is if you guys want to test for merging at some point, this fixes a little more than what #1073 is about . i'd rather make a new branch for the refactoring part.

scopatz:
> i'd rather make a new branch for the refactoring part.

Much agreed!

tests/test_main.py
```diff
@@ -13,13 +13,24 @@
...
     with patch('xonsh.main.Shell', Shell), mock_xonsh_env({}):
         xonsh.main.premain([])
         assert_true(builtins.__xonsh_env__.get('XONSH_LOGIN'))
+        # assert_true(builtins.__xonsh_env__.get('XONSH_INTERACTIVE'))
+
+    # with patch('xonsh.main.Shell', Shell), mock_xonsh_env({}):
+    #     for case in ('-i', '-i', '-il' ):
+    #         xonsh.main.premain([])
+    #         assert_true(builtins.__xonsh_env__.get('XONSH_LOGIN'))
+    #         assert_true(builtins.__xonsh_env__.get('XONSH_INTERACTIVE'))
```
scopatz: please remove code comments like this..

tests/test_main.py
```diff
@@ -32,6 +43,38 @@ def Shell(*args, **kwargs):
     with patch('xonsh.main.Shell', Shell), mock_xonsh_env({}):
         xonsh.main.premain(['-l'])
         assert_true(builtins.__xonsh_env__.get('XONSH_LOGIN'))
+        # assert_true(builtins.__xonsh_env__.get('XONSH_INTERACTIVE'))
```
scopatz: Same here

scopatz:
I think that if you change the `-D` part of the parser by removing the `nargs` kwarg and adding the kwarg `action='append'`, it will work.

scopatz:
Other than that and the extra comments this looks good to go!

laerus:
allrighty i will try out if this can fix `-D`
yes the extra comments were to point out the problem with `XONSH_INTERACTIVE`
and the need for a refactor, if someone wanted to give feedback for a solution.

scopatz:
you should ope. another issue for the interactive / refactor stuff

scopatz:
*open

scopatz:
Is this ready @laerus?

laerus:
yes i think so, i have an extra local commit i didnt' push that changes the tests to pytest but i can do this on the pytest_port branch too, your choice!

scopatz:
Looks great to me @laerus!  Thanks for all the back and forth!

Issue #1268, opened 2016-06-17
Slow commands on Windows


JohnLunzer:
This just started today on the master branch I believe somewhere between 426b1ff and 1ba3b50 though it could have been a little earlier.

On the current master branch any subprocessed/aliased command is taking multiple seconds to execute, even something like `exit`. By comparison in the version of xonsh on pypi this is taking only a fraction of a second (a not really noticeable amount of time).

JohnLunzer:
Checking out old commits the furthest I have to go back is to af1f01c before the commands are not slow. After that they are either slow or I am getting a `cannot import 'exit'` error.

scopatz:
Hey @adqm, this seems to be caused by the pipelining fix that you put in.  Any idea why?

adqm:
@scopatz are you sure?  I can't possibly imagine that making a difference.  In particular, for commands that are not part of a pipeline (just single commands), this should be identical to the old behavior.

@JohnLunzer: do you mean that af1f01ce92324bf6eeb61f2f803943021209f333 is the last one that works at normal speed?  And then ones after that are slow?  Or do you mean that af1f01ce92324bf6eeb61f2f803943021209f333 itself is slow?

melund:
I don't see this on my end. @JohnLunzer Do you rebuild when changing between versions?

JohnLunzer:
@Adam af1f01c was the last that worked at normal speed.

I was it driving around and I think I realized the issue. I use conemu for my console on Windows and I have never been able to use the git status in my command prompt because it made every command take multiple seconds. I think this has something to do with how python and conemu interact when it comes to subprocesses. The pipeline fix must create a similar interaction. I guess I'll see if I can change a setting in conemu, otherwise I'll be looking for a new Windows console.

adqm:
If af1f01ce92324bf6eeb61f2f803943021209f333 was working, then this probably isn't the pipeline fix (as it was merged in in that commit).

My guess would be that somehow the "amalgamation" stuff is having an unintended side effect on your machine.

scopatz:
Ahh true.  Sorry I didn't grok your original comment.  Are the commands still slow if you start xonsh in debug mode:

```
$ with ${...}.swap(XONSH_DEBUG=1):
    xonsh
```

scopatz:
Also, does this happen on current master? We have resolved a bunch of latent amalgamation issues.

PR #1281, opened 2016-06-18
typo for cygwin job control

scopatz:
This might have been causing the slowdown in #1268, but is definitely a bug.  @JohnLunzer @gforsyth, would you mind taking a look?


scopatz:
Ping on this, since I think it is needed for v0.4

gforsyth:
Sorry about that, @scopatz!  Looks good on my end.

scopatz:
Thanks @gforsyth!

PR #1288, opened 2016-06-19
Fix some windows test errors

melund:
This fixes a some windows test errors.

Still two more left. @VHarisop . Can you readily see the problem, then you welcome to fix them.

````
======================================================================
FAIL: test_tools.test_is_logfile_opt(True, False)
  File "C:\Users\mel\Anaconda3\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
AssertionError: True != False

======================================================================
FAIL: test_tools.test_to_logfile_opt('/dev/null', None)
  File "C:\Users\mel\Anaconda3\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
AssertionError: '/dev/null' != None

````

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -472,7 +472,6 @@ def test_env_path_to_str():
 
 
 def test_env_path():
-    """Test the EnvPath class."""
     # lambda to expand the expected paths
     expand = lambda path: os.path.expanduser(os.path.expandvars(path))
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks @melund! Since this did fix most of the test errors other than the ones that you mentioned, I will go ahead and merge it

PR #1271, opened 2016-06-17
Fix aliases that reference them self

melund:
I think this should fix the problem with aliases that reference them self. #1240 

@scopatz  You had this problem as well, could you verify that this work. 

Also @gforsyth maybe you better check that it doesn't break something else for you.

PR #1290, opened 2016-06-19
restore alias preference in alias binary collision

gforsyth:
#1271 reintroduced a problem where if an alias has the same name as an existing binary, xonsh prefers the binary over the alias.  This goes against our "Python always wins" philosophy.  

@melund can you check this to make sure that self-referential aliases still work on your end?  I never had that problem so can't tell if this reintroduces the same problem.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -440,7 +440,7 @@ def run_subproc(cmds, captured=False):
                 aliased_cmd = alias + cmd[1:]
             else:
                 aliased_cmd = cmd
-            if binary_loc is not None:
+            if binary_loc is not None and alias is None:
                 try:
                     aliased_cmd = get_script_subproc_command(binary_loc,
                                                              aliased_cmd[1:])


scopatz:
Code looks good to me!

melund:
I have no time to test now. But I think it will not work for me. Popen on windows does not consider the content of the PATH variable. Thus, we always need to replace a command which its full paths when available (`ls` with `full/path/to/ls.exe`). It may be a windows issue only, but @scopatz also saw this on his laptop. 

`ls -l`
becomes:
`['ls','--show-control-chars','--color', '-X'] + ['-l']`
becomes:
`['c:/:Program Files/Git/usr/bin/ls.exe', '--show-control-chars','--color', '-X', '-l']`

I think there is something fundamental here we need to sort out. At least on Windows the first element should always be replaced by the binary location if it exists.  I have no time the next many days due to work. So I appreciate if someone else looked at this.

scopatz:
So sadly, this does seem to reintroduce the alias bug.

gforsyth:
This was fixed in #1320 in a non-breaking way

PR #1291, opened 2016-06-20
no `import *`.  not even once

gforsyth:
"I've seen things you people wouldn't believe. Attack ships on fire off the shoulder of Orion. I watched C-beams glitter in the dark near the Tannhäuser Gate. All those moments will be lost in time, like tears...in...rain. Time to die." - `import *`


scopatz:
"It's too bad she won't live! But then again, who does?"

xonsh/tokenize.py
```diff
@@ -24,7 +24,15 @@
 from itertools import chain
 import re
 import sys
-from token import *
+from token import (AMPER, AMPEREQUAL, ASYNC, AT, AWAIT, CIRCUMFLEX,
```
scopatz: `ASYNC` and `AWAIT` should only be imported on Python v3.5 and above.

gforsyth:
Ready for another look, @scopatz -- and I guess AppVeyor will keep failing since the windows tests are... well... failing

scopatz:
Thanks @gforsyth!

PR #1292, opened 2016-06-20
Ensure that Which can handle PATHEXT

melund:
This should ensure that `which` can handle the new PathEnv type that PATHEXT gets


scopatz:
Thanks @melund!

PR #1296, opened 2016-06-20
Binary locate speedups

scopatz:
This should hopefully address the slowdown issues on windows. @JohnLunzer and @melund could you give it a spin?

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -39,7 +39,8 @@ from xonsh.tools import (
     is_dynamic_cwd_width, to_dynamic_cwd_tuple, dynamic_cwd_tuple_to_str,
     is_logfile_opt, to_logfile_opt, logfile_opt_to_str, executables_in,
     pathsep_to_set, set_to_pathsep, pathsep_to_seq, seq_to_pathsep,
-    is_string_seq,
+    is_string_seq, is_nonstring_seq_of_strings, pathsep_to_upper_seq,
+    seq_to_upper_pathsep,
 )
 
 
@@ -112,7 +113,8 @@ DEFAULT_ENSURERS = {
     'MOUSE_SUPPORT': (is_bool, to_bool, bool_to_str),
     'MULTILINE_PROMPT': (is_string_or_callable, ensure_string, ensure_string),
     re.compile('\w*PATH$'): (is_env_path, str_to_env_path, env_path_to_str),
-    'PATHEXT': (is_string_seq, pathsep_to_seq, seq_to_pathsep),
+    'PATHEXT': (is_nonstring_seq_of_strings, pathsep_to_upper_seq,
+                seq_to_upper_pathsep),
     'PRETTY_PRINT_RESULTS': (is_bool, to_bool, bool_to_str),
     'PROMPT': (is_string_or_callable, ensure_string, ensure_string),
     'RAISE_SUBPROC_ERROR': (is_bool, to_bool, bool_to_str),
@@ -232,7 +234,7 @@ DEFAULT_VALUES = {
     'MOUSE_SUPPORT': False,
     'MULTILINE_PROMPT': '.',
     'PATH': PATH_DEFAULT,
-    'PATHEXT': ['.com', '.exe', '.bat', '.cmd'] if ON_WINDOWS else [],
+    'PATHEXT': ['.COM', '.EXE', '.BAT', '.CMD'] if ON_WINDOWS else [],
     'PRETTY_PRINT_RESULTS': True,
     'PROMPT': DEFAULT_PROMPT,
     'PUSHD_MINUS': False,
@@ -410,8 +412,9 @@ DEFAULT_DOCS = {
         configurable=False),
     'PATH': VarDocs(
         'List of strings representing where to look for executables.'),
-    'PATHEXT': VarDocs('Sequence of extention strings (eg, ".exe") for '
-                       'filtering valid executables by.'),
+    'PATHEXT': VarDocs('Sequence of extention strings (eg, ".EXE") for '
+                       'filtering valid executables by. Each element must be '
+                       'uppercase.'),
     'PRETTY_PRINT_RESULTS': VarDocs(
             'Flag for "pretty printing" return values.'),
     'PROMPT': VarDocs(
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -37,7 +37,8 @@ from xonsh.tools import (
     csv_to_bool_seq, bool_seq_to_csv, DefaultNotGiven, print_exception,
     setup_win_unicode_console, intensify_colors_on_win_setter, format_color,
     is_dynamic_cwd_width, to_dynamic_cwd_tuple, dynamic_cwd_tuple_to_str,
-    is_logfile_opt, to_logfile_opt, logfile_opt_to_str, executables_in
+    is_logfile_opt, to_logfile_opt, logfile_opt_to_str, executables_in,
+    pathsep_to_set, set_to_pathsep,
 )
 
 
@@ -110,7 +111,7 @@ DEFAULT_ENSURERS = {
     'MOUSE_SUPPORT': (is_bool, to_bool, bool_to_str),
     'MULTILINE_PROMPT': (is_string_or_callable, ensure_string, ensure_string),
     re.compile('\w*PATH$'): (is_env_path, str_to_env_path, env_path_to_str),
-    'PATHEXT': (is_string_set, csv_to_set, set_to_csv),
+    'PATHEXT': (is_string_set, pathsep_to_set, set_to_pathsep),
     'PRETTY_PRINT_RESULTS': (is_bool, to_bool, bool_to_str),
     'PROMPT': (is_string_or_callable, ensure_string, ensure_string),
     'RAISE_SUBPROC_ERROR': (is_bool, to_bool, bool_to_str),
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -816,17 +816,19 @@ def _yield_executables(directory, name):
 
 
 def locate_binary(name):
-    if os.path.isfile(name) and name != os.path.basename(name):
-        return name
-    directories = builtins.__xonsh_env__.get('PATH')
-    # Windows users expect to be able to execute files in the same directory
-    # without `./`
-    if ON_WINDOWS:
-        directories = [_get_cwd()] + list(directories)
-    try:
-        return next(chain.from_iterable(_yield_executables(directory, name) for
-                    directory in directories if os.path.isdir(directory)))
-    except StopIteration:
+    """Locates an executable on the file system."""
+    if os.path.isfile(name):
+        if ON_WINDOWS:
+            # Windows users expect to be able to execute files in the same
+            # directory without `./`
+            return os.path.abspath(os.path.relpath(name, _get_cwd()))
+        elif name != os.path.basename(name):
+            return name
+    cc = builtins.__xonsh_commands_cache__
+    if name in cc:
+        # can be lazy here since we know name is already available
+        return cc.lazyget(name)[0]
+    else:
         return None
 
 
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -837,11 +837,17 @@ def locate_binary(name):
     elif os.path.isfile(name) and name != os.path.basename(name):
         return name
     cc = builtins.__xonsh_commands_cache__
-    if name in cc:
+    if ON_WINDOWS:
+        upname = name.upper()
+        if upname in cc:
+            return cc.lazyget(upname)[0]
+        for ext in exts:
+            upnamext = upname + ext
+            if cc.lazyin(upnamext):
+                return cc.lazyget(upnamext)[0]
+    elif name in cc:
         # can be lazy here since we know name is already available
         return cc.lazyget(name)[0]
-    else:
-        return None
 
 
 def get_git_branch():
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -110,7 +110,7 @@ DEFAULT_ENSURERS = {
     'MOUSE_SUPPORT': (is_bool, to_bool, bool_to_str),
     'MULTILINE_PROMPT': (is_string_or_callable, ensure_string, ensure_string),
     re.compile('\w*PATH$'): (is_env_path, str_to_env_path, env_path_to_str),
-    'PATHEXT': (is_env_path, str_to_env_path, env_path_to_str),
+    'PATHEXT': (is_string_set, csv_to_set, set_to_csv),
     'PRETTY_PRINT_RESULTS': (is_bool, to_bool, bool_to_str),
     'PROMPT': (is_string_or_callable, ensure_string, ensure_string),
     'RAISE_SUBPROC_ERROR': (is_bool, to_bool, bool_to_str),
@@ -230,7 +230,8 @@ DEFAULT_VALUES = {
     'MOUSE_SUPPORT': False,
     'MULTILINE_PROMPT': '.',
     'PATH': PATH_DEFAULT,
-    'PATHEXT': (),
+    'PATHEXT': frozenset(['.com', '.exe', '.bat']) if ON_WINDOWS else \
+               frozenset(),
     'PRETTY_PRINT_RESULTS': True,
     'PROMPT': DEFAULT_PROMPT,
     'PUSHD_MINUS': False,
@@ -408,7 +409,8 @@ DEFAULT_DOCS = {
         configurable=False),
     'PATH': VarDocs(
         'List of strings representing where to look for executables.'),
-    'PATHEXT': VarDocs('List of strings for filtering valid executables by.'),
+    'PATHEXT': VarDocs('Set of extention strings (eg, ".exe") for filtering '
+                       'valid executables by.'),
     'PRETTY_PRINT_RESULTS': VarDocs(
             'Flag for "pretty printing" return values.'),
     'PROMPT': VarDocs(
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -38,7 +38,8 @@ from xonsh.tools import (
     setup_win_unicode_console, intensify_colors_on_win_setter, format_color,
     is_dynamic_cwd_width, to_dynamic_cwd_tuple, dynamic_cwd_tuple_to_str,
     is_logfile_opt, to_logfile_opt, logfile_opt_to_str, executables_in,
-    pathsep_to_set, set_to_pathsep,
+    pathsep_to_set, set_to_pathsep, pathsep_to_seq, seq_to_pathsep,
+    is_string_seq,
 )
 
 
@@ -111,7 +112,7 @@ DEFAULT_ENSURERS = {
     'MOUSE_SUPPORT': (is_bool, to_bool, bool_to_str),
     'MULTILINE_PROMPT': (is_string_or_callable, ensure_string, ensure_string),
     re.compile('\w*PATH$'): (is_env_path, str_to_env_path, env_path_to_str),
-    'PATHEXT': (is_string_set, pathsep_to_set, set_to_pathsep),
+    'PATHEXT': (is_string_seq, pathsep_to_seq, seq_to_pathsep),
     'PRETTY_PRINT_RESULTS': (is_bool, to_bool, bool_to_str),
     'PROMPT': (is_string_or_callable, ensure_string, ensure_string),
     'RAISE_SUBPROC_ERROR': (is_bool, to_bool, bool_to_str),
@@ -231,8 +232,7 @@ DEFAULT_VALUES = {
     'MOUSE_SUPPORT': False,
     'MULTILINE_PROMPT': '.',
     'PATH': PATH_DEFAULT,
-    'PATHEXT': frozenset(['.com', '.exe', '.bat']) if ON_WINDOWS else \
-               frozenset(),
+    'PATHEXT': ['.com', '.exe', '.bat', '.cmd'] if ON_WINDOWS else [],
     'PRETTY_PRINT_RESULTS': True,
     'PROMPT': DEFAULT_PROMPT,
     'PUSHD_MINUS': False,
@@ -410,8 +410,8 @@ DEFAULT_DOCS = {
         configurable=False),
     'PATH': VarDocs(
         'List of strings representing where to look for executables.'),
-    'PATHEXT': VarDocs('Set of extention strings (eg, ".exe") for filtering '
-                       'valid executables by.'),
+    'PATHEXT': VarDocs('Sequence of extention strings (eg, ".exe") for '
+                       'filtering valid executables by.'),
     'PRETTY_PRINT_RESULTS': VarDocs(
             'Flag for "pretty printing" return values.'),
     'PROMPT': VarDocs(
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -820,13 +820,19 @@ def _yield_executables(directory, name):
 
 def locate_binary(name):
     """Locates an executable on the file system."""
-    if os.path.isfile(name):
-        if ON_WINDOWS:
-            # Windows users expect to be able to execute files in the same
-            # directory without `./`
-            return os.path.abspath(os.path.relpath(name, _get_cwd()))
-        elif name != os.path.basename(name):
-            return name
+    if ON_WINDOWS:
+        # Windows users expect to be able to execute files in the same
+        # directory without `./`
+        cwd = _get_cwd()
+        if os.path.isfile(name):
+            return os.path.abspath(os.path.relpath(name, cwd))
+        exts = builtins.__xonsh_env__['PATHEXT']
+        for ext in exts:
+            namext = name + ext
+            if os.path.isfile(namext):
+                return os.path.abspath(os.path.relpath(namext, cwd))
+    elif os.path.isfile(name) and name != os.path.basename(name):
+        return name
     cc = builtins.__xonsh_commands_cache__
     if name in cc:
         # can be lazy here since we know name is already available
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -1476,17 +1476,22 @@ def expanduser_abs_path(inp):
     return os.path.abspath(os.path.expanduser(inp))
 
 
-class CommandsCache(abc.Set):
-    """A lazy cache representing the commands available on the file system."""
+class CommandsCache(abc.Mapping):
+    """A lazy cache representing the commands available on the file system.
+    The keys are the command names and the values a tuple of (loc, has_alias)
+    where loc is either a str pointing to the executable on the file system or
+    None (if no executable exists) and has_alias is a boolean flag for whether
+    the command has an alias.
+    """
 
     def __init__(self):
-        self._cmds_cache = frozenset()
+        self._cmds_cache = {}
         self._path_checksum = None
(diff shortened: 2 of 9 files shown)


xonsh/environ.py
```diff
@@ -816,17 +819,19 @@ def _yield_executables(directory, name):
...
-    # without `./`
-    if ON_WINDOWS:
-        directories = [_get_cwd()] + list(directories)
-    try:
-        return next(chain.from_iterable(_yield_executables(directory, name) for
-                    directory in directories if os.path.isdir(directory)))
-    except StopIteration:
+    """Locates an executable on the file system."""
+    if os.path.isfile(name):
+        if ON_WINDOWS:
```
melund: I think there is an issue here. Imagine we have a `strombidae.exe` in the current directory. Then the if should trigger no matter what if the user writes `strombidae` or `STROMBIDAE` or  `Strombidae.exe`. Basically, it should also be able to launch any executable with an extension in `PATHEXT`. This thing was previously handled by `executables_in` https://github.com/xonsh/xonsh/blob/master/xonsh/tools.py#L487

melund:
I was also wondering how it would handle several executables with different extensions. Which has priority. Maybe the order in PATHEXT has a meaning.

xonsh/environ.py
scopatz: Thanks for pointing this out @melund. Can you try it now, please?

scopatz:
> I was also wondering how it would handle several executables with different extensions. Which has priority. Maybe the order in PATHEXT has a meaning.

That is why I asked :)

xonsh/environ.py
```diff
@@ -820,13 +820,19 @@ def _yield_executables(directory, name):
...
-        elif name != os.path.basename(name):
-            return name
+    if ON_WINDOWS:
+        # Windows users expect to be able to execute files in the same
+        # directory without `./`
+        cwd = _get_cwd()
+        if os.path.isfile(name):
+            return os.path.abspath(os.path.relpath(name, cwd))
+        exts = builtins.__xonsh_env__['PATHEXT']
+        for ext in exts:
```
melund: So if PATHEXT was a sequence, then it would not be random if `strombidae.bat` or `strombidae.exe`. Maybe it was a mistake to make it a Set.

scopatz:
I can change it to a list if you think it is needed. Not a big deal.

xonsh/environ.py
scopatz: Yeah, just let me know what you think is best. I have very little sense of the windows conventions

xonsh/xoreutils/_which.py
```diff
@@ -191,21 +192,21 @@ def whichgen(command, path=None, verbose=0, exts=None):
     # Windows has the concept of a list of extensions (PATHEXT env var).
     if sys.platform.startswith("win"):
         if exts is None:
-            exts = os.environ.get("PATHEXT", "").split(os.pathsep)
+            exts = builtins.__xonsh_env__['PATHEXT']
```
melund: I think we should keep `_which.py` oblivious to xonsh, and just use it as a module. Then we still have the option to move it to an independent module. Before we just passed in `PathExts` through the `ext` argument to `_whichgen`

melund: Alternatively, it could probably be rewritten completely. But there is something about old well tested code.

gforsyth: I would strongly recommend not rewriting `which` again.  That never goes well.

xonsh/environ.py
melund: Yes. Sorry about that. Google reveals that it should be a Sequence. http://environmentvariables.org/PathExt

xonsh/xoreutils/_which.py
melund: I agree:-) We should not change the file unless really necessary.

melund:
I get this error when starting xonsh: 

```
((x)) C:\Users\mel\Documents\xonsh>xonsh
  File "C:\Users\mel\Anaconda3\envs\x\lib\runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Users\mel\Anaconda3\envs\x\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\mel\Documents\xonsh\xonsh\main.py", line 270, in <module>
    main()
  File "C:\Users\mel\Documents\xonsh\xonsh\main.py", line 214, in main
    args = premain(argv)
  File "C:\Users\mel\Documents\xonsh\xonsh\main.py", line 201, in premain
  File "C:\Users\mel\Documents\xonsh\xonsh\__amalgam__.py", line 16077, in __init__
  File "C:\Users\mel\Documents\xonsh\xonsh\__amalgam__.py", line 16119, in _init_environ
    debug_level=debug)
  File "C:\Users\mel\Documents\xonsh\xonsh\__amalgam__.py", line 15763, in __init__
    load_builtins(execer=self, config=config, login=login, ctx=xonsh_ctx)
  File "C:\Users\mel\Documents\xonsh\xonsh\__amalgam__.py", line 15640, in load_builtins
    builtins.default_aliases = builtins.aliases = Aliases(make_default_aliases())
  File "C:\Users\mel\Documents\xonsh\xonsh\__amalgam__.py", line 14432, in make_default_aliases
    if not locate_binary('sudo'):
  File "C:\Users\mel\Documents\xonsh\xonsh\__amalgam__.py", line 11424, in locate_binary
    if name in cc:
  File "C:\Users\mel\Documents\xonsh\xonsh\__amalgam__.py", line 6494, in __contains__
    return key in self.all_commands
  File "C:\Users\mel\Documents\xonsh\xonsh\__amalgam__.py", line 6514, in all_commands
    alss = builtins.aliases
AttributeError: module 'builtins' has no attribute 'aliases'
```

melund:
And without amal
```
((x)) C:\Users\mel\Documents\xonsh>xonsh
  File "C:\Users\mel\Anaconda3\envs\x\lib\runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Users\mel\Anaconda3\envs\x\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\mel\Documents\xonsh\xonsh\main.py", line 270, in <module>
    main()
  File "C:\Users\mel\Documents\xonsh\xonsh\main.py", line 214, in main
    args = premain(argv)
  File "C:\Users\mel\Documents\xonsh\xonsh\main.py", line 201, in premain
  File "C:\Users\mel\Documents\xonsh\xonsh\shell.py", line 45, in __init__
  File "C:\Users\mel\Documents\xonsh\xonsh\shell.py", line 87, in _init_environ
    debug_level=debug)
  File "C:\Users\mel\Documents\xonsh\xonsh\execer.py", line 43, in __init__
    load_builtins(execer=self, config=config, login=login, ctx=xonsh_ctx)
  File "C:\Users\mel\Documents\xonsh\xonsh\built_ins.py", line 711, in load_builtins
    builtins.default_aliases = builtins.aliases = Aliases(make_default_aliases())
  File "C:\Users\mel\Documents\xonsh\xonsh\aliases.py", line 555, in make_default_aliases
    if not locate_binary('sudo'):
  File "C:\Users\mel\Documents\xonsh\xonsh\environ.py", line 837, in locate_binary
    if name in cc:
  File "C:\Users\mel\Documents\xonsh\xonsh\tools.py", line 1531, in __contains__
    return key in self.all_commands
  File "C:\Users\mel\Documents\xonsh\xonsh\tools.py", line 1551, in all_commands
    alss = builtins.aliases
AttributeError: module 'builtins' has no attribute 'aliases'

```

xonsh/xoreutils/_which.py
scopatz: Well, I think that looking things up in `os.environ` is a bug.  Perhaps an unhittable one given the interface, but still.

scopatz: Note that we have made minor modifications to code that we have forked from IPython, such as inpectors, to be able to function with xonsh correctly.

gforsyth: If it needs changing (and it sounds like it does), maybe we can leave a standard comment/note on any lines that are xonsh-specific for anyone who wants to use it outside of xonsh?

melund: The `os.environ` swap is not nice. We should probably change that.

scopatz: Sure thing, any suggestions?

scopatz:
I believe that I have fixed the alias issue. Sorry about that.

xonsh/xoreutils/_which.py
melund: Are we talking about different things? I was not talking about `exts = os.environ.get("PATHEXT", "").split(os.pathsep)` We don't use that. `exts` are passed in thought the environments. It is the way `PATH` is passed in by chaning the os.environ temporarily.  I think we can wait with such a  change

melund:
Another error:
```
((x)) C:\Users\mel\Documents\xonsh>xonsh
  File "C:\Users\mel\Anaconda3\envs\x\lib\runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Users\mel\Anaconda3\envs\x\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\mel\Documents\xonsh\xonsh\main.py", line 270, in <module>
    main()
  File "C:\Users\mel\Documents\xonsh\xonsh\main.py", line 214, in main
    args = premain(argv)
  File "C:\Users\mel\Documents\xonsh\xonsh\main.py", line 201, in premain
  File "C:\Users\mel\Documents\xonsh\xonsh\shell.py", line 45, in __init__
  File "C:\Users\mel\Documents\xonsh\xonsh\shell.py", line 87, in _init_environ
    debug_level=debug)
  File "C:\Users\mel\Documents\xonsh\xonsh\execer.py", line 43, in __init__
    load_builtins(execer=self, config=config, login=login, ctx=xonsh_ctx)
  File "C:\Users\mel\Documents\xonsh\xonsh\built_ins.py", line 711, in load_builtins
    builtins.default_aliases = builtins.aliases = Aliases(make_default_aliases())
  File "C:\Users\mel\Documents\xonsh\xonsh\aliases.py", line 555, in make_default_aliases
    if not locate_binary('sudo'):
  File "C:\Users\mel\Documents\xonsh\xonsh\environ.py", line 837, in locate_binary
    if name in cc:
  File "C:\Users\mel\Documents\xonsh\xonsh\tools.py", line 1531, in __contains__
    return key in self.all_commands
  File "C:\Users\mel\Documents\xonsh\xonsh\tools.py", line 1569, in all_commands
  File "C:\Users\mel\Documents\xonsh\xonsh\tools.py", line 511, in executables_in
    yield from func(path)
  File "C:\Users\mel\Documents\xonsh\xonsh\tools.py", line 498, in _executables_in_windows
    for fname in (x.name for x in scandir(path) if x.is_file()):
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:\\Users\\mel\\Anaconda3\\envs\\x\\Library\\usr\\bin'
```

scopatz:
OK, I have pushed what I believe is a fix for the file not found issue.

xonsh/xoreutils/_which.py
```diff
@@ -191,16 +192,16 @@ def whichgen(command, path=None, verbose=0, exts=None):
...
             # If '.exe' is not in exts then obviously this is Win9x and
             # or a bogus PATHEXT, then use a reasonable default.
             for ext in exts:
                 if ext.lower() == ".exe":
                     break
             else:
-                exts = ['.COM', '.EXE', '.BAT']
-        elif not isinstance(exts, MutableSequence):
-            raise TypeError("'exts' argument must be a list or None")
+                exts = ['.com', '.exe', '.bat', '.cmd']
```
melund: These are always Upper case on windows: 
```
C:\Users\mel>set PATHEXT
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC
```

xonsh/xoreutils/_which.py
```diff
@@ -191,16 +192,16 @@ def whichgen(command, path=None, verbose=0, exts=None):
...
+        elif not isinstance(exts, abc.Sequence):
+            raise TypeError("'exts' argument must be a set or None")
```
melund: change to "'exts' argument must be a set or None"

xonsh/tools.py
```diff
@@ -954,6 +961,42 @@ def set_to_csv(x):
...
+    """
+    if sort:
+        x = sorted(x)
+    return os.pathsep.join(x)
+
+
+def is_string_seq(x):
+    """Tests if something is a sequence of strings"""
+    return (isinstance(x, abc.Sequence) and
+            all(isinstance(a, str) for a in x))
```
melund: This will return `True` for both `is_string_seq('HalloWorld')` and  `is_string_seq(['Hallo','World']')`

melund:
Ahh. It is the `is_string_seq` that is causing this.

xonsh/tools.py
```diff
@@ -485,23 +486,29 @@ def _executables_in_posix(path):
...
+    if not os.path.isdir(path):
+        return
+    extensions = builtins.__xonsh_env__['PATHEXT']
     if PYTHON_VERSION_INFO < (3, 5, 0):
         for fname in os.listdir(path):
             fpath = os.path.join(path, fname)
             if (os.path.exists(fpath) and not os.path.isdir(fpath)):
                 base_name, ext = os.path.splitext(fname)
-                if ext.upper() in extensions:
+                if ext.lower() in extensions:
```
melund: This should be: `if ext.lower() in (e.lower() for e in extensions)`

xonsh/tools.py
```diff
@@ -485,23 +486,29 @@ def _executables_in_posix(path):
...
     else:
-        for fname in (x.name for x in scandir(path) if x.is_file()):
+        for x in scandir(path):
+            if x.is_file():
+                fname = x.name
+            else:
+                continue
             base_name, ext = os.path.splitext(fname)
-            if ext.upper() in extensions:
+            if ext.lower() in extensions:
```
melund: `if ext.lower() in (e.lower() for e in extensions):`

xonsh/environ.py
```diff
@@ -816,17 +819,25 @@ def _yield_executables(directory, name):
...
+            if os.path.isfile(namext):
+                return os.path.abspath(os.path.relpath(namext, cwd))
+    elif os.path.isfile(name) and name != os.path.basename(name):
+        return name
+    cc = builtins.__xonsh_commands_cache__
+    if name in cc:
+        # can be lazy here since we know name is already available
+        return cc.lazyget(name)[0]
+    else:
         return None
```
melund: I think there is an issue here. The commands are stored like: 'ex.exe' in cc, but the argument to `locate_binary` could have the wrong case and no extension. For example: `eX` and it should still find `ex.exe`. 

We could change the command cache to be case insensitive on windows. And then change the last part of `locate_binary()` to this: 
```python
    cc = builtins.__xonsh_commands_cache__
    if ON_WINDOWS:
        for ext in [''] + exts:
            nameext = (name + ext).lower()
            if nameext in (c.lower() for c in cc):
                return cc.lazyget(namext)[0]
    else:
        if name in cc:
            # can be lazy here since we know name is already available
            return cc.lazyget(name)[0]
    return None
```

xonsh/xoreutils/_which.py
scopatz: I think we were talking about different things then.

scopatz:
Ok @melund - I have taken another shot at this by incorporating the changes you suggested @melund.  Let me know how it goes!

melund:
Now it works well :)  I am little curious about the speed. Running `locate_binary()` twice should show a big speed improvement on the second run? 
```
snail@sea ~ $ builtins.__xonsh_commands_cache__._path_checksum = None
snail@sea ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.04190334118978182
snail@sea ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.04187055111814164
```

scopatz:
> Now it works well :)

Finally!

> I am little curious about the speed. Running locate_binary() twice should show a big speed improvement on the second run?

It should only be much faster on the second run if the cache was invalid when the command first ran.  If the cache was valid the first time you should get basically the same time. This is just the time it takes to check that the cache is valid.

scopatz:
So in your example, your cache probably started off as valid because on windows the sudo alias construction at startup triggers the first cache creation.

melund:
True. But I added a `builtins.__xonsh_commands_cache__._path_checksum = None` before the first run.

xonsh/tools.py
```diff
@@ -1480,52 +1546,68 @@ def expanduser_abs_path(inp):
...
+        allcmds = {}
+        for path in reversed(paths):
+            # iterate backwards so that entries at the front of PATH overwrite
+            # entries at the back.
+            for cmd in executables_in(path):
+                key = cmd.upper() if ON_WINDOWS else cmd
+                allcmds[key] = (os.path.join(path, cmd), cmd in alss)
+        only_alias = (None, True)
+        for cmd in alss:
+            if cmd not in allcmds:
```
melund: So on Windows entries in `allcmds` are uppercase :/

scopatz:
Ahh fair, Yeah, it seems like the cache may be invalidated sometimes:

```
scopatz@artemis ~ $ import builtins
scopatz@artemis ~ $ builtins.__xonsh_commands_cache__._path_checksum = None
scopatz@artemis ~ $ from timeit import timeit
scopatz@artemis ~ $ builtins.__xonsh_commands_cache__._path_checksum = None
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.06382862000100431
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.022337049998895964
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.016513843002030626
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.01689089700084878
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.0648717990006844
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.06180991799919866
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.03276195599755738
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.01691527699949802
```

I bet that the frozenset isn't returning a consistently sorted value.

xonsh/tools.py
scopatz: Well, it is the most consistent with PATHEXT, and it is just the key. Note that the command itself is stored with the original casing.

adqm:
> I bet that the frozenset isn't returning a consistently sorted value.

If you look at the source code that does hashing for frozensets, I don't think that can happen.

melund:
I can never get a valid cache.

scopatz:
> If you look at the source code that does hashing for frozensets, I don't think that can happen.

Yeah, you are right.

xonsh/tools.py
```diff
@@ -1480,52 +1546,68 @@ def expanduser_abs_path(inp):
...
         cache_valid = cache_valid and al_hash == self._alias_checksum
         self._alias_checksum = al_hash
         # did the contents of any directory in PATH change?
         max_mtime = 0
-        for path in paths:
+        for path in pathset:
             mtime = os.stat(path).st_mtime
             if mtime > max_mtime:
                 max_mtime = mtime
         cache_valid = cache_valid and max_mtime > self._path_mtime
```
melund: Should this not be: `cache_valid = cache_valid and max_mtime >= self._path_mtime`

melund:
Seems to fix it:
``` 
snail@sea ~ $ builtins.__xonsh_commands_cache__._path_checksum = None
snail@sea ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.038149075517510296
snail@sea ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.00710991469050537
```

melund:
Awesome :)

xonsh/tools.py
scopatz: Wait, shouldn't it be `cache_valid = cache_valid and max_mtime <= self._path_mtime`

scopatz: Yup that is it!

scopatz:
Now seeing speedups like:

```console
scopatz@artemis ~ $ import builtins
scopatz@artemis ~ $ from timeit import timeit
scopatz@artemis ~ $ builtins.__xonsh_commands_cache__._path_checksum = None
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.016165105000254698
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.00016302600124618039
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.0003268709988333285
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.00016243599748122506
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.00016327100092894398

# touch a path element to invalidate cache
scopatz@artemis ~ $ touch ~/.local/bin
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.01779462299964507
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.00017087300147977658
scopatz@artemis ~ $ timeit('locate_binary("ex")',setup='from xonsh.environ import locate_binary', number=1)
0.00016280799900414422
```

I think this is working now :)

xonsh/tools.py
melund: Well. I mean. It will never find a hit since all entries in `allcmds` are uppercase, and the aliases are probably not.

scopatz: I am a little confused, what do you think won't find a hit?

melund:
There is just a single issue about looking aliases up in ´allcmds` on windows. See the inline comment https://github.com/xonsh/xonsh/pull/1296#discussion_r67939099

scopatz:
@melund - yeah, but lookup happens on uppercase names: https://github.com/xonsh/xonsh/pull/1296/files#diff-b08c376dd29024ccd6f9aef90f2ea96fR841

melund:
@scopatz ... Yes. It was the `only_alias` stuff I was referring to. 

Everyhting in `alass` are normal case. Everything in `allcmds` are upper case on windows. 
```python
        for cmd in alss:
             if cmd not in allcmds:
                   allcmds[cmd] = only_alias
```

scopatz:
Ahh, Ok, I think I see what you are saying, but I believe that this is the correct implementation. Say we have a command called `foo` and an alias called `foo`, we'll end up with two entries in the cache:

```python
{
'FOO': ('/path/to/foo', True),  # executable entry
'foo': (None, True),   # alias entry
}
```

This is OK because `run_subproc()` will run `locate_binary()` and find the `/path/to/foo` https://github.com/xonsh/xonsh/blob/master/xonsh/built_ins.py#L425

The `'foo': (None, True)` entry doesn't hurt anything.  I suppose this could look up `/path/to/foo` on windows if that is what you are concerned about. By taking the upper of the alias, we'd have, 

```python
{
'FOO': ('/path/to/foo', True),  # executable entry
'foo': ('/path/to/foo', True),   # alias entry
}
```

However, I think that might be the wrong thing to do because, what if you you had a command Bar and an alias bar, you would want

```python
{
'BAR': ('/path/to/Bar', False),  # executable entry
'bar': (None, True),   # alias entry
}
```

since `bar` should not call `Bar`. Or should it?

xonsh/xoreutils/_which.py
Carreau: > Note that we have made minor modifications to code that we have forked from IPython, such as inpectors, to be able to function with xonsh correctly.

We can try to extract these in separate packages. I know Xonsh vendors all necessary things, but IPython have lot's of dependencies, so I have no real issues with make separate pacakges.

melund:
Ok. I think I see what you mean, but I am not completly convinsed. Probably because I don't know how `CommandCache` is or will be used.   But I can't come up with a better idea. 

Two things bug me though. Firstly, the CommandCache will behave a little different between platforms, even though it 'doesn't hurt anathing'. Secondly, it left with the 'user' of CommandCache to know that binaries should be converted to upper-case before a lookup. I suspect that could lead to many future Windows related bugs when people start to use the cache and forget this fact. 

Unless you have a good idea to solve this now, I think it is ready to merge. Then we can handle the other thing i other PR

melund:
If the sole purpose of `CommandCache` is to look-up binaries. Then the conversion to upper-case could be handled internally in the class.

 In that case, we shouldn't need to store the alias information either. It could just be looked-up on the fly, and added to result returned by CommandCache. 

Somehting like:
```python
def __contains__(self, key):
    if ON_WINDOWS:
        key= key.upper()    
    return key in self.all_commands

def __getitem__(self, key):
    if ON_WINDOWS:
        item = self.all_commands[key.upper()]
    else:
        item = self.all_commands[key]
    is_alias = key in builtins.aliases
    return item, is_alias
```

Is this not relevant for other case-insensitive platforms besides windows? Or is windows the only one.

scopatz:
> Probably because I don't know how CommandCache is or will be used. ... Secondly, it left with the 'user' of CommandCache to know that binaries should be converted to upper-case before a lookup

Well, I have been under the impression that CommandsCache is a xonsh internal interface.  It isn't really public facing, which is why there are other APIs, like `locate_binary()`. So really users shouldn't be using it, and devs and reviewers should be able to lookup how it works.

> Unless you have a good idea to solve this now, I think it is ready to merge.

My current "big" idea here would be to move `locate_binary()` into CommandsCache as a method. But I don't think this is an immediate need. For me, the whole point of this PR was to hotfix the breakage on windows so that we could move to a v0.4. It wasn't to refactor the code base's APIs :)

> Is this not relevant for other case-insensitive platforms besides windows? Or is windows the only one.

I may be wrong about this, but I believe that Windows is only truly case-insensitive platform*.  Mac OSX's default filesystem is case-preserving, not case-insensitive. Everything else (that we support currently) is case-sensitive.

Of course, we can address further case-insensitive systems as needed.

* Also, OS/2, but I would only port xonsh to OS/2 as a joke...which means it will probably happen :)

melund:
Windows is also case preserving. You just can't have two files with the same name but different case. And when calling a file the case is irrelevant. I think this is good to go in. Further 'nice to haves' are for an other PR

melund:
Thanks a lot for doing this @scopatz. Now xonsh on Windows really works well again :)

scopatz:
Thanks a million for the thorough review @melund!

scopatz:
>  Now xonsh on Windows really works well again :)

Great! I think that we only have the alias preference issue left before we can cut a v0.4 :)

scopatz:
Which I think this can help with.

PR #1298, opened 2016-06-20
fixes for windows test failures

gforsyth:
I don't think the `/dev/null` test is a) necessary or b) meaningful on
Windows.  And it should fail, if it's passed, since `/dev/null` isn't
writeable or creatable file on Windows (I think).

I believe that the second set of tests that I've changed also have
little to no meaning on Windows.

Definitely need @melund to take a look at this when he has a chance.


melund:
This seems fine :) Thanks a lot @gforsyth

PR #1305, opened 2016-06-21
Only run appveyor on Py35

melund:
Only run appveyor on Py35. This should make it run a bit faster so we avoid waiting for appveyor all the time


gforsyth:
It passed the Appveyor tests!  Woohoo!

PR #1309, opened 2016-06-21
update readme links for organizational account

gforsyth:
trivial update of scopatz -> xonsh in readme links


scopatz:
Thanks @gforsyth!

PR #1311, opened 2016-06-21
fix PTK check and add warning if pygments not installed

gforsyth:
kept getting weird errors when pygments wasn't installed in an environment but prompt_toolkit was.  this should take care of those and provide something a little more helpful.

### DIFF
--- a/xonsh/platform.py
+++ b/xonsh/platform.py
@@ -81,7 +81,7 @@ def pygments_version():
 @functools.lru_cache(1)
 def has_prompt_toolkit():
     """ Tests if the `prompt_toolkit` is available. """
-    spec = importlib.util.find_spec('pygments')
+    spec = importlib.util.find_spec('prompt_toolkit')
     return (spec is not None)
 
 
--- a/xonsh/shell.py
+++ b/xonsh/shell.py
@@ -9,7 +9,8 @@ from xonsh.xontribs import update_context
 from xonsh.environ import xonshrc_context
 from xonsh.execer import Execer
 from xonsh.platform import (best_shell_type, has_prompt_toolkit,
-                            ptk_version_is_supported, ptk_version_info)
+                            ptk_version_is_supported, ptk_version_info,
+                            HAS_PYGMENTS)
 from xonsh.tools import XonshError, to_bool_or_int
 
 
@@ -59,6 +60,10 @@ class Shell(object):
             elif not ptk_version_is_supported():
                 warn('prompt-toolkit version < v1.0.0 is not supported. '
                      'Please update prompt-toolkit. Using readline instead.')
+                shell_type = 'readline'
+            elif not HAS_PYGMENTS:
+                warn('prompt_toolkit is available but requires pygments.  '
+                     'Please install pygments. Using readline shell')
                 shell_type = 'readline'
         env['SHELL_TYPE'] = shell_type
         # actually make the shell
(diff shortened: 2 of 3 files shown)


xonsh/shell.py
```diff
@@ -60,6 +61,10 @@ def __init__(self, ctx=None, shell_type=None, config=None, rc=None,
                 warn('prompt-toolkit version < v1.0.0 is not supported. '
                      'Please update prompt-toolkit. Using readline instead.')
                 shell_type = 'readline'
+            elif not HAS_PYGMENTS:
+                warn('prompt_toolkit is available but requires pygments.  '
```
scopatz: Prompt toolkit no longer requires pygments.  These are separate, optional dependencies now. (Though ptk will use pygments if it is available.)

xonsh/platform.py
```diff
@@ -81,7 +81,7 @@ def pygments_version():
 @functools.lru_cache(1)
 def has_prompt_toolkit():
     """ Tests if the `prompt_toolkit` is available. """
-    spec = importlib.util.find_spec('pygments')
+    spec = importlib.util.find_spec('prompt_toolkit')
```
scopatz: Doh!

xonsh/shell.py
gforsyth: Yeah, that's what got me started on this.  Should it instead warn that "xonsh requires pygments alongside prompt_toolkit"?

scopatz: well I would argue that we shouldn't require that, but that change might be more involved:)

xonsh/shell.py
```diff
@@ -60,6 +61,10 @@ def __init__(self, ctx=None, shell_type=None, config=None, rc=None,
...
+                     'Please install pygments. Using readline shell')
+                shell_type = 'readline'
```
adqm: If we have ptk but not pygments, I think we shouldn't need to fall all the way back to readline.  I think we may be able just to turn `$COLOR_INPUT` and `$COLOR_RESULTS` off in this case, and still to use ptk.

That way, we don't have to say that we are falling back to readline, but just that syntax highlighting has been disabled, or something along those lines.

laerus:
hey @gforsyth !

i'm getting this when only prompt_toolkit exist and no pygments:

```
$ XONSH_DEBUG=1 xonsh
/home/laerus/.venvs/xonsh/lib/python3.4/site-packages/xonsh/shell.py:65: UserWarning: pygments not found. syntax highlighting disabled.
  warn('pygments not found. syntax highlighting disabled.')
  File "/home/laerus/.venvs/xonsh/bin/xonsh", line 3, in <module>
    main()
  File "/home/laerus/.venvs/xonsh/lib/python3.4/site-packages/xonsh/main.py", line 246, in main
    shell.shell.cmdloop()
  File "/home/laerus/.venvs/xonsh/lib/python3.4/site-packages/xonsh/ptk/shell.py", line 130, in cmdloop
  File "/home/laerus/.venvs/xonsh/lib/python3.4/site-packages/xonsh/ptk/shell.py", line 77, in singleline
    rprompt_tokens = self.rprompt_tokens(None)
  File "/home/laerus/.venvs/xonsh/lib/python3.4/site-packages/xonsh/ptk/shell.py", line 169, in rprompt_tokens
    toks = pyghooks.partial_color_tokenize(p)
NameError: name 'pyghooks' is not defined
```

scopatz:
This might need to be merged with master as well.

gforsyth:
Hey @laerus -- I honestly can't even remember what I was doing here -- I'm going to abandon this and start fresh.

Issue #1252, opened 2016-06-17
Prompt disappears after running `source-bash $(docker-machine env default)`


mscottford:
After I run `source-bash $(docker-machine env default)` the prompt is no longer displayed. I initially thought that the `source-bash` command had just never returned. It took me a while to figure out the the `$PROMPT` variable had been set to `''`. Here's some shell output that shows what's going on.

```
Last login: Thu Jun 16 22:48:13 on ttys000

mscottford@blaster ~
$ xonfig
+------------------+----------+
| xonsh            | 0.3.4    |
| Python           | 3.5.1    |
| PLY              | 3.7      |
| have readline    | True     |
| prompt toolkit   | None     |
| shell type       | readline |
| pygments         | None     |
| on posix         | True     |
| on linux         | False    |
| on darwin        | True     |
| on windows       | False    |
| on cygwin        | False    |
| is superuser     | False    |
| default encoding | utf-8    |
+------------------+----------+

mscottford@blaster ~
$ $XONSH_SHOW_TRACEBACK=True

mscottford@blaster ~
$ source-bash $(docker-machine env default)

xonfig
+------------------+----------+
| xonsh            | 0.3.4    |
| Python           | 3.5.1    |
| PLY              | 3.7      |
| have readline    | True     |
| prompt toolkit   | None     |
| shell type       | readline |
| pygments         | None     |
| on posix         | True     |
| on linux         | False    |
| on darwin        | True     |
| on windows       | False    |
| on cygwin        | False    |
| is superuser     | False    |
| default encoding | utf-8    |
+------------------+----------+

ls
Applications
Creative Cloud Files
Desktop
Documents
Downloads
Library
Movies
Music
Pictures
Projects
Public

$PROMPT
''

$PROMPT = '\n{BOLD_GREEN}{user}@{hostname} {BOLD_BLUE}{cwd}{branch_color}{curr_branch}\n{prompt_end}{NO_COLOR} '

mscottford@blaster ~
$ 
```

scopatz:
Hi @mscottford - thanks for reporting!  This might be a silly question, but what command are you running in your container? Is it xonsh?  

We recently started removing PROMPT from foreign shell commands (and stopped inheriting it from the calling env.) because it was causing fatal errors in many cases. The xonsh prompt is not necessarily valid in any other shell.

[...]

PR #1313, opened 2016-06-22
fixed newline env issue

scopatz:
This potentially addresses #1252


scopatz:
OK, I have updated this to use pytest if someone would please take a look!

melund:
LGTM

PR #1316, opened 2016-06-22
Sort results from globbing expressions

### LINKED ISSUE #1314: Xonsh wildcards should sort lexicographically
Bash uses lexicographical sort when a list is created via a wildcard.   We should, too.

This is useful because:
* it is a bashwards-facing feature
* it is useful 
* its drawback is minor (sort time)
* the alternative (current state) is disorder, where typically order is desired
* ordering by name is a very reasonable default
* it is a de-facto standard of shell behaviour

usage example:
``` cat *.sql | mysql -u foo -p -- somedb ```

..where the '*.sql' files are prefixed by a numeric token, such as `001_cows.sql, 002_add_cow_stomachs.sql, 003_cow_currently_masticating_colum.sql`, or are prefixed by a datestring, such as `20120105_monthly_update.sql, 20120201_monthly_update.sql, [...]`

..this is particularly significant in cases where people have already structured their filenames for use in scripting, as the scripting in Xonsh would add a needless step.   Granted, this can be done in python anyways, but the simplest syntax is to implicitly sort on wildcards:
```python
cat *.sql | mysql -u foo -p -- somedb
//or
cat @(' '.join(sorted(f for f in $(ls -1 *.sql).split('\n')))) | mysql -u foo -p -- somedb
// or 
from glob import glob
cat sorted(glob('*.sql')) | mysql -u foo -p -- somedb
```

pgoelz:
Fixes #1314.

I have tested it on my machine (Python 3.5, arch linux).
I haven't yet tested that the envvar documentation will turn out okay.

Should this functionality be tested? I have written a probabilistic check and pasted it in the referenced issue, but the functionality depends on the underlying file system.


xonsh/tools.py
```diff
@@ -1686,15 +1686,24 @@ def globpath(s, ignore_case=False, return_empty=False):
...
+    glob_sorted = builtins.__xonsh_env__.get('GLOB_SORTED')
     if ignore_case:
         s = expand_case_matching(s)
     if sys.version_info > (3, 5):
         if '**' in s and '**/*' not in s:
             s = s.replace('**', '**/*')
         # `recursive` is only a 3.5+ kwarg.
-        return glob.iglob(s, recursive=True), s
+        if glob_sorted:
+            paths = sorted(glob.glob(s, recursive=True))
```
scopatz: This should actually return an iterator, since it is an iglob replacement.  Something that is a bit more efficient would be:

```python
paths = glob.glob(s, recursive=True)
paths.sort()
paths = iter(paths)
```

I think that will work.

scopatz:
Also, this could certainly use some tests (non-probabilistic ones, preferably). I think you just need to test that this works in both sorted and unsorted cases.

pgoelz:
The problem seems to be that using `mock_xonsh_env` deletes the environment after leaving the block, while tests like `test_is_env_path` depend on the environment being present.

@scopatz, can you give me a hint on how I should set environment variables instead?

xonsh/tools.py
```diff
@@ -1686,15 +1686,29 @@ def globpath(s, ignore_case=False, return_empty=False):
 
 def _iglobpath(s, ignore_case=False):
```
melund: Would it be an idea to add sorted kwarg to the function to override the environment variable e.g `sorted=None`. If `sorted` were `True` or `False` it would override the environment variable.

pgoelz: That should work, a normal optional argument probably would be easier.
Still, I am unsure whether it would be good style to introduce such an argument just for testing.
Do you think that this would be alright?

scopatz: Yeah, I would also be in favor of a `sorted` kwarg.  This wouldn't be just for testing.  You should then remove the envvar lookup from this function completely, and then look up the envvar at the call site of this function.

scopatz: At least that is how I would do it :)

melund: I agree. Keep the environment stuff out of the function.

scopatz:
Let us know when this is ready for another look!

pgoelz:
I'll do :)

I have done as you have suggested. However, since the version change, `builtins.__xonsh_expand_path__` is no longer defined.

Is this expected behavior? Do you have hints on how I should include it?

I tried to manually set it to `built_ins.expand_path`, but this in turn depends on the global variable `ENV` being not None. I am afraid that an ugly hack would make other tests unpredictable. Do you have advice?

scopatz:
@pgoelz - `__xonsh_expand_path__()` does seem to still be available: https://github.com/xonsh/xonsh/blob/master/xonsh/built_ins.py#L688

scopatz:
Are you not seeing it? Can you send a traceback, please?  Also, you are always free to do a `from xonsh.built_ins import expand_path`, if you just need to grab the function object

pgoelz:
In the console, I can see it just fine. During test execution, however, it does not seem to be defined yet, while this was the case before v0.4.0.
The traceback looks like this:

    tests/test_tools.py:352:
    _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
    […]/python3.5/site-packages/xonsh/__amalgam__.py:6604: in iglobpath
        return _iglobpath(s, ignore_case)[0]
    _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
    
    s = '/var/folders/hw/5vj86t952lq22j1c5mg99lx40000gn/T/tmpf1cukt1a/*.test', ignore_case = False
    
        def _iglobpath(s, ignore_case=False):
    >       s = builtins.__xonsh_expand_path__(s)
    E       AttributeError: module 'builtins' has no attribute '__xonsh_expand_path__'

That means that `__xonsh_expand_path__` is read inside of the tested function (and I do not think we should tear this functionality outside of it).
I could import `expand_path` and set `builtins.__xonsh_expand_path__` to it, but then `expand_path` crashes, because the global variable `ENV` is None, but expand_path calls `ENV.get('EXPAND_ENV_VARS')` on it.

tests/test_tools.py
```diff
@@ -28,6 +28,8 @@
 
 from tools import mock_xonsh_env
 
+from tools import mock_xonsh_env
```
scopatz: Why is this imported twice in a row?

xonsh/completers/path.py
```diff
@@ -105,6 +105,7 @@ def _add_cdpaths(paths, prefix):
     """Completes current prefix using CDPATH"""
     env = builtins.__xonsh_env__
     csc = env.get('CASE_SENSITIVE_COMPLETIONS')
+    glob_sorted = builtins.__xonsh_env__.get('GLOB_SORTED')
```
scopatz: `builtins.__xonsh_env__` should just be env

xonsh/completers/path.py
```diff
@@ -247,6 +249,7 @@ def complete_path(prefix, line, start, end, ctx, cdpath=True, filtfunc=None):
     paths = set()
     env = builtins.__xonsh_env__
     csc = env.get('CASE_SENSITIVE_COMPLETIONS')
+    glob_sorted = builtins.__xonsh_env__.get('GLOB_SORTED')
```
scopatz: Should just be env

xonsh/tools.py
```diff
@@ -1593,26 +1614,50 @@ def expand_case_matching(s):
...
+        sort_result = builtins.__xonsh_env__.get('GLOB_SORTED')
+
+    o, s = _iglobpath(s, ignore_case=ignore_case, sort_result=sort_result)
     o = list(o)
     no_match = [] if return_empty else [s]
     return o if len(o) != 0 else no_match
 
 
-def _iglobpath(s, ignore_case=False):
+def _iglobpath(s, sort_result, ignore_case=False):
```
scopatz: This should be a kwarg and should come after `ignore_case`

xonsh/tools.py
```diff
@@ -1593,26 +1614,50 @@ def expand_case_matching(s):
...
-def iglobpath(s, ignore_case=False):
+        return paths, s
+
+
+def iglobpath(s, ignore_case=False, sort_result=None):
     """Simple wrapper around iglob that also expands home and env vars."""
-    return _iglobpath(s, ignore_case)[0]
+    if sort_result is None:
+        sort_result = builtins.__xonsh_env__.get('GLOB_SORTED')
+
```
scopatz: please remove excess vertical whitespace

tests/test_tools.py
```diff
@@ -623,7 +653,8 @@ def test_env_path():
 
 def test_env_path_slices():
     # build os-dependent paths properly
-    mkpath = lambda *paths: os.sep + os.sep.join(paths)
+    def mkpath(*paths):
+        return os.sep + os.sep.join(paths)
```
scopatz: I don't think this is a needed change.

tests/test_tools.py
```diff
@@ -566,8 +595,9 @@ def test_env_path_to_str():
 
 
 def test_env_path():
-    # lambda to expand the expected paths
-    expand = lambda path: os.path.expanduser(os.path.expandvars(path))
+    def expand(path):
+        return os.path.expanduser(os.path.expandvars(path))
```
scopatz: Also unneeded

xonsh/tools.py
```diff
@@ -1593,26 +1614,50 @@ def expand_case_matching(s):
     return ''.join(t)
 
 
-def globpath(s, ignore_case=False, return_empty=False):
+def globpath(s, ignore_case=False, return_empty=False, sort_result=None):
     """Simple wrapper around glob that also expands home and env vars."""
-    o, s = _iglobpath(s, ignore_case=ignore_case)
+    if sort_result is None:
+        sort_result = builtins.__xonsh_env__.get('GLOB_SORTED')
+
```
scopatz: remove white space line please

xonsh/tools.py
```diff
@@ -1593,26 +1614,50 @@ def expand_case_matching(s):
...
+        return paths, s
     else:
-        return glob.iglob(s), s
+        if sort_result:
+            paths = glob.glob(s)
+            paths.sort()
+            paths = iter(paths)
+        else:
+            paths = glob.iglob(s)
```
scopatz: Does this no return now?

xonsh/tools.py
```diff
@@ -1593,26 +1614,50 @@ def expand_case_matching(s):
...
+
         # `recursive` is only a 3.5+ kwarg.
-        return glob.iglob(s, recursive=True), s
+        if sort_result:
+            paths = glob.glob(s, recursive=True)
+            paths.sort()
+            paths = iter(paths)
+        else:
+            paths = glob.iglob(s, recursive=True)
+
```
scopatz: Remove vertical whitespace please

news/sortglob.rst
```diff
@@ -0,0 +1,13 @@
+**Added:** None
+* The results of glob expressions are sorted if `$GLOB_SORTED` is set.
```
scopatz: Double backticks needed, extra line between title, and remove  None, please

news/sortglob.rst
```diff
@@ -0,0 +1,13 @@
...
+
+**Changed:** None
+* `GLOB_SORTED` is enabled by default.
```
scopatz: Same here

scopatz:
Hi @pgoelz! Thanks again.  I think your tests need to be put inside of a `with mock_xonsh_env()`, like some other tests in the suite.  The first line of the with block is where you would set `builtins.__xonsh_expand_path__ = xonsh.built_ins.expand_path`

xonsh/tools.py
pgoelz: kwarg as in `**kwargs` or as in `sort_result=None`?

tests/test_tools.py
pgoelz: I can happily revert, but PEP8 complained¹.
If this is a general thing, it might make sense to put this into the Developer's Guide.

---

<sup>¹ "Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier." [(source with reasons)](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)</sup>

xonsh/tools.py
pgoelz: Yes, it does, the github window is unfortunate:

        if sort_result:
            paths = glob.glob(s, recursive=True)
            paths.sort()
            paths = iter(paths)
        else:
            paths = glob.iglob(s, recursive=True)
        return paths, s

tests/test_tools.py
scopatz: @pgoelz OK, that is fair.

xonsh/tools.py
scopatz: `sort_result=None`

scopatz: I think the consistency maters.

pgoelz:
Thank you for your helpful suggestions :)

Using `mock_xonsh_env` does not solve the problem with the global variable and brings back the problems from before:
Tests test_repath_backslash, test_man_completion and test_to_logfile_opt depend somehow on implicit environment configuration in the tests:

    _______________________________________ test_repath_backslash ________________________________________
    
        @pytest.mark.skipif(ON_WINDOWS, reason='Unix stuff')
        def test_repath_backslash():
            home = os.path.expanduser('~')
            built_ins.ENV = Env(HOME=home)
            with mock_xonsh_env(built_ins.ENV):
                exp = os.listdir(home)
                exp = {p for p in exp if re.match(r'\w\w.*', p)}
                exp = {os.path.join(home, p) for p in exp}
                obs = set(pathsearch(regexsearch, r'~/\w\w.*'))
    >           assert exp ==  obs
    E           assert set([]) == set(['~/\\w\\w.*'])
    E             Extra items in the right set:
    E             '~/\\w\\w.*'
    E             Full diff:
    E             - set()
    E             + {'~/\\w\\w.*'}
    
    tests/test_builtins.py:33: AssertionError
    ________________________________________ test_man_completion _________________________________________
    
        @pytest.mark.skipif(ON_WINDOWS, reason='No man completions on Windows')
        def test_man_completion():
            with tempfile.TemporaryDirectory() as tempdir:
                with mock_xonsh_env({'XONSH_DATA_DIR': tempdir}):
                    completions = complete_from_man('--', 'yes --', 4, 6, __xonsh_env__)
    >           assert ('--version' in completions)
    E           assert '--version' in set([])

scopatz:
Oh yeah, you can't set `built_ins.ENV` this is a global that should be removed.  Any envvar that you want to set, you should set in `mock_xonsh_env`, ie:

```python
with mock_xonsh_env({'HOME': home}):
```

or 

```python
with mock_xonsh_env(Env(HOME=home)):
```

pgoelz:
The problems is that `expand_path` calls the global variable directly: 
https://github.com/xonsh/xonsh/blob/849bf3ebf7cee7debbe5d82241f8274df37d8607/xonsh/built_ins.py#L90
Setting `EXPAND_ENV_VARS` is not enough.
Should I change `expand_path` in any way?

scopatz:
Yes, please change `expand_path()` to use `env = builtins.__xonsh_env__` instead of `global ENV`

pgoelz:
Thanks, that did the trick :)
I will still have to investigate why those three tests behave differently if my test runs alongside them.

scopatz:
Thanks @pgoelz! Just waiting on CI!

scopatz:
Thanks @pgoelz! This is great to have!

pgoelz:
… okay, I'll have to check my testing setup.
In a fresh docker environment, I still have those three cases failing 😕 

@scopatz, thank you very much for your help!

PR #1319, opened 2016-06-23
merge from master and replace some new nose test with pytest

laerus:
atm the only failling test i see in python3.5 is:

```
        for inp, exp in getitem_cases:
            obs = EnvPath(inp)[0]  # call to __getitem__
>           assert expand(exp) == obs
E           assert '/home/laerus/' == '~/'
E             - /home/laerus/
E             + ~/

tests/test_tools.py:572: AssertionError
```
which seems like a legit bug in EnvPath

### DIFF
--- a/tests/test_imphooks.py
+++ b/tests/test_imphooks.py
@@ -10,7 +10,9 @@ from xonsh.built_ins import load_builtins, unload_builtins
 from tools import mock_xonsh_env
 LOADED_HERE = False
 
-def setup_module():
+IMP_ENV = {'PATH': [], 'PATHEXT': []}
+
+def setup():
     global LOADED_HERE
     if built_ins.BUILTINS_LOADED:
         unload_builtins()  # make sure we have a clean env from other tests.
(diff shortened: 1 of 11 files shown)


tests/test_imphooks.py
```diff
@@ -10,7 +10,9 @@
 from tools import mock_xonsh_env
 LOADED_HERE = False
 
-def setup_module():
+IMP_ENV = {'PATH': [], 'PATHEXT': []}
+
+def setup():
```
gforsyth: should this be `setup_module` for pytest?

laerus: it was a conflict with the master merge and i kept the master part, will change it

tests/tools.py
```diff
@@ -22,6 +22,7 @@
 VER_MAJOR_MINOR = sys.version_info[:2]
 VER_FULL = sys.version_info[:3]
 ON_MAC = (platform.system() == 'Darwin')
+ON_WINDOWS = (platform.system() == 'Windows')
```
gforsyth: shoudl we just import this from `xonsh.platform`?  same for `ON_DARWIN` (instead of `ON_MAC`)

laerus: on it!

gforsyth:
Awesome.  Thanks @laerus !

PR #1320, opened 2016-06-23
fix alias preference issue

scopatz:
This is an alternative to #1290 that looks up the binary location of the *alias*, if applicable, not the binary location of the original command.

With some of the new command cache behaviour, this is a little more tricky to trigger.  I have been using the following: `aliases['/bin/ls'] = ['echo', '.']`  This should print a single dot, rather than the contents of the current directory:

```console
scopatz@artemis ~ $ aliases['/bin/ls'] = ['echo', '.']
scopatz@artemis ~ $ c = !(/bin/ls); c
CompletedCommand(stdin=None, stdout='.\n', stderr='', pid=7128, returncode=0, args=['/bin/ls'], 
    alias=['echo', '.'], stdin_redirect=None, stdout_redirect=None, stderr_redirect=None, 
    timestamp=(1466649889.561564, 1466649889.6136434), executed_cmd=['/bin/echo', '.'])
scopatz@artemis ~ $ /bin/ls
.
```

Note that in the above, the `executed_cmd` starts with `/bin/echo`, not `/bin/ls`.  Therefore we can be sure that the alias is taking precedence. On master, this is `/bin/ls`

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -417,14 +417,18 @@ def run_subproc(cmds, captured=False):
         elif builtins.__xonsh_stderr_uncaptured__ is not None:
             stderr = builtins.__xonsh_stderr_uncaptured__
         uninew = (ix == last_cmd) and (not _capture_streams)
-
+        # find alias
         if callable(cmd[0]):
             alias = cmd[0]
         else:
             alias = builtins.aliases.get(cmd[0], None)
-            binary_loc = locate_binary(cmd[0])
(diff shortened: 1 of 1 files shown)


gforsyth:
Works for me!  Appveyor decided to hang in a random place, so if @melund can confirm that this works on his end I think we're good to merge.

melund:
Tests works fine.

scopatz:
Thanks!

PR #1325, opened 2016-06-23
Pytest port

melund:
This fixes a failing test on my machine where `XONSH_ENCODING_ERROR` was missing in the mocked environment. 

It also adds a better integration of the assertion check_ast assertion functions, as well as getting rid of all the yield statements.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -27,6 +27,7 @@ def setup_module():
                     yacc_table='parser_test_table')
 
 def nodes_equal(x, y):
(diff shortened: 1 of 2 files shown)


gforsyth:
Thanks, @melund!  Curious why AppVeyor didn't have the same issue with the encoding error.  I guess we just need to add more tests?

melund:
I don't know either. Maybe because I ran it with `$XONSH_DEBUG=1`? If that can cause issues, then maybe we should do both for the CI tests.

PR #1326, opened 2016-06-23
fix `has_prompt_toolkit()`

gforsyth:
this same fix is in #1311 but I thought this should go in while I get the pygments fallback stuff working


melund:
LGTM

scopatz:
Thanks @gforsyth!

PR #1329, opened 2016-06-23
Make CommandsCache play well with Windows

BYK:
Currently when you do `'git' in builtins.__xonsh_commands_cache__` on Windows you get `False` since the version stored in commands cache is `GIT.EXE`.

This patch amends its getter methods to respond to any version available on Windows. This fixes `curr_branch` not working at all on Windows platform since it only looks for `'git'` in command cache and bailds if its not there.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -9,7 +9,6 @@ import builtins
 from collections import Sequence
 from contextlib import contextmanager
 import inspect
-from glob import iglob
 import os
 import re
 import shlex
@@ -21,7 +20,6 @@ import time
 
 from xonsh.lazyasd import LazyObject
 from xonsh.history import History
-from xonsh.tokenize import SearchPath
 from xonsh.inspectors import Inspector
 from xonsh.aliases import Aliases, make_default_aliases
 from xonsh.environ import Env, default_env, locate_binary
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -455,8 +455,8 @@ def run_subproc(cmds, captured=False):
         if (stdin is not None and
                 ENV.get('XONSH_STORE_STDIN') and
                 captured == 'object' and
-                'cat' in __xonsh_commands_cache__ and
-                'tee' in __xonsh_commands_cache__):
+                __xonsh_commands_cache__.locate_binary('cat') and
+                __xonsh_commands_cache__.locate_binary('tee')):
             _stdin_file = tempfile.NamedTemporaryFile()
             cproc = Popen(['cat'],
                           stdin=stdin,
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -32,9 +32,10 @@ from xonsh.proc import (ProcProxy, SimpleProcProxy, ForegroundProcProxy,
                         SimpleForegroundProcProxy, TeePTYProc,
                         CompletedCommand, HiddenCompletedCommand)
 from xonsh.tools import (
-    suggest_commands, expandvars, CommandsCache, globpath, XonshError,
+    suggest_commands, expandvars, globpath, XonshError,
     XonshCalledProcessError, XonshBlockError
 )
+from xonsh.commands_cache import CommandsCache
 
 
 ENV = None
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -455,8 +455,8 @@ def run_subproc(cmds, captured=False):
         if (stdin is not None and
                 ENV.get('XONSH_STORE_STDIN') and
                 captured == 'object' and
-                __xonsh_commands_cache__.locate_binary('cat') and
-                __xonsh_commands_cache__.locate_binary('tee')):
+                __xonsh_commands_cache__.lazy_locate_binary('cat') and
+                __xonsh_commands_cache__.lazy_locate_binary('tee')):
             _stdin_file = tempfile.NamedTemporaryFile()
             cproc = Popen(['cat'],
                           stdin=stdin,
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -1,0 +1,140 @@
+import os
+from collections import abc as abc
+
+from xonsh.dirstack import _get_cwd
+from xonsh.platform import ON_WINDOWS
+from xonsh.tools import executables_in
+
+
+class CommandsCache(abc.Mapping):
+    """A lazy cache representing the commands available on the file system.
+    The keys are the command names and the values a tuple of (loc, has_alias)
+    where loc is either a str pointing to the executable on the file system or
+    None (if no executable exists) and has_alias is a boolean flag for whether
+    the command has an alias.
+    """
+
+    def __init__(self):
+        self._cmds_cache = {}
+        self._path_checksum = None
+        self._alias_checksum = None
+        self._path_mtime = -1
+
+    def __contains__(self, key):
+        return key in self.all_commands
+
+    def __iter__(self):
+        return iter(self.all_commands)
+
+    def __len__(self):
+        return len(self.all_commands)
+
+    def __getitem__(self, key):
+        return self.all_commands[key]
+
+    def is_empty(self):
+        """Returns whether the cache is populated or not. Does not check cache validity."""
+        return len(self._cmds_cache) == 0
+
+    @staticmethod
+    def get_possible_names(name):
+        """Generates the possible `PATHEXT` extension variants of a given executable
+         name on Windows as a list, conserving the ordering in `PATHEXT`.
+         Returns a list as `name` being the only item in it on other platforms."""
+        if ON_WINDOWS:
+            name = name.upper()
+            return [
+                name + ext
+                for ext in ([''] + builtins.__xonsh_env__['PATHEXT'])
+            ]
+        else:
+            return [name]
+
+    @property
+    def all_commands(self):
+        paths = builtins.__xonsh_env__.get('PATH', [])
+        pathset = frozenset(x for x in paths if os.path.isdir(x))
+        # did PATH change?
+        path_hash = hash(pathset)
+        cache_valid = path_hash == self._path_checksum
+        self._path_checksum = path_hash
+        # did aliases change?
+        alss = getattr(builtins, 'aliases', set())
+        al_hash = hash(frozenset(alss))
+        cache_valid = cache_valid and al_hash == self._alias_checksum
+        self._alias_checksum = al_hash
+        # did the contents of any directory in PATH change?
+        max_mtime = 0
+        for path in pathset:
+            mtime = os.stat(path).st_mtime
+            if mtime > max_mtime:
+                max_mtime = mtime
+        cache_valid = cache_valid and (max_mtime <= self._path_mtime)
+        self._path_mtime = max_mtime
+        if cache_valid:
+            return self._cmds_cache
+        allcmds = {}
+        for path in reversed(paths):
+            # iterate backwards so that entries at the front of PATH overwrite
+            # entries at the back.
+            for cmd in executables_in(path):
+                key = cmd.upper() if ON_WINDOWS else cmd
+                allcmds[key] = (os.path.join(path, cmd), cmd in alss)
+        only_alias = (None, True)
+        for cmd in alss:
+            if cmd not in allcmds:
+                allcmds[cmd] = only_alias
+        self._cmds_cache = allcmds
+        return allcmds
+
+    def lazyin(self, key):
+        """Checks if the value is in the current cache without the potential to
+        update the cache. It just says whether the value is known *now*. This
+        may not reflect precisely what is on the $PATH.
+        """
+        return key in self._cmds_cache
+
+    def lazyiter(self):
+        """Returns an iterator over the current cache contents without the
+        potential to update the cache. This may not reflect what is on the
+        $PATH.
+        """
+        return iter(self._cmds_cache)
+
+    def lazylen(self):
+        """Returns the length of the current cache contents without the
+        potential to update the cache. This may not reflect precisely
+        what is on the $PATH.
+        """
+        return len(self._cmds_cache)
+
+    def lazyget(self, key, default=None):
+        """A lazy value getter."""
+        return self._cmds_cache.get(key, default)
+
+    def locate_binary(self, name):
+        """Locates an executable on the file system using the cache."""
+        # make sure the cache is up to date by accessing the property
+        _ = self.all_commands
+        return self.lazy_locate_binary(name)
+
+    def lazy_locate_binary(self, name):
+        """Locates an executable in the cache, without checking its validity."""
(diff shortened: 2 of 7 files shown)


xonsh/environ.py
```diff
@@ -823,31 +823,27 @@ def _yield_executables(directory, name):
...
+
+    cc = builtins.__xonsh_commands_cache__
+
     if ON_WINDOWS:
         # Windows users expect to be able to execute files in the same
         # directory without `./`
         cwd = _get_cwd()
         if os.path.isfile(name):
             return os.path.abspath(os.path.relpath(name, cwd))
+
```
scopatz: No need for all of this whitespace :)

xonsh/tools.py
```diff
@@ -1560,8 +1560,18 @@ def __init__(self):
         self._alias_checksum = None
         self._path_mtime = -1
 
+    def __get_possible_names(self, key):
```
scopatz: Please don't start attrs or methods with double underscore because then they are name mangled by Python and live only on the class

scopatz: See https://docs.python.org/3.5/tutorial/classes.html#private-variables

> Any identifier of the form __spam (at least two leading underscores, at most one trailing underscore) is textually replaced with _classname__spam, where classname is the current class name with leading underscore(s) stripped. This mangling is done without regard to the syntactic position of the identifier, as long as it occurs within the definition of a class.

BYK: I know, this was my intention but then I decided I don't really need a private method :)

scopatz:
Thanks for opening this @BYK!

xonsh/built_ins.py
```diff
@@ -455,8 +455,8 @@ def run_subproc(cmds, captured=False):
         if (stdin is not None and
                 ENV.get('XONSH_STORE_STDIN') and
                 captured == 'object' and
-                'cat' in __xonsh_commands_cache__ and
-                'tee' in __xonsh_commands_cache__):
+                __xonsh_commands_cache__.locate_binary('cat') and
+                __xonsh_commands_cache__.locate_binary('tee')):
```
scopatz: These should be lazy, right?

xonsh/tools.py
```diff
@@ -1560,6 +1560,17 @@ def __init__(self):
...
         self._path_mtime = -1
 
+    @staticmethod
+    def get_possible_names(key):
+        if ON_WINDOWS:
+            key = key.upper()
+            return {
+                key + ext
+                for ext in ([''] + builtins.__xonsh_env__['PATHEXT'])
+            }
```
scopatz: This should not be a set operation because PATHEXT is ordered by the windows spec.

xonsh/tools.py
```diff
@@ -1634,6 +1647,38 @@ def lazyget(self, key, default=None):
...
+        """Locates an executable in the cache."""
+
+        possibilities = self.get_possible_names(name)
+        matches = possibilities & self._cmds_cache.keys()
+
+        if matches:
+            return self._cmds_cache[matches.pop()][0]
+        elif os.path.isfile(name) and name != os.path.basename(name):
+            return name
+
```
scopatz: Lots of extra vertical whitespace in these functions.

xonsh/tools.py
```diff
@@ -1634,6 +1647,38 @@ def lazyget(self, key, default=None):
...
+        possibilities = self.get_possible_names(name)
+        matches = possibilities & self.all_commands.keys()
+
+        if matches:
+            return self.all_commands[matches.pop()][0]
+        elif os.path.isfile(name) and name != os.path.basename(name):
+            return name
+
+    def lazy_locate_binary(self, name):
+        """Locates an executable in the cache."""
```
scopatz: I think these two docstrings could be a little more descriptive and distinct

xonsh/tools.py
```diff
@@ -1634,6 +1647,38 @@ def lazyget(self, key, default=None):
...
+
+        possibilities = self.get_possible_names(name)
+        matches = possibilities & self._cmds_cache.keys()
```
scopatz: This is a pretty inefficient way to check if a single name is in the cache.  Assuming possibilities returns a list or names, a faster implementation would be:

```python
match = None
for n in self.get_possible_names(name):
    if n in self._cmds_cache:
        match = n
        break
```

xonsh/tools.py
```diff
@@ -1634,6 +1647,38 @@ def lazyget(self, key, default=None):
...
+    def locate_binary(self, name):
+        """Locates an executable on the file system."""
+
+        if ON_WINDOWS:
+            # Windows users expect to be able to execute files in the same
+            # directory without `./`
+            cwd = _get_cwd()
+
+            local_bins = [full_name for full_name in self.get_possible_names(name)
+                          if os.path.isfile(full_name)]
```
scopatz: This should only have to find the first match that is a file, not test all of the possible matches for being a file each time.

xonsh/tools.py
```diff
@@ -1634,6 +1647,38 @@ def lazyget(self, key, default=None):
...
+            if local_bins:
+                return os.path.abspath(os.path.relpath(local_bins[0], cwd))
+
+        possibilities = self.get_possible_names(name)
+        matches = possibilities & self.all_commands.keys()
+
+        if matches:
+            return self.all_commands[matches.pop()][0]
```
scopatz: This function should only access `self.all_commands` once, to prevent excessive re-computation of the cache validity.

xonsh/built_ins.py
scopatz: Or really, it maybe should be lazyin() thus actual path doen't matter here at all.  Or you could revert it to what it was.

BYK: Okay with `lazy` variant. Was mostly trying to keep it consistent with the current version: `in` is not lazy right?

xonsh/tools.py
BYK: Oh, you mean this ordering actually conveys a preference! Makes sense, will change.

xonsh/tools.py
BYK: Hah, I was thinking about generators and then ended up with a list. Will optimize :)

xonsh/tools.py
BYK: Good catch, will fix.

xonsh/tools.py
BYK: Sure.

xonsh/tools.py
BYK: Why would it be inefficient? I'm assuming set intersection does the same. Building a new set and using `pop` on it is rather inefficient though, yes.

xonsh/tools.py
BYK: I find new lines making functions easier to follow by sectioning them. Will try to keep them minimal but TBH as an outsider, I find most of Xonsh code hard to follow due to this reason.

xonsh/tools.py
scopatz: Yeah, it is the building of the set of keys that takes up the time, coupled with the fact that dictionary lookup is one of the fastest things that you can do in Python  Here is some simple timeit:

```
scopatz@artemis ~ $ timeit "next((x for x in range(1000) if x == 1), None)"
1000000 loops, best of 3: 600 ns per loop
scopatz@artemis ~ $ d = {i: i for i in range(1000)}
scopatz@artemis ~ $ p = {1}
scopatz@artemis ~ $ timeit "p & d.keys()"
100000 loops, best of 3: 11.5 µs per loop
scopatz@artemis ~ $ 11.5e3 / 600
19.166666666666668
scopatz@artemis ~ $ timeit "next((x for x in d if x == 1), None)"
1000000 loops, best of 3: 392 ns per loop
scopatz@artemis ~ $ 11.5e3 / 392
29.336734693877553
```

So this is about ~20x - 30x slower with a len-1000 sized cache.  Note that my cache is about 3600.

xonsh/tools.py
```diff
@@ -1572,6 +1574,20 @@ def __len__(self):
     def __getitem__(self, key):
         return self.all_commands[key]
 
+    def is_empty(self):
```
scopatz: This needs a docstring.

xonsh/tools.py
scopatz: I find them hugely distracting and harder to understand algorithms because less code fits on a page :). Also, PEP8 basically says to leave them out.  I think that this is mostly a style thing, so I defer to PEP8

melund: So do I. If you need to section different areas to make it easier to understand, you can do it with a comment line.

melund:
Thanks  for tackling this @BYK. I will give it a thorough look when you push again.

BYK:
@melund @scopatz - Ready for the next round. I really don't agree with you guys' "no empty lines" thing but I'll avoid them as much as possible since this is 1 against 2 :D (also yay consistency!)

xonsh/tools.py
```diff
@@ -1648,34 +1650,36 @@ def lazyget(self, key, default=None):
...
-                return os.path.abspath(os.path.relpath(local_bins[0], cwd))
+            local_bin = next((
+                full_name for full_name in possibilities
+                if os.path.isfile(full_name)
+            ), None)
 
-        possibilities = self.get_possible_names(name)
-        matches = possibilities & self.all_commands.keys()
+            if local_bin:
+                return os.path.abspath(os.path.relpath(local_bin, cwd))
```
melund: Should it not be enough to call `os.path.abspath(local_bin)`. What does the `relpath` do?

BYK: Don't have an opinion or preference on this. Just copied what the old code did :)

xonsh/tools.py
```diff
@@ -1648,34 +1650,36 @@ def lazyget(self, key, default=None):
...
 
-        if matches:
-            return self.all_commands[matches.pop()][0]
+        cmds = self.all_commands
+        cached = next((cmd for cmd in possibilities if cmd in cmds), None)
```
melund: Since you just accessed `self.all_commands` I think you can use the lazy methods here: 
`cached = next((cmd for cmd in possibilities if cmd in self._cmds_cache), None)`

xonsh/tools.py
```diff
@@ -1648,34 +1650,36 @@ def lazyget(self, key, default=None):
...
+        if cached:
+            return cmds[cached][0]
```
melund: Same here: `return self._cmds_cache[cached][0]`

xonsh/tools.py
BYK: Ah, very good point. Will do.

BYK: Well, on second thought not sure if that'd provide any benefits. And it would make the code look a bit more complex. What do you think?

xonsh/tools.py
```diff
@@ -1648,34 +1650,36 @@ def lazyget(self, key, default=None):
...
         elif os.path.isfile(name) and name != os.path.basename(name):
             return name
 
     def lazy_locate_binary(self, name):
```
melund: Should `lazy_locate_binary` not also look in the current directory, similar to what `locate_binary` does. I they share so much code that you could potentially call `lazy_locate_binary()` from `locate_binary()`. With the addition that `locate_binary()` should just update the command cache first.

BYK: >Should lazy_locate_binary not also look in the current directory, similar to what locate_binary does.

I'm not so sure about that. I feel like `lazy_` methods should be very fast and directory look up is not fast. It's also not using the cache so I decided not to add that logic here.

>I think they share so much code that you could potentially call lazy_locate_binary() from locate_binary(). With the addition that locate_binary() should just update the command cache first. 

This also occurred to me but I didn't spend too much time to find a good way to merge them. If you feel strongly I'll think harder ;)

melund:
Argghh. Did I just comment on something that was outdated. Hmm. I will just redo my comments. Sorry :)

BYK:
>Argghh. Did I just comment on something that was outdated. Hmm. I will just redo my comments. Sorry :)

Not sure if the places you commented on were outdated and they were certainly useful :)

melund:
It is the same code. I was just looking at old commits, so GitHub hides the comments.

xonsh/tools.py
melund: When ever you call the non-lazy methods, it will recompute all the hashes to see if it is still valid. So you only want to do that one time. 
`cmds = self.all_commands`

Afterwards you can just use the lazy methods.

xonsh/tools.py
```diff
@@ -1634,13 +1654,45 @@ def lazyget(self, key, default=None):
...
+            # directory without `./`
+            cwd = _get_cwd()
+            local_bin = next((
+                full_name for full_name in possibilities
+                if os.path.isfile(full_name)
+            ), None)
+            if local_bin:
+                return os.path.abspath(os.path.relpath(local_bin, cwd))
+
+        cmds = self.all_commands
```
melund: Basically, you only want to access `all_commands` property ones. Afterwards use the lazy methods.

xonsh/tools.py
```diff
@@ -1634,13 +1654,45 @@ def lazyget(self, key, default=None):
         """A lazy value getter."""
         return self._cmds_cache.get(key, default)
 
+    def locate_binary(self, name):
```
melund: If you move the stuff about the current directory to `lazy_locate_binary()`  and then here you just do

```
def locate_binary(self, name):
    cmd = self.all_commands
    return self.lazy_locate_binary(name)
```

BYK: Good idea :)

xonsh/commands_cache.py
```diff
@@ -0,0 +1,142 @@
+# -*- coding: utf-8 -*-
+import builtins
+import os
+from collections import abc as abc
```
scopatz: This should be `import collections.abc as abc` to allow this to be a lazy import

scopatz:
Also, `xonsh.commands_cache` needs a new `docs/api/commands_cache.rst` file.

xonsh/commands_cache.py
BYK: Ah, so far with PyCharm's automatic extraction feature :)

scopatz:
Thanks a million @BYK!

xonsh/commands_cache.py
```diff
@@ -0,0 +1,142 @@
...
+        if ON_WINDOWS:
+            # Windows users expect to be able to execute files in the same
+            # directory without `./`
+            cwd = _get_cwd()
+            local_bin = next((
+                full_name for full_name in possibilities
+                if os.path.isfile(full_name)
+            ), None)
+            if local_bin:
+                return os.path.abspath(os.path.relpath(local_bin, cwd))
```
melund: I know this is already merged. But I think could be simplified a bit. Especially if @scopatz want to leave out the `_get_cwd()` bit. `os.path.abspath()` already works relative to current directory. So we could just do:
```python 
return os.path.abspath(local_bin)
```

scopatz: Yeah, please feel free to clean this up!

PR #1337, opened 2016-06-24
Amalgamate install fix

scopatz:
Critical fix for pypi package.


scopatz:
Also, this is decidedly less 1337 than I would have hoped for :)

PR #1339, opened 2016-06-25
WIP Full Pytest Migration

laerus:
moving to pytest idioms  (some of these should be added to *testing* section on CONTRIBUTING) 

the conftest.py has a yield_fixture called `xonsh_builtins` that is a replacement for the old 
`mock_xonsh_env` context manager. `xonsh_builtins` sets and yields `builtins`
To use `xonsh_builtins` in a test we just specify it as an argument to the test, i.e:
```
def test_one(xonsh_builtins):
     ...
```
the xonsh builtins get set when the test starts and teared when the test ends.

note that there is no need to import that fixture from anywhere, fixtures that exist in conftest.py
pytest will pick up and use in all tests.

`xonsh_builtins` sets `__xonsh_env__` to be an empty dict. if we want to set the `__xonsh_env__` to a different initial value we set it directly from `xonsh_builtins.__xonsh_env__`
```
def test_one(xonsh_builtins):
    xonsh_env.__xonsh_env__ = ENV_1
    ...
```

if we want to test for different `__xonsh_env__` values we can use parametrization i.e:
```
@pytest.mark.parametrize('env', [ENV_1, ENV_2])
def test_one(xonsh_builtins, env):
    xonsh_builtins.__xonsh_env__ = env
    ...
```
this will run 2 times each time with the respective env setup.

if we want to set the same  `__xonsh_env__`  for different tests we can create a new fixture (ty @melund ! )
i.e:
```
@pytest.fixture
def special_env(xonsh_builtins):
     xonsh_builtins.__xonsh_env__['Strombus'] = "gigas"

def test_1(special_env):
     assert True

def test_2(special_env):
     assert True
```

i have added a TODO.txt to track progress.


melund:
Great. I was also in the process of doing this, but you beat me to the PR :) I know you based some of it on what I suggested in the previous post, but it was maybe a little hasty. At least I changed my mind on a number points. 

First of all, it can be confusing the `xonsh_env` refers to all builtins and `xenv` refers to the actual `__xonsh_env__`. I know we did that in the nose test, but I didn't like it. 

I suggest that we rename the yield_fixture to `mocked_builtins` (or `xonsh_builtins`) and have it yield the actual `builtins` module. Thus we could do:

```
def test_one(mocked_builtins):
     mocked_builtins.__xonsh_env__['Strombus'] = "gigas"
```

I also think it very verbose to set values in `__xonsh_env__` using `@pytest.mark.parametrize`. I think we should only use it if we want to parameterize test functions.

Tests functions can just set the environment variables they need them self. Similar to the example above. It is s much simpler and easier to remember.

Of course, if many tests in a module share the same extended or special setup, we make a local fixture in that module. 

```
@pytest.fixture
def special_env(mocked_builtins):
     mocked_builtins.__xonsh_env__['Strombus'] = "gigas"

def test_1(special_env):
     assert True

def test_2(special_env):
     assert True
```

laerus:
I absolutely agree on the naming, its very confusing atm.
the only problem i can see with the `special_env` fixture is the teardown for the `__xonsh_env__`,
we can either make the `special_env` a yield fixture but that would be cumbersome for each special_env we need or just del the `__xonsh_env__` at the teardown of `xonsh_builtins` fixture

laerus:
oh my bad, the `mock_builtins` yield fixture deals with deleting the `__xonsh_env__` already, so no need for change. I'm removing the xenv fixture. if we have a single test that uses different `__xonsh_env__` setups we can also just use a direct parametrize and not an indirect one.

```
@pytest.mark.parametrize('env', [SETUP1, SETUP2])
def test_1(mock_builtins, env):
    mock_builtins.__xonsh_env__ = env
    ...
```

laerus:
btw @melund if you want i can remove this PR so you can push yours and i can PR to your PR (inception :laughing:)

melund:
No no. This already better than what I started :)

laerus:
@scopatz @melund Hey guys, i'm a little confused about `tests/tools.py`, specifically about the `check_exec/eval/parse`. At the begging of the tests/tools.py file we instantiate `Execer` that uses `load_builtins` and then in these three check functions we use mock_xonsh_env to set the builtins, isn't that a little redundant ? or am i missing something.

scopatz:
@laerus Ahh, yeah that is redundant.  We probably should have a kwarg in the execers init that allows us to turn off loading builtins.  There is an unload kwarg that lets us turn off unloading, so maybe we just need a `load`.  That seems like the right path.

laerus:
i am trying to decouple these check functions from setting xonsh builtins. At the moment if i just remove `mock_xonsh_env` tests work fine.

tests/test_tools.py
```diff
@@ -931,6 +1164,7 @@ def test_argvquote():
     for st, esc in cases:
         obs = argvquote(st)
         assert esc == obs
+>>>>>>> master
```
scopatz: Seems like you merged in a merge conflict.

scopatz:
> i am trying to decouple these check functions from setting xonsh builtins. At the moment if i just remove mock_xonsh_env tests work fine.

What do you mean exactly?

laerus:
i'm trying to make it so that the loading of builtins or the xonsh enviroment in general happens before we call these check functions, then i can properly use the new xonsh_builtins pytest fixture on the tests tha call these check functions.

laerus:
atm the fact that before we call any of these check function we set up builtins through Execer initiation lets us remove the `mock_xonsh_env` without a problem.

scopatz:
>  lets us remove the mock_xonsh_env without a problem.

I think this might not be the approach we want.  We want to have a clean environment for each test.  This will minimize the possibility of sideeffects in the test suite. I think we therefore want to have a mock env and then also prevent the Execer from loading the builtins, like I mentioned above.  Does this make sense?

laerus:
Yes thats the plan! fixtures have scope and the default scope is 'function', so the fixture runs every time for each test. We can also make a fixture that uses a new Execer instance each time for each test and not change the implementation of Execer, this may be slower but we get more isolation.

laerus:
We better not change the implementation just for the tests unless we want to use the new 'feature' elsewhere or if we want a major refactor.

scopatz:
> Execer instance each time for each test and not change the implementation of Execer, this may be slower but we get more isolation.

Right, but you get even more isolation if the execer doesn't have the side effect of loading builtins.

> We better not change the implementation just for the tests unless we want to use the new 'feature' elsewhere or if we want a major refactor.

I think that this is a three line change that would be worth it :)

laerus:
We can always mock the `load_builtins` to w/e we want and i always think YAGNI when i have to touch the implementation just for the tests but thats your call! :D though it would be better the change to be a seperate PR and keep this PR only about the tests.

scopatz:
Yeah, I agree, let's get this into a mergable state :)

tests/test_environ.py
```diff
@@ -162,17 +165,17 @@ def test_load_static_config(s, exp, loaded, tmpdir, xonsh_builtins):
...
 @skip_if_on_unix
-def test_locate_binary_on_windows():
+def test_locate_binary_on_windows(xonsh_builtins):
     files = ('file1.exe', 'FILE2.BAT', 'file3.txt')
     with TemporaryDirectory() as tmpdir:
         for fname in files:
             fpath = os.path.join(tmpdir, fname)
             with open(fpath, 'w') as f:
                 f.write(fpath)
         env = Env({'PATH': [tmpdir], 'PATHEXT': ['.COM', '.EXE', '.BAT']})
```
melund: Could `xonsh_builtins.__xonsh_env__` not be an empty instance of the `Env` class by default (i.e. define in the fixture. Then we could just do the following here: 

```
xonsh_builtins.__xonsh_env__.update({
    'PATH': [tmpdir],
    'PATHEXT': ['.COM', '.EXE', '.BAT'],
})
```

laerus: yes `xonsh_builtins` fixture sets `__xonsh_env__` to be an empty dict.
i can replace this part in this test with your snippet.

laerus:
Hey @scopatz i'm stucked in test_parser.py
i've changed `nodes_equal` to use assert instead of *if* clauses and i get an error in:
```
for (xname, xval), (yname, yval) in zip(ast.iter_fields(x),
                                            ast.iter_fields(y)):
        assert xname == yname
        assert xval == yval
```
```
E       assert <_ast.Num object at 0xb4add92c> == <_ast.Num object at 0xb684612c>
```
any idea what's wrong?

seems like in current master in test_parser.py 'assert_nodes_equal` doesn't complain if DEBUG_LEVEL is 0

scopatz:
I have no idea, they may actually be different.  Can you call `pdump()` on both x and y please?

laerus:
```
Num(
 n=42
) Num(
 n=42
)
```

laerus:
the test is:
```
def test_int_literal():
    check_ast('42')
```

scopatz:
AST instances don't compare as equal, which is why you are getting this issue:

```
scopatz@artemis ~ $ import ast
scopatz@artemis ~ $ ast.parse('42').body[0].value
<_ast.Num at 0x7f56c6d00940>
scopatz@artemis ~ $ ast.parse('42').body[0].value == ast.parse('42').body[0].value
False
scopatz@artemis ~ $ ast.parse('42').body[0].value != ast.parse('42').body[0].value
True
```

I think in `check_ast()` you need to have the following:

```python
        assert xname == yname
        if not isinstance(xval, ast.AST):
            assert xval == yval
```

scopatz:
Also, I'd really recommend having a single PR for each test file in the future. You might even want to do that instead now :)

laerus:
in current master in test_parser.py `nodes_equal` checks for that equality and always return False but the test doesn't complain because `DEBUG_LEVEL = 0`, so this is a bug in the `assert_nodes_equal` which i'm removing.

you want to me to put different PR for each file under tests?

scopatz:
> you want to me to put different PR for each file under tests?

Yes, ideally.  It would make each PR much much easier to review and understand and it would be faster.  Big PRs like this take a long time to get in and are prone to merge conflicts.

Many small PRs are always better than one or tow big PRs.  Occasionally, big PRs are needed because the change is so fundamental or it is a totally new feature.  But these should be the exception, not the rule :).

I am suggesting that the testing refactor effort might see more consistent and faster progress if it was implemented in smaller chunks.  Just my two drachme, of course :).

scopatz:
Also, I really do appreciate your effort here @laerus! I want to make sure it succeeds.

laerus:
i am not sure how to got about splitting this atm, i wasn't expecting this to be so big but the tests have many redundant stuff. Most of the problems i run into were due to actual bugs that nose was hiding or bugs on the tests themselves. The main reason i got into this is so i can then make small PRs for raising coverage which would be easy after pytest migration. We can leave this PR open for review or splitting it into smaller ones at some point.

scopatz:
Yeah, I agree that at this point it is not worth splitting up.  I think that the main goal right now should just be to get the tests to pass.  My point about smaller PRs was mostly for the future.

laerus:
@scopatz the test_parser fix didnt work out, i've commented out the two lines in `nodes_equal` if you want to take a look, other than that and the persistent bug in EnvPath this PR seems stable atm.

news/pytest-migration.rst
```diff
@@ -0,0 +1,11 @@
+**Added:** xonsh_builtins, xonsh_execer fixtures in conftest.py
```
gforsyth: Small formatting issue here for the changelog amalgamater -- things should be bullet pointed with a single `*` so this should read:

```console
**Added:**

* xonsh_builtins, xonsh_execer fixtures in conftest.py
```

laerus:
strange, tests passing on my machine in 3.4, any way to restart the specific test?

scopatz:
Yeah, codecov sometimes times out, which is annoying.  We might want to remove the quiet flag. I have restarted the failing run.  You can also push up a trival change to force a restart too.

scopatz:
Looks like it passes!

news/pytest-migration.rst
```diff
@@ -0,0 +1,16 @@
+**Added:** 
+* xonsh_builtins, xonsh_execer fixtures in conftest.py
```
scopatz: newline needed between these two lines, and similarly below

laerus: just noticed my rst is wrong again, i'm a total rst noob :laughing:

tests/test_environ.py
```diff
@@ -108,22 +109,25 @@ def p():
         return '{user}'
     assert isinstance(partial_format_prompt(p), str)
 
-def test_HISTCONTROL():
+def test_HISTCONTROL_none():
```
scopatz: PEP8 functions names should be lowercase with underscores.  I know that you didn't write it this way originally @laerus, but if you don't mind cleaning it up since you touched these lines, that would be awesome! Thanks a ton.

tests/test_environ.py
```diff
@@ -156,41 +160,36 @@ def test_swap():
...
+                f.write(fpath)
+        xonsh_builtins.__xonsh_env__.update({
+            'PATH': [tmpdir],
+            'PATHEXT': ['.COM', '.EXE', '.BAT'],
+        })
+        assert ( locate_binary('file1') == os.path.join(tmpdir,'file1.exe'))
+        assert ( locate_binary('file1.exe') == os.path.join(tmpdir,'file1.exe'))
+        assert ( locate_binary('file2') == os.path.join(tmpdir,'FILE2.BAT'))
+        assert ( locate_binary('file2.bat') == os.path.join(tmpdir,'FILE2.BAT'))
+        assert ( locate_binary('file3') is None)
```
scopatz: PEP8 no space after `(`.  I think these outer parentheses can all be removed.

scopatz:
This looks good to me!

tests/test_environ.py
laerus: uh yes! it itched a little when i saw it but i got passed it.

scopatz: Well, do you want to clean it up, or should I go ahead and merge it?

tests/test_environ.py
laerus: my neovim flake8 died for some reason, and i thought i had an innate pep8 ability for a bit there. :laughing: 
(neomake gave up on me :()

tests/test_environ.py
laerus: cleaned it, commits arent pushed yet for some reason

laerus:
@scopatz are the new entries in the testing section on CONTRIBUTING good enough?

scopatz:
> @scopatz are the new entries in the testing section on CONTRIBUTING good enough?

Yep, looks good.  Also, looks like these tests fail now and weren't being picked up before :(.

laerus:
yes on it, a substitution gone bad.

scopatz:
Looks good now!

laerus:
python3.4 test stucked again

scopatz:
Look good now. Just let me know when it is ready to merge!

laerus:
it's ready on my part, there is always room for things to fix but this madness must stop at this point :laughing:

scopatz:
Thanks a million @laerus!

laerus:
thanks for bearing with me! will work on raising coverage next, small PRs this time.

scopatz:
Thank you for putting in this amazing effort on testing! I can't tell you how much I appreciate it!

melund:
Yes. Great job @laerus. Testing will be so much nicer now.

Issue #1338, opened 2016-06-25
completion fails


konsilanto:
Using the latest version (0.4.1) completion does not work al all (at least for me):

- If CDPATH is set with `[".", "/home/foo/workspace"]` and the HOME-directory as well as the directory  "~/workspace" contain a directory "bin", then performing the following command (from the HOME  directory) `cd work<TAB><TAB>` completes to `cd bin/`. Therefore, it can't be used.

- Removing CDPATH (`del $CDPATH`) the false completion does not occur but
  `cd work<TAB><TAB>` activates the first entry in the popup and therefore I am not 
  able to choose one of the subdirectories based on the starting letter.  Typing `cd workspace/<TAB>` does not activate the the first entry and I am able to choose one of the subdirectories by specifying one or several starting letters.

  Typing `cd workspace//<TAB>` popups the list of directories and does not activate the first
  entry but it is at least cumbersome.

scopatz:
Thanks for reporting @konsilanto! Can you please post the output of the `xonfig` command?

gforsyth:
Hey @konsilanto -- this won't help the `CDPATH` issue but if you want to check out #1340 I changed the completer behavior to display completions but not autoselect the first option

konsilanto:
Here the output of `xonfig`:
    +------------------+----------------+
     | xonsh             | 0.4.1           |
     | Python           | 3.5.1          |
     | PLY              | 3.7            |
     | have readline    | True           |
     | prompt toolkit   | 1.0.3          |
     | shell type       | prompt_toolkit |
     | pygments         | 2.1.3          |
     | on posix         | True           |
     | on linux         | True           |
     | distro           | unknown        |
     | on darwin        | False          |
     | on windows       | False          |
     | on cygwin        | False          |
     | is superuser     | False          |
     | default encoding | utf-8          |
     +------------------+----------------+

konsilanto:
@gforsyth -- `dont_select_first` works like a charm. thanks

scopatz:
Great!  Feel free to close this issue, then @konsilanto!

konsilanto:
I am sorry but it was a simple misunterstanding... the one particular problem (concerning selecting the first entry of the popup) is solved however the other just remains unresolved (CDPATH related)

adqm:
Sorry for the confusion, @konsilanto.  Yes, it does look like the completer isn't properly handling `$CDPATH`.

I'll take this one, unless someone else wants to jump in and beat me to it.

[...]

PR #1340, opened 2016-06-25
 stop autocomplete from autoselecting first entry

gforsyth:
stop autocomplete from autoselecting first entry
    
This was mentioned in #1338.  If I have the following directory
structure
    
```console
foo/
 -bar/
 -bag/
 -bing/
 -bong/
```
    
and have already completed to `cd foo/` and hit TAB, ptk will autoselect
the first entry of the completions.  This is incredibly annoying if
there are many possible selections and you would like to filter through
them by prefix.
    
This forks the `generate_completions` event from PTK and stops it from
autoselecting the first entry (but still displays possibilities)

### DIFF
--- a/news/completion.rst
+++ b/news/completion.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** 
+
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks @gforsyth!

PR #1344, opened 2016-06-26
Dirty version numbers

gforsyth:
Ok, here's my first crack at this.  When doing any of `setup.py [install | develop | sdist]`, check if we're in a git repo.  If there is a tag on the most recent commit assume it's a "clean" version and write an empty file `dev.githash`

If there's not a tag, append the number of commits since the most recent version (as given by `git describe --tags`) to `__version__` in `__init__.py` as per PEP440, write git short SHA to dev.githash file (which is ignored in `.gitignore`).  

Then do the actually install, then `checkout --` changes to `__init__.py` if any were made during the course of installation.  

`xonfig` reads the version number as usual.  If it finds a non-empty `dev.githash` file, it adds that on to the end of the `xonfig info` readout.

Possible pitfall:  If you make changes to `__init__.py`, don't commit them and run `setup.py` it will discard your changes.  This doesn't seem like a likely thing to me, but I'm all ears.

```console
+------------------+----------------+
| xonsh            | 0.4.1.dev13    |
| Python           | 3.5.1          |
| PLY              | 3.7            |
| have readline    | True           |
| prompt toolkit   | 1.0.1          |
| shell type       | prompt_toolkit |
| pygments         | 2.1.1          |
| on posix         | True           |
| on linux         | True           |
| distro           | arch           |
| on darwin        | False          |
| on windows       | False          |
| on cygwin        | False          |
| is superuser     | False          |
| default encoding | utf-8          |
| git SHA          | g65b3b06       |
+------------------+----------------+
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -112,12 +112,13 @@ def dirty_version():
     try:
         _version = subprocess.check_output(['git', 'describe', '--tags'])
         _version = _version.decode('ascii')
-        try:
-            base, N, sha = _version.strip().split('-')
-        except ValueError: #on base release
-            open('xonsh/dev.githash', 'w').close()
-            return False
     except subprocess.CalledProcessError:
+        return False
+
+    try:
+        base, N, sha = _version.strip().split('-')
+    except ValueError: #on base release
+        open('xonsh/dev.githash', 'w').close()
         return False
 
     replace_version(base, N)
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@ from __future__ import print_function, unicode_literals
 import os
 import sys
 import json
+import subprocess
 
 try:
     from tempfile import TemporaryDirectory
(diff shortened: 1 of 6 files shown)


scopatz:

setup.py
```diff
@@ -101,11 +102,60 @@ def install_jupyter_hook(prefix=None, root=None):
...
+def dirty_version():
+    """
+    If install/sdist is run from a git directory (not a conda install), add
+    a devN suffix to reported version number and write a gitignored file
+    that holds the git hash of the current state of the repo to be queried
+    by ``xonfig``
+    """
+    try:
+        _version = subprocess.check_output(['git', 'describe', '--tags'])
+        _version = _version.decode('ascii')
```
scopatz: All the lines bellow this can be moved outside of the top-level try-except block.

xonsh/platform.py
```diff
@@ -126,6 +126,26 @@ def is_readline_available():
 
 
 #
+# Dev release info
+#
+
+@functools.lru_cache(1)
+def githash():
+    from xonsh import main
+    install_base = main.__file__.rsplit('/', 1)[0]
```
scopatz: These two lines should just be `install_base = os.path.dirname(__file__)`.

xonsh/platform.py
```diff
@@ -126,6 +126,26 @@ def is_readline_available():
...
+    try:
+        with open('{}/dev.githash'.format(install_base), 'r') as f:
+            hash = f.readlines()
```
scopatz: Wouldn't `hash = f.read().strip()` be more robust here?

xonsh/platform.py
```diff
@@ -126,6 +126,26 @@ def is_readline_available():
...
+        if not hash:
+            hash = None
+        else:
+            hash = hash.pop()
+    except FileNotFoundError:
+        hash = None
+    return hash
```
scopatz: Probably should avoid overwritting the builtin name `hash()` in this function.

xonsh/xonfig.py
```diff
@@ -353,6 +353,8 @@ def _info(ns):
         ('is superuser', is_superuser()),
         ('default encoding', DEFAULT_ENCODING),
         ])
+    if githash():
+        data.append(('git SHA', githash()))
```
scopatz: It is probably OK to always have this as part of the `xonfig`. If it is None, we'll see none

scopatz:
This looks great to me, just some relatively minor cleanup work.

xonsh/platform.py
gforsyth: I knew there must be something to do that but didn't know what.  I learned a thing!

xonsh/platform.py
gforsyth: Yes

xonsh/platform.py
gforsyth: yep, changed it to sha everywhere else, but forgot this bit

xonsh/xonfig.py
gforsyth: ryokai

gforsyth:
Hey @scopatz -- pushed up the changes.  Ready for another look now

scopatz:
Looks great to me! Thanks!

Carreau:
Awesome !

PR #1346, opened 2016-06-26
Anaconda config

melund:
This updates the conda build configuration for the xonsh/channel/dev packages on Anaconda.org. The packages now get a version that matches what @gforsyth just made. 
```
...
The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    xonsh-0.4.2.dev1           |    py35_g84270d3         950 KB  xonsh
....
```

It also has a potential fix for missing Windows developer builds. Apparently, conda-convert have started to refuse to convert to windows because entry-point conversion is not supported. This is not a problem because we supply a xonsh.bat file in scripts as a fallback.

### DIFF
--- a/.binstar.yml
+++ b/.binstar.yml
@@ -12,7 +12,7 @@ script:
   - conda build recipe
 
 after_success:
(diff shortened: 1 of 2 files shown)


gforsyth:
Looks good, thanks @melund!

Issue #1330, opened 2016-06-23
piping(?) freezes xonsh


laerus:
both less and more don't show any output before freeze
`XONSH_SHOW_TRACEBACK=True` doesn't show anything

```                                                                                                    
$ python3 -m pytest --help | less
$ jobs
[1]+ stopped: python3 -m pytest --help | less (19744)
$ fg
[1]+ running: python3 -m pytest --help | less (19744)

^C
```

```
$ cat random_HUGE_file.txt | more
$ fg
[1]+ running: cat random_HUGE_file.txt | more (20426)
^C^C
```

anyone can reproduce?

scopatz:
Thanks for reporting @laerus!  I can reproduce, but I don't have a good sense of why this is happening.

PaulReiber:
Same occurs if you pipe something to "view -"

I also tried this:
ls | /usr/games/rain

...and couldn't even get control-c's or control-\'s to kill rain - had to
abort my ssh session to kill it.
Controlling terminal access maybe?  Session?  Not sure - but there's
definitely a bug here.

Kind regards,


*Paul Reiber*http://about.me/reiber

PaulReiber:
Hello xonsh-erers,

Happy to share results of further testing regarding pipes.

TL;DR is, xonsh's pipes don't seem to be answering as they should for a
pipeline. From the source subprocess's point of view, if stdout is a pipe,
sys.stdout.isatty() needs to return False.

pbr@mongo:~$ cat tty_test/idea

# comparing pipes test

in bash:
python -c "import sys; print(sys.stdout.isatty())" # should be True
python -c "import sys; print(sys.stdout.isatty())" | grep . # should be False

in xonsh:
python -c "import sys; print(sys.stdout.isatty())" # should be True
python -c "import sys; print(sys.stdout.isatty())" | grep . # should be False

...ok, so let's see what happens:
pbr@mongo:~$ python -c "import sys; print(sys.stdout.isatty())"
True
pbr@mongo:~$ python -c "import sys; print(sys.stdout.isatty())" |cat
False
pbr@mongo:~$ . xonsh/venv/bin/activate
(venv)pbr@mongo:~$ which xonsh
/home/pbr/xonsh/venv/bin/xonsh
(venv)pbr@mongo:~$ xonsh
(venv) pbr@mongo ~ $ import sys
(venv) pbr@mongo ~ $ print(sys.stdout.isatty())
True
[...]

PR #1347, opened 2016-06-26
Another pipeline fix

adqm:
I believe this should fix the original issue described in #1330, by making sure all the processes in a pipeline are part of the same process group.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -233,11 +233,6 @@ def get_script_subproc_command(fname, args):
     return interp + [fname] + args
 
 
-def _subproc_pre():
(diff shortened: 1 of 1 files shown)


adqm:
I have done some testing on my own and this seems to do the trick.  This might be a big enough bug that this warrants a micro-release, assuming it looks okay to everyone else.

scopatz:
looks good to me!

scopatz:
Thanks @adqm!

scopatz:
Also, feel free to do a micro release whenever you feel like it. I have already done one today, so I will tap out myself :)

PR #1349, opened 2016-06-27
fix #1342 readline case sensitive issue

### LINKED ISSUE #1342: Completion issue on readline shell with CASE_SENSITIVE_COMPLETIONS
When using readline and set `$CASE_SENSITIVE_COMPLETIONS = False`.

```
$ mkdir usr && mkdir Users
$ cd ./u<TAB>
```

The `u` in second cmd will be eaten. No such issue with PTK.

mitnk:
to fix #1342 

@adqm could you have a look, thanks!

### DIFF
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -43,7 +43,9 @@ RL_DONE = None
 RL_VARIABLE_VALUE = None
 _RL_STATE_DONE = 0x1000000
 _RL_STATE_ISEARCH = 0x0000080
-_RL_PREV_CASE_SENSITIVE_COMPLETIONS = None
+
+_RL_PREV_CASE_SENSITIVE_COMPLETIONS = 'to-be-set'
+
 
 def setup_readline():
     """Sets up the readline module and completion suppression, if available."""
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -121,6 +121,12 @@ def setup_readline():
             # this seems to fail with libedit
             print_exception('xonsh: could not load readline default init file.')
 
+    # handle case sensitive, see Github issue #1342 for details
+    if env.get('CASE_SENSITIVE_COMPLETIONS'):
+        readline.parse_and_bind("set completion-ignore-case off")
+    else:
+        readline.parse_and_bind("set completion-ignore-case on")
+
 
 def teardown_readline():
     """Tears down up the readline module, if available."""
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -26,7 +26,7 @@ from xonsh.base_shell import BaseShell
 from xonsh.ansi_colors import partial_color_format, color_style_names, color_style
 from xonsh.environ import partial_format_prompt, multiline_prompt
 from xonsh.tools import print_exception
-from xonsh.platform import HAS_PYGMENTS, ON_WINDOWS, ON_CYGWIN, ON_DARWIN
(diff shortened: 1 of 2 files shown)


xonsh/readline_shell.py
```diff
@@ -121,6 +121,12 @@ def setup_readline():
             # this seems to fail with libedit
             print_exception('xonsh: could not load readline default init file.')
 
+    # handle case sensitive, see Github issue #1342 for details
+    if env.get('CASE_SENSITIVE_COMPLETIONS'):
+        readline.parse_and_bind("set completion-ignore-case off")
+    else:
+        readline.parse_and_bind("set completion-ignore-case on")
```
scopatz: This will work fine for startup, but will not work when dynamically setting the value at the command line.  Something more involved, like having a custom validator, or having this run before each prompt is needed to truly fix this bug.

scopatz:
Thanks for putting this in @mitnk!

xonsh/readline_shell.py
mitnk: You're right! Should I move these lines to `ReadlineShell.completedefault()` which does fix dynamic setting. If so, is it too heavy to call `readline.parse_and_bind` once for every cmd?

scopatz: > If so, is it too heavy to call readline.parse_and_bind once for every cmd?

Yeah, you should only make the call to parse and bind if a change is detected.

mitnk:
@scopatz Thanks for your suggestions! I updated a bit and it's ready for another look.

xonsh/readline_shell.py
```diff
@@ -43,6 +43,7 @@
 RL_VARIABLE_VALUE = None
 _RL_STATE_DONE = 0x1000000
 _RL_STATE_ISEARCH = 0x0000080
+_RL_PREV_CASE_SENSITIVE_COMPLETIONS = None
```
mitnk: hmm, maybe I better to set this guy to something other than `None`, like a str `'to-be-set'`. Because `env.get()` could get `None` and `readline.parse_and_bind` would not be called (it should better to be called at least once). @scopatz thoughts here?

xonsh/readline_shell.py
```diff
@@ -232,9 +235,24 @@ def parseline(self, line):
         """Overridden to no-op."""
         return '', line, line
 
+    def _rebind_case_sensitive_completions(self):
+        # handle case sensitive, see Github issue #1342 for details
+        global _RL_PREV_CASE_SENSITIVE_COMPLETIONS
+        env = builtins.__xonsh_env__
+        case_sensitive = env.get('CASE_SENSITIVE_COMPLETIONS')
+        if case_sensitive == _RL_PREV_CASE_SENSITIVE_COMPLETIONS:
```
scopatz: I think you could safely have `is` here instead of `==`

xonsh/readline_shell.py
```diff
@@ -232,9 +235,24 @@ def parseline(self, line):
         """Overridden to no-op."""
         return '', line, line
 
+    def _rebind_case_sensitive_completions(self):
```
scopatz: This doesn't use self, so I think think that this can safely be a method at global scope.

scopatz:
Thanks @mitnk - just a couple minor changes!

mitnk:
@scopatz updated, thanks.

scopatz:
Great!  Could you please add a news entry @mitnk?

mitnk:
@scopatz added.

scopatz:
Thanks @mitnk!

Issue #1350, opened 2016-06-27
setup.py install fails if git is not installed


chronitis:
The `dirty_version()` function which tries to check if a random git revision is being built fails if `git` isn't in `$PATH`.

At `setup.py:112` (in v0.4.2) it checks for `CalledProcessError`, but git being unavailable instead throws (uncaught) `FileNotFoundError`.

melund:
Thanks for reporting this @chronitis. I have pushed a fix for it.

PR #1351, opened 2016-06-27
Catch FileNotFoundError when git is not on PATH

melund:
This should fix https://github.com/xonsh/xonsh/issues/1350


gforsyth:
Oops. Thanks for fixing this @melund

PR #1358, opened 2016-06-28
path fix for v3.4

scopatz:
Should fix #1356

PR #1359, opened 2016-06-28
Background thread module loading

scopatz:
This generalizes the notion of loading a module in a background thread and removes the custom `pkg_resources` background loader.  Additionally, the module proxy is actually useful if anyone has a reference to it instead of the actual module.

This also fixes a bug with setup.py checking out `__init__.py` and undoing the amalgamation hooks, when all that is needed is to replace the version line.

Also this accidentally includes #1358.  Sorry!

Maybe @gforsyth  or @Carreau or @laerus can take a look?


scopatz:
Ping on this one for anyone who is interested.

xonsh/lazyasd.py
```diff
@@ -166,3 +174,106 @@ def __bool__(self):
...
+
+    def run(self):
+        # wait for other modules to stop being imported
+        i = 0
+        last = -6
+        hist = [-5, -4, -3, -2, -1]
+        while not all(last == x for x in hist):
+            time.sleep(0.001)
+            last = hist[i%5] = len(sys.modules)
+            i += 1
```
melund: What does this do. It is not immediately clear.  Does it check that the number modules haven't changed for 0.005 seconds? Then assume that all modules have loaded?

melund:
Nice trick. The lazyasd module is really neat. I think the code looks good. 

Does it help startup time for you? I mean if we somehow use a part of pygments that relies on this it should just block until it is fully loaded. Maybe we can tweak our default style to avoid that if it is an issue.

 I have no computer at the moment, so just did code review on my phone. So I will pass on merging to someone who can run with it.

laerus:
uh sry @scopatz i missed this notification, this seems to work and i get a faster startup time. no sure how to test it thoroughly though.

melund:
Great. I will merge then.

xonsh/lazyasd.py
scopatz: Yeah, that is what it does.

scopatz:
Thanks both!

scopatz:
> xonsh startup in this branch is very close to bash on my machine

awesome!

scopatz:
> Does it help startup time for you? I mean if we somehow use a part of pygments that relies on this it should just block until it is fully loaded. Maybe we can tweak our default style to avoid that if it is an issue.

@melund yeah, it really helps startup time for me. It gets prompt toolkit much closer the the readline shell startup on my machine.  

The thing about pygments is that the only part that relies on `pkg_resources` is are the lexer and style search / finder /listing functions.  So xonsh doesn't rely on that at all at startup time.  Maybe if we get around to supporting custom styles this will become an issue and this code will block for people using custom styles.  But there isn't much we can do about that.  Pygments really should be using a simpler and faster model for finding styles.  Something like our xontribs and implicit namespces would be much faster than the `pkg_resources` pain.

PR #1360, opened 2016-06-28
update xonsh repo url

mitnk:
we may need to do this?


scopatz:
A million thanks @mitnk!

PR #1369, opened 2016-06-29
Proposal: a better way to add optional items to the prompt

BlaXpirit:
To describe the situation, let's say my $PROMPT is:
```
{last_return} | {hostname}@{user} | {curr_branch} | {cwd} {prompt_end}
```
`last_return` is just something I came up with - the return status of the last executed program, let's say I don't want it displayed if it's zero. Similarly, `curr_branch` is the current git branch, which is not applicable most of the time. So I don't want my prompt to look like this most of the time:
```
 | hostname@user | | ~ $
```
I know that currently this is alleviated by adding spaces to the formatter dict items themselves, so I am supposed to write my prompt like this:
```
{last_return}{hostname}@{user}{curr_branch} {cwd} {prompt_end}
```
... and rely on `last_return` and `curr_branch` appending a space. But as I demonstrated, that is not flexible enough to do what I want (without reimplementing the formatter dict items to append a pipe instead). Especially if I want some optional part to come at the beginning/end of the format string.

What I propose is allowing users to write their prompt strings like this:

`{last_return:{} | }{hostname}@{user}{curr_branch: | {}} | {cwd} {prompt_end}`

... so that any desired separator can be picked. So, the format string will be applied to the value, like so: `'{} | '.format(formatter_dict['last_return']())`, except if `formatter_dict['last_return']()` returns `None` in the first place, then nothing will be inserted.

See for yourself how it affects the user: https://github.com/xonsh/xonsh/pull/1369/commits/f911613d
(This commit basically keeps the default behavior the same but simplfies and generalizes the implementation of formatter dict items)

Removing the reliance on hardcoded space-based separation from the standard formatter dict values will facilitate flexible implementation of prompt strings, even ones similar to [powerline](https://github.com/powerline/powerline#vim-statusline), without having to redefine half of the formatter dict itself.

As a bonus, sometimes it might be desirable to apply format specs to the values, so that will be possible now: `{last_return:{:3} | }`

Backwards compatibility:
The format specs were previously ignored anyway as far as I can see so this will not cause errors. However, it will cause some whitespace to disappear around the optional parts of $PROMPT strings that are unaware of this change.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1197,32 +1197,6 @@ def _failover_template_format(template):
     return template
 
 
-def format_prompt(template=DEFAULT_PROMPT, formatter_dict=None):
-    try:
-        return _format_prompt_main(template, formatter_dict)
-    except Exception:
-        return _failover_template_format(template)
-
-
-def _format_prompt_main(template, formatter_dict):
-    """Formats a xonsh prompt template string."""
-    template = template() if callable(template) else template
-    fmtter = _get_fmtter(formatter_dict)
-    included_names = set(i[1] for i in _FORMATTER.parse(template))
-    fmt = {}
-    for name in included_names:
-        if name is None:
-            continue
-        if name.startswith('$'):
-            v = builtins.__xonsh_env__[name[1:]]
-        else:
-            v = fmtter[name]
-        val = v() if callable(v) else v
-        val = '' if val is None else val
-        fmt[name] = val
-    return template.format(**fmt)
-
-
 def partial_format_prompt(template=DEFAULT_PROMPT, formatter_dict=None):
     """Formats a xonsh prompt template string."""
     try:
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1219,15 +1219,13 @@ def _partial_format_prompt_main(template=DEFAULT_PROMPT, formatter_dict=None):
         if field is None:
             continue
         elif field.startswith('$'):
-            v = builtins.__xonsh_env__[field[1:]]
-            v = _FORMATTER.convert_field(v, conv)
-            v = _FORMATTER.format_field(v, spec)
-            toks.append(v)
-            continue
+            val = builtins.__xonsh_env__[field[1:]]
+            val = _format_value(val, spec, conv)
+            toks.append(val)
         elif field in fmtter:
             v = fmtter[field]
             val = v() if callable(v) else v
-            val = '' if val is None else val
+            val = _format_value(val, spec, conv)
             toks.append(val)
         else:
             toks.append(bopen)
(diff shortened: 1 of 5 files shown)


scopatz:
Thanks for putting this in @BlaXpirit!  I have to say, I was surprised by the fact that formats specs didn't work at all.  But you are totally right, they don't.

Going through this now.

xonsh/environ.py
```diff
@@ -937,13 +937,13 @@ def _first_branch_timeout_message():
           file=sys.stderr)
 
 
-def current_branch(pad=True):
+def current_branch():
```
scopatz: I don't think that the API should be here unless @melund approves.  I believe that he, and others, may use this.  Instead, maybe change `pad=NotImplemented` and then print a deprecation warning if `pas is not NotImplemented.`

xonsh/environ.py
```diff
@@ -1231,6 +1203,8 @@ def partial_format_prompt(template=DEFAULT_PROMPT, formatter_dict=None):
     except:
         return _failover_template_format(template)
 
+format_prompt = partial_format_prompt
```
scopatz: I think that you can kill this line.

xonsh/environ.py
```diff
@@ -1268,6 +1240,23 @@ def _partial_format_prompt_main(template=DEFAULT_PROMPT, formatter_dict=None):
     return ''.join(toks)
 
 
+def _format_value(val, spec, conv):
```
scopatz: Some tests for this function would be great!

xonsh/environ.py
BlaXpirit: I guess I just wasn't entirely sure that dropping it would be harmless. It's still used - https://github.com/xonsh/xonsh/blob/c345694b/xonsh/base_shell.py#L239  and BTW seems like format specs do work for that case.

xonsh/environ.py
BlaXpirit: @scopatz Will be done if the proposal is accepted

scopatz:
Hi @BlaXpirit, I think that the ideas here are sound.  However, I also think that this needs the full complement of QA stuff.  The new function could use some tests (and maybe that function could even be public, your call), [there should be a news entry](http://xon.sh/devguide.html#changelog), and this definitienly needs to have some docs, probably in the customizing the prompt section of the tutorial: http://xon.sh/tutorial.html#customizing-the-prompt

Though, at this point I think the prompt customization stuff could/should even be its own page in the docs. If you wanted to take this opportunity to split this out of the tutorial, that would be great! If not, no worries.  Thanks again!

xonsh/environ.py
scopatz: Yeah, I think if we are going to kill it, we should really kill it though :)

xonsh/environ.py
scopatz: I think it is good, and I can't think of a better way.

xonsh/environ.py
BlaXpirit: @scopatz  I don't understand 
>I don't think that the API should be here

If we're talking about deprecation... the default behavior of the function changed anyway, is it still worth doing?

scopatz: Well, I know that people were taking partials of this function with pad set to false. I just don't what their rcs to break overnight. But if the people who where doing this say it is not a big deal, then we can go ahead an break it. I think having a warning for a couple of releases is a reasonable way to approach this. We don't claim stability (yet) but we also don't want to knowingly make peoples xonsh's fail to start.

xonsh/environ.py
BlaXpirit: @scopatz  Maybe `partial_format_prompt` should be renamed to, and replace, `format_prompt`?

scopatz: Yeah, I think it is more correct as a partial format.

scopatz:
@BlaXpirit - please ping when this is ready for another look!

BlaXpirit:
@scopatz Ready!

Note that I slightly changed the basic idea: previously any falsey value (`None`, `""`, etc) would be considered "nothing", but then there would be no way to have an actual empty string be formatted, if someone desires that... anyway, now the functions must return None.

PR #1371, opened 2016-06-30
add entry for vox-tabcomplete xontrib

Granitas:
Added another tab completion that bugged me.
Maybe this should be in the core since vox is part of xonsh, right?

source: https://github.com/Granitas/xonsh-vox-tabcomplete


scopatz:
Thanks @Granitas!  And yes, I think this would be a good thing to just have in xonsh proper. If you would like to put in a PR that would be great!

PR #1373, opened 2016-06-30
add xonda to xontribs

gforsyth:
I wrote `xonda` as a thin wrapper around `conda`.  Instead of the (imo) messy activate and deactivate scripts, this just uses `xonsh` `PATH` magic to change environments.  And it has tab completion!  I have no idea if it works on Windows.  @melund -- do you want to try it out?

### DIFF
--- a/xonsh/xontribs.json
+++ b/xonsh/xontribs.json
@@ -60,6 +60,11 @@
   "url": "https://github.com/Granitas/xonsh-autoxsh",
   "description": ["Adds automatic execution of xonsh script files called",
 	          "`.autoxsh` when enterting a directory with `cd` function"]
+ },
+ {"name": "xonda",
+  "package": "xonda",
+  "url": "https://github.com/gforsyth/xonda",
+  "description": ["A thin wrapper around conda with tab completion"]
  }
  ],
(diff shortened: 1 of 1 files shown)


scopatz:
xool!  Should I wait on merging this in?  This is just the xontrib docs, so I think it could go in now...

gforsyth:
I think it can go in now.  The inevitable issues can be opened up on the `xonda` repo

scopatz:
Thanks @gforsyth!

melund:
@gforsyth.  Cool. I will try once I can bring my laptop home again. I have to leave at work because the hinges are broken and locked in an open position :(

I think it sounds cool. I don't  have a big problem with the current activation scripts on Windows though. But the tab completion is nice. Does it also work if I use standard conda?

gforsyth:
@melund 

> the hinges are broken and locked in an open position :(
best wishes for a speedy recovery

> Does it also work if I use standard conda?
it does now!

scopatz:
ironically there is no conda package, double ironically pip is a conda package :)

scopatz:
best wishes for your computer @melund!

PR #1375, opened 2016-07-01
Don't use bare `except:`

### LINKED ISSUE #1372: Raw $PROMPT
Sometimes it shows me a prompt template: `{cwd}{branch_color}{curr_branch} {prompt_end} `
Its very hard to reproduce and I cannot figure out a reason of this problem.

My `$PROMPT`:  

```
$PROMPT = '{BOLD_BLUE}{cwd}{branch_color}{curr_branch}{NO_COLOR} {BOLD_BLUE}{prompt_end}{NO_COLOR} '
```

And xonsh version:

```
~ $ xonsh --version
('xonsh/0.4.2.dev54',)
```

BlaXpirit:
Bare `except:` also [catches](https://docs.python.org/3/library/exceptions.html#exception-hierarchy) SystemExit, KeyboardInterrupt, GeneratorExit, causing unexpected problems.

Fixes #1372

### DIFF
--- a/xonsh/ansi_colors.py
+++ b/xonsh/ansi_colors.py
@@ -34,7 +34,7 @@ def partial_color_format(template, style='default', cmap=None, hide=False):
     """
     try:
         return _partial_color_format_main(template, style=style, cmap=cmap, hide=hide)
-    except:
+    except Exception:
         return template
 
 
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -551,7 +551,7 @@ def run_subproc(cmds, captured=False):
                 output = stdoutfile.read()
             try:
                 _nstdout.close()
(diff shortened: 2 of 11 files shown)


laerus:
Yeas bare except is evil! We can also be more specific about the Exception in case we get an unexpected error, easier for bug fixing too!

scopatz:
Thanks @BlaXpirit! Great job hunting this one down. In most of these cases, @laerus, we can't be more specific. Anything could happen in a prompt function, for example, and just because a prompt function fails doesn't mean that the whole shell should die.

scopatz:
Also @BlaXpirit - sorry I didn't notice this didn't have a news entry.  Would you mind PRing in one separately, please? Thanks!

PR #1378, opened 2016-07-02
Add vox completion to core

Granitas:
Ported https://github.com/Granitas/xonsh-vox-tabcomplete xontrib to core. The vox module had to be adjusted a bit to expose help text.

### DIFF
--- a/xonsh/completers/vox.py
+++ b/xonsh/completers/vox.py
@@ -1,9 +1,13 @@
 import os
 import builtins
+
+from xonsh.lazyasd import LazyObject
+
 from xonsh.platform import scandir
 from xonsh.vox import Vox
 
-DEFAULT_ENV_HOME = builtins.__xonsh_expand_path__('~/.virtualenvs')
+DEFAULT_ENV_HOME = LazyObject(lambda: builtins.__xonsh_expand_path__('~/.virtualenvs'),
+                              globals(), 'DEFAULT_ENV_HOME')
 
 
 def complete_vox(prefix, line, begidx, endidx, ctx):
--- a/xonsh/completers/vox.py
+++ b/xonsh/completers/vox.py
@@ -15,10 +15,11 @@ def complete_vox(prefix, line, begidx, endidx, ctx):
     """
     Completes Xonsh's Vox virtual environment manager
     """
+    line_len = len(line.split())
     cases = [
         not line.startswith('vox'),
-        len(line.split()) > 3,
-        len(line.split()) > 2 and line.endswith(' '),
+        line_len > 3,
+        line_len > 2 and line.endswith(' '),
     ]
     if any(cases):
         return
@@ -29,7 +30,7 @@ def complete_vox(prefix, line, begidx, endidx, ctx):
         env_dirs = list(x.name for x in scandir(venv_home) if x.is_dir())
         return env_dirs, len(prefix)
 
-    if len(line.split()) > 1 and line.endswith(' '):  # "vox new "
+    if (line_len > 1 and line.endswith(' ')) or line_len > 2:  # "vox new "
         return
 
     if prefix not in ALL_COMMANDS:
--- a/xonsh/completers/vox.py
+++ b/xonsh/completers/vox.py
@@ -7,31 +7,33 @@ from xonsh.vox import Vox
 import os
 
 DEFAULT_ENV_HOME = os.path.expanduser('~/.virtualenvs')
+ALL_COMMANDS = LazyObject(lambda: [c[0].split()[1] for c in Vox.help_commands],
+                          globals(), 'ALL_COMMANDS')
 
 
 def complete_vox(prefix, line, begidx, endidx, ctx):
     """
     Completes Xonsh's Vox virtual environment manager
     """
-    if not line.startswith('vox'):
+    cases = [
+        not line.startswith('vox'),
+        len(line.split()) > 3,
+        len(line.split()) > 2 and line.endswith(' '),
+    ]
+    if any(cases):
         return
+
     to_list_when = ['vox activate ', 'vox remove ']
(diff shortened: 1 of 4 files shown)


xonsh/completers/vox.py
```diff
@@ -0,0 +1,32 @@
+import os
+from xonsh.__amalgam__ import builtins
```
gforsyth: I think you should avoid importing from `__amalgam__` (explicitly). I think this is why you have a bunch of test failures

xonsh/completers/vox.py
```diff
@@ -0,0 +1,32 @@
...
+from xonsh.platform import scandir
+from xonsh.vox import Vox
+
+DEFAULT_ENV_HOME = os.path.expanduser('~/.virtualenvs')
```
gforsyth: this should use `builtins.__xonsh_expand_path__`

gforsyth:
Thanks for putting this in, @Granitas !  As I mentioned inline, tests are failing but I think it should be a simple fix.  

And I agree that `vox` completion should live wherever `vox` lives.  
@scopatz @melund : I'm not opposed to merging this in when fixed, but should we consider first whether we want to strip vox out of core and make it a xontrib?  @moigagoo , your input here is also welcomed.  
(also anybody else who has thoughts/comments on this)

melund:
@gforsyth I think vox should be a xontrib, but one that ships with xonsh.

There are probably other things which was added before xontribs were possible. The sudo stuff for windows is one of them.

scopatz:
I agree but I think vox xontrib could be a set pr though

moigagoo:
I am totally pro moving vox to xontrib. If xontrib had been around at the time, I would have made vox a xontrib from the start.

melund:
I agree that it is a separate PR.

xonsh/completers/vox.py
Granitas: Seems to be returning `AttributeError: module 'builtins' has no attribute '__xonsh_expand_path__'`
for the tests below: https://travis-ci.org/xonsh/xonsh/jobs/142168766#L211

Any idea what's causing this?

scopatz: That is probably because xonsh ins't started up by import time. You might be able to get away with making this a lazy object http://xon.sh/api/lazyasd.html#xonsh.lazyasd.LazyObject  Otherwise it is probably OK to leave as is or move to the usage site or make into a function.

xonsh/completers/vox.py
```diff
@@ -0,0 +1,36 @@
...
+from xonsh.platform import scandir
+from xonsh.vox import Vox
+
+DEFAULT_ENV_HOME = LazyObject(lambda: builtins.__xonsh_expand_path__('~/.virtualenvs'),
+                              globals(), 'DEFAULT_ENV_HOME')
+
+
+def complete_vox(prefix, line, begidx, endidx, ctx):
+    """
+    Completes xonsh's `vox` virtual environment manager
```
scopatz: Please remove backticks, this is not valid rst

scopatz:
Thanks @Granitas! I think all that this needs is a news entry and a couple of minor changes and it is good to go!

xonsh/vox.py
```diff
@@ -149,7 +161,9 @@ def remove_env(name):
             virtual environment name
         """
         if 'VIRTUAL_ENV' in __xonsh_env__:
-            print('This environment is currently active. If you really want to remove it, deactivate it first with "vox deactivate %s".\n' % name)
+            print('This environment is currently active. '
```
laerus: this can print to stderr

xonsh/completers/vox.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    if not line.startswith('vox'):
+        return
+    to_list_when = ['vox activate ', 'vox remove ']
+    if any(c in line for c in to_list_when):
+        venv_home = builtins.__xonsh_env__.get('VIRTUALENV_HOME', DEFAULT_ENV_HOME)
+        env_dirs = list(x.name for x in scandir(venv_home) if x.is_dir())
+        return set(env_dirs)
+
+    if (len(line.split()) > 1 and line.endswith(' ')) or len(line.split()) > 2:
```
laerus: it wouldn't rly make a big difference but it would be more sufficient to calculate `len(line.split())` only once.

xonsh/vox.py
scopatz: I agree this should go to stderr

xonsh/completers/vox.py
```diff
@@ -0,0 +1,37 @@
+import builtins
+
+from xonsh.lazyasd import LazyObject
+
+from xonsh.platform import scandir
+from xonsh.vox import Vox
+import os
```
scopatz: PEP8 Python standard library imports should be grouped together at the top.

scopatz:
Looks like there is a merge conflict on this branch @Granitas.

gforsyth:
Hey @Granitas so now that `vox` is a `xontrib`, I'm not sure if this belongs in core, or if it should be a `xontrib` that is bundled with core.  As it stands, if someone hasn't loaded `vox` and then tries to use it's completion functions things will go poorly.

gforsyth:
Actually, thinking for another 30 seconds on this, I think it should probably be a _part_ of the vox xontrib.  no need to separate them out.

scopatz:
+1, I think this should be part of the vox xontrib.  Also, this now has merge conflicts.

Granitas:
Gotcha, I'll try to reformat this and submit a PR fo vox xontrib. Actually I found it's significantly easier to write completion for a xontrib because you can do cool stuff like 
```
help_text = !(program --help)
commands = re.findall('\t(\w+)", help_text)
```
:)

santagada:
@Granitas can you close this branch then?

PR #1379, opened 2016-07-03
Implement basic comma-separated braces expansion

Cheaterman:
Everything in the title, yet for some reason `echo "{1,2}"` doesn't yield the expected result, while manually invoking `from xonsh.tools import expand_case_matching; expand_case_matching('{1,2}')` works.

Please help ^__^

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -123,7 +123,9 @@ def expand_braces(s):
                 [prefix + middle + suffix for middle in middles] +
                 suffixes
             )
-    return ' '.join(_split_braces(s))
+    if regex.fullmatch(s):
+        return ' '.join(_split_braces(s))
+    return s
 
 
 def reglob(path, parts=None, i=None):
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -98,6 +98,34 @@ def expand_path(s):
     return os.path.expanduser(s)
 
 
+def expand_braces(s):
+    # XXX: Doesn't handle escaping literal braces
+    """Takes a string and expands brace-enclosed comma-separated lists into
+    space-separated arguments.
+    """
+    regex = re.compile('(.*?){(.*)}(.*)')
+    def _split_braces(s, nesting=0):
+        matches = regex.fullmatch(s)
+        prefix, middle, suffix = matches.groups()
+        if regex.fullmatch(middle):
+            middles = _split_braces(middle, nesting + 1)
+        else:
+            middles = middle.split(',')
+        if not nesting:
+            return [prefix + middle + suffix for middle in middles]
+        else:
+            prefixes = prefix.split(',')
+            suffixes = suffix.split(',')
+            prefix, prefixes = prefixes[-1], prefixes[:-1]
+            suffix, suffixes = suffixes[0], suffixes[1:]
+            return (
+                prefixes +
+                [prefix + middle + suffix for middle in middles] +
+                suffixes
+            )
+    return ' '.join(_split_braces(s))
+
+
 def reglob(path, parts=None, i=None):
     """Regular expression-based globbing."""
     if parts is None:
@@ -692,6 +720,7 @@ def load_builtins(execer=None, config=None, login=False, ctx=None):
     builtins.__xonsh_regexsearch__ = regexsearch
     builtins.__xonsh_glob__ = globpath
     builtins.__xonsh_expand_path__ = expand_path
+    builtins.__xonsh_expand_braces__ = expand_braces
     builtins.__xonsh_exit__ = False
     builtins.__xonsh_stdout_uncaptured__ = None
     builtins.__xonsh_stderr_uncaptured__ = None
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -98,33 +98,41 @@ def expand_path(s):
     return os.path.expanduser(s)
 
 
+braces_regex = re.compile('(.*?){(.*)}(.*)')
+def _split_braces(s, nesting=0):
+    matches = braces_regex.fullmatch(s)
+    prefix, middle, suffix = matches.groups()
+    if braces_regex.fullmatch(middle):
+        middles = _split_braces(middle, nesting + 1)
+    else:
+        middles = middle.split(',')
+    if not nesting:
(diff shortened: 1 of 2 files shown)


xonsh/tools.py
```diff
@@ -1582,9 +1582,13 @@ def expanduser_abs_path(inp):
...
-    openers = {'[', '{'}
-    closers = {']', '}'}
-    nesting = 0
+    openers = {'[': 'brackets', '{': 'braces'}
+    closers = {']': 'brackets', '}': 'braces'}
+    nesting = {'brackets': 0, 'braces': 0}
+    separators = {'brackets': None, 'braces': ','}
+    brace_prefix = ''
+    brace_args = []
+    cur_brace_arg = ''
```
scopatz: I don't think that it is a good idea to modify this function.  Instead, we probably want a separate function that is independent of the case matching stuff.

scopatz:
@Cheaterman - what matters is that the globpath and iglobpath functions work, not the `expand_case_matching()`, which is not called unless case sensitive completions are turned off.

scopatz:
Also, thanks a ton for putting this in!

Cheaterman:
Gimme some time buddies, I'll be working on it next weekend unless anyone else feels like picking up where I left - very unlikely considering the current state of affairs. So, just a bit of patience, if it's as easy as @scopatz made it sound to be, it should be done this weekend. Thanks buddies ^__^

scopatz:
Hey @Cheaterman! No worries!  Thanks so much for working on this!  Just let us know when  it is ready for another look.

xonsh/tools.py
```diff
@@ -1582,9 +1582,13 @@ def expanduser_abs_path(inp):
 def expand_case_matching(s):
```
laerus: i know this comes a little late but i always feel like regular expressions is the best tool for parsing jobs like this, with regex the implementation is much cleaner and more efficient too! If you want we can try out implementing this with regex.

Cheaterman: If you believe so, I can implement using re.match of course! I thought per-character operation was faster but obviously that may not apply to Python as well as it would to C. Going to re-implement using regexps if you all think it's better ^__^

scopatz: I thi k either way is fine :)

laerus: the decision is your @Cheaterman ! Manual per character parsing in python is hella slow, cause you have to create a lot of intermediate objects, regex parsing is done on the C level.

Cheaterman: Aight @laerus this is more or less what I expected ^__^ I'll try and reimplement this with regexps! Thanks for the hint!

laerus: i know regex can be a little overwhelming, but they really pay off imho. it's easier to debug than the spaghetti per character parsing and much cleaner. also the very cool http://pythex.org/ to test your patterns!

Cheaterman: No worries buddi, it's not my first time writing regexps ^__^ check my IRCBot command system for some cool regexp stuff!
https://github.com/Cheaterman/IRCBot/blob/master/commands/parser.py
Thanks for the cool tool also ^__^ appreciated!

Cheaterman:
There you go buddies, new version of the brace expansion using regexps ^__^ it's still rudimentary but it should get the job done for basic things.

xonsh/built_ins.py
```diff
@@ -98,6 +98,34 @@ def expand_path(s):
     return os.path.expanduser(s)
 
 
+def expand_braces(s):
+    # XXX: Doesn't handle escaping literal braces
+    """Takes a string and expands brace-enclosed comma-separated lists into
+    space-separated arguments.
+    """
+    regex = re.compile('(.*?){(.*)}(.*)')
+    def _split_braces(s, nesting=0):
```
scopatz: I think you can define this function globally, also you can define the regex globally and lazily

xonsh/built_ins.py
```diff
@@ -692,6 +720,7 @@ def load_builtins(execer=None, config=None, login=False, ctx=None):
     builtins.__xonsh_regexsearch__ = regexsearch
     builtins.__xonsh_glob__ = globpath
     builtins.__xonsh_expand_path__ = expand_path
+    builtins.__xonsh_expand_braces__ = expand_braces
```
scopatz: so my recommendation is to rename the current expandpath to something else, then write a newexpand path that calls both the current expandpath and expand braces

xonsh/built_ins.py
Cheaterman: Most certainly can, dunno why I should? :-)

Cheaterman:
FWIW, echo "{1,2}" now yields the expected result, but the resulting string is considered a single argument. So mv file{,.bak} won't work :-/

xonsh/built_ins.py
```diff
@@ -98,6 +98,44 @@ def expand_path(s):
     return os.path.expanduser(s)
 
 
+braces_regex = re.compile('(.*?){(.*)}(.*)')
```
laerus: this can be lazy, a function decorated with `@lazyobject` from `xonsh.lazyasd` that returns the regex.

Also i think  that the second `.*` must be the non-greedy one, atm this regex matches weird cases with brackets like:
```
$ braces_regex.fullmatch(r'test test{1,2,3}}test test').groups()
('test test', '1,2,3}', 'test test')
```

laerus:
Hey @Cheaterman, ty for switching this to regex! it's a lot cleaner. Btw if you can add some tests it would be awesome, making it easier to debug your regex patterns too!

Cheaterman:
Owiii Laerus ^__^ thanks for letting me know about xonsh's implementation of a lazy-loaded object, I wasn't sure what you meant before. Testing is planned! Other than that, the regexp behavior you demonstrated above is the expected one, care to show what would differ if the second .* was non-greedy? Thanks in advance!

laerus:
with the greedy * you catch the biggest possible match thats inside two braces, this may inlcude other braces or anything, some tests may make this more clear, also we have more than one brace expansion in the line.

laerus:
this can now go in the bashisms xontrib and use the `events.on_transform_command`, @Cheaterman want to take a shot at it?

gforsyth:

PR #1381, opened 2016-07-04
add completer for pip

Granitas:
Seems like this should be in the core.

This completes commands from `--help` and completes `pip list` values for `pip show` and `pip uninstall`.

### DIFF
--- a/xonsh/completers/pip.py
+++ b/xonsh/completers/pip.py
@@ -22,10 +22,11 @@ def complete_pip(prefix, line, begidx, endidx, ctx):
     """
     Completes python's package manager pip
     """
+    line_len = len(line.split())
     case = [
         not PIP_RE.search(line),
-        len(line.split()) > 3,
-        len(line.split()) > 2 and line.endswith(' '),
+        line_len > 3,
+        line_len > 2 and line.endswith(' '),
     ]
     if any(case):
         return
@@ -34,7 +35,7 @@ def complete_pip(prefix, line, begidx, endidx, ctx):
         items = items.decode('utf-8').splitlines()
         return set(i.split()[0] for i in items)
 
-    if (len(line.split()) > 1 and line.endswith(' ')) or len(line.split()) > 2:
+    if (line_len > 1 and line.endswith(' ')) or line_len > 2:
         # "pip show " -> no complete (note space)
         return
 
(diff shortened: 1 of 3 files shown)


xonsh/completers/pip.py
```diff
@@ -0,0 +1,38 @@
+import re
+import subprocess
+
+RE_PIP = re.compile("pip(?:\d|\.)*")
+RE_PIP_LIST = re.compile("pip(?:\d|\.)* (?:uninstall|show)")
```
scopatz: Please make these lazy

xonsh/completers/pip.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def get_pip_commands():
+    help_text = str(subprocess.check_output(['pip', '--help'], stderr=subprocess.DEVNULL))
+    commands = re.findall("  (\w+)  ", help_text)
+    return [c for c in commands if c not in ['completion', 'help']]
+
+
+ALL_COMMANDS = get_pip_commands()
```
scopatz: same here

scopatz:
Thanks a ton for putting this in!

scopatz:
Yeah that is what I meant :)

Granitas:
For some reason the build on python 3.5 timeout, but everything seems to be fine. Could someone maybe restart it?

scopatz:
Hi @Granitas - I think that this needs a news entry http://xon.sh/devguide.html#changelog and the minor pip ordering change.

xonsh/completers/pip.py
```diff
@@ -22,7 +22,12 @@ def complete_pip(prefix, line, begidx, endidx, ctx):
     """
     Completes python's package manager pip
     """
-    if not PIP_RE.search(line):
+    case = [
+        not PIP_RE.search(line),
+        len(line.split()) > 3,
+        len(line.split()) > 2 and line.endswith(' '),
+    ]
```
laerus: if you do something like this `line_len = len(line.split())` at the start of the function you will create only one list and calculate its length only once. a little more sufficient!

Granitas: Good catch!

scopatz:
Thanks @Granitas!

Issue #1382, opened 2016-07-04
Typo in lazyasd?


janschulz:
I just stole the file for a different project and found that my IDE flagged this:

https://github.com/xonsh/xonsh/blob/master/xonsh/lazyasd.py#L131

```py
    def __delitem__(self, key):
        if key in self._d:
            del self._d[lkey] # <- lkey instead of key?
        else:
            del self._loaders[key]
            self._destruct()
```

IMO that's a typo?

scopatz:
@janschulz that is definitely a typo! Thanks for reporting. Would you mind putting in a PR for this?

janschulz:
See https://github.com/xonsh/xonsh/pull/1383

PR #1383, opened 2016-07-04
Fix typo in LazyDict.__delitem__

janschulz:
The only visible result of this error would probably be when not all values from this LazyDict were loaded and xonsh would have been closed at that point. So not very likely :-)

Closes: https://github.com/xonsh/xonsh/issues/1382

[PR done via the github web ui...]


scopatz:
Thanks!

Issue #385, opened 2015-09-09
unexpected behavior with subprocess mode and environment variables


zardus:
I might be misunderstanding how env variable expansion works, but it seems to me that something is not being parsed correctly:

```
yans@lath ~ master $ $ASDF = '/usr'
yans@lath ~ master $ echo $ASDF
/usr
yans@lath ~ master $ ls $ASDF
aarch64-linux-gnu    bin             include         mips-linux-gnu       powerpc64le-linux-gnu  sparc64-linux-gnu
alpha-linux-gnu      cris-linux-gnu  lib             mips64-linux-gnu     s390-linux-gnu         src
arm-linux-gnu        games           lib32           msp430               sbin                   vax-linux-gnu
arm-linux-gnueabi    hppa-linux-gnu  libx32          msp430-linux-gnu     sh-linux-gnu           x86_64-linux-gnu
arm-linux-gnueabihf  i386-linux-gnu  local           powerpc-linux-gnu    share                  xscale-linux-gnu
avr-linux-gnu        ia64-linux-gnu  m68k-linux-gnu  powerpc64-linux-gnu  sparc-linux-gnu
yans@lath ~ master $ ls $ASDF/bin
.................... 
```

The expected result, as far as I can tell, is for xonsh to execute `ls /usr/bin` in subprocess mode, but it seems to want more input. I thought that this was maybe due to `ls` being an alias, but it doesn't look like it:

```
yans@lath ~ master $ del aliases['ls']
yans@lath ~ master $ ls $ASDF/bin
.................... 
```

BTW, I also tried quoting the string:

```
yans@lath ~ master $ ls "$ASDF"/bin
/bin/ls: cannot access "$ASDF"/bin: No such file or directory
```

It looks like it did run this in subprocess mode, but without expanding the variable...

Am I missing something?

scopatz:
Hi @zardus - first off, sorry for the delay in responding.  (I am in the middle of proposal season at work and have been turning email off to get my writing done.)

Secondly, this seems like a bug or deficiency.  It should be fixed.  For the moment, you can get around this via the rather hackey:

```python
$ ls @($ASDF + "/bin")
```

Not satisfying, I know.  If you want to take a stab at fixing this, I'd really appreciate it.  The basic problem is that `os.expandvars` is not being run on every command argument that it could be / should be.  That would be one way to implement this. Another might be to give the $NAME operation the highest precedence in the parser. 

[...]

PR #1387, opened 2016-07-05
alternative expandvars + tests

laerus:
This changes the expandvars implementation to use regular expressions thus becoming faster.
The regex now matches cases like those mentioned in #385 but this doesn't actual fix this issue. Any ideas on how the requested behavior could be implemented would be appreciated, although this PR could just serve as a refactor of `expandvars` with added tests.


xonsh/tools.py
```diff
@@ -42,6 +42,13 @@
                             ON_LINUX, ON_WINDOWS, PYTHON_VERSION_INFO)
 
 
+# regular expressions for matching enviroment variables
+# i.e $FOO, ${'FOO'}
+posix_envvar_regex = re.compile(r'\$({(?P<quote>[\'\"])|)(?P<envvar>\w+)((?P=quote)}|(?:\1\b))')
```
melund: You should make these lazy. See the concept here: https://github.com/laerus/xonsh/blob/2dc7f2b5c322ddb640eacd7d814c48371cc6c445/xonsh/tools.py#L213

melund: I also think we usually give them upper case names

laerus: i was just checking lazyasd thinking about doing this !

melund:
I haven't got time to really look at it now. But I like the negative diff :)

laerus:
tried to make the regex string with comments about each part inside a raw triple quoted string while escaping newlines, didn't work out.

laerus:
there is still a bug from not using `_get_env_string' to detype the value of envvar to string, more tests incoming

scopatz:
Is this ready for another look @laerus?

laerus:
yes! i will add a news entry.

xonsh/tools.py
```diff
@@ -1434,121 +1435,41 @@ def check_for_partial_string(x):
...
+    Unknown variables are left unchanged."""
     env = builtins.__xonsh_env__
     if isinstance(path, bytes):
         path = path.decode(encoding=env.get('XONSH_ENCODING'),
                            errors=env.get('XONSH_ENCODING_ERRORS'))
     elif isinstance(path, pathlib.Path):
         # get the path's string representation
         path = str(path)
-    if '$' not in path and (not ON_WINDOWS or '%' not in path):
-        return path
```
scopatz: The above two lines should come back, I think

scopatz:
Thanks for putting this in @laerus!  I like the general idea,  but I think that this adds a regression.  On Windows, you are able to mix $VAR and %VAR% and this looks like it requires %VAR% on windows.  I think that the regex needs to be a little more involved to handle this.

laerus:
the second `if` that checks for '$' in path runs on both linux and windows independent  of the first `if` that only runs on windows, i prolly should add a test for windows that checks if both envvar syntaxes, in the same path, get expanded.

laerus:
hmm no idea how to start AppVeyor on this. @scopatz

melund:
@learus If you login to appweyor with your github  account you should have access to restart it . @xonsh/core-developers are added as administrators on the appweyor account. If it doesn't work I can add you directly as I did with scopatz

laerus:
hey @melund ! I can't make sense of AppVeyor interface atm. I logged in with my github account but there was no entry under *projects* and when i added my fork of xonsh i can't seem to make it test only this branch, even though i choose to build only this on settings

laerus:
managed to try this branch on a windows machine, now works as expected and all tests pass, something like this `'$foo %spam%'` on windows will be expanded in `'bar eggs'` @scopatz

laerus:
@melund when i push commits in this branch AppVeyor doesn't even care, only travis run tests here, am i doing something wrong?

melund:
No. Maybe I configured it wrong way. I will look at it later to night.

xonsh/tools.py
scopatz: Ahh nevermind your logic is good.

scopatz:
> managed to try this branch on a windows machine, now works as expected and all tests pass, something like this '$foo %spam%' on windows will be expanded in 'bar eggs' @scopatz

Ahh you are right! Mea culpa

scopatz:
Thanks a ton!

PR #1394, opened 2016-07-06
Fixing an error I encountered in the intro wizard

ericdill:
A `NameError` appeared in the xonsh wizard.  This PR fixes that.  Would it be
of interest to put a testing harness around this xonsh wizard code?  I'm not
really even sure how to go about doing that, but would be interested in working
on it if some guidance could be provided.

```
$ xonsh wizard
<snip>
Would you like to enable xontribs now, yes or no [default: no]? yes

distributed
url: http://xon.sh
package: xonsh
license: BSD 3-clause
installed? yes
The distributed parallel computing library hooks for xonsh.
Importantly this provides a subsitute 'dworker' command which enables
distributed workers to have access to xonsh builtins.

Furthermore, this xontrib adds a 'DSubmitter' context manager for
executing a block remotely. Moroever, this also adds a convienece
function 'dsubmit()' for creating DSubmitter and Executor instances
at the same time. Thus users may submit distributed jobs with::

    with dsubmit('127.0.0.1:8786', rtn='x') as dsub:
        x = $(echo I am elsewhere)

    res = dsub.future.result()
    print(res)

This is useful for long running or non-blocking jobs.
Add this xontrib, yes or no [default: no]? no
NameError: name 'Unstorable' is not defined
```

### DIFF
--- a/news/fix-name-error.rst
+++ b/news/fix-name-error.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Unqualified usage of Unstorable in xonsh setup wizard that was causing the
+  wizard to crash and burn
+
+**Security:** None
--- a/xonsh/xonfig.py
+++ b/xonsh/xonfig.py
@@ -240,7 +240,7 @@ def make_xontrib(xontrib, package):
     if msg.endswith('\n'):
         msg = msg[:-1]
     mnode = wiz.Message(message=msg)
-    convert = lambda x: name if to_bool(x) else Unstorable
+    convert = lambda x: name if to_bool(x) else wiz.Unstorable
     pnode = wiz.StoreNonEmpty(XONTRIB_PROMPT, converter=convert,
                               path=_xontrib_path)
     return mnode, pnode


scopatz:
Thanks @ericdill! would you please also add a news entry? see the developer docs / contributions doc for more details.

Also, we don't yet have a great way of testing our CLIs, ironically. We still need to figure this out.

laerus:
@scopatz  do you think we can do the cli testing with tmux like [https://github.com/DRMacIver/hecate](url) is doing it?

melund:
That is interesting :)  New idea for a pytest plugin?

melund:
Well. A pytest plugin is not really necessary. I say we try to use it. It seems really cool

laerus:
Hecate source is kinda of a hairball, there is also [pymux](https://github.com/jonathanslenders/pymux) that uses prompt_toolkit to do its magic. Maybe we can come up with something in the middle, a simpler version of hecate suited for xonsh.

melund:
Looks cool as well. But it not really target testing.

laerus:
yes pymux is big we don't need to use it, maybe just take some ideas if we want to make our own tmuxer for testing xonsh.

laerus:
@melund whay about [pyte](https://github.com/selectel/pyte) and maybe open a new issue about this :)

PaulReiber:
I think it'd be a shame to work on integrating xonsh and tmux with the end
use case being only for testing.

Tmux is super powerful - please let's consider how we might bring its
capabilities to bear for xonsh end-users too.

I'd love to have a succinct means of controlling various things in
managed tmux sessions via xonsh.


Kind regards,


ericdill:
Well this turned into quite the healthy discussion!  Testing the CLIs seems like it is quite the challenging project.  @scopatz Added a news page entry

gforsyth:
Thanks, @ericdill !  For the fix and the discussion starter!

PR #1398, opened 2016-07-07
Some more lazy improvements

scopatz:
Relatively small.

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -8,6 +8,7 @@ import os
 import sys
 import shlex
 
+from xonsh.lazyasd import LazyObject
 from xonsh.dirstack import cd, pushd, popd, dirs, _get_cwd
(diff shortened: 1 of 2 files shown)


xonsh/aliases.py
```diff
@@ -301,19 +301,22 @@ def xexec(args, stdin=None):
...
+def _bang_n_parser():
+    parser = ArgumentParser('!n', usage='!n <n>',
+                description="Re-runs the nth command as specified in the "
+                            "argument.")
+    parser.add_argument('n', type=int, help='the command to rerun, may be '
+                                            'negative')
+    return parser
+
+
+_BANG_N_PARSER = LazyObject(_bang_n_parser, globals(), '_BANG_N_PARSER')
```
melund: It is a very repeatable pattern. Would make sense to have lazy decorator?

melund:
Look good

PR #1399, opened 2016-07-07
update the osx guide

santagada:
- Python 3 default binaries instead of leaving it open for interpretation
- Moved the readline comment up (seems important)
- Put a line about bash-completion


scopatz:
Thanks @santagada!

PR #1400, opened 2016-07-07
remove unneded echo

santagada:
I'm not sure why it was there, but tested on bash and xonsh and it doesn't seem to be needed.


scopatz:
Thanks again!

Issue #1392, opened 2016-07-05
default value for $PATH seems strange in wizard


Qumeric:
It's  `<xonsh.__amalgam__.LazyObject object at 0x7f6705929eb8>`. Current value is okay though.

I run fresh xonsh 0.4.3 on Ubuntu 16.04.

laerus:
Hello, thank you for reporting this! this happens because we need to call the `__getattribute__` method on a LazyObject somehow in order to replace itself with the actual object. Simply putting something like `PATH_DEFAULT[0]' in `environ.py` before `DEFAULT_VALUES` dictionary is being set, fixes this problem, but there must be a more elegant way to this.

scopatz:
I think this is a matter of calling str or repr s directly

PR #1401, opened 2016-07-07
lazy decorators

scopatz:
Thanks for the idea @melund!


melund:
Cool. It looks awesome.

laerus:
is there a way to solve  #1392 without calling `__getattribute__` somehow?

scopatz:
@laerus, I think that issue is unrelated to this one

scopatz:
Ok!  I have finished messing around with this branch.  Now everything in the main xonsh dir should be lazy.  Sorry it is such a huge change.  Most of the line diffs are simple changes to pyghooks.

laerus:
Great work @scopatz ! i've been playing with this branch and everything seems to be working nice.

melund:
Great. Looks good to me as well. Love the decorators.

xonsh/proc.py
```diff
@@ -12,46 +12,49 @@
...
+import functools
+import threading
+import importlib
+import subprocess
+import collections
+import collections.abc as abc
 
 from xonsh.tools import (redirect_stdout, redirect_stderr, ON_WINDOWS, ON_LINUX,
                          fallback, print_exception, XonshCalledProcessError)
+from xonsh.teepty import TeePTY
```
melund: @scopatz . Sorry I didn't notice this in the review. 

This change causes xonsh to fail on windows when `XONSH_DEBUG` is set. TeePTY imports some pacakges which are linux only

scopatz: Do you know which ones?  It seems like the right fix is to make those imports explicitly lazy / non-windows.

melund:
termios is not available. Also termios is also imported by both tty and pty

scopatz:
OK, fix in #1417

PR #1405, opened 2016-07-08
Docs on how to get good colors in win ConHost

melund:
This adds info on how to get better colors in the default windows terminal. 

It also includes some minor adjustments to the color tweaks on windows. For example, it wasn't possible to disable `$INTENSIFY_COLORS_ON_WIN` in `xonshrc` since it relied on __XONSH_SHELL__ being available.

PR #1407, opened 2016-07-09
Fix color mixup

melund:
I messed up the registry colors in the last commit to #1405. Basically forgot to sort them correctly. 

This sorts it out.


gforsyth:
Looks good to me!

PR #1414, opened 2016-07-11
Fix windows title

melund:
$TERM is not set on Windows. This prevented xonsh from ever setting the title on the console window.


scopatz:
Thanks @melund!

PR #1417, opened 2016-07-11
fix for windows debug via lazy imports

scopatz:
Can you try this out please @melund?


melund:
Solves it. Thanks a lot.

scopatz:
Sorry about the failure!  More evidence that we need the CI to run tests in both debug and normal mode.

Issue #1430, opened 2016-07-14
Docs on how to contribute to Xonsh


santagada:
specially talk about how to setup a dev environment with amalgamation deactivated, I tried this steps and failed:
- created a vox env
- installed requirements-test
- python3 setup.py develop
- py.test

it resulted in this error:

```
 $  py.test
============================================================================================================ test session starts =============================================================================================================
platform darwin -- Python 3.5.2, pytest-2.9.2, py-1.4.31, pluggy-0.3.1
rootdir: /Users/santagada/devel/xonsh, inifile:
collected 1131 items

tests/test_aliases.py .........
tests/test_ast.py ...
tests/test_builtins.py s.....................
tests/test_contexts.py .........................
tests/test_dirstack.py ....
tests/test_environ.py ..............................s
tests/test_execer.py ss..................
tests/test_foreign_shells.py .....s
tests/test_history.py ......
tests/test_imphooks.py ....
tests/test_inspectors.py .
tests/test_lazyasd.py .
tests/test_lazyjson.py ...................
tests/test_lexer.py .....................................
tests/test_main.py .................
tests/test_man.py .
tests/test_news.py .................
tests/test_parser.py .....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
tests/test_ptk_history.py .
tests/test_ptk_multiline.py .......
tests/test_replay.py sss
tests/test_tools.py ................................................................................................................................................................................................................................................................................................ssss...s
tests/test_wizard.py ...
tests/test_xontribs.py .

================================================================================================== 1118 passed, 13 skipped in 7.14 seconds ===================================================================================================
Exception ignored in: <bound method Execer.__del__ of <xonsh.__amalgam__.Execer object at 0x10ed742b0>>
Traceback (most recent call last):
  File "/Users/santagada/devel/xonsh/xonsh/__amalgam__.py", line 16113, in __del__
TypeError: 'NoneType' object is not callable
Exception ignored in: <generator object get_tokens at 0x10f29b9e8>
Traceback (most recent call last):
  File "/Users/santagada/devel/xonsh/xonsh/__amalgam__.py", line 9931, in get_tokens
TypeError: catching classes that do not inherit from BaseException is not allowed
```

gforsyth:
Thanks for reporting, @santagada!  We can definitely use some more documentation on this.  
If you want to disable amalgamation for testing/development, you can do (from within `xonsh`)

```console
$XONSH_DEBUG=1
[...]

PR #1431, opened 2016-07-15
Better CI run

santagada:
I did a bunch of changes to make the continuous integration/development smoother

- use pytest-cov and pytest-flake8 (can run both locally easier)
- always run flake8 on test runs (it caches the results so only run for changed files)
- run xonsh in debug mode
- turn on branch coverage (slower but more complete)
- configure files to not do coverage
- ignore all errors present currently on the code base for flake8
- added a conftest.py to make python3 -m pytest and py.test to behave the same

I will send more pull requests or maybe update this one removing the flake8 errors one by one or all in a bunch.

I think it will make writting the contributing docs on #1430 much easier


.travis.yml
```diff
@@ -4,8 +4,8 @@ python:
   - 3.5
   - "nightly"
 install:
-  - pip install -r requirements-tests.txt
+  - pip install --upgrade -r requirements-tests.txt
 script:
-  - coverage run --source xonsh -m py.test -q
+  - py.test --flake8 --cov=. --cov-report=term
```
laerus: i think this should be `--cov=xonsh` then you can stop omitting tests, setup etc on .coveragerc

scopatz: shouldn't we have equivalent changes in appveyor?

scopatz:
Thanks a million for putting this in!

.travis.yml
santagada: I don't know if we shouldn't be harcoding exactly what we don't care about? maybe we want to have test for amalgamation that is outside, maybe tests for other examples and other things as well? if not I can totally change.

santagada:
changed appveyor although I don't know much about it and I don't know where to look for results.

.travis.yml
laerus: or we should hard code what we really want to test, we can always use more `--cov` arguments if we test stuff outside the xonsh package dir at a later time, seems easier than adding lines in `ommit:` each time we add something new to the root xonsh dir. The decision is yours though and i'm happy with either way!

scopatz:
I am happy with this @santagada. I leave it up to @laerus to merge though!

laerus:
Hey @santagada, do we still need the empty `conftest.py` in the root xonsh dir?

santagada:
@laerus yes unfortunately, so new developers can just run py.test and be happy. We could put a not on why it is there.

santagada:
@laerus it would be good to merge soon as new code just broke flake8 and I had to allow even more errors on the run.

laerus:
ty @santagada , merging this!

PR #1434, opened 2016-07-15
branch space fix

scopatz:
When the branch name in an empty string for the default prompt, there would sometime be two spaces.  This fixes that issue.  @Carreau and I discussed this at SciPy


melund:
Lgtm

PR #1433, opened 2016-07-15
`vox remove` accepts multiple arguments

stockninja:
I was surprised when `vox rm foo bar` failed, since it only accepts 1 argument. Here's a potential improvement.

I'm not sure if I properly documented the `names` argument to the `remove_envs` function.

Tried to update all references to `vox remove` in the docs.

PR #1435, opened 2016-07-16
Print to error to sys.stderr

stockninja:
https://github.com/xonsh/xonsh/pull/1433


scopatz:
Thanks for putting this in @stockninja!

Issue #1097, opened 2016-06-04
pygments style shows up on config styles but cannot be used


ypekka:
Installing pygments style from pypi doesn't  work fully.

Installing package "pygments-style-solarized" adds "solarizedlight" and "solarizeddark"  styles, they show up with "xonfig styles", but following exception is raised when used in config.json.

% /opt/xonsh/bin/xonsh
/opt/xonsh/lib/python3.5/site-packages/xonsh/pyghooks.py:306: RuntimeWarning: Could not find style 'solarizeddark', using default
  RuntimeWarning)

scopatz:
Thanks for reporting this @ypekka! There are sort of two things going on here:

1. We would need to add these styles to [`xonsh.pyghooks`](https://github.com/scopatz/xonsh/blob/master/xonsh/pyghooks.py) and probably also `ansi_colors`
2. Maybe we should allow the pygments color style to work and use the default style if the style is allowed by pygments.

Pull requests for both of these are most welcome!

stockninja:
While trying to replicate this, I may have unearthed another issue. 

When I run `$XONSH_COLOR_STYLE = 'foo'`, my whole terminal window crashes. Happens in both Terminal.app and iTerm2.app on OSX.

Tried setting `XONSH_DEBUG = 1` to see if I could catch an exception, but that didn't help.

scopatz:
@stockninja - ahh yeah, that is a separate issue.  We should probably issue a warning if the color style doesn't exist.  I actually thought we did this already.... Try running xonsh from within another xonsh or bash or etc instance to see the full traceback

airdrik:
in my personal opinion, I would prefer to leave the ANSI colors separate from the syntax highlighting.  

I found it a little odd that the colors in my prompt (defined using ANSI colors) changed each time I switched style.  I am also running into a problem, which I'm inclined to believe to be related, where the nano text editor is basically unusable as most things are showing up as dark-gray on dark-gray (doesn't matter which style I use.  Running a new bash shell and running nano from there produces the expected coloration).

So my vote would be to just use pygments for the syntax highlighting bits, and leave the ANSI colors alone.

scopatz:
Potential fix in #2230.

scopatz:
Closing due to merger of #2230, please reopen if needed.

PR #1439, opened 2016-07-16
Unrecognized `$XONSH_COLOR_SCHEME` values don't crash terminal

stockninja:
Previously, when unrecognized value given to `$XONSH_COLOR_STYLE`, the mysterious
`DEFAULT_STYLE` variable causes an unhandled `NameError`, leading to
terminal crash.

This gets rid of the mystery variable, and falls back to the "default" colorscheme -- which seems to be intent of current code.

https://github.com/xonsh/xonsh/issues/1097#issuecomment-233105328


xonsh/pyghooks.py
```diff
@@ -317,12 +317,12 @@ def style_name(self):
...
         if self._style_name == value:
             return
-        if value in STYLES:
-            cmap = STYLES[value]
-        else:
+        if value not in STYLES:
             warn('Could not find style {0!r}, using default'.format(value),
                  RuntimeWarning)
-            cmap = DEFAULT_STYLE
+            value = 'default'
```
stockninja: IMO `value = self._style_name or 'default'` would lead to better UX ... but that's an awful lot of indirection code-wise ...

news/xonsh-color-style-bugfix.rst
```diff
@@ -0,0 +1,13 @@
...
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Unrecognized ``$XONSH_COLOR_SCHEME`` values don't crash terminal.
```
scopatz: probably should be style instead of scheme

scopatz:
Thanks a million @stockninja!

Issue #1441, opened 2016-07-16
Unittests ocassionally hang


stockninja:
Occassionally unittests hang for me on CI. I think I recall someone else experiencing this elsewhere. I installed the `pytest-timeout` pip package, and ran tests with `--timeout=10` flag to try and get a stacktrace -> https://gist.github.com/stockninja/849134871d2e118055ff9feb2046b7f7

`test_ast.test_multilline_num` and `test_tools.test_partial_string` both appear to have `time.sleep` problems. I don't understand what's going on nearly well enough to solve, but this should help others solve the problem ...

stockninja:
closing and putting this info in https://github.com/xonsh/xonsh/issues/1335 ...

PR #1442, opened 2016-07-16
WIP: Add MacOS/OSX hosts to CI

stockninja:
Using https://docs.travis-ci.com/user/multi-os/ to test on MacOS/OSX with Python version 3.4 and 3.5.

As Travis doesn't support Python on MacOS/OSX yet, it get slightly hacky ... but not too bad.

I also added a `pytest-timeouts` testing dependency. I can definitely remove this, but I find it greatly helps find tests that hang ...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,25 @@
 language: python
-python:
-  - 3.4
-  - 3.5
-  - "nightly"
+
+matrix:
+    include:
+        - os: linux
+          python: "nightly"
+          env: TOXENV=py36
+        - os: linux
+          python: 3.5
+          env: TOXENV=py35
+        - os: linux
+          python: 3.4
+          env: TOXENV=py34
+        - os: osx
+          language: generic
+          env: TOXENV=py35
+        - os: osx
+          language: generic
+          env: TOXENV=py34
 install:
-  - pip install -r requirements-tests.txt
+    - ./install.sh
 script:
-  - coverage run --source xonsh -m py.test -q
+  - python3 -m coverage run --source xonsh -m py.test -v --timeout=10 tests/
 after_success:
   - codecov
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
+    case "${TOXENV}" in
+        py35)
+            # Install Python 3.5 from source
+            wget https://www.python.org/ftp/python/3.5.2/Python-3.5.2.tar.xz
+            tar -xf Python-3.5.2.tar.xz
+            cd Python-3.5.2
+            ./configure
+            make
+            make install
+            cd ..
+            rm -rf Python-3.5.2
+        ;;
+        py34)
+            # Install Python 3.4 from source
+            wget https://www.python.org/ftp/python/3.4.3/Python-3.4.3.tar.xz
+            tar -xf Python-3.4.3.tar.xz
+            cd Python-3.4.3
+            ./configure
+            make
+            make install
+            cd ..
+            rm -rf Python-3.4.3
+        ;;
+    esac
+fi
+
+pip3 install -r requirements-tests.txt
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -1,5 +1,6 @@
 ply
-pytest
+pytest==2.9.1
+pytest-timeout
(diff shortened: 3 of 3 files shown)


.travis.yml
```diff
@@ -1,11 +1,25 @@
...
+          env: TOXENV=py35
+        - os: osx
+          language: generic
+          env: TOXENV=py34
 install:
-  - pip install -r requirements-tests.txt
+    - ./install.sh
 script:
-  - coverage run --source xonsh -m py.test -q
+  - python3 -m coverage run --source xonsh -m py.test -v --timeout=10 tests/
```
stockninja: Is 10 seconds a reasonable cutoff for hanging tests?

install.sh
```diff
@@ -0,0 +1,30 @@
...
+        py34)
+            # Install Python 3.4 from source
+            wget https://www.python.org/ftp/python/3.4.3/Python-3.4.3.tar.xz
+            tar -xf Python-3.4.3.tar.xz
+            cd Python-3.4.3
+            ./configure
+            make
+            make install
+            cd ..
+            rm -rf Python-3.4.3
```
stockninja: Not a shell scripter ... which is why I love xonsh!

Any help here would be appreciated.

Installing from source b/c I don't know how to make brew can install arbitrary python3 options ... and I'd rather not also install something like pyenv ...

requirements-tests.txt
```diff
@@ -1,5 +1,6 @@
 ply
-pytest
+pytest==2.9.1
```
stockninja: On linux + python3.4, CI would mysteriously install `pytest` at 2.6.x ... which didn't work with `pytest-timeout` ... I should probably remove this `pytest-timeout` dependendy.

stockninja:
https://github.com/xonsh/xonsh/issues/1441 is killing the CI build ...

melund:
Cool. Thanks a lot for adding this.  I will let others take a look at this. I am holiday now so I only have my phone.

requirements-tests.txt
BlaXpirit: Why not `>=`

install.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
```
BlaXpirit:
```bash
set -o errexit
```
could be nice here

install.sh
BlaXpirit: Seems like a lot of duplicated code, could generalize it with something like this:
```bash
    py34)
        py_ver=3.4.3
```
```bash
wget -q -O - "https://www.python.org/ftp/python/$py_ver/Python-$py_ver.tar.xz" |
tar -xf
```
(also here I'm doing the download+extract in a stream, without saving the archive file)

.travis.yml
scopatz: Sounds good to me.  We can update this if needed.

install.sh
scopatz: It seems like we should just use miniconda here.  We definitely should not be building Python from source.

requirements-tests.txt
scopatz: Yeah, please use `>=`

scopatz:
Thanks for putting this in @stockninja!  This looks like it now has a merge conflict with master,

install.sh
asmeurer: You should definitely use conda. See http://conda.pydata.org/docs/travis.html.

santagada:
@stockninja can I help? I'm very interested in better testing for OSX

scopatz:
@santagada, you can always pull from @stockninja's branch into your own local repo/fork and then open up another PR - either to the mainline or into @stockninja's fork (which would then affect this PR).

stockninja:
By all means, help! I've been moving the last few days so haven't touched
it in a bit.

santagada:
anyone oppose to just install brew and latest python? It makes a ton of sense because most people on osx are either on brew or with latest python3 anyway (as it doesn't come with mac). It simplifies so much the install.

gforsyth:
@santagada -- is brew building things from source? Or grabbing binaries?  If the former, then miniconda is a better bet.  Actually, miniconda is probably just a better bet overall.

santagada:
Can we close this pull request?

scopatz:
Yup, just merged.

stockninja:
Thanks for the help, guys!

PR #1445, opened 2016-07-18
Trim the heading g in Git SHA of xonfig

mitnk:
The g in `Git SHA` makes no sense other than look similar to `[9a-f]`.. Let's trim it!

```
+------------------+--------------+
| xonsh            | 0.4.3.dev209 |
| Git SHA          | gd32819f     |
| Python           | 3.5.2        |
| PLY              | 3.7          |
| have readline    | True         |
| prompt toolkit   | 1.0.3        |
| shell type       | readline     |
| pygments         | 2.1.3        |
| on posix         | True         |
| on linux         | False        |
| on darwin        | True         |
| on windows       | False        |
| on cygwin        | False        |
| is superuser     | False        |
| default encoding | utf-8        |
+------------------+--------------+
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,9 +121,10 @@ def dirty_version():
         open('xonsh/dev.githash', 'w').close()
         print('failed to parse git version', file=sys.stderr)
         return False
+    sha = sha.strip('g')
     replace_version(base, N)
     with open('xonsh/dev.githash', 'w') as f:
-        f.write(sha.strip('g'))
+        f.write(sha)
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks for putting this in!

scopatz:
actually it seems like the correct place for this change is where the hash is written to a file in setup.py

mitnk:
right! updated.

scopatz:
thanks!

PR #1446, opened 2016-07-18
Remove flake8 exceptions

santagada:
Do most of the styles fixes so we don’t need to have exceptions for flake8 compliance.

I know this is a big and boring pull request, but believe me it was much more boring actually fixing these


melund:
Thanks for doing all the boring work. It is appreciated, but I am generally not a big fan of Pep8 as hard rules. And I don't believe adding it to an automatic checker is going to improve code quality. Especially not if we enforce it. I am Holliday now so I don't have time to present all my arguments for the case, but Raymond Hettinger framed it pretty well in his 2015 pycon talk. Please watch it: https://youtu.be/wf-BqAjZb8M

xonsh/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 # amalgamate exclude jupyter_kernel parser_table parser_test_table pyghooks
 # amalgamate exclude winutils wizard
-import os as _os
+import os as _os  # noqa: E402
```
scopatz: This will just be overwritten the next time amalgamate is run.

xonsh/aliases.py
```diff
@@ -298,8 +297,8 @@ def xexec(args, stdin=None):
 @lazyobject
 def _BANG_N_PARSER():
     parser = argparse.ArgumentParser('!n', usage='!n <n>',
-                description="Re-runs the nth command as specified in the "
-                            "argument.")
+                                     description="Re-runs the nth command as specified in the "
+                                                 "argument.")
```
scopatz: This extra indentation is not needed.

xonsh/base_shell.py
```diff
@@ -226,7 +226,7 @@ def reset_buffer(self):
 
     def settitle(self):
         """Sets terminal title."""
-        _ = self
+        _ = self  # noqa: F841
```
scopatz: This line should just be removed

xonsh/base_shell.py
```diff
@@ -266,7 +266,7 @@ def prompt(self):
 
     def _append_history(self, tee_out=None, **info):
         """Append information about the command to the history."""
-        _ = self
+        _ = self  # noqa: F841
```
scopatz: this line should be removed

xonsh/diff_history.py
```diff
@@ -193,16 +193,16 @@ def _cmd_out_and_rtn_diff(self, i, j):
...
         bout = self.b['cmds'][j].get('out', None)
         if aout is None and bout is None:
-            #s += 'Note: neither output stored\n'
+            # s += 'Note: neither output stored\n'
             pass
         elif bout is None:
             aid = self.a['sessionid']
             s += 'Note: only {red}{aid}{no_color} output stored\n'.format(red=RED,
-                                                            aid=aid, no_color=NO_COLOR)
+                                                                          aid=aid, no_color=NO_COLOR)
```
scopatz: Line too long

xonsh/environ.py
```diff
@@ -583,7 +590,7 @@ def DEFAULT_VALUES():
         'for more information and available options.',
         default="'surrogateescape'"),
     'XONSH_HISTORY_FILE': VarDocs('Location of history file (deprecated).',
-        configurable=False, default="'~/.xonsh_history'"),
+                                  configurable=False, default="'~/.xonsh_history'"),
```
scopatz: I disagree with almost all of these changes based on readability

scopatz: In this data structure.

xonsh/environ.py
```diff
@@ -1341,7 +1352,7 @@ def multiline_prompt(curr=''):
         elif newcount <= n:
             toks.append(tok)
         else:
-            toks.append((tok[0], tok[1][:n-count]))
+            toks.append((tok[0], tok[1][:n - count]))
```
scopatz: I don't think you need the extra spaces in a slice

xonsh/foreign_shells.py
```diff
@@ -11,12 +10,12 @@
 import warnings
 import functools
 import collections.abc as abc
+from collections import Sequence
```
scopatz: Please remove this line, this is a non-lazy import.  Sequence should be refered to via `abc.Sequence`

xonsh/foreign_shells.py
```diff
@@ -267,7 +266,7 @@ def foreign_shell_data(shell, interactive=True, login=False, envcmd=None,
                                        '__XONSH_ENV_END__', flags=re.DOTALL),
                     globals(), 'ENV_RE')
 ENV_SPLIT_RE = LazyObject(lambda: re.compile('^([^=]+)=([^=]*|[^\n]*)$',
-                                             flags=re.DOTALL|re.MULTILINE),
+                                             flags=re.DOTALL | re.MULTILINE),
```
scopatz: I don't think the spaces are needed here

xonsh/xoreutils/_which.py
```diff
@@ -150,16 +149,16 @@ def _cull(potential, matches, verbose=0):
...
-                sys.stderr.write("no executable access: %s (%s)\n"\
+                sys.stderr.write("no executable access: %s (%s)\n"
                                  % potential)
         else:
             matches.append(potential)
             return potential
 
 
-#---- module API
+# module API
```
scopatz: Why does this change matter?

scopatz:
Thanks for putting this in @santagada!  I have made some in-line comments.  However, I see two big issues with this as it stands:

1. can you find a way to please remove all of the `# noqa` comments.  I find these very bothersome and unhelpful, in-line.
2. Can you please make sure that the indendation doesn't push lines over 80 chars.  This seems to happen a lot.

Thanks!

xonsh/base_shell.py
santagada: I thought this was put there for pylint... but if you are ok to remove it I will 👍

xonsh/foreign_shells.py
santagada: lower in this file it does refer directly to it.

xonsh/xoreutils/_which.py
santagada: there should only be spaces after a "#"

santagada:
Thanks for the comments I tried to go through guessing what would make sense. I will fix whats need to be fixed.

about 1, how can I fix things like lazyload should not be needed or imports that are not used in the module (but are there so other people importing it can use it?

scopatz:
> about 1, how can I fix things like lazyload should not be needed or imports that are not used in the module (but are there so other people importing it can use it?

Can't you set this in a config file somewhere?

santagada:
only file wide... but not a problem I can do it for now if you want it we can talk about it later

scopatz:
> only file wide... but not a problem I can do it for now if you want it we can talk about it later

That is fine.

santagada:
sidenote:
@melund I'm also not a big fan of it as a hard rule, what I am is totally for automatic guarantees of consistency (not of beauty or unfortunately function). Auto pep8 makes all the code looks the same which was already what was not happening here. When you have automatic formatting (either using a program like yapf or gofmt or at least enforcing a standard) you get two things: All code looks the same, no one needs to do code reviews about pep8. People can actually focus on substance and not form.
I saw the talk and totally agree with everything Hettinger said, but exactly so we can focus on that and leave petty arguments about formatting to machine.
I really think  gofmt is the best contribution of go to the programming language space, with interfaces a distant second, and their implementation CSP a distant third. We can go over this in a later stage and hope you will be happy with the conclusion.

scopatz:
I haven't watched the talk @santagada, but I tend to agree with @melund.  The code should be approachable and hackable first.  I have yet to see a Python tool that really achieves this, but I am happy to keep accepting PRs to this end.

I think the most important thing is for us to not get upset about autoformatting. I am happy to have improvements, not happy to have unhappy devs :)

xonsh/base_shell.py
scopatz: yup!

xonsh/foreign_shells.py
scopatz: Thanks!

scopatz:
@santagada let us know when this is ready for another look!  Also, it would be much easier to do this one or two files per PR.  One giant PR is a big review burden....

scopatz:
... and a big maintenance burden when the code will inevitably change underneath you

santagada:
Fixed all the problems and merged in master.

@scopatz why did you turn off defaulting to run flake8 tests? I see that running flake8 doesn't add much or anything to test run time (py.test caches the results of non modified files) and it makes the errors faster to fix specially for people without an editor that doesn't have a linter.

scopatz:
Thanks!  As per the flake8 removal, it was causing a failure to run any tests locally, because many of us don't have flake8 installed. I think you just need to add --flake8 whenever run py.test. You could even have an alias for this.

santagada:
@scopatz Tried updating the patch with the new vox files but there were a lot of undefined things in there (and the very bizarre format when no variable on string). How can I fix this? vox.py seems to be accessing a lot of modules not defined

santagada:
just added one more ignore.

xonsh/ptk/shortcuts.py
```diff
@@ -32,7 +32,7 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
-        #self.reset()
+        # self.reset()
```
scopatz: this line can just be removed

xonsh/readline_shell.py
```diff
@@ -401,7 +403,7 @@ def prompt(self):
             # This is needed to support some system where line-wrapping doesn't
             # work. This is a bug in upstream Python, or possibly readline.
             RL_LIB.rl_reset_screen_size()
-        #return super().prompt
+        # return super().prompt
```
scopatz: this line can also be removed

xonsh/xoreutils/_which.py
```diff
@@ -232,7 +231,7 @@ def whichgen(command, path=None, verbose=0, exts=None):
                     elif i == 0:
                         fromWhere = "from current directory"
                     else:
-                        fromWhere = "from PATH element %d" % (i-1)
+                        fromWhere = "from PATH element %d" % (i - 1)
```
scopatz: PEP8 variable names should be lower case with underscores

scopatz:
Couple of more small issues

xonsh/xoreutils/_which.py
melund: I don't think we should change these. I left them on purpose when we moved the which.py module. Basically, I wanted to make minimal changes to the library.

melund: I think all the whitespace changes are OK. But the rest we really shouldn't change.

scopatz: OK, there are still the comment removals in some other modules to do, though.

santagada:
@scopatz done :)

scopatz:
Thanks!

PR #1447, opened 2016-07-18
Strip pygments

gforsyth:
This is a crack at decoupling `pygments` from the `prompt_toolkit` shell.  I tried to separate out functions that don't strictly require `pygments` into a separate module `xonsh.style`.  

If a user has `prompt_toolkit` but doesn't have `pygments` installed, everything works except for syntax hightighting and styles.  Styles could be made to work, but I think this is reasonable for a start.


news/pygments-separation.rst
```diff
@@ -0,0 +1,13 @@
+**Added:** None
+
+**Changed:** 
+
+* ``prompt_toolkit`` shell no longer explicitly requires ``pygments`` to run, although styles and syntax highlighting are unavailable without it
```
scopatz: Long line.

scopatz:
Hi @gforsyth - thanks for putting this in.  I am not sure I think all of these changes are a good idea.  First, all of the color token things in `style`  do apply to pygments directly.  Therefore, I think it makes sense for them to stay in pyghooks, since they could apply to any pygments based system.

The remaining things in `style` seem to apply to just ptk, and thus could probably be moved to the ptk subpackage.

Also, the `style` name is a little ambiguous since there is an entire parallel ANSI style system in `ansi_colors` that does not use pygments.  Hopefully this makes sense.

gforsyth:
Hey @scopatz -- valid concerns all.  I was trying to avoid duplicating a lot of code, with one version using `ptk` tokens and the other using `pygments` tokens.  I can take another crack at it.  And the name _is_ far too generic.

scopatz:
> Hey @scopatz -- valid concerns all. I was trying to avoid duplicating a lot of code, with one version using ptk tokens and the other using pygments tokens.

Ahh yeah, so I don't think it will be too bad if the PTK specifc stuff moved to the ptk subpackage.  I think this is safest too, because it is possible that ptk could drop pygements at some point in the future and we would still want a pure pygments version of some of these functions..

gforsyth:
I'm closing this to begin anew -- if the other core devs would be so kind as to not delete this branch straight away, that would be great.

PR #1448, opened 2016-07-18
history refactor + small feature

laerus:
a first try in refactoring history

sessions  are now options to the `show` subcommand i.e:
`history show bash`
`history show xonsh` for all xonsh history
`history` and `history show` implicitly means `history show session` and returns current session history as expected

history now accepts multiple slices arguments separated by spaces i.e:
`history show bash 1:3 10 11 14 15:30:5` returns the respective ranges or indexes etc.

changes in `tools.py`:
`ensure_int_or_slice` has now been deprecated for the new `ensure_slice` that always return a slice
and uses regular expressions, 

`is_int_as_str` and `is_slice_as_str` are now reimplemented in EAFP style.


tests/test_tools.py
```diff
@@ -853,24 +813,37 @@ def test_bool_or_int_to_str(inp, exp):
...
+    'test',
+    '6.53:100:5',
+    '4:-',
+    '2:15-:3',
+    '50:-:666',
+    object(),
+    ])
+def test_ensure_slice_invalid(inp):
+    obs = ensure_slice(inp)
+    assert obs is None
```
scopatz: Failures shouldn't return None, failures should raise an error

xonsh/history.py
```diff
@@ -357,8 +357,6 @@ def _bash_hist_parser(location=None):
                     bash_hist_formatted.append((command, 0.0, ind))
                 return bash_hist_formatted
     else:
-        import ipdb
-        ipdb.set_trace()
```
scopatz: Thanks for catching these.

xonsh/history.py
```diff
@@ -648,8 +617,7 @@ def show(self, *args, **kwargs):
         Valid options:
             `session` - returns xonsh history from current session
             `show`    - alias of `session`
-            `all`     - returns xonsh history from all sessions
-            `xonsh`   - alias of `all`
+            `xonsh`   - returns xonsh history from all sessions
```
scopatz: Please retain `all`

xonsh/history.py
```diff
@@ -680,39 +648,42 @@ def _hist_gc(ns, hist):
             continue
 
 
-_HIST_MAIN_ACTIONS = LazyObject(lambda: {
+@lazyobject
+def HIST_SESSIONS():
+    return ('session', 'xonsh', 'bash', 'zsh')
```
scopatz: It seems that this should be a frozenset rather than a tuple

xonsh/history.py
```diff
@@ -680,39 +648,42 @@ def _hist_gc(ns, hist):
...
     'diff': _dh_main_action,
     'gc': _hist_gc,
-    }, globals(), '_HIST_MAIN_ACTIONS')
+    }
 
 
-def _hist_main(hist, args):
-    """This implements the history CLI."""
-    if not args or args[0] in ['-r'] or ensure_int_or_slice(args[0]):
+def parse_args(args):
```
scopatz: Due to amalgamation, please don't have such generically named funtions.  Retain the use of `_hist` in the name.  Also, all public functions need a docstring

xonsh/tools.py
```diff
@@ -763,20 +763,6 @@ def is_int(x):
...
-    Tests if something is an integer
-    If not a string to begin with is automatically false
-    """
-    if isinstance(x, str):
-        try:
-            int(x)
-        except ValueError:
-            return False
-        return True
-    return False
```
scopatz: I don't see a reason to remove these functions or their tests, even if they are not used.  They could still be useful to users or us in the future for other environment variable validators.

xonsh/tools.py
```diff
@@ -792,25 +778,6 @@ def is_slice(x):
...
-            x = x.strip('[]()')
-            try:
-                slice(*(int(x) if len(x) > 0 else None
-                        for x in x.split(':')))
-            except ValueError:
-                return False
-        else:
-            return False
-        return True
-    return False
```
scopatz: Same here.  Please retain the function and the tests

xonsh/tools.py
```diff
@@ -956,19 +923,27 @@ def bool_or_int_to_str(x):
...
-    elif is_int_as_str(x):
-        return int(x)
-    return False
+        m = SLICE_REG.fullmatch(x)
+        if m:
+            groups = (int(i) if i else None for i in m.groups())
+            s = slice(*groups)
+    except TypeError:
+        pass
+    return s
```
scopatz: Ensurers should raise an error if they fail.

scopatz:
Thanks for writing this @laerus!

scopatz:
I am happy with the new functions and most of the refactor, but I think the old functions should stay too.

tests/test_tools.py
laerus: i wanted to ask you about this, the previous `ensure_int_or_slice` returned False, i prefer raising and error too.

scopatz: let's do that then :)

xonsh/tools.py
laerus: i can change this to just use the regex

scopatz: that would be even better

xonsh/tools.py
laerus: this can fail with TypeError, also the docstring is misleading. *something*

laerus:
in general these 2 functions i removed are not very well implemented imo and they existed only for the old `is_int_or_slice` which wasnt very useful either cause it required another check on the callers level. sry i rushed to a removal its just that i prefer EAFP. will bring them back though.

laerus:
arggg flake8 complains

scopatz:
the news test failure seems like a legitimate failure. Howeve, the flake8 indentation error seems incorrect. I think you can disable this in setup.cfg

laerus:
fixed news and the pep8 complain but imo it seems a little more ugly.

scopatz:
Oh yeah, I was suggesting you chabge setup.cfg to make flake8 not report those errors anymore :)

laerus:
i think this is ready. well history could use a lot more refactoring and tests but this is enough for a start.

scopatz:
Hi @laerus - this looks good except for the loss of the `all` type in the show command.  Please keep this around. Thanks!

laerus:
i have included `all` too, just forgot to update the docstring of `_hist_show`, btw `all` is just an alias for `xonsh` and a little misleading, cause you'd expect to return history from *all* xonsh/bash/zsh but it just returns all xonsh history.

scopatz:
Ahh I see.  Well I do think it needs to be in the docstrings so people know what it does.  Maybe you can change the docstring to say that all is an alias for xonsh.

laerus:
Yes i will include it in the docs except if we want to change it so `all` really returns history from all sessions.

scopatz:
I don't think that we need to change it in the short term

scopatz:
That is, we don't need to change the behaviour, we just need to change the docs.

laerus:
atm there is a small bug but i can't figure out yet where it comes from

when i type an invalid slice like `history 10:-` with `XONSH_SHOW_TRACEBACK` set to False the first time the error doesn't appear, any consecutive runs show the error msg normally though.

```
$ history 10:-
'10:-' is not a valid slice format
```

with `XONSH_SHOW_TRACEBACK` set to True this doesn't happen.

scopatz:
Is `display_error_message()` being called? https://github.com/xonsh/xonsh/blob/master/xonsh/tools.py#L639

laerus:
Doesn't seem so. i shouldn't call it manually right?

scopatz:
No, you shouldn't call it manually.  This seems like an unrelated issue.

scopatz:
Also, it seems to work just fine for me, even on the first time

scopatz:
```console
scopatz@localhost ~/temp $ xonsh --shell-type=readline -DXONSH_COLOR_STYLE=default
scopatz@localhost ~/temp $ $XONSH_SHOW_TRACEBACK = False
scopatz@localhost ~/temp $ history 10:-
'10:-' is not a valid slice format
scopatz@localhost ~/temp $ 
```

scopatz:
Thanks a ton @laerus!

PR #1456, opened 2016-07-20
Vox API

astronouth7303:
Adds a dict-like object for API access to Vox and virtual environments.

This also allows for testing the Vox machinery (the basics of which is included).

This also changes a few details of the algorithms, namely that instead of dynamically recalculating the old value of `PATH`, it stores the old value and restores it on deactivation. `$PYTHONHOME` is also handled, to bring Vox in line with traditional venv activate scripts.

### DIFF
--- a/xontrib/vox.py
+++ b/xontrib/vox.py
@@ -308,3 +308,11 @@ class VoxHandler:
     vox help (-h, --help)
         Show help
 """)
+
+
+def _vox(args, stdin=None):
+    """Runs Vox environment manager."""
+    vox = VoxHandler()
+    return vox(args, stdin=stdin)
+
+aliases['vox'] = _vox
--- a/xontrib/vox.py
+++ b/xontrib/vox.py
@@ -1,11 +1,14 @@
+"""Python virtual environment manager for xonsh."""
+
+import xontrib.voxapi as _voxapi
+
 class _VoxHandler:
     """Vox is a virtual environment manager for xonsh."""
 
     def __init__(self):
         """Ensure that $VIRTUALENV_HOME is defined and declare the available vox commands"""
 
-        import xontrib.voxapi
-        self.vox = xontrib.voxapi.Vox()
+        self.vox = _voxapi.Vox()
 
         self.commands = {
             ('new',): self.create_env,
@@ -104,7 +107,7 @@ class _VoxHandler:
         for name in names:
             try:
                 del self.vox[name]
-            except self.vox.EnvironmentInUse:
+            except _voxapi.EnvironmentInUse:
                 print('The "%s" environment is currently active. In order to remove it, deactivate it first with "vox deactivate %s".\n' % (name, name),
                       file=sys.stderr)
                 return
--- a/xontrib/vox.py
+++ b/xontrib/vox.py
@@ -1,180 +1,11 @@
-"""Python virtual environment manager for xonsh."""
-import os
-import sys
-import venv
-import shutil
-import builtins
-import collections
-import collections.abc
-
-from xonsh.platform import ON_POSIX, ON_WINDOWS, scandir
-
-VirtualEnvironment = collections.namedtuple('VirtualEnvironment', ['env', 'bin'])
-
-class EnvironmentInUse(Exception): pass
-
-class NoEnvironmentActive(Exception): pass
-
-class Vox(collections.abc.Mapping):
-    """API access to Vox and virtual environments, in a dict-like format.
-
-    Makes use of the VirtualEnvironment namedtuple:
-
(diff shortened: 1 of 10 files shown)


scopatz:
Thanks for putting this in @astronouth7303! Before merging this is, I think it makes sense to move vox to an officially supported xontrib (ie, in the xontrib/ dir of the main repo here).  If we are going to be making API changes like this, it makes sense to do them all in one go rather than having multiple API changes one after another. 

Would you be willing to migrate vox to a xontrib, please, as part of this?  I don't think it will be so hard.  Just a move vox.py to the xontrib dir and a couple of minor changes (like in aliases.py)

astronouth7303:
Yeah, I can do that. It kinda makes sense, too, since Vox isn't really a core bit to a shell.

scopatz:
Thanks a million!

scopatz:
Also, please comment when this is ready for another look.

astronouth7303:
I think I got everything covered code-side for the migration, including the publication in `xontribs.json`. Tests pass locally. Documentation was not updated.

scopatz:
Thanks @astronouth7303, I think you will need to update `docs/api/vox.api` at a minimum to point to xontrib rather than xonsh

scopatz:
Also, can you please add a news entry: http://xon.sh/devguide.html#changelog

xontrib/vox.py
```diff
@@ -0,0 +1,318 @@
...
+import os
+import sys
+import venv
+import shutil
+import builtins
+import collections.abc
+
+from xonsh.platform import ON_POSIX, ON_WINDOWS, scandir
+
+VirtualEnvironment = collections.namedtuple('VirtualEnvironment', ['env', 'bin'])
```
scopatz: I think it is a good idea to `import collections` explicitly too

xontrib/vox.py
```diff
@@ -0,0 +1,318 @@
...
+
+class EnvironmentInUse(Exception): pass
+
+class NoEnvironmentActive(Exception): pass
+
+class Vox(collections.abc.Mapping):
+    """API access to Vox and virtual environments, in a dict-like format.
+
+    Makes use of the VirtualEnvironment namedtuple:
+    1. ``env``: The full path to the environment
```
scopatz: newline needed in rst here

xontrib/vox.py
```diff
@@ -0,0 +1,318 @@
...
+        List all available environments
+
+    vox remove (rm, delete, del) <env> <env2> ...
+        Remove virtual environments
+
+    vox help (-h, --help)
+        Show help
+""")
+
+
```
scopatz: It might be a good idea to split this into a `voxapi.py` or `voxhooks.py` file for everything above this line and leave `vox.py` for everything below this line.  Currently all of the names and definitions will be loaded into the xonsh context. for classes like Vox this isn't a huge deal, but we probably don't want to define os, scandir, etc for everyone.

Note that the mpl.py and mplhooks.py in the xontribs dir takes this same dual file strategy to separate out the API from the xontrib interface.

astronouth7303:
Done, done-done, and done.

xontrib/vox.py
```diff
@@ -1,23 +1,11 @@
...
     """Vox is a virtual environment manager for xonsh."""
 
     def __init__(self):
         """Ensure that $VIRTUALENV_HOME is defined and declare the available vox commands"""
 
-        if not builtins.__xonsh_env__.get('VIRTUALENV_HOME'):
-            home_path = os.path.expanduser('~')
-            venvdir = os.path.join(home_path, '.virtualenvs')
-            builtins.__xonsh_env__['VIRTUALENV_HOME'] = venvdir
+        import xontrib.voxapi
```
scopatz: I think this is better as `import xontrib.voxapi as _voxapi` globally.

The reason the mpl hooks did imported lazily is because matplotlib and numpy take a long time to import.  That shouldn't be a problem here.

xontrib/voxapi.py
```diff
@@ -0,0 +1,168 @@
...
+
+    Makes use of the VirtualEnvironment namedtuple:
+
+    1. ``env``: The full path to the environment
+    2. ``bin``: The full path to the bin/Scripts directory of the environment
+    """
+
+    class EnvironmentInUse(Exception): pass
+
+    class NoEnvironmentActive(Exception): pass
```
scopatz: These should be defined globally, imho

astronouth7303:
Fair enough. I only did it as inner classes to minimize importing and be pristine about the global namespace.

scopatz:
Thanks a million!

PR #1457, opened 2016-07-20
Add sys import, and be more consistent about errors going to stderr.

astronouth7303:
Add a `sys` import a broke.

Also, all errors now go to stderr.


scopatz:
Thanks!

PR #1458, opened 2016-07-21
Add options and methods to bring voxapi to feature parity with venv.

astronouth7303:
Adds keyword arguments to `Vox.create()` and the method `Vox.upgrade()` so the Vox machinery is at feature parity with the venv module.

Note: The oddities with `Vox.upgrade()` is due to how venv works.

### DIFF
--- a/xontrib/voxapi.py
+++ b/xontrib/voxapi.py
@@ -32,16 +32,68 @@ class Vox(collections.abc.Mapping):
         else:
             self.venvdir = builtins.__xonsh_env__['VIRTUALENV_HOME']
 
-    def create(self, name):
+    def create(self, name, *, system_site_packages=False, symlinks=False, 
+               with_pip=True):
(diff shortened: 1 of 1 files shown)


xontrib/voxapi.py
```diff
@@ -32,16 +32,68 @@ def __init__(self):
...
-    def create(self, name):
+    def create(self, name, *, system_site_packages=False, symlinks=False, 
+               with_pip=True):
         """Create a virtual environment in $VIRTUALENV_HOME with python3's ``venv``.
 
         Parameters
         ----------
         name : str
             Virtual environment name
+        system_site_packages: bool
```
scopatz: These should be in numpydoc format which requires a space on both sides of the colon

scopatz:

astronouth7303:
Fixed.

scopatz:
Thanks a ton!

PR #1459, opened 2016-07-21
amalgamate handles time machine

scopatz:
This isn't super useful to xonsh directly, but we should keep in sync


scopatz:
And it may become useful in the...wait for it...future.

scopatz:
ping!

gforsyth:
Looks good to me.

PR #1460, opened 2016-07-21
Use real options for Vox

astronouth7303:
Ports Vox to use `argparse`.

This enables several knobs in the creation of virtual environments.


xontrib/vox.py
```diff
@@ -1,77 +1,99 @@
 """Python virtual environment manager for xonsh."""
 
 import sys as _sys
+import argparse as _argparse
 import xontrib.voxapi as _voxapi
 
 class _VoxHandler:
     """Vox is a virtual environment manager for xonsh."""
+    parser = _argparse.ArgumentParser(prog='vox', description=__doc__)
```
scopatz: It is probably better to have the parser be constructed the first time it is used (maybe as a property?) rather than at import time. You can see some examples of cached parser construction through out xonsh. We want our imports to be as fast as reasonably possible.

xontrib/vox.py
```diff
@@ -1,77 +1,99 @@
...
+
+    create = subparsers.add_parser('new', help='Create a new virtual environment')
+    create.add_argument('name', metavar='ENV',
+                        help='The environments to create')
+
+    create.add_argument('--system-site-packages', default=False,
+                        action='store_true', dest='system_site',
+                        help='Give the virtual environment access to the '
+                             'system site-packages dir.')
+    from xonsh.platform import ON_WINDOWS
```
scopatz: this could be at the top level

xontrib/vox.py
```diff
@@ -1,77 +1,99 @@
...
+    if ON_WINDOWS:
+        use_symlinks = False
+    else:
+        use_symlinks = True
```
scopatz: also you could set use symlinks to not on_windows, rather than going through an if statement

astronouth7303: Didn't even think about it! Was pretty much just copying what venv does.

xontrib/vox.py
```diff
@@ -1,77 +1,99 @@
...
-            Virtual environment name
         """
         print('Creating environment...')
-        self.vox.create(name)
+        self.vox.create(args.name)
         msg = 'Environment {0!r} created. Activate it with "vox activate {0}".\n'
         print(msg.format(name))
 
-    def activate_env(self, name):
+    def cmd_activate(self, args, stdin=None):
```
scopatz: do you need the cmd_ everywhere? I don't see any code related to looking these methods up by name, but I could be missing them

scopatz:
Thanks for putting this in!

xontrib/vox.py
astronouth7303: They're looked up dynamically in `__call__()`. I just use `cmd_` as a prefix to denote command-handling methods. (Like having `do_` in front of event handlers.)

scopatz: ahhh thanks, that is what I was missing

xontrib/vox.py
astronouth7303: Ok, wow. I just confirmed this with some quick timeit tests.

```python
timeit('reload(xontrib.vox)', 'from imp import reload; import xontrib.vox', number=10000)
```

With the code as it is (49ea0eee15a1e6cdcf93bdf885560031fca5848d), timeit returns a time of 32.5 seconds. When I comment out parser creation, it takes 1.98s.

(For reference, `timeit('reload(argparse)', 'from imp import reload; import argparse', number=10000)` should be 11.3s, but I had to run it at `number=1000` because 10000 was causing some kind of memory/gc issue.)

xontrib/vox.py
```diff
@@ -1,78 +1,102 @@
...
+                            'even when symlinks are the default for '
+                            'the platform.')
+    create.add_argument('--without-pip', dest='with_pip',
+                        default=True, action='store_false',
+                        help='Skips installing or upgrading pip in the '
+                             'virtual environment (pip is bootstrapped '
+                             'by default)')
+
+    activate = subparsers.add_parser('activate', aliases=['workon', 'enter'], help='Activate virtual environment')
+    activate.add_argument('name', metavar='ENV',
```
gforsyth: trailing whitespace here that flake8 is angry about

xontrib/vox.py
```diff
@@ -1,78 +1,102 @@
...
 import xontrib.voxapi as _voxapi
 
 
 class _VoxHandler:
     """Vox is a virtual environment manager for xonsh."""
+    parser = _argparse.ArgumentParser(prog='vox', description=__doc__)
+    subparsers = parser.add_subparsers(dest='command')
+
+    create = subparsers.add_parser('new', aliases=['create'],
+        help='Create a new virtual environment')
```
gforsyth: flake8 complaint here, too, that the line is under-indented.  if you have `help` directly under the first `'` it should be fine.

xontrib/vox.py
```diff
@@ -1,78 +1,102 @@
...
+                            'the platform.')
+    create.add_argument('--without-pip', dest='with_pip',
+                        default=True, action='store_false',
+                        help='Skips installing or upgrading pip in the '
+                             'virtual environment (pip is bootstrapped '
+                             'by default)')
+
+    activate = subparsers.add_parser('activate', aliases=['workon', 'enter'], help='Activate virtual environment')
+    activate.add_argument('name', metavar='ENV', 
+                        help='The environment to activate')
```
gforsyth: flake8 again, under-indented continuation line

astronouth7303:
Ok, got lazy loading working.

(The timeit score for ed5c1c7 is 2.07.)

astronouth7303:
Done and done.

scopatz:
Looks great! Thanks!

PR #1465, opened 2016-07-22
Add OSX tests

santagada:
Just updated @stockninja branch, I have simplifications to run only with python3 from homebrew, but I'm planning on sending that as another PR after this one is merged


gforsyth:
Hey @santagada, thanks for updating this!  Can we use miniconda for this instead as mentioned in the previous PR?  http://conda.pydata.org/docs/travis.html

As it stands, this would add around 9 minutes to every travis run and given the pace of development right now that might be prohibitive.

santagada:
I prefer to move this to homebrew as that is what most people use on osx and is already installed on travis. I will try it then.

gforsyth:
It doesn't seem like brew has a "bottle" for python3.4 -- as long we're supporting both 3.4 and 3.5 both should be tested

santagada:
Why? I think testing on 3.5 is enough no? People on macs usually have the latest python because it is updated independently of the system.

gforsyth:
I have broken the 3.4 builds a couple of times by using features exclusive to Python 3.5, and since we have a few OSX only tests, it is much preferable to test everything.  
I know that the vast majority of mac users use brew, but a lot of scientific python people use conda since it handles a bunch of other stuff and we don't want to leave them out in the cold.

gforsyth:
@santagada -- since I'm making more work for you, if you'd like I can PR your fork with the miniconda changes.  Thanks again for putting this in!

santagada:
yep, add miniconda for python 3.4 no problem

santagada:
it still takes 3 minutes... but at least its using a python version that other people use.

scopatz:
I am strongly in favor of using minicoda for a couple of reasons:

1. mindshare among the core-devs / maintainers. I think most of us know and use conda but few us us know brew, 
2. Consistency across platforms, we could use miniconda on all of our CI platforms, if we needed to.

santagada:
most also don't use osx, brew usage is way more widespread. Also it is already installed in osx and makes all the code simple (2 more lines on the .travis.yaml). If someone wants to do conda I would not oppose but I never used it and not even now what conda versions means.

santagada:
test are passing, timeouts work, sending coverage on only one of the test runs also works.

santagada:
this mostly a superset of #1442

scopatz:
Hi @santagada - there are now conflicts with master on this.  I think that @gforsyth implemented the main features (timeouts, only running coverage tests sometimes, etc) but I might have missed something.  Feel free to update or close! Thanks!

santagada:
Cache for pip is missing but that's easierto just add directly

gforsyth:
Hey @santagada -- does the cache thing help if there's only one build that uses pip?  I thought that travis would cache something, upload it to S3 and then pull it back down for subsequent build jobs?  Or does this persist across builds?

santagada:
You can use the cache to store miniconda install directory to speed things further... Caching pip will only help that one platform. The way cache works on Travis is per matrix entry and you can cache any directory. Also it's easy to clean up later on the ui

PR #1466, opened 2016-07-22
Improve first set of "Developer's Guide" instructions [updated]

santagada:
updated another pull request to speed the improvement of documentation

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -15,6 +15,22 @@ develop xonsh.
 
 .. note:: All code changes must go through the pull request review procedure.
 
+
+Making Your First Change
+========================
+
+First, install xonsh from source and open a xonsh shell in your favorite terminal application. See installation instructions for details.
+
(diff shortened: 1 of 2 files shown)


gforsyth:
Hey @santagada, thanks for putting this in!  I think it might be good to add a reference to installing with `setup.py develop` so that people can make edits to the git repo and not within their `site-packages`.  Or so they don't make changes in the repo directory and expect them to propagate out to `site-packages`.  Does that make sense?

laerus:
++ @gforsyth we should also change the developer mode to not amalgamate. i will try PR this soon, if noone else want to mess with that.

santagada:
@gforsyth and @laerus I think both make sense, but should be done after merging this in right?

gforsyth:
@santagada -- yes, I'll merge this now.  
@laerus -- it's all yours, or I'll try to tackle it later today.

gforsyth:
Thanks @santagada and @stockninja !

PR #1468, opened 2016-07-22
Updates to #1465 using miniconda everywhere

gforsyth:
This builds on @stockninja and @santagada 's work in #1465 and #1442, using miniconda for all CI builds on travis except for "nightly"

I'm breaking it out into a separate PR to check that a) it works and b) that it's faster (which I think it should be, but let's find out).  

Also it's a bit verbose at the moment, if this works I'll switch over to using an `environment.yml` file for `conda`


install.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
scopatz: Do we still need this install file?

scopatz:
Thanks @gforsyth!

install.sh
gforsyth: nope, I'm going to remove it.

scopatz: Great!

santagada:
is it supposed to only pass on nightly?

gforsyth:
Not by design :)

gforsyth:
@scopatz -- seems like miniconda is a little slower than just using travis python on Linux.  We can save around 40 seconds if we don't use miniconda on the linux builds.

also, the vox tests are a problem since `venv` and conda really don't get along very well.

scopatz:
> @scopatz -- seems like miniconda is a little slower than just using travis python on Linux. We can save around 40 seconds if we don't use miniconda on the linux builds.

I am impartial on this.  It might be nice to have faster tests, but it is also nice have consistency.  But I guess we already switch for nightlies....  Either way.

> also, the vox tests are a problem since venv and conda really don't get along very well.

Is there any way to skip these if we know we are in a conda env?  Maybe check if `__file__` or the path contains the word conda? Or that the `conda` cli is on the $PATH?

gforsyth:
wtf... so it looks like conda's python3.5 builds don't include `ensurepip` which is part of the standard library.  that seems like a bug to me.

scopatz:
I think this is one of the reasons why conda-forge starting building their own python.

gforsyth:
Yeah -- I just installed that to test it out and now I'm getting a 3 year old bug where the `sys.version` regex is wrong.  This is a fun little rabbit hole.

scopatz:
Uggh, well, I am in favor of whatever works :)

scopatz:
Let us know when you think this is ready!

scopatz:
Seems like there are merge conflicts now.

gforsyth:
@scopatz @santagada -- can you take a look at this now?  All tests passing on travis and appveyor.  Total test time is less than 6 minutes for the 5 builds, but delay between each test starting makes the total walltime closer to 10 minutes.  This may be a temporary travis thing, or something else at work.

scopatz:
I think that this is a Travis thing because of availability of machines.  Thanks a million @gforsyth!

PR #1473, opened 2016-07-23
history refactor

laerus:
refactoring `_hist_show()` to call the new `_hist_get()` function that uses generators to filter history commands.

`test_show_cmd` refactor to pytest style.

`ensure_slice` in tools now accepts iterables

new tests, more refactoring


xonsh/history.py
```diff
@@ -407,103 +406,84 @@ def _hist_create_parser():
...
-               start_time=None, end_time=None, location=None):
-    """Show the requested portion of shell history.
-    Accepts multiple history sources (xonsh, bash, zsh)
-
-    May be invoked as an alias with history all/bash/zsh which will
-    provide history as stdout or with __xonsh_history__.show()
-    which will return the history as a list with each item
-    in the tuple form (name, start_time, index).
+def _hist_get_session(session='session', location=None):
+    "Yield commands from history."
```
scopatz: PEP8, docstrings should be tripple quoted

xonsh/history.py
```diff
@@ -407,103 +406,84 @@ def _hist_create_parser():
...
+    cmds = _HIST_SESSIONS[session](location=location)
+    try:
+        yield from cmds
+    except TypeError:
```
scopatz: Can't you test if cmds is iterable or None or something rather than relying on a try-except?

xonsh/history.py
```diff
@@ -407,103 +406,84 @@ def _hist_create_parser():
...
+        pass
+
+
+def _hist_get_portion(commands, slices):
+    "Yield from portions of history commands. "
+    commands = list(commands)
```
scopatz: Why do you need to convert this to a list?

xonsh/history.py
```diff
@@ -407,103 +406,84 @@ def _hist_create_parser():
...
+    "Yield commands from history."
+    cmds = _HIST_SESSIONS[session](location=location)
+    try:
+        yield from cmds
+    except TypeError:
+        pass
+
+
+def _hist_get_portion(commands, slices):
+    "Yield from portions of history commands. "
```
scopatz: PEP8 docstring

xonsh/history.py
```diff
@@ -407,103 +406,84 @@ def _hist_create_parser():
...
+
+
+def _hist_filter_ts(commands, start_time=None, end_time=None):
+    """Yield only the commands between start and end time."""
+    if start_time is None:
+        start_time = 0.0
+    if end_time is None:
+        end_time = float('inf')
+    for cmd in commands:
+        if end_time > cmd[1] > start_time:
```
scopatz: shouldn't one of these be `>=`, probably the start time. So, maybe `end_time > cmd[1] >= start_time`

xonsh/history.py
```diff
@@ -407,103 +406,84 @@ def _hist_create_parser():
...
+    -------
+    list
+       A filtered list of commands
+    """
+    cmds = _hist_get_session(session, location)
+    if slices:
+        cmds = _hist_get_portion(cmds, slices)
+    if start_time or end_time:
+        cmds = _hist_filter_ts(cmds, start_time, end_time)
+    return list(cmds)
```
scopatz: List again?

xonsh/history.py
```diff
@@ -407,103 +406,84 @@ def _hist_create_parser():
...
+    """Show the requested portion of shell history. Accepts same parameters
+    with `_hist_get`."""
     try:
-        commands = _HIST_SESSIONS[ns.session](hist=hist, location=location)
-    except KeyError:
-        print("{} is not a valid history session".format(ns.action))
-        return None
+        commands = _hist_get(ns.session, ns.slices, **kwargs)
+    except ValueError as err:
+        print(err, file=sys.stderr)
```
scopatz: I think the previous error message was a little more informative.  Maybe you could merge the two.

xonsh/history.py
```diff
@@ -673,6 +653,9 @@ def _hist_parse_args(args):
     parser = _hist_create_parser()
     if not args:
         args = ['show']
+    elif args[0] in _HIST_SESSIONS:
+        print('Use the `show` subcommand to get history sessions.', file=sys.stderr)
```
scopatz: I don't think the backticks are needed here.  Maybe regular quotes are more readable

xonsh/history.py
```diff
@@ -673,6 +653,9 @@ def _hist_parse_args(args):
...
+        raise SystemExit
```
scopatz: Don't raise SystemExit or call sys.exit(), please.  This can/will take out the whole xonsh interpreter.

scopatz: Just return instead

xonsh/tools.py
```diff
@@ -926,9 +926,11 @@ def SLICE_REG():
 
 
 def ensure_slice(x):
-    """Convert a string or int to a slice."""
-    if x is None:
+    """Convert `x` to a slice. If `x` is already a slice object return as is."""
```
scopatz: Single backticks not valid rst.

xonsh/tools.py
```diff
@@ -941,7 +943,10 @@ def ensure_slice(x):
         else:
             raise ValueError('cannot convert {!r} to slice'.format(x))
     except TypeError:
-        raise TypeError('ensure_slice() argument must be a string or a number not {}'.format(type(x)))
+        try:
+            s = slice(*(int(i) for i in x))
+        except (TypeError, ValueError):
+            raise ValueError('cannot convert {!r} to slice'.format(x))
```
scopatz: I think this is a type error, not a value error

scopatz:
Thanks for putting this in @laerus! I am in super favor of this refactor. Comments in-line.

xonsh/history.py
laerus: we may iterate over commands more than once if we have more than one slice.

scopatz: But you don't need list for that right?  You just need iter?

scopatz: Or why isn't commands comming in as a list or tuple in the first place is my question

scopatz: Or can you yeild from `itertools.islice()` https://docs.python.org/3.5/library/itertools.html#itertools.islice

laerus: tried with islice at first, islice doesnt accept negative ints, we could hack the indexes to become positive but we need the len of commands anyway. i was thinking about itertools.tee but its an overskill.

scopatz: OK, sure, but we *do* know the length of the commands at one point because everything in `_HIST_SESSIONS` returns a list.  You can then use `slice.indicies()` to get positive indicies based on the length of the commands.  

I guess my point is that `_hist_get_session()` gets a list and then turns it into a iterator.  This function then gets an iterator and turns it back into a list with the same contents.  That seems a little excessive.  It would be better to just use the original list directly or have `_hist_get_session()` return the list it sees.

laerus: i was planning on changing the session functions to generator funcs aswell(they can be refactored aswell anyway).  if we have only one slice (which may be the most common case) we can just islice the commands in `_hist_get_portion` . i will play a little more with islice.

xonsh/tools.py
laerus: i gave a thought on that. i feel like TypeError would be if we didnt accept the type, its the value of x that we try to make into a slice and complain if it cannot be explicitly done. in short, we dont care about the type.

xonsh/history.py
scopatz: OK, yeah.  The thing about iterators is that if it is going to be iterators it usually has to be iterators all the way down :)

xonsh/tools.py
scopatz: Hmm yeah, I guess this isn't a big deal either way and is non-blocking.  It sort of depends how you look at it.  Maybe it should just retain the error type the was caught?

xonsh/history.py
laerus: Sadly i knew that when i started implementing this with generators. Getting only once slice would be easier but i like the multiple slice feature. I will try and improve this and see if the implementation can be clean else i will drop to using lists. (:heart:  generators)

xonsh/history.py
laerus: maybe both?

scopatz: Well, probably not both.  The Pythonic thing is to be inclusive on the lower end and exclusive on the upper end.

xonsh/tools.py
laerus: TypeError is being raised on `int()` not accepting lists and tuple etc. In the traceback you can see the chain of exceptions. This way it's easier for the caller to just except ValueError too.

xonsh/history.py
laerus: you're right it's a temporary hack mostly, i may change arg parsing a little.
(btw i think we should foolproof against SystemExit, many parsers raise this on bad arguments or when printing help msgs.)

xonsh/tools.py
laerus: i'm totally confused with rst, from the numpy doc example

```
Several sentences providing an extended description. Refer to
variables using back-ticks, e.g. `var`.
```

scopatz: I guess numpy doc is using a purposefully undefined part of rst: http://docutils.sourceforge.net/docs/user/rst/quickref.html#inline-markup

scopatz: I suppose this is valid, then.

laerus:
i think we can remove enumeration on session parsers and enumerate when we want to, maybe an option to show subcommand `-n` to choose to numerate the commands.

laerus:
argument parsing is a mess atm, will try and re-implement  `_hist_parse_args` to pass the tweaked tests.

scopatz:
Thanks, just let us know when you have it working!

laerus:
this is ready for another look.

gforsyth:
Hey @laerus -- one small thing in a couple of places: if you have multiline docstrings then the convention is to have the ending triple quote be on its own line

xonsh/history.py
```diff
@@ -301,28 +287,20 @@ def _curr_session_parser(hist=None, **kwargs):
...
     if location is None:
-        location = _find_histfile_var(location_list, default_location)
-    z_hist_formatted = []
-    if location and os.path.isfile(location):
-        with open(location, 'r', errors='backslashreplace') as z_file:
-            z_txt = z_file.read()
-            z_hist = z_txt.splitlines()
-            if z_hist:
-                for ind, line in enumerate(z_hist):
+        location = _find_histfile_var(['~/.zshrc', '~/.zprofile'], '~/.zsh_history')
```
gforsyth: I'm not sure, but this might cause problems for people using `zsh` on Windows

gforsyth: That is, having the joined path hardcoded in posix-style

xonsh/history.py
```diff
@@ -331,29 +310,23 @@ def _zsh_hist_parser(location=None, **kwargs):
...
-    bash_hist_formatted = []
-    if location and os.path.isfile(location):
-        with open(location, 'r', errors='backslashreplace') as bash_file:
-            b_txt = bash_file.read()
-            bash_hist = b_txt.splitlines()
-            if bash_hist:
-                for ind, command in enumerate(bash_hist):
-                    bash_hist_formatted.append((command, 0.0, ind))
-                return bash_hist_formatted
+        location = _find_histfile_var(['~/.bashrc', '~/.bash_profile'], '~/.bash_history')
```
gforsyth: This, too, should probably be `os.path.join`

gforsyth:
Hey @laerus, this is looking good.  Is the timestamp based filtering supposed to be exposed in the argument parser?

xonsh/history.py
laerus: oh yea that's any issue, i didn't change the paths, no idea where zsh saves rc and stuff on windows. will look that up.

xonsh/history.py
laerus: ok now i get it. forgot windows use backslash..

laerus:
@gforsyth we should expose it right? it won't be that hard, i was thinking about doing that in a different PR. the api is here anyway

gforsyth:
Ok, sounds good to me!  I'm also going to write completions for this

gforsyth:
@laerus -- this looks good to me.  I do think that we need some small changes to the docs as history is not automatically enumerated now, but I don't think that's blocking (as we have a few more history related PRs upcoming soon).  Anything else you wanted to get in here before merging?

laerus:
i feel like we need more tests but this can be another PR. i think we can merge this.

gforsyth:
Great!  Thanks, @laerus !

Issue #1470, opened 2016-07-22
Disable amalgamation on install develop


laerus:
@scopatz how can we go about doing this?

on `setup.py` can `amalagamate_source()` be safely moved out of `build_tables()` ?

(btw is amalagamate a typo?)

scopatz:
Yeah, it can be moved out, but the sys.path maniputlations need to be duplicated too.  It probably should be moved out of build tables, I was just lazy :)

PR #1475, opened 2016-07-23
no amalgmation on setup develop

laerus:
#1470

@scopatz something like this?

or can we get the sys.path manimulation directly into main with no implications?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,7 @@ import os
 import sys
 import json
 import subprocess
+import contextlib
 
 try:
     from tempfile import TemporaryDirectory
@@ -35,6 +36,12 @@ except ImportError:
     HAVE_JUPYTER = False
 
 
+@contextlib.contextmanager
+def set_path():
+    sys.path.insert(0, os.path.dirname(__file__))
+    yield
+    sys.path.pop(0)
+
 TABLES = ['xonsh/lexer_table.py', 'xonsh/parser_table.py',
           'xonsh/__amalgam__.py', 'xonsh/completers/__amalgam__.py']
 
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -35,6 +36,12 @@
     HAVE_JUPYTER = False
 
 
+@contextlib.contextmanager
+def set_path():
```
santagada: set_cwd_path() or maybe a docstring could  be good to know what this does

laerus: yes will eventually shape it up. if we stick with this

news/amalgamation.rst
```diff
@@ -0,0 +1,15 @@
+**Added:** None
+
+**Changed:**
+
+* Moved `amalgamate_source` outside `build_tables`
```
scopatz: Need double backticks in rst

setup.py
```diff
@@ -65,12 +72,9 @@ def amalagamate_source():
 def build_tables():
     """Build the lexer/parser modules."""
     print('Building lexer and parser tables.')
-    sys.path.insert(0, os.path.dirname(__file__))
     from xonsh.parser import Parser
```
scopatz: The parser tools need to be inside of a `with set_path()`

setup.py
```diff
@@ -184,7 +190,9 @@ class xsdist(sdist):
     """Xonsh specialization of setuptools sdist class."""
     def make_release_tree(self, basedir, files):
         clean_tables()
-        build_tables()
+        with set_path():
```
scopatz: Ahh I see.  I would move the set_path calls to the functions themselves.

scopatz:
This seems reasonable.  Just a couple of small changes.

laerus:
oh wow, the machine i'm currently on with a flash drive had its date at 2006..
i will fix this when i'm back on a normal machine.

scopatz:
This looks good to me!  Let us know when it is ready!

laerus:
@scopatz i think this is ready

scopatz:
Thanks!

PR #1476, opened 2016-07-23
Amalgamate relative imports

scopatz:
Not that we'll use this in xonsh, but this is support for amalgamating relative imports from the same package. With this both, 

```python
from xonsh.main import main
from .main import main
```

will be amalgamated equivalently.  Previously, only the top example would amalgamate.

See also xonsh/amalgamate#1 and xonsh/amalgamate#5

### DIFF
--- a/amalgamate.py
+++ b/amalgamate.py
@@ -185,7 +185,7 @@ def module_from_package(module, pkg, level):
     elif level == 1:
         return True
     else:
-       return False
+        return False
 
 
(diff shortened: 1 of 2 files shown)


gforsyth:
this looks good.  We may want to consider pulling in amalgamate and lazyasd as subtrees soon if there's much more work done on them.  maintaining parallel code bases isn't fun.

scopatz:
Yeah, hopefully the churn on these has mostly stopped!

Issue #1368, opened 2016-06-29
Background processes make stdin invisible


Granitosaurus:
Tested in v0.4.2 and ipython 4.1.2 and python 3.5

```
$ python
>> "foo"
>> Ctrl+z
$ python
>> <invisible text>
>> Ctrl+z or even Ctrl+d
$ fg python(first one)
>> "foo" # everything is visible here
>> Ctrl+z
$ killall python
# every stdin from this point is invisible and only a restart of xonsh can fix it
# seems like the original job with stdin takes the stdin to the grave with it :D
```

Even though the text is invisible it is still there and is interpreted as it would be normally.
Now it seems that the only way to resolve this is to restart xonsh shell completely since killing of the ipython  processes doesn't resolve the issue.

Edit: this is also an issue with standard python shell
Edit: It appears that stdin in `cat` is visible but is all messed up, like it doesn't allow you to delete characters or insert newlines.

scopatz:
Thanks for reporting @Granitas. The same basic thing happens to me in xonsh and Bash with the hiding of stdin.  Conceptually, I am not sure that it is well defined to show stdin on a background process (it is in the background) because its special file descriptors for stdin, stdout, and stderr are all being redirected.

What is the behaviour you were expecting and what changes can be made?  Sorry I haven't thought a lot about this corner case

Granitas:
@scopatz The thing here is that a suspended job shouldn't block stdin. I'm I'm not having this issue on bash or zsh and I'm not sure but I think this bug didn't exist in Xonsh 0.3.

scopatz:
Ahh that is fair. This is almost certainly a job control issue then.  Would you mind trying to find the commit that broke the behaviour for you with git bisect?  That would really help!

PaulReiber:
if the characters from stdin are being consumed and processed just fine,
but are "invisible" that means they're not getting echoed out onto the
screen as you type them.

That's an aspect of the terminal's mode.  When you control-Z there's no
telling what the terminal mode will be left in as the process suspends, and
usually, the process being suspended is pretty stupid about cleaning up
after itself.  Some do better than others - vi and emacs reset the terminal
mode nicely when backgrounding, for example.  But not all programs play
fair, so the xonsh shell needs to ensure it's setting the mode right in
these case, rather than leaving things to chance.

Kind regards,


*Paul Reiber*http://about.me/reiber

scopatz:
Pull requests very welcome here.

[...]

PR #1478, opened 2016-07-23
first attempt (non-working) at setting the terminal mode ECHO back on

PaulReiber:
Attempting to fix this issue:  https://github.com/xonsh/xonsh/issues/1368

The strategy is simple enough - set the terminal's mode to be what we want (typed characters should be echoed) at just the right time - before prompting and reading a line of input from the end user.

Unsure why it's not working.  Anthony had asked that I submit a PR anyway, so here it is.

Happy to help!

### DIFF
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -15,6 +15,7 @@ import sys
 import cmd
 import time
 import select
+import termios
 import builtins
 import importlib
 import threading
@@ -230,6 +231,14 @@ class ReadlineShell(BaseShell, cmd.Cmd):
     def __del__(self):
         teardown_readline()
 
+
+    def sttysane(self):
+        """Change settings of stdin to include ECHO so characters typed at the keyboard are echoed on screen"""
+        fd = sys.stdin.fileno()
+        si_attrs = termios.tcgetattr(fd)
+        si_attrs[3] = si_attrs[3] & termios.ECHO
+        termios.tcsetattr(fd, termios.TCSADRAIN, si_attrs)
+
     def singleline(self, store_in_history=True, **kwargs):
         """Reads a single line of input. The store_in_history kwarg
         flags whether the input should be stored in readline's in-memory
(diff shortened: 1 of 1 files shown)


xonsh/readline_shell.py
```diff
@@ -15,6 +15,7 @@
 import cmd
 import time
 import select
+import termios
```
scopatz: This isn't sufficiently cross platform.  You'd probably want to move the lazy imports from teepty ([see here](https://github.com/xonsh/xonsh/blob/master/xonsh/teepty.py#L25)) to lazyimps.py, and then import from lazyimps.py in both teepty and this module.

xonsh/readline_shell.py
```diff
@@ -241,6 +250,7 @@ def singleline(self, store_in_history=True, **kwargs):
             except ImportError:
                 store_in_history = True
             pos = readline.get_current_history_length() - 1
+        self.sttysane()
```
scopatz: This should not be run on windows.

xonsh/readline_shell.py
```diff
@@ -230,6 +231,14 @@ def __init__(self, completekey='tab', stdin=None, stdout=None, **kwargs):
     def __del__(self):
         teardown_readline()
 
+
+    def sttysane(self):
+        """Change settings of stdin to include ECHO so characters typed at the keyboard are echoed on screen"""
+        fd = sys.stdin.fileno()
+        si_attrs = termios.tcgetattr(fd)
+        si_attrs[3] = si_attrs[3] & termios.ECHO
+        termios.tcsetattr(fd, termios.TCSADRAIN, si_attrs)
```
scopatz: Yeah, not sure why this doesn't work, but maybe it has to do with readline taking over stdin.

scopatz:
Thanks a million for putting this in!

laerus:
btw this is still an issue on 0.5.2 and current master.

scopatz:
I believe that this is superceded by #2123. Please feel free to reopen.

PR #1482, opened 2016-07-25
more intuitive blocking loop

scopatz:
CC @janschulz, This is just to keep updated with xonsh/lazyasd#3


janschulz:
Isn't it easier to simply take a complete release?

scopatz:
Yeah, I'd rather copy things over sooner than later for exactly that reason.

melund:
@janschulz Nice simplification. When I first looked at the code it also took me a little while to grasp what this simple part of the code did.

gforsyth:
Looks good!  Thanks, @scopatz !

PR #1485, opened 2016-07-26
Fix build

santagada:
flake8 3.0.0 breaks the flake8 plugin, I’m opening a ticket there, but in the meantime this should fix it on our side


scopatz:
Thanks a ton!

PR #1486, opened 2016-07-26
base version should keep unchanged via setup

mitnk:
Let base version not depend on git tags which may not at the latest states. While the version in `xonsh/__init__.py` should be used instead, it is the correct version number (always).


setup.py
```diff
@@ -136,21 +136,24 @@ def dirty_version():
...
 
 
-def replace_version(base, N):
+def replace_version(N):
     """Replace version in `__init__.py` with devN suffix"""
     global ORIGINAL_VERSION_LINE
     with open('xonsh/__init__.py', 'r') as f:
         raw = f.read()
     lines = raw.splitlines()
+    assert '__version__' in lines[0], 'version should define at the top'
```
scopatz: Please make this message more explicit about the `__init__.py` file, for example `'__version__' must be the first line of the __init__.py`

scopatz:
Thanks @mitnk! Just a minor update assert error message is all that is needed.  Feel free to add a news entry too if you want.

mitnk:
updated, thanks!

scopatz:
@santagada - any ideas on why flake8 is failing to install here?

@mitnk - can you try merging from master, please?  I think on master this uses flake8<3.0, but here the version is >=3.0, which breaks the py.test compatibility.

mitnk:
still failing about merging

mitnk:
in #1485 it fixes on flake8 3.0.0. Now it's 3.0.2, things may have changed. let me have a try.

santagada:
I have a pull request that fixes this... now we fix versions so this will hopefully not happen again.

mitnk:
thanks!

PR #1491, opened 2016-07-27
change from set window title to window + icon (tab) title

santagada:
On iterm2 (and maybe other terminals) you have a way to set tab title and window title independent of each other. By changing this escape we set both to the same value, which should be ok until we have a better way to do this.

Documentation for it in http://tldp.org/HOWTO/Xterm-Title-3.html

### DIFF
--- a/news/set-title.rst
+++ b/news/set-title.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+* $TITLE now changes both icon (tab) and window title
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/xonsh/base_shell.py
+++ b/xonsh/base_shell.py
@@ -246,7 +246,7 @@ class BaseShell(object):
             with open(1, 'wb', closefd=False) as f:
                 # prevent xonsh from answering interative questions
                 # on the next command by writing the title
-                f.write("\x1b]2;{0}\x07".format(t).encode())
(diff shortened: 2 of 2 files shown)


gforsyth:
Hey @santagada -- nightly is failing because now it wants a version of `flake8 > 3.0` -- it looks like `pytest-flake8` just pushed out an update late yesterday to support `3.0` so if you update the `requirements-tests.txt` to set `flake8 >= 3.0` it _should_ work.

Otherwise, this change is fine with me, and importantly, doesn't seem to have any impact on the interactivity bug that was causing us so much trouble before.

santagada:
Ok I wil submit another patch @gforsyth don't you think we should start pinning our versions on requirements? then we don't have to do this fixes, just update things when we want to.

gforsyth:
@santagada -- i think that's a good idea.  we didn't used to, since the pace of development of both `xonsh` and `prompt_toolkit` was pretty rapid.  It's still pretty fast now, but things are more stable all around.  

I'm happy with anything that means less mucking around cancelling builds on Travis after some dependency mismatch causes blanket failures.

scopatz:
@asmeurer and I dug into this at one point and there was a bug in iterm2 that was preventing this from working.  I am surprised that this works for you now.  If this works, though, I am all for it.

scopatz:
Also, can you maybe please add a news entry for this too!

santagada:
done, and tested it on iterm2 and terminal.app both works

scopatz:
Thanks!

PR #1492, opened 2016-07-27
remove flake8 checks because they are broken

gforsyth:
the pytest flake8 extension is not working particularly well.  I am not opposed to having it enabled if it doesn't continuously break CI.  This strips it out.  I am happy to re-enable it if someone submits a PR that can pass CI


santagada:
just did that on #1493

gforsyth:
Nice!

PR #1493, opened 2016-07-27
update flake8 to 3.0.2

santagada:
during the update we have to ignore some new errors on our current
codebase


santagada:
@gforsyth here it is, updated to the latest flake8. Can I submit another pull to fix the package versions instead of using >= ?

gforsyth:
@santagada -- that's fine with me.  You can do it here or in a separate PR.

santagada:
done... this I think will be more stable

santagada:
somehow we hit a problem between flake8 3.0.2, timeouts and anaconda. As it was not happening for me it was impossible to fix so we have fixed on a old flake8

I will create a new pull request removing pytest-flake8 and running flake8 independently from py.test, that should fix it.

gforsyth:
Thanks @santagada!  This looks like a good fix until we rework the `flake8` behavior.

Issue #1490, opened 2016-07-26
Aliased commands don't have autocompletion


fisadev:
This is a feature request, inspired on zsh and fish, which already do this.

If I create this alias:

    aliases['g'] = "git"

I would like to have the proper git autocompletion for the new "g" command. 
Currently if I do:

    git br<TAB>

It correctly completes with "branch". But if I do:

    g br<TAB>

It does nothing.

scopatz:
thanks for reporting! I thought that this used to work, but it might have been lost in the transition to the new completer system.

gforsyth:
this definitely used to work.  The original issue was #391 and it was implemented in #526.  The new completer system seems to have undone some of that.

gforsyth:
Fixed in #1495.  Please reopen if the issue persists.  Thanks!

PR #1495, opened 2016-07-27
re-enable tab completion for aliases

gforsyth:
in re: #1490 

this re-enables tab completion for aliases by using the handy (and already present) `expand_alias` method


scopatz:
Thanks for putting this in @gforsyth! Does a similar change also need to go into jupyter_hook.py?

scopatz:
Confirmed that this seems to work locally.

gforsyth:
oo, nice catch.

PR #1497, opened 2016-07-27
Document History access fields

### LINKED ISSUE #1275: Document sequence attributes of the history object
```python
self.tss = CommandField('ts', self)
self.inps = CommandField('inp', self)
self.outs = CommandField('out', self)
self.rtns = CommandField('rtn', self)
```

It turns out these are supposed to be the public interface to the history object but they scared me away because they're missing in [the documentation](http://xon.sh/api/history.html#xonsh.history.History)

astronouth7303:
Fixes #1275.

### DIFF
--- a/xonsh/history.py
+++ b/xonsh/history.py
@@ -507,14 +507,14 @@ def _hist_show(ns=None, hist=None, start_index=None, end_index=None,
 class History(object):
     """Xonsh session history.
 
-    History is accessed through several sequence-like fields:
-
+    Attributes
+    ----------
     rtns : sequence of ints
         The return of the command (ie, 0 on success)
     inps : sequence of strings
         The command as typed by the user, including newlines
     tss : sequence of two-tuples of floats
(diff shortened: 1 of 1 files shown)


xonsh/history.py
```diff
@@ -505,7 +505,22 @@ def _hist_show(ns=None, hist=None, start_index=None, end_index=None,
 # Interface to History
 #
 class History(object):
-    """Xonsh session history."""
+    """Xonsh session history.
+
+    History is accessed through several sequence-like fields:
+
+    rtns : sequence of ints
```
scopatz: I think that these should be in an Attributes section as per numpydoc https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#class-docstring

scopatz:

astronouth7303:
Clean up.

scopatz:
Thanks a million!

PR #1500, opened 2016-07-28
Make `which -v` extremely vocal

### LINKED ISSUE #306: type builtin, or some equivalent
In bash, you can use the `type` builtin to see if a command is a file, alias, or shell function. It would be useful to have this in xonsh, especially to tell if a command is coming from xonsh or from subprocess mode. Right now, type calls `/usr/bin/type` (not sure what that is all about). This would also be useful if xonsh ever implements PATH hashing. 

Note that `type` is also a command on Windows which means something else entirely. Bash also has `command -V` which is the same as `type`.

astronouth7303:
Fixes #306 

When you pass the `-v` flag to `which`, it'll now call superhelp.

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -24,6 +24,7 @@ from xonsh.xontribs import xontribs_main
 from xonsh.xoreutils import _which
 from xonsh.completers._aliases import completer_alias
 
+
 class Aliases(abc.MutableMapping):
     """Represents a location to hold and look up aliases."""
 
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -24,7 +24,6 @@ from xonsh.xontribs import xontribs_main
 from xonsh.xoreutils import _which
 from xonsh.completers._aliases import completer_alias
 
-
 class Aliases(abc.MutableMapping):
     """Represents a location to hold and look up aliases."""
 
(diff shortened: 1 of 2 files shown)


xonsh/aliases.py
```diff
@@ -402,6 +401,7 @@ def which(args, stdin=None, stdout=None, stderr=None):
                     print(arg, file=stdout)
             else:
                 print("aliases['{}'] = {}".format(arg, builtins.aliases[arg]), file=stdout)
+                builtins.__xonsh_superhelp__(builtins.aliases[arg])
```
scopatz: I think you might want to check that `aliases[arg]` is callable() before running this line.  Otherwise you will get superhelp for a list of args.

scopatz:
Thanks for putting this in @astronouth7303!  Could you also please add a news entry:  http://xon.sh/devguide.html#changelog

xonsh/aliases.py
astronouth7303: Oh yup! I thought it did that already.

scopatz: I don't think so at this stage.

astronouth7303:
Done

xonsh/aliases.py
```diff
@@ -402,6 +402,8 @@ def which(args, stdin=None, stdout=None, stderr=None):
                     print(arg, file=stdout)
             else:
                 print("aliases['{}'] = {}".format(arg, builtins.aliases[arg]), file=stdout)
+                if not isinstance(builtins.aliases[arg], list):
```
scopatz: I think this is better as a duck-typed `if callable(builtins.aliases[arg])`

astronouth7303: It's the exact test used 4 lines up for the same thing.

scopatz: That one is probably wrong too.  It should at least be an `abc.Sequence` test above.  And this one is explicitly for callables

astronouth7303:
quack quack

scopatz:
Thanks @astronouth7303!

PR #1501, opened 2016-07-28
Implementation of Event Subsystem

### LINKED ISSUE #1489: Idea: Event System?
Would it be a good idea to have some form of hooking or event system?

* xontribs like avox and autoxsh do not have to shadow `cd` in order to work
* Ability to run code before or after prompt execution

So far, I've identified three events: changing directory, pre-command, and post-command.

astronouth7303:
Fixes #1489.

This is my stab at an event bus, after talking to @scopatz.

Please! Comments, questions, etc! I'm open to any ideas, or even scrapping this entire model of events.

Basic usage, as it stands:

```python
@events.on_precommand
def myhandler(cmd):
    print("About to run command:", cmd)
```

This covers stand-alone functions, static methods, and maybe class methods? Instance methods certainly do not work.

I believe the docstrings cover most of the usage, even if they currently ignore rst and numpy.


xonsh/events.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def untilFalse(self, *pargs, **kwargs):
+        """
+        Calls each handler until one returns something falsey.
+        """
+        for rv in self.calleach(*pargs, **kwargs):
+            if not rv:
+                return
+
+    def loopback(self, *pargs, **kwargs):
```
astronouth7303: I'm open to suggestions on this name in particular.

xonsh/events.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def __call__(self, *pargs, **kwargs):
+        """
+        The simplest use case: Calls each handler in turn with the provided
+        arguments and ignore the return values.
+        """
+        for _ in self.calleach(*pargs, **kwargs):
+            pass
+
+    def untilTrue(self, *pargs, **kwargs):
```
gforsyth: Pep8 here. No camelcase, use `until_true` instead

xonsh/events.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        Calls each handler until one returns something truthy.
+
+        Returns that truthy value.
+        """
+        for rv in self.calleach(*pargs, **kwargs):
+            if rv:
+                return rv
+
+    def untilFalse(self, *pargs, **kwargs):
```
gforsyth: Ditto on camelcase

xonsh/events.py
```diff
@@ -0,0 +1,113 @@
...
+
+        NOTE: Each handler must return the same number of values it was
+        passed, or nothing at all.
+        """
+        calls = self.calleach(*pargs, **kwargs)
+        newvals = next(calls)
+        while True:
+            if newvals is not None:
+                if len(pargs) == 1:
+                    pargs = newvals,
```
gforsyth: On my phone so can't test this at the moment but what's up with this trailing comma?

astronouth7303: Makes it a one-tuple. The assumption is that if the event takes one argument, it'll only return one value.

xonsh/events.py
```diff
@@ -0,0 +1,113 @@
...
+        This works as a generator. Each handler is called in turn and its
+        results are yielded.
+
+        If the generator is interupted, no further handlers are called.
+
+        The caller may send() new positional arguments (eg, to implement
+        modifying semantics). Keyword arguments cannot be modified this way.
+        """
+        for handler in self:
+            newargs = yield handler(*pargs, **kwargs)
```
melund: Neat . But what is the idea of altering the positional arguments of other handlers if you can't guarantee the order the are called.

xonsh/events.py
laerus: you can use parens to make one-tuples more readable.

xonsh/events.py
astronouth7303: A specific use case was in the implementation of command preprocessors.

Note that even if the event system guaranteed order, we're not guaranteed what order the user would load xontribs in.

xonsh/events.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    def __init__(self, doc=None):
+        self.__doc__ = doc
+
+    def handler(self, callable):
+        """
+        Registers a handler. It's suggested to use this as a decorator.
+        """
+        self.add(callable)
+        return callable
```
scopatz: Probably shouldn't overwrite a builtin.

xonsh/events.py
```diff
@@ -0,0 +1,113 @@
...
+
+In all likelihood, you want builtins.__xonsh_events__
+
+The best way to "declare" an event is something like::
+
+    __xonsh_events__.on_spam.__doc__ = "Comes with eggs"
+"""
+
+
+class Event(set):
```
scopatz: Shouldn't this be a Mapping, not a Set.  Also, please don't inherit from builtin types, but instead use `collections.abc`

xonsh/events.py
```diff
@@ -0,0 +1,113 @@
...
+    A given event that handlers can register against.
+
+    Acts as a ``set`` for registered handlers.
+
+    Note that ordering is never guarenteed.
+    """
+    def __init__(self, doc=None):
+        self.__doc__ = doc
+
+    def handler(self, callable):
```
scopatz: I would call this `register()`

xonsh/events.py
scopatz: Or actually this should be a Sequence and Events should be the Mapping.  Order matters here.

xonsh/events.py
```diff
@@ -0,0 +1,113 @@
...
+            except StopIteration:
+                break
+
+        if newvals is not None:
+            return newvals
+        else:
+            return pargs
+
+
+class Events:
```
scopatz: This should be a mapping.  I think the attribute access is a nice convenience, but the underlying mapping is more important.

xonsh/events.py
```diff
@@ -0,0 +1,113 @@
...
+
+        This works as a generator. Each handler is called in turn and its
+        results are yielded.
+
+        If the generator is interupted, no further handlers are called.
+
+        The caller may send() new positional arguments (eg, to implement
+        modifying semantics). Keyword arguments cannot be modified this way.
+        """
+        for handler in self:
```
scopatz: Also, I think that is is important that each handler function (optionally) has a validator function that takes the same arguments but returns a true/false.  The handler is then only called if the validator return true.  I think this is going to be really important for things like cd handlers, where you may not always want to run the handler function.

scopatz:
Thanks a million for putting this in @astronouth7303!

xonsh/events.py
astronouth7303: I inherited from a builtin type because I want the storage as well, not just implementing the interface, since I'd just be proxying a base storage type anyway.

As for order, I said elsewhere: "Note that even if the event system guaranteed order, we're not guaranteed what order the user would load xontribs in, or even what xontribs will be loaded."

Events is a plain object and not a mapping because of syntactic sugar, although there's no reason it can't be a mapping, too.

xonsh/events.py
astronouth7303: I think handler makes more sense when reading it in usage, since you use it on event handlers.

xonsh/events.py
astronouth7303: I can't imagine a use case where the handler couldn't fail fast instead of having an extra lambda (or worse, a full function).

cd, in particular, I don't agree with you. The two xontribs that would make use of the cd event so far (autoxsh, avox) have to call out to the file system to see if they can run or not. And the way avox is architected (all the filesystem calls are handled by `Vox`, so avox has no knowledge of how virtual environments are actually stored), it can't efficiently make use of a validator function.

In general, class-based commands could not make use of a validator function.

xonsh/events.py
astronouth7303: Events is a plain object and not a mapping because of syntactic sugar, although there's no reason it can't be a mapping, too.

I'd personally rather keep it available through attribute access. The use of the Events collection as a collection I expect to be extremely minimal.

xonsh/events.py
astronouth7303: Even when using events internally, how strong is the guarantee that certain modules would load (and register their events) before others?

xonsh/built_ins.py
```diff
@@ -714,6 +715,7 @@ def load_builtins(execer=None, config=None, login=False, ctx=None):
     builtins.__xonsh_ensure_list_of_strs__ = ensure_list_of_strs
     builtins.__xonsh_list_of_strs_or_callables__ = list_of_strs_or_callables
     builtins.__xonsh_completers__ = xonsh.completers.init.default_completers()
+    builtins.__xonsh_events__ = Events()
```
astronouth7303: This is a pretty terrible place to initialize this if events are supposed to be used by the core.

I'm also not a fan of the length of `__xonsh_events__`.

santagada:
Can't we have something like completers were we can specify if we want to add our handler in the beginning or the end of the call chain? I also think we should respect the order they are added as this will be mostly repeatable and not in a random order.

astronouth7303:
You can. GTK has a full thing where there's default handler, and handlers can register before or after it. JavaScript/DOM events have phases and bubbling.

I'm still not sure what guarantees we have about the load order of different modules. So if something is order dependent, how do you specify when it happens? An index? A priority? Relative to other modules? Does that mean that a developer needs to be aware of other users of the event?

santagada:
I think for now being able to say if you want to be on the top or bottom of the call stack for the event is both simple and powerful enough. It will depend on the order we import things but that is mostly fine for the uses we have now.

A better approach is the aspect oriented kind of technique that exist on emacs that you can schedule yourself to run before, after or around an event, in the last case you control the calling of the event itself which can be useful for things like profilling or exception catching.

I also don't like using dunder everywhere on xonsh... but I think we can discuss something in another PR/issue/Gitter

astronouth7303:
It'd be pretty easy to do phases. it's a matter of having `command` or `precommand`/`postcommand`.

Having before/after/meh just breaks up an unordered group into three unordered groups. You can run all the befores, all the mehs, and then all the afters, but within each group I think it's still unordered.

I think the real question is, what should the event system be capable of? I'm going with the assumption that vast majority of usage will be notify-only (the handler doesn't pass anything back to the code that fired the event). 

Do use cases like preprocessors make sense? Is that even something you want to do? How much power do you want to give xontribs, and how much do you want to protect from bad code?

Keep in mind that as xonsh grows in popularity, the more power xontribs have, the more potential for bad xontribs to muck with users and the more support tickets will end up in your laps.

There's also the possibility that a well-designed syntax will allow some of these advanced event models to happen later. It's entirely feasible to add `.handler('after')` or `.handler(after=True)` to a later revision without breaking backwards compatibility.

xonsh/events.py
scopatz: Right, you use it on event handler, it is not a handler itself.  Another take is that functions should be verbs, and this is a noun.  `handle()` would also be a reasonable alternative to `register()`

xonsh/events.py
scopatz: > I can't imagine a use case where the handler couldn't fail fast instead of having an extra lambda (or worse, a full function).

Git, hg, and other version control hooks.  It takes much longer and much more logic to figure out if you are in a repo than it does to execute whatever action you want to perform.

However, broadly speaking, validators help separate out the concern of "should I do x" from actually performing x.  You end up with two, smaller, better defined, and more testable functions  rather than one function that does everything.

Having optional validators is an essential feature for me.  If you don't wish to add them, that is fine, but I will probably put in a PR to add them.  Also it is only a few lines of code to implement, so it is not that big of a deal.  If devs don't want to use them , that is fine. That is why they are optional.

xonsh/built_ins.py
scopatz: Yeah, events is not going to be able to live here forever.  Also, I agree and think that since this is a public interface, like aliases, it should be called `events`

xonsh/events.py
gforsyth: For what it's worth, the `prompt_toolkit` key binding stuff uses a `@handle` decorator for adding bindings to the key registry

xonsh/events.py
scopatz: I don't think we can get away with extremely minimal.  Users are going to want to do a few things, such as look up all event names and collections.  I also think that having dispatch methods down to the Event methods for a given event name would also be useful.

xonsh/built_ins.py
astronouth7303: I wasn't sure to do that because it felt like a common name, although I suppose it's not that common.

scopatz: Well, we could also call is `registar` or something, but I feel like events is more intuitive

xonsh/events.py
asmeurer: Even if ordering is not guaranteed, it seems useful to have it be consistent. And there may be use-cases where you really do need things to run in a certain order.

xonsh/events.py
```diff
@@ -0,0 +1,113 @@
...
+"""
+
+
+class Event(set):
+    """
+    A given event that handlers can register against.
+
+    Acts as a ``set`` for registered handlers.
+
+    Note that ordering is never guarenteed.
```
asmeurer: guaranteed

xonsh/events.py
astronouth7303: Well, it's not randomized, it's hash order. So each session will run them in the same order, and a given installation and configuration may always run them in the same order.

Xontribs will load in any order. The real question is, how does amalgamate order modules?

xonsh/events.py
asmeurer: The attribute access seems a bit too magic to me. I was quite confused by the example in the docstring, until I saw this code. What's wrong with just using a dictionary (or wrapper around a dictionary)? 

Also, `__getattr__` doing a `setattr` is really confusing. That means any mistyped attribute is automatically set as a handler.

xonsh/events.py
```diff
@@ -0,0 +1,113 @@
+"""
+Events for xonsh.
+
+In all likelihood, you want builtins.__xonsh_events__
+
+The best way to "declare" an event is something like::
+
+    __xonsh_events__.on_spam.__doc__ = "Comes with eggs"
```
asmeurer: What does this do? Does the docstring have special significance?

xonsh/events.py
```diff
@@ -0,0 +1,113 @@
...
+            if rv:
+                return rv
+
+    def untilFalse(self, *pargs, **kwargs):
+        """
+        Calls each handler until one returns something falsey.
+        """
+        for rv in self.calleach(*pargs, **kwargs):
+            if not rv:
+                return
```
asmeurer: Should it `return rv`?

xonsh/events.py
astronouth7303: It forces the event to exist, and it sets the docstring for it. (Except that in this version, nothing will use that docstring. I'm working on it.)

xonsh/events.py
asmeurer: How is this different from what calleach already does?

xonsh/events.py
astronouth7303: I just liked the less typing of attributes.

Under this design, `__getattr__` must save the result, so that the same instance is returned later. And where better to save it than `__dict__`? (@scopatz has already convinced me that mappings are more conventional for xonsh.)

And yes, mistyped events are a problem with automatic event names. I figured that a singleton event registrar would be better suited for xonsh than having declared events requiring import.

xonsh/events.py
astronouth7303: eh? For symmetry yes, but I had a hard time imagining when a falsey value would be of interest to the caller.

xonsh/events.py
asmeurer: The elements will be functions, right? If so, the hash order will be randomized across different Python sessions (the default `hash` is based on `id`).

xonsh/events.py
asmeurer: The whole thing would be much clearer as `defaultdict(lambda: Event())`. I don't see why you need syntactic sugar for this. Are these events going to be used a lot?

xonsh/events.py
astronouth7303: Yeah, I'm sure how repeatable CPython is with allocations and function addresses.

(Technically, it could be any callable, but it would be expected to be a function.)

xonsh/events.py
asmeurer: I'm unclear how this function is even to be used, honestly. I thought the functions returned arguments for the next function.

xonsh/events.py
astronouth7303: Well, every xontrib I've written so far would use at least one. I'm not familiar enough with the internals of xonsh to know how much it would be used internally, but the expectation would be to refactor xonsh itself to make use of it. (Which is why I posted the idea first, and then solicited comments on a prototype, before going in and doing the real thing.)

asmeurer: But aren't you just using the events object itself just once? I don't see the use-case where you are accessing the members of this object a lot (which would be reason behind saving typing with `__getattr__`).

xonsh/events.py
astronouth7303: Depends on the semantics of the particular event.

For example, a completer-like system might want to call each handler until one returned a value.

`cd` just wants to notify that the directory changed.

A preprocessor would want to chain the handlers, so each one got the collectively modified output.

`calleach()` utilizes generators to support all of these semantics (as well as subtle variations). `untilTrue()`, `untilFalse()`, and `loopback()` are just utility functions to support some expected usages.

xonsh/events.py
astronouth7303: Well, probably once in a given file.

xonsh/events.py
astronouth7303: It actually instructs `calleach()` to do the looping. `calleach()` just supports using the generator method `send()` to change the positional arguments.

xonsh/events.py
scopatz: My original thought was that loopback would be something that could be looked up (in Events?) if an alternative version was provided.  This loopback() would simply be the default that is good enough for most use cases if another implementation is not provided.

tests/test_events.py
```diff
@@ -0,0 +1,77 @@
...
+    @e.test.handler
+    def _(spam):
+        nonlocal called
+        called = spam
+
+    e.test("eggs")
+
+    assert called == "eggs"
+
+def test_untilTrue():
```
laerus: pep8 lowercase names seperated by underscores

tests/test_events.py
```diff
@@ -0,0 +1,77 @@
...
+    def second():
+        nonlocal called
+        called += 1
+        return True
+
+    e.test.untilTrue()
+
+    assert called == 1
+
+def test_untilFalse():
```
laerus: same here, python doesn't like camel :D

xonsh/events.py
astronouth7303: So I had two thoughts on the syntax of validators:

```python
def v(spam, egg):
    return spam == egg

@events.on_spamegg.handler(validator=v)
def se(spam, egg):
    print(spam, egg)
```

```python
@events.on_spamegg.handler
def se(spam, egg):
    print(spam, egg)

@se.validator
def v(spam, egg):
    return spam == egg
```

I prefer the second one myself. I think the implementation will be simpler and leads to a more declarative style.

scopatz: I prefer the second as well.  It is similar to properties.  Though I would probably most prefer if `Event.__call__` was an alias for `handle`, so we could simply do:

```python
@events.on_spamegg
def se(spam, egg):
    print(spam, egg)

@se.validator
def v(spam, egg):
    return spam == egg
```

astronouth7303: That's probably not a bad idea, forcing event callers to explicitly pick an event style instead of having a default one.

xonsh/events.py
scopatz: I imagine that when we first start up, events will be created with a known suite of builtin events with a known order.  Yes, users will be able alter this dynamically, but that is a matter of their configuration (which xontribs they load, what they do in their xonshrc files, etc.)

xonsh/events.py
scopatz: > Xontribs will load in any order. 

That is, whatever order the user decides.

> The real question is, how does amalgamate order modules?

Amalgamate orders based on import dependencies, and then lexicographically on the module name.

But all of this is getting pretty far afield.  Event needs to be an ordered container so that users can inspect the execution order and manipulate it if they want.  There are almost certainly going to be cases where one handler should only be called if a previous handler was called.  This is impossible to implement in an unordered container.

xonsh/events.py
astronouth7303: The way I figured is that each subsystem would have its own event registrations, so what order handlers are registered depends on module load order, which is either import order or what order amalgamate decides. I would call this deterministic but chaotic, and not easily reasoned about.

Which makes me think that for the sake of future sanity, any kind of event semantics that depend on ordering should be avoided. That is, events could be no responses, one response, or many responses. The primary use case for chained responses--preprocessors--is problematic anyway because each preprocessor has to be able to deal with arbitrary unknown syntax because it doesn't know what other preprocessors might be installed on the system.

scopatz: >  I would call this deterministic but chaotic, and not easily reasoned about.

Exactly. I feel that this needs to be easily reasoned about. It is a public interface.

> Which makes me think that for the sake of future sanity, any kind of event semantics that depend on ordering should be avoided.

Yes, I agree. These kinds of things should be avoided if at all possible.  However, avoided doesn't mean forbidden.  There may even be cases in the core that may be most easily implemented in this way.  I don't think we need to implement a full DAG dependency solver for this, but we also don't want to explicitly disallow any ad hoc dependencies either.

scopatz:
> I think the real question is, what should the event system be capable of? I'm going with the assumption that vast majority of usage will be notify-only (the handler doesn't pass anything back to the code that fired the event).

I don't think notify-only enough. Especially for the core, it might be extremely useful to pass back-and-forth information when building up an execution context from rc files, xontribs, etc.

> Do use cases like preprocessors make sense? Is that even something you want to do? How much power do you want to give xontribs, and how much do you want to protect from bad code?

The short answer is that since this is Python and everything is public an users can do anything, I would assume that there is no reason to limit the power we give to users and xontrib authors.  They already have power. Why make them jump through hoops like monkey patching the Event class at run time?  I'd rather users to feel empowered by the xonsh tools than feel like they have to work around them.

xonsh/events.py
PaulReiber: On the whole "validators" think I think that's not pythonic let's keep it
simpler.  Someone said (and they're right) event handlers can exit on line
one if they've got a good validation function that applies.  Same
difference.

Xonsh is about providing outer infrastructure for future python+shell
end-user needs.

Letting imported modules do something new and better in xonsh than they do
in python is an exciting idea.

Maybe just like modules have __all__ for all exported functions in python,
in xonsh they could have __events__ for exported event handlers?

Kind regards,


scopatz:
@astronouth7303 - were you going to do another cut in this PR or another PR? Just curious. I think I have said my piece for the moment :)

xonsh/events.py
astronouth7303: I argue that the event semantics should be designed to avoid and prevent ordering issues. Having an ordered event subsystem just exposes the chaotic ordering of the overall system. Having the event system unordered forces those inherent issues to the forefront. It is better to try to sidestep a whole class of issues.

If the order of handlers must be preserved (because there's a needed use case that must be ordered), and that order must be exposed to the user to introspect and manipulate, then each handler needs a reasonable name that a user can refer to. I do not like this idea because it allows for some very messy configuration problems that could require an in-depth understanding of xonsh and the xontribs involved.

I, for one, would much rather design it to prevent the issues than hope they never occur.

asmeurer: But how can the user even manipulate them if the events object is a set?

astronouth7303: The point is to do everything we can to avoid the need.

astronouth7303:
I've just been waiting for the discussion to die down before moving forward, seeing as GitHub has a habit of hiding conversations when you push new commits.

astronouth7303:
One thing that I didn't really well explain is the event model used in this PR. NOTE: This is just meant to document what was implemented in 3bb099b05b27208d755b66c3732497a599d86077, not what was proposed up to this point.

## Event Discovery
There's a few ways to access events. This particular PR uses a central event registry with event autocreation.

* Event handlers do not need to track down a module or object to register events
* Events do not have an initialization, so handlers may be registered before the "owning" module is imported
* There is a single, global namespace for all events. Any namespacing is done by convention.
* Events are not intrinsically tied to objects or instances
* Bugs caused by mis-typed events will be silent

## Event Execution

This PR accepts any callable as an event handler. It (controversially) does make any sort of guarantees about the order of calling. (I'm going to attempt to summarize this debate separately.)

The core handler calling logic is based on generators, allowing for many semantics to be layered on top of the above conditions. Several are included:

* Call each handler and return their values
* Call each handler until one of them returns something truthy
* Call each handler until one returns something falsey
* Modification chain: Each handler is called and returns a new version of the arguments that are passed in, and the new arguments are passed to the next handler (IMHO, fairly problematic even with ordering)

astronouth7303:
So @scopatz and I have gone back and forth about Ordered vs Unordered. I'm going to attempt to summarize the technical points of each. Not discussed is @santagada's discussion to use bins (start/end/meh) or phases (before, during, after).

Some assumptions, though: 

* The greater xonsh system (including xontribs) is chaotically ordered, so we don't really know what order handlers will be registered in.
* Certain handler calling semantics (modification chains, I'm looking at you) could require specific orderings in order to get desperate components to work together.

## Ordered
AFAIK, maintaining event order is typical of event systems. This makes the event subsystem itself easy to reason about, and also potentially gives the user the ability to sort out problems stemming from event misordering.

However, this also exposes the inherent chaotic ordering of the greater system, and by allowing order-dependent semantics, and could force the user to dig into the messy details of xonsh and xontribs (going as far as digging in to source) and attempt to make sense of what precisely events do and suss out what order they need to go in.

## Unordered
While unconventional, unordered handler calling has the potential to simultaneously hide some of the messy details while forcing developers to deal with the practical upshots of it. If done correctly, this would present the end user with a simpler semantics.

However, by forcing handlers to be called this way (currently, a different order every session), some semantics are disallowed and developers must be more creative or explicitly handle ordering problems. For example:

* Modification chains must be carefully constructed, or just disallowed
* There is no precedence, so return one (ie, until true) must have some idea how to deal with multiple returns, or face inconsistent results.

Note that I don't think unordered introduces any new problems, just brings them to the forefront at development time instead of letting them lurk for the user to discover.

xonsh/events.py
astronouth7303: Today I learned that you can't subscript in decorator lines, so `__getattr__()` must stay if decaorators are to be used.

asmeurer: I didn't know that. But you can still make the event registration explicit (`events.register_event` or something). The actual number of events should be few, since they will only be for the handful of things that are explicitly implemented in xonsh.

santagada:
The chaotic ordering of imports is just a implementation details for speed, and I don't think it is that chaotic, if I xontrib load x modules or source them, it will be done in source order.

I'm still strongly on the ordered events camp.

astronouth7303:
By "chaotic" I mean "not random, but too complex to reasonably reason about."

Modules are loaded linearly. The linearization of import order is not well-defined, and the constraints on it are spread out through literally every file.

Any bugs stemming from import order will be fairly difficult to track down, solve, and keep from regressing.

asmeurer:
That's true, but it's also generally considered an antipattern to perform some action at import time.

xonsh/events.py
astronouth7303: I would say that if explicit event initialization is wanted for xonsh, individual Event objects in various modules should be used and not a central registry.

This way, you can guarantee that event initialization happens before usage and the dependency between modules is made explicit.

astronouth7303:
In this case, "some action" is "register a callback" which is extremely common at import time. If there is an event semantic dependent on order, there is a potential for an order-related bug. Since the implicit ordering of event handlers would be tied to import order, you have a few options to fix this:

1. Juggle the import order so that the order is correct
2. Sprinkle little bits of ordering constraints across the code base, creating quiet little dependencies
3. Force the user to insert an incantation into their config file to fix it up.
4. Change the event semantics so that order no longer matters

I think 1 would be delicate, 2 would be unmaintainable, and 3 is an onerous burden.

This leaves 4. By making the event system itself unordered, it forces you to do this upfront before weird bugs happen.

asmeurer:
Do we have a concrete example where event ordering matters? I think that would ground the discussion a bit.

astronouth7303:
It's potentially a problem any time there's precedence or potential for a handler to "win".

* A completer-like system without merging, just taking the first set of results 
* Any kind of preprocessing (although that has its own delicacies outside of proper ordering)

astronouth7303:
I've finally implemented most of the changes discussed above, with two notable exceptions:

* `Event` is still unordered, because heated discussion
* The `Events` registry is not a mapping, because `@events['spam']` is not allowed, and there was weirdness in making it a mapping-with-attributes

I also added event docstrings that will actually show up in `help()`. This involves some dynamic type magic, but should not require anyone doing anything special outside of the one line that does it.

scopatz:
> Do we have a concrete example where event ordering matters?

I have envisioned using the event system for start up, where each config and rc file gets its own event handler.  If events are ordered, an rc file can be used to register further rc files. If events are unordered this is impossible without creating another event category and then triggering that.

Said another way, a handler should be able to modify the event that is currently executed in a way that will be picked up in that iteration. An event system without this notion of a stack is not unnecessarily  limited, in my opinion.

Thanks for continuing to push this @astronouth7303. I  would really like to see an event system enter the code base.  I think it is overdue.

astronouth7303:
Oh! Modifying during iteration wasn't ever mentioned.

I'll have to do some research on how the built-in types handle modification during iteration, in particular when it would modify the iteration. Off-hand, it's either undefined, poorly defined, or just not allowed.

That's not to say it couldn't be made to work, ordered or not. It just could involve some tracking.

scopatz:
Right, it probably involves a while loop rather than an iterator.

astronouth7303:
The other problem is that I think that particular requirement only applies to loading, which has its own wrinkle:

If an `on_load` handler is registered after the load phase happened (say, if the user loads a xontrib from the shell), then that handler should be called as soon as convenient. Standard event mechanics (pretty much universally) would say that it wouldn't be called until the next time the event is fired (aka, never for `on_load`).

So either `on_load` needs to be a special case (Zen of Python #8), or load handlers need to have their own mechanism.

scopatz:
Well, it seems like there would be both pre- and post-xontrib loading events.  If a xontrib wanted something to happen after they were loaded, always, they would register the handler in the post-xontrib load event. That way it wouldn't matter when they loaded the contrib - from the config file or from the shell.

astronouth7303:
But no matter when you fired the post-xontrib load event, there's the possibility of a handler being added later.

Conventionally, an event is fired once, all its handlers are processed, and then its over. No more handlers are called until the event is fired again.

Futures behave more like the load event. The value is returned once, and all the handlers get called. Any handlers that get registered later are called immediately.

xonsh/events.py
```diff
@@ -0,0 +1,101 @@
...
+class Events:
+    """
+    Container for all events in a system.
+
+    Meant to be a singleton, but doesn't enforce that itself.
+
+    Each event is just an attribute. They're created dynamically on first use.
+    """
+
+    def __getattr__(self, name):
```
laerus: i think this is very confusing and implicit, can't we just have a `register` or similar?

astronouth7303: Ok, here's the thing:

If you have a central event registry, you need to have events be usable before the "owning" module (because such a concept doesn't really exist) is loaded, because otherwise you have an implicit dependency. If the event system is used in the core, what if the handling module is loaded before the "registering" module?

The alternative is to have event objects instantiated and scattered about the code base, such that:

```
from xonsh.exec import on_precommand

@on_precommand
def myhandler(cmd):
    ...
```

I'm not saying either one is wrong, but I will say that anything that causes implicit, untracable dependencies is bad.

(Note that by the time the _user_ sees the event registry, the list of available events will be known, except for xontribs that are loaded by the user later.)

astronouth7303: (Oh, and the thing to be careful of is circular imports. If two modules depend on events from each other, you're SOL. I don't think you could lazy load your way out of it, because you're accessing and modifying the handler lists at import time, by design.)

laerus: OTOH this requires that events exist before the handlers that register to them.

laerus:
Overall i think a pub-sub pattern would be a better choice, a central event registry is a lot more abstract.
Now events are coupled with handlers, and what about the same handler handling several events, do you pile up decorators?

Maybe the registry instance can do all the work of passing events to handlers, registering handlers or even firing events, it's the API anyway. Events don't have to know about handlers and vice versa.

Also lets not totally *reinvent the wheel*, there are [many projects](http://stackoverflow.com/questions/1092531/event-system-in-python) out there that have tackled this problem, maybe they have some good ideas.


(lets do this with asyncio :laughing:)

xonsh/events.py
```diff
@@ -0,0 +1,101 @@
...
+    def until_false(self, *pargs, **kwargs):
+        """
+        Calls each handler until one returns something falsey.
+        """
+        for rv in self.calleach(*pargs, **kwargs):
+            if not rv:
+                return rv
+
+
+class Events:
```
laerus: A better name would be nice, `Event` `Events` is prone to typos and harder to read, maybe `Events` can be renamed to `EventDispatcher` or `XonshRegistry` or `Reactor` :laughing:.

scopatz: I think Events is correct here, since it does hold and manage many Event instances.  Also, I'd rather we not get too caught up on bike shedding the names while we still have core architecture to work out :).

xonsh/events.py
scopatz: For the record also, I anticipate that core will need to manually register its initial handlers, and not use the decorator syntax, in order to get around this circular dependency issue.  I don't think this is a big deal because it isn't user facing, startup is strange, and it will still simplify startup relative to what it is now.

xonsh/events.py
```diff
@@ -0,0 +1,101 @@
...
+
+In all likelihood, you want builtins.__xonsh_events__
+
+The best way to "declare" an event is something like::
+
+    __xonsh_events__.on_spam.doc("Comes with eggs")
+"""
+
+
+class Event(set):
```
laerus: better to inherit from `collections.abc.Set` 
although i'm pro ordered collection aswell

scopatz:
> Overall i think a pub-sub pattern would be a better choice, a central event registry is a lot more abstract. Now events are coupled with handlers, and what about the same handler handling several events, do you pile up decorators?

Yes, you'd pile them up.  As per pub-sub, I am indifferent. It seems like that could be added later if desired.

> Also lets not totally reinvent the wheel, there are many projects out there that have tackled this problem, maybe they have some good ideas.

Any particular suggestions? I think the main issue here is retaining our "Python as only a dependency" status. We vend PLY because it is central to xonsh-the-language. We could end up vending an event system if it were central to xonsh-the-application.  Most other event systems (zmq, celery, etc) come with a lot of dependencies of their own.  This is why I myself never looked at these in detail for xonsh. Of course, I could be missing something small and obvious ;)

> (lets do this with asyncio :laughing:)

Don't you mean "this wido  asyncth lets io"? :)

laerus:
>  As per pub-sub, I am indifferent. It seems like that could be added later if desired.

Why to couple events and handlers on such a low level? it's harder to test and it makes `Event` harder to subclass. And decoupling seems easier on this stage than later. 

> Any particular suggestions?

i was thinking more of small PyPI projects that can be integrated in xonsh
[this url](http://stackoverflow.com/questions/1092531/event-system-in-python) that i mentioned before has some examples.

this will go to the core right? do we want a pluggable event dispatch system that can be even be turned off/on?

a decorator is nice an all but this `__call__` trick on the `Event` class is a little too implicit imo, we can have a decorator that does the same thing with a global decorator function.

xonsh/events.py
```diff
@@ -0,0 +1,101 @@
...
+    Acts as a ``set`` for registered handlers.
+
+    Note that ordering is never guaranteed.
+    """
+
+    @classmethod
+    def doc(cls, text):
+        cls.__doc__ = text
+
+    def __call__(self, func):
```
laerus: this `__call__` trick is a little too implicit imo, we can have a decorator that does the same thing with just a module level decorator function.

scopatz: I don't mind this.  This is just how Python works.

scopatz:
> Why to couple events and handlers on such a low level? it's harder to test and it makes Event harder to subclass. And decoupling seems easier on this stage than later.

I think that having something, if it isn't what everybody wants (myself included), is better than not having it at all.  

> this will go to the core right? 

Yeah it would have to go into the core.

> do we want a pluggable event dispatch system that can even be turned off/on?

I don't think that it should be able to be turned off.

Sorry I had missed that link before

scopatz:
Axel seems the most robust, except that it uses threads.  [This simple implementation](http://stackoverflow.com/questions/1092531/event-system-in-python/1094423#1094423) is closest to what I would like to have from the model side but with @astronouth7303's additional syntactic sugar.

scopatz:
Also, probably the main thing that is keeping me from merging this is that Event is unordered. I think everything else could be cleaned up as we go.

scopatz:
Also, @laerus, I think we could have another proposal that does pub-sub or we could add pub-sub stuff very soon after this one is merged.  My main concern right now is that we are in danger of never merging anything because of trying to meet all needs simultaneously.  I think we could unravel some of these or have competing proposals.  Personally, I would like for *something* to get in relatively soon, since it is blocking a few development paths.

xonsh/events.py
astronouth7303: The way I figure, the decorator will be attached to the thing doing the storage. That is, any time you could do it by hand, you could use the decorator.

@laerus As it is, yes, there's no way to catch typos. Although it occurs to me that if there's a definitive startup phase when the core is loaded before xontribs execute, that would be a good time to check for typos. Or make it part of the test suite.

xonsh/events.py
astronouth7303:
```python
@events.vox_on_activate
def myhandler(venv):
    ...
```

I like the readability of this.

astronouth7303:
@laerus As it stands, this is pretty decoupled. Handlers and owners are only coupled by the name and signature of the event. Just because there isn't a string to be seen doesn't mean it's not decoupled.

@scopatz I still hold that having ordered handlers just exposes calling order complexity of the underlying system. Unordered means that you're forced to solve the problems without having to deal with the problems.

scopatz:
> @scopatz I still hold that having ordered handlers just exposes calling order complexity of the underlying system. Unordered means that you're forced to solve the problems without having to deal with the problems.

Well, let's return the the problem of configuration and rc loading, which I think is a real issue. In this case, I think it is pretty clear that it may be necessary for certain files (which each have their own handler) to be loaded in a specific sequence.  For example, the system configuration should always be loaded before user configuration. 

If events are unordered, then this forces a model where several different event categories are present, and each event category is triggered in its own sequence.  The code in xonsh to fire these events might look something like:

```python
events.syswide_config.fire()
events.user_config.fire()
```

Each of these would presumably only have one or two handlers.  Now, what if we wanted to make a distinction between default user configuration and user specified configuration.  The core would then have to have something like:

```python
events.syswide_config.fire()
events.default_user_config.fire()
events.user_loaded_config.fire()
```

Again each of these having only one or two handlers. Now what if the users want to force an ordering because some of their configs need to be loaded in a sensible order?  The core then needs to know about those and provided more event loading phases *ad infinitum*.  Effectively this makes the core be able to predict all available event categories and unroll a loop explicitly. Furthermore, each event basically only has one handler, which makes it seem like events should not be a collection of handlers, but the handlers themselves (in this particular case).

I think it i s much more robust and clean for the core to have a single config event, with many handlers, that may be modified while it is being triggered.  The code the core would need to implement is then just:

```python
events.config.fire()
```

The event class is then responsible for the iteration, not the library author.  If the event class can make the above single line version work without ordering and with run-time modification, I'd love to see it!  I think that would be effectively a dependency DAG +/- a SAT solver.  (cough cough dask, first post). But I am not sure we need that level of complexity here, yet.

Alternatively, if the purpose of a set is uniqueness and not ordering, I am all in favor.  I think uniqueness is a desirable, if not strictly requires property. But that can be accomplished via an OrderedSet or UniqueSequence or something.

Hopefully this helps explain what I see as one of the major end game use cases for events.

astronouth7303:
We have a collection of known obvious use cases:

* `on_precommand`
* `on_postcommand`
* `vox_on_activate`
* `vox_on_deactivate`
* `on_chdir`

The semantics of these is pretty obvious, and they all behave similarly.

However, loading breaks this in two ways:

1. A given handler must only ever be called once
2. A given handler must be guarenteed to be called, after certain preconditions are made

Tring to shoehorn `on_load` in to the standard event semantics (or twisting the event semantics to support it) will only end in pain.

(I'm not against making `on_load` accessible through the `events` module or using a similar interface as events. But I don't think it's truly an event, in the same way as the others above.)

scopatz:
I guess my response to that is that I'd rather not have two event systems.  I think that would be even less intuitive to users.

astronouth7303:
Me either, but I'd rather have two simple event systems than one complicated one. Loading just doesn't behave like an event; it behaves like a future/promise.

Maybe there's a way to have some events be "one-off" events? A better understanding of initialization and #1550 would help. Would this be a time to introduce some form of async inversion of control type thing? Ideas for another thread.

scopatz:
@astronouth7303, all maybe we could chat about this over a google hangout or some such? I think we have gotten as far as we can go with this without some one changing their mind or some one else implementing this.

astronouth7303:
Ok, @scopatz and I had a live discussion.

1. Unordered is the way to go for notify/scatter-gather events (pretty much the same thing).
2. Load events have the same interface as notify events, but have completely different semantics, so trying to force them into the same code path is a nice route to insanity.

So the solution is that I'm going to implement the ability to transmogrify events. That is, you'll be able convert an event from one backing class to another. (There's going to be restrictions on this. It's not arbitrary.) I'll also add a prototype `LoadEvent` to prove that this works, but it won't implement any of the semantics or discussion that will come out of #1550 (although we discussed some of the shape of that as well).

I also need to implement some form of exception catching. For notify/scatter-gather, it will catch, log, warn, and move on. (With the note that there may be a way for the caller to get this information in some form in the future, but we couldn't come up with a concrete use case for notify events.)

scopatz:
Thanks for writing up the results of our discussion @astronouth7303. I look forward to the next review!

xonsh/built_ins.py
astronouth7303: I'm still not sure where this should happen.

If there's a specific spot we can add it to built-ins, could someone find it?

Another idea is that I could implement [`sh`](https://amoffat.github.io/sh/)-like magic where events get imported like:

```python
from xonsh.events import on_precommand
```

But my gut says that's an unnecessary amount of magic.

scopatz: For now, maybe make the instance at module level scope of events.py. That way the core can use it.  We can then attach it to builtins whenever we want.  Maybe in main.py?

astronouth7303: Probably the simplest way to do it. If all the core is importing it, then might as well keep it with the rest of the builtins.

laerus:
for loading events why not just have something like `loading_start` `loading_end`, also maybe events should pass the sender.

will we be able to subclass `Event` and make custom events? 

maybe events can have a context dict,  handlers can pass in w/e info they like so others can see and mutate it.

scopatz:
> will we be able to subclass Event and make custom events?

Yes this is basically the idea. This allow the user facing interface to Events to be the same, but then different Event categories can call their handlers in different ways, depending on the needs and the use case.

astronouth7303:
> for loading events why not just have something like loading_start loading_end, also maybe events should pass the sender.

The exact semantics of how loading will happen is getting punted to #1550, so we can just get events in general out there and start using them.

> will we be able to subclass Event and make custom events?

Technically, yes. In practice, I don't really see--nor encourage--this happening. We don't need 10 event species running around. We did the two because they're so fundamentally different that they couldn't be the same thing.

At some point, there might be a testing rig so that certain classes of errors can be sussed out.

> maybe events can have a context dict, handlers can pass in w/e info they like so others can see and mutate it.

?? Callers can pass in arguments, and handlers can return values (if the caller cares). I'm not sure what use case you're looking at here.

@laerus If you're planning in your head to bypass some of this discussion for _your_ events, keep in mind:

* Events are unordered to force developers to deal with order-related problems up-front. Changing this for one event won't make them go away, it'll just punt the problems to future you.
* I dropped mutation chains because the given use case (command preprocessing) is extremely problematic in implementation. I'm not against putting it back in, but any usage must be structured so that the output of any given mutator can be fed into input of any other mutator.

laerus:
So are we supposed to subclass `Event` or not?

As this is going to the core my main concerns are if this is scalable for feature additions and refactoring, if it is a little too much magic with `__getattr__` and `__call__` tricks (yes Python but also the Zen), maybe we leave most of the work for devs/users for stuff that should be already builtin.??? 

If this goes in in a provisional style that means we can break the API till it's clear and robust then lets merge this now and start working on it.

scopatz:
> If this goes in in a provisional style that means we can break the API till it's clear and robust then lets merge this now and start working on it.

More or less that is the idea.  @astronouth7303 is going to modify a couple of things to make it suitable for loading, and then we'll roll with that and see what works and what doesn't.

astronouth7303:
Ok, refactored and implemented under the current proposal. Still needs properly-formatted docstrings and probably a few pages in the docs. (I'm thinking at least two.)

astronouth7303:
And apparently horribly and mysteriously broke the build? Looking at my fork, something between when i merged in the updates and that last commit made pytest explode.

xonsh/events.py
```diff
@@ -0,0 +1,169 @@
+"""
+Events for xonsh.
+
+In all likelihood, you want builtins.__xonsh_events__
+
+The best way to "declare" an event is something like::
+
+    __xonsh_events__.on_spam.doc("Comes with eggs")
+"""
+import abc
```
scopatz: This is what is causing things to explode, because other parts of the code base `import collections.abc as abc`, which is probably the wrong thing to do. Alagamate then starts to treat them like the same import.

scopatz: Maybe we want to do `import collections.abc as cabc` or something.  Sorry about this @astronouth7303!

astronouth7303:
That's fixed.

xonsh/events.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def fire(self, *pargs, **kwargs):
+        raise NotImplementedError("See #1550")
+
+
+class EventManager:
+    """
+    Container for all events in a system.
+
+    Meant to be a singleton, but doesn't enforce that itself.
```
laerus: Maybe we should enforce this?

astronouth7303: It's good for testing to not use it as a singleton.

laerus: can the 'private' attribute that holds the reference to the singleton be used in tests?

astronouth7303: Having actual new instances helps ensure clean environments for tests.

It's the `Event` instances do all the real work anyway, so having multiple `EventManagers` just makes access weird. It shouldn't cause anything to actually break.

astronouth7303:
BTW, if anyone wants to help with docs, I would love it.

scopatz:
Is this ready for another review?

astronouth7303:
Code should be in the polishing phase. I think docstrings are ok, but I haven't really learned numpy format.

I think the biggest thing that needs to happen in this PR (assuming actually implementing events is a future PR) is the documentation pages. I've kinda started them, but the writing is being difficult.

xonsh/built_ins.py
```diff
@@ -714,6 +715,7 @@ def load_builtins(execer=None, config=None, login=False, ctx=None):
     builtins.__xonsh_ensure_list_of_strs__ = ensure_list_of_strs
     builtins.__xonsh_list_of_strs_or_callables__ = list_of_strs_or_callables
     builtins.__xonsh_completers__ = xonsh.completers.init.default_completers()
+    builtins.__xonsh_events__ = events
```
scopatz: Shouldn't this be public if we expect people to use it?  That is, `builtins.events = events`

xonsh/events.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        Fires each event, returning a non-unique iterable of the results.
+        """
+        vals = []
+        for handler in self._filterhandlers(self._handlers, *pargs, **kwargs):
+            try:
+                rv = handler(*pargs, **kwargs)
+            except Exception:
+                print("Exception raised in event handler; ignored.", file=sys.stderr)
+                traceback.print_exc()
```
scopatz: This should probably use xonsh.tools.print_traceback() https://github.com/xonsh/xonsh/blob/master/xonsh/tools.py#L590 so that it respects the user configuration settings.

xonsh/events.py
```diff
@@ -0,0 +1,169 @@
...
+        for handler in oldevent:
+            newevent.add(handler)
+
+    def __getattr__(self, name):
+        e = type(name, (Event,), {'__doc__': None})()
+        setattr(self, name, e)
+        return e
+
+
+events = EventManager()
```
scopatz: I think that the should wrapped as a lazyobject or have a comment about why it isn't

scopatz:
Just a couple of minor issues here.  I think that the docs could be in a separate PR if needed.

scopatz:
That is, the user/tutorial docs.  The docstrings are sufficient for ne,

xonsh/built_ins.py
astronouth7303: There's a commit missing. I could have sworn I changed the name and initialized it in the module (so that core could load it).

astronouth7303: No, i'ts here. Not sure how you commented on an old version, though.

xonsh/events.py
astronouth7303: Initialization is very nearly the cheapest it possibly can and it's expected to be used at load time. I'm pretty sure in this case using lazyobject would be more expensive because of overhead.

Comment will be in the next commit, though.

scopatz:
OK here goes nothing! Thanks for the amazing effort here @astronouth7303!

Issue #1506, opened 2016-07-28
xonfig AttributeError PLY version


laerus:
```
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/home/laerus/projects/xonsh/xonsh/proc.py", line 349, in wrapped_simple_command
    r = f(args, i)
  File "/home/laerus/projects/xonsh/xonsh/aliases.py", line 449, in xonfig
    return xonfig_main(args)
  File "/home/laerus/projects/xonsh/xonsh/xonfig.py", line 481, in xonfig_main
    return _XONFIG_MAIN_ACTIONS[ns.action](ns)
  File "/home/laerus/projects/xonsh/xonsh/xonfig.py", line 344, in _info
    ('PLY', ply.__version__),
AttributeError: 'module' object has no attribute '__version__'
```
anyone can reproduce?

gforsyth:
I believe that @aliceriot had the same issue yesterday while we were trying to figure out some `source-zsh` issues in #1134 

Reinstalling PLY seemed to do the trick, but it would also be nice to understand why this just started happening...

gforsyth:
@laerus -- do you have PLY installed externally?  Or are you using the the internal xonsh one?

laerus:
i have installed it through my distro package, externally that is. But i haven't upgrade it lately.

gforsyth:
Huh.  Well, I figured it out, but it's weird.
If I install PLY 3.7 externally (which is the one bundled with `xonsh`), then `ply.__version__` is (unsurprisingly) 3.7

3.8, though, seems to have dropped the `__version__` attribute

gforsyth:
No, it's even weirder... we do bundle 3.8, but the `__init__` file in our version has a `__version__` tag that says 3.7

gforsyth:
Ahhh, because that's what is on @dabeaz's PLY repo.  So PLY 3.8 reports itself as 3.7 if you install from github, but apparently not from distro package managers or conda

gforsyth:
Ok, here's what I've got:
PLY 3.5 doesn't report a version but is only compatible with python <= 3.4
PLY 3.6 reports
PLY 3.7 reports
PLY 3.8 doesn't

since PLY 3.5 " had some critical bugs and has been withdrawn. " I think we can safely assume for now that if there is no `__version__` attribute then someone is running 3.8.

laerus:
[...]

PR #1507, opened 2016-07-28
fix reported PLY version in bundled PLY

gforsyth:
As discussed in #1506 , PLY didn't bump its version number in `__init__.py` when they moved from 3.7 -> 3.8


scopatz:
Thanks @gforsyth!

PR #1510, opened 2016-07-29
fix xon.sh to make it take the current python and not a generic python3

santagada:
This is to make xon.sh work on os x homebrew or anywhere were the python you install is not the global python3

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -181,7 +181,7 @@ class xinstall(install):
             import traceback
             traceback.print_exc()
             print('Installing Jupyter hook failed.')
-        install.run(self)
+        super().run()
         if dirty:
             restore_version()
 
@@ -193,7 +193,7 @@ class xsdist(sdist):
         build_tables()
         amalgamate_source()
         dirty = dirty_version()
-        sdist.make_release_tree(self, basedir, files)
+        super().make_release_tree(basedir, files)
         if dirty:
             restore_version()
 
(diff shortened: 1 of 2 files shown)


scopatz:
Hi @santagada, thanks a million for putting this in! Can you please add a news entry

santagada:
I can, can anyone verify if there is a less crazy way to do this

My main questions:
1) why develop command doesn't call into install_scripts to install_scripts?
2) the windows fix is on write_script which is only called by setuptools in specific cases so does setup.py develop work on windows?
3) I just open the file without encoding, should we default to utf-8 or something else?

ps: distutils/setuptools are kind of a big mess.

laerus:
1) develop doesn't copy any files it creates an egg that is linked to the source

3) in most systems the default encoding is utf8 so its kinda safe.

indeed python is still in limbo on this matter

setup.py
```diff
@@ -213,11 +213,30 @@ def write_script(self, script_name, contents, mode="t", *ignored):
             contents = contents.replace(shebang, quoted_shebang)
         super().write_script(script_name, contents, mode, *ignored)
 
+
+class install_scripts_rewrite(install_scripts):
+    """Rewrite python3 to the current python executable"""
+    def run(self):
+        super().run()
+        if os.name == 'posix' and not self.dry_run:
```
scopatz: xon.sh is installed on non-posix systems, so the posix check should be removed.

setup.py
```diff
@@ -213,11 +213,30 @@ def write_script(self, script_name, contents, mode="t", *ignored):
...
+            for file in self.get_outputs():
+                if file.endswith('xon.sh'):
+                    bs_cmd = self.get_finalized_command('build_scripts')
+                    exec_param = getattr(bs_cmd, 'executable', None)
+
+                    with open(file, 'r') as f:
+                        content = f.read()
+
+                    processed = content.replace(' python3 ', ' {} '.format(exec_param))
```
scopatz: I think the format should be `' "{}" '.fomat(exec_param)`, to allow for bad actor cases such as when the user has a space in the abspath to the Python executable.

scopatz:
Thanks for putting this in @santagada!  Sadly, I think this is the minimal crazy way. Some inline comments. And also, could you please put in a news entry? Thanks!

santagada:
Great, thanks for the comments, I will try to document stuff and add a news entry.

santagada:
done

scopatz:
Thanks a ton!

Issue #1156, opened 2016-06-07
Starting the X server from .xonshrc


b52:
So, I switched to xonsh as my default shell. No I tried to install a way to `startx` without a login manager. Previously, I simply used `~/.zprofile` to do this, but it seems there is no such thing in xonsh. But I found the environment variable `$XONSH_LOGIN`, which should come close. So, I installed a similar script at the end of my `~/.xonshrc` to do the same:

``` python
if $XONSH_LOGIN and 'DISPLAY' not in __xonsh_env__ and $XDG_VTNR == '1':
    startx -- -nolisten tcp vt1
```

However, `$XONSH_LOGIN` is `False` during the invocation of the script. If I print it in the spawned login shell it's `True` however.

**Edit:** Furthermore, if I start terms in X and look at $XONSH_LOGIN it's still `True`, which shouldn't be the case in my opinion. I assume that the login shell is the first one spawned by me, e.g., the one I use to `startx`.

**Edit2:** Good [explanation](http://unix.stackexchange.com/questions/38175/difference-between-login-shell-and-non-login-shell) of the distinction between login and non-login shell.

adqm:
Thanks for reporting, @b52.  You are right that it looks like right now xonsh is treating `$XONSH_LOGIN` and `$XONSH_INTERACTIVE` as the same thing, which is not the right thing to do.  That logic could use some work.

adqm:
In the meantime, one way you could hack this would be to set a dummy variable in the environment that you could later check in the other shells (since all the shells after you have started x should inherit from the first shell you use to log in).  It's not pretty, but it should get the job done (tested a little bit on my machine):

```
if 'STARTED_X' in __xonsh_env__:
    # already started x; do nothing
    pass
elif any(i in __xonsh_env__ for i in {'SSH_TTY', 'SSH_CLIENT', 'SSH_CONNECTION'}):
    # ssh'ed in; don't start x
    pass
else:
    $STARTED_X = True
    startx -- -nolisten tcp vt1
```

Hope this helps!

astronouth7303:
Hm. Testing out some stuff mentioned in the StackExchange link above, and I don't get the leading `-` in `sys.argv` (or in procfs `cmdline`). However, I do see it when running zsh. So I'm wondering if CPython cleans this up, or if it's something weird about executing scripts vs binaries.

astronouth7303:
Ohhhh, from the [sys module reference](https://docs.python.org/3/library/sys.html):

[...]

PR #1512, opened 2016-07-29
Bring hashbang parsing in line with bash, zsh, sh, and Linux

astronouth7303:
In doing some testing for #1156, I discovered that the way xonsh parses shbangs is different than every other example I had installed (sh, bash, zsh, and Linux).

To test this, I used a small C program:

``` c
#include <stdio.h>

int main (int argc, char *argv[])
{
    for(int i = 0; i < argc; i++) {
        printf("%i\t%s\n", i, argv[i]);
    }
    getchar();
}
```

And a script:

```
#!/tmp/shbang hello world
```

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -230,7 +230,7 @@ def get_script_subproc_command(fname, args):
     else:
         interp = m.group(1).strip()
         if len(interp) > 0:
-            interp = shlex.split(interp)
+            interp = interp.split(' ', 1)
         else:
             interp = ['xonsh']
     if ON_WINDOWS:
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -8,7 +8,6 @@ import os
 import re
 import sys
 import time
-import shlex
 import signal
 import atexit
 import inspect


scopatz:
Thanks for putting this in @astronouth7303!  For the record though, I think restricting the shebang to two tokens is a mistake.  I can be convinced away from this, of course, but I don't think this should be merged just for consistency with other shells.

astronouth7303:
https://en.wikipedia.org/wiki/Shebang_(Unix)#Syntax

http://www.in-ulm.de/~mascheck/various/shebang/#splitting

I don't think this is something xonsh has jurisdiction on, since parsing it differently than the system can cause unexpected behavior or incompatibility in the edge cases.

I would call it debatable if xonsh should attempt parsing at all, or just hand it off to the system and let it do its thing (at least on POSIX-like systems).

gforsyth:
From wikipedia: 
>    (for reasons of portability, it should not contain any whitespace)

From http://stackoverflow.com/a/4304187
>    The shebang line has never been specified as part of POSIX, SUS, LSB or any other specification. AFAIK, it hasn't even been properly documented.

There doesn't seem to be a consensus on this.  

One thing that does seem to be relatively consistent (although I've never seen it in the wild) is that a space after the shebang should be supported

gforsyth:
Hmm, the table at the bottom of http://www.in-ulm.de/%7Emascheck/various/shebang/ seems to say that OSX splits the arguments...
And FreeBSD (and we do have at least one FreeBSD user)

krytarowski:
There is one NetBSD users.

gforsyth:
Hey @astronouth7303 -- I don't know if we've figured much out on this one.  Any thoughts on what we should do here?

scopatz:
I think we should leave the current behaviour as is.

astronouth7303:
It's all pretty arbitrary. Since there's no standardization, the only safe thing to do is to remove shbang parsing altogether. (I'm assuming it's in there for the sake of Windows?)

scopatz:
Hey @astronouth7303 - Do you think this is worth keeping open, given the lack-of-standardization discussion above?

astronouth7303:
I suppose not. I'm still wondering why xonsh does any shbang parsing at all, though.

scopatz:
It is so that we can execute things like bash scripts, Python scripts, and xonsh scripts. This is something that is done by the shell, and not the kernel, as I recall.

astronouth7303:
No, it's done by the kernel on Posix. It's why Python's calls work, and even C programs can do it. Nobody implements shbang parsing, except maybe on windows for unixy compatibility.

scopatz:
Yeah it is definitely needed by Windows.  If it becomes a problem, we can make it platform dependent, but I would rather not.

astronouth7303:
See my argument above. Basically: I think conformity to the underlying platform is important here.

PR #1514, opened 2016-07-30
Fix #1513

### LINKED ISSUE #1513: vox deactivate no longer resets my path
`which python3` returns the same thing before and after `vox deactivate` now.  `vox activate` correctly changes it, but `vox deactivate` no longer puts it back.  I am on the newest label of xonsh.

astronouth7303:
Forgot to copy a mutated value on save. Added a test to prevent regression.


scopatz:
Thanks!

PR #1516, opened 2016-07-30
Fix #1515

### LINKED ISSUE #1515: Invalid sys.argv when running a xonsh script
```sh
$ cat test.py
#!/usr/bin/env python3
import sys
print(sys.argv)
$ ./test.py --hello world
['./test.py', '--hello', 'world']
$ python3 test.py --hello world
['test.py', '--hello', 'world']
$ sed -i s/python3/xonsh/ test.py
$ ./test.py --hello world
['--hello', 'world']
$ xonsh test.py --hello world
['--hello', 'world']
$ execx($(cat test.py))
['/usr/local/bin/xonsh']
$
```

As you can see, program name is not passed into `argv[0]`. This behaviour differs from *every other* existing programming language, including Python. Thus, none of the option parsing libraries work with default settings, since they try to use `sys.argv[1:]`.
Hoping this is a bug that's going to be fixed, I'm using the following workaround for now:

```py
if semver.match(re.search(r"([\w.-]+)\W*$", $(xonsh --version)).group(1), "<=0.4.4"):
    args = sys.argv
else:
    args = sys.argv[1:]
```

By the way, why does `xonsh --version` print a `repr` of a tuple?

SirNickolas:


xonsh/main.py
```diff
@@ -224,8 +224,8 @@ def main(argv=None):
         # run a script contained in a file
         path = os.path.abspath(os.path.expanduser(args.file))
         if os.path.isfile(path):
-            sys.argv = args.args
-            env['ARGS'] = [args.file] + args.args
+            sys.argv = [args.file] + args.args
+            env['ARGS'] = sys.argv[:] # $ARGS is not sys.argv
```
laerus: why not? i think it's safe to have them both point to the same list.

xonsh/main.py
```diff
@@ -159,7 +159,7 @@ def premain(argv=None):
         parser.print_help()
         parser.exit()
     if args.version:
-        version = '/'.join(('xonsh', __version__)),
+        version = '/'.join(('xonsh', __version__))
```
laerus: nice catch!

laerus:
test failing to flake8
`/home/travis/build/xonsh/xonsh/xonsh/main.py:228:38: E261 at least two spaces before inline comment`
meh..

laerus:
@SirNickolas thank you for this!

xonsh/main.py
SirNickolas: Because there is no other `$ENV_VARIABLE` that is alias to something. Furthermore, one may modify `$ARGS`, reckoning to pass it to a subprocess, but not expecting `sys.argv` to be modified.

xonsh/main.py
SirNickolas: Liked that too!

xonsh/main.py
```diff
@@ -224,8 +224,8 @@ def main(argv=None):
         # run a script contained in a file
         path = os.path.abspath(os.path.expanduser(args.file))
         if os.path.isfile(path):
-            sys.argv = args.args
-            env['ARGS'] = [args.file] + args.args
+            sys.argv = [args.file] + args.args
+            env['ARGS'] = sys.argv[:]  # $ARGS is not sys.argv
```
SirNickolas: There is no clear relationship between `$ARGS` environmental variable and `sys.argv` global list (except that both hold the same elements at the startup). I doubt it's evident that any changes made to one of them should apply to the other as well. Explicit is better than implicit.

laerus: in my point of view $ARGS envvar is  an alias/shortcut for sys.argv so you can get the arguments in xonsh scripts without having to `import sys`. The explicit thing here imo is that both of these names point to the same object. One bug that may come up involves `argparser.ArgumentParser.parse_args()` which reads the arguments from `sys.stdin`, if a xonsh script mutates the `$ARGS` envvar the changes wont be visible to `parse_args()`

SirNickolas:
If `$ARGS is sys.argv`, a subtle bug turns up, since, strictly speaking, `$ARGS` is not an alias. Someone can easily reassign it without even noting. For example:

```py
$ARGS = $ARG0 + [...] + $ARGS[1:]
# This is not equivalent to $ARGS[1:1] = [...]!
```

Or more complicated one:

```py
def preprocess_args():
    switch = False
    unhandled = [$ARG0]
    for arg in $ARGS[1:]:
        if arg == "--switch":
            switch = True
        elif arg == "--no-switch":
            switch = False
        else:
            unhandled.append(arg)

    $ARGS = unhandled
    return { "switch": switch }

...

# Innocent code:
switches = preprocess_args()
args = parser.parse_args()
```

The very fact that `$ARGS += something` works but `$ARGS = $ARGS + something` doesn't is extremely counter-intuitive.

xonsh/main.py
SirNickolas: If `$ARGS is sys.argv`, a subtle bug turns up, since, strictly speaking, `$ARGS` is not an alias. Someone can easily reassign it without even noting. For example:

```py
$ARGS = $ARG0 + [...] + $ARGS[1:]
# This is not equivalent to $ARGS[1:1] = [...]!
```

Or more complicated one:

```py
def preprocess_args():
    switch = False
    unhandled = [$ARG0]
    for arg in $ARGS[1:]:
        if arg == "--switch":
            switch = True
        elif arg == "--no-switch":
            switch = False
        else:
            unhandled.append(arg)

    $ARGS = unhandled
    return { "switch": switch }

...

# Innocent code:
switches = preprocess_args()
args = parser.parse_args()
```

The very fact that `$ARGS += something` works but `$ARGS = $ARGS + something` doesn't is extremely counter-intuitive.

laerus: you're right, if you reassign $ARGS to a new list then it doesn't matter anyway but if you mutate in place then it is a feature more than a bug imo.

> 
The very fact that $ARGS += something works but $ARGS = $ARGS + something doesn't is, in my opinion, extremely counter-intuitive.

that is a python implementation detail, in-place addition uses `list.extend()` while the second one creates a new list.

SirNickolas: My solution is also, well, more consistent: it always introduces bugs. :) Thus making harder to miss them. The Zen says about that, too.

> that is a python implementation detail

I suppose, "corner case" using of implementation details is a kind of covert bug.

laerus: you can always use `list.insert()` instead

scopatz: I agree with the list copy that is written here.

scopatz:
Is this ready to go @SirNickolas @laerus?

xonsh/main.py
scopatz: If for no other reason than it was a copy beforehand and it could cause problems to change.

laerus:
oh yes i'm merging this

Issue #1517, opened 2016-08-01
No newline at end of .xsh file causes failure


mackeyja92:
Not having a new line at the end of my `.xsh` file causes all sorts of weird errors to be thrown. Here is an example when trying to do `bash-source nvm.sh` 

```
Traceback (most recent call last):
  File "/usr/local/Cellar/xonsh/0.4.3/libexec/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 10323, in wrapped_simple_command
    r = f(args, i)
  File "/usr/local/Cellar/xonsh/0.4.3/libexec/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 14407, in source_alias
    builtins.execx(fp.read(), 'exec', builtins.__xonsh_ctx__)
  File "/usr/local/Cellar/xonsh/0.4.3/libexec/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 16152, in exec
    transform=transform)
  File "/usr/local/Cellar/xonsh/0.4.3/libexec/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 16113, in compile
    tree = self.parse(input, ctx, mode=mode, transform=transform)
  File "/usr/local/Cellar/xonsh/0.4.3/libexec/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 16084, in parse
    tree, input = self._parse_ctx_free(input, mode=mode)
  File "/usr/local/Cellar/xonsh/0.4.3/libexec/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 16166, in _parse_ctx_free
    debug_level=(self.debug_level > 1))
  File "/usr/local/Cellar/xonsh/0.4.3/libexec/lib/python3.5/site-packages/xonsh/parsers/base.py", line 307, in parse
    tree = self.parser.parse(input=s, lexer=self.lexer, debug=debug_level)
  File "/usr/local/Cellar/xonsh/0.4.3/libexec/vendor/lib/python3.5/site-packages/ply/yacc.py", line 331, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/usr/local/Cellar/xonsh/0.4.3/libexec/vendor/lib/python3.5/site-packages/ply/yacc.py", line 1181, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
[...]

PR #1518, opened 2016-08-01
No error in parser error

scopatz:
Sometimes locs report lineno incorrectly, and we should fail correctly.

### DIFF
--- a/news/noerr.rst
+++ b/news/noerr.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fixed parser error line number exception from being raised while trying to
+  raise a SyntaxError.
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


laerus:
how can i reproduce this issue?

gforsyth:
@laerus, I think this is in response to #1517 but I can't reproduce that issue.

scopatz:
Yes, this is for #1517 and reproducing this involves setting you system encoding to something that doesn't support many code points, and then using those code points in an *.xsh file that you source.  What makes this tricky to test is that you need to set the system encoding before Python starts up at all, so that the default encoding is the wrong one.

gforsyth:
Hey @mackeyja92 -- would you mind trying out this fix to see if it helps your issue with the trailing newline error?  Thanks!

scopatz:
It would be nice to get this in too prior to v0.4.5

laerus:
I've tried to reproduce this but no luck. Think we can merge anyway?

scopatz:
Yeah, I think we can merge anyway.  It is a pretty obscure and hard to trigger error.

laerus:
thank you @scopatz! merging.

PR #1520, opened 2016-08-02
history show filter by timestamps

laerus:
added optional `history show` args:
`-T` `+T` to filter by timestamp
`-t` to print the timestamps
`-f` for the datetime format

some refactoring aswell and a `ensure_slice` bugfix

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -1588,6 +1588,9 @@ def iglobpath(s, ignore_case=False, sort_result=None):
 
 
 def ensure_timestamp(t):
-    t = float(t)
+    try:
+        t = float(t)
+    except TypeError:
+        t = t.timestamp()
     return t
 
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -933,7 +933,10 @@ def ensure_slice(x):
         return x
     try:
         x = int(x)
-        s = slice(x, x+1)
+        if x != -1:
+            s = slice(x, x+1)
+        else:
+            s = slice(-1, None, None)
     except ValueError:
         x = x.strip('[]()')
         m = SLICE_REG.fullmatch(x)
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -1585,3 +1585,9 @@ def _iglobpath(s, ignore_case=False, sort_result=None):
 def iglobpath(s, ignore_case=False, sort_result=None):
     """Simple wrapper around iglob that also expands home and env vars."""
     return _iglobpath(s, ignore_case=ignore_case, sort_result=sort_result)[0]
+
+
+def ensure_timestamp(t):
+    t = float(t)
+    return t
+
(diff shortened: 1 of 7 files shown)


scopatz:
Thanks for putting this in @laerus! I agree that a refactor is in order. I would prefer `history show bash -nr 5 6 7` too.  Also, it looks like there are some flake8 failures.

xonsh/tools.py
```diff
@@ -1585,3 +1585,12 @@ def _iglobpath(s, ignore_case=False, sort_result=None):
 def iglobpath(s, ignore_case=False, sort_result=None):
     """Simple wrapper around iglob that also expands home and env vars."""
     return _iglobpath(s, ignore_case=ignore_case, sort_result=sort_result)[0]
+
+
+def ensure_timestamp(t):
+    try:
```
scopatz: Can this be implemented without exceptions, please?  `isinstance()` maybe?

laerus: yes this the bare minimum to just pass the two tests. btw why do you prefer type checking over EAFP for this?

scopatz: Exceptions are horribly slow and if you use ABCs (like numbers https://docs.python.org/3.5/library/numbers.html) it is still generic.

scopatz:
@laerus my two cents is that this should just use `strptime()`, because it is well known and well documented.  We should have a sensible default for the format (probably isoformat), but then also allow users to pass in their own format string (probably using another command line argument). I think that covers all the cases without being too restrictive or writing and maintaining something that is already in the standard library.  Just my thoughts :)

laerus:
cool! that makes this easier to implement and doesn't make any assumptions about the format. Maybe we want an envvar that will hold the format string so we can use it in any place with `xonsh.tools.ensure_timestamp`? so you don't have to pass the custom format in `history show` args everytime.

scopatz:
> Maybe we want an envvar that will hold the format string so we can use it in any place with xonsh.tools.ensure_timestamp?

Sure, I think having the default as an envvar is a good idea, though the user should still be able to pass in a value on the command line too.

laerus:
allrighty! thinking of the `-f` optional arg for this, and maybe another option to show the timestamp when printing the commands. if the `-f` option is not present it would default to the envvar, any preference on the envvar name?

scopatz:
`-f` sounds good to me.  The envvar could be something like `$DEFAULT_DATETIME_FORMAT`. The most important thing is for it to be clear :)

xonsh/tools.py
laerus: i think we'll need at least one try/except block because if we get a string it may look like a float or a something that looks like datetime.

laerus:
one problem i encounter with `datetime.strptime()` at the moment is:
`datetime.strptime('15:33', '%H:%M')` this will default the year/month/day to 1900-1-1

laerus:
this is a very basic implementation of `ensure_timestamp` and leaves most of the configuration on the user, we may want at some point to take it a step further and add things like 'yesterday' 'today' which would be very cool!

gforsyth:
Yeah, that coverage thing is pretty weird...

@laerus, can you add a news entry for this?

scopatz:
Thanks a ton @laerus!

Issue #1464, opened 2016-07-21
Loading too diverse


astronouth7303:
There are currently 4 non-interactive ways of loading and executing code in xonsh (as of 84501037eb10d9bda963435e8a33166edb3aea01, shortly after the release of v0.4.4):

* Python import
* `xontrib load`
* `.xonshrc`
* sourcing

Currently, these all have slightly different semantics and rules. 

* Python `import` works as expected (the current virtual environment does not apply, but doing that would be a whole other can of worms)
* `source` loads up the entire file and executes it in the shell context
* `xontrib load` runs the xontrib in its own namespace, but uses slightly different importing rules than `from ... import *` (it ignores `__all__` and just copies everything not prefixed with a `_`)
* `xonshrc` is basically the same as sourcing, but it has a different code path

I propose to bring `xontrib load` in line with `import` by specifying `xontrib load` to behave the same as `from xontrib.<name> import *` (plus name resolution rules). Namely, by respecting `__all__` if the xontrib defines it. This reduces the effective number of types of executions to two (sourcing and importing, or shell-style and Python-style).

Currently, `xonsh.xontribs.xontrib_context()` handles this.

astronouth7303:
The other, more pragmatic, reason for the change is that it's currently really easy for xontribs to pollute the shell context by following standard Python coding conventions. Specifying it this way makes behavior clearer to Python developers, and gives xontributors a tool to directly control what gets exported.

scopatz:
Thanks for opening this up, @astronouth7303!  I think you more pragmatic proposal for `__all__` is something that should be discussed in a separate issue.  At least @gforsyth and I seem in favor of it.

As for this broader proposal, I am all for simplifying the code base.  However, I am not sure that there is much we can do here.  Each of these are implemented in slightly different ways because they have slightly different use cases.  That said, I am in favor of finding whatever commonality is there that hasn't already been implemented.  Ok, let's dive in.

* Python `import`s, as you said, should remain unchanged.
* sourcing should be like typing the file directly into the shell and executing it.  This has the important feature that if you source the same file twice in a row it should be fully executed twice in a row.  This is distinct from importing a file twice, which will only execute it once. 

I think we are all on board so far.

[...]

PR #1524, opened 2016-08-03
Add support for `__all__` in xontribs

astronouth7303:
In reference to #1464, allow xontribs to explicitly define their public API by defining `__all__`, just like regular Python modules.

The goal is to make `xontrib load spam` and `from xontrib.spam import *` effectively the same.

### DIFF
--- a/docs/tutorial_xontrib.rst
+++ b/docs/tutorial_xontrib.rst
@@ -69,8 +69,8 @@ default, they cannot be unloaded (easily).
 .. note::
 
     When a xontrib is loaded from a config file or via the xontrib command,
-    its public variables are placed in the current execution context, just
-    like variables set in run control files.
+    its public variables are placed in the current execution context unless
+    `__all__` is defined, just like in regular Python modules.
 
 Loading xontribs in the config file is as simple as adding a list of string
 xontrib names to the top-level ``"xontribs"`` key. For example, the following
--- a/docs/tutorial_xontrib.rst
+++ b/docs/tutorial_xontrib.rst
@@ -70,7 +70,7 @@ default, they cannot be unloaded (easily).
 
     When a xontrib is loaded from a config file or via the xontrib command,
     its public variables are placed in the current execution context unless
-    `__all__` is defined, just like in regular Python modules.
+    ``__all__`` is defined, just like in regular Python modules.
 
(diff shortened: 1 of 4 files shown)


astronouth7303:
(Travis failed due to unrelated timeout. Wish I could ask Travis to re-run it.)

scopatz:
Thanks a million for putting this in! This will be great to have. Relaunched.

docs/tutorial_xontrib.rst
```diff
@@ -69,8 +69,8 @@ default, they cannot be unloaded (easily).
 .. note::
 
     When a xontrib is loaded from a config file or via the xontrib command,
-    its public variables are placed in the current execution context, just
-    like variables set in run control files.
+    its public variables are placed in the current execution context unless
+    `__all__` is defined, just like in regular Python modules.
```
scopatz: rst requires double backticks

tests/test_xontribs.py
```diff
@@ -1,6 +1,56 @@
 """xontrib tests, such as they are"""
-from xonsh.xontribs import xontrib_metadata
+import pytest
+from xonsh.xontribs import xontrib_metadata, xontrib_context
+import sys
```
scopatz: This doesn't block because it is the tests, but PEP8 says that imports should be grouped.

xonsh/xontribs.py
```diff
@@ -36,7 +36,10 @@ def xontrib_context(name):
                           ImportWarning)
         return {}
     m = importlib.import_module(spec.name)
-    ctx = {k: getattr(m, k) for k in dir(m) if not k.startswith('_')}
+    if hasattr(m, '__all__'):
```
scopatz: Not that this is a huge deal, and not blocking, but I think you could save a lookup in m by doing:

```python
allnames = getattr(m, '__all__', None)
if allnames is None:
    ctx = {k: getattr(m, k) for k in dir(m) if not k.startswith('_')}
else:
    ctx = {k: getattr(m, k) for k in allnames}
```

scopatz:
Few minor inline comments, but otherwise looks good to go!

astronouth7303:
Ok, tweaked.

scopatz:
Thanks again!

PR #1527, opened 2016-08-04
Point at the right appveyor

astronouth7303:
README should point at the current appveyor.


melund:
Good catch. Thanks @astronouth7303

PR #1529, opened 2016-08-04
s.strip needs to be method call, 's.strip()', for the example to work

ericmharris:
s.strip needs to be method call, 's.strip()', for the example  correctly.  While not necessary, adding the newline ('world\n') puts 'hello
world' on its own line of output instead of having the xonsh prompt
appended to it.


melund:
Thanks for spotting this @ericmharris

PR #1545, opened 2016-08-11
Paste mode

Carreau:
Add option for carriage return to not autoindent.

Cannot be controlled through the $INDENT option as there is also the
copy_margin kwarg.

Mostly useful when programmatically inserting text in prompt_toolkit
from already indented source.

remove the `indent_` variable and lookup dynamically the $INDENT
variable as well for consistency and reflect the user config at any
instant.

`carriage_return`  now takes the kwonly: `autoindent=True[|False]`

--- 

Mostly need that for a prototype of "Faketyping" in a live Xonsh shell.

### DIFF
--- a/xonsh/ptk/key_bindings.py
+++ b/xonsh/ptk/key_bindings.py
@@ -36,7 +36,7 @@ def carriage_return(b, cli):
     # indent after a colon
     if (doc.current_line_before_cursor.strip().endswith(':') and
             at_end_of_line):
-        b.newline()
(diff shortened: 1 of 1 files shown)


Carreau:
Stupid me... don't merge yet.

Carreau:
The diff looks busy, but the commits themselves are easy to review, as the 2 first one are virtually no-op/renaming.

scopatz:
Thanks a ton @Carreau!

PR #1546, opened 2016-08-11
Improve documentation and messaging.

Carreau:
Develop info for aliases, and tell user that on first launch that if
they decide to configure now they can resume anytime.


scopatz:
Thanks @Carreau!

PR #1551, opened 2016-08-11
pip crash fix

### LINKED ISSUE #1548: `xonsh.completers.pip` explodes if `pip` is not on PATH
On my Windows installation, Python is not on PATH (because multiple Python madness), and therefore neither is pip. However, the pip completer [expects pip to be on the path](https://github.com/xonsh/xonsh/blob/master/xonsh/completers/pip.py#L14).

This causes the completer to blow up with a `FileNotFoundError` when it tries to complete.

scopatz:
@astronouth7303 does this fix #1548 for you?

### DIFF
--- a/news/pcrash.rst
+++ b/news/pcrash.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Made pip completer more robust to when pip is not installed.
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


astronouth7303:
I thought there was a way to get the filename of the python interpreter from `sys`, so you could do `sys.execname, '-m', 'pip'` and be much more likely to get pip (conda doesn't have pip).

But this should work. I'll check tomorrow.

gforsyth:
Hrm, conda has pip in Linux installs. Is it not present in Windows installs?

scopatz:
Well, you can always do silly things like change your PATH, etc. that would remove pip from being runnable.

astronouth7303:
@gforsyth Like I said in #1548. I have a Windows machine with Py2 and Py3, neither of which I put on the path so I don't get conflicts about which Python for what.

astronouth7303:
Yup, that fixes it.

gforsyth:
Thanks, @scopatz!

Issue #1533, opened 2016-08-05
Error starting xonsh from a protected directory


melund:
I tried to use xonsh on a machine where python was install for all users (i.e. system wide install on windows). If I run xonsh as admin I don't get the error. 

```
(C:\Anaconda3) C:\Users\kpr>xonsh
Traceback (most recent call last):
  File "C:\Anaconda3\Scripts\xonsh-script.py", line 5, in <module>
    sys.exit(xonsh.main.main())
  File "C:\Anaconda3\lib\site-packages\xonsh\main.py", line 209, in main
    args = premain(argv)
  File "C:\Anaconda3\lib\site-packages\xonsh\main.py", line 201, in premain
    setup_win_unicode_console(env.get('WIN_UNICODE_CONSOLE', True))
  File "C:\Anaconda3\lib\site-packages\xonsh\tools.py", line 1091, in setup_win_unicode_console
    import win_unicode_console
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 954, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 896, in _find_spec
  File "C:\Anaconda3\lib\site-packages\xonsh\imphooks.py", line 53, in find_spec
    if fname not in (x.name for x in scandir(p)):
PermissionError: [WinError 5] Access is denied: '.'
```

The error comes from `scandir` in `imphooks.py`. The error appears since some of the files are protected somehow. 

I think we need to wrap our `scandir` use in something that handles files not beeing accessible: 
```
def scandir_accessible(path='.'):
    gen = os.scandir(path)
    while True:
        try:
            yield next(gen)
        except PermissionError:
            pass
```

gforsyth:
Thanks for reporting, @melund!  

Was this Python < 3.5?  I'm curious, because if it was, then `scandir = os.scandir` and this might actually be a problem upstream.

gforsyth:
And if it was < 3.5, then maybe we should just vend `os.scandir` from Python 3.5?  or at least the important bits?

melund:
No. It was Python 3.5.2

```
$ xonfig
+------------------+----------------+
| xonsh            | 0.4.4          |
| Git SHA          | None           |
| Python           | 3.5.1          |
| PLY              | 3.7            |
| have readline    | True           |
| prompt toolkit   | 1.0.0          |
| shell type       | prompt_toolkit |
| pygments         | 2.1.3          |
[...]

PR #1552, opened 2016-08-12
import hooks access check

scopatz:
Attempts to address #1533.  @melund does this work for you?


melund:
Yep. Fixes the error for me. Thanks.

Issue #1550, opened 2016-08-11
include a specific "load" hook for Xontrib ?


Carreau:
I'm running into the question of whether Xontribs should do all their shenigan at import time, or not. 
My reasoning is that in the current one I'm developing, I need to have access to `__xonsh_shell__`, 
which is fine if I manually load the xontrib.

Among other I want to bind shortcuts, so I cannot use a lazy accessor, or I would have.  

Though, if I register the xontrib to be loaded at startup, it will be loaded during the initialisation of `__xonsh_shell__ = Shell(..)` leading to `__xonsh_shell__` not being in built-in yet, and me getting a name error. 

May I suggest to standardise with the **convention** that if a xontrib module define a function called `xontrib_init(parameters, to , define)` this get called at a well-known step in the initialisation process; possible in a state which will be identical as doing `xontrib load whatever` ?

I would _suggest_ not to pass any arguments for now, but to recommend implementation to allow `*args, **kwargs` parameters.

---

Also in general I don't like import having side effects., and I would prefer an init method that get passed a reference to a global namespace instead of implicitly accessing `__xonsh_*_`;  but that's a personal preference. 

Thoughts ?

astronouth7303:
I thought about rolling this as an event, but the problem is that xontribs loaded latter still need it to be called (behaving more like a future than an event).

Are the any documents that would describe the initialization process?

Carreau:
> I thought about rolling this as an event, but the problem is that xontribs loaded latter still need it to be called (behaving more like a future than an event).

Yeah, I was unsure of that either. I also want to modify this PR to actually call the equivalent of `xontrib load` instead of having the logic twice. I think that would be the most reasonable.

scopatz:
Yeah, as mentioned in the PR, I think something that is more amenable to an event system is preferable.

astronouth7303:
So the solution stubbed out in #1501 is that load hooks are part of the event subsystem as the `LoadEvent` species.

Basically, on fire, `LoadEvent` needs to call each registered handler. On subsequent calls, `fire()` will no-op. After the event fires, any handler that is registered is called immediately. (Note that any handlers registered while `fire()` is running must still be called.)

An implementation is like 80~90% there, but I wanted to make sure there was an opportunity to discuss the semantics.

astronouth7303:
[...]

PR #1554, opened 2016-08-12
[WIP] Draft of cleaner Xonfig loading.

Carreau:
Need to take care of tracking which xonfig have been loaded and don't
reload if the case.

Addresses (in part) #1550 

--- 
Open to discussion.

### DIFF
--- a/docs/tutorial_xontrib.rst
+++ b/docs/tutorial_xontrib.rst
@@ -27,6 +27,7 @@ took inspiration from for xonsh:
 
 Structure
 ==========
+
 Xontribs are modules written in either xonsh (``*.xsh``) or Python (``*.py``).
 Normally, these are stored and found in an
 `implicit namespace package <https://www.python.org/dev/peps/pep-0420/>`_
@@ -59,8 +60,21 @@ Here is a sample file system layout and what the xontrib names would be::
           |- done.py     # "mypkg.subpkg.done", full module name
 
 
+Xontribs **may** define functions named ``xontrib_init``, that will be called
+at xontrib load time. ``xontrib_init`` is recommended to accept an arbitrary
+number of parameters, but for now, no parameter will be passed to the function. 
+
+That is to say::
+
+
+    def xontrib_init(*args, **kwargs):
+        # init my xontrib
+        print('my contrib is being loaded')
+
+
 Loading Xontribs
 ================
+
 Xontribs may be loaded in a few different ways: from the config file,
 dynamically at runtime with the ``xontrib`` command, or by importing the
 module normally. Since these extensions are just Python modules, by
@@ -105,6 +119,7 @@ Of course, you have to use the full module name to import a xontrib:
 
 Listing Known Xontribs
 ======================
+
 In addition to loading extensions, the ``xontrib`` command also allows you to
 list the known xontribs. This command will report whether known xontribs are
 installed and if they are loaded in the current session. To display this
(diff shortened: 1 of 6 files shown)


xonsh/ptk/shell.py
```diff
@@ -39,6 +39,9 @@ def __init__(self, **kwargs):
         self.key_bindings_manager = KeyBindingManager(**key_bindings_manager_args)
         load_xonsh_bindings(self.key_bindings_manager)
 
+        self._prompt_tokens = None
```
gforsyth: Hey @Carreau -- re the `_prompt_tokens` stuff, I agree with the changes but is it part of this feature?

Carreau: No sorry, miss-commited.

gforsyth:
I like this idea -- it's especially nice for completion xontribs, since the `completer add` functionality isn't available until after startup, so some wacky stuff ends up being used to properly load new completions.  

I can't get it to work, though.  I tried to use `xontrib_init` to do a `completer add` but it doesn't get loaded into my completer list.

Carreau:
> I can't get it to work, though. I tried to use xontrib_init to do a completer add but it doesn't get loaded into my completer list.

We likely need to move the xontrib initialisation even later

Carreau:
> @astronouth7303 said: Should this be a dunder special function eg __init__()?

Well, i think that `__init__` would be confusing for various reasons:

1) it is rarely if ever at module scope.
2) It has a specific meaning for Python himself
3) A user coming across this has ~0 change to google the right thing.
4) if you are giving a call and said to your audience to create  `dunder-init` they will be confused. (assuming your users know the "dunder" word.)

if we want a magic method at module level I would thinkmore about `__main__`, but then that makes the module executable.

I'm fine changing but I think the explicit method name is better.

scopatz:
Thanks for putting this in @Carreau! And sorry for taking so long to review this....

In general, I think a named xontrib API is a slippery slope. I think it can become completely unmanageable as needs and use cases change and evolve.  A prime example of this is Vistrails where the API is practically unusable, while being essential to actually writing plugins.  Furthermore, special names should be special, so something like `__xontrib_init__()` would be better than a public variable name to me.

I would prefer some kind of metaprogramming technique.  Maybe a decorator that simply labels the functions to be run (for the time being) would avoid the special naming issue and also allow us to evolve the backend implementation without the xontrib authors needing to worry about a changing API. This could let us swap in an event system later transparently.  For example, I would see something like:

```python
from xonsh.xontrib import xontrib_init

@xontrib_init
def my_initer(*args, **kwargs):
    pass
```

This would even allow a xontrib to have multiple inits, if it needed them for some reason.  Or we could have a more robust system that has different loaders/unloaders

```python
from xonsh.xontribs import loader, unloader

@loader
def my_initer(*args, **kwargs):
    pass

@loader
def my_other_init(*args, **kwargs):
   pass

@unloader
def teardown(*args, **kwargs):
    pass
```

I think this kind of system is a bit more robust overall.  Feel free to reject any names, etc. The main point is that something closer to an event system interface I think is much better than special names.

Thanks again, @Carreau!

gforsyth:
I'm going to close this since we have an events system now.  Please reopen if desired!

astronouth7303:
Not exactly, but this sounds a lot like the ideas for #1550.

gforsyth:
got it, reopening then.

astronouth7303:
No, closed is good. The loading system should go through the event system (as I outlined over there), but there's missing infrastructure pieces before it can start being used.

Issue #1556, opened 2016-08-13
IGNOREEOF doesn't work


OJFord:
I set `IGNOREEOF` during the config wizard, but it's not having any effect - I can leave xonsh with Ctrl-D.

```sh
$ grep 'IGNOREEOF' "$XDG_CONFIG_HOME/xonsh/config.json"
  "IGNOREEOF": true,
$ xonsh --version
('xonsh/0.4.4',)
$ uname -r
4.7.0-1-ARCH
```

laerus:
this exists on current master as well and it's defo a bug. nice catch!

`$IGNOREEOF` doesn't affect `EOF` too

scopatz:
Thanks for reporting @OJFord! Potential fix in #1557

scopatz:
The fix for this has been merged.  Please feel free to reopen if needed.

PR #1557, opened 2016-08-13
EOF fix for prompt-toolkit.

scopatz:
Should address #1556, @OJFord @laerus  would you mind trying this out?

### DIFF
--- a/news/ptkeof.rst
+++ b/news/ptkeof.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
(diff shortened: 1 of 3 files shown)


laerus:
this fixes Ctrl-D, do we want to ignore `EOF` aswell?

scopatz:
Well, EOF is just an alias to exit so I don't think it is worth it. It isn't a literal end of  file.

laerus:
Cool, thank you! Merging this.

PR #1562, opened 2016-08-14
fix coverage

laerus:
figuring out whats wrong with coverage report

this PR is just for exploration, the fix will be a new PR with only 1 commit.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,7 +1,2 @@
 [run]
 branch = true
-omit =
-  tests/*
-  xonsh/__amalgam__.py
-  */parser*_table.py
-  setup.py
--- a/.gitignore
+++ b/.gitignore
@@ -41,5 +41,3 @@ include/
 *.komodo*
 .cache
 .idea
-
-.coverage
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ include/
 *.komodo*
 .cache
 .idea
+
+.coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ install:
       source activate test_env;
     else
       pip install -r requirements-tests.txt;
-    fi
+    fi;
     python setup.py install;
 
 script:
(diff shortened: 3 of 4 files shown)


scopatz:
I bet part of what is wrong is that both amalgamated and non-amalgamated code is not getting executed. This will probably reduce the coverage by 50%

scopatz:
Also thanks for taking this on!

laerus:
do we rly need to use conda for linux/mac builds?

laerus:
i'm totally confused right now :laughing: 

btw do we want to test coverage against a non amalgamated build?

scopatz:
> do we rly need to use conda for linux/mac builds?

Well, it is more consistent than using brew and something else.

> btw do we want to test coverage against a non amalgamated build?

Maybe, probably.  We probably want to make sure that the tests are successful in both amalgamated and non-amalgamated forms.  This *should* be as easy as running the tests twice and setting/exporting the envvar `XONSH_DEBUG=1` between them.

laerus:
>    >  do we rly need to use conda for linux/mac builds?
> 
> Well, it is more consistent than using brew and something else.

why can't we just  remove all the conda related stuff on `before_install`??
sry if i'm missing something but why do we need to setup conda for this?

gforsyth:
@laerus -- there are also speed considerations.  the `miniconda` install is generally faster, even with the extra download time

laerus:
@gforsyth do we use conda in general so we can test against MacOS builds?

btw i got this all wrong, the problem is that *No coverage report found for master*
whatever i do here the problem won't change until there is actually a coverage report for master.

i'm trying for a PR to this PR and see if i can make something out of it

laerus:
well this seems like it worked, https://github.com/laerus/xonsh/pull/2

but no sure what fixed this with all the travis.yml ripping :laughing:

laerus:
PR to PR doesn't help.

the issue may be the `--cov-report=term` on `script` section in `travis.yml`

also in the `install` section if the build is not nightly we pip install requirements but we never run `setup.py install` is that normal?

PR #1574, opened 2016-08-17
some more release abstractions

scopatz:
This makes it a little easier to move the release script around.  I might be worth thinking about breaking this out as its own project(?)

### DIFF
--- a/release.xsh
+++ b/release.xsh
@@ -30,6 +30,13 @@ WILL_DO = {
   'do_conda': True,
   'do_docs': True,
 }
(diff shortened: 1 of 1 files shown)


gforsyth:
Looks good!  Yeah, this could work as a standalone project.  Obviously would require `xonsh` as a dependency, but then we could have a setup wizard / argparser to create a `release.xsh` for a given project.

scopatz:
Oh yeah! That is a cool idea.  Have a bunch of template pieces that get glued together...

PR #1578, opened 2016-08-18
Use pylauncher on Windows when available

### LINKED ISSUE #1395: handle multiple python versions on windows
I have python2.7 (installed first, in c:\python27) and python3.5 (installed second, in c:\users\...).  Both are in the path, but python2.7 is first.  The following installs xonsh fine:

    pip3 install xonsh

but when trying to run it, you get the error `No module named xonsh`.  It seems it finds the `xonsh.bat` file in the python3 scripts directory, but the script executes the first python.exe found on the path (the 2.7 version).

As a workaround I modified xonsh.bat to:

    @echo off py -3 -m xonsh.main %*

so that it would use pylauncher ([pep397](https://www.python.org/dev/peps/pep-0397/)).  There may be a better way to do this, but using pylauncher in some way might be a good idea because it's included with python starting with 3.3.

BYK:
Uses the trick from http://stackoverflow.com/a/18982410/90297

Fixes #1395.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,10 @@ Xonsh Change Log
 
 .. current developments
 
+**Changed:**
+
+* ``xonsh.bat`` tries to use `pylauncher <https://www.python.org/dev/peps/pep-0397/>`_ when available.
+
 v0.4.5
 ====================
 
--- a/scripts/xonsh.bat
+++ b/scripts/xonsh.bat
@@ -1,2 +1,14 @@
 @echo off
-python -m xonsh %*
+call :s_which py.exe
+if not "%_path%" == "" (
+  py -3 -m xonsh %*
+) else (
+  python -m xonsh %*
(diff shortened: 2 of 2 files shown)


melund:
Doesn't seem that Anaconda/miniconda comes with pylauncher. So I can't test that it works. 

But I can confirm that it still works if pylauncher is not present. Remember to add a entry to the changelog.

BYK:
@melund I tested it with both pylauncher and without it and made sure it uses pylauncher when it exists :)

BYK:
@melund Updated changelog

melund:
Lgtm. I guess the hanging tests on MacOS is irrelevant....

melund:
Hmm... Can't merge from my phone when tests are not completed.

BYK:
@melund looks like this is a problem with Travis CI. I'll merge since you said LGTM.

PR #1581, opened 2016-08-18
change codeblock format from bash to console

### LINKED ISSUE #1580: Commented command in addditional setup
http://xon.sh/add_to_shell.html#additional-setup

the part to add it to the shells is commented out

Is it intended?

gforsyth:
resolves  #1580


melund:
Lgtm

SanketDG:
Looks like this is used in a lot of places. It isn't really bash, (its a whole different shell :smile:) so using `console` in all places should be better.

adqm:
Looks good to me!  Merging.

PR #1584, opened 2016-08-18
temporary files shouldn't cause the news-related test to break

adqm:
This small change modifies the news-related test to check that `rst` is a substring of the extension or all news files, rather than matching the extension directly.  This means that the presence of vim's backup files (`.rst~`) doesn't cause the test to fail.


scopatz:
Thanks @adqm!

Issue #1411, opened 2016-07-10
feature request: more history control


laerus:
edit: After #1585 The feature request is about changing the `__xonsh_history__` object API to resemble the functionality presented below and __not__ about __adding new syntax__ to xonsh.

bash history expansion has *event designators* which can be specified almost anywhere in the command line and they are references to a cmd entry in the history list i.e:

```
!n     Refer to command line n.

!string
              Refer  to the most recent command preceding the current position
              in the history list starting with string.
!?string
              Refer to the most recent command preceding the current  position
              in  the  history  list containing string.

^string1^string2^
    Quick substitution.   Repeat  the  previous  command,  replacing string1  with  string2.
```

after these event designators you can specify *word designators. which are used to select desired words from the event.  A : separates  the event specification from the word designator.* i.e:

```
!:n            the nth word
!:$            the last word
!:x-y         a range of words
```

any ideas on how can we go about implementing these for xonsh?

laerus:
related to #1221

scopatz:
Thanks for bringing this up, @laerus!  I am not sure that having a special syntax for these things is needed, since each of these cases seems to be do-able at worst by dropping into Python mode with `@()`.  Maybe we need better examples in the translation guide?

dylanmc:
I've got !$ and !* hard-wired into my bash-fingers, so having these work in xonsh would be great. Also :p at the end of a history-modifier prints that item, and adds it to your history, but doesn't execute it - also really useful.

scopatz:
Perhaps we could / should add common Bash idioms like this as a xontrib.

laerus:
my main concern is how we can implement this in a way that it requires a short syntax.

++ for a xontrib that adds common bash idioms like this.

laerus:
@scopatz do you think it would be proper to implement these in the `_` underscore ie:
`_[-1]` is the last command and `_[-1][-1]` is the last argument of the last command etc so we can do something like:
```
$ mkdir /home/laerus/xonsh/test/some_long_dir_name
$ cd @(_[-1][-1])
```

another way would be to change the the `!n` to return the nth command inp without executing.

atm i like the idea of overloading `_` cause i don't think is being used a lot for just accessing the last command which would just change to `_[-1].rtn`

scopatz:
I don't think it is safe to use `_` for this everywhere.  The fact is that `_` already has semantic (if not syntactic) meaning elsewhere in Python.  

[...]

PR #1585, opened 2016-08-18
history syntax

laerus:
a first draft on implementing history syntax, #1411 
currently `!{...}` is used but this is subject to changes.

### DIFF
--- a/xonsh/history.py
+++ b/xonsh/history.py
@@ -552,7 +552,7 @@ class History(object):
         self.rtns = CommandField('rtn', self)
 
     def __getitem__(self, item):
-        print("GETTING HISTORY:", item)
+        return self.show(slices=item)
 
     def __len__(self):
         return self._len
@@ -609,15 +609,11 @@ class History(object):
         return hf
 
     def show(self, *args, **kwargs):
-        """Return shell history as a list
+        """Return session history."""
+        return list(c for c, *_ in _hist_get(*args, **kwargs))
 
-        Valid options:
-            `session` - returns xonsh history from current session
-            `xonsh`   - returns xonsh history from all sessions
-            `zsh`     - returns all zsh history
-            `bash`    - returns all bash history
-        """
-        return list(_hist_get(*args, **kwargs))
+    def get(self, index, *args, **kwargs):
+        return list(c for c, *_ in _hist_get(*args, **kwargs))[index]
 
 
 def _hist_info(ns, hist):
--- a/xonsh/history.py
+++ b/xonsh/history.py
@@ -551,6 +551,9 @@ class History(object):
         self.outs = CommandField('out', self)
         self.rtns = CommandField('rtn', self)
 
+    def __getitem__(self, item):
+        print("GETTING HISTORY:", item)
+
     def __len__(self):
         return self._len
 
--- a/xonsh/lexer.py
+++ b/xonsh/lexer.py
@@ -203,6 +203,7 @@ def special_handlers():
     _make_matcher_handler('$(', 'DOLLAR_LPAREN', False, ')', sh)
     _make_matcher_handler('$[', 'DOLLAR_LBRACKET', False, ']', sh)
     _make_matcher_handler('${', 'DOLLAR_LBRACE', True, '}', sh)
+    _make_matcher_handler('!{', 'BANG_LBRACE', True, '}', sh)
     _make_matcher_handler('!(', 'BANG_LPAREN', False, ')', sh)
     _make_matcher_handler('![', 'BANG_LBRACKET', False, ']', sh)
     _make_matcher_handler('@(', 'AT_LPAREN', True, ')', sh)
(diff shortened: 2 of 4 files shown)


laerus:
i think we should probably remove `__xonsh_history__.show()` method in favor of `.get()` and it should get only session history

adqm:
I wasn't around for the original conversation here, but this is very very close to the syntax proposed (and partially implemented) for inline suites (see https://github.com/xonsh/xonsh/pull/1223#issuecomment-226321526).

I think it might be worth revisiting and potentially modifying one of those syntactic structures.

laerus:
this was just a proposal by @scopatz on #1411 
it's open for discussion and far far from getting close to even be considered for a merge.
i'm still not even sure if we should have special syntax for history (but i wanted so much to try and implement it :blush:)

scopatz:
Yeah, I would like to retract this syntax proposal for history.  If anything, I think access to the history object could be done via a public name, like how `aliases` goes into builtins.  Like @adqm this looks too similar to the `!{}!` syntax that we decided on.

Basically, the only reason to have a public name is because users feel uncomfortable using `__xonsh_history__` directly. Unfortunately, `history` is already taken as a command name.  Maybe this could be `hist`?  I don't know if I have a good name here....

scopatz:
Thanks for putting this in @laerus!

laerus:
i'll repeat myself but my main concern considering new syntax includes the special ENV syntax too, atm there are three ways for someone to access envvars, `$ENVVAR` `${'ENVVAR'}` `__xonsh_env__['ENVVAR']` , isn't that redundant for something that imo isn't even so widely used? History is probably used more on a daily basis than setting up envvars (at least by me). And eventually someone will come along and argue that they use something else (i.e `__xonsh_ctx__`) more than both ENV and HISTORY so why not have special syntax for that something. I'm not rly familiar with syntax parsing etc but maybe we can have a customizable operator that the user can choose which object it will return? Back to this PR, maybe we can use `H{...}` for history?

scopatz:
Well, I think that the environment and history have fundamentally different purposes and usages and I don't think that either of them really deserve their own special syntax for getting ahold of the underlying object.  (Also, I think usage patterns and what is used more depends a lot on the individual.) 

To explain my thinking a bit more, to me the fundemental instance of the environment is `__xonsh_env__`, the fundamental instance of history is `__xonsh_history__`, and the fundamental instance of aliases is `aliases`.  These are the names that are shoved into builtins and used to access the objects from the parser and elsewhere.

Now, unlike history, the environment is something that is modified and manipulated by the user.  There are a few different pieces of syntax that enable this, `$NAME` for name based lookup and `${expr}` for Python expression based lookup.  In both cases the parser goes and finds `__xonsh_env__`.  

Now because there is expression based lookup, we can co-opt looking up a literal Python `Ellipsis` to mean "look up the whole environment -> self".  So `${...}` really is `__xonsh_env__.get(...)` and that happens to return self (ie `__xonsh_env__`).  

However, history has no generic lookup syntax in xonsh right now.  So it is not possible to appropriate its lookup to also return itself.  Additionally, it is not entirely clear to me what a history lookup syntax should actually provide (this session, all sessions, other shell's history, etc).  This is why I think history lookup is better left as an API that can be modified and evolve rather than as custom syntax. 

Additionally, it seems like a history is of limited use in most xonsh scripts / modules.  It mostly useful at the command line.  Also, this would only be used as lookup, because modifying history should not be allowed.

So for those reasons, I think that history is actually much more like `aliases` or `execx` or `compilex` where a public name and API is the better option.  I'd be all in favor of renaming `__xonsh_history__` to something like `hist`, for example, to give people access to it more readily.

I hope this makes sense.  Maybe other folks can weigh in too.

laerus:
Thanks for your concern on this @scopatz! I'm convcinced, a better api on the `History` object can solve the need for special syntax, one can then simply assign `__xonsh_history__` to just `H` or whatnot and do a history session lookup (current session imo) with i.e indexing, depending on the api. The same can be apllied to `__xonsh_env__` though, `$NAME` is bashwards and all but it's a little reduntant to have special python syntax with `${...}` when you have the object itself with item getting/setting api, truly i stopped using `${...}` cause it's annoying to type as it requires holding down the shift-key, even `${'XONSH_DEBUG'}` is super annoying when used with single quotes, lots of finger pain. I just assign `e=__xonsh_env__` and use that for experssion based lookups and/or assigning. Maybe we should consider removing `${}` as a special syntax for the enviroment and even change `aliases` to `__xonsh_aliases__` for consistency with the xonsh builtins. I'm closing this PR and it can be a reference for similar issues.

scopatz:
Thanks @laerus! On one level it is a little sad that this didn't work out because I am totally in favor of having more people bang on the parser and lexer.

laerus:
no worries! it was very educational for me messing around with this, but having *one way to do it* is more important in these cases.

PR #1586, opened 2016-08-18
Wrap Bash's Tab Completions in Quotes

adqm:
This change causes the results of Bash's tab completion to be escaped (quoted) where necessary (e.g., filenames with spaces).


laerus:
works perfectly, thank you @adqm !! merging

PR #1587, opened 2016-08-18
coverage report fix

laerus:
maybe it's that simple? can't find a way to test this when no report already on master.


scopatz:
Seems reasonable...

scopatz:
Thanks fo r putting this in @laerus!

Issue #1589, opened 2016-08-19
UnboundLocalError when disabling script caching for s script that has already been run and cached once.


scopatz:
Maybe @adqm knows how to solve this off the top of his head? :)

```console
scopatz@artemis ~/temp $ $XONSH_CACHE_SCRIPTS = False
scopatz@artemis ~/temp $ env XONSH_DEBUG=1 xonsh t.xsh 
Traceback (most recent call last):
  File "/home/scopatz/.local/bin/xonsh", line 3, in <module>
    main()
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/main.py", line 233, in main
    loc=None, mode='exec')
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/codecache.py", line 144, in run_script_with_cache
    update_cache(ccode, cachefname)
UnboundLocalError: local variable 'cachefname' referenced before assignment
```

*edited*

adqm:
Possibly!  Could you please make a traceback with amalgam turned off?

adqm:
Nvm, found the relevant spot in the code.  Looking into it now.

scopatz:
Edited in original post

scopatz:
Thanks for looking into it!

adqm:
Simple fix.  PR incoming.

scopatz:
Amazing @adqm!

PR #1590, opened 2016-08-19
Fix broken code cache

adqm:
As mentioned in #1589, disabling the script cache leads to problems.  I think this little patch should fix.


scopatz:
Thanks a ton @adqm!

Issue #776, opened 2016-04-01
fdjskflj --arg=$ENV kills xonsh


asmeurer:
I was trying to debug some other issue when I got this. `fdjskflj` is not a real command.

```
xonshmaster=🐚 fdjskflj --arg=$ENVTraceback (most recent call last):
  File "/Users/aaronmeurer/anaconda3/bin/xonsh", line 6, in <module>
    sys.exit(xonsh.main.main())
  File "/Users/aaronmeurer/Documents/xonsh/xonsh/main.py", line 236, in main
    shell.cmdloop()
  File "/Users/aaronmeurer/Documents/xonsh/xonsh/ptk/shell.py", line 111, in cmdloop
    line = self.singleline(auto_suggest=auto_suggest)
  File "/Users/aaronmeurer/Documents/xonsh/xonsh/ptk/shell.py", line 80, in singleline
    display_completions_in_columns=multicolumn)
  File "/Users/aaronmeurer/Documents/xonsh/xonsh/ptk/shortcuts.py", line 104, in prompt
    document = cli.run(reset_current_buffer=False)
  File "/Users/aaronmeurer/anaconda/lib/python3.5/site-packages/prompt_toolkit-0.57-py3.5.egg/prompt_toolkit/interface.py", line 353, in run
  File "/Users/aaronmeurer/anaconda/lib/python3.5/site-packages/prompt_toolkit-0.57-py3.5.egg/prompt_toolkit/eventloop/posix.py", line 150, in run
  File "/Users/aaronmeurer/anaconda/lib/python3.5/site-packages/prompt_toolkit-0.57-py3.5.egg/prompt_toolkit/eventloop/posix.py", line 82, in read_from_stdin
  File "/Users/aaronmeurer/anaconda/lib/python3.5/site-packages/prompt_toolkit-0.57-py3.5.egg/prompt_toolkit/terminal/vt100_input.py", line 373, in feed
  File "/Users/aaronmeurer/anaconda/lib/python3.5/site-packages/prompt_toolkit-0.57-py3.5.egg/prompt_toolkit/terminal/vt100_input.py", line 292, in _input_par
  File "/Users/aaronmeurer/anaconda/lib/python3.5/site-packages/prompt_toolkit-0.57-py3.5.egg/prompt_toolkit/terminal/vt100_input.py", line 325, in _call_hanr
[...]

PR #1592, opened 2016-08-19
Handle More Syntax Errors

adqm:
This patch allows xonsh to gracefully handle more syntax errors than before.

There is a class of syntax errors that seem not to result from expressions that are invalid as specified by the grammar, but rather from using certain expressions in contexts they aren't made for (e.g., `del 7`).  Previously, xonsh did not report these as syntax errors; rather, they fell through the parsing stage and were reported as other `ValueError`s downstream, which prevented the automatic jump to subprocess mode.

Not a life-changing fix, probably, but it does make all (I think) syntax errors behave in the same way.  Partially addresses #776, though the real error with that particular expression was from the dollar sign.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1809,29 +1809,29 @@ def test_syntax_error_del_ifexp():
     with pytest.raises(SyntaxError):
         PARSER.parse('del x if y else z')
 
-def test_syntax_error_del_comps():
-    with pytest.raises(SyntaxError):
-        PARSER.parse('del [i for i in foo]')
-    with pytest.raises(SyntaxError):
-        PARSER.parse('del {i for i in foo}')
-    with pytest.raises(SyntaxError):
-        PARSER.parse('del {k:v for k,v in d.items()}')
-    with pytest.raises(SyntaxError):
-        PARSER.parse('del (i for i in foo)')
 
-def test_syntax_error_del_ops():
+@pytest.mark.parametrize('exp', ['[i for i in foo]',
+                                 '{i for i in foo}',
+                                 '(i for i in foo)',
+                                 '{k:v for k,v in d.items()}'])
+def test_syntax_error_del_comps(exp):
     with pytest.raises(SyntaxError):
-        PARSER.parse('del x + y')
-    with pytest.raises(SyntaxError):
-        PARSER.parse('del x and y')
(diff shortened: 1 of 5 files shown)


tests/test_parser.py
```diff
@@ -1782,3 +1782,153 @@ def test_redirect_error_to_output(r, o):
...
+def test_syntax_error_del_comps():
+    with pytest.raises(SyntaxError):
+        PARSER.parse('del [i for i in foo]')
+    with pytest.raises(SyntaxError):
+        PARSER.parse('del {i for i in foo}')
+    with pytest.raises(SyntaxError):
+        PARSER.parse('del {k:v for k,v in d.items()}')
+    with pytest.raises(SyntaxError):
+        PARSER.parse('del (i for i in foo)')
+
```
laerus: you can also use the `@pytest.mark.parametrize` decorator feature which is very handy, just for future reference!

adqm: Oh, neat.  I will do that now.

adqm:
okay, just updated to use the `parametrize` decorator

scopatz:
Hi @adqm! Thanks a million for this.  I am curious if the `check_contexts()` is in the right place, though.  It seems like this could be incorporated into the `xonsh.ast.CtxAwareTransformer` and save us from scanning over the whole tree twice.  Maybe, that class can/should be renamed to XonshTransformer or some such.  It already does more than just the context sensitive transformations.  

Obviously, this is a bigger deal for scripts and modules, rather than the REPL.

Other than that, it looks great!

adqm:
I thought about that myself; for me, consistency with Python won out over the (presumably slight) speed boost we would get from moving this.

I wanted this to be a part of the original `parse` call, though, since that's closer to the way it is handled in Python itself (trying to call `ast.parse` on one of these examples will fail rather than yielding an AST) and this should be an error for anyone calling `parse`, even if it's outside the normal xonsh setup with the context-sensitive transformations.  Originally I had hoped to do this in one go (error during the actual PLY parse), but I ran into a whole bunch of trouble with the way exceptions were being handled there...

I still think this is the "right" way (or at least closer to it), but I'm open to changing if you feel strongly about it.

scopatz:
That's fair and reasonable! Thanks again @adqm!

PR #1594, opened 2016-08-20
fix shebang to use python3 and remove unknown -u parameter

rbignon:


scopatz:
Thanks for putting this in @rbignon!  The `-u` parameter is pretty important since it tells Python to start in unbuffered mode, which is required on some systems.  +1 for the py3k fix, though.

rbignon:
The problem is that /usr/bin/env does not accept to get extra parameters, so I get this error:

$ xonsh
/usr/bin/env: ‘python3 -u’: No such file or directory

After googling, I get this answer:

https://stackoverflow.com/questions/3306518/cannot-pass-an-argument-to-python-with-usr-bin-env-python

scopatz:
Well sort of. `env` does accept multiple arguments.  However, whatever shell you are using to start up xonsh limits you to two arguments on the shebang line.  And for whatever reason pip, setuptools, distutils, etc) isn't rewriting your shebang line to point to the python interpreter you installed with. 

The fix is to use the `xon.sh` script instead, which is the guaranteed sh-lang compatible way of starting xonsh

rbignon:
Hm you're right. Note that I've linked ~/bin/xonsh to ~/src/xonsh/scripts/xonsh as I use the development version and I'm going to do some changes.

I change my link to ~/src/xonsh/scripts/xon.sh. Thank you.

scopatz:
Thanks @rbignon.

PR #1598, opened 2016-08-21
Function Macros

scopatz:
This PR implements function macros via a Rust-like syntax.  Important points to consider are, 

1. Any Python function/callable may be called as a macro (no special macro definition syntax)
2. Uses annotations to declare argument converters,
3. Partially satisfies the same needs as Jupyter/IPython magics, but with an expanded underlying system.
4. The `f!()` syntax was chosen to be consistent with Rust, minimizing what users may need to learn
5. This was a warm-up exercise for subproc-mode parser suppression and Block macros.

More documentation has been provided in the PR itself. Thoughts?  

I think there are some really interesting opportunities with a macro system, especially with respect to data display and foreign shells. Hopefully this isn't excessive dark wizardry...


scopatz:
Fixed some flake8 issues here

melund:
Fascinating. I know simple macros from C and a few other languages. Not familiar with Rust so this is quite interesting :) I think I need to see some more elaborate uses of it to really grok the power :)

scopatz:
Thanks @melund!  Yeah, I have a couple of examples brewing in my head that I won't be able to get to until tonight.  But for example, you could have a macro that autolabels plots with just `plot!(x, y)`.  Or you could have one that autojits, `jit!([a + b for a, b in x])`.  I think this starts to become extra powerful when mixed with singledispatch or multidispatch.

In any event, a more detailed review would be most appreciated!

blahgeek:
Hi @scopatz , I haven't tried the code yet but I really like the idea here.

I came up several use cases, like:

- `time!(cp a.bin b.bin)`, `time!(python_function())`: prints timer statistic for a shell command or python expr
- `explain!(...)`: Prints AST, explain how this line will be executed (python mode or subprocess mode)

Several suggestions:

- Maybe an option to mark an function macro only? Maybe a `@macro_only` decorator. For example, we may want to call `time!` as macro, but do not expect `time` to do anything.
- Personally I'd prefer `macro! arg1 arg2` syntax, after all we're in a shell, this may be cleaner

blahgeek:
PS: regarding `macro! arg1 arg2` syntax, maybe it can implemented by alias?

blahgeek:
PS: regarding `macro! arg1 arg2` syntax, maybe it can implemented by alias?

scopatz:
Thanks a ton for looking at this @blahgeek!  As per your suggestions, 

> Maybe an option to mark an function macro only? Maybe a @macro_only decorator. For example, we may want to call time! as macro, but do not expect time to do anything.

Yeah, I figured someone would want a @macro decorator.  I figured this could come later as a relatively small PR.

> Personally I'd prefer macro! arg1 arg2 syntax, after all we're in a shell, this may be cleaner... maybe it can implemented by alias?

Yes! This is definitely coming for subprocess mode - if/when this is merged.  This PR was a warm up that helped me clarify how exactly this stuff works, while hopefully also being useful.  I also want to add a `with!` sytnax for context managers.

scopatz:
Also, I am glad you like the idea.  That is reassuring ;)

scopatz:
OK, I have now updated this to allow macro argument escaping, via `f!(x, *, y)` syntax.  Everything before the `*` is a macro argument and everything after is a normal Python argument.

docs/tutorial_macros.rst
```diff
@@ -0,0 +1,350 @@
...
+
+    # Single arg
+    f!(x)
+    g!([y, 43, 44])
+
+    # Two args
+    f!(x, x + 42)
+    g!([y, 43, 44], f!(z))
+
+Not so bad, right?  So what actually happens when to the arguments when used
```
gforsyth: what actually happens ~~when~~ to the arguments

docs/tutorial_macros.rst
```diff
@@ -0,0 +1,350 @@
...
+in a macro call?  Well, that depends on the definition of the function. In
+particular, each argument in the macro call is matched up with the corresponding
+parameter annotation in the callable's signature.  For example, say we have
+an ``identity()`` function that is annotates its sole argument as a string:
```
gforsyth: function that ~~is~~ annotates its sole argument

docs/tutorial_macros.rst
```diff
@@ -0,0 +1,350 @@
...
+    >>> identity('me')
+    'me'
+
+    >>> identity(42)
+    42
+
+    >>> identity(identity)
+    <function __main__.identity>
+
+However, if we perform macro calls instead we are now guaranteed to get a
```
gforsyth: to get ~~a~~

docs/tutorial_macros.rst
```diff
@@ -0,0 +1,350 @@
...
+     - ``'exec'`` (default) or ``'single'``
+     - Execs the argument and returns None.
+   * - Type
+     - ``type``
+     - ``'t'`` or ``'type'``
+     -
+     - The type of the argument after it has been evaluated.
+
+These annotations allow you to hook into whichever stage of the compilation
+that you desire. It is important note that the string form of the arguments
```
gforsyth: important *to* note

xonsh/built_ins.py
```diff
@@ -680,6 +683,197 @@ def list_of_strs_or_callables(x):
...
+    Returns
+    -------
+    The converted argument.
+    """
+    # munge kind and mode to start
+    mode = None
+    if isinstance(kind, cabc.Sequence) and not isinstance(kind, str):
+        # have (kind, mode) tuple
+        kind, mode = kind
+    if isinstance(kind, str):
```
gforsyth: i'm probably missing something but couldn't these be `elif`s?

scopatz: They shouldn't be.  The first block unpacks a tuple and reassigns kind.  The second also reassigns kind.

gforsyth: ahh, right.  :+1:

gforsyth:
Hey @scopatz -- this looks good to me.  And after taking some time to really look it over I don't think it's blackmagicky at all.  Reasonably straight-forward given how seemingly wacky it can look at first glance.  

I have strong feelings about `margs`.  I can't decide if they're positive or negative.  

Waiting on Travis but then i think this can go in

gforsyth:
Thanks, @scopatz!

scopatz:
First off, thanks *so much* @gforsyth for giving this an in depth review. I think it was totally necessary.  And thanks a million for merging.  I know that this was kind of a large PR.  I am super happy that this made it in!

> I have strong feelings about margs. I can't decide if they're positive or negative.

What do you mean by `margs`, exactly?  I only found one reference in the code...

gforsyth:
Purely as a linguistic device.  If keyword arguments are `kwargs` then of course macro arguments are `margs`.  But I can't read or type that word without wanting to just say `margs` in a louder than appropriate monotone.

scopatz:
Haha, we are truly blazing new trails through the forest of Self Expression.

PR #1601, opened 2016-08-21
NetBSD support

scopatz:
@krytarowski can you try this out please?

### DIFF
--- a/news/netbsd.rst
+++ b/news/netbsd.rst
@@ -1,0 +1,13 @@
+**Added:**
+
+* NetBSD is now supported.
+
+**Changed:** None
+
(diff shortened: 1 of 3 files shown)


krytarowski:
`ls -G` is FreeBSD specific, we don't have a switch for colorized output

scopatz:
OK thanks for letting me know

scopatz:
Does this look correct?

krytarowski:
It works! Thanks!

scopatz:
Great! This should be ready to go in, then.

gforsyth:
Nice. Thanks @scopatz!

PR #1602, opened 2016-08-21
New implementation of bash completer

### LINKED ISSUE #1595: Bash completion: command not found
I'm using xonsh 0.4.5 on OS X, with bash-completion 1.3 installed with homebrew.

My `$BASH_COMPLETIONS` is set to `['/usr/local/etc/bash_completion']`, but bash completion does not working. So I looked at the code at `completers/bash.py` and tested the following code:

```
bash -c 'source "/usr/local/etc/bash_completion.d/ssh"\nCOMP_WORDS=(ssh M)\nCOMP_LINE=\'ssh M\'\nCOMP_POINT=${#COMP_LINE}\nCOMP_COUNT=6\nCOMP_CWORD=1\n_ssh ssh M ssh\nfor ((i=0;i<${#COMPREPLY[*]};i++)) do echo ${COMPREPLY[i]}; done\n'
/usr/local/etc/bash_completion.d/ssh: line 3: have: command not found
/bin/bash: line 6: _ssh: command not found
```

The `have` command in `bash_completion.d/ssh` is defined in `bash_completion` which is not loaded. I also checked bash-completion 2.3.5 in archlinux, its `completions/ssh` also uses some functions like `_init_completion` in `bash_completion`.

I have `source /usr/local/etc/bash_completion` in my `.bashrc` but it wont load while running bash by `bash -c`.

Did I do anything wrong or it may be a bug? Thanks!

blahgeek:
trying to fix #1595

### DIFF
--- a/news/fix-bash-completions.rst
+++ b/news/fix-bash-completions.rst
@@ -1,13 +1,13 @@
 **Added:** None
 
-**Changed:** None
+**Changed:**
+
+* New implementation of bash completer with better performance and compatibility.
 
 **Deprecated:** None
 
 **Removed:** None
 
-**Fixed:**
-
-* Source `bash_completion` at first on bash completion
+**Fixed:** None
 
 **Security:** None
--- a/news/fix-bash-completions.rst
+++ b/news/fix-bash-completions.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Source `bash_completion` at first on bash completion
+
+**Security:** None
--- a/xonsh/completers/bash.py
+++ b/xonsh/completers/bash.py
@@ -1,101 +1,38 @@
-import os
-import re
 import shlex
-import pickle
-import hashlib
 import pathlib
 import builtins
 import subprocess
 
-import xonsh.lazyasd as xl
 import xonsh.platform as xp
 
 from xonsh.completers.path import _quote_paths
 
-RE_DASHF = xl.LazyObject(lambda: re.compile(r'-F\s+(\w+)'),
-                         globals(), 'RE_DASHF')
-
-INITED = False
-
-BASH_COMPLETE_HASH = None
-BASH_COMPLETE_FUNCS = {}
-BASH_COMPLETE_FILES = {}
-
-CACHED_HASH = None
-CACHED_FUNCS = None
-CACHED_FILES = None
-
-BASH_COMPLETE_SCRIPT = """
-BASH_COMPLETION_DIR="nonexist"
-BASH_COMPLETION_COMPAT_DIR="nonexist"
-{completions_sources}
-source "{filename}"
+BASH_COMPLETE_SCRIPT = r'''
+{sources}
+if (complete -p "{cmd}" 2> /dev/null || echo _minimal) | grep --quiet -e "_minimal"
+then
+    declare -f _completion_loader > /dev/null && _completion_loader "{cmd}"
+fi
+_func=$(complete -p {cmd} | grep -o -e '-F \w\+' | cut -d ' ' -f 2)
 COMP_WORDS=({line})
 COMP_LINE={comp_line}
 COMP_POINT=${{#COMP_LINE}}
 COMP_COUNT={end}
 COMP_CWORD={n}
-{func} {cmd} {prefix} {prev}
+$_func {cmd} {prefix} {prev}
 for ((i=0;i<${{#COMPREPLY[*]}};i++)) do echo ${{COMPREPLY[i]}}; done
-"""
-
-
-def update_bash_completion():
-    global BASH_COMPLETE_FUNCS, BASH_COMPLETE_FILES, BASH_COMPLETE_HASH
-    global CACHED_FUNCS, CACHED_FILES, CACHED_HASH, INITED
-
-    completers = builtins.__xonsh_env__.get('BASH_COMPLETIONS', ())
-    BASH_COMPLETE_HASH = hashlib.md5(repr(completers).encode()).hexdigest()
-
-    datadir = builtins.__xonsh_env__['XONSH_DATA_DIR']
-    cachefname = os.path.join(datadir, 'bash_completion_cache')
-
-    if not INITED:
-        if os.path.isfile(cachefname):
-            # load from cache
-            with open(cachefname, 'rb') as cache:
-                CACHED_HASH, CACHED_FUNCS, CACHED_FILES = pickle.load(cache)
-                BASH_COMPLETE_HASH = CACHED_HASH
-                BASH_COMPLETE_FUNCS = CACHED_FUNCS
-                BASH_COMPLETE_FILES = CACHED_FILES
-        else:
-            # create initial cache
-            _load_bash_complete_funcs()
-            _load_bash_complete_files()
-            CACHED_HASH = BASH_COMPLETE_HASH
-            CACHED_FUNCS = BASH_COMPLETE_FUNCS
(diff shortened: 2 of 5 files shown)


xonsh/completers/bash.py
```diff
@@ -25,6 +25,10 @@
 CACHED_FUNCS = None
 CACHED_FILES = None
 
+BASH_COMPLETE_SCRIPT_PRE = """BASH_COMPLETION_DIR="nonexist"
+BASH_COMPLETION_COMPAT_DIR="nonexist"
+"""
```
scopatz: I think you can just add these to the `BASH_COMPLETE_SCRIPT` below.  No nee to have a 'pre' that is just going to be directly glued in anyways.

scopatz:
Thanks for putting this in @blahgeek!

blahgeek:
PS: I only tested this on my mac with bash-completion 1.3 installed by homebrew which is pretty outdated. This may need further review, especially on other platforms.

news/fix-bash-completions.rst
```diff
@@ -0,0 +1,13 @@
...
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Source `bash_completion` at first on bash completion
```
scopatz: rst requires two backticks.

xonsh/completers/bash.py
```diff
@@ -107,7 +111,8 @@ def complete_from_bash(prefix, line, begidx, endidx, ctx):
     script = BASH_COMPLETE_SCRIPT.format(
         filename=fnme, line=' '.join(shlex.quote(p) for p in splt),
         comp_line=shlex.quote(line), n=n, func=func, cmd=cmd,
-        end=endidx + 1, prefix=prefix, prev=shlex.quote(prev))
+        end=endidx + 1, prefix=prefix, prev=shlex.quote(prev),
+        completions_sources='\n'.join(_collect_completions_sources()))
```
scopatz: I think this is doing too much.  We only need the `bash_completions` file that is associated with (in the directory above) the filename we are trying to source. We don't need / want all of the `bash_completions` specified by the user for the whole system.  This could lead to some weird out-of-order errors, in theory.

scopatz:
Thanks @blahgeek! I'll be sure to test it once we have it working.

blahgeek:
> I think this is doing too much. We only need the bash_completions file that is associated with (in the directory above) the filename we are trying to source.

I think maybe you misunderstood what I'm trying to do here. As mentioned in #1595 , directly sourcing the completion script (e.g., for ssh, `/etc/bash_completion/bash_completions.d/ssh`) would cause an error because that script would use some functions defined in `/etc/bash_completion/bash_completion` (like `have`, `_init_completes`, etc.)

Normally when we use bash, we would add `source /etc/bash_completion/bash_completion` to `.bashrc`, which would find all the completion scripts in `/etc/bash_completion/bash_completions.d/` and loads them (and that's how xonsh find associated completion scripts, by source that file at startup).

So what I am trying to do is to source `/etc/bash_completion/bash_completion` (this is what `$BASH_COMPLETIONS` is, also `_collect_completions_sources()` returns) before sourcing individual completion script, to provide necessary helper functions. And, to prevent it finding and loading every scripts under `bash_completions.d` (which costs about 0.5 seconds on my machine), `BASH_COMPLETION_DIR` is set to a non-exist directory.

It works pretty well on my Macbook running OS X, with bash-completion v1.3. But for bash-completion v2.x (in most linux systems), setting `BASH_COMPLETION_DIR` may not be the right way to do the trick.

Thanks every much for your review, any suggestions are welcome.

gforsyth:
> And, to prevent it finding and loading every scripts under bash_completions.d (which costs about 0.5 seconds on my machine), BASH_COMPLETION_DIR is set to a non-exist directory.

This seems prone to all sorts of confusion in userspace.  I'm not suggesting we reinvent the wheel, but maybe in this case it makes more sense to pull out whatever helper functions are necessary and just include them independently of the system `bash_completion` file

blahgeek:
But those functions are shipped with `bash-completion` package, and may be different in every versions. Including those scripts in xonsh can break things

gforsyth:
I've just looked through the history of `bash_completion` and they haven't made any changes to functionality beyond bug-fixes since 2011.  I imagine they probably aren't fond of making big changes to their API since it would require rewriting all of their completion files.  

I don't think it's a necessity to include it, I just don't want the functionality to depend on the user either paying a load-time penalty or setting some otherwise unneeded envvar to a nonexistent directory.  

That's a fine hack for figuring out what needs to be fixed, but not for actual use.

blahgeek:
I took a look at bash-completion v2.x on my linux box, found some interesting points.

With bash-completion v2.x, the main script is `/usr/share/bash-completion/bash_completion` and the individual scripts is under `/usr/share/bash-completion/completions/`. Still, with `$BASH_COMPLETIONS` set to `/usr/share/bash-completion/bash_completion`, most of the commands cannot be autocompleted. But after I applied this patch (source the main script at first), there's still no luck.

Then I noticed that, with bash-completion v2.x, those individual completion scripts under `completions/` is lazy-loaded. In the main script, a default completion handler `_completion_loader` is registered by `complete -D -F _completion_loader`, which would load completion script on demand. So while xonsh starts up, individual completion scripts can not be found by `complete -p`.

Now the problem is much more easier.

We would only run the following commands if we want to complete `ssh w`:

```
# Source the main script, it runs pretty fast since all other scripts is lazy-loaded
source "/usr/share/bash-completion/bash_completion"
# Load the completion script for `ssh`
_completion_loader ssh
# Get the complete function for ssh
complete -p ssh
```

That would prints `complete -F _ssh ssh`, and finally we could:

```
COMP_WORDS=(ssh w)
COMP_LINE='ssh w'
COMP_POINT=${#COMP_LINE}
COMP_COUNT=6
COMP_CWORD=1
_ssh ssh w ssh
for ((i=0;i<${#COMPREPLY[*]};i++)) do echo ${COMPREPLY[i]}; done
```

Now it seems much more simpler, no `BASH_COMPLETE_FUNCS`, no `BASH_COMPLETE_FILES`, no caches...

This only works for bash-completion v2.x

gforsyth:
Hmm, well, bash-completion 2.x started in June 2012, so it's not completely unreasonable to only support more recent versions.

Similarly, we could check the version listed in `bash_completions` and only offer the new functionality if it's supported

blahgeek:
OK now the code is much more simpler.

It works well with bash-completion v2.x, good performance, individual completion scripts works.

It also works with bash-completion v1.x, just a little bit slow, due to the lack of lazy-loading.

gforsyth:
Whoah!  First impressions are that this is awesome!  As it stands, this obviates the need for my pacman completer since this works automatically.  Will test further but this is great.

xonsh/completers/bash.py
```diff
@@ -1,97 +1,42 @@
...
-BASH_COMPLETE_HASH = None
-BASH_COMPLETE_FUNCS = {}
-BASH_COMPLETE_FILES = {}
-
-CACHED_HASH = None
-CACHED_FUNCS = None
-CACHED_FILES = None
-
-BASH_COMPLETE_SCRIPT = """source "{filename}"
+BASH_COMPLETE_SCRIPT = r'''
```
scopatz: PEP8: triple double quotes instead of triple single quotes https://www.python.org/dev/peps/pep-0008/#string-quotes

scopatz:
Very impressive @blahgeek!  This even feels snappier on my main machine. Just the one cleanup issue from me.  @gforsyth might have more thoughts though...

gforsyth:
Thanks a million, @blahgeek !  This is absolutely awesome to have!

Issue #1521, opened 2016-08-02
Slow startup


jeLee6gi:
I tried xonsh on my 2011 Intel Atom netbook running Archlinux today and noticed the very slow startup times. It always takes between 12 and 15 seconds.

Here is some profiling output (pruned for readability), it seems that xonsh spends at least 10 seconds in time.sleep:

```

         98465 function calls (95969 primitive calls) in 14.570 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     82/1    0.068    0.001   14.579   14.579 {built-in method builtins.exec}
        1    0.000    0.000   14.579   14.579 xonsh:2(<module>)
        1    0.002    0.002   13.697   13.697 __amalgam__.py:16514(main)
        1    0.000    0.000   11.733   11.733 __amalgam__.py:2048(run_code_with_cache)
        1    0.000    0.000   11.683   11.683 __amalgam__.py:1967(compile_code)
        1    0.002    0.002   11.683   11.683 __amalgam__.py:15984(compile)
        1    0.000    0.000   10.449   10.449 __amalgam__.py:15944(parse)
        2    0.013    0.006   10.433    5.217 base.py:287(parse)
        1    0.000    0.000   10.425   10.425 __amalgam__.py:16040(_parse_ctx_free)
      652   10.222    0.016   10.222    0.016 {built-in method time.sleep}
     83/7    0.008    0.000    2.401    0.343 <frozen importlib._bootstrap>:966(_find_and_load)
     83/7    0.004    0.000    2.398    0.343 <frozen importlib._bootstrap>:939(_find_and_load_unlocked)
        1    0.000    0.000    1.963    1.963 __amalgam__.py:16460(premain)
        1    0.000    0.000    1.848    1.848 __amalgam__.py:16230(__init__)
        1    0.000    0.000    1.832    1.832 __amalgam__.py:16292(_init_environ)
        1    0.000    0.000    1.831    1.831 __amalgam__.py:15915(__init__)
        1    0.001    0.001    1.767    1.767 __amalgam__.py:15769(load_builtins)
    103/7    0.002    0.000    1.446    0.207 <frozen importlib._bootstrap>:214(_call_with_frames_removed)
     76/6    0.007    0.000    1.445    0.241 <frozen importlib._bootstrap>:659(_load_unlocked)
     59/5    0.003    0.000    1.440    0.288 <frozen importlib._bootstrap_external>:659(exec_module)
[...]

PR #1603, opened 2016-08-21
yacc_debug loads parser on same thread.

scopatz:
This should address some of this issues in #1521 with package creation.


gforsyth:
Confirmed that this works to include `parser_table.py` on installs on Arch (without the hackery added to the `PKGBUILD`).  Merging in.

gforsyth:
Thanks, @scopatz!

Issue #1134, opened 2016-06-06
problems sourcing zsh functions


aliceriot:
I have a number of functions written for `zsh` that I'd like to avoid reimplementing when switching to xonsh. When I try to use the `source-zsh` function to import a file that includes all of these functions I get this error:

![xonsh-error](https://cloud.githubusercontent.com/assets/6207644/15825915/9083e5f6-2bd3-11e6-8f8b-1b912b5d38c9.png)

(only the first line shown here is relevant).

Am I missing something? There's not much in the docs about `zsh` interop. The file I'm trying to source is here: https://github.com/aliceriot/dotfiles/blob/master/zsh/functions/functions.zsh

gforsyth:
Hey @aliceriot -- thanks for reporting!  Could you post the output of `xonfig` and also try running the `source-zsh` command after setting `$XONSH_SHOW_TRACEBACK=True` and paste the full traceback from the error?  Thanks!

gforsyth:
Mmm, also -- does it work if you comment out all of the `source ...zsh` lines at the beginning?  I wonder if that's tripping it up.

aliceriot:
I get the same error if I `source-zsh` on a file that just defines functions, and doesn't have any `import` statements, e.g: https://github.com/aliceriot/dotfiles/blob/master/zsh/functions/git.zsh

When I try to run `xonfig` I get this:

![xonfig-error](https://cloud.githubusercontent.com/assets/6207644/15826142/61d21fd8-2bd4-11e6-9ac4-628e02ff10f5.png)

setting `$XONSH_SHOW_TRACEBACK` doesn't seem to have any effect, sourcing a `.zsh` file still gives the same error.

Other details: I'm running `xonsh` from `zsh`, which is still set as my login shell (not being able to import existing setup is holding me back b/c I use these functions all day when working). My system is Arch Linux, terminal is xfce4-terminal.

aliceriot:
reinstalled `ply`, now `xonfig` gives me:

```
+------------------+----------------+
| xonsh            | 0.3.2          |
| Python           | 3.5.1          |
| PLY              | 3.7            |
| have readline    | True           |
| prompt toolkit   | 1.0.0          |
| pygments         | 2.1.3          |
| on posix         | True           |
| on linux         | True           |
| distro           | Antergos Linux |
| on darwin        | False          |
| on windows       | False          |
| is superuser     | False          |
| default encoding | utf-8          |
+------------------+----------------+
```

[...]

PR #1605, opened 2016-08-21
Foreign shell function empty filename fix

scopatz:
This partially address #1134


gforsyth:
Thanks, @scopatz!  I hope someone comes up with a new way for zsh to figure out filenames

Issue #1606, opened 2016-08-22
$PWD, and therefore, prompt string doesn't track change in real working directory.


bobhy:
If XONSH invokes a Python function which changes working directory, e.g, by ``os.chdir()``, $PWD and prompt string don't reflect it, leading to confusion all round.  Here's an example:

```xonsh
> xonsh
-------------------------- bobhy@BobHy-PC ~
$ $PWD
'C:\\Users\\bobhy'
-------------------------- bobhy@BobHy-PC ~
$ cmd /c 'cd'
C:\Users\bobhy
-------------------------- bobhy@BobHy-PC ~
$ import os
-------------------------- bobhy@BobHy-PC ~
$ os.getcwd()
'C:\\Users\\bobhy'
-------------------------- bobhy@BobHy-PC ~
$ os.chdir('Documents')  # <-- triggers the bug
-------------------------- bobhy@BobHy-PC ~          # <-- prompt string unaffected
$ $PWD
'C:\\Users\\bobhy'                                                     # <-- as is PWD
-------------------------- bobhy@BobHy-PC ~
$ os.getcwd()
'C:\\Users\\bobhy\\Documents'                                # <-- but working directory did change
-------------------------- bobhy@BobHy-PC ~
$ cmd /c 'cd'
C:\Users\bobhy\Documents
-------------------------- bobhy@BobHy-PC ~

**Comment by scopatz** [comment]
Thanks for reporting @bobhy. This is a subtle issue and I am not sure what can be done about it that doesn't involve excessive querying or monkey patching parts of the standard library.

Also, note that this is a problem *within* the standard library itself. For example, say I have a function which changes the directory from the C-level, `os.environ` and `os.getenv()`  will only see the original, old directory:

```python
>>> os.environ['PWD']
/a
>>> change_dir_from_c('/b')
>>> get_dir_from_c()
/b
>>> os.environ['PWD']
/a
>>> os.getenv['PWD']
/a
```

The safest thing to do from xonsh is to always use `__xonsh_env__` and `cd`.  Unfortunately, even setting `os.environ` to the xonsh env doesn't seem to work:

```console
scopatz@artemis ~ $ os.environ = __xonsh_env__
scopatz@artemis ~ $ os.chdir('/')
scopatz@artemis ~ $ $PWD
'/home/scopatz'
scopatz@artemis ~ $ os.environ['PWD']
'/home/scopatz'
scopatz@artemis ~ $ os.getcwd()
'/'
```

[...]

PR #1608, opened 2016-08-23
Track working directory changes in invoked commands

bobhy:
Proposed fix for the problem raised in #1606.


gforsyth:
Hi @bobhy -- thanks for putting this in!  I'm not a windows user and so will probably punt to one of the windows core devs.  

Before that, though, there seem to be a few different features in here -- could you break them up so there's one proposal per PR?  It makes it much easier to review and approve new features if we're more granular about it.  

Thanks!

bobhy:
Sorry, the intended change was only a few lines in ``base_shell``and a test.  Let me re-do the branch.

But the fix is not windows-specific.  The bug should be easily reproducible on Linux and so should the fix.

melund:
Looks good to me. Thanks a lot this @bobhy

PR #1610, opened 2016-08-23
Add experimental parentheses and bracket completion

rtpg:
This adds an experimental "autopair" functionality to the prompt-toolkit shell
that will insert the right parentheses/bracket when typing left ones (because you'll
probably need it).

 This feature is enabled only when the `XONSH_AUTOPAIR` environment variable is truthy

 I'm unsure of where in the documentation these sorts of features should be documented. Is there a list somewhere?


gforsyth:
Hey @rtpg, cool!  Thanks for putting this in!
 In the future, maybe we can be even smarter with these and automatically overwrite the closing paren/bracket if someone types it (but that's for another day).  

I think that the envvar `XONSH_AUTOPAIR` is a fine name.  Could you add it to the defaults in `environ.py`  (it should default to `False`, I think)?  

See the following spots in the code for how we do this for a similar envvar `AUTO_CD`: https://github.com/xonsh/xonsh/blob/master/xonsh/environ.py#L96
https://github.com/xonsh/xonsh/blob/master/xonsh/environ.py#L239

(Do please keep those lists alphabetical).  

Could you add an entry in the `news` folder?  Instructions are available here: https://github.com/xonsh/xonsh/blob/master/CONTRIBUTING.rst#changelog

gforsyth:
It's also worth considering if we'd like to match `'`s and `"`s (that's my default behavior in emacs, but other people find it pretty annoying)

santagada:
This is great, and I really like all the comments @gforsyth gave except one: I think it should default to True, specially if we support the overwrite behavior. It seems like we want the best possible shell in its default, and that is a pretty neat feature. People annoyed by it can always turn it off once forever, but the risk of people not paying attention to the changelog or who doesn't read all the docs might never see it.

gforsyth:
I agree with @santagada on it defaulting to True _if_ it supports overwrite.

scopatz:
I think this should default to False, if for no other reason that this is the single most infuriating feature (overwrites make it worse) in Jupyter, in my opinion.  Also, I think the usage goes the other way.  If people want it, they are neutral/happy to enable it.   If people don't want this, then they are annoyed that they have to disable it.

All that being said, I *am* very happy to have a feature like this in  xonsh.  I just think that default False is correct.  Of course, feel free to outvote me on this :)

scopatz:
Also, thanks a ton for putting this in @rtpg!

rtpg:
I think that defaulting it as `False` is the right solution until this is more functional (for example with overwriting). There can be a bit of frustration with this without overwriting, but some people would use this as-is.

 I'm a little reluctant to autopair `'` right now, because of things like inline git commit messages requiring "normal" parentheses (`git commit -m "There's a snake in my boot!"`) . _But_ I like the idea of autopairing `"` so I'll put in both.

 I'll add in the suggestions.

rtpg:
OK I've been using this for a couple of hours on my local machine, and without overwriting it does get pretty frustrating, so I'll look into adding that here if it's feasible later tonight

santagada:
Nice, as a second option we could create a page on the docs with "cool features to try" were we can point to nice things in xonsh not enabled by default or that need some other software, like bash completion, prompt-toolkit, this and other stuff like xontribs.

scopatz:
That is an awesome idea, @santagada

rtpg:
Alrighty, I added overwriting + wrote a little `VarDoc` that _should_ show up on the [Environment Variable](http://xon.sh/envvars.html) page later.

  I like @santagada's idea, people love new features and messing with settings after all! In the future, having a little setup wizard might not be bad either.

gforsyth:
Hey @rtpg -- sorry for the delay on this, I've been out of the world for the past 10 days or so.  

Can you merge from master?  We switched our Linux tests to CircleCI and this branch doesn't have the `circle.yml` file which is why they're failing

laerus:
thanks for this @rtpg! i've played around with it and it's very cool!

PR #1611, opened 2016-08-23
Option to turn off completion brackets

### LINKED ISSUE #1508: sys.stdin tab completes to sys.stdin[
`sys.stdin` tab completes to `sys.stdin[`. I don't see why you would ever want to automatically complete a square bracket, but for sys.stdin it especially doesn't make sense, since it isn't indexable.

gforsyth:
Add `$COMPLETIONS_BRACKETS` (default True) to determine if the python
completer should add opening parens and square brackets to its
completions.

Whether or not `$COMPLETION_BRACKETS` is True, `sys.stdin`, `sys.stdout`
and `sys.stderr` do not have brackets in their completions.

Resolves #1508


xonsh/environ.py
```diff
@@ -244,6 +245,7 @@ def DEFAULT_VALUES():
         'CDPATH': (),
         'COLOR_INPUT': True,
         'COLOR_RESULTS': True,
+        'COMPLETIONS_BRACKETS': True,
```
scopatz: Can you also add a VarDocs entry for this, please too?

scopatz:
Thanks for this @gforsyth!  Just the one minor issue.  Probably long overdue itself :)

gforsyth:
Hey @scopatz -- ready for another look!

scopatz:
Looks great! Just waiting for CI

laerus:
`/home/travis/build/xonsh/xonsh/xonsh/completers/python.py:99:25: E127 continuation line over-indented for visual indent` (can we turn this off for everything?)

also MacOS builds seem to take very long.

xonsh/completers/python.py
```diff
@@ -95,10 +95,14 @@ def attr_complete(prefix, ctx, filter_func):
...
         a = getattr(val, opt)
-        if callable(a):
-            rpl = opt + '('
-        elif isinstance(a, abc.Iterable):
-            rpl = opt + '['
+        if (builtins.__xonsh_env__['COMPLETIONS_BRACKETS'] and
+                        opt not in ['stdin', 'stdout', 'stderr']):
+            if callable(a):
+                rpl = opt + '('
+            elif isinstance(a, abc.Iterable):
```
scopatz: Actually, now that I think about it, this should `isinstance(a, (abc.Sequence, abc.Mapping))` and the special case file handles above should be removed.  We only want to add an `[` when there is a `__getitem__`.  As a counter example, this implementation will still add an `[` to sets, file handles, and other iterable but non-indexable  objects.

scopatz:
> also MacOS builds seem to take very long.

Yeah, Mac has been slow all day.  I noticed it on conda-forge ~18 hrs ago

scopatz:
Thanks a million @gforsyth!

Issue #1612, opened 2016-08-23
history not working


SanketDG:
I get the following while running history:

```
Did you mean one of the following?
    cc:  Command (/usr/bin/cc)
    w:   Command (/usr/bin/w)
    nc:  Command (/bin/nc)
    9:   Alias
    [:   Command (/usr/bin/[)

The program 'fc' is currently not installed. You can install it by typing:
sudo apt-get install f2c
```

`f2c` is a fortran to c compiler?

gforsyth:
Hey @SanketDG -- I believe this is a duplicate of #1577.  Do you use `zsh`?  They alias history to `fc` which is a builtin.  There is a workaround here: https://github.com/xonsh/xonsh/issues/1577#issuecomment-240600295

gforsyth:
And thanks for reporting!

SanketDG:
Uh I see, sorry for not doing  a quick search.

Although, why is `history` overrided? IMO it should be in a `ignore_list` where nothing can override it.

gforsyth:
Agreed.  I think @laerus is discussing just that in #1577

laerus:
also #1155, means it's time we go about fixing this. PRs are welcomed!

SanketDG:
@laerus Is this fixed? I still get the same thing on the latest master.

laerus:
i can't reproduce this anymore, when starting xonsh i get:
`aliases: alias 'history' of shell 'zsh' tries to override xonsh alias, xonsh wins!`
then `history` is the builtin alias. 

can you post the output of `xonfig` and `which -v history` please?

SanketDG:
> i can't reproduce this anymore, when starting xonsh i get:
> aliases: alias 'history' of shell 'zsh' tries to override xonsh alias, xonsh wins!
> then history is the builtin alias.

```
+------------------+-----------------+
| xonsh            | 0.4.5           |
| Git SHA          | 62cb089         |
| Python           | 3.4.3           |
| PLY              | 3.7             |
| have readline    | True            |
| prompt toolkit   | 1.0.3           |
| shell type       | prompt_toolkit  |
| pygments         | 2.1.3           |
| on posix         | True            |
| on linux         | True            |
| distro           | LinuxMint       |
| on darwin        | False           |
| on windows       | False           |
| on cygwin        | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
[...]

Issue #1155, opened 2016-06-07
`history` command shows 'fc' is not installed


metalseargolid:
When I run the `history` command, I get the following error:

```
$ fc
xonsh: subprocess mode: command not found: fc
Did you mean one of the following?
    cc:  Command (/usr/bin/cc)
    6:   Alias
    bc:  Command (/usr/bin/bc)
    cp:  Command (/bin/cp)
    gc:  Alias

The program 'fc' is currently not installed. You can install it by typing:
```

`fc` is a shell built-in AFAIK though, so I'm not sure what the problem is. I am running ('xonsh/0.3.3',). `fc` works when invoking from bash or zsh directly.

mitnk:
@metalseargolid Thanks for reporting. From what you pasted, it seems you were running `fc` not `history`. Could you have a confirm and also provide the output of `xonfig`?

here's mine:
```
$ fc
xonsh: subprocess mode: command not found: fc
Did you mean one of the following?
    cc:   Command (/usr/bin/cc)
    nc:   Command (/bin/nc)
    ucf:  Command (/usr/bin/ucf)
    cd:   Alias
    [:    Command (/usr/bin/[)
pi@pi1 ~$
pi@pi1 ~$ history
 0  which pc
 1  which fc
 2  fc
pi@pi1 ~$ xonfig
+------------------+----------+
| xonsh            | 0.3.3    |
| Python           | 3.5.1    |
| PLY              | 3.7      |
| have readline    | True     |
| prompt toolkit   | None     |
| shell type       | readline |
| pygments         | None     |
| on posix         | True     |
| on linux         | True     |
| distro           | debian   |
| on darwin        | False    |
| on windows       | False    |
| is superuser     | False    |
| default encoding | utf-8    |
+------------------+----------+
pi@pi1 ~$
```

metalseargolid:
Ahhhhhh I set xonsh up to import my .zshrc, and it seems ZSH (or oh-my-zsh during setup) aliases the `history` command to `fc -l 1`. I should be able to fix it in my .zshrc when I get home like so:

```
if [[ $SHELL == "/usr/local/bin/xonsh" ]]; then
   unalias history
fi
```

scopatz:
Note that `$SHELL` is your default login shell, though, not the shell you are running with.

laerus:
closing this, feel free to reopen!

mandarvaze:
[...]

PR #1616, opened 2016-08-24
ignore foreign aliases

laerus:
that match xonsh bultin aliases

#1612 #1155

### DIFF
--- a/xonsh/foreign_shells.py
+++ b/xonsh/foreign_shells.py
@@ -12,7 +12,7 @@ import warnings
 import functools
 import collections.abc as abc
 
-from xonsh.lazyasd import LazyObject
+from xonsh.lazyasd import lazyobject
 from xonsh.tools import to_bool, ensure_string
 from xonsh.platform import ON_WINDOWS, ON_CYGWIN
 
@@ -80,7 +80,9 @@ fi
 echo ${namefile}"""
 
 # mapping of shell name alises to keys in other lookup dictionaries.
-CANON_SHELL_NAMES = LazyObject(lambda: {
+@lazyobject
+def CANON_SHELL_NAMES():
+    return {
     'bash': 'bash',
     '/bin/bash': 'bash',
     'zsh': 'zsh',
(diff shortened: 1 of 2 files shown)


xonsh/foreign_shells.py
```diff
@@ -540,9 +541,20 @@ def load_foreign_aliases(shells=None, config=None, issue_warning=True):
...
-            aliases.update(shaliases)
+        if not shaliases:
+            continue
+        for alias in list(shaliases.keys()):
+            if alias not in xonsh_aliases:
+                continue
+            else:
+                del shaliases[alias]
+                print('aliases: alias {!r} of shell {!r} '
+                      'tries to override xonsh builtin alias, '
```
scopatz: These aren't only just builtin aliases.  They are just any pre-existing aliases.

xonsh/foreign_shells.py
```diff
@@ -540,9 +541,20 @@ def load_foreign_aliases(shells=None, config=None, issue_warning=True):
...
+                      'xonsh wins!'.format(alias, shell['shell']),
+                      file=sys.stderr)
```
scopatz: Also, couldn't this be:

```python
for alias in set(shaliases) & set(xonsh_aliases):
    del shaliases[alias]
    print(....)
```

It seems like this is a set operation, so maybe should be expressed as one.  Just a thought!

scopatz:
Thanks a ton @laerus! Just a couple of comments in-line!

xonsh/foreign_shells.py
laerus: nice!

xonsh/foreign_shells.py
laerus: right!

xonsh/foreign_shells.py
```diff
@@ -540,9 +541,17 @@ def load_foreign_aliases(shells=None, config=None, issue_warning=True):
...
     shells = _get_shells(shells=shells, config=config, issue_warning=issue_warning)
     aliases = {}
+    xonsh_aliases = builtins.aliases
     for shell in shells:
         shell = ensure_shell(shell)
         _, shaliases = foreign_shell_data(**shell)
-        if shaliases:
-            aliases.update(shaliases)
+        if not shaliases:
+            continue
```
scopatz: You don't need these the above two lines anymore either because `set(shaliases) == set()` and so the loop will be empty.

scopatz: This doesn't block the PR, just an opportunity to get rid of a couple more lines :)

laerus: Always good to remove useless lines!

scopatz:
Amazing! Thanks so much!

laerus:
thanks for the intersection hint!

PR #1617, opened 2016-08-24
New option to use <Enter> to confirm completion instead of running command

blahgeek:
Like fish shell, while completion menu is displayed, pressing `<Enter>` do not run the command.

### DIFF
--- a/xonsh/ptk/key_bindings.py
+++ b/xonsh/ptk/key_bindings.py
@@ -107,6 +107,15 @@ class EndOfLine(Filter):
         return bool(at_end and not last_line)
 
 
+class ShouldConfirmCompletion(Filter):
+    """
+    Check if completion needs confirmation
+    """
+    def __call__(self, cli):
+        return (builtins.__xonsh_env__.get('COMPLETION_CONFIRM', True)
+                and bool(cli.current_buffer.complete_state))
+
+
 # Copied from prompt-toolkit's key_binding/bindings/basic.py
 @Condition
 def ctrl_d_condition(cli):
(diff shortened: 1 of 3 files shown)


xonsh/ptk/key_bindings.py
```diff
@@ -107,6 +107,15 @@ def __call__(self, cli):
         return bool(at_end and not last_line)
 
 
+class ShouldConfirmCompletion(Filter):
+    """
+    Check if completion needs confirmation
+    """
+    def __call__(self, cli):
+        return (builtins.__xonsh_env__.get('COMPLETION_CONFIRM', False)
```
scopatz: You don't need to provide False here, since it is in the `Env` defaults already.  In general, we try to define the default only in one place.

xonsh/ptk/key_bindings.py
```diff
@@ -107,6 +107,15 @@ def __call__(self, cli):
...
+                and bool(cli.current_buffer.complete_state))
```
scopatz: Isn't the call to bool implicit after the `and`?

scopatz:
Thanks a million for this @blahgeek!  Much appreciated!

gforsyth:
Also looking good, @blahgeek!  Could you also add this to `DEFAULT_ENSURERS` on or about here: https://github.com/xonsh/xonsh/blob/master/xonsh/environ.py#L104 ?

Thanks again!

scopatz:
Looks like there is a merge conflict on this branch now....

scopatz:
Hi @blahgeek - would you mind merging from master please in order to get Circle passing?  Thanks!

PR #1623, opened 2016-08-25
more expand vars

### LINKED ISSUE #1615: Environment variable expansion
It is unclear to me whether this was a design choice or an oversight but current versions of xonsh require `ls "$FOO/bar"` whereas bash expands `ls $FOO/bar` naturally.

scopatz:
This should fix #1615.  It also makes the parser a bit simpler.

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1641,6 +1641,9 @@ def test_uncaptured_sub():
 def test_hiddenobj_sub():
     check_xonsh_ast({}, '![ls]', False)
 
+def test_slash_envarv_echo():
+    check_xonsh_ast({}, '![echo $HOME/place]', False)
+
 def test_bang_two_cmds_one_pipe():
     check_xonsh_ast({}, '!(ls | grep wakka)', False)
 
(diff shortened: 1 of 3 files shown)


tests/test_parser.py
```diff
@@ -1641,6 +1641,9 @@ def test_uncaptured_sub():
 def test_hiddenobj_sub():
     check_xonsh_ast({}, '![ls]', False)
 
+def test_slash_envarv_echo():
+    check_xonsh_ast({}, '![echo $HOME/place]', False)
```
astronouth7303: Does this actually verify the AST, or does it just test that it's parsable?

scopatz: When `run=False`, as is passed in here, this checks that it is parseable and that the AST can be compiled into bytecode. It skips the actual execution.  In this case, since the original issue was actually a parser bug, I think this is enough.

laerus:
this fixes it, thanks @scopatz!

scopatz:
Great!

scopatz:
Sweet, great to have this in. It was really bothering me!

PR #1624, opened 2016-08-25
more tilde expansions

scopatz:
It has been bothering me recently that tildes weren't expanded inside of expresions, like in Bash. I didn't know what was causing it. However, it is a pretty simple fix. Just a few lines.  You can read about the Bash semantics that I tried to match here: https://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html


scopatz:
I am a little confused why circle is running here since we haven't merged in the yaml file.

gforsyth:
Hey @scopatz -- so CircleCI seems to have this mantra of "it just works!" and they try to automagically determine how to run tests, etc...
This... doesn't work, so any pushes without the `circle.yml` file will still run in circle, but will fail.  
I didn't realize this would happen -- it's pretty annoying.  We can merge in the circle CI stuff with the yaml file, or I can also revoke the deploy key so they will stop.

scopatz:
Ahh ok, I will just gao ahead and merge in the corcle pr then

scopatz:
Merged in from master to pick up the tilde expansion.

scopatz:
Ping to whoever can merge it since @gforsyth is now AFK

laerus:
seems like nightly fails somewhere on `test_ast.py`

scopatz:
Ok seems to be working now

scopatz:
Thanks!

PR #1626, opened 2016-08-25
add a 'kill' alias to use jobs indexes

rbignon:
The zsh jobs management allow user to call a 'kill' builtin command with
'%' + index of the job, to kill a given job. This patch introduces the
same behavior in xonsh.

Unfortunately, as it calls the system 'kill' command, which doesn't exist
under Windows, I've added this alias only for unix systems.

A better way would be to re-implement the kill binary, and calling the
os.kill() function.

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -12,7 +12,7 @@ from xonsh.lazyasd import lazyobject
 from xonsh.dirstack import cd, pushd, popd, dirs, _get_cwd
 from xonsh.environ import locate_binary
 from xonsh.foreign_shells import foreign_shell_data
-from xonsh.jobs import jobs, fg, bg, clean_jobs
+from xonsh.jobs import jobs, fg, bg, clean_jobs, kill
 from xonsh.history import history_main
 from xonsh.platform import (ON_ANACONDA, ON_DARWIN, ON_WINDOWS, ON_FREEBSD,
                             ON_NETBSD, scandir)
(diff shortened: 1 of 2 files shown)


laerus:
Hey @rbignon, thanks for putting this in! 

Although i think we should not use a xonsh alias as
a wrapper around a pre-existing unix command for this.
We had a similar discussion before and the idea was to
include an optional argument or a subcommand to the `jobs` alias.
something like `jobs kill` or `jobs -k` that deals only with xonsh jobs.

scopatz:
Hi, @rbignon,. I agree with you and @laerus. I think a wrapper is a bad idea. Either a reimplementation in xoreutils or a different command is preferable here.

rbignon:
Ah I haven't see the xoreutils. By the way I think it's better to have our own implementation of kill which accepts jobs IDs parameter, firstly because it is a common feature in zsh and bash, then it is an opportunity for windows users to have a 'kill' command.

So, if I move my code to xutils, and write a custom kill command which uses os.kill, is it ok for you?

scopatz:
@rbignon - yep that would be perfectly acceptable!

scopatz:
Hey @rbignon - any interest in cleaning this up? I am happy to help guide this.

gforsyth:

mqualizz:
Is there still interest in having this implemented, or has the issue been solved in a different way?

scopatz:
@mqualizz - yes I think that there is still interest!

PR #1629, opened 2016-08-25
Adding prompt_ret_code to xontribs

Siecje:


scopatz:
Thanks a million @Siecje!

PR #1630, opened 2016-08-25
Enshorten attr comp

### LINKED ISSUE #937: Autocomplete: don't display full path
Using  `prompt_kit`, when completing a path such as `/var/log/<TAB>`, the autocompletion menu shows fully prefixed entries. The problem is that when the path is very deep, the autocomplete panel starts to give unreadable results (see attached screenshot).

The proposed solution is to only display the `os.path.basename` of the autocompleted path, suffixed by `/` if it is a directory.

screenshot from 2016-05-19 10-33-09

gforsyth:
Resolves #937 

This was bugging me.  This is only a problem with prompt_toolkit.

Before:
2016-08-25-163033_747x151_escrotum

Now:
2016-08-25-163322_724x153_escrotum

That ghost space isn't actually present... some weird artifact from interaction with wcwidth and scrot, I think.

And enshorten is a perfectly xromulent word.

### DIFF
--- a/xonsh/completers/python.py
+++ b/xonsh/completers/python.py
@@ -106,7 +106,8 @@ def attr_complete(prefix, ctx, filter_func):
             rpl = opt
         # note that prefix[:prelen-len(attr)] != prefix[:-len(attr)]
         # when len(attr) == 0.
-        comp = prefix[:prelen - len(attr)] + rpl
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks a short ton @gforsyth!

PR #1560, opened 2016-08-14
exec

scopatz:
This should resolve all of the pending exec issues.

PR #1633, opened 2016-08-26
correctly preserve parameters

rbignon:
In case there are arguments containing spaces, there were splited. For
example:

$ xonsh -c 'echo 123'
['/home/rom1/python/xonsh/__main__.py', '-c', 'echo', '123']

$

With this patch:

$ xonsh -c 'echo 123'
['/home/rom1/python/xonsh/__main__.py', '-c', 'echo 123']
123
$

### DIFF
--- a/news/pr-params.rst
+++ b/news/pr-params.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Correctly preserve arguments given to xon.sh, in case there are quoted ones.
+
+**Security:** None
--- a/scripts/xon.sh
+++ b/scripts/xon.sh
@@ -7,4 +7,4 @@ if [ -z "${LC_ALL+x}" ] && [ -z "${LC_CTYPE+x}" ] && \
 fi
 
 # run python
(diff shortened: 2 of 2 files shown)


rbignon:
It also fixes my problem on #1560.

scopatz:
Thanks for putting this in! I don't think that this is correct behavior for all cases, though, such as with other command line options. I am on my phone and cannot test it right now

scopatz:
though I could be wrong....

astronouth7303:
I'm pretty sure you're right. I think this will cause all options to be conjoined into a single argument.

scopatz:
... except that `"$@"` actually has a special meaning distinct from all other double quotes and all other dollar sign special meanings : https://www.gnu.org/software/bash/manual/html_node/Special-Parameters.html#Special-Parameters

scopatz:
OK, I have confirmed locally that this is the correct solution.  Thanks so much @rbignon!  Would you please add a news entry about this: http://xon.sh/devguide.html#changelog

scopatz:
ping on this @rbignon!

scopatz:
Also, would you mind merging in from master here, please, to pick up a hotfix to the test suite.  Thanks!

scopatz:
Thanks @rbignon!

PR #1634, opened 2016-08-26
remove pacman completion xontrib

gforsyth:
This xontrib is no longer needed thanks to improvements in
bash-completion imports.  Yay!  One less thing to maintain!


scopatz:
Sweet!

Issue #1194, opened 2016-06-10
`__file__` not defined when importing a `.xsh` file


willsALMANJ:
Creating a file `test.xsh` with content

```
print(__file__)
```

and then importing it with `import test` gives me `NameError: name '__file__' is not defined` whereas as mentioned in #1169 `__file__` should be set to the path to `test.xsh`.

scopatz:
Thanks for reporting @willsALMANJ! Looks like there is more work to do on the import hooks.

laerus:
its seem that `get_filename()` method in `XonshImportHook` isn't being called when importing a .xsh script.

PR #1636, opened 2016-08-27
xonsh module __file__ attribute

laerus:
#1194

### DIFF
--- a/xonsh/imphooks.py
+++ b/xonsh/imphooks.py
@@ -3,15 +3,19 @@
 
 This module registers the hooks it defines when it is imported.
 """
+import builtins
 import os
 import sys
-import builtins
+import types
 from importlib.machinery import ModuleSpec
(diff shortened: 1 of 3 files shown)


xonsh/imphooks.py
```diff
@@ -60,6 +64,16 @@ def find_spec(self, fullname, path, target=None):
     #
     # SourceLoader methods
     #
+    def create_module(self, spec):
+        """Create a xonsh module with the appropriate attributes."""
+        name = spec.name
+        parent = spec.parent
+        mod = XonshModule(name)
```
scopatz: Do we need a special module type? Can this line be `mod = types.ModuleType(name)`?

scopatz:
Thanks for putting this in @laerus!  I think it would a good idea to have a couple of tests related to this too.

scopatz:
Thanks!

PR #1637, opened 2016-08-27
More enhancement on bash completion

blahgeek:


scopatz:
Thanks for putting this in @blahgeek! Looks like we are getting some test failures on nightly.

xonsh/completers/bash.py
```diff
@@ -23,7 +25,6 @@
 for ((i=0;i<${{#COMPREPLY[*]}};i++)) do echo ${{COMPREPLY[i]}}; done
 """
```
gforsyth: Flake8 wants two empty lines here

adqm:
Thanks for the PR, @blahgeek!  I think there might be a cleaner fix here, though; rather than using `_quote_paths` directly, I think it might be cleaner to split `_quote_paths` into two pieces: one that does the adding of slashes/spaces and one that does the quoting (if necessary).

Then, I think we always want to call the "quote if necessary" function on the results of bash completions, but not the "add trailing slashes/spaces" function (since bash presumably takes care of that already).

Really, if we do this, we can make the top-level `Completer` handle the quoting (everything should be run through that function, regardless of where it came from), rather than requiring each completer to handle quoting itself.

blahgeek:
@scopatz I think this PR is ready to merge. We can maybe think about implementing an improved completion API as discussed in #1643 later. This PR is just some bug fixing for the new bash completion.

scopatz:
Thanks @blahgeek!

PR #1639, opened 2016-08-27
fix for envvar expansion completion in ptk

gforsyth:
I broke this in #1630 -- this is a quick fix to restore broken functionality.  Ideally I'd prefer to have path completion behave the same way as attribute completion and hide common prefixes in the dropdown.

### DIFF
--- a/xonsh/ptk/completer.py
+++ b/xonsh/ptk/completer.py
@@ -40,12 +40,17 @@ class PromptToolkitCompleter(Completer):
                     pass
                 elif len(os.path.commonprefix(completions)) <= len(prefix):
                     self.reserve_space()
-                prefix, _, compprefix = prefix.rpartition('.')
-                for comp in completions:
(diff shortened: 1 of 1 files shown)


blahgeek:
Hi @gforsyth , I've found several completion bugs in ptk that maybe related to your work (in current master, not including this PR):

- Filenames with dot: 

```
$ touch a.b.txt a.c.txt
$ ls a.<TAB>
```

- Filenames with spaces:

```
$ mkdir 'a and b'
$ touch 'a and b/test.txt'
$ ls 'a and b'<TAB>
```

In both cases, `complete_path` function in `xonsh.completers.path` returns the correct value (both the string and offset value), I think it maybe a bug in `ptk.completer`

scopatz:
Thanks @gforsyth! @blahgeek thanks for reporting.  Happy to keep merging in PR fixes from anyone :)

Issue #1632, opened 2016-08-26
Improve BASHward compatibility : history interaction


mandarvaze:
This was discussed over twitter with @scopatz [here](https://twitter.com/mandarvaze/status/764887040785735680) where he suggested a separate "bashism" xontrib to handle these

It seems this was discussed on the [google group](https://groups.google.com/d/msg/xonsh/wEs0zI2dHm8/gAS9DuQGBQAJ) and IRC as well (link not available)

Complete details of BASH history interaction feature listed [here](https://www.gnu.org/software/bash/manual/bashref.html#History-Interaction)  - as a reference.

My preference would be to support `!$` first,  since `!!` (to recall last command) already works.

laerus:
Hey! see also #1411

mandarvaze:
@laerus #1411 seems to have a lot more discussion/details
Feel free to close this as duplicate.

laerus:
We haven't discussed about how a xontrib will provide this functionality. Maybe a pre-parsing of some short. #1411 is about the History API now, which may help in the implementation a bashwards xontrib.

scopatz:
I don't think any preparsing is needed:

```console
scopatz@artemis ~/xonsh master $ aliases['!$'] = lambda a, s: 'banana\n'
scopatz@artemis ~/xonsh master $ !$
banana
```

laerus:
will this work like so: `echo !$` ?

scopatz:
Oh, no.  That won't work.  You'd have to do something like `echo @(__xonsh_history__.inps[-1])` right now

scopatz:
Maybe such a preprocessor would be a a good precmd handler (in a xontrib) once we get the event system.  It seems like this kind of expansion would only be useful in the shell, and not in scripts or modules so much anyway.

astronouth7303:
Maybe subprocess-mode macros would help?

scopatz:
Subprocess macros probably mess some of this up.  I think that this is really more a precmd  preprocessor kind of thing.

PaulReiber:
What would be initial goals for recognizers for history preprocessing?
Just !$ or other history subs as well? For example:

$ cd /ect/
cd: /ect/: No such file or directory
$ ^ct^tc

cd /etc

$ pwd
/etc


The above shows a common use case we could call "^wrong^right"
I guess the question is... are we aiming for full BASH level history
substitution?

Kind regards,


*Paul Reiber*http://about.me/reiber

[...]

PR #1640, opened 2016-08-27
History indexing API

laerus:
History object supports indexing with some rules for extra functionality:
* index must be one of string, int, slice or tuple of length two
* an int returns the appropriate command in order
* a string returns the most recent command that contains the string
* a slice returns a list of commands
* if the index is a tuple:
    * the first item follows the previous
              three rules.
    * the second item is the slice of the arguments to be returned
 
see also bash history event and word designators
#1411 #1632

equivalents:

bash  | xonsh
-------- | --------
`!$`      | `__xonsh_history__[-1, -1]`
`!?cmd?:$` | `__xonsh_history__['cmd', -1]`
`!4:2-5` | `__xonsh_history__[4, 2:5]`

more functionality can be added to support more bash history idioms like
`^word^replace`.


scopatz:
Looks like there is a flake8 error...

tests/test_history.py
```diff
@@ -185,3 +185,19 @@ def test_parser_show(args, exp):
               'timestamp': False}
     ns = _hist_parse_args(shlex.split(args))
     assert ns.__dict__ == exp_ns
+
+
+# CMDS = ['ls', 'cat hello kitty', 'abc', 'def', 'touch me', 'grep from me']
```
scopatz: Remove comment

xonsh/history.py
```diff
@@ -605,16 +615,43 @@ def flush(self, at_exit=False):
         self.buffer.clear()
         return hf
 
-    def show(self, *args, **kwargs):
-        """Return shell history as a list
+    def _get(self):
```
scopatz: This seems like an `__iter__` rather than a get.

scopatz:
Thanks for putting this in @laerus!

laerus:
still some minor typos on news, will fix that.

i was thinking about the `^replace^word` bashism and how i can add it in the indexing API,
would something like this `__xonsh_history__[-1, 'replace' : 'word']` be clear and intuitive?

gforsyth:
> would something like this `__xonsh_history__[-1, 'replace' : 'word']` be clear and intuitive? 

how about `__xonsh_history__[-1].replace('old', 'new')` ?

laerus:
@gforsyth yea it's an overkill mostly, just a little less verbose and it will also handle a list of commands. i'm fine with not implementing this, python is just an impulse to be creative.

scopatz:
I agree with @gforsyth that replacement isn't exactly indexing and probably shouldn't be forced into that role.  The replacement bashism can still be implemented in a bashism xontrib.  In light of that, this looks good to go to me.  Any reason to delay merging this @laerus?

laerus:
Maybe i should make `_cmd_filter` and `_args_filter` staticmethods? other than this no reason.

scopatz:
Sure thing, if you want.  Just let me know when it is ready

laerus:
this is ready!

scopatz:
Looks good, thanks!

laerus:
Thank you!

PR #1642, opened 2016-08-27
Add `on_chdir` Event

astronouth7303:
Fire an event when the current directory changes for any reason.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -39,7 +39,8 @@ def xonsh_builtins():
     builtins.execx = None
     builtins.compilex = None
     builtins.aliases = {}
-    # Unlike all the other stuff, this has to refer to the "real" one because all modules that would be firing events on the global instance.
+    # Unlike all the other stuff, this has to refer to the "real" one because all modules that would
+    # be firing events on the global instance.
     builtins.events = events
     yield builtins
     del builtins.__xonsh_env__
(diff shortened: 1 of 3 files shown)


tests/conftest.py
```diff
@@ -38,6 +39,8 @@ def xonsh_builtins():
     builtins.execx = None
     builtins.compilex = None
     builtins.aliases = {}
+    # Unlike all the other stuff, this has to refer to the "real" one because all modules that would be firing events on the global instance.
```
scopatz: PEP8 line too long

astronouth7303: Oh, right. It doesn't flake on the tests themselves.

scopatz:
Thanks @astronouth7303!  This looks good to me code wise.

I think that the docs need to have a page that contains a listing of all (core supported) event categories, their interfaces, etc.  This is so the people xontrib developers and users know how to hook in.  Given how the event docs side of things work, I think this can be autogenerated potentially.  We do something like this in `docs/conf.py` for envvars and for xontribs, if you are looking to model off of something.

astronouth7303:
Yeah, I was planning to do that in the events tutorial maybe, although I suppose a separate page wouldn't hurt.

(I was also thinking of doing a typo test, that would fail if there were any tests without docstrings.)

scopatz:
> Yeah, I was planning to do that in the events tutorial maybe, although I suppose a separate page wouldn't hurt.

I think it is better to have this kind of reference page be separate from the tutorial.

> (I was also thinking of doing a typo test, that would fail if there were any events without docstrings.)

oooh! I like that idea.

astronouth7303:
Done and done, in their own areas.

scopatz:
Looks great to me, thanks @astronouth7303!

PR #1643, opened 2016-08-27
Better ptk completion display

gforsyth:
Ok, fingers crossed, but I think I finally cracked this.  @blahgeek, could you take a look at this and see if it covers the corner cases you mentioned in #1639 

Quick summary:
`numpy.<TAB>` will show `abs(, absolute(...`
`cd xon` will show (for example) `xonsh/, xonda/`
`cd git/xon` will show (as above) `xonsh/, xonda/`

file/folder names with dots and spaces seem to work well.  Note that the quotes around file/folders will spaces won't be shown in the completion but will appear when you select that completion.

### DIFF
--- a/news/ptk-completion-display.rst
+++ b/news/ptk-completion-display.rst
@@ -2,7 +2,7 @@
 
 **Changed:** 
 
-* `prompt_toolkit` completions now only show the rightmost portion
+* ``prompt_toolkit`` completions now only show the rightmost portion
   of a given completion in the dropdown
 
 **Deprecated:** None
--- a/news/ptk-completion-display.rst
+++ b/news/ptk-completion-display.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** 
+
+* `prompt_toolkit` completions now only show the rightmost portion
+  of a given completion in the dropdown
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/xonsh/ptk/completer.py
+++ b/xonsh/ptk/completer.py
@@ -43,6 +43,8 @@ class PromptToolkitCompleter(Completer):
                 c_prefix = _commonprefix([a.strip('\'/').rsplit('/', 1)[0]
                                           for a in completions])
                 for comp in completions:
+                    if comp.endswith('/') and not c_prefix.startswith('/'):
+                        c_prefix = ''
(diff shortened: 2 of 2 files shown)


news/ptk-completion-display.rst
```diff
@@ -0,0 +1,14 @@
+**Added:** None
+
+**Changed:** 
+
+* `prompt_toolkit` completions now only show the rightmost portion
```
scopatz: double backticks

scopatz:
Thanks @gforsyth! This looks good to me, but I'll let @blahgeek chime in.

blahgeek:
Thanks @gforsyth !

I tried it with several edge cases I can think of, the completion results are as great, but the displayed text may still needs some tweaks, e.g.:

- `git sta<TAB>` menu shows `tus` and `ge`, I think we'd expect `status` and `stage`?
- `ls ~/Document/a<TAB>` menu displays prefix (which should not, like all other paths starts with `/`)

Beyond this PR, generally, I think it may not be the best idea to shorten/convert the completion results in `PromptToolkitCompleter`, as it does not have the information whether the completion result is a path, or a command, or whatever it is. As @adqm mentioned in #1637, we can make the top-level Completer handle the quoting, shortening, adding spaces, etc. (maybe with type information returned from complete functions).

gforsyth:
> Beyond this PR, generally, I think it may not be the best idea to shorten/convert the completion results in PromptToolkitCompleter, as it does not have the information whether the completion result is a path, or a command, or whatever it is. As @adqm mentioned in #1637, we can make the top-level Completer handle the quoting, shortening, adding spaces, etc. (maybe with type information returned from complete functions).

I agree that handling these things with type information would be a lot easier.  What's bothering me is that readline seems to handle this without trouble and is receiving the same completions as the ptk shell.

gforsyth:
I'm ok closing this out or merging it with fixes to come later.  I'm going to be AWOL for the next 10 days or so, so I'll leave it to everyone else.  

I'll put in my vote for a type-aware display system.

scopatz:
So @blahgeek - do you think that this OK to merge for the moment?  Happy to so, if it won't be too frustrating, and then we can build on top of it....

blahgeek:
@scopatz Sure, as this PR provides much better ptk completion display than before and does not actually break anything, I think it's good to merge. As for the several small issues I mentioned above, I will see what I can do later :P

scopatz:
Cool Thanks!

PR #1644, opened 2016-08-28
Subprocess Macros

scopatz:
This PR proved subprocess macros. It builds off of #1598, of course, and allows you to do things like:

```console
scopatz@artemis ~ $ timeit! "hello mom " + "dad, yo"
100000000 loops, best of 3: 8.24 ns per loop
```

that are currently impossible to express easily on master.


melund:
Awesome :)

scopatz:
Thanks @melund! Now that #1598 has been merged, this is a much smaller change.  Would you mind reviewing it?

docs/tutorial_macros.rst
```diff
@@ -344,7 +344,90 @@ Of course, many other more sophisticated options are available depending on the
...
+    # normally, xonsh will split on whitespace,
+    # so each argument is passed in separately
+    >>> echo x  y       z
+    x y z
+
+    # usually space can be preserved with quotes
+    >>> echo "x  y       z"
+    x  y       z
+
+    # however, subproces macros will pause and then strip
```
laerus: subprocesS

docs/tutorial_macros.rst
```diff
@@ -344,7 +344,90 @@ Of course, many other more sophisticated options are available depending on the
...
+    # all input after the exclamation point
+    >>> echo! x  y       z
+    x  y       z
+
+However, the macro will pause everything, including path and environment variable
+expansion, that might be present even with quotes.  For example:
+
+.. code-block:: xonshcon
+
+    # without macros, envrioment variable are expanded
```
laerus: enviroment

melund:
@scopatz Sure. I may have time for a closer look at the code tomorrow. That it usually turns out to be more for my own education than the sake of the code quality ;) 

But I tried it out now, and all the examples seem to work nicely.  

If @laerus says go. Then I am all for merging it in.

laerus:
i'm really not the right person the review this, i've played a little with macros and they still seem magical to me. I believe the only thing that's left now is implementing xonsh in xonsh :laughing:

scopatz:
Well, @laerus, I think any review is better than no review :).  I'd really appreciate it.  Perhaps the main question here should be if this is a valuable syntax that you think should be supported.

laerus:
This pretty much solves the need to escape everything with quotes.

>  This allows certain commands to function in cases where quoting or piping might be more burdensome. 

what about piping? are we supposed to do something like `$(echo! ...)`? this works but 
`$ echo $(echo! ...)` doesn't

also something like this:
`$ ls!` returns an error with ls trying to open `''`
is this the way it's supposed to work?

melund:
@scopatz I think it is good to go in unless you want to address the stuff raised by @laerus. 

New features like this are best tested in the wild :)

laerus:
Agreed. I've been playing around with this and I just had some questions. I'm all in for merging.

scopatz:
Yeah, please feel free to merge in :)  The short answer is that the `ls!` is correct behaviour, ls gets an empty name, and then *ls* fails.  The nested subprocs I just forgot to implement.  But it is not hard to add in (I don't think). But that is an additional feature, and so no reason to block :)  I'll try to get to it tonight, but I have sort of a backlog.

laerus:
Thanks @scopatz!

scopatz:
Woohoo!  Thanks to you both!

PR #1645, opened 2016-08-28
Add test to detect event typos

astronouth7303:
Just a test to detect undeclared core events, ie typos.


scopatz:
Thanks @astronouth7303!

PR #1646, opened 2016-08-28
Hot fix for pytest events

scopatz:
pytest magic was adding an extra event on my system. @astronouth7303, mind taking a look.

### DIFF
--- a/tests/test_events.py
+++ b/tests/test_events.py
@@ -111,5 +111,7 @@ def test_transmogrify_by_string(events):
     assert inspect.getdoc(events.on_test) == docstring
 
 def test_typos(xonsh_builtins):
-    for ev in vars(xonsh_builtins.events).values():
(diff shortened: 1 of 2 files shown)


astronouth7303:
Huh. `_pytestfixturefunction` is something applied to fixture functions themselves, I believe. How it got on the global `EventManager`, I'm not sure.

I would suggest filtering all private attributes (ie, anything starting with `_`), but this works fine.

astronouth7303:
Of course, the real solution to this would be to make `EventsManager` a real mapping.

scopatz:
> I would suggest filtering all private attributes (ie, anything starting with _), but this works fi

That was my first impulse as well, but then I thought that folks might want to have private events.

> Of course, the real solution to this would be to make EventsManager a real mapping.

I am all for that, but this is just a hotfix.  Merging in.

PR #1648, opened 2016-08-28
Add events to Vox

astronouth7303:
Adds four venv-related events.

Also improves the vox documentation.

### DIFF
--- a/xontrib/voxapi.py
+++ b/xontrib/voxapi.py
@@ -198,7 +198,7 @@ class Vox(collections.abc.Mapping):
         env_path = builtins.__xonsh_env__['VIRTUAL_ENV']
         if env_path.startswith(self.venvdir):
             name = env_path[len(self.venvdir):]
-            if name[0] == '/':
+            if name[0] in '/\\':
                 name = name[1:]
             return name
         else:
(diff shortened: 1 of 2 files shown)


xontrib/voxapi.py
```diff
@@ -7,16 +16,45 @@
 
 from xonsh.platform import ON_POSIX, ON_WINDOWS, scandir
 
+# This is because builtins aren't globally created during testing.
+# FIXME: Is there a better way?
```
scopatz: I am not sure that there is a better way right now, other than having a xontrib load event that does the importing.  But for that we'd need a xontrib load event :)

scopatz:
This looks good to me @astronouth7303! Is it OK to merge or did you want to add a xontrib event first?  This can always be refactored later.

astronouth7303:
Please merge.

scopatz:
Thanks!

xontrib/voxapi.py
astronouth7303: No, because voxapi is never loaded as a xontrib, just as a Python module.

The test infrastructure never installs builtins globally, so anything core that's tested can't expect `events` (or `aliases`) to be available.

PR #1650, opened 2016-08-29
fix path completion for empty directories

blahgeek:
Before this commit, `ls /tmp/empty_dir/<TAB>` would results `ls //`,
this commit fixes that bug


scopatz:
Thanks a million @blahgeek!

PR #1657, opened 2016-09-01
Fix win tests

melund:
This fixes some test problems I was having locally on windows. 

It skips the tests in 'test_dirstack_unc.py' if drive names are already mapped for other things in windows. 

It also disables `win_unicode_console` if that is installed. The module collides with pytests caputure functionality causing some test to error.


scopatz:
Thanks a ton @melund!

PR #1658, opened 2016-09-01
Misc fix - support subprocess running from different drive letter than PWD (on windows)

bobhy:
execer unit tests were failing on my (windows) PC because my repo is not on C:\\.

There is probably a corresponding bug in unix, since ```os.path.relpath(a, b)``` would fail if ``a`` and ``b`` are on different mountpoints.


melund:
I don't think the `relpath` is needed at all so feel free to change it for all. I think I have seen somewhere else in the code base also. But I may be mistaken.

melund:
No. It was actually this same place. I think I also spotted this problem a while back but forgot to do anything about it.
https://github.com/xonsh/xonsh/pull/1329#discussion_r68448471

Thanks for finding this again @bobhy

scopatz:
Thanks a ton @bobhy!  Sorry if this was a bit premature @melund, I thought it should be in the release, though

melund:
No problem...

xonsh/commands_cache.py
```diff
@@ -133,7 +133,10 @@ def lazy_locate_binary(self, name):
                 if os.path.isfile(full_name)
             ), None)
             if local_bin:
-                return os.path.abspath(os.path.relpath(local_bin, cwd))
+                if os.path.splitdrive(cwd)[0] != os.path.splitdrive(local_bin)[0]:  # if cwd not on same drive as bin
+                    return os.path.abspath(local_bin)                               # avoid ValueError in relpath()
+                else:
+                    return os.path.abspath(os.path.relpath(local_bin, cwd))
```
melund: @bobhy I tried to look at this in more detail to clean it up further, and I don't understand why cwd and the binary can ever be on different drives here. 

We should only be here if local_bin is a file in the current working directory. Right?

melund: is this related to the issue with xonsh not tracking os.getcwd() ?

melund: I made a further cleanup in #1665

bobhy:
The code cleanup in #1665 is certainly a better fix than my patch-the-patches, thanks @melund. I've deleted my branch.

Regarding why a fix was necessary in the first place, I was seeing unit test failures in ```test_execer.test_ipconfig``` when running pytest from my repo directory on another drive letter, e.g, ```n:\devwork\xonsh```.  Exception was ```ValueError``` thrown by ```os.path.relpath```.

PR #1662, opened 2016-09-02
fix completion of paths starting with '~'

blahgeek:
@scopatz  So sorry... spoke too soon... one more fix following #1637


scopatz:
Thanks!

PR #1669, opened 2016-09-02
Informative git status on prompt

blahgeek:
screen shot 2016-09-03 at 01 14 57

Much more informative git status!

Should this go into xonsh core or remains xontrib? And, can I add documentations in `docs/` (as this is a xontrib...) ?

Thanks for reviewing!


melund:
Nice. I want it in core if it isn't slower :)

xontrib/gitstatus.xsh
```diff
@@ -0,0 +1,140 @@
+'''informative git status for prompt'''
```
scopatz: PEP78 triple double quotes

xontrib/gitstatus.xsh
```diff
@@ -0,0 +1,140 @@
...
+import builtins
+import functools
+import subprocess
+
+__all__ = []
+
+_env = builtins.__xonsh_env__
+
+check_output = functools.partial(subprocess.check_output,
+                                 env=_env.detype(),
```
scopatz: I don't think this is valid at the global scope.  The env needs to be detyped each time, right before the call

xontrib/gitstatus.xsh
```diff
@@ -0,0 +1,140 @@
...
+                                 stderr=subprocess.DEVNULL,
+                                 timeout=_env['VC_BRANCH_TIMEOUT'],
```
scopatz: Same here. What if a user changes the timeout?

scopatz:
I agree. This would be great to have in the core!  Thanks for putting this in @blahgeek!

blahgeek:
It's now in core.

I alse moved all `FORMATTER_DICT` related functions to `xonsh/prompt/` sub-package, created a new amalgamate.

melund:
Thanks for the nice work @blahgeek. I agree that a restructure was probably needed here. The new structure also seems more logical.  But it  is with mixed feelings. I really hate that I can't no longer remember where things are located or easily see who coded what. The former can probably be fixed by reviewing this but I don't have time to look at this in details now.Will try to find time later this weekend

xonsh/prompt/base.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
```
laerus: No hashbang here, this will never be executed as a standalone script right?

xonsh/prompt/cwd.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# Created by i@BlahGeek.com at 2016-09-03
+
```
laerus: Module docstrings would be nice aswell

xonsh/prompt/gitstatus.py
```diff
@@ -0,0 +1,140 @@
...
+
+def check_output(*args, **kwargs):
+    kwargs.update(dict(env=builtins.__xonsh_env__.detype(),
+                       stderr=subprocess.DEVNULL,
+                       timeout=builtins.__xonsh_env__['VC_BRANCH_TIMEOUT'],
+                       universal_newlines=True
+                       ))
+    return subprocess.check_output(*args, **kwargs)
+
+_DEFS = {
```
laerus: this can be a function decorated with `lazyobject` from [lazyasd](https://github.com/xonsh/xonsh/blob/master/xonsh/lazyasd.py#L126) that returns the dict like:
```
@lazyobject
def _DEFS():
    return {...}
```

xonsh/prompt/cwd.py
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# Created by i@BlahGeek.com at 2016-09-03
```
scopatz: We don't have a custom of signing files in xonsh. If you could remove this line, that would be most appreciated!

xonsh/prompt/base.py
```diff
@@ -0,0 +1,44 @@
...
+# -*- coding: utf-8 -*-
+
+import os
+import socket
+
+from xonsh.lazyasd import LazyObject
+from xonsh.tools import is_superuser
+from xonsh.platform import ON_WINDOWS
```
scopatz: You should not do `from ... import` for modules outside of the subpackage.  This kinds of imports cannot be performed lazily.

xonsh/prompt/cwd.py
```diff
@@ -0,0 +1,80 @@
...
+
+import os
+import shutil
+import builtins
+
+from xonsh.tools import get_sep
+from xonsh.platform import ON_WINDOWS
```
scopatz: Same here

scopatz:
Thanks for this awesome change @blahgeek!   In addition to the inline notes, the autodoc hooks in the `docs/api` folder now need to be updated too.

blahgeek:
Sure. I just removed those inline comments, just old habits and editor plugins, never mind :P 

I would take a look at `docs/api` and the imports issues later.

blahgeek:
`from ... import ...` statements are removed. `docs/api/prompt` added. Any suggestions?

docs/api/prompt/index.rst
```diff
@@ -0,0 +1,12 @@
...
+=================
+Prompt API
+=================
+Functions in ``FORMATTER_DICT`` for prompt.
+
+.. toctree::
+    :maxdepth: 1
+
+    gitstatus
+    vc_branch
```
scopatz: base, cwd, and env need entries and doc files too please.

news/gitstatus.rst
```diff
@@ -0,0 +1,13 @@
+**Added:**
+
+* New ``FORMATTER_DICT`` entry ``gitstatus`` to provides informative git status
+
+**Changed:** None
+
```
scopatz: Please also add a changed entry for the new prompt subpackage.

xonsh/environ.py
```diff
@@ -45,6 +40,8 @@
     seq_to_upper_pathsep,
 )
 
+from xonsh.prompt.base import FORMATTER_DICT
```
scopatz: You should not do `from ... import ...` for elements outside of the current package as this will not be lazy.

xonsh/tracer.py
```diff
@@ -12,7 +12,7 @@
 from xonsh.platform import HAS_PYGMENTS
 from xonsh.tools import DefaultNotGiven, print_color, normabspath, to_bool
 from xonsh.inspectors import find_file, getouterframes
-from xonsh.environ import _replace_home
+from xonsh.prompt.cwd import _replace_home
```
scopatz: Same here, should not do `from ... import ...` for objects outside of the current subpackage.

xonsh/prompt/gitstatus.py
```diff
@@ -0,0 +1,149 @@
...
+    stashed = _get_stash(gitdir)
+    operations = _gitoperation(gitdir)
+
+    return (branch, num_ahead, num_behind,
+            untracked, changed, conflicts, staged, stashed,
+            operations)
+
+
+def gitstatus_prompt():
+    """Return str `[BRANCH|OPERATOR|numbers]`"""
```
scopatz: I don't think that the return should contain the bounding `[` and `]`, because this is something that the user may wish to have / not have in their own prompt.  I don't count this as blocking and I am happy to put in a separate PR to this effect.

scopatz:
Just a couple of more small issues, @blahgeek and then I think this is ready to go! Thanks again!

docs/api/prompt/index.rst
blahgeek: I thought about that... but those files do not include any functions that does not starts with `_` and have docstrings, should I include them as well?

scopatz: No, but the should have module level docstrings and they may have public members in the future.

scopatz:
This looks good to me! Let us know when you think it is ready!

blahgeek:
@scopatz I think it's good to go. Thanks a million!

scopatz:
Thanks again @blahgeek! This was a much needed refactor!

laerus:
cool! just a question @blahgeek, can we drop `vc_branch.get_git_branch` in favor of `gitstatus.gitstatus` which seems a lot faster, or am i missing something?

blahgeek:
I think, basically, yes. But it seems that `vc_branch.get_git_branch` handles more edge cases (on windows, when git executable do not exist, etc.) and surely we would not want to break any compatibility for existing users using `vc_branch`, so I just left the whole `vc_branch` intact for now. 

I think we can surely make more enhancement to `gitstatus` and maybe merge `gitstatus` and `vc_branch` some day in the future.

PR #1671, opened 2016-09-02
Macro Consistency

scopatz:
This changes macro function calls to default to strings, rather than eval.  This makes it more consistent with subprocess macros and context manager macros, which also default to strings.


melund:
LGTM. A flake 8 issue have sneaked it's way in but it has nothing to do with this PR so I am merging anyway.

PR #1672, opened 2016-09-02
Stop nightly timeouts

scopatz:
Just a test PR for now

### DIFF
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -3,7 +3,6 @@ import os
 import builtins
 import collections.abc as cabc
 
-from xonsh.dirstack import _get_cwd
 from xonsh.platform import ON_WINDOWS
 from xonsh.tools import executables_in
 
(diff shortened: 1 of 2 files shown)


scopatz:
No longer just a test! This seems to work

laerus:
thanks @scopatz !! hope baby travis will stop crying now.

xonsh/commands_cache.py
```diff
@@ -126,10 +125,12 @@ def lazy_locate_binary(self, name):
...
-            local_bin = next((fn for fn in possibilities if os.path.isfile(fn)), None)
+            local_bin = next((fn for fn in possibilities if os.path.isfile(fn)),
+                             None)
             if local_bin:
-                return os.path.abspath(local_bin)                              
-        cached = next((cmd for cmd in possibilities if cmd in self._cmds_cache), None)
+                return os.path.abspath(local_bin)
+        cached = next((cmd for cmd in possibilities if cmd in self._cmds_cache),
+                      None)
         if cached:
```
melund: Oh no.  Was I just PEP8'ed here? Can we change the line endings rules so it doesn't complain hard if we go a few characters above 80. That rule is not giving nicer code ...

laerus: totally agree

scopatz: Well, I have just be using PEP8.  Flake8 is set to 180 right now, which is clearly too high.  If we are going to deviate from PEP8, then we need to document how and in what circumstances.

laerus:
Maybe 90?

Issue #1661, opened 2016-09-02
Single backtick in news should cause test failure


scopatz:
It is too easy to type markdown

PR #1682, opened 2016-09-05
better news testing

laerus:
#1661

### DIFF
--- a/news/test-news.rst
+++ b/news/test-news.rst
@@ -1,7 +1,7 @@
 **Added:**
 
-* fail if no empty line before a category
-* more info on failures
+* test_news fails if no empty line before a category
+* more info on test_news failures
 
 **Changed:** None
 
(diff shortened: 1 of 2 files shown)


news/test-news.rst
```diff
@@ -0,0 +1,16 @@
+**Added:**
+
+* fail if no empty line before a category
+* more info on failures
```
scopatz: These should be more precise about being about news testing.

tests/test_news.py
```diff
@@ -1,37 +1,63 @@
 # -*- coding: utf-8 -*-
 """Testing that news entries are well formed."""
 import os
+import pytest
+import re
```
scopatz: PEP8 standard library imports should be first, then 3rd party packages, and then package or local imports.

scopatz:
Thanks a million @laerus! Just a couple of small changes.

scopatz:
Thanks @laerus!!

PR #1683, opened 2016-09-05
it's time then

laerus:
was staring at `run_subproc` for so long.
here are some tests, wanted to write a lot more and i can still can if you want me to @scopatz 
and these probably would be better with `tmpdir` fixture.

### DIFF
--- a/tests/test_run_subproc.py
+++ b/tests/test_run_subproc.py
@@ -1,0 +1,32 @@
+import sys
+import os
+
+import pytest
+
+from xonsh.built_ins import run_subproc
+
+
+@pytest.yield_fixture(autouse=True, scope='module')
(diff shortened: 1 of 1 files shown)


scopatz:
Thanks @laerus! I have started on a `run_subproc()` refactor, and so these will be super helpful.

laerus:
i will give more time on this later today or tomorrow, will try and write some more tests to include piping and such.

laerus:
note that i'm pushing this into your `rs` branch not in master

scopatz:
Ahh good point. Then I would expect this to fail right now, because that branch is currently totally broken :)

scopatz:
Thanks!

Issue #1695, opened 2016-09-08
Documentation build broken


Carreau:
Just pasting the end of the traceback:

```
  File "/Users/bussonniermatthias/anaconda/envs/test_env/lib/python3.5/site-packages/docutils/utils/code_analyzer.py", line 79, in merge
    for ttype, value in tokens:
  File "/Users/bussonniermatthias/anaconda/envs/test_env/lib/python3.5/site-packages/pygments/lexer.py", line 191, in streamer
    for _, t, v in self.get_tokens_unprocessed(text):
  File "/Users/bussonniermatthias/anaconda/envs/test_env/lib/python3.5/site-packages/pygments/lexer.py", line 634, in get_tokens_unprocessed
    for item in action(self, m):
  File "/Users/bussonniermatthias/anaconda/envs/test_env/lib/python3.5/site-packages/xonsh/pyghooks.py", line 41, in command_token_callback
    if cmd in builtins.__xonsh_commands_cache__ or \
AttributeError: module 'builtins' has no attribute '__xonsh_commands_cache__'
```

Trying to get docs to build on travis to avoid this issue.

gforsyth:
Fixed by @Carreau himself in #1697

PR #1697, opened 2016-09-08
Build docs on travis

Carreau:
The first commit test should fail see cause of #1695.
If it fails with the same cause I'll push a second commit that fix the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,9 @@ matrix:
     include:
         - os: linux
           python: "nightly"
-          env: RUN_COVERAGE=true
+          env:
+            - RUN_COVERAGE=true
+            - BUILD_DOCS=true
         - os: osx
           language: generic
           env: PYTHON="3.4" MINICONDA_OS="MacOSX"
@@ -31,6 +33,11 @@ install:
     else
       pip install -r requirements-tests.txt;
     fi
+  - if [[ $BUILD_DOCS = true ]]; then
+      python setup.py install;
+      pip install -r requirements-docs.txt;
+    fi
+
 
 script:
   - if [[ $RUN_COVERAGE = true ]]; then
@@ -38,6 +45,11 @@ script:
     else
       py.test --timeout=10;
     fi
+  - if [[ $BUILD_DOCS = true ]]; then
+        cd docs;
+        make html;
+    fi
+
 
 after_success:
   - if [[ $RUN_COVERAGE = true ]]; then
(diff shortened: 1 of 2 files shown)


Carreau:
Travis is failing for the god reason. Pushing a commit that should fix the doc build.

scopatz:
I am not sure what is going on with the failures here @Carreau, but thanks for putting this in!

Carreau:
> I am not sure what is going on with the failures here @Carreau, but thanks for putting this in!

Failure look (to me) like a spurious one. I'll restart this specific build.

Carreau:
> I am not sure what is going on with the failures here @Carreau, but thanks for putting this in!

Actually I wouldn't be astonished if this is due to the recent merge of compact dicts that have been merged yesterday in CPython, as it looks like a C-level error triggered by py-test

laerus:
compact dicts woohooo!, did they already merge it in python3.5 wow..

gforsyth:
maybe we should temporarily switch `nightly` to `3.5` then?

scopatz:
I would be fine merging things for the next couple of days if they exhibit only this failure. This pop issue is clearly not coming from us.

scopatz:
Maybe @gforsyth should be the one to make the call on this PR since it is more in his bailiwick

Carreau:
We can mark nightly as "allow_failure". 

There already a patch pending upstream, so it should be fixed soon.

http://bugs.python.org/issue28040

gforsyth:
> We can mark nightly as "allow_failure".

That seems like a better temporary fix to me.  

On another note, is it time to call the `doctr`?

Carreau:
> On another note, is it time to call the doctr?

Yes, yes, but I want to get the url scheme right for multiple versions of the docs.

Carreau:
> That seems like a better temporary fix to me.

I'm leaving in 2 h and need to prepare my stuff , I might offline until tuesday. You should have the right ito push on this branch ("Allow edits from maintainers") if needed.

gforsyth:
Ok, I'll take care of it.  Thanks, @Carreau!

gforsyth:
Hey @scopatz -- if you're ok with this as it is, I say we merge it.  The docs are building correctly and this will also get the updated `travis.yml` into master which should stop a bunch of the spurious error messages based on the new 3.6 generator comprehensions

scopatz:
If you think it is ready, I am happy with it :)

laerus:
allright i'm merging this to stop the nightly failure madness, thanks @Carreau & @gforsyth !!

Issue #1681, opened 2016-09-05
`~` should not be expanded at the middle of string


blahgeek:
So that `scp another-user-name@example.com:~/a.txt ./` would not fail.

This is a bug that introduced at #1624 

The document reads: "Each **variable assignment** is checked for unquoted tilde-prefixes immediately following a ‘:’ or the first ‘=’." Only variable assignment commands.

scopatz:
Thanks for reporting @blahgeek! Xonsh does not have a notion of variable assignment commands in subprocess mode.  I am not entirely sure what to do here. This is probably less than ideal, but for the moment you could do:

```
$ scp @("another-user-name@example.com:~/a.txt") ./
```

laerus:
maybe we should expand tilde only if it is after a space?

scopatz:
Well, after thinking about this more, I think we should only perform colon expansions after an `=`. I think that is what the Bash docs mean by assignment

blahgeek:
For me, the only use case that needs tilde expansion after ':' on bash is when setting `$PATH` (or `$..._PATH`) variable, in which case #1624 is not needed. What's your original intention in #1624 ?

scopatz:
The purpose of #1624 was to be consistent with Bash, because it was driving me nuts.  I overshot that a bit...

blahgeek:
Actually I don't think that being consistent with bash/zsh/or any other shells is 100% needed for xonsh. I'd prefer xonsh more like python instead of bash... But that's a big general topic that needs discussion beyond this issue

scopatz:
Sure, I agree in general.  But this is about token expansion in subprocess mode.  I thought that what you were asking for is more like what Bash has anyway. I could be misreading this, though...

blahgeek:
Well... I think what I was asking for is that, if being 100% like bash is too hard to implement or do not suites well with python syntax, it may be better to just leave it. In this case, correctly handling tilde expansion like bash may be too much work, there's always edge cases (and also, even in subproc mode, users may not expect literal strings altered by some unclear rules), may be it would be better to just do not expand it...

scopatz:
Ahh OK.  The one thing I really want to work is `--arg=~/mydir`.  I think the "leading tilde and after equals or after a colon after an equals" is probably the behaviour we want. I think that is what Bash does, even though their docs do not make that clear.

blahgeek:
[...]

PR #1706, opened 2016-09-09
Fixed tilde expansion

scopatz:
Should address #1681


gforsyth:
Looks like some new generator comprehension stuff hit nightly and is unhappy, but I don't think that really matters for the moment.  Looks good to me, merging

gforsyth:
Thanks, @scopatz !

Issue #1698, opened 2016-09-08
just rebuilt, tested - xonsh wins!


PaulReiber:
Greetings,

Happy to share what I'm seeing w/ latest build.  
```
pbr@mongo:~$ xonsh
aliases: alias 'fgrep' of shell 'bash' tries to override xonsh alias, xonsh wins!
aliases: alias 'grep' of shell 'bash' tries to override xonsh alias, xonsh wins!
aliases: alias 'egrep' of shell 'bash' tries to override xonsh alias, xonsh wins!
aliases: alias 'ls' of shell 'bash' tries to override xonsh alias, xonsh wins!
pbr@mongo ~ $ 

```
...wondering why xonsh has to win here.

gforsyth:
Hey @PaulReiber, I know one specific problem we were trying to avoid with this (there are probably others).  `zsh` has a `history` alias that itself aliases a `zsh` command `fc` which `xonsh` doesn't have, so the alias just breaks and gives a message about `fc` not installed.  If there are conflicts with builtin `xonsh` aliases, we override to avoid these sorts of problems.  

You can definitely re-override them by setting them in your `xonshrc`.

scopatz:
Thanks for reporting @PaulReiber. In addition to what @gforsyth said, I think that we can change the text of "xonsh wins!" to be more clear that xonsh is ignoring that alias and using the existing xonsh one.

laerus:
i did that 'xonsh wins' cause everything else was too long, what exactly do you want to message to say @scopatz ?

PaulReiber:
Can we look at this from, say, 50 feet up instead of so close?
My concern's not with the message - it's with what's going on.

This is the xonsh shell, not zsh, not bash, not csh, not tcsh, not dash.

My first thought when I learned that xonsh was reading my .bashrc was:
 WHY?!?!?  (yes, the thought in my head had the exclamation marks and
question marks along with it)

Why's it reading in startup files designed to be used by OTHER shells as it
starts up?  I'm looking at this and pondering:

What are the responsibilities of a shell?

Is xonsh (a) missing anything it should be doing or (b) doing anything it
shouldn't be doing?

Is xonsh maintaining a proper separation of concerns?

Traditionally the environment's the only thing a parent shares with the
child process, data-sharing wise.  There's nothing in Linux that says child
processes which happen to be shells should process the config files of
$SHELL as well.

Maybe I'm missing something?

Kind regards,


*Paul Reiber*http://about.me/reiber

[...]

PR #1712, opened 2016-09-10
control overriding of foreign aliases

laerus:
#1698

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1176,7 +1176,8 @@ function with the ``xonsh.proc.foreground`` decorator.
 Aliasing is a powerful way that xonsh allows you to seamlessly interact to
 with Python and subprocess.
 
-.. warning:: Foreign shell aliases that try to override xonsh aliases will be ignored
+.. warning:: if ``FOREIGN_ALIASES_OVERRIDE`` enviroment variable is False then
+             foreign shell aliases that try to override xonsh aliases will be ignored.
 
 Up, Down, Tab
 ==============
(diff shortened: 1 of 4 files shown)


scopatz:
Thanks @laerus!

scopatz:
Ahh, Ok I thought that this would run the other way. Basically that the new alias would override and only print a warning message if debug is on. I am happy with this though for the moment if you just want it merged in.

laerus:
do we ever want foreign aliases to override xonsh aliases?

laerus:
we can always add an envvar so the user can control what happens.

scopatz:
Yeah, a new envvar is probably best

laerus:
any good name for the envvar? i'm thiking something like $XONSH_OVERRIDE_ALIASES but this is kind of confusing.

scopatz:
Maybe `FOREIGN_ALIASES_OVERRIDE`?

laerus:
so if `$FOREIGN_ALIASES_OVERRIDE` is False then xonsh will ignore foreign aliases that clash with xonsh aliases, also i think False should be the default.

scopatz:
Yep, that is what I think too

laerus:
uh this deosn't work if i set `FOREIGN_ALIASES_OVERRIDE=True` in my .xonshrc,
cause i just realized that `load_foreign_aliases` is being called before .xonshrc has been parsed.

scopatz:
Yeah, if you want this apply to the config file, you'd have to set this envvar in the "env" section of the config file.

melund:
I think this needs #1716 merged first, or integrated into this.

melund:
Cool. I think this is good to go.

Issue #1691, opened 2016-09-08
Path Completer infinite recursion on windows


astronouth7303:
```
...
    return _splitpath_helper(folder, sofar)
  File "C:\Users\James\AppData\Local\Programs\Python\Python35\lib\site-packages\xonsh\completers\path.py", line 193, in _splitpath_helper
    return _splitpath_helper(folder, sofar)
  File "C:\Users\James\AppData\Local\Programs\Python\Python35\lib\site-packages\xonsh\completers\path.py", line 193, in _splitpath_helper
    return _splitpath_helper(folder, sofar)
  File "C:\Users\James\AppData\Local\Programs\Python\Python35\lib\site-packages\xonsh\completers\path.py", line 191, in _splitpath_helper
    if not folder or folder == xt.get_sep():
  File "C:\Users\James\AppData\Local\Programs\Python\Python35\lib\site-packages\xonsh\tools.py", line 422, in get_sep
    if ON_WINDOWS and builtins.__xonsh_env__.get('FORCE_POSIX_PATHS'):
  File "C:\Users\James\AppData\Local\Programs\Python\Python35\lib\site-packages\xonsh\environ.py", line 856, in get
    return self[key]
  File "C:\Users\James\AppData\Local\Programs\Python\Python35\lib\site-packages\xonsh\environ.py", line 827, in __getitem__
    cabc.MutableMapping)):
  File "C:\Users\James\AppData\Local\Programs\Python\Python35\lib\abc.py", line 188, in __instancecheck__
    subclass in cls._abc_negative_cache):
  File "C:\Users\James\AppData\Local\Programs\Python\Python35\lib\_weakrefset.py", line 75, in __contains__
    return wr in self.data
RecursionError: maximum recursion depth exceeded in comparison
```

I believe the input was

```
$ cd ~/.ven<TAB>
```
or similar.

astronouth7303:
(I'm guessing it's climbing up the directory tree looking for an empty path, but on windows it's rooted in `'C:'`, not `''`)

gforsyth:
What happens if you have `$FORCE_POSIX_PATHS=False` ?

astronouth7303:
It's already `False`.

gforsyth:
Hey @astronouth7303 -- when you're next on a windows box, can you bisect this?  I think windows path completion was working not that long ago

astronouth7303:
Bisect blames b3b38eee611a8e5456540e41caabedbc64a7584b:

```
[...]

PR #1713, opened 2016-09-11
remember windows when checking for abs paths

gforsyth:
potential fix for #1691 

@astronouth7303 -- not sure if path completion will *work* but I think this will stop the infinite recursion

### DIFF
--- a/xonsh/completers/path.py
+++ b/xonsh/completers/path.py
@@ -188,7 +188,8 @@ def _splitpath_helper(path, sofar=()):
     folder, path = os.path.split(path)
     if path:
         sofar = sofar + (path, )
-    if not folder or folder == xt.get_sep():
+    if (not folder or folder == xt.get_sep() or
+       (xp.ON_WINDOWS and os.path.splitdrive(path)[0])):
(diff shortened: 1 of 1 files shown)


scopatz:
Should we wait for @astronouth7303? Or is this good to merge now?

gforsyth:
I'd like for someone who had the error to confirm that it's fixed -- I think this will do it but there could be other things I haven't accounted for

scopatz:
Makes sense @gforsyth!

bobhy:
The fix works for me, FWIW.  I was able to repro by:
```
dir ./abc<tab>    # no files abc* in this directory
```
Before the fix, I got the recursion.  After the fix, cursor just stayed put at end of line, as expected.

Fix doesn't break: 
``` 
dir abc<tab>
```
Cursor stays put here, too.

melund:
Great. I will merge then

astronouth7303:
After pulling from `master`, I'm still getting infinite recursions if `$SUBSEQUENCE_PATH_COMPLETION` is `True`.

PR #1717, opened 2016-09-12
Fix async comprehension on 3.6

Carreau:
Seem simple enough, and avoid writing a full 3.6 parser for now. 

-- 

I tried to modify the parser to actually recognize actual `async for` comprehensions, but the lexer insist on seeing async as a NAME instead of an async token and thus refuses to match the `async_comp_for` rule...

### DIFF
--- a/tests/tools.py
+++ b/tests/tools.py
@@ -94,15 +94,15 @@ def check_parse(input):
 
 def nodes_equal(x, y):
     __tracebackhide__ = True
-    assert type(x) == type(y)
+    assert type(x) == type(y) , "Ast nodes do not have the same type: '%s' != '%s' " % (type(x), type(y))
     if isinstance(x, (ast.Expr, ast.FunctionDef, ast.ClassDef)):
-        assert x.lineno == y.lineno
-        assert x.col_offset == y.col_offset
+        assert x.lineno == y.lineno, "Ast nodes do not have the same line number : %s != %s" % (x.lineno, y.lineno)
+        assert x.col_offset == y.col_offset, "Ast nodes do not have the same column offset number : %s != %s" % (x.col_offset, y.col_offset)
     for (xname, xval), (yname, yval) in zip(ast.iter_fields(x),
                                             ast.iter_fields(y)):
-        assert xname == yname
-        assert type(xval) == type(yval)
+        assert xname == yname, "Ast nodes fields differ : %s (of type %s) != %s (of type %s)" % (xname, type(xval), yname, type(yval))
+        assert type(xval) == type(yval), "Ast nodes fields differ : %s (of type %s) != %s (of type %s)" % (xname, type(xval), yname, type(yval))
     for xchild, ychild in zip(ast.iter_child_nodes(x),
                               ast.iter_child_nodes(y)):
-        assert nodes_equal(xchild, ychild)
+        assert nodes_equal(xchild, ychild) , "Ast node children differs"
     return True
--- a/xonsh/parser.py
+++ b/xonsh/parser.py
@@ -6,8 +6,10 @@ from xonsh.platform import PYTHON_VERSION_INFO
 
 @lazyobject
 def Parser():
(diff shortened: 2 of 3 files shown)


xonsh/parsers/base.py
```diff
@@ -2324,7 +2325,10 @@ def p_comp_for(self, p):
         else:
             targ = ensure_has_elts(targs)
         store_ctx(targ)
-        comp = ast.comprehension(target=targ, iter=it, ifs=[])
+        if sys.version_info > (3,6):
+            comp = ast.comprehension(target=targ, iter=it, ifs=[], is_asyc=False)
```
gforsyth: *`is_asyc` -> `is_async`

gforsyth:
I love simple solutions.  And yeah, it's nice to not have to worry about a 3.6 parser just yet

xonsh/parsers/base.py
Carreau: Doh.

gforsyth: That's the problem with asynchronous typing

scopatz:
I think we should really just bite the bullet and make a v36 class.  Early versions of the parser took the if-else strategy and it became *really* hard to maintain or think about.  I am happy to make such a class if you don't want to @Carreau.

scopatz:
Thanks a million for putting this in @Carreau! I really appreciate it.  It will be great to roll out xonsh on v3.6 same day.

Carreau:
I'll make a 3.6 Parser then. Thanks !

scopatz:
No problem and thanks! The easy way  is to probably inherit from the v35 parser :)

Carreau:
> The easy way is to probably inherit from the v35 parser :)

Yeah, not sure if ply will like that.

Carreau:
New parser pushed.

gforsyth:
Hey @Carreau -- the parser looks good to me, but there are still some test failures on Travis for 3.6.  Annoyingly, tests are timing out before it actually says which tests are failing, but it looks like at least 3 tests in `test_contexts` aren't passing.  
It's also suspicious that the tests always hang at the same point in the history tests... not sure what's going on there.

Carreau:
> but it looks like at least 3 tests in test_contexts aren't passing.

Yeas that context one I don't know how to fix. They rely on `global` in a context manager assuming they are block though this is a SyntaxError in Python 3.6 (Syntax Warning in 3.4, 3,5), and I'm not sure how to fix that. It might  get stuck because of some 3.6 issue on IPython side the build get killed because of MemoryError at the same point every time.

gforsyth:
> They rely on global in a context manager assuming they are block though this is a SyntaxError in Python 3.6 (Syntax Warning in 3.4, 3,5),

Ahh, right.  Sorry, I knew about that but I forgot.  So this is more a problem with pytest and 3.6, then

scopatz:
This looks great to me @Carreau! Any reason not to merge it?

Carreau:
> Ahh, right. Sorry, I knew about that but I forgot. So this is more a problem with pytest and 3.6, then

Why did I wrote "They", no Xonsh rely on global in a context manager not being a syntax error. I'm not sure we __can__ do what we are current y doing in 3.6 for blocks. But I'm not super familar with that piece of the code.

gforsyth:
Thanks, @Carreau !

donnemartin:
Hi @Carreau, @scopatz, @gforsyth: We're running into an issue with `gitsome` on Python 3.6 that might be related to this PR: https://github.com/donnemartin/gitsome/issues/105.

https://github.com/donnemartin/gitsome/issues/25 is still on my radar, I hope to dive into it in the next month or two.  Sorry I know this is overdue.

>Seem simple enough, and avoid writing a full 3.6 parser for now.

Was there a simple patch that might work for `gitsome` before the creation of the full 3.6 parser?

Thanks for the help!

ryukinix:
Would be great a fix for gitsome

PR #1719, opened 2016-09-12
Fix context-manager tests on 3.6

### LINKED ISSUE #1718: Global after use is SyntaxError in 3.6
Cf http://bugs.python.org/issue27999, it seem like some tests likely might reuse a variable name or some other weird issue in `tests/test_contexts.py`

Carreau:
Don't reuse variables names in tests.

Since http://bugs.python.org/issue27999, global after use is a
SyntaxError; reusing variable name in tests/test_contexts.py
lead to failure.

Should close #1718

--- 

On top of #1717 that fix other things and is necessary for tests to pass.

### DIFF
--- a/tests/test_contexts.py
+++ b/tests/test_contexts.py
@@ -236,38 +236,38 @@ def test_block_func_trailing_triple_string():
 # Functor tests
 #
 
-X2_WITH = ('x = 1\n'
+X2_WITH = ('{var} = 1\n'
            'with Functor() as f:\n'
            '{body}'
-           'x += 1\n'
+           '{var} += 1\n'
            '{calls}\n'
            )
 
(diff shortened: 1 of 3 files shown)


Carreau:
Gnnn. No does'nt work. `Functor()` with `globals` are (will be) invalid in Python 3.6 if there is a global statement inside as they compile to Python AST. Or we need to nuke the content at parse time to make it empty, or replace by macros. AFAICT.

CC @scopatz .

scopatz:
Thanks for fixing this up @Carreau! I means a lot to me; I am totally overwhelmed this week.  Code looks good here, just waiting for #1717

gforsyth:
Thanks again, @Carreau !

scopatz:
Woohoo!

PR #1720, opened 2016-09-13
Completes commands after pipe

blahgeek:
completes `ls | gre<TAB>`

### DIFF
--- a/xonsh/completers/commands.py
+++ b/xonsh/completers/commands.py
@@ -7,7 +7,7 @@ import xonsh.platform as xp
 from xonsh.completers.tools import get_filter_function
 
 SKIP_TOKENS = {'sudo', 'time', 'timeit', 'which', 'showcmd', 'man'}
-GOTO_TOKENS = {'|', '||', '&&', 'and', 'or'}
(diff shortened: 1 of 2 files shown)


xonsh/completers/commands.py
```diff
@@ -7,6 +7,7 @@
 from xonsh.completers.tools import get_filter_function
 
 SKIP_TOKENS = {'sudo', 'time', 'timeit', 'which', 'showcmd', 'man'}
+GOTO_TOKENS = {'|', '||', '&&', 'and', 'or'}
```
scopatz: I don't know that "GOTO" fits my brain here, because we aren't referring to jump statements.  Maybe this it nitpicking too much but perhaps we could call them  `end_proc_tokens` or something?

scopatz:
Thanks a million BlahGeek! This is certainly much needed! Just the one small nitpick about the name from my end.  Otherwise the code looks great!

blahgeek:
Updated. Naming things is hard :P

gforsyth:
Thanks, @blahgeek! This looks great!

Issue #1542, opened 2016-08-09
Could not create '.git/index.lock': File exists.


Droggelbecher:
The above error message happens to me a lot when running git commands since using xonsh. Apart from the xonsh prompt I there is no other process meddling with my repo that could be responsible for it.

Although frequent enough to be remarkable, it does not seem to occur deterministically, but has a random component (maybe a timing/parallelization issue?).

laerus:
I get that too from time to time, never thought it is cause of xonsh. You can try using `fuser` or `lsof` on the file when it happens to see what os holding it. maybe this happens cause of the branch prompt mechanism allthough this is just a guess.

astronouth7303:
I also get this intermittently. I'm using this work-around:

```python
def _gitfix(a, s=None):
    if !(pidof git):
        return "", "Git is currently running; cowardly refusing\n", 1
    elif not os.path.exists('.git/index.lock'):
        return "", "No fixing required\n"
    else:
        ![rm .git/index.lock]
aliases['fixgit'] = _gitfix
```

So far, it hasn't cowardly refused.

mitnk:
Remove git related info from your xonsh prompt :)

astronouth7303:
Does `xonsh.environ.current_branch()` kill the process if it times out?

laerus:
Doesn't seem so. `get_git_branch` uses `subprocess.check_output` and even though python docs explicitly say that `subprocess.run` kills the child proc on timeout, it doesn't say anything about `checkout_output` except that the signature is the same. The `timeout` parameter is passed on  `Popen.communicate` that explicitly states 

> If the process does not terminate after timeout seconds, a TimeoutExpired exception will be raised. Catching this exception and retrying communication will not lose any output.
> 
> The child process is not killed if the timeout expires, so in order to cleanup properly a well-behaved application should kill the child process and finish communication

so we may have to look a little deeper in the CPython source except if someone knows what's going on.

astronouth7303:
I'm pretty sure killing `git` would cause this problem.

scopatz:
I believe that this is because of the timeout.  You can fix this either by removing the version control stuff from the $PROMPT or by bumping the timeout length with `$VC_BRANCH_TIMEOUT` http://xon.sh/envvars.html#vc-branch-timeout 

A fix that cleans up the lock file in the event of a time out (if it exists) would be most welcome!

astronouth7303:
[...]

PR #1724, opened 2016-09-13
vc_branch prompt

laerus:
maybe a fix for #1542


gforsyth:
hey @laerus, thanks for putting this in!

As it is, this breaks my `$PROMPT` when I'm _not_ in a git repo

xonsh/prompt/base.py
```diff
@@ -20,7 +20,9 @@
 from xonsh.prompt.gitstatus import gitstatus_prompt
 
 
-FORMATTER_DICT = xl.LazyObject(lambda: dict(
+@xl.lazyobject
+def FORMATTER_DICT():
+    return dict(
     user=os.environ.get('USERNAME' if xp.ON_WINDOWS else 'USER', '<user>'),
```
astronouth7303: This indentation is probably a PEP8 violation.

laerus:
not sure why this is happening atm, will investigate more.

xonsh/prompt/gitstatus.py
```diff
@@ -110,20 +119,28 @@ def gitstatus():
...
 
 def gitstatus_prompt():
     """Return str `BRANCH|OPERATOR|numbers`"""
     try:
-        (branch, num_ahead, num_behind,
-         untracked, changed, conflicts, staged, stashed,
-         operations) = gitstatus()
+        status = gitstatus()
     except subprocess.SubprocessError:
         return None
```
gforsyth: maybe if this returns an empty string instead?

gforsyth: hmm, nope, that didn't work

laerus:
found what's wrong (pdb to the rescue)
the bad news is that i've encountered the index.lock bug while i was messing around with this.
i'm having doubts about this PR fixing this.

scopatz:
Yeah, as long as their are time outs, I doubt that there is a silver bullet.  A more bullet proof option might be to do the git check on a background thread, which we ignore if it goes beyond the timeout.

laerus:
Threading and bullet proof in the same sentence is a paradox :laughing:
I don't get how threading will make a difference, will we wait for the thread to finish before printing the prompt?

-
after some benchmarking:
```
$ timeit! get_git_branch()
100 loops, best of 3: 4.36 ms per loop
$ timeit! gitstatus()
100 loops, best of 3: 7.24 ms per loop
```

the older `get_git_branch` is faster after all, so no reason to change `vc_branch.current_branch` to use `gitstatus`

melund:
> I don't get how threading will make a difference, will we wait for the thread to finish before printing the prompt?

No. The lock issue happens when we kill the process after a timeout. 

What @scopatz  is suggesting is to run the git query in a separate thread. We can still have the normal time-out and give up if it doesn't return fast enough. But since it is running in a thread we don't have to kill the process but just leave the process to finish on its own. That would prevent the lock issue.

laerus:
Thanks for clearing this out for me @melund. Do we really want to introduce threading and all their goodies like overhead on the prompt though? I think there is still space for making the implementation a little more sufficient.

i've rewritten `get_git_branch` and `get_dirty_working_directory`

```
$ timeit! get_git_branch()
100 loops, best of 3: 3.04 ms per loop
```

```
$ timeit! git_dirty_working_directory() <--- old one
100 loops, best of 3: 3.64 ms per loop

$ timeit! git_dirty_working_directory() <--- new one
100 loops, best of 3: 2.92 ms per loop
```

Also it seems that we are calling both of these function for each prompt which results in calling git at least two times, while we can get all the info (that is branch name and if its dirty or not) just from calling git once, right?

melund:
Combining the calls would be a significant improvement. Any other code optimizations in xonsh is probably not worth it. The problem is not xonsh but the subproc call to git. On large git repositories, it can be very slow, I have seen almost half 0.4 seconds on my machine :(  In these cases it is best to give up, and not wait for it to finish.

 I haven't had the git lock issues. But for people that have that problem running in a thread is the best solution.

laerus:
or we can just remove the `index.lock` when catching `TimeoutExpired` as previously suggested and not involve threads which imo will make this more ugly.

laerus:
i can try doing this with threads, do we want to keep this version of the functions or bring back the old ones?

xonsh/prompt/vc_branch.py
```diff
@@ -129,39 +100,28 @@ def current_branch(pad=NotImplemented):
     cmds = builtins.__xonsh_commands_cache__
     if cmds.lazy_locate_binary('git') or cmds.is_empty():
         branch = get_git_branch()
-    if (cmds.lazy_locate_binary('hg') or cmds.is_empty()) and not branch:
+    elif (cmds.lazy_locate_binary('hg') or cmds.is_empty()) and not branch:
```
laerus: fixme, return this to `if`

laerus:
i don't know if this is robust enough or not but i've already encountered this:

```
Exception in thread Thread-8:
  File "/usr/lib/python3.5/subprocess.py", line 695, in run
    stdout, stderr = process.communicate(input, timeout=timeout)
  File "/usr/lib/python3.5/subprocess.py", line 1059, in communicate
    stdout = self.stdout.read()

  File "/usr/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.5/threading.py", line 862, in run
  File "/home/laerus/projects/xonsh/xonsh/prompt/vc_branch.py", line 135, in _git_dirty_working_directory
    stderr=subprocess.DEVNULL)
  File "/usr/lib/python3.5/subprocess.py", line 626, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.5/subprocess.py", line 708, in run
    output=stdout, stderr=stderr)
  File "/usr/lib/python3.5/subprocess.py", line 984, in __exit__
    self.stdout.close()
```

well subprocess spawns it's thread for `communicate`

also is there a possibility that git will hang and never let the thread die?
in large repos maybe there is a problem with too many threads spawning
if the user just hits enter a few times.

generally i would like to avoid threads but if coredevs think that's the best solution, i'll keep it that way.

astronouth7303:
Finding `index.lock` in all cases (eg, multiple workspaces) is non-trivial.

Given that git is already an external process, do we actually need a xonsh thread to babysit it? Could we just... let it wander off?

xonsh/prompt/gitstatus.py
```diff
@@ -1,13 +1,21 @@
...
 import builtins
+import collections
+import os
 import subprocess
 
 import xonsh.lazyasd as xl
 
 
+@xl.lazyobject
+def GitStatus():
```
scopatz: PEP8 function name.

astronouth7303: PEP8 variable name. It's a lazy object.

scopatz: Yeah, this is a weird case.

scopatz: But named tuples are just type definitions, so they probably shouldn't be lazily implemented.  We don't hide class definitions in lazy objects, for example.  I think this is overdoing it.

xonsh/prompt/gitstatus.py
```diff
@@ -110,20 +119,28 @@ def gitstatus():
...
 
+    branch = status.branch
+    num_ahead = status.num_ahead
+    num_behind = status.num_behind
+    untracked = status.untracked
+    changed = status.changed
+    conflicts = status.conflicts
+    staged = status.staged
+    stashed = status.stashed
+    operations = status.operations
```
scopatz: Why not just unpack the status namedtuple?

astronouth7303: I usually use namedtuples as compact data structs. The point is that you kinda forget the order of it.

xonsh/prompt/vc_branch.py
```diff
@@ -9,49 +9,37 @@
...
+            q.put(status[3])
+        else:
+            q.put(status[2])
 
 def get_git_branch():
-    """Attempts to find the current git branch.  If no branch is found, then
-    an empty string is returned. If a timeout occured, the timeout exception
-    (subprocess.TimeoutExpired) is returned.
-    """
+    # import pdb; pdb.set_trace()
```
scopatz: Excess comment.

xonsh/prompt/gitstatus.py
astronouth7303: The bit I find odd is that he unpacked it at all.

scopatz: Yeah that is better said :)

xonsh/prompt/vc_branch.py
```diff
@@ -137,30 +125,32 @@ def current_branch(pad=NotImplemented):
...
+    except subprocess.CalledProcessError:
+        q.put(None)
+    if status is not None:
+        if b'nothing to commit' in status:
+            return q.put(False)
+        else:
+            return q.put(True)
+
+
+def git_dirty_working_directory():
```
scopatz: Public functions should have docstrings.

scopatz:
Thanks @laerus! Overall this looks pretty good to me.  I think the the namedtuple should be moved globally (not in a lazyobject) and then the unpacking either shouldn't happen or should use Python's unpacking syntax.  Other than that, this is probably good to go!

xonsh/prompt/gitstatus.py
laerus: oh right, i forgot to clean this up

laerus:
So are we keeping threads in this? what about the seemingly random OSError.

xonsh/prompt/vc_branch.py
```diff
@@ -9,49 +9,40 @@
...
+import xonsh.prompt
+
+import threading, queue
+
+
+def _get_git_branch(q):
+    try:
+        status = subprocess.check_output(['git', 'status'],
+                                         stderr=subprocess.DEVNULL)
+    except subprocess.CalledProcessError:
```
scopatz: I think that if you caught OSError here too, you could avoid the error later on. You might also need to use PIPE, but let'd try one thing at a time: http://stackoverflow.com/questions/18439712/subprocesss-popen-closes-stdout-stderr-filedescriptors-used-in-another-thread-w

scopatz:
I think we need to try to keep the threads. Sorry for the delay getting back to you @laerus!

laerus:
np @scopatz, hope you're doing good! Fixed the conflicts and catching OSError on subprocess.

gforsyth:
I think this looks good, but can @melund or @astronouth7303 confirm that it's working reasonably well on Windows?

astronouth7303:
I'll have to look. I've never gotten this issue on Windows, though.

laerus:
ping for this!

scopatz:
Thanks @laerus!

PR #1726, opened 2016-09-13
Restore bash git completions on Windows

melund:
I noticed that git completions had stopped working for me on windows. I haven't figure when it happen. Maybe xonsh or maybe a restructure of Git For Windows. 

Anyway, this fix makes it work again by adding the `` 'GitForWidnows\\mingw64\\share\\git\\completion\\git-completion.bash'` to the default completion list.


scopatz:
Thanks a million @melund!

PR #1728, opened 2016-09-14
Add on_precommand and on_postcommand

astronouth7303:
Add events bracketing the execution of commands.

The arguments are heavily based on what's required to implement history.

Note: There's no test because shell itself isn't really tested, and that's a big can of worms.

### DIFF
--- a/xonsh/base_shell.py
+++ b/xonsh/base_shell.py
@@ -159,6 +159,9 @@ class BaseShell(object):
         src, code = self.push(line)
         if code is None:
             return
+
+        events.on_precommand.fire(src)
+
         hist = builtins.__xonsh_history__  # pylint: disable=no-member
         ts1 = None
         store_stdout = builtins.__xonsh_env__.get('XONSH_STORE_STDOUT')  # pylint: disable=no-member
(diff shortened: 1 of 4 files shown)


xonsh/shell.py
```diff
@@ -11,6 +11,20 @@
 from xonsh.platform import (best_shell_type, has_prompt_toolkit,
                             ptk_version_is_supported)
 from xonsh.tools import XonshError, to_bool_or_int
+from xonsh.events import events
+
+
+events.doc('on_precommand', """
```
scopatz: I am in favor of changing these to just `{pre,post}command` since the `on_` is implied. It is your PR so I defer to your naming, but I thouhgt i would register my preference.

xonsh/base_shell.py
```diff
@@ -159,6 +159,9 @@ def default(self, line):
         src, code = self.push(line)
         if code is None:
             return
+
+        events.on_precommand.fire(src)
+
```
scopatz: This seems like excess vertical whitespace.

astronouth7303: A little grouping never hurt anyone.

scopatz:
Thanks @astronouth7303! I think the code looks good all-in-all and will enable some needed features. Just a couple of small things to consider that are non-blocking.

xonsh/base_shell.py
scopatz: PEP8 says to use these sparingly, and I agree because you can get more code on the screen simeltaneously the fewer blank lines you have.  I don't think this blocks merging.

xonsh/shell.py
astronouth7303: Yeah, I see what you mean and I kinda agree. But at the same time, I like having the `on_` convention.

Eh, it can go either way.

scopatz: Your call :) If you wanna keep the on convention that is fine by me.

astronouth7303: I think we should keep it. None of the existing events grammatically flow; they all read like "On event <opaque identifier>" and not like sentences.

scopatz: ok I will go ahead and merge then!

scopatz:
thanks a million!

PaulReiber:
+1 and thank you too Anthony.  Looking forward to exploring what can be
done now w this. 😎

On Sep 13, 2016 8:45 PM, "Anthony Scopatz" <notifications@github.com> wrote:

thanks a million!

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/xonsh/xonsh/pull/1728#issuecomment-246898215>, or mute
the thread

Issue #1731, opened 2016-09-14
More robust prompt formatter


laerus:
As is it now if there is a bug on any of the functions that return a value for the formatter dict the whole prompt turns ugly. Maybe we want a more robust way to handle the formatting and explicitly show an error about the specific value and not turn the prompt useless without any warnings.

scopatz:
Yup, I would definitely be in favor of this. I think it is probably just a small modification to the except block in the formatter func.

laerus:
also all the prompt formatting stuff should be moved to the `xonsh.prompt` package.

PR #1735, opened 2016-09-15
prompt formatting error handling + refactor

laerus:
#1731 

this is a first draft, the error message can be more friendly/informative,
ideas are appreciated.

moved functions that deal with prompt formatting from `environ` to `prompt.base`

### DIFF
--- a/xonsh/prompt/base.py
+++ b/xonsh/prompt/base.py
@@ -94,11 +94,12 @@ def _failover_template_format(template):
 
 def partial_format_prompt(template=DEFAULT_PROMPT, formatter_dict=None):
     """Formats a xonsh prompt template string."""
-    try:
-        return _partial_format_prompt_main(template=template,
-                                           formatter_dict=formatter_dict)
-    except Exception:
-        return _failover_template_format(template)
+    # try:
+    return _partial_format_prompt_main(template=template,
+                                       formatter_dict=formatter_dict)
+    # except Exception:
+    #     print("EXCEPTING")
+    #     return _failover_template_format(template)
 
 
 def _partial_format_prompt_main(template=DEFAULT_PROMPT, formatter_dict=None):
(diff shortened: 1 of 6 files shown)


gforsyth:
Hey @laerus -- I think you missed some imports in the tests:
```console
=================================== ERRORS ====================================
___________________ ERROR collecting tests/test_environ.py ____________________
tests\test_environ.py:12: in <module>
    from xonsh.environ import (Env, load_static_config,
E   ImportError: cannot import name 'partial_format_prompt'
============== 2687 passed, 15 skipped, 1 error in 36.90 seconds ==============
Command exited with code 1
```

laerus:
yeah i am moving the tests from `test_environ.py` to a new `test_prompt.py`
and some tests are failing, i think cause of the try-except block.

laerus:
okay it feels like there is some serious test leakage,
if i use `xonsh_builtins` fixture on `test_format_prompt_with_invalid_func` __without__ setting 
`__xonsh_env__` to an actual `xonsh.environ.Env` instance then it complains about not finding `is_manually_set` on the env object which is correct, but if i just remove `xonsh_builtins` it doesn't complain at all which means that it finds an `Env` instance afterall. (so much for isolation :sob:)

xonsh/prompt/base.py
```diff
@@ -1,13 +1,20 @@
...
 import socket
+import string
+import sys
 
 import xonsh.lazyasd as xl
 import xonsh.tools as xt
 import xonsh.platform as xp
 
+from xonsh.tools import format_color, print_exception
+from xonsh.platform import ON_WINDOWS, ON_CYGWIN
```
gforsyth: Since `xonsh.tools` and `xonsh.platform` are already imported and renamed above, should probably just use those.

laerus: thanks for catching this! i've cleanup them up

gforsyth:
Hey @laerus -- I finally got around to trying this and it's really nice!  I know you said it's a first draft, but I'm inclined to merge it in unless there's anything else you'd like to change.

laerus:
hey @gforsyth , if you think the error printing is clear and not too annoying then there is nothing else to add/change in this PR.

PR #1736, opened 2016-09-15
try,finally in xonsh_builtins fixture

laerus:
this came up in [comment](https://github.com/xonsh/xonsh/pull/1735#issuecomment-247352575)

not tested yet if that's the case but this seems more safe

not sure if this is excessive, maybe pytest handles that anyway

there is also this usual suspect

```
Exception ignored in: <bound method Execer.__del__ of <xonsh.execer.Execer object at 0x7f2bd3fd9a20>>
  File "/home/laerus/projects/xonsh/xonsh/execer.py", line 46, in __del__
  File "/home/laerus/projects/xonsh/xonsh/built_ins.py", line 1000, in unload_builtins
  File "/usr/lib/python3.5/abc.py", line 187, in __instancecheck__
AttributeError: 'NoneType' object has no attribute '_abc_invalidation_counter'
```

### DIFF
--- a/news/conftest-xonsh_builtins.rst
+++ b/news/conftest-xonsh_builtins.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+* ``xonsh_builtins`` fixture in conftest now uses try-finally block for more isolation
+
+**Deprecated:** None
+
+**Removed:** None
+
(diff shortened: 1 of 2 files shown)


laerus:
yeap that is excessive and it's not solving anything, closing this PR

PR #1740, opened 2016-09-17
Build docs on 3.5 otherwise we'll never caught failures.

Carreau:
Fix docs building and move docs building to a non-allow-failure version of Python.


laerus:
thank you @Carreau! i'm merging this.

PR #1746, opened 2016-09-21
help method on Env

laerus:
to format and print the Vardocs of an envvar

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -5,6 +5,7 @@ import re
 import sys
 import json
 import pprint
+import textwrap
 import locale
 import builtins
 import warnings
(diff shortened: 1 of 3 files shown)


xonsh/environ.py
```diff
@@ -773,6 +781,17 @@ def get_docs(self, key, default=VarDocs('<no documentation>')):
             self._docs[key] = vd
         return vd
 
+    def help(self, key):
+        """Get information about a specific enviroment variable."""
+        vardocs = self.get_docs(key)
+        width = width=min(79, os.get_terminal_size()[0])
```
gforsyth: drop one of the `width`s

gforsyth:
Hey @laerus, looks good!  Just the one inline comment.  It would be cool to have completion for the ENV keys but that can come later

xonsh/environ.py
laerus: oops thanks for catching this!

PR #1747, opened 2016-09-21
test_cdpath_expansion leaving stray dirs

laerus:
we should probably use the pytest fixture tmpdir but lets fix that for now.


gforsyth:
Thanks, @laerus !

Issue #1432, opened 2016-07-15
to quote or not to quote


laerus:
A summary of what seems to be the same issue for #413 #462 #621 #1092 #1148 #1388 #1390 #1402 #1084 #1742 and maybe more. (list growing)

EDIT: The answer at this moment is to 'quote', it seems that xonsh is drifting away from implementing a backslash escape mechanism, since quotes do exactly the same thing and its much clearer imo. Also some of these issues can be solved with subprocess macros. The best solution at this moment is to include an entry to the tutorial (or a separate tutorial) that explains how xonsh differs from other shells in that matter.
## 

I think this is a design decision that we will have to make sooner or later.

Bash has the motto 'in doubt use quotes', maybe we want to enforce this on xonsh and talk about it extensively in the tutorial. This will instantly solve almost all of the related issues.

Every other solution that comes in mind involves changes on how each command is being parsed, either at the Execer or at a lower level.

Ideas on this matter are appreciated!

scopatz:
> I think this is a design decision that we will have to make sooner or later.

I think the design decision has essentially been made. If someone magically handed us a backslah implementation, I am sure we would seriously discuss it.  But even so, that doesn't prevent the 

> 'in doubt use quotes',

motto at all.  So I am all for having more and clear documentation about this.  Maybe it deserves its own tutorial page because it is sort of a gotcha compared to other shell languages.

mitnk:
a bit of OT here, `git checkout -b foo-and-bar` to me is not an `in doubt`, but it does require me to use quotes to get what I wanted. Use quotes here is unnatural for me.

laerus:
> I think the design decision has essentially been made.
@scopatz is the parsing implementation at the Execer level subject to changes or is it off limits?

scopatz:
well, it is on the parser level, not the execer level. it is subject to changes.

scopatz:
Maybe I should clarify now that I am not on my phone :)  It is subject to change because we are pre-v1.0.  I think it is unlikely to change given the choices the core devs have made to this point. But if someone has a compelling reason for it, that would be great.  And if we had an implementation, that wouldn't hurt.

laerus:
can we implement something like #378 without changing the implementation of the `Execer.parse` ?

scopatz:
that particular example should just use @(f) which is more explicit

laerus:
[...]

Issue #1721, opened 2016-09-13
Auto quote file names with and/or


gforsyth:
We already do this for names with spaces. If a filename has the word `and` in it then the resulting command is almost certainly not what the user expects. Instead, the name should be wrapped in quotes automatically.

laerus:
with #1432 and all maybe we should generalize this a little more and quote more args in subproc mode?

gforsyth:
That's a good conversation to have.  i should clarify that I mean this specifically in completion results

scopatz:
I am pro this!

gforsyth:
Closed via #1748

PR #1748, opened 2016-09-21
auto quote completion paths that contain 'and', 'or'

laerus:
#1721

### DIFF
--- a/news/completions.rst
+++ b/news/completions.rst
@@ -8,6 +8,8 @@
 
 **Fixed:**
 
+* Auto quote path completions if path contains 'and' or 'or'
+
 * Completion now works on subcommands after pipe, ``&&``, ``||`` and so on.
 
 **Security:** None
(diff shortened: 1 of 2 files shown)


gforsyth:
Ha.  I was _just_ working on this.

gforsyth:
Yeah, this works well.  Thanks, @laerus, you beat me to it!  Just waiting on CI

laerus:
this is a little confusing if you don't know how this works:

```
>>> mkdir -p 'test/and/or/okay'
>>> tree
.
└── test
    └── and
        └── or
            └── okay
>>> cd 'test/and/' <---cursor here
>>> cd 'test/and/'o <--- hitting tab here
>>>cd 'test/and/or/' <--- works as it should
```
one may be tempted to remove the last single quote to continue the completion
even though it doesn't have to.
(maybe we should mention this in the docs, well after we resolve #1432 imo)

gforsyth:
I agree that it can use a mention in the docs.  The same behavior shows up for completing path names with other quoted characters

gforsyth:
Thanks, @laerus !  Glad to have this in.

PR #1749, opened 2016-09-21
add support for windows drives in subsequence matching

gforsyth:
Subsequence matching was crawling up a directory tree and looking for an empty base, but on windows it's going to be a drive letter.  

pinging @astronouth7303 to see if this works.

### DIFF
--- a/xonsh/completers/path.py
+++ b/xonsh/completers/path.py
@@ -267,7 +267,7 @@ def complete_path(prefix, line, start, end, ctx, cdpath=True, filtfunc=None):
             if p[0] == '':
                 basedir = ('', )
                 p = p[1:]
-            elif re.match(RE_WIN_DRIVE, p[0]):
+            elif xp.ON_WINDOWS and re.match(RE_WIN_DRIVE, p[0]):
                 basedir = (re.match(RE_WIN_DRIVE, p[0]).group(), )
                 p = p[1:]
             else:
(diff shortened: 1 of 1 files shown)


astronouth7303:
What's the user-visible change?

gforsyth:
ah, sorry, that you shouldn't get infinite recursion on path completion when `$SUBSEQUENCE_PATH_COMPLETION=True`.

And ideally, also that subsequence completion works

xonsh/completers/path.py
```diff
@@ -265,6 +267,9 @@ def complete_path(prefix, line, start, end, ctx, cdpath=True, filtfunc=None):
             if p[0] == '':
                 basedir = ('', )
                 p = p[1:]
+            elif re.match(RE_WIN_DRIVE, p[0]):
```
laerus: maybe check `if ON_WINDOWS and ...` here?

astronouth7303:
I thought the infinite recursion bug was already fixed?

gforsyth:
You mentioned in #1713 that you were still having recursion issues with the subsequence matching.  If you aren't, we can scrap this.

astronouth7303:
Nope. This branch doesn't fix #1691 either.

gforsyth:
Damn.  Oh well.  I can take a crack at this sometime when I have a windows machine to play with.

PR #1752, opened 2016-09-22
do not quote unnessasery and/or

blahgeek:
`ls and/or` should be quoted, but `ls andor` should not. only regex `\band\b|\bor\b` is searched

### DIFF
--- a/xonsh/completers/path.py
+++ b/xonsh/completers/path.py
@@ -1,4 +1,5 @@
 import os
+import re
 import ast
 import builtins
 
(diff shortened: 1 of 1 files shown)


laerus:
Hey @blahgeek, why do you think this is an issue? i guess we can change the behavior again but this quoting thing is already out of hand with #1432.

blahgeek:
Hi @laerus , I didn't intent to change the behavior to quote "and"/"or" when it would otherwise recognized as python keyword. What I am trying to fix here is, for example, in `vim xonsh/history.py`, the quoting for `xonsh/history.py` is unnessasery (even if there's "or" inside "history.py").

blahgeek:
I just changed `and`/`or` to `\band\b`/`\bor\b` (regex whole word) in effect.

laerus:
Alright, thank you @blahgeek!

PR #1755, opened 2016-09-24
Allow empty XONSH_GITSTATUS_ defs

t184256:
builtins.__xonsh_env__.get('XONSH_GITSTATUS_' + key) or _DEFS[key]
swallowed not only None, but also empty strings, and was replaced
with an explicit check for None.

### DIFF
--- a/xonsh/prompt/gitstatus.py
+++ b/xonsh/prompt/gitstatus.py
@@ -36,7 +36,8 @@ def _DEFS():
 
 
 def _get_def(key):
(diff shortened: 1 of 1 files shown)


laerus:
thanks for putting this in @t184256! I thinkwe should provide the `$XONSH_GITSTATUS` as a dictionary just like `$FORMATTER_DICT` so they are all on one place and we can actually inspect what's what.

scopatz:
Thanks @t184256! @laerus is there any reason not to merge this for the time being, though?

laerus:
nope let's merge it, cause i'm working on a prompt refactor.

PR #1757, opened 2016-09-24
implements $BOTTOM_TOOLBAR to create a bottom toolbar

santagada:
This implements support for using prompt toolkit bottom toolbar functionality.

It needs tests, docs, news entry but more important do you guys think we should do it this way? Bottom toolbar on prompt-toolkit support being updated per keystroke and this might be very useful (a user types a command we might show the path to the binary for example, or file mode)


santagada:
screenshot 2016-09-24 15 40 57

using it on my powerline xontrib, it seems useful to have a lot of info about the current position without taking space specially when running multiple small commands.

xonsh/environ.py
```diff
@@ -266,6 +267,7 @@ def DEFAULT_VALUES():
         'PUSHD_SILENT': False,
         'RAISE_SUBPROC_ERROR': False,
         'RIGHT_PROMPT': '',
+        'BOTTOM_TOOLBAR': '',
```
scopatz: Can you also please add a VarDocs entry :)

scopatz:
This is awesome @santagada! I think this is definitely worth having. Updating every keystroke might be the right thing to do here.  Or maybe we should have a flag for this sort of thing.

scopatz:
The code looks good to me other than the docs issue.

laerus:
ppl are banging at the prompt these days, sorry @santagada can you please fix the conflicts and add the vardoc entry so we can merge this asap?

santagada:
I'm on vacation, if anyone can do this before I'm back on Saturday please be my guest

santagada:
@laerus updated the PR and added vardoc and a news entry, please take a look again :)

xonsh/ptk/shell.py
```diff
@@ -157,6 +162,23 @@ def rprompt_tokens(self, cli):
...
 
+    def bottom_toolbar_tokens(self, cli):
+        """Returns a list of (token, str) tuples for the current bottom
+        toolbar.
+        """
+        p = builtins.__xonsh_env__.get('BOTTOM_TOOLBAR')
+        # partial_format_prompt does handle empty strings properly,
+        # but this avoids descending into it in the common case of
+        # $TOOLBAR == ''.
+        if isinstance(p, str) and len(p) == 0:
```
laerus: This can be just `if p == ''`

laerus:
Thank you @santagada, I think hit is ready to go in! Just one comment, i know the other prompt parts are written like this, you can change the other checks as well or we can merge now. There should be a proper prompt refactor at some point.

santagada:
I think it is better to have it as a copy+paste of the right_prompt right now, later we can improve it all.

I want to improve how creating a new env var works, adding the same thing on three different places (definition, default value, vardoc) seems unnecessarily complicated.

Issue #1743, opened 2016-09-19
Can't keyboard interrupt while loop


asmeurer:
If you have a while loop with a command, like

```
while True:
    sleep 1
```

you can't interrupt it. Ctrl-c just restarts the loop.

asmeurer:
Even 

```py
while True:
     try:
         sleep 1
     except KeyboardInterrupt:
         break
```

doesn't work. And the only apparent way to exit is to kill xonsh from another terminal.

scopatz:
This seems to be a subproc / job control issue since 

```
scopatz@artemis ~ $ import time
scopatz@artemis ~ $ while True:
`·.,¸,.·*¯`·.,¸,.·*     time.sleep(1)
`·.,¸,.·*¯`·.,¸,.·*     
^Cscopatz@artemis ~   
```

Can be Ctrl+C'd just fine. Thanks for reporting @asmeurer!

adqm:
I'm not sure I would classify this as a bug, nor am I sure there's an easy way to fix it.

The "problem" (I suspect, based on what I've seen in other cases) is that the Ctrl+C is killing the subprocess (`sleep`) and so is never even reaching xonsh (since we give the terminal over to child processes when they start).  Then another `sleep` process starts up almost immediately thereafter, so Ctrl+C again reaches that process instead of xonsh.

I've had some amounts of success getting around this with:
* press and hold Ctrl+C and wait until it happens to catch xonsh in between killing one `sleep` and starting the next
* press Ctrl+\

That said, I'm not sure there is a real right way to "fix" this, unless we run the whole loop in a subshell.

scopatz:
For what is it worth this is fixed on the run subprocess branch.  The fix involves sending SIGINT to the subprocess, and then if this caused the suprocess to exit, PopenThread will call the original handler from Python.  This will cause Python to walk back up the stack until the shell KeyboardInterrupt handling is reached.

scopatz:
Though Crtl+\ also works!

gforsyth:
Closed via #1758

asmeurer:
This is still happening for me in master. It's especially annoying when running a script, because it becomes difficult to kill (you have to spam Ctrl-C until it dies.

gforsyth:
Damn.  Sorry about that!

scopatz:
Ok, something weird is going on here. On master this works fine for me.  On the new perf branch, the behaviour is as described...

asmeurer:
[...]

PR #1758, opened 2016-09-24
Run Subproc Refactor

scopatz:
This is a major overhaul of the way subprocs are executed.  

The big idea here is that sub processes are now split into three pieces:

* A process specification, which describes the process to be run and may be manipulated prior to execution
* A process object, like Popen or ProcProxy, responsible for running a single command.
* A command pipeline, which holds all specs and procs and in responsible for making sure that all procs in the pipeline read from and cleaned up correctly.

To enable the streaming and iterating features that we desired, a new PopenThread class has been introduced.  This adheres to the Popen API and should be the last non-callable proc in the pipeline.  This is responsible for running a regular Popen proc on a separate thread, and reading its output in a non-blocking manner (which was tricky!).

This PR also removes the TeePTY class and related psuedo-terminal handling.  This PR should thus enable Storing stdout on all platforms now!

Please bang on this as there are things I have probably missed.


scopatz:
OK, this seems to work on Mac and Linux.  Appveyor was hanging indefinitely, though, for reasons I don't understand.  Has anyone seen this behaviour before?

laerus:
this is an awesome work! I will get my hands on this branch asap.

scopatz:
OK, I have figured out why this is failing on windows and I have a fix.  Just need to make the fix robust...

scopatz:
Well, CI finally passes!

laerus:
piping to `less` causes `less` to be interrupted, for example `cat somefile | less`

this branch solves #1743 !

no sure if this is related:
there is something strange going on with `which`, it doesn't return anything when i `which less` also 
on `which --help` i need to Ctrl-C to get out of the output.

gforsyth:
Hey @laerus, I believe the `which` issue is related -- all of the `print` statements in `which` have a `file=stdout` at the end.  If we remove it then the output is visible.  Not sure why the help is hanging yet.  
Still going through the code here to figure out how to pass the right output handles into `which`

I can confirm the problem with `less`.  Also, it seems like `htop` and probably other ncurses interfaces are not being granted control over the keyboard.  Only Ctrl-C seems to work to exit out.

gforsyth:
Hmm, another issue related to `stdout` stuff -- in `IPython`, the "are you sure?" exit prompt doesn't get printed.

laerus:
i can confirm the htop issue, it seems like tty echo is turned on, i can exit with q + enter

iterating over `CommandPipeline` gets output as it's produced, so something like:
```
for line in  !(yes):
    print(line)
```
actually works! thanks for this @scopatz

scopatz:
Thanks for banging on this @laerus and @gforsyth! I have most of the issues that you mentioned.

* The Crtl+C in less was caused by not handling SIGINT correctly.  This has been fixed.
* The which issue was caused by not redirecting stdout and stderr for proxies. It is now.
* The IPython issue was caused because I forgot to stream stderr :)

The htop echo issue is partially solved. If I turnoff echo on stdin, I can't seem to turn it back on successfully.  Which is problematic.

Also, I can definitely `cat file | less` and everything works as expected.  I can even `less < somefile`. I am curious if this issue is fixed now...

Also, I have expanded the alias API and reduced the number of proxy classes.  I haven't documented this yet though.

scopatz:
I am glad that this fixes #1743 and the iteration works. There is some progress here :)  There are an insane number of things to keep track of with this.

laerus:
i still get interrupted when `cat | less`
also the output of some commands is getting cut off
like single `ls`, 
note i'm on a poor 6yo netbook with debian-stable atm and the machine it's pretty slow itself.

laerus:
`!(ls)` sometimes leaks output.
same with `!(cat filename)`

usually on new pwd/filename

laerus:
there is some tee output redirection on `base_shell.py` too, maybe it is interfering?

scopatz:
The tee output on base shell shouldn't affect this since it is only for Python / Xonsh objects.  

I think I understand what is causing the `while True: sleep 1` issue.  Basically, the fix for SIGINT doesn't raise an error if the process failed because of a Crtl+C.  It probably should.

gforsyth:
Hey @scopatz -- not _entirely_ sure why, but if you run the tests with the `-s` flag and disable stdout capture, then all of the tests pass on my machine.

scopatz:
Thanks for pointing out `-s` @gforsyth!  I think this is a result of turning echo off, or pytest running in its own pipe.  A work around should be pretty easy, but I was busy squashing other bugs last night :).

scopatz:
@laerus - Thanks for these test cases.  I think I have most of these working - with the exception of Bash, which I'll address tonight, hopefully.

Note that:

```
$ outs = []
$ for i in range(30):
.     outs.append(!(echo @('TEST'*100)).out)
.     outs.append(!(echo @('TEST'*100)).out)
.     outs.append(!(echo @('TEST'*100)).out)
.
$ len(set(outs)) 
5 <-- pretty random
```

This is actually correct.  This is because `!(cmd)` is a streaming operator and so you need to iterate through it at least once to get the full result.  `out.append()` won't perform this streaming so you are just getting whatever partial state the command is in.  `!(cmd)` is more like map(), range(), iter() etc than anything else.

scopatz:
Ok! Things *seem* to work now.  I am hopeful that this is now good enough.  But please keep banging on it :)  

I am not super pleased by the trick to get bash and other shells working, but I can't think of another way to do it, other than having users run bash et al. through `$[sh]`, which I am sure people would be less than thrilled by.

gforsyth:
Hey @scopatz, amazing work here.  
Everything I tried works except for the completion in `xonda`.  
I'm only posting it in case it indicates some larger potential issue.  I'm happy to rewrite the completion routines to work around this, although I can't quite figure out what it's unhappy about since the `get_envs` command works fine when entered on its own.

This is triggered when I hit TAB after typing out `conda activate `

```console
gil@theo ~ 🐚   conda activate Exception in thread Thread-45:                 
  File "/home/gil/anaconda/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/home/gil/anaconda/lib/python3.5/threading.py", line 862, in run
  File "/home/gil/anaconda/lib/python3.5/site-packages/prompt_toolkit/interface.py", line 836, in run
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/ptk/completer.py", line 42, in get_completions
    self.ctx)
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/completer.py", line 36, in complete
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonda-0.1.5-py3.5.egg/xontrib/xonda.xsh", line 93, in xonda_completer
    possible = set(get_envs())
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonda-0.1.5-py3.5.egg/xontrib/xonda.xsh", line 13, in get_envs
    return [env for env in $(ls @(envs[0])).split('\n')[:-1]]
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/built_ins.py", line 797, in subproc_captured_stdout
    return run_subproc(cmds, captured='stdout')
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/built_ins.py", line 753, in run_subproc
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/built_ins.py", line 496, in run
    p = self._run_binary(kwargs)
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/built_ins.py", line 505, in _run_binary
    p = self.cls(self.cmd, bufsize=bufsize, **kwargs)
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/proc.py", line 276, in __init__
    self.old_int_handler = signal.signal(signal.SIGINT, self._signal_int)
  File "/home/gil/anaconda/lib/python3.5/signal.py", line 47, in signal
    handler = _signal.signal(_enum_to_int(signalnum), _enum_to_int(handler))
ValueError: signal only works in main thread
```

scopatz:
> Hey @scopatz, amazing work here.

Thanks @gforsyth!  

That issue you found with xonda was a legitimate bug with running subproc commands on a background thread and signal handling. I have pushed up some changes that make this more thread safe. 

I am not sure why appveyor thinks this is unmergable, since it clearly is.  Note that the branch build does succeed.

gforsyth:
Yurgh.  Sorry.  
So this is really weird.  Those fixes handle the signal problem, but now when the list of completions comes up I can't select them.  TAB just advances the cursor by 4 spaces and I have to Control-D to escape to a new line.

Then I tried it in a readline shell and the first command I enter doesn't get written to screen.  It's captured and works, but it's invisible.

scopatz:
Ok, I'll have to play around with this later today.  It is not immediately apparent what is causing this. I can't seem to reproduce it with just a thread that launches a subprocess. There must be something funny going on with ptk / readline fighting for control over stdin with xonsh while inputting code. It is like the cbreak mode is never restored... If you happen to see an error message, let me know!

gforsyth:
Possibly unrelated issue with echoing stdin:  The prompt when exiting IPython prints the "Are you sure?" message but the user response doesn't appear.  
Similar problem when entering info for `cookiecutter`

PaulReiber:
Anthony wrote:
> There must be something funny going on with ptk / readline fighting for
control over stdin with xonsh while inputting code

Indeed, I think maybe it's the same echo concern we hit here:
https://github.com/xonsh/xonsh/issues/1368

In there I attempted to use tcsetattr to correct the mode on stdin but it
didn't help.

Just now I pulled down the source for gnureadline and grepped for tcsetattr

$ cd ~/readline_source; grep -R tcsetattr .
Binary file ./readline-7.0.tar matches
./readline-7.0/rltty.c:#    define SETATTR(tty, tiop)   (tcsetattr
(tty, TCSANOW, tiop))
./readline-7.0/rltty.c:#    define SETATTR(tty, tiop)   (tcsetattr
(tty, TCSADRAIN, tiop))
./readline-7.0/examples/rlptytest.c:  if(tcsetattr(fd, TCSAFLUSH, &buf) < 0)
./readline-7.0/examples/rlptytest.c:  if(tcsetattr(fd, TCSAFLUSH, &buf) < 0)
./readline-7.0/examples/rlptytest.c:   if(tcsetattr(fd, TCSAFLUSH,
&save_termios) < 0)
./readline-7.0/examples/rlfe/rlfe.c:  tcsetattr(STDIN_FILENO, TCSANOW,
&orig_term);
./readline-7.0/examples/rlfe/rlfe.c:  tcsetattr(STDIN_FILENO, TCSANOW, &t);
./readline-7.0/examples/rlfe/rlfe.c:          tcsetattr(STDIN_FILENO,
TCSANOW, &orig_term);
./readline-7.0/examples/excallback.c:    if( tcsetattr(STDIN_FILENO,
TCSANOW, &term) < 0 ) {
./readline-7.0/examples/excallback.c:        perror("tcsetattr");
./readline-7.0/examples/excallback.c:    if( tcsetattr(STDIN_FILENO,
TCSANOW, &term) < 0 ) {
./readline-7.0/examples/excallback.c:        perror("tcsetattr");
pbr@mongo:~/readline_source$ grep -R SETATTR .
./readline-7.0/rltty.c:#    define SETATTR(tty, tiop)   (tcsetattr
(tty, TCSANOW, tiop))
./readline-7.0/rltty.c:#    define SETATTR(tty, tiop)   (tcsetattr
(tty, TCSADRAIN, tiop))
./readline-7.0/rltty.c:#  define SETATTR(tty, tiop)     (ioctl (tty,
TCSETAW, tiop))
./readline-7.0/rltty.c:  while (SETATTR (tty, tiop) < 0)


Reviewing the C code from there, I found  _rl_echoing_p which looks
promising - I'm continuing to review the code.  The README for "rlfe"
example appears to have identified similar concerns.  Possibly you guys
want to get readline's maintainer(s) in the loop on this?  They may be able
to help with sussing out the remainder of what needs to be done.


Kind regards,


gforsyth:
Define an alias function that calls something in subproc mode:
```console
gil@theo ~ 🐚  def _pwd(args, stdin=None):
..............     pwd
..............     
gil@theo ~ 🐚  aliases['mypwd'] = _pwd
```
Run it once
```console
gil@theo ~ 🐚  mypwd
/home/gil
Hangs here
```

Ctrl-C to break the hang and run it again
```console
gil@theo ~ 🐚  mypwd
/home/gil
```
Works as expected

Run it again
```console
gil@theo ~ 🐚  mypwd
Exception in thread Thread-44:
  File "/home/gil/anaconda/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 9810, in run
    sp_stdout.flush()
```
This error kills the current xonsh instance, traceback continues in a parent instance:

```console
gil@theo ~ 🐚                                                                                            
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/ptk/shortcuts.py", line 108, in prompt
  File "/home/gil/anaconda/lib/python3.5/site-packages/prompt_toolkit/interface.py", line 389, in run
  File "/home/gil/anaconda/lib/python3.5/site-packages/prompt_toolkit/eventloop/posix.py", line 110, in run
    fds = self._ready_for_reading(remaining_timeout)
  File "/home/gil/anaconda/lib/python3.5/site-packages/prompt_toolkit/eventloop/posix.py", line 183, in _ready_for_reading
    fds = select_fds(read_fds, timeout)
  File "/home/gil/anaconda/lib/python3.5/site-packages/prompt_toolkit/eventloop/select.py", line 35, in select_fds
    result = _python3_selectors(read_fds, timeout)
  File "/home/gil/anaconda/lib/python3.5/site-packages/prompt_toolkit/eventloop/select.py", line 69, in _python3_selectors
    sel.register(fd, selectors.EVENT_READ, None)
  File "/home/gil/anaconda/lib/python3.5/selectors.py", line 411, in register
    self._epoll.register(key.fd, epoll_events)

  File "/home/gil/anaconda/bin/xonsh", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/home/gil/git/xonsh/scripts/xonsh", line 3, in <module>
    main()
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 17925, in main
    shell.shell.cmdloop()
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 118, in cmdloop
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 87, in singleline
    line = self.prompter.prompt(**prompt_args)
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/ptk/shortcuts.py", line 113, in prompt
    eventloop.close()
  File "/home/gil/anaconda/lib/python3.5/site-packages/prompt_toolkit/eventloop/posix.py", line 252, in close
    os.close(schedule_pipe[0])
```

scopatz:
@PaulReiber The saddest phrase I have read today is "the readline maintainer" :)

scopatz:
Ho-kay! Here we go again. I think I have fixed the proxy deadlock problem.  I also moved the cbreak mode to only trigger for alternate mode, and every thing seemed to work.  I haven't tested it with xonda.  I dare to dream that this is it again....

gforsyth:
Hey @scopatz, I am not relishing my role as crusher of dreams...

`xonda` works, input on IPython works, breaking while loops works, proxy deadlocks are gone, but now backgrounding jobs is broken...

scopatz:
I'd take a dream crusher over a Wesley, any day. (:snap: 90s ref.)

Ok, so by backgrounding, do you mean starting a process in the background with `&` or putting it there with ^Z and then bg?

gforsyth:
Shut up, Wesley!

I meant ^Z and bg, but starting processes in the background also appears to not work.

gforsyth:
Ahh, damn.  Also, piping between xonsh builtins and bash seems to be broken, e.g.

`history show bash | grep git`

which works on master but not here.

laerus:
`cat somefile | less` still gets interrupted for me.

this still happens:
```
$ for i in range(100):
.     ls1 = $(ls)
.     ls2 = $(ls)
.     if ls1 != ls2:
.         print('MISMATCH')
.         break
.     
MISMATCH
```

on this slow machine in this branch i'm experiencing some slowdown on fullscreen commands
like man, top, less(with big files) and it's like i can see the screen refreshing in slow motion.

scopatz:
@gforsyth did you have a specific command that you were looking at for ^Z? I tried it last night and it seemed pretty broken on both.

@laerus - OK that is probably a matter of tuning the timeout. Maybe I need a shorter value, but I could also make this user-tunable too.

gforsyth:
@scopatz -- The only thing I ever really use it for is connecting to the university VPN using `openconnect`

xonsh/proc.py
```diff
@@ -571,6 +571,166 @@ def __repr__(self):
...
+
+    @property
+    def available(self):
+        """True if the thread is available in the registry."""
+        return threading.get_ident() in self.registry
+
+    @property
+    def handle(self):
+        """Gets the current handle for the thread."""
+        return self.registry(threading.get_ident(), self.default)
```
laerus: missing `.get` ?

scopatz: Yep, but that was fixed a few commits ago

laerus:
oh you actively on this. sorry
ps reviewing system is odd

scopatz:
Yeah, it is.

laerus:
this is a lot more complex than i thought it would be

scopatz:
> this is a lot more complex than i thought it would be

You and me both! This is probably some of the most complex code I have ever written. Basically, everything is an  edge case. But, on the other hand, what we are asking xonsh to do *is* sort of crazy.

scopatz:
Ho-kay! I want to believe that all the concerns have mentioned to date have now been addressed.  Preparing for my dreams to be crushed...

laerus:
```
$ xonfig
Exception in thread Thread-10:
  File "/usr/lib/python3.4/threading.py", line 920, in _bootstrap_inner
    self.run()
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/proc.py", line 1020, in run
    self.returncode = parse_proxy_return(r, sp_stdout, sp_stderr)
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/proc.py", line 816, in parse_proxy_return
    stdout.write(r)
ValueError: I/O operation on closed file.

^C
$ # xonsh okay
```

laerus:
```
$ for i in range(100):
.     ls1 = ''.join(!(ls))
.     ls2 = ''.join(!(ls))
.     if ls1 != ls2:
.         print('MISMATCH')
.         
. 
MISMATCH
MISMATCH
MISMATCH
MISMATCH
MISMATCH
MISMATCH
MISMATCH
MISMATCH
```

man/top/less a lot more responsive

`bash` works!

`cat fname | less` works!

-

a little random on new xonsh instance:
```
$ !(ls)
CommandPipeline(stdin=<_io.BytesIO object at 0xb569205c>, stdout=<_io.BytesIO object at 0xb56920bc>, stderr=<_io.BytesIO object at 0xb56920ec>, pid=24157, returncode=0, args=['ls'], alias=['ls', '--color=auto', '-v'], stdin_redirect=['<stdin>', 'r'], stdout_redirect=[7, 'wb'], stderr_redirect=[9, 'w'], timestamps=[1475327836.8005934, None], executed_cmd=['/bin/ls', '--color=auto', '-v'], input=None, output=, errors=None)
$ # if i hit Ctrl-C here
  File "/home/laerus/.local/bin/xonsh", line 3, in <module>
    main()
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/main.py", line 221, in main
    shell.shell.cmdloop()
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/readline_shell.py", line 397, in cmdloop
    self._cmdloop(intro=intro)
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/readline_shell.py", line 357, in _cmdloop
    line = self.singleline()
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/readline_shell.py", line 249, in singleline
    rtn = input(self.prompt)
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/proc.py", line 476, in _signal_int
    self.send_signal(signum)
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/proc.py", line 599, in send_signal
    return self.proc.send_signal(signal)
  File "/usr/lib/python3.4/subprocess.py", line 1681, in send_signal
    os.kill(self.pid, sig)
ProcessLookupError: [Errno 3] No such process
```
and xonsh crashes

scopatz:
Hi @laerus - thanks for noticing these issues.  I think I have squashed them now with the exception of the ls mismatch issue.  I cannot seem to reproduce that.  Even if I go up to 1000 iterations.  Can you maybe try it with a different command, like echo, that doesn't depend on the state of your file system.  It is possible that something is adding and removing files durring these calls.

astronouth7303:
Running the test suite locally fails for me. This could be local weirdness, though.

```
____________________________________________________________________ test_bin_ls ____________________________________________________________________

    @skip_if_on_windows
    def test_bin_ls():
>       assert check_eval('/bin/ls -l')

tests/test_execer.py:26: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/tools.py:125: in check_eval
    builtins.__xonsh_execer__.eval(input)
xonsh/execer.py:132: in eval
    return eval(code, glbs, locs)
<xonsh-code>:1: in <module>
    ???
xonsh/built_ins.py:821: in subproc_captured_hiddenobject
xonsh/built_ins.py:756: in run_subproc
xonsh/built_ins.py:496: in run
    p = self._run_binary(kwargs)
xonsh/built_ins.py:505: in _run_binary
    p = self.cls(self.cmd, bufsize=bufsize, **kwargs)
xonsh/proc.py:281: in __init__
    self._set_pty_size()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <PopenThread(Thread-235, initial)>

    def _set_pty_size(self):
        """Sets the window size of the child pty based on the window size of
            our own controlling terminal.
            """
        if not os.isatty(self.stdout_fd):
            return
        # Get the terminal size of the real terminal, set it on the
        #       pseudoterminal.
        buf = array.array('h', [0, 0, 0, 0])
        # 1 = stdout here
>       fcntl.ioctl(1, termios.TIOCGWINSZ, buf, True)
E       OSError: [Errno 25] Inappropriate ioctl for device

xonsh/proc.py:467: OSError
```

I'm not sure why this doesn't turn up in CI.

laerus:
kitten-hang-in-there

-

```
$ for i in range(100):
.     echo1 = ''.join(!(echo @('TEST'*100)))
.     echo2 = ''.join(!(echo @('TEST'*100)))
.     if echo1 != echo2:
.         print('MISMATCH')
.   
MISMATCH
... x10+      
```

astronouth7303:
@laerus so far I've been able to reproduce at a rate of about .05%, and it looks like the failure mode is incomplete STDOUT on the part of `ls1`. (Although a second run is showing a much higher rate.)

I think this is a related failure:

```
🐚 for i in range(10000):
.     if not ''.join(!(ls)).endswith('\n'): 
.         errors += 1
.         print(errors)
```

(I got an error rate of .1% in this case.)

laerus:
i'm on a pretty slow machine atm, i even dropped to readline but it's still happening.

astronouth7303:
I also get errors with the command `echo @("*"*500)`.

PaulReiber:
> @PaulReiber <https://github.com/PaulReiber> The saddest phrase I have read
 today is "the readline maintainer" :)

I'll revisit the code base again to see if I can glean anything further.

Kind regards,


gforsyth:
Hey @scopatz -- my remaining problems seem to have been mopped up somewhere along the way.  The only thing I have left (I think) is a strange little thing.  I can open `bash` from `xonsh` if I'm using the prompt_toolkit shell, but if I use readline then it immediately exits with

```console
gil@bad_cat ~ $ bash
bash: cannot set terminal process group (-1): Inappropriate ioctl for device
bash: no job control in this shell
┌[gil] [bad_cat]:~ 
└>exit
```

scopatz:
With @astronouth7303's example, I see no breakages.  This seems to be machine dependent somehow.

```
scopatz@artemis ~ $ errors = 0
scopatz@artemis ~ $ for i in range(10000):
`·.,¸,.·*¯`·.,¸,.·*     if not ''.join(!(ls)).endswith('\n'):
`·.,¸,.·*¯`·.,¸,.·*         print(errors)
`·.,¸,.·*¯`·.,¸,.·*         
`·.,¸,.·*¯`·.,¸,.·* 
scopatz@artemis ~ $ errors
0
```

scopatz:
> I'm not sure why this doesn't turn up in CI.

This issue is with some pytest weirdness and them swapping out stdout...apparently with something that is a tty, but doesn't have a size set.  I suspect it might even be a pytest 2 vs pytest 3 issue.  In any event, I'll make this more robust.

scopatz:
@laerus @astronouth7303 - when you are missing output, are you missing it from the begining or end of the command?  I have added some more sleep to the output, just in case.

@PaulReiber - thanks! I think I figured out a way around that issue, though. Also the original phrase was meant as a joke, but it wasn't funny. All apologies.

@gforsyth - I have no idea what is going on with bash and readline. I have tried every combination of settings I can think of and everything works.  Also, it doesn't make sense that readline shell would affect anything since the for bash here it to *avoid* PopenThread and just use Popen normally.  Do you have some fancy inputrc or envvars or similar that could affect readline or Bash?

scopatz:
@laerus @astronouth7303 - I have added an envvar called `$XONSH_PROC_FREQUENCY` that controls the sleeps in the process loops. The default is 1e-4, which is what I have been using. However, I found that if I drove this down to 1e-8, it would break consistently in the way described, but leaving off the end of the command.  Can you try setting this value higher and seeing if/when it goes away, please?  This may just need to be the sort of thing users need to tune to their machine. We can probably provide an autotuner to help.  But I would like to keep as much out of this PR as possible.  I am pretty much at end of my rope for this.  I think it already is a vast improvement over what is on master presently...

astronouth7303:
From the end. I figured there was one chunk in the buffer not getting read after the process exited.

I'm still getting the error at a rate of .295%.

scopatz:
> At 0.0005, it drops to .115%.

What if you do order of magnitude bumps, like 1e-3 or 1e-2?

> And frankly, we shouldn't have such a tweakable. No other shell has this problem, and users shouldn't have to performance tune to get their shell reliable.

I understand it is a hack to get things working for the moment. I am interested if there is a frequency at which this never happens.

astronouth7303:
Ok, today:

|`$XONSH_PROC_FREQUENCY` | Error Rate |
| -------------------------------------------- | --------------- |
| 0.0001 | .795% |
| 0.0005 | .0725% |
| 0.001 | | .016% |

I'm currently using this version of the test:
```
for i in range(20000):
    if len(''.join(!(echo @("*"*500)))) != 501: 
    errors += 1
    print(errors)
```

laerus:
100 iterations of this test version:
0.0001 -> 0.07%
0.0005 -> 0.06%
0.001 -> 0.02%
0.005 -> 0.01%
0.01 -> 0% (also tested with 1k iterations)

tests/test_run_subproc.py
```diff
@@ -0,0 +1,47 @@
...
+import pytest
+
+from xonsh.platform import ON_WINDOWS
+from xonsh.built_ins import run_subproc
+
+from tools import skip_if_on_windows
+
+
+@pytest.yield_fixture(autouse=True)
+def chdir_to_test_dir(xonsh_builtins):
```
melund: Better use an actual temp dir here instead. This is something we need to change many places in the test suite. 

```
def chdir_to_test_dir(tmpdir):
    with tmpdir.as_cwd():
        yield
```

tests/test_run_subproc.py
```diff
@@ -0,0 +1,47 @@
...
+
+@skip_if_on_windows
+def test_runsubproc_simple(xonsh_builtins, xonsh_execer):
+    new_cwd = os.path.dirname(__file__)
+    xonsh_builtins.__xonsh_env__['PATH'] = os.path.join(new_cwd, 'bin') + \
+        os.pathsep + os.path.dirname(sys.executable)
+    xonsh_builtins.__xonsh_env__['XONSH_ENCODING'] = 'utf8'
+    xonsh_builtins.__xonsh_env__['XONSH_ENCODING_ERRORS'] = 'surrogateescape'
+    xonsh_builtins.__xonsh_env__['XONSH_PROC_FREQUENCY'] = 1e-4
+    if ON_WINDOWS:
```
melund: The whole test is skipped on windows, so why this?

scopatz: I put this in there for when someone fixes this test for windows.

tests/test_run_subproc.py
scopatz: Well, the purpose of this is to get the xonsh/tests/bin dir on the relative path, not to create or modify files. The redirect one below does do some of that, but I think that this is fine for now.  A refactor of that would be nice.

tests/test_run_subproc.py
melund: No problem. Maybe, something I could do when this goes into master.

melund:
This is really crazy impressive.

tests/test_run_subproc.py
melund: I see. I do have some unpushed fixes to the test suite. But I think it should wait until this goes in.

scopatz:
> This is really crazy impressive.

Thanks!

Also @astronouth7303 and @laerus, I have added some code that I think should prevent the dropping of output when the frequency is too low.  This get all of the output on my machine even when `f = 1e-20`.  Would you mind trying this out?  Thanks!

gforsyth:
Still some issues with `XONSH_STORE_STDOUT`:

```console
gil@bad_cat ~/git/xonsh rs 🐚  xonsh --no-rc
gil@bad_cat ~/git/xonsh rs $ $XONSH_STORE_STDOUT=True
gil@bad_cat ~/git/xonsh rs $ $XONSH_SHOW_TRACEBACK
True
gil@bad_cat ~/git/xonsh rs $ ls
  File "/home/gil/miniconda3/lib/python3.5/site-packages/xonsh/base_shell.py", line 176, in default
  File "/home/gil/miniconda3/lib/python3.5/site-packages/xonsh/codecache.py", line 65, in run_compiled_code
    func(code, glb, loc)
  File "/home/gil/miniconda3/lib/python3.5/site-packages/xonsh/built_ins.py", line 821, in subproc_captured_hiddenobject
  File "/home/gil/miniconda3/lib/python3.5/site-packages/xonsh/built_ins.py", line 789, in run_subproc
    command.end()
  File "/home/gil/miniconda3/lib/python3.5/site-packages/xonsh/proc.py", line 1528, in end
    for _ in self.tee_stdout():
  File "/home/gil/miniconda3/lib/python3.5/site-packages/xonsh/proc.py", line 1468, in tee_stdout
    sys.stdout.buffer.write(line)
TypeError: string argument expected, got 'bytes'
```

I changed line 1468 in `proc.py` to read

`sys.stdout.buffer.write(line.decode(encoding=enc, errors=err))`

which then does write the output to the history file but neglects to actually write it to stdout

astronouth7303:
I'm now getting incomplete stdout (at 0.0001 wait) at the rate of .002%.

scopatz:
Ok @gforsyth, it turns out that @laerus was right about the Tee classes in base shell.  I have fixed these up. 

@astronouth7303 @laerus I have also taken a more convergence approach to the last read.  Could you try it again please? Thanks!

astronouth7303:
With default settings, I'm now getting <0.002% failure (no failures on 50,000 iterations).

I'm still disturbed this comes up at all. Does every shell do this or is there a Python- or xonsh-specific reason this happens?

derekbrokeit:
@astronouth7303 The issue you describe of getting incomplete output is actually happening to me in `zsh`, but I am not experiencing this in `xonsh`. Specifically, in `zsh` I am loosing the last line of output after using the shell for too long. Do you import from zsh configurations at all? 

Also, with default settings, but running `del $FORMATTER_DICT['gitstatus']`, do you still see this problem?

astronouth7303:
@derekbrokeit Nope. I steer clear of foreign shell interfaces on the premise that it'll probably bite me and the less bash/zsh I have to maintain, the happier I am.

I can't test if `'gitstatus'` has any effect because the incidence has been reduced to levels I can't detect, but I'm not sure how you figure, since my testing doesn't invoke the prompt again until it's done.

adqm:
Sorry for being OOTL for such a long period of time.  As of Friday evening, I'm down from teaching 2.5 classes to now only 1.5 for the rest of the term, so I may finally have some free time.

This is really nice work!

But...there are a couple of cases that seem to break.

Most seriously, running `vim` at all screws everything up for the rest of the session (I have to close the terminal and open a new one to get back to a usable state).  `emacs -nw` works fine (some bias in the codebase? :smile:), as does `mutt`.

I had been working on a reimagining of `ProcProxy` myself before the semester kicked in (not this crazy of a rewrite, but was reworking to allow job control of aliases and the same kind of async comunication), and so I figured I'd try my test cases on this branch.  What I had used for testing are:

```
def dummy(args, inn, out, err):
    time.sleep(2)
    out.write('hey!')
    return 0

def dummy2(args, inn, out, err):
    time.sleep(2)
    out.write(inn.read())
    return 0

aliases['d'] = dummy
aliases['d2'] = dummy2
```

One thing to note is that the `@(function)` syntax seems to be broken:
```
hartz@lappy-587 ~ $ @(dummy)
[...snip...]
TypeError: object of type 'function' has no len()
```

The other is that, while piping _from_ one of these things seems to work, piping _to_ them is reliably broken for me, both from other procproxies and from real subprocs:
```
hartz@lappy-587 ~/xonsh $ d | cat
hey!hartz@lappy-587 ~/xonsh $ 
hartz@lappy-587 ~/xonsh $ d | d2
  File "/usr/local/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 10061, in run
    r = self.f(self.args, sp_stdin, sp_stdout, sp_stderr)
  File "/home/hartz/.xonshrc", line 115, in dummy2
    out.write(inn.read())
io.UnsupportedOperation: not readable
hartz@lappy-587 ~/xonsh $ echo "hi" | d2
  File "/usr/local/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 10061, in run
    r = self.f(self.args, sp_stdin, sp_stdout, sp_stderr)
  File "/home/hartz/.xonshrc", line 115, in dummy2
    out.write(inn.read())
io.UnsupportedOperation: not readable
```

There are a whole bunch of crazy edge cases involved here (as I've found out on my own branch), and it seems like the timeout is going to cause problems no matter what.  I was pretty close to having my `ProcProxy` rewrite with job control and async IO working, so I may revisit that (though on top of this branch) and see whether I can help to kill these edge cases.

scopatz:
> I'm still disturbed this comes up at all. Does every shell do this or is there a Python- or xonsh-specific reason this happens?

The answer to this is complicated, but I agree that it is disturbing, and I will try. Basically, there are three pieces to this which lead to a contradiction that is then solved with threads:

1. We ask xonsh to capture stdout and stderr, almost always.
2. Python subprocesses are relatively opaque and cannot be reliable suspended/resumed.
3. We ask xonsh to be interactive.

`$()`, `![]`, and `!()` all by definition capture and/or tee the output of command.  Only `$[]` streams directly and does not capture. Furthermore, while `$()` and `![]` can wait for the end of the command right after they have been started, `!()` has to delay until the iteration has started to begin popping bytes out of its output buffers.  `!()` if both non-blocking and capturing. I don't know of another shell that has this in the language.  (Yes, `screen` is both capturing and non-blocking, but it has restrictions on the kinds of (interactive) commands that you can use.)

Popen has two ways of enabling interactivity: (1) pass in the PIPE flag and use communicate(), or (2) pass in a file handle that happens to be a pipe and handle this yourself. Because of we require interactivity, xonsh cannot use the first option because communicate() only lets you send a predetermined input. Thus we must use our own pipes and read/write to them ourselves.  

This is where things get tricky.  OK, fine we are going to manage our own pipes.  But the question arises, *who is responsible for waiting on the end of the process?*  For processes that are in a pipeline but not the last one, the kernel and Popen will take care of everything for use, assuming everything is appropriately piped.  

For the last command in a captured pipeline, it probably makes sense that whatever is consuming the captured output (CommandPipeline here) should be driving the last command, which it turn drives the previous commands. 

Now, because `!()` is non-blocking / streaming, 1. the driver should be iterating through the lines (`iterraw()` here), and 2. We can't just wait for the last process to end. In fact, the user may never look at any output from `!()`. Thus, CommandPipeline can't forcibly close the captured output or kill the last process, unless the user explicitly asks.  It has to wait for the process to end itself. 

Suppose that the process ends but there is still data left in the queue waiting to be processed. Because we remain interactive, we have to have a layer where the data from the output pipe is read into a queue. **The contradiction is that the queue can be empty, the pipe can still have valid data, and the process can be complete.**

Thus, before/while you do the last few reads, you have to wait. But you can't wait indefinitely because that blocks, the process is already over (so you can't wait on that), and you can't wait on the pipe (because it is the thing you are reading from). 

The strategy here is the wait for 55 times the timeout for no new data to come into the queue. If any data does come in ever, the timer reset.  As heuristics go, I think it is a pretty effective one. It also has the benefit of being the only way forward I saw that didn't end in actually intractable contradictions. 

All that being said, there might be better ways that are platform specific.  POSIX (or at least Linux) allows you to suspend and resume a process.This could guarantee that the pipe isn't being modified while we are processing data from it. This would also allow us to delay starting the process from executing until we are ready to read data from it.  This, sadly, has no hope of working on WIndows where the best you can do is lower the process priority to a minimum.  You'd still leak some output into the pipe.  The code here works basically the same everywhere.

I hope this helps explain things.  I think this mess is partially a result of all of the things that we would like to do that are waaaay beyond the scope of a normal shell.

astronouth7303:
Wow, that is complicated. I may have to actually dig in to the code.

I had a whole thing written up on how the queue is the source of problems and we should get it out of the processing path, but then I remembered that Python doesn't have non-blocking synchronous IO, so we need a thread to deal with it. I blame monday morning.

Could you run End-of-Stream through the queue as well? Instead of relying on the fullness of the queue and having the queue reader check the process state, have the queue itself be aware of when it will get no more data. Then the reader knows when it can quit without having to interrogate farther up the data pipeline?

Going on step further, could each stream get its own process thread and we just have an abstraction object that takes an open file at one end and produces non-blocking IO at the other? It would manage the reading thread and queue, and then neither the process spawner nor the calling code would have to care about the mechanics of actually schlepping the data.

This is all contingent on Python actually telling us when the stream is closed...

PaulReiber:
> Suppose that the process ends but there is still data left in the queue
waiting to be processed

As the parent process, xonsh will receive SIGCHLD signals when children
processes exit
...for example,
http://www.microhowto.info/howto/reap_zombie_processes_using_a_sigchld_handler.html

That could trigger something to read on the pipe until it exhausts itself.


Kind regards,


scopatz:
@adqm 

> Sorry for being OOTL for such a long period of time. As of Friday evening, I'm down from teaching 2.5 classes to now only 1.5 for the rest of the term, so I may finally have some free time.

OMG! I can barely keep up with one class++. Welcome back to the world of sanity :)

> This is really nice work!

Thanks!  I think that I have managed to squash the two callable alias bugs that you found.  Those weren't so hard.  Just a couple of minor logic errors.  

Note that `d  | cat` is/was working correctly. This is the same behaviour as in Bash. Also, if you do `d | less`, you'll see less pop up and be empty for two seconds before it reads the "hey!".

The only one that has me a little concerned is vim.  vim seems to work and exit properly for me, leaving the shell in a fully usable state. Maybe vim is entering alternate mode and then not exiting it properly.  Still this shouldn't cause problems because this PR makes extra special care to clean up alternate mode. 

That said, I have put in a fix for vim which effectively just uses Popen (and not PopenThread) in all cases. This should be fine because there is basically no reason to capture output of vi and we know that ahead of time.

Your async rewrite sounds cool and I certainly welcome any improvements to job control!  I agree that that stuff should probably go on top of this one.  I think the overall architecture here is much much closer to what we ultimately want.  Job control can then take advantage of it.

scopatz:
@astronouth7303,

> Could you run End-of-Stream through the queue as well? Instead of relying on the fullness of the queue and having the queue reader check the process state, have the queue itself be aware of when it will get no more data.

Yes, in theory there are the `task_done()` and `join()` methods: https://docs.python.org/3/library/queue.html#queue.Queue.task_done. However, when I tried to use it by putting the task done call in the byte reading thread and the join in the PopenThread, it was blocking, IIRC.  I may have been using it wrong.

I think these kinds of changes or the abstraction object or moving to async io(?) are all great ideas. I definitely welcome them. However, this PR was (at least initially) meant as a minimum viable refactor. The diff is currently at 3K+ and has closed many open bugs and added a few features, so I think it has failed in that original goal :). That said, I think what we have here 

1. works, in an engineering sense,
2. doesn't have platform specialization
3. is a improvement over what is in master for the user experience.

I doubt that all of the possible bugs here are squashed and it certainly could be improved - especially in platform specific ways, or with respect to job control, or with respect to testing. But I think we should build up towards perfection, and this is just one "good enough" block.

@PaulReiber - SIGCHLD is posix, and maybe even Linux specific.  I am purposefully avoiding all platform specific code here in order to establish a baseline.  This may be a good candidate for the future, though, so thanks!

scopatz:
Also, obligatory inspirational poster until this is merged :)

scopatz:
Also, thanks again to everyone for all of your thoughtful feedback!

laerus:
(i thought all heuristics should have a prime number!)


pretty random:

```
$ cd src/xonsh
<xonsh-code>:1:3 - cd src/xonsh
<xonsh-code>:1:3 + ![cd src/xonsh]
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/base_shell.py", line 274, in default
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/codecache.py", line 65, in run_compiled_code
    func(code, glb, loc)
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/built_ins.py", line 821, in subproc_captured_hiddenobject
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/built_ins.py", line 789, in run_subproc
    command.end()
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/proc.py", line 1528, in end
    for _ in self.tee_stdout():
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/proc.py", line 1465, in tee_stdout
    for line in self.iterraw():
  File "/home/laerus/.local/lib/python3.4/site-packages/xonsh/proc.py", line 1403, in iterraw
    if ((stdout is None or not stdout.readable()) and
ValueError: I/O operation on closed file
$ # xonsh okay 
```

-


on a really slow machine:

small dir:
```
$ pwd
/home/laerus/src/xonsh/scripts
$ outs = []
$ for i in range(1000):
.     outs.append(len($(ls)))
.     
     
$ collections.Counter(outs)
Counter({22: 1, 23: 999}) 
$ # good enough
```

big dir:
```
$ pwd
/usr/lib
$ outs = []
$ for i in range(1000):
.     outs.append(len($(ls)))
.

$ counter = collections.Counter(outs)
$ counter.most_common(1)
[(13851, 267)]
$ # OUCH!
```

-

now on a fast machine with debian-testing:

on a small directory:
```
$ outs = []
$ for i in range(1000):
.     outs.append($(ls))
.

$ len(set(outs))
1
$ # COOL
```

on a huge dir
```
$ outs = []
$ for i in range(1000):
.     outs.append($(ls))
.
$ len(set(outs))
4
$ # :(
```


fun part, i tried to run these tests while doing a 1GB distro upgrade
and resulted to:

```
OSError: [Errno 24] Too many open files
```

btw even prompt_toolkit uses `os.pipe` somewhere

gforsyth:
I'm inclined to merge this as it stands, deal with the fallout (such as it may be) and then build upon it.  

@scopatz -- *cough* it does need a news entry

@xonsh/core-developers -- this is obviously a huge change, so if y'all could chime in with yays/nays/conditionals

scopatz:
+1 to what @gforsyth said!

@laerus - I have pushed up a hopeful fix for the readable issue.  The too many open files issue I would like to delay worrying about until after this PR.

xonsh/built_ins.py
```diff
@@ -317,322 +306,494 @@ def _redirect_io(streams, r, loc=None):
...
+        spec.resolve_binary_loc()
+        spec.resolve_auto_cd()
+        spec.resolve_executable_commands()
+        spec.resolve_alias_cls()
+        return spec
+
+    def redirect_leading(self):
+        """Manage leading redirects such as with '< input.txt COMMAND'. """
+        while len(self.cmd) >= 3 and self.cmd[0] == '<':
+            self.stdin = safe_open(seld.cmd[1], 'r')
```
gforsyth: this should be `self`, yes?

scopatz: Which part?  `safe_open()` is a function at module level scope. See https://github.com/xonsh/xonsh/pull/1758/files/965d2aa2569312cee9fcb3440450e35a81c56cc1#diff-32a5e51471ce3482f01745a4592320c1R267

gforsyth: I meant that the `seld.cmd[1]` should be `self.cmd[1]`

melund:
+1 for merging sooner than later...

adqm:
Sorry to (always) be the contrarian, but I am against merging while we have known timing issues as were reported above.  Sooner rather than later, sure, but not while we have known bugs, I would argue.

melund:
Hmm. We also have a bunch of other known bugs that we know this will close. So I am still for merging.

scopatz:
@adqm I don't think we have known timing issues.  The last thing that @astronouth7303 reported was a zero error rate on default.  Also, if @laerus has reported that if he bumps the frequency from the default to 1e-3 or 1e-2, then his issues go away too, unless I am mistaken about that. 

Meanwhile, there are issues on master that this branch fixes. 

I think I fixed the leaking open file handle issue of @laerus's too, but I haven't rigorously tested it.

astronouth7303:
I think the timing issue is resolved well enough until we can fix it properly (I think we'll need some platform-specific work, unfortunately).

The bigger concern to me is the reports that some major programs (like vim) are just up and failing.

Breaking <.5% of calls is a problem, but something fixable. Breaking 100% of a major app? That's a blocker and _must not_ make it to a release.

scopatz:
@astronouth7303 I have put in a fix for vim. Also, I think this is a good time to merge because it is right after a release.  So we can bang on this in master before the next release. I agree that we shouldn't break anything major.  But on some level it needs to have more exposure than a PR to find some of the harder to reach corner cases.

laerus:
vim is broken on this branch for me too.
```
$ old = tty.tcgetattr(sys.stdout)
$ vim
$ new = tty.tcgetattr(sys.stdout)
$ old == new
False
```

also i noticed that just running `top` on this branch, xonsh goes up to +35% cpu usage 
and i think i don't see that happening on current master. 
can someone confirm this please?

yes i think we should bring this in, even though it's a little unusable for me atm and i'll stay in .4.7 until this gets more stable, that's what the release was all about.

scopatz:
@laerus - can you try vim now? I was only fixing vi before... sorry!

laerus:
yep that's fixed, that list will grow.

backgrounding jobs is a little broken too.

scopatz:
> yep that's fixed, that list will grow.

Absolutely.

> backgrounding jobs is a little broken too.

It is a little broken on master too :)

scopatz:
Ok, I have added a new file and updated the docs a bit.

news/rs.rst
```diff
@@ -0,0 +1,42 @@
+**Added:**
+
+* New subprocess specification class ``SubprocSpec`` is sued for specifiying
```
astronouth7303: "sued"

news/rs.rst
```diff
@@ -0,0 +1,42 @@
...
+* The ``run_subproc()`` function has been replaced with a new implementation.
+* Piping between processes now uses OS pipes.
+* ``$XONSH_STORE_STDIN`` now uses ``os.pread()`` rather than ``tee`` and a new
+  file.
+
+**Deprecated:** None
+
+**Removed:**
+
+* ``CompletedCommand`` and ``HiddenCompletedCommand`` classes have been removed.
```
astronouth7303: So what does `!()` return now?

news/rs.rst
scopatz: dyslexicai not I am.

news/rs.rst
scopatz: Pipelines.  I updated the news to mention this.

laerus: just `!(ls)` sometimes leaks output after printing the pipeline object.

scopatz: It doesn't though, right, if you adjust your `$XONSH_PROC_FREQUENCY`... But I should add that to the news

scopatz: Done.

scopatz:
Are there any other objections? I really do think that I have addressed everything that everyone has brought up that isn't platform specific, a performance or testing issue, or itself a refactor of this.  If I am missing something, please let me know!

gforsyth:
I think this is ready to go -- in the interest of letting people get a word in if they need to, I'll hold off on merging this until noon EST

gforsyth:
Ok, here we go.

gforsyth:
Thanks, @scopatz !  

:tada::tada::tada:

scopatz:
Thanks @gforsyth! All!

Issue #1754, opened 2016-09-24
Can't point to a specific version for pip


rtpg:
Writing this here for posterity, Hopefully I'll be able to find the root cause this weekend.

Basically I can't run pip properly. I'm assuming this is a parsing issue related to mode detection, but the first part of the command is very "command-y" so I'm a bit surprised.

```
>>> pip install django==1.9.9
...
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 16237, in _parse_ctx_free
    debug_level=(self.debug_level > 1))
  File "/usr/local/lib/python3.5/site-packages/xonsh/parsers/base.py", line 314, in parse
    tree = self.parser.parse(input=s, lexer=self.lexer, debug=debug_level)
  File "/usr/local/lib/python3.5/site-packages/ply/yacc.py", line 331, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/usr/local/lib/python3.5/site-packages/ply/yacc.py", line 1181, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/usr/local/lib/python3.5/site-packages/ply/yacc.py", line 193, in call_errorfunc
    r = errorfunc(token)
  File "/usr/local/lib/python3.5/site-packages/xonsh/parsers/base.py", line 2355, in p_error
    self.xonsh_code)
  File "/usr/local/lib/python3.5/site-packages/xonsh/parsers/base.py", line 420, in _parse_error
    raise err
  File "<string>", line None
SyntaxError: <xonsh-code>:1:20: ('code: ==',)
![pip install django==1.9.9]
                    ^

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 99, in push
    locs=None)
  File "/usr/local/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 16184, in compile
    tree = self.parse(input, ctx, mode=mode, transform=transform)
  File "/usr/local/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 16155, in parse
    tree, input = self._parse_ctx_free(input, mode=mode)
[...]

PR #1759, opened 2016-09-24
Addressed double equals issue, and others

scopatz:
This is a fix for #1754


laerus:
this doesn't seem to fix the issue still getting an error in subproc mode:

```
$ pip3 install django==1.9.9
.......
SyntaxError: <xonsh-code>:1:21: ('code: ==',)
![pip3 install django==1.9.9]
                     ^
```

scopatz:
hmm, what if you do `echo django==1.9.9`?  The test I added passed....

laerus:
I tried a plain `echo ==` and I still get the issue, but now that I think of it I always had issues on develop mode with not catching up the changes on the parser, will try again. Maybe I must delete the parser table.

scopatz:
Oh! That might be a separate bug :)

laerus:
allright! the problem was local after all, this works great, thanks @scopatz !

PR #1760, opened 2016-09-24
Source alias newline fix

scopatz:
This should address #1517, for real.


gforsyth:
I could never reproduce this error but this seems straightforward enough.  I'm sure someone will let us know if it keeps being an issue.

Thanks @scopatz!

Issue #1048, opened 2016-05-31
Expose vi input-mode to prompt


mattbowen:
In ZSH, I use vim bindings for my prompt. It's really useful to display which mode I'm in in my prompt. It doesn't look like xonsh exposes my mode when $VI_MODE = True.

In ZSH, I have this setup:

```
bindkey -v

function zle-line-init zle-keymap-select {
    VIM_PROMPT="%{$fg_bold[yellow]%} [% NORMAL]% %{$reset_color%}"
    PROMPT_TIME="[%D{%K:%M:%S}]"
    RPS1="${${KEYMAP/vicmd/$VIM_PROMPT}/(main|viins)/}$PROMPT_TIME"
    zle reset-prompt
}

zle -N zle-line-init
zle -N zle-keymap-select
```

It'd be cool to be be able to do something similar, although the way of doing it could definitely be more beautiful in how you spell it.

scopatz:
Hey @mattbowen! Thanks for asking about this.   I believe that you can do something like this already.  If you take a look at the customizing the prompt part of the tutorial (http://xon.sh/tutorial.html#customizing-the-prompt) you'll see that all you need to do is add a function to the `$FORMATTER_DICT` that implements the desired behaviour. Then you are free to use the key that you use in your `$PROMPT` variable.  All of this is something that you can do in your xonshrc file.

Additionally, something like this sounds like it would be useful more broadly.  A pull request that implements this function (defined in `xonsh.environ`) would be much appreciated!

t184256:
@scopatz, your creation is pure wonder and your wisdom is deep.

However that's probably not all that is required here. How often is the prompt updated? It seems that it is only reevaluated between inputs.

That's what I've tried in my `.xonshrc`:

```xsh
import builtins
def vi_mode_prompt():
    try:
        m = builtins.__xonsh_shell__.prompter.cli.vi_state.input_mode
    except AttributeError:
        return '[???]'
    return '[INSERT]' if m == 'vi-insert' else '[NORMAL]'
$FORMATTER_DICT['vi_mode'] = vi_mode_prompt
$RIGHT_PROMPT = '{vi_mode}'
```

Simple and expressive, but not working. On first start it's `[???]`. Then it changes to whatever mode prompt-toolkit's prompt was in where you pressed Enter. It's not what we want; I guess we want prompt reevaluation on mode change. Sounds pretty niche and radical.

Any advices from here?

scopatz:
[...]

PR #1761, opened 2016-09-25
Add $XONSH_UPDATE_PROMPT_ON_KEYPRESS option

t184256:
$XONSH_UPDATE_PROMPT_ON_KEYPRESS = True (False by default)
disables prompt caching, so that it will be reevaluated on each keypress.
It is crucial, e.g. for exposing vi-input-mode to the prompt (see #1048).
Performance with slow prompts will probably suffer.


scopatz:
Thanks @t184256, it looks like this is what is in #1762.  My comments on the name apply here too,  Thanks again!

t184256:
Thank you for xonsh!

I can't get my head around how dependent pull requests are filed on github. I'll change the name and rebase everything for now.

t184256:
Done.

By the way, should I now write a readline backend change to attain feature parity? Can readline reevaluate the prompt on keypress? It doesn't look like xonsh/readline_shell.py has some prompt caching to disable =)

scopatz:
Thanks @t184256!  And, readline does not have a similar feature to the best of my knowledge, so it is probably not worth trying to replicate.

Issue #1564, opened 2016-08-15
Catch  and log exception in fromatter_dict, and prompt formatting.


Carreau:
In particular the example in tutorial.rst fais as `$FORMATTER_DICT['test'] = lambda: random.randint(1,9)` does not return a string, and other failures might be hard to debug.

adqm:
Seems to work fine in the `readline` shell, so adding the `prompt-toolkit` label.

scopatz:
This seems to fail for me on both shells:

```console
scopatz@artemis ~ $ import random
scopatz@artemis ~ $ $FORMATTER_DICT['test'] = lambda: random.randint(1,9)
scopatz@artemis ~ $ $PROMPT = '{test} ' + $PROMPT
{test} {env_name:{} }{user}@{hostname} {cwd}{branch_color}{curr_branch: {}} {prompt_end} 
```

scopatz:
Thanks for reporting @Carreau!  I distinctly recall having string conversion originally, so this must have been lost at some point.

adqm:
oh, shoot; i was testing the wrong thing.  my bad.  agree this affects both shells.

laerus:
Testing with ptk atm, it seems that we're calling `prompt.base.partial_format_prompt` 3 times each time we print the prompt. Once in `ptk.shell.PromptToolkitShell.prompt_tokens`, once in `ptk.shell.PromptToolkitShell.rpromt_tokens` and once in `base_shell.BaseShell.settitle`, that seems a little excessive, will try and refactor this out, putting this here as a note.

scopatz:
@laerus - I agree that we could certainly try to do better caching, but somethings like CWD and the last return code actually can and do change each time the prompt is printed.  I think we'd have to clear the cache pretty frequently, but we might be able to cut down on the prompt and rprompt overhead a bit.

scopatz:
Of course this would only help if prompt and rprompt display the same information.  However, this may be more useful if / when we also add a bottom toolbar.  Also, maybe different entries should have a way of flagging how often they need to be revaluated. For example, once (for things like user and host name), every prompt (for repo status, cwd, etc), and every keystroke (for run time validation of input).  This might be a bit of an overhaul.  But it would give us the chance to change the name of `$FORMATTER_DICT` :)

laerus:
Was thinking that `$FORMATTER_DICT` is not a nice name too. Think we should bring events on this?

scopatz:
[...]

PR #1765, opened 2016-09-26
stringify values that come from FORMATTER_DICT envvar

laerus:
#1564

### DIFF
--- a/news/prompt-formatter.rst
+++ b/news/prompt-formatter.rst
@@ -11,6 +11,7 @@
 
 **Fixed:**
 
+* non string type value in $FORMATTER_DICT turning prompt ugly
 * whole prompt turning useless when one formatting function raises an exception
 
 **Security:** None
--- a/xonsh/prompt/base.py
+++ b/xonsh/prompt/base.py
@@ -120,7 +120,7 @@ def _partial_format_prompt_main(template=DEFAULT_PROMPT, formatter_dict=None):
         elif field in fmtter:
             v = fmtter[field]
             try:
-                val = v() if callable(v) else v
+                val = str(v() if callable(v) else v)
             except Exception as err:
                 print('prompt: error: on field {!r}'
                       ''.format(field), file=sys.stderr)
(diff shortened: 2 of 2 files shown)


scopatz:
Looks like we have cross-the board-failures @laerus. What was the type that wasn't a string?

scopatz:
For instance, I think if val is None then it should be skipped, or replaced with an empty string.

laerus:
i'm looking for a temporal fix here so i can then try for a refactor, i think we need a proper `XonshPrompt` object in place of `$FORMATTER_DICT` that inherits from `MuttableMapping` and maybe `string.Fromatter` and later maybe move most of the work being done in the shells on the `XonshPrompt`, what do you think? also what would be a good name in place of `FORMATTER_DICT`, something like `$XONSH_PROMPT` ?

laerus:
btw can i move `prompt.base` to `prompt.__init__` so we can just `import xonsh.prompt` or does this interfere with amalgamation ?

derekbrokeit:
I'm new to xonsh, but have been very interested in the positive direction I see this shell moving. I was about to submit a bug report regarding a similar issue, but I found this issue and wanted to add my two cents.

Instead of stringify results of `$FORMATTER_DICT` it should attempt to use arbitrary objects. Here's an example of a useful application to have a universal date formatter:

```
import datetime
$FORMATTER_DICT['date'] = datetime.datetime.now
$PROMPT = "{date:%H:%M} >>> "
$RIGHT_PROMPT = "{date:%Y-%m-%d}"
```

A nice advantage of this would be improvements to the customizability of builtin formatters such as `gitstatus`. Instead of `"{gitstatus}"` the gitstatus formatter could return a `GitStatus` customized object with a `__format__` method that allows for customization: `{gitstatus:%(branch)|%(status)}` or something similar. Another possible addition would be that the object is turned into an empty string if the returned object is `None`. 

I think this would be a better long term step than just standard string conversion, which 529d933 seems to produce.

scopatz:
Thanks @laerus!

Also @derekbrokeit, thanks for using xonsh!  Personally, I am all in favor of having improved / better prompt formatting mechanisms. Your suggestions do sound useful.  Pull requests are very welcome in this department, and more broadly.

PR #1767, opened 2016-09-26
Use SetConsoleTitleW() on Windows

### LINKED ISSUE #1753: BaseShell.settitle() uses os.system()
And [arguments are inserted without escaping](https://github.com/xonsh/xonsh/blob/2e5234da02b5b6590c21dd932bc50b8f614ade9f/xonsh/base_shell.py#L292).

Additionally, this only works under CMD, not PowerShell.

My recommendation is to call [SetConsoleTitleW()](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686050(v=vs.85).aspx) via [ctypes](https://docs.python.org/3/library/ctypes.html).

astronouth7303:
Closes #1753.

This should improve performance (yay one less process to wait on) and potentially improve security (There aren't known issues with the Windows escaping, but shell escaping isn't always clean...)

`kernel32` could be made to be lazy, but it'll be used as part of the first prompt anyway, and since it's already loaded into the process, I'm not sure significant time would be saved.

### DIFF
--- a/news/win-setconsoletitle.rst
+++ b/news/win-setconsoletitle.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+* Use ``SetConsoleTitleW()`` on Windows instead of a process call.
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


melund:
Good work. I think the main performance issue that we care about is startup time. So it better be lazy.

astronouth7303:
1. The expensive operation (loading the DLL in to memory) is already loaded, because `kernel32` also includes functions dealing with threads, file IO, registry access, sockets, and [a few other things](http://www.geoffchappell.com/studies/windows/win32/kernel32/api/).
2. If a lazy object is guaranteed to be resolved before the prompt is displayed, is any performance gained?

melund:
Yes. That is true. LGTM

PR #1770, opened 2016-09-26
which

laerus:
moved `aliases.which` into `xoreutils.which`
small refactor
`which -a` now searches in `__xonsh_ctx__` too

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -22,9 +22,9 @@ from xonsh.timings import timeit_alias
 from xonsh.tools import (XonshError, argvquote, escape_windows_cmd_string,
                          to_bool)
 from xonsh.xontribs import xontribs_main
-from xonsh.xoreutils.which import which_main
 
 import xonsh.completers._aliases as xca
+import xonsh.xoreutils.which as xxw
 
 
 class Aliases(cabc.MutableMapping):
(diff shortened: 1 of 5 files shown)


xonsh/xoreutils/which.py
```diff
@@ -0,0 +1,162 @@
+import argparse
+import builtins
+import functools
+import os
+
+from xonsh.xoreutils import _which
+from xonsh.platform import ON_WINDOWS, scandir
```
scopatz: I think this should be `import xonsh.platform as xp` to be consistent with other subpackages.

xonsh/aliases.py
```diff
@@ -22,7 +22,7 @@
 from xonsh.tools import (XonshError, argvquote, escape_windows_cmd_string,
                          to_bool)
 from xonsh.xontribs import xontribs_main
-from xonsh.xoreutils import _which
+from xonsh.xoreutils.which import which_main
```
scopatz: This should be `import xonsh.xoreutils.which` so that this can be a lazy import.  You should only use `from... import...` if the code is in  the same package.

scopatz:
Thanks for putting this in @laerus! Just a couple of small import changes.

laerus:
ping for this!

xonsh/xoreutils/which.py
```diff
@@ -144,6 +148,7 @@ def which(args, stdin=None, stdout=None, stderr=None):
 
 
 class AWitchAWitch(argparse.Action):
+    """The Ducstring, the mother of all ducs."""
```
gforsyth: :heart:

gforsyth:
Looks good, @laerus !

Issue #1776, opened 2016-09-29
git_dirty_working_directory always returns True if git not in english


xsteadfastx:
im using the latest version from master (commit 1c96dc4ab23ad67). i found out about the changes for `git_dirty_working_directory`. it checks for a dirty git directory in a different way. the old way was with `git status --porcelain` (https://github.com/xonsh/xonsh/blob/e10c38deccf94189012cca9864cac2bff77dd984/xonsh/environ.py#L1036) and right now it checks for the bytestring `nothing to commit` (https://github.com/xonsh/xonsh/blob/master/xonsh/prompt/vc_branch.py#L139). my git is in german so it always returns True. has it a deeper reason that it uses a new way of finding out about a dirty git directory? i thought i ask first before i send a pull request for using the old way again.

laerus:
Sorry for this @xsteadfastx, this change was part of a fix on the `vc_branch.py`, i should have kept the function in it's older version.

PR #1777, opened 2016-09-29
using --porcelain to find out if git directory dirty

xsteadfastx:
a pull request for #1776

### DIFF
--- a/news/vc_branch.rst
+++ b/news/vc_branch.rst
@@ -4,13 +4,17 @@
 
 * ``xonsh.prompt.gitstatus.gitstatus`` now returns a namedtuple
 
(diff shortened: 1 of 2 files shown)


laerus:
thanks for this @xsteadfastx! Can you add an entry in the `news/vc_branch.rst` in the `Fixed` section?

xsteadfastx:
@laerus done. im thinking about to add tests as well. i never contributed anything to xonsh before... so i will see how hard it is to write tests for xonsh.

xsteadfastx:
and another question is... if we should readd the `include_untracked` argument.

laerus:
Thank you! Another PR with tests would be awesome, i guess we should mock out the subprocess call to git to test this.

Issue #1778, opened 2016-09-29
Wrong git branch name in edge case.


Carreau:
```
path/to/repo master $ git checkout origin/pr/132
path/to/repo origin/pr/132 $ # do changes
path/to/repo origin/pr/132 $ git commit -am 'stuff'
...
path/to/repo from $ #WTF ? Why "from" branch ?
path/to/repo from $ git branch
* (HEAD detached from origin/pr/132)
  106
  community-page
  fix-css
  foo
  master
```

Oh, make sens.

laerus:
Sorry about this @Carreau, on it!

Carreau:
> Sorry about this @Carreau, on it!

No need to apologize. I can likely have a shot at it but I wanted to open the issue before forgetting.

laerus:
i introduced this bug recently, potential fix in #1779

PR #1779, opened 2016-09-29
prompt git branch

laerus:
#1778

### DIFF
--- a/xonsh/prompt/vc_branch.py
+++ b/xonsh/prompt/vc_branch.py
@@ -16,16 +16,13 @@ import threading, queue
 
 def _get_git_branch(q):
     try:
-        status = subprocess.check_output(['git', 'status'],
+        status = subprocess.check_output(['git', 'status', '--porcelain', '--branch'],
                                          stderr=subprocess.DEVNULL)
     except (subprocess.CalledProcessError, OSError):
         q.put(None)
     else:
-        status = status.decode().split()
-        if status[2] == 'at':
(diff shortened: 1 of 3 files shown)


laerus:
nope, porcelain is strange.

xonsh/prompt/vc_branch.py
```diff
@@ -16,16 +16,13 @@
...
                                          stderr=subprocess.DEVNULL)
     except (subprocess.CalledProcessError, OSError):
         q.put(None)
     else:
-        status = status.decode().split()
-        if status[2] == 'at':
-            q.put(status[3])
-        else:
-            q.put(status[2])
+        status = status.decode().splitlines()[0].split()[1]
```
scopatz: Calls to decode should use the `XONSH_ENCODING` and `XONSH_ENCODING_ERRORS` env vars

scopatz:
Thanks @laerus!

xonsh/prompt/vc_branch.py
laerus: shouldn't `tools.decode` respect these envvars too?

laerus: ok it seems that `tools.decode_bytes` does this, but the docs are little misleading.

scopatz: Ahh yeah! Thanks for fixing this up.

scopatz:
Thanks @laerus!

PR #1788, opened 2016-10-01
Expand user for non-default data directories

### LINKED ISSUE #1782: Creation of anomalous folders when setting `$XDG_DATA_DIR` and `$XONSH_DATA_DIR` in config.json
In the `config.json` file, I had the following items:

``` json
"XDG_DATA_HOME": "~/.local/share",
"XONSH_DATA_DIR": "~/.local/share/xonsh",
```

I found that whenever a new terminal was created, it would create a folder `~` in whatever the initial open folder is. So, there will end up being `~` folders everywhere if I open up `tmux` or start new terminals. This doesn't really affect me because I do not need to set these items, which are just defaults, but it may be good to have proper way to reference `$HOME` from the config files.

gforsyth:
Attempt to fix #1782.  
We were using `expanduser` on the default values but if non-default values are specified in `config.json` then `expanduser` isn't run on them and a bunch of spurious `~` directories might be created.

### DIFF
--- a/news/expand_user.rst
+++ b/news/expand_user.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Bug where non-default locations for ``XDG_DATA_HOME`` and ``XONSH_DATA_DIR``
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks @gforsyth!

Issue #374, opened 2015-08-27
iTerm2 shell integration with xonsh


asmeurer:
In case you haven't seen it, iTerm2 has some nice [shell integration](https://iterm2.com/shell_integration.html) that enables some nice features.  These work by echoing some [special escape sequences](https://groups.google.com/forum/#!msg/iterm2-discuss/URKCBtS0228/rs5Ive4PCAAJ) in the shell. I have written a Python library called [iterm2-tools](https://github.com/asmeurer/iterm2-tools) that has an [API](https://iterm2-tools.readthedocs.org/en/latest/shell_integration.html) for these sequences, and implemented integration with [IPython](https://iterm2-tools.readthedocs.org/en/latest/ipython.html). 

I would like to implement integration with xonsh. This is effectively the only thing keeping me from using Xonsh in day-to-day work. 

The integration by printing special sequences before and after the prompt, and before and after the output of a command. The prompt integration is easy: you generally just need to wrap your `$PROMPT` with them. The tricky part is the sequence that has to be printed before the command output. 

[Here](https://github.com/gnachman/iterm2-website/tree/master/source/misc) are the integrations with other shells. Bash implements this using the `DEBUG` trap (bash also uses `PROMPT_COMMAND` instead of editing `PS1` directly). zsh has `preexec` functions. tcsh has `postcmd` (I guess). And fish I'm not sure about (in each case, look how the `;C` sequence is being echoed). 

I looked a little bit in xonsh and didn't see any obvious ways to do this without editing xonsh itself. I figured it would be easier to ask here than dig too hard, though. 

Secondly, would you like for some or any of this to be part of xonsh? I'd like to use `iterm2-tools` as a dependency. We could either make it so that xonsh does this automatically if iterm2-tools is installed (technically it could work in terminals other than iTerm2, such as the now defunct FinalTerm, where the integration codes originated). Otherwise, I could add it to iterm2-tools and people would need to add a couple of lines to their .xonshrc to enable it. I don't care very strongly either way about this. I mostly just want to get it working. 

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

scopatz:
Hi @asmeurer - This would be awesome to have! And you are right that there would not be an easy way to do this without editing xonsh itself.  We could add `preprecmd()` and `postpostcmd()`, but I am not sure that is a great option either.  

I kinda would like it if it just worked automatically if `iterm2-tools` was installed.  

[...]

PR #1790, opened 2016-10-02
xontrib-mpl makes a copy of the current figure before modifying it

derekbrokeit:
This is a small change to mplhooks that ensures that the figure that is used for converting to a string is a copy of the original. This makes sure that the modifications made to the pixel size, ticks, etc are not put into the original figure. That way, `mpl` can be used to quickly iterate and modify the current figure before saving or plotting the figure in a different manor that can show text or use the `matplotlibrc` for beautification.

In order to do this, I am importing pickle, I hope that is not a problem. This is my first pull request to xonsh, and I wanted to make it a relatively simple one.


astronouth7303:
Does [`copy`](https://docs.python.org/3/library/copy.html) not work for this?

derekbrokeit:
That's a good question. This was an old technique that I had to use in a much older version of `matplotlib` way back, so I was used to it. I just checked and it still seems that the `copy` module only partially copies the `Figure` object. When I used the `copy` module, the margins of the figure were removed by `mpl`.

astronouth7303:
Even with `deepcopy()`? What are they doing?

Could you add a comment explaining to future generations that `copy` doesn't work? Possibly some test cases if you can.

I don't use matplotlib, so I don't know if this is a _good_ idea. That's my only comment on the quality of the code itself.

scopatz:
Hi All, I think this is highly usual - as a matplotlib user - to copy a figure in this way. Maybe @tacaswell can either approve or offer an alternative? Thanks in advance!

derekbrokeit:
I did some digging, and found that there is an open ticket for this functionality https://github.com/matplotlib/matplotlib/issues/3056 . The exception that is thrown is `NotImplementedError`. 

I will fix up this commit soon to include tests, documentation, etc to help people understand the reason for this less-readable method. In the meantime, I will also wait for a reply if there is a safer way to make a copy of a figure in the above ticket.

derekbrokeit:
I made a few more changes, do these need to be rebased and a new pull request created?

melund:
@derekbrokeit No need to do anything special. It is enough to push the changes and the PR will update

melund:
It seems like the test setup is missing the numpy package.

melund:
Well, also `matplotlib` itself I guess.

derekbrokeit:
There were no tests previously for this xontrib, so there were no previous requirements that `matplotlib`/`nump` were present I suppose. Being as `mpl` ships with xonsh, does that make them requirements of xonsh?

melund:
You can add it here https://github.com/xonsh/xonsh/blob/master/requirements-tests.txt . The only question is how much slower it makes the CI tests. (The extra time it takes for mpl and numpy to download and install) 

If really changes the test time a lot these tests may not be worth it - Never thought I should have that about tests..

melund:
xonsh have many optional features that only work when certain packages are present. The mpl xontrib is one of them. Numpy and Mpl are not dependencies of xonsh

derekbrokeit:
That makes sense, I wanted to make sure.

derekbrokeit:
As a test of the ... tests, I will also send a commit change of testing requirements.

gforsyth:
I suggest that we only test numpy+matplotlib with the mpl xontrib on CircleCI, since it's fast and uses conda and we won't take a hit building them.

melund:
@gforsyth . Good idea. @derekbrokeit  You can import `numpy`and `matplotlib` with pytest `importorskip` machinery in the test module : 
```
mpl= pytest.importorskip("matplotlib")
np = pytest.importorskip("numpy")
```

That way the test should skip if modules are not available. Then we only need to add `matplotlib` and `numpy` to Circle CI setup.

melund:
@derekbrokeit. I think you can add mpl and numpy to the Circle CI tests in this file: https://github.com/xonsh/xonsh/blob/master/.circle_miniconda.sh

tacaswell:
Obligatorily applying `tight_layout` is not a great idea, not everything that can end up in the figure is respected by it (which is the reason it is not enabled by default upstream).  The rcparam `figure.autolayout` can be used to control the default `tight_layout` setting of new figures.

For everything else, I suggest using a context manager to ensure that the get rolled back to what they were before.

This also seems to assume that there is only ever one `Axes` in the figure.

Most of this function could be replaced by

```python
from io import BytesIO
bb = BytesIO()
fig.savefig(bb, dpi=fig.dpi, format='raw')
bb.seek(0)
buf = np.frombuffer(bb.read(), dtype='uint8').reshape(*fig.canvas.get_width_height(), *(4,))
```

derekbrokeit:
@tacaswell saving the figure as a byte stream is not a bad idea either. My intention was not to refactor the entire xontrib, which could be done for greater clarity. Instead, my main hope was simply to not have `mpl` mangle figures.

derekbrokeit:
What `mpl` should actually do instead of tight layout is remove the margins entirely since it refuses to write text anyway. Also, `mpl` works fine with multiple axes.

tests/test_mpl.py
```diff
@@ -1,7 +1,11 @@
+import pytest
+
+# make sure to skip these tests entirely if numpy/matplotlib is not present
```
astronouth7303: "are not present"

tacaswell:
It is probably better to just set the visibility of the text to `False`

You can do

```python
fig.savefig(bb, format='png', bbox_inches='tight')
```
and then just decode the png which will, I think, get what you want with much less work.

tacaswell:
And you _have_ completely refactored this in a way that is actively fighting the underlying library.

derekbrokeit:
@tacaswell As far as I can see it, we must adjust the parameters of the figure before making the buffer, so your suggestion of a bytes stream only affects section labeled ` Draw the renderer and get the RGB buffer from the figure`. Unfortunately, the figure must be copied before the adjustments because otherwise the user cannot iterate on the figure.

And, ya, at this point the function is undergoing refactoring compared to that first commit.

derekbrokeit:
I just wanted to give an update. After having thought about this for a bit, I found there's a relatively harmless way of making these changes and reversing them such that the `mpl` figure prints out maximizing space without having to copy the figure. I'll prepare a couple patches that will make this possible. I also wonder if the user may want to specify (through environment variable) to not manipulate the figure for the reason that they just generally want to see the layout. Also, the test figure will be more robust in the next pushes. I'll get these changes ready later when I have some time. Thanks @tacaswell for pushing back on this, I think it will be better than the first push.

Also, in the meantime I also found that iTerm2 allows inline images, which would end up being a superior solution, but will only work for a subset of users. I don't know if xonsh is interested in trying to support such terminal-specific features in the future, but I wanted to make note of it in case anyone is interested.

tacaswell:
Great!

scopatz:
Thanks @derekbrokeit! Defintiely looking forward to it. 

And yes, we are interested in terminal emulator specific code, if you have it.  I think that iTerm2 integration like this would be great to have and better than the hack here :)

tacaswell:
There is an iterm2 mpl backend floating around someplace.

derekbrokeit:
I have made most of the changes discussed above except that the test improvement hasn't been done yet. I figured, most of the other changes could get a look at. I am using `iterm2_tools` package, which isn't exactly necessary because the required tools are relatively simple. However, it was mentioned in #374 as a possible way to integrate xonsh and iterm2 seamlessly. Also, I don't yet have a good way to detect the iTerm2 terminal application, so I thought this could be a way to do it.

scopatz:
Hi @derekbrokeit, this looks good to me. Is it ready to go in?

derekbrokeit:
Hi @scopatz , unless there was an objection to using `iterm2_tools` in this way, I was not planning any additional changes.

scopatz:
No objections from me!

scopatz:
Thanks again!

PR #1792, opened 2016-10-03
Allow activation of virtual environments in relative directories.

astronouth7303:
So that venvs made by other means (eg, `.venv`) are usable within xonsh.

(Also disallow the creation of venvs with reserved names.)

I'm pretty sure this is a good idea, but I'm not certain, so I wanted a second opinion.


scopatz:
Thanks @astronouth7303!

PR #1795, opened 2016-10-04
Fix pygment lexer when used by third party programs

melund:
This fixes crash when using the pygment plugin from third-party programs. 

It happens because `__xonsh__commands__cache__` and `__xonsh_env__` are not available unless it is used from xonsh. 

In Spyder I got the follow error when opening a '.xsh' file: 
```
...
 File "C:\Users\mel\Anaconda3\lib\site-packages\spyder\utils\syntaxhighlighters.py", line 989, in guess_pygments_highlighter
    lexer = get_lexer_for_filename(filename)
  File "C:\Users\mel\Anaconda3\lib\site-packages\pygments\lexers\__init__.py", line 150, in get_lexer_for_filename
    return res(**options)
  File "C:\Users\mel\Anaconda3\lib\site-packages\pygments\lexer.py", line 582, in __call__
    return type.__call__(cls, *args, **kwds)
  File "c:\users\mel\documents\xonsh\xonsh\pyghooks.py", line 74, in __init__
    _ = builtins.__xonsh_commands_cache__.all_commands
AttributeError: module 'builtins' has no attribute '__xonsh_commands_cache__'
``` 

Although this PR fixes the immediate problem, the syntax highlighting looks a bit wonky in Spyder. 

image

Probably a separate issue.

### DIFF
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -50,10 +50,11 @@ class CommandsCache(cabc.Mapping):
          name on Windows as a list, conserving the ordering in `PATHEXT`.
          Returns a list as `name` being the only item in it on other platforms."""
         if ON_WINDOWS:
+            pathext = builtins.__xonsh_env__.get(['PATHEXT'],['.EXE','.BAT','.CMD'])
             name = name.upper()
             return [
                 name + ext
-                for ext in ([''] + builtins.__xonsh_env__['PATHEXT'])
+                for ext in ([''] + pathext)
             ]
         else:
             return [name]
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -50,7 +50,7 @@ class CommandsCache(cabc.Mapping):
          name on Windows as a list, conserving the ordering in `PATHEXT`.
          Returns a list as `name` being the only item in it on other platforms."""
         if ON_WINDOWS:
-            pathext = builtins.__xonsh_env__.get('PATHEXT', ['.EXE','.BAT','.CMD'])
+            pathext = builtins.__xonsh_env__.get('PATHEXT')
             name = name.upper()
             return [
                 name + ext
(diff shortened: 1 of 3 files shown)


xonsh/commands_cache.py
```diff
@@ -50,10 +50,11 @@ def get_possible_names(name):
          name on Windows as a list, conserving the ordering in `PATHEXT`.
          Returns a list as `name` being the only item in it on other platforms."""
         if ON_WINDOWS:
+            pathext = builtins.__xonsh_env__.get('PATHEXT', ['.EXE','.BAT','.CMD'])
```
scopatz: You don't need to supply the default here do you?

xonsh/pyghooks.py
```diff
@@ -21,11 +21,17 @@
...
 from xonsh.tools import (ON_WINDOWS, intensify_colors_for_cmd_exe,
                          expand_gray_colors_for_cmd_exe)
-from xonsh.tokenize import SearchPath
-
+                         
+# If the lexor is loaded as a pygment plugin, we have to mock  
+# __xonsh_env__ and __xonsh_commands_cache__ 
+if not hasattr(builtins, '__xonsh_commands_cache__'):
+    if not hasattr(builtins, '__xonsh_env__'):
+        setattr(builtins, '__xonsh_env__', {})
```
scopatz: Shouldn't the default here be os.environ?

scopatz:
Thanks @melund, other than these coupld of changes, I think this looks good.  Though I would prefer if the commands cache and env logic were moved to the lexer class somehow.  However, since pyghooks is not amalgamated, this is not strictly necessary.

xonsh/commands_cache.py
melund: Well it is needed if  `__xonsh_env__` is mocked as a simple empty dict. But I guess it is better to just add needed data where we mock `__xonsh_env__`

melund:
No. Problem. I will try to move them into the class.

xonsh/pyghooks.py
melund: No. That would be a detyped environment. What the `command_cache` need is access to a few entries in the xonsh form.

scopatz: Ahh good point!

scopatz:
Thanks @melund!

PR #1796, opened 2016-10-04
fix pip completion

gforsyth:
pip completion results for install/uninstall were returning the entire
list of installed packages in a given python install no matter the
contents of the completion prefix.

Now the results are filtered by a `startswith(prefix)`


scopatz:
looks great to me!  waiting for ci

scopatz:
Thanks @gforsyth!

PR #1800, opened 2016-10-05
[Experiment] Python 3.3 Support

astronouth7303:
So that PyPy is usable, backport to 3.3

### DIFF
--- a/.circle.yml
+++ b/.circle.yml
@@ -2,7 +2,7 @@ machine:
     environment:
         PATH: /home/ubuntu/miniconda/envs/test_env/bin:/home/ubuntu/miniconda/bin:$PATH
     post:
-        - pyenv global 3.4.4 3.5.1
+        - pyenv global 3.3.5 3.4.4 3.5.1
 
 dependencies:
     pre:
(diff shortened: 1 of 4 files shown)


astronouth7303:
So pathlib (introduced in 3.4) is currently required to even run setup.py. I'm going to leave this here for now in case anyone else feels like working on it, but it's mostly just an idle project.

Or we could just wait for [PyPy3.5](https://morepypy.blogspot.com/2016/08/pypy-gets-funding-from-mozilla-for.html).

santagada:
I would wait for pypy 3.5. Also the startup time of pypy was bigger than cpython and I don't see much place for jit optimizations on xonsh so don't know if there is a strong case for it yet.

(maybe we should support micropython to run xonsh on embeded devices)

astronouth7303:
The big use case was to make use of pypy sandboxing as the container for live xonsh.

santagada:
Can't we just compile it with emscripten and run it on javascript? That seems even safer :)

scopatz:
Well, emscripten won't connect it with a real TTY, right?  That is, you wouldn't be able to run any commands...

santagada:
Sorry maybe I don't know what is live xonsh, I think it was supposed to be a xonsh running on the website so people could play with it. If that is the case we can even run a whole linux emulator with files and all (that way people can play without fear of destroying anything). But if it is something else sorry.

santagada:
Can we close this branch?

astronouth7303:
Yeah.

To future generations: Feel free to reopen if you decide to take a go at it.

PR #1802, opened 2016-10-06
Performance Improvements

scopatz:
This prevents xonsh from idly spinning it wheels in many cases where it doesn't need to, while still remaining responsive.  For me CPU usage for an open `less` with no input from the user went from 13.5%-45% down to 0.0%-0.5%.  This is more in the acceptable range of most applications. 

@laerus, you mentioned you were also seeing issues.  Would you mind seeing if this helps?

### DIFF
--- a/news/perf.rst
+++ b/news/perf.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* CPU usage of ``PopenThread`` and ``CommandPipeline`` has been brought
+  down significantly.
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


laerus:
on new xonsh instance just running `top`:

on current master: 40% cpu usage
on this branch: 25%
on 0.4.7: 1%

scopatz:
Well, it is still better :)

laerus:
i'm merging, thanks @scopatz! just a quick question, can't we give direct control of stdout/stdin for a program that runs in the foreground?

astronouth7303:
@laerus unless we're logging stdin/stdout.

scopatz:
Yeah, like Jamie said, there are two problems here. The first is that you can't swapout the I/O handles for a subprocess once it has started. The second is that we don't necessarily know ahead of time if a process is going to enter alternate mode or not before it is executed. 

Right now, we assume that the user wants to capture I/O unless (1) they use the $[cmd] operartor,or (2) the CommandsCache predicts that the command is not "backgroundable" meaning that it cannot be run on a background thread and should just use Popen, directly.  The user can modify these as they see fit, but we should probably provide more defaults than we have now.

PR #1803, opened 2016-10-06
More which and improvements

scopatz:
This expands the callable alias API a bit and makes it so that ``$(which cmd)`` does the right thing.


scopatz:
Would anyone mind taking a look at this, please? Thanks!

melund:
I did. Didn't I? LGTM, except for my single comments which may of course not be valid.

scopatz:
Hi @melund - I am not seeing any of your comments on this PR, except for the one I am responding to :)  Maybe they are held up in "review changes"

melund:
Oh.. you are right. I just haven't understood Github's new review system.

melund:
Well. It does improve which and it also fixes the root problem in #1826. The general issue with aliases is still there ( #1828 ) and I don't think I know how to fix it.

PR #1810, opened 2016-10-07
Cleanup refactor

gforsyth:
This is the first of two PRs -- this one is more important than the other.  

We still allow failures on Travis, which means we haven't had any flake checks running (or they run, but don't really get reported).  This should obviously be changed.  

After manually checking on my local machine, I discovered some issues with the recent refactor, like doubly defined methods.  

This removes those in what I hope is the correct way.  Tests still pass locally and I think it's ok, but I would really appreciate it if everyone could try out their edge cases to make sure this behaves the same as master (or hell, even better?)

Next PR will just be simple flake cleanups removing unused imports, etc...

### DIFF
--- a/xonsh/base_shell.py
+++ b/xonsh/base_shell.py
@@ -6,8 +6,8 @@ import sys
 import time
 import builtins
 
-from xonsh.tools import (XonshError, escape_windows_cmd_string, print_exception,
-                         DefaultNotGiven, check_for_partial_string)
+from xonsh.tools import (XonshError, print_exception, DefaultNotGiven,
+                         check_for_partial_string)
 from xonsh.platform import HAS_PYGMENTS, ON_WINDOWS
 from xonsh.codecache import (should_use_cache, code_cache_name,
                              code_cache_check, get_cache_filename,
@@ -156,13 +156,6 @@ class _TeeStd(io.TextIOBase):
         loc = self.std.tell()
         self.std.seek(loc + len(s))
         return s
-
-    def write(self, s):
-        """Write a string to both streams and return the length written to the
-        in-memory stream.
-        """
-        self.std.write(s)
-        return self.mem.write(s)
 
 
 class Tee:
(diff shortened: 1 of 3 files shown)


xonsh/base_shell.py
```diff
@@ -157,13 +157,6 @@ def readline(self, size=-1):
         self.std.seek(loc + len(s))
         return s
 
-    def write(self, s):
-        """Write a string to both streams and return the length written to the
-        in-memory stream.
-        """
-        self.std.write(s)
-        return self.mem.write(s)
```
scopatz: This method should not be removed because it may be used by downstream users.

gforsyth: Hey @scopatz -- I'm happy to restore it, but there's also a method `self.write` defined here: https://github.com/xonsh/xonsh/pull/1810/files#diff-8534328692230305271758ac1f48a813R114 which doesn't return the `self.mem.write(s)` -- should the `write` on line 114 be removed instead?  I'll confess that I couldn't grok how the difference between them would propagate down the pipeline

scopatz: OMG @gforsyth, you are totally write(). I don't think that it matters much which one is used.  Let's proceed with this version.  Sorry I was AWOL the past few days.

xonsh/proc.py
```diff
@@ -1008,9 +1009,6 @@ def run(self):
         if self.f is None:
             return
         last_in_pipeline = self._wait_and_getattr('last_in_pipeline')
-        if last_in_pipeline:
-            capout = self._wait_and_getattr('captured_stdout')
-            caperr = self._wait_and_getattr('captured_stderr')
```
scopatz: These shouldn't be removed per se, because they guarantee that the pipeline can read from stdout & stderr file pipes. Also, I think that this has been slightly refactored in one of my other PRs.

laerus:
note that #1812 is already merged.

gforsyth:
Huh, so it has.  Well... I guess I'll close this.  We can either add in the `caperr` and `capout` in a separate PR or just defer to @scopatz refactored version

PR #1812, opened 2016-10-07
More cleanup refactor

gforsyth:
As promised, this fixes all the existing PEP8 violations that we missed since we aren't reporting the results of flake8 in CI.  

This is built on top of #1810 so it shouldn't be merged until that's sorted.


melund:
Lgtm. Thanks for doing all the boring work  :)

PR #1813, opened 2016-10-07
Ci fix

gforsyth:
Moves flake8 checks to circle, stops allowing failures on travis nightly and removes coverage checks from travis since they weren't working anyway.  

I'm happy to have coverage reports back if someone wants to put in the time to get them working properly but otherwise it's just noise.


gforsyth:
Ok -- this is working correctly now.  Note that the failures on Circle CI are the flake8 failures that should be fixed in #1810 and #1812

scopatz:
Looks like this is failing in some places @gforsyth. Thanks for putting this in!

scopatz:
Thanks @gforsyth! Just the one small issue and this has some failure on CI. Not sure if those are my fault, though...

gforsyth:
Ok, this is ready for another look.  Allows failures on Python nightly since that's now 3.7 -- we could add a 3.6 check somewhere if we felt like it.  flake8 checks done on 3.5 on circle CI

astronouth7303:
I feel like we should have a 3.6 check, so that we know if there's problems with the upcoming version. Assuming any of the CIs support it.

gforsyth:
Travis has a "3.6-dev" option.  I'm happy to add that, although we might need to allow failures for the moment.  I'd like to get this in sooner, so that we're properly checking other PRs

requirements-tests.txt
```diff
@@ -10,5 +10,3 @@ pygments==2.1.3
 codecov==2.0.5
 flake8==2.6.2
 coverage==4.2
-numpy==1.10.4
-matplotlib==1.5.1
```
astronouth7303: Are we not testing xontribs?

gforsyth: I don't think it's worth it on nightly or 3.6 dev. The install time is ugly on travis

scopatz:
Thanks @gforsyth!

PR #1823, opened 2016-10-09
Fix getting git branch name when status.short = true

### LINKED ISSUE #1822: Prompt `curr_branch` breaks when git config has status.short
If git config contains:
```ini
[status]
    short = true
```
then [the current method of getting the git branch](https://github.com/xonsh/xonsh/blob/619ae11486f2e4febfc2218c0d53c7b3f3c9ea49/xonsh/prompt/vc_branch.py#L16-L25) fails, even if `status.branch = true` is also set.

`git status` is a heavy-handed way of doing this anyway; better would be `git rev-parse --abbrev-ref HEAD` IMO. (PR to follow.)

OJFord:
This PR fixes #1822.

### DIFF
--- a/news/git_branch.rst
+++ b/news/git_branch.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* ``curr_branch`` reports correctly when ``git config status.short true`` is used
+
+**Security:** None
--- a/xonsh/prompt/vc_branch.py
+++ b/xonsh/prompt/vc_branch.py
@@ -15,17 +15,25 @@ import xonsh.tools as xt
 
 def _get_git_branch(q):
     try:
-        branch = subprocess.check_output(
-            ['git', 'rev-parse', '--abbrev-ref', 'HEAD'],
+        branches = xt.decode_bytes(subprocess.check_output(
+            ['git', 'branch'],
             stderr=subprocess.DEVNULL
-        )
+        )).splitlines()
     except (subprocess.CalledProcessError, OSError):
         q.put(None)
     else:
-        info = xt.decode_bytes(branch)
-        branch = info.splitlines()[0].split()[-1]
-        q.put(branch)
+        for branch in branches:
+            if not branch.startswith('* '):
+                continue
+            elif branch.endswith(')'):
+                branch = branch.split()[-1][:-1]
+            else:
+                branch = branch.split()[-1]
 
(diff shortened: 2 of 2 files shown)


laerus:
Hey @OJFord! The only problem with `git rev-parse --abbrev-ref HEAD` is that when your HEAD is detached on a commit what you get back is just `HEAD` and not the actually commit hash. I think we can get around #1822 by adding `--long` : `git status --long`.

OJFord:
Ah, true. `--long` works. Another alternative would be `git branch`, and grabbing the line that starts with `* ` - I mention it only because it's probably faster, though I'm not sure how often it would be noticeable in the real-world.

laerus:
For slower machines it would really make a difference cause we call git in each prompt update. Though keep in mind that we can't parse the line using English words because we had problems with ppl having a different default language for git, but it could be done in the same way the current implementation grabs the branch, that is strip/split and get the last 'word'.

OJFord:
@laerus Yep, this has no reliance on specific words, just the `* ` present at the beginning of the current branch.

I have tested this locally with `git config status.short true` on this branch and on master (where the `curr_branch` in prompt becomes "Dockerfile" 😆 ) and with a detached `HEAD`, where this branch also now shows the commit SHA.

OJFord:
I'm not sure what the lint problem is by the way - it doesn't link anything and when I run pylint locally nothing it reports was introduced by my commits.

scopatz:
> I'm not sure what the lint problem is by the way - it doesn't link anything and when I run pylint locally nothing it reports was introduced by my commits.

Thanks for putting this in @OJFord!  This is our own PR-linter, not a code linter.  It says that you need to put in a new entry as per our dev guide here: http://xon.sh/devguide.html#changelog  This is not hard but really helps users.

OJFord:
Ah! Sorry about that, should've read more closely. Fixed.

news/git_branch.rst
```diff
@@ -0,0 +1,13 @@
...
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* `curr_branch` reports correctly when `git config status.short true` is used
```
scopatz: Need to use double backticks in rST.

OJFord: Bah, thanks!

scopatz:
@laerus - I'll leave this to you since I haven't really been following the git status stuff at all :).

laerus:
This works great, thank you @OJFord!

Issue #1828, opened 2016-10-11
streaming aliases fail with !(alias)


melund:
I found an issue with the streaming aliases. 

```
def _test_stream(args, stdin, stdout, stderr):
    print('hallo on err', file=stderr)
    print('hallo on out', file=stdout)
    return 1

aliases['test-stream'] = _test_stream
```

The `![]` works fine:
```
$ ![test-stream] 
hallo on out
hallo on err
```

But `!()` gives me this error: 
```
$ !(test-stream)
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "c:\users\mel\documents\xonsh\xonsh\base_shell.py", line 267, in default
    run_compiled_code(code, self.ctx, None, 'single')
  File "c:\users\mel\documents\xonsh\xonsh\codecache.py", line 65, in run_compiled_code
    func(code, glb, loc)
  File "<xonsh-code>", line 1, in <module>
  File "c:\users\mel\documents\xonsh\xonsh\main.py", line 132, in _pprint_displayhook
    printed_val = pretty(value)
  File "c:\users\mel\documents\xonsh\xonsh\pretty.py", line 120, in pretty
    printer.pretty(obj)
  File "c:\users\mel\documents\xonsh\xonsh\pretty.py", line 375, in pretty
    return _default_pprint(obj, self, cycle)
  File "c:\users\mel\documents\xonsh\xonsh\pretty.py", line 494, in _default_pprint
    _repr_pprint(obj, p, cycle)
  File "c:\users\mel\documents\xonsh\xonsh\pretty.py", line 691, in _repr_pprint
    output = repr(obj)
  File "c:\users\mel\documents\xonsh\xonsh\proc.py", line 1397, in __repr__
    s += ', '.join(a + '=' + str(getattr(self, a)) for a in self.attrnames)
  File "c:\users\mel\documents\xonsh\xonsh\proc.py", line 1397, in <genexpr>
    s += ', '.join(a + '=' + str(getattr(self, a)) for a in self.attrnames)
  File "c:\users\mel\documents\xonsh\xonsh\proc.py", line 1704, in returncode
    self.end()
  File "c:\users\mel\documents\xonsh\xonsh\proc.py", line 1553, in end
    for _ in self.tee_stdout():
  File "c:\users\mel\documents\xonsh\xonsh\proc.py", line 1490, in tee_stdout
    for line in self.iterraw():
[...]

Issue #1826, opened 2016-10-11
`which` return code is always 0


derekbrokeit:
In xonsh, I am currently receiving `0` return code even when the command cannot be found. This makes it impossible to use !(which program-that-doesnt-exist) as a boolean or other similar effects. Ideally, `which` should exit with status `1` when no item is found.

Example use case
```
$PAGER = "vimpager" if !(which vimpager) else "less"
```

melund:
This used to work. Probably broke in one of the refactors. I will try to take a look. @derekbrokeit  Thanks for spotting it.

melund:
There seems to be a bug with the return values of callable aliases. The which function does return non-zero if it doesn't find anything. But xonsh seems to ignore this. 

I checked, and this was also a problem before the recent refracture of the subproc system. But after the refracture I can't enter pdb breakpoints in the code. So it is a little difficult to quickly debug. @scopatz How do you get `pdb` working?

melund:
Ahh. A bug was introduced in the 'which' refactor 14 days ago. I fix is incoming

melund:
Still no idea why I can't enter the `pdb` break point on master. Anyone.?

melund:
@derekbrokeit  So there is a fix here #1829. But it probably won't help you much as `!(which)` fails due to some other bug with streaming aliases #1828 . Probably needs to be fixed first.

melund:
So. This problem was also fixed in #1803 which I just merged. But the problem with `!(fun_alias)` #1828 is still there...

scopatz:
> So it is a little difficult to quickly debug. @scopatz How do you get pdb working?

Sorry @melund! I was out for the holiday!  I am back now.  I am not sure why pdb isn't working... That is concerning.  I'll take a look at this and the other problems in their own issues, which seems to have been opened up/

PR #1829, opened 2016-10-11
Which returns

melund:
This should fix the regression with `which` not giving the correct return code. See #1826 

Unfortunately, it also revealed another bug with streaming aliases, see #1828 . So `!(which)` will fail with this fix....


melund:
This is actually also fixed in the waiting #1803 ..  So I am just goint to close this.

PR #1834, opened 2016-10-11
fix for aliases that start with a comma

gforsyth:
For 0.4.7 I added in logic to split on commas so that completion worked
inline with comma separated terms.  This inadvertently broke completion
behavior for aliases starting with a comma.  The rest of the characters
that we split on aren't valid syntax for beginning an alias, but commas
are.


melund:
LGTM

PR #1837, opened 2016-10-12
Fix which prints 

melund:
So I am just opening  a new PR for the `which` stuff, which wasn't addressed in @scopatz's #1803 PR

This fixes the print statements when `which` doesn't find any matches, and ensures they go to stderr.

### DIFF
--- a/xonsh/xoreutils/which.py
+++ b/xonsh/xoreutils/which.py
@@ -97,7 +97,10 @@ def which(args, stdin=None, stdout=None, stderr=None, spec=None):
 
     pargs = parser.parse_args(args)
     verbose = pargs.verbose or pargs.all
-    captured = spec.captured in xproc.STDOUT_CAPTURE_KINDS
+    if spec is not None:
+        captured = spec.captured in xproc.STDOUT_CAPTURE_KINDS
+    else:
+        captured = False
     if pargs.plain:
         verbose = False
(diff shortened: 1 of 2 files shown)


xonsh/xoreutils/which.py
```diff
@@ -97,7 +97,10 @@ def which(args, stdin=None, stdout=None, stderr=None, spec=None):
 
     pargs = parser.parse_args(args)
     verbose = pargs.verbose or pargs.all
-    captured = spec.captured in xproc.STDOUT_CAPTURE_KINDS
+    if spec is not None:
+        captured = spec.captured in xproc.STDOUT_CAPTURE_KINDS
+    else:
+        captured = False
```
scopatz: I think these do the same thing since None is not in `xproc.STDOUT_CAPTURE_KINDS`, right?

scopatz:
Thanks for this @melund!  Just the one small comment, but otherwise, I think is this good to go.

xonsh/xoreutils/which.py
melund: Exactly. The `spec` could be None if not passed as a kwarg.

scopatz: whoops! my bad

scopatz:
Thanks @melund

Issue #1830, opened 2016-10-11
Clearscreen (cls and clear) broken on master (windows)


melund:
Clear screen `cls` or `clear` (from MSYS) are both broken on master. I believe it comes after the sub-proc refactor. 

![image](https://cloud.githubusercontent.com/assets/1038978/19263848/a646b5e2-8f9e-11e6-914a-fa398c5e9c8a.png)

Instead of clearing the screen it just prints an unrecognizable character.

laerus:
i cannot reproduce this in linux, probably some windows specific code is causing this?

gforsyth:
Hey @melund, is there any difference if you try a different terminal emulator?

astronouth7303:
Is this even something xonsh can control?

melund:
Apparently, since it works on 0.4.7. I have to read up on how `cls` works for Windows conhost.

melund:
I am thinking that maybe subprocess are executed in their own process so that the does affect state of xonsh. Maybe that is also why I can't add pdb brakepoints in callable aliases. @scopatz could that be the reason for what I observe with the `cls` alias not working.

scopatz:
Potential fix in #1845

melund:
Fixed in master

PR #1845, opened 2016-10-14
added clear to foreground, since it is not really needed

scopatz:
And also cls.  @melund, would you mind seeing if this helps out #1830, please? Thanks!

### DIFF
--- a/news/not_backgroundable.rst
+++ b/news/not_backgroundable.rst
@@ -1,0 +1,29 @@
+**Added:** None
+
+**Changed:**
+
+* The following commands are, by deafult, predicted to be not backgroundable
+  in some circumastances:
+
+    * bash
+    * csh
+    * clear
+    * cls
+    * fish
+    * ksh
+    * sh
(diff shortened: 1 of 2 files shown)


melund:
@scopatz . The code isn't quite working since aliases (`aliases['cls'] = ['cmd', '/c', 'cls']`) are always predicted as background-able: 
https://github.com/xonsh/xonsh/blob/master/xonsh/commands_cache.py#L173

I fixed that locally so it doesn't return True before reaching the `backgroundable_predictors` check. The next problem was that all keys in command_cache are stored in upper case on windows. So I needed to add the following: 

```python
if ON_WINDOWS:
    name = name.lower()
predictor = self.backgroundable_predictors[name]
```
Then it fixed the problem, and `cls` works again.

melund:
Actually, it may be better to make all keys in `default_backgroundable_predictors` uppercase on Windows, so it match the keys in the command_cache.

PR #1847, opened 2016-10-14
No background certain commands

scopatz:
This builds on #1845 and should really help with performance for common commands. @laerus, you might be most affected by some of these, so would you mind trying it out?

### DIFF
--- a/news/not_backgroundable.rst
+++ b/news/not_backgroundable.rst
@@ -11,12 +11,19 @@
     * cls
     * fish
     * ksh
+    * less
+    * man
+    * more
     * sh
     * ssh
     * startx
+    * sudo
     * tcsh
+    * top
     * vi
     * vim
+    * vimpager
+    * xo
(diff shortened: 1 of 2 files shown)


laerus:
never had problems with this on my main machine, i'll get on the slow netbook and try this out sometime today.

laerus:
yes this branch feels more stable on these commands! But does this mean that we can't background them with ctrl-Z anymore?

scopatz:
You can still background.  This just means that they don't run on a background thread.  I plan on cleaning up the name to make it more understandable. Sorry!

laerus:
i will have to go over the codebase some more times to grok what's going on, job backgrounding isn't working atm right? any hint on where is the problem?

scopatz:
I think job backgrounding is working, but if it is broken, it might depend on the command. In any event, that is a separate issue from this, which is mainly for performance improvements.

PR #1848, opened 2016-10-14
fix for streaming aliases

scopatz:
This should address #1828.  @melund, would you mind taking it for a spin, please? Thanks!

### DIFF
--- a/news/sa.rst
+++ b/news/sa.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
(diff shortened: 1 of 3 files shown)


melund:
Thanks. Fixes it. It seems like the CircleCI failures are not caused by this PR. It can be merge once Travis finishes.

gforsyth:
Merging this.  Travis OSX builds are down again and there's nothing here that's OSX specific

PR #1852, opened 2016-10-15
fix command_cache aliases on Windows

melund:
This PR fixes two things. 

1. A bug where tab-completing aliases on windows would convert the name to upper-case. (Keys in the command_cache are upper-case on Windows. We do handle that nicely, but not in the case of aliases. )

2. This fixes the cls/clear bug on windows (#1830). The `cls`/`clear` aliases were detected as being background-able because they `CLEAR`/`CLS` keys in the command_cache did not match the lower case keys in `default_backgroundable_predictors`


laerus:
circle complains cause of flake8

melund:
I think only one if them was my fault. We better fix the lingering pep8 issues in separate PR.

melund:
OK. Maybe two of them :/

xonsh/commands_cache.py
```diff
@@ -163,15 +162,18 @@ def lazy_locate_binary(self, name):
...
             return self._cmds_cache[cached][0]
-        elif os.path.isfile(name) and name != os.path.basename(name):
+        elif os.path.isfile(name) and name != pathbasename(name):
             return name
 
     def predict_backgroundable(self, cmd):
         """Predics whether a command list is backgroundable."""
         name = self.cached_name(cmd[0])
-        path, is_alias = self.lazyget(name, (None, None))
-        if path is None or is_alias:
```
laerus: isn't this supposed to run on linux aswell?

melund: I don't see the point for this on Linux/Mac but prove me wrong. The default dict backgroundable_predicters will return true anyway if the command is not found.

laerus: right! thanks

scopatz: The first clause is a speed up because calling a function is relatively expensive, the second clause was because this whole formulation doesn't apply to callable aliases at all.

scopatz:

xonsh/commands_cache.py
melund: @scopatz How can we see the difference between a alias and a callable alias from the commans_cache. Just so I haven't broken something else in this PR

scopatz: Right now we can't.  I think that this is fine though.  I might put the `path is None` check back in, though.

melund: Oh... But it is there.  We only need it for Windows where we need to check the original cased name for the command.

PR #1854, opened 2016-10-15
make safe readble even more safe

scopatz:
@laerus from https://gitter.im/xonsh/xonsh?at=58022358457ae29b71e0506d


gforsyth:
Thanks, @scopatz!

PR #1855, opened 2016-10-15
Replaced backgroundable with threadable

scopatz:
This should make it more clear whether or not a process is run on a thread and will hopefully prevent confusion with backgrounding a process.


gforsyth:
Thanks, @scopatz !

scopatz:
Thanks!

PR #1858, opened 2016-10-16
Heal test ptk_multiline/replay/run_subproc

laerus:
worst offender in landscape health.


laerus:
TFW tests that have not been touched fail

laerus:
`tests/test_replay.py` fails for some reason

running only `test_replay` passes
running only `test_ptk_multiline` passes

running first `test_ptk_multiline` and then `test_replay` fails
but not vice versa

*pytest unicorn*

melund:
Yes. We have a few of those in the test framework. Do have any clues to what is causing it here? I have an idea that we should use monkeypatch when constructing the xonsh_builtins fixture. There may also be some classes in xonsh that have some unintended side effects when imported in the test code.

laerus:
We have some objects sticking in `builtins` between tests for sure, no idea which ones yet.

laerus:
maybe we can add a fixture that does a builtins teardown after each test module.

laerus:
circle complaining on `test_run_subproc`, seems like a legit bug, 
`builtins.__xonsh_env__` seems to get deleted before xonsh_builtins fixture 
tries to clean it up.

scopatz:
I think a better solution to isolation is to run the tests in a separate process.  This is what we are doing now in `test_integrations.py`.  Maybe these tests can/should be migrated there.

laerus:
a small refactor for the setup logic in test_integrations could be a seperate PR

tests/test_integrations.py
```diff
@@ -108,3 +110,74 @@ def test_script(case):
...
+                         )
+    try:
+        out, err = p.communicate(input=cmd, timeout=10)
+    except subprocess.TimeoutExpired:
+        p.kill()
+        raise
+    if callable(fmt):
+        out = fmt(out)
+    assert out == exp
+    assert p.returncode == 0
```
scopatz: Is there anyway that we could generealize this function so that it is essentially not copied three times in the file?  Maybe make one `check_command()` or something?

laerus: we can make a setup fixture.

laerus:
ok i've added a function for the setup logic. Still some mismatching expected cases on running pwd, will have to figure this out.

scopatz:
The new fixture looks great, thanks!

laerus:
seems like echo throws something that looks like a hash before the actual output. strange.

scopatz:
Well, we have debug mode on, which prints the autowrapping of lines to stderr, which is what you are seeing.  Since no filename is given, because we are reading from a string, the has is used instead.  Are you redirecting stderr to stdout in any of the command that are being tested?  That would cause this.

laerus:
right! yes i'm using the same `Popen` setup that the rest of the test were using.

tests/test_integrations.py
```diff
@@ -24,6 +27,30 @@
...
+    env['XONSH_SHOW_TRACEBACK'] = '1'
+    env['RAISE_SUBPROC_ERROR'] = '1'
+    env['PROMPT'] = ''
+    xonsh = 'xonsh.bat' if ON_WINDOWS else 'xon.sh'
+    xonsh = shutil.which(xonsh, path=PATH)
+    proc = subprocess.Popen([xonsh, '--no-rc'],
+                            env=env,
+                            stdin=subprocess.PIPE,
+                            stdout=subprocess.PIPE,
+                            stderr=subprocess.STDOUT,
```
scopatz: Maybe this should be PIPE?

laerus:
allright! i've added parameters to `run_xonsh` to redirect the streams

scopatz:
Sweet! This looks great to me, thanks!

laerus:
i think we can at least run `pwd` on windows through these tests.

scopatz:
Yeah, I added that tests/bin/pwd script so that we could. We might want to add others, like echo, too.

Issue #1820, opened 2016-10-09
`parallel` command failed on xonsh


blahgeek:
Running `seq 10 | parallel echo` fails on xonsh (works on bash) while `seq 10 | xargs echo` works

also seems related to #1758

astronouth7303:
What version of xonsh?

Where is parallel from?

blahgeek:
I'm using current master version of xonsh (b8e4adbaf3b)

parallel is installed from homebrew (GNU parallel 20160922).

I did some more tests:

```
seq 10 | parallel echo  # does not work
seq 10 > tmp.txt
parallel echo < tmp.txt  # this works
```

I think this bug may be related some complex behaviors in parallel (e.g. fork, stdout/stderr redirect, etc.) but is not specific to parallel 

I can confirm that this bug is introduced by #1758 , so maybe @scopatz would like to take a look? Thanks!

blahgeek:
ps: `ls | parallel ls` also fails (it's more generic since `echo` is a shell builtin)

scopatz:
Yeah, I'll take a look. Sorry about the issues you have been having

scopatz:
This might actually be a bug with parallel exiting too soon if the stdin buffer is not full / contains an EOF.  Note that both `seq 100000 | parallel echo` and `sleep 0.1 | seq 10 | parallel echo` work for me.  Also, echo is not a builtin in xonsh (or in Bash I don't think if it is an argument to another command).

scopatz:
Note that the bug seems to be present no matter if we use Popen or PopenThread.

scopatz:
Or maybe we are just doing it wrong...

scopatz:
Ok, I have a tentative fix, but I am not happy with it yet.  Let me think on this one some more....

scopatz:
Hi @blahgeek, can you try out #1859, which fixes this issue for me, please? Thanks!

blahgeek:
@scopatz Thanks a lot! But it does not seem work for me. If I run `seq 10 | parallel echo`, xonsh freezes about 1 second and then prints only the header part of parallel and exit:

```
> seq 10 | parallel echo
Academic tradition requires you to cite works you base your article on.
When using programs that use GNU Parallel to process data for publication
please cite:

  O. Tange (2011): GNU Parallel - The Command-Line Power Tool,
  ;login: The USENIX Magazine, February 2011:42-47.

This helps funding further development; AND IT WON'T COST YOU A CENT.
[...]

PR #1859, opened 2016-10-16
Parallel delay

scopatz:
This should address #1820


laerus:
going in then, ty @scopatz !

scopatz:
Thanks @laerus!

PR #1862, opened 2016-10-17
First class path objects

chronitis:
This changeset provides syntactic sugar for using `pathlib.Path` objects in xonsh directly, enabling them to be written as literals and generated by backtick searches.

```
>>> p"foo/bar"
PosixPath("foo/bar")
>>> p`foo/.*`
[PosixPath("foo/bar")]
```

The rationale being that a great deal of shell scripting involves performing some sort of manipulation on paths, and being able to get path objects directly means you can benefit from manipulation methods directly rather than needing to deal with parsing and reforming the string representation manually.


news/first-class-paths.rst
```diff
@@ -0,0 +1,20 @@
+**Added:**
+
+Since shells are all about paths, some syntactic sugar to make using
```
astronouth7303: You can reduce this to a a few-line summary and link to the tutorial.

news/first-class-paths.rst
```diff
@@ -0,0 +1,20 @@
...
+``pathlib.Path`` objects easy is added:
+
+* p-strings: ``p'/foo/bar'`` is short for ``Path("/foo/bar")``
+* p-backticks: the ``p`` modifier to a backticks search (``p`.*```) returns a
+  list of ``Path`` objects instead of strings. This can be combined with the
+  existing modifiers, eg ``pg`**``` to use glob instead of regex. (This only
```
astronouth7303: Mention this in the documentation

docs/tutorial.rst
```diff
@@ -953,6 +953,42 @@ The following example shows the form of these functions:
...
+
+Path objects can be used without further conversion in subprocess mode:
+
+.. code-block:: xonshcon
+
+    >>> x = p"some/path"
+    >>> echo @(x)
+    some/path
+
+
```
astronouth7303: Please mention the combining and if order matters.

docs/tutorial.rst
```diff
@@ -953,6 +953,42 @@ The following example shows the form of these functions:
     ['aa', 'aaa', 'aab', 'aabb']
 
 
+Path Objects
+============
+
+Shells spend a lot of their time manipulating paths, so you may find it useful
+to handle paths as objects instead of string. A shorthand is provided to
+instantiate ``pathlib.Path`` objects:
```
astronouth7303: Link, please. `pathlib` is not the most well-known module yet.

astronouth7303:
In general, I really like this idea. I think I would use it a lot.

I'm not sure if we want to do this by default or in this PR, but what about doing `expanduser()` in string forms? Should this be a different modifier or is part of `p""` ok?

scopatz:
First off, thanks a ton @chronitis for putting this in!

> In general, I really like this idea. I think I would use it a lot.

I am +1 to the idea.

> I'm not sure if we want to do this by default or in this PR, but what about doing expanduser() in string forms? Should this be a different modifier or is part of p"" ok?

I am happy either way, though automatic tilde expansion would make the most sense to me.

docs/tutorial.rst
scopatz: I also think this should come right before custom globbing section.  Also, it should `----` as the section underline, rather than `===`

xonsh/built_ins.py
```diff
@@ -150,16 +154,19 @@ def globsearch(s):
...
-    or arbitrary search function).
+    or arbitrary search function). If pathobj=True, the return is a list of
+    pathlib.Path objects instead of strings.
     """
     if (not callable(func) or
             len(inspect.signature(func).parameters) != 1):
         error = "%r is not a known path search function"
         raise XonshError(error % func)
     o = func(s)
+    if pathobj and pymode: #this doesn't make sense in subprocess mode
```
scopatz: PEP8 two spaces before `#` and one space after.

xonsh/parsers/base.py
```diff
@@ -1973,9 +1977,15 @@ def p_nocloser_many(self, p):
...
         """string_literal : string_tok"""
         p1 = p[1]
-        s = ast.literal_eval(p1.value)
-        cls = ast.Bytes if p1.value.startswith('b') else ast.Str
-        p[0] = cls(s=s, lineno=p1.lineno, col_offset=p1.lexpos)
+        if p1.value.startswith('p'):
+            s = ast.Str(s=ast.literal_eval(p1.value[1:]), lineno=p1.lineno,
+                        col_offset=p1.lexpos)
+            p[0] = xonsh_call('__xonsh_path_literal__', [s],
+                              lineno=p1.lineno, col=p1.lexpos)
```
scopatz: This is not in the correct spot.  Or rather, I am against this.

scopatz:
And sorry for closing.

scopatz:
So, let me amend what I previously wrote.  I think that the backtick glob syntax is great.  I am less fond of the string literal syntax.  This is something that Python itself could conceivably implement in the future and may have a different syntax, or worse, take the p"" syntax and then give it a different meaning.  

So on the whole, I think that there are two issues here.  I think that the backtick part could go in as-is.  The string part probably warrants further discussion and buy-in from the other core-devs.  If we go ahead with it, I think it needs a big red flag that it is subject to change based on what Python does. 

Personally, I am +0 on the string form, since it won't really affect me much.  Another thought is that the string form might make a good PEP ~_~.

To get the backtick stuff merged in faster, I would recommend redoing this as two separate PRs.

chronitis:
Thanks for the reviews.

 * The current implementation of the path literal does not combine with any other string modifiers (`u, r, b`). Raw strings might well be worth combining (windows paths, perhaps) and probably wouldn't be too hard to handle, if desirable. It is also worth thinking how this will interact with f-strings in python 3.6 - presumably the string substitution should be done first and the result cast to Path.

 * The idea of path literals has come up occasionally on python ideas - the most recent thread I could find starts at [1]. It's hard to draw a conclusion from the thread; it seems to get bogged down in lots of unrelated issues, but there is a -1 from Guido [2] which makes me think this is unlikely to become core syntax, and therefore xonsh is *probably* free to use the syntax as desired.

I am still inclined to think the path literals are useful as well, but splitting the PR seems like a good idea for the moment.

[1]: https://mail.python.org/pipermail/python-ideas/2016-March/039132.html
[2]: https://mail.python.org/pipermail/python-ideas/2016-March/039163.html

xonsh/parsers/base.py
chronitis: How would you prefer to go about this?

It is necessary to have some logic at this point since `p` is part of the string token but `ast.literal_eval` can't handle it. You could perhaps introduce a custom AST node and generate the path later, but that seems like it would introduce more complexity.

scopatz: Well, you could have p-strs have a different token than python supported strings.  Then there would be a different parser rule for handling them rather that the rule for handling normal strings.  I think that this is probably the way to go because it will help us be able to support separating out a pure-python parser rules from the extension rules, which is still a goal that we maintain.  

Additionally, doing it this way is slightly faster since you don't have to eval an if-statement that has already effectively been evaluated by the parser.

news/first-class-paths.rst
```diff
@@ -1,13 +1,6 @@
...
 
-Since shells are all about paths, some syntactic sugar to make using
-``pathlib.Path`` objects easy is added:
-
-* p-strings: ``p'/foo/bar'`` is short for ``Path("/foo/bar")``
-* p-backticks: the ``p`` modifier to a backticks search (``p`.*```) returns a
-  list of ``Path`` objects instead of strings. This can be combined with the
-  existing modifiers, eg ``pg`**``` to use glob instead of regex. (This only
-  applies in python mode).
+* p-strings: ``p'/foo/bar'`` is short for ``pathlib.Path('/foo/bar')``
```
astronouth7303: I would still suggest linking.

scopatz:
Fair points @chronitis on the p-strings.  Maybe the rest of @xonsh/core-developers can comment on their feelings about it.

gforsyth:
I don't have strong feelings on this -- I don't think I'll use them, but it doesn't bother me if they're in there.  And I do like and feel that xonsh can operate as a testing ground for new Python features.  Given what @chronitis linked in re: p-strings, it does seem unlikely that they'll make it in to Python, so no conflict there.  But if `p""` gets used for something else, well, we're still sub-1.0 and there are no promises about syntax.  

So I think that's me tacitly approving?

astronouth7303:
Part of me wants to make raw default, because windows and you very rarely need escapes, but I'm 100% certain that will bite us later.

chronitis:
Yes, I'm pretty sure raw-by-default would be a problem. Supporting optionally-raw strings shouldn't be a problem though - I think `ast.literal_eval` deals with the details for us.

With respect to the current implementation and whether it would be better to have a seperate token, I had a look and I'm a bit uncertain whether it would actually be beneficial - you replace a branch in the AST node with an extra branch in the tokenisation loop, which might arguably cost more, although path tokens can probably use simplified string parsing - multiline or triple quoting can probably be dispensed with.

I wonder if it makes sense to defer the implementation (or at least, do it in parallel with) python 3.6 f-strings, which have pretty much the same parser and AST issues - an extra string modifier which needs to generate a runtime call, and perhaps both can be handled more efficiently together?

astronouth7303:
Any progress on this?

chronitis:
Sorry, this seems to have been forgotten rather after the other bit got merged. I'll take another look at it in the next few days.

astronouth7303:
Also, with 3.6 now rc, f-strings should be addressed.

My thought is that `pf"mypath/{myvar}"` should map to `Path(f"mypath/{myvar}")` so that it becomes evaluated. Doing it in the inverse fashion would essentially create a closure and have all kinds of spooky action-at-a-distance gotchas.

chronitis:
This should fix allowing raw path strings (`rp'/foo'`). While doing this, I also noticed that the current implementation does not handle byte strings properly - only `b'foo'` works, not anything else (`B`, `rb`, etc), which I've fixed and added tests for, since it requires the same prefix checking logic.

Is anything else required? As far as I can see, f-strings don't yet exist in the codebase so I'm not sure how to address them short of writing the support as part of this PR. I would anticipate that they could be supported using more of less the same logic - detecting the prefix and generating a function call in `p_string_literal`, and the logic of ensuring `p` and `f` nest correctly shouldn't be too complicated.

(There is a slight inefficiency at the moment that we end up re-tokenising the string in `p_string_literal` to extract the prefix, but I'm not sure whether adding a separate `StringPrefix` token would actually be either simpler or faster).

scopatz:
This looks pretty good to me. But I will let @astronouth7303 and @gforsyth make the call.

astronouth7303:
I try to stay out of the parser.

That said, some tests to check that `pr` actually does do raw strings would be good.

Otherwise, I think it's a great idea and a fine addition to xonsh.

gforsyth:
Thanks, @chronitis! This is great to have!

PR #1865, opened 2016-10-18
Add support for path-yielding backticks

chronitis:
This is a split of #1862 containing only the path backticks part.


scopatz:
Sweet, thanks!

PR #1869, opened 2016-10-18
Test cleanup

gforsyth:
I know I should never look at landscape.

But, in this case, maybe not so bad.  Fixed some "code smells", but there were a number of straight up duplicated tests which could cause problems down the line if we needed to change something.


laerus:
tyvm @gforsyth !

PR #1872, opened 2016-10-19
Alt enter

gforsyth:
Enables Alt+Enter to force execute a code block in PTK irrespective of cursor position.

Also changes all of the filter definitions to be consistent and reordered the key_bindings file a bit to group things more sensibly.


scopatz:
This seems reasonable to me @gforsyth! Just waiting on Travis...

PR #1873, opened 2016-10-19
Use doctr for dev docs

gforsyth:
Uses [doctr](https://github.com/drdoctr/doctr) to build and deploy docs after each PR is merged to `xonsh.github.io/dev`.

Adds in an encrypted deploy key to the `.travis.yml` and also adds `doctr` as a requirement for the `BUILD_DOCS` env.  

For the moment, this is just for dev docs -- note that we won't actually know if it works completely until the PR is merged, since documents aren't deployed otherwise.  
If everything goes as planned, I can add in some logic to also deploy release docs when there's a release tag added, but that can come later.

Also to do: add a link to the dev docs on the main `xon.sh` page (or have a menu for different version of the docs?)


news/doctr.rst
```diff
@@ -1,6 +1,6 @@
 **Added:**
 
-* Use `doctr` to deploy dev docs to github pages
+* Use ``doctr`` to deploy dev docs to github pages
```
astronouth7303: Link?

gforsyth: added.  I am bad at promotion

scopatz:
Sweet! I have been waiting for this @gforsyth!

scopatz:
Thanks @gforsyth!

gforsyth:
Right -- I disabled push builds a while ago since Travis was lagging so hard, but `doctr` only pushes after PRs are merged.  

I've reenabled the push builds -- should trigger the next time there's a commit made to master.  And since we now offload some tests to Circle it shouldn't be as much of a problem.

gforsyth:
Ok... so it works and it doesn't.  
When we wrote `doctr` we were targeting docs being served from `gh-pages` branches of the same repo.  So we only deploy to the gh-pages branch.  That said, the organizational github pages sites only support serving from master.  
I'll add support for deploying to any branch, then we'll have to push a new `doctr` release and then this should work as expected.

gforsyth:
Ugh, there's a whole load of problems with having `doctr` deploy to master.  Any objections to just creating a `xonsh_docs` repo that isn't a `github.io` organizational site?

scopatz:
I am fine with it as long as it is `xonsh-docs` :)

PR #1882, opened 2016-10-22
Fix for windows quoting issue with machines

janschulz:
The problem is that `argvquote(r'\\machine\dir\file.bat', force=True)` returns a wrongly quoted string: `"\\\machine\dir\file.bat"` (note the three instead of two `\` in the beginning). This fix adds a testcase for such paths and reimplements the mentioned algorithm in argvquote more closely.

CC: @melund

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -915,19 +915,25 @@ def test_escape_windows_cmd_string(st, esc):
     assert esc == obs
 
 
-@pytest.mark.parametrize('st, esc', [
-    ('', '""'),
-    ('foo', 'foo'),
+@pytest.mark.parametrize('st, esc, forced', [
+    ('', '""', None),
+    ('foo', 'foo', '"foo"'),
     (r'arg1 "hallo, "world""  "\some\path with\spaces")',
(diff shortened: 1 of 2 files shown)


janschulz:
Original bugreport in https://github.com/Cadair/jupyter_environment_kernels/issues/25 by @bytesemantics

melund:
Ahh, I see where I went wrong when porting the algorithm. It is very dangerous to be creative when doing such things. 

Big kudos for spotting and fixing this error @janschulz. And thanks a lot for all the tests. It is missing a news item, but I can add that. Thanks for this fix.

melund:
And nice to see that the code snippet is used somewhere else :)

janschulz:
If xonsh would put that part of the code (basically `execute a command and get a back the env variables`) into it's own package, I (and probably the conda-kernel people) would be very happy :-)

scopatz:
@janschulz - that is a great idea.  We have already broken out some other projects. We could make a repo int the xonsh org and give you collaborator access if you'd be willing to help out with that.  Otherwise it will probably have to wait until all of these subproc issues get worked out for me (personally) have the opportunity to split anything out myself.

PR #1883, opened 2016-10-22
Add news item for #1882

melund:


astronouth7303:
Will the @'s link when docs are generated?

scopatz:
I don't think so, sadly, though there is probably some fancy sphinx extension for that

scopatz:
So it actually doesn't look like there is one, but it probably wouldn't be hard to write...

Issue #1824, opened 2016-10-10
Delay when autocompleting file/directory


rnovacek:
When I'm using autocompletion for files and directories, there is a delay between pressing the <TAB> button and the completion appearing. The delay is ~1 second. 

Example:
$ touch somefile
$ ls somefil<TAB>
Nothing happens for like one second, then completion appears. I would prefer the completion to appear immediately. Other completions (command name, environment variable) appear immediately when I press the <TAB> button.

Any ideas how to debug the delay?

xonfig:
```
+------------------+-----------------+
| xonsh            | 0.4.7           |
| Git SHA          | None            |
| Python           | 3.5.1           |
| PLY              | 3.7             |
| have readline    | True            |
| prompt toolkit   | 1.0.7           |
| shell type       | prompt_toolkit  |
| pygments         | 2.1.3           |
| on posix         | True            |
| on linux         | True            |
| distro           | fedora          |
| on darwin        | False           |
| on windows       | False           |
| on cygwin        | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
+------------------+-----------------+
```

rnovacek:
I just found out that it only affects some commands:

```
$ mkdir somedir; touch somefile
$ ls somefi<TAB>  # delay appears, completion appears after ~1s
$ cd som<TAB>  # competion to 'somedir/' appears immediately
```

gforsyth:
Hi @rnovacek , thanks for reporting!  
We definitely don't want a 1 second delay for anything, really.  
Are there a *lot* of files in the directory where you see the delay?  What happens if you try `ls somefi<TAB>` in directory that only has one or two folders/files?

gforsyth:
Sorry, my brain is fried -- I see that that's what you did and there is a delay.  Ok, I'll try to dig around and see if I can see what's up.

scopatz:
My guess is that this delay is coming from Bash completions being slow for some reason.  Do you happen to know if you are using V1 or V2 Bash completions?  Also, can you try turning off Bash completions entirely and see if that improves performance.  See http://xon.sh/tutorial_completers.html#removing-a-completer

rnovacek:
Thanks for the tip. Disabling bash completion with `completer remove bash` helps. Now there is no delay when I press the TAB button.

[...]

PR #1888, opened 2016-10-25
Proper default value and documentation for $BASH_COMPLETIONS

blahgeek:
This is related to #1824

### DIFF
--- a/xonsh/completers/bash.py
+++ b/xonsh/completers/bash.py
@@ -9,7 +9,7 @@ import xonsh.platform as xp
 from xonsh.completers.path import _quote_paths
 
 BASH_COMPLETE_SCRIPT = r"""
-{sources}
+{source}
 
 # Override some functions in bash-completion, do not quote for readline
 quote_readline()
@@ -58,8 +58,8 @@ for ((i=0;i<${{#COMPREPLY[*]}};i++)) do echo ${{COMPREPLY[i]}}; done
 
 def complete_from_bash(prefix, line, begidx, endidx, ctx):
     """Completes based on results from BASH completion."""
-    sources = _collect_completions_sources()
-    if not sources:
+    source = _get_completions_source()
+    if not source:
         return set()
 
     if prefix.startswith('$'):  # do not complete env variables
@@ -83,7 +83,7 @@ def complete_from_bash(prefix, line, begidx, endidx, ctx):
         prefix_quoted = shlex.quote(prefix)
 
     script = BASH_COMPLETE_SCRIPT.format(
-        sources='\n'.join(sources), line=' '.join(shlex.quote(p) for p in splt),
+        source=source, line=' '.join(shlex.quote(p) for p in splt),
         comp_line=shlex.quote(line), n=n, cmd=shlex.quote(cmd),
         end=endidx + 1, prefix=prefix_quoted, prev=shlex.quote(prev),
     )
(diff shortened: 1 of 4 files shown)


scopatz:
LGTM @blahgeek! Waiting for CI. Thanks!

xonsh/platform.py
```diff
@@ -305,16 +305,13 @@ def BASH_COMPLETIONS_DEFAULT():
...
-        else:
-            bcd = ('/usr/share/bash-completion', )
+        bcd = ('/usr/share/bash-completion/bash_completion', )
     elif ON_DARWIN:
         bcd = ('/usr/local/share/bash-completion/bash_completion',  # v2.x
                '/usr/local/etc/bash_completion')  # v1.x
     elif ON_WINDOWS and git_for_windows_path():
         bcd = (os.path.join(git_for_windows_path(),
-                            'usr\\share\\bash-completion'),
+                            'usr\\share\\bash-completion\\bash_completion'),
```
melund: Not sure that file exist. On my machine the directory looks like this: 
``` 
snail@sea C:\Program Fi\Git\usr\share $ tree /F /A bash-completion
Folder PATH listing
Volume serial number is 00000125 FA73:929B
C:\PROGRAM FILES\GIT\USR\SHARE\BASH-COMPLETION
|   python-argcomplete.sh
|
\---completions
        subversion
        svn
        svnadmin
        svndumpfilter
        svnlook
        svnsync
        svnversion
        vim
```

blahgeek: Those files under `..bash-completion/completions/` are installed along with its corresponding  packages (like vim), usually (as in linux) an additional package `bash-completion` is required to be installed which provides `..bash-completion/bash_completion` to lazy load those scripts. I don't know about windows, is there such a package?

astronouth7303: @blahgeek that's [Git for windows](https://git-for-windows.github.io/), which is a distribution of git and msys so windows users have a useful development environment.

I can confirm the file layout given by @melund for GFW.

melund: There doesn't seem to be a file called `bash-completions`. Maybe this is just leftovers from packaging the msys environment with git.  The git completion stuff is handled by this file `mingw64\\share\\git\\completion\\git-completion.bash`.

Are there any bash command that I could run that I could run to reveal how it is configured?

scopatz: Can we just include both paths in the default?  If it is not there, it should skip it right?

melund: Yes. It shouldn't matter. If the 'Git for Windows' project ever decide to include it it will start working.

blahgeek: Yes, and the original value `...\\bash-completion` is wrong anyway.

melund: That is true. Merging in now.

xonsh/completers/bash.py
```diff
@@ -116,14 +116,9 @@ def complete_from_bash(prefix, line, begidx, endidx, ctx):
...
-def _collect_completions_sources():
-    sources = []
+def _get_completions_source():
     completers = builtins.__xonsh_env__.get('BASH_COMPLETIONS', ())
-    paths = (pathlib.Path(x) for x in completers)
-    for path in paths:
+    for path in map(pathlib.Path, completers):
         if path.is_file():
-            sources.append('source "{}"'.format(path.as_posix()))
-        elif path.is_dir():
```
mitnk: @blahgeek I think this `elif` was mistakenly missed, and that's why bash completion now is not functional? :)

Issue #1880, opened 2016-10-22
No output from scp command


blahgeek:
While running scp in xonsh, the progress does not showed up:

https://asciinema.org/a/322p80uvb0pjyaic2e51iqmhq

I'm using version 3f45378

gforsyth:
Hi @blahgeek -- could you try out the fix in #1890 and see if that resolves your issue?

gforsyth:
Also, sorry for not responding before thsi!

blahgeek:
Works for me, thanks a million!

PR #1890, opened 2016-10-25
Scp predict false

gforsyth:
fix for #1880

### DIFF
--- a/news/scp_predict_false.rst
+++ b/news/scp_predict_false.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fix ``scp`` progress not being outputted
+
+**Security:** None
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -258,6 +258,7 @@ def default_threadable_predictors():
         'less': predict_help_ver,
         'man': predict_help_ver,
         'more': predict_help_ver,
+        'scp': predict_false,
         'sh': predict_shell,
         'ssh': predict_false,
         'startx': predict_false,


laerus:
this list will keep growing, maybe we can find some heuristic for this?

scopatz:
> this list will keep growing, maybe we can find some heuristic for this?

I have been thinking a lot about this. I haven't hit upon a consistent system.  Cases like "git merge" are the weird corner cases that this has to be robust to. I think we do need some better prediction system, though, likely based on the history of commands going into alternate mode or not.

laerus:
not groking everything yet but maybe we can turn this around and have everything be non threadable by default and only switch when we want to do capturing?

scopatz:
> not groking everything yet but maybe we can turn this around and have everything be non threadable by default and only switch when we want to do capturing?

I am -1 on that because it will lead to a lot of bugs involving not streaming output during iteration or calculation.

laerus:
we only want the non blocking capturing behavior when we call `__iter__` right? do we spawn threads each time a command is run? even on plain 'ls' and 'pwd'?

scopatz:
> we only want the non blocking capturing behavior when we call __iter__ right? 

Not only then.  If you want `![cmd]` and thus `_` to have access to the output of a command, then we have to do what we are doing now whenever we stream.  The argument for skipping any capture/stream mechanics for shells, pagers, and text editors is that they don't produce any information worth storing or viewing later, and we know that ahead of time.

> do we spawn threads each time a command is run? even on plain 'ls' and 'pwd'?

Yes.

scopatz:
Thanks @gforsyth!

@laerus - we can keep the discussion going, though, here if you want.

laerus:
But `![cmd]` is by definition non capturing, we had access to the output of command with `_` before too. I think most of the time we run commands that we except to block, just a plain `ls` or `pwd` should block right? even a pipe that's not wrapped with `!()` is excepted to block until we get all the output from the last command.

I've inserted a pdb breakpoint in `iterraw` and it seems that it runs (and spawns a `NonBlockingFDReader`) even with no command, just hitting enter makes gdb hit the breakpoint 3-4 times before returning to the prompt.

I think we should go through the capturing/streaming mechanism only when we explicitly ask for it by calling the __iter__ of `CommandPipeline` somehow. Atm the only use case i can think for this behavior is when using commands/pipes in a for-loop, in every other case we just want to output to stdout or capture the whole thing with `$()` which is a blocking operation anyway.

scopatz:
So previously, we had mixed the notions of capturing and streaming. This mixing got us into a lot trouble with weird edge cases and commands blocking until they were finished, etc. For clarity, let's define

* capturing as xonsh acquiring the output (and error)
* streaming as the process of writing output or reading input as it is generated.

In this regards, the relevant distinction is that `![cmd]` is streaming and `!(cmd)` is non-streaming.  Because we want to iterate over the output, `!(cmd)` via capturing output and then not streaming it stdout. 

Also, if we want to have a single mechanism for handling all process pipelines (which I think leads to fewer and more obvious bugs), and we want capturing anywhere, we basically have to have capturing everywhere.  

And on that note, I personally *want* capturing everywhere. One of my main motivations for working on xonsh is to have a provenance system (in scientific computing lingo) that does all of the work for me. I really do want my shell to be total record of all of the little software experiments I do. So having the option of capturing and storing everything is a major use case for me. (Also, this isn't a feature of any other shell that I know of, so points to us!) I fully expect for most people to not care about this feature, but those of us who do, care deeply. It is part of the reason Jupyter is popular, etc.

The only things we shouldn't capture are when the command tells us not to capture (by going into alternate mode) or if the user says say so (via the commands cache predictors). We defintiely need a better way to predict if a command shouldn't be captured because it won't produce any non-alternate mode output, but I think this is a hard problem. I was working on some ideas before all the pipe issues arose.

Furthermore, I think we should be wary of assuming that simple commands can block. I have regularly had `ls` take 30 - 60 *minutes* on certain systems. These are big HPC machines with a parallel file system (lustre, usually) and `ls` is aliased to `ls --color=auto` (like usual).  So I do think it is a good idea to be non-blocking for everything.  You never know how painfully slow some system is going to be, and if the command is fast, it won't hang around the pipeline loop very long anyway. 

So these are the reasons I wrote it how I did. I am, of course, open to other options and implementations. I'd like for any implementation to hit all of the same features that we are getting now. 

I hope this helps!

laerus:
Alright, sorry for the misunderstanding, when i wrote capturing i meant non-blocking streaming of output either on stdout or when iterating of the command with a for loop. I want capturing of everything too, that's a very useful feature not doubt about that, but this can be done with teeing and without threads right?

My problem is with spawning threads for every command, I don't see the point of streaming the output through a `NonBlockingFDReader` of a simple `ls` (or even a pipe of cmmands) when that's what `ls` is doing anyway, it feels like going through threads just to print to stdout is making it slower than faster. Does it make a difference if you are on these big machines that `ls` takes 30-60mins if you go through threads, correctly if i'm wrong but if `ls` doesn't print anything to stdout then neither our non-blocking reading will read anything. I mean its up to the command to print/redirect it's output either as it produces it either all at once when the command ends and we can't change that.

Personally i would expect commands to block until they are finished except if i add `&` in the end which is the explicit way of making them work in the background and not block my shell. So in a slow machine you can do something like `ls  > filelist &` then continue to work in the foreground. Which atm in my slow netbook wont work cause the output it's not reliable even with low PROC_FREQUENCY.

Btw any hint about how we can make jobs start woking again?

laerus:
some simple benchmarking on my main pc with is pretty fast.
```
import time

cd /usr/bin
outs = set()
start = time.time()
for i in range(1000):
    outs.add($(ls))
print('command took: {}seconds'.format(time.time() - start))
print('difference in outs:', len(outs))
```

on 0.4.7
```
command took: 19.81511902809143seconds
difference in outs: 1
```

on current master
```
command took: 791.6223046779633seconds
difference in outs: 23
```

scopatz:
I'll respond to the rest tonight when I get a chance, but for now...

> Btw any hint about how we can make jobs start woking again?

It isn't clear right now *where* the slowdown is coming from.  If we could find that out, we could address that. 

Also, another thing I have thought of is that it might finally be time to cythonize our amalgamated files.  This should help xonsh overall, but might help a lot here.

scopatz:
@laerus Sorry for taking so long to get back to you!

>  I want capturing of everything too, that's a very useful feature not doubt about that, but this can be done with teeing and without threads right?

This is a great question! The answer is sadly no, and we know this because we have already tried.  @adqm had attempted this a while back in an attempt to replace the the now-removed TeePTY class.  Ignoring the platform dependent issues for the moment, the main issue with "tee-everywhere" is that it consume the stdout.  Say you do something like, 

``` console
$ ls | tee /tmp/myfile
```

And then you read in `/tmp/myfile`.  Now, `ls`'s stdout is not longer at TTY.  Thus you don't get color output, formatted columns, etc. Using `git diff` instead of `ls`, will not open up the PAGER, and will just dump everything to the file / screen.  For `less somefile`, you'd see similar failures.  

Additionally, because [`tee` doesn't let you set its stdin as a TTY](http://man7.org/linux/man-pages/man1/tee.1.html) the above issues problem are basically insurmountable.  You can say something very similar for using `screen` instead of `tee`.

So unfortunately, if we want to have stream, capture, and TTY-like behaviour everywhere we have to manage it ourselves.  (And it is a lot to manage!)

> it feels like going through threads just to print to stdout is making it slower than faster

Something is definitely slowing us down.  I am not sure whether it is the threads themselves or the fact that we are simply asking xonsh to do more real work for each command. I have been wrong before, but I am pretty confident that we can work out the performance issues. Maybe in the short term we could have an envvar flag that toggles using (Popen, stdout=None) and (PopenThread, stdout=TTY pipe). I think that the default should be the current behaviour, but it might help some of these performance issues.

laerus:
@scopatz thank you very much for taking the time to explain this! Some more thoughts, what about instead of an envvar we make the xonsh subproc operators act differently, for example the `$()` operator is capturing but not streaming/non-blocking, the `!()` is for streaming or something among these lines? We can assume that `$()` is a blocking operator so we can skip going through threads. (I'm still going through the code trying to grok stuff)

scopatz:
> what about instead of an envvar we make the xonsh subproc operators act differently, for example the $() operator is capturing but not streaming/non-blocking, the !() is for streaming or something among these lines?

It already does this, actually. It is a little confusing and buried pretty deep, but if either `$(cmd)` or `$[cmd]` is used, the spec is such that this block is entered: https://github.com/xonsh/xonsh/blob/master/xonsh/proc.py#L1447

```python
        if not stdout or not safe_readable(stdout):
            # we get here if the process is not threadable or the
            # class is the real Popen
            wait_for_active_job()
            proc.wait()
            self._endtime()
            if self.captured == 'object':
                self.end(tee_output=False)
            raise StopIteration
```

Basically this waits for the process to end (blocks) and makes sure the correct output in set on the pipeline, if needed.  This is the reason why all of the threadable predictors work.  The main iterraw() loop can be totally skipped.  This gives effectively the same behaviour that we had before.  No extra threads are spawned and output is not tee'd.

> I'm still going through the code trying to grok stuff

No worries, there is a lot to grok!  Thanks for diving into it.  Trust me, I don't want to be the only person who understands how all this stuff works together :).

scopatz:
OK, I may have found where 90% of the slowdowns are coming from.  But it is so simple that I am not sure what else it breaks :).  It *might* even be good enough to not need a heuristic threadable predictor...

PR #1889, opened 2016-10-25
Doc fixes

gforsyth:
Fixing some lingering errors in the sphinx build and removing a few defunct bits of the docs.

PR #1891, opened 2016-10-25
Howtofaq

gforsyth:
This adds a general landing point for "How do I ...?" questions that arise frequently.  

For now, it covers styles, importing local modules into xonsh, making xonsh the default shell and a few other tips/fixes.  

There's a bunch of stuff missing, but at least this gives people a place to stick things and should make it easier for new contributors to add on their own tips/fixes.

This builds on #1889 and so shouldn't go in before that.


laerus:
poor travis... merging this, ty @gforsyth !

scopatz:
Travis is definitely overworked and underpaid :)

Issue #1821, opened 2016-10-09
Pygments highlighting doesn't use terminal colour scheme


OJFord:
What's the correct way to change syntax highlighting colour scheme?

![term colours](https://cloud.githubusercontent.com/assets/3946014/19222818/fc7e868e-8e58-11e6-9f76-5e35eec02aa3.png)

Ignoring that some of my colours are broken in the screenshot above, it's clear that the environment variable is a completely different colour to any of them, and barely legible (this is after I changed my background colour specifically so that I could read it at all).

IMO it would be best if pygments was set to use the terminal emulator's scheme by xonsh, at least as a default.

laerus:
You can change the colors in general by setting `$XONSH_COLOR_STYLE`, you can see the available styles by running `xonfig styles`. Atm i don't think you can change only the highlighting colors.

OJFord:
@laerus So it can only be changed among presets? And none of those presets is "use the terminal emulator's settings"?

I find it odd that anyone would want xonsh to be outputting a different colour scheme to that of any tool like `ls`, for example.

Even if you don't agree this should be the default, is there at least a way I can _manually_ set them to be the same as everything else (i.e. add a new entry to that list)?

melund:
@OJFord I think this will work once pygments 2.2 is released. 

Can you try to run with pygments master?

You can see that it should use the ansicolors (terminal colors) but only if it detects that pygments support it. 
https://github.com/xonsh/xonsh/blob/master/xonsh/pyghooks.py#L646

OJFord:
@melund With pygments 2.2.dev20161011, it's fixed in the prompt (presumably anything using the `{COLOR}` formatters), but not the syntax highlighting:

![screen shot 2016-10-11 at 03 29 44](https://cloud.githubusercontent.com/assets/3946014/19256911/27ab5cc8-8f63-11e6-86f1-30bf1e46bfb0.png)
notice that `$FOO` is the same colour as it and the prompt were before starting a new `xonsh`.

scopatz:
Also note that that you have to use the "default" style to get the terminal colors.  

I agree that the color selection and customization interface could benefit from user-facing improvements. Pull requests very welcome!

OJFord:
This is with default set @scopatz:
![screen shot 2016-10-11 at 05 07 10](https://cloud.githubusercontent.com/assets/3946014/19258564/9ccf6816-8f70-11e6-8f42-eaf4ea205f08.png)

[...]

PR #1894, opened 2016-10-26
Add ansi-color codes to the default style

melund:
This PR updates the PTK styles. So now we don't use pygments when the style is set to 'default'. Instead,  we use the `XONSH_BASE_STYLE` dict, which has been updated to match the default pygments style. 

This has the advantage that we can make a version with ansi color code names. 
See this issue: https://github.com/xonsh/xonsh/issues/1821

At the moment, this is a WIP because I think there is a bug in prompt-toolkit which swap the dark and light ansi color names on Windows. https://github.com/jonathanslenders/python-prompt-toolkit/pull/408 

So I think we should wait until that is fixed.

### DIFF
--- a/news/ansi_colors.rst
+++ b/news/ansi_colors.rst
@@ -1,0 +1,15 @@
+**Added:** None
+
+**Changed:**
+
+* The default style will now use the color keywords (#ansired, #ansidarkred) 
+  to set colors that follow the terminal color schemes. Currently, this requires
+  prompt_toolkit master (>1.0.8) and pygments master (2.2) to work correctly. 
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/xonsh/pyghooks.py
+++ b/xonsh/pyghooks.py
@@ -719,7 +719,7 @@ def _default_style():
             Color.INTENSE_YELLOW: '#ansiyellow',
             Color.NO_COLOR: 'noinherit',
             Color.PURPLE: '#ansipurple',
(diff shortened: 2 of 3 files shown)


gforsyth:
Hey @melund -- overall looks good, just a couple of linting errors

melund:
Thanks. I will fix those lint errors..

scopatz:
Thanks for putting this in @melund! I guess we should wait until the next PTK release it sounds like...

melund:
@OJFord The fix have been merged into prompt_toolkit master. So I think this is ready to go. 

Do you mind trying it? Let me know if you think any of the styles here https://github.com/xonsh/xonsh/pull/1894/files#diff-2316634a699f0d18ebe4a69e7d69ea9cR497 should have different colors...

scopatz:
Happy to merge this whenever @melund

OJFord:
Happy to try it later - but what am I trying?

I don't have a Windows machine at the moment, so I can't test that issue, do I still need to run it with PTK master?

melund:
No need for Windows. The color-names was also swapped for VT100 in PTK. So it wasn't a windows only problem. That is fixed in PTK now.

You should just test that the colors follow your terminal colors when you use the 'default' style in xonsh. You need to use PTK master and pygments master for it to work.

Also, if you think that any of the colors choices does not work well in your terminal I would like some suggestions. 

@scopatz. I believe we can merge this now. Then everyone can help test it if they use pygments and PTK master.

laerus:
ty @melund !!

Issue #1397, opened 2016-07-07
Path Completion on Common Prefixes


mitnk:
Say we have the following dirs in pwd:

```
$ ls -F
foo/    foo bar/
```

Then in bash:

```
$ cd f<TAB>
```

would get:

```
$ cd foo
```

While in xonsh, it remains with `cd f` because xonsh currently returns `{"'foo bar'", "foo"}`, so that readline treat them as different prefix (they are different).

I think we should make it same as bash.

Thoughts?

PaulReiber:
My first thought is, I think your example is very good for demonstrating why filenames just plain shouldn't have spaces in them. Spaces already have an important job to do, separating commandline arguments from each other, and when you put them into filenames it introduces a bunch of special cases.  

Let's consider.  You type 'f', then tab.   It completes to "foo".  You type a SPACE character, then tab.

How could the completer possibly know if you meant...
 (a) I'm happy with foo on my commandline, and I'd like to complete '' (empty string) for the next argument (which is a totally valid thing to do), 
or
 (b) Although foo is a totally valid and existing directory, I want the completer to throw in whatever's needed ( a \ before the space?) and complete 'foo bar' for me.

Should an initial tab do (a) then a second tab do (b)?  Special cases.  Yuck.

I can't say I agree regarding making it the same as bash.  I'm for simplifying this instead.

What's in bash is pretty complex - the 'programmable completion' section alone, in the bash manual, is 80+ lines long, and a rough check shows around a dozen end-user-preference variables which affect its operation.  

While I do think xonsh needs powerful tab-based completion... I'm not at all sure that mimicking what's in bash is the right approach.

mitnk:
IMO, at least we need to complete `f` to `foo`, what next up to the user. This is more natural for any shell I think (because they are with the same prefix, any shell doesn't complete like this?).

You're right, xonsh speak a different language, even `cd foo\ bar` don't work here (yet). The thing is I don't know whether this is a `will-support` or a `won't-support`. If latter, xonsh may need to put a word into the docs to tell users/devs the philosophy behind and probably a `won't-fix` list.

I think it's not about complex or simple, it's about what's the right thing. For me, 1) completes `f` to `foo` is the right thing, 2) support `foo\ bar` or not is choosable, up to xonsh own philosophy.

mitnk:
[...]

PR #1897, opened 2016-10-26
Make readline completes to max prefix

mitnk:
Trying address #1397. May not be the right way, but open it here for discussions.

### DIFF
--- a/news/quotes-paths.rst
+++ b/news/quotes-paths.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:**
+
+* Quote all paths when completion if any of the paths needs be quoted,
+  so that bash can automatically complete to the max prefix of the paths.
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/xonsh/completers/path.py
+++ b/xonsh/completers/path.py
@@ -135,8 +135,10 @@ def _quote_paths(paths, start, end):
     orig_start = start
     orig_end = end
     # quote on all or none, to make readline completes to max prefix
-    need_quotes = (any([re.search(PATTERN_NEED_QUOTES, x) for x in paths]) or
-        any([backslash in x and slash != backslash for x in paths]))
+    need_quotes = any(
+        re.search(PATTERN_NEED_QUOTES, x) or
+        (backslash in x and slash != backslash)
+        for x in paths)
 
     for s in paths:
         start = orig_start
--- a/xonsh/completers/path.py
+++ b/xonsh/completers/path.py
@@ -134,12 +134,14 @@ def _quote_paths(paths, start, end):
     slash = xt.get_sep()
     orig_start = start
     orig_end = end
+    # quote on all or none, to make readline completes to max prefix
+    need_quotes = (any([re.search(PATTERN_NEED_QUOTES, x) for x in paths]) or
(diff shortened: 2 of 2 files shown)


scopatz:
Thanks for putting this in @mitnk! It seems like this is failing on linux for some reason.  Is this related to #1899 at all?

gforsyth:
Just a flake error

FLAKE8-check(ignoring E122 E402) _______________________
/home/ubuntu/xonsh/xonsh/completers/path.py:139:9: E128 continuation line under-indented for visual indent

mitnk:
@scopatz I'll check the failed cases; no, it is not related with #1899 which is fixing readline side issue, and this PR is for the procedures of computing paths in xonsh before hand out to readline.

@gforsyth Thanks. I'll check the lint stuff, when we feel the direction of this PR is ok.

mitnk:
Putting more details here

say we have two dirs:
```
$ ls -F
foo/     foo bar/
```
Before this commit:
```
$ cd f<TAB>
(Nothing happened)
$ cd f<TAB><TAB>
$ cd f    # <---- no further completes 
'foo bar/'  foo/
````=
After this commit:
```
$ cd f<TAB>
$ cd 'foo  # <---- completes to 'foo by readline

$ cd 'foo<TAB><TAB>
'foo bar/'  'foo/'
```

**Comment by scopatz** [comment]
I think @adqm and/or @gforsyth should handle this one, since my mind is pretty far from completions right now :)

**Comment by adqm** [comment]
Haven't had a chance to test this, but how does it work if we are already completing something with a starting quote other than the default single quote `'`?

E.g., if we do `r"""something with spa<TAB>`, will everything get the right `r"""` as start and `"""` as end, to make everything have the same prefix?

**Comment by mitnk** [comment]
@adqm it seems so:

```
$ cd r"""f<Tab>
$ cd r"""foo<Tab><Tab>
r"""foo bar/"""  r"""foo/"""

$ cd r'foo<Tab><Tab>
r'foo bar/'  r'foo/'

$ cd r'''foo<Tab><Tab>
r'''foo bar/'''  r'''foo/'''

$ cd 'foo<Tab><Tab>
'foo bar/'  'foo/'
$ cd "foo<Tab><Tab>
"foo bar/"  "foo/"
```

**Comment by adqm** [comment]
I am totally on board with this; makes good sense to me and makes completion a lot more usable in readline.  I haven't tried with PTK, but I assume this is a win there as well, or at least a break-even.

Not sure which way is actually faster, but the `needs_quotes` line could be refactored to:
```
need_quotes = any(re.search(PATTERN_NEED_QUOTES, x) or (backslash in x and slash != backslash) for x in paths)
```

Definitely getting rid of the `[]` in those expressions will speed things up (making them generator expressions instead of list comprehensions).  It's also likely that having just the one generator expression is going to be faster, but I'm not 100% on that.

mitnk:
@adqm agreed on the suggestions, thanks. new commit added.

gforsyth:
Hey @mitnk -- this seems ready to go.  Could you add an entry in the `news` folder?  Instructions are available here: https://github.com/xonsh/xonsh/blob/master/CONTRIBUTING.rst#changelog

mitnk:
@gforsyth added, thanks.

gforsyth:
Great.  Thanks again, @mitnk!

Oddly, this seems to have no effect on PTK, but it's not a regression in any case and it's a big improvement on the readline side.

PR #1896, opened 2016-10-26
fix #1886 - readline completer with space

mitnk:
As titled for #1886

PR #1899, opened 2016-10-26
additional fixes for tab completion with spaces in readline

adqm:
I think #1896 fixed some things partially w.r.t. tab completion in readline with spaces, but it seemed to cause other issues for me when:
- the first completion resulted in a fully-quoted directory, and
- I hit tab again to continue completing

I believe this patch should fix that behavior.

### DIFF
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -25,7 +25,7 @@ from xonsh.lazyasd import LazyObject
 from xonsh.base_shell import BaseShell
 from xonsh.ansi_colors import ansi_partial_color_format, ansi_color_style_names, ansi_color_style
 from xonsh.prompt.base import partial_format_prompt, multiline_prompt
-from xonsh.tools import print_exception
+from xonsh.tools import print_exception, check_for_partial_string
 from xonsh.platform import ON_WINDOWS, ON_CYGWIN, ON_DARWIN
 from xonsh.lazyimps import pygments, pyghooks
 
@@ -258,18 +258,25 @@ class ReadlineShell(BaseShell, cmd.Cmd):
     def completedefault(self, text, line, begidx, endidx):
         """Implements tab-completion for text."""
         rl_completion_suppress_append()  # this needs to be called each time
-        _rebind_case_sensitive_completions()
-
-        line = builtins.aliases.expand_alias(line)
-        mline = line.rpartition(' ')[2]
-        offs = len(mline) - len(text)
+        offset = 0
+        _s, _e, _q = check_for_partial_string(line)
+        if _s is not None and _e is None:
+            # we are in a partial string; we should complete based
+            # on this part now
+            offset = begidx - _s
+            begidx = _s
+        elif _e == endidx:
+            offset = begidx - _s
+            begidx = _s
+            endidx -= 1
         if self.completer is None:
             x = []
         else:
-            x = [(i[offs:] if " " in i[:-1] else i)
-                 for i in self.completer.complete(text, line,
-                                                  begidx, endidx,
-                                                  ctx=self.ctx)[0]]
+            comps = self.completer.complete(text, line,
+                                            begidx, endidx,
+                                            ctx=self.ctx)
+            x = [(i[offset:] if " " in i[:-1] else i)
+                 for i in comps[0]]
         return x
 
     # tab complete on first index too
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -269,6 +269,12 @@ class ReadlineShell(BaseShell, cmd.Cmd):
             offset = begidx - _s
             begidx = _s
             endidx -= 1
+        elif (_s is not None and
+                _e is not None and
+                not any(i == ' ' for i in line[_e:begidx])):
+            line = line[:_e] + line[_e+1:]
+            offset = begidx - _s
+            begidx = _s
         if self.completer is None:
             x = []
         else:


gforsyth:
Thanks for putting this in @adqm.  Sorry for the last patch breaking stuff on your end.  

There's a bit of weird behavior here, if I do

```console
mkdir -p '/tmp/path with spaces/ok'
mkdir -p '/tmp/path with spaces/pok'
```

`cd`ing to either of those dirs, if I complete as far as `/tmp/path with spaces/` and hit tab twice, the trailing quote disappears and I can complete from the listed options without issue.

```console
gil@bad_cat ~ $ cd '/tmp/path with spaces/
spaces/ok/'   spaces/pok/'  
gil@bad_cat ~ $ cd '/tmp/path with spaces/pok/'
```

but if I already know I want to go into, say, `pok`, and I do

```console
gil@bad_cat ~ $ cd 'tmp/path with spaces/'p<TAB>
gil@bad_cat ~ $ cd '/tmp/path with '/tmp/path with spaces/pok/'
```

adqm:
> Sorry for the last patch breaking stuff on your end. 

It's actually always been broken, just in a different way...

> but if I already know I want to go into, say, pok, and I do

Ah, I see.  I can try to work on that.  This whole thing is a hack, as readline's completion system specifically delimits based on spaces.  But I think we can get something to work that fixes all of these issues.

scopatz:
This page might help: https://robots.thoughtbot.com/tab-completion-in-gnu-readline#adding-quoting-support, if you haven't seen it already.

adqm:
The hard part there, @scopatz, is things like the `r'''` that aren't likely to work with anything built in to readline...

scopatz:
Yeah, readline can be tricky that way.

adqm:
I just pushed another commit that should (I think) account for the case you mentioned, @gforsyth.  Do you mind giving it a try?

mitnk:
I tried the cases from @gforsyth , the latest commit fixed the issue.

Thanks @adqm for the extra fixes in this PR!

gforsyth:
Hey @adqm -- almost resolved, I think it's off by one character.  Same dir structure:

```console
/tmp/path with spaces/ok
/tmp/path with spaces/pok
```

```console
ls '/tmp/path with spaces/'po<TAB>
ls '/tmp/path with spaces/ok'
ls '/tmp/path with spaces/'Gpo<TAB>'
ls '/tmp/path with spaces/pok'
```

mitnk:
Hi @adqm I try simpler way to fix the issue:

```
diff --git a/xonsh/readline_shell.py b/xonsh/readline_shell.py
index 45370bf..1fa330b 100644
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -25,7 +25,7 @@ from xonsh.lazyasd import LazyObject
 from xonsh.base_shell import BaseShell
 from xonsh.ansi_colors import ansi_partial_color_format, ansi_color_style_names, ansi_color_style
-from xonsh.tools import print_exception
+from xonsh.tools import print_exception, check_for_partial_string
 from xonsh.platform import ON_WINDOWS, ON_CYGWIN, ON_DARWIN
 from xonsh.lazyimps import pygments, pyghooks

@@ -255,19 +255,21 @@ class ReadlineShell(BaseShell, cmd.Cmd):
         """Overridden to no-op."""
         return '', line, line

-    def completedefault(self, text, line, begidx, endidx):
+    def completedefault(self, prefix, line, begidx, endidx):
         """Implements tab-completion for text."""
         rl_completion_suppress_append()  # this needs to be called each time
         _rebind_case_sensitive_completions()
-
-        line = builtins.aliases.expand_alias(line)
-        mline = line.rpartition(' ')[2]
-        offs = len(mline) - len(text)
+        _s, _e, _q = check_for_partial_string(line)
+        if _s is not None:
+            mline = line[_s:]
+        else:
+            mline = line.rpartition(' ')[2]
+        offs = len(mline) - len(prefix)
         if self.completer is None:
             x = []
         else:
             x = [(i[offs:] if " " in i[:-1] else i)
-                 for i in self.completer.complete(text, line,
+                 for i in self.completer.complete(prefix, line,
                                                   begidx, endidx,
                                                   ctx=self.ctx)[0]]
         return x
```

It fixes the follow issue on master introduced by me (#1896) and the cases from @gforsyth:

```
# set up
$ mkdir -p '/tmp/path with spaces/ok'
$ mkdir -p '/tmp/path with spaces/pok'
$ mkdir -p '/tmp/foo'
$ mkdir -p '/tmp/foo bar'

# cases (breaks in current master)
$ ls '/tmp/foo b<TAB>
$ cd  '/tmp/foo b<TAB>
$ ls '/tmp/path with spaces/'po<TAB>
$ ls '/tmp/path with spaces/'Gpo<TAB>'  # nothing should happen here, right? @gforsyth
```

mitnk:
And based on the changes of my previous diff, it seems this line:
```
x = [(i[offs:] if " " in i[:-1] else i)
     for i in ...
```
can be changed to
```
x = [i[offs:]
     for i in ...
```
without any bad effects. (do I miss anything here?)

adqm:
Yes, @mitnk, that is better in all dimensions!  Simpler code, and seems (to me, at least) to hit all the cases.

Would you mind submitting a pull request with your version?  Then I can close this one.

mitnk:
Thanks @adqm, I added the changes inside #1951 with a bit more changes. You could help review it if you got time :)

PR #1903, opened 2016-10-27
Windows non-threadable fix

scopatz:
This should address the issues with less, vim, etc and other non-threadable commands on Windows. @melund  and @BYK, would you mind taking a look please?  

This issue with `git diff` and other similar threadable commands are more deeply seeded with alternate mode. I am looking in to them now.


scopatz:
PS I don't think this needs a new entry

melund:
Wuhoo... It fixes the less problem. Merging

PR #1906, opened 2016-10-27
more entries in threadable predictors

gforsyth:
I know this is just a bandaid, but for now we can cover a lot of commonly used programs until we come up with a robust way to predict these things.  Also, `pacman` and `macvim` will let us close out at least two issues.


scopatz:
Thanks @gforsyth!

astronouth7303:
Given that threadability hasn't been released yet, I'm not sure this needed the news item.

PR #1908, opened 2016-10-27
Fix SubprocSpec repr

gforsyth:
Quick fix for an annoyance I ran into while trying to debug some alias stuff


scopatz:
Whoops! Thanks!

PR #1910, opened 2016-10-29
heal completers pip

laerus:
worst offender in landscape.. mostly just disabled pylint errors


laerus:
i don't think that need a news entry.

scopatz:
Thanks a ton @laerus!

PR #1911, opened 2016-10-29
Windows Capture & Alt Mode 

scopatz:
This is a fix for allowing Windows to both stream subprocs normally and to, with the exception of some edge cases, capture the output.

To implement this required digging into the Windows API, wrapping some functions and structs in winutils, and then exposing this as a non-blocking parallel reader. 

I don't think this fixes all of the issues, but it does let you do `git diff` and `x = ![ls]; print(x.output)` and have everything work.  I have only tested this on the anaconda console. I have no idea if it will work in other places (like ConEmu).

Programs are always executed and output is always displayed.  The exceptions are for caputiring, which are detailed in the PR.  But basically, if we are at the end of the screen buffer, we may not be able to capture all of the output until a resize is complete.  Unfortunately, this fix for this rare occurrence is to set the cursor position, and doing so triggers a bug (in PTK _I think_) that renders the next prompt unusable. So the tradeoff is to always have a working prompt, but maybe miss some data at the end of the screen buffer occasionally.  I think this is reasonable for the time being.

Sorry this is so huge.


melund:
Awesome. It seems to solve all the issues I had. I will test it more thoroughly.

scopatz:
Woohoo! Thanks @melund!

xonsh/proc.py
```diff
@@ -234,8 +216,256 @@ def __init__(self, fd, buffer=None, chunksize=1024):
...
+    #        x = new_offset%cols
+    #        y = new_offset//cols
+    #        continue
+    #
+    # So this method is imperfect and only works as long as the sceen has
+    # room to expand to.  Thus the trick here is to expand the screen size
+    # when we get close enough to the end of the screen. There remain some
+    # async issues related to not being able to set the cursor position.
+    # but they just affect the alignment / capture of the output of the
+    # first command run after a screen resize.
```
melund: Arghh.. Crazy. I knew windows behaved this way, but I must say I ca hardly follow the code below.

melund:
I have read the code, but not to level where I really understand it. I haven't got time to do proper review the next weeks time. But I am all for merging this if other Windows agree. It is a huge improvement. 

Can @jonathanslenders advise on the possible PTK, or do @scopatz have an idea how to solve it?

xonsh/proc.py
scopatz: Yeah, this code definitely evolved.... If only they had used newlines, thins whole thing would be like 20 lines.

scopatz:
> I have read the code, but not to level where I really understand it. I haven't got time to do proper review the next weeks time. But I am all for merging this if other Windows agree. It is a huge improvement.

Awesome, maybe @BYK or @astronouth7303 can chime in...

> ICan @jonathanslenders advise on the possible PTK, or do @scopatz have an idea how to solve it?

I don't know enough about the PTK code base to say.  I dove into the code base but I could find if there were any places where the cursor position is set inside of PTK.  So if @jonathanslenders could confirm or deny that, it would be great!

melund:
Ok. I got impatient. I will just merge it in. It solves a lot of issues, and It only affect Windows anyways. It is best tested in the wild.

jonathanslenders:
Hi @scopatz,
For Windows, the cursor position is set here: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/prompt_toolkit/terminal/win32_output.py#L236
So far, I did not read through this pull request, so not sure whether this is related.

FYI: regarding Windows+prompt_toolkit. The rendering needs to be rewritten at some point to be more performant. See: https://github.com/jonathanslenders/python-prompt-toolkit/issues/241

scopatz:
Thanks @melund!

Thanks @jonathanslenders, I did see that function.  My question is more, "where is its call site?" I didn't have time to really dig into the PTK code to find out.  (Also that win32, vt100 abstraction is incredible)

BYK:
Sorry for the radio silence. I really wanna test this and I hope to provide some feedback (both the workings and on the code) on Thursday :)

scopatz:
No worries, thanks @BYK! Looking forward to your comments

PR #1914, opened 2016-10-30
Hack a rel=canonical in the templates.

Carreau:
We do not want non-canonical versions of xonsh docs to be indexed. OR at
least they should point to the canonical version.

Technically we should even try to have the stable version under a
/stable/ url (to leave us with some freedom of changing things) So that
later we can replace stable with X.y and have a version switcher.

-- 
From experience with the IPython docs.

### DIFF
--- a/docs/_templates_overwrite/layout.html
+++ b/docs/_templates_overwrite/layout.html
@@ -1,0 +1,6 @@
+{% extends '!layout.html' %}
+
+{% block extrahead %}
+    {{ super() }}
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks a ton @Carreau! This was definitely needed...

Carreau:
Thanks @scopatz ! It's good to have some time free and to start contributing back !

scopatz:
Sweet!

PR #1915, opened 2016-10-31
[WIP] Version selector for documentation.

Carreau:
That's still highly in progress, for now poking around to see how it could look like. 

This basically setup a marker for the version-switcher in the sidebar and search/replace it script.
I try to be smart enough and if the page does not exists in another version of the docs, it redirect to `index.html`. The main point of research/replace being that you can update "old" docs without rebuilding. 

We will need at some point to move the documentation to a subfolder like /stable/, maybe set a meta-redirect, but this is extremely frowned upon and not good for SEO.


scopatz:
This would be awesome to have! Just let us know when this is no longer WIP or if you need us to test it out locally.

Carreau:
Meh... I'm unsure it should works but doctr does trigger on PR...

Carreau:
I'll see if I can patch doctr to actually have a dry-run and not push on PR instead of Just bailing out.

.travis.yml
```diff
@@ -46,11 +46,16 @@ install:
 
 script:
   - if [[ $BUILD_DOCS = true ]]; then
+      echo "buildign docs"
```
Carreau: Yeah.. just debug... trying stuff. I'll rbase clean later once I'm happy. I might do a few change to doctr first.

Carreau:
I think I know have an idea of all the changes that are needed in doctr to deploy that. It will take me a couple of days to finish implementing that and maybe get a release of doctr.

scopatz:
I am really looking forward to it!

scopatz:
Hi @Carreau - I am pretty sure more releases of doctr have come out :) Is this more or less ready to go?

Carreau:
I'll hack on that on the week of March 6, we're doing a docathon in BIDS, eScience Institute and NY.

scopatz:
Sounds like a plan and like fun! Thanks @Carreau

Carreau:
Note to myself: https://pypi.python.org/pypi/sphinxcontrib-versioning

gforsyth:
Hey @Carreau -- should we leave this PR open?  Do you think you'll have time at some point to shepherd this through?

Carreau:
let's close it for now, I can come back to it later and reopen.

Thanks !

PR #1916, opened 2016-10-31
Fix "Fork me on GitHub" ribbons.

Carreau:
The current ribbon just don't react on click on some platform for
some reason. Mainly because the Image is in the corner, but not the
link.

This fixes it

Though it still look ugly on hi-dpi screen

One might want to consider a pure-css method like
https://simonwhitaker.github.io/github-fork-ribbon-css/

-- 

The hover style is not inline as ~ no browser support an inline hover style.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -266,4 +266,10 @@ Helpful Links
 
 .. raw:: html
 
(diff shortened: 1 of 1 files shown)


gforsyth:
Sweet!  Thanks, @Carreau!

PR #1918, opened 2016-10-31
Pure-css "Fork me on GitHub" ribbon.

Carreau:
It's actually a bit more css that I thought it would be. But it looks nice on hi-dpi screen, and don't need to touch Photoshop/Gimp to be modified.
screen shot 2016-10-31 at 07 22 24
screen shot 2016-10-31 at 07 22 14

Also fix a bug where the ribbon would "jump" when clicked.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -266,10 +266,63 @@ Helpful Links
 
 .. raw:: html
 
-    <a href="https://github.com/xonsh/xonsh" style="position: absolute; top: 0; right: 0; border: 0;" id="forkmeongithub"><img src="https://camo.githubusercontent.com/52760788cde945287fbb584134c4cbc2bc36f904/68747470733a2f2f73332e616d617a6f6e6177732e636f6d2f6769746875622f726962626f6e732f666f726b6d655f72696768745f77686974655f6666666666662e706e67" alt="Fork me on GitHub" data-canonical-src="https://s3.amazonaws.com/github/ribbons/forkme_right_white_ffffff.png"></a>
+    <a href="https://github.com/xonsh/xonsh" class='github-fork-ribbon' title='Fork me on GitHub'>Fork me on GitHub</a>
 
     <style>
-    #forkmeongithub:hover {
-        background: none;
+    /*!
+     * Adapted from
+     * "Fork me on GitHub" CSS ribbon v0.2.0 | MIT License
+     * https://github.com/simonwhitaker/github-fork-ribbon-css
+     */
+
+    .github-fork-ribbon, .github-fork-ribbon:hover, .github-fork-ribbon:hover:active {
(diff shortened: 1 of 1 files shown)


Carreau:
Also if you feel it's too complicated I'm fine not merging.

Carreau:
OSX * ( Chrome, Firefox, Safari) approved: 

screen shot 2016-10-31 at 09 25 08

gforsyth:
Looks good on Firefox and Chromium on Linux.

gforsyth:
Thank you for making us beautiful, @Carreau

astronouth7303:
Moz, blink, and webkit should cover it. It would be nice to get IE/Edge in there, but given the audience, I don't consider it a high enough priority to do it myself.

Carreau:
> Thank you for making us beautiful, @Carreau

:-) 

Also one thing I was thinking about is "hiding" a qr-code in the xonsh logo that point to https://xon.sh.

gforsyth:
> Also one thing I was thinking about is "hiding" a qr-code in the xonsh logo that point to https://xon.sh.

That would be awesome.  Although we should probably get https working first...

gforsyth:
In fact, I think the QR-code thing needs to happen before we make stickers.  Because clearly stickers are happening, but stickers with a hidden qr code are better

Carreau:
I'm guessing you know about qart-code that use data-correction to embed images in qrcodes ?
qart1

gforsyth:
I did not know, but now I'm not going to get anything done today.

Carreau:
> I did not know, but now I'm not going to get anything done today.

Don't even try: 

http://www.thonky.com/qr-code-tutorial/introduction
http://speedata.github.io/luaqrcode/docs/qrencode.html#section-5
https://github.com/sylnsfar/qrcode/

And don't even bother reading about ECC codes on wikipedia and Low Density Parity Check.

I'm +1 on stickers, I want to open a global sticker shop for the scientific community were we can Just order "Jet kills", "use viridis"... and not only stickers, but tshirts, but mugs... etc. 

Have to think of long term funding right ?

astronouth7303:
> That would be awesome. Although we should probably get https working first...

That would be really fantastic. But I'm not sure github is ready to deal with that.

gforsyth:
Would this kind of thing work?  (Using Cloudflare free services)
https://developer.ubuntu.com/en/blog/2016/02/17/how-host-your-static-site-https-github-pages-and-cloudflare/

Carreau:
Yes. ipython.org and Jupyter.org use fastly and do the same.  There is a tiny bit of maintenance overhead though.

scopatz:
So glad I didn't read this thread until after work :).  Also I now need "Jet Kills" on everything...

PR #1919, opened 2016-10-31
add ranger to thread list, clean up news entries

gforsyth:
`ranger` should've been added a while ago.  I also collapsed the various
different news items about default-not-backgroundable programs into a
single file so it will make a little more sense at the next release.


laerus:
thank you @gforsyth for cleaning things up!!

PR #1923, opened 2016-11-01
Fix swapped background colors on PTK completions.

melund:
I realized that I swapped the background colors for PTK completions. 

This doesn't need a news item as it is just a fix to a previous PR.


scopatz:
LGTM @melund, thanks!

scopatz:
@laerus - wasn't travis still building?

laerus:
yep but this was a minor change and travis.. is very slow  :(

scopatz:
True, travis is very slow :(

PR #1924, opened 2016-11-02
added commit date in xonfig

mitnk:
I feel this info would be helpful for xonsh developers. Does this make sense to you?

```
$ xonfig
| xonsh            | 0.4.7.dev1734       |
| Git SHA          | f78a0e1             |
| Commit Date      | Nov 2 01:44:25 2016 |
| Python           | 3.5.1               |
| prompt toolkit   | 1.0.0               |
| shell type       | readline            |
| pygments         | 2.1.3               |
| distro           | debian              |
```

### DIFF
--- a/xonsh/platform.py
+++ b/xonsh/platform.py
@@ -172,6 +172,7 @@ def pathbasename(p):
 
 @functools.lru_cache(1)
 def githash():
+    """Returns a tuple contains two strings: the hash and the date."""
     install_base = os.path.dirname(__file__)
     sha = None
     date_ = None
--- a/xonsh/platform.py
+++ b/xonsh/platform.py
@@ -173,14 +173,14 @@ def pathbasename(p):
 @functools.lru_cache(1)
 def githash():
     install_base = os.path.dirname(__file__)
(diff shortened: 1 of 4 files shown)


xonsh/platform.py
```diff
@@ -173,14 +173,14 @@ def pathbasename(p):
 @functools.lru_cache(1)
 def githash():
```
scopatz: While you are here, would you mind c please adding a docstring that mentions this now returns a tuple (more than just the hash), please?

scopatz:
Thanks a ton for putting this in @mitnk! This will be great to have, just the one small suggestion.

xonsh/platform.py
mitnk: sure, and maybe I could also change this function name to `get_git_version_info()`?

mitnk: maybe not ;) docstring added.

mitnk:
Fixed an issue in last commit.

scopatz:
Thanks @mitnk! Great to have this!

PR #1926, opened 2016-11-02
prompt mechanism refactor

laerus:
new PromptFormatter class

renamed `FORMATTER_DICT` to `PROMPT_FIELDS`

some caching when building the prompt (we can do better, some details on #1564)

(some shell refactoring on this matter would prolly come after this)

### DIFF
--- a/xonsh/prompt/base.py
+++ b/xonsh/prompt/base.py
@@ -34,45 +34,45 @@ class PromptFormatter:
     def __init__(self):
         self.cache = {}
 
-    def __call__(self, template=DEFAULT_PROMPT, formatter_dict=None):
+    def __call__(self, template=DEFAULT_PROMPT, fields=None):
         """Formats a xonsh prompt template string."""
+        if fields is None:
+            self.fields = builtins.__xonsh_env__.get('FORMATTER_DICT', FORMATTER_DICT)
+        else:
+            self.fields = fields
         try:
-            prompt = self._format_prompt(template=template,
-                                         formatter_dict=formatter_dict)
+            prompt = self._format_prompt(template=template)
+            # keep cache only during building prompt
             self.cache.clear()
             return prompt
         except Exception:
             return _failover_template_format(template)
 
-    def _format_prompt(self, template=DEFAULT_PROMPT, formatter_dict=None):
+    def _format_prompt(self, template=DEFAULT_PROMPT):
         template = template() if callable(template) else template
         toks = []
-        if formatter_dict is None:
-            fmtter = builtins.__xonsh_env__.get('FORMATTER_DICT', FORMATTER_DICT)
-        else:
-            fmtter = formatter_dict
         for literal, field, spec, conv in _FORMATTER.parse(template):
             toks.append(literal)
-            field = self._format_field(field, spec, conv, fmtter)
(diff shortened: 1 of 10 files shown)


astronouth7303:
Notably, this does not change the prompt syntax, as we were discussing elsewhere.

scopatz:
Thanks for putting this in @laerus! Because lot of people use `$FORMATTER_DICT`, I think we need to alias it to the new name and issues a warning about the deprecation when the old name is used for at least one release.  Would you be willing to add that to this? I know it is a pain.... We can do that in a different PR, too, if you want.

laerus:
good point! i'll get to it on this PR

laerus:
this was the easiest way i could find to add the warning, tried with the warnings modules but there is some strange noise.
```
/home/laerus/src/xonsh/xonsh/environ.py:849: UserWarning: FORMATTER_DICT is an alias of PROMPT_FIELDS and will be removed in the next release
  warnings.warn('FORMATTER_DICT is an alias of '
```

and if i change it to `PendingDeprecationWarning` the noise is
```
/home/laerus/.local/bin/xonsh:1: PendingDeprecationWarning: FORMATTER_DICT is an alias of PROMPT_FIELDS and will be removed in the next release
  #!/usr/bin/python3 -u

```

laerus:
i think we can even refactor this to have a function that we can use in case of envvar deprecations in the future.

xonsh/prompt/base.py
```diff
@@ -24,8 +24,70 @@
 from xonsh.prompt.gitstatus import gitstatus_prompt
 
 
+@xt.lazyobject
+def DEFAULT_PROMPT():
+    return default_prompt()
+
+
+class PromptFormatter:
+    """Class used by base_shell"""
```
scopatz: Could this be more descriptive, please?

xonsh/prompt/base.py
```diff
@@ -24,8 +24,70 @@
...
+        self.cache = {}
+
+    def __call__(self, template=DEFAULT_PROMPT, fields=None):
+        """Formats a xonsh prompt template string."""
+        if fields is None:
+            self.fields = builtins.__xonsh_env__.get('PROMPT_FIELDS', PROMPT_FIELDS)
+        else:
+            self.fields = fields
+        try:
+            prompt = self._format_prompt(template=template)
```
scopatz: I think only this line should be in the try-block.  Please move the other two lines to after the except.

scopatz:
Overall, this looks great, @laerus! Just a couple of small cleanups.

xonsh/prompt/base.py
```diff
@@ -24,8 +24,70 @@
...
+            return
+        elif field.startswith('$'):
+            val = builtins.__xonsh_env__[field[1:]]
+            return _format_value(val, spec, conv)
+        elif field in self.fields:
+            val = self._get_field_value(field)
+            return _format_value(val, spec, conv)
+        else:
+            # color or unkown field, return as is
+            return '{{{}}}'.format(field)
```
laerus: i thought string concatenation is slower in python

laerus: concatenation is way faster is this case..

PR #1927, opened 2016-11-03
lazyasd update from repo

scopatz:
Turns out xonsh wasn't even at the latest laziness. Fitting(?)


scopatz:
PS, I don't think this needs news.

PR #1929, opened 2016-11-03
Patch setup:install_jupyter_root to respect the root param

chronitis:
`install_jupyter_hook(prefix, root)` doesn't honour the `root` parameter (since `KernelSpecManager.install_kernel_spec` has only a `prefix` option). When `root` is set (eg, distribution builds), this results in the jupyter kernel being installed in the wrong place.

### DIFF
--- a/news/jupyter-install-root.rst
+++ b/news/jupyter-install-root.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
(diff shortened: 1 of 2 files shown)


gforsyth:
Thanks for putting this in, @chronitis !  This looks good to go, but could you add an entry in the `news` folder?  Instructions are available here: https://github.com/xonsh/xonsh/blob/master/CONTRIBUTING.rst#changelog

scopatz:
Thanks a ton!

PR #1932, opened 2016-11-05
1kb

scopatz:
This PR provides about a 75% reduction in speed from master, and doesn't *seem* to break anything.

Using @laerus's script (slightly modified)

**prof.xsh**

```
import time

cd /usr/bin
outs = set()
start = time.time()
for i in range(100):
    outs.add($(ls))
print('command took: {} seconds'.format(time.time() - start))
print('difference in outs:', len(outs) - 1)
```

With this I am seeing:

|                     | command took [s] | difference in outs |
|---------------------|------------------|--------------------|
| 3545358a4b (master) | 20.03104758      | 17                 |
| 495d04f             | 4.724117279      | 0                  |
| 5aec225             | 4.678267479      | 0                  |

The problem before hand was that before hand we were reading from the pipe byte-by-byte.  This generated a lot of overhead in system calls.  Performance seems to level off above 100 byte reads on my machine, but I think that 1 kb is a typical amount, so I am going with that here.  

Also added some refactor to reduce duplicated code.

### DIFF
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -590,38 +590,37 @@ class PopenThread(threading.Thread):
         return getattr(self, name)
 
     def _read_write(self, reader, writer, stdbuf):
-        """Read from a buffer and write into memory or back down to
-        the standard buffer, line-by-line, as approriate. Returns the number of
-        lines read.
+        """Reads a chunk of bytes from a buffer and write into memory or back
+        down to the standard buffer, as approriate. Returns the number of
+        successful reads.
         """
         if reader is None:
             return 0
         i = -1
-        #for i, line in enumerate(iter(reader.readline, b'')):
-        for i, line in enumerate(iter(reader.read_queue, b'')):
-            self._alt_mode_switch(line, writer, stdbuf)
+        for i, chunk in enumerate(iter(reader.read_queue, b'')):
+            self._alt_mode_switch(chunk, writer, stdbuf)
         if i >= 0:
             writer.flush()
             stdbuf.flush()
         return i + 1
 
-    def _alt_mode_switch(self, line, membuf, stdbuf):
+    def _alt_mode_switch(self, chunk, membuf, stdbuf):
         """Enables recursively switching between normal capturing mode
         and 'alt' mode, which passes through values to the standard
         buffer. Pagers, text editors, curses applications, etc. use
         alternate mode.
         """
-        i, flag = findfirst(line, ALTERNATE_MODE_FLAGS)
+        i, flag = findfirst(chunk, ALTERNATE_MODE_FLAGS)
         if flag is None:
-            self._alt_mode_writer(line, membuf, stdbuf)
+            self._alt_mode_writer(chunk, membuf, stdbuf)
         else:
             # This code is executed when the child process switches the
             # terminal into or out of alternate mode. The line below assumes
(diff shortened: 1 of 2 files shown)


melund:
Nice. Seems to do the trick on Windows as well. 

I would say it is an increase in speed. But I don't know how a physicist defines the speed of software :)

This is with running `ls` on `c:/windows/system32`, but the ratio is the same. 

|                     | command took [s] | difference in outs |
|---------------------|------------------|--------------------|
| 3545358  (master) | 93.802318      | 3                 |
| b117853       |  20.3957757     | 0                  |

xonsh/proc.py
```diff
@@ -68,51 +68,15 @@ def RE_VT100_ESCAPE():
...
-            reader.closed = True
-            break
-
-
-class NonBlockingFDReader:
-    """A class for reading characters from a file descriptor on a background
-    thread. This has the advantages that the calling thread can close the
-    file and that the reading does not block the calling thread.
-    """
+class QueueReader:
```
melund: I see you removed the `__init__` function. It seems a bit weird to have a parent class that will only work if the child classes remember to initialize a number of variables. Like `self.timeout`, and `self.closes`.

melund: I think it should have a `__init__` function that sets the class methods that the parent class expects to be present. The child class `NonBlockingFDReader` can then overload them if it wants.

xonsh/proc.py
```diff
@@ -172,6 +136,50 @@ def readable():
         return True
 
 
+def populate_fd_queue(reader, fd, queue):
+    """Reads single characters from a file descriptor into a queue.
```
melund: Should be: `Read 1kb of data from a file descriptor...`

scopatz:
Thanks for the feedback, @melund! I am actually a little surprised that this effects Windows, but I am glad it helps!  I have updated the PR as requested.

melund:
Why would think it wouldn't work on windows? I double checked and those are the results I get. 

How far are we from a 0.5 release with this?

laerus:
woohoo! that really makes a difference ty @scopatz! (Continuing the discussion on #1890) I cannot find the cause that we lose output on some commands, even though `$()` is blocking until the command is finished. any idea about that?

laerus:
@melund i think we should also make backgrounding jobs work before 0.5

scopatz:
> Why would think it wouldn't work on windows? I double checked and those are the results I get.

@melund This was a misunderstanding on my part. I thought the PopenThread class was a little less tightly coupled on Windows than it is. Now I know :)

> How far are we from a 0.5 release with this?

Much closer. I have some further speed improvements, I am trying to implement. Other than this, my wishlist for 0.5 is just v3.6 support working (which shouldn't be that hard) and shell-in-a-box on the website.

laerus:
Tried with 10k iterations on the test script, not a single output was lost! (i'll assume that's the case for slower machines too, but i'll test it on my slow netbook at some point), also the cpu usage got halved on the latest master! yey

backgrounding jobs is still not working for me though

laerus:
@scopatz do you think we can make the magic number 1024 of chunksize an envvar too?

scopatz:
> @scopatz do you think we can make the magic number 1024 of chunksize an envvar too?

We could, I was too lazy :).  Glad to hear that everything is working.

PR #1933, opened 2016-11-05
New PTK menu style

melund:
This updates the PTK styles to bring the scrollbar back. PTK expect it to defined by the styles  `Token.Scrollbar` and `Token.Scrollbar.Button`. 

It also updates the toolbar styles to more neutral lightgray/darkgray color, instead of the previous turquoise/teal colors. This is a of course a matter of taste, but it is more in line with the default colors in prompt toolkit. 

image

Let me know what you think.

### DIFF
--- a/news/ptk_menu.rst
+++ b/news/ptk_menu.rst
@@ -1,0 +1,15 @@
+**Added:** None
+
+**Changed:**
+
+* The style for the prompt toolkit completion menu is now lightgray/darkgray instead of turquoise/teal
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** 
+
+* The scroll bars in the PTK completions menus are back. 
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


melund:
There are also other options:

Black on darkgray
image

Light gray on darkgray:
image

scopatz:
I am definitely for changing this (and even more so for allowing users some way of setting these, but that doesn't have to be here.) I think @gforsyth should probably weigh in though, one way or the other.

melund:
It is not impossible for users to change these although it is not really easy: 

image

It not robust either. Xonsh will crash if we set an invalid color code.

gforsyth:
Hey @melund, this looks great!  I think this is a much better default.  It would be awesome to have a simpler way for users to customize this (in a non-crashy way) but that can definitely be a separate feature.

PR #1934, opened 2016-11-05
Update win color registry file

melund:
This an update to the registry file for setting nicer console colors on windows. 

It is an update to previous unreleased PR, so it doesn't need a news item.


scopatz:
Thanks a ton!

PR #1940, opened 2016-11-08
fixed small regression on linux caused by windows fixes

scopatz:
I don't think this needs a news entry, but basically we shouldn't over-close handles on Linux because it can loose data from buffers.  Windows does not seem to operate this way.


melund:
Lgtm

PR #1936, opened 2016-11-06
get_hg_branch runs 'hg status' first

laerus:
noticed while stracing xonsh that `get_hg_branch` is crawling all the way up to root dir to check if we
are in a repo, this can be slow if we are very deep in a dir tree.

```
$ pwd
/home/laerus/lab/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p
```
part of strace on every prompt update:
```
...
stat("/home/laerus/lab/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/.hg", 0x7ffca79cef30) = -1 ENOENT (No such file or directory)
stat("/home/laerus/lab/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/.hg", 0x7ffca79cef30) = -1 ENOENT (No such file or directory)
stat("/home/laerus/lab/a/b/c/d/e/f/g/h/i/j/k/l/m/n/.hg", 0x7ffca79cef30) = -1 ENOENT (No such file or directory)
stat("/home/laerus/lab/a/b/c/d/e/f/g/h/i/j/k/l/m/.hg", 0x7ffca79cef30) = -1 ENOENT (No such file or directory)
stat("/home/laerus/lab/a/b/c/d/e/f/g/h/i/j/k/l/.hg", 0x7ffca79cef30) = -1 ENOENT (No such file or directory)
stat("/home/laerus/lab/a/b/c/d/e/f/g/h/i/j/k/.hg", 0x7ffca79cef30) = -1 ENOENT (No such file or directory)
stat("/home/laerus/lab/a/b/c/d/e/f/g/h/i/j/.hg", 0x7ffca79cef30) = -1 ENOENT (No such file or directory)
stat("/home/laerus/lab/a/b/c/d/e/f/g/h/i/.hg", 0x7ffca79cef30) = -1 ENOENT (No such file or directory)
stat("/home/laerus/lab/a/b/c/d/e/f/g/h/.hg", 0x7ffca79cef30) = -1 ENOENT (No such file or directory)
stat("/home/laerus/lab/a/b/c/d/e/f/g/.hg", 0x7ffca79cef30) = -1 ENOENT (No such file or directory)
stat("/home/laerus/lab/a/b/c/d/e/f/.hg", 0x7ffca79cef30) = -1 ENOENT (No such file or directory)
stat("/home/laerus/lab/a/b/c/d/e/.hg", 0x7ffca79cef30) = -1 ENOENT (No such file or directory)
[...]

PR #1941, opened 2016-11-08
run git/hg only if we can locate binary

laerus:
this came out after #1936 was merged.


gforsyth:
Looks good to me, @laerus.  

@Siecje -- would you mind giving this PR a try?  I recall you had some mixed git/mercurial thing going on and it would be good to know that this doesn't break your setup.

Siecje:
For Mercurial repos I have to increase ```VC_BRANCH_TIMEOUT``` but other than that it works. 

The branch timeout message is great!

gforsyth:
Awesome.  Thanks for trying it out, @Siecje !

And thanks for putting this in @laerus!

PR #1942, opened 2016-11-09
get_hg_status runs 'hg root'

laerus:
@Siecje can you check this one out pls?

i believe `_get_parent_dir_for` was a little excessive in that you had to stats every dir for `.hg` up to the filesystem root and that causes a slowdown when not in a hg repo.

there is more room for speed improvements, the plan is to run git/hg only once to get both the branch and if the working directory is 'dirty', i'm currently trying out ways for this.

### DIFF
--- a/news/hg_branch.rst
+++ b/news/hg_branch.rst
@@ -4,7 +4,7 @@
 
 **Changed:**
 
-* ``get_hg_branch`` runs ``hg status`` first to check if we are in a repo
+* ``get_hg_branch`` runs ``hg root`` to find root dir and check if we're in repo
 
 **Deprecated:** None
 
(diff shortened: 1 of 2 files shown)


Siecje:
One options might be to use `hg summary` to determine if it is an hg repo and get the branch, bookmark, and repo state.

scopatz:
This looks good to me @laerus, should it be merged?

laerus:
i think this currently conflicts with #1943 and i'm not sure what's the best approach to this yet.

scopatz:
Ok, I defer to you then :)

Siecje:
#1943 Is to be added to these changes, it still uses the ```hg root``` call. 

The biggest thing is the ```strip()``` on the root returned by ```hg root```.

It wasn't working without it for me.

The ```command_cache``` changes were not working. I was able to to navigate to a repo without seeing the branch and dir state in the prompt. 

I fixed the error when ```hg``` was not installed by handling the ```FileNotFoundError``` exception from the ```subprocess``` call.

PR #1943, opened 2016-11-09
Fixing branch and repo state...

Siecje:
Previously you had to run a git or hg command in order for the branch and repo
state to be displayed in the prompt.

### DIFF
--- a/xonsh/prompt/vc_branch.py
+++ b/xonsh/prompt/vc_branch.py
@@ -65,11 +65,14 @@ def get_hg_branch(root=None):
                                        stderr=subprocess.DEVNULL)
     except subprocess.TimeoutExpired:
         return subprocess.TimeoutExpired(['hg'], timeout)
+    except FileNotFoundError:
+        # hg is not found
+        return None
     except subprocess.CalledProcessError:
         # not in repo
         return None
     else:
-        root = xt.decode_bytes(root)
+        root = xt.decode_bytes(root).strip()
     if env.get('VC_HG_SHOW_BRANCH'):
         # get branch name
         branch_path = os.path.sep.join([root, '.hg', 'branch'])
@@ -81,7 +84,7 @@ def get_hg_branch(root=None):
     else:
         branch = ''
     # add bookmark, if we can
-    bookmark_path = os.path.sep.join([root, '.hg', 'bookmarks.current'])
+    bookmark_path = os.path.join(root, '.hg', 'bookmarks.current')
     if os.path.exists(bookmark_path):
         with open(bookmark_path, 'r') as bookmark_file:
             active_bookmark = bookmark_file.read()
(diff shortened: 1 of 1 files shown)


melund:
Will this mean that the `command_cache` is updated when xonsh starts? Can you test what this PR do to the startup time.

Siecje:
Currently your prompt will be different depending if you have recently run git or hg.

This is a PR for the vc branch.

This is is removing checking ```builtins.__xonsh_commands_cache__``` so it does not update ```command_cache``` when ```xonsh``` starts.

This should not impact the startup time.

laerus:
this will fail if we have no git/hg installed though.

melund:
Yes. The problem now is that it doesn't show the branch info until after the `command_cache` has been updated. That is why it starts working when after you have run a 'git/hg' command. Actually, any subproc command would do.

But we don't want the `command_cache` to be updated at startup to keep the startup time low. Maybe we can evaluate the `command_cache` in a separate thread on startup.

Siecje:
@laerus Good catch. I've updated the code to handle when hg is not installed.

@melund This PR doesn't use the ```command_cache```. It does try to run both ```git``` and ```hg``` if you are not in a repo. Using a cache might make ```xonsh``` faster when not in a repo. We can address this if it is an issue.

Another speed increase would be to figure out if it is a git or hg repo once, not in ```current_branch()``` and ```dirty_working_directory()```. 

Both of those function try to get the value for ```git``` and if it is not found get it for ```hg```.

laerus:
We should handle the case when git is not installed aswell, also if we do not indeed have hg/git installed this will keep trying to run them on every prompt update and that is what `lazy_locate_binary` is speeding up.

Siecje:
@laerus I was not able to trigger an error when ```git``` was not installed.

laerus:
hey @siecje can you please add/change the news in `git_branch.rst` `hg_branch.rst` to cover this changes?

Siecje:
This was a pull request for the ```vc``` branch in order for VCS branches and repo state to work for me.

I see you have merged ```vc```. If you don't need these changes then you can close this.

laerus:
I've merged so we can merge this, just needs news tweaking so we dont lose track of whats what. i'll take a closer look when i will have more time.

laerus:
my bad @siecje i thought this was a PR for master, if you want to add news and push to master i will merge asap.

laerus:
@melund apparently `__xonsh_commands_cache__.is_empty` populates the cache when it returns True (strange), so this was always a part of the startup when formatting the first prompt.

melund:
Well spotted @laerus. Maybe this will help speed up windows startup times then

Siecje:
I'm going to close this because I may have been confused with the hg branch not working due to https://github.com/xonsh/xonsh/issues/1948

I'm sure this will come up again if it is still an issue.

PR #1945, opened 2016-11-10
Backtab completion

gforsyth:
As per discussion on gitter, shift+Tab will move backwards in the prompt toolkit completion list.  when the completion buffer isn't active, shift+tab will still insert an indent


scopatz:
3 minutes to write, hours for CI

laerus:
cooool ty @gforsyth !

PR #1938, opened 2016-11-07
Sleepless

scopatz:
Here are some more performance improvements for running subprocesses. Using the same example script from #1932, I am now seeing,

|                     | command took [s] | difference in outs |
|---------------------|------------------|--------------------|
| 3545358a4b (master) | 20.03104758      | 17                 |
| 5aec225 (1kb)       | 4.678267479      | 0                  |
| 8603909be           | 1.552735806      | 0                  |

The two main points here are that:

1. the PopenThread now does a proper blocking read after the process has ended, which should prevent losing data, and 
2. Many sleep() calls have been replaced with a proper usage of queue, which used operating system mutexes, locking, etc under the covers to timeout instead of Python's relatively expensive sleep().

There is still more to do. I think that the CommandPipeline.iterread() could also get rid of its sleep() in the inner loop and replace it with a queue (sometimes).  Other than that, we'd need to move to Cython, find faster regexes, etc get faster my preliminary profiling tells me.

PR #1946, opened 2016-11-11
some munging speed ups

scopatz:
I have rearranged some of the string and bytes munging of output to yield a faster runtimes in the test case given in #1932 and #1938. This is about a 25% performance improvement over the sleepless branch, and takes < 6% of the runtime of the original.

|                       | command took [s] | difference in outs |
|-----------------------|------------------|--------------------|
| 3545358a4b (master)   | 20.03104758      | 17                 |
| 5aec225 (1kb)         | 4.678267479      | 0                  |
| 8603909be (sleepless) | 1.552735806      | 0                  |
| b5d63ce               | 1.150137424      | 0                  |

My remaining ideas that don't involve optional external dependencies or writing a regular expression library haven't panned out.  They all added to the runtime, sometimes significantly. There is probably more profiling and fine tuning that can be done, but I would like to work on something other than run_subproc eventually too :)

### DIFF
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -62,10 +62,15 @@ ALTERNATE_MODE_FLAGS = LazyObject(
 RE_HIDDEN_BYTES = LazyObject(lambda: re.compile(b'(\001.*?\002)'),
                              globals(), 'RE_HIDDEN')
 
-
 @lazyobject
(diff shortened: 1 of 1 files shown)


scopatz:
Also, not sure this needs a separate news entry

laerus:
ty @scopatz !!

Issue #1947, opened 2016-11-11
Subprocess output is not captured using $(...) for some commands


cryzed:
```
cryzed@arch ~ $ xonsh --version
xonsh/0.4.7.dev538
cryzed@arch ~ $ a = $(echo "Hello")
cryzed@arch ~ $ a
'Hello\n'
cryzed@arch ~ $ b = $(pacman -Qmq)
anything-sync-daemon
arc-dark-suite-git
arc-suite-git
[...]
cryzed@arch ~ $ b
''
cryzed@arch ~ $ !(pacman -Qmq)
anything-sync-daemon
arc-dark-suite-git
arc-suite-git
[...]
CommandPipeline(stdin=None, stdout=None, stderr=None, pid=30659, returncode=0, args=['pacman', '-Qmq'], alias=None, stdin_redirect=['<stdin>', 'r'], stdout_redirect=['<stdout>', 'a'], stderr_redirect=['<stderr>', 'r'], timestamps=[1478876453.635026, 1478876453.9975464], executed_cmd=['/usr/bin/pacman', '-Qmq'], input=, output=, errors=None)
```

So the stdout/stderr output of the pacman command seems to be entirely uncaptured. This has definitely worked in previous versions.

scopatz:
Thanks for reporting, @cryzed. Are you on master? What is the output of the xonfig command?

cryzed:
Yes, I am, check the first few lines of the output I posted. [This](https://github.com/xonsh/xonsh/commit/28359582f3b71f3ef6cce4a592ab3d71f4e9ec60) commit to be specific.

```
cryzed@arch /mnt/sdb1/makepkg $ xonfig
+------------------+----------------------+
| xonsh            | 0.4.7.dev538         |
| Git SHA          | 2835958              |
| Commit Date      | Nov 11 13:46:35 2016 |
| Python           | 3.5.2                |
| PLY              | 3.9                  |
| have readline    | True                 |
| prompt toolkit   | 1.0.8                |
| shell type       | prompt_toolkit       |
| pygments         | 2.1.3                |
| on posix         | True                 |
| on linux         | True                 |
| distro           | arch                 |
| on darwin        | False                |
| on windows       | False                |
| on cygwin        | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+
```

scopatz:
[...]

PR #1953, opened 2016-11-14
remove pacman from threadable predictors

gforsyth:
as reported in #1947, `pacman` in the `default_threadable_predictors`
disables capture of all `pacman` commands.


scopatz:
Thanks @gforsyth!

scopatz:
I am going to merge this because there is basically no way that this effects mac. Thanks a ton @gforsyth!

Issue #1952, opened 2016-11-13
Exception when handle proc signals


mitnk:
I encountered this exception once, when I hit `Ctrl + C`:

```
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.5/bin/xonsh", line 3, in <module>
    main()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xonsh/main.py", line 221, in main
    shell.shell.cmdloop()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xonsh/readline_shell.py", line 399, in cmdloop
    self._cmdloop(intro=intro)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xonsh/readline_shell.py", line 359, in _cmdloop
    line = self.singleline()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xonsh/readline_shell.py", line 249, in singleline
    rtn = input(self.prompt)
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/site-packages/xonsh/proc.py", line 731, in _signal_int
    if self.proc.poll() is not None:
AttributeError: 'NoneType' object has no attribute 'poll'
```

Checked the code in `proc.py`:

```
        self.proc = None  # has to be here for closure for handles
        self.old_int_handler = self.old_winch_handler = None
        self.old_tstp_handler = self.old_quit_handler = None
        if on_main_thread():
            self.old_int_handler = signal.signal(signal.SIGINT,
                                                 self._signal_int)
            if ON_POSIX:
                self.old_tstp_handler = signal.signal(signal.SIGTSTP,
                                                      self._signal_tstp)
                self.old_quit_handler = signal.signal(signal.SIGQUIT,
                                                      self._signal_quit)
            if CAN_RESIZE_WINDOW:
                self.old_winch_handler = signal.signal(signal.SIGWINCH,
                                                       self._signal_winch)
        # start up process
        if ON_WINDOWS and stdout is not None:
            os.set_handle_inheritable(stdout.fileno(), False)
        self.proc = proc = subprocess.Popen(*args,
                                            stdin=stdin,
                                            stdout=stdout,
                                            stderr=stderr,
                                            **kwargs)

```

Seems the signal handlers are registered before `self.proc` are initialized. So that there a very rare chances got the exception I found earlier.

[...]

PR #1954, opened 2016-11-15
signal barrier

scopatz:
Hopeful fix for #1952


scopatz:
PS this works but doesn't have a news entry,

gforsyth:
thanks, @scopatz !

Issue #1956, opened 2016-11-16
new mercurial error on master


santagada:
I have mercurial installed, but would prefer that it is not even called (I have maybe one mercurial repo) but anyway it fails on my machine

```
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/xonsh/prompt/__amalgam__.py", line 649, in _get_field_value
    value = field_value() if callable(field_value) else field_value
  File "/usr/local/lib/python3.5/site-packages/xonsh/prompt/__amalgam__.py", line 472, in current_branch
    branch = get_hg_branch()
  File "/usr/local/lib/python3.5/site-packages/xonsh/prompt/__amalgam__.py", line 406, in get_hg_branch
    stderr=subprocess.DEVNULL)
  File "/usr/local/Cellar/python3/3.5.2_3/Frameworks/Python.framework/Versions/3.5/lib/python3.5/subprocess.py", line 626, in check_output
    **kwargs).stdout
  File "/usr/local/Cellar/python3/3.5.2_3/Frameworks/Python.framework/Versions/3.5/lib/python3.5/subprocess.py", line 693, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/local/Cellar/python3/3.5.2_3/Frameworks/Python.framework/Versions/3.5/lib/python3.5/subprocess.py", line 947, in __init__
    restore_signals, start_new_session)
  File "/usr/local/Cellar/python3/3.5.2_3/Frameworks/Python.framework/Versions/3.5/lib/python3.5/subprocess.py", line 1551, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: 'hg'
```

santagada:
I don't see it anymore on master.

PR #1957, opened 2016-11-16
vc fix

laerus:
added git and mercurial repo in tests dir
some testing for getting current branch name
renamed `xonsh.prompt.vc_branch` to `xonsh.prompt.vc`

 #1956


laerus:
in my machine this is where the commands cache is being populated, this adds some time to startup if we have no cache yet, there is also the chance that we may never have a cache file at startup.

xonsh/prompt/vc_branch.py
```diff
@@ -125,9 +125,9 @@ def current_branch(pad=NotImplemented):
                       "and will be removed in the future")
     branch = None
     cmds = builtins.__xonsh_commands_cache__
-    if cmds.lazy_locate_binary('git'):
+    if cmds.lazy_locate_binary('git') or cmds.locate_binary('git'):
         branch = get_git_branch()
-    if cmds.lazy_locate_binary('hg') and not branch:
+    if cmds.lazy_locate_binary('hg') or cmds.locate_binary('hg') and not branch:
```
melund: If I haven't got mercurial installed, then this will search my entire path for the `hg` binary for every prompt. 

I think the only solution is to have the `cmd_cache` update on xonsh startup in a separate thread. Then when the command cache is populated the branch labels will start to work. Of course the branch may then not work on the first prompt, but that the price to pay for fast startup.

laerus: you're right but i am not sure about using threads for updating the cache, maybe there is another thread which messes with the cache at the same time. caching will happen anyway at some point so we may want to do it on a higher level than this module.

laerus:
i think we can just run locate binary once and then always use lazy locate or just a global flag for a quick fix, but we can do better.

laerus:
Does subprocess uses the xonsh commands cache? if not then it searches the PATH and thats our job.
There are also the edges cases of removing/installing git/hg while in xonsh session.

i'm missing something but it seems that if we have no cache file then `__xonsh_commands_cache.is_empty()` returns True but the cache is then populated

melund:
Not sure I get what you mean. But if `is_empty()` then it seems like a bug.

laerus:
i had an occasion where if i used `is_empty()` twice it returned True and then False, but i can't reproduce this on a different machine though. Also i thought we were caching the commands on disk too.

i generally try to avoid calling subprocess/threading for minor things, the overhead feels excessive.

scopatz:
> Not sure I get what you mean. But if is_empty() then it seems like a bug.

I agree that seems like a bug.

laerus:
circle had a hiccup, i've restarted and it seems okay...

laerus:
this PR is mostly tests, i'm freezing this. I still think xonsh must do manually some of the job vc systems do, calling the binaries on each prompt update is slow. Getting the branch name from the dotdir of the repo and caching the root directory could be a speed boost, the hard part is knowing if the directory is dirty, that stats all files in the repo and building the prompt on big repos can be slow. 

We can always get away with this with by adding a note/warning sign on the documents.

scopatz:
> calling the binaries on each prompt update is slow.

It would be nice to have some timing stats on this, but it might be too hard to collect since it is going to be highly dependent on the system.

Thanks @laerus, I am going to merge this. Great tests!

PR #1966, opened 2016-11-20
Make history show all sorted

mitnk:
In this PR, I want to make the history sorted. It still work in progress, but I open it here for comments.

Currently when you run `history show all -t | head`, `history show all -t | tail`, xonsh mess up the order of results (when you have tons of history files). The first commit is sorting them by modify time. Maybe sorting by `ts` attr of cmds is more accurate but needs more work or sacrifice the performance. Ideas on this?

Also fixed a small issue on deleting empty history files.

### DIFF
--- a/xonsh/history.py
+++ b/xonsh/history.py
@@ -2,7 +2,6 @@
 """Implements the xonsh history object."""
 import os
 import sys
-import glob
 import json
 import time
 import uuid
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks for putting this in @mitnk!  I think sorting by the ts of each command is probably too much for the moment.  Maybe that could be an option in HISTCONTROL

xonsh/history.py
```diff
@@ -70,6 +70,15 @@ def _gc_bytes_to_rmfiles(hsize, files):
     return rmfiles
 
 
+def _get_history_files(reverse=False):
+    data_dir = builtins.__xonsh_env__.get('XONSH_DATA_DIR')
+    data_dir = expanduser_abs_path(data_dir)
+    files = [os.path.join(data_dir, f) for f in os.listdir(data_dir)
+             if f.startswith('xonsh-') and f.endswith('.json')]
+    files.sort(key=lambda x: os.path.getmtime(x), reverse=reverse)
```
scopatz: Can't you open up the files and look at the session timestamp? See http://xon.sh/tutorial_hist.html#how-is-xonsh-history-different

mitnk: That will be as the same much of work as check ts attrs, isn't it?

scopatz: Maybe, but you probably also need to filter out the locked files, which means you need to open up the file anyway. Or does that filtering happen later?

mitnk: We only need to check `lock` when deleting `gc`, not for listing `show`

mitnk:
@scopatz Updated a bit. It's ready for another look, thanks.

scopatz:
LGTM, thanks @mitnk! Just waiting for CI

scopatz:
Thanks!

PR #1968, opened 2016-11-20
Bashisms

scopatz:
This starts a bashisms xontrib that performs the mininal transformation.  In order to get it working, some refactoring and cleanup was needed.

### DIFF
--- a/xonsh/shell.py
+++ b/xonsh/shell.py
@@ -1,7 +1,9 @@
 # -*- coding: utf-8 -*-
 """The xonsh shell"""
 import os
+import sys
 import random
+import difflib
 import builtins
 import warnings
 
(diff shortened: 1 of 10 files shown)


scopatz:
ping!

melund:
I haven't had a chance to try it but I did have some time to read the code on my phone. It looks good to me. The precommand event is quite interesting. I think it will be extremely useful for a lot of application specific things.

Unless you want to give others a chance to test I am OK with merging this in.

gforsyth:
I was traveling yesterday but I'll take a look at this today

xonsh/shell.py
```diff
@@ -27,6 +29,34 @@
...
+        lst = src
+        srcs = events.on_precommand.fire(src)
+        for s in srcs:
+            if s != lst:
+                src = s
+                break
+        i += 1
+        if i == limit:
+            print_exception('Modifcations to source input took more than '
+                            'the recurssion limit number of interations to '
```
gforsyth: recursion

scopatz: Got it, thanks.

xonsh/xontribs.json
```diff
@@ -1,4 +1,25 @@
...
+ },
+ {"name": "bashisms",
+  "package": "xonsh",
+  "url": "http://xon.sh",
+  "description": [
+    "Enables additional Bash-like syntax while at the command prompt. For ",
+    "example, the ``!!`` syntax for running the previous command is now usable.",
+    "Note that these features are implemented as precommand events and these ",
+    "additions do not affect the xonsh language when run as script. That said, ",
+    "you might find them useful if you have strong musucle memory."]
```
gforsyth: muscle

gforsyth:
Ok, this works and I like it.  

Thought and question: I had been working on something similar and while there is no way around it, transforming input before it's executed would seem to open `xonsh` up to malicious xontribs.  Should there be a warning the first time that a xontrib is going to modify what was entered by the user before execution?

gforsyth:
Also, to be fair, there could also be a xontrib that just does `rm -rf /` whenever you load it and that would have nothing to do with transforming `src`

scopatz:
Also, this is pretty simple and the actual implementation could be really expanded.

scopatz:
The test failure might be related to circle's github service being down.  @laerus, do you have any ideas here since I think this it related to your recent VC work.  Is it OK to not do the assert if we get back a None?

gforsyth:
I think we can also take the time to review xontribs that are submitted to make sure they aren't malicious.  Just waiting on CI for this, then

gforsyth:
The circle failure seems to have been a fluke...?

scopatz:
> The circle failure seems to have been a fluke...?

I added the workaround :)

scopatz:
CI finally finished :)

gforsyth:
Awesome!  Thanks @scopatz!

Issue #1961, opened 2016-11-19
ENH: tag history


CJ-Wright:
It might be nice to tag xonsh sessions so that we can search our histories for the session we want.

laerus:
True that would be a nice feature, thanks for the proposal @CJ-Wright !

CJ-Wright:
Would a reasonable syntax look like:
```python
history tag ['tag1', 'tag2']
history search 'tag1'
```

scopatz:
Thanks @CJ-Wright!  I am +1 on this idea.  It should probably follow the usual subprocess parsing conventions.

PR #1970, opened 2016-11-22
New History Backend - Sqlite

mitnk:
This is currently just an POC, I open it here for collecting ideas, to prevent the direction of it from too wrong. Eventually I hope It can introduce a second history backend - sqlite, and some public interfaces of the history backend (or call it engine?) be defined.

The APIs, the naming of all aspects are still need to be determined. A lot of details need to be determined, e.g. Table fields defines in sqlite.

Please leave any ideas, concerns. Thanks!


scopatz:
Thanks for putting this in @mitnk! I think this is a great idea and definitely a step in the right direction. (And probably long overdue.)

setup.py
```diff
@@ -306,7 +306,7 @@ def main():
         classifiers=['Programming Language :: Python :: 3'],
         packages=['xonsh', 'xonsh.ply.ply', 'xonsh.ptk', 'xonsh.parsers',
                   'xonsh.xoreutils', 'xontrib',
-                  'xonsh.completers', 'xonsh.prompt'],
+                  'xonsh.completers', 'xonsh.history', 'xonsh.prompt'],
```
scopatz: See above in this file for the hooks

xonsh/built_ins.py
```diff
@@ -23,7 +23,7 @@
 
 from xonsh.ast import AST
 from xonsh.lazyasd import LazyObject, lazyobject
-from xonsh.history import History
+from xonsh.history import get_history_backend
```
scopatz: This should be something like, 

```python
import xonsh.history as xh
```

like we do with other imports that are outside of the current subpackage. See https://github.com/xonsh/amalgamate for more details

xonsh/history/__init__.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
```
scopatz: The amalgamate hooks should go in this file.

xonsh/history/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+from xonsh.history.sqlite import History as SqliteHistory
+
+_BACKENDS = {
+    'dummy': DummyHistory,
+    'json': JsonHistory,
+    'sqlite': SqliteHistory,
+}
+
+
+def get_history_backend(env, ts, locked, gc=True, filename=None):
```
scopatz: Maybe this can be `construct_history()` or similar.

xonsh/history/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+    env = builtins.__xonsh_env__
+    backend = env.get('XONSH_HISTORY_BACKEND', 'json')
+    try:
+        kls_history = _BACKENDS[backend]
+    except KeyError:
```
scopatz: No need to use try-except here. This can be done with `if backend not in BACKENDS`

xonsh/history/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+"""Implements the xonsh history object."""
+import builtins
+import time
+
+from xonsh.history.dummy import History as DummyHistory
+from xonsh.history.json import History as JsonHistory
+from xonsh.history.json import history_main
+from xonsh.history.sqlite import History as SqliteHistory
+
+_BACKENDS = {
```
scopatz: This should be public because xonrtribs and users may want to register their own backends.

xonsh/history/__init__.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+"""Implements the xonsh history object."""
+import builtins
+import time
+
+from xonsh.history.dummy import History as DummyHistory
+from xonsh.history.json import History as JsonHistory
+from xonsh.history.json import history_main
+from xonsh.history.sqlite import History as SqliteHistory
```
scopatz: Also, when you amalgamate, these class names should be changed to their aliases so that they don't override each other.

xonsh/history/dummy.py
```diff
@@ -0,0 +1,30 @@
...
+        self.rtns = None
+        self.last_cmd_rtn = None
+        self.last_cmd_out = None
+
+    def __iter__(self):
+        for cmd, ts, index in []:
+            yield (cmd, ts, index)
+
+    def append(self, cmd):
+        print('DummyHistory append cmd: {}'.format(cmd))
```
scopatz: Should these go to stderr?

xonsh/history/dummy.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def flush(self, at_exit=False):
+        print('DummyHistory flush ...')
+
+    def get_history_items(self):
```
scopatz: I would prefer if this was just called `items()` and was a generator (yielded items, rather than returned a list).

xonsh/history/json.py
```diff
@@ -635,6 +635,18 @@ def flush(self, at_exit=False):
         self.buffer.clear()
         return hf
 
+    def get_history_items(self):
```
scopatz: Same as in dummy.

xonsh/history/sqlite.py
```diff
@@ -0,0 +1,92 @@
+# -*- coding: utf-8 -*-
+"""Implements the xonsh history backend via sqlite3."""
+import builtins
+import os.path
```
scopatz: os imports path, so this just needs to be `import os`

xonsh/history/sqlite.py
```diff
@@ -0,0 +1,92 @@
...
+import sqlite3
+import threading
+import time
+
+from xonsh.tools import expanduser_abs_path
```
scopatz: importing from outside of package, so import should be `import xonsh.tools as xt`

xonsh/history/sqlite.py
```diff
@@ -0,0 +1,92 @@
...
+
+__all__ = ['History']
```
scopatz: probably should remove since it won't survive amalgamation.

xonsh/history/sqlite.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def _get_conn():
+    data_dir = builtins.__xonsh_env__.get('XONSH_DATA_DIR')
+    data_dir = expanduser_abs_path(data_dir)
+    db_file = os.path.join(data_dir, 'xonsh-history.db')
```
scopatz: I'd prefer if this was `*.sqlite`

xonsh/history/sqlite.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def _insert_command(cursor, cmd):
+    cursor.execute("""
+        INSERT INTO xonsh_history VALUES(?, ?, ?, ?)
+    """, (cmd['inp'].rstrip(), cmd['rtn'], cmd['ts'][0], cmd['ts'][1]))
+
+
+def _get_records(cursor):
+    cursor.execute("""SELECT inp FROM xonsh_history ORDER BY tsb""")
```
scopatz: This should be a single quoted string.

astronouth7303:
The `on_precommand` and `on_postcommand` should provide all the data necessary to implement this with minimum invasion.

I would suggest refactoring the JSON backend to use these and add a config option for which backend to enable (or have individual options). This also provides a template for xontribs to implement their own backends. (eg, I might like a CouchDB backend for myself so I can aggregate my history.)

(This is, of course, pushing my own agenda of extensibility, modularity, and not-in-core. I still think it's a good one.)

scopatz:
You are right @astronouth7303 that most of the adding items to history can be done with postcommand. However, we still need a place to add the backend specializations. Also, different backends need to be able to read back in data through a common interface. I am not sure that can be done through an event, but maybe I am missing something....

mitnk:
Still WIP, but comments are welcome!

xonsh/history/dummy.py
```diff
@@ -0,0 +1,17 @@
...
+class DummyHistory(HistoryBase):
+    def append(self, cmd):
+        print('DummyHistory append: {}'.format(cmd))
+
+    def flush(self, at_exit=False):
+        print('DummyHistory flush ...')
+
+    def items(self):
+        for item in [{'inp': 'dummy in action\n'}]:
+            yield item
```
scopatz: I think the loop here can just be replaced by `yield {'inp': 'dummy in action\n'}`

xonsh/history/json.py
```diff
@@ -635,6 +398,58 @@ def flush(self, at_exit=False):
...
+                json_file = xlj.LazyJSON(f, reopen=False)
+            except ValueError:
+                # Invalid json file
+                continue
+            commands = json_file.load()['cmds']
+            for c in commands:
+                yield (c['inp'].rstrip(), c['ts'][0], ind)
+                ind += 1
+
+    # TODO: merge methods all_items() and items() to one
```
scopatz: I am not sure I understand the distinction here...

xonsh/history/main.py
```diff
@@ -0,0 +1,323 @@
...
+    'json': History,
+    'sqlite': SqliteHistory,
+}
+
+
+def construct_history(env, ts, locked, gc=True, filename=None):
+    env = builtins.__xonsh_env__
+    backend = env.get('XONSH_HISTORY_BACKEND', 'json')
+    if backend not in HISTORY_BACKENDS:
+        print('Unknown history backend: {}. Use Json version'.format(backend))
```
scopatz: This should be printed to `file=sys.stderr`, since it is a warning message.

xonsh/history/sqlite.py
```diff
@@ -0,0 +1,79 @@
...
+        t = time.time()
+        xh_sqlite_append_history(cmd)
+        print('history cmd: {} took {:.4f}s'.format(cmd, time.time() - t))
+
+    def flush(self, at_exit=False):
+        print('SqliteHistory flush() called')
+
+    def items(self):
+        records = xh_sqlite_items()
+        return [{'inp': x[0]} for x in records]
```
scopatz: This should yield these values.

scopatz:
Thanks for making all of the amalgamate changes @mitnk! I think this needs a couple more robustness features before it would be good to go:

* The autodoc hooks in the docs/api need to be updated.
* Tests that make sure that we can read/write to each backend.

I don't think we need too much in the way of testing, but at least something to verify that it works.

mitnk:
Yes, this PR is far from good to go. I'd say it's about 10% complete. Will let you guys know when it's ready for testing.

scopatz:
Thanks, I am super excited to have something like this go in. Please let us know whenever you want us to review

astronouth7303:
@scopatz Events can scatter/gather, but you'd need a way to merge history from multiple sources.

The option of multiple, first-class history backends opens up the possibility of complex history setups (blending local, remote, indexed, etc).

On consideration, having the `history` command being able to merge query results from multiple backends isn't a bad idea. If you have both local history and remote aggregated history (basically, history from all your boxes in one place), it would be nice if `history` (and history-using features) could handle the redundancy of results. (Note that this might not be an actual remote service. Could be a dropbox folder or `rsync` on a cron job.)

I'm also wondering how the current interface works for "richer" searching? (eg, fed through Lucene or ElasticSearch)

xonsh/history/main.py
astronouth7303: "Using JSON version"

CJ-Wright:
I think this may also help to address #1961

xonsh/history/main.py
mitnk: Thanks @astronouth7303 those suggestions are what I always needed :)

mitnk:
@CJ-Wright Yes, I think we need to consider the extension ability when thinking the APIs of the backend.

xonsh/history/json.py
mitnk: One of them were used to feed history items to readline (when do history search etc); one is for `history show all`, I think we can merge them into one. I will work on them.

mitnk:
For sql table design, some quick idea of mine. Considered simple, extensible, and migration[1] in mind. Tables are:

**xonsh_history**: this is the main table, store the basic input informations, like `inp`, `rtn`, `ts`, `session` (the uuid string or a ID) etc.

**xonsh_history_cmd_info**: this table could store extra information for commands, like `output`, or `tags`? With default setting, this table won't be used. three column: FK of an xonsh_history item, name, value.

**xonsh_history_session_info**: same idea as `xonsh_history_cmd_info` but for storing information for sessions.

We may or may not need a `xonsh_history_session` table.

[1]  Migration can be very complex. If we changed the table schema at some version of xonsh, we have to provide some migration tool for users. Whether the migration is via external cmd, or automatically done by sqlite history backend itself, it would be very complex. So we better design a schema for the first time and work forever (if possible).

Ideas on this? cc @scopatz

astronouth7303:
Why put output in it's own table like that?

We could half-ass migrations by only allowing the additions of columns, never their removal. Given that I don't think the history schema has changed recently (and is now in the event signatures, too), I don't think we should expect many changes.

mitnk:
@astronouth7303 you mean "why not", right? we could. I was thinking it was not in use by default and the output size could be large - let's put it in the main table - if no other ideas here.

Columns-append-only is good strategy. Thanks!

astronouth7303:
@mitnk Check how SQLite handles large blobs. Input can be just as large.

mitnk:
I think this PR is ready for the first batch of code review, before I start work on fixing CI, adding tests and adding doc strings.

One thing need to be done before merging it is that making history backend be set in `xonshrc` file. Until my last commit, `$XONSH_HISTORY_BACKEND` works like `$XONSH_DEBUG`, only works with setting it prior to stating xonsh.

mitnk:
Oh, haven't update it for PTK, please use readline shell if want to have some testing. I will update it for PTK soon.

scopatz:
I am probably not the primary target for the SQL backend, and so @astronouth7303 and others so probably comment on table design. My general opinion is that the fewer tables the better though, because joins are so expensive.  But maybe there are exceptions for performance reasons.   Append-only columns seems like the right strategy though.

xonsh/shell.py
```diff
@@ -87,7 +89,12 @@ def __init__(self, ctx=None, shell_type=None, config=None, rc=None,
         self._init_environ(ctx, config, rc,
                            kwargs.get('scriptcache', True),
                            kwargs.get('cacheall', False))
+
         env = builtins.__xonsh_env__
+        # build history backend before creating shell
+        builtins.__xonsh_history__ = xhm.construct_history(
+            env=env.detype(), ts=[time.time(), None], locked=True)
```
scopatz: Why was this moved here from the `load_builtins()`?

xonsh/history/base.py
```diff
@@ -0,0 +1,99 @@
...
+        """Display history sessionid."""
+        if not self.sessionid:
+            return
+        print(str(self.sessionid), file=stdout)
+
+    def on_file(self, ns, stdout=None, stderr=None):
+        """Display history file name if it exists."""
+        if not self.filename:
+            return
+        print(str(self.filename), file=stdout)
```
scopatz: These `on_*()` methods seem tied to the CLI. I am not sure this base history object is the right place for them. Maybe they should be closer to the CLI and just operate on the current `__xonsh_history__`.

xonsh/history/base.py
```diff
@@ -0,0 +1,99 @@
...
+        """Display all history items."""
+        raise NotImplementedError
+
+    def session_items(self):
+        """Display history items of current session."""
+        raise NotImplementedError
+
+    def on_info(self, ns, stdout=None, stderr=None):
+        """Display information about the shell history."""
+        pass
```
scopatz: Maybe there can just be an `info()` method which returns a dict of the string keys and values that hold the interesting information.

xonsh/history/base.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class HistoryGC(threading.Thread):
+    pass
+
+
+class HistoryBase:
+    def __init__(self, sessionid=None, gc=True, **kwargs):
+        self.sessionid = uuid.uuid4() if sessionid is None else sessionid
+        self.gc = HistoryGC() if gc else None
```
scopatz: HistoryGC doesn't seem to do anything, so maybe gc should just be set to None on the base?

xonsh/history/json.py
```diff
@@ -0,0 +1,459 @@
...
+    ----------
+    rtns : sequence of ints
+        The return of the command (ie, 0 on success)
+    inps : sequence of strings
+        The command as typed by the user, including newlines
+    tss : sequence of two-tuples of floats
+        The timestamps of when the command started and finished, including
+        fractions
+    outs : sequence of strings
+        The output of the command, if xonsh is configured to save it
```
scopatz: Note that these attrs are part of the base interface, but they only seem to be present on the JSON history, unless I am missing something....

scopatz: It would be nice if they somehow got pushed down to `HistoryBase` so that all subclasses would get them

xonsh/xonfig.py
```diff
@@ -353,6 +353,7 @@ def _info(ns):
     data.extend([
         ('Python', '{}.{}.{}'.format(*PYTHON_VERSION_INFO)),
         ('PLY', ply.__version__),
+        ('history backend', env.get('XONSH_HISTORY_BACKEND', 'json')),
```
scopatz: `$XONSH_HISTORY_BACKEND` should be added as a full-fledged environment variable in environ.py. That way you will only have to set the default in one place.

xonsh/shell.py
mitnk: I moved it here so that history backend is created after loading xonshrc and knows which one need to be created (but before shell got created). Should I achieve this through other ways?

xonsh/history/base.py
mitnk: removed HistoryGC.

xonsh/history/json.py
mitnk: Thanks! Will work on here later.

xonsh/shell.py
scopatz: hmm this might be OK for this PR.  We really need a pre-shell/post-load event, I think.

astronouth7303: I'll get on that...

astronouth7303: See #1989.

xonsh/shell.py
```diff
@@ -131,7 +131,8 @@ def __init__(self, ctx=None, shell_type=None, config=None, rc=None,
         self.shell = shell_class(execer=self.execer,
                                  ctx=self.ctx, **kwargs)
         # allows history garbace colector to start running
```
astronouth7303: "garbace"

scopatz:
Also @mitnk - let us know when you think it is ready for another look!

mitnk:
@scopatz I think this PR is ready for another look and it also ready for testing the two history backends.

Discussed with @astronouth7303 in Gitter and he has some ideas on make history plugin-able (not the core of xonsh). And I think it does make sense. In this PR I've make history object exports as few APIs as possible. Then when we want to move the history around I think it won't be hard.

The new backend sqlite does not support `diff` and `replay` action. 1) The current implement of these two are coupled with JSON features and 2) I don't find my own cases to use diff & replay. So ... but still we can support them if anyone want to in future.

A final note is - if you find you're in a weird limbo thing (do some testing on new backend) when switching between master and this branch, a `sudo rm -rf ./build/` may help to clean things up.

mitnk:
Here is a tip for anyone who want to try sqlite backed history:

```
# set $XONSH_HISTORY_BACKEND = 'sqlite' in xonshrc file
$ pip3.5 install sqlite-web
...
$ sqlite_web @$(history file)
```

xonsh/environ.py
```diff
@@ -299,6 +299,7 @@ def DEFAULT_VALUES():
         'XONSH_DEBUG': False,
         'XONSH_ENCODING': DEFAULT_ENCODING,
         'XONSH_ENCODING_ERRORS': 'surrogateescape',
+        'XONSH_HISTORY_BACKEND': 'json',
```
scopatz: Please add some validators above, thanks!

xonsh/history/base.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def items(self):
+        """Get history items of current session."""
+        raise NotImplementedError
+
+    def all_items(self):
+        """Get all history items."""
+        raise NotImplementedError
+
+    def info(self, ns, stdout=None, stderr=None):
```
scopatz: I don't think that info() should have any arguments

xonsh/replay.py
```diff
@@ -105,17 +105,17 @@ def _rp_create_parser(p=None):
...
 
-def _rp_main_action(ns, h=None):
+def replay_main_action(h, ns, stdout=None, stderr=None):
     replayer = Replayer(ns.path)
     hist = replayer.replay(merge_envs=ns.merge_envs, target=ns.target)
     print('----------------------------------------------------------------')
     print('Just replayed history, new history has the following information')
     print('----------------------------------------------------------------')
-    _hist_info(ns, hist)
+    hist.on_info(ns, stdout=stdout, stderr=stderr)
```
scopatz: Shouldn't this just be "info()"

scopatz: Also, info should return a dict, which this then uses.

scopatz:
Hi @mitnk - This is looking really great. There are a couple of minor issue and then one big one.  It looks like replay might be broken for all backends. I pointed this out where this happens in the code.  

It also seems like the `construct_history()` kwarg filename should be renamed to sessionid, and then the constructor is responsible for finding the session with that ID. I think that and making sure that replay and diff work for the sqlite backend can be saved for another PR.

mitnk:
@scopatz Thanks, I mentioned the `diff` & `replay` in [my previous comment](https://github.com/xonsh/xonsh/pull/1970#issuecomment-264185528). I will update for other comments soon :)

mitnk:
Updated the code related for info and replay. Could you help to do some tests on JSON's replay action to make sure I didn't break it - I find myself hard to understand what the `replay` action do :-) (haven't dig into it yet)

I actually do not get your idea on dropping the `filename` from `construct_history()`. Drop `filename` from it means drop `filename` from `History.__init__()` right? Then do you mean we use something like this instead? (since we need a way to hardcode the filename in `replay` action and in `tests/`.) -
```
def __init__(self, sessionid):
    if str(sessionid).endswith('.json'):
        self.filename = self.sessionid
    else:
        self.filename = str(self.sessionid) + '.json'
```

tests/test_history_sqlite.py
```diff
@@ -0,0 +1,178 @@
...
+
+    xonsh_builtins.__xonsh_env__['HISTCONTROL'] = 'ignoredups,ignoreerr'
+    assert len(hist) == 0
+
+    # An error, items() remains empty
+    hist.append({'inp': 'ls foo', 'rtn': 2})
+    assert len(hist) == 0
+    assert len(hist.inps) == 1
+    assert len(hist.rtns) == 1
+    assert 2 == hist.rtns[-1]
```
mitnk: @scopatz From the APIs point view, do you think this is correct? i.e. when err or dups encountered, `.inps` and `.rtns` will be appending while `items()` will not?

xonsh/history/base.py
```diff
@@ -0,0 +1,171 @@
...
+            A dict contains informations of a command. It should contains
+            the following keys like ``inp``, ``rtn``, ``ts`` etc.
+        """
+        pass
+
+    def flush(self, **kwargs):
+        """Flush the history items to disk from a buffer."""
+        pass
+
+    def items(self):
```
mitnk: @scopatz in this API, it yields items like `{'inp': 'ls foo', 'rtn': 1}`. My question is do we need to make it strict like which key it must provide? (e.g. `inp`, `rtn`, `out`, `tss` ...). or some keys are must, and some are optional?

tests/test_history_sqlite.py
scopatz: Yeah, I think that is correct. Also, sorry about not getting to this sooner. I have been sick the past couple of days and had a lot of end-of-the-semester work to do.

scopatz:
Thanks for this awesome feature, @mitnk! I think that it was in a stable enough state to merge in. We can go ahead and clean it up in future PRs.

tests/test_history_sqlite.py
mitnk: never mind, no complaints on the speed, we have no deadlines, ha.. 

Sorry to hear that. I hope you feel better now.

mitnk:
Thanks @scopatz, not expecting this merge happens so soon :) But really nice to see this it got merged!

I had plans to fix the `__xonsh_history__.rtns` and `ignoredups` (not introduced in this PR tho). Will put them in another PR.

scopatz:
Well, I wanted to avoid conflicts with the store-stdout issue @gforsyth brought up. That failed, but hopefully this will still reduce the workload overall

xonsh/history/dummy.py
Russell-Jones-OxPhys: yield or yield from?

astronouth7303: `yield`. `yield from` would be:

```python
yield from [{'inp': 'dummy in action\n'}]
```

Russell-Jones-OxPhys: Yep, I read it too quickly :| I was fooled by the for loop into thinking it was a set or dict comprehension with several members.

PR #1971, opened 2016-11-23
Added ability to format stderr pre- and post-fixes

scopatz:
This should supersede the implementation in #1277, which shockingly was only in June.  I think this might have fixed some errant issues with some pygments formatting, but I haven't tested it rigorously.


laerus:
ty @scopatz!

PR #1972, opened 2016-11-24
Fix ON_ANACONDA

melund:
The ON_ANACONDA returned false when python was installed from the conda-forge branch.


laerus:
ty @melund!

scopatz:
Thanks!

PR #1978, opened 2016-11-26
Add FileNotFoundError exception to the handler returning None

anddam:
After installing de9e1686353ec1d7a76e76e2c43185406de88e73 I got

    FileNotFoundError: [Errno 2] No such file or directory: 'hg'

at every new shell instance. This is due to the fact that my tools are in /opt/local since I'm using MacPorts on macOS but at the time the subprocess is called in get_hg_branch() PATH is just

    PATH=/usr/bin:/bin:/usr/sbin:/sbin

I'm adding FileNotFoundError to the None case handler of subprocess.check_output calls where needed, one of those already had the exception.

The rationale is that if the tool is missing then just return None like if there wasn't a repository at all.

### DIFF
--- a/xonsh/prompt/vc.py
+++ b/xonsh/prompt/vc.py
@@ -18,7 +18,7 @@ def _get_git_branch(q):
             ['git', 'branch'],
             stderr=subprocess.DEVNULL
         )).splitlines()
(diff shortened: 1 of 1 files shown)


scopatz:
Seems reasonable to me, thanks!

PR #1988, opened 2016-11-28
front page hooks to xonsh-live

scopatz:
Hopefully my javascript is half-way decent.


scopatz:
No news needed.

docs/index.rst
```diff
@@ -56,15 +56,44 @@ that you are used to from Bash and IPython. It works on all major systems includ
...
+        border-radius: 8px;
+        position:relative;
+        float: right;
+        right: 35%;
+        bottom: 240px;
+    }
+    </style>
+    <div id="trydiv"><p style="text-align:center;">
+    <iframe id="tryframe" data-src="http://hermit.astro73.com/"
+            src="_static/xonsh-live.png" width="80%" height="480px"
```
astronouth7303: I didn't know you could do that with an iframe.

gforsyth:
This works locally, so i guess it's just time to merge it and see what happens.

Question re: persistence -- I `touch`ed a few files and they stick around between sessions -- is there a timeout when everything gets reset?

gforsyth:
Thanks @scopatz !  And thanks @astronouth7303 for putting together the shell-in-a-box!

gforsyth:
Hmm... it shows up but when I click it isn't interactive, just an image that zooms in and out:
https://xonsh.github.io/xonsh-docs/dev/

astronouth7303:
> is there a timeout when everything gets reset?

Not at this time. See xonsh/inabox

> Hmm... it shows up but when I click it isn't interactive, just an image that zooms in and out:
https://xonsh.github.io/xonsh-docs/dev/

For me, it's refusing to load the iframe because of mixed HTTP/HTTPS content.

docs/index.rst
scopatz: It is pretty sneaky. I had to look it up :)

scopatz:
Hmm how do we get around this then? https://hermit.astro73.com/ doesn't seem to work...

astronouth7303:
No, I didn't set up HTTPS. Unfortunately, the only solution appears to be implement HTTPS.

See xonsh/inabox#1

scopatz:
dang

adqm:
Is there any kind of sandboxing on this?  It seems that everyone who logs in can do whatever they want (and that the user has sudo access if they can figure out the password).

I like the idea of an interactive demo, but this makes me really nervous (unless I'm missing something, it doesn't seem like there is anything in place to prevent dangerous things or misuse of those computing resources, either accidental or intentional).

PR #1989, opened 2016-11-29
Load events

astronouth7303:
Implement #1550.

Add an event species that is single-shot: It is fired once, but guarantees that all handlers (no matter when they are registered) will be called as long as it is fired ever.

Also added a collection of events:

* `on_post_init`
* `on_pre_cmdloop`
* `on_pre_rc`
* `on_post_rc`
* `on_ptk_create`

(As an extra, created a makefile with some of the management commands, because I got tired of remembering the amalgamation-related incantations.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,16 @@
+# Make GNU Make xonshy
+SHELL=xonsh
+.SHELLFLAGS=-c
+.ONESHELL:
+
+xonsh/ply:
+	git subtree pull --prefix xonsh/ply https://github.com/dabeaz/ply.git master --squash
+
+
+.PHONY: clean
+clean:
+	find xonsh -name __amalgam__.py -delete -print
+
+.PHONY: amalgamate
+amalgamate:
+	python3 amalgamate.py xonsh
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,18 @@
 SHELL=xonsh
 .SHELLFLAGS=-c
 .ONESHELL:
+.SILENT:
+
+# Unlike normal makefiles: executes the entire body in one go under xonsh, and doesn't echo
+
+.PHONY: help
+help:
+	print("""
+	Utility file for xonsh project. Try these targets:
+	* amalgamate: Generate __amalgam__.py files
+	* clean: Remove generated files (namely, the amalgamations)
+	* xonsh/ply: Pull down most recent ply
+	""")
 
 xonsh/ply:
 	git subtree pull --prefix xonsh/ply https://github.com/dabeaz/ply.git master --squash
@@ -13,4 +25,7 @@ clean:
 
 .PHONY: amalgamate
 amalgamate:
-	python3 amalgamate.py xonsh
+	sys.path.insert(0, '.')
+	import setup
+	setup.amalgamate_source()
+	_ = sys.path.pop(0)
--- a/update_ply.xsh
+++ b/update_ply.xsh
@@ -1,4 +1,0 @@
-#!/usr/bin/env xonsh
-
-# update the ply repo bundled with xonsh
-git subtree pull --prefix xonsh/ply https://github.com/dabeaz/ply.git master --squash
--- a/xonsh/events.py
+++ b/xonsh/events.py
@@ -145,7 +145,12 @@ class LoadEvent(AbstractEvent):
     """
     An event species where each handler is called exactly once, shortly after either the event is
     fired or the handler is registered (whichever is later).
+
+    Note: Does not support scatter/gather, due to never knowing when we have all the handlers.
+
+    Note: Maintains a strong reference to pargs/kwargs in case of the addition of future handlers.
     """
+    # NOTE: This is currently NOT THREAD SAFE.
     def __init__(self):
         self._fired = set()
         self._unfired = set()
@@ -167,7 +172,11 @@ class LoadEvent(AbstractEvent):
 
         This has no effect if the element is already present.
         """
-        self._fired.add(item)
+        if self._hasfired:
+            self._call(item)
+            self._fired.add(item)
+        else:
+            self._unfired.add(item)
 
     def discard(self, item):
         """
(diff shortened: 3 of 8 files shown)


Makefile
```diff
@@ -0,0 +1,16 @@
...
+	git subtree pull --prefix xonsh/ply https://github.com/dabeaz/ply.git master --squash
+
+
+.PHONY: clean
+clean:
+	find xonsh -name __amalgam__.py -delete -print
+
+.PHONY: amalgamate
+amalgamate:
+	python3 amalgamate.py xonsh
```
astronouth7303: I have no idea which subpackages do and do not get amalgamated.

update_ply.xsh
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env xonsh
-
-# update the ply repo bundled with xonsh
-git subtree pull --prefix xonsh/ply https://github.com/dabeaz/ply.git master --squash
```
scopatz: I am not sure I am a fan of removing this file.

astronouth7303: I could redo it to point at the Makefile.

xonsh/events.py
```diff
@@ -145,7 +145,12 @@ class LoadEvent(AbstractEvent):
     """
     An event species where each handler is called exactly once, shortly after either the event is
     fired or the handler is registered (whichever is later).
+
+    Note: Does not support scatter/gather, due to never knowing when we have all the handlers.
+
+    Note: Maintains a strong reference to pargs/kwargs in case of the addition of future handlers.
     """
+    # NOTE: This is currently NOT THREAD SAFE.
```
scopatz: Maybe this should be in the docstring

xonsh/main.py
```diff
@@ -19,6 +19,39 @@
...
+
+Fired just before  the command loop is started, if it is.
+""")
+
+
+events.transmogrify('on_pre_rc', 'LoadEvent')
+events.doc('on_pre_rc', """
+on_pre_rc() -> None
+
+Fired just before xonshrc is loaded, if it is.
```
scopatz: There isn't a single RC file, maybe this (and the following) docstring should be a little clearer in this regards.

xonsh/main.py
```diff
@@ -19,6 +19,39 @@
...
+on_post_init() -> None
+
+Fired after all initialization is finished and we're ready to do work.
+
+NOTE: This is fired before the wizard is automatically started.
+""")
+
+events.transmogrify('on_pre_cmdloop', 'LoadEvent')
+events.doc('on_pre_cmdloop', """
+on_pre_cmdloop() -> None
```
scopatz: Also, maybe there should be a post cmdloop event too?

xonsh/ptk/shell.py
```diff
@@ -20,6 +20,15 @@
 from xonsh.ptk.history import PromptToolkitHistory
 from xonsh.ptk.key_bindings import load_xonsh_bindings
 from xonsh.ptk.shortcuts import Prompter
+from xonsh.events import events
+
+
+events.transmogrify('on_ptk_create', 'LoadEvent')
+events.doc('on_ptk_create', """
+on_ptk_create(prompter: Prompter, history: PromptToolkitHistory, completer: PromptToolkitCompleter, bindings: KeyBindingManager) ->
```
scopatz: PEP8: line too long

xonsh/ptk/shell.py
```diff
@@ -37,6 +46,8 @@ def __init__(self, **kwargs):
             }
         self.key_bindings_manager = KeyBindingManager(**key_bindings_manager_args)
         load_xonsh_bindings(self.key_bindings_manager)
+        # This assumes that PromptToolkitShell is a singleton
+        events.on_ptk_create.fire(self.prompter, self.history, self.pt_completer, self.key_bindings_manager)
```
scopatz: PEP8 line too long

xonsh/main.py
astronouth7303: I figured atexit would cover those use cases? I suppose it wouldn't be hard to use atexit to make sure closing events fire.

xonsh/events.py
```diff
@@ -178,8 +187,22 @@ def discard(self, item):
...
+    def _call(self, handler):
+        try:
+            handler(*self._pargs, **self._kwargs)
+        except Exception:
+            print_exception("Exception raised in event handler; ignored.")
+
     def fire(self, *pargs, **kwargs):
-        raise NotImplementedError("See #1550")
+        if self._hasfired:
+            return
```
scopatz: This seems like the event is only fireable only once, not just that each handle is callable only once. Should there be a reset method or some such so that the event can be fired more than once.  I don't have a compelling use case, so this is a non-blocking suggestion.

Makefile
scopatz: They are listed in the setup.py

xonsh/events.py
astronouth7303: Well, if each handler is only callable once, how do you handle multiple firings? What does that even mean?

A reset method is easy (merge fired into unfired, reset fired, clear hasfired), but I'm going to claim YAGNI in this case. I can't imagine a time you'd need it except _maybe_ for testing.

xonsh/main.py
scopatz: I think it might be a little different (depending on when things are registered) because all of the `__xonsh_*__` builtins may be unloaded by the time atexit is called.  If there was a post cmdloop we would guarantee that the state was still sufficiently xonshy.

xonsh/events.py
scopatz: Very fair.

scopatz:

xonsh/main.py
astronouth7303: Fair enough. I'll add that.

update_ply.xsh
astronouth7303: We already rely on make to generate docs. GNU Make is not an onerous dependency on pretty much every environment. Hell, I think I accidentally got it on my Windows box from Git for Windows.

xonsh/ptk/shell.py
astronouth7303: It'll get a `#noqa` because of how documentation is generated.

update_ply.xsh
scopatz: Yeah, and since it is for devs only, it is not that big of a deal. You might want to make sure that there are no lingering references in the devs guide

xonsh/ptk/shell.py
scopatz: Sound good.

update_ply.xsh
astronouth7303: Nope, no references anywhere in the repo.

Makefile
```diff
@@ -0,0 +1,31 @@
+# Make GNU Make xonshy
+SHELL=xonsh
+.SHELLFLAGS=-c
+.ONESHELL:
+.SILENT:
```
scopatz: This is really cool, I had no idea about this!

scopatz:
This looks good to me! Thanks @astronouth7303!

Makefile
astronouth7303: GNU Make has some gems. I figured a xonshified Makefile would be greatly appreciated.

PR #1991, opened 2016-12-01
import xonsh.main to fix events docs generation

gforsyth:
sphinx will implicitly import main, but not before we call
`make_events()`, so only `on_chdir` shows up on the events listing.
explicitly importing `xonsh.main` fixes this up so all of the available
events are documented properly.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -18,6 +18,7 @@ from xonsh import __version__ as XONSH_VERSION
 from xonsh.environ import DEFAULT_DOCS, Env
 from xonsh.xontribs import xontrib_metadata
 from xonsh.events import events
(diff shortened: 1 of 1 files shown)


gforsyth:
no news item on this

astronouth7303:
Don't think a news item is required.

scopatz:
And the world grinds to a halt as travis macs images are down.

gforsyth:
Yeah... I think we could just cancel those mac jobs since there's no doc generation in them.  that only happens on the linux images on travis

scopatz:
done!

scopatz:
Thanks!

PR #1992, opened 2016-12-01
quick docs fix for events

gforsyth:
`on_post_rc` happens after rc files are loaded.  

no news entry and probably don't need to wait on CI, either.


scopatz:
Thanks!

PR #1993, opened 2016-12-02
Stdout storage

gforsyth:
somewhere along the way history went a little bit sideways and we started storing all STDOUT by default.  This can definitely be desirable in some situations but oppressive in others, so this restores respect of the `XONSH_STORE_STDOUT` envvar.


scopatz:

xonsh/base_shell.py
```diff
@@ -343,11 +343,14 @@ def _append_history(self, tee_out=None, **info):
...
         """
         hist = builtins.__xonsh_history__  # pylint: disable=no-member
+        env = builtins.__xonsh_env__
         info['rtn'] = hist.last_cmd_rtn
         tee_out = tee_out or None
         last_out = hist.last_cmd_out or None
         if last_out is None and tee_out is None:
             pass
+        elif not env.get('XONSH_STORE_STDOUT'):
+            pass
```
astronouth7303: If it only changes history? Probably, or the thing that sends data to history.

gforsyth: Clarifying question: we want `__xonsh_history__.outs` to remain accessible but only write those outs to file if `XONSH_STORE_STDOUT` is `True`?

scopatz: Yes, I think so.  I think we were muddling getting output and storing output before. The purpose of this envvar to me is to make / prevent history form storing output only.

I think there are consistency issues with prevent obtaining the output, for example:

```python
cat myfile
for line in _:
     ....
```

If we prevent obtaining output, this would only work sometimes. It seems reasonable to me that users should always have that work and not have to remember to set an env var (which is then stored!) just to have it work.  

Or if we do want to prevent obtaining output, that is a wholly independent env var.

mitnk: Well, I asked the quite same question here: https://github.com/xonsh/xonsh/pull/1970#discussion_r90779531

We are using `__xonsh_history__.rtns[-1]` as the counterpart of `$?` in Bash. Currently for json based history, if we set `ignoredups`, we will have an issue with it:

```
last-rtn[<not-set>]$ ls foo  # foo not exist, rtn is 1

# in another xonsh session, we create the file `foo` out.

last-rtn[1]$ ls foo  # foo now exists, rtn 0, but __xonsh_history__.rtns[-1] not got updated.
last-rtn[1]$ 
```
This is an issue because I think xonsh users usually use `__xonsh_history__.rtns[-1]` with colors in prompt string to know whether last cmd is succeed or not.

scopatz:
Sorry @gforsyth! This now has conflicts due to me merging in #1970

gforsyth:
No worries, going to have to redo it anyway. I figure we can wait until the
flush to file and then choose whether to strip the outs then.

gforsyth:
Ok, this is ready for another look.  I moved my changes to `json.py` in the new history backend.  @mitnk has already added a similar check into `sqlite.py`.

mitnk:
works for me.

gforsyth:
ping!

scopatz:
Thanks @gforsyth! Sorry about the delay!

Issue #2001, opened 2016-12-07
Support PEP 519


astronouth7303:
[PEP 519](https://www.python.org/dev/peps/pep-0519/) adds the `__fspath__()` special method (with the accompanying `PathLike` ABC and some utility functions) and will be available with 3.6 (now rc).

The following areas should be adjusted to accept `PathLike` objects:

* [x] `@()` substitution
* [x] Env vars that take paths
* [x] Vox

(Feel free to add to this list if I've forgotten something.)

astronouth7303:
My personal rationale: Using `pathlib` in my own Python work, I've found it much, _much_ more concise than `os.path` and generally easier to read and use.

In general, enabling xonsh to support them through out (eg #1862) will be useful. `Path` objects have really fantastic utility methods and bring together a lot of code from other places in the standard library: `.match()`, `.glob()`, `.read_{bytes,text}()`, and others. 

Given that a large chunk of a shell's job is working with files and file names, anything we can do to make this easier for users will be helpful.

laerus:
+1

scopatz:
I am for any changes folks want to make in this direction.

melund:
So am I. I haven't used the Path objects yet, but I will certainly try.

scopatz:
These have been implemeted!

PR #2004, opened 2016-12-10
PEP 519 for Vox

astronouth7303:
Part of #2001.

If Vox is given a `PathLike` (including a `pathlib.Path`), it will not use the user's venv store and assume it points to the location in the filesystem that the virtual env is.


gforsyth:
Hey @astronouth7303 -- there were some small flake issues and a weird looking docs failure that ended up just being a ref to the old location of `history_main` so I fixed them and pushed them up.

gforsyth:
Hmm... here's a weird one.  If I try to run `xonsh` after installing this branch, it crashes with this:

```console
  File "/home/gil/anaconda/bin/xonsh", line 3, in <module>
    main()
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 17832, in main
    shell.shell.cmdloop()
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 132, in cmdloop
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 71, in singleline
    prompt_tokens_cached = self.prompt_tokens(None)
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 154, in prompt_tokens
    self.settitle()
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 17509, in settitle
    with open(1, 'wb', closefd=False) as f:
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 229, in open
    return io.open(fspath(file), *pargs, **kwargs)
  File "/home/gil/anaconda/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 190, in fspath
    + path_type.__name__)
TypeError: expected str, bytes or os.PathLike object, not int
```

I tried with `XONSH_DEBUG` set so I could look at the code in question, and then it doesn't throw the error.  But it seems to be coming from `settitle` https://github.com/xonsh/xonsh/blob/vox-fs/xonsh/base_shell.py#L446

astronouth7303:
Oh, right. Amalgamation.

`fs.py` defines an `open()`.

xonsh/fs.py
```diff
@@ -0,0 +1,94 @@
+"""
```
scopatz: This is non-lazy in a lot of small ways. Anyway that you can make it more lazy, please?

scopatz:
Thanks for putting this in @astronouth7303! I think the only thing I would like to see is for fs.py to become lazy.

xonsh/fs.py
```diff
@@ -0,0 +1,94 @@
...
+                return filename.decode(encoding, errors)
+            else:
+                return filename
+
+        return fsencode, fsdecode
+
+    fsencode, fsdecode = _fscodec()
+    del _fscodec
+
+    def open(file, *pargs, **kwargs):
```
scopatz: Do we really need this function, it doesn't seem to be used.

scopatz: This is non-blocking, BTW

astronouth7303: It was modified as part of the PEP, so I included it.

I did have the thought of turning this into a package so it doesn't get amalgamated, and therefore the rest of the system has to opt in to it.

xonsh/fs.py
astronouth7303: Compiling a <100-line file is expensive? The only fresh import I'm seeing is `pathlib` (possibly, depending on how lazy #1862 is), and that becomes a non-issue as more of #2001 is implemented.

The only run-time code executed is trivial (register `Path` as a `PathLike`, and get the system encoding).

After the `pathlib` import, the most expensive thing I'm seeing is the actual parsing and compilation, which isn't something we can lazify. Unless defining an ABC is more involved than I thought?

I can move the other imports into the `except` block, although they should all be already loaded.

xonsh/fs.py
astronouth7303: I'm not seeing what you mean? Unless you mean "make it a package, and then don't add it to be amalgamated".

I'm more concerned with the fact that as it stands, this `open()` will shadow `builtin.open()` when amalgamation is on. It _shouldn't_ make a difference, but it still feels wrong.

scopatz: Oh weird, yeah, this shouldn't shadow builtins. I am definitely leaning towards not amalgamating this. I think it would solve a lot of my concerns and mirror how it will be used going forward, since we will eventually be grabbing these from os directly.

xonsh/fs.py
astronouth7303: Yeah, I was just copying what `os.py` does. The closure _is_excessive. Although I will still stand by my argument that no part of that is non-trivial. (ie, I think it's more of a memory excess than a CPU excess.)

xonsh/fs.py
astronouth7303: The only thing is that I expect it will eventually be used in a number of places in the core (#2001), so loading it with the amalgamation _will_ have benefits, but we will have more consistent and correct behavior if it's not in the same flat namespace.

scopatz:
Thanks @astronouth7303!

PR #2008, opened 2016-12-11
update info print of debug level

mitnk:
There are debug level `+-` outputs with commands like `history info` when `$XONSH_DEBUG = 1`. I believe these outputs should be printed with a higher debug level as the other places of `execer.py` related with `self.debug_level`.

no news items is needed.

### DIFF
--- a/xonsh/foreign_shells.py
+++ b/xonsh/foreign_shells.py
@@ -584,7 +584,7 @@ def load_foreign_aliases(shells=None, config=None, issue_warning=True):
             shaliases = {} if shaliases is None else shaliases
             for alias in set(shaliases) & set(xonsh_aliases):
                 del shaliases[alias]
-                if builtins.__xonsh_env__.get('XONSH_DEBUG', False) > 1:
+                if builtins.__xonsh_env__.get('XONSH_DEBUG') > 1:
                     print('aliases: ignoring alias {!r} of shell {!r} '
                           'which tries to override xonsh alias.'
                           ''.format(alias, shell['shell']),
--- a/xonsh/foreign_shells.py
+++ b/xonsh/foreign_shells.py
@@ -584,7 +584,7 @@ def load_foreign_aliases(shells=None, config=None, issue_warning=True):
             shaliases = {} if shaliases is None else shaliases
             for alias in set(shaliases) & set(xonsh_aliases):
                 del shaliases[alias]
-                if builtins.__xonsh_env__.get('XONSH_DEBUG'):
+                if builtins.__xonsh_env__.get('XONSH_DEBUG', False) > 1:
                     print('aliases: ignoring alias {!r} of shell {!r} '
                           'which tries to override xonsh alias.'
                           ''.format(alias, shell['shell']),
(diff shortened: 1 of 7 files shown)


scopatz:
Hi @mitnk, thanks for submitting. I kinda think that these should be available at debug level 1, since they tell you how the code transforms.  We don't really have a standard meaning for the different debug levels. Maybe we should? 

The reason we put these at the base debug level, unlike other parts of the execer, is knowing how xonsh is transforming the input it gets seems like the most basic form of debugging.  Maybe there is another way of thinking about this that is more logical. But that was the idea.

mitnk:
I put it here is that - when I do not care about speed and I want amalgamate disabled but not printing anything debug info out. In this case, I can use xonsh as normal and will see better traceback in case there are some errors raised.

So how about make `1` do the thing I want and showing more debug info (like input transforming) with a `2` or higher?

mitnk:
Or do we want a separated dedicated ENV for disabling amalgamate? :)

scopatz:
I guess I am OK with that. Can you please add something to the docs though, about what the different debug levels mean though? And maybe change the env var to validate to type int? Right now, I often set `$XONSH_DEBUG=True` and, since that evaluates to 1, my behaviour (and possible that of users) will need to change. Sorry if this is getting beyond a one line change....

mitnk:
Since in Python bool is comparable with integer, so I think we don't need it to be int, and thus not breaking your RC.

But by defining `1` to be disable amalgamate has some side effects. Before with debug level `1` will let inputs like `!!` (with bashisms enalbed) or `history info`, we will get some diff like information. Since we now increase it to `2`, and `2` was the level when we will print a lot of PLY parsing debug messages. So I also increase that level to `3` in last commit. Ideas on this?

xonsh/foreign_shells.py
```diff
@@ -584,7 +584,7 @@ def load_foreign_aliases(shells=None, config=None, issue_warning=True):
             shaliases = {} if shaliases is None else shaliases
             for alias in set(shaliases) & set(xonsh_aliases):
                 del shaliases[alias]
-                if builtins.__xonsh_env__.get('XONSH_DEBUG'):
+                if builtins.__xonsh_env__.get('XONSH_DEBUG', False) > 1:
```
scopatz: You shouldn't need the `False` part here. The env will fill in with its own default.

xonsh/shell.py
```diff
@@ -49,7 +49,8 @@ def fire_precommand(src, show_diff=True):
             print_exception('Modifcations to source input took more than '
                             'the recursion limit number of interations to '
                             'converge.')
-    if show_diff and builtins.__xonsh_env__.get('XONSH_DEBUG') and src != raw:
+    debug_level = builtins.__xonsh_env__.get('XONSH_DEBUG', 0)
```
scopatz: Same, you shouldn't need the 0 here.

scopatz:
Hi @mitnk, I am happy with anything as long as there is documentation :)

mitnk:
Thanks for the comments. Updated.

scopatz:
Thanks again @mitnk!

PR #2014, opened 2016-12-13
stop travis if build fails

gforsyth:
@asmeurer reported that travis doesn't stop a build if the docs fail to build unless there's a `set -e` in there.  Adding it in so that we don't accidentally push broken docs, potentially deleting what we have.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,8 @@ install:
       pip install -r requirements-tests.txt;
     fi
 
+before_script:
+  - rvm get head || true
 
 script:
   - set -e
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ install:
 
 
 script:
+  - set -e
   - if [[ $BUILD_DOCS = true ]]; then
       cd docs;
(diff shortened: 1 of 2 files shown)


scopatz:
LGTM, thanks @gforsyth and @asmeurer!

scopatz:
I think that the travis error `/Users/travis/build.sh: line 148: shell_session_update: command not found` might be spurious. @gforsyth, any thoughts?

gforsyth:
That does seem spurious, although the error in the docs is legit, I think?

scopatz:
@gforsyth are we comfortable with merging this now that the docs build?

gforsyth:
I don't think that I am.  I agree that the error is spurious but I'd like to track it down so we don't have every PR failing (especially since the failing tests we'll wait for are the OSX ones that take forever anyway)

scopatz:
That is fair!

gforsyth:
Woohoo! Fixed it. This should be good to go.

scopatz:
Sweet!

PR #2016, opened 2016-12-15
added view to unthreadable

### LINKED ISSUE #2002: Troublesome command entering python mode
I'm trying to run this command which I used to use often in zsh:
```
svn diff ./test.py | vim -R -
```
but it keeps trying to evaluate in Python mode. 

I can usually wrestle these types of troublesome commands into submission with `$()` and `@()` but I can't figure this one out.

scopatz:
This probably won't fix #2002, but it seems like the right thing to do.


gforsyth:
Couldn't hurt, certainly.  Thanks, @scopatz!

PR #2018, opened 2016-12-15
Xompletions

gforsyth:
Adds completions for `xonfig` and `xontrib`


xonsh/completers/xompletions.py
```diff
@@ -0,0 +1,46 @@
+import xonsh.xontribs as xx
```
scopatz: Can you add a module docstring, please? Also can you add a sphinx autodoc hook for this module too, please? Thanks!

xonsh/completers/xompletions.py
```diff
@@ -0,0 +1,46 @@
...
+from xonsh.tools import color_style_names
```
scopatz: This should be `import xonsh.tools as xt` to get the benefits of lazy module amalgamation.

scopatz:
Thanks @gforsyth! Just a couple of small issues. Really looking forward to this!

gforsyth:
Hey @scopatz -- fixed the issues you raised.  Also an unrelated fix I tacked on for the events docs generation -- was raising a bunch of warnings in Sphinx

scopatz:
Looks great, thanks!

Issue #1525, opened 2016-08-03
Suggestion: Add fallback shell


derphilipp:
As xonsh fails if a configuration error exists, it would be nice to define a fallback shell that would be used if xonsh crashes / can not start.

This could lead to more adopters being able to set xonsh as their default shell.

xsteadfastx:
Thats a good idea. I had the problem when my filesystem got remounted read-only and it couldnt write the history file no more. no chance to login. happily i had a ssh server open and could `ssh foo@bar /bin/sh` into it.

scopatz:
I am definitely in favor of this idea. It just needs an implementation.  PRs welcome!

PaulReiber:
Logistically, one can't guarantee a fallback would even work - the fallback
shell could easily be affected by the same problem Xonsh faced.

Scope/responsibility wise, the OS defines the users shell - it's there, not
within Xonsh, that one might introduce a fallback feature/capability.
Interesting idea, but it'd be down to init/login/getty to do the right
things when a user shell gets hung - not the user shell itself.

It's easy enough to ensure a user can login via another shell.

A Xonsh user can have multiple entries in /etc/passwd to ensure they can
always login with another shell:

pbr:x:1000:1000:Reiber Labs,,,:/home/pbr:/usr/local/bin/xonsh
pbrbash:x:1000:1000:Reiber Labs,,,:/home/pbr:/bin/bash

Note the UID and home directory are the same - these are effectively the
same user, just logging in using different shells.

Kind regards,


*Paul Reiber*http://about.me/reiber

scopatz:
> Logistically, one can't guarantee a fallback would even work - the fallback
shell could easily be affected by the same problem Xonsh faced.

I agree that this isn't about gaurantees.  Rather it is about "Oh no!" moments and debugging.  If dropping into a backup shell works, great! If not, then things were broken already.  The backup shell would still be run from within the xonsh process, so I don't see this as stepping on the toes of the OS.

asmeurer:
Most terminal emulators let you define different profiles that start different shells. Maybe this is useful for ssh, though?

asmeurer:
Also, I'd argue that xonsh already does have a backup shell, namely, itself. If .xonshrc can't be loaded, it falls back to xonsh with no configuration.

scopatz:
[...]

PR #2020, opened 2016-12-17
Make xonsh able to failback to other shells

mitnk:
Try addressing #1525

### DIFF
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -263,11 +263,15 @@ def main(argv=None):
         shells_file = '/etc/shells'
         if not os.path.exists(shells_file):
             return
+        excluded_list = ['xonsh', 'screen']
         with open(shells_file) as f:
             for line in f:
                 if not line.strip() or line.strip().startswith('#'):
                     continue
-                if line.strip().endswith('xonsh'):
+                if '/' not in line:
+                    continue
+                _, shell = line.strip().rsplit('/', 1)
+                if shell in excluded_list:
                     continue
                 foreign_shell = line.strip()
                 break
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -253,22 +253,40 @@ def premain(argv=None):
     return args
 
 
-def _failback_to_other_shells():
+def _failback_to_other_shells(argv):
+    args = None
+    try:
+        args = premain(argv)
+    except Exception:
+        pass
+    # only failback for interactive shell; if we cannot tell, treat it
+    # as an interactive one for safe.
+    if hasattr(args, 'mode') and args.mode != XonshMode.interactive:
+        sys.exit(1)
+        # put a harmless verbose `return` here so that unit test is able to
+        # break here when mock up `sys.exit()`
+        return
+
+    print('Xonsh encountered an issue during launch', file=sys.stderr)
     foreign_shell = None
     shells_file = '/etc/shells'
     if not os.path.exists(shells_file):
+        # right now, it will always return here on Windows
         return
     excluded_list = ['xonsh', 'screen']
     with open(shells_file) as f:
         for line in f:
-            if not line.strip() or line.strip().startswith('#'):
+            line = line.strip()
+            if not line or line.startswith('#'):
                 continue
             if '/' not in line:
                 continue
-            _, shell = line.strip().rsplit('/', 1)
+            _, shell = line.rsplit('/', 1)
             if shell in excluded_list:
                 continue
-            foreign_shell = line.strip()
+            if not os.path.exists(line):
+                continue
+            foreign_shell = line
             break
     if foreign_shell:
         print('Failback to {}'.format(foreign_shell), file=sys.stderr)
@@ -279,17 +297,10 @@ def main(argv=None):
     try:
         return main_xonsh(argv)
     except Exception:
+        # use traceback module here instead of xonsh.tools.print_exception()
+        # to gain a bit more stability.
         traceback.print_exc()
-        args = None
-        try:
-            args = premain(argv)
-        except Exception:
-            pass
-        if hasattr(args, 'mode') and args.mode != XonshMode.interactive:
-            sys.exit(1)
-        else:
-            print('Xonsh encountered an issue during launch', file=sys.stderr)
-            _failback_to_other_shells()
+        _failback_to_other_shells(argv)
 
 
 def main_xonsh(argv=None):
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -267,6 +267,8 @@ def main(argv=None):
             for line in f:
                 if not line.strip() or line.strip().startswith('#'):
                     continue
+                if line.strip().endswith('xonsh'):
+                    continue
                 foreign_shell = line.strip()
                 break
         if foreign_shell:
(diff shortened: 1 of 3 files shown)


melund:
Cool. I could be useful on Windows in some cases where xonsh is started within other programs. There is no 'shells' file on Windows but maybe you could make it fall back to 'cmd.exe' if it happened here.

mitnk:
I'm not sure how to do with Windows. I thought Windows maybe need the fail-back, only a headless server could be unrecoverable with xonsh as default and errors occurred? Windows should have other ways to update the filesystem to fix the issue.

I don't have a Windows to use and don't know whether `os.execlp()` works on it.

astronouth7303:
I think this is only the correct behavior for interactive invocations.

xonsh/main.py
```diff
@@ -253,6 +254,33 @@ def premain(argv=None):
 
 
 def main(argv=None):
+    try:
+        return main_xonsh(argv)
+    except Exception:
+        traceback.print_exc()
```
scopatz: You should probably use `xonsh.tools.print_exception()`

xonsh/main.py
```diff
@@ -253,6 +254,33 @@ def premain(argv=None):
...
+                if '/' not in line:
+                    continue
+                _, shell = line.strip().rsplit('/', 1)
+                if shell in excluded_list:
+                    continue
+                foreign_shell = line.strip()
+                break
+        if foreign_shell:
+            print('Failback to {}'.format(foreign_shell), file=sys.stderr)
+            os.execlp(foreign_shell, foreign_shell)
```
scopatz: It would be good if the exception body here were broken out into its own function.  try-excepts should only be a couple of lines long, ideally.

scopatz:
I agree with @melund and @astronouth7303 that this should be only for interactive mode and have better cross-platform support.  Thanks for tackling this, @mitnk!

xonsh/main.py
mitnk: How should I know xt is still available? after all it's doing fail over. So I guess use `traceback` module is safer?

mitnk:
Thanks @astronouth7303 for the interactive comment. Updated.

About the Windows, how about a following up PR (probably by somebody else) to do it. I do not have a Windows for testing :)

xonsh/main.py
scopatz: If you import it at the top, you should still have a reference to it.  But I think I see what you are saying, that xonsh builtins might not be around anymore and so `xt.print_exception()` could itself fail.  That is fair.

xonsh/main.py
```diff
@@ -253,37 +253,47 @@ def premain(argv=None):
...
-                if shell in excluded_list:
-                    continue
-                foreign_shell = line.strip()
-                break
-        if foreign_shell:
-            print('Failback to {}'.format(foreign_shell), file=sys.stderr)
-            os.execlp(foreign_shell, foreign_shell)
+        args = None
+        try:
+            args = premain(argv)
```
astronouth7303: `premain()` is not side-effect free, but I'm not sure it matters since we're about to bail?

scopatz:
> About the Windows, how about a following up PR (probably by somebody else) to do it. I do not have a Windows for testing :)

Well, I think this needs to be disabled on windows at the very least. The problem is that if it is enabled as it is, it just won't work.  Notice that AppVeyor is failing.  I am happy if an implementation for windows comes later.

mitnk:
The CI failure is due to my last commit. Checking it. But the code did have "be disabled on Windows" with `if os.path.exists()`.

xonsh/main.py
```diff
@@ -252,10 +253,47 @@ def premain(argv=None):
...
 
+def _failback_to_other_shells():
+    foreign_shell = None
+    shells_file = '/etc/shells'
+    if not os.path.exists(shells_file):
+        return
+    excluded_list = ['xonsh', 'screen']
+    with open(shells_file) as f:
+        for line in f:
+            if not line.strip() or line.strip().startswith('#'):
```
scopatz: Only should strip the line once.

mitnk:
It's ready for another look.

mitnk:
@melund How people are using shells at Windows? I just could fire a cmd up and have to type xonsh to start a xonsh - in this case fail-back is not necessary. Maybe you are using some featured terminals so that they can set xonsh as default shell when start up?

I've tried (within a cmd) `os.exexx` on a Windows, but seems not work as expected. Maybe you could help it out for Windows (if it does need a fail-back) in future PRs, or even send send a PR to this one?

xonsh/main.py
```diff
@@ -252,10 +253,58 @@ def premain(argv=None):
...
+def _failback_to_other_shells(argv):
+    args = None
+    try:
+        args = premain(argv)
+    except Exception:
+        pass
+    # only failback for interactive shell; if we cannot tell, treat it
+    # as an interactive one for safe.
+    if hasattr(args, 'mode') and args.mode != XonshMode.interactive:
+        sys.exit(1)
```
scopatz: shouldn't this reraise the exception above?

xonsh/main.py
```diff
@@ -252,10 +253,58 @@ def premain(argv=None):
...
+            _, shell = line.rsplit('/', 1)
+            if shell in excluded_list:
+                continue
+            if not os.path.exists(line):
+                continue
+            foreign_shell = line
+            break
+    if foreign_shell:
+        print('Failback to {}'.format(foreign_shell), file=sys.stderr)
+        os.execlp(foreign_shell, foreign_shell)
```
scopatz: Also, I think there should be ab else clause that reraises the exception above.

xonsh/main.py
mitnk: Since `premain()` is the first expression (from code or logic phases) in `main_xonsh()`, so if it raises errors the traceback is already printed; if not, then there should be no exceptions too.  That's my idea here, I can add reraise here is you still think the reason I just put is not valid enough :)

xonsh/main.py
mitnk: same reason here I guess. what errors should printed should have already printed.

xonsh/main.py
scopatz: That is a fair point. But maybe we should just rely on the return below and and call sys.exit().  sys.exit is problematic if you want to call the main function programtically - which is why I suggested reraising. I guess I think we should just return then.  Really, I just want to avoid the sys.exit if we can.

mitnk: Ah, I didn't know `sys.exit(1)` is not suitable here .. and what the concern about it? I use sys.exit so that xonsh can exit with code 1. I have checked `return 1` cannot made the exit code to 1.

I'm ok with a just return without updating exit code - tho that will make `xonsh -c 'list(?'` exited with exit code 0.

astronouth7303: Raising an exception will cause a non-zero process return.

mitnk: sys.exit [is actually raising exception](https://docs.python.org/3.6/library/sys.html#sys.exit). But I guess raising the original one is better here. will update it later. thanks!

mitnk:
@scopatz Updated. New code seemed turn out exactly what you have suggested LOL..

scopatz:
@mitnk hahah, awesome ;)

PR #2025, opened 2016-12-21
Fixed xonfig raising error when xonsh is not installed from source

mitnk:
Found xonfig is broken when I was trying 0.5.0 in a clean env..

Sorry @scopatz, maybe you could do a 0.5.1 release for it soon?

### DIFF
--- a/tests/test_platform.py
+++ b/tests/test_platform.py
@@ -1,0 +1,15 @@
+from contextlib import contextmanager
+from unittest.mock import MagicMock
+
+import builtins
+import xonsh.platform as xp
+
+
+def test_githash_value_error(monkeypatch):
+    @contextmanager
+    def mocked_open(*args):
+        yield MagicMock(read=lambda: 'abc123')
+    monkeypatch.setattr(builtins, 'open', mocked_open)
+    sha, date_ = xp.githash()
+    assert date_ is None
+    assert sha is None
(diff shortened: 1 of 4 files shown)


tests/test_platform.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def test_githash_value_error(monkeypatch):
+    @contextmanager
+    def mocked_open(*args):
+        yield MagicMock(read=lambda: 'abc123')
+    monkeypatch.setattr(builtins, 'open', mocked_open)
+    sha, date_ = xp.githash()
+    assert date_ is None
+    assert sha is None
```
mitnk: Tried to add a normal case, but haven't found a way to mock up `functools.lru_cache`. So only bad case added.

gforsyth:
Thanks, @mitnk! I'll try to push a microrelease but I'm on _SLOW_ internet and might have to have one of the other core devs do it

mitnk:
I think it's okay, not a urgent one :) And better this release could include this one https://github.com/xonsh/xonsh/pull/2028

gforsyth:
well, too late, but we can always do another one

gforsyth:
and I, for one, would rather `xonfig` were working after a major release since it's the first thing we ask of users reporting issues

mitnk:
haha. agreed. that one is really minor compare to xonfig. Thank you!

PR #2027, opened 2016-12-21
Tests python shebangs

chronitis:
These files contain python3 syntax, which causes the associated tests to fail on a system for which `/usr/bin/python` is python2.


scopatz:
Thanks for the fix!

PR #2029, opened 2016-12-21
use doctr to also manage release docs

gforsyth:
now that we have a release that contains the encrypted deploy key, we
can also have doctr always push the most recent release docs to
`xonsh-docs`.

there may be more fiddling to do to get a documentation selector
working but this is a start

Usual caveat that we won't know if this is working until it's merged, but if anyone spots any obvious problems in the yaml logic, let me know.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,13 +53,12 @@ script:
       make html;
       cd ..;
       doctr deploy --deploy-repo xonsh/xonsh-docs --gh-pages-docs dev;
+      git checkout $(git describe --tags `git rev-list --tags --max-count=1`);
+      cd docs;
+      make html;
+      cd ..;
+      doctr deploy --deploy-repo xonsh/xonsh-docs --gh-pages-docs .;
     else
       py.test --timeout=10;
     fi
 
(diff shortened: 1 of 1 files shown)


astronouth7303:
Go for it and see if it breaks something?

gforsyth:
No time like the present

.travis.yml
```diff
@@ -53,13 +53,12 @@ script:
       make html;
       cd ..;
       doctr deploy --deploy-repo xonsh/xonsh-docs --gh-pages-docs dev;
+      git checkout $(git describe --tags `git rev-list --tags --max-count=1`);
+      cd docs;
+      make html;
```
asmeurer: You might want to put a `make clean` here to clear out any cruft from the dev build above.

scopatz:
Thanks for this @gforsyth! Just the one small change @asmeurer suggested.

gforsyth:
Good call, @asmeurer. Thanks.  Updated.

scopatz:
Thanks @gforsyth! Also, I am not sure if you avoided this on purpose, but you can also do `make clean html`

PR #2032, opened 2016-12-21
POC - hot pluggable completers

mitnk:
I think instead of alway implementing a next completer and add it to xonsh codebase, we can make completers writing easy as bash completion scripts. You can put as many as new scripts as your want, bash will load them all. Program author then could pack a xonsh-myapp.py as its completer script for xonsh like they will do for bash.

This PR is a quick & dirty POC, without safe check or anything. Just put it here to collect ideas.

With a PR like this, xonsh users can drop their completer files into `~/.xonsh/completers/`, and they will just work. This will be even convenient than xontrib. Demo files would be:

file: **~/.xonsh/completers/foo.py**:
```
def complete(prefix, line, begidx, endidx, ctx):
    if not line.startswith('foo'):
        return
    L = ['bar', 'baz']
    return {x for x in L if x.startswith(prefix)}
```

file: **~/.xonsh/completers/brew.py**:
```
def complete(prefix, line, begidx, endidx, ctx):
    if not line.startswith('brew'):
        return
    L = ['install', 'info', 'uninstall']
    return {x for x in L if x.startswith(prefix)}
```

### DIFF
--- a/xonsh/completers/__init__.py
+++ b/xonsh/completers/__init__.py
@@ -14,6 +14,8 @@ else:
         _sys.modules['xonsh.completers.tools'] = __amalgam__
         xompletions = __amalgam__
         _sys.modules['xonsh.completers.xompletions'] = __amalgam__
+        xonsh = __amalgam__
+        _sys.modules['xonsh.completers.xonsh'] = __amalgam__
         _aliases = __amalgam__
         _sys.modules['xonsh.completers._aliases'] = __amalgam__
         commands = __amalgam__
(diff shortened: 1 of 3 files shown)


astronouth7303:
Ideally, completers would live in xontribs.

scopatz:
Thanks for putting this in, @mitnk! I think we might want to avoid having too many ways to load files. If a user wanted this behaviour, couldn't they add this into one of their xonshrc files?  Then if they wanted to share this, the right place for this would be a xontrib.  

I can see how the workflow presented here can be more convenient, potentially, than xonshrc files. So maybe this itself could be a xontrib? What do you think?

mitnk:
Thanks for the comments. Make sense to me.

PR #2039, opened 2016-12-23
checkout HEAD before attempting second deploy

gforsyth:
I think the errors that doctr encounters when pushing the second set of
docs has to do with trying to add to the gh-pages branch when we're
starting from non-HEAD location?

I'm hoping that by manually putting us back onto HEAD that this should
avoid that...

If this does work, it's probably worth putting into doctr itself.

Usually caveats that we can't test this until it's merged...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,9 +55,9 @@ script:
       doctr deploy --deploy-repo xonsh/xonsh-docs --gh-pages-docs dev;
       git checkout $(git describe --tags `git rev-list --tags --max-count=1`);
       cd docs;
-      make clean;
-      make html;
+      make clean html;
       cd ..;
+      git checkout HEAD;
       doctr deploy --deploy-repo xonsh/xonsh-docs --gh-pages-docs .;
     else
       py.test --timeout=10;


scopatz:
Let's try it! Thanks so much @gforsyth!

asmeurer:
Hmm, I remember putting code in doctr to check out the old HEAD again. Is it not working?

asmeurer:
Specifically [this](https://github.com/drdoctr/doctr/blob/4be34bbc86488248ba5067ecc1142e4989179253/doctr/__main__.py#L161).

gforsyth:
Sorry for the delay -- travel and holidays and whatnot. 

@asmeurer -- I didn't see that in `doctr` before, but it also isn't working as expected (I don't think). 
I wasn't aware of this behavior of `git`:
1. get current hash of HEAD using `git rev-parse HEAD`
2. checkout previous commit/tag or whatever
3. checkout old head using hash from step 1
4. end up with HEAD detached at the correct hash

I'm going to try to figure out how to return to HEAD without the detaching bit, since I think that's related to part of the trouble here.

asmeurer:
Yeah if you checkout a hash it detaches the HEAD. I didn't think it would be a problem, since we aren't going to be making any commits in Travis.

gforsyth:
Ok, I think I have this figured out? It will require a small change to `doctr`. If you look at the log from one fo the failed doc builds on xonsh master (https://travis-ci.org/xonsh/xonsh/jobs/187278382#L1872) you can see that `doctr` reports it is doing a `git pull` but what that is doing is a recursive merge of the entire xonsh repo into the local (to travis) `gh-pages` branch.  Then when it tries to push that things go very wrong.

I don't know why it's pulling in xonsh master, since that isn't even the correct remote, but if we change https://github.com/drdoctr/doctr/blob/master/doctr/travis.py#L382 to explicitly do `git pull doctr_remote gh-pages` then I _think_ that should avoid this?

asmeurer:
That seems like a reasonable change.

PR #2040, opened 2016-12-24
Minor updates on source alias

mitnk:
When we write xonsh scripts, we may write something like this:
```
# file foobar.xsh
...
xonshrc_local = '~/.xonshrc_local'
source xonshrc_local
...
```
Current implementation will make it parsing as `source 'xonshrc_local'`, and try to source the content of file 'xonshrc_local' in current dir & $PATH.

In this PR, it would make it parsing as `source the value string of xonshrc_local python var`, and fail back to source file in $PATH. And the context take priority here is because in xonsh, we always say Python first and then subprocess, IIRC.

I'd like to hear from you about this. If it's ok to go, then I will add some tests and docs etc.


scopatz:
Hi @mitnk, is there some reason that `source @(xonshrc_local)` doesn't work? I like this way a little more because it makes it explicit that the user is going into Python mode. 

Something like this would probably be OK with me (but others should absolutely chime in) if there was a generic way to declare (on the alias?) that a subprocess argument could also come from the context.  This is probably relatively easy to apply to all arguments but hard to apply to a single, arbitrary argument. If we had a generic way to do this, then you could just apply it to the source alias. 

How do other people feel about this?  It seems like it could cause some errors-at-a-distance. Do we just need to figure out a way to make the Python-mode `@(expr)` operator more concise?

mitnk:
I must haven't got used to `@()` grammars :) Yes, `source @(xonshrc_local)` work fine and is more explicit here. Sorry for interrupting.

Then I will update this PR for some minor changes for the source alias. Commits coming.

mitnk:
It' ready for another look for a different fix ..

scopatz:
LGTM, thanks!

PR #2043, opened 2016-12-29
Improved bash completions

nd-net:
Python 3.6 breaks bash completions because now os.path.commonprefix expects an indexable argument.

This change fixes that and extends the functionality of $BASH_COMPLETIONS:
* Added tests for the bash completer
* All valid files in $BASH_COMPLETIONS are used instead of just the first one
* Paths to directories cause the whole directory to be imported

### DIFF
--- a/news/improved-bash-completions.rst
+++ b/news/improved-bash-completions.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fixed a bug where the bash completer raised an exception when using Python 3.6.
+
+**Security:** None
+
--- a/news/improved-bash-completions.rst
+++ b/news/improved-bash-completions.rst
@@ -13,4 +13,3 @@
 * Fixed a bug where the bash completer raised an exception when using Python 3.6.
 
 **Security:** None
-
--- a/news/improved-bash-completions.rst
+++ b/news/improved-bash-completions.rst
@@ -1,4 +1,6 @@
-**Added:** None
+**Added:**
+
+* The ``$BASH_COMPLETIONS`` environment variable now supports multiple valid paths.
 
 **Changed:** None
 
--- a/news/improved-bash-completions.rst
+++ b/news/improved-bash-completions.rst
@@ -1,6 +1,6 @@
 **Added:**
 
-* The ``$BASH_COMPLETIONS`` environment variable now supports multiple valid paths.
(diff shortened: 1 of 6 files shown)


gforsyth:
Hi @nd-net -- thanks for putting this in! Python 3.6 fixes are a great thing to have. I'm a little concerned about the directory expansion bit -- I think that it might hurt performance since we won't be taking advantage of the lazy loading built in to completions 2.x

Maybe @blahgeek has thoughts / input on this?

gforsyth:
It may be enough to make the documentation more explicit in re: speed tradeoffs when using directories to specify bash completions

scopatz:
Thanks @nd-net! I agree with @gforsyth.  I think we need a way if detecting (or the user specifying) if there are v1 or v2 completions and then supporting them in different code paths.  There doesn't seem to be a one-size-fits-all solutions and enough people are using v1 still. 

@nd-net would you mind either (1) trying to tackle the v1-v2 distinction here, or (2) reducing this to just the Python v3.6 fixes? Thanks a ton!

nd-net:
I need to investigate how to differentiate between bash completions v1 and v2 sensibly, but I will look into it and report back.

How should xonsh deal with a v1 completion? Have a config setting or environment variable that explicitly enables v1 completions? Or logging the use of v1 completions somewhere (maybe stderr)?

In the meantime, if you just want the Python v3.6 fixes: that is the initial commit of the branch: https://github.com/nd-net/xonsh/commit/a8cc49c5cecf26b305bcd2c357db280d644c2500 - it's only a ``list(…)`` call around a ``set``, but the commit also contains an initial unit test for the fix.

scopatz:
> I need to investigate how to differentiate between bash completions v1 and v2 sensibly, but I will look into it and report back.

Thanks!

> How should xonsh deal with a v1 completion? Have a config setting or environment variable that explicitly enables v1 completions? Or logging the use of v1 completions somewhere (maybe stderr)?

Any of those seem reasonable.  Also, just opaquely handling it would be fine too.  As for the code itself, what is in xonsh now is for v2, and what you and @mitnk have both recently proposed is for v1, just to be clear.  

Also, would you mind please enabling CiricleCI fr your fork of xonsh? That way our linux test suite will run too. Thanks!

scopatz:
Also, please let us know when this is ready for another look.

scopatz:
This now has conflicts given the recent updates to bash / bash_completions.

scopatz:
Pinging this @nd-net!

gforsyth:

Issue #2036, opened 2016-12-23
multiline SyntaxError


laerus:
this appears to happen only with prompt_toolkit
```
$ print(
. 
$ <--- hitting return here
  File "<string>", line None
SyntaxError: <xonsh-code>:0:0: EOF in multi-line statement
$ 
```

scopatz:
Thanks for reporting @laerus.  I am not sure what is going on here, but I haven't done much work on the PTK side of things...

gforsyth:
hey @laerus, can you see if #2045 fixes this for you?

gforsyth:
Fixed in #2045.

PR #2045, opened 2016-12-31
use ptk-specific `push` when executing commands

gforsyth:
When events were introduced, the PTK shell's `_push` was left behind.
This is the cause of #2036 as the buffer isn't cleared between separate
prompt inputs, leading to very, very strange behavior.

This restores the use of the PTK `_push` that properly resets the buffer
if the AST fails to compile and also adds in the precommand src
transformation so that the bashisms xontrib still works as expected.

### DIFF
--- a/news/fix_multiline_execution.rst
+++ b/news/fix_multiline_execution.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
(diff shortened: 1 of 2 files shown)


scopatz:
LGTM, @gforsyth! Should we wait for @laerus, or should this just go in?

gforsyth:
I think it's good to go.  I could reproduce the error @laerus found and this fixed it on my end.

scopatz:
Sounds good, thanks!

Issue #1887, opened 2016-10-25
Tab completion fails on Python 3.6 due to PEP 519 changes


jonathanhogg:
One for the future: tab completion throws exceptions on Python 3.6b2 because of changes made in `os.path.commonprefix()` as part of PEP 519 (file system path-like objects). It appears that a `set()` is being passed into this somewhere (no stack trace handy, but will re-create and post one after this message) and `commonprefix()` now requires an indexable object – in fairness, the 3.5.2 docs do state that this function takes a list.

I opened a Python bug on the regression, but they don't seem to be interested.

http://bugs.python.org/issue28527

jonathanhogg:
Promised traceback:

```
Traceback (most recent call last):
  File "/opt/python36/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/opt/python36/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/python36/lib/python3.6/site-packages/prompt_toolkit/interface.py", line 841, in run
    completions = list(buffer.completer.get_completions(document, complete_event))
  File "/opt/python36/lib/python3.6/site-packages/xonsh/ptk/completer.py", line 42, in get_completions
    self.ctx)
  File "/opt/python36/lib/python3.6/site-packages/xonsh/completer.py", line 36, in complete
    out = func(prefix, line, begidx, endidx, ctx)
  File "/opt/python36/lib/python3.6/site-packages/xonsh/completers/bash.py", line 85, in complete_from_bash
    commprefix = os.path.commonprefix(out)
  File "/opt/python36/lib/python3.6/genericpath.py", line 76, in commonprefix
    if not isinstance(m[0], (list, tuple)):
TypeError: 'set' object does not support indexing
```

jonathanhogg:
Sufficient – but slightly displeasing – fix:

```
*** bash.py.orig	2016-10-25 14:54:26.440015117 +0100
--- bash.py	2016-10-25 14:53:36.120445475 +0100
***************
*** 82,88 ****
  
      # From GNU Bash document: The results of the expansion are prefix-matched
      # against the word being completed
!     commprefix = os.path.commonprefix(out)
      strip_len = 0
      while strip_len < len(prefix):
          if commprefix.startswith(prefix[strip_len:]):
--- 82,88 ----
  
      # From GNU Bash document: The results of the expansion are prefix-matched
      # against the word being completed
[...]

PR #2047, opened 2017-01-01
Ensure input to `commonprefix` is a list (now required by Python 3.6)

jonathanhogg:
As per issue #1887.


scopatz:
Sorry for the delay here, @jonathanhogg! Thanks a ton for this.

PR #2050, opened 2017-01-03
escape backslash in Tutorial warning

das-g:
There is no notion of an escaping character in xonsh, but there is in reStructuredText


laerus:
Thank you @das-g !

PR #2051, opened 2017-01-03
Dynamic cwd elision char

jb-leger:
By default the path as '/aaaaa/bbbbb/ccccc/ddddd' is shortened to obtain e.g.:
'/aaa/bbb/ccc/ddddd' which can be confusion.

The first commit add a DYNAMIC_CWD_ELISION_CHAR env variable which default value
empty string, keeping the same behavior described before.

Setting DYNAMIC_CWD_ELISION_CHAR to '…', for the same path, we obtain
'/aa…/bb…/cc…/ddddd'.

The second commit use the DYNAMIC_CWD_ELISION_CHAR when non empty for truncature. The default behavior is not changed.


astronouth7303:
Odd. CircleCI appears to be missing.

scopatz:
Yeah, Circle needs to be [enabled on all forks now](http://conda-forge.github.io/docs/ciservices.html#enabling-circle-on-your-fork). Could you please do this @jb-leger? Thanks for putting this in!

jb-leger:
@scopatz: I did. Do I need to do something more to have the check on this PR? (I am a git user, not a github one).

scopatz:
There are instructions on that link, but basically you need to go to CircleCI's website and enable builds for your fork of xonsh. After that, you just need to make a push to this PR to trigger a build on circle.  We use Circle to test Linux, which is why it would be good for you to do this.  Note that you only have to enable it once and it will affect all future PRs

astronouth7303:
@scopatz We should mention CircleCI in the contributor docs.

xonsh/environ.py
```diff
@@ -246,6 +247,7 @@ def DEFAULT_VALUES():
         'COMPLETIONS_MENU_ROWS': 5,
         'DIRSTACK_SIZE': 20,
         'DYNAMIC_CWD_WIDTH': (float('inf'), 'c'),
+        'DYNAMIC_CWD_ELISION_CHAR': '',
```
scopatz: Can you please add VarDocs for this below? Thanks!

scopatz:
@astronouth7303 yes we should :).

jb-leger:
@scopatz: it was the last step I did'nt know for running CircleCI, i.e. push again on the branch. Done now.

@scopatz: I have added VarDocs (dans squashed the 3 documentation related commits in one)

@astronouth7303: Is VarDocs a kind of "doc string that will eventually end up in the envvar reference"?

scopatz:
Great! Thanks!

PR #2053, opened 2017-01-03
Tutorial: Rephrase Bash/Xonsh comparison warning concerning backticks

das-g:
to fix its grammar and increase its clarity


laerus:
Thank you @das-g !

PR #2054, opened 2017-01-03
Tutorial: replace regexpath by regexsearch

das-g:
`__xonsh_regexpath__` doesn't seem to exist in Xonsh 0.5.1, but `__xonsh_regexsearch__` has the mentioned implementation according to `??`

Also replaced `??` output with the actual (and user-agnostic) one from a Xonsh installed system-wide with `sudo pip install ...`


gforsyth:
Yup. Thanks a ton @das-g !

Issue #2056, opened 2017-01-04
Artefacts in ncurses applications


xsteadfastx:
i just came back from the holidays and updated some tools i use in a daily bases. for example my beloved xonsh. i started weechat and it throws so many artefacts that its not usable. they come up everytime i move from channel to channel.  i made a screenshot. 

![xonshartefact](https://cloud.githubusercontent.com/assets/168397/21640870/5cc947c8-d278-11e6-9658-269cc5e0fc90.png)

i tried it with bash and dash and weechat runs just fine. any ideas?

gforsyth:
hey @xsteadfastx -- sorry about that! fix incoming.

gforsyth:
see #2057

scopatz:
Also @jb-leger may have a general fix for this class of problems in #2061. I am hopeful!

xsteadfastx:
this sounds good. i hope it gets merged soon.

scopatz:
@xsteadfastx @gforsyth's fix should now be in 0.5.2 for you

xsteadfastx:
@scopatz this looks great. no artefacts right now <3

scopatz:
Awesome!

PR #2057, opened 2017-01-04
add weechat to default predictors (ncurses based)

gforsyth:
this should fix weechat in xonsh, but the larger issues around ncurses based applications raised in #2056 remains


scopatz:
LGTM @gforsyth Thanks!

PR #2059, opened 2017-01-04
Tutorial: fix stdin handling in anonymous aliases

das-g:
In Xonsh 0.5.1, an alias function's `stdin` argument receives a `_io.TextIOWrapper` object, which cannot be stripped or be treated as a string as-is. We have to `.read()` it first and then `.strip()` the resulting string instead.

Also, fix the second lambda to output the space and EOL correctly.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -504,7 +504,7 @@ result is automatically converted to a string. For example,
     4
     >>> echo @([42, 'yo'])
     42 yo
-    >>> echo "hello" | @(lambda a, s=None: s.strip() + " world\n")
+    >>> echo "hello" | @(lambda a, s=None: s.read().strip() + " world\n")
     hello world
 
 This syntax can be used inside of a captured or uncaptured subprocess, and can
@@ -1256,7 +1256,7 @@ as aliases, by wrapping them in ``@()``.  For example:
 
     >>> @(_banana)
     'My spoon is tooo big!'
-    >>> echo "hello" | @(lambda args, stdin=None: stdin.strip() + args[0]) world
+    >>> echo "hello" | @(lambda args, stdin=None: ' '.join([stdin.read().strip(), *args]) + '\n') world
     hello world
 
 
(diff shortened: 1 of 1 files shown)


docs/tutorial.rst
```diff
@@ -1256,7 +1256,7 @@ as aliases, by wrapping them in ``@()``.  For example:
 
     >>> @(_banana)
     'My spoon is tooo big!'
-    >>> echo "hello" | @(lambda args, stdin=None: stdin.strip() + args[0]) world
+    >>> echo "hello" | @(lambda args, stdin=None: ' '.join([stdin.read().strip(), *args]) + '\n') world
```
das-g: This lambda is more generic (will happily append multiple arguments if given) but lacks the elegance and simplicity of the previous revision :slightly_frowning_face:

docs/tutorial.rst
```diff
@@ -1256,7 +1256,7 @@ as aliases, by wrapping them in ``@()``.  For example:
 
     >>> @(_banana)
     'My spoon is tooo big!'
-    >>> echo "hello" | @(lambda args, stdin=None: stdin.strip() + args[0]) world
```
das-g: Even if previous Xonsh version might have passed something else than an `_io.TextIOWrapper` as `stdin` or an `_io.TextIOWrapper` might have behaved differently (more string-like), I don't quite see how this implementation ever resultet in the desired output with a space between the words and an `EOL` at the end, so the prompt wouldn't eat the output on the next CLI input ... :confused:

scopatz: No, this wasn't always broken this probably broke in the refactor on v0.5. Thanks for catching it!

docs/tutorial.rst
scopatz: I think that since this is a lambda applied in place, generality is not needed here.  Would you mind simplifying it please? Thanks again for catching this.

scopatz:
Thanks for the PR @das-g! It is much appreciated.

docs/tutorial.rst
das-g: Just wondering: Where did the space between "hello" and "world" and the EOL come from back then?

scopatz: I don't think that there was one.

docs/tutorial.rst
das-g: done: 0e4cb0b

scopatz:
Thanks!

Issue #2061, opened 2017-01-04
Autodetect ncurse and program which use directly tty


jb-leger:
Hi,

There is a predictor dict used to predict if a command will be runnable in background (i.e. do not use the tty), by manually adding commands. Command have to be manually added as view in #2056, with associated PR #2057.

I had some problems with other commands which use tty, and I wonder why we do not try to detect ncurses / tty usage automatically?

I wrote that, in https://github.com/jb-leger/xonsh/tree/background_preditor_from_bin, it seems to work fine on POSIX systems.

Is it useful?

scopatz:
Hi @jb-leger - that would be extraordinarily useful! Could you put in a PR so that we can talk about it there, please? Thanks so much!

jb-leger:
Associated PR merged. I close this issue.

PR #2062, opened 2017-01-05
Background predictor generator from binary analysis

jb-leger:
Build a background predictor from analyzing the binary file.
    
This commit introduce a predictor generator from analyzing the binary of the executable.
It works for emacs, weechat, vim, nano, mutt for example.

The main idea is to detect the use of ncurses or manual tty handling with isatty, tcgetattr, and tcgetattr.

It should works on POSIX, fully tested on Linux, predictor generator function is also tested on MacOS.

For me it is WIP, not to be merged now, but in #2061 @scopatz ask me to submit PR.


xonsh/commands_cache.py
```diff
@@ -184,9 +183,62 @@ def predict_threadable(self, cmd):
                 pre, ext = os.path.splitext(name)
                 if pre in predictors:
                     predictors[name] = predictors[pre]
+        if name not in predictors:
+            predictors[name] = self.default_predictor(name, cmd[0])
```
scopatz: Why move away from the defaultdict?

scopatz: Oh, I think I see why, nm.

xonsh/commands_cache.py
```diff
@@ -184,9 +183,62 @@ def predict_threadable(self, cmd):
...
+        else:
+            return predict_true
+
+    def default_predictor_readbin(self, name, cmd0, timeout, failure):
+        """Make a defautt predictor by
+        analyzing the content of the binary. Should only works on POSIX.
+        Return failure if the analysis fails.
+        """
+
+        fname = cmd0 if os.path.isabs(cmd0) else self.locate_binary(name)
```
scopatz: This should probably call `lazy_locate_binary()` instead. The cache should be filled by SubprocSpec by the time we get to the prediction stage.

xonsh/commands_cache.py
```diff
@@ -184,9 +183,62 @@ def predict_threadable(self, cmd):
...
+        try:
+            fd = os.open(fname, os.O_RDONLY | os.O_NONBLOCK)
+        except:
+            return failure  # opening error
+
+        search_for = {
+            (b'ncurses',): [False, ],
+            (b'isatty', b'tcgetattr', b'tcsetattr'): [False, False, False],
+        }
+        tstart = os.times()[4]
```
scopatz: Is there a reason to us this rather than just `time.time()`?

xonsh/commands_cache.py
```diff
@@ -184,9 +183,62 @@ def predict_threadable(self, cmd):
...
+        block = b''
+        while os.times()[4] < tstart + timeout:
+            previous_block = block
+            block = os.read(fd, 2048)
+            if len(block) == 0:
+                os.close(fd)
+                return predict_true  # no keys of search_for found
+            analyzed_block = previous_block + block
+            for k, v in search_for.items():
+                for i in range(len(k)):
```
scopatz: Might be nice to have an `if v[i]: continue` block here so we don't recheck for keys we have already found.

scopatz:
This is really clever! Thank you so much for putting this in @jb-leger. Other than the stylistic changes, I think that this definitely deserves a news entry. 

Also, I think a test or two is probably warranted.  You can add some fake files to the test suite that contain and don't contain the keys that we are looking for and then test against those.  If you could do that, I would really appreciate it!

scopatz:
Thanks for the updates @jb-leger. Please let us know when you think it is ready.

jb-leger:
In my POV, I think it is ready for me. But only the idea of the generator was tested on Mac, without xonsh, so I think it could be a good idea to have a test from a Mac user before merging.

jb-leger:
Also, in CommandCache, the command are cached using the basename. Maybe there are a good reason for that, but I think it is a bad idea. For command which are in the PATH, basename is the same as the command name, for others, why we should presume that binary in different path (e.g. compiled program ``./a.out``) have the same behavior?

scopatz:
Thanks a ton for this!

scopatz:
> Also, in CommandCache, the command are cached using the basename. Maybe there are a good reason for that, but I think it is a bad idea.

The primary/original purpose of the cache is to resolve command names to file system paths without having to actually search the file system each time a command is run. If a user runs an absolute or relative path, the command cache should not be looked into, since the path is already known. I hope this makes sense.

The threadable predictors also make sense to key on basename since default locations are platform dependent. Maybe there is an edge case here that we are missing, though, with relative and absolute paths here. This is probably a separate issue.

Issue #2012, opened 2016-12-13
events.on_precommand broken


laerus:
without bashisms xontrib loaded

```
$ @events.on_precommand
. def _(*args):
.     print('pre1')
.     
$ 
pre1
pre1
$ echo okpre1
pre1
$ echo ok                                                                             
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/home/laerus/src/xonsh/xonsh/ptk/shortcuts.py", line 109, in prompt
    document = cli.run(reset_current_buffer=False)
  File "/usr/lib/python3/dist-packages/prompt_toolkit/interface.py", line 394, in run
    self.eventloop.run(self.input, self.create_eventloop_callbacks())
  File "/usr/lib/python3/dist-packages/prompt_toolkit/eventloop/posix.py", line 164, in run
    t()
  File "/usr/lib/python3/dist-packages/prompt_toolkit/eventloop/posix.py", line 83, in read_from_stdin
    inputstream.feed(data)
  File "/usr/lib/python3/dist-packages/prompt_toolkit/terminal/vt100_input.py", line 395, in feed
    self._input_parser.send(c)
  File "/usr/lib/python3/dist-packages/prompt_toolkit/terminal/vt100_input.py", line 304, in _input_parser_generator
    self._call_handler(match, prefix)
  File "/usr/lib/python3/dist-packages/prompt_toolkit/terminal/vt100_input.py", line 337, in _call_handler
    self.feed_key_callback(KeyPress(key, insert_text))
  File "/usr/lib/python3/dist-packages/prompt_toolkit/interface.py", line 1023, in feed_key
    cli.input_processor.process_keys()
  File "/usr/lib/python3/dist-packages/prompt_toolkit/key_binding/input_processor.py", line 201, in process_keys
    self._process_coroutine.send(key_press)
  File "/usr/lib/python3/dist-packages/prompt_toolkit/key_binding/input_processor.py", line 158, in _process
    self._call_handler(matches[-1], key_sequence=buffer)
  File "/usr/lib/python3/dist-packages/prompt_toolkit/key_binding/input_processor.py", line 228, in _call_handler
    handler.call(event)
  File "/usr/lib/python3/dist-packages/prompt_toolkit/key_binding/registry.py", line 31, in call
    return self.handler(event)
[...]

PR #2063, opened 2017-01-05
Create on_transform_command event

astronouth7303:
Should deal with the issues discussed in #2012.

* [x] Write tests
* [x] News item


xonsh/parsers/base.py
```diff
@@ -19,13 +22,55 @@
...
+
+Fired to request xontribs to transform a command line. Return the transformed
+command, or the same command if no transformaiton occurs.
+
+This may be fired multiple times per command, so design any handlers for this
+carefully.
+""")
+
+
+def transform_command(src, show_diff=True):
```
scopatz: I don't know that the parser is the best place for this, since it is specifically for things beyond the parser. Maybe a new module in the base xonsh package would be better.

xonsh/parsers/base.py
```diff
@@ -19,13 +22,55 @@
 from xonsh.tokenize import SearchPath, StringPrefix
 from xonsh.lazyasd import LazyObject
 from xonsh.parsers.context_check import check_contexts
+from xonsh.events import events
+from xonsh.tools import print_exception
```
scopatz: Also, the parser itself should not need to know about events or exception printing.

scopatz:

xonsh/parsers/base.py
```diff
@@ -19,13 +22,55 @@
...
 
 
+events.doc('on_transform_command', """
+on_command_transform(cmd: str) -> str
+
+Fired to request xontribs to transform a command line. Return the transformed
+command, or the same command if no transformaiton occurs.
+
+This may be fired multiple times per command, so design any handlers for this
+carefully.
```
scopatz: I thought the idea here was that we could have a different event kind that was ordered so we don't have to run this multiple times.

xonsh/parsers/base.py
astronouth7303: The parser seemed like the best place to put `on_transform_command` so that it's handled consistently across ptk, readline, jupyter, and scripts.

xonsh/parsers/base.py
astronouth7303: I wasn't messing with that. I was just trying to unbreak `on_precommand` and make sure that `on_transform_command` works consistently across all execution contexts.

xonsh/parsers/base.py
astronouth7303: Either in the parser, in the executor, or piece-meal in shells, script handlers, and a few other places.

xonsh/parsers/base.py
scopatz: Ok, that is great then.

xonsh/parsers/base.py
scopatz: Maybe put it in a new module then in the parser subpackage? I still think that is a little weird. The parser doesn't actually use this event though. Could you clarify how this helps consistency?  I had put this in base_shell originally because that is what gets called by all of the frontends.

astronouth7303: It's down at 375.

I didn't think shells applied to script execution, though? I would like to be able to apply it before history, so that history backends may choose to log pre- and post-transformation.

astronouth7303: Or, alternatively, we can say transformers only apply to interactive contexts.

scopatz: > Or, alternatively, we can say transformers only apply to interactive contexts.

This has been my thought all along. The only transformations we have been asked for don't really make sense in a script environment. This could change if we try to reimagine the context sensitive transformer as a transform event. Then it would need to be applied to scripts too.I was hoping to worry about that later though :)

I agree about the history backends too.

astronouth7303: Oh, ok, then.

astronouth7303:
Ok, what? Apparently, resetting events is causing infinite recursions in parser tests.

This is officially the wackiest thing I've seen in a long time.

xonsh/events.py
```diff
@@ -248,22 +253,35 @@ def transmogrify(self, name, klass):
...
         # This is only called if the attribute doesn't exist, so create the Event...
-        # (A little bit of magic to enable docstrings to work right)
-        e = type(name, (Event,), {'__doc__': None})()
+        e = self._mkevent(name)
         # ... and save it.
         setattr(self, name, e)
         # Now it exists, and we won't be called again.
         return e
 
+    def clean_between_tests(self):
```
laerus: i don't think it's a good idea to have code that is for testing only,
maybe this is a code smell, can't this be done in tests?

astronouth7303: Only if you'd like the complexity of how to clean out events buried in the unrelated `xonsh_builtins` fixture.

astronouth7303: I could name it `reset_all_events()` instead and it would do the same thing and have the same usefulness.

laerus: can't we have a pytest.fixture that creates new events and/or EventManager for each test?

astronouth7303: I suppose? You can't create a new `EventManager` each test because all of xonsh depends on its existence at load time.

astronouth7303:
Well, I moved it into a fixture, but I'm still not sure how or why I'm breaking the ABC framework.

astronouth7303:
So apparently, very deep trees (ie, 293 MRO) seems to break ABCs, but it's not clear why or why the break level is so low.

astronouth7303:
Anything else for this?

tests/conftest.py
```diff
@@ -34,7 +34,17 @@ def xonsh_execer(monkeypatch):
 
 
 @pytest.yield_fixture
-def xonsh_builtins():
+def xonsh_events():
+    yield events
+    for name, oldevent in vars(events).items():
+        # Heavily based on transmogrification
+        klass = type(oldevent).__bases__[0]  # events.on_chdir -> <class on_chdir> -> <class Event>
+        newevent = events._mkevent(name, klass, klass.__doc__)
```
scopatz: Maybe this should pass in all bases, just in case, even if it only takes the docstring from the first base.

xonsh/events.py
```diff
@@ -228,6 +228,12 @@ def doc(self, name, docstring):
         """
         type(getattr(self, name)).__doc__ = docstring
 
+    @staticmethod
+    def _mkevent(name, klass=Event, doc=None):
```
scopatz: klass might need to be bases

scopatz:
Thanks @astronouth7303! Just the one small bases request.

scopatz:
Note that I reported an upstream error at http://bugs.python.org/issue29221

xonsh/events.py
scopatz: pinging this @astronouth7303

tests/conftest.py
astronouth7303: YAGNI. I consider the fact that I have to make a new subclass an implementation detail.

astronouth7303:
... and now I"m getting unrelated CI failures.

scopatz:
Thanks @astronouth7303!

PR #2065, opened 2017-01-05
made sending signals on mac safer

### LINKED ISSUE #2064: ls | more does not work on mac
xonsh version: 0.5.2
python version 3.6.0

*****************************************************************************************************************************
```
>>> ls | more
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 17380, in default
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 1964, in run_compiled_code
    func(code, glb, loc)
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 16159, in subproc_captured_hiddenobject
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 16127, in run_subproc
    command.end()
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 11116, in end
    for _ in self.tee_stdout():
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 11041, in tee_stdout
    for line in self.iterraw():
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 10940, in iterraw
    wait_for_active_job()
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 3694, in wait_for_active_job
    _continue(active_task)
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 3607, in _continue
    _send_signal(job, signal.SIGCONT)
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 3542, in _send_signal
    os.kill(pid, signal)
ProcessLookupError: [Errno 3] No such process
```
*****************************************************************************************************************************

Also, I created a alias function ls-more on python version 3.5... now that does not work either.

def _lsmore(args, stdin):                                                                                                           
    ls -lt @(args) >> .ls_more                                                                                                      
    more .ls_more                                                                                                                   
    rm .ls_more   

aliases['ls-more'] = _lsmore

scopatz:
Hopefully will fix #2064

### DIFF
--- a/news/sigmac.rst
+++ b/news/sigmac.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Made sending job control signals on Mac safer.
+
+**Security:** None
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -792,9 +792,9 @@ def run_subproc(cmds, captured=False):
             'obj': proc,
             'bg': spec.background,
         })
-    #if _should_set_title(captured=captured):
+    if _should_set_title(captured=captured):
         # set title here to get currently executing command
-    #    pause_call_resume(proc, builtins.__xonsh_shell__.settitle)
+        pause_call_resume(proc, builtins.__xonsh_shell__.settitle)
     # create command or return if backgrounding.
     if spec.background:
         return
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -792,9 +792,9 @@ def run_subproc(cmds, captured=False):
             'obj': proc,
             'bg': spec.background,
         })
-    if _should_set_title(captured=captured):
+    #if _should_set_title(captured=captured):
         # set title here to get currently executing command
-        pause_call_resume(proc, builtins.__xonsh_shell__.settitle)
+    #    pause_call_resume(proc, builtins.__xonsh_shell__.settitle)
     # create command or return if backgrounding.
     if spec.background:
         return
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -355,7 +355,7 @@ def no_pg_xonsh_preexec_fn():
     pipeline group.
     """
     os.setpgrp()
-    signal.signal(signal.SIGSTOP, default_signal_pauser)
+    signal.signal(signal.SIGTSTP, default_signal_pauser)
 
 
 class SubprocSpec:
@@ -550,7 +550,7 @@ class SubprocSpec:
             def xonsh_preexec_fn():
                 """Preexec function bound to a pipeline group."""
                 os.setpgid(0, pipeline_group)
-                signal.signal(signal.SIGSTOP, default_signal_pauser)
+                signal.signal(signal.SIGTSTP, default_signal_pauser)
         kwargs['preexec_fn'] = xonsh_preexec_fn
 
     #
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -355,7 +355,7 @@ def no_pg_xonsh_preexec_fn():
     pipeline group.
     """
     os.setpgrp()
-    signal.signal(signal.SIGTSTP, default_signal_pauser)
+    signal.signal(signal.SIGSTOP, default_signal_pauser)
 
 
 class SubprocSpec:
@@ -550,7 +550,7 @@ class SubprocSpec:
             def xonsh_preexec_fn():
                 """Preexec function bound to a pipeline group."""
                 os.setpgid(0, pipeline_group)
-                signal.signal(signal.SIGTSTP, default_signal_pauser)
+                signal.signal(signal.SIGSTOP, default_signal_pauser)
         kwargs['preexec_fn'] = xonsh_preexec_fn
 
     #
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -355,7 +355,7 @@ def no_pg_xonsh_preexec_fn():
     pipeline group.
     """
     os.setpgrp()
-    signal.signal(signal.SIGTSTP, default_signal_pauser)
+    #signal.signal(signal.SIGTSTP, default_signal_pauser)
 
 
 class SubprocSpec:
@@ -550,7 +550,7 @@ class SubprocSpec:
             def xonsh_preexec_fn():
                 """Preexec function bound to a pipeline group."""
                 os.setpgid(0, pipeline_group)
-                signal.signal(signal.SIGTSTP, default_signal_pauser)
+                #signal.signal(signal.SIGTSTP, default_signal_pauser)
         kwargs['preexec_fn'] = xonsh_preexec_fn
 
     #
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -27,7 +27,10 @@ if ON_DARWIN:
         for pid in job['pids']:
             if pid is None:  # the pid of an aliased proc is None
                 continue
-            os.kill(pid, signal)
+            try:
+                os.kill(pid, signal)
+            except Exception:
+                pass
 elif ON_WINDOWS:
     pass
 elif ON_CYGWIN:
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -1735,8 +1735,6 @@ class CommandPipeline:
             # we get here if the process is not threadable or the
             # class is the real Popen
             PrevProcCloser(pipeline=self)
-            if ON_POSIX:
-                proc.send_signal(signal.SIGCONT)
             wait_for_active_job()
             proc.wait()
             self._endtime()
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -1735,6 +1735,8 @@ class CommandPipeline:
             # we get here if the process is not threadable or the
             # class is the real Popen
             PrevProcCloser(pipeline=self)
+            if ON_POSIX:
+                proc.send_signal(signal.SIGCONT)
             wait_for_active_job()
             proc.wait()
             self._endtime()


mitnk:
I not quite familiar with code of proc/jobs. Here is what I got so far:

For `ls | more`, firstly, the status of `ls` is becoming "zombie" and `more` "stopped"; and when sending SIGCONT, `ls` Popen is already gone.

With the try-catch in this PR, it will enter `jobs.wait_for_active_job()` for the second time, and found `active_task is None` and returns. And now `ls` proc is gone, and `more` is with "stopped" status, and code stuck on `proc.wait()` (more) line in `proc.CommandPipeline.iterraw()` for ever.

mitnk:
Is there any reason using `os.pipe()` but not `subprocess.PIPE` and [subprocess.Popen.communicate](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) in xonsh?

Refer 1: https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

Refer 2: http://stackoverflow.com/a/2761237/665869

scopatz:
Ahh, Ok.  So maybe this is the wrong fix.  Instead we really need to be able to find the ls and more processes so that we can continue/wake them and kill them as needed. Do we know why it can't find the process? Does mac change the pid somehow?

Unfortunately, we cannot use communicate() and have the kind of interactivity and redirection that we want. We need to be in complete control of the pipes that we create and manage. I don't know of another shell that doesn't manually creates pipes.

mitnk:
I don't think mac would change pids. `ls` proc seems just become "zombie", and then terminated.

scopatz:
Yeah, the problem is that more is stopped for some reason and isn't getting continued. I get an `S` status, for sleeping idle on Linux in htop.  Not a `T` for suspended. Is this what you are seeing?

mitnk:
I alway saw `T`
```
15095 s006  T      0:00.01 /usr/bin/more
```

scopatz:
Fascinating. Ok that is the problem!

scopatz:
Alright @mitnk, can you try pulling down the latest change and seeing if it works now? Expect a few more of these. Thanks!

mitnk:
Nope .. more is still stuck at T, and ls proc is gone :)

scopatz:
Ok, can you try it now please?

mitnk:
Issue still exists. Send SIGCONT (inside or outside of xonsh) to the T more does not have effects. I can dig more when i got time and sync-up here.

scopatz:
Wow, ok that is bad. Can you try it now please? I have swapped out the relevant `SIGTSTP` (typed stop) with `SIGSTOP` (untyped). There are some places online that seem to suggest other people are having problems with `SIGTSTP`.  Also, since the not setting the title had no effect, I am reasonably sure that `SIGSTOP` works because of how `pause_call_resume()` is implemented.

If this doesn't work, I have one more semi-obvious thing to try, ie not setting a `pre_exec_fn`.

mitnk:
Got an exception now:
```
$ ls | more
  File "/Users/hugo/mitnk/xonsh/xonsh/base_shell.py", line 319, in default
  File "/Users/hugo/mitnk/xonsh/xonsh/codecache.py", line 65, in run_compiled_code
    func(code, glb, loc)
  File "/Users/hugo/mitnk/xonsh/xonsh/built_ins.py", line 846, in subproc_captured_hiddenobject
  File "/Users/hugo/mitnk/xonsh/xonsh/built_ins.py", line 783, in run_subproc
  File "/Users/hugo/mitnk/xonsh/xonsh/built_ins.py", line 510, in run
    p = self._run_binary(kwargs)
  File "/Users/hugo/mitnk/xonsh/xonsh/built_ins.py", line 520, in _run_binary
    p = self.cls(self.cmd, bufsize=bufsize, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 707, in __init__
    restore_signals, start_new_session)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 1327, in _execute_child
    raise child_exception_type(err_msg)
subprocess.SubprocessError: Exception occurred in preexec_fn.
```

scopatz:
Yeah, that didn't work at all.  Here is an attempt that tries to turn off signal pausing.

mitnk:
no errors now, but still stuck

laerus:
python subprocess docs warn about using a preexec function with threads but it is not obvious if there is a race condition somewhere.

the examples on replacing shell pipes also close the previous subproc stdout to propagate SIGPIPE, dunno if this has any effect on the current issue

still going around the subproc code on spare time to figure out this cute beast :)

scopatz:
This might have been related to the Ctrl+Z and backgrounding issue. Is this still happening on master?

mitnk:
latest master break at a different way:
```
$ sudo py setup.py install
^Z
$ ls | more
^Z
$ ls | wc -l
      31
$ jobs
$ jobs
[12]+ stopped: sudo py setup.py install (49839)
[11]- stopped: ls | more (49837)
[10]  stopped: sudo py setup.py install (49828)
[9]  stopped: ps | more (49815)
[8]  stopped: ls | more (49813)
[7]  stopped: ls | more (49804)
[6]  stopped: ls | more (49801)
[5]  stopped: ls | more (49798)
[4]  stopped: ls | more (49796)
[3]  stopped: ls | more (49794)
[2]  stopped: ls | more (49792)
[1]  stopped: ls | more (49790)
```

mitnk:
Got where related to this issue - `pipeline_group` stuff. If we make `prep_preexec_fn()` or `xonsh_preexec_fn()` just return at the first line of its body (do nothing), then `ls | more` will work properly.

mitnk:
So continued with previous comment. The following code mimic the less-stuck-at-status-T issue:
```
import sys
import os
import subprocess

stage = 0
lead_pid = 0

def get_gid(pid):
    try:
        return os.getpgid(pid)
    except Exception:
        pass

def pre_func():
    if stage == 0:
        os.setpgrp()
    else:
        try:
            os.setpgid(0, lead_pid)
        except:
            pass

r, w = os.pipe()
p1 = subprocess.Popen(
    ['ls', ],
    stdout=w,
    preexec_fn=pre_func,
)
print('ls - pid:{} gid:{}'.format(p1.pid, get_gid(p1.pid)), file=sys.stderr)
lead_pid = p1.pid
stage = 1

p2 = subprocess.Popen(
    ['less', ],
    stdin=r,
    preexec_fn=pre_func,
)
print('less - pid:{} gid:{}'.format(p2.pid, get_gid(p2.pid)), file=sys.stderr)

p1.wait()
os.close(w)
p2.wait()
os.close(r)
```

And code will work properly, when `pre_func()` is empty:
```
def pre_func():
    return
```

Basically it's that the process group lead (ls) has exited (so that this process group marked exited?), which caused the less precess, which belongs to this (exited?) process group, stuck.

I'm currently not sure how to fix this in xonsh. But hope this could leave a clue for you.

mitnk:
> If this doesn't work, I have one more semi-obvious thing to try, ie not setting a pre_exec_fn.

Ah, you had already mentioned that :)

scopatz:
This is good to know @mitnk. So if we have an empty `pre_exec_fn()` are you still abel to background the process with ^Z?

mitnk:
`Ctrl-Z` seems cannot make jobs to background on Mac with master branch.

Also in a Linux box, I find `fg` not works well to bring stoped jobs (with `Ctrl+Z`) to foreground. Is this the case on your side?

mitnk:
Opened a separated `^Z` issue here: #2133

scopatz:
Hi @mitnk,  ^Z and fg seem to work on linux for me with some of the latest fixes that just went in a day or two ago. Thanks for opening the other issue.  What should we do here, though?  Is the fix to not have a `pre_exec_fn()` on mac at all?

mitnk:
For fixing this `ls | less` on mac issue only, do not call `os.setpgid()` and `os.setpgrp()` in `no_pg_xonsh_preexec_fn()` and `xonsh_preexec_fn()` is sufficient.

Shells like bash does create new process group for each job, so I'm not sure it's good or not fix this issue this way. (May harm job control).

What I'm not clear about is how (why) the `less` proc get STOPPED, when the process group leader (proc `ls`) terminated (and would not stop on Linux).

mitnk:
Now I think I know why `less` proc in job `ls | less`, i.e. ...
```
$ ls | less
Applications
Desktop
Documents
:^Z
$  # <-- xonsh now takes back the terminal
```
... get stopped:

```
 PID  PPID  PGID TPGID STAT COMMAND
1856  1821  1856  1856 S+   xonsh
1865  1856  1864  1856 T    less
```
ls (PID 1864) terminated normally before xonsh run `CommandPipeline.end()`, which calls `wait_for_active_job()`, where the new process group (PGID 1864) could get a chance to be set as foreground process group (by `tcsetpgrp()`).

Since `less` (PID 1865, GID 1864) program needs receiving SDTIN to work, but its process group doesn't control the terminal - and be made `STOPPED` (by kernel?).

And now when running into `wait_for_active_job()`, we cannot find any active task and the code decides to give terminal back the shell (xonsh proc). 

And it works if we don't create new progress group, I.E.
```
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -362,7 +362,7 @@ def no_pg_xonsh_preexec_fn():
     """Default subprocess preexec function for when there is no existing
     pipeline group.
     """
-    os.setpgrp()
+    # os.setpgrp()
     signal.signal(signal.SIGTSTP, default_signal_pauser)

@@ -557,7 +557,7 @@ class SubprocSpec:
         else:
             def xonsh_preexec_fn():
                 """Preexec function bound to a pipeline group."""
-                os.setpgid(0, pipeline_group)
+                # os.setpgid(0, pipeline_group)
                 signal.signal(signal.SIGTSTP, default_signal_pauser)
         kwargs['preexec_fn'] = xonsh_preexec_fn
```
Then less works because its process group has the controlling terminal:
```
 PID  PPID  PGID TPGID STAT COMMAND
2668  2655  2668  2668 S+   xonsh
2676  2668  2668  2668 S+   less
```

scopatz:
Interesting. Would you mind sending a PR to replace this one with something that works for you, please? I have no way to develop/test this on mac so I think it would be easier if you opened it up. Thanks!

mitnk:
Ok, I will open a new PR. Thanks.

mitnk:
Opened new PR #2150, closing this one.

PR #2068, opened 2017-01-06
History fix - use tss instead of buffer

mitnk:
Fixed an issue with sqlite history backend that a second "exit" do not kill all the unfinished jobs when exiting xonsh.

Very minor one, so I guess no news items is needed.

### DIFF
--- a/news/hist-tss-fix.rst
+++ b/news/hist-tss-fix.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks! I think this does need a news item since it fixes an issue that is in a release. LGTM otherwise.

mitnk:
@scopatz added, thanks!

scopatz:
Thanks!

PR #2072, opened 2017-01-09
source alias improvements

scopatz:
Some small additions I saw when trying to port conda de/activate to xonsh

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -259,8 +259,12 @@ def source_alias(args, stdin=None):
         fpath = fname
         if not os.path.isfile(fpath):
             fpath = locate_binary(fname)
-            if fpath is None and env.get('XONSH_DEBUG'):
-                print('source: {}: No such file'.format(fname), file=sys.stderr)
+            if fpath is None:
+                if env.get('XONSH_DEBUG'):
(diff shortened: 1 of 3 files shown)


xonsh/aliases.py
```diff
@@ -255,18 +255,19 @@ def source_alias(args, stdin=None):
     env = builtins.__xonsh_env__
     encoding = env.get('XONSH_ENCODING')
     errors = env.get('XONSH_ENCODING_ERRORS')
-    for fname in args:
+    for i, fname in enumerate(args):
         fpath = fname
         if not os.path.isfile(fpath):
             fpath = locate_binary(fname)
-            if fpath is None:
+            if fpath is None and env.get('XONSH_DEBUG'):
```
laerus: if the path is None and we haven't set `XONSH_DEBUG` this will try to source the file,
is that what we want? the previous behavior seems very different

scopatz: Whoops! Thanks for catching this.

scopatz: fixed.

scopatz:
ping!

laerus:
Merging, ty @scopatz !

PR #2073, opened 2017-01-09
Add Python 3.6 to Continuous Integration

astronouth7303:
Seems to be some oddities with Appveyor, though. Help?

Internal. No news.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,7 +1,23 @@
 version: 0.5.2.{build}
 os: Windows Server 2012 R2
+environment:
+
+  matrix:
+
+    # For Python versions available on Appveyor, see
+    # http://www.appveyor.com/docs/installed-software#python
+    # The list here is complete (excluding Python 2.6, which
+    # isn't covered by this document) at the time of writing.
+
+    - PYTHON: "C:\\Python34"
+    - PYTHON: "C:\\Python35"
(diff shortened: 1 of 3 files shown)


.appveyor.yml
```diff
@@ -1,7 +1,19 @@
 version: 0.5.2.{build}
 os: Windows Server 2012 R2
+environment:
+
+  matrix:
+    # http://www.appveyor.com/docs/installed-software#python
+    - PYTHON: "C:\\Python34"
```
scopatz: We can probably stop testing on Python v3.4

scopatz:
Theee seem to be legitimate errors on Python v3.6 on Windows. It looks like some of the context management stuff might not work yet.

.appveyor.yml
astronouth7303: Probably? Although I know it's the version installed by default on Amazon Linux. (Aka, what's default on EC2.)

astronouth7303:
That's pretty weird that something that doesn't involved system interaction is only broken on Windows and 3.6.

.appveyor.yml
laerus: debian-stable will still be python3.4 for some time too.

scopatz: That is fair. I guess we can keep it around

scopatz:
Potential fix in #2074

PR #2074, opened 2017-01-10
Macro contexts refactor

scopatz:
This includes #2073, which I hope will fix the appveyor issues.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,7 +1,23 @@
 version: 0.5.2.{build}
 os: Windows Server 2012 R2
+environment:
+
+  matrix:
+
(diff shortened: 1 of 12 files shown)


astronouth7303:
The `distributed` xontrib uses these, right? Does it need to be rewritten?

scopatz:
I looked at the distributed xontrib last night when I was writing this and I couldn't find anything that needed changing. Usage of the distributed xontrib will need to change, but this raises an exception to that effect.

astronouth7303:
Do we want to make note of that in the news item?

scopatz:
done

PR #2079, opened 2017-01-11
xonsh script.xsh should not fail over

### LINKED ISSUE #2077: failback shell should only work with interactive mode
atm if there a a `fail.xsh` script that raises any exception we run `sh`

```
# fail.xsh 
raise ValueError('just because')
```

```
$ xonsh fail.xsh 
  File "/home/laerus/src/xonsh/xonsh/main.py", line 298, in main
    return main_xonsh(argv)
  File "/home/laerus/src/xonsh/xonsh/main.py", line 336, in main_xonsh
    loc=None, mode='exec')
  File "/home/laerus/src/xonsh/xonsh/codecache.py", line 145, in run_script_with_cache
  File "/home/laerus/src/xonsh/xonsh/codecache.py", line 65, in run_compiled_code
    func(code, glb, loc)
  File "fail.xsh", line 1, in <module>
    raise ValueError('just because')
ValueError: just because
Xonsh encountered an issue during launch
Failback to /bin/sh
$ 
```

mitnk:
fixes #2077

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -9,6 +9,7 @@ import sys
 
 import xonsh.main
 import pytest
+from tools import TEST_DIR
 
 
 def Shell(*args, **kwargs):
@@ -91,7 +92,7 @@ def test_xonsh_failback(shell, monkeypatch):
     assert failback_checker == ['/bin/xshell', '/bin/xshell']
 
 
-def test_xonsh_failback_non_interactive(shell, monkeypatch):
+def test_xonsh_failback_single(shell, monkeypatch):
     class FakeFailureError(Exception):
         pass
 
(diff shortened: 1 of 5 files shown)


tests/test_main.py
```diff
@@ -103,3 +104,17 @@ def mocked_main(*args):
 
     with pytest.raises(FakeFailureError):
         xonsh.main.main()
+
+
+def test_xonsh_failback_script_from_file(shell, monkeypatch):
```
mitnk: Not an issue of this PR, but if this new test is run based on the old version of `xonsh/main.py` file.

This is pass (20 passed)
```
py.test tests/test_main.py
```

And this will fail (1 failed, 19 passed)
```
py.test -s tests/test_main.py
```

quite confused to me.

laerus: this is probably due to monkeypatching stderr, `-s` messes with stdout/stderr, can't recall in what way.

laerus: do you think we can just test the failback function and not run main at all in these tests?

mitnk: What's the problem you see in current status of these tests to propose this change?

laerus: too much monkeypatching going on, feels like we are testing more that we should on these tests.
failback only makes sense on interactive mode anyway.

mitnk: "non-interactive should not fail back" is also a case, right?

Any suggests in code to write these cases without writting mocks?

laerus: there is also the case where we want to run a script and then go into interactive
mode like the `-i` option of python, something that doesn't work atm and
 it didn't work the last time i checked way back before the failback PR.

maybe if we isolate the try/except that calls the failback function 
to the `XonshMode.interactive` clause of main, we can then just mock out
`execlp` and call directly the `_failback_to_other_shells` function in tests.

laerus:
i can confirm that this fixes the issue @mitnk, thank you !
we can do the refactoring of tests or even main at some other PR.

mitnk:
Mac has /etc/shells. Feel free to send PR for those you mentioned. Thanks @laerus.

PR #2080, opened 2017-01-12
^Z and fg Fixes

scopatz:
I am not sure that this fixes all use cases, but it does address the vi one for me on Linux.


laerus:
no time to test this atm but seems legit.

Issue #323, opened 2015-07-29
Redirecting stdout to stderr doesn't work


kseistrup:
In bash I usually write error messages to `stderr`, e.g.:

``` sh
[ CONDITION_THAT_SHOULD_SUCCEED ] || {
  echo 'we failed' >&2
  exit 1
}
echo 'we succeeded'
```

Following [the tutorial](http://xonsh.org/tutorial.html#input-output-redirection), I would expect at least one of these to work:

``` sh
echo 'we failed' out>err
echo 'we failed' >&2        # included for bash compatibility
```

but none of them do.

scopatz:
Hi @kseistrup - this looks like a legitimate bug.  We'll try to get to it quickly.  Alternatively, I can tell you where to start looking, if you want to lend a hand with debugging.

kseistrup:
I know where to look (and I did look).  The reason I didn't try to fix it is because I don't understand how the lexer works and what to do about it. ;)

scopatz:
Ahh OK.  I'll take a look tonight.  That is all @adqm's work :)

adqm:
I don't know that I would quite qualify this as a bug; this was intentionally left out as a feature.  It isn't by any means impossible to implement, but it does require some effort.  I tried to make this clear in the tutorial, but, evidently, I failed.

The constructor for `subprocess.Popen` objects accepts `stderr=STDOUT` to make stderr go to stdout, but it doesn't offer the opposite.  We will have to implement that ourselves if we want it (this will, I think, make the lexer less complicated, but it will make the handling of those subprocesses harder, as we'll have to monkey with the pipes ourselves).

scopatz:
Ahh I see.  It should still be possible to set `stdout=sys.stderr`, though.  But you are right, and I missed this, that we'll have to do something different.

kseistrup:
@adqm, I'm sorry if I didn't read the docs thoroughly enough, but as @scopatz said: it should still be possible to set `stdout=sys.stderr` in the `Popen()` call.

adqm:
Is `stdout=sys.stderr` actually what we want?  It's not clear to me that that will interact with other redirects properly (that is, if we did `o>e` and also sent stderr or stdout to a file, would we still get the right thing?). 

We can always try it, but I think we may need to dig in a little deeper (in the `_get_handles` method of the `Popen` and `ProcProxy` classes, where `stderr=STDOUT` is handled).

scopatz:
[...]

PR #2081, opened 2017-01-12
Allow stdout to redirect to stderr

scopatz:
This is waaaaay overdue and meant to address #323. Really not so bad after the infamous subprocess refactor.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -725,7 +725,7 @@ with the following syntax:
     >>> COMMAND err>o
     >>> COMMAND e>out
     >>> COMMAND e>o
-    >>> COMMAND 2>&1 # included for Bash compatibility
+    >>> COMMAND 2>&1  # included for Bash compatibility
 
 This merge can be combined with other redirections, including pipes (see the
 section on `Pipes`_ above):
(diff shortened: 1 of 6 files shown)


laerus:
i can confirm that this works!
although redirecting stderr to stdout doesn't
```
$ cd test
cd: no such file or directory: test
$ cd test 2>1
  File "/home/laerus/src/xonsh/xonsh/base_shell.py", line 319, in default
  File "/home/laerus/src/xonsh/xonsh/codecache.py", line 65, in run_compiled_code
    func(code, glb, loc)
  File "/home/laerus/src/xonsh/xonsh/built_ins.py", line 858, in subproc_captured_hiddenobject
  File "/home/laerus/src/xonsh/xonsh/built_ins.py", line 826, in run_subproc
    command.end()
  File "/home/laerus/src/xonsh/xonsh/proc.py", line 1914, in end
    for _ in self.tee_stdout():
  File "/home/laerus/src/xonsh/xonsh/proc.py", line 1839, in tee_stdout
    for line in self.iterraw():
  File "/home/laerus/src/xonsh/xonsh/proc.py", line 1746, in iterraw
    if ((stderr is None or spec.stderr is None or not safe_readable(stderr))
  File "/home/laerus/src/xonsh/xonsh/proc.py", line 1638, in safe_readable
    status = handle.readable()
AttributeError: 'int' object has no attribute 'readable'
```

btw bash doesn't seem to define `2>1` and `1>2` redirections without the ampersand
but this is prettier (if only we could drop bash redirections all together) :)

scopatz:
Thanks @laerus! Can you please open another issue for this. I believe that is a bug in `cd`, as other 2>1 redirects work.

PR #2083, opened 2017-01-12
Add environment variable for hg binary name

MatthewTurk:
This adds an environment variable for specifying the hg binary name to use; this can, for instance, allow specification of the `chg` binary for (considerably) faster queries of the status.


gforsyth:
Hey @MatthewTurk , thanks for putting this in!  There are a few test failures and you'll also need to add a news entry. 

Also, if we move forward with this we'll also need a `VarDocs` entry for the new envvar. 

Question, though: is this something that can be done by using an alias instead? e.g. `aliases['hg'] = 'chg'`

MatthewTurk:
Thanks, @gforsyth !  I will try out with `aliases`, which would be much nicer.  If that works (and I bet it will, based on you suggesting it :)), I'll rescind the PR, and if it doesn't, I'll update with your comments.

gforsyth:
Cool! If it _does_ work and there's an improvement in prompt time, this would be a great tip to put on the customizations page in the docs!

scopatz:
Thanks for putting this in @MatthewTurk! It looks like this was branched from an incomplete topical branch I was working on last night. That might explain some of the CI issues.

MatthewTurk:
Whoops!  My mistake.  I wasn't able to get aliases to work, which I'll
investigate further before rebasing/updating/blahblah.  Thanks folks!

scopatz:
Hey @MatthewTurk, any luck with this? Happy to help however I can. I do want to make sure that the hg support is top-notch

scopatz:
Hi All, should we close this given #2233?

MatthewTurk:
Fine with me!

scopatz:
Cool thanks!

Issue #367, opened 2015-08-20
tab completion on empty line shows many files


melund:
When I press tab on an empty terminal xonsh prints more than 5524 files. That is long list. I guess that is all the files I have in my path. In my normal terminal there is only 813 files. Seems that Xonsh also includes non executables. 

Also my normal terminal has a some nice behaviour. First is always ignores the first tab-press. On the second press it asks:

```
Display all 813 possibilities? (y or n)
```

Maybe xonsh should do something similar?

scopatz:
Are you on readline or prompt-toolkit?

scopatz:
I vaguely recall seeing this feature before with xonsh in readline, but it has been a while.

melund:
It was using readline. prompt-toolkit only shows five completions in which you can scroll. Do you know of any way to ask the user if there are many completions.

scopatz:
The magic variable is apparently `rl_completion_query_items`, which you can read about at http://www.delorie.com/gnu/docs/readline/rlman_47.html

This is probably not set, or set way too high, or something.  Something like what is done for `RL_COMPLETION_SUPPRESS_APPEND` should be done for this variable.

scopatz:
This should be fixed by #2086

gforsyth:
Fixed in #2086

PR #2086, opened 2017-01-13
Query Completions 

scopatz:
This should address #367.


gforsyth:
Thanks @scopatz ! This seems to work well.

scopatz:
Awesome! I am glad to have this in.

PR #2094, opened 2017-01-15
[RFC] Events keywords

astronouth7303:
In discussing the future of command transformation, I realized that history backends might want to log the pre- and post-transformation commands. But changing the signature of `on_precommand` will break  existing users.

So, moving forward, I propose that events are keyword-argument only, and that all handlers have a `**kw` against future expansion. (There is precedent for this. Django--and the PyDispatch it was based on--does the same thing.)

* [x] Implement in core usage
* [x] Update documentation
* [x] News

### DIFF
--- a/xonsh/base_shell.py
+++ b/xonsh/base_shell.py
@@ -309,7 +309,7 @@ class BaseShell(object):
         if code is None:
             return
 
-        events.on_precommand.fire(src)
+        events.on_precommand.fire(cmd=src)
 
         env = builtins.__xonsh_env__
         hist = builtins.__xonsh_history__  # pylint: disable=no-member
--- a/xonsh/base_shell.py
+++ b/xonsh/base_shell.py
@@ -372,7 +372,7 @@ class BaseShell(object):
             builtins.__xonsh_env__['PWD'] = cwd      # track it now
             if old is not None:
                 builtins.__xonsh_env__['OLDPWD'] = old  # and update $OLDPWD like dirstack.
-            events.on_chdir.fire(old, cwd)              # fire event after cwd actually changed.
+            events.on_chdir.fire(olddir=old, newdir=cwd)              # fire event after cwd actually changed.
 
     def push(self, line):
         """Pushes a line onto the buffer and compiles the code in a way that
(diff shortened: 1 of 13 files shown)


laerus:
+1

xonsh/base_shell.py
```diff
@@ -375,7 +375,7 @@ def _fix_cwd(self):
             builtins.__xonsh_env__['PWD'] = cwd      # track it now
             if old is not None:
                 builtins.__xonsh_env__['OLDPWD'] = old  # and update $OLDPWD like dirstack.
-            events.on_chdir.fire(old, cwd)              # fire event after cwd actually changed.
+            events.on_chdir.fire(olddir=old, newdir=cwd)              # fire event after cwd actually changed.
```
laerus: maybe just `old` `new` ? (naming itches :p)

xonsh/events.py
```diff
@@ -49,37 +63,41 @@ def __call__(self, handler):
         rtn : callable
             The handler
         """
-        #  Using Pythons "private" munging to minimize hypothetical collisions
+        #  Using Python's "private" munging to minimize hypothetical collisions
         handler.__validator = None
+        if debug_level():
+            if not has_kwargs(handler):
```
laerus: shouldnt this check also run when `XONSH_DEBUG=0`?

xonsh/events.py
```diff
@@ -8,11 +8,25 @@
...
 
 
+def has_kwargs(func):
+    return any(p.kind == p.VAR_KEYWORD for p in inspect.signature(func).parameters.values())
+
+
+def debug_level():
+    if hasattr(builtins, '__xonsh_env__'):
+        return builtins.__xonsh_env__.get('XONSH_DEBUG')
+    # FIXME: Under py.test, return 1(?)
```
laerus: i dont think this module should worry about pytest

scopatz:
This looks great to me! Thanks a million @astronouth7303!

scopatz:
Thanks for reviewing @laerus!

xonsh/base_shell.py
astronouth7303: `new` always feels like a keyword to me, even if it isn't in Python.

xonsh/events.py
astronouth7303: Shouldn't, but if you've got a better way to deal with it, I'm ears.

xonsh/events.py
astronouth7303: It's not required to run, and does technically slow things down. And it's not like end-users could do something about it.

PR #2100, opened 2017-01-16
Traceback logfile expand path

aweltsch:
I moved `built_ins.expand_path` to `tools.expand_path`. I also changed the implementation of
`tools.expandpath` (now `tools._expandpath`) to use `tools.expand_path`. To preserve the distinct behaviour, I added the argument `expand_user=True` to `tools.expand_path` which is checked before performing user/tilde expansion.

Still don't know a good way to test this, since we can not guarantee, that some file like `~/xonsh.log` will be writable by the current user.


xonsh/tools.py
```diff
@@ -81,14 +81,33 @@ def __init__(self, returncode, command, output=None, stderr=None,
         self.completed_command = completed_command
 
 
-def expandpath(path):
+def expand_path(s, expand_user=True):
```
mitnk: Ideas on adding test cases(for expand_user part): like `p = expand_path('~/foo/bar.txt'); assert p.endswith('/foo/bar.txt') and not p.startswith('~')`.

mitnk:
Thanks @aweltsch, I think this PR is good to go if you could add some basic unit tests for `tools.expand_path()`.

scopatz:
ping @aweltsch , any movement on @mitnk's tests request? It would be nice to merge this in so we have it for the next micro release.

aweltsch:
@scopatz , sorry I was occupied with a lot of stuff for university. I will continue to work on it today.

aweltsch:
I added some tests that test different combinations of EXPAND_ENV_VARS and expand_user. I hope those test cases are enough for now. I'm not sure why circleci fails. I can't see a relation to my changes.

scopatz:
Hi @aweltsch - that is a bug that has been fixed on master and is not your fault. Could you merge from master please? That should clear this up.

aweltsch:
Merge is done.

mitnk:
@aweltsch Thanks a lot for spending time on fixing this issue!

PR #2101, opened 2017-01-18
Add a vox custom prompt section

### LINKED ISSUE #2090: Vox prompt variable
`vox` is cool and it's handy to be able to show the current venv name in the prompt like `virualenv` does.

So far I'm using this in my `$PROMPT` function:
```python
venv = __xonsh_env__.get('VIRTUAL_ENV')                                                     
venvname = '({})'.format(os.path.basename(venv)) if venv else '' 
```

Do you guys think this is worth turning into a prompt variable?

tgoodlet:
Aids those of use who like custom prompts but also want the same
prompt behaviour as `virtualenv`.

Resolves #2090

### DIFF
--- a/docs/python_virtual_environments.rst
+++ b/docs/python_virtual_environments.rst
@@ -74,3 +74,16 @@ To see all available commands, run ``vox help``, ``vox --help``, or ``vox -h``::
 
         vox help (-h, --help)
             Show help
+
+
+``virtualenv`` like prompt
+--------------------------
+Although it's included in the default prompt, you can customize your prompt
+to automatically update in the same way as ``virtualenv``.
+
+Simply add the ``'{env_name}'`` variable to your ``$PROMPT``::
+
+    $PROMPT = '{env_name: {}}' + restofmyprompt
+
+Note that you do **not** need to load the ``vox`` xontrib for this to work.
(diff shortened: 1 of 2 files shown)


astronouth7303:
It should probably be noted that this is included in the default prompt and does _not_ require loading the vox xontrib.

(No news is needed for web-only changes.)

tgoodlet:
@astronouth7303 good point. So maybe just enter with "Although it's included in the default prompt, ..."?

I don't see a default prompt section in the docs I can link to however.

tgoodlet:
@astronouth7303 made the changes I think you wanted?

If there's anything else let me know :)

astronouth7303:
It looks fine, other than the tests being weird.

gforsyth:
Test failures are due to incompatibilities with the mpl xontrib and matplotlib 2, which is now live in the conda repos. 

I guess we should pin matplotlib to 1.5.1 or thereabouts until we've updated the xontrib

scopatz:
Thanks @tgoodlet! I merged since the MPL bugs were addressed by @gforsyth and this was a docs only change.

tgoodlet:
@scopatz :+1:

PR #2109, opened 2017-01-26
changed history indexing api

laerus:
made history indexing a lot simpler now returns a `SimpleNamespace` `HistoryEntry` with the corresponding fields.

a try to redeem my previously failed attempt to make the indexing API
resemble bash's history shortcuts, no need for this now that we have command transformations.

maybe this is also a chance to make this the standard way to get command entries from history
for example instead of `__xonsh_history__.inps[-1]` we can now do the same with
`__xonsh_history__[-1].cmd`


laerus:
where are ma tests...

astronouth7303:
Given the possibility of adding pre- and post-transformation items to the data fed to history, should `namedtuple` be used? Or don't document it as indexable? (This was the impetus for #2094.)

laerus:
not sure what's the best approach.. this PR can wait till after #2094 is merged, no pressure.

astronouth7303:
#2094 shouldn't directly affect this. I'm just saying future proofing.

laerus:
so should we not make history indexable? we can always changed namedtuple for a class.

astronouth7303:
(I mean, unless `SimpleNamespace` is a poor choice for other reasons.)

scopatz:
Thanks for putting this in @laerus! @astronouth7303 I think that namedtuples are more memory efficient than simple namespaces because they are backed by slots rather than dicts.  Both seem OK to me.

astronouth7303:
There's pros and cons for each. I think `SimpleNamespace` is slightly more CPU efficient because it goes directly to the dict instead of detouring through a property descriptor.

scopatz:
Wow yeah, it is almost twice as fast:

```console
scopatz@artemis ~ $ import types
scopatz@artemis ~ $ from collections import namedtuple
scopatz@artemis ~ $ T = namedtuple('T', ['x'])
scopatz@artemis ~ $ t = T(42)
scopatz@artemis ~ $ n = types.SimpleNamespace()
scopatz@artemis ~ $ n = types.SimpleNamespace(x=42)
scopatz@artemis ~ $ t.x
42
scopatz@artemis ~ $ n.x
42
scopatz@artemis ~ $ timeit! t.x
10000000 loops, best of 3: 52.4 ns per loop
scopatz@artemis ~ $ timeit! n.x
10000000 loops, best of 3: 29.6 ns per loop
```

astronouth7303:
Once again proving that function calls in Python can be surprisingly expensive.

laerus:
sad but what can you do, better than 1000 loc functions.

no idea why circle is not happy.

scopatz:
Circle looks like it just has a couple of small flake8 issues.

laerus:
at this point maybe we should just use a plain class with slots? i like the contrain on the fields.

astronouth7303:
Eh, it's the same level of constraining as like 99% of Python objects, including a great deal of xonsh.

laerus:
i mean when creating a HistoryEntry object, atm `__init__` is overriding `SimpleNamespace` functionality.

laerus:
seems that  instantiating a `SimpleNamespace` is faster than instantiating a custom class (no idea why).
I think this is ready to go in.

scopatz:
LGTM but I'll leave it for @astronouth7303

astronouth7303:
It's probably faster because `SimpleNamespace` is 100% C.

Does subclassing `SimpleNamespace` cancel that performance boost.

laerus:
subclassing makes it 10% slower but we get the name on the repr..

laerus:
ping!

scopatz:
@astronouth7303 Any objections to merging this in?

Issue #2111, opened 2017-01-26
Documentation Change: Use bash-completion2 in OSX


EntilZha:
Creating an issue to document the issue/reason for changing documentation, so I can link to it from the documentation before submitting a PR, and if anyone runs into the issue they can see the solution.

The original issue came up on macOS 10.12.2 using bash 4.4, xonsh 0.5.2, and bash completion 1.3 (installed via `brew install bash-completion` as per the docs). The basic issue is xonsh sometimes expanding `~` when it shouldn't or doing it incorrectly. Xonsh tab completed the following when `prompt-toolkit` was installed:

1. `ls ~/Down` -> `ls ~/Downloads`
2. `vim ~/Down` -> `~/Down/Users/pedro/Downloads/`

Similar behavior to (1) was observed with most commands, but with `emacs` behavior (2) was observed. When using `readline` tab completed to `/Users/pedro/Downloads/` which works but not desirable to expand the `~`.

After running these commands to upgrade bash completion to v2 the problem went away and expected behavior re-emerged

```
brew unlink bash-completion
brew install bash-completion2
```

I'll be submitting a PR with the documentation change in a moment.

scopatz:
Thanks @EntilZha, feel free to close this

EntilZha:
Thanks.

PR #2112, opened 2017-01-26
Change OSX Installation instructions to use bash-completion2

EntilZha:
Info on PR is at https://github.com/xonsh/xonsh/issues/2111


scopatz:
Thanks @EntilZha! Not sure why matplotlib is failing here, but this is a documentation-only change and so shouldn't affect that.

Issue #2099, opened 2017-01-16
Jupyter kernel: I/O operation on closed file


kalekundert:
I had to make an awful hack to get the Jupyter Xonsh kernel working, I'm hoping somebody can either find a real solution or just tell me that I'm doing something wrong.

When I enter a python-mode command (e.g. "1+1") into jupyter, everything works as expected.  When I enter a shell-mode command (e.g. "ls"), I get the following stack trace (in the notebook server, not the client):

```
[IPKernelApp] ERROR | Exception in message handler:
Traceback (most recent call last):
  File "/home/kale/.local/lib/python3.6/site-packages/xonsh/base_shell.py", line 319, in default
    run_compiled_code(code, self.ctx, None, 'single')
  File "/home/kale/.local/lib/python3.6/site-packages/xonsh/codecache.py", line 65, in run_compiled_code
    func(code, glb, loc)
  File "<xonsh-code>", line 1, in <module>
  File "/home/kale/.local/lib/python3.6/site-packages/ipykernel/displayhook.py", line 34, in __call__
    sys.stdout.flush()
  File "/home/kale/.local/lib/python3.6/site-packages/xonsh/base_shell.py", line 170, in flush
    self.std.flush()
  File "/home/kale/.local/lib/python3.6/tempfile.py", line 709, in flush
    self._file.flush()
ValueError: I/O operation on closed file.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/kale/.local/lib/python3.6/site-packages/ipykernel/kernelbase.py", line 228, in dispatch_shell
    handler(stream, idents, msg)
  File "/home/kale/.local/lib/python3.6/site-packages/ipykernel/kernelbase.py", line 390, in execute_request
    user_expressions, allow_stdin)
  File "/home/kale/.local/lib/python3.6/site-packages/xonsh/jupyter_kernel.py", line 55, in do_execute
    shell.default(code)
  File "/home/kale/.local/lib/python3.6/site-packages/xonsh/base_shell.py", line 328, in default
    print_exception()
  File "/home/kale/.local/lib/python3.6/site-packages/xonsh/tools.py", line 630, in print_exception
    sys.stderr.write('xonsh: For full traceback set: '
  File "/home/kale/.local/lib/python3.6/site-packages/xonsh/base_shell.py", line 165, in write
    self.std.write(std_s)
  File "/home/kale/.local/lib/python3.6/tempfile.py", line 766, in write
[...]

PR #2113, opened 2017-01-26
potential fix for jupyter

scopatz:
I think we were redirecting too much. See #2099.

### DIFF
--- a/news/jk.rst
+++ b/news/jk.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Jupyter kernel was no longer properly running subprocess commands.
+  This has been fixed.
+
+**Security:** None
--- a/tests/test_mpl.py
+++ b/tests/test_mpl.py
@@ -7,6 +7,8 @@ plt = pytest.importorskip("matplotlib.pyplot")
 
 from xontrib import mplhooks
 
+skip_if_mpl2 = pytest.mark.skipif(matplotlib.__version__.startswith('2'),
+                                  reason='Bug in matplotlib v2')
 
 # some default settings that are temporarily changed by mpl
 FONT_SIZE    = 22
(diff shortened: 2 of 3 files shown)


gforsyth:
Hey @scopatz -- I did this, too, and it definitely makes things work again, which is good, but should subprocess commands print to Outputs or stdout?  I can't remember the old behavior

scopatz:
I think the intended behaviour is outputs, right?  Like you wouldn't want to have to go and look at the terminal if you are in the notebook. Also, this is handled by ipykernel internally since they are the ones redirecting stdout & stderr to outputs. We are just reporting it back.

scopatz:
Also, great minds think alike :)

gforsyth:
Right, so should the outputs here have `Out[1]`, `Out[2]`, `Out[3]` by them?

2017-01-26-170513_1071x314

Obviously returning the jupyter kernel to any level of usability is better than the current state.

scopatz:
Ahh right.  Let me dig into that.

scopatz:
Ok, I couldn't figure out how to get `Out` back.  I am not sure it was ever there, actually.  I think it has something to do with changing `stream` to a different message type, but I couldn't figure out which message was the appropriate one (the [docs didn't help much](http://jupyter-client.readthedocs.io/en/latest/messaging.html)) and looking through IPython and ipykernel I couldn't find what value was really being set. @Carreau probably knows ;).  But for the time being, I think this is an improvement.

Carreau:
Hum, haven't touched that in a while. I'll have a look.

scopatz:
I think this is worth merging in even without the Out fix for the time being.

melund:
On windows, I don't see any output. But the problem is the same on Master. 

However, this branch does fix `which ls`, which didn't work on master. So I am OK with merging..

image

scopatz:
Yeah I think the Out entries have always been missing.

melund:
> Yeah I think the Out entries have always been missing.

I believe you are right. But the output itself is missing on Windows :( 

This PR still fixes some issues with `which ls` failing. So I think it is OK to merge if you have no ideas why it is misbehaving on Win. Then I will create a new separate issue for it.

scopatz:
Yeah, I think creating a separate PR would be best

scopatz:
Err, a separate issue.

melund:
Good. I will merge

PR #2114, opened 2017-01-26
pin mpl to 1.5.3 for tests

gforsyth:
Pinning matplotlib to 1.5.3 avoids the test failures on CircleCI for now. We'll have to update the `mpl` xontrib to work with matplotlib v2.0 soon


scopatz:
Thanks @gforsyth! Yeah, I actually think this is a bug deep in matplotlib. I'll report it and may have a simple fix.

scopatz:
Submitted a fix at matplotlib/matplotlib#7959. Not sure if it is a complete fix, but it at least will get us to the next error :)

scopatz:
And travis is finally done :)

scopatz:
Thanks for the hot fix!

PR #2116, opened 2017-01-27
Remove strict pygments dependency

gforsyth:
Take... 4(?) on this.  

Allows the `prompt_toolkit` shell to work even if Pygments isn't installed. 
Doesn't allow style selection (that could be added later to work with the PTK style framework instead of the Pygments one) but selects ansicolor style from the terminal settings. 
Syntax highlighting doesn't work without Pygments.

I've checked as many things as I can think to and there's no loss of functionality to the Pygments-enabled version or the readline shell.  I'm happy to rearrange things or rename them as others see fit.  "style_tools" isn't exactly inspired.


astronouth7303:
Don't forget to update `setup.py` (extras) and the documentation.

gforsyth:
Does it make sense to create `xonsh.style` and put all of the style related stuff in there? It would be `pyghooks`, `ansi_colors`, `style_tools`

scopatz:
Thanks for keeping this moving @gforsyth! I think conceptually `xonsh.style` makes sense.  However, pyghooks is something that we we lazily import because otherwise we would have to lazily create many of the classes in that module.  Neither here nor there, but it would be some restructuring.  Maybe we need a `lazyclass` decorator? IIRC most of the data in `ansi_colors` is already lazily created. I guess I think that `xonsh.style` could be a separate PR :)

scopatz:
This LGTM @gforsyth. Let us know when you think it is ready.

gforsyth:
I agree that there's a lot more that can be done, but it can be done separately. This at least gets us to a place where `pygments` isn't tied in unequivocally everywhere, which seems like a better jumping off point for further changes. 
A `lazyclass` decorator sounds like a great idea to me.  
I think this is ready to go.

scopatz:
Sweet. Glad to have this in!

PR #2117, opened 2017-01-28
Add `xip` alias

### LINKED ISSUE #2110: Add `xip`
Add a core alias `xip` ("kip") that will always run pip in xonsh's python environment (as opposed to the system, the current venv, or any other).

The interface is exactly the same as `pip`.

This is a problem because some tools (eg, homebrew) will install xonsh in an isolated environment, which makes installing xontribs or shell-accessible modules challenging.

The goal is to make it so that eg `xip install requests` will make the `requests` module usable from the shell context.

As a bonus: This should also try to detect if a `sudo` is necessary and automatically use it.

The alias is approximately `@(sys.executable) -m pip @(argv)`.

astronouth7303:
Allows easily calling `pip` for xonsh itself. Fixes #2110.

* [x] Automatically call `sudo` if needed.
* [ ] Tests

### DIFF
--- a/docs/aliases.rst
+++ b/docs/aliases.rst
@@ -126,6 +126,12 @@ Provides an interface to printing lines of source code prior to their execution.
 .. command-help:: xonsh.tracer.tracermain
 
 
+``xip``
+=================
+Runs the ``pip`` package manager for xonsh itself. Useful for installations where xonsh is in an
+isolated environment (eg homebrew). Pronounced "kip".
+
+
 ``xonfig``
 =================
 Manages xonsh configuration information.
--- a/news/xip.rst
+++ b/news/xip.rst
@@ -1,0 +1,13 @@
+**Added:**
+
+* Add alias ``xip`` ("kip") so that xonsh's Python environment (whatever that is) can be modified.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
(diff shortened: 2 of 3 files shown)


astronouth7303:
This really needs someone on homebrew to test it. And no, I haven't the foggiest idea how to get a dev version in homebrew.

scopatz:
This looks good to me. Why do you think we need the homebrew test? I tested it with conda and it did pick up the conda pip it looks like:

```console
scopatz@artemis ~ $ which xip
/home/scopatz/miniconda3/bin/python3 -m pip
```

scopatz:
Thanks again!

astronouth7303:
homebrew is the major example where xonsh is installed in an isolated environment by default.

I'm also not sure how to handle the case on Windows if Python is installed system-wide. eg, I think you'd need admin permissions to install modules. But I'm not sure how'd you would have installed xonsh in the first place, then.

There's a lot of different ways Python (and therefore xonsh) get installed, and while this will _probably_ work, I want to double check.

scopatz:
Makes sense to me.

astronouth7303:
So I'm not getting anyone testing on homebrew. Do we just want to merge this and hope it's right?

scopatz:
I am OK with that. Thanks @astronouth7303

PR #2125, opened 2017-01-29
wrong gray used in ANSI pygments

scopatz:
I don't think this needs a news entry

### DIFF
--- a/xonsh/pyghooks.py
+++ b/xonsh/pyghooks.py
@@ -484,9 +484,20 @@ if hasattr(pygments.style, 'ansicolors'):
         Token.Scrollbar: 'bg:#ansidarkgray',
         Token.Scrollbar.Button: 'bg:#ansiblack',
         Token.Scrollbar.Arrow: 'bg:#ansiblack #ansiwhite bold',
-        Token.AutoSuggestion: '#ansilightgray',
-        Token.Aborted: '#ansilightgray',
+        Token.AutoSuggestion: '#ansidarkgray',
+        Token.Aborted: '#ansidarkgray',
     }
+    if not ON_WINDOWS:
+        # light / dark gray seem to be swapped on linux/posix for some reason
+        # windows apepars to be correct. The following should be removed if
+        # this is ever fixed.
+        PTK_STYLE.update({
+            Token.Menu.Completions: 'bg:#ansidarkgray #ansiblack',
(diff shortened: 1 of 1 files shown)


melund:
Hmm. I am not sure this works well on Windows. Let me check...

melund:
Well. On windows, the `#ansilightgray` is the normal color, so with this we can't distinguish between suggestions and normal text. 

This is how it used to look: 
image

And this is now. 

image

I think I was the one who touched this last, and the colors are deliberate. How does it look on windows? Do you have pygments 2.2?

scopatz:
Interesting @melund, I am seeing the exact opposite on linux. The current master looks like normal text and this PR grays it out. And yes, I am on pygments v2.2 and this problem only started after I upgraded.

scopatz:
light/dark gray might be swapped on linux for some reason.

scopatz:
Ok @melund, I have updated this PR.

melund:
Wait. This was a bug in prompt_toolkit, which I fixed three months back. It has looked good for me because I was running the development version of prompt-toolkit. 

https://github.com/jonathanslenders/python-prompt-toolkit/commit/a9fc8981be2d4465f309e878216f2797a2bbaf0e

I would have thought it should have been out by now, but it seems @jonathanslenders have been a little slow with the PTK releases lately :)  It will work correctly as soon as new PTK release is out. 

@jonathanslenders is there anything we can do to help procure a new release?

@scopatz .. Otherwise we will do you first temp fix until PTK is updated.

jonathanslenders:
Hi @melund, I didn't realise that it was that long ago. I'll try to push a new release tonight.

(FYI: Right now, I'm mostly working on the 2.0 branch of prompt_toolkit, which refactors some of the "internal" architecture. It's mostly relevant for full-screen applications. 2.0 is still unstable and backwards incompatible, but I'll send a merge request to Xonsh before I release 2.0. Probably in a couple of months.)

scopatz:
Thanks @jonathanslenders, a new release would be most appreciated! xonsh+ptk are really hard to use right now because of this issue. The completion menus are unreadable and the autosuggestion is hard to distinguish from normal text.

jonathanslenders:
Hi @scopatz, a new version (1.0.10) has been pushed.
Please let me know if you experience any issues!
Thanks!

scopatz:
It seems to work great for me! Thanks to both of you!

Issue #344, opened 2015-08-03
Python completion using jedi


asmeurer:
See http://jedi.jedidjah.ch/en/latest/. It shouldn't be hard to get the basic thing working, since it's designed to be used as a library. 

Interestingly, Jedi's parser is designed to support partial parsing for the cases when you are writing code that isn't completely syntactically correct. As a result, it already seems to work on Xonsh code (at least emacs-jedi is giving me completions for my .xonshrc, which has dollar signs all over the place).  It doesn't complete on environment variables, though. I don't know if Jedi can be easily extended with the full xonsh grammar.

scopatz:
Hi @asmeurer - given the failed attempt in #2127 and the merger of a replacement in #2153. I am going to close this. Feel free to reopen if you think it is needed. I think if someone really wanted jedi completions for some reason, the right place would be in a xontrib, perhaps in jedi itself.

asmeurer:
Jedi is way smarter than just command line completion. If you are just using xonsh for shell stuff it probably is more than you need, but if you use it as a Python REPL, it's very nice. Basically, it lets you write any kind of expression, type `.` at the end, and have a decent hope of getting actual completions after the `.` (without literally evaluating the expression first).

scopatz:
Yeah, my claim isn't that jedi isn't valuable and smart with respect to attr access. Rather that it didn't seem valuable enough for the additional completions in provided while at the command line. On my desktop at home it seemed like it was exceeding my personal 0.1 sec limit for completions. And it definitely didn't seem to work with functions defined in the shell session itself (which was one of the main things I was looking for).

I think we still could have jedi completion, if we wanted, but it should probably be in a xontrib. If people want it, they should have to enable it in my opinion, because it comes with a performance penalty. Such a xontrib is basically written in #2127, if anyone wants to take up the banner.

asmeurer:
I just checked and `??` doesn't work on functions defined in the shell either. Assumedly the same sort of fix would be needed to solve either one.

scopatz:
I am not sure that the semantics of `??` make sense in that case, just like they don't for C-extensions. Feel free to open up an issue for that as there is probably some history hack we can get away with there. 

I am not sure how we'd express that to jedi, but maybe there is a way with a custom file-like object that makes history look like a single file.

astronouth7303:
Could we store the source on the function (or code) object and then use that if it's defined? And presumably extend Jedi to do the same.

scopatz:
[...]

PR #2127, opened 2017-01-29
Added jedi completions

scopatz:
This is an attempt to satisfy #344.  However, I am not sure that it really should go in (yet?) for a couple of reasons. 

1. Jedi seems super slow (like 15 ms for simple completions). I think this might be because it is built for a use case when there is a file being edited.  

2. The interpreter interface seems like an afterthought and somewhat buggy. For example if you dynamically define a function, jedi doesn't seem like it can complete kwargs from that function.

3. The only thing we really get from jedi is the function parameter names. Imports, attribute access, etc, we have other tools for.

All things said, I think this is a lot of extra parsing and another optional dependency for not a whole lot of differential value. Also it is hard/annoying to put this in a xontrib since to even get that differential value it has to hook into xonsh's existing python completer.

I think we could do better ourselves with a judicious use of inspect and `__xonsh_ctx__`. Let's keep this open, though, until we have an alternative.

### DIFF
--- a/news/jedi.rst
+++ b/news/jedi.rst
@@ -1,0 +1,13 @@
+**Added:**
+
+* Added jedi completion to Python-mode completions.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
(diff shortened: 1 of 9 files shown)


astronouth7303:
The main reason to use this (imho) would be to reduce our maintenance load; ie the less xonsh code there is, in theory the less we have to debug and maintain against.

Not that there's also maintenance risks with 3rd party code. But considering what a beast even the core parsing and execution is (xonsh's MVP), efforts to reduce that load should be welcome.

Of course, if we have to maintain our own fork to make it useful, it's probably worse.

scopatz:
Good points.  I do think that performance counts though, too.

scopatz:
Closing in favor of #2153

PR #2131, opened 2017-01-30
Event on_envvar

laerus:
fired after an envvar has changed

@astronouth7303 ping!

### DIFF
--- a/news/event-envvar.rst
+++ b/news/event-envvar.rst
@@ -1,6 +1,7 @@
 **Added:**
 
-* new event ``on_envvar`` that fires after an enviroment variable has been changed.
+* ``on_envvar_new`` event that fires after a new envvar is created.
+* ``on_envvar_change`` event that fires after an envvar is changed.
 
 **Changed:** None
 
--- a/news/event-envvar.rst
+++ b/news/event-envvar.rst
@@ -1,0 +1,13 @@
+**Added:**
+
+* new event ``on_envvar`` that fires after an enviroment variable has been changed.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/tests/test_environ.py
+++ b/tests/test_environ.py
@@ -184,3 +184,20 @@ def test_event_on_envvar_change_from_none_value(xonsh_builtins):
     env['TEST'] = 1
 
     assert share == ['TEST', None, 1]
+
+
+@pytest.mark.parametrize('val', [1, None, True, 'ok'])
+def test_event_on_envvar_change_no_fire_when_value_is_same(val, xonsh_builtins):
+    env = Env(TEST=val)
+    xonsh_builtins.__xonsh_env__ = env
+    share = []
+    # register
+    @xonsh_builtins.events.on_envvar_change
+    def handler(name, oldvalue, newvalue, **kwargs):
+        share.extend((name, oldvalue, newvalue))
+
+
+    # trigger
+    env['TEST'] = val
+
+    assert share == []
--- a/tests/test_environ.py
+++ b/tests/test_environ.py
@@ -140,3 +140,19 @@ def test_locate_binary_on_windows(xonsh_builtins):
         assert locate_binary('file2') == os.path.join(tmpdir, 'FILE2.BAT')
         assert locate_binary('file2.bat') == os.path.join(tmpdir, 'FILE2.BAT')
         assert locate_binary('file3') is None
+
+
+def test_event_on_envvar(xonsh_builtins):
+    env = Env(TEST=0)
+    xonsh_builtins.__xonsh_env__ = env
+    share = []
+    # register
+    @xonsh_builtins.events.on_envvar
+    def handler(name, oldvalue, newvalue, **kwargs):
+        share.extend((name,oldvalue, newvalue))
+
+    # trigger
+    env['TEST'] = 1
+
+    assert share == ['TEST', 0, 1]
+    # assert len(share) > 0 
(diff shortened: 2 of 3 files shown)


laerus:
i feel this can lead to a recursion?

tests/test_environ.py
```diff
@@ -140,3 +140,33 @@ def test_locate_binary_on_windows(xonsh_builtins):
...
+
+
+def test_event_on_envvar(xonsh_builtins):
+    env = Env(TEST=0)
+    xonsh_builtins.__xonsh_env__ = env
+    share = []
+    # register
+    @xonsh_builtins.events.on_envvar
+    def handler(name, oldvalue, newvalue, **kwargs):
+        share.extend((name,oldvalue, newvalue))
```
astronouth7303: PEP8 moment.

news/event-envvar.rst
```diff
@@ -0,0 +1,13 @@
+**Added:**
+
+* new event ``on_envvar`` that fires after an enviroment variable has been changed.
```
astronouth7303: "New" (alt-grammar moment)

xonsh/environ.py
```diff
@@ -42,6 +43,13 @@
 import xonsh.prompt.base as prompt
 
 
+events.doc('on_envvar', """
+on_envvar(name: str, oldvalue, newvalue) -> None
+
+Fires after an enviromental variable has changed.
```
astronouth7303: The term is "environment variable".

tests/test_environ.py
```diff
@@ -140,3 +140,33 @@ def test_locate_binary_on_windows(xonsh_builtins):
...
+
+
+def test_event_on_envvar_new(xonsh_builtins):
+    env = Env()
+    xonsh_builtins.__xonsh_env__ = env
+    share = []
+    # register
+    @xonsh_builtins.events.on_envvar
+    def handler(name, oldvalue, newvalue, **kwargs):
+        share.extend((name,oldvalue, newvalue))
```
astronouth7303: PEP8 moment.

astronouth7303:
Hmm. Yes, it can lead to recursion if an underlying handler updates a environment variable. I'm not sure if the event system should be modified to queue up events, or just suck it up with a note in the documentation "This event will happily recurse if you modify environment variables in handlers. Design appropriately."

Oh! One thing that I will block on: Only fire if `oldvalue` != `newvalue`. This should make dealing with recursion simpler and generally cut down event firings.

astronouth7303:
Nope. We should suck it up. Queueing up events will prevent a `RecursionError`, but won't actually prevent handlers from doing the wrong thing.

laerus:
should we split this event to two `on_envvar_new` and `on_envvar_change` ?

we may need some kind of defense against wild handlers at some point.

astronouth7303:
Maybe have two events? Since envvars can have any value, there isn't really a sentinel that can be used.

We may need to defend against it in the future, but this problem is _not_ unique to `on_envvar`, so don't worry about it in this PR. (`on_chdir` has the same problem, as well as any future "change notification" events).

xonsh/environ.py
```diff
@@ -930,7 +943,12 @@ def __setitem__(self, key, val):
                     self.replace_env()
                 else:
                     os.environ[key] = ensurer.detype(val)
-        events.on_envvar.fire(name=key, oldvalue=old_val, newvalue=val)
+        if old_value is no_value:
+            events.on_envvar_new.fire(name=key, value=val)
+        else:
+            events.on_envvar_change.fire(name=key,
+                                         oldvalue=old_value,
+                                         newvalue=val)
```
astronouth7303: Still need to handle the case where `newvalue == oldvalue`.

news/event-envvar.rst
```diff
@@ -1,6 +1,7 @@
 **Added:**
 
-* new event ``on_envvar`` that fires after an enviroment variable has been changed.
+* ``on_envvar_new`` event that fires after a new envvar is created.
+* ``on_envvar_change`` event that fires after an envvar is changed.
```
astronouth7303: My preference would be for english order, not OOP order. (ie `on_new_envvar` or `on_envvar_created`), but I'm not blocking on it.

xonsh/environ.py
```diff
@@ -42,6 +43,24 @@
...
+Warning! if a handler tries to set an envvar
+you'll get a nice recursion until the limit.
+""")
+
+
+events.doc('on_envvar_change', """
+on_envvar_change(name: str, oldvalue: Any, newvalue: Any) -> None
+
+Fires after an enviroment variable is changed.
+Warning! if a handler tries to set an envvar
```
astronouth7303: I would prefer a less hopeless warning, since it'll only recurse if they make unconditional changes.

xonsh/environ.py
```diff
@@ -768,6 +768,8 @@ class Env(cabc.MutableMapping):
     def __init__(self, *args, **kwargs):
         """If no initial environment is given, os.environ is used."""
         self._d = {}
+        # sentinel value for non existing envvars
+        self._no_value = object()
```
astronouth7303: Does this produce a measurable difference?

laerus: not so really, seems tidier this way

scopatz: wouldn't None be a better sentinel?

astronouth7303: `None` is a valid value for an environment variable. As is `Ellipsis` and every other global singleton.

I would disagree on the "tidier" front, but that is pretty purely taste.

laerus:
ping!

scopatz:
Feel free to merge this whenever you think it is ready @astronouth7303

PR #2132, opened 2017-01-30
adding code coverage back

laerus:
@gforsyth ping!

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,4 +4,5 @@ omit =
   tests/*
   xonsh/__amalgam__.py
   */parser*_table.py
+  xonsh/ply/*
   setup.py
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,8 +1,9 @@
 [run]
 branch = true
+source =
+  xonsh/
 omit =
-  tests/*
-  xonsh/__amalgam__.py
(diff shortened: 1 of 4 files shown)


scopatz:
LGTM

.coveragerc
```diff
@@ -4,4 +4,5 @@ omit =
   tests/*
   xonsh/__amalgam__.py
   */parser*_table.py
+  xonsh/ply/*
```
gforsyth: Do we want coverage on ply?

gforsyth:
Hey @laerus -- i'm good with this, just the one question about whether or not to include PLY in the coverage report

.coveragerc
laerus: nope that's why i've added it in the ommit section
am i missing something?

gforsyth: My bad, early morning review is a bad idea

scopatz:
also it is worth omitting lazyasd.py and amalgamate.py since these are tested in their own repos.

laerus:
we also need to ommit all `__amalgam__.py` files in the subpackages

laerus:
i don't think this needs a news entry

gforsyth:
Thanks, @laerus!

PR #2136, opened 2017-01-31
ignore vc aliases in command_cache

### LINKED ISSUE #2135: Branch not shown if 'git' is aliased to 'hub'
I you do `brew install hub` and add `aliases['git'] = ['hub']` to ~/.xonshrc, git branches will no longer be shown in directories.

This is very strange since Hub is just a wrapper around git, and `hub branch` yields exactly the same output as `git branch`. In fact the former just invokes the latter.

gforsyth:
I don't *think* this will have any side effects.

When we build the ``$PROMPT``, we lazily check for ``git`` and ``hg``
binaries to see if we should generate a ``<branch>`` field.

At the same time ``lazy_locate_binary`` doesn't return the path to a
command if it detects that that command is an alias (since aliases won't
have a path).

In the case that ``git`` or ``hg`` is aliased (the specific case here
was replacing ``git`` with ``hub``), ``lazy_locate_binary`` will return
``None`` because the command is an alias.

This breaks the logic in the branch detection. We don't want to change
the logic in the branch detection since there are many potential impacts
on prompt generation speed, which makes for a bad user experience.

So, this is a mildly ugly hack, but it's the best option I could come up
with in order to preserve prompt speed and still keep the branch
information available for this slightly weird edge case.

I'm all ears to alternative solutions.

Should resolve #2135

### DIFF
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -14,7 +14,11 @@ import collections.abc as cabc
 
 from xonsh.platform import ON_WINDOWS, ON_POSIX, pathbasename
 from xonsh.tools import executables_in
-from xonsh.lazyasd import lazyobject
+from xonsh.lazyasd import lazyobject, LazyObject
+
+
+IGNORE_ALIAS_CMDS = LazyObject(lambda: frozenset(['git', 'hg']), globals(),
+                               'IGNORE_ALIAS_CMDS')
 
 
 class CommandsCache(cabc.Mapping):
@@ -104,6 +108,11 @@ class CommandsCache(cabc.Mapping):
             if cmd not in allcmds:
                 key = cmd.upper() if ON_WINDOWS else cmd
                 allcmds[key] = (cmd, True)
+        for key in IGNORE_ALIAS_CMDS:
+            if key in allcmds:
+                key = key.upper() if ON_WINDOWS else key
+                pathkey, _ = allcmds[key]
+                allcmds[key] = (pathkey, False)
         self._cmds_cache = allcmds
         return allcmds
 
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -14,11 +14,7 @@ import collections.abc as cabc
 
 from xonsh.platform import ON_WINDOWS, ON_POSIX, pathbasename
 from xonsh.tools import executables_in
-from xonsh.lazyasd import lazyobject, LazyObject
-
-
-IGNORE_ALIAS_CMDS = LazyObject(lambda: frozenset(['git', 'hg']), globals(),
-                               'IGNORE_ALIAS_CMDS')
+from xonsh.lazyasd import lazyobject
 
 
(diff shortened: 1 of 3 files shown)


astronouth7303:
Ewww gross. "Special cases aren't special enough to break the rules."

Is there a cleaner way to do this? New method on `CommandsCache`? I don't `CommandsCache` all that well, unfortunately.

astronouth7303:
(Especially since it's on my todo list to move things like VCS support out of core proper.)

scopatz:
Hi @gforsyth, first thanks for putting this in! 

I think maybe we could change the data structure and interface a bit to get a cleaner solution. Right now it looks like the `_cmds_cache` / `allcmds` have values that are `(string path, bool if alias)`.  

Maybe we could change this to be `(string path, alias or None or False)`. That way we'd preserve the path on the file system, if it exists, and have the alias information. After this change, we could then have another method or kwarg to `lazy_locate_binary()` that ignores the alias value. We'd then use that option when getting git etc. for the purposes indicated.

Does that make sense?

gforsyth:
Ok, this is ready for another look. Travis is cancelling all osx jobs at the moment, but the rest of the tests pass

xonsh/commands_cache.py
```diff
@@ -99,11 +99,11 @@ def all_commands(self):
             # entries at the back.
             for cmd in executables_in(path):
                 key = cmd.upper() if ON_WINDOWS else cmd
-                allcmds[key] = (os.path.join(path, cmd), cmd in alss)
+                allcmds[key] = (os.path.join(path, cmd), alss.get(key, None))
         for cmd in alss:
             if cmd not in allcmds:
                 key = cmd.upper() if ON_WINDOWS else cmd
-                allcmds[key] = (cmd, True)
+                allcmds[key] = (cmd, None)
```
melund: Why? Would this indicate that this is not an alias?

gforsyth: Ahh, that's an interesting point. In this case, there's an alias that doesn't exist in `allcmds` for some reason -- the second value of the tuple is (now) the name of the alias if it exists and the first value is the full string path of the command, so in this case, the "alias" value is the same as the cmd value. I just set it to None, but it could be `(cmd, cmd)`

melund: But on line 183 you check: 
```
(path, alias) = self._cmds_cache[cached]
             if not alias or ignore_alias:
...
```
Which would evaluate to True since `alias` would be `None`.

gforsyth: Ahh, gotcha. Yes, I agree. I'll change it.

melund:
Ok. Now I think it is good to go. Is travis ever gonna finish

gforsyth:
Eventually? It looks like at least one of the OSX runs went through, so it's doing something...

scopatz:
I don't think we should wait for travis today since they canceled all their mac jobs

scopatz:
LGTM boss. Thanks!

Issue #2134, opened 2017-01-31
Using rsync over ssh to a destination using xonsh causes protocol errors


jonathanhogg:
Hi (again!),

I just noticed today that using rsync over ssh with xonsh at the destination seems to result in various different protocol and buffer overflow errors. I suspect that this is a problem with how xonsh is feeding through the stdin/stdout of rsync in server mode. I don't think the actual errors are necessarily useful, but here's an example:

```
resistance$ rsync --archive --verbose futile.local:/working/boot/ boot/
receiving incremental file list
hard-link reference out of range: 121 (9)
rsync error: protocol incompatibility (code 2) at flist.c(798) [Receiver=3.1.2]
unexpected tag -7 [Receiver/inc]
```

If I switch back to bash as the login shell at the remote end, then the problem goes away.

Does xonsh do universal newlines conversion or similar on the output of commands? [I should be able to tell this having just had the bonnet up…]

Jonathan

gforsyth:
Hi @jonathanhogg , thanks for reporting! This sounds like it's at least related to #2098 -- it could well be a newline problem.  I should also be able to tell this, too, but I'll have to take a look to confirm.

scopatz:
I agree with @gforsyth. I think this is highly related to #2098.

jonathanhogg:
Hi @gforsyth, yes, it might well be related to that bug. I'm beginning to suspect it may actually be the use of `io.TextIOWrapper`s on the stdin/stdout. I'm still examining the code, but it looks like it is assumed in lots of places that "last" commands are always producing/consuming text and not binary data.

scopatz:
So that isn't entirely true. What is displayed to the user normally is binary, if you look in `CommandPipeline.tee_stdout()`. The value of output on the command pipeline is what is converted to unicode after the data has been streamed to the user. Of course, this shouldn't affect ssh at all because ssh is marked as unthreadable and uses Popen directly.  It very well might be a universal newlines thing, but I thought that was set to false for unthreadable commands. It is worth checking.

jonathanhogg:
@scopatz: Wrong end – the client ssh will be executed directly by rsync, that connects to a xonsh running at the remote end and then instructs *that* to run another rsync in server mode. So it'll be the piping of stdin/stdout of the rsync server through the stdin/stdout of xonsh that'll be the problem.

Sorry if I'm making assumptions about the code, I'm still trying to inhale it…

scopatz:
[...]

Issue #2098, opened 2017-01-15
xonsh breaks git for me (protocol error)


Droggelbecher:
**Environment**

The "Server" machine is a Ubuntu 14.04 (however with various packages from other releases/ppa's),
with xonsh from git like this (ill explain in a bit what server/client means here):
```
$ git log -n 1
commit 6eeac872724e2353ddeb84b5f6ffeed7c68d2996
Author: Jamie Bliss <astronouth7303@gmail.com>
Date:   Sat Jan 14 18:22:15 2017 -0500

    Re-establish `on_precommand`. (Thought I fixed this in the PR?)
$ xonsh --version
xonsh/0.5.2.dev61
```

Xonsh is set as the default shell (/usr/bin/local/xonsh set using chsh).

A recent git is installed:
```
$ git --version
git version 2.11.0
```

The following issue was observed with 2 different "client" machines, one isa macbook (git 2.11.0), the other an archlinux system (also a recent git, iirc also 2.11.x). Both observe the same issue

**The issue**

When using a "git pull" or "git push" command from one of the client machines, targeting the server machine, this error is observed:
```
Fatal: protocol error: bad line length character: 
003
fatal: The remote end hung up unexpectedly                        
```
The problem does **not** occur when I set bash as the default shell on the server machine.
Some googling turned out that something seems to meddle with the 4th byte that git transmits.
Eg. https://www.spinics.net/lists/git/msg133582.html says there is 0x01-byte expected that might be mangled somehow?

I ensured that I don't have any extra output generated that could be the cause (eg. ssh me@server echo x, just echos "x" and nothing else).

Does anyone have a clue what might be happening here?

EDIT: Formatting

mitnk:
I can reproduce this issue, but no ideas what's the reason :)

Droggelbecher:
Well thats already a relief, I was worried I might have something weird going on in my specific environment.

Is there (as a workaround) a way to tell git to use a different shell on the remote end when pushing/pulling so I don't have to login and change my default shell before every pull/push?

mitnk:
I think we need to find out the reason why xonsh caused this.

But for now, here is a workaround - (creating an aliased user which use bash as the login shell on your server](https://github.com/xonsh/xonsh/issues/1525#issuecomment-237478842). And then use `user-aliased@host` as your git remote. (If you count this as a workaround :-)

Droggelbecher:
[...]

PR #2137, opened 2017-01-31
Bad PTY

scopatz:
This is an attempt to address #2098 and #2134. It seems to work for me testing with localhost. This one was super obscure.


mitnk:
LGTM.

Issue #2141, opened 2017-02-01
Loading history fails with readline shell (on windows) 


melund:
When I start xonsh with the readline shell on windows it fails 20 seconds after showing the prompt. 

Probably something to do with loading the history on a separate thread. It doesn't happen with the PTK shell.

```
snail@sea ~ $ Exception in thread Thread-3:
Traceback (most recent call last):
  File "C:\Users\mel\Anaconda3\lib\threading.py", line 914, in _bootstrap_inner
    self.run()
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\__amalgam__.py", line 18968, in run
    if line == readline.get_history_item(i - 1):
  File "C:\Users\mel\Anaconda3\lib\site-packages\pyreadline\rlmain.py", line 154, in get_history_item
    return self.mode._history.get_history_item(index)
  File "C:\Users\mel\Anaconda3\lib\site-packages\pyreadline\lineeditor\history.py", line 52, in get_history_item
    item = self.history[index - 1]
IndexError: list index out of range
```

scopatz:
Thanks for opening this @melund. I think that this was a legitimate logic error. #2145 should fix this.

PR #2145, opened 2017-02-02
Readline len

scopatz:
Should address #2141. CC @melund

### DIFF
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -572,7 +572,9 @@ class ReadlineHistoryAdder(threading.Thread):
         i = 1
         for h in hist.all_items():
             line = h['inp'].rstrip()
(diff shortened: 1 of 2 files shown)


xonsh/readline_shell.py
```diff
@@ -572,7 +572,9 @@ def run(self):
         i = 1
         for h in hist.all_items():
             line = h['inp'].rstrip()
-            if line == readline.get_history_item(i - 1):
+            if i == 1:
+                pass
+            elif line == readline.get_history_item(i - 1):
```
mitnk:
```
if i > 1 and line == readline.get_history_item(i - 1):
```
put i test on the same line?

melund:
I don't think @mitnk comment is blocking and it fixes it for me. I will merge it in.

PR #2138, opened 2017-02-01
Docs - Tutorial - Write Your Own History Backend

mitnk:
- [x] working demo code
- [x] explain each sections

PR #2146, opened 2017-02-02
Make XONSH_HISTORY_BACKEND accept class type

mitnk:
Just a POC. Will continue if direction is ok.

With some changes like this would [simplify code in xonshrc](https://github.com/xonsh/xonsh/pull/2138#issuecomment-276942858) if the users want to [write their own history backend](https://github.com/xonsh/xonsh/pull/2138#issuecomment-276942858).

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -38,7 +38,7 @@ from xonsh.tools import (
     is_dynamic_cwd_width, to_dynamic_cwd_tuple, dynamic_cwd_tuple_to_str,
     is_logfile_opt, to_logfile_opt, logfile_opt_to_str, executables_in,
     is_nonstring_seq_of_strings, pathsep_to_upper_seq,
-    seq_to_upper_pathsep, print_color
+    seq_to_upper_pathsep, print_color, is_history_backend, to_itself,
 )
 import xonsh.prompt.base as prompt
 
@@ -176,7 +176,7 @@ def DEFAULT_ENSURERS():
     'XONSH_DEBUG': (always_false, to_debug, bool_or_int_to_str),
     'XONSH_ENCODING': (is_string, ensure_string, ensure_string),
     'XONSH_ENCODING_ERRORS': (is_string, ensure_string, ensure_string),
-    'XONSH_HISTORY_BACKEND': (is_string_or_callable, ensure_string, ensure_string),
+    'XONSH_HISTORY_BACKEND': (is_history_backend, to_itself, ensure_string),
     'XONSH_HISTORY_FILE': (is_string, ensure_string, ensure_string),
     'XONSH_HISTORY_SIZE': (is_history_tuple, to_history_tuple, history_tuple_to_str),
     'XONSH_LOGIN': (is_bool, to_bool, bool_to_str),
@@ -937,6 +937,7 @@ class Env(cabc.MutableMapping):
         if not ensurer.validate(val):
             val = ensurer.convert(val)
         # existing envvars can have any value including None
+
         old_value = self._d[key] if key in self._d else self._no_value
         self._d[key] = val
         if self.detypeable(val):
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -176,7 +176,7 @@ def DEFAULT_ENSURERS():
     'XONSH_DEBUG': (always_false, to_debug, bool_or_int_to_str),
     'XONSH_ENCODING': (is_string, ensure_string, ensure_string),
     'XONSH_ENCODING_ERRORS': (is_string, ensure_string, ensure_string),
-    'XONSH_HISTORY_BACKEND': (is_string, ensure_string, ensure_string),
+    'XONSH_HISTORY_BACKEND': (is_string_or_callable, ensure_string, ensure_string),
     'XONSH_HISTORY_FILE': (is_string, ensure_string, ensure_string),
     'XONSH_HISTORY_SIZE': (is_history_tuple, to_history_tuple, history_tuple_to_str),
     'XONSH_LOGIN': (is_bool, to_bool, bool_to_str),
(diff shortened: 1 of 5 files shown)


astronouth7303:
One question on `$XONSH_HISTORY_BACKEND`. What do they mean by "use"? For everything? Does xonsh not support parallel history backends?

mitnk:
@astronouth7303 which "use"?

Xonsh doesn't support parallel history BEs. And I don't think we need xonsh to do it (let's keep things simple on our side).

I think I understand what you want, to use two history BEs at the same time can easily achieve by user like this:
```
from xonsh.history.sqlite import SqliteHistory

class CouchDBHistory(SqliteHistory):
    def append(self, cmd):
        super().append(cmd)  # save to sqlitedb locally
        self._save_to_couchdb(cmd)  # save to your new couchdb too
```

scopatz:
> Xonsh doesn't support parallel history BEs. And I don't think we need xonsh to do it (let's keep things simple on our side).

This is a little off topic, but I think we'll eventually want to have multiple backends running.  I can think of a couple use cases personally. One that I'll probably hit eventually myself is to have a server that logs history locally and at a remote logging server. That way, even if the server goes down you still have a copy of the history.  This is kind of beyond the concern of this PR, though.

mitnk:
No "-1" received yet, so I'm pushing this PR to the code review phase. Thanks.

xonsh/history/main.py
```diff
@@ -26,7 +27,9 @@ def construct_history(**kwargs):
     """Construct the history backend object."""
     env = builtins.__xonsh_env__
     backend = env.get('XONSH_HISTORY_BACKEND')
```
mitnk: And we even could make `XONSH_HISTORY_BACKEND` also receive a class instance. But that may go too far?

scopatz:
Sorry for not getting to this sooner @mitnk. My only concern with this is (and I am not convinced it is that important) that if the envvar is set to a class, that class won't be passed on correctly to any subshells.  For example, if you open up xonsh inside of xonsh then the history class will be a string repr() in the second xonsh's environment. This repr string won't match as known history backend name. 

The reason this might not be a problem is that if the xonshrc hasn't been modified, the xonshrc will run and set it back to the actual class.

Still, I think if backend classes had a 'name' or similar class attribute, this could be used to set or persist the envvar as a string. This is another thing that has to go in the interface, but I think it is an improvement

astronouth7303:
@scopatz Is there a round-tripping precedent on other envvars?

xonsh/environ.py
```diff
@@ -176,7 +176,7 @@ def DEFAULT_ENSURERS():
     'XONSH_DEBUG': (always_false, to_debug, bool_or_int_to_str),
     'XONSH_ENCODING': (is_string, ensure_string, ensure_string),
     'XONSH_ENCODING_ERRORS': (is_string, ensure_string, ensure_string),
-    'XONSH_HISTORY_BACKEND': (is_string, ensure_string, ensure_string),
+    'XONSH_HISTORY_BACKEND': (is_string_or_callable, ensure_string, ensure_string),
```
scopatz: `ensure_string` is probably no longer the right thing here. It is probably worth writing a custom to-string converter here that reads the history classes `.name` attribute. Of course, we'd need to add the name attr to the existing backends.

mitnk:
> that class won't be passed on correctly to any subshells. For example, 
> if you open up xonsh inside of xonsh ...
> The reason this might not be a problem is that if the xonshrc hasn't been modified,
> the xonshrc will run and set it back to the actual class.

Yes, this is what I find for current situation. Is there any other subshell example this will break?

> if backend classes had a 'name' or similar class attribute, 
> this could be used to set or persist the envvar as a string

I'm not sure I get your ideas here. Doesn't this mean we need this class to be known by `xonsh.history.main.HISTORY_BACKENDS`?

xonsh/environ.py
mitnk: Oh, I see `$PROMPT` can take a string or a function, but it is defined as:
```
    'PROMPT': (is_string_or_callable, ensure_string, ensure_string),
```
So ENVs like this was not defined properly?

xonsh/history/main.py
scopatz: You could, I don't see a problem with that. It wouldn't be much different.

scopatz:
@astronouth7303 - I think almost all other envvars are roundtrippable. The only ones that aren't are liks $PROMPT, which can accept a callable.  Such callables are not converted to a string representation.

scopatz:
@mitnk I mean something like this:

```python
class JsonBackend:

    name = 'json'

    def __init__(self, ...):
        ...
```

Then you can use the name whenever you need to convert to a string.

xonsh/environ.py
scopatz: Well, it just skips it if you set it to a callable. In theory we could use pickle or somesuch to make it through the process barrier, but no one has really complained about this yet.

mitnk: Yes, that's what I mean, this `PROMPT` is like `XONSH_HISTORY_BACKEND` will be refresh to what user want them to be in xonshrc every time xonsh read xonshrc :) so no one would complain (found) assign a class type to it.

scopatz: That's right, I don't think it is a huge deal.  But we may as well do the "right" thing while we are thinking about it instead of having a hard to debug issue in the future.

mitnk:
@scopatz I still not quite understand here:
So if a user defined a custom class `CouchBE` class, and has a `name = 'couchbe'` attr. Then user did this:
```
from xxx import CouchBE
$XONSH_HISTORY_BACKEND = CouchBE
```
or you may mean:
```
$XONSH_HISTORY_BACKEND = 'couchbe'
```
Then for some reason, xonsh run a subshell and made `$XONSH_HISTORY_BACKEND` as a `'couchbe'`, then xonsh MUST need to know where `CouchBe` file located. A string `'couchbe'` doesn't tell anything about this. Then how could xonsh load `CouchBE` class back?

xonsh/environ.py
mitnk: ok, got your idea here, agreed!

mitnk:
Or do you mean we should just let users do:

```
from history_couchdb import CouchDBHistory
from xonsh.history.main import HISTORY_BACKENDS
HISTORY_BACKENDS['couchdb'] = CouchDBHistory
$XONSH_HISTORY_BACKEND = 'couchdb'
```

and not attempt to simplify it to

```
from history_couchdb import CouchDBHistory
$XONSH_HISTORY_BACKEND = CouchDBHistory
```

Then it is a string, and users need to know a second xonsh stuff `HISTORY_BACKENDS`?

scopatz:
Ohh I see the issue. You are right that the history class would still potentially not be registered in the subshell. This is would then cause a bug of its own unless we did something very magical (with pickle or saved an import string). It probably isn't worth it as it is likely to cause more bugs than it avoids. 

OK, I am totally on board with skipping it if it is a class.  Thanks for putting up with the back and forth.  Let us know when you think this is ready to go in.

mitnk:
Thanks for the explanations too.

I've updated the code a bit. It's ready for code review now.

mitnk:
code froze.

scopatz:
Thanks @mitnk! This is great to have!

PR #2147, opened 2017-02-03
Fix bash completions on Windows when WSL is installed

### LINKED ISSUE #2140: git completion on windows broken
Git completions on windows has stopped working:
```
snail@sea ~\Documents\xonsh master $ $BASH_COMPLETIONS
['C:\\Program Files\\Git\\usr\\share\\bash-completion\\bash_completion', 
'C:\\Program Files\\Git\\mingw64\\share\\git\\completion\\git-completion.bash']
```

I guess it could be some of the recent fixes to bash completion scripts. Do we require bash completion 2.0? 

On windows we use the git bash completions that come bundled with 'Git for Windows'. I don't think that 2.0

melund:
This fixes the problem I had with bash completion on windows because 'Windows Subsystem for Linux' was installed. 

Fixes #2140

### DIFF
--- a/xonsh/platform.py
+++ b/xonsh/platform.py
@@ -292,13 +292,13 @@ def windows_bash_command():
         out = subprocess.check_output([bash_on_path, '--version'],
                                       stderr=subprocess.PIPE,
                                       universal_newlines=True)
-        if 'pc-linux-gnu' in out.splitlines()[0]: 
+        if 'pc-linux-gnu' in out.splitlines()[0]:
             gfwp = git_for_windows_path()
             if gfwp:
                 bashcmd = os.path.join(gfwp, 'bin\\bash.exe')
                 if os.path.isfile(bashcmd):
                     wbc = bashcmd
-        else: 
+        else:
             wbc = bash_on_path
     return wbc
 
--- a/xonsh/platform.py
+++ b/xonsh/platform.py
@@ -286,7 +286,7 @@ def windows_bash_command():
     # Check that bash is on path otherwise try the default directory
     # used by Git for windows
     wbc = 'bash'
-    bash_on_path = builtins.__xonsh_commands_cache__.lazy_locate_binary('bash', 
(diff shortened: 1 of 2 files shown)


xonsh/platform.py
```diff
@@ -6,14 +6,14 @@
...
 import pathlib
+import builtins
 import platform
 import functools
 import subprocess
 import importlib.util
 
 from xonsh.lazyasd import LazyBool, lazyobject, lazybool
 
-
```
mitnk: I guess two black lines are for pep8, or we will break tools like `flake8` etc.

scopatz: Yeah that seems right to me.

scopatz:
I don't think we need to wait for Travis today. Thanks for putting this in @melund! LGTM

xonsh/platform.py
```diff
@@ -285,16 +285,21 @@ def windows_bash_command():
...
-        subprocess.check_call([wbc, '--version'],
-                              stdout=subprocess.PIPE,
-                              stderr=subprocess.PIPE)
-    except (FileNotFoundError, subprocess.CalledProcessError):
-        gfwp = git_for_windows_path()
-        if gfwp:
-            bashcmd = os.path.join(gfwp, 'bin\\bash.exe')
-            if os.path.isfile(bashcmd):
-                wbc = bashcmd
+    bash_on_path = builtins.__xonsh_commands_cache__.lazy_locate_binary('bash')
```
scopatz: Maybe we want to add `ignore_alias=True` here since we want the real path

xonsh/platform.py
melund: Thanks. I wondered what it was complaining about.

xonsh/platform.py
melund: Good idea.

melund:
I away from my computer until tomorrow, but I tired to make the commit on my phone....

scopatz:
Looks like you got it. Thanks!

Issue #2148, opened 2017-02-04
Auto history GC doesn't seem to work.


melund:
I have noticed that history seemed to take longer and longer time before starting up. It would take several minutes before history from previous sessions became available. My configuration is:

```
snail@sea ~ $ $XONSH_HISTORY_SIZE
(8128, 'commands')
```

But when I looked for the session files I had more than 8700 session files. 
```
snail@sea ~ $ histfiles= g`~/.local/share/xonsh/xonsh-*`
snail@sea ~ $ len(histfiles)
8735
```
I don't know if that many files would be a problem on Linux, but it is on win. 

Running `history gc` manually to limit the files simply stalls xonsh: 
```
snail@sea ~\.local\share\xonsh $ history gc --size 100 files
```
I tried to ctrl-z/ctrl-d, and after 10 sec it gave me this tb, but it didn't give me control back. So I ended up closing the terminal manually. 

``` 
Traceback (most recent call last):
  File "C:\Users\mel\Documents\xonsh\xonsh\base_shell.py", line 322, in default
    run_compiled_code(code, self.ctx, None, 'single')
  File "C:\Users\mel\Documents\xonsh\xonsh\codecache.py", line 65, in run_compiled_code
    func(code, glb, loc)
  File "<xonsh-code>", line 1, in <module>
  File "C:\Users\mel\Documents\xonsh\xonsh\built_ins.py", line 859, in subproc_captured_hiddenobject
    return run_subproc(cmds, captured='hiddenobject')
  File "C:\Users\mel\Documents\xonsh\xonsh\built_ins.py", line 827, in run_subproc
    command.end()
  File "C:\Users\mel\Documents\xonsh\xonsh\proc.py", line 1908, in end
    for _ in self.tee_stdout():
  File "C:\Users\mel\Documents\xonsh\xonsh\proc.py", line 1833, in tee_stdout
    for line in self.iterraw():
  File "C:\Users\mel\Documents\xonsh\xonsh\proc.py", line 1796, in iterraw
    time.sleep(timeout * cnt)
  File "C:\Users\mel\Documents\xonsh\xonsh\proc.py", line 749, in _signal_int
    self.send_signal(signum)
  File "C:\Users\mel\Documents\xonsh\xonsh\proc.py", line 882, in send_signal
    time.sleep(1e-7)
  File "C:\Users\mel\Documents\xonsh\xonsh\proc.py", line 749, in _signal_int
    self.send_signal(signum)
[...]

PR #2152, opened 2017-02-04
Uptime

scopatz:
Hopefully this addresses some of the issue brought up in #2148

### DIFF
--- a/.circle.yml
+++ b/.circle.yml
@@ -1,6 +1,7 @@
 machine:
     environment:
         PATH: /home/ubuntu/miniconda/envs/test_env/bin:/home/ubuntu/miniconda/bin:$PATH
+        XONSH_DEBUG: 1
     post:
         - pyenv global 3.4.4 3.5.2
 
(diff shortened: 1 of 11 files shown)


scopatz:
OK, I think this is good to go. The mac failures are not correct. If you open it up you'll see that it passes.

mitnk:
amazing! that's a lot of code for getting the boot time :) 

For Unix-like systems, could we just Popen('uptime') and parse output get the result?

scopatz:
> For Unix-like systems, could we just Popen('uptime') and parse output get the result?

We could, but subprocess is relatively slow compared to just reading from `/proc/stat`. Also if uptime is not installed for some reason, it would fail. I think that this is a better approach ultimately, even if it is a lot of work for one number :)

melund:
Yes. Pretty cool. I will restore the json files I delete and see if they are now cleaned up.

melund:
It fixes the GC for me. Cool.

scopatz:
Thanks @melund!

PR #2153, opened 2017-02-05
Argument completion

scopatz:
This adds the ability for Python function argument names to be completed. This is instead of having jedi completions. See #2127 for more discussion.

### DIFF
--- a/news/argcomp.rst
+++ b/news/argcomp.rst
@@ -1,0 +1,14 @@
+**Added:**
+
+* The Python tab completer will now complete the argument names of functions
+  and other callables.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/tests/test_python_completers.py
+++ b/tests/test_python_completers.py
@@ -1,0 +1,51 @@
+import builtins
+
+import pytest
+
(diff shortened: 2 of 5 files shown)


scopatz:
Again, I think the travis failures are not correct

scopatz:
Updated this from master

scopatz:
pinging this one!

melund:
What am I doing wrong: 
image

scopatz:
You should see `x=` and `y=` in that completions list.  If you start typing an argument name and then press tab, you should only get the arguments with the prefix you have already typed

melund:
So it does work, but it seem to mix in other completions as well. 
image

melund:
I guess that is intended right. The big completion list after an opening bracket is not that useful though...

astronouth7303:
Do completions have any concept of priority?

scopatz:
> I guess that is intended right. The big completion list after an opening bracket is not that useful though...

I agree, but it is correct since you pass in any value to a function.

> Do completions have any concept of priority?

Not yet. That is a separate issue because it would involve some change to the model, most likely.

melund:
Ok. This all looks good to me. Nice indeed.

scopatz:
Thanks @melund!

PR #2157, opened 2017-02-05
Adding some tests back

mitnk:
Adding some tests back, which was removed in #2109 (which, BTW, is a good one!)

Though it was the same test code for json and sqlite, but we need duplicate it in `test_history_sqlite.py` and `test_history.py` since they use different `hist` feature one for json and one for sqlite. Since there's two built-in backend, we have to cover them both :)

cc @laerus

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,4 +1,4 @@
-[pytest]
+[tool:pytest]
 flake8-max-line-length = 180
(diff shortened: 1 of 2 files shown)


scopatz:
LGTM. Thanks @mitnk!

PR #2161, opened 2017-02-07
bash_completions is now manually lazy...

scopatz:
....and does not override other parts of the completions subpackage on amalgamation.

I think the current state of master will block a release without this fix. Hopefully it isn't too bad to review.

### DIFF
--- a/xonsh/completers/bash_completion.py
+++ b/xonsh/completers/bash_completion.py
@@ -56,7 +56,7 @@ def _bash_completion_paths_default():
     """A possibly empty tuple with default paths to Bash completions known for
     the current platform.
     """
-    platfom_sys = platform.system()
+    platform_sys = platform.system()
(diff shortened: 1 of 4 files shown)


xonsh/completers/bash_completion.py
```diff
@@ -37,66 +35,88 @@ def windows_bash_command():
                               stdout=subprocess.PIPE,
                               stderr=subprocess.PIPE)
     except (FileNotFoundError, subprocess.CalledProcessError):
```
scopatz: Ahh good catch! I have updated this to include the new version.

melund:
@scopatz  Awesome. Now I think it is good to go right. Are we still ignoring Travis?

scopatz:
Yeah, I think it is good to go. Travis *seems* to have been working in the past 24 hours. But it is probably OK to ignore until we are sure.

PR #2162, opened 2017-02-07
Xonsh timings

melund:
This adds code to show timings for xonsh based on when events occur. 

To use run xonsh with:

```
PS ~ > xonsh --timings
Debug level: 0
|------------------|-----------|-----------|
|Event name        | Time (s)  | Delta (s) |
|------------------|-----------|-----------|
|startup           |   0.000   |   0.000   |
|pre_execer_init   |   0.266   |   0.266   |
|post_execer_init  |   0.430   |   0.163   |
|on_pre_rc         |   0.430   |   0.000   |
|on_post_rc        |   0.554   |   0.125   |
|on_ptk_create     |   0.720   |   0.166   |
|on_post_init      |   0.721   |   0.001   |
|on_pre_cmdloop    |   0.721   |   0.001   |
|on_pre_prompt     |   1.237   |   0.516   |
|------------------|-----------|-----------|
snail@sea ~\Documents\xonsh startuptime $
```

It is easy to add more custom markers by firing the `on_timingprobe` event anywhere in the code:

```python 
events.on_timingprobe.fire('pre marker name')
some_slow_code()
events.on_timingprobe.fire('post marker name')
```


melund:
Right now I add the events manually: https://github.com/xonsh/xonsh/pull/2162/files#diff-9c2723247eace91c7239099592206185R246

It could be done automatically if there were an event for when events are registered. (Meta events).

xonsh/timings.py
```diff
@@ -223,3 +225,87 @@ def timeit_alias(args, stdin=None):
...
     return
+
+# Can be used to insert custom timings into the code
+# events.on_timingprobe.fire(name='prope_test')
+# anywehere in the code base
+#@events.on_time_probe
+#def _time_prope_timing(name, **kw):
+#    if hasattr(builtins, '__xonsh_starttime__'):
+#        builtins.__xonsh_starttime__[name] = clock()
+##
```
scopatz: Please remove commented out code.

astronouth7303:
@melund That crossed my mind. I also wouldn't be against doing some kind of special hook in `LoadEvent`.

xonsh/__init__.py
```diff
@@ -1,5 +1,11 @@
 __version__ = '0.5.3'
 
+import sys as _sys
+if '--timings' in _sys.argv:
+    from xonsh.timings import setup_timings
+    setup_timings()
+del _sys
```
scopatz: I think this should be removed because it breaks amalgamation (`xonsh.timings` is imported and then overwritten later). The `setup_timings()` call can then be moved to main.py, right before the first event is fired.  

Also, given the way that out clock works with `resource.getrusage()`, it is actually better to assume that the first time is 0.0. That will capture the Python interpreter startup time. This version doesn't.

xonsh/timings.py
```diff
@@ -223,3 +225,87 @@ def timeit_alias(args, stdin=None):
...
+def setup_timings():
+    setattr(builtins, '__xonsh_timings__', {'startup': clock()})
```
scopatz: This initial call to clock() is not needed, since it should be assumed to be 0.0 to capture the interpreter startup time.

scopatz:

xonsh/__init__.py
melund: Ahh. Yes. I was afraid it was so. 

Is the clock thing also true for Windows. Don't worry I will check my self ...

scopatz: The docs seem to indicate that it works on windows, but I didn't boot into windows to find out :) https://docs.python.org/3/library/resource.html#resource.getrusage

melund: So, on Windows clock() reports the time since it was first called. I have adapted the code to only call it on windows and assume 0 on other platforms.

melund: `resource` is unix only package.

xonsh/timings.py
melund: That has been fixed.

xonsh/timings.py
melund: done

melund:
@astronouth7303 
> That crossed my mind. I also wouldn't be against doing some kind of special hook in LoadEvent
I am not sure what you have in mind with a special hook, but a something like `on_event_register` could be useful in this particular case. It should, of course, provide the name of the new event.

astronouth7303:
Well, there's `on_new_event` and `on_event_transmogrify`. I assume you only care about load events and not other kinds.

scopatz:
@melund, this is so awesome! Thanks so much for putting this in.  I think this is a game changer for debugging, like `xonfig` was.

Issue #2165, opened 2017-02-07
$(...) is not working with gpg


flavius:
The following script used to work with xonsh 0.5.2, I've just downgraded and tested it again, it works. But with the beta release of 0.5.3, the command `$(gpg2 --fingerprint)` returns an empty string.

```
import tempfile
import xonsh
import os
import sys
import getpass

source ~/.config/xonsh/autoload/00_storedconfig.xsh

realname_config=StoredConfig("realname", "Real Name")
email_config=StoredConfig("email", "E-mail Address")

$GPGHOME = '/home/homediruser/.gnupg'
GPGAGENTCONF = $GPGHOME + '/gpg-agent.conf'

gpgdir=[p for p in p`$GPGHOME` if p.is_dir()]
bitlength = 4096

configtpl = (
            "Key-Type: 1\n"
            "Key-Length: " + str(bitlength) + "\n"
            "Subkey-Type: 1\n"
            "Subkey-Length: " + str(bitlength) + "\n"
            "Subkey-Usage: encrypt\n"
            "Name-Real: {realname}\n"
            "Name-Email: {email}\n"
            "Name-Comment: comment\n"
            "Handle: handle\n"
            "Expire-Date: 50w\n"
            "Passphrase: {passphrase}"
            )
revocationtpl = (
            "y\n"
            "0\n"
            "{comment}\n"
            "\n"
            "y\n"
            )

if len(gpgdir) == 0:
    pkill gpg-agent
    print("===== You are about to set up a \"perfect GPG keypair\" for security. Please follow the instructions.")
    realname=realname_config.readvalue_plain()
    email=email_config.readvalue_plain()
    passphrase=False
    passphrase_confirm=True
    while passphrase != passphrase_confirm:
        passphrase=getpass.getpass("Master key passphrase (NEVER FORGET THIS): ")
        passphrase_confirm=getpass.getpass("Repeat passphrase: ")

    mkdir $GPGHOME
    chmod 0700 $GPGHOME

    fp = tempfile.NamedTemporaryFile()
    configdata=configtpl.format(realname=realname, email=email, passphrase=passphrase)
    fp.write(configdata.encode())
    fp.flush()
    print("===== Generating master keypair, this may take a few minutes...")
    gpg2 --batch --pinentry-mode loopback --generate-key @(fp.name)
    fp.close()

    fpr=$(gpg2 --fingerprint).split("\n")[3].strip()
    print("===== Generating signing key")
[...]

PR #2167, opened 2017-02-08
Allow unthreadable subprocs to be captured.

scopatz:
This should address #2165. I would appreciate it if folks could bang on this for a bit. I don't think it breaks anything else, but I have only done preliminary tests. I am not sure if it will work when the output is larger than the buffer size.


scopatz:
Also, I should mention that the basic idea here is that just because something is unthreadable (and can't be teed to Python), doesn't mean that we can't capture the output if we are willing to wait. Since this is exactly what `$(cmd)` is supposed to do (ie wait and capture all output), the threadability shouldn't matter. Mea culpa.

mitnk:
@scopatz 
OT here - I'm currently struggling with job control stuff. Could you provide some (short is ok) information why `threadable/unthreadable/ThreadPopen` was introduced (instead use Popen directly)? Thanks :)

scopatz:
@mitnk - Basically, it is because Popen will block the thread it is on if you read from the pipes that you send to it. In certain situations, such as `![]` and `!()`,  we want to be able to tee stdout and stderr into an in-memory buffer and print it out and remain fully interactive. You basically can't get all of that on a single thread normally.  You either need to use multiple threads (what we do) or use an event loop, like asyncio provides.

scopatz:
Ok the circle tests seems to pass now, but don't merge yet because I have noticed some bugs locally.

scopatz:
Alright I took care of the local bug. If this passes CI, it is ready for review.

melund:
The test fails. It seems it is because 'echo' on windows return a '\r\n' to the when it is captured. 

```
   def test_import():
        import sample
>       assert ('hello mom jawaka\n' == sample.x)
E       assert 'hello mom jawaka\n' == 'hello mom jawaka\r\n'
E           hello mom jawaka
```

It is weird because on don't see the same behaviour on my computer. 
```
snail@sea ~\Documents\xonsh capunthread $ x = $(echo hello)
snail@sea ~\Documents\xonsh capunthread $ x
'hello\n'
```

melund:
Ahh.. Echo is aliased to a print statement in the test files: 
https://github.com/xonsh/xonsh/blob/master/tests/xpack/sample.xsh#L2
```
aliases['echo'] = lambda args, stdin=None: print(' '.join(args))
```

melund:
There seem to be a universal newline issue here. 

If I capture a normal echo command it get a universal newline:
snail@sea ~\Documents\xonsh capunthread $ x = $(echo hello)
snail@sea ~\Documents\xonsh capunthread $ x
'hello\n'
```
But if I capture the ouptut of a aliased print statement then I don't get universal newlines:
```
snail@sea ~\Documents\xonsh capunthread $ aliases['echo'] = lambda args, stdin=
......................................... None: print(' '.join(args))
snail@sea ~\Documents\xonsh capunthread $ x = $(echo hello)
snail@sea ~\Documents\xonsh capunthread $ x
'hello\r\n'
```

scopatz:
Added `universal_newlines` fix. Let's hope it works!

scopatz:
Ok, that seemed to fix the windows issues.

melund:
Great.

scopatz:
Thanks @melund!

PR #2171, opened 2017-02-08
added a missing param

mitnk:
Just found a missing parameter. new news item is needed.


gforsyth:
Thanks, @mitnk!  Nice catch.

PR #2172, opened 2017-02-08
fix tcsetattr: Interrupted system call

### LINKED ISSUE #2170: script that calls ssh gets interrupted
```
laerus@xonsh$ cat autossh 
ssh laerus@192.168.1.10
laerus@xonsh$ ./autossh 
^Z <--- didn't press anything here, this comes from xonsh or ssh
```

tried again with a non-lan remote
```
laerus@xonsh$ ./autossh
^Z
tcsetattr: Interrupted system call
```

mitnk:
try fixes #2170

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -535,6 +535,9 @@ class SubprocSpec:
         """Prepares the 'preexec_fn' keyword argument"""
         if not (ON_POSIX and self.cls is subprocess.Popen):
             return
+        env = builtins.__xonsh_env__
+        if not env['XONSH_INTERACTIVE']:
+            return
         if pipeline_group is None:
             xonsh_preexec_fn = no_pg_xonsh_preexec_fn
         else:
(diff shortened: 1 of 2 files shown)


xonsh/built_ins.py
```diff
@@ -535,6 +535,9 @@ def prep_preexec_fn(self, kwargs, pipeline_group=None):
         """Prepares the 'preexec_fn' keyword argument"""
         if not (ON_POSIX and self.cls is subprocess.Popen):
             return
+        env = builtins.__xonsh_env__
+        if not env['XONSH_INTERACTIVE']:
```
scopatz: This might need to be env.get('XONSH_INTERACTIVE')

astronouth7303: `'XONSH_INTERACTIVE' in env`?

mitnk: I think this is safe. this env is always set during the start up phase, and here is a run-sub-proc phase. But I could update a bit :)

mitnk: Ok, I see the points. CI was failing because of this. In tests we may only mock part of the ENVs.

scopatz:
This seems like a reasonable change. thanks @mitnk!

Issue #2089, opened 2017-01-13
bpython integration.


AkhterAli:
Could there be a bpython like text prediction on xonsh?  This would be awesome, annoying too at times if not imported properly, but that would make the use case of xonsh much more desirable.

astronouth7303:
Could you clarify?

AkhterAli:
<img width="2550" alt="screen shot 2017-01-13 at 10 48 55 pm" src="https://cloud.githubusercontent.com/assets/7011993/21952137/8cf7b828-d9e2-11e6-947d-3ca8b3daf7a5.png">

I've taken a picture of it as it's not easy to explain.  If you see my cursor, I've only typed a portion of my client variable creation, which calls a method on boto3.  As I type the class, I get all of the available methods that can be used with this class and tab through them to select what I want.

I was wondering if such a feature could be possible on xonsh.

astronouth7303:
Ah, python-mode completions.

scopatz:
I am all for this feature! We may have another issue related to this, namely #344.  Pull requests most welcome!

scopatz:
Please see #2175 for an implementation of this feature. Note that it is restricted to the prompt-tollkit shell for now. If someone wants to implement this in readline they are more than welcome to :)

PR #2175, opened 2017-02-09
Complete while typing

scopatz:
This implements the desired feature in #2089. Thanks to @jonathanslenders for helping me debug!


melund:
I am not the right person to test this. On windows, this gives huge performances issue. 1-second delay on every keypress is hard to work with. It may be unavoidable on Windows, but we better mention it in the docs.

melund:
It is weird. I wonder if the bash completer gets called every time... That could explain why it is so slow.

melund:
No wait.

melund:
I think I accidently activated `$UPDATE_PROMPT_ON_KEYPRESS` instead. Dohh. That explains it.

melund:
This works nicely.  I think will have this on by default in my setup.

melund:
I think this is good to merge :) Awesome feature.

scopatz:
Excellent! Thanks.

Issue #2071, opened 2017-01-08
jobs and backgrounding broken


laerus:
```
$ vim 
( hitting Ctrl-Z)

$                                     
$                        
$ fg
[1]+ running: vim (1966)
$ bg
[1]+ running: vim (1966)
$ fg
$ ( hitting Ctrl-D)
xonsh: there is an unfinished job
-----
[1]+ running: vim (1966)
-----
Type "exit" or press "ctrl-d" again to force quit.
```

same happens with `top`

scopatz:
Thanks for reporting @laerus! I am not sure what is going on here.

scopatz:
Though I am seeing the same issues.

laerus:
been trying to troubleshoot this from time to time with no luck,
this was an issue before the subproc refactor

adqm:
I'm not sure when this became an issue.  In case it helps, I'm running a version forked from 0.3.4 and I don't have this issue.  That said, probably a lot has changed with job control since then.

scopatz:
Thanks @adqm. If someone feels like bisecting, that could be helpful to know when it broke :)

gforsyth:
I think there are two separate issues here (that I've noticed). 
One is that backgrounding doesn't seem to work (the big issue). The other is that while the SIGSTOP is sent by Ctrl-Z, the prompt doesn't always return control to the user. 

Bisecting and running `vim` and then hitting `Ctrl-Z` points to 008085e as the culprit for control not being returned to the user. And commenting out https://github.com/xonsh/xonsh/blob/master/xonsh/proc.py#L1739 does fix that issue (although it may introduce other problems). 

Still working on tracking down exactly where bringing vim back to the foreground broke...

scopatz:
Thanks for hunting this down @gforsyth! I am formulating an idea on how to fix this.  I think the proc.wait() might only be needed in we are running an alias (Proxy or ProxyThread). Or maybe this needs to be `proc.communicate()` if we have a Popen class.

scopatz:
Ok, communicate didn't work.

scopatz:
Alright, so I think part of the issue might be whether we want to call wait() and end() depend on what is returned from `wait_for_active_job()`. Unfortunately right now, Crtl-z and a process exiting normally will both return None. So, I think that job control needs to be updated to reflect the kind of signal it sees before we can fix the iterraw() implementation.

scopatz:
[...]

PR #2123, opened 2017-01-29
Sanity Restoration on Background

scopatz:
Hopefully this will further address #2071. @laerus, would you mind taking a look, please?

PR #2177, opened 2017-02-10
Update controlling terminal and process group

mitnk:
This PR try fixing  #2064, which is an Mac issue, describes as the following:
```
$ ls | less
Applications
Desktop
Documents
:^Z  # <-- printed by xonsh, not typed by user
$  # <-- xonsh now takes back the terminal
```
The processes relationship would be:
```
 PID  PPID  PGID TPGID STAT COMMAND
1856  1821  1856  1856 S+   xonsh
1865  1856  1864  1856 T    less
```
The proc `ls` (PID 1864) terminated normally before xonsh run `CommandPipeline.end()`, which calls `wait_for_active_job()`, where the new process group (PGID 1864) could get a chance to be set as foreground process group (by `tcsetpgrp()`).

Since `less` (PID 1865, GID 1864) program needs receiving `stdin` and `stdout` to work, but its process group doesn't control the terminal. Thus the `less` process is set to `STOPPED` by OS.

And now when running into `wait_for_active_job()`, we cannot find any active task and the code decides to give terminal back to the shell (xonsh proc).

------

In this PR, we fix the issue this way - 1) always create process group as possible (as long as new subprocess(es) is created), try be more POSIX as possible. 2) give controlling terminal as soon as possible to the new foreground process group, by giving it when it is just created. That way when `less` process created, for the above example, it will be set into the new process group as `ls`, which have already has the controlling terminal). and 3) moved the job "returning control terminal to xonsh proc" from `wait_for_active_job()` to the `CommandPipeline.end()`, take advantage of its state-ful compare to the stateless `wait_for_active_job()`.

@scopatz I want to collect your thoughts on those changes or this direction.

I've tested various sub process commands, it seems the new code have worked properly.

--------

This PR is still working in progress because one thing I notice the code breaks the following `Ctrl-Z` functionality:
```
$ ls | less
^Z   # <--  user hit Ctrl-Z to stop it and put it into background
$ fg  # <-- user want bring `ls | less` back and continue it to running, which failed
```
So the WIP is mostly for the `Ctrl-Z` and `fg` stuff.


xonsh/jobs.py
```diff
@@ -170,22 +180,14 @@ def wait_for_active_job(last_task=None, backgrounded=False):
         active_task = get_next_task()
         # Return when there are no foreground active task
         if active_task is None:
-            _give_terminal_to(_shell_pgrp)  # give terminal back to the shell
-            if backgrounded and hasattr(builtins, '__xonsh_shell__'):
-                # restoring sanity could probably be called whenever we return
-                # control to the shell. But it only seems to matter after a
-                # ^Z event. This *has* to be called after we give the terminal
-                # back to the shell.
-                builtins.__xonsh_shell__.shell.restore_tty_sanity()
```
mitnk: These block of code is one of the WIP, needing to put back to somewhere.

xonsh/jobs.py
```diff
@@ -354,31 +356,31 @@ def fg(args, stdin=None):
...
+    tasks.remove(tid)
+    tasks.appendleft(tid)
 
-    job = get_task(act)
+    job = get_task(tid)
     job['bg'] = False
     job['status'] = "running"
-    print_one_job(act)
+    print_one_job(tid)
     wait_for_active_job()
```
mitnk: The alias `def fg()` may need to re-implemented, because it's running in a thread and changing controlling terminal only can be done in a main thread. But now sure, still digging.

mitnk:
I think I've figured out how to update `fg`, but now kind of blocked by issue #2179

scopatz:
Hi @mitnk - in general, I totally agree with you about the changes to the statefulness of job control. Job control is sort of in the wrong place at this point and probably needs a rewrite

xonsh/built_ins.py
```diff
@@ -793,34 +806,37 @@ def run_subproc(cmds, captured=False):
...
         # set title here to get currently executing command
         pause_call_resume(proc, builtins.__xonsh_shell__.settitle)
     # create command or return if backgrounding.
-    if spec.background:
+    if background:
         return
     if captured == 'hiddenobject':
         command = HiddenCommandPipeline(specs, procs, starttime=starttime,
-                                        captured=captured)
+                                        captured=captured, term_pgid=term_pgid)
```
scopatz: I don't think this additional interface is needed. This is already stored on the specs or procs, right? So the CommandPipline can just pull it off from those, right?

xonsh/jobs.py
```diff
@@ -153,12 +159,16 @@ def _give_terminal_to(pgid):
...
                                  ctypes.byref(omask), None)
     else:
-        def _give_terminal_to(pgid):
-            st = _shell_tty()
-            if st is not None and os.isatty(st):
-                oldmask = signal.pthread_sigmask(signal.SIG_BLOCK,
-                                                 _block_when_giving)
-                os.tcsetpgrp(st, pgid)
+        def give_terminal_to(pgid):
+            oldmask = signal.pthread_sigmask(signal.SIG_BLOCK, _block_when_giving)
```
scopatz: PEP8: This line is now over 80 characters

xonsh/jobs.py
```diff
@@ -153,12 +159,16 @@ def _give_terminal_to(pgid):
...
+            try:
+                os.tcsetpgrp(sys.stderr.fileno(), pgid)
```
scopatz: I think it is best to stick with `_shell_tty()` here. Is there a reason that didn't work?

xonsh/jobs.py
```diff
@@ -153,12 +159,16 @@ def _give_terminal_to(pgid):
...
+                return True
+            except Exception as e:
+                print('tcsetpgrp error {}: {}'.format(e.__class__.__name__, e),
+                      file=sys.stderr)
```
scopatz: You probably want to use `xonsh.tools.print_exception()` here, but I still think that `_shell_tty()` should be used if possible.

xonsh/jobs.py
scopatz: You should be able to decorate it with `xonsh.tools.unthreadable`

xonsh/proc.py
```diff
@@ -1912,12 +1914,10 @@ def _decode_uninew(self, b, universal_newlines=None):
     # Ending methods
     #
 
-    def end(self, tee_output=True):
+    def end_internal(self, tee_output):
```
scopatz: Since this is a private method it is probably best to call it `_end()`

scopatz:
Thanks for the awesome effort here, this is a particularly tricky part of the code base

xonsh/built_ins.py
mitnk: It could not be pulled from its procs or specs. For examples: `ls | cat` and `o = $(ls | cat)` has the same proc and specs but one has `term_pgid` to the pgid of `ls`, one is `None`. Unless we repeat the `if` tests inside `CommandPipeline`:
```
        if proc.pid and pipeline_group is None and not spec.is_proxy and \
                captured != 'object':
            pipeline_group = proc.pid
            if update_fg_process_group(pipeline_group, background):
                term_pgid = pipeline_group

```

scopatz: Ahh yeah, there is no good place for this.  But I think it makes more sense (to me) conceptually if that block is moved to `CommandPipeline.__init__()`.  CommandPipeline represents a command actually being run, and we probably don't want to hand off to the terminal until we are running the command. Does that make sense or am I crazy?

mitnk: That makes sense, but in this direction, we probably need to move `for spec in specs: spec.run()` into `CommandPipeline.__init__()`, and also keep `give terminal to new pgrd` done after the first `proc` run (which to fix the original issue).

scopatz: Yeah, I think that would be the correct move.

scopatz: That loop probably should have been in `CommandPipeline.__init__()` from the start :)

mitnk: Let's do it now, later better than never :)

scopatz: Very true :)

xonsh/jobs.py
```diff
@@ -123,21 +130,21 @@ def _set_pgrp(info):
     def _shell_tty():
```
mitnk: @scopatz On `_shell_tty()` thing, I'll do self review on it again later. but basically it was not needed since we starting record what happened to the owner of terminal, and an "am I own the terminal" is not needed anymore.

xonsh/built_ins.py
```diff
@@ -962,7 +955,7 @@ def convert_macro_arg(raw_arg, kind, glbs, locs, *, name='<arg>',
             ctx |= set(locs.keys())
         mode = mode or 'eval'
         arg = execer.parse(raw_arg, ctx, mode=mode, filename=filename)
-    elif kind is types.CodeType or kind is compile:
+    elif isinstance(kind, types.CodeType) or kind is compile:
```
mitnk: This is a `flake8` complain fix - hope it's correct.

xonsh/jobs.py
mitnk: @scopatz what was the case (any command example?) for this block of code to exist? I want to put them back into `CommandPipeline`, but haven't find a place yet.

xonsh/built_ins.py
mitnk: nope.. reverted.

mitnk:
The [following code](https://github.com/xonsh/xonsh/blob/master/xonsh/jobs.py#L174-L179) which was removed by this PR
```
            if backgrounded and hasattr(builtins, '__xonsh_shell__'):
                # restoring sanity could probably be called whenever we return
                # control to the shell. But it only seems to matter after a
                # ^Z event. This *has* to be called after we give the terminal
                # back to the shell.
                builtins.__xonsh_shell__.shell.restore_tty_sanity()
```
was introduced by #2123 and was try fix #2071, the two cases were not working:
```
$ vim
^Z  # <-- by user, didn't work
$ fg
```
and
```
$ python3
^Z  # <-- by user, didn't work
$fg
```
The above two case is current working in this PR of the last commit.
So I think these `restore_tty_sanity()` code may not be needed anymore.

---------

O[ne message](https://github.com/xonsh/xonsh/issues/2071#issuecomment-275934727) in #2071 mentioned the following command do not work:
```
$ bash -c 'while true; do echo hey; sleep 1; done'
^Z  # <-- Do not work
``` 
Well, this is still not working on this PR, and not work on master. The reason I think you've mentioned is because `PopenThread` process do not run into `wait_for_active_job()` and `^Z` is not getting listened.

I think we can fix this issue in further PRs.

xonsh/jobs.py
mitnk: see [this comment](https://github.com/xonsh/xonsh/pull/2177#issuecomment-279401696).

mitnk:
In last commit, I removed the following code and others just for consistence :
```
# check for shell tty
-    @functools.lru_cache(1)
-    def _shell_tty():
-        try:
-            _st = sys.stderr.fileno()
-            if os.tcgetpgrp(_st) != os.getpgid(0):
-                # we don't own it
-                _st = None
-        except OSError:
-            _st = None
-        return _st
```
We can add them back if we found it's necessary.

My idea here the `give_terminal_to` should work properly unless we had a bug in somewhere else, say `CommandPipeline` (e.g. should give terminal but didn't, or vice versa). Or do I miss anything here?

mitnk:
So this PR is leaving WIP status. Ready for code review. What's left from my side are some more tests and other platforms (Windows).

scopatz:
Hi @mitnk, thanks so much for this amazing effort!  Everything seems to work as expected *except* backgrounding Python on the readline shell. When I ^Z a running Python interpreter, I can longer see the text I type. I think we need to put the call to `restore_pty_sanity()` back in.  This is why I added it in the first place. 

You'll notice that PTK correctly resets the TTY at the start of a prompt. This is why PTK's `restore_tty_sanity()` is a no-op. However, with readline, we have to do this manually.  

If you agree, once you add this back, I think it is ready to go.

mitnk:
Ah.. yes, all my tests are based on PTK. Readline is what I missed! Ok, added tty sanity back. But I found weird thing about this: I can ^Z on `python` or `python2` or `python3` normally, but `^Z` would stuck with `python3.6`, `python2.7`, is there any string matching thing in our code?

mitnk:
Another issue I just found is the following pipeline will stuck (I'll fix it soon):
```
$ ls | command-not-exist
```

scopatz:
Ok! Just let me know when it is ready for another look.

scopatz:
It looks like there are now conflicts with master.  But it should be fixed by a reinstall/reamalgamation. Also we should check that `sudo` works normally and we don't have to type in the password after the first time

mitnk:
Conflicts solved. I think it's ready for another look thanks. checked `sudo`, seems works properly.

xonsh/__init__.py
```diff
@@ -26,8 +26,6 @@
         _sys.modules['xonsh.platform'] = __amalgam__
         pretty = __amalgam__
         _sys.modules['xonsh.pretty'] = __amalgam__
-        jobs = __amalgam__
```
mitnk: @scopatz is the amalgam serialisation orders not reliable, or am I doing something wrong?

scopatz: They are totally reliable and deterministic. The issue is that changes to the code (imports specifically) can change the dependency graph of the different modules. This is then reflected in the ordering here.

scopatz:
This is great @mitnk! Just waiting for CI.

mitnk:
Great! Thanks for your help @scopatz!

Issue #2178, opened 2017-02-10
Provide __file__


flavius:
There should be a way to determine the location of the current script, no matter how it's executed:

1. as a stand-alone script with a xonsh shabang header
2. as a parameter to xonsh on the command line
3. via `source`

There is `$XONSH_SOURCE` (related #1169), but that one is not defined in case 3.

scopatz:
Hi @flavius, thanks for reporting. It is unclear to me which one of these cases is failing. Can you please give a more detailed example. Thanks!

flavius:
@scopatz Complete transcript of my terminal session, showcasing the source and the three cases.

```
⋊> /tmp cat -n test.xsh                                                                                                                                                                        22:12:51
     1	#!/usr/bin/xonsh
     2	
     3	print(__file__)
⋊> /tmp /tmp/test.xsh                                                                                                                                                                          22:12:57
Traceback (most recent call last):
  File "/usr/bin/xonsh", line 3, in <module>
    main()
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 18676, in main
    _failback_to_other_shells(args, err)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 18640, in _failback_to_other_shells
    raise err
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 18674, in main
    return main_xonsh(args)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 18711, in main_xonsh
    loc=None, mode='exec')
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 2165, in run_script_with_cache
    run_compiled_code(ccode, glb, loc, mode)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 2085, in run_compiled_code
    func(code, glb, loc)
  File "./test.xsh", line 3, in <module>
    print(__file__)
NameError: name '__file__' is not defined
⋊> /tmp xonsh /tmp/test.xsh                                                                                                                                                                    22:13:13
Traceback (most recent call last):
  File "/usr/bin/xonsh", line 3, in <module>
    main()
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 18676, in main
    _failback_to_other_shells(args, err)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 18640, in _failback_to_other_shells
[...]

PR #2181, opened 2017-02-11
Added __file__ and __name__ to xonsh scripts

scopatz:
This should address #2178

### DIFF
--- a/tests/tools.py
+++ b/tests/tools.py
@@ -39,6 +39,11 @@ skip_if_on_unix = pytest.mark.skipif(not ON_WINDOWS, reason='Windows stuff')
 
 skip_if_on_darwin = pytest.mark.skipif(ON_DARWIN, reason='not Mac friendly')
 
+skip_if_on_travis = pytest.mark.skipif('TRAVIS' is os.environ and 'CI' in os.environ,
+                                       reason='not Travis CI friendly')
+
+
+
 
 def sp(cmd):
     return subprocess.check_output(cmd, universal_newlines=True)
--- a/tests/tools.py
+++ b/tests/tools.py
@@ -43,8 +43,6 @@ skip_if_on_darwin = pytest.mark.skipif(ON_DARWIN, reason='not Mac friendly')
 skip_if_on_travis = pytest.mark.skipif(ON_TRAVIS, reason='not Travis CI friendly')
 
 
-
-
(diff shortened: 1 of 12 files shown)


scopatz:
Ok, the tests finally pass CI here

tests/tools.py
```diff
@@ -25,6 +25,7 @@
 ON_WINDOWS = (platform.system() == 'Windows')
 ON_CONDA = True in [conda in pytest.__file__ for conda
                     in ['anaconda', 'miniconda']]
+ON_TRAVIS = 'TRAVIS' is os.environ and 'CI' in os.environ
```
laerus: `'TRAVIS' in* os.environ` ?

xonsh/tools.py
```diff
@@ -805,6 +805,21 @@ def swap(namespace, name, value, default=NotImplemented):
     else:
         setattr(namespace, name, old)
 
+
+@contextlib.contextmanager
+def swap_values(d, updates, default=NotImplemented):
+    """Updates a dictionary (or other mapping) with values from another mapping,
+    and then restores the original mapping when the context is exited.
+    """
+    old = {k: d.get(k, NotImplemented) for k in updates}
```
laerus: `NotImplemented` can be the value of an existing key in the old dict,
isn't an instance of `object` the suggested sentinel value?

tests/tools.py
scopatz: Oh shoot! Good catch @laerus!

xonsh/tools.py
scopatz: That is a good idea.  Where did you see that suggested, though? Just out of curiosity...

scopatz:
Thanks for the review @laerus, I think that this is ready for another look!

xonsh/tools.py
laerus: picked that up from a python core dev talk (iirc Raymond Hettinger on idiomatic python)
and kind of took it for granted, no idea if there is anything on the official docs.

laerus:
thanks @scopatz !

xonsh/tools.py
scopatz: Nice, it is definitely the way to do it.

scopatz:
Thanks!

PR #2182, opened 2017-02-12
[RFC] moved unthreadable etc decorators to tools

mitnk:
Moved decorators `unthreadable`, `foreground`, `uncapturable` from `xonsh.proc` to `xonsh.tools`. Because,

- Can reduce potential circle imports, since proc.py have more deps. (major reason)
- These decorators have no deps, and all one/two line of code, suitable in tools.py
- `xonsh.tools` is more suitable than `xonsh.proc` for using by xonsh itself or xonsh users

ideas?

cc @scopatz

### DIFF
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -31,6 +31,8 @@ from xonsh.tools import (redirect_stdout, redirect_stderr, print_exception,
 from xonsh.lazyasd import lazyobject, LazyObject
 from xonsh.jobs import wait_for_active_job
 from xonsh.lazyimps import fcntl, termios, _winapi, msvcrt, winutils
+# these decorators are imported for users back-compatible
+from xonsh.tools import unthreadable, foreground, uncapturable
 
 
 @lazyobject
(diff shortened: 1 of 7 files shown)


xonsh/proc.py
```diff
@@ -31,6 +31,8 @@
 from xonsh.lazyasd import lazyobject, LazyObject
 from xonsh.jobs import wait_for_active_job
 from xonsh.lazyimps import fcntl, termios, _winapi, msvcrt, winutils
+# these decorators are imported for users back-compatible
+from xonsh.tools import unthreadable, foreground, uncapturable
```
mitnk: not sure whether I need to add this.

xonsh/proc.py
```diff
@@ -31,6 +31,8 @@
 from xonsh.lazyasd import lazyobject, LazyObject
 from xonsh.jobs import wait_for_active_job
 from xonsh.lazyimps import fcntl, termios, _winapi, msvcrt, winutils
+# these decorators are imported for users back-compatible
+from xonsh.tools import unthreadable, foreground, uncapturable  # NOQA
```
mitnk: do we need to add this back compatibility?

scopatz: Yes we do. Thanks for doing that.  You can keep `foreground` in proc, though, since that was deprecated already.

scopatz:
This looks good to me, let me know if you think it is ready

mitnk:
@scopatz updated for `foreground`, ready for another look. no work left from my side in this PR.

scopatz:
Thanks!

Issue #830, opened 2016-05-10
SIGINT is not passed to subprocess when piped into an alias


azag0:
See the code below. I eventually killed the bash process from another terminal.

```
🐌 10:21 her@air ~ def _mycat(args, stdin):
.................    sys.stdout.write(stdin) 
.................    
🐌 10:22 her@air ~ aliases['mycat'] = _mycat
🐌 10:22 her@air ~ bash -c 'while true; do echo hey; sleep 1; done' | cat
hey
hey
hey
^C
🐌 10:22 her@air ~ bash -c 'while true; do echo hey; sleep 1; done' | mycat
^C^C^C^C^C^Chey
hey
hey
hey
hey
hey
hey
hey
🐌 10:22 her@air ~ 
```

adqm:
Yes, this is true.  Job control doesn't currently work on aliases.  As part of #742, which I will be coming back to soon, I had started to work some of this in.  That's still a work in progress.

adqm:
(more specifically, job control doesn't currently work on aliases specified as Python functions)

scopatz:
Potential fix in #2184

mitnk:
Have xonsh changed something on alias since then? I got following when try reproduce the issue:
```
$ bash -c 'while true; do echo hey; sleep 1; done' | mycat
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/proc.py", line 1330, in run
    r = self.f(self.args, sp_stdin, sp_stdout, sp_stderr, spec)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/proc.py", line 1159, in proxy_two
    return f(args, stdin)
  File "<xonsh-code>", line 2, in _mycat
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/proc.py", line 1044, in write
    r = h.write(s)
TypeError: write() argument must be str, not _io.TextIOWrapper
```

scopatz:
Yes, xonsh has changed.  My version of mycat now looks like:

```python
def _mycat(args, stdin):
    for line in stdin:
        sys.stdout.write(line) 
```

scopatz:
#2184 has been merged and so this is cloasable. Please feel free to reopen or open a new issue if problems persist.

azag0:
Amazing. Time to give xonsh another round of trial!

Issue #2133, opened 2017-01-31
Some jobs control issues


mitnk:
1. On Mac, run background & fg:

```
$ tail -f /tmp/goixy.log &
[1]+ running: tail -f /tmp/some.log & & (30817)

$ fg
[1]+ running: tail -f /tmp/some.log & (30817)
(**Here fg seems brought something into foreground, but stuck, let's ^C it**)

^C
$ Exception in thread Thread-16:
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/proc.py",n
    r = self.f(self.args, sp_stdin, sp_stdout, sp_stderr, spec)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/proc.py",o
    return f(args, stdin)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/jobs.py",g
    wait_for_active_job()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/jobs.py",b
    _, wcode = os.waitpid(obj.pid, os.WUNTRACED)
ChildProcessError: [Errno 10] No child processes

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py", line 916, in _r
    self.run()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/proc.py",n
    print_exception()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/tools.py"n
    traceback.print_exc()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/traceback.py", line 159, in pc
    print_exception(*sys.exc_info(), limit=limit, file=file, chain=chain)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/traceback.py", line 101, in pn
    print(line, file=file, end="")
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/base_shele
    self.std.write(std_s)
AttributeError: 'NoneType' object has no attribute 'write'
```

2. on Mac, run foreground & ^Z

`^Z` doesn't work.

[...]

PR #2184, opened 2017-02-12
Interruptable Proc Proxies

scopatz:
This should address #830, finally.


mitnk:
In this PR the `bash` process not killed fully. It is left as a zombie (on Mac at least) after `^C`:
```
USER   PID  PPID  PGID TPGID STAT COMMAND
hugo 25960 25959 25959 25959 S+   Python -u xonsh
hugo 25965 25960 25965 25959 Z    (bash)   # <-- sleep process was killed, but not bash
```

This issue #830 is not an issue in PR #2177 branch (without the patch of this PR). And `bash`, `sleep` processes was killed successfully. So I suggest PR #2177 to go master first then we check how it affects the jobs control issues including these: #2133 and this one?

scopatz:
> In this PR the bash process not killed fully. It is left as a zombie (on Mac at least) after ^C

Hmm that is very strange. On Linux the process it completely gone.  I am happy with waiting, but in any event it does seem to fix the problem on Linux.

mitnk:
So this PR also clean the pipes opened, so I think it is good to go. no zombie issue anymore with current master.

gforsyth:
Merge away, @mitnk !

mitnk:
@gforsyth I thought we may have different plans, any way, merged! :)

PR #2185, opened 2017-02-12
Xoreutils Reloaded

scopatz:
This is a cleanup and modernization of #742 with the goal of closing that one out.  Note that:

* #2184 takes care of the job control part of #742, this takes care of the xoreutils part.
* I have branched off of @adqm's original branch, so that he will get credit for the effort here.

As part of the cleanup, I made sure that the xoreutils were amalagamatable. As part of the modernization, I moved the loading piece of this to a `coreutils` xontrib, as per the discussions @astronouth7303 and I have been having recently with respect to modularity.  I think I actually wanted something like this in #742, but I don't think we had xontribs at that point, so it wasn't coming across.


xonsh/xoreutils/yes.py
```diff
@@ -0,0 +1,25 @@
...
+    while True:
+        print(*to_print, file=stdout)
+
+    return 0
+
+
+YES_HELP = """This version of yes was written in Python for the xonsh project: http://xon.sh
+Based on yes from GNU coreutils: http://www.gnu.org/software/coreutils/
+
+Usage: /usr/bin/yes [STRING]...
```
gforsyth: not in `/usr/bin`

xonsh/xoreutils/yes.py
```diff
@@ -0,0 +1,25 @@
...
+  or:  /usr/bin/yes OPTION
```
gforsyth: not in `/usr/bin`

xontrib/coreutils.py
```diff
@@ -0,0 +1,30 @@
+"""Additional core utilites that are implemened in xonsh. The current list
```
gforsyth: *implemented

gforsyth:
huh... ok, the `yes` issue is more complicated. If I run it amalgamated, it works and exits fine, but then something is hanging `xonsh` when I try to `exit`.

If I run in debug mode, then the `SIGINT` to `yes` does kill `xonsh` and then I get this error in the parent `xonsh`

```
[snip]
File "/home/gil/miniconda3/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 10357, in _signal_int
    signal.pthread_kill(threading.get_ident(), signal.SIGINT)
  File "/home/gil/miniconda3/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 10357, in _signal_int
    signal.pthread_kill(threading.get_ident(), signal.SIGINT)
  File "/home/gil/miniconda3/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 10354, in _signal_int
    if self.proc is not None and self.proc.poll() is not None:
  File "/home/gil/miniconda3/lib/python3.5/subprocess.py", line 1082, in poll
    return self._internal_poll()
  File "/home/gil/miniconda3/lib/python3.5/subprocess.py", line 1588, in _internal_poll
    if pid == self.pid:
RecursionError: maximum recursion depth exceeded in comparison
```

so it seems like something in `yes` isn't getting cleaned up quite right.

astronouth7303:
Wasn't really what I was thinking about, but having this as a bundled xontrib is pretty useful.

Should all of these live in `xonsh.xoreutils` or maybe in `xontribs.xoreutils`? The current `xontribs/xoreutils.py` would just be `xontribs/xoreutils/__init__.py` with no problems. Since these utilities are strictly unnecessary value-adds, should they live with all the stuff that's some level of necessary?

laerus:
maybe pwd/echo/which must be on their own package/module, these are the ones implemented as builtins in bash.

astronouth7303:
@laerus I think `which` is the only one that's required to be a shell builtin, though. (In order to pick up aliases). `pwd` and `echo` can be external programs with no loss of functionality.

scopatz:
Thanks for the review all. 

> yes works, but when you send a SIGINT it kills the entire xonsh instance.

This should be fixed by #2184, and is a broader problem than just the yes alias here

xonsh/xoreutils/echo.py
```diff
@@ -0,0 +1,27 @@
...
+        return
+    ender = opts['end']
+    args = map(str, args)
+    if opts['escapes']:
+        args = map(lambda x: x.encode().decode('unicode_escape'), args)
+    print(*args, end=ender, file=stdout)
+
+
+def _echo_parse_args(args):
+    out = {'escapes': False, 'end': '\n'}
```
scopatz: I don't think so. I didn't want to change the code too much for no reason, but I am not a huge fan of `arg_handerler()` personally.

scopatz:
Thanks @gforsyth! I have updated the PR based on your review.

gforsyth:
Thanks, @scopatz !

gforsyth:
and @adqm!

Issue #870, opened 2016-05-12
xonsh --config-path fails


scopatz:
And so does `xonsh -DXONSHCONFIG=/...`  If the default config file exists, these will just pick up that file instead.

scopatz:
See #2187

scopatz:
Closing due to merge of #2187

PR #2187, opened 2017-02-13
Xonsh startup loading refactor

scopatz:
This refactor of the loading system linearizes the setup of xonsh. The Shell class is now no longer responsible for creating an execer and loading the rc files. Additionally, it unifies the load path for xonsh-lang rc files and the JSON-based static configuration files. Lastly, this PR should address #870.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ tests/lexer_table.py
 tests/parser_table.py
 tests/lexer_test_table.py
 tests/parser_test_table.py
+tests/testfile
 build/
 dist/
 xonsh.egg-info/
--- a/news/lr.rst
+++ b/news/lr.rst
@@ -18,7 +18,11 @@
 * The ``--config-path`` command line option is now deprecated in favor of
   ``--rc``.
 
-**Removed:** None
+**Removed:**
+
+* ``xonsh.environ.DEFAULT_XONSHRC`` has been removed due to deprecation.
+  For this value, please check the environment instead, or call
+  ``xonsh.environ.default_xonshrc(env)``.
 
 **Fixed:** None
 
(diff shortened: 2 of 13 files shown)


scopatz:
Can someone more familiar with pytest and windows help me debug the remaining errors here, please? I don't understand why these pass on Linux and Mac, but not on Windows.

scopatz:
Feel free to push fixes directly to this branch!

melund:
I can reproduce this on my machine. But what causing it I can't say. I will try to dig deeper.

melund:
Maybe a clue. It seems that monkey patching the `load_builtins` doesn't work
https://github.com/xonsh/xonsh/blob/lr/tests/conftest.py#L28

At least when it is called in `excer.py`
https://github.com/xonsh/xonsh/blob/lr/xonsh/execer.py#L47

It calls the original function.

melund:
Apparently, monkeypatch can't work with all types of imports.
http://stackoverflow.com/questions/28403380/py-tests-monkeypatch-setattr-not-working-in-some-cases
Hack incoming.

scopatz:
Ughh Thanks for hunting this down @melund!  It is very disconcerting that this happens only on Windows.

melund:
This could also explain a lot of the weird issue we have had with pytest previously. Like some tests only working if they were in the correct order.

melund:
Now all test pass locally for me. They have never done that before :)

scopatz:
Yeah it could.  Let's hope they pass on CI :)

melund:
I am heading to bed. Feel free to take over. I think this good to merged btw..

scopatz:
Thanks again, @melund!

scopatz:
Ok great! It looks like all of the tests pass here.  Any objection to this going in?

PR #2188, opened 2017-02-13
provide core events for started and finished commands

flavius:
The purpose of the new events is to make it easy for users to detect executed commands, before and after the fact, as they are seen by the OS, after all the xonsh magic has been done.

It's just a working example. There are at least two rough edges I am aware of:

* the name of the two events is confusing; we need better names
* the "before" event just passes over the internal representation of the command - the purpose of these events is to offer a simple hooking point for basic needs, so I would prefer a quick and dirty access to the command(s) as lists of strings with the commands and their parameters

I have experience with formal languages, but my idiomatic python is rusty (pun intended) and I've never written a shell, so any advice is welcome.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -792,6 +792,7 @@ def run_subproc(cmds, captured=False):
     Lastly, the captured argument affects only the last real command.
     """
     specs = cmds_to_specs(cmds, captured=captured)
+    events.on_command_started.fire(cmd=specs)
     captured = specs[-1].captured
     procs = []
     proc = pipeline_group = None
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -33,6 +33,7 @@ from xonsh.jobs import wait_for_active_job
 from xonsh.lazyimps import fcntl, termios, _winapi, msvcrt, winutils
 # these decorators are imported for users back-compatible
 from xonsh.tools import unthreadable, uncapturable  # NOQA
+from xonsh.events import events
 
 # foreground has be deprecated
 foreground = unthreadable
@@ -535,6 +536,8 @@ class PopenThread(threading.Thread):
         self.proc = None  # has to be here for closure for handles
         self.old_int_handler = self.old_winch_handler = None
         self.old_tstp_handler = self.old_quit_handler = None
+        self.command_args = args
+        self.previous_signal = None
         if on_main_thread():
             self.old_int_handler = signal.signal(signal.SIGINT,
                                                  self._signal_int)
@@ -649,6 +652,7 @@ class PopenThread(threading.Thread):
         # kill the process if it is still alive. Happens when piping.
         if proc.poll() is None:
             proc.terminate()
+        events.on_command_finished.fire(cmd=self.command_args, sig=self.previous_signal)
 
     def _wait_and_getattr(self, name):
         """make sure the instance has a certain attr, and return it."""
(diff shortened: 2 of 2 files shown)


flavius:
@astronouth7303 I've noticed the pre/post convention, I just don't know how to name this pair of events. Once I have a name, I'll add it to news and to docs. I've been thinking hard, so any suggestion is welcome.

xonsh/built_ins.py
```diff
@@ -792,6 +792,7 @@ def run_subproc(cmds, captured=False):
     Lastly, the captured argument affects only the last real command.
     """
     specs = cmds_to_specs(cmds, captured=captured)
+    events.on_command_started.fire(cmd=specs)
```
scopatz: The keyword here should be `specs`, not `cmd`.  Also this isn't really after the command started. It is more like "after commands specified" or "before pipeline execution"

xonsh/proc.py
```diff
@@ -649,6 +652,7 @@ def run(self):
         # kill the process if it is still alive. Happens when piping.
         if proc.poll() is None:
             proc.terminate()
+        events.on_command_finished.fire(cmd=self.command_args, sig=self.previous_signal)
```
scopatz: I don't think this should be on PopenThread.  It makes the most sense for me for this to be in `CommandPipeline.end()`

scopatz:
Thanks for putting this in @flavius!

astronouth7303:
> There are only two hard things in Computer Science: cache invalidation and naming things. -- Phil Karlton

I would suggest `on_pre_process`, but that sounds a lot like the function of `on_transform_command`.

Maybe `on_pre_execution` or `on_pre_exec`?

@scopatz from a design and API usability standpoint, would it make more sense to do this the moment before or the moment after processes are executed?

flavius:
Well, the decision whether to trigger the pre event before or after opening the process depends on the semantics.

* a pre event before the execution was started is more like an intent. It has its own right to exist
* a pre event after the execution was started, is a (post) fact

Both of them could have their usages. Case 1 could prepare a special environment in some way for the command, case 2 could be blocked by the command to do that, in case of locked resources for instance.

For the sake of symmetry with the post event, I prefer Case 1, which can only fail when the command does not exist, which would be easy to check by the hook itself, if it needs to.

scopatz:
> @scopatz from a design and API usability standpoint, would it make more sense to do this the moment before or the moment after processes are executed?

@astronouth7303 I don't think so. I think it would make it more confusing.  Because we have 4 different ways of running commands, it would be unnecessarily duplicated code.  Also do they each get their own event or different events?  Also what happens if someone aliases `git` form a subprocess to some internal command: which event picks that up? How does the event handler know which element of a pipeline a command is if it only gets the single command?  

I think that those are some of the hard questions that would need to be answered if we went with per command events instead of per pipeline events.  Add to that that fact that the SubprocSpec and CommandPipeline are the only "public" parts of the API. That is, I have been working off the model that once a CommandPipeline has started it is in complete control and the only side effects are job control related.

In short, it is much easier to reason about pipelines than commands. I think that is what we should expose to the user too. At least at a first pass, since we'll always need to provide a way to get the pipeline.

scopatz:
@flavius - Just wanna make sure that this doesn't get lost.  Let us know when you think it is ready for another look.

flavius:
@scopatz Sorry for the silence, I got caught in some work. I'll be back to this quite soon and I'll update this PR.

gforsyth:
Hi @flavius -- I'm going through and closing all pull requests that have been idle for more than a year.  We'd love to have your contribution and if you come back to it, please feel free to either open a new PR or you can ping here to have us reopen the existing PR.

Thanks!

Issue #2186, opened 2017-02-13
git and/or vim breaking stdout


rtpg:
I've been having a bit of an issue with my shell output for a while (at least since a release ago), but I only just isolated some of the cause:

When I do `git commit` and use my default editor to edit the commit message (`vim`), after I commit, my output suddenly has tabs being "added" to it.

![image](https://cloud.githubusercontent.com/assets/1408472/22868976/f003cfe2-f1de-11e6-9965-63ad33541d85.png)

This problem persists for the rest of the shell session, and requires me to open a new terminal:

![image](https://cloud.githubusercontent.com/assets/1408472/22868989/038e4b96-f1df-11e6-8359-f0b014cf3673.png)

I have this problem when `git` is the one invoking `vim` (a similar issue when Github's `hub` tool invokes `vim`) but not when I just invoke `vim` by itself.

My theory would be that the issue comes from a process invoking another process, and then some state modifications because of that are missing (maybe `xonsh` doesn't properly reset the state in recursive execution?)

I hope this helps anyone working on the subprocess bug-squashing!

scopatz:
Hi @rtpg - thanks for reporting!  Can you try adding the following to your xonshrc:

```python
__xonsh_commands_cache__.threadable_predictors['git'] = lambda args: False
```

I think that this is because of PopenThread getting in the way somehow.  Though a couple things should be noted:

1. I don't see this issue with the text editors I regularly use (less, nano, xo), so this makes me think it is a vim-specific issue.

>>> xonsh doesn't properly reset the state in recursive execution

2. The way shells work, xonsh actually doesn't have anything to do with nested subprocess calls. So when git opens your editor, all xonsh know is that git is sending bytes to stdout & stderr. 

This seems like it might be sanity thing, though. So another thing to try would be after the git command, try running:

```python
>>> __xonsh_shell__.shell.restore_tty_sanity()
```

Let us know if either of those work.  Also, it would be great to know if other vim users are having this problem.

mitnk:
I've been encountering this issue for a few years (since started using vim?). And I think it may not related to xonsh. It also happens in Bash, IIRC. So it may not be a bug in shell level but in terminal level (not sure). But we still can dig it.

rtpg:
`__xonsh_commands_cache__.threadable_predictors['git'] = lambda args: False` did the trick on this issue. TTY sanity didn't help, though.

[...]

PR #2191, opened 2017-02-14
Fix for alias resolution in threadable predition

scopatz:
This should address some of the issued pointed out in #2186


gforsyth:
Thanks, @scopatz! This looks good, now we settle in for the long CI wait.

PR #2197, opened 2017-02-15
Prevent main proc from stopping

### LINKED ISSUE #2196: git commit stuck on Mac
When type `git commit` without an `-m` will got xonsh stuck on Mac OS. It doesn't happen on Linux. (I'd say Linux is too kind for devs :))

Xonsh stuck is because the xonsh process has hand off the controlling terminal to git process, but it still want write something to stdout/stderr. Thus it got a SIGTTOU and stopped by OS.

```
USER   PID  PPID  PGID TPGID STAT COMMAND
hugo 23584 23583 23583 23601 T    Python -u xonsh
hugo 23601 23584 23601 23601 S+   git commit
hugo 23602 23601 23601 23601 S+   vim COMMIT_EDITMSG
```

This is a new issue introduced by #2177, my bad. Fix is on the way.

mitnk:
Try fixes #2196. The current committed code actually fixed the stuck issue, but, obviously, is not the correct way.

The correct way I think should be 1) "not read from stdin or write to stdout/stderr when hand out the terminal" and/or 2) "do not settcattr when should not".

### DIFF
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -6,6 +6,7 @@ import enum
 import argparse
 import builtins
 import contextlib
+import signal
 import traceback
 
 from xonsh import __version__
@@ -340,6 +341,11 @@ def main(argv=None):
 
 def main_xonsh(args):
     """Main entry point for xonsh cli."""
+    def func_sig_ttin_ttou(n, f):
+        pass
+    signal.signal(signal.SIGTTIN, func_sig_ttin_ttou)
+    signal.signal(signal.SIGTTOU, func_sig_ttin_ttou)
+
     events.on_post_init.fire()
     env = builtins.__xonsh_env__
     shell = builtins.__xonsh_shell__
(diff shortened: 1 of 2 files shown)


scopatz:
Is this still a work in progress? I think that #2196 should block the weekly release for the moment, since we can't break git. FWIW, I tried this branch out and it seems to work well on Linux

scopatz:
Also thanks for putting in the fix!

mitnk:
Hi @scopatz, with the last commit added, I think we can merge it into master. This partly fix the `git commit` stucking issue on mac. We can do further clean up in later PRs.

mitnk:
no news item needed, since this is a new issue after 0.5.4.

scopatz:
Is the windows if-statement not needed?

mitnk:
It is needed, SIGTTIN are POSIX things.

scopatz:
Oh whoops I was looking at the wrong diff :)

scopatz:
Thanks!

PR #2205, opened 2017-02-16
Link the events tutorial and the events references

astronouth7303:
Because they really should be cross-linked.

(Docs-only change. No news.)


scopatz:
Thanks!

Issue #2202, opened 2017-02-15
Xonsh clears screen after open/close text editor


scopatz:
Whereas the previous commands and output should still be visible.  @mitnk any ideas? Or have you noticed this?

mitnk:
cannot reproduce this. any details?

scopatz:
Sorry Yeah, I was using `xo`

mitnk:
Checked with xo, and reproduced the issue. The solution is to revert #2191 - it will fix 1) this issue 2) #2198. And the revert won't break #2186, which #2191 was trying to fix (at least not an issue from my testing). I've added some context tof the reason to revert it [here](https://github.com/xonsh/xonsh/issues/2198#issuecomment-280087026). Thoughts @scopatz ?

scopatz:
Potential fixes in #2206 and #2211

Issue #2198, opened 2017-02-15
issue when entering xonsh multiple times


mitnk:
current master; repeat enter `xonsh` in `xonsh` will got errors:
```
$ xonsh
$ xonsh
$ xonsh
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/base_shell.py", line 322, in default
    run_compiled_code(code, self.ctx, None, 'single')
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/codecache.py", line 65, in run_compiled_code
    func(code, glb, loc)
  File "<xonsh-code>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/built_ins.py", line 856, in subproc_captured_hiddenobject
    return run_subproc(cmds, captured='hiddenobject')
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/built_ins.py", line 797, in run_subproc
    command = HiddenCommandPipeline(specs)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/proc.py", line 1704, in __init__
    if update_fg_process_group(pipeline_group, background):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/proc.py", line 1640, in update_fg_process_group
    return give_terminal_to(pipeline_group)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/jobs.py", line 139, in give_terminal_to
    os.tcsetpgrp(sys.stderr.fileno(), pgid)
OSError: [Errno 25] Inappropriate ioctl for device
```

This issue is introduced both by #2177 and #2191 - revert changes of #2191 will fix the issue (I'm not saying this is the solution thou), but I need to check what really happening and how to fix it.

scopatz:
`sys.stderr.fileno()` should probably be replaced with `2` for stability reasons. Thanks for reporting this, @mitnk.

mitnk:
In command_cache we have
```
predictors = {
    ...
    'xonsh': predict_shell,
    ...
}
```

mitnk:
[...]

PR #2206, opened 2017-02-17
Revert "Fix for alias resolution in threadable predition"

gforsyth:
Reverts xonsh/xonsh#2191

Putting this in as a place to continue discussion started by @mitnk in #2198 #2202 and elsewhere.

It does solve #2198 for me.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -676,7 +676,7 @@ def _update_last_spec(last):
     if callable_alias:
         pass
     else:
-        thable = builtins.__xonsh_commands_cache__.predict_threadable(last.cmd)
+        thable = builtins.__xonsh_commands_cache__.predict_threadable(last.args)
         if captured and thable:
             last.cls = PopenThread
         elif not thable:


scopatz:
Thanks for putting this in @gforsyth. I think that this restores the other bug with resolving aliases. I'll look into the underlying cause of the bug this weekend, because I think the code on master right now is correct.

mitnk:
Thanks @gforsyth for open this discussion.

> underlying cause of the bug .. I think the code on master right now is correct.

@scopatz I agree on the "reverting solution" is a temporary/not-ideal one. We need to make `Popen` and `PopenThread` both to work for same programs (i.e. fix the underlying issues).

> I think the code on master right now is correct.

Agreed. But a bit of small issues here in `threadable_predictors`, we defined `xonsh`, `xo` etc in the `predictors` mapping which would never be matched because or of them are `#!/usr/bin/python` and will use `python`s predictor. For other script-ed program too (e.g. I have a tool named `psj`, which is a bash script with `#! /bin/bash`, it will go through `bash`'s predictor).

scopatz:
See #2211

mitnk:
replaced by #2211, thanks again @gforsyth!

Issue #2130, opened 2017-01-30
readline windows issues.


melund:
It is has been a while since I used the readline shell on windows. I only switch to it when I need speed. 

Now it is giving me some issues: 

If I press tab when it launches, the 'Display all ## possibilities' question shows up with ansi escapes: 

![image](https://cloud.githubusercontent.com/assets/1038978/22416966/e59d5018-e6d1-11e6-9019-5c462b34723f.png)

If I press `n` I get this error: 
```
Readline internal error
Traceback (most recent call last):
  File "C:\Users\mel\Anaconda3\lib\site-packages\pyreadline\console\console.py", line 768, in hook_wrapper_23
    res = ensure_str(readline_hook(prompt))
  File "C:\Users\mel\Anaconda3\lib\site-packages\pyreadline\rlmain.py", line 571, in readline
    self._readline_from_keyboard()
  File "C:\Users\mel\Anaconda3\lib\site-packages\pyreadline\rlmain.py", line 536, in _readline_from_keyboard
    if self._readline_from_keyboard_poll():
  File "C:\Users\mel\Anaconda3\lib\site-packages\pyreadline\rlmain.py", line 556, in _readline_from_keyboard_poll
    result = self.mode.process_keyevent(event.keyinfo)
  File "C:\Users\mel\Anaconda3\lib\site-packages\pyreadline\modes\emacs.py", line 243, in process_keyevent
    r = self.process_keyevent_queue[-1](keyinfo)
  File "C:\Users\mel\Anaconda3\lib\site-packages\pyreadline\modes\emacs.py", line 286, in _process_keyevent
    r = dispatch_func(keyinfo)
  File "C:\Users\mel\Anaconda3\lib\site-packages\pyreadline\modes\basemode.py", line 257, in complete
    completions = self._get_completions()
  File "C:\Users\mel\Anaconda3\lib\site-packages\pyreadline\modes\basemode.py", line 200, in _get_completions
    r = self.completer(ensure_unicode(text), i)
  File "C:\Users\mel\Anaconda3\lib\cmd.py", line 275, in complete
    self.completion_matches = compfunc(text, line, begidx, endidx)
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\__amalgam__.py", line 18735, in completedefault
    show_completions = self._querycompletions(completions, endidx - begidx)
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\__amalgam__.py", line 18695, in _querycompletions
[...]

PR #2207, opened 2017-02-18
windows don't complete fix

scopatz:
This should partially address the issue in #2130. Particularly when the tab completer is responded to with a "no" for more completions. cc @melund


scopatz:
Note that saying yes doesn't seem problematic.

melund:
This fixes it :)

Issue #398, opened 2015-09-29
Pyreadline and color code issue in windows 


jsarver:
First off, this tool is really awesome and I love it. That being said I am having an unusual issue and I think I've narrowed it down to pyreadline. 

I am running windows 7 Miniconda3 python 3.4.3 64 bit and here are the steps I do to recreate the issue.

```
conda create -n shelltest -c scopatz xonsh
```

It then asks me to install the follwoing which I say yes.

```
The following NEW packages will be INSTALLED:

pip:        7.1.2-py34_0
ply:        3.7-py34_0
python:     3.4.3-0
setuptools: 18.3.2-py34_0
wheel:      0.26.0-py34_1
xonsh:      0.2.0-py34_1
```

Everything installs as expected so I run "Activate shelltest" to activate my virtualenv and type "xonsh" to run my xonsh shell and get the following prompt:

```
☺←[1;32m☻jsarver@mycomputername☺←[1;34m☻ ~☺←[1;32m☻ ☺←[1;34m☻$☺←[0m☻
```

Soo.. I get the odd color codes instead of the default green and blue prompt, that is strange. After some fooling around I discoverd if I install pyreadline then run xonsh I get this prompt:

```
jsarver@mycomputername ~ $
```

Ok now things are looking up, but now if I try to utilize any multiline functionality like defining a function in the xonsh command line like so:

```
def foo():
```

I get the following error.

```
jsarver@mycomputername ~ $ def foo():
pre_input_hook failed
Readline internal error
Traceback (most recent call last):
  File "C:\Miniconda3\envs\shelltest\lib\site-packages\pyreadline\modes\basemode.py", line 124, in readline_setup
    self.pre_input_hook()
  File "C:\Miniconda3\envs\shelltest\lib\site-packages\xonsh\readline_shell.py", line 81, in inserter
    readline.redisplay()
AttributeError: 'module' object has no attribute 'redisplay'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Miniconda3\envs\shelltest\lib\site-packages\pyreadline\console\console.py", line 768, in hook_wrapper_23
    res = ensure_str(readline_hook(prompt))
  File "C:\Miniconda3\envs\shelltest\lib\site-packages\pyreadline\rlmain.py", line 569, in readline
    self.readline_setup(prompt)
  File "C:\Miniconda3\envs\shelltest\lib\site-packages\pyreadline\rlmain.py", line 564, in readline_setup
[...]

PR #2208, opened 2017-02-18
Windows Terminal ANSI color fixes

scopatz:
This fixes the other bug in #2130, where ANSI color sequences don't display properly in windows terminals. cc @melund @laerus This builds on #2207, but shouldn't be a big deal

### DIFF
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -232,6 +232,10 @@ def rl_on_new_line():
         func = getattr(RL_LIB, name, None)
         if func is not None:
             break
+    else:
+        def print_for_newline():
+            print()
+        func = print_for_newline
     return func
 
 
@@ -304,10 +308,7 @@ class ReadlineShell(BaseShell, cmd.Cmd):
         show_completions = to_bool(yn)
         print()
         if not show_completions:
-            if rl_on_new_line is None:
-                rl_on_new_line()
-            else:
-                print()
+            rl_on_new_line()
             return False
         w, h = shutil.get_terminal_size()
         lines = columnize(completions, width=w)
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -302,7 +302,10 @@ class ReadlineShell(BaseShell, cmd.Cmd):
         show_completions = to_bool(yn)
         print()
         if not show_completions:
-            rl_on_new_line()
+            if rl_on_new_line is None:
+                rl_on_new_line()
+            else:
+                print()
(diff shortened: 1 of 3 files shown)


scopatz:
Also should address #398

xonsh/readline_shell.py
```diff
@@ -302,7 +304,10 @@ def _querycompletions(self, completions, loc):
         show_completions = to_bool(yn)
         print()
         if not show_completions:
-            rl_on_new_line()
+            if rl_on_new_line is None:
```
laerus: `is not* None` ?

laerus: uh that comes from the previous PR..

xonsh/winutils.py
```diff
@@ -249,6 +249,14 @@ def set_console_mode(mode, fd=1):
     SetConsoleMode(hcon, mode)
 
 
+def enable_virtual_terminal_processing():
+    """Enables virtual terminal processing on Windows.
+    This inlcudes ANSI escape sequence interpretation.
+    See http://stackoverflow.com/a/36760881/2312428
+    """
+    SetConsoleMode(GetStdHandle(-11), 7)
```
melund: Shouldn't we set this for both the readline shell and the PTK shell? Just so ANSI out is always handled correctly. 

Also, I think this can fail if we are on old versions of windows. In that case we can assume that ConHost doesn't have the capabilities.

See my question here: 
https://github.com/Microsoft/BashOnWindows/issues/1356#issuecomment-260704549

melund:
This solves the issues, so I am OK with merging. Just wondering if there were any benefits of `ENABLE_VIRTUAL_TERMINAL_PROCESSING` on the PTK shell.

xonsh/readline_shell.py
scopatz: Thanks for noticing.

xonsh/winutils.py
scopatz: > Shouldn't we set this for both the readline shell and the PTK shell? Just so ANSI out is always handled correctly.

I think we can do that easily if we find any cases that fail.  In general, though, it seems like PTK handles all of this stuff internally for us. PTK does set the console mode in a much a places.

> Also, I think this can fail if we are on old versions of windows. In that case we can assume that ConHost doesn't have the capabilities.

So right now, if it fails it is just a silent failure.  We could put in a warning in debug mode if you think it is needed.

scopatz:
OK! I have updated this PR.

PR #2211, opened 2017-02-18
new threadable prediction lookup

scopatz:
This is an alternative to #2206 and should address  #2198 #220. Thanks to @gforsyth and @mitnk for pointing this out


mitnk:
Yeah, this make more sense. Thanks!

scopatz:
Thanks! And sorry about the mix up originally.

Issue #956, opened 2016-05-20
Cannot Crlt-C out of wizard in readline shell


scopatz:
This is part of deeper issues with Crtl-C and readline that I can't figure out right now. I have tried a bunch of things, but none of them seem to work.

GuillaumeLeclerc:
It may worth nothing and I don't know what you tried but if you send some signal using kill at some point you may want to try sending the signals to the whole process group instead of a specific process.

This is what fixed all my problems regarding signals when I was trying to fix #855 and other issues related to job control

scopatz:
See potential fix in #2212

scopatz:
Closing this due to #2212, please feel free to reopen!

PR #2212, opened 2017-02-18
Readline wizard usability fixes

scopatz:
This is an attempt to address #956. It turns out there is an underlying issue with readline that eats Ctrl+C in Python (see https://bugs.python.org/issue24266).  Unfortunately, the fix - which has already been merged in - is only applicable to readline v7, which most people are not on yet. 

So as a usability fix, the wizard now prints the correct command to exit. This depends on the shell backend.

Additionally, the fixes another minor issue with the shell type not being set correctly on the shell class.

### DIFF
--- a/news/rlwiz.rst
+++ b/news/rlwiz.rst
@@ -1,0 +1,18 @@
+**Added:** None
+
+**Changed:**
+
+* ``Shell.stype`` has been renamed to ``Shell.shell_type``.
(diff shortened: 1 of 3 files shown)


scopatz:
ping! Anyone willing to take a look at this, please? It is a pretty small change

laerus:
legit, thanks for this @scopatz !

scopatz:
Thanks @laerus!

Issue #959, opened 2016-05-20
Print correctly the caracters typed before xonsh has started


GuillaumeLeclerc:
Hello,

Usually, since xonsh is usually slower than the other shells I tried, I end up typing before xonsh is fully started. Then my command is wrongly displayed on the terminal. The behavior is the following
- I type something, let's say 'Hello `
- Xonsh finishes loading
- The prompt line is append after `Hello`
- `Hello` is printed after the prompt line
- The cursor stays after the prompt line
- I type `how`. and since the cursor is just after the prompt line, everything under the cursor is replaced so I end up with : 

```
Hello ${PROMPT}howlo
                /\
             Cursor Here
                \/
${PROMPT}Hello how
   => This should be the correct output I think (note that the position of the cursor is correct. It's what I typed before, since it was duplicated, that shifts the text)
```

For the moment the only way I found to recover from this is to press `Ctrl-L` because everything is reprinted. Is there anything we could do on the xonsh side to fix this. `fish-shell` is behaving correctly in this situation.

scopatz:
Crazy, I didn't know that Ctrl-L would fix this.  There are probably prompt-toolkit and readline functions that we can use to signal to do this at the end of the prompt.  A PR for this would be most welcome.

GuillaumeLeclerc:
I think Ctrl-L would be a temporary solution.

I think we would need a more generic solution because:

- Some people would not like their screen wiped if they start xonsh (they might start it from another shell and still have somthing interesting on their screen)
- The issue occur also when you type while a command is executing. Example just do `$ sleep 5` and type `Hello` and it will have the exact same behavior.

Any idea on how we could solve that ?

scopatz:
Potential fix in #2214

scopatz:
Closing based on merger of #2214, please reopen if needed.

PR #2214, opened 2017-02-18
carriage return on the first prompt to properly capture all input

scopatz:
This should address #959 for both ptk and readline.

### DIFF
--- a/xonsh/ptk/shell.py
+++ b/xonsh/ptk/shell.py
@@ -172,7 +172,7 @@ class PromptToolkitShell(BaseShell):
         toks = partial_color_tokenize(p)
         if self._first_prompt:
             carriage_return()
-            self._first_message = False
+            self._first_prompt = False
         self.settitle()
         return toks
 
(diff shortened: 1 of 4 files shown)


xonsh/ptk/shell.py
```diff
@@ -169,6 +170,9 @@ def prompt_tokens(self, cli):
         except Exception:  # pylint: disable=broad-except
             print_exception()
         toks = partial_color_tokenize(p)
+        if self._first_prompt:
+            carriage_return()
+            self._first_message = False
```
gforsyth: should this be `self._first_prompt`?

gforsyth:
Hey @scopatz -- seems to work on PTK! I've never had this problem with readline so can't judge.

xonsh/ptk/shell.py
scopatz: Nice, thanks, it should be!

scopatz:
Ok the fix is in. I was seeing this problem on both ptk and readline, and this branch does fix both.

scopatz:
Also, this was somewhat tricky to figure out given how easy the solution is.

gforsyth:
> Also, this was somewhat tricky to figure out given how easy the solution is.

That's always the way of these things, isn't it?

scopatz:
> That's always the way of these things, isn't it?

Its it true! Especially for our oldest bugs....

melund:
Lgtm

scopatz:
Thanks @melund!

Issue #981, opened 2016-05-22
Confusing autocompletion, ghosted hint mismatches tab completion


mwiebe:
Running 0.3.0 on Windows. When I type "git ", it shows a ghosted autocomplete, but then if I hit Tab, the tab completion options don't include that. For example:

![image](https://cloud.githubusercontent.com/assets/399551/15451720/47b24a96-1f87-11e6-9b2b-e73792f51d51.png)

scopatz:
Hi @mwiebe - thanks for reporting this!  The ghosted value comes from your history.  My guess is that your Bash completions are not present in the `$BASH_COMPLETIONS` env variable: http://xon.sh/envvars.html#bash-completions. Thus , TAB is just giving you the standard Python completion guesses... which is obviously incorrect.  Can you check that env var and report back, please?

mwiebe:
Here's what my $BASH_COMPLETIONS has:

```
mwiebe@prax C:\Dev\autobuild master $ echo $BASH_COMPLETIONS
xonsh: subprocess mode: command not found: echo
mwiebe@prax C:\Dev\autobuild master $ print($BASH_COMPLETIONS)
[]
```

scopatz:
Ahh that is it then.  Do you have git's bash completions installed?  If so you can simply set this list in the xonshrc file to contain a string that points to them.  I think the Windows default is an empty tuple because there is no standard place to install these to.  @melund and @BYK may know more since they are the xonsh devs that actively use windows.

mwiebe:
I've tried adding the bash-completions directory from the git install, but that doesn't seem to help. Will try also switching all the `\` to `/`.

```
mwiebe@prax C:\Dev\autobuild master $ $BASH_COMPLETIONS.append("C:\Program Files\Git/mingw64/share/git/completion")
mwiebe@prax C:\Dev\autobuild master $ print($BASH_COMPLETIONS)
['C:\\Program Files\\Git/mingw64/share/git/completion']
mwiebe@prax C:\Dev\autobuild master $ gfind 'C:\\Program Files\\Git' "\\*bash\\*"
mwiebe@prax C:\Dev\autobuild master $ git status
                                             StopAsyncIteration staticmethod
                                             StopIteration      str
```

mwiebe:
That didn't make a difference:

![image](https://cloud.githubusercontent.com/assets/399551/15451773/d3fa139c-1f89-11e6-84d3-9cdd143aa466.png)

[...]

PR #2218, opened 2017-02-19
Complete based on autosuggestion

scopatz:
This is attempt to solve the long running completion-based-off-of-autosuggestion issue from #981. Now, if `$AUTO_SUGGEST`  is enabled, the first completion will always be the current token of the autosugestion.


melund:
Awesome. Thanks a lot for adding this @scopatz

melund:
Code looks good to me, and it works as I would expect.

PR #2223, opened 2017-02-19
raw path completer fixes

scopatz:
This should enable more robust raw string path completers on windows. It should address the remaining issues in #983.

### DIFF
--- a/news/rawpath.rst
+++ b/news/rawpath.rst
@@ -1,0 +1,15 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fixed issues pertaining to completing from raw string paths.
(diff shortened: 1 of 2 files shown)


melund:
Not completely sure what issue this fixes. But I can't see anything wrong with the code and haven't seen any problems working with it. So I will go ahead and merge. Maybe you can reveal what it solves :)

scopatz:
Thanks @melund. Basically, its solves some problems when completing a path that starts with `r"`, like `r'C:\\Docu<TAB>`. There were many issues here:

1. Sometimes the completer would enter an infinite recursion because it wasn't handling the drive name correctly.
2. If you didn't have an ending quote, this would add it (which was good). If you did have an ending quote, it would still add (which was bad) because you would end up accumulating ending quotes at the end of the completion.

I think there were some other small issues that were also handled, but these were the big ones.

Issue #1043, opened 2016-05-30
Matplotlib interactive mode freezes the figures


melund:
I have tried to use matplotlib from xonsh. As far as creating the figure it works fine. 

```
snail@sea ~ $ from matplotlib import pyplot as plt                                    
snail@sea ~ $ plt.plot([1,3])                                                         
[<matplotlib.lines.Line2D at 0x15eeb318dd8>]
snail@sea ~ $ plt.show()                                                              
```

This blocks xonsh similar to what it does in python, and the figure is shown. It also behaves nicely if I close the figure. But xonsh crashes if I instead try to close the figure by pressing ctrl-c in xonsh.

```
snail@sea ~ $ plt.show()                                                              
<ctrl-c>
QObject::killTimers: timers cannot be stopped from another thread
```

If I try to use interactive mode: 

```
snail@sea ~ $ plt.ion()                                                            
snail@sea ~ $ plt.plot([1,2])                                                        
[<matplotlib.lines.Line2D at 0x1df8de65ac8>]
snail@sea ~ $
```

Then xonsh does not block, but the figure opens  in a non-responsive state:

![image](https://cloud.githubusercontent.com/assets/1038978/15650464/0c681d28-2679-11e6-86be-9664de0fef41.png)

if I try to close the figure (clicking x on the windows) I get a Python crashed dialog: 
![image](https://cloud.githubusercontent.com/assets/1038978/15650481/2a98da12-2679-11e6-9369-3450edec3e65.png)

The figure window stays in an unresponsive state. If I then try to run `plt.close()` from xonsh then the figure windows closes down nicely.

JohnLunzer:
I did some research into getting pyqtgraph to run interactively in xonsh as well. I was drawn to [how this is achieved in IPython](http://ipython.readthedocs.io/en/stable/config/eventloops.html) ,  which has what looks to be an extensive system for integrating GUI event loop hooks.

It seems like it would be no small feat to integrate this type of system in Xonsh. I think it would make sense to pull in IPython's system if this effort were to be undertaken.

scopatz:
Hmm strange, if I run `ion()` I don't see any of the figures after this, even if I run `show()`.

scopatz:
Potential fix in #2225

melund:
#2225 fixes part of the problem on Windows. Will close this one and open a new issue.

scopatz:
Thanks @melund!

PR #2225, opened 2017-02-19
Import Events and Matplotlib Interactive Fixes

scopatz:
This PR adds six new events for hooking into Python's import mechanism.  Now you can register a handler for before/after a module is search for, created, or exec'd. There are probably some pretty crazy abuses of this. 

The initial application of this is to help fix some of the interactivity issues with matplotlib. This should close out #1043. Some of the problems mentioned in that issue seemed to related to mpl v1.x and have since been fixed in mpl v2.x.  This PR tries to address the remaining problems.

### DIFF
--- a/news/ie.rst
+++ b/news/ie.rst
@@ -1,0 +1,29 @@
+**Added:**
+
+* New events for hooking into the Python import process are now available.
+  You can now provide a handler for:
+
+  - ``on_import_pre_find_spec``
+  - ``on_import_post_find_spec``
+  - ``on_import_pre_create_module``
+  - ``on_import_post_create_module``
+  - ``on_import_pre_exec_module``
+  - ``on_import_post_exec_module``
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* The ``mpl`` xontrib has been updated to improve matplotlib
+  handling. If ``xontrib load mpl`` is run before matplotlib
+  is imported and xonsh is in ineteractive mode, matplotlib
+  will automatically enter interactive mode as well. Additionally,
+  ``pyplot.show()`` is patched in interactive mode to be non-blocking.
+  If a non-blocking show fails to draw the figre for some reason,
+  a regular blocking version is called.
+
+**Security:** None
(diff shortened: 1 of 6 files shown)


astronouth7303:
This seems absurdly dangerous. Are there other ways to accomplish this without new events? Or with events that are not so broad in scope and power?

scopatz:
I don't know that this is any more or less dangerous than what Python's import system already provides. Anyone could write an import finder and stick it at the start of the `sys.meta_path` that would totally disrupt the import mechanism, if they wanted.  All this really provides is a translation layer between the two powerful systems. It doesn't stop bad behaviour but it does make good behaviour easier.

melund:
@scopatz : I can't get this working: 

image

It just shows a blank figure. If I then run `plt.show()` the figure renders but then the prompt it blocked. Closing the figure returns the prompt. But if I press `ctrl-c` instead I get the following error. 
```
snail@sea ~\Documents\xonsh ie
$ plt.show() (CTRL-C)
QObject::killTimers: timers cannot be stopped from another thread
PS C:\Users\mel\Documents\xonsh>
```

scopatz:
Thanks for trying it out @melund. I think this is almost certainly a matplotlib & windows issue. I know that they have had issues like this open in the past (https://github.com/matplotlib/matplotlib/issues/3052).  Perhaps it is best to report this there?  

This PR does fix some of the issues I was seeing on Linux, though.

Having dived into the matplotlib code to write this PR, I am not convinced that their interactive mode is really well thought out. They special case a lot of stuff to work for just IPython. It would be much better to have a real API to work with.

melund:
Then I am OK with merging.

scopatz:
Thanks @melund! I *do* think this is worth reporting over at matplotlib. They seem to be trying to move towards something better. Bringing it up again couldn't hurt.

melund:
I agree. Also maybe a new issue for xonsh so we don't forget. I don't have time to do it right now thus.

scopatz:
Sure, but let's close out the old bug and label the new one as a feature.

melund:
Agreed

Den tir. 21. feb. 2017 05.47 skrev Anthony Scopatz <notifications@github.com

Issue #1064, opened 2016-06-02
Shell foreign interface doesn't work or warn for overriding xonsh builtin functions


yangchenyun:
It is normal in bash profile to override some builtin bash command, such as `cd`: 

``` sh
cd() {
  builtin cd "$@" && ls -AF
}
```

But the imported `cd` overrides what `xonsh` has (without any warning) and then the new `cd` failed to change the `$PWD` of the shell.

Doest the foreign interface understands `builtin` syntax? A possible solution maybe remapping shell's builtin command to `xonsh`'s builtin function.

willsALMANJ:
I noted a similar issue in #1056. The problem with `cd` in an alias is that `xonsh` runs the foreign shell alias in the foreign shell run as a subprocess. Any output from the alias gets printed back to `xonsh` normally but changing environment variables or the working directory in the alias changes the foreign shell inside the subprocess and that is not read back into the parent `xonsh` shell.

scopatz:
Thanks @yangchenyun and @willsALMANJ.  I am thinking about this one.  I am not sure that it is safe to read back in the env and aliases because this would overwrite all other envs and alaises that happen to be defined in the same file as the foreign function.  There is no current way of knowing which envvars a foreign function changes, and therefore which ones we should allow to be changed.  

Maybe we could allow running some in an unsafe mode or perhaps allow for the alias object to have a set of envvar names and alias names it reads back in?

Additionally, we are considering dropping the loading of the bash configuration by default.  This will make it so that this won't happen to users without them explicitly asking for it.

scrool:
Another example and closer to python usage. [Using Autoenv: Directory-based Environments](https://github.com/kennethreitz/autoenv):

What autoenv's activate does:
```
(bash) $ source /usr/bin/activate.sh
(bash) $ type cd
cd is a function
cd () 
{ 
    autoenv_cd "$@"
}
(bash) $ type autoenv_cd
autoenv_cd is a function
autoenv_cd () 
{ 
    if builtin cd "$@"; then
        autoenv_init;
        return 0;
    else
        return $?;
    fi
}
```

Enable it in bashrc and run xonsh:

```
(bash) $ echo 'source /usr/bin/activate.sh' >> ~/.bash_profile
(bash) $ xonsh
(xonsh) $ pwd
/
(xonsh) $ cd /tmp
(xonsh) $ pwd
/
```

scopatz:
See #2226 for a potential fix.

[...]

PR #2226, opened 2017-02-19
Skip existing foreign shell aliases

scopatz:
This PR adds some logic to skip existing foreign shell aliases. This should hopefully address #1064


melund:
Lgtm.

Issue #1091, opened 2016-06-04
@$() doesn't split quoted strings correctly


asmeurer:
The docs claim that `@$(cmd)` does `@([i.strip() for i in $(cmd).split()])`, which is correct, but not helpful, because if `cmd` contains quoted strings with spaces, they are not parsed as you would hope. In general, there is no way to escape spaces with this.

scopatz:
Yep, I agree. This is a first pass. We need something that splits on whitespace that is not in a string, by default. This should be a do-able regex.

adqm:
FWIW, bash seems to do what is described in the docs:
```
$ ls `echo \"this dir\"`
ls: cannot access "this: No such file or directory
ls: cannot access dir": No such file or directory
```

That said, I think better handling of strings would be nice, since that's almost certainly what people intend.

wreed4:
shlex.split does exactly what you need instead of the default str.split used above.

scopatz:
I think this should be implemented with the lexer itself.

scopatz:
Potential fix in #2227

gforsyth:
Closing this out since #2227 was merged but please reopen if the issue persists.

PR #2227, opened 2017-02-20
Lexer splitter

scopatz:
This adds a new split() method to the lexer. This is then applied to the `@$(cmd)` syntax. This should address #1091.


gforsyth:
Thanks, @scopatz !

scopatz:
Thanks @gforsyth!

Issue #2224, opened 2017-02-19
Bash On Windows causes platform.windows_bash_command() to raise CalledProcessError.


ericmharris:
Bash On Ubuntu On Windows (BOW)  is a new feature available in Windows 10 Anniversary Update with Developer Mode enabled.  Background here: https://msdn.microsoft.com/en-us/commandline/wsl/about

Installing BOW creates C:\WINDOWS\system32\bash.exe, and this causes the bash_completer to choke because it's expecting to find git-bash.

I've created a demo fix that's probably a really bad idea, but it'll help show the problem.

https://github.com/ericmharris/xonsh.git

[emharris@eharris-laptop xonsh]# ls Exception in thread Thread-113:
Traceback (most recent call last):
  File "C:\tools\Anaconda3\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "C:\tools\Anaconda3\lib\threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "C:\tools\Anaconda3\lib\site-packages\prompt_toolkit\interface.py", line 840, in run
    existing complete_state if we had one. (The user could have
  File "C:\Users\emharris\Dropbox\Development\xonsh\xonsh\ptk\completer.py", line 49, in get_completions
    self.ctx)
  File "C:\Users\emharris\Dropbox\Development\xonsh\xonsh\completer.py", line 35, in complete
    out = func(prefix, line, begidx, endidx, ctx)
  File "C:\Users\emharris\Dropbox\Development\xonsh\xonsh\completers\bash.py", line 13, in complete_from_bash
    command = xp.bash_command()
  File "C:\Users\emharris\Dropbox\Development\xonsh\xonsh\platform.py", line 329, in bash_command
    bc = windows_bash_command()
  File "C:\Users\emharris\Dropbox\Development\xonsh\xonsh\platform.py", line 309, in windows_bash_command
    universal_newlines=True)
  File "C:\tools\Anaconda3\lib\subprocess.py", line 336, in check_output
    **kwargs).stdout
  File "C:\tools\Anaconda3\lib\subprocess.py", line 418, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['C:\\WINDOWS\\system32\\bash.exe', '--version']' returned non-zero exit status 4294967295

scopatz:
[...]

PR #2228, opened 2017-02-20
Handle if WSL bash misbehaves on Windows

melund:
This should address #2224, where @ericmharris reported that `Windows subsystem for Linux` (WSL) bash did not return a correct version string.

### DIFF
--- a/news/wsl_bash.rst
+++ b/news/wsl_bash.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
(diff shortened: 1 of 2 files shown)


scopatz:
LGTM @melund, thanks for the fix!

scopatz:
It might be worth making a similar change to py-bash-completion. cc @imomaliev

imomaliev:
@scopatz I think it might, will look into this later

PR #2229, opened 2017-02-20
codecov raise threshold for commit status

laerus:


codecov.yml
```diff
@@ -0,0 +1,5 @@
+coverage:
+  status:
+    project:
+      default:
+        threshold: 1
```
astronouth7303: Unit?

laerus: yeah i had it at first then codecov docs made me change that
https://docs.codecov.io/docs/commit-status#section-threshold
ambiguity wins here

astronouth7303: So it's in percentage points.

astronouth7303: (YAML allows comments)

scopatz:
Thanks @laerus! I really appreciate it!

PR #2230, opened 2017-02-21
Xolors

scopatz:
This PR refactors and merges in the important parts of the xonsh/xolors repo. This is done to enable dynamically creating style color maps from pygments styles. This itself is necessary to be able to pick up 3rd party pygments styles that may be installed on the user's system. This should address #1097.  Like most color handling, it is probably overly verbose for what it actually is.


melund:
Just reading through the code. So creating a new style corresponds to updating the  BASE_XONSH_COLORs to their closest match in whatever colours are available in the pygments-style. 

I guess that makes sense. Couldn't we do the same for all pygment styles instead of listing them manually in the code: 
https://github.com/xonsh/xonsh/blob/master/xonsh/pyghooks.py#L974

melund:
The PTK shell seems to behave nicely with this. I am probably not the one to test the readline backend.

melund:
On a side note. Would be possible to expand the XONSH_BASE_COLORS to include all the colours from the xkcd color survey https://xkcd.com/color/rgb/

melund:
I will go ahead and merge this in. The PTK backend behaves nicely, and the readline code seems identical

melund:
Thanks for adding this @scopatz

scopatz:
Hey @melund, thanks for merging and sorry for not responding earlier.  

So the `BASE_XONSH_COLORS` just map to the the 16 base names (RED, etc) to the basics color codes as defined on wikipedia. Like you mention, we then find the closest values to these colors. `BASE_XONSH_COLORS` probably shouldn't be added to, though a different basis could be chosen. 

> Couldn't we do the same for all pygment styles instead of listing them manually in the code

We could, with the exception of default, but then that would be more expensive on startup. precomputing them seems like the right thing to do for known styles.

Speaking of which, we can/should/must have an xkcd style :)

PR #2232, opened 2017-02-21
Replace deprecated calls to `time.clock()` by `time.perf_counter()`

### LINKED ISSUE #2163: time.clock() is deprecated
Should remove and replace in timings module.

StreakyCobra:
Close #2163

### DIFF
--- a/news/fix_time_clock.rst
+++ b/news/fix_time_clock.rst
@@ -10,5 +10,7 @@
 
 * Replace deprecated calls to ``time.clock()`` by calls to
   ``time.perf_counter()``.
+* Use ``clock()`` to set the start time of ``_timings`` in non-windows instead
+  of manually setting it to ``0.0``.
 
-**Security):** None
+**Security:** None
--- a/news/fix_time_clock.rst
+++ b/news/fix_time_clock.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Replace deprecated calls to ``time.clock()`` by calls to
+  ``time.perf_counter()``.
+
+**Security):** None
--- a/xonsh/timings.py
+++ b/xonsh/timings.py
@@ -28,7 +28,7 @@ def _HAVE_RESOURCE():
         have = True
     except ImportError:
         # There is no distinction of user/system time under windows, so we
-        # just use time.clock() for everything...
+        # just use time.perf_counter() for everything...
         have = False
     return have
 
@@ -44,12 +44,11 @@ def clocku():
     if _HAVE_RESOURCE:
         def clocku():
             """clocku() -> floating point number
-            Return the *USER* CPU time in seconds since the start of the process.
-            This is done via a call to resource.getrusage, so it avoids the
-            wraparound problems in time.clock()."""
+            Return the *USER* CPU time in seconds since the start of the
+            process."""
             return resource.getrusage(resource.RUSAGE_SELF)[0]
     else:
-        clocku = time.clock
+        clocku = time.perf_counter
     return clocku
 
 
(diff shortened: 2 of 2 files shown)


xonsh/timings.py
```diff
@@ -46,10 +46,10 @@ def clocku():
             """clocku() -> floating point number
             Return the *USER* CPU time in seconds since the start of the process.
             This is done via a call to resource.getrusage, so it avoids the
-            wraparound problems in time.clock()."""
+            wraparound problems in time.perf_counter()."""
```
StreakyCobra: I'm not sure to understand what are the «wraparound problems» there, so I don't know if a deeper refactoring should be done, or if the comment still make sense after changing the call to `time.perf_counter()`.

melund:
Thanks for fixing this @StreakyCobra .  You probably need to remove the ` if ON_WINDOWS else 0.0` part here: 

https://github.com/xonsh/xonsh/blob/master/xonsh/timings.py#L231

This is used for the `xonsh --timings` command. 

As I understand the new `perf_counter` doesn't have a well-defined starting time. Previously, that was only an issue on Windows, while `clock()` on Linux would always be zero when the process started.

xonsh/timings.py
melund: I don't know what it refers to either. Probably not an issue, but @scopatz may know.

news/fix_time_clock.rst
```diff
@@ -0,0 +1,13 @@
...
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Replace deprecated calls to `time.clock()` by calls to `time.perf_counter()`.
```
melund: This has to be ` ``time.clock()`` ` in restructured text.

StreakyCobra: Indeed!

StreakyCobra:
@melund:

> You probably need to remove the if ON_WINDOWS else 0.0 part

Not sure to understand all this, I'm not experienced with xonsh. For me it should already have been `clock()` on linux instead of setting it manually to `0.0`, event before this change. I've put this in a separate if you don't want to include it.

scopatz:
> Not sure to understand all this, I'm not experienced with xonsh. For me it should already have been clock() on linux instead of setting it manually to 0.0 

This was just a small performance boost since clock() was guaranteed to be zero

news/fix_time_clock.rst
```diff
@@ -0,0 +1,16 @@
...
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Replace deprecated calls to ``time.clock()`` by calls to
+  ``time.perf_counter()``.
+* Use ``clock()`` to set the start time of ``_timings`` in non-windows instead
+  of manually setting it to `0`.
```
scopatz: Still have a single backtick around the zero :)

StreakyCobra: Rhaaaaaaaaa

StreakyCobra:
> This was just a small performance boost since clock() was guaranteed to be zero

@scopatz So this mean it should be kept then, no?

scopatz:
@StreakyCobra What you have in the PR right now is correct because of the differences between clock() and `perf_counter()`  clock() was guaranteed to start at zero.

StreakyCobra:
Ok got it :+1:

xonsh/timings.py
scopatz: I think this is fine to keep as it is. The wraparound comment is not valid for `perf_counter()`, but still makes sense for clock().  You should either restore the comment to say clock() (which we are still avoiding) or remove the whole comment.  

The problem with clock is that it uses uint32s. So it was possible to reset times back to zero if you got to big.  `perf_counter()` uses floats under the covers and so doesn't have this issue.

scopatz:
Thanks a million for putting this in! I really appreciate it!

xonsh/timings.py
StreakyCobra: So, as `clock` is deprecated and won't be used again anyway, we can probably just remove the comment?

scopatz: probably

news/fix_time_clock.rst
```diff
@@ -0,0 +1,16 @@
...
+**Removed:** None
+
+**Fixed:**
+
+* Replace deprecated calls to ``time.clock()`` by calls to
+  ``time.perf_counter()``.
+* Use ``clock()`` to set the start time of ``_timings`` in non-windows instead
+  of manually setting it to ``0.0``.
+
+**Security):** None
```
scopatz: Please remove the extra parentheses here.

StreakyCobra: What the hell… done :+1:

melund:
This looks good @StreakyCobra. Thanks a lot for fixing this.

Issue #2151, opened 2017-02-04
Refactor into xoretribs


astronouth7303:
Several builtin functionalities aren't necessary for xonsh to be usable, just useful. These should be refactored into stand alone xontrib-like modules: xoretribs (xonsh+core+contrib).

They are structured and loaded like xontribs, but they don't appear in the xontrib registry, cannot be managed through the `xontrib` command, and are loaded at startup unconditionally. (After core initialization and before rc files.)

The biggest requirement is that xonsh must still load and run commands without a xoretrib.

* [ ] VCS support
* [ ] `which`
* [ ] History
* [ ] Most of the builtin aliases?

(Note: I'm not saying that you'd ever _want_ to use a shell without these, just that the shell _can_ run without these.)

The goal is to provide better modularity within xonsh, so that different aspects depend less on each other. Hopefully, this should 1. reduce code complexity, 2. Improve APIs, and 3. improve the experience of developing actual xontribs.

astronouth7303:
(This is another reminder issue. I'll probably do this myself, I just want it out there that it's going to happen.)

scopatz:
I am 100% for more modularity. But just out of curiosity, what are you thinking in terms of implementation? I think it still makes sense for many of these pieces (perhaps on a case-by-case basis) to have their implementation in `xonsh/` and then provide a xontrib shim for loading and unloading. Moving the implementation to a xontrib would be kind of annoying for xonsh-as-a-library.

Also to the same degree as history, you can also add completion.

astronouth7303:
I figure moving them into `xonsh.xoretribs` or `xonsh.xoreutils`, and just using an `__init__.py` to import them. (We can also amalgamate then, unless we go crazy with subpackages.) This is pretty similar to what `xontribs` does, except for updating the shell context.

Yeah, the specific completion implementations can be moved as well, but not the completion framework itself. I'll add that to the list.

astronouth7303:
Things I don't think should be moved: Parser, executor, initialization, the xontribs system (even though it _could_), the completions system, shells, jupyter support.

Another aspect should be that APIs used by xoretribs should be public, and they should be valid as xontribs. eg if someone is looking to write a xontrib, they can also look at the xoretribs as examples.

PR #2233, opened 2017-02-21
Fast hg branch

gforsyth:
This needs more work, but I wanted folks to take a look at it (ping @ngoldbaum @laerus) to make sure I'm not missing any `hg` specific issues. 

Also, the usage of `$VC_BRANCH_TIMEOUT` will probably have to change (or be removed entirely)

Preliminary checks on my machine show something like a 1000x speed increase, which ain't too bad.

```
In [1]: import subprocess

In [2]: from xonsh.prompt.vc import _get_hg_root

In [3]: import queue

In [4]: pwd
Out[4]: '/home/gil/git/pygments-main/pygments/styles'

In [5]: subprocess.check_output(['hg', 'root'])
Out[5]: b'/home/gil/git/pygments-main\n'

In [6]: %timeit subprocess.check_output(['hg', 'root'])
10 loops, best of 3: 80.4 ms per loop

In [7]: q = queue.Queue()

In [8]: _get_hg_root(q)
   ...: q.get_nowait()
   ...: 
Out[8]: '/home/gil/git/pygments-main'

In [9]: %%timeit 
   ...: _get_hg_root(q)
   ...: q.get_nowait()
   ...: 
The slowest run took 6.81 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 65.4 µs per loop
```

### DIFF
--- a/xonsh/prompt/vc.py
+++ b/xonsh/prompt/vc.py
@@ -79,19 +79,6 @@ def get_hg_branch(root=None):
         root = q.get_nowait()
     except queue.Empty:
         return None
-#    if t.is_alive():
-#        return subprocess.TimeoutExpired(['hg'], timeout)
-
-#    try:
-#        root = subprocess.check_output(['hg', 'root'], timeout=timeout,
-#                                       stderr=subprocess.DEVNULL)
-#    except subprocess.TimeoutExpired:
-#        return subprocess.TimeoutExpired(['hg'], timeout)
-#    except (subprocess.CalledProcessError, FileNotFoundError):
-#        # not in repo or command not in PATH
-#        return None
-#    else:
-#        root = xt.decode_bytes(root).strip()
     if env.get('VC_HG_SHOW_BRANCH'):
         # get branch name
         branch_path = os.path.sep.join([root, '.hg', 'branch'])
--- a/xonsh/prompt/vc.py
+++ b/xonsh/prompt/vc.py
@@ -52,6 +52,18 @@ def get_git_branch():
         branch = None
     return branch
 
+def _get_hg_root(q):
+    _curpwd = os.getcwd()
+    while True:
+        if any([b.name == '.hg' for b in os.scandir(_curpwd)]):
+            q.put(_curpwd)
+            break
+        else:
+            _oldpwd = _curpwd
+            _curpwd = os.path.split(_curpwd)[0]
+            if _oldpwd == _curpwd:
+                return False
+
 
 def get_hg_branch(root=None):
     """Try to get the mercurial branch of the current directory,
@@ -59,16 +71,27 @@ def get_hg_branch(root=None):
     """
     env = builtins.__xonsh_env__
     timeout = env['VC_BRANCH_TIMEOUT']
+    q = queue.Queue()
+    t = threading.Thread(target=_get_hg_root, args=(q,))
+    t.start()
+    t.join(timeout=timeout)
     try:
-        root = subprocess.check_output(['hg', 'root'], timeout=timeout,
-                                       stderr=subprocess.DEVNULL)
-    except subprocess.TimeoutExpired:
-        return subprocess.TimeoutExpired(['hg'], timeout)
-    except (subprocess.CalledProcessError, FileNotFoundError):
-        # not in repo or command not in PATH
+        root = q.get_nowait()
+    except queue.Empty:
         return None
-    else:
-        root = xt.decode_bytes(root).strip()
+#    if t.is_alive():
+#        return subprocess.TimeoutExpired(['hg'], timeout)
+
+#    try:
+#        root = subprocess.check_output(['hg', 'root'], timeout=timeout,
+#                                       stderr=subprocess.DEVNULL)
+#    except subprocess.TimeoutExpired:
+#        return subprocess.TimeoutExpired(['hg'], timeout)
+#    except (subprocess.CalledProcessError, FileNotFoundError):
+#        # not in repo or command not in PATH
+#        return None
+#    else:
+#        root = xt.decode_bytes(root).strip()
     if env.get('VC_HG_SHOW_BRANCH'):
         # get branch name
         branch_path = os.path.sep.join([root, '.hg', 'branch'])
--- a/xonsh/prompt/vc.py
+++ b/xonsh/prompt/vc.py
@@ -52,6 +52,7 @@ def get_git_branch():
         branch = None
     return branch
 
+
 def _get_hg_root(q):
     _curpwd = os.getcwd()
     while True:
--- a/news/fast_hg_branch.rst
+++ b/news/fast_hg_branch.rst
@@ -1,0 +1,15 @@
+**Added:** None
+
+**Changed:**
+
+* ``prompt.vc.get_hg_branch`` now uses ``os.scandir`` to walk up the filetree
+  looking for a ``.hg`` directory. This results in (generally) faster branch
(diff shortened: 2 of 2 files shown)


gforsyth:
Oh, and ping @melund to see if this offers any speedups on windows

scopatz:
Code looks good to me.  Also pining @MatthewTurk since I know he is interested.

MatthewTurk:
This looks nice!  I haven't returned to #2083  (hit a roadblock with one aspect of it), but this looks to me like it might make a lot of that unnecessary anyway.

xonsh/prompt/vc.py
```diff
@@ -52,23 +52,33 @@ def get_git_branch():
...
+    _curpwd = os.getcwd()
+    while True:
+        if any([b.name == '.hg' for b in os.scandir(_curpwd)]):
+            q.put(_curpwd)
+            break
+        else:
+            _oldpwd = _curpwd
+            _curpwd = os.path.split(_curpwd)[0]
+            if _oldpwd == _curpwd:
+                return False
```
ngoldbaum: What's strange to me is that this is how I originally implemented this:

https://github.com/xonsh/xonsh/pull/239/files#diff-b08c376dd29024ccd6f9aef90f2ea96fR220

But for some reason it was subsequently changed. I have vague memories of someone complaining about issues on Windows, but I can't find the pull request where this behavior was subsequently changed.

melund: Yes. But that wasn't why it was changed. At least that is what I remember. The difference here is that @gforsyth runs it in a thread. Or am I misreading it?

melund: Ok. Looking at the code, it doesn't seem the queue thing makes a difference for the speed. I also recall this was changed originally due to speed issues. Can't recall who handled it. `git blame` can probably help us find the right issue.

melund:
@gforsyth Should you time the function with a blocking call. Will `timeit` show if it raises an `Empty` exception? Also, isn't it more interesting to time how long it takes when you are not in a mercurial repository?

melund:
Here is the issue from the last time this was changed: https://github.com/xonsh/xonsh/pull/1942

ngoldbaum:
On most filesystems traversing the filesystem to look for a .hg directory will be substantially faster than waiting for hg to startup. However on e.g. a network filesystem or any fs where stats are slow, it might be very slow to traverse all parent directories to look for a .hg directory. Personally I'd err on the side of looking for a .hg directory since it will be faster for most users, but that's just me.

gforsyth:
> Personally I'd err on the side of looking for a .hg directory since it will be faster for most users, but that's just me.

I concur. 

> Should you time the function with a blocking call? Will timeit show if it raises an Empty exception? Also, isn't it more interesting to time how long it takes when you are not in a mercurial repository?

I think the timing was legit given that I'm asking for the queue to respond and `timeit` will barf if it throws an exception. 
I ran some preliminary timing stuff in deep non-hg directories -- I'll rerun those and post up some timings.

The threading stuff may be completely unnecessary -- I was following the template from the git check as a way to have the `VC_BRANCH_TIMEOUT` still mean something.

melund:
It was @laerus  who originally reported that directory crawling was slow. See #1936 

Maybe he can chime in here...

gforsyth:
master
```
gil@bad_cat ...b/a/c/c/cha <branch-timeout> 🐚  pwd
/home/gil/tmp/foo/bar/baz/bag/bin/bwa/buu/bee/boo/ah/cha/cha/cha
gil@bad_cat ...b/a/c/c/cha 🐚  for i in range(20):
.............................     tic = time.time()
.............................     get_hg_branch()
.............................     print(time.time() - tic)
.............................     
0.10229611396789551
0.06147575378417969
0.061022281646728516
0.09182953834533691
0.07280611991882324
0.0730445384979248
0.06133460998535156
0.10456323623657227
0.060903310775756836
0.05923032760620117
subprocess.TimeoutExpired(['hg'], 0.1)
0.11464619636535645
0.0662076473236084
0.061403751373291016
0.08989596366882324
0.06216120719909668
0.08715391159057617
subprocess.TimeoutExpired(['hg'], 0.1)
0.11542820930480957
0.06349849700927734
0.06175732612609863
0.10113906860351562
```

this branch
```
gil@bad_cat ...b/a/c/c/cha 🐚  pwd
/home/gil/tmp/foo/bar/baz/bag/bin/bwa/buu/bee/boo/ah/cha/cha/cha
gil@bad_cat ...b/a/c/c/cha 🐚  for i in range(20):
.............................     tic = time.time()
.............................     get_hg_branch()
.............................     print(time.time() - tic)
.............................     
0.0016951560974121094
0.0016422271728515625
0.0015931129455566406
0.0014965534210205078
0.001451730728149414
0.0016002655029296875
0.0013582706451416016
0.001369476318359375
0.0014560222625732422
0.00140380859375
0.0012578964233398438
0.0019445419311523438
0.0013194084167480469
0.0021021366119384766
0.002000093460083008
0.0012085437774658203
0.0009868144989013672
0.0011036396026611328
0.0011038780212402344
0.0011391639709472656
```

so, excepting possible network drive hiccups, still seems like at least an order of magnitude faster and I might be able to remove the `try-except` and get it faster?

laerus:
nice @gforsyth! sadly i can't help you much, i am not so familiar with hg, not using it tbh.

xonsh/prompt/vc.py
laerus: i think i touched this code too at some point, my prompt was getting too slow when i was deep in the dir tree cause of all the recursive searches up to root for .hg

laerus:
btw maybe we should make this a xontrib? same for git

astronouth7303:
@laerus I mention it in #2151.

gforsyth:
I agree this should be a `xontrib`  but I think that should go in a separate PR. I'm happy with how this is working. If folks can try to time it's behavior vs. master, that's great. If there are no objections otherwise, I think it's ready to go in.

scopatz:
Code looks good to me, @gforsyth! Just waiting on CI. 

As per the xontrib, I disagree with that. I don't see version control as something that is optional and integration with version control tools should be the default. I think people should have to turn it off, rather than on.  Maybe this is my experience teaching undergrads who aren't CS majors, but the better information you can provide out of the gate without being overwhelming, the better.

I think the same goes for the venv/conda env name. This should also be something that is on by default. Many new and novice users won't even have it occur to them that this is something they *could* turn on.

gforsyth:
> As per the xontrib, I disagree with that. I don't see version control as something that is optional and integration with version control tools should be the default. I think people should have to turn it off, rather than on. Maybe this is my experience teaching undergrads who aren't CS majors, but the better information you can provide out of the gate without being overwhelming, the better.
> 
> I think the same goes for the venv/conda env name. This should also be something that is on by default. Many new and novice users won't even have it occur to them that this is something they could turn on.

I agree with all of that -- I think the idea was more along the `xoretrib` thing that @astronouth7303 proposed. Less that it's a xontrib and more that it's just more modular (but load them by default)

scopatz:
Ahh I see! Thanks for the clarification!

scopatz:
Thanks a million @gforsyth!

Siecje:
FWIW before `get_hg_branch` is called `get_git_branch` is called (if git is installed).

And then the same thing happens for the `branch_color`, and then again for `branch_bg_color`.

The hg functions are only called if the git functions are not successful. So git does not have this problem.

PR #2235, opened 2017-02-21
Fix how `$PWD` is managed in order to work with symlinks gracefully

### LINKED ISSUE #1741: Don't resolve symlinks for CWD in prompt
I am not entirely sure if this is a subjective matter, however I think that the path returned by using the `{cwd}` variable in `$PROMPT` shouldn't resolve symbolic links to their real paths. I use symbolic links to abstract away complexities and simplify folder hierarchies and would like this abstraction to be preserved in the prompt.

For example I created symbolic links for each of the common XDG desktop directories in my home-directory (Desktop, Documents, Downloads etc.) so that they point to folders on my HDD instead of my SSD to reduce writes. Currently when changing the directory to such a symbolic link, I would prefer for my prompt to display either: `/home/cryzed/Downloads` or `~/Downloads` instead of fully resolving the path.

I am aware that I can customize the `$FORMATTER_DICT` with my own additions, but I am not entirely sure how to go about this: the code handling the tracking of the CWD and PWD seems rather hard to make sense of to me. It would be awesome if there was a prompt-variable that preserved the symbolic links on the path and maybe optionally one that resolves all symbolic links (the current behaviour).

StreakyCobra:
Add the ``RESOLVE_SYMLINKS`` environment variable to control whether the change directory function should follow symlinks or not.

Close #1741

### DIFF
--- a/xonsh/base_shell.py
+++ b/xonsh/base_shell.py
@@ -372,14 +372,15 @@ class BaseShell(object):
             hist.last_cmd_rtn = hist.last_cmd_out = None
 
     def _fix_cwd(self):
-        """Check if the cwd changed out from under us"""
+        """Check if the cwd changed out from under us."""
+        env = builtins.__xonsh_env__
         cwd = os.getcwd()
-        if cwd != builtins.__xonsh_env__.get('PWD'):
-            old = builtins.__xonsh_env__.get('PWD')  # working directory changed without updating $PWD
-            builtins.__xonsh_env__['PWD'] = cwd      # track it now
+        if 'PWD' in env and os.path.realpath(cwd) != os.path.realpath(env['PWD']):
+            old = env['PWD']  # working directory changed without updating $PWD
+            env['PWD'] = cwd  # track it now
             if old is not None:
-                builtins.__xonsh_env__['OLDPWD'] = old  # and update $OLDPWD like dirstack.
-            events.on_chdir.fire(olddir=old, newdir=cwd)              # fire event after cwd actually changed.
+                env['OLDPWD'] = old  # and update $OLDPWD like dirstack.
+            events.on_chdir.fire(olddir=old, newdir=cwd)  # fire event after cwd actually changed.
 
     def push(self, line):
         """Pushes a line onto the buffer and compiles the code in a way that
--- a/xonsh/dirstack.py
+++ b/xonsh/dirstack.py
@@ -129,11 +129,15 @@ def _get_cwd():
         return None
 
 
-def _change_working_directory(newdir):
+def _change_working_directory(newdir, follow_symlinks=False):
     env = builtins.__xonsh_env__
     old = env['PWD']
     new = os.path.join(old, newdir)
     absnew = os.path.abspath(new)
+
+    if follow_symlinks:
+        absnew = os.path.realpath(absnew)
+
     try:
         os.chdir(absnew)
     except (OSError, FileNotFoundError):
@@ -180,6 +184,11 @@ def cd(args, stdin=None):
     oldpwd = env.get('OLDPWD', None)
     cwd = env['PWD']
 
+    follow_symlinks = False
+    if len(args) > 0 and args[0] == '-P':
+        follow_symlinks = True
+        del args[0]
+
     if len(args) == 0:
         d = os.path.expanduser('~')
     elif len(args) == 1:
@@ -207,7 +216,9 @@ def cd(args, stdin=None):
             else:
                 d = _try_cdpath(d)
     else:
-        return '', 'cd takes 0 or 1 arguments, not {0}\n'.format(len(args)), 1
+        return '', ('cd takes 0 or 1 arguments, not {0}. An additional `-P` '
+                    'flag can be passed in first position to follow symlinks.'
+                    '\n'.format(len(args))), 1
     if not os.path.exists(d):
         return '', 'cd: no such file or directory: {0}\n'.format(d), 1
     if not os.path.isdir(d):
@@ -223,7 +234,7 @@ def cd(args, stdin=None):
         pushd(['-n', '-q', cwd])
         if ON_WINDOWS and _is_unc_path(d):
             d = _unc_map_temp_drive(d)
-    _change_working_directory(d)
+    _change_working_directory(d, follow_symlinks)
     return None, None, 0
 
 
--- a/news/fix_pwd.rst
+++ b/news/fix_pwd.rst
@@ -1,0 +1,16 @@
+**Added:**
+
+* Add a ``-P`` flag to the ``cd`` function in order to change directory and
+  following symlinks.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fix how ``$PWD`` is managed in order to work with symlinks gracefully
+
+**Security:** None


StreakyCobra:
I set `RESOLVE_SYMLINKS` to `True` by default to keep the same behaviour as now. But IMHO it should be set to `False`. I tested with bash and zsh, they are not resolving symlinks by default.

(for the rest I assume that the directory `/tmp/to/dst` is a symlink that points to `/tmp/from/src/`)

I finally decided not to add new variables to the $PROMPT for the following reason: the `{cwd}` variable should always show what is the current directory:

- If the user does not want to resolve symlinks, then it will do `cd /tmp/to/dst` and stay in this directory. `{cwd}` should show `/tmp/to/dst` because the user is in this directory, and doing `cd ..` will put him in `/tmp/to/`. It would be weird to show `/tmp/from/src` and having `cd ..` putting him to `/tmp/to`

- If the user want to resolve symlinks, then it will do `cd /tmp/to/dst` and arrive in `/tmp/from/src`. The later should be displayed as `{cwd}` because if the user does `cd ..`, it will arrive in `/tmp/from`. Showing `/tmp/to/dst` in the prompt and having `cd ..` putting the user in `/tmp/from/` would be weird.

So what this change does is that is let the user decide if symlinks should be resolved or not at `cd` time, and set the `{cwd}` variable accordingly.

melund:
I also vote for `RESOLVE_SYMLINKS=False` being the default :)

xonsh/prompt/cwd.py
```diff
@@ -28,7 +28,13 @@ def _replace_home(x):
 
 
 def _replace_home_cwd():
-    return _replace_home(builtins.__xonsh_env__['PWD'])
+    env = builtins.__xonsh_env__
+    if 'RESOLVE_SYMLINKS' in env and not env['RESOLVE_SYMLINKS'] \
```
melund: I don't think you have to check that `RESOLVE_SYMLINKS` and `LITERAL_PWD` are in the environment.

xonsh/dirstack.py
```diff
@@ -131,7 +131,12 @@ def _get_cwd():
 
 def _change_working_directory(newdir):
     env = builtins.__xonsh_env__
-    old = env['PWD']
+    # If user does not want to resolve symlinks, start from the symlink pwd.
+    if 'RESOLVE_SYMLINKS' in env and not env['RESOLVE_SYMLINKS'] \
+       and 'LITERAL_PWD' in env:
```
melund: You don't need to check that they are present in the `__xonsh_env__`.

melund:
@StreakyCobra  You will need to add `LITERAL_PWD` to the environment when it is constructed. Just like we do with `PWD`.

https://github.com/StreakyCobra/xonsh/blob/57d0412b8f59077329d27a057598de70bda02187/xonsh/environ.py#L1194

That should fix the failing tests. 

Just realized that we may have a problem with `PWD` and now `LITERAL_PWD`. Since they don't have default values in the Environment, they can be deleted completely. That crashes Xonsh. I have created a new issue for this: #2236

xonsh/prompt/cwd.py
StreakyCobra: I was not sure if I could assume that the variables are present in the env without checking them. I'll remove these checks then :-)

xonsh/dirstack.py
StreakyCobra: Same here.

StreakyCobra:
So, after a discussion with @dragon788 on gitter, I search for the bash behaviour concerning `cd` and `symlinks`. bash *doesn't follow symlinks by default*, but provide a [`cd -P` flag](http://unix.stackexchange.com/questions/55713/make-cd-follow-symbolic-links) to do so. The `pwd` command has a [similar `-P` flag](http://stackoverflow.com/questions/7665/how-to-resolve-symbolic-links-in-a-shell-script), returning the working directory with symlink resolved.

To recap, there are these two situations (`/tmp/to/dst` being a symlink to `/tmp/from/src`):

- `cd /tmp/to/dst`
  - `pwd` and `$PWD` return `/tmp/to/dst`
  - `pwd -P` return `/tmp/from/src`
  - The default prompt shows `dst`
- `cd -P /tmp/to/dst`
  - `pwd` and `$PWD` return `/tmp/from/src`
  - `pwd -P` return `/tmp/from/src`
  - The default bash prompt shows `src`

So the current xonsh `cd` behviour is the one of `cd -P`. If we want to mimic bash behaviour, we should:

- Remove the `$RESOLVE_SYMLINK` environment variable (people would be able to enable it globally by setting `cd` as an alias to `cd -P`)
- Make `cd` not follows symlink
- Add a `-P`flag to `cd` in order to force following symlinks when changing directory

But the problem is that the Python `os.chdir` function doesn't allow us to **not** follow symlinks, so we can not do like `bash`: `cd /tmp/to/dst` in bash put us in `/tmp/to/dst`, `os.chdir('/tmp/to/dst)` in Python put us to `/tmp/from/src/` :-(

Any idea how to solve this?

gforsyth:
> If we want to mimic bash behaviour

As a rule, we don't want to mimic bash or zsh just for the sake of mimicking. That doesn't meant that we don't want to mimic it here, just that we should decide on the merits of the behavior itself.

dragon788:
You'd need a pre-os.chdir check using os.path.islink if you are going to implement a follow_symlink=False kwarg or equivalent alias like `cd -P`.

```
os.path.islink(path)
Return True if path refers to a directory entry that is a symbolic link. Always False if symbolic links are not supported by the Python runtime
```

StreakyCobra:
> As a rule, we don't want to mimic bash or zsh just for the sake of mimicking.

I understand. Let say it this way: I don't care actually if it's mimicking bash/zsh or not, it's just it makes more sense from my point of view to not follow symlinks by default :-) When I want to go to `/tmp/to/dst`, I don't want to arrive in `/tmp/from/src/` :smile: 

-----

@dragon788 I know it is possible to check if a path is a symlink or not. But then, assume that a user typed `cd /tmp/to/dst`, and you detected that `/tmp/to/dst` is a symlink, how can you change the directory to `/tmp/to/dst`? Using python `os.chdir('/tmp/to/dst')` will result being in `/tmp/from/src`, with the `pwd` command returnig `/tmp/from/src`. 

It should be possible in a way to really **go** to `/tmp/to/dst`, as shows this bash session with the `pwd` binary and `-L` flag:

```shell
[fabien@asus ~]$ cd /tmp/to/dst
[fabien@asus dst]$ /usr/bin/pwd
/tmp/from/src
[fabien@asus dst]$ /usr/bin/pwd -P
/tmp/from/src
[fabien@asus dst]$ /usr/bin/pwd -L
/tmp/to/dst
[fabien@asus dst]$ echo $PWD
/tmp/to/dst
```

pwd is not part of bash, so if `pwd -L` can return the symlink path, it should be somewhere :smile: Whereas python `os.chdir` does not expose this behavior in xonsh:

```shell
fabien@asus ~ $ import os
fabien@asus ~ $ os.chdir('/tmp/to/dst')
fabien@asus /tmp/from/src $ /usr/bin/pwd
/tmp/from/src
fabien@asus /tmp/from/src $ /usr/bin/pwd -P
/tmp/from/src
fabien@asus /tmp/from/src $ /usr/bin/pwd -L
/tmp/from/src
fabien@asus /tmp/from/src $ echo $PWD
/tmp/from/src
```

Also compare these two examples. First we use python to change the directory, and then we call `/usr/bin/pwd`:

```python
[fabien@asus ` ~ ]$ python
>>> import os
>>> os.chdir('/tmp/to/dst')
>>> os.getcwd()
'/tmp/from/src'
>>> os.system('/usr/bin/pwd')
/tmp/from/src
0
>>> os.system('/usr/bin/pwd -P')
/tmp/from/src
0
>>> os.system('/usr/bin/pwd -L')
/tmp/from/src
0
```

Compare it to when python is started (from bash) already from the `/tmp/to/dst` folder:

```python
[fabien@asus ~]$ cd /tmp/to/dst
[fabien@asus dst]$ python
>>> import os
>>> os.system('/usr/bin/pwd')
/tmp/from/src
0
>>> os.system('/usr/bin/pwd -L')
/tmp/to/dst
0
>>> os.system('/usr/bin/pwd -P')
/tmp/from/src
0
>>> 
```

So the problem is coming from the python `os.chdir` function that does something weird with paths.

astronouth7303:
xonsh makes efforts to recover when `os.chdir()` is used (eg, after the command exits, it checks the current directory and fires `on_chdir` if necessary). So I have the question of what happens to `os.chdir()`?

1. Follow existing policy of not monkeypatching the standard library and just suffer the problems when aliases and modules use it.
2. Attempt to cover `os.chdir()` and `os.getcwd()` to be symlink unaware, potentially breaking Python modules (seems unlikely, but is of course theoretically possible).

xonsh/prompt/cwd.py
scopatz: Hi @StreakyCobra - yeah, our Env class will correctly apply the default if it doesn't exist.

StreakyCobra:
From my point of view it looks like a feature that should be added on CPython side, no? Like adding a keyword argument to `os.chdir`.

    os.chdir(follow_symlinks=True)

That will allow to have the same behaviour as `cd` vs `cd -P`

But I'm not really aware of Python/CPython development process, maybe it doesn't worth it or is too complicated.

scopatz:
So the way that I interpret #1741 is that `cd` should always resolve symlinks. However, what gets displayed in the prompt should be what the user typed in. So `cd()` will need to store what the user typed in somewhere, but it should still resolve when it goes to actually change the directory

astronouth7303:
Time to look at how other shells resolve this?

scopatz:
> it looks like a feature that should be added on CPython side, no?

The problem with this method is that even if CPython added this, it would still be a while before we could drop support for older version of Python that don't have it.

Also thanks a ton for putting this in @StreakyCobra!

StreakyCobra:
> The problem with this method is that even if CPython added this, it would still be a while before we could drop support for older version of Python that don't have it.

Indeed.

So to recap:

- `os.chdir` does not behave like the `cd` function in other shells, but behaves like `cd -P`, i.e. it resolves symlinks before changing the directory. It doesn't seem possible to disable that.
- Because of this, we can not get the current symlink path with `/usr/bin/pwd -L`, the information of the symlink path is lost during the `os.chdir`.
- To correct way to fix this would be on CPython/Python side, but it would take time and would still not work on older Python version that are supported by xonsh.
- Workaround solution would be to mimic this behaviour by manually keeping track of the symlink path in a variable, then using this variable for displaying the symlink path in the prompt, and also using it as base path for `cd` calls.

scopatz:
Yep, I pretty much :)

xonsh/base_shell.py
```diff
@@ -373,13 +373,21 @@ def _append_history(self, tee_out=None, **info):
 
     def _fix_cwd(self):
         """Check if the cwd changed out from under us"""
+        env = builtins.__xonsh_env__
         cwd = os.getcwd()
-        if cwd != builtins.__xonsh_env__.get('PWD'):
-            old = builtins.__xonsh_env__.get('PWD')  # working directory changed without updating $PWD
-            builtins.__xonsh_env__['PWD'] = cwd      # track it now
+        if cwd != env['PWD']:
```
melund: You need to check that PWD exists or use something like `env.get('PWD', None)`. `$PWD` is not in the list of variables that the `Env` class handles. Maybe it should be, but that can be a different PR. 

The same comment applies to all the other places you access `$PWD` below.

melund: Otherwise, you get the error I reported here: $2236

StreakyCobra:
So here we are: http://bugs.python.org/issue29635

There is only one system `chdir` function, and it's following symlinks. Dot.

The `cd -P` builtin of bash is calling `chdir` directly, whereas the `cd` builtin is playing internally with `$PWD`. The `pwd` binary, when called with `-L`, will print the `$PWD` environment variable.

So what we have to do to emulate `cd` and `cd -P`, to print the correct working directory in the prompt, and to make `/usr/bin/pwd -L` to work is to deal correctly with the $PWD variable :smiley:

StreakyCobra:
Proposal updated. Rewritten from scratch, totally different approach, more correct, more elegant, working better.

Actually symlinks support could have been done in one line… The rest is just to support the `-P` flag to `cd`.

```diff
-if cwd != builtins.__xonsh_env__.get('PWD'):
+if os.path.realpath(cwd) != os.path.realpath(builtins.__xonsh_env__.get('PWD')):
```

:man_facepalming:

scopatz:
Great work here @StreakyCobra! Thanks so much for diving in an solving this problem.

PR #2240, opened 2017-02-24
flag nvim as unthreadable

### LINKED ISSUE #2237: Can't use neovim (previously: Alias resolution in threadable prediction is broken again?)
I use neovim (aliased to vim as aliases['vi'] = 'nvim'). Finding a commit where it works without touching `__xonsh_commands_cache__.threadable_predictors` is not an easy task:

* at debian's 0.5.2 it was working OK.
* at 7a44848d4cfbb542deb87fd282f1dbc945959e11 (recent and seemingly relevant commit by @scopatz) it works OK.
* at intermediate commits from [builtins.py history](https://github.com/xonsh/xonsh/commits/master/xonsh/built_ins.py) (e.g. 80d4afdd91797ecff3ffb9f4b63afd30c57ad907) neovim doesn't even start (`termios.error: (5, 'Input/output error')`)
* at 77d30597a6c5f21545f051a0f121de967d2666f0 and later commits it misbehaves badly, all keypresses are output in bold instead of commands being executed. unusable.

Seems like you've introduced those "threadable predictors" kludge, whatever magic it is, then recently they started to support aliases, but got broken nearly immediately after that.

This threadable prediction thing has to be some really worthy and unavoidable dark magic if you want to maintain it... How do other shells live without it?

If my findings are not well described enough, or my bug hunting was too sloppy, please tell me the direction in which to improve it.

mitnk:
fixes #2237

### DIFF
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -353,6 +353,7 @@ def default_threadable_predictors():
         'mvim': predict_help_ver,
         'mutt': predict_help_ver,
         'nano': predict_help_ver,
+        'nvim': predict_false,
         'psql': predict_false,
         'python': predict_shell,
         'python2': predict_shell,
--- a/news/no-thread-nvim.rst
+++ b/news/no-thread-nvim.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+* nvim (Neovim) has been flagged as unthreadable
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None


xonsh/commands_cache.py
```diff
@@ -353,6 +353,7 @@ def default_threadable_predictors():
         'mvim': predict_help_ver,
         'mutt': predict_help_ver,
         'nano': predict_help_ver,
+        'nvim': predict_false,
```
scopatz: I think this should be `predict_help_ver`, like the other vims

scopatz:
Thanks @mitnk! This is basically what I was thinking would be the correct fix.

xonsh/commands_cache.py
mitnk: well, `nvim` here is more like `vim`, not `mvim` etc.

scopatz: Oh yeah ok.

scopatz:
Travis is all backed up. Going to go ahead and merge this

PR #2241, opened 2017-02-24
zsh v5.2 fixes

scopatz:
This is an attempt to address #1134, which fixes some of the foreign shell issues with zsh v5.2.


melund:
I can't review the zsh stuff, but the show and dry-run stuff look good. Great addition. I am just going to merge this anyway. Then we can test in the wild.

scopatz:
Thanks @melund!

PR #2244, opened 2017-02-25
Revised spell miss (overwite -> overwrite)

harukaeru:
It causes an exception when `source-bash foo/bin/activate` to activate python virtual environment in my terminal (Mac OS X 10.11.2).

The content of the exception is here.
```python
AttributeError: 'Namespace' object has no attribute 'overwite_aliases'
```

It seems this value `overwite` is not used in other place (I grepped this), so I changed.

I don't write test for this. but `source-bash` does work well to me.


harukaeru:
Now I've knew you or someones already had implemented `vox` for this. So I'll use this.
However I assume the spell miss is not intended to, so I still open this pull request.

scopatz:
Thanks for fixing this, @harukaeru!

Issue #2033, opened 2016-12-21
xonsh 0.5.1 does not support Python 3.6 f-string syntax


alexis-layton:
xonsh seems to believe that any expression involving an f-string, such as the simple f'hi mom' should be evaluated in a subprocess context. It believes this for `'' + f'hi mom'` as well.

scopatz:
Thanks for reporting this @alexis-layton! Our Python 3.6 support is only partial right now.  Full support is coming soon.  Sorry about the delay.

scopatz:
PS once we have v3.6 support, we'll do a micro release to get it out there for people.

pgoelz:
Can this issue be closed? F-strings work in my setup and commit https://github.com/xonsh/xonsh/commit/936f738516f850c6d710d46ff6b3584c4c27de12 seems to refer to it.

PR #2246, opened 2017-02-25
Wrapping Triple Quotes for Subprocs

### LINKED ISSUE #1152: Command cannot use multiple lines values
```
$ git commit -am 'Fixed issue xxx - <hit RETURN key>
fatal: Paths with -a does not make sense.
$
```

When writing git commit message, we sometime need to type into multiple messages for good. But this is break in xonsh when hit Enter.

scopatz:
This PR contains a number of fixes for wrapping subprocesses that happen to have partially open triple quotes. This is in an attempt to fix #1152.

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -360,6 +360,9 @@ LOGICAL_LINE_CASES = [
 14 \\
 + 2
 """, 1, '14 + 2', 2),
+('''x = """wow
+mom"""
+''', 0, 'x = """wow\nmom"""', 2),
 ]
 
 @pytest.mark.parametrize('src, idx, exp_line, exp_n', LOGICAL_LINE_CASES)
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -24,7 +24,7 @@ from xonsh.tools import (
     is_string_seq, pathsep_to_seq, seq_to_pathsep, is_nonstring_seq_of_strings,
     pathsep_to_upper_seq, seq_to_upper_pathsep, expandvars, is_int_as_str, is_slice_as_str,
     ensure_timestamp, get_portions, is_balanced, subexpr_before_unbalanced,
-    swap_values, get_logical_line, replace_logical_line
+    swap_values, get_logical_line, replace_logical_line, check_quotes,
     )
 from xonsh.environ import Env
 
@@ -208,6 +208,11 @@ def test_subproc_toks_hello_mom_second():
     assert exp == obs
 
 
+def test_subproc_toks_hello_bad_quotes():
+    with pytest.raises(SyntaxError):
+        obs = subproc_toks('echo """hello', lexer=LEXER, returnline=True)
+
+
 def test_subproc_toks_comment():
     exp = None
     obs = subproc_toks('# I am a comment', lexer=LEXER, returnline=True)
@@ -352,6 +357,17 @@ def test_replace_logical_line(src, idx, exp_line, exp_n):
     obs = '\n'.join(lines).replace('\\\n', '').strip()
     assert exp == obs
 
+
+@pytest.mark.parametrize('inp, exp', [
+    ('f(1,10),x.y', True),
+    ('"x"', True),
+    ("'y'", True),
+    ('b"x"', True),
+    ("r'y'", True),
+])
+def test_check_quotes(inp, exp):
+    obs = check_quotes(inp)
+    assert exp is obs
 
 @pytest.mark.parametrize('inp', [
     'f(1,10),x.y',
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -208,9 +208,24 @@ def test_subproc_toks_hello_mom_second():
     assert exp == obs
 
 
-def test_subproc_toks_hello_bad_quotes():
-    with pytest.raises(SyntaxError):
-        obs = subproc_toks('echo """hello', lexer=LEXER, returnline=True)
+def test_subproc_toks_hello_bad_leading_triple_quotes():
+    obs = subproc_toks('echo """hello', lexer=LEXER, returnline=True)
+    assert obs is None
+
+
+def test_subproc_toks_hello_bad_trailing_single_quotes():
+    obs = subproc_toks('echo hello"', lexer=LEXER, returnline=True)
+    assert obs is None
+
+
+def test_subproc_toks_hello_bad_leading_triple_quotes():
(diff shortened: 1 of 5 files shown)


xonsh/tools.py
```diff
@@ -346,6 +348,44 @@ def subproc_toks(line, mincol=-1, maxcol=None, lexer=None, returnline=False):
...
+        m = RE_COMPLETE_STRING.match(s)
+        ok = m is not None
+    return ok
+
+
+def _have_open_triple_quotes(s):
+    if s.count('"""') % 2 == 1:
+        open_triple = '"""'
+    elif s.count("'''") % 2 == 1:
+        open_triple = "'''"
```
melund: Maybe I misunderstand the use of this function. I am wondering if there is a logic issue here ... If you pass the a string with this content: `''' I like """ quotes '''` 

Would it return `"""` even thus the internal quotes should not count.

Or maybe such a string is not possible with Python and the inner quotes must be escaped

mitnk:
seem a step ahead, but does not work yet:
```
$ git commit -am 'fix issue #123 -                                          [62027]
....................
.................... this is a good one'
....................
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/ptk/shell.py", line 137, in _push
    locs=None)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 19242, in compile
    transform=transform)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 19212, in parse
    tree, input = self._parse_ctx_free(input, mode=mode, filename=filename)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 19349, in _parse_ctx_free
    raise original_error
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 19297, in _parse_ctx_free
    debug_level=(self.debug_level > 2))
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/parsers/base.py", line 336, in parse
    tree = self.parser.parse(input=s, lexer=self.lexer, debug=debug_level)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/ply/ply/yacc.py", line 331, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/ply/ply/yacc.py", line 1199, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/ply/ply/yacc.py", line 193, in call_errorfunc
    r = errorfunc(token)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/parsers/base.py", line 2695, in p_error
    column=p.lexpos))
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/parsers/base.py", line 467, in _parse_error
    raise err
  File "<string>", line None
SyntaxError: <xonsh-code>:1:4: ('code: commit',)
git commit -am 'fix issue #123 -
    ^
```

scopatz:
This isn't meant for single quotes, which I didn't implement because that wouldn't follow the python convention for single quoted string. It is a feature I suppose we could add to suprocesses, if people wanted but I would err on the side of not having different string conventions for different mode.

mitnk:
Ok, `git commit -am '''` works though, but I guess that may not close issue #1152. Since it is a issue user can encountered at the first try in xonsh and would hope single quote with multiple line supported in xonsh. unless we define this issue as an WONT-FIX :)

xonsh/tools.py
```diff
@@ -1526,7 +1571,7 @@ def format_std_prepost(template, env=None):
     return s
 
 
-_RE_STRING_START = "[bBrRuU]*"
+_RE_STRING_START = "[bBprRuU]*"
```
melund: I assume p is path strings.What about f strings.

melund: Of course. They are proably not allowed for subproc..

melund: Of course. They are probably not allowed for subproc..

tests/test_tools.py
```diff
@@ -208,6 +208,33 @@ def test_subproc_toks_hello_mom_second():
...
+    obs = subproc_toks('echo """hello', lexer=LEXER, returnline=True)
+    assert obs is None
+
+
+def test_subproc_toks_hello_bad_trailing_single_quotes():
+    obs = subproc_toks('echo hello"', lexer=LEXER, returnline=True)
+    assert obs is None
+
+
+def test_subproc_toks_hello_bad_leading_triple_quotes():
```
scopatz: Nice catch!

xonsh/tools.py
scopatz: Well, we don't actually support f-strings yet at all. See #2033

xonsh/tools.py
scopatz: So the `''' I like """ quotes '''` case should not come up because that is a valid Python string. `get_logical_lines()` is only called when there is a SyntaxError and so it won't trigger. This really only comes up when someone has typed in a partial string, like `''' I like """ qu`, and then needs a continuation line. 

In the worst case, `get_logical_lines()` will treat the rest of the input as a single line, eating up the rest of the input the user wrote at the command line. However, `subroc_toks()` then re-tokeninzes the line and will correctly see the complete string and newlines.

scopatz:
> Unless we define this issue as an WONT-FIX :)

@mitnk, I think we could move #1152 to a feature request. Personally, I am not interested in implementing it. Also, as discussed above, I think there are good reasons for marking it as WONTFIX since Python strings and subproc strings would diverge. I think we'd need a broader conversation about it.

scopatz:
Note I updated this from master to fix the merge conflicts

xonsh/tools.py
melund: Ahh. Now I get it :)

melund:
This looks good to go.

scopatz:
Thanks @melund!

Issue #1174, opened 2016-06-08
Can't start process in background and redirect its streams


cryzed:
I am trying to run the following script:

``` python
#!/usr/bin/env xonsh

import os
import time


def main():
    killall plasmashell 2>&1 > /dev/null
    nohup plasmashell 2>&1 & > /dev/null

    while not os.path.exists('nohup.out'):
        time.sleep(0.1)

    os.remove('nohup.out')


if __name__ == '__main__':
    main()
```

However, I get the following traceback when executing it:

``` python-traceback
cryzed@arch ~ $ restart-plasmashell
Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/xonsh/execer.py", line 147, in _parse_ctx_free
    debug_level=(self.debug_level > 1))
  File "/usr/lib/python3.5/site-packages/xonsh/parsers/base.py", line 289, in parse
    tree = self.parser.parse(input=s, lexer=self.lexer, debug=debug_level)
  File "/usr/lib/python3.5/site-packages/ply/yacc.py", line 331, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/usr/lib/python3.5/site-packages/ply/yacc.py", line 1181, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/usr/lib/python3.5/site-packages/ply/yacc.py", line 193, in call_errorfunc
    r = errorfunc(token)
  File "/usr/lib/python3.5/site-packages/xonsh/parsers/base.py", line 2326, in p_error
    self.xonsh_code)
  File "/usr/lib/python3.5/site-packages/xonsh/parsers/base.py", line 390, in _parse_error
    raise err
SyntaxError: /home/cryzed/bin/restart-plasmashell:9:31: ('code: >',)
    ![nohup plasmashell 2>&1 & > /dev/null]
                               ^

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/xonsh", line 9, in <module>
    load_entry_point('xonsh==0.3.3', 'console_scripts', 'xonsh')()
  File "/usr/lib/python3.5/site-packages/xonsh/main.py", line 247, in main
    run_script_with_cache(args.file, shell.execer, glb=shell.ctx, loc=None, mode='exec')
  File "/usr/lib/python3.5/site-packages/xonsh/codecache.py", line 135, in run_script_with_cache
[...]

PR #2247, opened 2017-02-25
print jobs when interactive

scopatz:
This PR only prints job info when in interactive mode when preforming backgrounding and foregrounding tasks.  The `jobs` command will always display, of course. This addresses part of what is in #1174.


melund:
Looks good. Thanks for putting it in @scopatz.

Issue #1234, opened 2016-06-15
del with parens parses Load instead of Del


scopatz:
So there is a minor bug here:

```
scopatz@artemis ~/xonsh amal $ x, y = 1, 2
scopatz@artemis ~/xonsh amal $ del (x, y)
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/ptk/shell.py", line 101, in push
    locs=None)
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/execer.py", line 108, in compile
    code = compile(tree, filename, mode)
ValueError: expression must have Del context but has Load instead
```

scopatz:
This is valid Python

scopatz:
Potential fix in #2248

PR #2248, opened 2017-02-25
Fixed del (x, y)

scopatz:
Should address #1234, but not #5678


melund:
Looks good, and works for me

Issue #1262, opened 2016-06-17
Output redirection doesn't work in aliases


Cheaterman:
Hello buddies, I hope you're all doing good ^__^

Apparently output redirection doesn't work in aliases:

```
cheaterman@cmans-laptop ~ $ aliases['bli'] = 'echo bli &> /dev/null'
cheaterman@cmans-laptop ~ $ bli
bli &> /dev/null
```

Thanks in advance for investigating/fixing ^__^

BlaXpirit:
It's not specific to output redirection, it's just defined to space-separate and run the command with the exact arguments as written

scopatz:
Thanks for reporting @Cheaterman!

Cheaterman:
Thanks to you for acknowledging @scopatz ^__^

scopatz:
Potential fix for this in #2250

gforsyth:
Closed via #2250

PR #2250, opened 2017-02-25
Enable alias redirections

scopatz:
This should address #1262. Basically a two-line change.


gforsyth:
Sweet! Thanks, @scopatz !

Issue #1481, opened 2016-07-25
Really, really bizarre issue with xontribs that use signal


kirbyfan64:
This is really weird.

To reproduce:
- Create a new `xontrib` directory.
- Create a file `xontrib/bug.py` containing a single line: `import xonsh.__main__`.
- Create `setup.py` containing:
  
  ```
  from distutils.core import setup
  
  setup(
  name='xontrib-bug',
  packages=['xontrib'],
  package_dir={'xontrib': 'xontrib'},
  )
  ```
- Run `python setup.py install`.

Now, if you run `xontrib load bug` twice, _this_ happens:

``` python
ryan@DevPC-LX /media/ryan/stuff/xonsh/xontrib-bug master $ xontrib load bug
ValueError: signal only works in main thread
ryan@DevPC-LX ~ master $ pwd # Why does the prompt say just `~`?
/media/ryan/stuff/xonsh/xontrib-bug
ryan@DevPC-LX ~ master $ xontrib load bug # This will crash Xonsh.
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK = True
AttributeError: 'module' object has no attribute '__xonsh_config__'
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK = True
AttributeError: 'module' object has no attribute '__xonsh_history__'
Traceback (most recent call last):
  File "/media/ryan/stuff/anaconda/bin/xonsh", line 9, in <module>
    load_entry_point('xonsh==0.3.0', 'console_scripts', 'xonsh')()
  File "/home/ryan/stuff/anaconda/lib/python3.4/site-packages/xonsh/main.py", line 254, in main
    shell.cmdloop()
  File "/home/ryan/stuff/anaconda/lib/python3.4/site-packages/xonsh/ptk/shell.py", line 120, in cmdloop
    self.default(line)
  File "/home/ryan/stuff/anaconda/lib/python3.4/site-packages/xonsh/base_shell.py", line 168, in default
    self._append_history(inp=src, ts=[ts0, ts1], tee_out=tee.getvalue())
  File "/home/ryan/stuff/anaconda/lib/python3.4/site-packages/xonsh/base_shell.py", line 258, in _append_history
    hist = builtins.__xonsh_history__  # pylint: disable=no-member
AttributeError: 'module' object has no attribute '__xonsh_history__'
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
[...]

PR #2257, opened 2017-02-25
Unthreadable xontrib command

scopatz:
This enables signals to be used inside of xontribs on load. Should address #1481


melund:
Looks good.

Issue #2253, opened 2017-02-25
echo @('a', 'b')


t184256:
As an end-user, I want `echo @('a', 'b')` to work.

Currently only `echo @(('a', 'b'))` does, but I find it unwieldy. `'a', 'b',` works, so why not `echo @('a', 'b')`?

scopatz:
This should now be available on master, please feel free to reopen if needed!

Issue #2209, opened 2017-02-18
Add generators to @(...) evaluation 


scopatz:
See https://github.com/xonsh/xonsh/issues/1217#issuecomment-228021152

t184256:
While you are pondering at what can go inside `@(...)`, please consider making `echo @('a', 'b')` work.

scopatz:
@t184256 please feel free to open another issue for that!

t184256:
OK, requested in #2253.

scopatz:
This should now be available on master.  Please feel free to reopen if needed, and thank @adqm :)

PR #2259, opened 2017-02-25
allow generators and bare tuples in @(...)

adqm:
I believe this addresses the issues in #2209 and #2253, or can at least serve as a starting point for continuing work.  Quick port of changes made on my fork.

```
user@machine $ echo @('hi')
hi
user@machine $ echo @(7)
7
user@machine $ echo @([7,8])
7 8
user@machine $ echo @(i**2 for i in range(7))
0 1 4 9 16 25 36
```

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -559,8 +559,10 @@ class SubprocSpec:
         # modifications that do not alter cmds may come before creating instance
         spec = kls(cmd, cls=cls, **kwargs)
         # modifications that alter cmds must come after creating instance
+        # perform initial redirects
         spec.redirect_leading()
         spec.redirect_trailing()
+        # apply aliases
         spec.resolve_alias()
         spec.resolve_binary_loc()
         spec.resolve_auto_cd()
@@ -623,13 +625,16 @@ class SubprocSpec:
     def resolve_executable_commands(self):
         """Resolve command executables, if applicable."""
         alias = self.alias
-        if callable(alias):
+        if alias is None:
+            pass
+        elif callable(alias):
             self.cmd.pop(0)
             return
-        elif alias is None:
-            pass
         else:
             self.cmd = alias + self.cmd[1:]
+            # resolve any redirects the aliases may have applied
+            self.redirect_leading()
+            self.redirect_trailing()
         if self.binary_loc is None:
             return
         try:
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -890,10 +890,11 @@ def list_of_strs_or_callables(x):
     """Ensures that x is a list of strings or functions"""
     if isinstance(x, str) or callable(x):
         rtn = [x]
-    elif isinstance(x, cabc.Sequence):
-        rtn = [i if isinstance(i, str) or callable(i) else str(i) for i in x]
     else:
-        rtn = [str(x)]
+        try:
+            rtn = [i if isinstance(i, str) or callable(i) else str(i) for i in x]
+        except:
+            rtn = [str(x)]
     return rtn
 
 
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -890,11 +890,10 @@ def list_of_strs_or_callables(x):
     """Ensures that x is a list of strings or functions"""
     if isinstance(x, str) or callable(x):
         rtn = [x]
+    elif isinstance(x, cabc.Iterable):
+        rtn = [i if isinstance(i, str) or callable(i) else str(i) for i in x]
     else:
-        try:
-            rtn = [i if isinstance(i, str) or callable(i) else str(i) for i in x]
-        except:
-            rtn = [str(x)]
+        rtn = [str(x)]
     return rtn
 
 
--- a/news/generator_at.rst
+++ b/news/generator_at.rst
@@ -1,0 +1,13 @@
+**Added:** 
+
+* ``@(...)`` syntax now supports generators and tuples without parentheses.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/news/rediral.rst
+++ b/news/rediral.rst
@@ -1,0 +1,14 @@
+**Added:**
+
+* Redirections may now be used in string and list-of-strings
+  aliases.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -1548,6 +1548,12 @@ def test_ls_dot():
 def test_lambda_in_atparens():
     check_xonsh_ast({}, '$(echo hello | @(lambda a, s=None: "hey!") foo bar baz)', False)
 
+def test_generator_in_atparens():
+    check_xonsh_ast({}, '$(echo @(i**2 for i in range(20)))', False)
+
+def test_bare_tuple_in_atparens():
+    check_xonsh_ast({}, '$(echo @("a", 7))', False)
+
 def test_nested_madness():
     check_xonsh_ast({}, '$(@$(which echo) ls | @(lambda a, s=None: $(@(s.strip()) @(a[1]))) foo -la baz)', False)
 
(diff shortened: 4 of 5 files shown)


scopatz:
Thanks for putting this in @adqm! Can you please add a news entry?

xonsh/built_ins.py
```diff
@@ -885,10 +885,11 @@ def list_of_strs_or_callables(x):
     """Ensures that x is a list of strings or functions"""
     if isinstance(x, str) or callable(x):
         rtn = [x]
-    elif isinstance(x, cabc.Sequence):
```
scopatz: I think you can just change this to be `cabc.Iterable` from Sequence,

xonsh/built_ins.py
```diff
@@ -885,10 +885,11 @@ def list_of_strs_or_callables(x):
...
-        rtn = [i if isinstance(i, str) or callable(i) else str(i) for i in x]
     else:
-        rtn = [str(x)]
+        try:
+            rtn = [i if isinstance(i, str) or callable(i) else str(i) for i in x]
+        except:
```
scopatz: I'd really prefer not to use try-except blocks for things that are deterministically knowable.  Also, we don't allow bare excepts, so at the very least, this should be `except Exception`.

scopatz:
Just those couple of quick comments!

adqm:
Okay, updated and added a news entry.  Should be ready for another look, I think.

scopatz:
Thanks @adqm! Looks good to me implementation-wise. It would be nice to have parser tests for these, but I suppose those can be added later.

adqm:
I can throw in some tests.

scopatz:
Awesome! Thanks. I really appreciate it!

adqm:
I just added a couple of quick tests.  Not sure if they're sufficient, but one checks a generator inside `@()`, and another checks a tuple with no parens in `@()`.  Happy to add more if that's preferred.

<p>&nbsp;</p>

A note: I didn't have the patience to wait for the tests on my machine.  xonsh startup seems to be really slow on master, so the tests are really taking forever.

My fork from 0.3.4 starts up in ~.25 seconds on my machine, whereas xonsh takes ~2.25 seconds.  Is that normal?  I'm happy to help track down the slowness if I can, but I'm not sure where to look.  I would have guessed `prompt_toolkit` to be the culprit, but that shouldn't be loaded in the case of running a command with `-c`, and it seems about as slow using `readline`.

```
hartz@desky-iii:~/xonsh$ time xonsh -c exit


real	0m2.377s
user	0m2.372s
sys	0m0.016s
```
```
hartz@desky-iii:~/xonsh$ time tako -c exit


real	0m0.260s
user	0m0.256s
sys	0m0.000s
hartz@desky-iii:~/xonsh$ 
```

scopatz:
> I just added a couple of quick tests. Not sure if they're sufficient, but one checks a generator inside @(), and another checks a tuple with no parens in @(). Happy to add more if that's preferred.

This is great, thanks!

> My fork from 0.3.4 starts up in ~.25 seconds on my machine, whereas xonsh takes ~2.25 seconds.

This seems potentially related to #1521. You still use arch right?  I have not been able to reproduce this personally. My average runtime from `xonsh -c exit` is 9.36 ms on master (a4de707).  I don't really know what is going on here but I would really love it if you (or anyone) could hunt down what the issue is.

@melud recently added the a `--timings` flag that helps profile startup timings.

adqm:
I've always been Debian stable and still am :)  but I am running Python 3.5.2 on this machine (Python 3.6 on some others).

Interestingly, the `--timings` thing seems to put the blame on three different hooks on three different runs:

```
 Debug level: Off
|------------------|-----------|-----------|
|Event name        | Time (s)  | Delta (s) |
|------------------|-----------|-----------|
|start             |   0.000   |   0.000   |
|pre_execer_init   |   0.024   |   0.024   |
|post_execer_init  |   0.120   |   0.096   |
|on_pre_rc         |   0.124   |   0.004   |
|on_post_rc        |   0.132   |   0.008   |
|on_pre_cmdloop    |   2.436   |   2.304   |
|on_ptk_create     |   2.436   |   0.000   |
|on_post_init      |   2.436   |   0.000   |
|on_pre_prompt     |   2.540   |   0.104   |
|------------------|-----------|-----------|
```
```
|------------------|-----------|-----------|
|Event name        | Time (s)  | Delta (s) |
|------------------|-----------|-----------|
|start             |   0.000   |   0.000   |
|pre_execer_init   |   0.024   |   0.024   |
|post_execer_init  |   0.120   |   0.096   |
|on_pre_rc         |   0.124   |   0.004   |
|on_post_rc        |   0.132   |   0.008   |
|on_ptk_create     |   2.420   |   2.288   |
|on_pre_cmdloop    |   2.420   |   0.000   |
|on_post_init      |   2.420   |   0.000   |
|on_pre_prompt     |   2.524   |   0.104   |
|------------------|-----------|-----------|
```
```
|------------------|-----------|-----------|
|Event name        | Time (s)  | Delta (s) |
|------------------|-----------|-----------|
|start             |   0.000   |   0.000   |
|pre_execer_init   |   0.024   |   0.024   |
|post_execer_init  |   0.200   |   0.176   |
|on_pre_rc         |   0.208   |   0.008   |
|on_post_rc        |   0.208   |   0.000   |
|on_post_init      |   2.424   |   2.216   |
|on_pre_cmdloop    |   2.424   |   0.000   |
|on_ptk_create     |   2.424   |   0.000   |
|on_pre_prompt     |   2.528   |   0.104   |
|------------------|-----------|-----------|
```

Note that this is 2.3 seconds on a pretty beefy machine.  So I'm not entirely sure where it could be eating time.

The fact that it is putting the blame on different things makes me think it's something else, and not the hooks themselves.  Maybe loading history or something like that?  Is there anything else happening async at startup?

Also checked with both Python 3.6.0 and 3.5.2, seemingly no difference.

melund:
@adqm Maybe you misunderstand how the timings work. 

It uses the existing event system to save some `perf_counter()` values during startup. Thus, it gives you time values for different stages of the startup sequence. 

Here are my timings on Windows (normal medium laptop). 

|Event name        | Time (s)  | Delta (s) |
|------------------|-----------|-----------|
|start             |   0.000   |   0.000   |
|pre_execer_init   |   0.041   |   0.041   |
|post_execer_init  |   0.170   |   0.129   |
|on_pre_rc         |   0.171   |   0.001   |
|on_post_rc        |   0.293   |   0.122   |
|on_ptk_create     |   0.445   |   0.153   |
|on_post_init      |   0.446   |   0.001   |
|on_pre_cmdloop    |   0.446   |   0.000   |
|on_pre_prompt     |   0.803   |   0.357   |

It is a bit slower than what I experienced on Python 3.5.

In your case `on_ptk_create`, `on_post_init` and `on_pre_cmdloop` happen so close that the system can't even figure out in which order they occur. That weird. I thought `perf_counter()` was accurate enough.

You timing results indicate that the main part of the startup time is spent between `on_post_rc` and `on_ptk_create`. It is most likely something related to PTK. 

I also added a new event to make custom timing points. For example. Here I fire some events before and after the [history construction](https://github.com/xonsh/xonsh/blob/master/xonsh/shell.py#L115)

``` python
events.on_timingprobe.fire(name='before_history')
builtins.__xonsh_history__ = hist = xhm.construct_history(
     env=env.detype(), ts=[time.time(), None], locked=True)
 events.on_timingprobe.fire(name='after_history')
```

This gives me these timings which show that this part of the code takes around 6 ms...

|Event name        | Time (s)  | Delta (s) |
|------------------|-----------|-----------|
|start             |   0.000   |   0.000   |
|pre_execer_init   |   0.036   |   0.036   |
|post_execer_init  |   0.165   |   0.128   |
|on_pre_rc         |   0.166   |   0.001   |
|on_post_rc        |   0.305   |   0.140   |
|before_history    |   0.305   |   0.000   |
|after_history     |   0.311   |   **0.006**   |
|on_ptk_create     |   0.459   |   0.147   |
|on_post_init      |   0.459   |   0.000   |
|on_pre_cmdloop    |   0.459   |   0.000   |
|on_pre_prompt     |   0.781   |   0.321   |

scopatz:
@adqm, could you try to use profimp to hunt down if a particular import is taking most of this 2 seconds? That would help us focus this down further.

Issue #1628, opened 2016-08-25
Problems with redirections


rbignon:
Here are some issues with redirections:

<pre>
$ ls > /dev/null
$ ls >/dev/null
................................. 
</pre>

It may not be needed to have a space between the '>' and the file path.

Also, redirection from stderr to stdout works:

<pre>
$ ls uie > /dev/null
/bin/ls: cannot access 'uie': No such file or directory
$ ls uie 2>&1 > /dev/null
$
</pre>

But not from stdout to stderr:

<pre>
 $ ls uie 1>&2 > /dev/null
................................. 
</pre>

I think this is related to #1536, and the problem with vim's syntastic plugin which tells at startup:

<pre>
syntastic: error: your shell /usr/local/bin/xonsh doesn't use traditional UNIX syntax for redirections
</pre>

scopatz:
Thanks for reporting.  Yes, not all redirections have been implemented yet.

scopatz:
Potential fix for the remaining issue in #2261

melund:
@rbignon This is implemented on master now. Reopen if you think something is missing.

PR #2261, opened 2017-02-26
Redirection no longer require a space

scopatz:
This expands the capability for redirections to not require a space. This should address the remaining issue in #1628

### DIFF
--- a/news/nsredir.rst
+++ b/news/nsredir.rst
@@ -1,0 +1,15 @@
+**Added:**
+
+* Subprocess redirection may now forego the whitespace between the
+  redirection and a file name.  For example,
+  ``echo hello world >/dev/null``.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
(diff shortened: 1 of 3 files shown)


laerus:
so something like `echo ok err>out` will redirect stderr to stdout or stderr to a filename 'out'? what if i want to redirect to a filename 'err' or the filename '1' will these be special cases?

scopatz:
Hi @laerus, so in ambiguous cases, the redirection token always wins (because it is tokenized first). Thus `err>out` will redirect to standard out. It will not create a file called out. I should note that I don't think that we should advertise this too strong because of these ambiguous cases. I put this in only so that programs like vim and emacs would be able to use xonsh without too much extra effort.  It inches us towards more sh-lang comparability, just so that we can interoperate with other tools.

melund:
This looks fine to me. You are really killing issues these days @scopatz

scopatz:
Ummhmm! I want to polish everything up and step one is to fix all of the bugs. Thanks for the reviews!

Issue #1649, opened 2016-08-29
Invalidate Code Caches on Install


scopatz:
I feel like the code cace should be invalidated whenever `__init__.py` or similar is changed.  This could prevent some non-obvious issues.

laerus:
totally agree, i has some issues before that were solved by removing the caches.

laerus:
i'm also having a hard time in develop mode when making changes to the parser, restarting isn't enough, i need to delete `parser_tables.py` to make it work. Should i open a new issue about this?

astronouth7303:
Does `parser_tables.py` need to be written? Or just a `.pyc` cache?

laerus:
should we just remove `xonsh_code_cache` and `parser_tables.py` on each install?

gforsyth:
I think that's a reasonable and simple solution.

laerus:
so `parser_tables.py` is already getting removed in `setup.py` my bad,
the hard part would be to find out if and where $XONSH_DATA_DIR is in `setup.py`

scopatz:
Yeah, I am not sure that we should remove the code cache when we install, because we can't be sure which one the user will use at run time.  This is probably going to have to be a run time check.  You can imagine different xonsh version installs which point to different caches, potentially.

scopatz:
Potential fix for this in #2262

PR #2262, opened 2017-02-26
Invalidate code cache whenever the xonsh version changes

scopatz:
This should address #1649

### DIFF
--- a/news/ic.rst
+++ b/news/ic.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:**
+
+* Xonsh's script and code caches will are now invalidated whenever the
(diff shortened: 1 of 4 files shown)


adqm:
Should this also happen is the Python version changes (`sys.version_info`), just in case?  (If I'm not mistaken, for example, 3.5.2 has a different magic number than 3.5.0-1, despite having the same cache filename, even in `__pycache__`s)

scopatz:
@adqm - ahh yeah that is a good point. I'll add that.

scopatz:
Done!

melund:
Looks good.

PR #2265, opened 2017-02-26
Interactive mode with command or script

laerus:
this patch adds functionality similar to the python repl
when forcing interactive mode with command or script :
`xonsh -ic '...'`
`xonsh -i script.xsh`

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -43,6 +43,7 @@ def xonsh_events():
 @pytest.yield_fixture
 def xonsh_builtins(xonsh_events):
     """Mock out most of the builtins xonsh attributes."""
+    old_builtins = set(dir(builtins))
     builtins.__xonsh_env__ = DummyEnv()
     if ON_WINDOWS:
         builtins.__xonsh_env__['PATHEXT'] = ['.EXE', '.BAT', '.CMD']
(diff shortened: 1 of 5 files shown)


xonsh/main.py
```diff
@@ -386,6 +372,20 @@ def func_sig_ttin_ttou(n, f):
             code = sys.stdin.read()
             run_code_with_cache(code, shell.execer, glb=shell.ctx, loc=None,
                                 mode='exec')
+        if args.force_interactive:
```
scopatz: Shouldn't this be an `elif`?

scopatz: Also, why move away from `args.mode`?  Shouldn't args.force_interactive just set `args.mode` to be interactive in premain?

scopatz:
Thanks for putting this in @laerus! Just a couple of comments in-line. Also, it looks like there are some test failures.

xonsh/main.py
laerus: we want xonsh to run the command or script before going into interactive,
modes are exclusive, we need new modes that correspond to 'interactive with script' and 'interactive with command'. With command/script argument present premain just sets mode accordingly so we can't check against mode for interactive. I was checking both against `args.force_interactive` and `args.mode` but it seemed redundant so i removed it.

tests/conftest.py
```diff
@@ -72,40 +73,8 @@ def xonsh_builtins(xonsh_events):
...
-    if hasattr(builtins, '__xonsh_history__'):
-        del builtins.__xonsh_history__
-    del builtins.__xonsh_enter_macro__
-    del builtins.evalx
-    del builtins.execx
-    del builtins.compilex
-    del builtins.aliases
-    del builtins.events
+    for attr in set(dir(builtins)) - old_builtins:
+        delattr(builtins, attr)
```
scopatz: Nice fix :)

xonsh/main.py
scopatz: This seems like you could end up running a script from a file and *then* entering a command loop. That shouldn't happen, even if it is in interactive mode. The command loop should be entered only if we are in interactive mode and we are not running a script  or a command (`-c`).

scopatz: Yeah, I just tried it.

```console
scopatz@artemis:~$ xonsh -i -c "echo hi"   # in bash
hi
scopatz@artemis ~ $  exit # in xonsh 
scopatz@artemis:~$  # back in bash
```

We should never enter into xonsh fully. Take the inverse, for example,

```console
scopatz@artemis ~/xonsh/news nullsub $ bash -i -c "echo hello"  # in xonsh
bash: cannot set terminal process group (-1): Inappropriate ioctl for device
bash: no job control in this shell
hello
scopatz@artemis ~/xonsh/news interactive $ # still in xonsh
```

Does this make sense?

scopatz:
Hi @laerus, I am seeing the following issue when I try to launch from bash:

```console
scopatz@artemis:~$ XONSH_DEBUG=1 xonsh
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/main.py", line 337, in main
    return main_xonsh(args)
  File "/home/scopatz/.local/lib/python3.5/site-packages/xonsh/main.py", line 354, in main_xonsh
    if args.mode == XonshMode.single_command:
AttributeError: 'Namespace' object has no attribute 'mode'
Xonsh encountered an issue during launch
Failback to /bin/sh
```

xonsh/main.py
laerus: yeah, that is how the python interactive option works, you can run a script and then enter the REPL in the same context. It seems like in bash it doesnt make any difference if you use the `-i` flag. Isnt this feature something we want in xonsh?

scopatz: I don't think so. I think we want to match the behaviour of bash and other shells. Text editors and other tools rely on the these command line options to work in a certain way and we want to have comparability with those tools. Changing `-i` from non-blocking to blocking would probably break a bunch of these things.

scopatz:
Hi @laerus I am still in favor of the test refactor here, if you are interested in cleaning that up.

laerus:
Hey @scopatz, this started out as a general xonsh.main and tests refactor when i noticed the difference in interactive mode with the python REPL but you're right about sticking with what most programs expect from a shell (darn you bash). I will open another PR with these!

melund:
Couldn't you just start in interactive mode and then execute the lines in the script. Sort'a like a pasting everything into an interactive session.

laerus:
yeah or `source` the script for more convenience,  i am used to python's `-i` flag when experimenting and i thought bash did the same but apparently it sets up interactive related things except entering the repl.

scopatz:
Maybe we could use `-I` for this instead?

laerus:
i'm okay with the current functionality, i think it can wait until more ppl need this feature.

Issue #1808, opened 2016-10-06
Setting values for BASH_COMPLETIONS in xonfig wizard leads to TypeError, "not JSON serializable"


gotgenes:
```
$BASH_COMPLETIONS
This is a list (or tuple) of strings that specifies where the
``bash_completion`` script may be found. For better performance,
bash-completion v2.x is recommended since it lazy-loads individual
completion scripts. Paths or directories of individual completion
scripts (like ``.../completes/ssh``) do not need to be included here.
The default values are platform dependent, but sane. To specify an
alternate list, do so in the run control file.
default value:
Normally this is:

    ``('/etc/bash_completion', )``

But, on Mac it is:

    ``('/usr/local/etc/bash_completion', )``

And on Arch Linux it is:

    ``('/usr/share/bash-completion/bash_completion', )``

Other OS-specific defaults may be added in the future.
current value: ['/usr/local/share/bash-completion/bash_completion', '/usr/local/etc/bash_completion']
>>> /usr/local/etc/bash_completion
Converted input '/usr/local/etc/bash_completion' to ['/usr/local/etc/bash_completion'].

[…]

'`-.,_,.-*'`-.,_,.-*'`-.,_,.-*'`-.,_,.-*'`-.,_,.-*'`-.,_,.-*'`-.,_,.-*'

xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK = True
TypeError: ['/usr/local/etc/bash_completion'] is not JSON serializable
```

I have also entered the following values; I end up with the same error:
- `('/usr/local/etc/bash_completion', )`
- `['/usr/local/etc/bash_completion']`

```
clasher@clasher-MBP ~/development $ xonfig
+------------------+-----------------+
| xonsh            | 0.4.7.dev175    |
| Git SHA          | 9e67c21         |
| Python           | 3.5.2           |
| PLY              | 3.8             |
| have readline    | True            |
| prompt toolkit   | 1.0.7           |
| shell type       | prompt_toolkit  |
| pygments         | 2.1.3           |
| on posix         | True            |
| on linux         | False           |
| on darwin        | True            |
| on windows       | False           |
| on cygwin        | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
+------------------+-----------------+
```

laerus:
hey @gotgenes, thanks for reporting! This seems like a legitimate bug in the wizard.

[...]

PR #2270, opened 2017-02-26
New JSON utils module

scopatz:
This should address the issue in #1808


melund:
Nice. Thanks @scopatz.

Issue #1871, opened 2016-10-19
File glob appears to work incorrectly with and


Wideshanks:
I want to use `and` with two commands and use file globs in both, but ~~it appears to be re-using the same glob in the second command~~ it's not getting both globs correctly. What am I missing here?

``` bash
$ ls -l
Actually_test.txt
Actually.txt
Complete_test.txt
Complete.txt

$ echo *.txt and echo *_test.txt # expect second command to give only two files
Actually.txt Actually_test.txt Complete.txt Complete_test.txt
Actually.txt Actually_test.txt Complete.txt Complete_test.txt

$ echo *_test.txt # these are the two files I wanted
Actually_test.txt Complete_test.txt

$ echo *_test.txt and echo *.txt # same result if reversed
Actually.txt Actually_test.txt Complete.txt Complete_test.txt
Actually.txt Actually_test.txt Complete.txt Complete_test.txt

$ bash
$ echo *.txt && echo *_test.txt # bash gives me the files I wanted in the second command
Actually_test.txt Actually.txt Complete_test.txt Complete.txt
Actually_test.txt Complete_test.txt

$ echo *_test.txt && echo *.txt
Actually_test.txt Complete_test.txt
Actually_test.txt Actually.txt Complete_test.txt Complete.txt

$ exit
$ xonfig
+------------------+-----------------+
| xonsh            | 0.4.5           |
| Git SHA          | None            |
| Python           | 3.4.0           |
| PLY              | 3.8             |
| have readline    | True            |
| prompt toolkit   | 1.0.7           |
| shell type       | prompt_toolkit  |
| pygments         | 2.0.2           |
| on posix         | True            |
| on linux         | True            |
| distro           | Ubuntu          |
| on darwin        | False           |
| on windows       | False           |
| on cygwin        | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
+------------------+-----------------+
```

gforsyth:
Hey @Wideshanks -- thanks for reporting!  Just to note that the current xonsh release is 0.4.7, but I can reproduce this on current master. 

And it's not reusing the first glob, so much as not parsing the second one correctly, it seems:

```console
gil@bad_cat ~/tmp $ ls
bar  foo  one.txt  one_test.txt  two.txt  two_test.txt
gil@bad_cat ~/tmp $ echo *.txt and echo *_test.txt
[...]

PR #2273, opened 2017-02-27
Bad globs

scopatz:
This improves autowrapping to include more ambiguous cases. This is intended to address #1871


scopatz:
I should note that this fixes a bug introduced by subshell, too

melund:
I think this looks OK, but I haven't been that thorough.

Issue #2024, opened 2016-12-19
On Windows, some environment variable names change case


beetleb:
I use xonsh on Windows (0.4.7)

Specifically, I open a command prompt, and then run xonsh explicitly under it.

I noticed some (but not all) environment variables that existed in mixed case show up as upper case once in Xonsh.

As an example,

in Windows prompt:

FrameworkDIR64 

Once in Xonsh, this becomes

FRAMEWORKDIR64

This is causing a few things to break on our side.

Not sure why this happens. Can I control this? Also not sure why it doesn't happen on all variables.

Thanks.

scopatz:
Thanks for reporting @beetleb. I am not a Windows person, but this sounds like a windows environment thing.  Does the same thing happy in Python, when you look at the environment there? Maybe @melund @BYK or @astronouth7303 have more experience in this area?

astronouth7303:
CPython does this to `os.environ`. For xonsh to circumvent this would be.... troublesome.

beetleb:
Hmm... Looking around more, I find this:

http://stackoverflow.com/questions/7797269/python-environment-case-senstivity-os-environ

And a ticket in Python for it:

https://bugs.python.org/issue28824

According to the first link: On Windows, although the OS preserves the case, apparently it is case insensitive. So I can create an environment variable called "abc", but if I set ABC, it will update the one in "abc".

So Windows environments should not have more the same environment variable in multiple cases.

However, Xonsh does allow this. If I go into xonsh, abc becomes ABC due to os.environ. But then I can create a new environment variable called "abc" and I'll have both.

For my needs, this duplicate is the issue (the case probably doesn't matter). Here are the steps:

1. Xonsh inherits an environment variable (let's call it abc), and converts to upper case ($ABC). 

2. Then as part of some build script we have (probably a batch file somewhere), it tries to update abc. 

3. Since in Xonsh, $abc does not exist, instead of updating $ABC, it creates a new environment variable called $abc. 

4. Later on in the build script, some tool is storing all the environment variables, and apparently quits if it finds duplicate environment variables. I really don't know the internals of that tool, but the error message from that tool complains about the key already existing, and points to that environment variable.

Do you see the problem?

The "proper" solution is not obvious, but I can sympathize with the comments to the question on the StackOverflow page. Somehow Xonsh should not allow the same environment variable to show up more than once in multiple cases in Windows, as it can easily break things in Windows. I do not know if Xonsh has any "say" when a .bat file is called, though.

[...]

PR #2276, opened 2017-02-27
Use nt.environ on windows

scopatz:
This is meant to address #2024.

### DIFF
--- a/news/ntenv.rst
+++ b/news/ntenv.rst
@@ -1,0 +1,16 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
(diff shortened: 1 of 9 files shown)


melund:
Awesome. Thanks a lot for fixing this. 

What do you think. Should we have a check Windows, to detect if two environment values have same name but different case? I guess one of them would implicitly be discarded when starting sub proceses. This should be a separate issue.

scopatz:
> I guess one of them would implicitly be discarded when starting sub proceses. This should be a separate issue.

Yeah this is probably a separate issue.

PR #2277, opened 2017-02-27
only remove attribute if it exists

### LINKED ISSUE #2264: "history play" does not work when not installed "pygments"
I found it if the library does not installed, `history replay` does not work by diving in source code.
screenshot 2017-02-26 16 29 06

Since there is no explanation about `pygments` or `style` in the document(http://xon.sh/tutorial_hist.html), I want you to add the sentences about that.

gforsyth:
Attempt to fix #2264 which only rears its head during history replay (on
linux).

### DIFF
--- a/news/fix_history_replay.rst
+++ b/news/fix_history_replay.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks for the fix @gforsyth! Waiting on CI

scopatz:
Circle actually ran. Not sure why it isn't reported.

scopatz:
Thanks!

PR #2281, opened 2017-02-28
A few doc fixes

gforsyth:
A couple small things here:

* `on_ptk_create` now shows up in the documentation. This does require a non-beautiful solution wherein we import `events` via `xonsh.ptk.shell` so that the events all register correctly. That said, `prompt_toolkit` is not a strict dependency for building the docs, just certain events won't show up for those who build them locally without it.

* `jsonutils` doc page was misnamed and was causing build warnings in sphinx

No news entry for this, I think


scopatz:
Thanks @gforsyth! Going to merge because the S3 outage is making this take forever.

scopatz:
and this is a doc-only change

Issue #2122, opened 2017-01-29
history json OSError bad file descriptor


laerus:
occasionally on xonsh startup:
```
Exception in thread Thread-5:
Traceback (most recent call last):
  File "/usr/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/home/laerus/src/xonsh/xonsh/history/json.py", line 103, in run
    files = self.files(only_unlocked=True)
  File "/home/laerus/src/xonsh/xonsh/history/json.py", line 126, in files
    fs = _xhj_get_history_files(sort=False)
  File "/home/laerus/src/xonsh/xonsh/history/json.py", line 70, in _xhj_get_history_files
    files = [os.path.join(data_dir, f) for f in os.listdir(data_dir)
OSError: [Errno 9] Bad file descriptor: '/home/laerus/.local/share/xonsh'
```

scopatz:
Hmm Odd. Thanks for reporting @laerus. I assume that the directory actually exists?

laerus:
haven't deleted anything manually for some time, only reinstalled from master

scopatz:
It would be nice if we could find a way to reliably reproduce this. We can always put a try-except around this, but it would be nice to know what is going wrong.

scopatz:
See safety fix in #2284

melund:
#2284 is merged. If anyone sees the issue again feel free to reopen.

PR #2284, opened 2017-03-01
Defense against collecting history files

scopatz:
This is meant to prevent the kind of issue seen in #2122


melund:
Looks good.

Issue #2254, opened 2017-02-25
Subprocess-mode line continuations eat spaces in scripts


t184256:
I thought it should be somewhere in the tutorial or bash translation guide, but I can't find it there. Let's suppose I wanna convert this piece of a script to xonsh:
```
echo --option1 value1 \
    --option2 value2 \
...
    --optionZ valueZ
```

Currently this eats spaces: `--option1 value1--option2 value2...`. The tutorial kinda warns about that, so cool, everybody needs something to eat.

I can use trailing spaces and `$[` to overcome this:
```
$[
echo --option1 value1 
    --option2 value2 
...
    --optionZ valueZ
]
```
But you know, _trailing spaces_.

I can also get creative and build it like that:
```
$[
echo @((
    '--option1', 'value1',
    '--option2', 'value2',
...
    '--optionZ', 'valueZ'
))
]
```
That's cool and flexible, but overly bracely and magical.

Let's suppose I don't need that flexibility and I just want to write an extremely long command in a script without trailing spaces or three layers of braces. What's the best practice for running long shell commands in a script?

scopatz:
Hi @t184256, it seems like 

```
echo --option1 value1 \
    --option2 value2 \
```

should work.  There have been a couple of fixes to this part of the code but apparently there is more work to do.

Personally, I don't have a problem with the trailing spaces. It is a fairly Pythonic idiom.

t184256:
So the first one is the preferred way, but it's not working at the time? Ugh, I fail to name issues all the time. Not even close!
```
$ echo -n 'echo --option1 value1 \\\n--option2 value2' > ml.xsh
$ xonsh ml.xsh
--option1 value1 --option2 value2
```

> Personally, I don't have a problem with the trailing spaces. It is a fairly Pythonic idiom.

Whoa. With all respect, this particular opinion doesn't seen Pythonic at all to me. The first strong opinion that springs to mind is, um, PEP 8:

> Avoid trailing whitespace anywhere. Because it's usually invisible, it can be confusing: e.g. a backslash followed by a space and a newline does not count as a line continuation marker. Some editors don't preserve it and many projects (like CPython itself) have pre-commit hooks that reject it.

scopatz:
> Whoa. With all respect, this particular opinion doesn't seen Pythonic at all to me. The first strong opinion that springs to mind is, um, PEP 8:

[...]

PR #2287, opened 2017-03-02
Backline

scopatz:
This fixes an issue in the lexer and in wrap subproc that was preventing the correct usage of line continuations in subporc mode. It should address #2254

### DIFF
--- a/news/backline.rst
+++ b/news/backline.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Line continuations (``\\``) is subproc mode will no longer consume the
+  surrounding whitespace.
+
+**Security:** None
--- a/tests/test_ast.py
+++ b/tests/test_ast.py
@@ -47,7 +47,7 @@ def test_gather_load_store_names_tuple():
     # the second line wil be transformed still even though l exists.
     'l = 1',
 ])
-def test_multilline_num(line1):
+def test_multilline_num(xonsh_builtins, line1):
     code = line1 + '\nls -l\n'
     tree = check_parse(code)
     lsnode = tree.body[1]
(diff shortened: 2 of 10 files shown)


scopatz:
Anyone able to review this, please?

gforsyth:
Hey @scopatz -- thanks for putting this in! 

So, this works:
```
gil@theo ~ 🐚  echo --test 1 \
.............. --test2
.............. 
--test 1 --test2
```

but not this (note no space between `test` and `1`)
```
gil@theo ~ 🐚  echo --test1 \
.............. --test2
.............. 
xonsh: subprocess mode: command not found: --test2
```

melund:
Also. Does this still handle the fact that there should always be a white space before the backslash on windows?

melund:
I see the same problem @gforsyth reports.

scopatz:
> Also. Does this still handle the fact that there should always be a white space before the backslash on windows?

Yes, that should not be any different

scopatz:
Thanks @gforsyth for finding that! I am actually surprised that I was able to fix that. That expression is fully ambiquous with Python and the error was happening by wrapping the second line on its own.  Luckily the node the AST Transformer got represented both lines...

scopatz:
This should be ready for another look.

scopatz:
Ok! This is back to passing.

gforsyth:
Nice!

```console
gil@bad_cat ~ $ echo --test1        \
...............      --test2   \
...............      --test3      \
...............      --test4
...............      
--test1 --test2 --test3 --test4
```

I noticed that multiline mode isn't triggered if there's a trailing space _after_ the `\`, but I think that's fine and also I imagine that if it _did_ trigger a newline it would cause problems on Windows.

Looks good to me!

scopatz:
> I noticed that multiline mode isn't triggered if there's a trailing space after the \

Yeah that is the rule for trailing backslashes in most languages :) Thanks for the merge!

PR #2295, opened 2017-03-06
Never use os.scandir

### LINKED ISSUE #2294: Python 3.4 no longer supported?
Hi all,

First of all, thank you all for your great work. 

I have noticed that the version bump to 0.5.7 introduced a call to `os.scandir` which is not supported by Python <3.5 afaik. As I am still using Ubuntu 14.04 with Python 3.4 on a few machines, this is a little bit of a headache... I don't know the codebase, but it looks like `xonsh.platform.scandir` could be used instead?

melund:
Fixes #2294. `os.scandir()` is not supported on py34, so we should use our internal version instead.

### DIFF
--- a/news/scandir_bug.rst
+++ b/news/scandir_bug.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fixed a bug on py34 where os.scandir was used by accident. 
+
+**Security:** None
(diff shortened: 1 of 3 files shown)


astronouth7303:
Why didn't CI pick this up when the code was originally written?

melund:
Well. @gforsyth may know. I think he added the code. Maybe this part of the code isn't covered by tests.

gforsyth:
I thought it was covered... interesting that it apparently isn't? We should fix that.  Thanks for putting in the fix, @melund

gforsyth:
Hey @melund -- I added a check to make sure the `hg` stuff doesn't barf when there's no repo there if you want to take a look.

melund:
Awesome. Thanks a lot @gforsyth

PR #2303, opened 2017-03-07
Release cwd on Windows while prompt is displayed

### LINKED ISSUE #2269: Working directory is locked (Windows)
Python somehow locks the current working directory on Windows. 
```
snail@sea ~\Documents\xonsh delcwd
$ rm -rf ~/x; mkdir ~/x; cd ~/x; rm -r ~/x
rm: remove directory 'C:\Users\mel/x'? y
rm: cannot remove 'C:\Users\mel/x': Device or resource busy
```

Or if I try from the GUI:

image

Does anyone know a way to prevent this? I have search but haven't come up with anything.

melund:
Fixes #2269

The fact that xonsh locks directories on windows and prevents them from being deleted or renamed is pretty annoying. Especially when you have many open shells around. 

This is an attempt at a hack to solve it. On Windows, the current directory is no longer locked when the prompt is shown. This is accomplished by resetting the CWD to the users home directory temporarily while the prompt is displayed. The CWD is restored when the prompt returns so it doesn't affect any commands being run. I.e. xonsh still can't remove its own working directory.


astronouth7303:
-1 This has the potential to cause weirdness with xontribs (eg, if they hook PTK).

I would only support this if its standard behavior for shells.

Unfortunately, a quick google doesn't seem to indicate a way to just release this lock.

scopatz:
Thanks for putting this in @melund. I think having this as optional (default off) is probably the best option. This is clearly bothering some @melund and probably others, so I have no problem with this in theory.  My only other suggestion is to maybe allow the target "reserved" directory be customizable. The user's home seems reasonable.  Maybe Progam Files or the some other system directory or the xonsh install location would be even better?

This probably needs an extra environment variable, to manage this behaviour.

Also, I think the only other shells that makes Windows console & system API calls directly are cmd and powershell. These are probably the main ones worth comparing.

melund:
> @astronouth7303 : -1 This has the potential to cause weirdness with xontribs (eg, if they hook PTK). I would only support this if its standard behaviour for shells.

That is my biggest concern as well. It would be a problem if a xontrib tried to use `os.getcwd()` in any of the PTK callbacks. But it could use `__xonsh_env__['PWD']` without an issue. 

I don't know if this is standard behaviour from shells on Windows. I just know that this is a real killer feature. One that PowerShell has, but CMD.exe doesn't have. How PowerShell accomplish it I don't know. PowerShell is probably not constrained by the concept of a 'current working directory' that Python enforces on us. I wish there was another way to accomplish the same thing. Alas, I haven't found it. 

>@scapatz : think having this as optional (default off) is probably the best option. 

I agree. I will add that. 

> My only other suggestion is to maybe allow the target "reserved" directory be customizable. The user's home seems reasonable. Maybe Progam Files or the some other system directory or the xonsh install location would be even better?

Well.The xonsh install directory could cause havoc with havoc with conda. I don't think I see the reason for being able to customise this. It is just a matter of getting `cwd` out of the way. Maybe the root directory of the current drive is the best option. We could use `os.path.splitdrive` for that. 

I will give this another iteration when I find time.

astronouth7303:
> That is my biggest concern as well. It would be a problem if a xontrib tried to use os.getcwd() in any of the PTK callbacks. But it could use __xonsh_env__['PWD'] without an issue.

Part of the power of xonsh is that it's also all Python. I don't think a list of gotchas is conducive to this (esp if you develop on *nix and then a windows user is confused when it gives really weird error messages). This is especially problematic if its not the xontrib's code but a module its using/wrapping.

Actually, this might also break any prompt-time xontrib that calls commands. How does this work with VCS support and `$UPDATE_PROMPT_ON_KEYPRESS`?

melund:
>Actually, this might also break any prompt-time xontrib that calls commands. How does this work with VCS support and $UPDATE_PROMPT_ON_KEYPRESS?

It will not affect any xontribs using the `ON_PRE_PROMPT`/`ON_POST_PROMPT` events. But it can't work when `$UPDATE_PROMPT_ON_KEYPRESS` is enabled.

gforsyth:
Hmm... should this then go in as a xontrib and we can also add in something to deal with conflicts between certain xontribs? (`xonflicts`?)

At the same time, `$UPDATE_PROMPT_ON_KEYPRESS` has always been completely unusable for me and I'm on a pretty fast machine.

melund:
Yes. `$UPDATE_PROMPT_ON_KEYPRESS` is not really usable unless you have a very simple prompt. I don't think this will conflict with anything atm. It resets the working directory when prompt_toolkit makes a callback to the completer. If the user enables `$UPDATE_PROMPT_ON_KEYPRESS` then this features automatically disables. 

I could probably argue that this feature could be part of prompt_toolkit. I can try to make a feature request there. But until that happens I think it fits here.

melund:
Well. Maybe not true. There is also the keybindings and history. I guess you are right. This belongs more in a xontrib. 

I will try to move it when I get time.

melund:
It is really a nice feature. Too bad most users won't get by default. I moved the code to xontrib now.

melund:
@gforsyth and @astronouth7303. I moved the code to a xontrib, so it is ready for another look.

gforsyth:
Hey @melund -- looks like some jupyter notebook checkpoints and some spyder config files were added in? Can you remove/rebase those out?

melund:
Argh. Not again. Visual studio code has a really bad default behavior on this point. I will rebase

melund:
@gforsyth . I have rebased.

melund:
Ping

melund:
Thanks a lot @laerus ..

Issue #1480, opened 2016-07-25
Recall last argument from previous command


amitkot:
I'm looking for a way to recall the last argument of the previous command, similar to `Alt+.` on Bash.
I'm using vi mode, so `Esc+_` does not work.
Is there a way to get this on Xonsh?

Posted this on the IRC channel a couple of weeks ago and got no reply, so I assume this is currently not implemented.

gforsyth:
Hey @amitkot, thanks for opening this.  There is not currently a shortcut for doing this, although a naive implementation wouldn't be too hard at this point.  
In the interim, you could do something like this as a temporary workaround:

```console
gil@bad_cat ~ $ aliases['lc'] = lambda args, stdin=None: _.inp.split()[-1]
gil@bad_cat ~ $ ls tmp
file1  file2  file3
gil@bad_cat ~ $ cd $(lc)
gil@bad_cat ~/tmp $ 
```

where `_.inp.split()[-1]` will return the last argument of the previous command

Cadair:
Where in the code base would a key binding like this belong?

scopatz:
@Cadair - It would go in the PTK shell key bindings  (`xonsh.ptk.key_bindings`) or in the readline startup
`xonsh.readline_shell.setup_readline()`

Cadair:
This seems to have been implemented?

scopatz:
It might have been. @gforsyth  would know more ;)

duerrp:
Something like this should do the trick right? I guess one should also be able to implement Alt + m like this (another one that's hard to erase from muscle memory...)

```
from prompt_toolkit.keys import Keys
from prompt_toolkit.filters import Condition, EmacsInsertMode, ViInsertMode

@events.on_ptk_create
def custom_keybindings(bindings, **kw):
    handler = bindings.registry.add_binding

    @handler(Keys.Escape, '.')
    def say_hi(event):
        event.current_buffer.insert_text(_.inp.split()[-1])
```

astronouth7303:
@duerrp Looks about right, although I will note that `_` is not always the last command. (It can be the result of a Python expression, or if it's the first line, nothing at all.)

gforsyth:
I wonder if this should go into the bashisms xontrib?

duerrp:
[...]

PR #2310, opened 2017-03-11
add Alt-. cmd to recall last arg to bashisms

gforsyth:
As requested in #1480 this adds the `Alt .` shortcut to insert the last argument from the previous command into the current prompt buffer at cursor position.


laerus:
works great ty @gforsyth ! do we need a news entry?

gforsyth:
Hey @laerus -- just added the news entry so this should be good to go in once CI finishes. Thanks!

PR #2317, opened 2017-03-14
WIP: Error out when a path in $PATH is not a directory

### LINKED ISSUE #2313: prompt: error: on field 'curr_branch' on running xonsh
Hey, I wanted to try out xonsh and installed it using homebrew, but I see this error every time I get into a `xonsh` shell. It looks like something related to getting the current git branch. 

```
~  xonsh
prompt: error: on field 'curr_branch'
  File "/usr/local/Cellar/xonsh/0.5.8/libexec/lib/python3.6/site-packages/xonsh/prompt/__amalgam__.py", line 671, in _get_field_value
    value = field_value() if callable(field_value) else field_value
  File "/usr/local/Cellar/xonsh/0.5.8/libexec/lib/python3.6/site-packages/xonsh/prompt/__amalgam__.py", line 486, in current_branch
    has_git = bool(cmds.locate_binary('git', ignore_alias=True))
  File "/usr/local/Cellar/xonsh/0.5.8/libexec/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 7790, in locate_binary
    _ = self.all_commands
  File "/usr/local/Cellar/xonsh/0.5.8/libexec/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 7733, in all_commands
  File "/usr/local/Cellar/xonsh/0.5.8/libexec/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 5881, in executables_in
    yield from func(path)
  File "/usr/local/Cellar/xonsh/0.5.8/libexec/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 5849, in _executables_in_posix
    yield from _yield_accessible_unix_file_names(path)
  File "/usr/local/Cellar/xonsh/0.5.8/libexec/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 5829, in _yield_accessible_unix_file_names
    for file_ in scandir(path):
NotADirectoryError: [Errno 20] Not a directory: '/usr/local/bin/jamf'
```

Here are my `xonfig` and `.xonshrc`:

```
r@GOHG ~ (ERROR:curr_branch) $ xonfig
| xonsh            | 0.5.8           |
| Python           | 3.6.0           |
```

```
$ cat ~/.xonshrc
$XONSH_SHOW_TRACEBACK = True
```

Let me know if you need any other information and/or need help fixing this. Thanks!

rajathagasthya:
If $PATH variable contains a path which is not a directory, this
raises an error.

Fixes #2313.

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -136,6 +136,15 @@ def findfirst(s, substrs):
     return i, result
 
 
+def _is_directory(path):
+    """Takes a path or a list of paths and checks if it is a directory.
+    """
+    if isinstance(path, collections.Iterable):
+        return all(os.path.isdir(p) for p in path)
+    else:
+        return os.path.isdir(path)
+
+
 class EnvPath(collections.MutableSequence):
     """A class that implements an environment path, which is a list of
     strings. Provides a custom method that expands all paths if the
@@ -187,6 +196,8 @@ class EnvPath(collections.MutableSequence):
         self._l.__delitem__(key)
 
     def insert(self, index, value):
+        if not _is_directory(value):
+            raise ValueError('Path should be a directory')
         self._l.insert(index, value)
 
     @property
@@ -203,6 +214,28 @@ class EnvPath(collections.MutableSequence):
         if len(self) != len(other):
             return False
         return all(map(operator.eq, self, other))
+
+
+class EnvPathDir(EnvPath):
+    """
+    Inherits from EnvPath class and checks if all paths are directories when
+    setting the paths.
+    """
+    def __init__(self, args=None):
+        super().__init__(args=args)
+        if not _is_directory(self._l):
+            raise ValueError('Path(s) should be a directory or a list of '
+                             'directories')
+
+    def __setitem__(self, index, item):
+        if not _is_directory(item):
+            raise ValueError('Path should be a directory')
+        super().__setitem__(index, item)
+
+    def insert(self, index, value):
+        if not _is_directory(value):
+            raise ValueError('Path should be a directory')
+        super().insert(index, value)
 
 
 class DefaultNotGivenType(object):
@@ -1001,12 +1034,28 @@ def is_env_path(x):
     return isinstance(x, EnvPath)
 
 
+def is_env_path_dir(x):
+    """
+    Tests if something is an environment path and the path(s) are all
+    directories.
+    """
+    return isinstance(x, EnvPathDir)
+
+
 def str_to_env_path(x):
     """Converts a string to an environment path, ie a list of strings,
     splitting on the OS separator.
     """
     # splitting will be done implicitly in EnvPath's __init__
     return EnvPath(x)
+
+
+def str_to_env_path_dir(x):
+    """
+    Converts a string to an enviroment path, splitting on the OS separator,
+    and ensures path(s) are all directories.
+    """
+    return EnvPathDir(x)
 
 
 def env_path_to_str(x):
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -39,7 +39,7 @@ from xonsh.tools import (
     is_logfile_opt, to_logfile_opt, logfile_opt_to_str, executables_in,
     is_nonstring_seq_of_strings, pathsep_to_upper_seq,
     seq_to_upper_pathsep, print_color, is_history_backend, to_itself,
-    swap_values,
+    swap_values, is_env_path_dir, str_to_env_path_dir,
 )
 import xonsh.prompt.base as prompt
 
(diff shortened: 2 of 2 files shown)


rajathagasthya:
Hey @gforsyth! This is an attempt to fix #2313. Am I on the right track with this change? This still prints an ugly exception to the terminal, so let me know how to fix that. And I still need to add tests.

xonsh/tools.py
```diff
@@ -136,6 +136,15 @@ def findfirst(s, substrs):
     return i, result
 
 
+def _is_directory(path):
```
laerus: this can be a 'public' method, someone may use it outside tools.

xonsh/tools.py
```diff
@@ -205,6 +216,28 @@ def __eq__(self, other):
         return all(map(operator.eq, self, other))
 
 
+class EnvPathDir(EnvPath):
```
laerus: i don't think we need to subclass here, all the functionality below can be part of the EnvPath class above, PATH should always be a list of dirs anyway.

gforsyth:
Hey @rajathagasthya -- thanks for putting this in! You're definitely on the right track. I agree with @laerus on his suggestions. You can also use `xonsh.tools.XonshError` for the errors instead of `ValueError`. 

Alternatively, it could be a warning instead and simply print a message that the proposed entry has not been added to `$PATH` since it's an invalid entry.

gforsyth:
Also, you should add an entry in the `news` folder describing the changes. Instructions are available here: http://xon.sh/devguide.html#changelog

laerus:
it seems that bash just ignores non directory entries

xonsh/tools.py
rajathagasthya: So that was my original implementation. But then I found out that `BASH_COMPLETIONS` uses the same class to validate/convert the paths in https://github.com/xonsh/xonsh/blob/master/xonsh/environ.py#L123. So if the file `/usr/local/etc/bash_completion` was one of the paths in `BASH_COMPLETIONS`, it would fail the directory check.

laerus: you're right, maybe we must solve this on a higher level as in when traversing the $PATH looking for commands [here](https://github.com/xonsh/xonsh/blob/master/xonsh/commands_cache.py#L97), it may also need a small cleanup/fixing. We either need to change this in `commands_cache.py` or change how we represent the `BASH_COMPLETION` envvar, maybe @gforsyth can chime in on that decision?

gforsyth: I think I'd prefer to see `BASH_COMPLETIONS` change. To my mind, a `PATH` is a collection of directories containing executables -- the `BASH_COMPLETIONS` isn't, necessarily, because of how the lazy-loading in the later versions work.

rajathagasthya: I'm curious, what would the `BASH_COMPLETIONS` changes look like?

laerus: we can try making `BASH_COMPLETIONS` just a list, although in general we should not crash xonsh here with and an error raised in `Env` coming from `EnvPath`, we whould print a message and ignore the entry as @gforsyth said before. You can try just making the `EnvPath` tweaks and push the changes in a commit so we can see tests run and then we can go about messing with `BASH_COMPLETIONS`

laerus:
oh and some tests about these changes would be nice :)

xonsh/tools.py
rajathagasthya: Sorry, just resuming this work. I'm still not quite clear how `BASH_COMPLETIONS` can just be a list. Right now, `is_env_path` and `str_to_env_path` are used to validate/convert entries. So if I change the `EnvPath` to only include directories, wouldn't `BASH_COMPLETIONS` fail?

gforsyth: Hey @rajathagasthya -- sorry for the radio silence. Yes, changing `EnvPath` to only include directories will cause `BASH_COMPLETIONS` to fail. My suggestion was to change the ensurers for `BASH_COMPLETIONS` to instead check that it's contents are either a directory or a file. 

In short, PATHS shouldn't contain files, generally, but `BASH_COMPLETIONS` is an exception. But we should special case the exception, not the general type. 

Does that make sense? (I'm not sure if it does, but it's in my head)

rajathagasthya: @gforsyth Yeah, it makes sense. I'm still have to realize the implementation, which shouldn't be too hard. 

The only concern for me is how to get the tests working. If we have checks like `os.path.isdir()` or `os.path.isfile()` and we parametrize pytest test methods with dummy values, it's bound to fail these checks. We might have to use http://doc.pytest.org/en/latest/tmpdir.html to make it work, but other suggestions are welcome.

gforsyth:
Hey @rajathagasthya -- sorry, I've been busy at work the last weeks -- is this ready for another look? I'm happy to help figure out a good way forward.

gforsyth:
Hi @rajathagasthya  -- I'm going through and closing all pull requests that have been idle for more than a year.  We'd love to have your contribution and if you come back to it, please feel free to either open a new PR or you can ping here to have us reopen the existing PR.

Thanks!

Issue #2318, opened 2017-03-14
Better document installation of PTK and the usage of xip


AstraLuma:
```
19:42:17 <rodarmor> I'm trying to get $RIGHT_PROMPT working, which I think requires prompt-toolkit integration. What do I need to do to turn on prompt-toolkit integration? (Besides installing it with pip)
19:45:40 <astronouth7303> rodarmor: Make sure you have both ptk and pygments installed (i think pygments is still required). After that, it should be automatically detected at startup (unless you set it manually through xonfig or .xonshrc)
19:49:09 <rodarmor> Hmm, I did `pip3 install prompt-toolkit` which worked, and I can do `import prompt_toolkit` inside of a plain python3 instnace
19:49:33 <astronouth7303> which platform? installation method?
19:49:39 <rodarmor> But having `$SHELL_TYPE = 'prompt_toolkit'` in my .xonshrc gives an error
19:49:45 <astronouth7303> what error?
19:50:56 <rodarmor> `prompt_toolkit is not available, using readline instead`
19:51:05 <rodarmor> I'm on macos using python3 and pip3 from homebrew
19:51:12 <astronouth7303> did you install pygments?
19:51:20 <astronouth7303> and how did you install xonsh?
19:51:29 <rodarmor> Yup, `pip3 install pygments` worked
19:52:23 <rodarmor> I installed xonsh through homebrew, and that looks like it was the problem.
19:52:40 <rodarmor> I uninstalled xonsh using homebrew, then reinstalled it using pip3
19:52:45 <astronouth7303> yeah
19:53:02 <astronouth7303> iirc, homebrew xonsh uses an isolated environment, not the default system one
19:53:18 <astronouth7303> use `xip` to install things you want to use from xonsh
19:54:43 <rodarmor> xip?
19:55:14 <astronouth7303> it's an alias for xonsh's pip
19:55:17 <rodarmor> Ahh, okay, xip is a xonsh command. On macos there's also a `/usr/bin/xip` which is completely different
19:55:27 <astronouth7303> .... of course there is.
19:55:29 <astronouth7303> crap.
19:55:56 <rodarmor> Haha, yeah,  the unix command namespace is pretty much the wild west, so it's hard to avoid collisions.
```

melund:
@astronouth7303 . Should we rename it to `xpip`. It may help user understand what the command is for.

astronouth7303:
[...]

PR #2319, opened 2017-03-15
Recommend `xip` to install my xontribs

astronouth7303:
xontribs should be installed with xip now, so recommend that for the xontribs I'm responsible for.

This is mostly to try to improve on #2318.

(No news required.)

### DIFF
--- a/xonsh/xontribs.json
+++ b/xonsh/xontribs.json
@@ -211,14 +211,14 @@
    "license": "GPLv3",
    "url": "https://github.com/astronouth7303/xontrib-avox",
    "install": {
-    "pip": "pip install xontrib-avox"
+    "pip": "xip install xontrib-avox"
    }
   },
   "xontrib-z": {
    "license": "GPLv3",
    "url": "https://github.com/astronouth7303/xontrib-z",
    "install": {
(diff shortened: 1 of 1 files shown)


gforsyth:
This looks good to me, @astronouth7303 -- I recognize you not wanting to summarily change all xontrib installs to this, but maybe we should? Is there a downside to doing that?

astronouth7303:
Other than `xip` itself being largely untested (afaik), I don't think so.

Although, to be honest, I think this needs to be refactored into a more structured form to support wizard and package management functions in xonfig and xontribs.

gforsyth:
Well, we can always test `xip` in production.  My inclination is to recommend people use `xip` for all `xontribs` -- it will hopefully save us a bunch of headaches around homebrew, if nothing else

melund:
@astronouth7303 Will you change to `xip` for all xontribs or should we just leave it for someone else to add that change?

astronouth7303:
Leave it for a future change? There's a bunch of other documentation that needs to happen, too.

melund:
Sounds good. Thanks for adding this ..

PR #2323, opened 2017-03-15
Pretty path

### LINKED ISSUE #2321: Add _repr_pretty_ method to EnvPath
The `$PATH` variable doesn't pretty print like other python lists.

We should add the `def _repr_pretty_` to the class. see https://github.com/xonsh/xonsh/blob/master/xonsh/pretty.py#L29

melund:
Fixes #2321 

This will pretty print $PATH

```python
EnvPath(
['C:\\Users\\mel\\Anaconda3\\Library\\bin',
 'C:\\Program Files\\ImageMagick-7.0.4-Q16',
 'C:\\Program Files\\Docker\\Docker\\Resources\\bin',
   <snip>
 'C:\\Users\\mel\\AppData\\Local\\Microsoft\\WindowsApps',
 'C:\\Program Files\\CMake\\bin',
 '']
)
```

### DIFF
--- a/news/pretty_path.rst
+++ b/news/pretty_path.rst
@@ -1,0 +1,13 @@
+**Added:** 
+
+* Pretty printing of the $PATH variable
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


laerus:
seems good, ty @melund !

PR #2324, opened 2017-03-15
add missing default ensurers for all default envs of type int or bool

gforsyth:
Fixing a bug reported on gitter.

Need tests to make sure that all of the default envvars have ensurers -- these issues seem to crop up more for those who use `config.json` to set things...

ping @SolarAquarion


gforsyth:
Ok -- I think I got all of the remaining default environment variables that are expected to be either `int` or `bool`, so that should stop these errors from popping up.

news/dirstack_ensurer.rst
```diff
@@ -8,6 +8,14 @@
 
 **Fixed:**
 
-* Added ensurer to make sure ``DIRSTACK_SIZE`` is an int
+* Added missing ensurers to make sure that ``bool`` env_vars are bools and
+  ``int`` env_vars are integers:
+  * ``DIRSTACK_SIZE``
```
scopatz: For rst, doesn't there need to be a blank line here?

gforsyth: Will do -- my rst-fu is lacking

scopatz:
Thanks @gforsyth! This looks great other than the newline thing, which I'll fix.

gforsyth:
Thanks, @scopatz!

PR #2298, opened 2017-03-06
Make os_environ case-preserving and case-insensitive on Windows

melund:
fixes #2297 

This PR wraps `os.environ` with a case-preserving case-insensitive class on Windows.

PR #2325, opened 2017-03-15
Fix win paths

### LINKED ISSUE #2320: xonsh $PATH does not include user env vars on Windows
xonsh $PATH does not include user environment variables on Windows. As a result, some commands that can be found correctly in `cmd.exe` cannot be found in xonsh.

For example, my global PATHs are:
```python
['C:\\Windows\\System32', 'C:\\Windows', ...]
```
And my user PATHs are:
```python
['D:\\Programs\\nodejs', '...']
```

In xonsh, `$PATH` contains only the global PATHs, while `os.environ['PATH']` contains everything.

In Windows `cmd.exe`, `npm` works, but `xonsh` says `command not found`.
After typing `$PATH = os.environ['PATH'].split(';')`, it works.
BTW, if a command refers to a `.exe` file, it will be found even if its path not in `$PATH`, but not `.bat`.

So the question is, should I set `$PATH = os.environ['PATH'].split(';')` manually or do I miss anything?

melund:
Fixes #2320 (Hopefully)

It ensures that if the `Path` variable from windows is renamed to `PATH`.

### DIFF
--- a/news/win_path.rst
+++ b/news/win_path.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fixed a bug on Windows which meant xonsh wasn't using PATH enrivonment variable but instead relying on a default
+  value from the widnows registry. 
+
+**Security:** None
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -794,8 +794,10 @@ class Env(cabc.MutableMapping):
             args = (os_environ,)
         for key, val in dict(*args, **kwargs).items():
             self[key] = val
-        if ON_WINDOWS and 'Path' in self._d:
-            self._d['PATH'] = self._d.pop('Path')
+        if ON_WINDOWS:
+            path_key = next((k for k in self._d if k.upper() == 'PATH'), None)
+            if path_key:
+                self['PATH'] = self._d.pop(path_key)
         if 'PATH' not in self._d:
             # this is here so the PATH is accessible to subprocs and so that
             # it can be modified in-place in the xonshrc file
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -794,6 +794,8 @@ class Env(cabc.MutableMapping):
             args = (os_environ,)
         for key, val in dict(*args, **kwargs).items():
             self[key] = val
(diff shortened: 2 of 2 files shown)


astronouth7303:
Why didn't #2298 fix this? Will this cause problems when commands are executed down-stream?

gera2ld:
@melund You might have to use `self['PATH'] = self._d.pop('Path')`, otherwise `$PATH` will be a string instead of a list.

gera2ld:
BTW, I think it's not safe to just check `Path`, because it's possible to be `path` or `PAth` or `pATh` and they all work in Windows. Can we just transform them to uppercase before using them?

melund:
> Why didn't #2298 fix this? Will this cause problems when commands are executed down-stream?

@astronouth7303 . It did fix some of the problems, but only for `os.environ` which we wrapped. All the keys are copied into our `Env` class which is still case sensitive. The question is whether the xonsh environment variables should continue to be case sensitive on Windows. I really think it should... We could consider printing warnings if a subprocess is started on Windows, and two environment variables have the same name but different case. 

> @melund You might have to use self['PATH'] = self._d.pop('Path'), otherwise $PATH will be a string instead of a list.

@gera2ld You are right. We try to fix that.

melund:
@gera2ld ... I have updated the PR. See if this is better :)

astronouth7303:
My concern with this PR is that we'll run into the same issue with some other (less critical) envvar.

melund:
@astronouth7303 Well see this as a hotfix for #2276.

We could use a more general solution. The only thing I think of is to make xonsh environment case-preserving/case-insensitive on windows. But that would create a big a difference between platforms.

melund:
Well. We need the hot-fix . Right now it doesn't pick up the Path variable when starting xonsh. The only reason it works on Windows is that it falls back to the default PATH from windows registry. So it is a really bad bug.

melund:
Maybe @scopatz can chime in once he gets back from holiday

laerus:
windows y u not case sensitive? i'm +1 for patching this now and see how it goes.

melund:
I think it is time we patch this. I am all for making bigger changes to handle the underlying problem later. But we shouldn't leave it broken on windows any longer. @scopatz Are you back from holiday?

melund:
Thanks a lot @laerus

Issue #2322, opened 2017-03-15
`cmd /c echo %PATH%` fails on Windows


melund:
$ cmd /c echo %PATH%
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\__amalgam__.py", line 15470, in default
    run_compiled_code(code, self.ctx, None, 'single')
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\__amalgam__.py", line 3634, in run_compiled_code
    func(code, glb, loc)
  File "<xonsh-code>", line 1, in <module>
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\__amalgam__.py", line 5282, in expand_path
    s = expandvars(s)
  File "C:\Users\mel\Anaconda3\lib\site-packages\xonsh\__amalgam__.py", line 6936, in expandvars
    path = WINDOWS_ENVVAR_REGEX.sub(value, path, count=1)
  File "C:\Users\mel\Anaconda3\lib\re.py", line 326, in _subx
    template = _compile_repl(template, pattern)
  File "C:\Users\mel\Anaconda3\lib\re.py", line 317, in _compile_repl
    return sre_parse.parse_template(repl, pattern)
  File "C:\Users\mel\Anaconda3\lib\sre_parse.py", line 944, in parse_template
    addgroup(int(this[1:]), len(this) - 1)
  File "C:\Users\mel\Anaconda3\lib\sre_parse.py", line 888, in addgroup
    raise s.error("invalid group reference %d" % index, pos)
sre_constants.error: invalid group reference 8 at position 842

melund:
It seems like environment substitution with `%var%` is broken. 

Should we consider dropping the support for substituting `%PATH%` variables like this in xonsh. I didn't even know we attempted to support this.

laerus:
hey @melund, i've added a new test about this [here](https://github.com/xonsh/xonsh/blob/expandvars/tests/test_tools.py#L1195) in the expandvars branch that seems to pass on appveyor.

laerus:
can you try while in xonsh
```
from xonsh.tools import expandvars

expandvars('cmd /c echo %PATH%')
```
and see what it returns?

laerus:
i can't get my hands on a windows machine atm :(

melund:
```
[...]

PR #2326, opened 2017-03-15
remove %foo% auto expansion on Windows

melund:
This removes the automatic variable expansion on windows for environment variables wrapped in %. 

It has some bugs and it makes it difficult to execute certain commands without macro escapes.  

See: #2322


laerus:
thank you @melund!

PR #2336, opened 2017-03-20
Make sudo on Windows work again

BYK:
Ideally, we'd move this to a `xontrib` but at least this patch makes it unbroken.

### DIFF
--- a/news/sudo.rst
+++ b/news/sudo.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fixed a regression in the Windows ``sudo`` command, that allows users to run elevated commands in xonsh.
+
+**Security:** None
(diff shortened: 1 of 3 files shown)


astronouth7303:
1. Please add a [news item](http://xon.sh/devguide.html#changelog).
2. What was the problem being fixed?

melund:
@astronouth7303 
There must have been a regression so we just got this error when using this on windows.
```
snail@sea ~\Documents\scratch
$ sudo mkdir newdir
NotImplementedError: Redirection is not supported
```

I can confirm that this PR fixes it. Thanks a lot doing that @BYK. I agree that this is really suited for a xontrib. So make another PR when you get time. Take a look at the other built-in xontribs in `xonsh/xontribs`.

melund:
@BYK. I can add the news item. Thanks for fixing this.

BYK:
@melund thanks! Sorry for the slow response.

I think I've found a way to stream the cosole output back to `xonsh` too but it requires a DLL so more incentives for a `xontrib` :)

melund:
@BYK. That sounds awesome and could be a real killer feature.

Issue #2159, opened 2017-02-05
[Tracking] Re-rewrite Executer


astronouth7303:
Xonsh is having a lot of trouble with fundamental low-level stuff (unthreadable, misalignment of standard out/error, signal issues, etc). Let's fix it.

Some of the issues come from using `subprocess` as the underlying executer, and others from conflating interfacing with the OS with the process of turning an abstract pipeline into concrete system objects.

1. [ ] Write a abstraction library to deal with OS API issues and generally manage resources. ([xonsh/slug](https://github.com/xonsh/slug))
2. [ ] Rewrite xonsh to vendor and use said abstraction library.

mitnk:
IMHO, using `subprocess` may not be a bad thing. After all it's tweaked in years by the community for various platforms.

However `xonsh/slug` looks cool!

astronouth7303:
It crossed my mind to use `subprocess.Popen` as the underlying implementation of a Process primitive. But we still need good management of process groups, integration of virtual processes (Python code as a "process"), and some well-tested utility pipes (tapped pipes, going between pipes and buffers, etc).

mitnk:
I'd say this `xonsh/slug` may go through all the hard things which the current or the previous job controls / subproc running. But you can try this direction out, though it may as hard as the current one.

Another obvious direction is fix issues based on what systems we currently have, it works at over 90% cases, and @scopatz and users have spent countless time to implementing & testing it.

> Rewrite xonsh to vend and use said abstraction library.

For this part, I'd like to know @scopatz 's idea: do you think it's ok to make xonsh's exec-or job-control part replaceable, like history backends :) Anyway this is the heart of xonsh (except parsing part).

scopatz:
Thanks for the concern @mitnk. For the record I am all for any efforts to make the code base more modular and maintainable. The sorts of things that @astronouth7303 is proposing have a lot of merit. 

My understanding / guess is that the initial version of slug would essentially be a copy of what we have in xonsh right now anyway. Also, now that we have an executor that *mostly* works, I think it is smart to do further development outside of mainline xonsh. That way we won't block releases for a couple of months or risk breaking everything until we really know it works. 

> xonsh's exec-or job-control part replaceable, like history backends

Well I don't think it will be replaceable like the history backends or shell frontends. It is going to be more like lazyasd, where there is a separate project that is a core component of xonsh that other people can use if they want. We'll still only have one execution model. At least that was how I understood the plan.

[...]

PR #2337, opened 2017-03-21
[WIP] Slugify

AstraLuma:
Implement #2159.

Trying to rewrite the executor so that it doesn't bend over backwards to make subprocess work.

Instead, use a library for the POSIX process/piping model ([slug](https://github.com/xonsh/slug)) to more concisely express what we're doing with processes and make the code somewhat readable.

TODO

* [x] Aliases
* [ ] PTYs (Single global PTY? PTY per job?) (Blocked on PTY support in slug)
* [ ] Output is not making it out
* [ ] Process stderr along with stdout (will probably change under PTYs)
* [ ] Add `FossilizedJob` for purposes of history backends?
* [ ] Confirm usages of bytes vs text (We should be using bytes internally, except for when consumers--xsh code--want text.)


astronouth7303:
BTW, does any of the other coredevs have commentary on this direction? If I died in 3 weeks and you had to fix a bug, would you be ok with that or would it be terrible?

mitnk:
It's not terrible, if I need fix a bug in it, rather, I'll die for a second time for it. The first time happened when I touched @scopatz 's version ;) I mean this part is (one of) the most intricate part of xonsh. I doubt it can be simplified that much.

My idea is when you feel it's ready for a beta test, we can join and test it. I believe you will encounter a lot of issues/bugs during it. If this direction (slug) can solve most of them easily, then I'd say I support it. Please ping me when you implemented the PTY part.

Nice +/- in diffs!

melund:
@astronouth7303 ... I newer spent time on the old Excer so I wouldn't have much said in this. I always like abstractions, and I am starting to grasp the complexity by following the Slug repo,

What I am trying to say is, that I am all for this, if you convince @scopatz :)

laerus:
only 3 tests failing on linux! 
oddly py36 passes all

pytests fails on app veyor, cannot import winapi from ctypes

astronouth7303:
Unfortunately, those tests point to a bug in slug, which is going to be uber-fun to track down.

astronouth7303:
So I added tests to slug for this output problem, and that seems to be working fine.

Unfortunately, output is still getting dropped somewhere by something.

laerus:
feels like a buffer needs flushing, can't spot where yet.

astronouth7303:
I don't think so? All of the buffers are kernel-native pipes.

astronouth7303:
1. Syncronization usually means locks and such, which won't help.
2. Python does buffering around text-mode pipes, because read-by-line and UTF-8 decode. Slugify moves xonsh to be using binary under the hood, which I have not seen any signs of buffering yet.
3. I'm generally careful about EOF conditions in slug, which should be flushing things out. (Prove me wrong!)

laerus:
do we run tests with `-u`? 
i will try to dig in this w/e.

astronouth7303:
No, `-u` is never used (afaik)

laerus:
we use -u on the shebang of xonsh startup script

gforsyth:
Hey @astronouth7303 -- I recognize that the slugification of xonsh is a massive undertaking.  Do you have a preference between keeping this PR open vs closing it and reopening one when work continues?

astronouth7303:
That's... complicated. I know you want to do cleanup, though.

melund:
@astronouth7303 I read the other day that Windows 10 is now getting a true Pseudo Console (PTY). 

Hopefully, this would allow us to drop support for the Win32 API, and make the Slugify work a lot simpler.

laerus:
This would be awesome but then xonsh won't work on windows7 and older. (i don't mind)

melund:
> This would be awesome but then xonsh won't work on windows7 and older. (i don't mind)

I don't think that would matter. It would still be possible to use an older version of xonsh in that case.

astronouth7303:
Umm... why? Did I write something not Windows 7 compatible?

astronouth7303:
Actually, the slug library doesn't support PTYs at all yet.

It's using Win32 for true process pausing/continuing and process groups.

astronouth7303:
I mean, that was super annoying. But no, I've maintained feature parity between Posix and Windows.

The real challenge is the less mainstream kernels. Currently don't test Cygwin, MSYS, WSL, *BSD, Haiku, etc.

gforsyth:
Closing out old PRs, thanks!

PR #2344, opened 2017-03-27
codecov raised threshold at 5%

laerus:
so we don't get so many failing builds.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,4 +2,4 @@ coverage:
   status:
     patch:
       default:
-        threshold: 1%
+        threshold: 5%
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,4 +2,4 @@ coverage:
   status:
     patch:
       default:
-        threshold: 5%
+        threshold: 2%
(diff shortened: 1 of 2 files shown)


astronouth7303:
Just to be clear, this means that it won't fail until coverage has fallen at least 5 percentage points (eg, from 80->75, 50->45, etc)?

laerus:
yeah, i'm not very fan of this but having so many failing commits due to codecov isn't good either.

laerus:
well apparently this PR somehow raises coverage...

melund:
I am +1 on this. Let us hear what the other say.

mitnk:
What was the reason to introduce codecov? Does 5% mean we would bypass the codecov check without writing any tests in 99.9% PRs :) ?

mitnk:
I mean if 1% is too low to be reasonable, maybe double it would work?

laerus:
yea maybe 2-3% is more reasonable.

mitnk:
LGTM, Thanks @laerus!

Issue #2302, opened 2017-03-07
 Bash Brace Expansion


pando85:
I can't live without [brace expansion](http://www.linuxjournal.com/content/bash-brace-expansion).

TeamSpen210:
You could define a function or two and stick it in builtins in your `xonshrc`, then invoke with `@(func())`. Macros would help there if you want to get rid of quotes around strings.

pando85:
No, I'm talking about expand commands like this:
```bash
rm /a/long/path/{foo,bar} 
# expand as this:
rm /a/long/path/foo /a/long/path/bar
```

melund:
There is an open (never finished) PR for this. https://github.com/xonsh/xonsh/pull/1379

Any help moving it along would be great.

gforsyth:
This should probably go in the bashisms xontrib or a similar space.

JohnLunzer:
I agree with @gforsyth I think bashisms sounds like a good home. It seems like enabling this feature would supplant python set notation syntax. 

That said, perhaps there isn't a reason to maintain the `{}` syntax.

Additionally, is there are more  "xonshic" way to accomplish this in similarly terse syntax?

adqm:
In my fork, I made brace expansion happen inside of ``g`...` `` "strings" but not outside, making use of the [braceexpand](https://pypi.python.org/pypi/braceexpand) module for the actual implementation.  This avoids the conflict with dictionaries at the cost of the three extra characters (``g, `, and ` ``).

If that's a happy medium, I'm happy to port it over.

JohnLunzer:
@adqm It sounds better than not having brace expansion at all. But I'd let others chime in on other possible solutions first.

amstan:
> I can't live without [brace expansion](http://www.linuxjournal.com/content/bash-brace-expansion).

Indeed, one the frequent patterns I do is stuff like `mv /etc/pacman.conf{.pacnew,}`. I essentially type in the common filename, then the before,after inside the {}.

PR #2350, opened 2017-04-03
add support for brace expansion inside "glob-strings"

adqm:
This is a partial implementation of #2302 (brace expansion), relying on the `braceexpand` module for the actual heavy lifting.  Right now there is no syntax for using brace expansion in globs directly in subprocess mode, but this patch allows brace expansion syntax inside of `g`` ` globs.

It should be possible, too, to let `{` and `}` delimit brace expansion in subprocess mode directly, as well, potentially using some of the same rules we already use for disambiguation where necessary.  I'm happy to hack on that here, or if this is a reasonable stepping stone, then that piece could come in a separate PR.

### DIFF
--- a/xonsh/__init__.py
+++ b/xonsh/__init__.py
@@ -10,6 +10,8 @@ else:
     import sys as _sys
     try:
         from xonsh import __amalgam__
+        braceexpand = __amalgam__
+        _sys.modules['xonsh.braceexpand'] = __amalgam__
         completer = __amalgam__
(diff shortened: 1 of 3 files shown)


gforsyth:
Hey @adqm -- I'm going through and closing all pull requests that have been idle for more than a year.  Please feel free to reopen the PR as you see fit.

Thanks!

scopatz:
This is something I still would really like to see happen!

Issue #2356, opened 2017-04-15
xonsh --rc <rcfile> doesn't load properly


chiahaoliu:
Hi,

I was trying to run desired control file with ``--rc`` option. I created the ``xonshrc`` and the ``xonshrc.xsh`` files, both with a simple ``print('hi')``, and place them after ``--rc`` and tried to run both of them.

``xonsh`` can be launched without error but the ``xonshrc`` file wasn't loaded.

I also confirmed if I move the ``xonshrc`` file into ``home`` then the print statement can be run successfully.  

att: @CJ-Wright

laerus:
hey @chiahaoliu thanks for the report!

can confirm that `--rc` doesn't work
cannot confirm the second part that moving xonshrc to home folder solves this.

gforsyth:
Hi @chiahaoliu -- I've got a fix for this in #2359 if you'd like to try that out.

gforsyth:
#2359 was merged into master -- please ping this issue if it's still a problem.  And thanks for reporting!

PR #2359, opened 2017-04-21
fix `--rc` option for xonsh startup

gforsyth:
Quick fix for #2356 

I guess we just forgot to actually implement it?


astronouth7303:
Can we have a test at least confirming that xonsh will load an rc file?

gforsyth:
Tests added -- one for `--rc` and one for `--no-rc`.

PR #2360, opened 2017-04-21
Fix at dollar

### LINKED ISSUE #2340: @$ result is not satisfying
When I run `@$(which ls)` I expect to see this result:
`some_file  some_other_file`

But I see this:
`ls: invalid option -- '
'
Try 'ls --help' for more information.
`

xonsh version: 0.5.9
os : linux (Ubuntu 16.04.2)

gforsyth:
This was borking out when using `@$` with aliased commands (due, I think, to a
trailing newline character that wasn't being stripped out).

Should resolve #2340

Testing this is tricky since it touches a lot of the xonsh internals...

### DIFF
--- a/news/fix_at_dollar.rst
+++ b/news/fix_at_dollar.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** 
+
+* ``@$`` operator now functions properly when returned command is an alias
+
+**Security:** None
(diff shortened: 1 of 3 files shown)


astronouth7303:
Well, what script would have failed before that now succeeds?

gforsyth:
`@$(which ls)`

astronouth7303:
That test seems to work.

gforsyth:
Sweet.  Thanks, @astronouth7303 -- I'll go ahead and merge this.

Issue #2361, opened 2017-04-23
Proposal - Set Unthread-able as Default When Calling Programs


mitnk:
Hi all,

As you may notice, several issues recently are related with run program in thread [1]. Besides the WIP slug, I'm thinking what we can do for now.

I've tried changing default from threadable to unthreadable. This way, we could fix all these issues, without having to update predictors-threadable-dict every time we found one more unthreadable program. I've manually tested this change with varies cases, and haven't found any breaking effects.

So ideas on this proposal? Or @scopatz did I missed something here?

[1] using the same list of @adqm pointed out in email: #2353 #2348 #2358 #2355

laerus:
+1

melund:
Wouldn't this mean capturing stdout would be off by default?

mitnk:
@melund this may what I missed. Could you put more details about it here? Is the "stdout capturing" something related history? or something like `foo = $(ls -l | wc -l)`?

astronouth7303:
I believe under the current implementation, threading is required for piping, capturing stdout, running multiple aliases in the same job, and generally other things shells are expected to do.

laerus:
xonsh captures everything by default, something bash (and others?) don't do but we like it, you must be able to opt-out though. 

Non-blocking looping over `!()` (`iterrraw`) is based on threads too, a killer feature imo and one that probably noone uses (has some issues still).

laerus:
@mitnk no this is not related to history at first. `HiddenCommandPipeline` and `CommandPipeline` capture stdout.

mitnk:
Thanks for comments guys.

@laerus yes, `!()` seems break when set unthreadable as default. i.e. the behavior of the following changed:
```
for x in !(ls -lh):
    print('=={}=='.format(x))
```
And this is only case broken by this change. (`$()`, `@$()` are not affected).

@astronouth7303 piping like `ls`, `ls | wc | cat | wc` would work as before. multiple aliases in same job also works:
```
$ ls | cat | mycat | cat | mycat | wc | mycat
86      87     926
$ tail .xonshrc
def _cat(args, stdin=None):
    for line in stdin:
        print(line.strip())
aliases['mycat'] = _cat
```

@gforsyth So do you mean `!()` by saying "It would break several existing bits of functionality", I haven't found any other breaking cases.

I'm now not going to push the PR to merge, but collect the background info for it. If you know any cases break by changing unthreadable as default, you could let me know.

mitnk:
[...]

PR #2362, opened 2017-04-23
[Do NOT Merge] make unthreadable as default

mitnk:
see #2361 cc @scopatz

### DIFF
--- a/tests/test_commands_cache.py
+++ b/tests/test_commands_cache.py
@@ -44,39 +44,3 @@ FALSE_SHELL_ARGS = [
 @pytest.mark.parametrize('args', FALSE_SHELL_ARGS)
 def test_predict_shell_false(args):
     assert not predict_shell(args)
-
-
-PATTERN_BIN_USING_TTY_OR_NOT = [
-    (False, {10: b'isnotatty'}),
-    (False, {12: b'isatty'}),
-    (False, {10: b'isatty', 100: b'tcgetattr', }),
-    (False, {10: b'isatty', 100: b'tcsetattr'}),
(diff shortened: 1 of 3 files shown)


laerus:
maybe we need this as a temporary solution

gforsyth:
I'm strongly against merging this in. It would break several existing bits
of functionality in exchange for warding off occasional times when we have
to intercede and add things to threadable predictors.

adqm:
I would argue that the threadable predictors is not a real solution to begin with.

Beginners (or people who just want a simple shell they can do list comprehensions in) don't have a shell that works as they expect with xonsh.  That's a problem.  Basic functionality is broken.

I would rather have something that requires the folks who want the fancy feature at the expense of some basic functionality to have to set an option to turn it on, rather than the opposite (currently, beginners who just want a functioning shell potentially have to turn something off (_potentially for each program!_) to make it work).  And there's no error message or warning that suggests that this is the case; xonsh just feels broken to them (hence the bug reports).

To me, it totally makes sense to have the defaults be something that doesn't break job control, and to have the the threaded stuff be something that can be switched on selectively by those who want it.

Then again, this is, in many ways, the same old disagreement...

mitnk:
At least `!()` syntax is broken by making unthreadable as default, so this PR won't be merged.

gforsyth:
@adqm -- if it doesn't also break `!()` and the only 'breakage' is the loss of stdout storing then I don't have a problem with this sort of change. 

And I agree that continuously adding things to `threadable_predictors` is untenable -- although there is some merged code somewhere that helps predict which programs need it.

gforsyth:
Ha, that "merged code", of course, being the code being removed in this pull request... always good to check those things first...

Issue #2371, opened 2017-05-02
Foreign shell set to empty string crashes xonsh


jkw:
sh-4.4$ sudo pip install -U xonsh
Requirement already up-to-date: xonsh in /usr/lib/python3.6/site-packages
sh-4.4$ xonsh 
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 20476, in main
    args = premain(argv)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 20429, in premain
    env = start_services(shell_kwargs)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 20380, in start_services
    xonshrc_context(rcfiles=rc, execer=execer, ctx=ctx, env=env, login=login)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 16772, in xonshrc_context
    login=login)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 16810, in static_config_run_control
    issue_warning=False)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 9204, in load_foreign_envs
    shenv, _ = foreign_shell_data(**shell)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 8888, in foreign_shell_data
    shkey = CANON_SHELL_NAMES[shell]
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 332, in __getitem__
    return obj[item]
KeyError: ''
Xonsh encountered an issue during launch
Failback to /bin/sh
sh-4.4$ which xonsh
/usr/bin/xonsh
sh-4.4$

laerus:
hey @jkw thanks for the report!

this doesn't seem like an amalgam issue, amalgam is just a big file with the whole xonsh codebase.
you can see what is actually going on if you start xonsh with ```XONSH_DEBUG=1``` envvar.

jkw:
Then I get this:
sh-4.4$ XONSH_DEBUG=1 xonsh
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/xonsh/main.py", line 336, in main
    args = premain(argv)
  File "/usr/lib/python3.6/site-packages/xonsh/main.py", line 289, in premain
    env = start_services(shell_kwargs)
[...]

PR #2378, opened 2017-05-14
[WIP] Fix foreign xonfig wizard empty shell name entry

laerus:
#2371
a little hacky  but it does the job.

I think the wizard visitor needs some more functionality, we need some way 
to break traversing the tree and return to the parent.

With the current wizard implementation i can't think of a way to 'cancel' setting 
the foreign shell in case of an empty value from the user (which can be an often
case if you just spam return while in wizard.) and return to the main wizard.

(tests :cry:)

### DIFF
--- a/news/xonfig-wizard.rst
+++ b/news/xonfig-wizard.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Foreign shell xonfig wizard empty shell name entry
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


gforsyth:
Hey @laerus, thanks for putting this in! 

I agree that a blank entry should just cancel the foreign shell entry -- I'm not sure if we should put this in in the meantime?

laerus:
ping

news/xonfig-wizard.rst
```diff
@@ -0,0 +1,13 @@
...
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Foreign shell xonfig wizard empty shell name entry
```
gforsyth: Hey @laerus -- would you mind adding a note here that an empty foreign shell entry will now default to `bash`?

laerus:
i will try to come up with a better solution.

laerus:
closing this for now

PR #2384, opened 2017-05-18
Add xontrib-fzf-widgets

shahinism:
I'm trying to port some of useful [fzf](https://github.com/junegunn/fzf) widgets developed for Bash/ZSH to Xonsh And thought that it might be a good idea to share it with the Xonsh community as a Xontrib.

### DIFF
--- a/news/add_fzf_widgets_extension.rst
+++ b/news/add_fzf_widgets_extension.rst
@@ -1,6 +1,6 @@
-**Added:** None
+**Added:**
 
-* Add `fzf-widgets` xontrib which provides fuzzy search productivity widgets
+* Add "fzf-widgets" xontrib which provides fuzzy search productivity widgets
   with on custom keybindings to xontrib list.
 
 **Changed:** None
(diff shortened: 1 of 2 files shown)


astronouth7303:
Can you add a news item?

Did we ever form an opinion about formally using `xip` in the xontribs listing?

laerus:
No, but we can do that now?

astronouth7303:
@shahinism Please change two things:

* Add a [news item](http://xon.sh/devguide.html#changelog)
* Please use `xip` instead of `pip` in the package installation directions (`xip` works better with homebrew and conda installations)

laerus:
all your CI are belong to us.

news/add_fzf_widgets_extension.rst
```diff
@@ -0,0 +1,14 @@
+**Added:** None
```
gforsyth: Hey @shahinism -- you need to remove the `None` from this line to resolve the test failures.  Thanks!

laerus:
funny how tests fail about news item but news PR lint is OK

laerus:
now the problem is that single backticks are not valid rst you need double backticks     ``

maybe we need a news formatter

laerus:
bummer that seems like a bug with 3.6 and Travis, I think we can merge this.

PR #2392, opened 2017-05-30
Don't show the repr of the LazyBool for whether or not xonsh is running on Linux.

nathan-hoad:
Before this change:

```
nhoad@nhoad-laptop ~ $ xonfig | grep linux
| on linux         | <xonsh.lazyasd.LazyBool object at 0x7f869e059e10> |
```

After this change:

```
nhoad@nhoad-laptop ~ $ xonfig | grep linux
```

### DIFF
--- a/news/xonfig-output.rst
+++ b/news/xonfig-output.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* ``xonfig`` now displays the proper value for "on linux"
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


news/xonfig-output.rst
```diff
@@ -0,0 +1,13 @@
...
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* `xonfig` now displays the proper value for "on linux"
```
gforsyth: Needs double backtick for rst (hence the test failure on circle)

gforsyth:
Hey @nathan-hoad -- thanks for putting this in! I think there's just the test failures from the news entry formatting, but otherwise this looks great!

nathan-hoad:
Ah! Silly me, I didn't think to run the tests for something like this :) should be sorted after this build!

laerus:
great! this was bothering me some months now.

PR #2398, opened 2017-06-07
Spelling Fixes

Jitsusama:
Hello,

I've been an avid user of Xonsh for a few weeks now and found myself reviewing the code in order to understand a few things about how the shell works more fully. In the process of doing this, I came across a few spelling typos. As I discovered a number of these, I decided it might be a good idea to more thoroughly discover these errors and provide fixes for them.

None of these errors were especially bad. They were mostly common misspellings that can occur while bashing out amazing code.

This PR includes my fixes for all of the spelling errors that I was able to discover. This includes spelling fixes for documentation (restructuredText files), comments/docstrings, text strings, local variable names, and a single function declaration; each in a separate commit.

I've run the py.test suite on my Macbook and it passes, so I don't believe that I broke anything (hooray!) :smile:

I tried to be careful with what typos I corrected, and how I corrected them. My hope is that this will be a simple PR to merge, but feel free to comment if there is anything in this PR that you would like to have changed.

### DIFF
--- a/amalgamate.py
+++ b/amalgamate.py
@@ -386,9 +386,9 @@ def rewrite_imports(name, pkg, order, imps):
             for n in a.names:
                 p, dot, m = n.name.rpartition('.')
                 if p == pkg and m in order:
-                    msg = ('Cannot amalgamate amalgam import of '
-                           'amalgamated module:\n\n  import {0}.{1}\n'
-                           '\nin {0}/{2}.py').format(pkg, n.name, name)
+                    msg = ('Cannot amalgamate import of amalgamated module:'
+                           '\n\n  import {0}.{1}\n\nin {0}/{2}.py').format(
+                        pkg, n.name, name)
                     raise RuntimeError(msg)
                 imp = (Import, n.name, n.asname)
                 if imp not in imps:
--- a/amalgamate.py
+++ b/amalgamate.py
@@ -386,7 +386,7 @@ def rewrite_imports(name, pkg, order, imps):
             for n in a.names:
                 p, dot, m = n.name.rpartition('.')
                 if p == pkg and m in order:
-                    msg = ('Cannot amalgamate almagate import of '
+                    msg = ('Cannot amalgamate amalgam import of '
                            'amalgamated module:\n\n  import {0}.{1}\n'
                            '\nin {0}/{2}.py').format(pkg, n.name, name)
                     raise RuntimeError(msg)
--- a/xonsh/base_shell.py
+++ b/xonsh/base_shell.py
@@ -506,8 +506,8 @@ class BaseShell(object):
         return p
 
     def format_color(self, string, hide=False, force_string=False, **kwargs):
-        """Formats the colors in a string. This base implementation's colors are
-        based on ANSI color codes.
+        """Formats the colors in a string. ``BaseShell``'s default implementation
+        of this method uses colors based on ANSI color codes.
         """
         style = builtins.__xonsh_env__.get('XONSH_COLOR_STYLE')
         return ansi_partial_color_format(string, hide=hide, style=style)
--- a/xonsh/base_shell.py
+++ b/xonsh/base_shell.py
@@ -213,10 +213,10 @@ class _TeeStd(io.TextIOBase):
 
 
 class Tee:
-    """Class that merges tee'd stdout and stderr into a single strea,.
+    """Class that merges tee'd stdout and stderr into a single stream.
 
     This represents what a user would actually see on the command line.
-    This class as the same interface as io.TextIOWrapper, except that
+    This class has the same interface as io.TextIOWrapper, except that
     the buffer is optional.
     """
     # pylint is a stupid about counting public methods when using inheritance.
@@ -480,7 +480,7 @@ class BaseShell(object):
             kernel32.SetConsoleTitleW(t)
         else:
             with open(1, 'wb', closefd=False) as f:
-                # prevent xonsh from answering interative questions
+                # prevent xonsh from answering interactive questions
                 # on the next command by writing the title
                 f.write("\x1b]0;{0}\x07".format(t).encode())
                 f.flush()
@@ -506,15 +506,15 @@ class BaseShell(object):
         return p
 
     def format_color(self, string, hide=False, force_string=False, **kwargs):
-        """Formats the colors in a string. This base implmentation colors based
-        on ANSI color codes
+        """Formats the colors in a string. This base implementation's colors are
+        based on ANSI color codes.
         """
         style = builtins.__xonsh_env__.get('XONSH_COLOR_STYLE')
         return ansi_partial_color_format(string, hide=hide, style=style)
 
     def print_color(self, string, hide=False, **kwargs):
-        """Prints a string in color. This base implmentation will color based
-        ANSI color codes if a string was given as input. If a list of token
+        """Prints a string in color. This base implementation's colors are based
+        on ANSI color codes if a string was given as input. If a list of token
         pairs is given, it will color based on pygments, if available. If
         pygments is not available, it will print a colorless string.
         """
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -191,8 +191,8 @@ def get_script_subproc_command(fname, args):
     if not os.access(fname, os.X_OK):
         raise PermissionError
     if ON_POSIX and not os.access(fname, os.R_OK):
-        # on some systems, some importnat programs (e.g. sudo) will have
-        # execute permissions but not read/write permisions. This enables
+        # on some systems, some important programs (e.g. sudo) will have
+        # execute permissions but not read/write permissions. This enables
         # things with the SUID set to be run. Needs to come before _is_binary()
         # is called, because that function tries to read the file.
         return [fname] + args
@@ -348,7 +348,7 @@ def no_pg_xonsh_preexec_fn():
 
 
 class SubprocSpec:
-    """A container for specifiying how a subprocess command should be
+    """A container for specifying how a subprocess command should be
     executed.
     """
 
@@ -381,7 +381,7 @@ class SubprocSpec:
         args : list of str
             Arguments as originally supplied.
         alias : list of str, callable, or None
-            The alias that was reolved for this command, if any.
+            The alias that was resolved for this command, if any.
         binary_loc : str or None
             Path to binary to execute.
         is_proxy : bool
@@ -549,7 +549,7 @@ class SubprocSpec:
 
     def _fix_null_cmd_bytes(self):
         # Popen does not accept null bytes in its input commands.
-        # that doesn;t stop some subproces from using them. Here we
+        # That doesn't stop some subprocess' from using them. Here we
         # escape them just in case.
         cmd = self.cmd
         for i in range(len(cmd)):
@@ -562,8 +562,8 @@ class SubprocSpec:
     @classmethod
     def build(kls, cmd, *, cls=subprocess.Popen, **kwargs):
         """Creates an instance of the subprocess command, with any
-        modifcations and adjustments based on the actual cmd that
-        was recieved.
+        modifications and adjustments based on the actual cmd that
+        was received.
         """
         # modifications that do not alter cmds may come before creating instance
         spec = kls(cmd, cls=cls, **kwargs)
@@ -662,7 +662,7 @@ class SubprocSpec:
         cls = ProcProxyThread if thable else ProcProxy
         self.cls = cls
         self.threadable = thable
-        # also check capturablity, while we are here
+        # also check capturability, while we are here
         cpable = getattr(alias, '__xonsh_capturable__', self.captured)
         self.captured = cpable
 
@@ -781,7 +781,7 @@ def cmds_to_specs(cmds, captured=False):
             specs[-1].background = True
         else:
             raise XonshError('unrecognized redirect {0!r}'.format(redirect))
-    # Apply boundry conditions
+    # Apply boundary conditions
     _update_last_spec(specs[-1])
     return specs
 
@@ -942,7 +942,7 @@ def convert_macro_arg(raw_arg, kind, glbs, locs, *, name='<arg>',
     Parameters
     ----------
     raw_arg : str
-        The str reprensetaion of the macro argument.
+        The str representation of the macro argument.
     kind : object
         A flag or type representing how to convert the argument.
     glbs : Mapping
@@ -966,7 +966,7 @@ def convert_macro_arg(raw_arg, kind, glbs, locs, *, name='<arg>',
     if isinstance(kind, str):
         kind = _convert_kind_flag(kind)
     if kind is str or kind is None:
-        return raw_arg  # short circut since there is nothing else to do
+        return raw_arg  # short circuit since there is nothing else to do
     # select from kind and convert
     execer = builtins.__xonsh_execer__
     filename = macroname + '(' + name + ')'
@@ -1035,9 +1035,9 @@ def call_macro(f, raw_args, glbs, locs):
     f : callable object
         The function that is called as ``f(*args)``.
     raw_args : tuple of str
-        The str reprensetaion of arguments of that were passed into the
+        The str representation of arguments of that were passed into the
         macro. These strings will be parsed, compiled, evaled, or left as
-        a string dependending on the annotations of f.
+        a string depending on the annotations of f.
     glbs : Mapping
         The globals from the call site.
     locs : Mapping or None
@@ -1110,7 +1110,7 @@ def enter_macro(obj, raw_block, glbs, locs):
     raw_block : str
         The str of the block that is the context body.
         This string will be parsed, compiled, evaled, or left as
-        a string dependending on the return annotation of obj.__enter__.
+        a string depending on the return annotation of obj.__enter__.
     glbs : Mapping
         The globals from the context site.
     locs : Mapping or None
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -549,7 +549,7 @@ class SubprocSpec:
 
     def _fix_null_cmd_bytes(self):
         # Popen does not accept null bytes in its input commands.
-        # That doesn't stop some subprocess' from using them. Here we
+        # That doesn't stop some subprocesses from using them. Here we
         # escape them just in case.
         cmd = self.cmd
         for i in range(len(cmd)):
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -992,7 +992,7 @@ def convert_macro_arg(raw_arg, kind, glbs, locs, *, name='<arg>',
         arg = type(execer.eval(raw_arg, glbs=glbs, locs=locs,
                                filename=filename))
     else:
-        msg = ('kind={0!r} and mode={1!r} was not recongnized for macro '
+        msg = ('kind={0!r} and mode={1!r} was not recognized for macro '
                'argument {2!r}')
         raise TypeError(msg.format(kind, mode, name))
     return arg
--- a/xonsh/color_tools.py
+++ b/xonsh/color_tools.py
@@ -397,17 +397,17 @@ def color_dist(x, y):
     return math.sqrt((x[0]-y[0])**2 + (x[1]-y[1])**2 + (x[2]-y[2])**2)
 
 
-def find_closest_color(x, pallette):
-    return min(sorted(pallette.keys())[::-1],
-               key=lambda k: color_dist(x, pallette[k]))
+def find_closest_color(x, palette):
+    return min(sorted(palette.keys())[::-1],
+               key=lambda k: color_dist(x, palette[k]))
 
 
 def make_pallete(strings):
-    """Makes a color pallete from a colection of strings."""
-    pallette = {}
+    """Makes a color palete from a collection of strings."""
+    palette = {}
     for s in strings:
         while '#' in s:
             _, t = s.split('#', 1)
             t, _, s = t.partition(' ')
-            pallette[t] = rgb_to_ints(t)
-    return pallette
+            palette[t] = rgb_to_ints(t)
+    return palette
--- a/xonsh/color_tools.py
+++ b/xonsh/color_tools.py
@@ -9,6 +9,7 @@ import re
 import math
 
 from xonsh.lazyasd import lazyobject, LazyObject
+from xonsh.tools import deprecated
 
 
 RE_BACKGROUND = LazyObject(lambda: re.compile('(BG#|BGHEX|BACKGROUND)'),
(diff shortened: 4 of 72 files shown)


adqm:
This is fantastic; thanks for putting it in!  I'll try go through with a fine-toothed comb sometime in the near future.

Jitsusama:
@adqm, thanks for the thanks! ;)

I realize that this is a large amount of changes. Little did I realize how many i would find when I started down this path. Feel free to give me a poke if you find anything untoward.

I find it quite amusing that codecov is complaining about this set of commits, as I only added one statement. I think this is because ``xonsh.color_tools.make_pallette()`` never included any unit/whatever tests against it.

astronouth7303:
You can ignore codecov and news for this PR. I think because of the amount of heuristic code we have, codecov has some jitter that shows up in these kinds of PRs.

Jitsusama:
Thanks for that info @astronouth7303.

I ended up adding the news when I first saw that check fail, but I couldn't find causality between the codecov report and my changes, so I didn't go any further with that.

astronouth7303:
@laerus codecov doesn't do any processing itself; it just collects information from other CI tools.

laerus:
@astronouth7303 yeah and it fails if a CI fails but doesn't have a way to refresh in case of a restarted and successful CI job.

Jitsusama:
@laerus, thanks for the kudos! And yes, the word environment seemed to be a fan favourite as far as typos are concerned 😄

amalgamate.py
```diff
@@ -386,7 +386,7 @@ def rewrite_imports(name, pkg, order, imps):
             for n in a.names:
                 p, dot, m = n.name.rpartition('.')
                 if p == pkg and m in order:
-                    msg = ('Cannot amalgamate almagate import of '
+                    msg = ('Cannot amalgamate amalgam import of '
```
adqm: To me, it feels like this message should just be "Cannot amalgamate import of ..." rather than "Cannot amalgamate amalgam import of".

xonsh/built_ins.py
```diff
@@ -549,7 +549,7 @@ def xonsh_preexec_fn():
 
     def _fix_null_cmd_bytes(self):
         # Popen does not accept null bytes in its input commands.
-        # that doesn;t stop some subproces from using them. Here we
+        # That doesn't stop some subprocess' from using them. Here we
```
adqm: This should probably be "subprocesses" instead of "subprocess'"

xonsh/color_tools.py
```diff
@@ -397,17 +397,19 @@ def color_dist(x, y):
...
     for s in strings:
         while '#' in s:
             _, t = s.split('#', 1)
             t, _, s = t.partition(' ')
-            pallette[t] = rgb_to_ints(t)
-    return pallette
+            palette[t] = rgb_to_ints(t)
+    return palette
+
+make_pallette = make_palette
```
adqm: The original function was misspelled as `make_pallete` (two Ls, one T).  So this should be `make_pallete` for backward compatibility?  Right now it's got two Ls _and_ two Ts.  Not sure whether the news entry has this same inconsistency.

Then again, I don't know anything about the PTK interface, so is `make_palette` a function we expect people to use directly?  If not, maybe we don't even have to bother with deprecating it (i.e., maybe we should just fix it here and not worry about keeping the typoed version around).  @astronouth7303, @laerus, @gforsyth, @scopatz, any thoughts?

xonsh/history/base.py
```diff
@@ -103,8 +103,8 @@ def append(self, cmd):
         Parameters
         ----------
         cmd: dict
-            A dict contains informations of a command. It should contains
-            the following keys like ``inp``, ``rtn``, ``ts`` etc.
+            A dict contains information on a command. It should contain
+            the following keys, like ``inp``, ``rtn``, ``ts`` &etc.
```
adqm: This sentence still isn't grammatically correct.  Maybe "...should contain the following keys: `inp`, `rtn`, `ts`, etc."  Also if it's a list of the keys that it _should_ contain, shouldn't we list them all?  Anyone know what keys are actually guaranteed to be there?  Maybe we could list all of the keys that are guaranteed to be there, and also all of the keys that _might_ be there (assuming different entries might not all have the same keys).

The ampersand also isn't necessary.  _et cetera_ should usually be abbreviated as "&c" or "etc", since the ampersand represents the _et_.

xonsh/parsers/base.py
```diff
@@ -2098,13 +2098,13 @@ def _attach_nocomma_tok_rules(self):
     # The following grammar rules are no-ops because we don't need to glue the
     # source code back together piece-by-piece. Instead, we simply look for
     # top-level commas and record their positions. With these positions and
-    # the bounding parantheses !() positions we can use the source_slice()
+    # the bounding parenthesis !() positions we can use the source_slice()
```
adqm: I think this wants to be plural possessive (i.e., "the bounding parentheses' respective positions").

xonsh/proc.py
```diff
@@ -220,8 +220,8 @@ def __init__(self, fd, timeout=None):
 
 def populate_buffer(reader, fd, buffer, chunksize):
     """Reads bytes from the file descriptor and copies them into a buffer.
-    The reads happend in parallel, using pread(), and is thus only
-    availabe on posix. If the read fails for any reason, the reader is
+    The reads happened in parallel, using pread(), and is thus only
```
adqm: This sentence confuses me.  Maybe this should be something like the following, instead?

"The reads happen in parallel using the pread() syscall, which is only available on POSIX systems."

xonsh/wizard.py
```diff
@@ -114,7 +114,7 @@ def __init__(self, prompt='>>> ', converter=None, show_conversion=False,
 class While(Node):
     """Computes a body while a condition function evaluates to true.
     The condition function has the form cond(visitor=None, node=None) and
-    should return an object that is convertable to a bool. The beg attribute
+    should return an object that is convertible to a bool. The beg attribute
```
adqm: This phrasing feels awkward to me.  Maybe "...should return an object that can be converted to a bool."?

xonsh/wizard.py
```diff
@@ -128,7 +128,7 @@ def __init__(self, cond, body, idxname='idx', beg=0, path=None):
             Function that determines if the next loop iteration should
             be executed. The condition function has the form
             cond(visitor=None, node=None) and should return an object that
-            is convertable to a bool.
+            is convertible to a bool.
```
adqm: Same as above.

xonsh/color_tools.py
astronouth7303: If we're being strict, this should be flagged as a deprecation with warnings, etc etc.

Honestly, there are probably like 2 places in the world third party code uses this. Call it out in a news item, and either remove this or leave comment saying when it should be removed.

amalgamate.py
Jitsusama: Yes, the sentence is definitely a bit awkward. There are a number of sentences that felt a little bit off, but I didn't want to risk offending someone with my own grammatical style, or, worse yet, ruin the sentence by introducing technical inaccuracies.

Now, that said, I don't know the code well enough to authoritatively say that the sentence as is (with the word amalgam instead of almagate) is not actually accurate. Is the code performing a super-amalgamate function, such that it is amalgamating an amalgamated bundle of code (just writing that out makes me realize how easy it is to misspell amalgamate 😉 )? If so, then the sentence might be better left alone. If you believe the sentence is more accurate with that word removed, I definitely don't mind including that in this PR. I just don't want to make

xonsh/built_ins.py
Jitsusama: I agree. I can make that change.

xonsh/color_tools.py
Jitsusama: I left a comment in the news item stating that it is deprecated. That said, is there some sort of special decorator or other method in Python that allows you to mark something as deprecated? I believe I saw an exception for this at some point, but I've never dug into it.

xonsh/history/base.py
Jitsusama: I do believe that a colon or semicolon would be better here.

Also, I learned something new about etc today. I did some internet searching and found out that I've been using it wrong all of my life. Thanks! 😆 

I'll make those changes.

xonsh/parsers/base.py
Jitsusama: So, are you thinking of it being written as "the bounding parenthesis' !() positions...", or are you thinking of removing !() from the text?

xonsh/proc.py
Jitsusama: I wasn't paying much attention to tense and grammar as I went through, just the full list of misspellings. That said, looking over the sentence now, it definitely could use some work. I do think your sentence is better all around. I can make this change as well.

xonsh/wizard.py
Jitsusama: As far as technical accuracy, shouldn't this be more of a statement than a question? IE: "The condition function has the `form cond(visitor=None, node=None)` and must return an object that responds to the Python magic method `__bool__`"

xonsh/wizard.py
Jitsusama: Same as above for me too! 😉

xonsh/color_tools.py
Jitsusama: Oh, @adqm, I just realized what you meant about the spelling. It turns out that I misspelled an intentional misspelling! 😆 

Let me know how you would like for me to handle this. If we are to keep it in, in some form or another, I'll make sure to keep the naming consistent with what it was before.

amalgamate.py
laerus: +1 adqm

adqm: I think removing it is more correct.

xonsh/color_tools.py
adqm: Yes, okay.  Let's leave it marked as deprecated, and "correct" the misspelling so that it reads `make_pallete = make_palette`.  Can you also add a comment TODO: remove in 0.6.0 or similar?

Then we'll have to remember in 0.6 to remove it.

xonsh/parsers/base.py
adqm: I would remove the `!()` (I'm not sure it actually adds any real clarity).

I would say:

"With these positions and the respective positions of the bounding parentheses, we can use the source_slice()..."

xonsh/wizard.py
adqm: Right, sorry for the confusion.  I didn't intend to suggest that it should be a question.  Your phrasing sounds fine.

amalgamate.py
Jitsusama: Sounds good, that's what I'll do.

xonsh/color_tools.py
Jitsusama: Sounds good. Would you like for me to investigate how to cause static analysis to report usage as deprecated? I believe there is a way to do this, as pylama has started giving me Deprecation errors for third party libraries on the latest release.

xonsh/parsers/base.py
Jitsusama: Sounds good.

xonsh/wizard.py
Jitsusama: No worries. I'll put in my text.

Jitsusama:
It sounds like @adqm and I are on the same page on a number of changes. I'm going to start working on these now.

I should note, it did become apparent during our discussions that there might be a number of more grammatical errors throughout the source code and documentation. I'm not promising to do this, but, would there be interest in having a more full audit performed on the grammar of the source?

xonsh/history/base.py
Jitsusama: I was thinking about this a bit more and I was wondering what you thought of this text:

> This dict contains information about a command. It should contain the following keys; ``inp``, ``rtn``, ``ts`` and others as needed.

xonsh/color_tools.py
Jitsusama: Ok, I've thought about this and did a little bit of research. It looks like raising a `DeprecationWarning` will not cause an actual exception unless Python is running with the `-W` flag or the `PYTHONWARNINGS` environment variable is set. With that in mind, what do you think of the following bit of code?

```python
# TODO: deprecated in 0.5.10, remove in 0.6.0.
def make_pallete(*args, **kwargs):
    """Makes a color palette from a collection of strings.

    Deprecated since version 0.5.10. Use make_palette instead.
    """
    make_palette(*args, **kwargs)
    raise DeprecationWarning(
        "make_pallete has been deprecated in version 0.5.10 and will be "
        "removed in 0.6.0. Please use make_palette instead.")
```

Jitsusama: Actually, it looks like the following might be the more Pythonic way of doing this:

```python
# TODO: deprecated in 0.5.10, remove in 0.6.0.
def make_pallete(*args, **kwargs):
    """Makes a color palette from a collection of strings.

    Deprecated since version 0.5.10. Use make_palette instead.
    """
    import warnings

    make_palette(*args, **kwargs)

    warnings.warn(
        "make_pallete has been deprecated in version 0.5.10 and will be "
        "removed in 0.6.0. Please use make_palette instead.",
        DeprecationWarning)
```

Jitsusama:
I've thrown in the changes that we seemed to be in agreement on. There are 2 more changes that I have questions on. I'll hold on to them until we have a consensus on them.

xonsh/color_tools.py
astronouth7303: Can you make that a generic wrapper/decorator? (Does one already exist somewhere?)

Jitsusama: There is a third party API for this, but not one that I could find builtin.

The [deprecation](https://pypi.python.org/pypi/deprecation) library on PyPI provides a decorator for this. I didn't think you'd want to support a third party library just for this, but it does seem to be a nice library for encapsulating this logic. It even updates the docstring to take note of the deprecation information.

Or, are you asking if I could write a custom decorator for this purpose? That should be pretty easy. If I did, where would you like for it to appear? In a separate module?

astronouth7303: This should be a utility decorator. It should go either in some kind of utilities/decorators module, or we should vendor the `deprecation` library.

Thoughts on which?

Jitsusama: Is vendoring what is responsible for the low test coverage? I'd hate to vendor something just to see the project's code coverage get slammed.

Jitsusama: I'm working on my own decorator now. I'll post a snippet up in a second so you can let me know what you think of it.

adqm: I think making a decorator sounds like a good plan, and putting it in `tools.py` might be fine.

Jitsusama: Nifty. I'm thinking of making 2 decorators, `deprecated` and `replaced_by`. I have `deprecated` pretty much done, and I'll have `replaced_by` call `deprecated`. We'll use `replaced_by` for this and will have `deprecated` available in case we ever want to gracefully deprecate a function in the future.

xonsh/history/base.py
adqm: I don't actually know any of the details about the history setup here, so I'm not sure what's optional and what's required.  And the sentence currently doesn't actually describe what any of those things _mean_.

Maybe:

"This dict contains information about a command.  It should contain keys `'inp'`, `'rtn'`, `'out'`, and `'ts'`, which have the same forms and meanings as the corresponding instance variables of the `HistoryEntry` class."

It's a pretty ugly sentence, but it is a bit more descriptive.

adqm:
> I'm not promising to do this, but, would there be interest in having a more full audit performed on the grammar of the source?

I think that would be great (though I'm certainly not volunteering for it, either :smile:).  Docs are often neglected because they're no fun, but I imagine there are grammatical issues all over the docs and the API reference...

Jitsusama:
@adqm, that's good to know. Changing grammar can be prone to bruised egos, so it might not be for the thin skinned. I'll think about it though, as my skin is quite thick 😉. The other problem is how subjective grammar can be.

xonsh/color_tools.py
Jitsusama: How does this look?

```python
import functools
import warnings


def deprecated(deprecated_in=None, removed_in=None):
    """Deprecate a function in a graceful manner.

    Updates the decorated function's docstring to mention the version
    that deprecation occurred in and the version it will be removed
    in if both of these values are passed.

    In all circumstances, call ``warnings.warn`` with details, while
    raising ``DeprecationWarning``.

    Usage:

        >>> @deprecated(deprecated_in='0.5.10', removed_in='0.6.0')
        ... def my_function(thing1, thing2):
        ...     print(thing1, thing2)
        >>> my_function('hello', 'world')
        hello world
        >>> ('my_function has been deprecated in version 0.5.10 and will be '
        ...  'removed in version 0.6.0' == my_function.__doc__)
        True
    """
    message_suffix = _message_suffix(deprecated_in, removed_in)
    if not message_suffix:
        message_suffix = ''

    def decorated(func):
        warning_message = '{} has been deprecated'.format(func.__name__)
        warning_message += message_suffix

        @functools.wraps(func)
        def wrapped(*args, **kwargs):
            func(*args, **kwargs)
            warnings.warn(warning_message, DeprecationWarning)

        wrapped.__doc__ = (
            wrapped.__doc__ + '\n\n' + warning_message
            if wrapped.__doc__ else warning_message)

        return wrapped
    return decorated


def _message_suffix(deprecated_in, removed_in):
    if deprecated_in and removed_in:
        message_suffix = (
            ' in version {} and will be removed in version {}'.format(
                deprecated_in, removed_in))
    elif deprecated_in and not removed_in:
        message_suffix = (
            ' in version {}'.format(deprecated_in))
    elif not deprecated_in and removed_in:
        message_suffix = (
            ' and will be removed in version {}'.format(removed_in))
    else:
        message_suffix = None

    return message_suffix
```

astronouth7303: Add it to the PR so we can review it properly.

Jitsusama: Will do.

Jitsusama: It has been added as commit 5c33aac, please let me know what you think.

xonsh/timings.py
```diff
@@ -211,7 +211,7 @@ def timeit_alias(args, stdin=None):
             worst = max(worst, worst_tuning)
         # Check best timing is greater than zero to avoid a
         # ZeroDivisionError.
-        # In cases where the slowest timing is lesser than 10 micoseconds
+        # In cases where the slowest timing is lesser than 10 microseconds
```
gforsyth: lesser -> less

xonsh/base_shell.py
```diff
@@ -506,15 +506,15 @@ def prompt(self):
         return p
 
     def format_color(self, string, hide=False, force_string=False, **kwargs):
-        """Formats the colors in a string. This base implmentation colors based
-        on ANSI color codes
+        """Formats the colors in a string. This base implementation's colors are
```
laerus: maybe *The colors of this implementation*

xonsh/color_tools.py
Jitsusama: FYI; I've released another commit that extends the functionality of the `@deprecated` decorator. It will cause an `AssertionError` to be raised whenever an `@deprecated` decorated function is called when its `removed_in` version is equal to or less than the current `xonsh.__version__`.

xonsh/main.py
```diff
@@ -170,8 +170,8 @@ def parser():
                    choices=('readline', 'prompt_toolkit', 'best', 'random'),
                    default=None)
     p.add_argument('--timings',
-                   help='Prints timing infomation before the prompt is shown. '
-                        'This is usefull to track down perfomance issues '
+                   help='Prints timing information before the prompt is shown. '
+                        'This is usefull to track down performance issues '
```
Jitsusama: Why oh why did my spell checker not catch that?! I will resolve this post haste!

xonsh/parsers/base.py
```diff
@@ -190,7 +190,7 @@ def empty_list_if_newline(x):
 
 def lopen_loc(x):
     """Extracts the line and column number for a node that may have anb opening
```
Jitsusama: My spell checker failed me again 😢

xonsh/proc.py
```diff
@@ -1408,7 +1410,7 @@ def wait(self, timeout=None):
 
     def _signal_int(self, signum, frame):
         """Signal handler for SIGINT - Ctrl+C may have been pressed."""
-        # check if we have already be interrupted to prevent infintie recurrsion
+        # check if we have already be interrupted to prevent infinite recursion
```
Jitsusama: At least this is a grammatical error and not another spelling error 😄 

I'll fix this too!

xonsh/pyghooks.py
```diff
@@ -360,7 +360,7 @@ def _partial_color_tokenize_main(template, styles):
 
 
 class CompoundColorMap(MutableMapping):
-    """Looks up color tokes by name, potentailly generating the value
+    """Looks up color tokes by name, potentially generating the value
```
Jitsusama: You mean it's not talking about taking a drag off a cigarette? 😉 

I'll take care of this too!

xonsh/pyghooks.py
```diff
@@ -1399,13 +1399,13 @@ def _xcode_style():
      _vim_style, _vs_style, _xcode_style)
 
 
-# dynamic syles
-def make_pygments_style(pallette):
-    """Makes a pygments style based on a color pallete."""
+# dynamic styles
+def make_pygments_style(palette):
+    """Makes a pygments style based on a color palete."""
```
Jitsusama: That word is the bane of my existence.

xonsh/timings.py
Jitsusama: 👍

xonsh/history/base.py
Jitsusama: I gave it another shot, using your rephrased sentence as a base, and further descriptions I drudged up from the `JsonHistory` and `HistoryEntry` classes. Since `JsonHistory` does not inherit the `append()` doctoring from `History`, I've updated it as well. Let me know if you'd still like to give it another round.

xonsh/base_shell.py
Jitsusama: Since this method gets inherited by other classes which seem to change the colour code standard that they receive as input, it does seem important to differentiate here. I think the word "this" could be changed to "the" though.

astronouth7303:
Stylistically, we've traditionally kept news bullets to one-line. I'm not sure if I'm for or against the more verbose items.

Jitsusama:
@astronouth7303, let me know what you prefer and I'll accomodate. If you would like them one-lined, please let me know the character limit you'd like per line.

Jitsusama:
@astronouth7303, I'm sure you're quite busy, but could you let me know if you'd like me to change my news entry?

astronouth7303:
I'm not sure. I was hoping some of the other xore member might chime in?

The problem is that the information is important and should go _somewhere_, but I'm not sure where I would rather it go.

Jitsusama:
Ok. I'll put this out of my mind while ya'll deliberate. Just give me a poke once you reach a decision.

laerus:
i don't mind, changelog verbosity is not a bad thing.
can we merge this?

Jitsusama:
I'm definitely cool with it getting merged ;)

astronouth7303:
Thank you, @Jitsusama, for sticking with this! I know we asked a lot of you.

Jitsusama:
Hey @astronouth7303, no worries, I enjoyed working with all of you to make this happen!

PR #2402, opened 2017-06-12
gistatus: Fix hash not being shown when in detaced HEAD and there are…

nh2:
… no tags.

The `git describe --exact-match` used until fails with
  fatal: No names found, cannot describe anything.
if there are no tags at all in the repository.
In that case, it exits with exit code 128, and thus the
control flow was aborted, and the fallback `git rev-parse`
was not executed.

As a workaround, `git describe --always` can be used,
which achieves the intended result without increasing the
number of `git` invocations.
See also:
  https://stackoverflow.com/questions/4916492/git-describe-fails-with-fatal-no-names-found-cannot-describe-anything

### DIFF
--- a/news/fix-gistatus-for-detached-head-no-tags.rst
+++ b/news/fix-gistatus-for-detached-head-no-tags.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks again @nh2!

PR #2406, opened 2017-06-17
gistatus: Don't SIGKILL git to avoid leftover .git/index.lock.

nh2:
Instead, ask it nicely to terminate using SIGTERM.

### DIFF
--- a/xonsh/prompt/gitstatus.py
+++ b/xonsh/prompt/gitstatus.py
@@ -17,11 +17,33 @@ GitStatus = collections.namedtuple('GitStatus',
 
 def _check_output(*args, **kwargs):
     kwargs.update(dict(env=builtins.__xonsh_env__.detype(),
+                       stdout=subprocess.PIPE,
                        stderr=subprocess.DEVNULL,
-                       timeout=builtins.__xonsh_env__['VC_BRANCH_TIMEOUT'],
                        universal_newlines=True
                        ))
-    return subprocess.check_output(*args, **kwargs)
+    timeout = builtins.__xonsh_env__['VC_BRANCH_TIMEOUT']
+    # See https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate
+    with subprocess.Popen(*args, **kwargs) as proc:
+        try:
+            out, err = proc.communicate(timeout=timeout)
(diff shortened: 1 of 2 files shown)


xonsh/prompt/gitstatus.py
```diff
@@ -17,11 +17,27 @@
...
+                raise subprocess.CalledProcessError(proc.returncode, proc.args, output=out, stderr=err)  # note err will always be empty as we redirect stderr to DEVNULL abvoe
+            return out
+        except subprocess.TimeoutExpired:
+            # We use `.terminate()` (SIGTERM) instead of `.kill()` (SIGKILL) here
+            # because otherwise we guarantee that a `.git/index.lock` file will be
+            # left over, and subsequent git operations will fail.
+            # We don't want that.
+            # As a result, we must rely on git to exit properly on SIGTERM.
+            proc.terminate()
+            proc.wait()  # we ignore what git says after we sent it SIGTERM
```
astronouth7303: Do we need to wait on the process here? Can we just assume it'll clean itself up?

astronouth7303:
Thank you! This should fix a long-time bug that was very annoying.

xonsh/prompt/gitstatus.py
nh2: We need to wait, because otherwise `index.lock` will still exist, and so the next git operation (e.g. when you press Enter again, forcing xonsh to try generate a new gistatus) will fail with the `index.lock` error.

I believe that this is OK also from a user experience point of view, because git will exit quickly on SIGTERM. If it doesn't, it's a bug, and we have to know about it.

---

Technically, we don't even have to write `proc.wait()` here, because the `with Popen` context manager wil call `wait()` anyway when that block is left, but I wrote it anyway in order to be explicit about the waiting. I'll add a comment to that regard.

scopatz:
This is a really nice fix to have @nh2! Thanks for putting it in.  Is there any way that you can add a news entry, please? http://xon.sh/devguide.html#changelog Thanks!

nh2:
> Is there any way that you can add a news entry, please?

Done.

scopatz:
Thanks @nh2!

Issue #2403, opened 2017-06-12
How to jump across words with Ctrl+Left/Right


nh2:
In bash, zsh and fish I can use Ctrl+Left or Ctrl+Right to jump across whole words, and Alt+Left or Alt-Right to jump around inside words.

For example, for the contents of the prompt `hello you-person` with the caret at the end, Ctrl+Left would jump across `you-person` and Alt-Left would jump across `person`.

How to achieve the same in xonsh?

I imagine this needs some `prompt_toolkit` configuration but I can't quite figure out what I have to do.

nh2:
OK, I figured it out:

The `find_start_of_previous_word()` function in `python-prompt-toolkit` has an argument `WORD=False` that if set to `True` jumps over whole words, like I want it. See its [code here](https://github.com/jonathanslenders/python-prompt-toolkit/blob/46d07b00530e76eee8126d38ecb3f9c043bf2eae/prompt_toolkit/document.py#L421).

The code for [`forward-word`](https://github.com/jonathanslenders/python-prompt-toolkit/blob/46d07b00530e76eee8126d38ecb3f9c043bf2eae/prompt_toolkit/key_binding/bindings/named_commands.py#L84) doesn't use that.

I could add this feature to xonsh by putting into `.xonshrc`:

```python
# Key bindings for jumping over whole words (everything that's not
# white space) using Ctrl+Left and Ctrl+Right;
# Alt+Left and Alt+Right still jump over smaller word segmetns.
# See https://github.com/xonsh/xonsh/issues/2403

@events.on_ptk_create
def custom_keybindings(bindings, **kw):
    handler = bindings.registry.add_binding

    @handler(Keys.ControlLeft)
    def ctrl_left(event):
        buff = event.current_buffer
        pos = buff.document.find_previous_word_beginning(count=event.arg, WORD=True)
        if pos:
            buff.cursor_position += pos

    @handler(Keys.ControlRight)
    def ctrl_right(event):
        buff = event.current_buffer
        pos = buff.document.find_next_word_ending(count=event.arg, WORD=True)
        if pos:
            buff.cursor_position += pos
```

Could we add this as some easy-to-use setting to xonsh somehow?

E.g. as some recipe in the documentation, or should it be a setting or a `xontrib`?

gforsyth:
[...]

PR #2407, opened 2017-06-18
Add whole_word_jumping xontrib. Implements #2403

nh2:


xontrib/whole_word_jumping.py
```diff
@@ -0,0 +1,30 @@
+"""Jumping across whole words (non-whitespace) with Ctrl+Left/Right.
+
+Alt+Left/Right remains unmodified to jump over smaller word segments.
+"""
+from prompt_toolkit.keys import Keys
+
```
astronouth7303: Please add an empty `__all__` to not polute the shell context.

astronouth7303:
This is solid work. Thank you!

For the record, I debated requesting this be an external xontrib, but I think this is pretty useful, and is also a fine example of working with ptk in xontribs.

xontrib/whole_word_jumping.py
nh2: Done.

nh2:
> EDIT: Please do include a "new xontrib" news item.

Done.

scopatz:
This looks good to me!

astronouth7303:
What's up with the failing build? @gforsyth is this docs related?

gforsyth:
Seems to be similar to the problem in #2408 -- something in the install is touching `xonsh/__init__.py` -- is there something added to amalgamation that we aren't aware of? 

And if there is, we _really_ need a test that isn't a failing docs build to catch that...

scopatz:
I think we can go ahead and merge, right?

astronouth7303:
I think so. The failed test is Travis's Linux/Py3.6, which I believe is also covered by CircleCI.

PR #2408, opened 2017-06-18
Implement an alias alias, matching what sh's alias does.

nathan-hoad:
I thought this might be useful for others as a means to minimise typing, e.g. `alias foo="echo hey"` instead of `aliases['foo'] = ['echo', 'hey']`. I find myself creating aliases a lot, so this is something I find pretty useful.

### DIFF
--- a/news/alias-command.rst
+++ b/news/alias-command.rst
@@ -1,0 +1,13 @@
+**Added:**
+
+* Added an alias command, matching bash's implementation, available as part of bashisms.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/xontrib/bashisms.py
+++ b/xontrib/bashisms.py
@@ -1,6 +1,11 @@
 """Bash-like interface extensions for xonsh."""
+import shlex
+import sys
+
 from prompt_toolkit.keys import Keys
 from prompt_toolkit.filters import Condition, EmacsInsertMode, ViInsertMode
+
+__all__ = ()
 
 
 @events.on_transform_command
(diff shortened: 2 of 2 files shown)


gforsyth:
Hey @nathan-hoad -- thanks for putting this in! I'm all for making people more comfortable when switching shells.  Rather than putting this in xonsh core, I think this would be a great addition to the bundled bashisms `xontrib`.  Or maybe we should have a sh `xontrib`?

nathan-hoad:
Good point! I'll move it into bashisms, that's a much more sensible place for it. Thanks for taking a look!

scopatz:
Hi @nathan-hoad - I agree that bashisms is the right place for this to live. Thanks for putting it in.

It looks like doctr is failing. Is this expected right now @gforsyth?

xonsh/aliases.py
```diff
@@ -423,9 +423,32 @@ def detect_xip_alias():
         return basecmd
 
 
+def alias(args, stdin=None):
```
scopatz: Probably should change the signature here to add `stdout=None, stderr=None` so that...

xonsh/aliases.py
```diff
@@ -423,9 +423,32 @@ def detect_xip_alias():
...
+        for arg in args:
+            if '=' in arg:
+                # shlex.split to remove quotes, e.g. "foo='echo hey'" into
+                # "foo=echo hey"
+                name, cmd = shlex.split(arg)[0].split('=', 1)
+                builtins.aliases[name] = shlex.split(cmd)
+            elif arg in builtins.aliases:
+                print('{}={}'.format(arg, builtins.aliases[arg]))
+            else:
+                print("alias: {}: not found".format(arg), file=sys.stderr)
```
scopatz: ... we can set `file=stderr` here, instead of `sys.stderr`

asmeurer:
Doctr failure is suggesting that somehow your docs build created an untracked change in `xonsh/__init__.py`, which prevents git from checking out the `gh-pages` branch of `xonsh-docs`.

nathan-hoad:
I've also thrown in a `__all__ = ()` into bashisms, so that it doesn't leak the objects into the shell. If you'd rather a separate PR for that, let me know.

gforsyth:
Hey @nathan-hoad -- thanks a ton! And definitely ok to stop the bashisms leaking everywhere.

PR #2412, opened 2017-06-22
Add $APPEND_IF_NO_NEWLINE environment variable

### LINKED ISSUE #448: Prompt displayed on the same line as command output
Consider the following code:

```
echo -n foo
```

. This will result in

```
foozyx@zyx-desktop ~ $
```

on the next line in place of just

```
zyx@zyx-desktop ~ $
```

. Zsh in this case outputs per cent sign:

highlighted per cent after foo

(note: attaching a screenshot to show that this sign is highlighted). (Of course, in zsh this is controlled by some option. I do not see why this should be optional though, but in zsh you can customize not only sign presence, but also how it is highlighted and what text should be there exactly (including running some command and displaying its output).)

Xonsh version 0.2.2.

rossnomann:
Closes #448

### DIFF
--- a/xonsh/base_shell.py
+++ b/xonsh/base_shell.py
@@ -339,8 +339,12 @@ class BaseShell(object):
                 hist.last_cmd_rtn = 1  # return code for failure
         finally:
             ts1 = ts1 or time.time()
-            self._append_history(inp=src, ts=[ts0, ts1], tee_out=tee.getvalue())
+            tee_out = tee.getvalue()
+            self._append_history(inp=src, ts=[ts0, ts1], tee_out=tee_out)
             self.accumulated_inputs += src
+            append_newline = env.get('XONSH_APPEND_NEWLINE')
+            if append_newline and not tee_out.endswith(os.linesep):
+                print(os.linesep, end='')
             tee.close()
             self._fix_cwd()
         if builtins.__xonsh_exit__:  # pylint: disable=no-member
--- a/news/issue-448-add-xonsh_append_newline-var.rst
+++ b/news/issue-448-add-xonsh_append_newline-var.rst
@@ -1,0 +1,13 @@
+**Added:**
+
+* Added ``$XONSH_APPEND_NEWLINE`` environment variable
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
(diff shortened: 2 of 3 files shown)


laerus:
Thanks for adding this feature!!

A better name would be `XONSH_APPEND_NEWLINE` or something like this but with the `XONSH_` prefix.

scopatz:
Thanks for putting this in @rossnomann! Please let us know when it is ready for another look!

rossnomann:
@scopatz @laerus I've updated PR.

scopatz:
Hi @rossnomann - thanks for putting this in! I tend agree with @laerus, though, that `$XONSH_APPEND_NEWLINE` is a better name.  Any reason not to change it?

rossnomann:
Hi @scopatz  Well, I'll rename it - I don't see the difference.

scopatz:
Adding XONSH means that it shouldn't collide with other variables in the environment.

rossnomann:
@scopatz I know. I meant APPEND_**IF_NO**_NEWLINE vs APPEND_NEWLINE.

scopatz:
Oh, it is generally a good idea to keep flag names positive, since some people find them easier to reason about.

rossnomann:
Good point :+1: . I already heard about it, but cant remember. Thank you!

xonsh/base_shell.py
```diff
@@ -339,8 +339,12 @@ def default(self, line):
                 hist.last_cmd_rtn = 1  # return code for failure
         finally:
             ts1 = ts1 or time.time()
-            self._append_history(inp=src, ts=[ts0, ts1], tee_out=tee.getvalue())
+            tee_out = tee.getvalue()
+            self._append_history(inp=src, ts=[ts0, ts1], tee_out=tee_out)
             self.accumulated_inputs += src
+            append_newline = env.get('XONSH_APPEND_NEWLINE')
+            if append_newline and not tee_out.endswith(os.linesep):
+                print(append_newline, end='')
```
laerus: isn't this suppose to print a new line and not the envvar?

laerus: if you want to print arbitary text after a command you can always use the on_post_cmd event.

rossnomann: > isn't this suppose to print a new line and not the envvar?
We need to print the envvar, of course.
> you can always use the on_post_cmd event
Can I get previous output and check for os.linesep at the end in this case?

rossnomann: Can't find `on_post_cmd` event.

laerus: sorry it is `events.on_postcommand`, yeah you can check various attributes of the last command.
try:
`print(events.on_postcommand.__doc__)`

i think we should just print a newline and leave more complicated cases for the events

rossnomann: Ok. So, I should print `os.linesep` right here, right?

rossnomann:
@laerus I've updated PR

laerus:
thanks @rossnomann !!!

PR #2415, opened 2017-06-24
pass env into git branch detection

scopatz:
Pretty minor change, but init file is coming along for the ride.


scopatz:
Note that this fixes the docs build

PR #2419, opened 2017-06-30
Fire on_pre_prompt before any prompt calculations are done.

astronouth7303:
Under PTK, any changes made to `$PROMPT` in an `on_pre_prompt` handler take effect with a one-prompt delay.

This moves when `on_pre_prompt` is fired to before any prompt calculations are made.

The readline shell already behaved this way (all of its logic being in `.prompt`). The base shell doesn't raise the event at all.

### DIFF
--- a/news/compute-prompt-after-event.rst
+++ b/news/compute-prompt-after-event.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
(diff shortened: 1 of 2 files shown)


laerus:
lgmt! that codecov though, should take a look at conf again.

PR #2424, opened 2017-07-06
Fix incorrect branch colorization

moigagoo:
If 'git' or 'hg' commands are aliases, repository state was determined incorrectly.

This commit adds explicit alias ignoring.


astronouth7303:
Your news item needs to have a line between the heading and the item. There should be plenty of examples in the directory.

laerus:
thanks for this @moigagoo !!

PR #2434, opened 2017-07-15
Don't load xonshrc is xonsh is running a script

gforsyth:
As per discussions with @asmeurer -- if xonsh is being used to launch a script,
e.g.

```
xonsh run_my_cool_deploy.xsh
```

then it should not load the `xonshrc` file since this should be reserved for
interactive functionality and we don't want side effects in the script.

Basically, if you need a xontrib to be loaded for your script, load it in your
script

### DIFF
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -236,6 +236,9 @@ def start_services(shell_kwargs):
     env = builtins.__xonsh_env__
     rc = shell_kwargs.get('rc', None)
     rc = env.get('XONSHRC') if rc is None else rc
+    if shell_kwargs['shell_type'] != 'none':
+        #  Don't load xonshrc if not interactive shell
+        rc = None
     events.on_pre_rc.fire()
     xonshrc_context(rcfiles=rc, execer=execer, ctx=ctx, env=env, login=login)
     events.on_post_rc.fire()
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -218,7 +218,7 @@ class XonshMode(enum.Enum):
     interactive = 3
 
 
-def start_services(shell_kwargs):
+def start_services(shell_kwargs, args):
     """Starts up the essential services in the proper order.
     This returns the environment instance as a convenience.
     """
@@ -236,7 +236,7 @@ def start_services(shell_kwargs):
     env = builtins.__xonsh_env__
     rc = shell_kwargs.get('rc', None)
     rc = env.get('XONSHRC') if rc is None else rc
-    if shell_kwargs['shell_type'] != 'none':
+    if args.mode != XonshMode.interactive:
         #  Don't load xonshrc if not interactive shell
         rc = None
     events.on_pre_rc.fire()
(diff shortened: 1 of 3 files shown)


xonsh/main.py
```diff
@@ -236,6 +236,9 @@ def start_services(shell_kwargs):
     env = builtins.__xonsh_env__
     rc = shell_kwargs.get('rc', None)
     rc = env.get('XONSHRC') if rc is None else rc
+    if shell_kwargs['shell_type'] != 'none':
```
scopatz: Can this not check `$XONSH_INTERACTIVE` for some reason?

gforsyth: We set `XONSH_INTERACTIVE` in `premain` after we've returned the `env` from this function, so I just went with this -- can also rework `premain` a bit

melund:
Seems reasonable to me. Is there a shortcut to loading the `.xonshrc` file in a script in an easy way?

gforsyth:
Yeah, I need to tweak this so that a user can force using xonshrc by
specifying interactive mode

laerus:
@melund you can just `source .xonshrc` from xonsh scripts

gforsyth:
Hey folks -- so I'm going to have to tweak some of the tests since they aren't necessarily aligned with what I _think_ is the correct behavior here, but I wanted to ask some other members of xore what they think should happen re: loading rc files or not.

My take:
- if xonsh is used to run a `xsh` file, it should ignore any rc files, unless the `-i` flag is passed to explicitly force interactive mode

Question:
- if xonsh is used to run a `xsh` file and the user specifies a custom rc file using `--rc`, should we default to interactive mode since the rc file has been explicitly given?  Or require them to also pass in `-i`?

astronouth7303:
How do we handle things like path additions? If I have eg `~/.local/bin`, gem directories, etc, where do I set those?

asmeurer:
The xonsh process should inherit the environment from wherever you run it, so if your PATH is already set in your shell, it shouldn't be an issue.

scopatz:
@gforsyth 

> Or require them to also pass in -i?

I think we should do this one under the assumption that all that `--rc` does is specify which rc file to use, not whether or not it should be used.

gforsyth:
Finally got back around to this.  Rebased on to master.

Calling `xonsh my_script.xsh` will not load any rc file
Calling `xonsh --rc myrc my_script.xsh` will also not load any rc file
Calling `xonsh -i my_script.xsh` will run the script using the default rc files
Calling `xonsh -i --rc myrc my_script.xsh` will run the script using the `myrc` rc file

I've also changed the envvar `XONSH_INTERACTIVE` to reflect `True` if _either_ the `-i` flag is set or if we infer `XonshMode.interactive` during startup (this was a red herring in testing, since previously it only reflected the status of the force flag)

gforsyth:
pinging on this, it's ready for another look.

scopatz:
Thanks @gforsyth!

PR #2438, opened 2017-07-16
Ipython threadable

jb-leger:
`IPython` is considered as threadable by the default predictor, and strange behavior occur when the terminal is resized, in particular when `IPython` is using `ptk`.


scopatz:
Whoops! Sorry @melund...

scopatz:
This should be predict shell, probably.  Can you please put in another PR @jb-leger :) Thanks!

scopatz:
Fixed in #2440

PR #2440, opened 2017-07-16
ipython not threadable: predict_shell should be used

jb-leger:
See #2438. Use `predict_shell` instead of `predict_help_ver`. Sorry for this mistake on my last commit.

I didn't add new item in changes, I think it is not necessary.


scopatz:
No news item is needed here because it was in the previous PR

Issue #2436, opened 2017-07-15
""" -> command not found


asmeurer:
```
$ """
xonsh: subprocess mode: command not found:

```

It should start a docstring, like in Python.

scopatz:
This produces the following traceback for me:

```
scopatz@artemis ~ $ """
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/home/scopatz/.local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 19854, in _parse_ctx_free
    debug_level=(self.debug_level > 2))
  File "/home/scopatz/.local/lib/python3.6/site-packages/xonsh/parsers/base.py", line 348, in parse
    tree = self.parser.parse(input=s, lexer=self.lexer, debug=debug_level)
  File "/home/scopatz/miniconda/lib/python3.6/site-packages/ply/yacc.py", line 331, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/home/scopatz/miniconda/lib/python3.6/site-packages/ply/yacc.py", line 1199, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/home/scopatz/miniconda/lib/python3.6/site-packages/ply/yacc.py", line 193, in call_errorfunc
    r = errorfunc(token)
  File "/home/scopatz/.local/lib/python3.6/site-packages/xonsh/parsers/base.py", line 2723, in p_error
    column=p.lexpos))
  File "/home/scopatz/.local/lib/python3.6/site-packages/xonsh/parsers/base.py", line 481, in _parse_error
    raise err
  File "<string>", line None
SyntaxError: <xonsh-code>:1:0: EOF in multi-line string
"""
^

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/scopatz/.local/lib/python3.6/site-packages/xonsh/ptk/shell.py", line 137, in _push
    locs=None)
  File "/home/scopatz/.local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 19798, in compile
    transform=transform)
  File "/home/scopatz/.local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 19767, in parse
    tree, input = self._parse_ctx_free(input, mode=mode, filename=filename)
[...]

PR #2441, opened 2017-07-16
fixed triple quote issue when open triple quotes are on the last line

scopatz:
This should address #2436


melund:
Fixes it for me. Thanks @scopatz

PR #2452, opened 2017-07-22
autosuggest in completions

### LINKED ISSUE #2445: xonsh autocomplete suggests files that don't exists
```
touch a
ls a
rm a
```

After that, xonsh (with prompt_toolkit backend) will suggest a completion that doesn't make sense:

If I press `ls <TAB>`, it will suggest `a` in the completion menu even though that file doesn't exist.

Why is that, and how can I turn it off?

Thanks!

scopatz:
Closes #2445


melund:
Works for me. I think I going to use this setting.

PR #2456, opened 2017-07-25
Substitute xip for pip in places it should be used.

astronouth7303:
Try to use `xip` instead of `pip` in places where it should be used.

There is a known conflict with the name `xip`, though. Mac has an archival program [xip](https://en.wikipedia.org/wiki/.XIP).

Considering renaming `xip` to `xpip`.

### DIFF
--- a/xonsh/xontribs.json
+++ b/xonsh/xontribs.json
@@ -170,14 +170,14 @@
    "url": "http://exofrills.org",
    "install": {
     "conda": "conda install -c conda-forge xo",
-    "pip": "pip install exofrills"}
+    "pip": "xip install exofrills"}
    },
   "xonsh": {
    "license": "BSD 3-clause",
    "url": "http://xon.sh",
    "install": {
     "conda": "conda install -c conda-forge xonsh",
-    "pip": "pip install xonsh",
+    "pip": "xip install xonsh",
     "aura": "sudo aura -A xonsh",
     "yaourt": "yaourt -Sa xonsh"}
    },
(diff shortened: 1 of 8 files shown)


xonsh/xontribs.json
```diff
@@ -170,71 +170,71 @@
    "url": "http://exofrills.org",
    "install": {
     "conda": "conda install -c conda-forge xo",
-    "pip": "pip install exofrills"}
+    "pip": "xip install exofrills"}
```
scopatz: I don't know that all of the instances in this file should be changed, should they?

scopatz: Or is it actually needed here to circumvent some pip bug?

scopatz:
Thanks for putting this in! Also, is there anyway to make xip/xpip invisible to the user? Or is that a bad idea?

xonsh/xontribs.json
astronouth7303: Pretty much yes, because `pip` generally refers to the system (or venv) pip, whereas `xip` refers to the pip matching xonsh.

Running `pip install xontrib-spam` will only put it in the right place if xonsh is installed in the system environment. Which is not the case for conda and homebrew.

The only entry that you might not want to do this is the installation for `xonsh` itself, and frankly I'm not even sure why that's in this file.

astronouth7303:
@scopatz Define "invisible"?

No, you shouldn't alias `pip` to `xpip` because of anyone doing Python development.

Yes, we should streamline the installation of xontribs more, but that is outside the scope of this PR.

xonsh/xontribs.json
scopatz: Ahh that makes sense!

scopatz:
Yeah, by invisible I meant aliasing pip to xip, but I can see how that is a bad idea in general.  

> Yes, we should streamline the installation of xontribs more, but that is outside the scope of this PR.

Much agreed!

astronouth7303:
I think this is ready to go as soon as CI finishes.

scopatz:
Sounds great and LGTM!

gforsyth:
Thanks, @astronouth7303 !

PR #2460, opened 2017-07-29
Don't capture so much

### LINKED ISSUE #2404: How to capture output of programs that use alternate mode
In `bash` you can do

```bash
OUTPUT=$(fzf)
```

then select an entry in the `fzf` list (which runs in alternate mode), and afterwards you can 

```
echo $OUTPUT
```

and will obtain what you selected.

In xonsh, this doesn't work:

* `OUTPUT = $(fzf)` doesn't show the alternate mode (`fzf` is running but invisible)
* `OUTPUT = $[fzf]` makes it visible but that doesn't capture the final stdout that `fzf` produces

How to do this?

nathan-hoad:
This fixes #2404, covering another situation I also discovered.

My additional discovery was that stdout and stderr of hiddenobjects are also being captured, which the documentation states shouldn't be happening - http://xon.sh/tutorial.html#uncaptured-subprocess-with-and. I'm slightly less sure about this change, because the code quite explicitly refers to them as captured:

```
        elif p1 == '![':
            p0 = xonsh_call('__xonsh_subproc_captured_hiddenobject__', p2,
                            lineno=lineno, col=col)
```
and
```
def subproc_captured_hiddenobject(*cmds):
    """Runs a subprocess, capturing the output. Returns an instance of
    HiddenCommandPipeline representing the completed command.
    """
```

But I think it might a copy-paste thing from nearby code. Either way the capture is definitely causing issues, e.g. here's two invocations of `fzf` that select the same file:

```
nathan@susannah ~/projects/xonsh master $ fzf
MANIFEST.in
nathan@susannah ~/projects/xonsh master $ fzf
nathan@susannah ~/projects/xonsh master $
```

Given that the documentation states it shouldn't be captured, and with this PR it really isn't captured anymore, I should probably also update the code to state that it's uncaptured. If you find the change agreeable let me know and I'll make those changes.

### DIFF
--- a/news/dont-capture-so-much.rst
+++ b/news/dont-capture-so-much.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* stderr is now longer captured when using $().
+
+**Security:** None
--- a/news/dont-capture-so-much.rst
+++ b/news/dont-capture-so-much.rst
@@ -9,5 +9,6 @@
 **Fixed:**
 
 * stderr is now longer captured when using $().
+* stdout and stderr are no longer captured when using ![].
 
 **Security:** None
--- a/tests/test_builtins.py
+++ b/tests/test_builtins.py
@@ -6,6 +6,7 @@ import re
 import builtins
 import types
 from ast import AST
+from unittest.mock import Mock, patch
 
 import pytest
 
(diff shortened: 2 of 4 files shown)


gforsyth:
Hey @nathan-hoad -- sorry for the long delay in getting to this.  The changes sound reasonable to me, and yes, the uncaptured commands should probably be uncaptured.  Do you want to go ahead and update the code to reflect that?

scopatz:
Hello All, Thanks for the PR @nathan-hoad!  So I think that there might be some confusion over terminology here. un/capturing only refers to whether the output is piped to stdout and stderr.  The distinction between `$[cmd]` and `![cmd]` -- even though both will stream stdout -- is that:

* `$[cmd]` will always return None
* `![cmd]` will always return a rich HiddenCommandPipeline instance, which even let's you inspect the output of the command you have just run.

For example on master, we see what I think is the correct behaviour:

```console
scopatz@artemis ~ $ $[echo hello] is None
hello
True
scopatz@artemis ~ $ ![echo hello].out
hello
'hello\n'
```

On this branch, we no longer have access to the full output after the command is run:

```console
scopatz@artemis ~ $ $[echo hello] is None
hello
True
scopatz@artemis ~ $ ![echo hello].out
hello
''
```

Right now, all subprocess commands are automatically wrapped in `![cmd]`, rather than `$[cmd]`. If this is problematic for some reason, maybe we should allow folks to set which way they want uncaptured commands wrapped (either with an object or without). I'd welcome a PR to this effect. I hope that this helps!

PS super sorry about taking so long to get to this.

gforsyth:
I'm going to close this out in light of @scopatz' comments, but please ping here if anyone wants to continue this discussion and/or change the direction of this PR.

PR #2471, opened 2017-08-10
stupid file extension logic error

scopatz:
mea culpa. This does not need a news item since it fixes a bug introduced since the last release.


gforsyth:
Thanks, @scopatz !

Issue #2477, opened 2017-08-13
xonsh website improvement


ahundt:
I think there might be a couple of small changes that can make the [xon.sh](http://xon.sh) homepage and docs much easier to browse. Unfortunately, I kept getting lost and could not find my way back to pages I saw before. What do you think of the following?

1. `Contents` section on the homepage be added as a page next to index at the top right, or on the left under the xonsh icon
2. Update `Contents` section to show three levels instead of just two
   - Installation
        - ...
   - Guides
      - [Tutorial](http://xon.sh/tutorial.html)
        - [Customizing the prompt](http://xon.sh/tutorial.html#customizing-the-prompt)
        - ...
      - ...
    - ...
3. "Index" link be renamed "API Index"

gforsyth:
Hi @ahundt -- any improvements to the website are always welcome! If you're familiar with sphinx we'd love to have any help you can offer.

ahundt:
I tried dropping in a theme I happen to have on hand but I'm getting errors I don't know how to resolve:

```
writing output... [  0%] advanced_events
Recursion error:
maximum recursion depth exceeded in comparison

This can happen with very large or deeply nested source files.  You can carefully increase the default Python recursion limit of 1000 in conf.py with e.g.:
    import sys; sys.setrecursionlimit(1500)
make: *** [html] Error 1

```

I tried setting the recursion limit to 10000 to no effect...

ahundt:
@gforsyth done! see #2480

gforsyth:
Hey @ahundt, if you think your changes in #2480 meet the points raised feel free to close this out.  Thansk again!

ahundt:
Yep essentials are there, thanks. :-)

PR #2479, opened 2017-08-13
Improved doc theme

ahundt:
based on https://github.com/xonsh/xonsh/issues/2477, this adds the sphinx readable theme to improve the accessibility and appearance of the documentation.

Here is an example:
https://sphinx-readable-theme.readthedocs.io/en/latest/

The version in this pr is tweaked to also handle changing width a bit better.


ahundt:
#2480 is much better, closing this

Issue #2383, opened 2017-05-16
Resize in mc (ncurses?) not working


ulno:
```
$ xonfig
+------------------+-----------------+
| xonsh            | 0.5.9           |
| Python           | 3.5.3           |
| PLY              | 3.9             |
| have readline    | True            |
| prompt toolkit   | 1.0.9           |
| shell type       | prompt_toolkit  |
| pygments         | 2.2.0           |
| on posix         | True            |
| on linux         | True            |
| distro           | Ubuntu          |
| on darwin        | False           |
| on windows       | False           |
| on cygwin        | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
+------------------+-----------------+
```
I am trying to use mc (Midnight Commander, https://midnight-commander.org/) and its editor mcedit in xonsh. It works in general, but resizes of the terminal window (xfce4-terminal) after starting mc or mcedit are not reflected. The size stays the same. When executing bash from xonsh and then mc, the resize works instantaneously.

Some ncurses updates getting lost here?

Example of not/wrongly resized window:
![image](https://cloud.githubusercontent.com/assets/72434/26106451/f1264fd0-3a13-11e7-9a31-48d00a6465ea.png)

DanielFGray:
I too experience this, seems like xonsh eats SIGWINCH without passing it to child processes

laerus:
Hey guys thanks for the feedback! Can you try to reproduce with readline?

ulno:
Same issues when running xonsh --shell-type readline

Both methods (readline and prompt_toolkit + start of an ncurses application) destroy mouse selection in the terminal where it was executed - even when leaving xonsh.

ulno:
Interesting observation.
byobo (based on tmux) resize seems to run correctly in xonsh.
If I start byobo in xonsh and mx in byobu (which opens bash), mc resizing works correctly.
Starting byobo, and then xonsh in it and then mc shows same issues as before.

ulno:
Ranger (consoel based file manager) actually works too.

I wonder, if ncurses works but slang based programs don't.

ulno:
neovim does also not work and eats most of special keys and mouse events (just showing messed up characters) - similar to #2461

gforsyth:
Hi @ulno -- thanks for reporting these.  Ncurses and slang tend to mess things up for us. 

Two things: 
[...]

PR #2493, opened 2017-09-05
Adding `libgpm` in pattern used for bin analysis in default backgroundable predictors

jb-leger:
When ``gpm`` is used, we can assume the command is not backgroundable. I had the problem with ``links``, but I had a look on why the predictor with binary analysis fails. Adding in the exception list is not a general fix, so I prefer fix the predictor.

### DIFF
--- a/news/predictor_libgpm.rst
+++ b/news/predictor_libgpm.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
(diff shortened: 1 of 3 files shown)


jb-leger:
In #2383 we see Midnight Commander have the same problem. After testing, this PR solve this issue.

gforsyth:
Hey @jb-leger -- thanks for putting this in! And sorry for the delay in getting to it!
There was a failing test, but it looks like it was a random timeout and restarting it seems to have fixed things.  This looks good to me.

PR #2507, opened 2017-10-09
PEP-515 Support (Underscores in Numeric Literals)

chronitis:
This adds support for PEP-515 (underscores in numeric literals). Support
is added unconditionally on underlying python version by patching the
tokenizer with the regexes used in cpython 3.6, then modifying the
parser to unconditionally remove underscores before evaluating numeric
literals, which avoids needing to check for the underlying version.


gforsyth:
Sorry for the delay in getting to this!

scopatz:
Thanks a million @chronitis!

PR #2508, opened 2017-10-09
bash_to_xsh.rst add command line args to table

ahundt:


astronouth7303:
Looks good to me, assuming the syntax comes out right.

Issue #2520, opened 2017-11-02
Improvements to 'autopair' behavior


Danmou:
Currently, if the `XONSH_AUTOPAIR` setting is enabled, entering any of the characters `' " ( [` will also insert its matching counterpart to the right of the cursor. This behavior is often more annoying than useful, for example when you are trying to wrap some text in quotes after writing it. To make it more useful, and in line with the behavior in popular editors like Sublime Text, I would suggest the following changes:

* For quotes `'` and `"`, the matching quote should only be inserted if there is whitespace (or BOL/EOL) on both sides of the cursor.
* For brackets `(`, `[` and `{` (note that `{` is not currently included in the autopair behavior), the matching bracket should only be inserted if there is whitespace (or EOL) to the right of the cursor.
* Pressing backspace when the cursor is between a matching pair, e.g. `'|'` or `(|)` should delete both characters.

Examples (where `|` represents the cursor):

`echo |some text {that needs to be put in quotes}` -> press `'`
* Current behavior: `echo '|'some text {that needs to be put in quotes}`
* Desired behavior: `echo '|some text {that needs to be put in quotes}`

`echo 'some text {that needs to be put in quotes}|` -> press `'`
* Current behavior: `echo 'some text {that needs to be put in quotes}'|'`
* Desired behavior: `echo 'some text {that needs to be put in quotes}'|`

`|2 + 2) / 5` -> press `(`
* Current behavior: `(|)2 + 2) / 5`
* Desired behavior: `(|2 + 2) / 5`

`function(|)` -> press backspace
* Current behavior: `function|)`
* Desired behavior: `function|`

### Edit:
To make autopair work properly inside nested brackets, we have to allow a closing bracket instead of whitespace on the right side of the cursor, and an opening bracket on the left side of the cursor. For example:

`function(|)` -> press `'`
* `function('|')`

`function()|` -> press `[`
* `function()[|]`

Danmou:
I've submitted a PR that implements this.

scopatz:
Thanks for this @Danmou!

PR #2522, opened 2017-11-02
Improvements to 'autopair' behavior

Danmou:
Implements improvements suggested in #2520

### DIFF
--- a/news/improved_autopair.rst
+++ b/news/improved_autopair.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+* Improved the autopair behavior to match that of popular code editors.
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


astronouth7303:
Great work! Can you add a news item per http://xon.sh/devguide.html#changelog ?

Danmou:
I just made a small improvement, so nested brackets should work as expected. It should be all ready now.

astronouth7303:
I don't know PTK well. Is this code testable?

Danmou:
Maybe, but to be honest I have no experience with unit testing. The changes I've made are pretty simple, since it's just a restriction of the cases where the autopair behavior should take place. It's difficult to define a 'correct' behavior, because it is somewhat dependent on personal preference, but I've made it so it pretty closely follows the behavior in Sublime Text (with the exception that ST sometimes changes it's behavior if there are unmatched brackets or quotes, which I think would be overkill in our case).

There's a more detailed description of the behavior in #2520

gforsyth:
It's not easily testable, since the keybindings are all inner functions and rely on a lot of the prompt_toolkit machinery.  

On the (long) todo list is to rewrite the way we define the keybindings so it's more reusable and more easily tested. This code looks reasonable to me right now and seems to work as expected. If there's some gotcha in there, it will only be an annoyance so I'm good to merge this in as is.

PR #2526, opened 2017-11-09
fix for eating semicolons

scopatz:
This should address #2511

### DIFF
--- a/news/semieat.rst
+++ b/news/semieat.rst
@@ -1,0 +1,15 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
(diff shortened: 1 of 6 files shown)


gforsyth:
Thanks, @scopatz !  This seems to fix the semicolons being eaten and should go in either way -- one little bit of inconsistent behavior is whether or not the `;` itself is `echo`ed when there are wildcards around:

```
~/tmp/foo 🐚  echo spam * ; echo eggs
spam file1 file2
eggs
~/tmp/foo 🐚  echo * spam ; echo eggs
file1 file2 spam ;
eggs
```

scopatz:
Thanks for noticing!

scopatz:
Ho-kay, I have put in a fix for this weird case.

gforsyth:
Thanks, @scopatz !

Issue #2514, opened 2017-10-19
Eats control sequences


maxigas:
Seems xonsh eats control sequences such as ^N rather than passing it to the running application as expected. In fact the literal characters ^ and N appear on the screen. Some other sequences seem to work though.

Test case:

    sudo pip3 install xonsh prompt_tools
    xonsh
    emacs25-nox foo.txt
    # Above works as expected
    emacs25 --daemon
    emacsclient -t foo.txt
    # In emacsclient ^N does not move caret to next line as expected, but outputs literally ^N

Test system:

* Debian Stable
* xonsh/0.5.12 installed via pip3 with prompt_tools backend
* GNU emacs 25.1.1
* X window system
* urxvtcd terminal emulator
* Original shell can be sh, bash or fish, did not test others

Seems to be similar to #2461 and maybe #2383 but it is not clear.

Thanks for developing xonsh.

gforsyth:
Thanks for reporting @maxigas -- I agree that this is related to #2461. For a possible temporary workaround, you can try adding this to your `xonshrc`
```
__xonsh_commands_cache__.threadable_predictors['emacsclient'] = lambda args: True
```

scopatz:
I think this is one that we can go ahead and just add.

scopatz:
Potential fix in #2529

scopatz:
Closing because #2529 has been merged.

PR #2529, opened 2017-11-10
emacslient no longer threadable

scopatz:
Should address #2514


gforsyth:
Appveyor is complaining but that doesn't make any sense and I'm merging anyway.

gforsyth:
Thanks, @scopatz !

PR #2532, opened 2017-11-11
this never should have been moved

scopatz:
This is probably partially responsible for some of the slowdowns


gforsyth:
Oof.  Yeah, that'll do it.  Nice spot, @scopatz

Issue #2534, opened 2017-11-14
Python lists sometimes get interpreted as regex in completer


Danmou:
I've found a weird bug: If I write `[1-0.1]` in xonsh, I will get a traceback like the following upon writing the last square bracket:

```
Exception in thread Thread-2577:
Traceback (most recent call last):                                                                        
  File "/home/daniel/anaconda3/lib/python3.6/threading.py", line 916, in _bootstrap_inner                
    self.run()                                                                                           
  File "/home/daniel/anaconda3/lib/python3.6/threading.py", line 864, in run                             
    self._target(*self._args, **self._kwargs)                                                            
  File "/home/daniel/anaconda3/lib/python3.6/site-packages/prompt_toolkit/interface.py", line 840, in run 
    completions = list(buffer.completer.get_completions(document, complete_event))
  File "/home/daniel/Git/xonsh/xonsh/ptk/completer.py", line 49, in get_completions
    self.ctx)
  File "/home/daniel/Git/xonsh/xonsh/completer.py", line 35, in complete
    out = func(prefix, line, begidx, endidx, ctx)
  File "/home/daniel/Git/xonsh/xonsh/completers/path.py", line 268, in complete_path
    sort_result=glob_sorted):
  File "/home/daniel/Git/xonsh/xonsh/tools.py", line 1872, in iglobpath
    return _iglobpath(s, ignore_case=ignore_case, sort_result=sort_result)[0]
  File "/home/daniel/Git/xonsh/xonsh/tools.py", line 1854, in _iglobpath
    paths = glob.glob(s, recursive=True)
  File "/home/daniel/anaconda3/lib/python3.6/glob.py", line 20, in glob
    return list(iglob(pathname, recursive=recursive))
  File "/home/daniel/anaconda3/lib/python3.6/glob.py", line 55, in _iglob
    yield from _glob1(dirname, basename, dironly)
  File "/home/daniel/anaconda3/lib/python3.6/glob.py", line 83, in _glob1
    return fnmatch.filter(names, pattern)
  File "/home/daniel/anaconda3/lib/python3.6/fnmatch.py", line 52, in filter
    match = _compile_pattern(pat)
  File "/home/daniel/anaconda3/lib/python3.6/fnmatch.py", line 46, in _compile_pattern
    return re.compile(res).match
  File "/home/daniel/anaconda3/lib/python3.6/re.py", line 233, in compile
    return _compile(pattern, flags)
[...]

PR #2535, opened 2017-11-15
Escape regex in path completion

gforsyth:
Fix for #2534 

We append a `*` to the prefix before trying path globbing and for prefixes with `[]` that was throwing errors. There's a `glob.escape` to fix that.

### DIFF
--- a/news/escape_regex.rst
+++ b/news/escape_regex.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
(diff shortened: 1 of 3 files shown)


scopatz:
Thanks @gforsyth! looks like we have a lot of CI failures...

gforsyth:
I wonder if we can move the news item formatting checks out of the unit tests?  Anyway, an issue for another time.  This should be good to go now

scopatz:
Super sorry I missed this @gforsyth!

Issue #2548, opened 2017-12-01
Non-portable path manipulation behavior 


paulschellin:
I discovered an (maybe) unlikely crash related to the cwd_release_wrapper in the xontrib module "free_cwd".

On line 47, 48 of free_cwd.py:
```
rootdir = os.path.splitdrive(os.getcwd())[0] + '\\'
os.chdir(rootdir)
```

When the split drive function is called on a home directory in macOS (and I'm guessing anything aside from Windows as well), the "drive" element of the pair is actually just an empty string (and I just confirmed it here: https://docs.python.org/3.6/library/os.path.html#os.path.splitdrive). Basically it tries to cd into "\\", which caused Xonsh to crash.

I think the pathlib.Path objects have a portable way to do this, using their "anchor" method -- it returns the drive on Windows and root elsewhere.

I think I just misunderstood what that xontrib module was for, but now I understand it is intended for Windows users only.

I figure that eventually someone will make the same wrong turn as I did, so perhaps they'll be able to find it more easily.

This is a really intriguing project, it's impressive how far it's come!

astronouth7303:
Yes, that is a windows-only xontrib and should not be enabled on non-Windows.

We should probably add a guard to the xontrib to prevent it from being loaded if not being run on windows (I think specifically Win32 or Cygwin).

melund:
@paulschellin I have updated the ``free_cwd`` xontrib, (see #2549)  to make it clear that this is intended for Windows. Thanks for the hints about ``pathlib``. I really should start to use that exclusively.

scopatz:
Closing this sinc #2549 has been merged.  Please feel free to reopen!

PR #2549, opened 2017-12-03
Free cwd on win

melund:
This fixes the problem reported by @paulschellin in #2548 

It ensures that ``free_cwd`` xontrib can only be active on windows.


scopatz:
Thanks @melund!

Issue #2554, opened 2017-12-05
tab completion error


pyeguy:
hi all

i'm on windows 10 using `cmder` (essentially ConEmu) sometimes i get the following error when tab completing directories:

```python
 cd monomersException in thread Thread-163:
Traceback (most recent call last):
  File "C:\Users\camer\Anaconda3\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "C:\Users\camer\Anaconda3\lib\threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\camer\Anaconda3\lib\site-packages\prompt_toolkit\interface.py", line 860, in run
    completions = list(buffer.completer.get_completions(document, complete_event))
  File "C:\Users\camer\Anaconda3\lib\site-packages\xonsh\ptk\completer.py", line 49, in get_completions
    self.ctx)
  File "C:\Users\camer\Anaconda3\lib\site-packages\xonsh\__amalgam__.py", line 124, in complete
    out = func(prefix, line, begidx, endidx, ctx)
  File "C:\Users\camer\Anaconda3\lib\site-packages\xonsh\completers\__amalgam__.py", line 1467, in complete_cd
    return complete_dir(prefix, line, start, end, ctx, True)
  File "C:\Users\camer\Anaconda3\lib\site-packages\xonsh\completers\__amalgam__.py", line 1201, in complete_dir
    filtfunc=os.path.isdir)
  File "C:\Users\camer\Anaconda3\lib\site-packages\xonsh\completers\__amalgam__.py", line 1159, in complete_path
    sort_result=glob_sorted):
  File "C:\Users\camer\Anaconda3\lib\site-packages\xonsh\__amalgam__.py", line 7052, in iglobpath
    return _iglobpath(s, ignore_case=ignore_case, sort_result=sort_result)[0]
  File "C:\Users\camer\Anaconda3\lib\site-packages\xonsh\__amalgam__.py", line 7034, in _iglobpath
    paths = glob.glob(s, recursive=True)
  File "C:\Users\camer\Anaconda3\lib\glob.py", line 20, in glob
    return list(iglob(pathname, recursive=recursive))
  File "C:\Users\camer\Anaconda3\lib\glob.py", line 55, in _iglob
    yield from _glob1(dirname, basename, dironly)
  File "C:\Users\camer\Anaconda3\lib\glob.py", line 83, in _glob1
    return fnmatch.filter(names, pattern)
[...]

PR #2555, opened 2017-12-07
Make iglobpath() more robust

scopatz:
This should address #2554

### DIFF
--- a/news/iglob.rst
+++ b/news/iglob.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Made an exceptional case in ``iglobpath()`` more robust when Python globbing
+  fails for due to strange scrandir issue.
+
+**Security:** None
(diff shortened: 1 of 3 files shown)


astronouth7303:
Is there a matching Python bug?

scopatz:
I didn't easily find one.  It is probably worth reporting upstream, if it is reproducible.

astronouth7303:
well, we should probably have our own test case so we make sure it doesn't happen again.

scopatz:
I agree, but I am not sure how to write one since it involves `os.scandir()` returning an empty string inside of `glob.iglob()`  I guess we could monkey path scandir and listdir for the test....

scopatz:
Does anyone know the proper pytest-foo to do that?

astronouth7303:
I would guess https://pypi.python.org/pypi/pytest-mock

melund:
I would say something like: 

```
import os
import glob

def test_iglob(monkeypatch):
    def mockscandir(path):
        yied ''
    monkeypatch.setattr(os, 'scandir', mockscandir)
    assert glob.iglob('some/path')
```

scopatz:
Thanks @melund @astronouth7303! I have added a monkeypatch test.

gforsyth:
Thanks @scopatz !

Issue #2327, opened 2017-03-15
Endless loop of an input/output error


refi64:
Still trying to accurately reproduce this, but here's the gist of it:

![xonsh_gem](https://cloud.githubusercontent.com/assets/1690697/23972507/14e78db6-09a0-11e7-8c35-dc89c3a3199a.jpg)

Was trying to install a `gem` package. So innocent, yet so much pain in return...

:O

gforsyth:
I've seen this on my machine before -- happened at shutdown and I had to power cycle the machine. Haven't been able to reproduce it, though, and haven't seen it in a while.

Thanks for reporting, @kirbyfan64 !

tekacs:
This happens to me frequently - on OSX and Linux of a variety of descriptions, all under `prompt_toolkit`.

astronouth7303:
Do we need a master tracker bug? We've got a pile of issues of similar symptoms.

scopatz:
Hello All, do we have a full traceback for this?  I have never seen it :(

tekacs:
@scopatz I'll run with `$XONSH_SHOW_TRACEBACK = True` for a few days and see if I can get you one.

scopatz:
Thanks!

tekacs:
@scopatz Here we go!

```
amar@<snip> ~/repos/<snip> master $ kubectl logs -f <some-kubernetes-pod> -p
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):                                                                                                                                                                                                                                                                                                                       
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 14797, in default
    run_compiled_code(code, self.ctx, None, 'single')
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 2322, in run_compiled_code
    func(code, glb, loc)                          
  File "<xonsh-code>", line 1, in <module>                                             
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 19353, in subproc_captured_hiddenobject
    return run_subproc(cmds, captured='hiddenobject')                                                   
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 19316, in run_subproc
    command.end()                       
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 12622, in end                                                                                                                                                                                                                                                                       
[...]

PR #2557, opened 2017-12-07
Give sensible defaults when encoding bytes

scopatz:
Potential fix for #2327

### DIFF
--- a/news/teenc.rst
+++ b/news/teenc.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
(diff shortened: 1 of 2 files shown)


gforsyth:
Hey @scopatz -- this looks like a reasonable change to me.  Since it's so hard to trigger this error I'm inclined to merge it in and then see if anyone can still get the error to pop.
Just waiting on CI -- there's a weird race condition in the OSX tests on Travis that seems to be causing random failures, but re-running the build usually sorts them.

Issue #2222, opened 2017-02-19
Alias output redirect issue - I/O operation on closed file


mitnk:
This is a test case in tests/test_integrations.py.
```
def _f(args, stdout):
    print('Wow Mom!', file=stdout)
aliases['f'] = _f
```
With this, run `f o>e` several times would reproduce the following error:
```
$ f o>e
$ f o>e
$ f o>e
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/site-packages/xonsh/base_shell.py", line 322, in default
    run_compiled_code(code, self.ctx, None, 'single')
  File "/usr/local/lib/python3.5/site-packages/xonsh/codecache.py", line 65, in run_compiled_code
    func(code, glb, loc)
  File "<xonsh-code>", line 1, in <module>
  File "/usr/local/lib/python3.5/site-packages/xonsh/built_ins.py", line 858, in subproc_captured_hiddenobject
    return run_subproc(cmds, captured='hiddenobject')
  File "/usr/local/lib/python3.5/site-packages/xonsh/built_ins.py", line 826, in run_subproc
    command.end()
  File "/usr/local/lib/python3.5/site-packages/xonsh/proc.py", line 1991, in end
    self._end(tee_output=tee_output)
  File "/usr/local/lib/python3.5/site-packages/xonsh/proc.py", line 1999, in _end
    for _ in self.tee_stdout():
  File "/usr/local/lib/python3.5/site-packages/xonsh/proc.py", line 1915, in tee_stdout
    for line in self.iterraw():
  File "/usr/local/lib/python3.5/site-packages/xonsh/proc.py", line 1797, in iterraw
    stdout = NonBlockingFDReader(stdout.fileno(), timeout=timeout)
ValueError: I/O operation on closed file
```
Both Linux & Mac have this issue.

scopatz:
Thanks for reporting @mitnk. I thought I had fixed that... oh well.

scopatz:
Potential fix in #2559

gforsyth:
Closed from #2559

PR #2559, opened 2017-12-08
fixed callable alias e>o redirection

scopatz:
This should address #2222

### DIFF
--- a/news/e2o.rst
+++ b/news/e2o.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fixed failing redirections from stderr to stdout when the command
+  being executed was a callable alias.
+
+**Security:** None
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -83,7 +83,7 @@ def _f():
 aliases['f'] = _f
 f
 """, "hello\n", 0),
-# test redirecting a function alias
+# test redirecting a function alias to a file
 ("""
 def _f():
     print('Wow Mom!')
@@ -95,6 +95,14 @@ with open('tttt') as tttt:
     s = tttt.read().strip()
 print('REDIRECTED OUTPUT: ' + s)
 """, "REDIRECTED OUTPUT: Wow Mom!\n", 0),
+# test redirecting a function alias from stderr -> stdout
+("""
+def _f(args, stdin, stdout, stderr):
+    print('The Truth is Out There', file=stderr)
+
+aliases['f'] = _f
+f e>o
+""", "The Truth is Out There\n", 0),
 # test system exit in function alias
 ("""
 import sys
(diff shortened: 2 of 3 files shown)


scopatz:
OK, I think this is ready for review.

mitnk:
Seems still has this issue:

```
$ f o>e
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/base_shell.py", line 328, in default
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/codecache.py", line 68, in run_compiled_code
    func(code, glb, loc)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/built_ins.py", line 881, in subproc_captured_hiddenobject
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/built_ins.py", line 844, in run_subproc
    command.end()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/proc.py", line 2010, in end
    self._end(tee_output=tee_output)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/proc.py", line 2018, in _end
    for _ in self.tee_stdout():
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/proc.py", line 1934, in tee_stdout
    for line in self.iterraw():
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/xonsh/proc.py", line 1816, in iterraw
    stdout = NonBlockingFDReader(stdout.fileno(), timeout=timeout)
ValueError: I/O operation on closed file
```

xonfig:

```
| xonsh            | 0.6.0.dev3984       |
| Git SHA          | 7af4ca53            |
| Commit Date      | Dec 8 14:36:33 2017 |
| Python           | 3.6.2               |
| prompt toolkit   | 1.0.13              |
```

scopatz:
Hi @mitnk - are you using the following `f()`?

```
def _f(args, stdout):
    print('Wow Mom!', file=stdout)
```

One of the problems with this alias is that the `stdout` arg is actually `stdin`.  We only match on argument position, not argument name (though that is a nice idea!).  So to properly set stdout, we'd have to use:

```
def _f(args, stdin, stdout):
    print('Wow Mom!', file=stdout)
```

I am not suprised that there are weird closed file errors when writing stdin and redirecting stdout to stderr.  Perhaps the error message here could be better too...

mitnk:
Oh, I didn't notice that :)

However same error raises with `def _f(args, stdin, stdout):`.

scopatz:
Hi @mitnk - that is very strange.  What error is it giving? The same one?  The test suite has a fix for o>e that works on my machine.

mitnk:
Yes, the same error on Mac. Just tested on a Linux box, the issue indeed was fixed. So I think we can merge this PR.

gforsyth:
Tests are all passing and it works locally on a mac over here, so I'm going to go ahead and merge this.  @mitnk, if problems persist for you on OSX please let us know.

gforsyth:
Thanks, @scopatz!  And sorry for the delay!

scopatz:
No worries! Happy to have it in!

PR #2567, opened 2017-12-19
Remove Static Configuration

### LINKED ISSUE #2542: Get rid of static config files
I didn't want to close #2540 until there was a ticket open. So here it is.  #2510 also references this.

Good idea, bad idea? Well at least we've got a ticket to collect that information in.

scopatz:
This PR removes static configuration file support from xonsh. Instead, it relies solely on xonshrc files.  The `xonfig wizard` command has been adjusted to write out to the user's xonshrc file instead. This should close #2542


scopatz:
Please bang on this since it is a major, backwards incompatible change.

xonsh/xonfig.py
```diff
@@ -110,6 +112,76 @@
 Thanks for using the xonsh configuration wizard!"""
 
 
+@lazyobject
+def _XONFIG_SOURCE_FOREIGN_SHELL_COMMAND():
+    return collections.defaultdict(lambda: 'source-foreign',
```
astronouth7303: I'm pretty sure that `defaultdict` is as cheap or cheaper than `lazyobject`. This isn't doing IO or complex calculation.

astronouth7303:
```
🐚 timeit("lazyobject(lambda: None)", "from xonsh.lazyasd import lazyobject")
0.5423036129999446
🐚 timeit("defaultdict(lambda: 'source-foreign', bash='source-bash', cmd='source-cmd', zsh='source-zsh')", "from collections import defaultdict")
0.385492806000002
```

astronouth7303: oh yeah, `defaultdict` is in C.

https://github.com/python/cpython/blob/c5ae169e1b73315672770517bf51cf8464286c76/Lib/collections/__init__.py#L42

scopatz:
> I'm not following why there still exists JSON functionality?

I am planning on using it in some downstream projects :)

xonsh/xonfig.py
scopatz: You are right!

scopatz: Changed.

astronouth7303:
Do you need to vendor the wizard framework?

(Also, we need actual vendoring tooling.)

scopatz:
Just to be clear, are suggesting breaking out the wizard into its own project? I am not opposed in theory, but there are some parts that are dependent on xonsh.tools, so that project would itself depend on xonsh.

astronouth7303:
oh, I had assumed "I am planning on using it in some downstream projects :)" implied things not tied to xonsh.

scopatz:
Oh, yeah, they have xonsh as a dependency: https://regro.github.io/

CJ-Wright:
:+1: to having a broken out wizard

scopatz:
Pinging this again!  Also Appveyor seems to be totally stalled right now...

gforsyth:
With the caveat that I don't use foreign shell imports or static config files, for those that do use foreign shell imports, doesn't this make it harder to customize that behavior? And we can decide that that's less important and that's fine, but eyes wide open and all that.

Also, should we, in the interim, have a "late" deprecation notice that informs users who have a static config file that it's being ignored and they should move stuff to `xonshrc`?

xonsh/environ.py
```diff
@@ -249,13 +244,18 @@ def default_xonshrc(env):
...
-                os.path.join(os_environ['ALLUSERSPROFILE'],
+        dxrc = (os.path.join(os_environ['ALLUSERSPROFILE'],
                              'xonsh', 'xonshrc'),
                 xcdrc,
                 os.path.expanduser('~/.xonshrc'))
     else:
-        dxrc = (xonshconfig(env), '/etc/xonshrc', xcdrc, os.path.expanduser('~/.xonshrc'))
+        dxrc = ('/etc/xonshrc', xcdrc, os.path.expanduser('~/.xonshrc'))
+    # Check if old config file exists and issue warning
+    old_config_filename = xonshconfig(env):
```
gforsyth: extra `:`

scopatz: Thanks!

scopatz:
>  for those that do use foreign shell imports, doesn't this make it harder to customize that behavior?

This is a great question @gforsyth! If we ignore what has come before, I think for new users not having the config is easier and more intuitive actually.  Rather than having two disticnt ways of starting up from foreign shells, we would now have only one.  Namely, inside the xonshrc, one could always run `source-bash`, `source-zsh`, etc.

I think it is worth it to streamline this.

gforsyth:
Ok -- this looks good to me.  I'm happy to merge and then people can hammer on it and find out what, if anything, isn't working quite right.  
I think we also need to write up a section of docs about foreign sourcing, since most of the documentation about that is contained in the disappearing section about static configuration.

One more thing for the manual, I suppose.

gforsyth:
Thanks, @scopatz!

PR #2571, opened 2017-12-20
added setup function

scopatz:
This will be useful in rever and elsewhere.


astronouth7303:
How does this relate to being able to use `xsh` importables in other packages? Was that ever really a thing?

scopatz:
This makes that easier to do.  rever is basically a fully xonsh project, and this function represents the hoops that need to be jumped through to enable that. See https://github.com/regro/rever/blob/master/rever/__init__.py

Issue #2370, opened 2017-05-02
xonsh not working on cygwin


fmv1992:
xonsh is not working on cygwin:

```
data_utilities/tests [develop●] » echo $XONSH_SHOW_TRACEBACK
True
data_utilities/tests [develop●] » xonsh --version
xonsh/0.5.1
data_utilities/tests [develop●] » xonsh
user@computer ~/path $ ls
                               [...] hangs [...]
```

some normal commands like ls, l, rm, etc remain in a irresponsive state. Ctrl + c does not work.

Python commands are working though:

```
/tmp » xonsh
user@computer /tmp $ print('hey')
hey
user@computer /tmp $
```

Finally commands can be taken out of the hanging state by using Ctrl + Z.

When quitting one of those sessions full of hangs I got:

```
xonsh: there are unfinished jobs
-----
[7]+ stopped: ls -a (10448)
[6]- stopped: rm ./bool_0.png (11000)
[5]  stopped: ls (10352)
[4]  stopped: ls (9600)
[3]  stopped: ls (10984)
[2]  stopped: ls (10080)
[1]  stopped: ls (9536)
```
(I have the same problem with the newest version and versions 0.4.x)
Please let me know how can I be of more help.

I can't wait to have it working on my computer : )

Best,

Felipe.

laerus:
hey @fmv1992 thanks for opening this!

i think Cygwin was always one of xonsh's weak spots.
We've gone back and forth about how we manage subprocesses trying to be
cross-platform and all and the main lesson was:
> Things are more complicated that we expected.

(it's going to be on the xonsh t-shirt too)

Jokes aside there is a major effort going on to fix this.
The best i can give you right now is to roll back to a xonsh version <=0.4.7
and see if this works out.

astronouth7303:
Unfortunately, xonsh/slug#5

fmv1992:
Can you outline as to what the problem is?

I'm willing to contribute even though I have no experience with designing a shell program.

anki-code:
I'm going to close this because we have many changes since 2017. Please open the separate issues with repeatable examples.

PR #2573, opened 2017-12-25
add MSYS2 support and fix Cygwin support

koehlma:
Hey there,

while running xonsh within an MSYS2 environment on Windows 10 I encountered issues similar to #2370 and in addition short-running processes seemed to randomly suspend themselves to the jobs list. When processes freezed, they did however not consume any resources, but instead seemed to have just stopped. Since this happens nondeterministically (and after some debugging) I believe there is some race condition within MSYS2 (Cygwin) with calling `Popen.send_signal` in ` pause_call_resume` in `proc.py` which unexpectedly suspends processes.

Anyway, this pull request adds support for MSYS2 and fixes those issues (at least for me :D).

Thank you all very much for this great piece of software. Windows became finally somewhat usable with xonsh. ;)

Best regards,
Maximilian

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -11,9 +11,11 @@ environment:
       DISTUTILS_USE_SDK: "1"
     - PYTHON: "C:\\Python35-x64"
     - PYTHON: "C:\\Python36-x64"
+    - XONSH_TEST_ENV: "MSYS2"
+      MSYS2_PATH: "C:\\msys64"
     # TODO: Miniconda
 install:
-- "%PYTHON%\\Scripts\\pip install -r requirements-tests.txt"
+  - ".appveyor_install.cmd"
 build: off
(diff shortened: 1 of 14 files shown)


astronouth7303:
Can MSYS2 and Cygwin please be added to the appveyor testing matrix?

koehlma:
Here you go. However, I do not know how to automate the process of installing Python and pip within the Cygwin environment provided by AppVeyor, so for now, I only added MSYS2 to the testing matrix. Furthermore, I allowed the MSYS2 tests to fail. I think this is a good idea until someone fixed the remaining issues with MSYS2.

scopatz:
Hey! This looks great! Sorry for taking so long to review.

scopatz:
Thanks a million for putting this in @koehlma!

PR #2589, opened 2018-02-15
Set prompt_toolkit version to <2.0.0 in requirements.

jonathanslenders:
The current xonsh version will break when prompt_toolkit 2.0 is released.
Set version to <2.0.0 until 2.0 compatibility has been merged.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -330,7 +330,7 @@ def main():
         }
         skw['cmdclass']['develop'] = xdevelop
         skw['extras_require'] = {
-            'ptk':  ['prompt-toolkit'],
+            'ptk':  ['prompt-toolkit<2.0.0'],
             'pygments': ['pygments'],
             'win': ['win_unicode_console'],
             'mac': ['gnureadline'],


melund:
No ideal why circle is failing. This looks good to me. I guess we need to do the same for the conda package...

gforsyth:
Thanks, @jonathanslenders !  Could you give us a heads-up before you release `2.0.0` so we can make sure to push out a new `xonsh` release before then with the updated reqs?  Thanks!

melund:
Is that enough. Wouldn't conda, for example, just grab an older version which doesn't explicitly state  PTK < 2....

gforsyth:
Hey @melund -- yeah, we'll have to update the conda forge recipe, too

melund:
I have updated the conda-forge recipe as well. https://github.com/conda-forge/xonsh-feedstock/pull/53

melund:
I could be afraid that it is not enough even with an updated conda package. @gforsyth Wouldn't conda just grab an older package where we haven't required PTK<2.0? 

We could delete the older conda-forge packages, but I wonder if that is bad practice. The best would be to support PTK2 as soon as possible.

PR #2592, opened 2018-03-12
Circle2

gforsyth:
Updating Circle to version 2.0

Added a `ci/` subdir to hold conda environment files.  Circle 2 has some pretty nice caching stuff so this takes advantage of that for the saving and loading of the test environments.

Running flake8 on python 3.6 also revealed a bunch of pep8 errors that we weren't catching on 3.5 for some reason.  In the interests of keeping this just related to CI, I've added them to the `setup.cfg` exceptions list.  Most of them are `ambiguous variable names` or other things that are relatively benign, but it would be nice to fix them as we go.


scopatz:
Thanks @gforsyth! Not sure why appveyor was failing but this PR shouldn't have affected that.

PR #2605, opened 2018-04-03
Better Append Newline Default

### LINKED ISSUE #2603: cat file without eol
In bash:
```
worldmind@host:~$ cat file
FILEDATASTRworldmind@host:~$
```
in xonsh
```
worldmind@host ~ master $ cat file
worldmind@host ~ master $
```
we don't see string from file
looks like problem with buffering

scopatz:
Should close #2603

### DIFF
--- a/news/pktnl.rst
+++ b/news/pktnl.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* ``$XONSH_APPEND_NEWLINE`` will now default to True when in interactive mode.
+
+**Security:** None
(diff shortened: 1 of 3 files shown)


gforsyth:
lgtm, thanks @scopatz!

gforsyth:
Hmm, so travis passed here but then failed on Python3.4 on OSX when it tried to build off of master...

scopatz:
Hmm weird.  There seem to be some test instabilities I can't reproduce locally.

gforsyth:
I can't reproduce them either.  Could possibly be a weird travis hiccup?

scopatz:
Yeah, it is weird. I am looking at the rawlambda failure in #2606 and it seems like it isn't actually picking up the changes.

Issue #2593, opened 2018-03-15
xonsh crashes over python 3.7


jgmtzmd:
hi. i have to installs of xonsh: over python3.4 (installed by my package manager zypper of opensuse 42.3), and over python3.7 (i will call this xonsh37)
xonsh37 crashes all the time.

example:
`jg@Osuse ~ $ top -o COMAND
`works fine. But when i exit (either q or ctrl-c)..
this happens..
```
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/opt/python/lib/python3.7/site-packages/xonsh/proc.py", line 1837, in iterraw
    raise StopIteration
StopIteration

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/python/lib/python3.7/site-packages/xonsh/base_shell.py", line 328, in default
    run_compiled_code(code, self.ctx, None, 'single')
  File "/opt/python/lib/python3.7/site-packages/xonsh/codecache.py", line 68, in run_compiled_code
    func(code, glb, loc)
  File "/opt/python/lib/python3.7/site-packages/xontrib/vox_tabcomplete.xsh", line 1, in <module>
    def _vox_completer(prefix, line, begidx, endidx, ctx):
  File "/opt/python/lib/python3.7/site-packages/xonsh/built_ins.py", line 877, in subproc_captured_hiddenobject
    return run_subproc(cmds, captured='hiddenobject')
  File "/opt/python/lib/python3.7/site-packages/xonsh/built_ins.py", line 840, in run_subproc
    command.end()
  File "/opt/python/lib/python3.7/site-packages/xonsh/proc.py", line 2010, in end
    self._end(tee_output=tee_output)
  File "/opt/python/lib/python3.7/site-packages/xonsh/proc.py", line 2018, in _end
    for _ in self.tee_stdout():
  File "/opt/python/lib/python3.7/site-packages/xonsh/proc.py", line 1934, in tee_stdout
    for line in self.iterraw():
RuntimeError: generator raised StopIteration
jg@Osuse ~ $                                                                                                    
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/opt/python/lib/python3.7/site-packages/xonsh/ptk/shortcuts.py", line 109, in prompt
    document = cli.run(reset_current_buffer=False)
[...]

PR #2608, opened 2018-04-06
Replace StopIteration with return

WouldYouKindly:
[PEP-479](https://www.python.org/dev/peps/pep-0479/) says 
"Finally, the proposal also clears up the confusion about how to terminate a generator: the proper way is return, not raise StopIteration."

This addresses [https://github.com/xonsh/xonsh/issues/2593](https://github.com/xonsh/xonsh/issues/2593)


WouldYouKindly:
Not sure if it deserves a news item...

scopatz:
Hi @WouldYouKindly - thanks for the fix!  I think this defintiely needs a news item (especially if it solves another issue).  

One small issue here is that this would break compatabilty with Python 3.4 because PEP-479 didn't go into effect until v3.5 (if I am not mistaken).  Also v3.4 [doesn't reach end-of-life until 2019-03-16](https://devguide.python.org/#status-of-python-branches). So I think we need an if-statement here that checks if we are on Python v3.4

scopatz:
Oh! I thought it would! If it doesn't, that's great.

scopatz:
Thanks a million @WouldYouKindly!

PR #2611, opened 2018-04-11
Added support for configuration through ~/.config/xonsh/rc.xsh

### LINKED ISSUE #2609: XDG Directory Specification
Xonsh already supports XDG Directory Specification. So why doesn't xonsh implement it also for per-user run control file `(~/.xonshrc)`?

scopatz:
Should resolve #2609

### DIFF
--- a/news/xdgconf.rst
+++ b/news/xdgconf.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:**
+
+* Xonsh run control now also looks for the XDG-compliant file
+  ``~/.config/xonsh/rc.xsh`` at startup.
+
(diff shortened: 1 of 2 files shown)


gforsyth:
Circle has communication issues.  Looks good.  Thanks, @scopatz!

con-f-use:
Not document in https://xon.sh/xonshrc.html#snippets-for-xonshrc

I almost missed it. Not pointing fingers or anything. I totally understand, that documentation is hard. However, just wanted to mention that this is not the first time a feature was added without any documentation. The awesomeness of xonsh should be discoverable.

PR #2622, opened 2018-04-23
Document path helper on macOS.

jaraco:
Inspired by [this discussion](https://gitter.im/xonsh/xonsh?at=5ade2a2b3fe1be3704cafde0), provide users and package managers with guidance on configuring the path helper under xonsh.

I use the preferred "macOS" moniker even though "OSX" is used in other parts of the document. I'd recommend using macOS throughout the document, but avoided making that change as it's incidental to the primary goal of this contrib.

### DIFF
--- a/docs/osx.rst
+++ b/docs/osx.rst
@@ -20,7 +20,7 @@ You can install xonsh using homebrew, conda, pip, or from source.
 
     $ conda config --add channels conda-forge
     $ conda install xonsh
-    
+
 
 **pip:**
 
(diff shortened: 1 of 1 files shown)


scopatz:
Since this just docs, you don't need a news item.  Thanks for putting this in @jaraco!

scopatz:
LGTM @jaraco - let us know when it is ready to merge

jaraco:
Ready to merge!

PR #2623, opened 2018-04-23
Don't print in color

scopatz:
Before the shell object has been fully instantiated.  Hot fix for code that hasn't been released yet, so does not need news


gforsyth:
Thanks, @scopatz !

Issue #2030, opened 2016-12-21
$BASH_COMPLETIONS Does not Support directories 


mitnk:
In #1888 @blahgeek fixed a delay issue in #1824, and made $BASH_COMPLETIONS do not support dirs anymore.

The original idea of $BASH_COMPLETIONS I think is that if I have this:
```
$ ls /usr/local/etc/bash_completion.d
aria2c                     cargo                      tmux
brew                       task.sh                    youtube-dl.bash-completion
```
And I set `$BASH_COMPLETIONS = ['/usr/local/etc/bash_completion.d']`, then I'll get all apps (brew, aria2c, etc) completion work.

Now I have to put all full paths of these scripts.

If we do want to make them to be files for speed (like #1824 mentioned), we may need to update the docstring of `$BASH_COMPLETIONS` to make it clear.

Any thoughts here?

scopatz:
Yeah, I am not entirely sure about the logic of quitting after the first successful completer is found. Maybe @blahgeek @melund or @astronouth7303 remember.

astronouth7303:
I never messed with completions, sorry.

scopatz:
Sorry about the noise @astronouth7303. 

@mitnk I think what the code it doing now is wrong but I am still not sure what it was fixed like that. I probably need to reread it all.

I think there are two issues here: 1) We should probably at least try to load all of the completions files listed in BASH_COMPLETIONS, and 2) We should decide if we should load directories.

blahgeek:
Hi @scopatz @mitnk Sorry for the delay, I've been busy at work for quite some time and missed all these notifications... Let me explain.

So the basic idea of the new bash completion is to take advantage of bash-completion's lazy loading to improve performance, which is available from bash-completion 2.x.

The single `bash_completion` script would (lazy) load all the completion files for you (in /etc/bash_completion.d and some other dirs). In that case, xonsh should not load all those scripts, just that single script is enough.

The document of BASH_COMPLETIONS also explains this (https://github.com/xonsh/xonsh/blob/master/xonsh/environ.py#L363)

But if this is not working for you, I would recommend:

- check BASH_COMPLETIONS variable, set to the path of `bash_completion` script, for me it's `/usr/share/bash-completion/bash_completion`
- try it in bash: check the completion after `source /usr/share/bash-completion/bash_completion`

mitnk:
> The single bash_completion script would (lazy) load all the completion files 
[...]

PR #2627, opened 2018-04-24
macOs bash_completion documents

bestlem:
This adds documentation for bash_completion to the osx page as per #2030.

I also changed all OSX to macOS in the file as per Apple's naming conventions.
I added Macports instructions.


scopatz:
No news item needed, since it is just docs. Thank @bestlem!

PR #2629, opened 2018-04-25
Fix bug in free_cwd xontrib

melund:
Fixed a bug/regression in the `free_cwd <http://xon.sh/xontribs.html#free-cwd>`__  Xontrib, which caused the prompt to error if the current directory i  deleted/renamed by other processes on Windows. 

Turns out that __xonsh_env__['PWD'] no longer handles pathlike objects well.


gforsyth:
Thanks, @melund!

PR #2630, opened 2018-04-25
Update xontrib description

melund:
This updates xontrib description to leave out the installation section when the xontrib is bundled with xonsh. 

It also adds a 'Usage' section with the command to enable the xontrib. I always thought this was missing from here. 

Finally, it fixes a few issues with building the docs on Windows. 

## New
image

## Old
image

(Disregard the change in fonts. This is from building locally).


scopatz:
Thanks @melund!

PR #2656, opened 2018-05-28
Run os.environ.clear() for test_environ.py

jcrben:
Before I added this I was seeing this in the test failure:
```
E           assert 'LINES' not in {'ANDROID_SDK_ROOT': '/Users/bencreasy/.config/android', 'ANKI_BASE': '/Users/bencreasy/dotfiles/local/anki', 'ANSIBLE_CONFIG': '/Users/bencreasy/.config/ansible', 'ATOM_HOME': '/Users/bencreasy/.config/atom/', ...}

tests/test_environ.py:228: AssertionError
````

I'm not sure if this was actually causing the test to fail, however - it seems that Python is loading some other `default_env()` function?

My original motivation is to fix the package for nixpkgs where I see this:
```
11 failed, 2985 passed, 23 skipped, 4 deselected, 1 warnings in 72.71 seconds =
builder for '/nix/store/1vbsjwvlmclqm1nvrr00nf7wfc3di5z1-xonsh-0.6.4.drv' failed with exit code 1
```

Is there an open issue tracking setting up darwin CI? I might be willing to help out with that. Willing to help a bit in fixing tests if I can figure out how to debug the code...

### DIFF
--- a/tests/test_environ.py
+++ b/tests/test_environ.py
@@ -15,6 +15,8 @@ from xonsh.environ import Env, locate_binary, DEFAULT_ENSURERS, DEFAULT_VALUES, 
 
 from tools import skip_if_on_unix
 
+os.environ.clear()
+
 def test_env_normal():
     env = Env(VAR='wakka')
     assert 'wakka' == env['VAR']


scopatz:
Hey @jcrben - thanks for putting this in!

> Is there an open issue tracking setting up darwin CI?

We actually use Travis to test on macOS.  So if you look at travis you'll see the mac builds there.

It also seems to me that calling `os.environ.clear()` is going to cause a lot of problems with respect to setting up an running subprocesses.  Certain varaibles will need to be maintained.  Maybe we can go back to the underlying issues and look at all of the tests that are failing.  

What is the output of the `xonfig` command?  What is the output of `pytest -v`?  Let's work from there!

jcrben:
Retaining certain variables seems reasonable, but I do hope you can keep an open mind about figuring out those variables and whitelisting them. Anyway:

xonfig:
| xonsh            | 0.5.12              |
| Git SHA          | 8941dfdd            |
| Commit Date      | Jun 7 16:04:39 2017 |
| Python           | 3.6.1               |
| PLY              | 3.7                 |
| prompt toolkit   | 1.0.3               |
| pygments         | 2.1.3               |

py.test -v opener:
platform darwin -- Python 3.6.1, pytest-2.9.2, py-1.4.31, pluggy-0.3.1 -- /Users/bencreasy/.local/pyenv/versions/3.6.1/bin/python3.6
cachedir: .cache
rootdir: /Users/bencreasy/code/open-source/xonsh, inifile: setup.cfg
plugins: xonsh-0.5.12, timeout-1.0.0, flake8-0.5, cov-2.3.0

Should I put the entire log into a gist or something?

jcrben:
OK, I figured out most of the problem. I had an old xonsh installed with pip and the test was importing that code instead of the local project code.

scopatz:
> Retaining certain variables seems reasonable, but I do hope you can keep an open mind about figuring out those variables and whitelisting them. 

Yeah, I'd love to see a PR that does this. 

> OK, I figured out most of the problem. I had an old xonsh installed with pip and the test was importing that code instead of the local project code.

Ahh that makes sense.  Glad you figured it out!

jcrben:
Don't think I'll have time to get into this anytime soon 😄

PR #2660, opened 2018-06-03
Added pygments-cache

scopatz:
This adds the `pygments-cache` project (which is being developed at https://github.com/xonsh/pygments-cache) to xonsh.  The purpose of this project is to enable faster initial searches for pygments components (lexers, styles, etc).  

In xonsh, we use pygments styles in a couple of places. This PR helps reduce the startup time of the interactive modes of xonsh by using cached entries to find styles.  It is hard to get timings on this stuff, but I have estimated the reduction to be ~200 ms.

To maintain no dependencies other than Python, this project has been included in the xonsh source tree.

Let me know if you have any questions!

### DIFF
--- a/news/pc.rst
+++ b/news/pc.rst
@@ -1,0 +1,13 @@
+**Added:**
+
+* Added ``pygments-cache`` project in order to reduce startup time.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,7 @@
 [pytest]
 flake8-max-line-length = 180
+flake8-exclude =
+    xonsh/pygments_cache.py
 flake8-ignore =
     *.py E122
     *.py E402
--- a/xonsh/__init__.py
+++ b/xonsh/__init__.py
@@ -2,7 +2,7 @@ __version__ = '0.6.5'
 
 
 # amalgamate exclude jupyter_kernel parser_table parser_test_table pyghooks
-# amalgamate exclude winutils wizard pytest_plugin fs macutils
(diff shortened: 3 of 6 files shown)


scopatz:
pinging this!

melund:
Looks good. Startup times are very variable on my machine but I don't sense any startup improvements on Windows. But the code looks good to me, and I can see how it should have potential to speed things up. 

Btw. Would it be sensible to use a `literaleval()` in the load function, or is that irrelevant here?

The errors on travis puzzles me a bit. Is that because prompt_toolkit 2 is released? 

On an other note; It seems like PTK2 has moved away from pygments in many ways. Is that something you have considered for this?

melund:
Ok. I see that it never created a cache.. Of course, this is only activated if I use a pygments style. So it does work :)

gforsyth:
Thanks for putting this in, @scopatz!  I'm seeing a reliable 200ms improvement in startup time, which is great.

scopatz:
> Btw. Would it be sensible to use a literaleval() in the load function, or is that irrelevant here?

Ahh that is a great suggestion! Issue opened at https://github.com/xonsh/pygments-cache/issues/2

> It seems like PTK2 has moved away from pygments in many ways. Is that something you have considered for this?

Yeah, they have moved to their own system - which I believe is compatible with pygments to some degree. @jonathanslenders could be more specific. I think it would be nice in general to have an alternative to pygments that was designed for performance.

> I'm seeing a reliable 200ms improvement in startup time, which is great.

@gforsyth - woohoo! that is awesome!

jonathanslenders:
Prompt_toolkit is now agnostic to whatever lexer is used. Internally it has its own data types which are more flexible. First we were using Pygments tokens all over the place, even if lexing was not needed at all.

So, yes, Pygments will always be supported, but if another lexing library is created, we can support that as well.

PR #2666, opened 2018-06-05
non directories in $PATH are now ignored

### LINKED ISSUE #2313: prompt: error: on field 'curr_branch' on running xonsh

gsaga:
A typo on line 97 was the reason behind #2313 

Fixes #2313

### DIFF
--- a/news/fixCommandsCacheGen.rst
+++ b/news/fixCommandsCacheGen.rst
@@ -1,0 +1,7 @@
+**Changed:**
+    built_ins.py:
+        corrected a typo. 
+    
+
+**Fixed:**
+    issue 2313
--- a/news/fixCommandsCacheGen.rst
+++ b/news/fixCommandsCacheGen.rst
@@ -1,19 +1,18 @@
-**Added:** None
+**Fixed:**
 
-**Changed:** None
+* issue 2313
+
+**Changed:**
 
 * built_ins.py
 
 corrected a typo. 
-    
+
+**Added:** None
 
 **Deprecated:** None
 
 **Removed:** None
-
-**Fixed:** None
-
-* issue 2313
 
 **Security:** None
 
--- a/news/fixCommandsCacheGen.rst
+++ b/news/fixCommandsCacheGen.rst
@@ -1,7 +1,19 @@
-**Changed:**
-    built_ins.py:
-        corrected a typo. 
+**Added:** None
+
+**Changed:** None
+
+* built_ins.py
+
+corrected a typo. 
     
 
-**Fixed:**
-    issue 2313
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+* issue 2313
+
+**Security:** None
+
--- a/news/fixCommandsCacheGen.rst
+++ b/news/fixCommandsCacheGen.rst
@@ -4,9 +4,7 @@
 
 **Changed:**
 
-* built_ins.py
-
-corrected a typo. 
+* built_ins.py, corrected a typo. 
 
 **Added:** None
 
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -73,7 +73,7 @@ class CommandsCache(cabc.Mapping):
     @property
     def all_commands(self):
         paths = builtins.__xonsh_env__.get('PATH', [])
-        pathset = frozenset(x for x in paths if os.path.isdir(x))
+        pathset = tuple(x for x in paths if os.path.isdir(x))
         # did PATH change?
         path_hash = hash(pathset)
         cache_valid = path_hash == self._path_checksum
@@ -94,7 +94,7 @@ class CommandsCache(cabc.Mapping):
         if cache_valid:
             return self._cmds_cache
         allcmds = {}
-        for path in reversed(paths):
+        for path in reversed(pathset):
             # iterate backwards so that entries at the front of PATH overwrite
             # entries at the back.
             for cmd in executables_in(path):
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -70,19 +70,22 @@ class CommandsCache(cabc.Mapping):
         else:
             return [name]
 
+    @staticmethod
+    def remove_dups(p):
+        ret = list()
+        for e in p:
+            if e not in ret:
+                ret.append(e)
+        return ret
+    
+
     @property
     def all_commands(self):
         paths = builtins.__xonsh_env__.get('PATH', [])
-        def removeDups(p):
-            ret = list()
-            for e in p:
-                if e not in ret:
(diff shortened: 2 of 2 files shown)


scopatz:
Thanks for putting this in @gsaga!

xonsh/commands_cache.py
```diff
@@ -73,7 +73,7 @@ def get_possible_names(name):
     @property
     def all_commands(self):
         paths = builtins.__xonsh_env__.get('PATH', [])
-        pathset = frozenset(x for x in paths if os.path.isdir(x))
+        pathset = tuple(x for x in paths if os.path.isdir(x))
```
scopatz: This should probably be a set, so that it removes duplicated entries

scopatz: Or maybe a tuple is a better option, but you need to create the tuple after the the set

gsaga: But `frozenset`s can't be reversed. Maybe should just remove the duplications old fashioned way.

gsaga: The problem is that frozensets don't preserve order. Duplications could be removed in EnvPath. Are there any environment path variables which might cause trouble if we remove duplications.

news/fixCommandsCacheGen.rst
```diff
@@ -0,0 +1,7 @@
+**Changed:**
```
gforsyth: For the news entry file, you should leave the unused headers alone (along with their trailing `None`).  items under a header should have a leading `*` followed by a single space before any text

xonsh/commands_cache.py
```diff
@@ -73,7 +73,14 @@ def get_possible_names(name):
     @property
     def all_commands(self):
         paths = builtins.__xonsh_env__.get('PATH', [])
-        pathset = frozenset(x for x in paths if os.path.isdir(x))
+        def removeDups(p):
```
gforsyth: We (and PEP8) prefer snake_case for function names, can this be `remove_dups` please?

xonsh/commands_cache.py
```diff
@@ -73,7 +73,14 @@ def get_possible_names(name):
...
+            ret = list()
+            for e in p:
+                if e not in ret:
+                    ret.append(e)
+            return ret
+        path = removeDups(path)
+        pathset = tuple(x for x in paths if os.path.isdir(x))
```
gforsyth: Not a blocker, but if this is no longer a set I would prefer if the variable name changed to reflect that.

xonsh/commands_cache.py
scopatz: Also, can you please move this function to be a method on the class or to part of the `xonsh.tools` module?  `all_commands` is a performance ciritical pathway and so we don't want to redefine this function each time.

xonsh/commands_cache.py
scopatz: Yeah, I agree! Thanks for bringing it up.  maybe `valid_path` is a better name

scopatz:
Thanks for continuing to bang on this @gsaga!

gsaga:
@scopatz @gforsyth I've updated the pull request.

xonsh/commands_cache.py
```diff
@@ -70,12 +70,22 @@ def get_possible_names(name):
...
 
+    @staticmethod
+    def remove_dups(p):
+        ret = list()
+        for e in p:
+            if e not in ret:
+                ret.append(e)
+        return ret
+    
+
```
scopatz: PEP8: I think there is only a single blank line between methods

scopatz:
This looks a ton better @gsaga! Thanks a million.  It looks like Circle is complaining about a couple of small style errors:

```
=================================== FAILURES ===================================
__________________ FLAKE8-check(ignoring E122 E402 F841 E741) __________________
/home/circleci/project/xonsh/commands_cache.py:80:1: W293 blank line contains whitespace
/home/circleci/project/xonsh/commands_cache.py:82:5: E303 too many blank lines (2)
```

Once these are fixed, I think this is ready to go in!

gsaga:
@scopatz @gforsyth The `all_commands`  performs the following checks to decide if the cache is outdated:

1. $PATH has changed
2. aliases dict has changed
3. **Any** of the directories in path have been modified

And if any of these are true the whole command cache is regenerated. Shouldn't the function regenerate commands cache for only those sources which have been modified?

It would probably be better if the EnvPath constructor converts the `str`s and `bytes` passed to it to `pathlib.Path`, cos path entries may contain multiple path separators, for ex:
    '/home/user/Music' != '/home/user//Music'
Converting strings and bytes to `Path` would probably solve that problem.

scopatz:
> I'll get on it.​

That'd be amazing!

PR #2668, opened 2018-06-06
fixed buggy rst file checker

### LINKED ISSUE #2667: A bug in test/test_news.py
The `test_news.py` script fails with an `index out of bounds error` for this job:
https://travis-ci.org/xonsh/xonsh/jobs/388607928

That's likely a bug in the test script. A file parsing script shouldn't fail with this error. This is the rst file for which it failed:

    **Fixed:**
    
    * issue 2313
    
    **Changed:**
    
    * built_ins.py, corrected a typo. 
    
    **Added:** None
    
    **Deprecated:** None
    
    **Removed:** None
    
    **Security:** None
    

It failed on the second last line. There should be a check before line 40 in `test_news.py` to check if the current entry is the last entry.

One way I could think of is to convert the file to a string based on the number of stars with which it's lines start, so the above file will become "2010201020202020", and then do a regex search on this string.

gsaga:
Uses regex to check changelog file format. 

Fixes #2667

### DIFF
--- a/news/fixTestNews.rst
+++ b/news/fixTestNews.rst
@@ -15,3 +15,4 @@
   and was needlessly tedious
 
 **Security:** None
+
--- a/news/fixTestNews.rst
+++ b/news/fixTestNews.rst
@@ -1,0 +1,17 @@
+**Added:** None
+
+**Changed:**
+
+* test/test_news.py
+  It now uses regex to verify the format of rst files
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Previous implementation had a lot of nested if-elses
+  and was needlessly tedious
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


gsaga:
I'm getting an AssertionError on this line in the travis tests:
https://github.com/xonsh/xonsh/blob/9a9be8b259e31017504790cc8aae05a0bffe0b8e/tests/test_integrations.py#L243

Though I'm not getting any error if I run this test on my machine!

scopatz:
Thanks for putting this in @gsaga!  The assertion error you were seeing was a random runtime failure that sometimes happens due to threading in the test suite. We haven't tracked it down yet.  Restarting the CI usually makes it pass.  So now this is all green.

news/fixTestNews.rst
```diff
@@ -0,0 +1,18 @@
...
+  It now uses regex to verify the format of rst files
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Previous implementation had a lot of nested if-elses
+  and was needlessly tedious
```
scopatz: BTW, the news files are all merged together when a release happens.  This comment doesn't make a lot of sense out of context, so I will trim it down.

PR #2679, opened 2018-06-12
Added repl.it to the docs

scopatz:
Regains the xonsh live idea!  It isn't perfect, but I think it is good enough.  

Thanks again to @astronouth7303! The events system was essential here to routing around some weird bugs in repl.it

### DIFF
--- a/docs/comparison.rst
+++ b/docs/comparison.rst
@@ -1,4 +1,3 @@
-
 Comparison
 ==========
 Xonsh is significantly different from most other shells or shell tools. The following
--- a/docs/devguide.rst
+++ b/docs/devguide.rst
@@ -1,1 +1,1 @@
-.. include:: ../CONTRIBUTING.rst
+../CONTRIBUTING.rst
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,5 +1,6 @@
-
 .. include:: intro.rst
+
+.. include:: tryitnow.rst
 
 .. include:: comparison.rst
 
--- a/docs/tryitnow.rst
+++ b/docs/tryitnow.rst
@@ -1,0 +1,9 @@
+Try It Now!
+===========
+Try out xonsh right here in the browser. Just press the triangle play button in
+the bottom right of the `Repl.it <https://repl.it>`_ frame below.
+**Note:** this is not a fully featured xonsh instance due to repl.it limitations.
+
+.. raw:: html
+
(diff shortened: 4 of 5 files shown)


scopatz:
Feedback is much appreciated!

melund:
This is interesting. It is kind of a shame that much of the shell part is missing. Seems subprocess are not possible on repl.it as well as tab completion. Or am I wrong? That is a big part of what makes up xonsh.

scopatz:
subprocesses worked for me, things like `ls`, `echo`, `python` etc.  It is a little funny because the `xonsh` command doesn't actually get installed.

Yeah, it is a real shame that tab completion fails.  I have opened an issue with them.  A big part of the definicency is that they haven't implement a full xterm terminal, they just kinda did a partial job. My hope is that things improve.

melund:
He he. I was too hasty. `ls` does work for me. Just clear didn't do its thing.

melund:
I think we should just add this. On a side note; I do miss the asciinema video we once had on frontpage. @scopatz Did you make some ghosttyping once. Could we get something like that to produce the asciinema videos. I imagine it was crazy hard to record.

scopatz:
Yeah we could add that too. It wasn't too hard to record, I don't think. The annoying thing was just executing the script correctly.

I am not suprised that clear doesn't work, since it isn't a real terminal. I'll open an issue.

Also it looks like some of the file contents were deleted?! Weird. Maybe this isn't ready for prime-time...

scopatz:
Well, I think I got it working again.

scopatz:
Apperently they are working on ANSI escape codes: https://repl.it/bugs/p/clear-command-doesnt-work-in-terminal  Hopefully all of the current issues will be fixed when that occurs.

timmyichen:
Hey all, Tim from Repl.it here.  Just to give some more details and context, right now we're using [jqconsole](https://github.com/replit/jq-console) (which we created a while ago) but are moving to [xtermjs](https://xtermjs.org/) to provide bash support + ANSI escape sequences, etc.  Let me know if you'd be interested in testing it when we have a beta version ready. :)

scopatz:
Hey @timmyichen! Thanks for reaching out. I really appreciate all of the hard work that your team has put into repl.it!  We'd love to test it when there is a beta version available. That'd be great. I'm glad youare moving to xtermjs!

melund:
I am all for getting this on the development docs now.

scopatz:
Thanks @melund! It would be nice to put out a release this week. Are there any objections to having this hit the front page?

PR #2683, opened 2018-06-16
Fix `$CDPATH` completion misfire

### LINKED ISSUE #2677: Paths relative to `CDPATH` are completed even for commands that aren't `cd`, and are buggy
Steps to reproduce:
1. Navigate to an empty directory
2. Set your `CDPATH` to include your home directory
3. Type `cat `
4. Start typing the name of some directory in your home, then press  tab
5. It will be completed, and a space will be inserted after instead of the usual slash
6. Delete the space, add a slash and start typing the name of a file or a directory in said directory, then press tab again
7. The whole thing will be replaced with the name of the file/subdirectory, once again, with a space after.

This is probably two bugs, but I can't reproduce one without the other.

Xonfig:
```
| xonsh            | 0.6.6.dev1          |
| Git SHA          | 4c78866e            |
| Commit Date      | Jun 6 18:21:58 2018 |
| Python           | 3.6.5               |
| prompt toolkit   | 1.0.15              |
| distro           | debian              |
```

gforsyth:
The `path` completer runs for many commands (which is the correct
behavior) but we are adding in possible `CDPATH` completions for commands
other than `cd` which doesn't make any sense since only `cd` supports
path resolution via `CDPATH`.

This adds a naive check to only add in the `CDPATH` completions if the
current command is `cd`.

This resolves #2677 

There's a trade-off here.  It will prevent invalid path completion for non `cd` commands but will also disable path competion relative to `CDPATH` if `cd` is part of a chained command with an `and` or `or`

### DIFF
--- a/news/cdpath_completion_fix.rst
+++ b/news/cdpath_completion_fix.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* completions relative to ``CDPATH`` only trigger when used with ``cd``
+
+**Security:** None
--- a/tests/test_path_completers.py
+++ b/tests/test_path_completers.py
@@ -1,7 +1,15 @@
+import builtins
+
 import pytest
+from unittest.mock import patch
 
 from xonsh.environ import Env
 import xonsh.completers.path as xcp
+
+
+@pytest.fixture(autouse=True)
+def xonsh_execer_autouse(xonsh_builtins, xonsh_execer):
+    return xonsh_execer
 
 
 def test_pattern_need_quotes():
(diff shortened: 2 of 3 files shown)


scopatz:
Thanks for putting this in @gforsyth! I'd like to suggest something slightly more sophisticated in order to manage the trade off. Rather, than looking at just the first split (which would also fail for cases like `![cd /dir]`), we could use xonsh's tokenizer to check if cd is a command that is used. I'd suggest another function, something like:

```python
def cd_in_command(line):
    lexer = builtins.__xonsh_execer__.parser.lexer
    lexer.reset()
    lexer.input(line)
    have_cd = False
    for tok in lexer:
        if tok.type == 'NAME' and tok.value == 'cd':
            have_cd = True
            break
    return have_cd
```

You could even extend this to look for aliases of `cd`, but that probably isn't necessary for the moment.

gforsyth:
Hey @scopatz -- good call.  I just realized, too, that the naive way would break the `CDPATH` completions in a multi-line code block which is not cool.  I've added the checker you suggested.

scopatz:
Thanks a million @gforsyth!

PR #2685, opened 2018-06-17
Remove duplicate `pip` completer

gforsyth:
`xpip` just uses the regular `pip` completer and it doesn't need its own entry
in the `__xonsh_completers__` dict.  That will just run the same checks twice
and slow down completions overall.


scopatz:
Thanks @gforsyth!

PR #2686, opened 2018-06-17
Update py-bash-completion with prefix length fix

gforsyth:
Porting over from `xonsh/py-bash-completion`

Should fix the assertion errors popping up in the prompt toolkit shell


scopatz:
Thanks again @gforsyth!

PR #2693, opened 2018-06-25
Windows redirection bug fix

robdefriesse:
The call to os.set_handle_inheritable expects a Windows handle, but it is being passed a posix file descriptor.
fileno() returns a posix file descriptor which is a small integer. The method os.set_inheritable() does accept a posix file descriptor. The fix is to simply replace set_handle_inheritable to set_inheritable.

This PR includes an regression test that will fail without the fix described.

### DIFF
--- a/news/windows_redirection.rst
+++ b/news/windows_redirection.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+Fixed error when using redirection (e.g., >) on Windows. #2470
+
+**Security:** None
--- a/news/windows_redirection.rst
+++ b/news/windows_redirection.rst
@@ -8,6 +8,6 @@
 
 **Fixed:**
 
-Fixed error when using redirection (e.g., >) on Windows. #2470
+* Fixed error when using redirection (e.g., >) on Windows.
 
 **Security:** None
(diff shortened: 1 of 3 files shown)


news/windows_redirection.rst
```diff
@@ -0,0 +1,13 @@
...
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+Fixed error when using redirection (e.g., >) on Windows. #2470
```
melund: @robdefriesse I think the checker is a little picky about the format in the news file. Hence the failures. The line must begin with an asterix (eg. `* Fi...` a). Also,you can remove the hashtag for the PR, that will not be rendered nicely anyway.

melund: Hopefully, that should fix the failures.

robdefriesse:
Thank you, @melund.  I made the edit to fix the news file on my computer and pushed to my fork.  Is that all I need to do, or do I need to push from my fork as well?

Thank you.

gforsyth:
Hey @robdefriesse -- thanks for this!  Pushing the changes up to your fork automatically updates the PR here, so we're all set.  And it looks like all the CI has passed.  Merging!

robdefriesse:
Thank you @scopatz, @melund, and @gforsyth for your assistance.

scopatz:
Great to see this go in! Thanks!

PR #2694, opened 2018-06-28
add indir and subprocess tools

CJ-Wright:


CJ-Wright:
I think there may be an issue here, since usually we need to run `xonsh.main.setup()` in the `__init__.py` to get python to recognize xonsh modules, but that may run into recursion? Or maybe this needs some amalgam magic? I'm a bit stuck at this point.

CJ-Wright:
Attn: @scopatz

xonsh/extra_tools.xsh
```diff
@@ -0,0 +1,25 @@
...
+    ![cd @(old_d)]
+
+
+def run(cmd, cwd=None, check=False):
+    """Stub for ``subprocess.run`` like functionality"""
+    if cwd is None:
+        cwd = '.'
+    with indir(cwd), ${...}.swap(RAISE_SUBPROC_ERROR=check):
+        ![@(cmd)]
+
```
scopatz: PEP8 - two lines between functions

xonsh/extra_tools.xsh
```diff
@@ -0,0 +1,25 @@
+import os
```
scopatz: Please add a docstring

scopatz:
Thanks for putting this in @CJ-Wright!  Here are a few thoughts:

1. You didn't seem to modify the setup.py to actually install `*.xsh` files, so that may be why the tests are failing.
2. Documentation hooks should be added.
3. I kind of like the idea of having a standard library subpackage (maybe `lib`) that os made up of `*.xsh` files for use by down stream packages.  To this end, it makes sense to split up what is in `extra_tools` into and `os.xsh` and `subprocess.xsh` files.

Thanks again!

xonsh/lib/os.xsh
```diff
@@ -0,0 +1,20 @@
+"""Xonsh extension of the standard library os module, using xonsh for
+subprocess calls"""
+import sys
+
+from xonsh.imphooks import install_hook
+
+install_hook
```
scopatz: I don't think you need to do anything with install hooks here.

xonsh/lib/os.xsh
```diff
@@ -0,0 +1,20 @@
...
+
+import os
+from contextlib import contextmanager
+
+@contextmanager
+def indir(d):
+    """Context manager for temporarily entering into a directory.
+     Copyright (c) 2017, Anthony Scopatz
+     Copyright (c) 2018, The Regro Developers"""
```
scopatz: I would throw the copyrights into comments, rather than the docstring.  They don't really help the user. Also, you can leave off mine :)

xonsh/lib/subprocess.xsh
```diff
@@ -0,0 +1,19 @@
+"""Xonsh extension of the standard library subprocess module, using xonsh for
+subprocess calls"""
+from xonsh.imphooks import install_hook
+
+install_hook()
```
scopatz: Same here, you don't need the install hook, since you have it in the `__init__.py`

xonsh/lib/os.xsh
```diff
@@ -0,0 +1,20 @@
...
+import sys
+
+from xonsh.imphooks import install_hook
+
+install_hook
+
+import os
+from contextlib import contextmanager
+
+@contextmanager
```
scopatz: two blank lines after imports

xonsh/lib/subprocess.xsh
```diff
@@ -0,0 +1,19 @@
...
+
+from xonsh.tools import XonshCalledProcessError as CalledProcessError
+from xonsh.lib.os import indir
+
+def run(cmd, cwd=None, check=False):
```
scopatz: two blank lines after import

xonsh/lib/subprocess.xsh
```diff
@@ -0,0 +1,19 @@
...
+    """Stub for ``subprocess.run`` like functionality"""
```
scopatz: This isn't really a stub, it is more of a drop-in replacement (where not all functionality has been implemented)

xonsh/lib/subprocess.xsh
```diff
@@ -0,0 +1,19 @@
...
+    if cwd is None:
+        cwd = '.'
+    with indir(cwd), ${...}.swap(RAISE_SUBPROC_ERROR=check):
+        ![@(cmd)]
+
+def check_call(cmd, cwd=None):
+    """Stub for ``subprocess.check_call`` like functionality"""
```
scopatz: Likewise, this is a replacement

xonsh/lib/subprocess.xsh
```diff
@@ -0,0 +1,19 @@
...
+    run(cmd, cwd=cwd, check=True)
```
scopatz: Newline at the end of the file please.

xonsh/lib/os.xsh
```diff
@@ -0,0 +1,20 @@
...
+def indir(d):
```
gforsyth: I think using `pushd` and `popd` will work better here.  Then you don't have to keep track of the "old" directory

xonsh/lib/subprocess.xsh
scopatz: Also, Python's `check_call()` returns the return code (0, an int), not None, as here.

scopatz: ... if it runs sucessfully.

xonsh/lib/subprocess.xsh
```diff
@@ -0,0 +1,19 @@
...
+
+from xonsh.tools import XonshCalledProcessError as CalledProcessError
+from xonsh.lib.os import indir
+
+def run(cmd, cwd=None, check=False):
+    """Stub for ``subprocess.run`` like functionality"""
+    if cwd is None:
+        cwd = '.'
+    with indir(cwd), ${...}.swap(RAISE_SUBPROC_ERROR=check):
+        ![@(cmd)]
```
scopatz: `run()` should return the commandpipeline object, so 

```
with ...:
    p = ![@(cmd)]
return p
```

CJ-Wright:
I think one of the travis builds needs to be re-run (conda didn't install)

xonsh/lib/os.xsh
```diff
@@ -0,0 +1,12 @@
+"""Xonsh extension of the standard library os module, using xonsh for
+subprocess calls"""
+from contextlib import contextmanager
+
+
+@contextmanager
+def indir(d):
+    """Context manager for temporarily entering into a directory."""
+    # Copyright (c) 2018, The Regro Developers
```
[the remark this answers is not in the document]
CJ-Wright: Most likely not I'll remove it.

xonsh/lib/subprocess.xsh
```diff
@@ -0,0 +1,19 @@
...
+        cwd = '.'
+    with indir(cwd), ${...}.swap(RAISE_SUBPROC_ERROR=check):
+        p = ![@(cmd)]
+    return p
+
+
+def check_call(cmd, cwd=None):
+    """Drop in replacement for ``subprocess.check_call`` like functionality"""
+    run(cmd, cwd=cwd, check=True)
+    return 0
```
scopatz: This isn't exactly what I meant :)  I would like to see, 

```
p = run(...)
return p.returncode
```

news/indir.rst
```diff
@@ -0,0 +1,14 @@
+**Added:**
+
+* ``indir`` a context manager for temporarily entering into a directory
+* ``run`` and ``check_call`` subprocess stubs using ``xonsh`` as the backend
```
scopatz: Can you give the full object path here, ie `xonsh.lib.indir`.  It would also probably be good to start with an item that says something like ```* New xonsh standard library `xonsh.lib` subpackage```

xonsh/lib/subprocess.xsh
CJ-Wright: Ok, I was trying to mirror the extract syntax offered by check call.

scopatz: Right, so my concern is that if for some reason run returns with a non-zero return code, the current version of `check_call()` would mask us from ever knowing.

CJ-Wright:
@scopatz @gforsyth I think this is ready for another round!

scopatz:
Thanks @CJ-Wright!

PR #2701, opened 2018-07-05
Bash completion spacetime continuum

gforsyth:
*Note*: This PR is tracking xonsh/py-bash-completion#8 

The current behavior of this code in xonsh right now is as follows:
```
mkdir -p "this is a test"  # in otherwise empty dir to avoid multiple
matches
ls th[tab] -> ls t'this is a test/'
ls 'th[tab] -> ls 't'this is a test/'
ls "th[tab] -> no results
```

Initially the leading double-quote seemed to work in xonsh but that is
actually the bash completer failing completely and then xonsh falling
back on to the path completer which works correctly for this completion.
Wild.

If I directly import the `bash_completions` function in the same test
directory with that folder name and call it directly:

```
gil@bad_cat ~/tmp shell  bash_completions('"th', 'ls "th', 5, 6)
(set(), 0)
gil@bad_cat ~/tmp shell  bash_completions("'th", "ls 'th", 5, 6)
({"'this is a test/'"}, 1
```

Note that the double-quoted lead returns nothing.  Also, the
single-quoted lead says that the number of characters to strip off of
the prefix is 1, which will leave a leading `'t` as evidenced above.

So, two main fixes in here:

* Make the underlying bash completion script also look for leading
double-quotes
* Correctly count the strip-length depending on if there are or aren't
leading quotes (irrespective of quote type)

### DIFF
--- a/news/bash_completion_spacetime_continuum.rst
+++ b/news/bash_completion_spacetime_continuum.rst
@@ -1,0 +1,15 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fixed issue with incorrect strip lengths for prefixes with quotes in them
+* Fixed bash script to also consider leading double quotes and not just single
+  quotes
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


melund:
This looks good @gforsyth. Thanks for fixing this.

PR #2704, opened 2018-07-06
Change default for `AUTO_SUGGEST_IN_COMPLETIONS`

gforsyth:
Up for debate -- I think it causes more confusion since it provides
"wrong" completions.

### DIFF
--- a/news/change_auto_suggest_default.rst
+++ b/news/change_auto_suggest_default.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:**
+
+* set default value of ``$AUTO_SUGGEST_IN_COMPLETIONS=False``
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks for putting this in @gforsyth! I don't recall, does this only affect PTK?

gforsyth:
Yeah, this is only for ptk stuff

melund:
I agree. It is a bit annoying and confusing. I have this to `False` in my .xonshrc, so I support changing the default behaviour here.

gforsyth:
Ping! Anyone have any arguments for or against this?

scopatz:
So IIRC the argument agains this (from @asmeurer originally) is that if you have the auto-suggest up and then press TAB, you won't complete against the next token that is displayed on the screen.

melund:
I am for this. But I do see the other argument. The problem is really that new users don't understand the concepts, and don't know how to use the ctrl-e history completion. 

Maybe history completion  should be off by default  and the ghosted text should just show the first normal entry from tab completion. 

Then we could have an option to use the ghosted text for separate  history completion for those who understand  concepts.

scopatz:
I am happy to be out-voted here, esp. since the capability is still there.

scopatz:
Thanks @gforsyth!

PR #2714, opened 2018-07-14
SciPy sprint new aliases.

AaronV77:
Added an aliases to clear the pythonic variables within the xonsh shell.


scopatz:
Thanks @AaronV77! I will fix up flake error on master

PR #2715, opened 2018-07-14
WIP SciPy sprint, printing special methods to prompt.

AaronV77:
Added a hook for a special method printing on an object. The method is called xonsh_display.

### DIFF
--- a/xonsh/pretty.py
+++ b/xonsh/pretty.py
@@ -115,6 +115,9 @@ def pretty(obj, verbose=False, max_width=79, newline='\n', max_seq_length=MAX_SE
     """
     Pretty print the object's representation.
     """
+    if hasattr(obj, 'xonsh_display'):
+        obj.xonsh_display()
+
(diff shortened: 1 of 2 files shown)


xonsh/pretty.py
```diff
@@ -115,6 +115,9 @@ def pretty(obj, verbose=False, max_width=79, newline='\n', max_seq_length=MAX_SE
     """
     Pretty print the object's representation.
     """
+    if hasattr(obj, 'xonsh_display'):
+        obj.xonsh_display()
```
scopatz: I think this needs to `return obj.xonsh_display()`

scopatz:
Thanks for putting this in @AaronV77!

xonsh/pretty.py
AaronV77: Yes, I copied my example and I had it printing from the xonsh_display().

scopatz:
Thanks @AaronV77!

PR #2729, opened 2018-07-18
Fixed a regression with xonfig styles

melund:
The "xonfig styles" command was reporting
AttributeError: module 'pygments' has no attribute 'styles'

Now we import the styles package from pygments explicitly

### DIFF
--- a/news/fix-xonfig-styles.rst
+++ b/news/fix-xonfig-styles.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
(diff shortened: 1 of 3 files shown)


gforsyth:
@scopatz has made me terrified of all pygments imports.  I'll defer to him on this one.
That said, working is better than broken.

And, for what it's worth, this particular import seems to only cost around `5e-05` seconds on my work laptop.

gforsyth:
Thanks for putting this in, @melund !

melund:
> @scopatz has made me terrified of all pygments imports. 

I agree. I this case I don't think it can't be avoided, since we really want to know what styles pygments provide. Also, this is really only called when the users interact with the `xonfig` command.

scopatz:
This looks correct to me!  There is probably a faster way to do this in `pygments_cache`, but it hasn't been exposed yet. Thanks @melund!

PR #2731, opened 2018-07-19
Fix regression for INTENSIFY_COLORS_ON_WIN due to PTK2

melund:
This fixes a problem with the [`INTENSIFY_COLORS_ON_WIN` hack](http://xon.sh/envvars.html#intensify-colors-on-win), which stopped working when we add PTK2 support. 

The hack is still useful for older versions of Windows. 
But there is a nicer workaround on Win10 where PTK2 now uses ANSI escapes to give full color support. The default colour scheme in the terminal still have unreadable blue and red colors. Instead, we could create a win10 style with hard-coded colour values, and use that as the default style on Win10. I will try to look into that.

### DIFF
--- a/news/color_win.rst
+++ b/news/color_win.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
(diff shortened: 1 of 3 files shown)


scopatz:
This looks great @melund! Just waiting on CI

scopatz:
Thanks @melund! This is great!

PR #2732, opened 2018-07-19
Python 37

scopatz:
This fixes some Python v3.7 bugs and also enables CI for Python 3.7.  

I guess this is as good of a time to ask as any if we should drop CI for Python 3.4 and/or Python 3.5.


scopatz:
Looking at https://docs.python.org/3/whatsnew/3.7.html, this actually seems to the only changes needed for Python 3.7.  Also! It looks like we are working on nightly again!

melund:
Awesome.  Thanks @scopatz

Issue #2734, opened 2018-07-20
xonsh version 0.7.0 does not start with prompt_toolkit >= 2.0 when no pygments is installed


Achimh3011:
xonsh 0.7.0 was installed though `brew upgrade`. Release notes for 0.7.0 state that prompt_toolkit>=2.0 is supported.  However, xonsh does not start with `prompt_toolkit==2.0.3` (falling back to bash).  With `prompt_toolkit==1.0.15` xonsh starts , but also shows an error message (due to API changes).

## xonfig
$ xonfig

_As xonsh doesn't start with prompt_toolkit installed, this xonfig output was generated without it. xonsh complains about prompt_toolkit missing, but at least it starts._

````
+------------------+-----------------+
| xonsh            | 0.7.0           |
| Python           | 3.7.0           |
| PLY              | 3.9             |
| have readline    | True            |
| prompt toolkit   | None            |
| shell type       | readline        |
| pygments         | None            |
| on posix         | True            |
| on linux         | False           |
| on darwin        | True            |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
+------------------+-----------------+
````

## Expected Behavior
xonsh should start up (and work) without error message.

## Current Behavior

`$` means running xonsh, `>` means bash prompt.

From a xonsh installation w/o prompt_toolkit:

````
$ xpip install 'prompt_toolkit==2.0.3'
Collecting prompt_toolkit==2.0.3
  Using cached https://files.pythonhosted.org/packages/bb/c5/e5c7098fa2441d669bba512c4b7a6291333db4af218e4365e73b7919c9ee/prompt_toolkit-2.0.3-py3-none-any.whl
Requirement already satisfied: wcwidth in /usr/local/Cellar/xonsh/0.7.0/libexec/lib/python3.7/site-packages (from prompt_toolkit==2.0.3) (0.1.7)
Requirement already satisfied: six>=1.9.0 in /usr/local/Cellar/xonsh/0.7.0/libexec/lib/python3.7/site-packages (from prompt_toolkit==2.0.3) (1.11.0)
Installing collected packages: prompt-toolkit
Successfully installed prompt-toolkit-2.0.3
$ ^D
> env XONSH_DEBUG=1 xonsh
Traceback (most recent call last):
[...]

PR #2736, opened 2018-07-20
Make ptk xontribs compatible with ptk1 and ptk2

gforsyth:
This should take care of the remaining issues raised in #2734 -- all of the xontribs (that we bundle) that make use of custom keybindings should now work in both ptk1 and ptk2


scopatz:
Lgtm! Thanks @gforsyth!

PR #2737, opened 2018-07-20
Update pytest syntax in setup.cfg

gforsyth:
This warning has been popping up for ages, so I've finally made the
change.  Time to see what else breaks.

No news entry needed here.


scopatz:
Thanks @gforsyth! This is indeed long overdue!

PR #2739, opened 2018-07-21
fixed $ARG<N> subprocess issue

### LINKED ISSUE #2718: $ARG1 doesn't work in subprocess mode
It seems like $ARGS works in both subprocess and python mode, but naming a specific argument such as $ARG1 does not. For example:

```
$ cat a.xsh
#!/usr/bin/env xonsh

echo $ARGS
echo $ARG1

print($ARGS)
print($ARG1)

$ ./a.xsh 1 2 3 4
['./a.xsh', '1', '2', '3', '4']
$ARG1
['./a.xsh', '1', '2', '3', '4']
1
```

xonfig output:
```
$ xonfig
| xonsh            | 0.6.8                |
| Git SHA          | a88a485f             |
| Commit Date      | Jun 28 20:10:30 2018 |
| Python           | 3.5.2                |
| PLY              | 3.9                  |
| prompt toolkit   | 1.0.15               |
| distro           | neon                 |
```

scopatz:
Should fix #2718

### DIFF
--- a/news/argn.rst
+++ b/news/argn.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Fixed issue with ``$ARG<N>`` varaibles not being passed to subprocesses correctly.
+
+**Security:** None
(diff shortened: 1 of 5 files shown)


gforsyth:
Looks good to me, just waiting on CI

PR #2741, opened 2018-07-21
Pygments get_all_styles

melund:
Add the `get_all_styles()`  function which @scopatz added to the ``pygments_cache`` module

### DIFF
--- a/news/pygments_cache.rst
+++ b/news/pygments_cache.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:**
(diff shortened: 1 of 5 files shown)


melund:
The test errors seem to be flake8 not respecting the line which should exclude `pygments_cache.py` https://github.com/xonsh/xonsh/blob/master/setup.cfg#L3-L4

Any ideas why that could be?

scopatz:
Thanks a million @melund!

Issue #2746, opened 2018-07-26
Add option for history builtin to output null-delimited results


laloch:
As of now the `history show` builtin can only print EOL-delimited list of commands. This makes it very hard for external filters to properly handle multi-line commands. For instance this example from the Bash to Xonsh Translation Guide:
```
with ${...}.swap(N=V):
    command
```
if used like `history show | fzf` will inevitably end up in fzf like
```
>'with ${...}.swap(N=V):'
>'    command'
```
Please consider adding an optional parameter to switch the output to null-delimited format.

scopatz:
This has been implemented in #2747

PR #2747, opened 2018-07-26
history show - add optional null-delimited output

laloch:
Resolves feature request #2746 - "Add option for history builtin to output null-delimited results"
For example usage see [this commit](https://github.com/laloch/xontrib-fzf-widgets/commit/313e163767dbe80d3ecc5b36fde980113d2032e5) in my fork of xontrib-fzf-widgets.

### DIFF
--- a/news/null-delim-history.rst
+++ b/news/null-delim-history.rst
@@ -1,7 +1,7 @@
 **Added:** None
 
 * ``history show`` builtin now supports optional ``-0`` parameter that switches
-  the output to null-delimited. Useful for pipng history to external filters.
+  the output to null-delimited. Useful for piping history to external filters.
 
 **Changed:** None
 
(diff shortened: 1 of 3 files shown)


TeamSpen210:
Any particular reason why you didn't flip your if-elses? That would make it a bit more clear without the negation.

laloch:
Well, yes. I don't particularly like the Python's ternary syntax, because it's just too "chatty" to my taste, but in this case it seemed to me to suite the need and I prefer the default case ('\n') to always go first regardless of the language.

gforsyth:
Hey @laloch -- thanks for putting this in!  Could you please add a news entry for our changelog?
http://xon.sh/devguide.html#changelog

Thanks!

laloch:
Done. Sorry for not doing it right away.

scopatz:
Thanks @laloch!

PR #2748, opened 2018-07-28
Return to original directory if exception during indir

justcalamari:


justcalamari:
@scopatz It would be great to get this in (+ a new release) ASAP since it is holding up a lot of migrations. I can also add a test for this special case if you want.

scopatz:
Thanks so much for this! I plan on doing a release on Monday, is that soon enough?

PR #2752, opened 2018-07-31
Do not mixin default pygment style on PTK2

melund:
This fixes a problem with styles in PTK2. PTK2 (by default) merges in the default pygments style with the style we provide. 

This is not what we want since the xonsh style is already complete.


gforsyth:
Thanks, @melund !

PR #2754, opened 2018-08-01
Win 10 color fixes

melund:
This PR adds a new way to compensate for unreadable default terminal colors in Win10 with prompt toolkit 2. 

Previously, we would swap the unreadable colors with more readable alternatives (dark blue->cyan). 

But there is a better alternative now since Windows 10 now support VT100 escapes and True colour in the terminal with PTK2. 

With this PR we swap all terminal ansi-colours with hard-coded colors.  The system is still controlled by the `$INTENSIFY_COLORS_ON_WN`  environment variable. But it uses the new approach if VT-100 support is available. 

This PR also removes support for pygments <2.2 because it didn't support ansi-colours and it was a mess. 

( Don't merge this yet. I still need to test this on different older versions of WIN)

Xonsh experience with `$INTENSIFY_COLORS_ON_WN=False`:
image

Xonsh experience with `$INTENSIFY_COLORS_ON_WN=True` (default):
image


scopatz:
Wow, this looks great @melund! Let us know when you think it is no longer WIP

melund:
@scopatz This is ready now. 

It builds on #2758 which needs to be merged first...

scopatz:
Thanks again @melund!

PR #2756, opened 2018-08-02
Jupyter Xernel

### LINKED ISSUE #2712: Issues with the Jupyter kernel
The Jupyter kernel seems nice, but it doesn't seem to work too well in a notebook:

- Output isn't printed until the command exits
- Commands like a git clone that clear the line aren't cleared (compare this to the standard IPython `!git clone`)
- It doesn't know that the output isn't a tty. So something like `git log` will not work, and it messes up the terminal that the notebook is running from

scopatz:
This fixes #2712, in so far as that issue is actually addressable. 

I have completely removed the dependence on `ipykernel`, and instead forked the pyzmq-based [simple kernel](https://github.com/dsblank/simple_kernel/blob/master/simple_kernel.py).  I needed to do this in order to avoid some of the AST transformations that were interfering with xonsh's syntax. 

Things should more or less be working now.

### DIFF
--- a/xonsh/jupyter_kernel.py
+++ b/xonsh/jupyter_kernel.py
@@ -226,7 +226,8 @@ class XonshKernel:
             parts = identities + parts
         self.dprint(3, "send parts:", parts)
         stream.send_multipart(parts)
-        stream.flush()
+        if isinstance(stream, zmqstream.ZMQStream):
+            stream.flush()
 
     def deserialize_wire_message(self, wire_message):
         """Split the routing prefix and message frames from a message on the wire"""
@@ -295,10 +296,12 @@ class XonshKernel:
     def handle_execute_request(self, message, identities):
         """Handle execute request messages."""
         self.dprint(2, "Xonsh Kernel Executing:", pformat(message['content']["code"]))
+        # Start by sending busy signal
         content = {'execution_state': "busy"}
         self.send(self.iopub_stream, 'status', content,
                   parent_header=message['header'])
 
+        # confirm the input that we are executing
         content = {
             'execution_count': self.execution_count,
             'code': message['content']["code"],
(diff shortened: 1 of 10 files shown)


xonsh/jupyter_kernel.py
```diff
@@ -1,72 +1,398 @@
...
+
+    def control_handler(self, wire_message):
+        """Handles control requests"""
+        self.dprint(1, "control received:", wire_message)
+        identities, msg = self.deserialize_wire_message(wire_message)
+        if msg["header"]["msg_type"] == "shutdown_request":
+            self.shutdown()
+
+    def stdin_handler(self, message):
+        dprint(2, "stdin received:", message)
```
gforsyth: This should be `self.dprint`

xonsh/jupyter_kernel.py
```diff
@@ -1,72 +1,398 @@
...
+        stream.send_multipart(parts)
+        if isinstance(stream, zmqstream.ZMQStream):
+            stream.flush()
+
+    def deserialize_wire_message(self, wire_message):
+        """Split the routing prefix and message frames from a message on the wire"""
+        delim_idx = wire_message.index(DELIM)
+        identities = wire_message[:delim_idx]
+        m_signature = wire_message[delim_idx + 1]
+        msg_frames = wire_message[delim_idx + 2 :]
```
gforsyth: pep8 space before colon

xonsh/jupyter_kernel.py
```diff
@@ -1,72 +1,398 @@
...
+                    continue
+                else:
+                    raise
+            else:
+                break
+
+    def shell_handler(self, message):
+        """Dispatch shell messages to their handlers"""
+        self.dprint(1, "received:", message)
+        position = 0
```
gforsyth: `position` is unused

xonsh/jupyter_kernel.py
scopatz: Unfortunately, this is technically [valid PEP8](https://www.python.org/dev/peps/pep-0008/#pet-peeves), 

> However, in a slice the colon acts like a binary operator, and should have equal amounts on either side (treating it as the operator with the lowest priority)

I just used `black` to reformat the file.  flake8 incorrectly says that this is an error :(

scopatz:
> I'm looking forward to testing it out!

Yay!

xonsh/jupyter_kernel.py
gforsyth: That's obnoxious.

scopatz: Yeah it is pretty much the only part of black I don't like... But since I don't actually have to write code that way , I don't think I mind it :)

scopatz:
@melund, yeah it is totally great for that!

scopatz:
Pinging this issue, hoping for a merge :)

melund:
@scopatz I tried it briefly the other day, but it didn't work on Windows. I thought that maybe it was just picking up the old version and I didn't have time to check. 

Just tried it again now, and I still get nothing back when running subprocesses on Windows. 

image

Since @gforsyth confirms that it works on Linux, I don't mind merging it. Then we can dig into the Win problem later.

scopatz:
Ahh OK! Thanks for noting @melund! I'll make an issue.

PR #2761, opened 2018-08-09
Jupyter printing with cat

scopatz:
This addresses some of the further issues brought up in #2712

PR #2763, opened 2018-08-10
Fixed a bug when converting new ptk2 ansi-color names to old names

melund:
While reviewing @scopatz's #2761 I stumble on a bug I think I caused. 

First, the new PTK ansi-color names should also be converted back when the shell-type is 'jupyter'. This fixes a bug with the pygments and the `XonshTerminal256FormatterProxy` in the notebook.

Fixing this, it turned out there was another bug in the code causing ansi-color names to be converted multiple times. Resulting in bugs like: `wrong color format '##ansidarkblue'` 

This PR fixes both. @scopatz you should probably merge it into #2761

### DIFF
--- a/news/convert_color_names.rst
+++ b/news/convert_color_names.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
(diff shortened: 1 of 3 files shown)


scopatz:
Thanks a million @melund!

PR #2768, opened 2018-08-13
Fix recent history item ordering in ptk2 shell

### LINKED ISSUE #2766: History not saved from session to session on xonsh 0.7.3
I now use xonsh as my main shell, and whenever I quit, I no longer get the history from the previous session when I hit "up" on the keyboard. Here is a good example:

```
~ $ echo "lol"                                                                                                                      
lol
~ $ exit
(reopen shell, type )
~ $ (push 'up' on keyboard)
~ $ ls (random command from days ago)
```

The history that it brings up is from days ago. I think history started getting weird once I installed prompt_toolkit version 2. Could this be a permissions issue in saving the history to a file?

Here is my xonfig

```
| xonsh            | 0.7.3               |
| Git SHA          | 9bbbf07b            |
| Commit Date      | Aug 2 18:08:44 2018 |
| Python           | 3.5.2               |
| prompt toolkit   | 2.0.4               |
| distro           | neon                |
```

gforsyth:
Looking at the existing code, it seems like history entries should always be
loaded oldest first (which hasn't been happening, so whaaa?). This is a fix to
reverse the sort order so that we get the newest entries first when browsing
through command history.

Should fix #2766

### DIFF
--- a/xonsh/history/json.py
+++ b/xonsh/history/json.py
@@ -384,7 +384,7 @@ class JsonHistory(History):
         for item, tss in zip(self.inps, self.tss):
             yield {'inp': item.rstrip(), 'ts': tss[0]}
 
-    def all_items(self, **kwargs):
+    def all_items(self, reverse=False, **kwargs):
         """
         Returns all history as found in XONSH_DATA_DIR.
 
@@ -392,7 +392,7 @@ class JsonHistory(History):
         """
         while self.gc and self.gc.is_alive():
             time.sleep(0.011)  # gc sleeps for 0.01 secs, sleep a beat longer
-        for f in _xhj_get_history_files():
+        for f in _xhj_get_history_files(reverse=reverse):
             try:
                 json_file = xlj.LazyJSON(f, reopen=False)
             except ValueError:
--- a/news/fix_ptk2_reverse_history.rst
+++ b/news/fix_ptk2_reverse_history.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* Recent command history in ptk2 prompt now returns most recently executed
+  commands first (as expected)
+
+**Security:** None
--- a/xonsh/ptk2/history.py
+++ b/xonsh/ptk2/history.py
@@ -25,7 +25,7 @@ class PromptToolkitHistory(prompt_toolkit.history.History):
         hist = builtins.__xonsh_history__
         if hist is None:
             return
-        for cmd in hist.all_items():
+        for cmd in hist.all_items(reverse=True):
             line = cmd['inp'].rstrip()
             strs = self.get_strings()
             if len(strs) == 0 or line != strs[-1]:


melund:
Thanks @gforsyth .. This fix would need to be applied to the sqlite backend as well. 

But are we sure this is a bug in xonsh??

>Looking at the existing code, it seems like history entries should always be
loaded oldest first (which hasn't been happening, so whaaa?). 

This doesn't correspond with [PTK2 docs](https://python-prompt-toolkit.readthedocs.io/en/master/pages/reference.html#prompt_toolkit.history.History.load_history_strings): 
> **load_history_strings()**
> This should be a generator that yields str instances.
>
>It should yield the most recent items first, because they are the most important. (The history can already be used, even when it’s only partially loaded.)

Wouldn't be more resonable that this is a bug in PTK2 @jonathanslenders

melund:
@gforsyth Maybe we should also inherit from the `prompt_toolkit.history.ThreadedHistory` class instead.

jonathanslenders:
This is interesting. Right now, prompt_toolkit assumed that the history loads in order, oldest first, because that's what makes sense if the history is written to a file. You can't read a file backwards.

However, if you have an SQLite database or something like that, it could indeed make sense to load the history in reverse order, I'll see whether there's a way to support that.

melund:
@jonathanslenders ... Ahh OK. But then the docs seem wrong I guess.

melund:
Or is it us that are wrong.. I am a bit confused here. @gforsyth your fix only reverses the order in which the json files are loaded, not the individual commands in the files. 

So it seems to me that we are serving history with oldest commands first in the JSON backend.

melund:
@jonathanslenders It seems to me the docs are correct. PTK 2 needs history newest first. And xonsh was serving the oldest commands first.

melund:
@gforsyth I have pushed updated fix ([`fix_ptk2_reverse_history_take2`](https://github.com/xonsh/xonsh/tree/fix_ptk2_reverse_history_take2)) which also reverses the actual commands and works with the sqlite backend as well. 

https://github.com/xonsh/xonsh/compare/fix_ptk2_reverse_history...fix_ptk2_reverse_history_take2

It does seem like PTK2 needs entries newest first as the docs says. And we were serving them oldest first.

melund:
Actually, I can just push the fix to your branch. Hope it is OK with you. Better to keep all discussions here..

xonsh/history/json.py
```diff
@@ -379,20 +379,24 @@ def flush(self, at_exit=False):
...
-    def items(self):
+    def items(self, reverse=False):
         """Display history items of current session."""
-        for item, tss in zip(self.inps, self.tss):
+        if reverse:
+            items = zip(reversed(self.inps), reversed(self.tss))
+        else:
+            items = zip(self.inps, self.tss)
+        for item, tss in items:
             yield {'inp': item.rstrip(), 'ts': tss[0]}
```
melund: Can anyone see a more elegant way of doing this?

gforsyth:
> Hope it is OK with you. Better to keep all discussions here..

Hey @melund, thanks for helping out with this!  Definitely ok with me and I agree with your changes.  

I don't see a better way to reverse the iterator, but at least this is all being threaded out so it shouldn't impact startup time?

jonathanslenders:
@melund,
That's right. I was confused for a moment, but there's a good reason for it that it consumes the history latest item first. Imagine that the history is stored on the network in a database, then it's possible to stream/read the database from latest to oldest. Given that ``History.load_history_strings`` is a generator, the UI will already be usable while the history is loading, and the newest will be available right away.

melund:
Maybe `reverse=False/True` is a little non-saying. I would say `newest_first=False/True` would be better... But  it is probably more important to get the fix out.

melund:
Ping. @scopatz we better get this in the next hotfix release as well. It is pretty annoying that history doesn't work.

xonsh/history/json.py
melund: I wouldn't think that was possible since zip is an iterator and could in principle be infinite. It would really have to be `reversed(list(items))` which isn't nice.

scopatz: Oh, sorry yeah, from itertools you can use `izip()`, but I think you run into the same problem with reverse.

PR #2771, opened 2018-08-15
Fix ValueError when executing `vox list`

knowsuchagency:


scopatz:
Thanks for putting this in @knowsuchagency! Just waiting on CI!

Issue #2776, opened 2018-08-19
cache scripts by realpath or disable caching for scripts with zero mtime


t184256:
Sorry if that's too much of a corner case, but I have an annoying problem with caching my .xonshrc. I use nixos and home-manager to manage it, and it gets deployed as a symlink to a file in /nix/store, which has a zero mtime. This way when a symlink starts pointing to a new revision of file (again, with a zero mtime), a cache does not get invalidated.

I see two simple (though not particularly useful to anyone else) workarounds:

1. cache scripts by their 'realpath', after following all symlinks

2.  add a kludge for zero-mtime files

Any thoughts?

adqm:
Interesting!  When I wrote this, I hadn't imagined this scenario.  Using the `realpath` in [`run_script_with_cache`](https://github.com/xonsh/xonsh/blob/master/xonsh/codecache.py#L147) seems like it should result in the right behavior to me, in all cases.  I think that beats a hack around files with 0 mtime.

t184256:
Since neither you nor me sees a problem with the realpath approach, here's a trivial pull request for it.

PR #2784, opened 2018-08-21
cache scripts by realpath

t184256:
Cache scripts by realpath, resolving symlinks,
as per discussion in #2776.
I chose realpath(...) over abspath(realpath(...)),
as realpath does call abspath internally.


scopatz:
Thanks @t184256!

Issue #2762, opened 2018-08-09
Allow Up/Down Arrows To Match Anywhere in Line, not Just from the start


scopatz:
## Expected Behavior
From [this tweet](https://twitter.com/i8br/status/1026496707725549569), it would be nice to have the up and down arrows match from anywhere in the line.

## Current Behavior
Only matches from the start of the line

gforsyth:
Pasting this in here for reference:

```
i love how you can start typing something in fish and just hit the up/down arrow to search for that anywhere in the history is there a way to enable that in xonsh? all i've seen so far is searching in the history for stuff that starts the same way
```

gforsyth:
Here's a side-loaded version that seems to work.  The method we need to override is `Buffer._history_matches` in `prompt_toolkit` but we don't use `Buffer` directly.  

For now, you can stick that in your `xonshrc` and I think it does the right thing.

```
from functools import partial

@events.on_post_init
def _monkeypatchme():
    def _cust_history_matches(self, i):
        return (self.history_search_text is None or
                self.history_search_text in self._working_lines[i])

    _partial_cust_history_matches = partial(_cust_history_matches, 
				            __xonsh_shell__.shell.prompter.default_buffer)

    __xonsh_shell__.shell.prompter.default_buffer._history_matches = _partial_cust_history_matches

    del _cust_history_matches
```

X41:
Awesome! Thanks a lot, that makes it a lot more comfortable :D 
the only thing missing from that now is highlighting of the matched part

scopatz:
It would be great if we could get a PR for this :wink:

scopatz:
Closing due to merge of #2787

PR #2787, opened 2018-08-23
Add history matching toggle to search entire line

gforsyth:
Default behavior of prompt_toolkit on an up-arrow history search is to use a
`startswith` check to match previous entries.
This introduces an environment variable `XONSH_HISTORY_MATCH_ANYWHERE` that, if
`True`, will instead perform a history search with an `is in` to allow matching
anywhere.

This could potentially have performance impacts depending on how many history
items are being searched, so I've defaulted it to `False` but I'm open to input
on that (or any of the rest of it).

This addresses #2762


scopatz:
Thanks Gil!

PR #2550, opened 2017-12-04
Add pypy3 to travis testcases

traverseda:

PR #2792, opened 2018-08-26
Pypytravis

gforsyth:
Rebased @traverseda's work from #2550 against current master to take care of the changes to the `setup.cfg` file and tweaked some defaults -- seems to be "passing" now (with allowed failures for the pypy test)

No news item needed for this.  This can also supercede #2550


scopatz:
Amazing! Thanks @gforsyth!

Issue #2793, opened 2018-08-27
Ctrl-R not working


davidbrochart:
## xonfig

```
+------------------+----------------------+
| xonsh            | 0.7.6                |
| Git SHA          | 3939a31b             |
| Commit Date      | Aug 21 16:03:45 2018 |
| Python           | 3.6.3                |
| PLY              | 3.9                  |
| have readline    | True                 |
| prompt toolkit   | 2.0.4                |
| shell type       | prompt_toolkit2      |
| pygments         | None                 |
| on posix         | True                 |
| on linux         | True                 |
| distro           | debian               |
| on darwin        | False                |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+
```
## Expected Behavior

With `Ctrl-R`, one can search a pattern backward and select it with `Enter`.
## Current Behavior

One cannot select the pattern with `Enter`, it remains in search mode.
## Steps to Reproduce

Type e.g. `ls` and `Enter`. Then type `Ctrl-R` and `l`, it finds the previous command `ls`, but typing `Enter` has no effect.

scopatz:
Thanks for reporting @davidbrochart! I can confirm that this behaviour is broken

scopatz:
This seems like it might be an issue with prompt-toolkit, rather than xonsh.  Can you please open a related issue at https://github.com/jonathanslenders/python-prompt-toolkit/ Thanks!

PR #2796, opened 2018-08-27
Fix ctrlr

gforsyth:
This may be an issue upstream in PTK or it might be because we override
all sorts of keybindings to allow for a friendlier multiline experience.

In any case, registering an explicit keybinding to break out of the
search mode restores the functionality.

Fix for #2793

### DIFF
--- a/xonsh/ptk2/key_bindings.py
+++ b/xonsh/ptk2/key_bindings.py
@@ -2,9 +2,10 @@
 """Key bindings for prompt_toolkit xonsh shell."""
 import builtins
 
+from prompt_toolkit import search
 from prompt_toolkit.enums import DEFAULT_BUFFER
 from prompt_toolkit.filters import (Condition, IsMultiline, HasSelection,
-                                    EmacsInsertMode, ViInsertMode)
+                                    EmacsInsertMode, ViInsertMode, IsSearching)
 from prompt_toolkit.keys import Keys
 from prompt_toolkit.application.current import get_app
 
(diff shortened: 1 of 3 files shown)


scopatz:
Thanks @gforsyth! 6e876118a5def8962776d5af5226974644fbc4ca fixes the pypy3 build to allow failures again, if you want to cherry-pick that in.

scopatz:
This works for me!

xonsh/ptk2/key_bindings.py
```diff
@@ -339,6 +340,11 @@ def wrap_cursor_forward(event):
         b.cursor_left(count=abs(relative_begin_index))
         b.cursor_down(count=1)
 
+    @handle(Keys.ControlM, filter=IsSearching())
+    @handle(Keys.ControlJ, filter=IsSearching())
```
scopatz: Not sure how this translates to ControlR, but I won't question it :)

gforsyth:
> Thanks @gforsyth! 6e87611 fixes the pypy3 build to allow failures again, if you want to cherry-pick that in.

Ah cool, I'll do that

xonsh/ptk2/key_bindings.py
gforsyth: ahh, but ControlR was working fine, it was that pesky Enter/Return key that was misbehaving

scopatz: Oh.....I see. Uggh

PR #2800, opened 2018-08-30
Fix multi-item path completion

### LINKED ISSUE #2791: Configure xonsh to directly complete instead of displaying a menu if there's a single completion entry.
Is there a way to configure xonsh so that it directly completes the word instead of displaying a menu if there's a single completion entry? It's very difficult to browse directory tree without it.

## xonfig

    | xonsh            | 0.7.6.dev5           |
    | Git SHA          | 837c714a             |
    | Commit Date      | Aug 25 18:35:16 2018 |
    | Python           | 3.7.0                |
    | PLY              | 3.9                  |
    | prompt toolkit   | 1.0.15               |
    | shell type       | prompt_toolkit1      |
    | pygments         | 2.2.0                |

gforsyth:
The keybinding logic for multiple `TAB`s brought up the selection menu if the event was repeated, hence why tabbing into a deep folder tree would bring up the selection menu every other time.

We had originally forked the code there from prompt_toolkit -- simply removing the keybinding seems to have fixed it (change only implemented for ptk2).

Fixes #2791

### DIFF
--- a/news/tab_deep.rst
+++ b/news/tab_deep.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* PTK tab-completion now auto-accepts completion if only one option is present
+  (note that fix is only for PTK2)
+
+**Security:** None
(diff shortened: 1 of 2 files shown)


scopatz:
LGTM! Thanks @gforsyth! I have seen every Windows build succeed.

PR #2804, opened 2018-09-05
Remove extra cleans and resets from doc builds

gforsyth:
Removing some of the lines that @scopatz added to the `travis.yml` file to get the build to work right -- should work now without those now that installing `xonsh` doesn't mess with repo state


scopatz:
Spurious AppVeyor failures.  Thanks for this @gforsyth!

PR #2809, opened 2018-09-11
add chain db

CJ-Wright:


scopatz:
Very cool! Thanks @CJ-Wright!

CJ-Wright:
Thank you!

melund:
@CJ-Wright Just out of curiosity, what is the use of this compared to the default chainmap in Python?

CJ-Wright:
I'd you chain map two dicts of lists where the keys are the same one list takes precedence. With chain DB the lists are combined. This allows one to form a map which is the super set of maps for both keys and values. The current use cases in production are making a single collection from multiple databases (each with the same top level keys and potentially lower level keys, but different permission levels) in regolith and for combining conda meta.yaml files rendered for different platforms creating the super set of dependencies which is needed to run jobs on packages in topological order.

PR #2811, opened 2018-09-12
skip foreign

### LINKED ISSUE #2810: How to disable 'skipping application of alias' message and let xonsh aliases take precedence
## xonfig

```
| xonsh            | 0.7.7           |
| Python           | 3.6.5           |
| distro           | Ubuntu          |
```

## Expected Behavior

I have just started using xonsh and would like to make it my main shell, but still be able to use bash aliases in both places. Ideally, xonsh should ignore any bash aliases that have the same name. If an alias has been defined in xonsh, I would like it to take precedence over any foreign aliases.

I don't want to see this error message every time I open the shell, and the `--overwrite-alias` option seems to do the opposite of what I want. Perhaps the ideal solution is to only print this message when `XONSH_DEBUG` is set.

## Current Behavior

I get this message every time I open xonsh:
```
Skipping application of 'ls' alias from 'bash' since it shares a name 
with an existing xonsh alias. Use "--overwrite-alias" option to 
apply it anyway.
```

## Steps to Reproduce

Open xonsh with bash sourced as a foreign shell, since bash usually has `alias ls='ls --color=auto'` defined by default.

scopatz:
Closes #2810

### DIFF
--- a/news/fssup.rst
+++ b/news/fssup.rst
@@ -1,0 +1,16 @@
+**Added:**
+
+* New environment variable ``$FOREIGN_ALIASES_SUPPRESS_SKIP_MESSAGE``
+  enables the removal of skipping foreign alias messages.
+* New ``--suppress-skip-message`` command line option for skipping
+  foreign alias messages when sourcing foreign shells.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -261,6 +261,13 @@ def _SOURCE_FOREIGN_PARSER():
         "replace the current xonsh aliases.",
     )
     parser.add_argument(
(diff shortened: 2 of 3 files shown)


gforsyth:
Looks good to me.  The Appveyor failure is a spurious timeout thing and isn't worth waiting on Appveyor to run again.

Time to get us some of those Azure pipelines running.

Thanks, @scopatz!

scopatz:
> Time to get us some of those Azure pipelines running.

Definitely!

comfortablynick:
Thank you for adding this option, @scopatz!

jeremyepling:
@scopatz / @gforsyth , I'm a product manager on Azure Pipelines. Let me know if you have any questions or suggestions when you set it up.

scopatz:
Thanks @jeremyepling! Will do.

scopatz:
When trying to register the xonsh org, I got:

> VS850012: The specified name is reserved: xonsh

Any ideas?

PR #2818, opened 2018-09-17
fixes empty source

### LINKED ISSUE #2814: if _get_bash_completions_source() is None, 'set()' gets into bash completion script
Once again, this is a borderline non-issue.

I'm running NixOS, which does not follow FHS and _get_bash_completions_source() returns `None` there.

While debugging the resulting lack of bash completion, I stumbled upon

https://github.com/xonsh/xonsh/blob/743a8513899cf5a27f0b3aae52e7fec16879f530/xonsh/completers/bash_completion.py#L313

which gets interpolated here

https://github.com/xonsh/xonsh/blob/743a8513899cf5a27f0b3aae52e7fec16879f530/xonsh/completers/bash_completion.py#L313

and sure enough, the bash script now starts with `set()`.

I'm sure that was not @scopatz's original intention.

scopatz:
closes #2814

### DIFF
--- a/news/emptysrc.rst
+++ b/news/emptysrc.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
+* In Bash completions, if there are no files to source, a ``set()`` will
+  no longer be inserted into the completion script.
+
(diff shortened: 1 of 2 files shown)


gforsyth:
Thanks, @scopatz !  Spurious failures on Appveyor

PR #2821, opened 2018-09-17
@(expr) in subprocess tokens

### LINKED ISSUE #2640: Python expression evaluation using @() not allowed as subexpression
I am trying to do something like this:

```
def upload_keys(user, server):
    scp /home/@(user)/.ssh/* @(user)@@(server):.ssh/.
```
Unfortunately it seems that the @() operator needs to be surrounded with spaces. 
```
def upload_keys(user, server):
    scp /home/ @(user) /.ssh/* @(user) @ @(server) :.ssh/.
```
This however doesn't work, because now the scp commands sees separate arguments.

A work around could be this
```
def upload_keys(user, server):
    $USER=user
    $SERVER=server
    scp /home/$USER/.ssh/* $USER@$SERVER:.ssh/.
```
But this pollutes the environment and results in a lot of copying.

Any clean solution to work with Python expressions as part of shell commands?

scopatz:
This closes #2640 

This was fun to implement.

### DIFF
--- a/xonsh/parsers/base.py
+++ b/xonsh/parsers/base.py
@@ -3008,7 +3008,9 @@ class BaseParser(object):
         p[0] = p0
 
     def p_subproc_atom_pyeval(self, p):
-        """subproc_atom : at_lparen_tok testlist_comp RPAREN"""
+        """subproc_atom : at_lparen_tok testlist_comp RPAREN
+           subproc_arg_part : at_lparen_tok testlist_comp RPAREN
+        """
         p1 = p[1]
         p0 = xonsh_call(
             "__xonsh_list_of_strs_or_callables__",
--- a/xonsh/parsers/base.py
+++ b/xonsh/parsers/base.py
@@ -3123,7 +3123,9 @@ class BaseParser(object):
             # has an expanding function call, such as @(x)
             p0 = xonsh_call(
                 "__xonsh_list_of_list_of_strs_outer_product__",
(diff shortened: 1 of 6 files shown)


xonsh/parsers/base.py
```diff
@@ -3085,7 +3149,19 @@ def p_subproc_arg_single(self, p):
     def p_subproc_arg_many(self, p):
         """subproc_arg : subproc_arg subproc_arg_part"""
         # This glues the string together after parsing
-        p[0] = p[1] + p[2]
+        p1 = p[1]
+        p2 = p[2]
+        if isinstance(p1, ast.Str) and isinstance(p2, ast.Str):
+            p0 = ast.Str(p1.s + p2.s, lineno=p1.lineno, col_offset=p1.col_offset)
+        elif isinstance(p1, list):
```
gforsyth: Is there a case here where both `p1` and `p2` are `list`s?  Does that need to be cased?

melund:
This is just crazy :)   I can't wait to use the outer-product feature though.

xonsh/parsers/base.py
scopatz: I think that isn't possible given how the parser reads tokens from left to right, but it is good to be robust here.  Fix incoming.

scopatz: Fix pushed up!

melund:
I won't say I understand the PLY stuff. But I believe @gforsyth looked it that part through.

PR #2824, opened 2018-09-18
Allow 'off' as falsy value

MrQubo:
Just a simple one.
I think some users might want to use 'off' as a falsy value (e.g. me).


MrQubo:
Wrong branch. Created new PR #2825.

PR #2829, opened 2018-09-20
Perform environment variable substitution in f-strings

### LINKED ISSUE #2822: Environment variables inside f-strings
It's hard to describe what the exact bug is. See the _Additional info_ section below.

## xonfig
```
| xonsh            | 0.7.8           |
| shell type       | prompt_toolkit1 |
```

## Expected Behavior
It's a correct code, so it shouldn't cause the SyntaxError.

## Current Behavior
It causes the following SyntaxError (with `XONSH_DEBUG=1`):
```
  File "<string>", line None
SyntaxError: /home/arch/.local/lib/python3.7/site-packages/xontrib/prompt_vi_mode.xsh:1:21: ('code: ]',)
[f'{$HOME}', $PWD, '']
                     ^
```

## Steps to Reproduce
Just type `[f'{$HOME}', $PWD, '']` in the prompt.

### General problem with environment variables inside f-strings
Typing `f'{$HOME}'` inside the prompt doesn't trigger the SyntaxError, but the result isn't printed under the prompt as usual. `print(f'{$HOME}')` doesn't output anything as well.
e.g.
```
   f'{$HOME}s'                       
   print(f'{$HOME}s')      
                               
```

If I try to save the result inside the variable it seems to have no effect at all.
e.g.
```
   y = f'{$HOME}'                    
   y                                                
xonsh: subprocess mode: command not found: y
    z:  Alias
    X:  Command (/usr/bin/X)
    w:  Command (/usr/bin/w)
    [:  Command (/usr/bin/[)
    R:  Command (/usr/bin/R)
   print(y)                                    
NameError: name 'y' is not defined
```
```
   y = 42                 
   y = f'{$HOME}'        
   print(y)            
42     
```

Instead of `f'{$HOME}'` one of the following can be used and the behavior will be the same. They don't trigger the SyntaxError as well.
```
[f'{$HOME}', '']

[f'{$HOME}', $PWD]
```

### Examples of working code
Following are the examples of code that __does__ work as expected:
```
x=2; [f'{x}', $PWD, '']

['{$HOME}', $PWD, '']
```

The first one uses f-string and evaluates normal python code inside it.
The second one doesn't involve f-strings at all.

### Another issue with f-strings
It seems there are problems with f-strings in general too. While using them I've run into the following error several times. I've found out it is triggered every time by typing `f""`, without pressing <kbd>Enter</kbd>.
```
   f""Exception in thread Thread-2566:    
  File "/usr/lib/python3.7/threading.py", line 917, in _bootstrap_inner                                                                                               
    self.run()                                                                                                                                                        
  File "/usr/lib/python3.7/threading.py", line 865, in run                                                                                                             
  File "/usr/lib/python3.7/site-packages/prompt_toolkit/interface.py", line 865, in run                                                                                
  File "/usr/lib/python3.7/site-packages/xonsh/ptk/completer.py", line 45, in get_completions
    prefix, line_ex, begidx + expand_offset, endidx + expand_offset, self.ctx
  File "/usr/lib/python3.7/site-packages/xonsh/__amalgam__.py", line 125, in complete
  File "/usr/lib/python3.7/site-packages/xonsh/completers/__amalgam__.py", line 1274, in complete_path
    p = _path_from_partial_string(line, end)
  File "/usr/lib/python3.7/site-packages/xonsh/completers/__amalgam__.py", line 1064, in _path_from_partial_string
    val = ast.literal_eval(_string)
  File "/usr/lib/python3.7/ast.py", line 91, in literal_eval
    return _convert(node_or_string)
  File "/usr/lib/python3.7/ast.py", line 90, in _convert
    return _convert_signed_num(node)
  File "/usr/lib/python3.7/ast.py", line 63, in _convert_signed_num
    return _convert_num(node)
  File "/usr/lib/python3.7/ast.py", line 55, in _convert_num
    raise ValueError('malformed node or string: ' + repr(node))
ValueError: malformed node or string: <_ast.JoinedStr object at 0x7f82d1549d68>
```

scopatz:
Closes #2822

### DIFF
--- a/xonsh/parsers/base.py
+++ b/xonsh/parsers/base.py
@@ -26,6 +26,7 @@ RE_SEARCHPATH = LazyObject(lambda: re.compile(SearchPath), globals(), "RE_SEARCH
 RE_STRINGPREFIX = LazyObject(
     lambda: re.compile(StringPrefix), globals(), "RE_STRINGPREFIX"
 )
+RE_ENVVAR = LazyObject(lambda: re.compile("\$(\w+)"), globals(), "RE_ENVVAR")
 
 
 class Location(object):
@@ -206,6 +207,26 @@ def hasglobstar(x):
             return False
     else:
         return False
+
+
+def _repl_sub_env_vars_single(matchobj):
+    return "__xonsh_env__.detype()['" + matchobj.group(1) + "']"
+
+
+def _repl_sub_env_vars_double(matchobj):
+    return '__xonsh_env__.detype()["' + matchobj.group(1) + '"]'
+
+
+def sub_env_vars(fstring):
+    """Takes an fstring that may contain environment variables and
+    substitues them for a valid environment lookup call. Roughly,
+    for example, this will take f"{$HOME}" and transform it to
+    be f"{__xonsh_env__.detype()['HOME']}".
+    """
+    repl = (
+        _repl_sub_env_vars_single if fstring[-1] == '"' else _repl_sub_env_vars_double
+    )
+    return RE_ENVVAR.sub(repl, fstring)
 
 
 class YaccLoader(Thread):
@@ -2382,7 +2403,8 @@ class BaseParser(object):
                 "__xonsh_path_literal__", [s], lineno=p1.lineno, col=p1.lexpos
             )
         elif "f" in prefix or "F" in prefix:
-            s = pyparse(p1.value).body[0].value
+            s = sub_env_vars(p1.value)
+            s = pyparse(s).body[0].value
             s = ast.increment_lineno(s, p1.lineno - 1)
             p[0] = s
         else:
--- a/news/fenv.rst
+++ b/news/fenv.rst
@@ -1,0 +1,16 @@
+**Added:**
+
+* Format strings (f-strings) now allow environment variables to be looked up.
+  For example, ``f"{$HOME}"`` will yield ``"/home/user"``. Note that this will
+  look up and fill in the ``detype()``-ed version of the environment variable,
+  i.e. it's native string representation.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -122,6 +122,13 @@ def test_f_literal():
     check_ast('F"{yo}"', run=False)
 
 
+@skip_if_lt_py36
+def test_f_env_var():
+    check_xonsh_ast({}, 'f"{$HOME}"', run=False)
+    check_xonsh_ast({}, "f'{$XONSH_DEBUG}'", run=False)
+    check_xonsh_ast({}, 'F"{$PATH} and {$XONSH_DEBUG}"', run=False)
+
+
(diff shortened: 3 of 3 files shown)


MrQubo:
What about
```
f'$HOME'
```
or
```
f"{'$HOME'}"
```

It's more complicated. I think we need to parse the content of an f-string and run the xonsh parser on the content of `{...}`. It would also support something like
```
f'{$(find -name fancy | dirname)}/morefancy'
```
etc.

gforsyth:
Ok, so this is awesome and when we add in the `pf` strings it will be incredible.

Works on my end, but I found a couple of corner cases:

```
gil@bad_cat ~ $ f'$HOME'
'__xonsh_env__.detype()["HOME"]'
```

and the string `f"{'$HOME'}"` doesn't generate any output, just a new prompt.

wait... 

```
gil@bad_cat ~ $ print(f"{'$HOME'}")
gil@bad_cat ~ $ _
'$HOME'
```

Ahh, that's a misleading output, the command is failing to register completely and the `_` is pointing to an earlier command

```
gil@bad_cat ~ $ which ls
ls --color=auto -v
gil@bad_cat ~ $ a = ![f"{'$HOME'}"]
gil@bad_cat ~ $ a.output
'ls --color=auto -v\n'
gil@bad_cat ~ $ 
```

MrQubo:
Yeah, I meant that my first two examples don't work with current code as they should.

I think that we could convert f-strings to regular strings concatenated with commands. We could take every `...{something_here}...` inside an f-string and convert it to `... + (something_here) + ...`.

Some examples:
```xonsh
f'{$HOME}' == '' + ($HOME) + ''
f'aa{$HOME}bb' == 'aa' + ($HOME) + 'bb'
f"aa{$X}bb{$A + '$' + $B}cc{'$Y'}dd" == 'aa' + ($X) + 'bb' + ($A + '$' + $B) + 'cc' + ('$Y') + 'dd'
f'{$(find -name fancy | dirname)}/morefancy' == '' + ($(find -name fancy | dirname)) + '/morefancy'
```

Not sure if I described clearly what I have in mind.

scopatz:
Thanks for noticing the corner cases @gforsyth! I'll fix those.

scopatz:
Ok! I have fixed the corner cases, I believe

xonsh/parsers/base.py
```diff
@@ -26,6 +26,7 @@
 RE_STRINGPREFIX = LazyObject(
     lambda: re.compile(StringPrefix), globals(), "RE_STRINGPREFIX"
 )
+RE_FSTR_ENVVAR = LazyObject(lambda: re.compile("\{\s*\$(\w+)"), globals(), "RE_FSTR_ENVVAR")
```
gforsyth: `black` wants this line to be shorter (sorry, i need to add that check to CI)

scopatz: done! Thanks

scopatz: I meant to run black, but forgot :(

scopatz:
@MrQubo - yeah you are right that those cases should be handled too.  Please open another issue for those. I think you are right that we will start having to parse f-strings ourselves.

PR #2830, opened 2018-09-21
fix path completer issue with f-strings

laloch:
Fix path completer crash on attempted f-string completion. Fixes the second part of #2822 .
`ast.literal_eval()` can't handle f-strings and raises `ValueError` for them. The patch just silences the exception and gives up on the completion.

### DIFF
--- a/news/path_completer_exception.rst
+++ b/news/path_completer_exception.rst
@@ -1,0 +1,13 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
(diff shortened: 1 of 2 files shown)


MrQubo:
Shouldn't we just check if a string isn't an f-string? The ValueError might be caused by some other things.
It might be slower than the catch though.

laloch:
@MrQubo We would have to check for Python version just for the sake of being defensive and, TBH, I couldn't care less about the real reason for the ValueError. I just want whatever exception can come from the eval to be handled.

PR #2831, opened 2018-09-21
fix wizard crash when an unknown foreign shell is given

### LINKED ISSUE #2823: Xonsh wizard crashed
I installed xonsh onto macos using homebrew, started it, and ran the wizard. I told it I wanted to import fish as a foreign shell and accepted all the defaults for every other question.

My wizard experience ended like this:

```
Would you like to enable xontribs now, yes or no [default: no]?

'`-.,_,.-*'`-.,_,.-*'`-.,_,.-*'`-.,_,.-*'`-.,_,.-*'`-.,_,.-*'`-.,_,.-*'


TypeError: sequence item 0: expected str instance, NoneType found
```

scopatz:
We weren't using defaultdict correctly. Closes #2823

### DIFF
--- a/news/fswiz.rst
+++ b/news/fswiz.rst
@@ -1,0 +1,14 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+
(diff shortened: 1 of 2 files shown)


gforsyth:
Thanks, @scopatz!

PR #2832, opened 2018-09-21
fix tab completion of fstrings

scopatz:
closes tab issue of #2822


laloch:
It's appreciated for sure, but it's exactly the same patch as in my PR #2830, isn't it?

scopatz:
Yes! Whoops! Don't know how I missed that.

laloch:
No problem. I don't really care about which one gets merged :)

PR #2839, opened 2018-09-23
Add black check to CircleCI

gforsyth:
There's one line in `xonsh/tools.py` that snuck through our manual reviews, which is actually great.  This should fail when we run `black` now and it finds changes to make, then I can push up the fix to confirm it's working as expected.


gforsyth:
Sweet, it worked!  No news item required for this one, I don't think.

scopatz:
Thanks @gforsyth! Not going to wait for CI here

PR #2842, opened 2018-09-25
Add DragonFly BSD support

strangelittlemonkey:

### DIFF
--- a/news/dragonfly_support.rst
+++ b/news/dragonfly_support.rst
@@ -1,0 +1,13 @@
+**Added:**
+
+- DragonFly BSD support
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -103,7 +103,7 @@ class Aliases(cabc.MutableMapping):
         if word in builtins.aliases and isinstance(self.get(word), cabc.Sequence):
             word_idx = line.find(word)
             expansion = " ".join(self.get(word))
-            line = line[:word_idx] + expansion + line[word_idx + len(word):]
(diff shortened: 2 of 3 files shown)


news/dragonfly_support.rst
```diff
@@ -0,0 +1,11 @@
+**Added:** DragonFly BSD support
```
gforsyth: The note needs to be on it's own line, so something like

```
**Added:**

- DragonFly BSD support
```

xonsh/aliases.py
```diff
@@ -13,7 +13,7 @@
 from xonsh.environ import locate_binary, make_args_env
 from xonsh.foreign_shells import foreign_shell_data
 from xonsh.jobs import jobs, fg, bg, clean_jobs
-from xonsh.platform import ON_ANACONDA, ON_DARWIN, ON_WINDOWS, ON_FREEBSD, ON_NETBSD
+from xonsh.platform import ON_ANACONDA, ON_DARWIN, ON_WINDOWS, ON_FREEBSD, ON_NETBSD, ON_DRAGONFLY
```
gforsyth: Ahh, and I think `black` is unhappy about the length of this line -- can you run `black` on `aliases.py` to fix it?  Thanks!

scopatz:
Thanks for putting this in @strangelittlemonkey!

xonsh/aliases.py
strangelittlemonkey: Eh, so for my OS py36-black pulled in like 4GB of deps and then it was broken and I had to install something else to fix it, got it up, but it seems that the lint also changed more than just that line. If you'd prefer I can take those changes back out, but I left them for the moment. Your call.

scopatz: Sorry about the black issue, that seems uncharacteristically large. I'll fix the remaining black issue

scopatz:
Thanks @strangelittlemonkey!

Issue #2584, opened 2018-02-02
NameError: name 'cmd2' is not defined for `cmd1; cmd2 && cmd3`, but `cmd1; cmd2 arg && cmd3` works


MrMetric:
These commands:
```
$ clear; make && ./program
$ clear && make && ./program
```
result in this error: `NameError: name 'make' is not defined`. These commands work, but none of them do what I want:
```
$ clear; make
$ clear; make; ./program
$ clear && make
$ clear && make; ./program
```
Note: a workaround is `clear; ![make] && ./program`.

I found that this has nothing to do with clearing the screen, running `make`, or running my program. These do not work:
```
$ echo; echo && echo
$ echo; echo && echo a
$ echo a; echo && echo
$ echo a; echo && echo a
```
but these do:
```
$ echo; echo a && echo
$ echo; echo a && echo a
$ echo a; echo a && echo
$ echo a; echo a && echo a
```

`xonfig` output:
```
+------------------+----------------------+
| xonsh            | 0.6.0.dev41          |
| Git SHA          | fa61b69a             |
| Commit Date      | Jan 13 13:37:39 2018 |
| Python           | 3.6.4                |
| PLY              | 3.9                  |
| have readline    | True                 |
| prompt toolkit   | 1.0.15               |
| shell type       | prompt_toolkit       |
| pygments         | 2.2.0                |
| on posix         | True                 |
| on linux         | True                 |
| distro           | arch                 |
| on darwin        | False                |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+```

astronouth7303:
Sounds a lot like #2430

gforsyth:
Closed in #2845

Issue #2430, opened 2017-07-12
Weird error about variable not being defined when using &&


nh2:
Does anybody know what's going on here?

Note, if you can't reproduce, I suspect it's something in my tree, but I haven't found anything yet so I'm first asking if it happens for you, too.

```
$ rm -f a && make && ./c
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK = True
NameError: name 'make' is not defined
```

```
$ rm -f a && make && ./c
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/home/niklas/python35env/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 15472, in default
    run_compiled_code(code, self.ctx, None, 'single')
  File "/home/niklas/python35env/lib/python3.5/site-packages/xonsh/__amalgam__.py", line 3634, in run_compiled_code
    func(code, glb, loc)
  File "/home/niklas/src/xontrib-fzf-widgets/xontrib/fzf-widgets.xsh", line 1, in <module>
    import os
NameError: name 'make' is not defined
```

```
$ xonfig
+------------------+----------------------+
| xonsh            | 0.5.12.dev4          |
| Git SHA          | 06c57eec             |
| Commit Date      | Jun 30 13:56:45 2017 |
| Python           | 3.5.2                |
| PLY              | 3.9                  |
| have readline    | True                 |
| prompt toolkit   | 1.0.14               |
| shell type       | prompt_toolkit       |
| pygments         | 2.2.0                |
| on posix         | True                 |
| on linux         | True                 |
| distro           | Ubuntu               |
| on darwin        | False                |
| on windows       | False                |
| on cygwin        | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+
```

astronouth7303:
Yup, I can reproduce on v0.5.12 with:

```
true && false && true
```

@scopatz Sounds like a parser bug?

scopatz:
Thanks for reporting.  @astronouth7303  it isn't a parser bug because the code is being compiled and executed. It is simply failing to autowrap the second command for some reason

gforsyth:
Closed in #2845

PR #2845, opened 2018-09-29
Fix issues #2584 and #2430

jmcrawford45:
Fix issues #2584 and #2430

### DIFF
--- a/news/subproc-whitespace-fix.rst
+++ b/news/subproc-whitespace-fix.rst
@@ -6,6 +6,8 @@
 
 **Removed:** None
 
-**Fixed:** try_subproc_toks now works for subprocs with trailing and leading whitespace
+**Fixed:**
+
+* try_subproc_toks now works for subprocs with trailing and leading whitespace
 
 **Security:** None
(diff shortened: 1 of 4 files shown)


jmcrawford45:
Any advice on fixing the build_black failure? From the test output, it's not clear what the issue is.

jmcrawford45:
@scopatz after the black format changes, the AppVeyor build is failing on Python36. I don't think the format changes should be affecting the build. Perhaps this was an intermittent failure? Help would be appreciated.

gforsyth:
Hey @jmcrawford45 -- thanks for putting this in!  The Appveyor failure is a known intermittent issue and not related to this PR as you suspect.

It also looks like `black` was updated to handle [Numeric literals](https://github.com/ambv/black#numeric-literals) since we used it to reformat everything.  I'll open an issue to pin the version of `black ` we're using so this doesn't just bite whomever happens to open a PR after there's a new version of `black` released that changes something.

Thanks again for this!

scopatz:
Thanks again @jmcrawford45!

Issue #2849, opened 2018-10-01
Add to "projects that use Xonsh" in README and docs


gforsyth:
All we have listed right now is `gitsome` but there's also 
* rever: https://regro.github.io/rever-docs/
* the regro autotick bot: https://github.com/regro/cf-scripts

and maybe more?  who knows?

scopatz:
Much agreed!

gforsyth:
Closed via #2850 but if there are more, please add them!

PR #2850, opened 2018-10-01
update projects

dbxnr:
#2849 Add to "projects that use Xonsh" in README and docs


gforsyth:
Awesome! Thanks @dbxnr!

PR #2852, opened 2018-10-02
don't capture cryptop output

### LINKED ISSUE #2517: Ncurses based python app does not run correctly on Xonsh, runs fine on Bash
```
pip install cryptop

...

$ cryptop
```
The app's UI is shown, but pressing any key (A to add a coin or Q to quit) does not work, instead the letter is written to the screen.

Runs fine from Bash.

scopatz:
closes #2517


gforsyth:
Thanks, @scopatz !

PR #2853, opened 2018-10-02
Set up CI with Azure Pipelines

gforsyth:
Wild.  They have built-in support for conda environments.  No idea if any of this will work and very much a work in progress, but let's see how this goes...

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,11 +1,6 @@
-# Python package
-# Create and test a Python package on multiple Python versions.
-# Add steps that analyze code, save the dist with the build record, publish to a PyPI-compatible index, and more:
-# https://docs.microsoft.com/azure/devops/pipelines/languages/python
-
 jobs:
 
-- job: 'Test'
+- job: 'xonsh_windows'
   pool:
     vmImage: 'VS2017-Win2016'
   strategy:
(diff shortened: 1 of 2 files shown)


gforsyth:
lol, not well

scopatz:
Amazing! Thanks for pushing on this @gforsyth!

scopatz:
@mariusvniekerk has pointed me to this, which he says is less broke than the built-in condas: https://github.com/mariusvniekerk/psutil-feedstock/blob/master/azure-pipelines.yaml

scopatz:
We don't have C extensions and probably don't need to build mac on this yet, but this may be a useful point of reference.

gforsyth:
Wow.  That was... actually pretty easy.  Cool!  

I have a few changes to make, but question: should we have this run in tandem with Appveyor for a bit and then disable Appveyor later?  Or disable it now?

gforsyth:
Also pinging @melund to see if he has any thoughts on windows testing stuff I may have overlooked

scopatz:
I think we can run at least the msys stuff on appveyor for the moment, since this doesn't test that on azure yet (which is fine).  Also, yeah, we should definitely defer to @melund to merge this :)

scopatz:
@gforsyth - woohoo! It works!

melund:
Looks good. We could think about exporting the test results directly to Azure, but I will make a separate PR for that.

PR #2855, opened 2018-10-02
Publish results from tests on Azure

melund:
This adds a publishing step to the azure pipeline. That may make it a little easier to see what tests fail, maybe there is a chance it can be reported right back to GitHub.

No news item for this is necessary.


gforsyth:
Thanks, @melund!

Issue #2840, opened 2018-09-23
Change auto_completer's color style


vaaaaanquish:
I want change auto_completer's complete color style using prompt-toolkit 2.x. Because, default white is not good for my eyes.
<img width="232" alt="2018-09-24 1 15 19" src="https://user-images.githubusercontent.com/4796081/45930169-64635b00-bf97-11e8-9a0b-5606aba152f2.png">

According to ptk document, It seems to give `style` and `selected_style` to prompt_toolkit.completion.Completion's args.
https://python-prompt-toolkit.readthedocs.io/en/master/pages/reference.html#prompt_toolkit.completion.Completion

I will make pull requests.

laloch:
This problem is very similar to #2838. Both the completion and the toolbar styles are defined by the PTK's default style and there seems to be no way to redefine the styles in xonsh. I think it would be better to provide a hook for overriding the PTK stylesheet globally.

laloch:
@vaaaaanquish I just did
```
import sys
defstyle = sys.modules['prompt_toolkit.styles.defaults']                   
defstyle.PROMPT_TOOLKIT_STYLE.append(('completion-menu','bg:#333333 #eeeeee'))
```
and it worked. So it's definitely possible.

melund:
We used to have a way to control those things: 
https://github.com/xonsh/xonsh/blob/master/xonsh/pyghooks.py#L515-L528

But it seems that PTK2 new way of styling doesn't respect those when they come through the pygments link. 

The problem is that everything in xonsh still relies on styles being `(Token, style_str)` whereas PTK2 now uses `(class_string, style_str)`.

@jonathanslenders is there anyway to control the PTK2 styles (e.g. `completion-menu`) when they are passed in through the pygments interface (`style_from_pygments_cls`, or `style_from_pygments_dict`).

vaaaaanquish:
Thank you. I also think that using PTK2's style is smart.
I will look at [pyghooks.py](https://github.com/xonsh/xonsh/blob/master/xonsh/pyghooks.py#L515-L528)

jonathanslenders:
@melund,

Unfortunately, it's not possible out of the box, because ``Token.X.Y`` translates into ``class:pygments.x.y`` instead of ``class:x.y``. So you can't use a Pygments Token to style non pygments stuff.

However, the pgcli developers created the following translation table:
```python

scopatz:
[...]

PR #2861, opened 2018-10-03
PTK2 style overrides

### LINKED ISSUE #2838: Overriding Prompt Toolkit style
PTK2 defaults to reverse the colors of the bottom toolbar, which visually breaks the Powerline extension:
screenshot_20180923_120221
In order to get rid of that annoyance, one needs to set `'bottom-toolbar': 'noreverse'` PTK style. My question is: What's the correct way to do it in xonsh? I got lost in the many style translation layers and the best I could come with is the below which does the trick, but is obviously wrong.
```
index 4a55664c..144d6f1b 100644
--- a/xonsh/ptk2/shell.py
+++ b/xonsh/ptk2/shell.py
@@ -144,6 +144,7 @@ class PromptToolkit2Shell(BaseShell):
             else:
                 style = style_from_pygments_dict(DEFAULT_STYLE_DICT)
 
+            style.style_rules.append(('bottom-toolbar', 'noreverse'))
             prompt_args["style"] = style
 
         line = self.prompter.prompt(**prompt_args)
```
screenshot_20180923_121133

laloch:
Introduces `$XONSH_STYLE_OVERRIDES` env-var, which is a Python dictionary of prompt_toolkit style definitions.
Closes #2838 and #2840.

### DIFF
--- a/news/style-overrides.rst
+++ b/news/style-overrides.rst
@@ -2,14 +2,14 @@
 
 * Added new env-var ``XONSH_STYLE_OVERRIDES``. The variable is
   a dictionary containing custom prompt_toolkit style definitions.
-  For instance
-  ```
+  For instance::
+
   $XONSH_STYLE_OVERRIDES['completion-menu'] = 'bg:#333333 #EEEEEE'
-  ```
-  will provide for more visually pleasing completion menu style whereas
-  ```
+
+  will provide for more visually pleasing completion menu style whereas::
+
   $XONSH_STYLE_OVERRIDES['bottom-toolbar'] = 'noreverse'
-  ```
+
   will prevent prompt_toolkit from inverting the bottom toolbar colors
   (useful for powerline extension users)
 
--- a/news/style-overrides.rst
+++ b/news/style-overrides.rst
@@ -1,14 +1,14 @@
 **Added:**
 
-* Added new env-var ``XONSH_STYLE_OVERRIDES``. The variable is
+* Added new env-var ``PTK_STYLE_OVERRIDES``. The variable is
   a dictionary containing custom prompt_toolkit style definitions.
   For instance::
 
-  $XONSH_STYLE_OVERRIDES['completion-menu'] = 'bg:#333333 #EEEEEE'
+    $PTK_STYLE_OVERRIDES['completion-menu'] = 'bg:#333333 #EEEEEE'
 
   will provide for more visually pleasing completion menu style whereas::
 
-  $XONSH_STYLE_OVERRIDES['bottom-toolbar'] = 'noreverse'
+    $PTK_STYLE_OVERRIDES['bottom-toolbar'] = 'noreverse'
 
   will prevent prompt_toolkit from inverting the bottom toolbar colors
   (useful for powerline extension users)
--- a/news/style-overrides.rst
+++ b/news/style-overrides.rst
@@ -1,0 +1,26 @@
+**Added:**
+
+* Added new env-var ``XONSH_STYLE_OVERRIDES``. The variable is
+  a dictionary containing custom prompt_toolkit style definitions.
+  For instance
+  ```
+  $XONSH_STYLE_OVERRIDES['completion-menu'] = 'bg:#333333 #EEEEEE'
+  ```
+  will provide for more visually pleasing completion menu style whereas
+  ```
+  $XONSH_STYLE_OVERRIDES['bottom-toolbar'] = 'noreverse'
+  ```
+  will prevent prompt_toolkit from inverting the bottom toolbar colors
+  (useful for powerline extension users)
+
+  Note: This only works with prompt_toolkit 2 prompter.
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
--- a/xonsh/ptk2/shell.py
+++ b/xonsh/ptk2/shell.py
@@ -152,7 +152,7 @@ class PromptToolkit2Shell(BaseShell):
                 try:
                     style_overrides = Style.from_dict(style_overrides_env)
                     prompt_args["style"] = merge_styles([style, style_overrides])
-                except Exception:  # pylint: disable=broad-except
+                except (AttributeError, TypeError, ValueError):
                     print_exception()
 
         line = self.prompter.prompt(**prompt_args)
--- a/xonsh/ptk2/shell.py
+++ b/xonsh/ptk2/shell.py
@@ -147,7 +147,7 @@ class PromptToolkit2Shell(BaseShell):
 
             prompt_args["style"] = style
 
-            style_overrides_env = env.get("XONSH_STYLE_OVERRIDES")
+            style_overrides_env = env.get("PTK_STYLE_OVERRIDES")
             if style_overrides_env:
                 try:
                     style_overrides = Style.from_dict(style_overrides_env)
--- a/xonsh/ptk2/shell.py
+++ b/xonsh/ptk2/shell.py
@@ -25,6 +25,7 @@ from prompt_toolkit.shortcuts import print_formatted_text as ptk_print
 from prompt_toolkit.shortcuts import CompleteStyle
 from prompt_toolkit.shortcuts.prompt import PromptSession
 from prompt_toolkit.formatted_text import PygmentsTokens
+from prompt_toolkit.styles import merge_styles, Style
 from prompt_toolkit.styles.pygments import (
     style_from_pygments_cls,
     style_from_pygments_dict,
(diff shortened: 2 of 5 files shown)


melund:
Well. Maybe this is the best simple hack that solves the immediate problem. But not very future proof. You should at least call the `env` something like`PTK2_STYLE_OVERRIDES` instead. To indicate that this is specific only to the PTK2 backend. 

Another idea for a solution: Maybe we could create a subclass of the(`__xonsh__.shell.shell.styler.styles`) chainmap so it automatically return the new `{class_str, style_str}`  on PTK2 and the old `{Token, style_str}' when using PTK1/readline. I.e. implement the functionality of [pygments_token_to_classname()](https://github.com/jonathanslenders/python-prompt-toolkit/blob/a4a010e4e4bf08755ed1efd587de34843cfe2f07/prompt_toolkit/styles/pygments.py#L53-L61) our selves. It could then be legal to put both `{class_str, style_str}` and `{Token, style_str}'  in the styles dict. 

@scapatz do any other good ideas spring to your mind.

laloch:
@melund You are absolutely right about this being just a simple hack. I just could not devise anything better. What I especially did not want to do was to create another bunch of artificial PTK1-like tokens just to do another round of pointless back-and-forth token to class translations, mimicking a feature that is already deprecated.

news/style-overrides.rst
```diff
@@ -0,0 +1,26 @@
+**Added:**
+
+* Added new env-var ``XONSH_STYLE_OVERRIDES``. The variable is
+  a dictionary containing custom prompt_toolkit style definitions.
+  For instance::
+
+  $XONSH_STYLE_OVERRIDES['completion-menu'] = 'bg:#333333 #EEEEEE'
```
scopatz: I think this line needs to be indented 4 spaces.

news/style-overrides.rst
```diff
@@ -0,0 +1,26 @@
...
+
+  will provide for more visually pleasing completion menu style whereas::
+
+  $XONSH_STYLE_OVERRIDES['bottom-toolbar'] = 'noreverse'
```
scopatz: As does this one.

scopatz:
I think this is good enough for now, with the exception that this should be called `$PTK_STYLE_OVERRIDES`. And yeah, don't worry about implementing this for PTK v1.

scopatz:

melund:
@laloch  Could you add the [old PTK style](https://github.com/xonsh/xonsh/blob/master/xonsh/pyghooks.py#L519-L532) (which is now igonored) as default values `$PTK_STYLE_OVERRIDES` environment var:

```
 {
 "completion-menu": "bg:ansigray ansiblack",
 "completion-menu.completion": "",
 "completion-menu.completion.current": "bg:ansibrightblack ansiwhite",
 "scrollbar.background": "bg:ansibrightblack",
 "scrollbar.arrow": "bg:ansiblack ansiwhite bold" ,
 "scrollbar.button": "bg:ansiblack",
 "auto-suggestion": "ansibrightblack",
 "aborting": "ansibrightblack",
 }
```

melund:
We originally overrode the PTK default values with our own color settings. Those are no longer used since they are given as pygments tokens. 

We had own colors because we specify colors with the ansi-color specifications. The ansi-colors allow users to change the color scheme directly from their terminal app. 

We could consider adding all PTK classes which are possibly used in xonsh to `$PTK_STYLE_OVERRIDES` but with ansi-colors instead. The env variable is more or less hack to allow us to specify PTK2 own style again which can't be modified with the pygments tokens. 

And even though it has a default value it can still be used for overriding stuff.

laloch:
I'm afraid I still don't see your point. The `PTK_STYLE` gets applied [here](https://github.com/xonsh/xonsh/blob/master/xonsh/pyghooks.py#L469) through the ChainMap. It is preceded and thus possibly overridden in the ChainMap by the selected xonsh style (the `cmap` argument). Now `$PTK_STYLE_OVERRIDES` contents are merged in at the very last step before creating the prompt. If it would contain token-to-class translation of `PTK_STYLE`, it would possibly nullify the effect of the xonsh style selection.

melund:
> I'm afraid I still don't see your point. The PTK_STYLE gets applied here through the ChainMap. It is preceded and thus possibly overridden in the ChainMap by the selected xonsh style (the cmap argument). 

Those styles have no effect on PTK2, since its style system changed. So the `$PTK_STYLE_OVERRIDES` would the only way to control PTK2 internal styles.

melund:
If the PTK style still worked there would be no need for $PTK_STYLE_OVERRIDES. In that case you could just have modified ` __xonsh__.shell.shell.styler.styles` directly.

laloch:
Now I see! Thanks for your patience @melund. OK, I can do it.

melund:
> Now I see! Thanks for your patience @melund. OK, I can do it.

No problem. You help is really appreciated :)

laloch:
@melund Is it OK now? Now it's getting really hackish. I already regret that I opened this one.

xonsh/ptk2/shell.py
```diff
@@ -146,6 +147,14 @@ def singleline(
 
             prompt_args["style"] = style
 
+            style_overrides_env = env.get("PTK_STYLE_OVERRIDES")
+            if style_overrides_env:
+                try:
```
melund: What is the reason for this broad try/catch?

laloch: 1) The PTK_STYLE_OVERRIDES can be pretty much anything (it's a user input).
2) I didn't really take the time to check what can be thrown from `Style.from_dict` and `merge_styles`.

melund: True, but I think it better to let it fail than to suppress all errors. 

If you want to verify the content of the `PTK_STYLE_OVERRIDES` you should add an entry to the [`DEFAULT_ENSURERS`](https://github.com/xonsh/xonsh/blob/c97fc91c284f4a2b7fd57c7d1508afe71373e45c/xonsh/environ.py#L164-L273) 

These functions are used to validate the content when a value is set: 
See for example the [__setitem__](https://github.com/xonsh/xonsh/blob/c97fc91c284f4a2b7fd57c7d1508afe71373e45c/xonsh/environ.py#L1157-L1174)

laloch: > True, but I think it better to let it fail than to suppress all errors.

The eventual exception get printed in the handler.

> If you want to verify the content of the PTK_STYLE_OVERRIDES you should add an entry to the DEFAULT_ENSURERS

I can do it, but I don't really think it makes much sense, because it's a dictionary of string keys with string values. Both are interpreted at some point somehow and both can contain possibly invalid values.

xonsh/style_tools.py
```diff
@@ -434,3 +434,18 @@ def norm_name(name):
     globals(),
     "DEFAULT_STYLE_DICT",
 )
+
+PTK2_STYLE = LazyObject(
```
scopatz: It doesn't make sense for this to be lazy if the this is just going to be copied first thing into the environment.

scopatz: as the default

laloch: Right. Too much copy-pasting ;)

xonsh/ptk2/shell.py
scopatz: Well, it makes sense if you want these settings to pass to a xonsh shell that is being run from another xonsh shell.

xonsh/style_tools.py
scopatz: No problem :)

scopatz:
I am not wedded to the ensurers issue, and it can always be added later

scopatz:
But it would be nice / consistent to have.

laloch:
All objections should be resolved. Please recheck.

xonsh/tools.py
```diff
@@ -36,6 +36,7 @@
 import traceback
 import warnings
 import operator
+from ast import literal_eval
```
scopatz: This needs to be `import ast` for amalgamate to work properly

xonsh/tools.py
```diff
@@ -1472,6 +1473,45 @@ def to_completions_display_value(x):
...
+        isinstance(k, str) and isinstance(v, str) for k, v in x.items()
+    )
+
+
+def to_dict(x):
+    """Converts a string to a dictionary"""
+    if isinstance(x, dict):
+        return x
+    try:
+        x = literal_eval(x)
```
scopatz: And then this needs to be `ast.literal_eval`

laloch: done already

scopatz:
Thanks for this @laloch! And thanks for dealing with the back-and-forth :wink:

scopatz:
Thanks for the great review @melund!

laloch:
@scopatz, @melund Thanks for the review and for your patience.

PR #2862, opened 2018-10-03
Fix deprecation warnings when running pytest

melund:
Fixes the deprecation warning reported by pytest. No news item needed I think.


scopatz:
Thanks @melund!

PR #2864, opened 2018-10-04
Fix string index

scopatz:
copied from @gforsyth in py-bash-completion


melund:
Looks good. Thanks @scopatz

Issue #2863, opened 2018-10-04
Spurious DeprecationWarning


laloch:
I'm currently on master at 4f9ab5d6c37aa82537cf06a249115c71575ae389.
While hunting for all the new sources of `DeprecationWarning` in the external xontribs I use, I stumbled upon very strange issue with [xonsh-vox-tabcomplete](https://github.com/Granitosaurus/xonsh-vox-tabcomplete). I have replaced the regular sources of the warning:
```
-__xonsh_completers__['vox'] = _vox_completer
-#bump to top of list (otherwise bash completion interferes)
-__xonsh_completers__.move_to_end('vox', last=False)
+completer add vox _vox_completer start
```
Now when I try to load the xontrib, I'm getting this very unusual `DeprecationWarning`:
```
$ xontrib load vox_tabcomplete
<unknown>:1: DeprecationWarning: invalid escape sequence \w
```
So far I failed to even find the place where the warning gets fired from, but if I modify the regular expression [here](https://github.com/laloch/xonsh-vox-tabcomplete/blob/master/xontrib/vox_tabcomplete.xsh#L17) and [here](https://github.com/laloch/xonsh-vox-tabcomplete/blob/master/xontrib/vox_tabcomplete.xsh#L20) to use "escaped" `\w` meta-character like `\\w`, the problem goes away and the extension works as expected.

melund:
I also get a lot of these now. But I don't have any contribs. There are few cases in xonsh, where regex strings were missing the `r` prefix. I have PR incoming to fix those. But they don't account for all the ones I see. 

I guess some are in the dependencies of xonsh. But it is hard to know where seeing they show up like this: 
```
<unknown>:1: DeprecationWarning: invalid escape sequence \|
```

scopatz:
With the merge of #2867, these have gone away for me.  Let's keep this open for a day or so to make sure.

Ayehavgunne:
I still had to do what the OP did to fix the warning.

scopatz:
what is the output of xonfig for you?

laloch:
Hi @Ayehavgunne, the xonsh-vox-tabcomplete xontrib seems to be unmaintained. You can install fixed version from my fork by running
```
xpip uninstall -y xonsh-vox-tabcomplete
xpip install git+https://github.com/laloch/xonsh-vox-tabcomplete.git@proxy-warnings#egg=xonsh-vox-tabcomplete
```

Ayehavgunne:
[...]

PR #2867, opened 2018-10-05
Escaped deprecation

melund:
This fixes some deprecation warnings from unallowed escape sequences as well as importing abstract base classes directly from `collections` . 

It may solve some of the problems reported in #2863. But probably not all, since some of the deprecation warnings may come from xonsh dependencies.


scopatz:
Thanks @melund!

PR #2868, opened 2018-10-06
created code for doing bash-like brace expansion.

ninjaaron:
Hi. The thing that bothers me the most in xonsh is the lack of Bash-like brace expansion. I have a lot of files with yy.mm.dd format dates in the title, and brace expansion is very useful for selecting certain date ranges and file types, moreso than regex. consider the following

```bash
$ touch 18.10.{01..30}
$ ls 18.10.{05..23}
18.10.05  18.10.08  18.10.11  18.10.14  18.10.17  18.10.20  18.10.23
18.10.06  18.10.09  18.10.12  18.10.15  18.10.18  18.10.21
18.10.07  18.10.10  18.10.13  18.10.16  18.10.19  18.10.22
```
As far as I am aware, doing something like this is non-trivial in xonsh.

This patch provides only a module with a function which can expand strings in this way. I have not made an attempt to merge this into current globbing system, which I believe is a task the maintainers will have a better handle on anyway.

Please let me know if this PR needs additional work before it is usable.


melund:
Interesting. We should definitely have this. 

But doing something like this is not that non-trivial in xonsh ( [even though a regression broke it on master](https://github.com/xonsh/xonsh/pull/2869) )

```
$ touch 18.10.@(f"{x:02d}" for x in range(1,31))
$ ls 18.10.@(f"{x:02d}" for x in range(5,24))
18.10.05  18.10.07  18.10.09  18.10.11  18.10.13  18.10.15  18.10.17  18.10.19  18.10.21  18.10.23
18.10.06  18.10.08  18.10.10  18.10.12  18.10.14  18.10.16  18.10.18  18.10.20  18.10.22
```
It is not as nice as the pure brace expansion, but maybe more powerfull. It can even do the outer product if there are more `@()` in one line. See the [tutorial](https://xon.sh/tutorial.html#python-evaluation-with)
```
$ echo @(['a', 'b']):@('x', 'y')
a:x a:y b:x b:y
```

ninjaaron:
Nice! The last time I tried something like this, it didn't work, but it's been a while since I've tried.

However, I would point out two important differences:

1. it still takes some extra fuffing with string formatting to add the extra zero padding.
2. The current mechanism cannot be combined with globbing.

My implementation does both, though it wouldn't surprise me if the specific way I've implemented the globbing needs to be adjusted (bog-standard glob.iglob; I noticed xonsh wraps it a bit, but I didn't look too deeply into it.)

gforsyth:
@adqm had a PR that tied in brace expansion -- I'm on mobile so can't really look it over but it might provide a good reference for where to tie things in

https://github.com/xonsh/xonsh/pull/2350

ninjaaron:
Thanks for the link, @gforsyth. I'll try to look at it this week, and I have some vacation the next week so _hopefully_ I can try to bang something out.

scopatz:
Thanks for pushing on this @ninjaaron! This is an awesome and much needed feature!

xonsh/brace_expansion.py
```diff
@@ -0,0 +1,93 @@
+import glob
+import itertools
+import re
+
+# regex to be used for splitting on braced regions
+BRACED = re.compile(r"(?<!\\)(\{.*?(?<!\\)\})")
```
scopatz: This should probably be a lazyobject, see some of the other examples in the code base.

scopatz:
@ninjaaron - can you add some unit tests for this too please? Thanks!

scopatz:
Also you might want to play around with replacing / augmenting the `__xonsh__.glob()` and `__xonsh__.globsearch()` functions with this capability if these match expressions make it through the parser

ninjaaron:
Thanks for the review, @scopatz. I'll write some tests for sure. That's the easy part. Getting everything plugged in to the existing xonsh machinery is the part that scares me!

scopatz:
Just let us know if you have any questions or run into any stumbling blocks

ninjaaron:
I've spent a little time on it this morning. I figured out how to integrate expansion into glob strings (the non-regex ones), but I have a couple questions regarding interface.

1. Glob strings don't return anything if there was no matching filename. In Bash, brace expansion can be used without globbing to create the products of arbitrary strings. Is this something we'd like to expose in xonsh, or is matching only existing filenames enough? (the addition of something like an e-string might be required to facilitate this, though I haven't yet found where syntax for these special strings are implemented. I'm afraid to look at the parser!)
2. At present, having braces in unquoted subprocess-mode arguments makes the parser angry. Once again, I fear the parser and don't even know where to begin looking to change this. I believe I do see where to add brace expansion for unquoted subprocess-mode arguments, but they won't get though the parser anyway. Thoughts?

scopatz:
Hi @ninjaaron - yeah, I think there is no avoiding the parser for some of this stuff. My guess is that the presence of certain tokens in [this set](https://github.com/xonsh/xonsh/blob/master/xonsh/parsers/base.py#L3206) is what is preventing `{` and `}` getting through.

anki-code:
This issue could be closed because it's outdated and we have two ways for solve it:

1. Bash-way:
```bash
bash -c! touch 18.10.{01..30}
ls
#18.10.01  18.10.03  18.10.05  18.10.07  18.10.09  18.10.11  18.10.13  18.10.15  18.10.17  18.10.19  18.10.21  18.10.23  18.10.25  18.10.27  18.10.29
#18.10.02  18.10.04  18.10.06  18.10.08  18.10.10  18.10.12  18.10.14  18.10.16  18.10.18  18.10.20  18.10.22  18.10.24  18.10.26  18.10.28  18.10.30

```
2. Xonsh-way:
```bash
touch 18.10.@(f"{x:02d}" for x in range(1,31))
ls
#18.10.01  18.10.03  18.10.05  18.10.07  18.10.09  18.10.11  18.10.13  18.10.15  18.10.17  18.10.19  18.10.21  18.10.23  18.10.25  18.10.27  18.10.29
#18.10.02  18.10.04  18.10.06  18.10.08  18.10.10  18.10.12  18.10.14  18.10.16  18.10.18  18.10.20  18.10.22  18.10.24  18.10.26  18.10.28  18.10.30
```
I'll add this examples to the docs in the future.

Please close.

ninjaaron:
Yeah. I basically took one look at the parser, and I was like, "nah". Now, two years on, I'm not even really using xonsh anymore.

anki-code:
@ninjaaron thank you for the response! Could you please tell us in the nutshell why you're not using xonsh?

ninjaaron:
@anki-code I think there are a few reasons, and none of them are really the fault of xonsh.

- I started noticing that whenever I wanted to test something in the REPL, I was opening a new Python REPL anyway to have a "clean" environment.
- While xontribs are great, there are a lot of benefits to using more widely-used shells from an application integration perspective. I started using Poetry for project management and there was no xonsh support at the time. Meanwhile, I also started programming more in OCaml, which also has shell integration for a lot of the tooling, and obviously none of this was designed to work with xonsh. Same with smart completions and stuff. Bash/Fish/Zsh just have so many more applications that target supporting them.
- At this point, I started using another shell primarily, but then popping into xonsh when I wanted to do some "shell work" (e.g. mass file operations).
- Over time, my workflow for this changed. I seldom use the shell for interactive programming anymore, and I now prefer to write Python or Julia scripts for anything bigger than a pipeline, even if it's just a one-off job.
- As I used xonsh less, I would experience changes between releases that would break my config files. This would have been fine if I was using it every day, but sometimes multiple releases would go by between when I used it, and it sometimes would take 20-30 minutes to get things working again. At this point, I was spending more time fixing my configuration files than I was actually using xonsh.

So, nothing wrong with xonsh. Just little changes in my own workflow that made it less useful to me over time.

anki-code:
Many thanks for the detailed answer! It's interesting feedback! Fyi @scopatz

gwhitney:
I put the following in my .xonshrc:
```
def b(s) :
     mainparts = s.split(':');
     for i in range(len(mainparts)):
         if ',' in mainparts[i]:
             ans = []
             for subpart in mainparts[i].split(','):
                 mainparts[i] = subpart;
                 ans.append(''.join(mainparts));
             return ans
     return s
```

and now I can write for example

```
mv @b`Very long filename with spaces.:PDF,pdf`
```

to achieve `mv 'Very long filename with spaces.PDF' 'Very long filename with spaces.pdf'`, which is most of what I used brace expansion for anyway, in a quite economical command line, thanks to the "custom path searches" facility.

PR #2873, opened 2018-10-11
test fixes

scopatz:
I was seeing a bunch of strange test failures locally. This fixes them.


gforsyth:
Thanks, @scopatz !  This is a weird bunch of changes but the tests pass and there's nothing that seems too untoward in here (although that `builtins.__xonsh__` assignment is pretty weird)

PR #2875, opened 2018-10-12
Spec Index

scopatz:
I think I need this for some downstream code, but it may help with pipelines of callable aliases that want to exchange rich Python objects


gforsyth:
This looks fine to me, but it is consistently failing on 3.5 because of that weird inability to insert `__xonsh__` into the builtins

gforsyth:
Ok, having resolved the weirdo test failures in #2886 I think this can go in.  

Thanks, @scopatz !  Can we just close out #2874 or does it have other stuff in it?

PR #2880, opened 2018-10-15
Patch 1

michaelaye:
Addresses #2823


gforsyth:
Thanks for putting this in @michaelaye !  Just a quick fix needed for the news entry, can you double the backticks around `fish`?  That's why circle is complaining

gforsyth:
Thanks for putting this in @michaelaye !  Merging now -- test failures have been fixed on master and are definitely not related to the docs.

PR #2882, opened 2018-10-15
Pin black version in CI config

dbarbuzzi:
Resolves xonsh#2846 by updating CircleCI config to explicitly use version 18.9b0 (latest at the time of this writing).

When staging this PR in my local fork, the Circle CI builds for `master` had the same output as this branch: 

* `build_black` and `build_34` pass
* `build_35` fails 1 test
```
tests/test_ast.py:54: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

input = 'l = 1\nls -l\n'

    def check_parse(input):
>       tree = builtins.__xonsh__.execer.parse(input, ctx=None)
E       AttributeError: module 'builtins' has no attribute '__xonsh__'
```

* `build_36` fails the lint check

```
/home/circleci/project/xontrib/bashisms.py:35:14: W504 line break after binary operator
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -197,7 +197,7 @@ jobs:
           name: configure conda
           command: |
               export PATH="$HOME/miniconda/bin:$PATH"
-              pip install black
+              pip install black==18.9b0
       - save_cache:
           key: miniconda-v1-black
           paths:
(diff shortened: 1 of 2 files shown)


gforsyth:
Thanks @dbarbuzzi !  Yeah, I think we've fixed those weird test failures on master but they aren't impacted either way by this change.

Thanks for putting this in!

PR #2886, opened 2018-10-16
Potential fix for ast test failures from attribute error

gforsyth:
Potential fix for the test failures on master and several open PRs (the `AttributeError` when trying to access `builtins.__xonsh__`)

No news entry for this one (if it works).  Deeply confused by this but the test with a call to `check_parse` that didn't have a `xonsh_builtins` argument wasn't failing, so thought this was worth a shot.

### DIFF
--- a/tests/test_ast.py
+++ b/tests/test_ast.py
@@ -49,7 +49,7 @@ def test_gather_load_store_names_tuple():
         "l = 1",
     ],
 )
-def test_multilline_num(xonsh_builtins, line1):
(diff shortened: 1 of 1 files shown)


gforsyth:
Ok, it looks like this worked.  The failures on AppVeyor are spurious and I believe that re-running the Azure job should fix the failures there.

scopatz:
Totally weird.  Thanks @gforsyth!

PR #2890, opened 2018-10-19
Fixed output wording for bg command

jay-tuckey:
Hi,

This PR should fix the `bg` command from printing error messages related to the foreground `fg` command.

In particular:
```
jay@jay-alienware-ubuntu:~:x$ bg                                                                                                                                          
Cannot bring nonexistent job to foreground.
```

I didn't think this one needed a changelog entry, but let me know if you would like one.

Jay


scopatz:
Thanks for the PT @jay-tuckey - just the one comment about the API, which I think forces you to have a function that wraps around (and returns) the current function

scopatz:
Also, it probably is worth having a news entry

jay-tuckey:
@scopatz Ah yes, I see, making the change in the `def fg` changes it's function signature.

I've changed my PR a bit, to put most of the logic in a separate function, and then have `fg` and `bg` call that. I've also simplified the wording a bit. What are your thoughts on this approach?

Also, added a news entry.

jay-tuckey:
No worries, I hope you had a great holiday! 😃

PR #2892, opened 2018-10-20
f-strings, again

### LINKED ISSUE #2870: F-string still aren't working correctly.
The #2829 was a fix for #2822. But it handles only a simple case where the variable is the only thing inside braces which is a small subset of possible usage. The code inside braces might be almost any valid xonsh expression.
The #2829 also breaks some things. For example, the first two cases below which is escaping braces with double braces and using f-strings inside an f-string.

Here I provide some cases that aren't currently handled correctly:
```
f"{{$X}}"
f"{f'{$X}'}"
f"{$X + $Y}"
f"{${'X'}}"
f"{repr($X)}"
f"{$(uname)}"
```

Those cases _are_ currently handled correctly but might be useful for testing:
```
f"{'$X'}"
f"a{$X}b"
f"{'}'}"
f"{'$'}"
```

As mentioned in #2829 we might need to parse the f-string content to correctly handle all possible cases. As a reference for implementation, we could check the f-string related code from the cpython's [parser implementation](https://github.com/python/cpython/blob/656d52dbfde3223cd2a3525d652b6cccb02fa991/Python/ast.c#L4927).

scopatz:
This fixes #2870, and also updates the news format to the newer format which confuses fewer people with the None removal

### DIFF
--- a/news/fstrenv.rst
+++ b/news/fstrenv.rst
@@ -4,7 +4,8 @@
 
 **Changed:**
 
-* <news item>
+* Now there is only a single instance of ``string.Formatter()`` in the
+  code base, which is called ``xonsh.tools.FORMATTER``.
 
 **Deprecated:**
 
--- a/news/fstrenv.rst
+++ b/news/fstrenv.rst
@@ -18,7 +18,7 @@
 **Fixed:**
 
 * f-strings (``f"{expr}"``) are now fully capable of executing xonsh expressions.
-  The one exception to this is that ``![cmd]`` and ``!(cmd)`` do work because
+  The one exception to this is that ``![cmd]`` and ``!(cmd)`` don't work because
   the ``!`` character interferes with Python string formatting. If you need to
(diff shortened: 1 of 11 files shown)


news/fstrenv.rst
```diff
@@ -0,0 +1,27 @@
...
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* f-strings (``f"{expr}"``) are now fully capable of executing xonsh expressions.
+  The one exception to this is that ``![cmd]`` and ``!(cmd)`` do work because
```
MrQubo:
```suggestion
  The one exception to this is that ``![cmd]`` and ``!(cmd)`` don't work because
```
Should be `do NOT work`.

scopatz: nice catch!

scopatz:
I think this is ready now!

scopatz:
Thanks @gforsyth! I added some CLI options to force pip to upgrade, which seems to have worked.

gforsyth:
Looks good to me! Thanks, @scopatz!

Issue #2364, opened 2017-04-27
tab completion with arrow selection like zsh


ahundt:
Is there a way in xsh to enable tab completion like the following where it lets me select from the list with the arrow key buttons? (up and down works too if there are many items)

Here is a video of what I mean:

[![zsh tab completion](https://cloud.githubusercontent.com/assets/55744/25497561/06bfe6aa-2b53-11e7-831f-233ef80623a9.png)](https://youtu.be/ew51b7n7QOQ)

Also perhaps there is a color scheme like the one in the video?

astronouth7303:
Completing file names should be default. There's a variety of color schemes available.

ahundt:
The names appear but the arrow doesn't work by default, sorry if that wasn't clear.

gforsyth:
Hi @ahundt -- up and down arrow keys should work with prompt_toolkit, but they don't support left-right as of yet.  See #917

ahundt:
Thanks for the link, am I correct that the issue is closed but the functionality doesn't yet exist?

gforsyth:
@ahundt, I closed the previous issue since it's an upstream thing. But yeah, no right-left arrows yet.

ahundt:
Oh I see that functionality is part of python-prompt-toolkit rather than xonsh 
https://github.com/jonathanslenders/python-prompt-toolkit/issues/147

ok fair enough, thanks!

ahundt:
Ah, these already exist:
```
$XONSH_HISTORY_MATCH_ANYWHERE = True
$COMPLETIONS_CONFIRM = True
```
https://xon.sh/envvars.html#completions-confirm

Any chance these or some even better set of automation options could be enabled by default? :-)

gforsyth:
Hey @ahundt -- I'm definitely game for setting `$COMPLETIONS_CONFIRM=True` as the default -- if we get a quorum of @xonsh/xore to agree to that we can change the default.

I'm a little more wary of the match anywhere, only because it's more computationally intensive, but I'm open to being convinced if you want to run some profiling on search timings with long histories!

scopatz:
I am in favor of `$COMEPLETIONS_CONFIRM = True` if someone wants to put in a PR

anki-code:
I also use True value and PR is here.

PR #2904, opened 2018-10-26
xonshrc.xsh now links to more details

ahundt:
This partially addresses https://github.com/xonsh/xonsh/issues/2364#issuecomment-433540042, but I'm not sure if there is another place I'd need to make a change to include this in the setup wizard.

I'm not sure this really deserves a news item, does it?


melund:
Thanks @ahundt. This can probably go straight in

PR #2907, opened 2018-10-29
Update arch linux install instructions

JakeHedman:
pacman uses the official repositories, not the AUR.


gforsyth:
Thanks, @JakeHedman !  The test failures are unrelated (as you might imagine).  yay docs!

PR #2914, opened 2018-11-01
Fix typo in docs

jsphpl:


melund:
Thanks @jsphpl. This is the right place :)

Test failures here are unrelated.

PR #2921, opened 2018-11-02
Fix import statement with trailing comma within parenthesis

### LINKED ISSUE #2920: A trailing comma in the import list raises syntax error
A trailing comma in the import list raises syntax error (it is valid Python)

```
$ from x import (y, z,)
.
  File "<string>", line None
SyntaxError: <xonsh-code>:1:20: ('code: )',)
from x import (y, z,)
                    ^
```
I have tried and failed to fix it my self :( 

I thought it should be handled by this: 
https://github.com/xonsh/xonsh/blob/master/xonsh/parsers/base.py#L1471-L1478

Or am I wrong here...

laloch:
Fixes #2920.

### DIFF
--- a/news/fix-import-names.rst
+++ b/news/fix-import-names.rst
@@ -18,6 +18,7 @@
 
 * xonsh can now properly parse import statements with trailing comma within
   parentheses, e.g.::
+  
     from x import (y, z,)
 
 **Security:**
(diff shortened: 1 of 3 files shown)


scopatz:
Thanks for putting this in @laloch! Can you please add a test to `tests/test_parser.py` that tests the failing case.  Thanks!

laloch:
I had in fact the test ready. I only forgot to commit it. Thanks for reminding me.

scopatz:
Thanks for fixing this @laloch!

PR #2923, opened 2018-11-02
Set stack level to 3 for meaningful warning messages

hmaarrfk:
Otherwise, I think the warning is coming from that line, but in reality, powerline is using a deprecated API.

Feel free to close the PR if you don't want this.


scopatz:
Thanks @hmaarrfk! This would be great!

scopatz:
I merged in from master

hmaarrfk:
but i had just forked.... weird.

PR #2924, opened 2018-11-02
Update vended version of `ply`

gforsyth:
I got a candy bar from the vendoring machine.

Fixed the makefile, too, so it actually does what it claims to do.


scopatz:
Thanks @gforsyth!

PR #2932, opened 2018-11-09
fix no candidates if no permission files in PATH

### LINKED ISSUE #2931: There is no candidates if no permission files in PATH
## xonfig

```
| Python           | 3.7.1               |
| prompt toolkit   | 2.0.7               |
| distro           | unknown             |
```

## Expected Behavior

All executable files in PATH directory should appear in candidates.

## Current Behavior

There is no candidates after PermissionError file in its path.

## Steps to Reproduce

```
$ ls /usr/local/bin

$ touch /usr/local/bin/aaa
$ chmod 700 /usr/local/bin/aaa
$ sudo chown root /usr/local/bin/aaa
```

There are no tab completion candidates in /usr/local/bin.

--

It perhaps is able to fix by adding errors to catch on [xonsh.tools:L759](https://github.com/xonsh/xonsh/blob/master/xonsh/tools.py#L757-L759).

`file_.is_file()` on L757 may raise `OSError`. Thus, L759 should catch `OSError` (or, `PermissionError` in this case).

philopon:
fixed #2931.

### DIFF
--- a/news/fix-exec-candidates.rst
+++ b/news/fix-exec-candidates.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* fix no candidates if no permission files in PATH
+
+**Security:**
+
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks!

PR #2934, opened 2018-11-10
ExecAlias

scopatz:
This provides a callable alias that enables the execution of multiple-statement code without having to manually write a function.  Note that this is exec'd in the globals & locals of the call site, not where it is defined. So for example, the following is now possible, 

```console
scopatz@artemis ~ $ x = 42                                                               
scopatz@artemis ~ $ aliases["yo"] = "echo hello; echo @(x)"                              
scopatz@artemis ~ $ yo                                                                   
hello
42
```

Of course, writing the alias yourself will give you more control over stdin, stdout, and stderr.  There are probably code transformers we could write to do more implicit redirection here, but this is definitely good enough for a first cut.


scopatz:
@gforsyth @melund - I *finally* have the test suite passing here and all fixed up for the various platforms.

gforsyth:
Hey @scopatz -- sorry for the pain with the test suite.  The `ExecAlias` stuff looks good to me and the testing changes all make sense.  I'd like to get to the bottom of the `ptk highlight` test thing, but that doesn't need to languish here.

gforsyth:
Thanks for putting this in @scopatz !

scopatz:
Thanks @gforsyth! Yeah, I agree!

Issue #2929, opened 2018-11-08
vox recognises only venvs with pyvenv.cfg


halloleo:
`vox list` (and `vox activate` for that matter) recognises only virtual environments with `pyvenv.cfg`.

However for the functionality of the virtual environment the config file is not needed. Pipenv for example creates venvs *without* a config file - and `vox` can even activate them if I add an `pyvenv.cfg` by hand.

I think `vox` should recognise these virtual environments by default.

## xonfig

    +------------------+---------------------------------+
    | xonsh            | 0.8.3                           |
    | Git SHA          | 07eb341c                        |
    | Commit Date      | Nov 6 07:39:16 2018             |
    | Python           | 3.6.4                           |
    | PLY              | 3.11                            |
    | have readline    | True                            |
    | prompt toolkit   | 2.0.7                           |
    | shell type       | readline                        |
    | pygments         | None                            |
    | on posix         | True                            |
    | on linux         | True                            |
    | distro           | Red Hat Enterprise Linux Server |
    | on darwin        | False                           |
    | on windows       | False                           |
    | on cygwin        | False                           |
    | on msys2         | False                           |
    | is superuser     | False                           |
    | default encoding | utf-8                           |
    | xonsh encoding   | utf-8                           |
    | encoding errors  | surrogateescape                 |
    +------------------+---------------------------------+

gforsyth:
Closed via #2937

halloleo:
Cool! #2937 fixes this. 

With some forum threads I have asked in the community about `pyvenv.cfg` and virtual environments to get a clearer picture. Basically, Pipenv won't change and there is no general consensus of what defines a virtual environment. So going with testing for the `activate` script (as #2937 does) sounds sensible.

For reference here the links to the threads:

* Pipenv-Github: [Issue #3303](https://github.com/pypa/pipenv/issues/3303) and [ Issue #15](https://github.com/pypa/pipenv/issues/15)

* Stackoverflow: [What makes in Python a virtual environment a virtual environment?](https://stackoverflow.com/questions/53281113/what-makes-in-python-a-virtual-environment-a-virtual-environment?noredirect=1#comment93466983_53281113)

PR #2937, opened 2018-11-13
addressed issue #2929

gsaga:


melund:
I am no familiar with how pyenv work, but will this change also work on Windows?

gforsyth:
Hey @gsaga -- thanks for putting this in!  
Sorry for the delay -- it's probably a good idea to ping on the PR after you push up a new commit.  I think most if not all of the maintainers have disabled the commit-notification emails so we don't get inundated with notifications.
I'm restarting AppVeyor to see if we can get that passing.

gforsyth:
Also -- I see that you put in code to handle pyenv on windows (great!) but do we have any tests that are actually hitting that line of code?

gforsyth:
Ah, I see that we do.  Great!  Thanks again for putting this in @gsaga !

PR #2959, opened 2018-12-13
dumb terminal support

### LINKED ISSUE #2891: xonsh on a dumb terminal (because it runs in an Emacs comint buffer)
I run `xonsh`  from a `bash` inside an Emacs `comint` buffer. 

## xonfig

```

| xonsh            | 0.7.9           |
```

## Expected Behavior

When starting `xonsh` from the `bash` shell the environment variable `TERM` is set to `dumb`, so I would hope `xonsh` doesn't use any coloring or other escape sequences.  Or if that's not possible I would like to know how I can manually switch off the use of all escape sequences, say, via the `.xonshrc` file.

## Current Behavior

`xonsh` seems to emit escape sequences despite a dumb terminal. (The prompt seems to be written in invisible colors). Also I get the following message:

    WARNING: your terminal doesn't support cursor position requests (CPR).

## Steps to Reproduce

In a graphical GNU Emacs 26.1 on macOS start a shell comint buffer, then run `xonsh`.

scopatz:
This should fix the issues with emacs support. closes #2891

### DIFF
--- a/docs/api/dumb_shell.rst
+++ b/docs/api/dumb_shell.rst
@@ -1,0 +1,10 @@
+.. _xonsh_dumb_shell:
+
+******************************************************
+Dumb Shell (``xonsh.dumb_shell``)
+******************************************************
+
+.. automodule:: xonsh.dumb_shell
+    :members:
+    :undoc-members:
+    :inherited-members:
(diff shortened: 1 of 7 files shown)


scopatz:
Pinging this!

gforsyth:
Thanks, @scopatz !  I don't use `comint-run` but I tried it out and can confirm that it's broken on master and that it works on this branch.

scopatz:
Excellent! Now I can uninstall emacs :wink:

PR #2954, opened 2018-12-06
Fix interpretation of color names with PTK2 and Pygments 2.3

laloch:
Fixes #2948.
@melund It turned out to be fairly simple. Let's hope your upstream PR gets merged before 2.3.1 ;)

Edit: Maybe it requires clarification that I intentionally set the upper bound to 2.3.0. The upstream PR has already been approved, so I expect the color monkey patching to not be needed for 2.3.1 any more.

PR #2963, opened 2018-12-16
Fix color names for PTK2 and Pygments 2.3.1

laloch:
This is a follow-up to #2954. The [upstream PR](https://bitbucket.org/birkenfeld/pygments-main/pull-requests/777/change-ansi-color-names-to-more-saying/diff) was already merged for Pygments 2.4, so this should be the last time we need to take care about it.

### DIFF
--- a/news/ptk2-pygments-2_3_1-colors.rst
+++ b/news/ptk2-pygments-2_3_1-colors.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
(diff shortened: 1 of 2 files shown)


melund:
Thanks a lot, @laloch. I am really sorry the fix to Pygments didn't enter 2.3, but if 2.4 is really coming at the end of Januar, it may not be so bad. Pygments just have a record of really slow release cycles :(

scopatz:
Thanks @laloch @melund for working on this!

PR #2947, opened 2018-11-28
Doco how to update xonsh

halloleo:
Doco about how to update xonsh and how to unset environment variables

Relates to later comments in issue #2928.

Issue #2928, opened 2018-11-07
Executing xonsh in xonsh results in AttributeError


halloleo:
When I'm executing `xonsh` in a running xonsh shell and I load the `vox` xontrib in the `.xonshrc` I get a strange error:

    xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
    Traceback (most recent call last):
      File "/working/software/venv/xonsh/lib/python3.6/site-packages/xonsh/environ.py", line 1300, in xonsh_script_run_control
        run_script_with_cache(filename, execer, ctx)
      File "/working/software/venv/xonsh/lib/python3.6/site-packages/xonsh/codecache.py", line 162, in run_script_with_cache
        run_compiled_code(ccode, glb, loc, mode)
      File "/working/software/venv/xonsh/lib/python3.6/site-packages/xonsh/codecache.py", line 67, in run_compiled_code
        func(code, glb, loc)
      File "/home/halloleo/.xonshrc", line 3, in <module>
        xontrib load vox 
      File "/working/software/venv/xonsh/lib/python3.6/site-packages/xonsh/built_ins.py", line 952, in subproc_captured_hiddenobject
        return run_subproc(cmds, captured="hiddenobject")
      File "/working/software/venv/xonsh/lib/python3.6/site-packages/xonsh/built_ins.py", line 904, in run_subproc
        pause_call_resume(proc, builtins.__xonsh__.shell.settitle)
    AttributeError: 'NoneType' object has no attribute 'settitle'
    error running xonsh run control file '/home/halloleo/.xonshrc': 'NoneType' object has no attribute 'settitle'

The `.xonshrc` file to make this bug happen is:

    $XONSH_SHOW_TRACEBACK = True
    $SHELL_TYPE='readline'
    xontrib load vox 

My `xonfig` output is:

    +------------------+---------------------------------+
    | xonsh            | 0.8.1                           |
    | Git SHA          | 23246604                        |
    | Commit Date      | Oct 18 09:55:19 2018            |
    | Python           | 3.6.4                           |
    | PLY              | 3.9                             |
    | have readline    | True                            |
    | prompt toolkit   | 2.0.7                           |
    | shell type       | readline                        |
    | pygments         | None                            |
    | on posix         | True                            |
    | on linux         | True                            |
    | distro           | Red Hat Enterprise Linux Server |
    | on darwin        | False                           |
    | on windows       | False                           |
    | on cygwin        | False                           |
[...]

PR #2964, opened 2018-12-18
Doco update xonsh

halloleo:
Doco about how to update xonsh and how to set/unset environment variables.

Relates to later comments in issue #2928.

Replace the pull request #2947.


docs/customization.rst
```diff
@@ -1,11 +1,36 @@
 =====================
-Customizing ``xonsh``
+Updating and customizing xonsh
 =====================
```
scopatz: The `===` need to go all the way below and above the title text.

melund:
Looks good @halloleo. Thanks a lot for adding this.

docs/customization.rst
```diff
@@ -1,11 +1,36 @@
...
 
 .. contents::
    :local:
 
-How do I...
+Updating xonsh
+===========
```
scopatz: same here

melund: Ahhh. I missed that. Maybe we should have our docs build fail if there are warnings.

docs/customization.rst
halloleo: Sure, makes sense. Will push the update in  a minute.

docs/customization.rst
halloleo: Probably a good idea to let the build fail if you want to enforce this (I don't think the RST format needs it, but I might be wrong).

scopatz: rest definitely needs this, and we do have a doc test on Travis, so I think this might just be a warning that renders incorrectly.  Maybe there is a way to turn that into a hard failure. Thanks again!

halloleo:
If rest needs it why can I build the documentation with shorter “=“ 
lines via sphinx (I.e.  `make html`) then?

I’m just curious about rest, sphinx and the like; I don’t want to 
question the style standard.

melund:
Well. Sphinx will issue warnings when the lines isn't long enough. With too many warnings it is hard to catch the ones that matter.

scopatz:
Yeah, that is true.  And even though it will produce a document, it often has incorrect formatting.

PR #2967, opened 2018-12-19
cat exit

### LINKED ISSUE #2827: Xonsh crashes when interrupting running process
## xonfig

```
| xonsh            | 0.7.8                |
| Git SHA          | ecca7a7c             |
| Commit Date      | Sep 11 13:48:32 2018 |
| Python           | 3.7.0                |
| prompt toolkit   | 2.0.4                |
```

## Expected Behavior

When running a command that takes longer (or forever, like cat /dev/random) interrupting it (with ctrl-c) will cause the running process to terminate.
## Current Behavior

When running a command that takes longer (or forever, like cat /dev/random) interrupting it (with ctrl-c) will cause the xonsh shell to close.

## Steps to Reproduce

execute ```cat /dev/random```, then ctrl-c

Please note that I'm using xonsh in byoubu (screen) with mutliple xonsh instances open. When the bug occurs, the affected instance (tab) is closed.
I just tested it without screen, the shell window doesn't close but the process (cat) doesn't terminate either.

scopatz:
This is kind of an involved fix, but it turns out that cat was failing in some pretty bad ways. I have fixed up all of the use cases I could imagine. It should be noted that cat is used pretty heavily in the test suite so these fixes are pretty well tested.

Fixes #2827


scopatz:
@xonsh/xore this is ready for review!

melund:
It seems to work in the situations I have tested. But I don't think I have seen the bad cases. I am OK with merging even though I can't readily comprehend all the fixes ;-)

scopatz:
Thanks!

Issue #2347, opened 2017-03-28
vox image not found on OSX


brianmay:
```
$ vox new abc
Creating environment...
subprocess.CalledProcessError: Command '['/Users/brianmay/.virtualenvs/abc/bin/python3.6', '-Im', 'ensurepip', '--upgrade', '--default-pip']' died with <Signals.SIGABRT: 6>.
```

Running this copy of python3.6 manually I get:

```
$ /Users/brianmay/.virtualenvs/abc/bin/python3.6
dyld: Library not loaded: @executable_path/../.Python
  Referenced from: /Users/brianmay/.virtualenvs/abc/bin/python3.6
  Reason: image not found
Aborted (core dumped)
```

I believe this should work...

mitnk:
Does venv work with bash?

mitnk:
Could you try following command in bash to see if it works:
```
bash $ python3.6 -m venv foo
bash $ source foo/bin/activate
```
I think it may not related xonsh.

brianmay:
The above lines work perfectly. I tried both with both versions of Python3.6 (I have no idea why the brew xonsh package for OSX installs its own version of xonsh - might be better to use the pip package). Odd, I wasn't expecting that...

I note that the python3.6 binary inside the chroot is normally a symlink:

```
$ /usr/local/Cellar/xonsh/0.5.9/libexec/bin/python3.6 -m venv foo                     
$ source foo/bin/activate
$ ls -l foo/bin 
total 48
-rw-------  1 brianmay  staff  2147 29 Mar 09:21 activate
-rw-------  1 brianmay  staff  1263 29 Mar 09:21 activate.csh
-rw-------  1 brianmay  staff  2427 29 Mar 09:21 activate.fish
lrwx------  1 brianmay  staff     9 29 Mar 09:21 python -> python3.6
lrwx------  1 brianmay  staff     9 29 Mar 09:21 python3 -> python3.6
lrwx------  1 brianmay  staff    51 29 Mar 09:21 python3.6 -> /usr/local/Cellar/xonsh/0.5.9/libexec/bin/python3.6
```

However via vox I don't get a symlink:

```
$ vox new abc
Creating environment...
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK = True
[...]

Issue #692, opened 2016-02-19
Feature Request: Vox: Choose Python version with `-p`, similar to virtualenv and other tools


moigagoo:
Currently, Vox invokes `venv.create` from Python 3 to create environments. This means that you can't create Python 2 environments.

Workaround: Create the environment in $VIRTUALENV_HOME (~/.virtualenvs by default) manually with `virtualenv -p python2 ~./virtualenvs`.

moigagoo:
The problem with implementing this feature it we have to add virtualenv as a dependency.

scopatz:
@moigagoo, as long as it is an optional dependency I don't see the problem.

eykd:
+1 to this. I still need to be able to work with py2.7 virtualenvs, and I seem to be shut out from even using existing virtualenvs because bin/activate no longer works. 😢

scopatz:
Pull requests most welcome!  Preferably by someone who uses virtual envs :)

eykd:
Who doesn't use virtualenvs? 😉  I don't understand xonsh's or virtualenv's inner workings well enough to dash anything off. Also, if pyenv is involved, it gets weird (the OP workaround doesn't work for me). I did, however, figure out a workaround, which may or not be helpful to anyone else.

Put this in your `.xonshrc`:

```
def _venv(args, stdin=None):
    """venv <python-version> <env-name>

    Activate with `vox activate <env-name>`
    """
    version, name = args
    root = $PYENV_ROOT
    pyversion = '%(root)s/versions/%(version)s/bin/python' % locals()
    venv_path = path('~/.virtualenvs').expanduser() / name
    print("Creating a virtualenv with Python %(version)s in %(venv_path)s" % locals())
    virtualenv -p @(pyversion) @(venv_path)


aliases['venv'] = _venv
```

scopatz:
Cool. Maybe you could add this the docs as a potential work around: http://xon.sh/python_virtual_environments.html

Granitas:
@eykd why break the `virtualenv` syntax though? 

So this code is more straigh-forward and elegant:

```python
#!/usr/bin/env xonsh
import os

def _venv(args, stdin=None):
    """
    venv <env-name> <python>
    Activate with `vox activate <env-name>`
    """
    name, version = args
    venv_path = os.path.join(os.path.expanduser('~/.virtualenvs'), name)
    print('Creating a virtualenv "{}" with "{}"'.format(name, version))
[...]

PR #2968, opened 2018-12-21
allow user to invoke specific interpreter for vox virtualenv creation

knowsuchagency:
This PR allows one to pass a `-p` or `--interpreter` argument to `vox new` in order to specify a specific interpreter to virtual environment creation.

If not passed, the interpreter will be determined by the value of the `$VOX_DEFAULT_INTREPRETER` environment variable. If not set, vox will default to using `xonsh`'s `sys.executable`.

This PR addresses the following issues:

#2347
#692

### DIFF
--- a/tests/test_news.py
+++ b/tests/test_news.py
@@ -29,7 +29,7 @@ def check_news_file(fname):
                 pytest.fail(
                     "{}:{}: single grave accents"
                     " are not valid rst".format(name, i + 1),
-                    pytrace=False,
+                    pytrace=True,
                 )
 
         # determine the form of line
@@ -40,7 +40,7 @@ def check_news_file(fname):
                     "{}:{}: {!r} not a proper category "
                     "must be one of {}"
                     "".format(name, i + 1, cat, list(CATEGORIES)),
-                    pytrace=False,
+                    pytrace=True,
                 )
             if l.endswith("None"):
                 form += "3"
@@ -53,14 +53,14 @@ def check_news_file(fname):
         elif l.strip() == "":
             form += "0"
         else:
-            pytest.fail("{}:{}: invalid rst".format(name, i + 1), pytrace=False)
+            pytest.fail("{}:{}: invalid rst".format(name, i + 1), pytrace=True)
     # The file should have:
     #   empty lines around categories
     #   at least one content line in a non null category
     reg = re.compile(r"^(3(0|$)|20(1|4)(1|0|4)*0|204$)+$")
     if not reg.match(form):
         print(form)
-        pytest.fail("{}: invalid rst".format(name), pytrace=False)
+        pytest.fail("{}: invalid rst".format(name), pytrace=True)
 
 
 @pytest.mark.parametrize("fname", list(scandir(NEWSDIR)))
--- a/xontrib/vox.py
+++ b/xontrib/vox.py
@@ -1,6 +1,7 @@
 """Python virtual environment manager for xonsh."""
 
 import sys
+import textwrap
 import xontrib.voxapi as voxapi
 import xonsh.lazyasd as lazyasd
 
@@ -30,8 +31,13 @@ class VoxHandler:
         create.add_argument(
             "-p",
             "--interpreter",
-            default=voxapi.VOX_DEFAULT_INTERPRETER,
-            help=f"The Python interpreter used to create the virtual environment default: {voxapi.VOX_DEFAULT_INTERPRETER}",
+            default=None,
+            help=textwrap.dedent(
+                """
+                The Python interpreter used to create the virtual environment. 
+                Can be configured via the $VOX_DEFAULT_INTERPRETER environment variable.
+                """
+            ).strip(),
         )
 
         from xonsh.platform import ON_WINDOWS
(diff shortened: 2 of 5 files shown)


xontrib/vox.py
```diff
@@ -27,6 +27,13 @@ def parser():
                             help='Give the virtual environment access to the '
                                  'system site-packages dir.')
 
+        create.add_argument(
+            "-p",
+            "--interpreter",
+            default=voxapi.VOX_DEFAULT_INTERPRETER,
+            help=f"The Python interpreter used to create the virtual environment default: {voxapi.VOX_DEFAULT_INTERPRETER}",
```
scopatz: Since we still need to support v3.4, we can't use f-strings just yet :cry:

xontrib/voxapi.py
```diff
@@ -22,6 +23,10 @@
 # FIXME: Is there a better way?
 from xonsh.events import events
 
+VOX_DEFAULT_INTERPRETER = builtins.__xonsh__.env.get(
+    "VOX_DEFAULT_INTERPRETER", sys.executable
+)
```
scopatz: It is usually better to pull the env vars out right before they are used, rather than at module-level scope.

xontrib/voxapi.py
```diff
@@ -114,14 +119,23 @@ def __init__(self):
         else:
             self.venvdir = builtins.__xonsh__.env['VIRTUALENV_HOME']
 
-    def create(self, name, *, system_site_packages=False, symlinks=False,
-               with_pip=True):
+    def create(
+        self,
+        name,
+        *,
+        interpreter=VOX_DEFAULT_INTERPRETER,
```
scopatz: I would maybe change this to `interpreter=None` and then in the body of the function, have 

```
interpreter = builtins.__xonsh__.env.get("VOX_DEFAULT_INTERPRETER", sys.executable) if interpreter is None else interpreter
```

xontrib/voxapi.py
```diff
@@ -138,14 +152,23 @@ def create(self, name, *, system_site_packages=False, symlinks=False,
...
         events.vox_on_create.fire(name=name)
 
-    def upgrade(self, name, *, symlinks=False, with_pip=True):
+    def upgrade(
+        self,
+        name,
+        *,
+        symlinks=False,
+        with_pip=True,
+        interpreter=VOX_DEFAULT_INTERPRETER,
```
scopatz: Same here

scopatz:
Thanks for putting this in @knowsuchagency!  Just a couple of small structural comments, and as you said needs documentation. Also, could you please add a new entry? See https://xon.sh/devguide.html#changelog

knowsuchagency:
Thanks for the feedback @scopatz , I'll submit those changes along with the documentation piece :)

scopatz:
Awesome! Just let us know when it is ready

knowsuchagency:
Should be good-to-go @scopatz 

Appveyor seems to be failing on the special test `['pytest', '--ignore', 'tests/test_ptk_highlight.py']`. You probably have more context on that than I

tests/test_news.py
```diff
@@ -29,7 +29,7 @@ def check_news_file(fname):
                 pytest.fail(
                     "{}:{}: single grave accents"
                     " are not valid rst".format(name, i + 1),
-                    pytrace=False,
+                    pytrace=True,
```
knowsuchagency: I actually didn't mean to commit this file, but it may not be a bad idea since the reason I altered the file was to save myself time debugging my mistakes with the `.rst` file I created. I'm sure I won't be the last

xontrib/voxapi.py
```diff
@@ -155,33 +171,71 @@ def upgrade(self, name, *, symlinks=False, with_pip=True):
...
+        interpreter_major_version = int(version_output.split()[-1].split(".")[0])
+        module = "venv" if interpreter_major_version >= 3 else "virtualenv"
+        system_site_packages = "--system-site-packages" if system_site_packages else ""
+        symlinks = "--symlinks" if symlinks and interpreter_major_version >= 3 else ""
+        with_pip = "" if with_pip else "--without-pip"
+        upgrade = "--upgrade" if upgrade else ""
+
+        cmd = shlex.split(
+            "{interpreter} -m {module} {env_path} {system_site_packages} {symlinks} {with_pip} {upgrade}".format(
+                **locals()
```
scopatz: I think that this might be messing up on windows due to `shlex.slpit()` removing backslashes and thus the args list looks like https://ci.appveyor.com/project/AnthonyScopatz/xonsh-1m7j9/builds/21195384/job/ppc2cdb02lgi0u5h#L890

Would it be OK if I pushed to your branch to fix this?

astronouth7303: I would strongly encourage you _not_ using `shlex` for this use.

Use a list form of the command instead of string interpolation.

There is just too high a risk of unintended consequences and side effects when using string commands. Always use lists.

knowsuchagency:
Feel free to push to the branch @scopatz or @astronouth7303 if you'd like to make any changes, but I think it's otherwise finished. Thanks, guys!

scopatz:
Thanks @knowsuchagency! I made one small change, but otherwise this was good to go!

PR #2970, opened 2018-12-21
Add `ls` to `threadable_predictors` as `predict_true`

### LINKED ISSUE #2946: Why do cat FILE and ls DIR behave differently in !(...)?
Why do `cat FILE` and `ls DIR` behave differently in `!(...)`?

## Expected Behavior
`!(cat FILE-LISTING_OF_DIR)` and `!(ls -1 DIR)` behave the same in Python mode giving a list of file names to iterate over.

## Current Behavior

`!(cat FILE-LISTING_OF_DIR)` works, `!(ls -1 DIR)` doesn't. Details in the next section:

## Steps to Reproduce

If I do 

    $ ls -1 $VIRTUALENV_HOME > ~/tmp/venvs.txt
    $ venvs = [ d for d in map(str.strip, !(cat ~/tmp/venvs.txt)) if Path(d).is_dir ]

I get - as expected - in `venvs` a Python list of all VENV directories

However, if I issue in one go

    $ venvs = [ d for d in map(str.strip, !(ls -1 $VIRTUALENV_HOME)) if Path(d).is_dir ]

I get the venvs listed on stdout and the variable `venvs` is empty.

Why on earth does this happen? And can that be right?

## xonfig

    | Python           | 3.7.0               |
    | prompt toolkit   | 2.0.4               |
    | shell type       | readline            |

gforsyth:
Fixes #2946

### DIFF
--- a/news/ls_predict_true.rst
+++ b/news/ls_predict_true.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Set ``ls`` to ``predict_true`` in ``default_threadable_predictors``.  This prevents ``ls`` on OSX
+  from being flagged on OSX as unthreadable (incorrectly) because it relies on ``ncurses``.
+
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks @gforsyth!

PR #2972, opened 2018-12-25
Add `pf` strings

### LINKED ISSUE #2524: Add pf-strings
Add support for pf-strings. That is, an f-string that is cast to a `Path`.

gforsyth:
Ok, this still needs some documentation but I got it working and wanted to put it up for review (plus it's fun!).

This resolves #2524 and adds a new bit of syntax which combines p-strings and f-strings.  
This means you can now do things like:

```
gil@bad_cat ~ $ repos = 'github.com'                                           
gil@bad_cat ~ $ fp"~/{repos}"
PosixPath('/home/gil/github.com')
```

but since @scopatz also enabled xonshy f-strings, you can also do things like

```
gil@bad_cat ~ $ pF"{$HOME}"                                              
PosixPath('/home/gil')
```

or you can keep going deeper and try

```
gil@bad_cat ~ $ pf"/home/${'US' + 'ER'}"                                 
PosixPath('/home/gil')
```

This currently doesn't support combining the `pf` prefix with the raw-string prefix because I couldn't think of a reason for them off the top of my head, but if someone wants it, just holler.

Also, just following along with our existing p-string work we support only lowercase `p` prefixes, but either case of `f` -- do we want to remove that ambiguity?

### DIFF
--- a/news/pf_strings.rst
+++ b/news/pf_strings.rst
@@ -1,0 +1,34 @@
+**Added:**
+
+* New xonsh syntax ``pf`` strings -- combining path strings with f-strings.
+
+  Usage:
+
+  .. code-block:: bash
+       gil@bad_cat ~ $ repos = 'github.com'
+       gil@bad_cat ~ $ pf"~/{repos}"
+       PosixPath('/home/gil/github.com')
+       gil@bad_cat ~ $ pf"{$HOME}"
+       PosixPath('/home/gil')
+       gil@bad_cat ~ $ pf"/home/${'US' + 'ER'}"
+       PosixPath('/home/gil')
+
+**Changed:**
+
+* <news item>
+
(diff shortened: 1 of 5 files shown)


gforsyth:
Hmm, so actually you could already do the middle example -- this works:

```
gil@bad_cat ~ $ p"$HOME"                                                 
PosixPath('/home/gil')
```

but not the other two.  And you _definitely_ couldn't do this (not that you should, necessarily...):

```
gil@bad_cat ~ $ repos = 'github.com'                                     
gil@bad_cat ~ $ p"/home" /  pf"${'US'+'ER'}" / pf"{repos}"               
PosixPath('/home/gil/github.com')
```

scopatz:
Amazing! Thanks so much @gforsyth!

PR #2979, opened 2019-01-05
Adds xontrib which supports direnv to xontrib list

74th:
I wants to use [direnv](https://direnv.net/) with xonsh.
I made a xontrib, and published to [PyPI package xonsh-direnv](https://pypi.org/project/xonsh-direnv/).
I'm grad if you add it to `xontrib list`.

### DIFF
--- a/news/xontrib-direnv.rst
+++ b/news/xontrib-direnv.rst
@@ -1,0 +1,3 @@
+**Added:**
+
+* A `xontrib <https://github.com/74th/xonsh-direnv>`_ which adds support for `direnv <https://direnv.net/>`_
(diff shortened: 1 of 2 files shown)


gforsyth:
Hey @74th -- thanks so much for putting this in!  The test failures are coming from the single grave accents you used in your news file.  I think our check for valid uses of grave accents is wrong, though, and is incorrectly flagging your RST links as incorrect.  We'll fix that up and then we can re-run CI

74th:
I fixed rst error, but one build error remains (at tests/test_ptk_highlight.py). 
What can i do for this?

74th:
I ran AppVeyor tests for my same commit, one test succeeded, but one test failed.
https://github.com/74th/xonsh/pull/1
I ran tests for current master(0.8.8), same test failed.
https://ci.appveyor.com/project/74th/xonsh/builds/21591386
It seems current master has an error by ptk.

scopatz:
Thanks! And sorry for the delay from the holidays and working with CIs.  We'll try to cut a release soon.

PR #2992, opened 2019-01-28
fixed eval aliases

### LINKED ISSUE #2981: Why does a normal alias to 'history show all' fails in a pipe and how can I create a pipe-able alias?
Why does a normal alias to 'history show all' fails in a pipe and how can I create a pipe-able alias?

## Expected Behavior

With the alias

    aliases['hsa'] = ['history','show', 'all']

I expect that 

    hsa | head 

gives me the first 10 history entries as

    history show all | head

does. (`head` is here only an example. The problem occurs with all sorts of pipe filters like `grep`, `sed`, etc)

## Current Behavior

Instead when issuing

    hsa | head 

I get many lines of history output (but not all) followed by the error text:

    Unhandled exception in thread started by 

Note that after the `by` comes nothing...

## My Questions

1. **Why does this happen?** Ok, `history` is a function, not a subproces command, but why does this affect the piping *only* in an alias not when using the function directly?

2. **How can I create an alias for 'history show all' which does work with a pipe?** (I tried a callable with `return $('history show all')`, but this fails of course because history is not a subproces command.)

## xonfig

```
+------------------+---------------------------------+
| xonsh            | 0.8.3                           |
| Git SHA          | 07eb341c                        |
| Commit Date      | Nov 6 07:39:16 2018             |
| Python           | 3.6.4                           |
| PLY              | 3.11                            |
| have readline    | True                            |
| prompt toolkit   | 2.0.7                           |
| shell type       | readline                        |
| pygments         | None                            |
| on posix         | True                            |
| on linux         | True                            |
| distro           | Red Hat Enterprise Linux Server |
| on darwin        | False                           |
| on windows       | False                           |
| on cygwin        | False                           |
| on msys2         | False                           |
| is superuser     | False                           |
| default encoding | utf-8                           |
| xonsh encoding   | utf-8                           |
| encoding errors  | surrogateescape                 |
+------------------+---------------------------------+
```

scopatz:
closes #2981

### DIFF
--- a/news/hsah.rst
+++ b/news/hsah.rst
@@ -1,0 +1,26 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Some minor ``history show`` efficiency improvements.
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed issue with recursive aliases not being passes all keyword arguments
+  that are part of the callable alias spec. This allows commands like
+  ``aliases['hsa'] = "history show all"; hsa | head`` to no longer fail
+  with strange errors.
+
+**Security:**
+
+* <news item>
--- a/tests/test_aliases.py
+++ b/tests/test_aliases.py
@@ -13,7 +13,7 @@ from xonsh.environ import Env
 from tools import skip_if_on_windows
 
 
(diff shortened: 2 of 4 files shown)


scopatz:
OK CI has been fixed here.

gforsyth:
Thanks, @scopatz !

PR #2995, opened 2019-01-29
https for github templates

techtonik:
It is worth to update repo description as well.


scopatz:
Just docs no need for CI or news. Thanks!

PR #2993, opened 2019-01-28
Broken symlinks

scopatz:
fixes  #2982

PR #2996, opened 2019-01-29
env_prefix and env_postfix

scopatz:
This came up in https://github.com/pypa/virtualenv/issues/1299 and is probably the right way to implement the env name printing.

@gforsyth - will this mess you up for xonda?

I have based this on #2993, mostly so I don't have redo all of the CI work.  Also I ran black on the prompt subpackage

### DIFF
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -2,7 +2,7 @@ ply
 py
 pytest
 flake8
-pyflakes
+pyflakes<2.1.0,>=2.0.0
 pytest-flake8
 pytest-cov
 pytest-timeout
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -1,12 +1,12 @@
 ply
 py
 pytest
+flake8
+pyflakes
 pytest-flake8
 pytest-cov
 pytest-timeout
 prompt-toolkit
 pygments>=2.2
 codecov
-flake8
-pyflakes
 coverage
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -2,7 +2,6 @@ ply
 py
 pytest
 flake8
-pyflakes<2.1.0,>=2.0.0
 pytest-flake8
 pytest-cov
 pytest-timeout
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -23,6 +23,7 @@ jobs:
   # Create and activate a Conda environment.
   - task: CondaEnvironment@1
     inputs:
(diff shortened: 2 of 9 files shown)


requirements-tests.txt
```diff
@@ -1,12 +1,12 @@
 ply
 py
 pytest
+flake8
+pyflakes<2.1.0,>=2.0.0
```
asottile: (sorry for the driveby) -- if you remove `pyflakes` from this file, `flake8` will automatically resolve the correct version (it has ~an equivalent requirements in its `setup.py`)

scopatz: Ahh OK thanks!

asottile: without this, you're hitting a bug in `pip` (which doesn't actually implement dependency resolution)

basically first-pass `pip` greedily sees `pyflakes` and says "latest version of pyflakes please!", then later it sees in the second pass that flake8 wants pyflakes>=2.0,<2.1 and says "I've already seen pyflakes, satisfied!" even when it's a conflict :sob:

scopatz: Yeah.... I figured that was happening. :frowning_face:

bskinn:
Hrm -- I was picturing implementing it such that the prefix and postfix wouldn't need to be explicitly included in the `$PROMPT` markup... that `{env_name}` would pull in `{env_prefix}` and `{env_postfix}` by default.

That would make `$PROMPT` more concise; and I can't picture any scenarios offhand where someone would want to reuse, e.g., `{env_prefix}` in a prompt.

scopatz:
@bskinn - that is what is on the current master now, but I guess you are saying that the `env_name()` function would actually look up `env_prefix` and `env_postfix` in the `$PROMPT_FIELDS` and then evaluate them? 

That is a nice idea... and a little cleaner.

bskinn:
> I guess you are saying that the `env_name()` function would actually look up `env_prefix` and `env_postfix` in the `$PROMPT_FIELDS` and then evaluate them?

Yep, exactly -- the same behavior as now, except that the values are configurable, rather than hard-coded.

scopatz:
OK, got it.  Just pushed up that change.

scopatz:
>  The latest release of conda claims to have xonsh support. Now, that support is broken and doesn't work, but still, we're almost there.

haha yeah, it is probably a good idea to just move xonda into conda :)

gforsyth:
looks good. thanks @scopatz !

PR #3000, opened 2019-01-30
 send SIGCONT in run_subproc

### LINKED ISSUE #2999: XONSH_STORE_STDOUT=True causes some commands to hang
## xonfig
```
| xonsh            | 0.8.8                |
| Git SHA          | 47585920             |
| Commit Date      | Jan 30 12:30:54 2019 |
| Python           | 3.7.2                |
| prompt toolkit   | 2.0.7                |
| pygments         | 2.3.1                |
```

## Expected Behavior
Xonsh shouldn't hang

## Current Behavior
xonsh hangs

## Steps to Reproduce
1. Put this in `rc.xsh`:
```
$XONSH_STORE_STDOUT = True
```
2. run `xonsh`
3. run `cat` command

Might be related to #2965 and some other issues.

sagartewari01:
It's a hack to fix #2999 and other freaky issues. This is what's causing #2999:
If XONSH_STORE_STDOUT is True
1. [this](https://github.com/sagartewari01/xonsh/blob/27067b6d4e55cc6503c52e221d7ac6aa84176eaa/xonsh/built_ins.py#L863) function returns False.
2. The call to `pause_call_resume` is skipped (https://github.com/sagartewari01/xonsh/blob/27067b6d4e55cc6503c52e221d7ac6aa84176eaa/xonsh/built_ins.py#L907)
3. `pause_call_resume` sends `SIGCONT` to the process.
So it appears to be a `RAII` problem.
As far as I know, sending multiple `SIGCONT`s shouldn't be a problem, so this PR shoudn't have any side effects.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -905,7 +905,8 @@ def run_subproc(cmds, captured=False):
     if _should_set_title(captured=captured):
         # set title here to get currently executing command
         pause_call_resume(proc, builtins.__xonsh__.shell.settitle)
-    proc.send_signal(signal.SIGCONT)
+    else:
+        pause_call_resume(proc, lambda : 0)
     # create command or return if backgrounding.
     if background:
         return
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -906,7 +906,11 @@ def run_subproc(cmds, captured=False):
         # set title here to get currently executing command
         pause_call_resume(proc, builtins.__xonsh__.shell.settitle)
     else:
-        pause_call_resume(proc, lambda : 0)
+        # for some reason, some programs are in a stopped state when the flow
+        # reaches this point, hence a SIGCONT should be sent to `proc` to make
+        # sure that the shell doesn't hang. This `pause_call_resume` invocation
+        # does this
+        pause_call_resume(proc, int)
     # create command or return if backgrounding.
     if background:
         return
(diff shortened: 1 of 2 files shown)


scopatz:
Hey @sagartewari01 - thanks for putting this in!  This is a really cool fix and thanks so much for hunting this down.  I think you are right that sending multiple `SIGCONT`s is harmless.

xonsh/built_ins.py
```diff
@@ -905,6 +905,8 @@ def run_subproc(cmds, captured=False):
     if _should_set_title(captured=captured):
         # set title here to get currently executing command
         pause_call_resume(proc, builtins.__xonsh__.shell.settitle)
+    else:
+        pause_call_resume(proc, lambda : 0)
```
scopatz: A couple of small things:

1. Can you put some comments above this line that discuss why this case is here. I think it might be confusing otherwise.
2. can you replace the `lambda: 0` with `int`, please.  `int()` [with no arguments] returns zero and is already defined.

Thanks!

scopatz:
Also, I would love to do a release today or tomorrow, so if you could make the changes I suggest soon, that'd be awesome!

xonsh/built_ins.py
sagartewari01: @scopatz done

sagartewari01:
@scopatz I pushed the changes as soon as I could today. Hope it wasn't late.

scopatz:
Thanks so much!

PR #3006, opened 2019-02-01
Add `<details>` block to issue template

gforsyth:
Just noticed this on a conda-forge feedstock and it's nice to not have the issue
text all cluttered up with tracebacks and config outputs


scopatz:
Thanks!

PR #3007, opened 2019-02-01
Add SVG of xonsh sticker to docs/_static

gforsyth:
I put in a trial run of these to Stickermule and they came out great!
Putting it here so it's available for others, although I'm planning to
put in a big order so I can bring some to PyCon


scopatz:
Woohoo! 🎉 thanks Gil!!!

PR #3002, opened 2019-01-31
fixed completer issue

scopatz:
This closes #2380

PR #3010, opened 2019-02-04
Tweak `command_skipper` to avoid overwriting tokens with completer output

gforsyth:
Followup to #3002 -- in the event that the last token on a line (without a trailing space) is one of `&&`, `|`, `||`, `and` or `or`, completer will now insert a single space, allowing the next completer call to complete without issue.


scopatz:
Thanks @gforsyth!

PR #3017, opened 2019-02-09
Fix detection of Anaconda

melund:
This fixes a regression with `ON_ANACONDA` in `platform.py`. This can no longer detect if we are using an Anaconda based Python. The new method to detect conda is more robust. 

This fixes the environment not showing up in the prompt, and the `activate`/`deactivate` script which has stopped working on Windows.


scopatz:
Thanks @melund!

PR #3018, opened 2019-02-10
PLY update

melund:
Ply is no longer distributed as an installable package. David Beazley recommends that ply is purely vended from now on:
https://github.com/dabeaz/ply/blob/master/CHANGES#L3

This PR updates ply with a `git subtree pull` and removes the code that tries to import installed versions before the one we ship with xonsh. 

We could consider only including the ply/ply directory to reduce the size a bit (instead of the whole package with docs and tests etc.). David Beazley recommends  this approach. But I still think it is good practice to include the whole thing.


scopatz:
Thanks a million @melund!

melund:
Thanks, once we drop Python 3.5 support we could consider David Beazley's follow up to PLY. 

https://sly.readthedocs.io/en/latest/

It looks really interesting...

scopatz:
Yeah, I agree SLY looks cool and it would be great to use!

PR #3028, opened 2019-02-18
Fix problem with partial aliases not having __name__ attribute

melund:
No news items needed for this. 

This fixes a problem with Partial aliases not having a `__name__` attribute in the same location as the previous aliases. @scopatz I don't know if this is the correct fix. 

Without this I get the following error when running a simple aliases: 
```
snail@sea ~\Documents\xonsh
$ cdx
  File "C:\Users\mel\anaconda3\lib\site-packages\xonsh\proc.py", line 1856, in __init__
  File "C:\Users\mel\anaconda3\lib\site-packages\xonsh\built_ins.py", line 538, in run
    event_name = self._cmd_event_name()
  File "C:\Users\mel\anaconda3\lib\site-packages\xonsh\built_ins.py", line 612, in _cmd_event_name
    return self.alias.__name__
AttributeError: 'PartialEvalAlias2' object has no attribute '__name__'
```

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -208,6 +208,7 @@ class PartialEvalAliasBase:
         """
         self.f = f
         self.acc_args = acc_args
+        self.__name__ = self.__class__.__name__
 
     def __call__(
         self, args, stdin=None, stdout=None, stderr=None, spec=None, stack=None
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -208,7 +208,7 @@ class PartialEvalAliasBase:
         """
         self.f = f
         self.acc_args = acc_args
-        self.__name__ = self.__class__.__name__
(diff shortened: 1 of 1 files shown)


xonsh/aliases.py
```diff
@@ -208,6 +208,7 @@ def __init__(self, f, acc_args=()):
         """
         self.f = f
         self.acc_args = acc_args
+        self.__name__ = self.__class__.__name__
```
scopatz: Should this be `self.__name__ = getattr(f, "__name__", self.__class__.__name__)` instead? That is, should it take the function name?

scopatz:
Thanks @melund! One minor suggestion

melund:
Thanks, @scopatz. I didn't think of taking the name from the wrapped function. That is, of course, the correct fix.

scopatz:
Thanks @melund!

PR #3048, opened 2019-03-12
PTK colors refactor

### LINKED ISSUE #3047: Enable new color modifiers on PTK
See https://github.com/xonsh/xonsh/pull/3045#issuecomment-471631973

scopatz:
Fixes  #3047

### DIFF
--- a/xonsh/pyghooks.py
+++ b/xonsh/pyghooks.py
@@ -44,6 +44,7 @@ from xonsh.color_tools import (
     RE_XONSH_COLOR,
     make_palette,
     find_closest_color,
+    iscolor,
 )
 from xonsh.style_tools import norm_name
 from xonsh.lazyimps import terminal256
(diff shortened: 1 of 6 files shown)


xonsh/pyghooks.py
```diff
@@ -278,6 +280,63 @@ def color_by_name(name, fg=None, bg=None):
     return tok, fg, bg
 
 
+@lazyobject
+def PYGMENTS_MODIFIERS():
+    # pygments doesn't support all modifiers.
+    # use None to represent unsupported
```
melund: I know pygments doesn't  support all the modifiers but prompt toolkit support `reverse` and `blink`. So maybe there was a way to use them still.

xonsh/pyghooks.py
```diff
@@ -293,34 +352,22 @@ def code_by_name(name, styles):
     code : str
         Pygments style color code.
     """
-    fg, _, bg = name.lower().partition("__")
-    if fg.startswith("background_"):
+    fg, _, bg = name.upper().replace("HEX", "#").partition("__")
```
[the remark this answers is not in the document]
scopatz: Yeah, I don't know why this was implemented in this way. For this PR, I just tried to match what was there already.

xonsh/pyghooks.py
scopatz: Yeah, I think we might want to have a separate module for computing PTK 2 colors directly from color names.

scopatz:
Thanks for the review and merge @melund!

PR #3056, opened 2019-03-16
Implement more bang commands: !$ !* !^ !<cmd>

kalekundert:
While the 'bashisms' xontrib provides the `!!` command, several related commands are notably missing:
```
!$        # The last argument of the previous command.
!*        # All arguments but the first of the previous command.
!^        # The first argument of the previous command.
!<cmd>    # The last command beginning with the string "cmd".
```
This commit implements these commands in the same fashion as `!!`.  Note that this remains far from a complete implementation of the bash bang command syntax, but it covers the comands that I use on a daily basis and is hopefully a step in the right direction.

If there's interest, I could also implement more of the bash syntax, e.g.:
```
!!:2      # The second argument of the previous command.
!!:2-3    # The second and third arguments of the previous command.
!!:2-$    # The second through final arguments of the previous command.
!!:-2     # The first two arguments of the previous command.
!-2       # The command before last.
!-2:2     # The second argument of the command before last.
```

### DIFF
--- a/news/bang-commands.rst
+++ b/news/bang-commands.rst
@@ -1,0 +1,5 @@
+**Added:**
+
+* Implemented the following "bang command" bashisms: ``!$``, ``$*``, ``!^``, 
+  and ``!<str>``.  These are in addition to ``!!``, which was already 
+  implemented.
--- a/news/bang-commands.rst
+++ b/news/bang-commands.rst
@@ -3,3 +3,4 @@
 * Implemented the following "bang command" bashisms: ``!$``, ``$*``, ``!^``, 
   and ``!<str>``.  These are in addition to ``!!``, which was already 
   implemented.
+
--- a/tests/test_bashisms.py
+++ b/tests/test_bashisms.py
@@ -2,11 +2,43 @@
 import pytest
 
 
-@pytest.mark.parametrize("inp, exp", [("x = 42", "x = 42"), ("!!", "ls")])
-def test_preproc(inp, exp, xonsh_builtins):
+@pytest.mark.parametrize(
+        "history, inp, exp", [
+            # No history:
+            ([], '!!', ''),
+            ([], '!$', ''),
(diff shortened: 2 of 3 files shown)


gforsyth:
Hey @kalekundert -- thanks so much for putting this in!  I'm going to try to carve out some time this week to review it (unless someone else gets to it).  Could you also add a news item to the PR?  There's info on that here: https://xon.sh/devguide.html#changelog

I think adding in these commonly used bashisms is a good feature for the recent bash converts (or the not-so-recent ones who still want to use these shorthands).  I'm not sure we want to delve in to the other history commands because we start dipping in to a lot of the bash syntax we were trying to get away from in xonsh.  But generally opening PRs is a good way to promote a cause.

I'd be interested in thinking about what some more intuitive (to python) ways to handle some of the history parsing / grabbing patterns.   Maybe grab history items from `__xonsh__.history` using slicing syntax and then insert them into the current line?  Obviously we can already do this but it can be a bit ungainly typing out `__xonsh__.history[-1][:-1]`

Anyway, thanks again and I'll look this over soon (i hope!)

kalekundert:
I agree with your point that it'd be nice to have a more pythonic syntax for performing complicated dissections of past commands.  Maybe make `!` a shorthand for `reversed(__xonsh__.history)` (unless that `!` is part of `!!`, `!$`, etc).  To give some concrete examples:

- `![0]`: The current command.
- `![1]`: The previous command.
- `![1][-1]`: The last argument of the previous command.

I do think `!$`, `!*`, and `!<str>` are useful enough to merit to more terse syntax, because I (at least) use them all the time.  (To be honest I hardly ever use `!^`, but once I'd implemented the others it just felt wrong to leave it out.)

gforsyth:
And I'm definitely open to continued discussions around easier history access.  `![]` already has an assigned meaning in `xonsh` but I'm sure we can figure something out.

PR #3066, opened 2019-03-31
Fix cat empty file

virusbb001:
Fixed causing infinite loop when doing `cat` empty file 
and add test about `cat`.

Reproduce problem:

```sh
xonsh --no-rc
xontrib load coreutils
touch empty_file
cat empty_file
# Hang up. Maybe it needs to kill this process from other terminal.
```

Reproduce `xonfig`:
```
| xonsh            | 0.8.12.dev19         |
| Git SHA          | 5d87fc76             |
| Commit Date      | Mar 23 23:55:33 2019 |
| Python           | 3.7.3                |
| pygments         | 2.3.1                |
```

### DIFF
--- a/news/fix-cat-empty-file.rst
+++ b/news/fix-cat-empty-file.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* fix causing infinite loop when doing ``cat`` empty file
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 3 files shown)


virusbb001:
The reason that failed of [ci/circleci: build_34](https://circleci.com/gh/xonsh/xonsh/5409) seems `pytest-timeout` is old.(Confirmed passing when using `pytest-timeout-1.3.0`)
I will try to fixing this issue and works test on windows.

melund:
Thanks for tackling this @virusbb001. @scopatz last fixed some issues (#2967) with `cat` when files did not a have a size (like ``/dev/random``). So he would better review this one.

virusbb001:
Oh, sorry, I missed that case (/dev/urandom) . `cat /dev/urandom | head -n 10` didn't work in my patch. I will try to fix it.

virusbb001:
I think this PR is ready. Please review this. 🙏

PR #3079, opened 2019-04-18
Set PTY size to same size as main terminal for threadable commands

jb-leger:
Commands which are threadable are run with separated PTY. For some usage, PTY should have the same size as the command is not threadable. For example, before this branch:

```
jben@garance ~ $ __xonsh__.commands_cache.threadable_predictors['tput']=xonsh.commands_cache.predict_false # we want tput as NOT threadable                    
jben@garance ~ $ tput cols                                                                                                                                     
159
jben@garance ~ $ __xonsh__.commands_cache.threadable_predictors['tput']=xonsh.commands_cache.predict_true # we want tput threadable                            
jben@garance ~ $ tput cols                                                                                                                                     
80
```

With this branch:

```
jben@garance ~ $ __xonsh__.commands_cache.threadable_predictors['tput']=xonsh.commands_cache.predict_false # we want tput as NOT threadable                    
jben@garance ~ $ tput cols                                                                                                                                     
159
jben@garance ~ $ __xonsh__.commands_cache.threadable_predictors['tput']=xonsh.commands_cache.predict_true # we want tput threadable                            
jben@garance ~ $ tput cols                                                                                                                                     
159
```

Many command which have artifacts with threading are more tolerant with this behavior. But those commands still need to be added to exception in command_cache. For some case, the size is the only reason why the command is not threadable.

### DIFF
--- a/news/pty_size.rst
+++ b/news/pty_size.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* PTYs created for running threadable command have now size set to same size
+  than main terminal.
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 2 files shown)


melund:
This looks good to me. Thanks @jb-leger

PR #3080, opened 2019-04-18
Not threadable asciinema

jb-leger:
`asciinema` (terminal recorder tool) can have very strange behavior when threaded. Seems to be resolved when pty size is correctly specified (see #3079) but, I think it is safer to add this command in not threadable commands.


melund:
Thanks a lot @jb-leger

PR #3081, opened 2019-04-18
Not threadable tput

jb-leger:
Usually when a user play with `tput`, they want to interact with the main terminal, not a PTY created between the command and the terminal.


melund:
Thanks a lot @jb-leger

PR #3068, opened 2019-04-07
Add `-S` argument to `env` in scripts/xon.sh

carmenbianca:
I ran into #2881 while packaging this for Fedora. This is a super simple fix.

https://bugzilla.redhat.com/show_bug.cgi?id=1695139

Fixes #2881.

PR #3082, opened 2019-04-19
Tests usrbinenv args

jb-leger:
Arguments passed to python in `scripts/xonsh` and in `scripts/xonsh-cat` are now passed by a portable hack in sh, not anymore by `/usr/bin/env`, because `/usr/bin/env` can be not able to handle arguments.

Some tests in `test_integrations.py` was failing on my system, and are now passing.

Similar to #3068, but this PR concerns other scripts.

### DIFF
--- a/scripts/xonsh
+++ b/scripts/xonsh
@@ -1,3 +1,8 @@
-#!/usr/bin/env python3 -u
+#!/bin/sh
+'''#'
+exec python3 -u "$0" "$@"
+'''#'
+# Portable trick to pass option [-u] to python3, more portable than /usr/bin/env.
+
 from xonsh.main import main
 main()
--- a/scripts/xonsh-cat
+++ b/scripts/xonsh-cat
@@ -1,3 +1,8 @@
-#!/usr/bin/env python3 -u
+#!/bin/sh
+'''#'
+exec python3 -u "$0" "$@"
+'''#'
+# Portable trick to pass option [-u] to python3, more portable than /usr/bin/env.
+
 from xonsh.xoreutils.cat import cat_main as main
 main()
--- a/news/tests_usrbinenv_args.rst
+++ b/news/tests_usrbinenv_args.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Arguments passed to python in 'scripts/xonsh' and in 'scripts/xonsh-cat' are
+  now passed by a portable hack in sh, not anymore by /usr/bin/env.
+
+**Security:**
+
+* <news item>


melund:
@jb-leger This is a little out of my league since I primarily use Windows.  @gforsyth or @scopatz  better review this

scripts/xonsh
```diff
@@ -1,3 +1,9 @@
-#!/usr/bin/env python3 -u
+#!/bin/sh
+'''#'
+exec python3 -u "$0" "$@"
+'''#'
+# Portable trick to pass option [-u] to python3, more portable than /usr/bin/env.
+# vim: syn=python
```
gforsyth: I don't think we need the vim syntax hint in here

scripts/xonsh-cat
```diff
@@ -1,3 +1,9 @@
-#!/usr/bin/env python3 -u
+#!/bin/sh
+'''#'
+exec python3 -u "$0" "$@"
+'''#'
+# Portable trick to pass option [-u] to python3, more portable than /usr/bin/env.
+# vim: syn=python
```
gforsyth: I don't think we need the vim syntax hint in here

scripts/xonsh
jb-leger: I add this line by force of habit. It is not necessary.

scripts/xonsh-cat
jb-leger: removed.

jb-leger:
(I forgot to write this comment this morning).

Thanks @gforsyth for review. I removed these lines, I added only by force of habit, each time I do something which broke syntax recognition in vim I add this line without thinking.

carmenbianca:
I'm not sure this works quite right. I ran `python setup.py develop`, then ran the following:

```
$ scripts/xonsh                                                                                                                                                           
scripts/xonsh: line 2: #: command not found
```

A little confusingly, it appears to run correctly, but the error is shown every time.

Machine details:

```
$ inxi -b -z                                                                                                                                                              
System:    Host: carmenbianca-thinkpad Kernel: 5.0.7-300.fc30.x86_64 x86_64 bits: 64 Desktop: Gnome 3.32.1 
           Distro: Fedora release 30 (Thirty) 
Machine:   Type: Laptop System: LENOVO product: 20HDCTO1WW v: ThinkPad T470 serial: <filter> 
           Mobo: LENOVO model: 20HDCTO1WW serial: <filter> UEFI: LENOVO v: N1QET82W (1.57 ) date: 03/07/2019 
Battery:   ID-1: BAT0 charge: 0.9 Wh condition: 20.1/23.9 Wh (84%) 
           ID-2: BAT1 charge: 14.7 Wh condition: 66.8/71.1 Wh (94%) 
CPU:       Dual Core: Intel Core i5-7200U type: MT MCP speed: 800 MHz min/max: 400/3100 MHz 
Graphics:  Device-1: Intel HD Graphics 620 driver: i915 v: kernel 
           Display: wayland server: Fedora Project X.org 1.20.4 driver: i915 resolution: 1920x1080~60Hz 
           OpenGL: renderer: Mesa DRI Intel HD Graphics 620 (Kaby Lake GT2) v: 4.5 Mesa 19.0.2 
Network:   Device-1: Intel Ethernet I219-V driver: e1000e 
           Device-2: Intel Wireless 8265 / 8275 driver: iwlwifi 
Drives:    Local Storage: total: 238.47 GiB used: 73.31 GiB (30.7%) 
Info:      Processes: 315 Uptime: 3d 8h 14m Memory: 7.69 GiB used: 5.61 GiB (73.0%) Shell: python3 inxi: 3.0.32 

$ xonsh --version                                                                                                                                                         
xonsh/0.8.12
```

PR #3083, opened 2019-04-19
Tests git commit no gpg sign

jb-leger:
The command `gpg commit` is used in some tests. I have a `gitconfig` with `gpgsign = true`, therefore, `git` interact to `gpg` during tests, tests fails or skipped if I do not sign.

For testing, there is no reason to sign commits, so this PR add `--no-gpg-sign` to `git commit` commands.

Since the change is only on `tests/`, I think no news item is necessary. I have commited a empty news item to satisfy the tests on CI.


melund:
Thanks @jb-leger

PR #3075, opened 2019-04-11
CI: use `conda` directly instead of CondaEnvironment

virusbb001:
CondaEnvironment is deprecated.
Rewrite to follow this instruction:
https://docs.microsoft.com/en-us/azure/devops/pipelines/languages/anaconda?view=azure-devops&tabs=vs2017

This PR fixes causing ImportError in Azure Pipelines Python 3.7 Conda Environment ([like this](https://dev.azure.com/xonsh/xonsh/_build/results?buildId=565)) , but the results of some tests become unstable. (I checked some testcase causes `AttributeError: module 'builtins' has no attribute '__xonsh__'
`)

Issue #3070, opened 2019-04-08
test_main fails non-deterministically


carmenbianca:
## xonfig

<details>

```
$ xonfig
+------------------+----------------------+
| xonsh            | 0.8.12               |
| Git SHA          | 986bf28d             |
| Commit Date      | Mar 11 14:37:52 2019 |
| Python           | 3.7.3                |
| PLY              | 3.11                 |
| have readline    | True                 |
| prompt toolkit   | 2.0.9                |
| shell type       | prompt_toolkit2      |
| pygments         | 2.2.0                |
| on posix         | True                 |
| on linux         | True                 |
| distro           | fedora               |
| on darwin        | False                |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+
```

</details>

## Expected Behavior

The test_main tests run correctly.

## Current Behavior

Some test_main tests fail, sometimes. I run into this while building and testing xonsh for Fedora: https://bugzilla.redhat.com/show_bug.cgi?id=1695139

The tests don't always fail, and not always the same tests fail. But when the tests fail, they all have something in common: `AttributeError: module 'builtins' has no attribute '__xonsh__'`

### Traceback (if applicable)

[out.txt](https://github.com/xonsh/xonsh/files/3053525/out.txt)

Spec file added for context:

[xonsh.spec.txt](https://github.com/xonsh/xonsh/files/3053530/xonsh.spec.txt)

## Steps to Reproduce

Run the tests.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

scopatz:
Yes, this is defintiely true.  This is likely a bug in pytests and their weird monkey patching.  Additonally since xonsh relies on shoving things into builtints, there is some weird interplay.

scopatz:
Thanks for reporting @carmenbianca

jnoortheen:
I have done some [improvements to the test suit](https://github.com/xonsh/xonsh/pull/4581) . Now the test_main is stable, but the integration tests fail sometimes but have a retry decorator.

jnoortheen:
Feel free to reopen if the issue still appears.

PR #3084, opened 2019-04-19
Solve AttributeError in test_main and test_ptk_highlight

jb-leger:
I had same problems than in #3070. For my understanding, it is because objects are not properly cleaned between different tests. So I propose this patch.

I use a empty news item, because I think this PR doesn't not fix a bug for a user, but a bug in the tests (and the CI).

Should solve issue #3070 and complete PR #3075.


melund:
I think it looks reasonable. Let us see if it helps the weird test failures

PR #3085, opened 2019-04-20
[WIP] Tests and CI

jb-leger:
This PR is only a branch containing #3075, #3082, #3083, #3084. To check if this PR together, pass the tests on CI.


jb-leger:
It seems that those PR together solves the problem. I keep the [WIP], because I think that orignal PR will be merged one by one after review, so, this PR will not be merged, and I keep it only as a demo that tests works.

jb-leger:
not necessary anymore. With PRs the CI success.

PR #3087, opened 2019-04-23
.hg lookup, OSError, reduce the Try/Except to the minimum

jb-leger:
Following the comment on #3061, I reduce the Try/Except to the minimum. I tried to reopen the PR, but it seems not possible, so I open a new one.


melund:
Thanks, @jb-leger :)

PR #3093, opened 2019-04-26
fix E126 error in xontrib/bashisms.py

virusbb001:
I don't know why Flake 8 check failure caused in CI although nobody touched xontrib/bashisms.py...

I hope this patch passes flake8 check again.

PR #3094, opened 2019-04-26
Black xontribs

gforsyth:


gforsyth:
This should supercede #3093 -- I realize that we weren't running `black` checks on the `xontribs` directory which eventually bit us when `flake8` complained.

Added it to the checks and also ran `black` against all the bundled `xontribs`

melund:
Looks good

PR #3098, opened 2019-04-27
The black check no longer skips files buried deeper in the directory tree

carmenbianca:
This is a really trivial fix. I changed black to run on directories instead of `*.py`.

I also ran black once on all the files that had previously been skipped over. For security, please verify that I did not inject any malicious code.


gforsyth:
Thanks for putting this in @carmenbianca !  Looks good to me!  And much simpler than my hacky solution.

Merging in -- appveyor failures are spurious.

PR #3102, opened 2019-04-29
Add process to avoid overwrite files that specified in options in `run-tests.xsh`

virusbb001:
## Problem

Executing `run-tests.xsh` with setting files in options cause unwanted overwriting such like `junit/test-results.xml`.

## Solutions

Add processing `%d` for avoid overwriting in `run-tests.xsh`.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -36,7 +36,7 @@ jobs:
   - script: |
       call activate conda-test-$(python.version)
       pip install .
-      xonsh run-tests.xsh --timeout=10 --junitxml=junit/test-results.xml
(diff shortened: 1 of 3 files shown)


gforsyth:
Hey @virusbb001 -- thanks for putting this in!  Always nice to not overwrite our test results :smile:

I'm not certain if this is working as expected -- I see a couple lines in the Azure logs that have this:

```
D:\a\1\s\junit\test-results.d.xml
```
which looks like the `%d` is just being translated to `d` instead of an integer.

virusbb001:
Thanks for checking this! I  forgot to add escape percent sign...
This fix works.

gforsyth:
Thanks again for putting this in @virusbb001 !  And sorry for the delay in getting to it!

Issue #3103, opened 2019-04-29
`env xonsh` from within Xonsh directly backgrounds certain graphical apps


sevens-ef:
## xonfig

<details>

```
$ xonfig
+------------------+----------------------+
| xonsh            | 0.8.12               |
| Git SHA          | 986bf28d             |
| Commit Date      | Mar 11 14:37:52 2019 |
| Python           | 3.6.7                |
| PLY              | 3.11                 |
| have readline    | True                 |
| prompt toolkit   | 2.0.9                |
| shell type       | prompt_toolkit2      |
| pygments         | 2.3.1                |
| on posix         | True                 |
| on linux         | True                 |
| distro           | ubuntu               |
| on darwin        | False                |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+
```
</details>

Though I've seen the same issue in earlier Xonsh versions (at least in 0.8.x) and with earlier Python 3.6 versions (and on different distros as well).

## Expected Behavior
A nested Xonsh using `env xonsh` to behave the same as a regular Xonsh session.

## Current Behavior
When starting a nested Xonsh via `env xonsh`, many programs that try to draw to the screen are immediately sent to the background (no drawings were ever visible, `^Z` is printed to the shell and a new prompt is shown); e.g. running `man`, `vim` or `less` (though at least `tmux` and `screen` seem fine).  Running the same commands in a `env bash` launched shell works fine.

Starting a nested Xonsh via plain `xonsh` also works fine.

## Steps to Reproduce

```
xonsh --no-rc
# Works fine
man man

env xonsh --no-rc
# Prints `^Z`, returns to prompt, shows `man man` in `jobs`
man man
```

For completeness: starting `env xonsh` from a Bash session works fine, and starting a nested Xonsh as `env -u PAGER -u MANPAGER xonsh --no-rc` seems to make no difference.

I can fix the behavior by marking `env` as not threadable in my `.xonshrc`.

May be related to #2759 (at least the same workaround fixes both issues).

jb-leger:
Should be fixed by PR #3106. @sevens-ef can you confirm that it is fixed now?

sevens-ef:
[...]

PR #3106, opened 2019-04-29
Command cache env pred

### LINKED ISSUE #2759: env /bin/sh results in "sh: cannot set terminal process group (-1): Inappropriate ioctl for device
When using emacs TRAMP remote shell functionality, it connects over `ssh` in the following way, in an attempt to make it easier to find and parse the shell:

`ssh <server> "env 'TERM=dumb' 'PROMPT_COMMAND=' 'PS1=#$ ' /bin/sh"`

When using xonsh, this unfortunately errors because sh complains about being unable to set the terminal process group.

I did some digging, and it seems to come from `tcgetpgrp(shell_tty)` returning ENOTTY [here](https://github.com/bminor/bash/blob/64447609994bfddeef1061948022c074093e9a9f/jobs.c#L4201), with shell_tty = stderr. This is explained as:

```
ENOTTY
The calling process does not have a controlling terminal, or it has one but it is not described by fd
```

So I made a small program to reproduce this:

```c
#include <unistd.h>
#include <stdio.h>
#include <errno.h>
int main() {
    pid_t = tcgetpgrp(2);
    printf("tcgetpgrp(2) = %d, errno = %d\n", p, errno);
    return 0;
}
```

- sh results: `tcgetpgrp(2) = 10837, errno = 0`
- xonsh results: `tcgetpgrp(2) = -1, errno = 25`

There is clearly a gap in behavior between sh and xonsh here. I'm unsure where to go from here. I believe xonsh is missing a `tcsetpgrp` call somewhere, but I'm not quite sure where it needs to be added. I notice bash calls it in `initialize_job_control`, in addition to `give_terminal_to`. Maybe this is what xonsh is missing?

## xonfig
```
| xonsh            | 0.6.0           |
| Python           | 3.6.6           |
```

## Expected Behavior

I expect sh to not complain about the failing ioctl. 

## Current Behavior

sh complains about the failing ioctl.

## Steps to Reproduce

From xonsh: `env sh`. Note that running the same command from another shell, such as sh, will not produce an error in sh.

jb-leger:
For `env`, a special predictor based on the predictor for the executed command.

Fixes #2759 and #3103.

### DIFF
--- a/news/command_cache_env_pred.rst
+++ b/news/command_cache_env_pred.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* threadable predictor for 'env' command based on predictor from the executed
+  command. Fixes #2759 and #3103.
+
+**Security:**
+
+* <news item>
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -380,6 +380,15 @@ def predict_hg(args):
         return not ns.interactive
 
 
+def predict_env(args):
+    """Predict if env is launching a threadable command or not"""
+
+    for i in range(len(args)):
+        if args[i] and args[i][0] != "-" and "=" not in args[i]:
+            return builtins.__xonsh__.commands_cache.predict_threadable(args[i:])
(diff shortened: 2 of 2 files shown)


jb-leger:
I have added comments. I do not know if the logic is more clear, but I tried.

gforsyth:
Thanks so much @jb-leger !  And sorry for the delay!

Issue #1783, opened 2016-09-30
Prefix environment variable and temporarily disable alias


ywangd:
I noticed that following two bash usages are not covered in xonsh. 
- Prefixing a command with temporary environment variables, e.g. `DEBUG=1 make`. I do realise that the same effect can be achieved by using environment `swap(...)`. 
- Temporarily disable an alias. For an example, if I have `alias ls=ls -a`, I can use `\ls` to temporarily ignore the alias and call to the actual `ls` command. 

As I understand it, xonsh is not trying to replicate every Bash feature/usage. So my question is: Are above usages out of scope for xonsh? I am trying to avoid effort on something if it is not aligned with xonsh's overall direction. Thanks!

astronouth7303:
I didn't know about `swap()`. More visibility on "user-facing" functions would be good.

ywangd:
I personally think `swap()` is more clunky to use. 

It also does not work exactly the same as prefixing env variable. But this could be an advantage because Bash's behaviour is a bit confusing.`swap` does inline substitution, such as the follows
```
with ${...}.swap(SOMEVAR='foo'):
     echo $SOMEVAR
```
While `SOMEVAR=foo echo $SOMEVAR` does not work.

scopatz:
Hi @ywangd @astronouth7303, for what it is worth, you can use the `env` command instead of swap: `env SOMEVAR=foo echo $SOMEVAR`.  `swap()` is really for many subprocess statements, though it can be used in a small context, too.

The guiding principle has been to not have more syntax than necessary.  For the env/swap issue, `env` gets the job done with out additional syntax and is explicit rather than implicit.  Personally, I am not opposed to adding implicit envvar setting, but I am not sure that I want to implement it myself. A PR would be welcome.  Also, we'd need to decide on the semantics for how long the envvars apply.  That is, do they apply for just a single command, or all commands in pipeline? My preference is  the former.

As per alias disabling, we probably should have something like swap() on the aliases object.  Custom syntax for this, though, is probably not a good idea.  Like Bash's ability to disable history by starting a command with a space, I personally feel that it is this sort of arcana that makes Bash hard to learn, hard to use, and hard to explain. If you want to avoid an alias, it is best to be explicit about this.  This can be done *either* by using the absolute path to the executable (`/bin/ls`) or by using the `which -s` command, which avoids aliases. For example `@$(which -s ls)` will run ls without an alias.

[...]

PR #3107, opened 2019-04-30
Do not attempt to expand raw strings in subprocess mode

### LINKED ISSUE #2943: `xonsh.__env__` and `env` subproc command precedence issue
If you use `env` to set a temporary env var for a one-line command, that works as expected
```
 ~ 🐚  env BAI="HAI" bash -c "echo $BAI"
"HAI"
```
and `$BAI` isn't available in the current session after the command completes
```
 ~ 🐚  $BAI
KeyError: 'Unknown environment variable: $BAI'
```

But if `$BAI` is set already and then `env` is used, the value in `__xonsh__.env` overrides the temporary value:
```
 ~ 🐚  $BAI="BYE"
 ~ 🐚  env BAI="HAI" bash -c "echo $BAI"
BYE
```

## xonfig

```
| Python           | 3.6.5               |
| prompt toolkit   | 2.0.6               |
```
## Expected Behavior

```
 ~ 🐚  $BAI="BYE"
 ~ 🐚  env BAI="HAI" bash -c "echo $BAI"
HAI
```

laloch:
Closes #2943 
Previously, there was no way to prevent Xonsh from expanding environment variables in strings:
```
~ 🐚  $BAI="BYE"
~ 🐚  env BAI=HAI bash -c "echo $BAI"
BYE
~ 🐚  env BAI=HAI bash -c r"echo $BAI"
BYE
```
Now:
```
~ 🐚  $BAI="BYE"
~ 🐚  env BAI=HAI bash -c "echo $BAI"
BYE
~ 🐚  env BAI=HAI bash -c r"echo $BAI"
HAI
```
Edit: relates also to #1783

### DIFF
--- a/news/noexpand-raw-string.rst
+++ b/news/noexpand-raw-string.rst
@@ -1,0 +1,28 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Xonsh now does not attempt to expand raw strings, so now::
+
+    $ echo "$HOME"
+    /home/user
+    $ echo r"$HOME"
+    $HOME
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -7,7 +7,7 @@ import itertools
 
 import pytest
 
-from xonsh.ast import AST, With, Pass, pdump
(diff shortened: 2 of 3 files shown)


laloch:
Yep, no problem, but ENOTIME right now. I'll try to get it done soonish.

laloch:
The test turns out to be harder to implement than it sounds. No matter what I put into `xonsh_builtins.__xonsh__.env`, I can't make the parser (via `check_xonsh_ast`) to expand any variables whatsoever.

PR #3112, opened 2019-05-06
Added emphasis to the error source

Ad115:
Final touches to close issue `#1564`.


scopatz:
We have dropped 3.4 support.  Thanks so much @Ad115!

PR #3113, opened 2019-05-06
Drop support for Python 3.4

### LINKED ISSUE #3110: Drop support for Python 3.4 because it is now end of life?
Given that [Python 3.4 is now end of life](https://devguide.python.org/devcycle/#end-of-life-branches) would it be appropriate to drop testing and support?

rodrigogolive:
Minimal required Python version is now 3.5.
Also, closes #3110.


scopatz:
Woohoo! Thanks so much @rodrigogolive!

Issue #3058, opened 2019-03-22
AttributeError: 'functools.partial' object has no attribute '__name__' (aliases)


jaraco:
## xonfig

<details>

```
$ xonfig
+------------------+----------------------+
| xonsh            | 0.8.11               |
| Git SHA          | fe5c016b             |
| Commit Date      | Feb 27 11:19:35 2019 |
| Python           | 3.7.2                |
| PLY              | 3.11                 |
| have readline    | True                 |
| prompt toolkit   | 2.0.7                |
| shell type       | prompt_toolkit2      |
| pygments         | None                 |
| on posix         | True                 |
| on linux         | False                |
| on darwin        | True                 |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+
```
</details>

## Expected Behavior
Aliases should accept `functools.partial` objects (or other callables) as they did in 0.8.10.

## Current Behavior
After upgrading to Xonsh 0.8.11 (but also in 0.8.12), my aliases no longer work. Here's how I define some aliases:

```
def add_mfa(alias, system, user):
	aliases[alias] = functools.partial(get_oath, system, user)

add_mfa('dropbox-mfa', 'Dropbox MFA', 'jaraco@jaraco.com')
add_mfa('microsoft-mfa', 'Microsoft MFA', 'jaraco@jaraco.com')
```

Running an alias emits an error:

```
~ $ dropbox-mfa                                                                                                                                                                                          
AttributeError: 'functools.partial' object has no attribute '__name__'
```

Setting `XONSH_DEBUG=1` prior to running xonsh has no effect (same output).

Downgrading to xonsh 0.8.10 works around the issue.

## Steps to Reproduce

```
~ $ def simple(): 
...     print("hello!") 
...                                                                                                                                                                                                       
~ $ aliases['simple'] = functools.partial(simple)                                                                                                                                                         
~ $ del simple                                                                                                                                                                                            
~ $ simple                                                                                                                                                                                                
AttributeError: 'functools.partial' object has no attribute '__name__'
```

Note, I tried to reproduce the issue with `print` but encountered a different, seemingly-related error:

```
~ $ aliases['foo'] = functools.partial(print)                                                                                                                                                             
~ $ foo                                                                                                                                                                                                   
[...]

PR #3114, opened 2019-05-07
Fixed problem with aliases composed of functions wrapped with functools.partial

Ad115:
This solves part of issue #3058 as adressed by @jaraco.


Ad115:
This only solves part of the problem. The issue also has to do with the problem found when executing
```
import functools
aliases['print_my_name'] = functools.partial(print, 'my name')
```

scopatz:
Thanks @Ad115!

PR #3124, opened 2019-05-10
python_requires

scopatz:
@gaborbernat - is this the kind of thing that you are looking for?  Happy to do a release after this gets in.  Also, for the python v3.4 tests, is it possible for us to just pin xonsh to `=0.8.12`?  I don't think it makes sense for the 0.9 series to support Python 3.4 anymore.

### DIFF
--- a/news/pyreq.rst
+++ b/news/pyreq.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* We no longer manually check the Python version in ``setup.py``,
+  but instead use the setuptools ``python_requires`` feature.
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/setup.py
+++ b/setup.py
@@ -325,8 +325,6 @@ if HAVE_SETUPTOOLS:
 
 def main():
     """The main entry point."""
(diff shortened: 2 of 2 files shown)


gaborbernat:
Well this is what will work going ahead, but will no longer help people on Python 3.4 so on its own insufficient (without a release having this that's compatible with python 3.4). Thanks for making the pr though.

melund:
Looks good.

gaborbernat:
Note this will only be helpful for python3.5 down the line, not for python3.4 users who'll be left needing to manually pin to 0.8.12.

scopatz:
I could do a revert and add the Python 3.4 to Python requires and push it out as xonsh 0.8.13 if that would help.

gaborbernat:
That will not help as pip will pick latest package that has a matching Python requires. And 0.9.0 did not said it does not support 3.4. Maybe delete 0.9.0 and release this as 0.9.1.

scopatz:
OK, we just released a 0.9.1 too.  So would it be enough to release a 0.9.2 that reverts this and then release a 0.9.3 that undoes the revert? Happy to do that if it helps, even if it is ugly :wink:

gaborbernat:
that would work too, for python 3.4 people. In virtualenv I've disabled xonosh on Python 3.4 to fix the ci. But that doesn't help users stuck on it.

scopatz:
Done! Sorry about this pain point and thanks for helping me work through a solution

Issue #3125, opened 2019-05-13
xonsh 0.9.0 can't start as default shell on Mac


mqualizz:
The "portable trick passing args" requires that python3 is on the path, which means that it needs to be in a path listed in /etc/paths if it's to be used as the default shell.  This breaks if python3 is installed, for example, via homebrew.

Example:
Create an iTerm2 profile that uses xonsh as the shell.  Then try to use that profile.

jb-leger:
Sorry for that, it's my patch.

Anyway, we have a dilemma:

 - Passing args with shebang using `/usr/bin/env` as
   ```
   #!/usr/bin/env python3 -u
   ```
   Some version of `env` try to find `python3 -u`, not `python3` with arg `-u`.
 - Passing args with sheebang using `/usr/bin/env` and `-S`
   ```
   #!/usr/bin/env -S python3 -u
   ```
   `-S` is not supported by all versions of `/usr/bin/env`
 - Passing args with portable trick:
   ```
   !/bin/sh
   ''''exec python3 -u "$0" "$@"
   '''#'
   ```
   I do not understand why `env` can find `python3` when it is not in `PATH`
 - Stop passing `-u`, and do very dirty thing in python to open stdout and stderr without buffering.

Maybe there is a alternative. I saw in the past some strange behaviors with `env`. Could you test this alternative:
   ```
   !//usr/bin/env sh
   ''''exec python3 -u "$0" "$@"
   '''#'
   ```

I have the hope that `env` will start `sh` with the good `PATH`. I do not use a system without a system side `python3`.

jb-leger:
I did this in [this branch](https://github.com/jb-leger/xonsh/tree/tests_usrbinenv_args_tryfixbug_3125), if you can test this option, it will be helpfull.

scopatz:
Yeah I hav noticed that the portability trick has caused other problems, such as using my system Python even though I have conda installed and on my PATH.  Also it seems weird to have to use `sh` to start xonsh, which is essentially a Python script.  Also, this is why we have the (little known) `xon.sh` script, for when we need to start xonsh from a `sh` script

I agree that the buffering thing is pretty serious. I think it might be better for folks who need to ensure the unbuffering behavior to rely on `xon.sh` for the moment, and revert the very impressive portability trick.

Thanks both @mqualizz & @jb-leger!

jb-leger:
I agree, the better idea is to revert the commit. Only some tests fail, but no use case is broken as far I know.

[...]

PR #3128, opened 2019-05-15
Revert portable env args trick

jb-leger:
Revert portable trick to pass args. See #3125.


scopatz:
This looks good @jb-leger! Thanks for putting it in!  Also, since you are now a member of @xore, please feel free to push branches up to this repo rather than your fork

PR #3143, opened 2019-05-29
aliases can be any sequence, not only list

jb-leger:
When I wrote threadabilty predictor for aliases in #3086, I was considering "normal" aliases as a `list`. Today I add a aliases with a tuple, I works perfectly with xonsh, but the predictor was returning the default prediction.

This is a very minor change, I think this does not desserve a news item.

### DIFF
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -274,7 +274,9 @@ class CommandsCache(cabc.Mapping):
         alss = getattr(builtins, "aliases", dict())
         while cmd0 in alss:
             alias_name = alss[cmd0]
-            if not isinstance(alias_name, list):
+            if isinstance(alias_name, (str, bytes)) or not isinstance(
+                alias_name, cabc.Sequence
+            ):
                 return predict_true
             for arg in alias_name[:0:-1]:
                 first_args.insert(0, arg)


xonsh/commands_cache.py
```diff
@@ -274,7 +274,7 @@ def default_predictor_alias(self, cmd0):
         alss = getattr(builtins, "aliases", dict())
         while cmd0 in alss:
             alias_name = alss[cmd0]
-            if not isinstance(alias_name, list):
+            if not isinstance(alias_name, cabc.Sequence):
```
scopatz: should this be `if isinstance(alias_name, (str, bytes)) or not isinstance(alias_name, cabc.Sequence)`?

scopatz:
Thanks for putting this in @jb-leger!

xonsh/commands_cache.py
jb-leger: Yes, I were right. I changed my commit.

scopatz:
This is weird. The setup for those two tests consistently fails on windows on Python 3.7.  Maybe it is time to move test_main to a separate testing instance.

scopatz:
Thanks @jb-leger!

PR #3144, opened 2019-05-30
Fix `pip` completer misfires on `bagpipes` and similar

### LINKED ISSUE #3142: Pip completer sneaks into cd when directory name contains 'pip'
## Expected Behavior
When triggering autocompletion in a `cd` command to enter a subdirectory of a directory containing 'pip', it should complete the subdirectory name. 

## Current Behavior
Instead, I get suggestions for `pip` subcommands like freeze, config, download, etc.

## Steps to Reproduce
Starting from a clean directory and an empty .xonshrc:
```
mkdir bagpipes
mkdir bagpipes/subdir
cd bagpipes/su[press tab here]
```
When tab is pressed in the above command, xonsh will not complete 'su' to 'subdir', but instead suggest pip subdommands.

## Workaround
Transposing the completers for `cd` and `pip` in the completer list fixes this specific issue, but it would not fix the similar issue of trying to complete another bash command, such as `ls bagpipes/su[press tab here]`. I haven't tried but assume that placing the pip completer after the bash completer would fix that too, although the proper fix seems to be making the pip completer not trigger when it shouldn't.

(For newbies like myself, you can transpose the completers by:

 ```
pip_completer = __xonsh__.completers['pip']
completer remove pip
completer add pip pip_completer ">cd"

gforsyth:
Requires a word-ending before `pip`, also adding in support for `xpip`

Resolves #3142

### DIFF
--- a/news/fix_pip_completer.rst
+++ b/news/fix_pip_completer.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* ``pip`` completer no longer fires when ``pip`` happens to appear within a word
+  like ``bagpipes``
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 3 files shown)


gforsyth:
Well, I lied about the `xpip` part, apparently.

scopatz:
Thanks a million @gforsyth!

PR #3145, opened 2019-05-31
test main to run seprately

scopatz:
no news needed

### DIFF
--- a/run-tests.xsh
+++ b/run-tests.xsh
@@ -19,7 +19,10 @@ run_separately = [
     'tests/test_main.py',
     'tests/test_ptk_highlight.py',
     ]
-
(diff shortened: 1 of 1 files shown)


scopatz:
@xonsh/xore - it looks like this works!

jb-leger:
It seems in `xonsh.xonsh` on azure that only separately tests are run, not the others. Or I can find where there are run.

scopatz:
Ahh excellent point @jb-leger! Can't believe that I missed that

scopatz:
Alright, this is fixed now

melund:
Great that is fixed. Well spotted @jb-leger

PR #3150, opened 2019-06-04
remove matplotlib from 3.5 deps

scopatz:
Just a CI fix, no news needed


laloch:
Great! Thanks for the fix.

Issue #2817, opened 2018-09-17
Gitstatus doesn't report deleted files


shadow-light:
## xonfig
```
+------------------+---------------------+
| xonsh            | 0.7.3               |
| Git SHA          | 9bbbf07b            |
| Commit Date      | Aug 2 18:08:44 2018 |
| Python           | 3.6.6               |
| PLY              | 3.9                 |
| have readline    | True                |
| prompt toolkit   | None                |
| shell type       | readline            |
| pygments         | None                |
| on posix         | True                |
| on linux         | True                |
| distro           | Ubuntu              |
| on darwin        | False               |
| on windows       | False               |
| on cygwin        | False               |
| on msys2         | False               |
| is superuser     | False               |
| default encoding | utf-8               |
| xonsh encoding   | utf-8               |
| encoding errors  | surrogateescape     |
+------------------+---------------------+

```

## Expected Behavior
Using `{gitstatus}` in `$PROMPT` should report deleted files that were previously commited.

## Current Behavior
It reports all changes to files and new files etc _except_ for deleted files.

```shell
~/path master|~1+1 git status --porcelain
 D filea
 M fileb
?? filec
~/path master|~1+1 

```

## Steps to Reproduce
Delete a previously committed file.

shadow-light:
I did a quick patch for my own use. Let me know if you want a pull request
https://github.com/shadow-light/xonsh/commit/95d393331dd823ecb6e6eeea4aecfe2011d52039

gforsyth:
Hey @shadow-light, great! Yes, please open a PR for this, we'd love to have it in!

scopatz:
closed by #3151

PR #3151, opened 2019-06-05
Patch gitstatus delete

shadow-light:
Fixes gitstatus prompt so that it also reports deleted files

Finally got round to submitting this #2817


scopatz:
Awesome! Thanks @shadow-light!

PR #3152, opened 2019-06-05
Change default environment value of document.

virusbb001:
+ Show default value instead of `<function function_name at FUNC_ID_HEX>`.
+ Show default PATH instead of the PATH which set when building document.

This is just document update so I didn't add news file.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -977,7 +977,32 @@ def DEFAULT_DOCS():
         "LANG": VarDocs("Fallback locale setting for systems where it matters"),
         "LS_COLORS": VarDocs(
             "Color settings for ``ls`` command line utility",
-            default="``" + LsColors(LsColors.default_settings).detype() + "``",
+            default="``*.7z=1;0;31:*.Z=1;0;31:*.aac=0;36:*.ace=1;0;31:"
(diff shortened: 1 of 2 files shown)


xonsh/environ.py
```diff
@@ -975,7 +975,10 @@ def DEFAULT_DOCS():
             configurable=ON_WINDOWS,
         ),
         "LANG": VarDocs("Fallback locale setting for systems where it matters"),
-        "LS_COLORS": VarDocs("Color settings for ``ls`` command line utility"),
+        "LS_COLORS": VarDocs(
+            "Color settings for ``ls`` command line utility",
+            default="``" + LsColors(LsColors.default_settings).detype() + "``",
```
scopatz: It isn't great that this will execute on import, effectively.  Is there any way this could just be a static string?

scopatz:
Thanks for putting this in @virusbb001!

xonsh/environ.py
virusbb001: Thanks for reviewing this!
The reason I used this method is to avoid peeling between documents and the real default value and make it easy to maintenance documents. But I agree that executing on import is not good. I will try to fix this.

scopatz:
Thanks @virusbb001!

Issue #2755, opened 2018-08-02
Captured subprocess syntax issue


laloch:
I do quite often find myself in need of running something like
```
gdb --pid=$(pidof krunner)                                                                                                                                                                                                                                      
....
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/xonsh/ptk/shell.py", line 138, in _push
    locs=None)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 20084, in compile
    transform=transform)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 20053, in parse
    tree, input = self._parse_ctx_free(input, mode=mode, filename=filename)
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 20153, in _parse_ctx_free
    raise original_error from None
  File "/usr/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 20140, in _parse_ctx_free
    debug_level=(self.debug_level > 2))
  File "/usr/lib/python3.6/site-packages/xonsh/parsers/base.py", line 351, in parse
    check_contexts(tree)
  File "/usr/lib/python3.6/site-packages/xonsh/parsers/context_check.py", line 57, in check_contexts
    raise e
  File "<string>", line None
SyntaxError: can't assign to operator
```
In the above example the captured subprocess syntax `$(<expr>)` didn't get parsed well and I can't find a way around the issue other than `gdb @('--pid=' + $(pidof krunner).rstrip())`, which is, well, inconvenient.
Also, I feel that the `$()` operator should automagically .rstrip() the resulting string like, for instance, Bash does.
There's already issue #1453 opened regarding similar behavior of the `@()` operator. Feel free to mark this as duplicate of the #1453 if they're related.
## xonfig
```
+------------------+-----------------+
| xonsh            | 0.7.2           |
| Python           | 3.6.6           |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | 1.0.15          |
| shell type       | prompt_toolkit1 |
| pygments         | 2.2.0           |
| on posix         | True            |
| on linux         | True            |
| distro           | arch            |
| on darwin        | False           |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
[...]

PR #3156, opened 2019-06-06
Fix @$ subprocess operator

### LINKED ISSUE #2328: Syntax for *just* stripping last newline
`@$` is nice, but, oftentimes, you need everything to stay together as *one* argument, in which case the only way to do it is `@($(mycommand).rstrip('\n'))`. (Or at least for me, I often want it that way!) It'd be nice if there were a syntax akin to `@$` that just strips the ending newline, something that would be like bash's `$(my-command)` syntax.

laloch:
`@$(cmd)` operator now returns a list of tokens created by splitting the command's output at the newline character.

Closes #2328

### DIFF
--- a/news/fix-captured-inject.rst
+++ b/news/fix-captured-inject.rst
@@ -1,0 +1,28 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* ``@$()`` subprocess operator now properly strips newline characters off
+  the lines of multiline output.
+
+* ``@$()`` subprocess operator does not require leading and trailing whitespace
+  anymore, so expansions like ``cd /lib/modules/@$(uname -r)/kernel`` or
+  ``gdb --pid=@$(pidof crashme)`` are now possible.
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -312,6 +312,34 @@ echo @$(which ls)
         0,
     ),
     #
+    # test @$() without leading/trailig WS
+    #
+    (
+        """
+def _echo(args):
+    print(' '.join(args))
+aliases['echo'] = _echo
+
+echo foo_@$(echo spam)_bar
+""",
+        "foo_spam_bar\n",
+        0,
+    ),
+    #
+    # test @$() outer product
+    #
+    (
+        """
+def _echo(args):
+    print(' '.join(args))
+aliases['echo'] = _echo
+
+echo foo_@$(echo spam sausage)_bar
+""",
+        "foo_spam_bar foo_sausage_bar\n",
+        0,
+    ),
+    #
     # test redirection
     #
     (
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -2250,6 +2250,13 @@ def test_rhs_nested_injection():
     check_xonsh_ast({}, "$[ls @$(dirname @$(which python))]", False)
 
 
+def test_merged_injection():
+    tree = check_xonsh_ast({}, "![a@$(echo 1 2)b]", False, return_obs=True)
+    assert isinstance(tree, AST)
+    func = tree.body.args[0].right.func
+    assert func.attr == "list_of_list_of_strs_outer_product"
+
+
 def test_backtick_octothorpe():
     check_xonsh_ast({}, "print(`#.*`)", False)
 
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -992,8 +992,12 @@ def subproc_captured_inject(*cmds):
     The string is split using xonsh's lexer, rather than Python's str.split()
     or shlex.split().
     """
-    s = run_subproc(cmds, captured="stdout")
-    toks = builtins.__xonsh__.execer.parser.lexer.split(s.strip())
+    o = run_subproc(cmds, captured="object")
+    o.end()
+    toks = []
+    for line in o:
+        line = line.rstrip(os.linesep)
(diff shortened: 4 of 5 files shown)


laloch:
The implementation is currently wrong. It breaks the `@$(which ls)` example from the tutorial and it is not consistent with the bash's behavior either.

scopatz:
Thanks for pushing on this

laloch:
OK, the `@$(cmd)` operator's behavior is now consistent with `$(cmd)` in bash.
It still can't do basic `ls /lib/modules/@$(uname -r)/kernel` though. Parser adjustment required...

laloch:
The @$ operator does not require the leading white space character any more. Now I can do
```
$ echo %%%%@$(which ls)
%%%% ls --color=auto -v --group-directories-first
```
There is, however still an extra space between the literal argument and the output of the subprocess. @scopatz, do you have a clue about where this extra space comes from?

scopatz:
Maybe it is coming from https://github.com/xonsh/xonsh/blob/master/xonsh/built_ins.py#L975?

laloch:
@scopatz, thanks for the suggestion, but no, the `toks` list does not contain the spaces.

laloch:
There is something that's ensuring the `subproc_arg` tokens to get space-separated.
```
$ echo a@$(which cp)@$(which cp)
a /usr/bin/cp /usr/bin/cp
```

scopatz:
Yeah, I wonder if the lexer itself is adding this token (WS)?

scopatz:
Yeah, maybe it is this line: https://github.com/xonsh/xonsh/blob/master/xonsh/lexer.py#L349

laloch:
So, I'm back to this PR after several months.
I still can't understand where the surrounding space characters come from. The `@$` operator is translated internally by the parser to a call to `__xonsh__.subproc_captured_inject` which by itself does not return any extra spaces with this patchset applied. Yet, as can be seen below, the `@$` operator yields both extra leading and trailing space.
```
$ echo _@(1)_
_1_
$ echo _@($(pidof init).rstrip())_
_1_
$ echo _@$(pidof init)_
_ 1 _
$ __xonsh__.subproc_captured_inject(["pidof", "init"])
['1']
$ echo _@(__xonsh__.subproc_captured_inject(["pidof", "init"]))_
_1_
```

laloch:
Never mind. I know what the problem is - it's that I'm parsing the `@$` operator without leading whitespace as a separate subprocess argument, which is obviously wrong. I need to concatenate the resulting expression with the previous argument instead.

laloch:
This is now ready to land.
The `@$()` operator now behaves accordingly to the [tutorial](https://xon.sh/tutorial.html#command-substitution-with) and consistently with Bash `$()` operator. It doesn't require leading and trailing whitespace anymore, so `cd /lib/modules/@$(uname -r)/kernel` and similar are now possible.

laloch:
Closes #2755 
Closes #3457

melund:
@laloch Any idea some test fail. Are they related to this?

laloch:
> @laloch Any idea some test fail. Are they related to this?

Yeah, I would like to know too, but I don't think they might be related.

laloch:
Please do not merge. This is wrong once again. It breaks `@()` outer product expansion.

laloch:
Now the `@$()` operator behaves like `@()` including the outer product expansion, so now `_@$(ls)_` yields
```
$ cd @$(mktemp -d) and touch a b c d
$ echo _@$(ls)_
_a_ _b_ _c_ _d_
```
, which is not desirable, since I aim for consistency with `$()` operator in Bash.

scopatz:
Thanks so much for this @laloch! Can you add a couple of cases to `test_parser.py`?

laloch:
I guess I can. The question is what is the desired behavior :wink:

scopatz:
Well, it looks like you are having `@$(ls)` return a list of stripped strings now, rather than a string. Is that the fundemental question?

laloch:
Nope, the question now is whether we want it to behave like `@()`:
```
$ showargs a:@$(echo a b)
['a:a', 'a:b']
```
or rather like `$()` in Bash:
```
$ showargs a:@$(echo a b)
['a:a', 'b']
```
The first case is the current state and is consistent with how `@()` behaves. The second case is consistent with how `$()` in Bash behave and with what a user would most probably expect.
The more I think about it, the more I prefer the first.

scopatz:
Oh yeah, thanks for the clear example.  Xonsh is all about dumping terrible Bash behavior.  I think consistency with `@()` is more important, and so we should go with the first options

laloch:
I'm pretty sure this is now ready to land - this time for real.

scopatz:
Windows on Python 3.8 keeps failing. Can you look into it please?

laloch:
The failures seem all to be of similar nature - either `builtins` or `__xonsh__` attribute doesn't exist without any previous error. That happens quite often and I don't have the slightest clue about how that can be possible.

scopatz:
Hey @laloch - maybe we can get these tests to pass by merging from master.  Not sure if you did that just now?

laloch:
Yeah, I just did a rebase on master, that was supposed to trigger the GitHub CI. It didn't.

scopatz:
Yeah. Weird.  Maybe push an empty commit to try to trigger it?

laloch:
I did some interactive rebasing, rewording and squashing - still the same. Let's try a dummy commit now...

laloch:
Nope. Still the same :(

scopatz:
OK, I am going to try to close and reopen

scopatz:
Yeah, didn't work. I'll just merge once this passes

laloch:
I can try to open a new PR for the same branch...

scopatz:
Yeah, if you want that would be nice

laloch:
No problem. Let's try it.
Closing to open a new one with the same contents.

Issue #3160, opened 2019-06-11
Alt + . does not cycle through recent arguments


con-f-use:
In most shells you can circle through the last arguments of last few commands with <kbd>Alt</kbd> + <kbd>.</kbd>, or apparently <kbd>ESC</kbd>+<kbd>_</kbd> (see #1480). 

In the current version of xonsh, this only appends the last argument of most recent command, over and over, instead of cycling through the commands in reverse order.

## xonfig

<details>
```
+------------------+----------------------+
| xonsh            | 0.9.5                |
| Git SHA          | 550d70d2             |
| Commit Date      | May 22 01:42:07 2019 |
| Python           | 3.6.7                |
| PLY              | 3.11                 |
| have readline    | True                 |
| prompt toolkit   | 2.0.9                |
| shell type       | prompt_toolkit2      |
| pygments         | 2.4.0                |
| on posix         | True                 |
| on linux         | True                 |
| distro           | ubuntu               |
| on darwin        | False                |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+
```
</details>

## Expected Behavior
```
┌─05-18:54 [confus@confusion /tmp/xonsh]                          😱  master|+2…1 😱
└$ echo three                                                                   
three
┌─11-12:29 [confus@confusion /tmp/xonsh]                          😱  master|+2…1 😱
└$ echo two                                                                     
two
┌─11-12:29 [confus@confusion /tmp/xonsh]                          😱  master|+2…1 😱
└$ echo one                                                                     
one
┌─11-12:29 [confus@confusion /tmp/xonsh]                          😱  master|+2…1 😱
└$ echo one   # First press of Alt + .
┌─11-12:29 [confus@confusion /tmp/xonsh]                          😱  master|+2…1 😱
└$ echo two   # Second press of Alt + .      
┌─11-12:29 [confus@confusion /tmp/xonsh]                          😱  master|+2…1 😱
└$ echo three # Thrid press of Alt + .      
```

## Current Behavior
```
┌─05-18:54 [confus@confusion /tmp/xonsh]                          😱  master|+2…1 😱
└$ echo three                                                                   
three
┌─11-12:29 [confus@confusion /tmp/xonsh]                          😱  master|+2…1 😱
└$ echo two                                                                     
two
[...]

PR #3163, opened 2019-06-12
Remove hard coded Alt+. binding

con-f-use:
Is part of prompt toolkit now

fixes https://github.com/xonsh/xonsh/issues/3160.

### DIFF
--- a/xontrib/bashisms.py
+++ b/xontrib/bashisms.py
@@ -3,7 +3,6 @@ import shlex
 import sys
 import re
 
-from prompt_toolkit.keys import Keys
 from prompt_toolkit.filters import Condition, EmacsInsertMode, ViInsertMode
 
 from xonsh.platform import ptk_shell_type
@@ -53,13 +52,9 @@ def custom_keybindings(bindings, **kw):
             return len(__xonsh__.history) > 0
 
     else:
-        handler = bindings.registry.add_binding
-
         @Condition
         def last_command_exists(cli):
             return len(__xonsh__.history) > 0
-
-    insert_mode = ViInsertMode() | EmacsInsertMode()
 
 
 def alias(args, stdin=None):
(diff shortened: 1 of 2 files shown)


xontrib/bashisms.py
```diff
@@ -3,9 +3,6 @@
 import sys
 import re
 
-from prompt_toolkit.keys import Keys
-from prompt_toolkit.filters import Condition, EmacsInsertMode, ViInsertMode
-
 from xonsh.platform import ptk_shell_type
```
gforsyth:
```suggestion

```

con-f-use: The question is, should we keep this for prompt_toolkit1 users:

```
from xonsh.platform import ptk_shell_type

@events.on_ptk_create
def custom_keybindings(bindings, **kw):
    if ptk_shell_type() != "prompt_toolkit2":
        from prompt_toolkit.keys import Keys
        from prompt_toolkit.filters import Condition, EmacsInsertMode, ViInsertMode
        
        handler = bindings.registry.add_binding

        @Condition
        def last_command_exists(cli):
            return len(__xonsh__.history) > 0

        insert_mode = ViInsertMode() | EmacsInsertMode()

        @handler(Keys.Escape, ".", filter=last_command_exists & insert_mode)
        def recall_last_arg(event):
            arg = __xonsh__.history[-1].cmd.split()[-1]
            event.current_buffer.insert_text(arg)
```

gforsyth: ptk1 is deprecated, as far as I'm concerned.  The only time we should touch those files is when we strip them out.

gforsyth:
Thanks for putting this in @con-f-use !

con-f-use:
... or ripping it out.

gforsyth:
Thanks again @con-f-use !

Issue #3165, opened 2019-06-14
completer DeprecationWarning


6syun9:
# xonfig
<details>

```
+------------------+----------------------+
| xonsh            | 0.9.6.dev6           |
| Git SHA          | 610d4c1f             |
| Commit Date      | Jun 13 03:12:18 2019 |
| Python           | 3.6.8                |
| PLY              | 3.11                 |
| have readline    | True                 |
| prompt toolkit   | 2.0.9                |
| shell type       | prompt_toolkit2      |
| pygments         | 2.3.1                |
| on posix         | True                 |
| on linux         | False                |
| on darwin        | True                 |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+
```

</details>

## Expected Behavior
Do not issue Warning when doing completer for pandas.

## Current Behavior
```
$ d = pd.DataFrame([{'a': 10, 'b': 2}, {'a': 11, 'b': 0}])
$ d[[[tab]]] ./Users/6syun9/.pyenv/versions/3.6.8/bin/xonsh:1: DeprecationWarning:
.ix is deprecated. Please use
.loc for label based indexing or
.iloc for positional indexing

See the documentation here:
http://pandas.pydata.org/pandas-docs/stable/indexing.html#ix-indexer-is-deprecated
  #!/Users/6syun9/.pyenv/versions/3.6.8/bin/python3.6 -u
/Users/6syun9/.local/lib/python3.6/site-packages/xonsh/completers/__amalgam__.py:1564: DeprecationWarning:
.ix is deprecated. Please use
.loc for label based indexing or
.iloc for positional indexing

See the documentation here:
http://pandas.pydata.org/pandas-docs/stable/indexing.html#ix-indexer-is-deprecated
  a = getattr(val, opt)
```

## Steps to Reproduce
I'm using pandas==0.24.1
.ix method is deprecated, but xonsh's completer doing evaluation by `getattr`.
This issue makes the operation heavy and display is misaligned.
I think that it is necessary to fix not show Warning when getattr in completer.
https://github.com/xonsh/xonsh/blob/master/xonsh/completers/python.py#L229

6syun9:
reference
```
$ getattr(df, 'ix')
[...]

PR #3166, opened 2019-06-14
completer turn off warning

6syun9:
fix: https://github.com/xonsh/xonsh/issues/3165

We can use `warnings.catch_warnings`. But It was useless if I tried.

> Once the context manager exits, the warnings filter is restored to its state when the context was entered. This prevents tests from changing the warnings filter in unexpected ways between tests and leading to indeterminate test results. The showwarning() function in the module is also restored to its original value. Note: this can only be guaranteed in a single-threaded application. If two or more threads use the catch_warnings context manager at the same time, the behavior is undefined.

ref: https://docs.python.org/3/library/warnings.html#testing-warnings

So I added `Decorator to turn off warning temporarily` .


scopatz:
Thanks @6syun9! Can you please merge from master to get the correct CI configuration.

6syun9:
@scopatz um...CI test passed.
But I'm warnings have become visible after completer.

For example, after completion...
```
 $ cd tmp/
/Users/6syun9/.pyenv/versions/3.6.8/lib/python3.6/threading.py:916: ResourceWarning: unclosed file <_io.TextIOWrapper name=7 encoding='utf-8'>
  self.run()
```
I need to review the warning settings. WIP.

6syun9:
@scopatz Sorry, It was caused by the fact that I did not have the latest version.
The behavior was fixed!

melund:
Thanks a lot @6syun9.

PR #3170, opened 2019-06-17
Xontrib zipapp support

astronouth7303:
Make the xontrib machinery zipapp friendly.

The resulting checking is more complicated than I'd like, but there's a lot of edge cases. Good news: All of it can be removed when 3.5, 3.6 are no longer supported.

Full disclosure: This is part of my own use of building xonsh zipapps and using them in my homelab. They should make xonsh more robust against being installed with other applications and being started inside of active venvs.

### DIFF
--- a/xonsh/xontribs.py
+++ b/xonsh/xontribs.py
@@ -92,12 +92,12 @@ def xontrib_metadata():
                 pass
 
     if impres:
-        with impres.open_text('xonsh', 'xontribs.json') as f:
+        with impres.open_text("xonsh", "xontribs.json") as f:
             md = json.load(f)
     elif pkg_resources:
         # Despite the name, this is a bytes
-        bytesdata = pkg_resources.resource_string('xonsh', 'xontribs.json')
-        md = json.loads(bytesdata.decode('utf-8'))
+        bytesdata = pkg_resources.resource_string("xonsh", "xontribs.json")
+        md = json.loads(bytesdata.decode("utf-8"))
     else:
         path = os.path.join(os.path.dirname(__file__), "xontribs.json")
         with open(path, "r") as f:
--- a/xonsh/xontribs.py
+++ b/xonsh/xontribs.py
@@ -11,11 +11,6 @@ import importlib.util
 from xonsh.tools import print_color, unthreadable
 
 
-@functools.lru_cache(1)
-def xontribs_json():
-    return os.path.join(os.path.dirname(__file__), "xontribs.json")
-
-
 def find_xontrib(name):
     """Finds a xontribution from its name."""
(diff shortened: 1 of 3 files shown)


astronouth7303:
I'm not sure why the windows builds are failing? They seem to be failing on code I didn't touch.

scopatz:
Thanks a million @astronouth7303! And those failures on windows look a lot like the random test failures caused by pytest (see tracking issue https://github.com/pytest-dev/pytest/issues/5348).

xonsh/xontribs.py
```diff
@@ -76,8 +71,38 @@ def update_context(name, ctx=None):
...
+        import importlib.resources as impres
+    except ImportError:
+        try:
+            # Optional backport for <3.7
+            import importlib_resources as impres
+        except ImportError:
+            try:
+                # Try the slower and clunkier pkg_resources
+                # This is only available if setuptools is part of the environment
+                import pkg_resources
```
scopatz: Is `pkg_resources` really needed to support zipapp? I would like to avoid this package if at all possible.

astronouth7303: I mean, you could also write the zip-reading and path traversal code yourself. But this already exists in `pkg_resources`, it's stdlib counterpart `importlib.resources`, and the stdlib backport `importlib_resources`. Hence all the variants.

All of these are completely optional dependencies. However, `pkg_resources` is part of `setuptools`, which is a dependency of `pip`, so I expect many real-world environments to include it. (Exceptions being things like some condas and several stock Linux distros.)

In the case where none of the three modules are available, it falls back to using the filename. While this is completely incompatible with any form of zip-based execution, it should mean nothing will break. (And it's easy enough to include extra dependencies when building the zip.)

The [`importlib.resources` docs](https://docs.python.org/3/library/importlib.html#module-importlib.resources) includes some discussion about these modules.

astronouth7303: There's also a reason why it's my 3rd choice of package to use.

astronouth7303: Oh, and xonsh implicitly optionally depends on `pkg_resources` because it's the only way I know of to utilize entry points, and xonsh registers entry points with pygments and pytest.

scopatz: Yeah - I know, we have just goine to crazy lengths to avoid importing `pkg_resources` (or importing it in a thread) on startup. My concern is that for folks using xontribs in their xonshrc's, this could really slow down their startup times.  That said, it is worth merging as is for now, and let's keep an eye on it.  Thanks again @astronouth7303!

astronouth7303: If it's bad, that version can easily be removed.

PR #3172, opened 2019-06-18
xontrib-ssh-agent added

dyuri:
Thanks for this awesome shell, I just discovered it last week and I'm already in love with it.

Since I use `ssh` a lot, I've created a xontrib to start `ssh-agent` and add ssh identities "automatically".


scopatz:
This is awesome! Thanks @dyuri!

PR #3173, opened 2019-06-18
[DO NOT MERGE] trigger error with assert plain

scopatz:
This is for https://github.com/pytest-dev/pytest/issues/5348


jnoortheen:
This is no longer the case after using singletons

PR #3179, opened 2019-06-23
Allow prompt_ret_code xontrib to set background color

m-col:
I discovered xonsh last week and I've been loving it.

This change extends the prompt_ret_code xontrib. It looks for `$PROMPT_RET_CODE_BACKGROUND` and if it is `True` then the return code will set the prompt background color, rather than foreground color.

I wasn't sure how to look for the existence of a variable nicely and ending up using a `try` block. Is this the best way?

This PR also fixes a typo in xonsh/aliases.py.

### DIFF
--- a/news/prompt_ret_code_background.rst
+++ b/news/prompt_ret_code_background.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* added $PROMPT_RET_CODE_BACKGROUND which causes the prompt_ret_code xontrib to
+  change background rather than foreground color when true.
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/news/prompt_ret_code_background.rst
+++ b/news/prompt_ret_code_background.rst
@@ -4,8 +4,7 @@
 
 **Changed:**
 
-* added $PROMPT_RET_CODE_BACKGROUND which causes the prompt_ret_code xontrib to
-  change background rather than foreground color when true.
+* added ``$XONTRIB_PROMPT_RET_CODE_COLOR_ZERO`` and ``$XONTRIB_PROMPT_RET_CODE_COLOR_NONZERO`` which can be used to set the color format for the prompt_ret_code xontrib.
 
 **Deprecated:**
 
--- a/xontrib/prompt_ret_code.xsh
+++ b/xontrib/prompt_ret_code.xsh
@@ -1,32 +1,35 @@
 from xonsh.tools import ON_WINDOWS as _ON_WINDOWS
+from xonsh.tools import is_string, ensure_string
+from xonsh.environ import Ensurer
+
+
+${...}.set_ensurer("XONTRIB_PROMPT_RET_CODE_COLOR_ZERO",
+    Ensurer(is_string, ensure_string, ensure_string))
+
+${...}.set_ensurer("XONTRIB_PROMPT_RET_CODE_COLOR_NONZERO",
+    Ensurer(is_string, ensure_string, ensure_string))
 
 
 def _ret_code_color():
-    try:
-	color_bg = $PROMPT_RET_CODE_BACKGROUND
-    except:
-	color_bg = False
-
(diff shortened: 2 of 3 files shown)


xontrib/prompt_ret_code.xsh
```diff
@@ -2,20 +2,31 @@ from xonsh.tools import ON_WINDOWS as _ON_WINDOWS
 
 
 def _ret_code_color():
+    try:
+	color_bg = $PROMPT_RET_CODE_BACKGROUND
```
gforsyth: For xonsh env-vars, we try to start them with `XONSH_` in an effort to namespace them and keep things a bit organized -- this is a xontrib, although a bundled xontrib.  What do you think about prefixing it with `XONTRIB_`?

xontrib/prompt_ret_code.xsh
```diff
@@ -2,20 +2,31 @@ from xonsh.tools import ON_WINDOWS as _ON_WINDOWS
...
+    except:
+	color_bg = False
+
+    if color_bg:
+	prefix = 'BACKGROUND'
```
gforsyth: Another option here is to allow the env-var to take the values of `BOLD` or `BACKGROUND` (and potentially others) and then use the value directly.

Take a look in `xonsh.environ` at the `DEFAULT_ENSURERS` for how we type and validate env-var inputs

You can create an ensurer for the acceptable value types and then add that ensurer using

```
${...}.set_ensurer(env_var_name, ensurer_func)
```

news/prompt_ret_code_background.rst
```diff
@@ -0,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* added $PROMPT_RET_CODE_BACKGROUND which causes the prompt_ret_code xontrib to
```
gforsyth:
```suggestion
* added ``$PROMPT_RET_CODE_BACKGROUND`` which causes the prompt_ret_code xontrib to
```

Just a small formatting thing, but adding double backticks makes the release notes look nicer

xontrib/prompt_ret_code.xsh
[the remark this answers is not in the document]
m-col: OK - that is much nicer that the try block!

m-col: Seems reasonable.

xontrib/prompt_ret_code.xsh
m-col: Good idea. My proposed changes extend hard coded colors to slightly more hard coded colors, but the functionality should probably be flexible enough to have any color string that would be valid in `$PROMPT_FIELDS`. This would only require a variable for each of zero or non-zero return codes.

I'll have a go setting up an ensurer for these two variables to make sure they are valid color strings.

m-col: Actually - it would be 4 variables. zero and non-zero for background and foreground.

m-col:
I boiled down the settings for color format to two variables, one for each of zero and non-zero return codes:
```
$XONTRIB_PROMPT_RET_CODE_COLOR_ZERO
$XONTRIB_PROMPT_RET_CODE_COLOR_NONZERO
```
I added ensurers for these too to make sure they are strings. Please let me know if this wasn't done correctly!

I also moved the ```ON_WINDOWS``` test out of the prompt function so it's only tested when the xontrib is loaded rather than during every draw. I assume this was there originally due to different color codes on windows, and I figured simply setting the default for windows in this way would be fine.

The last thing to add is documentation (and update the news file). I read around in your link to environ.py to see how documentation is handled. I didn't manage to add documentation using an equivalent method - I suppose this is what you meant by the lack of "public-facing" env-var documentation setting. Is there a temporary workaround until such a functionality exists?

Of course please do let me know of any improvements to my commits!

news/prompt_ret_code_background.rst
scopatz: Also, this doesn't seem to be the name of the variable any longer.

laloch:
What's up with this PR? Did we reject, abandon or just forget it?

m-col:
I thought you'd forgotten about it as I didn't hear back from that last comment. I stopped using xonsh since then so happy to close it.

PR #3184, opened 2019-06-29
Add $PROMPT_REFRESH_INTERVAL

6syun9:
for example...
```
from time import strftime
def get_time():
    return strftime('[ %H:%M:%S ]')
$RIGHT_PROMPT = get_time
$UPDATE_PROMPT_ON_KEYPRESS=True
$PROMPT_REFRESH_INTERVAL=1.0
```
sample mov
ref:https://github.com/prompt-toolkit/python-prompt-toolkit/blob/master/examples/prompts/clock-input.py
  
But this is so heavy...I have plans to use, but I do not know if there is a demand.
What do you think?


astronouth7303:
Handy, for git stuff, too.

scopatz:
Thanks @6syun9! Is this still a WIP?

6syun9:
@scopatz I think we can add ! It's a bit heavy, but spreads idea!

scopatz:
thanks!

micimize:
this seems to only work with `$UPDATE_PROMPT_ON_KEYPRESS`

PR #3185, opened 2019-06-29
add $COMPLETION_IN_THREAD

6syun9:
This is very early.
https://python-prompt-toolkit.readthedocs.io/en/master/pages/asking_for_input.html#asynchronous-completion
I think by default True is fine, but I have not tried working in multi platform.


scopatz:
Thanks @6syun9! Sorry this took so long to get to!

Issue #3199, opened 2019-07-03
xontrib `cat` fails to exit when reading from sysfs


traverseda:
`/usr/bin/cat /sys/class/power_supply/BAT0/power_now ` works as expected
`cat /sys/class/power_supply/BAT0/power_now` never exits when using

```
$ cat --help    
This version of cat was written in Python for the xonsh project: http://xon.sh
```

scopatz:
Thanks for reporting! I can read from certain things in `/sys/power`  but was having issues exiting

laloch:
This seems to be fixed by #3200. @traverseda, can you please confirm?

gforsyth:
Closing this out now -- if it's still an issue please ping here.

PR #3200, opened 2019-07-05
Fix `cat` can't read pseudo files with zero size

### LINKED ISSUE #3182: coreutils cat can't read /proc/*
Hey ! cat from coreutils doesn't return anything when reading files from `/proc` :
```
$ cat /proc/cpuinfo                                                                                                                                                                            
$ /usr/bin/cat /proc/cpuinfo                                                                                                                                                                   
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 60
model name      : Intel(R) Core(TM) i5-4590 CPU @ 3.30GHz
```

Here is my config :
```
| xonsh            | 0.9.6               |
| Git SHA          | dabd4bd3            |
| Commit Date      | Jun 6 09:44:17 2019 |
| Python           | 3.7.3               |
| prompt toolkit   | 2.0.9               |
| pygments         | 2.4.2               |
| distro           | unknown             |
```

virusbb001:
I hope this patch will fix #3182 and #3199.
I tried `cat /proc/cpuinfo` in my machine.

### DIFF
--- a/xonsh/xoreutils/cat.py
+++ b/xonsh/xoreutils/cat.py
@@ -1,6 +1,5 @@
 """Implements a cat command for xonsh."""
 import os
-import stat
 import time
 import builtins
 
@@ -39,7 +38,7 @@ def _cat_single_file(opts, fname, stdin, out, err, line_count=1):
     enc = env.get("XONSH_ENCODING")
     enc_errors = env.get("XONSH_ENCODING_ERRORS")
     read_size = 0
-    file_size = fobj = None
+    fobj = None
     if fname == "-":
         f = stdin
     elif os.path.isdir(fname):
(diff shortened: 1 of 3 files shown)


xonsh/xoreutils/cat.py
```diff
@@ -49,15 +52,11 @@ def _cat_single_file(opts, fname, stdin, out, err, line_count=1):
...
-        fstat = os.stat(fname)
-        file_size = fstat.st_size
-        if file_size == 0 and not stat.S_ISREG(fstat.st_mode):
-            file_size = None
         fobj = open(fname, "rb")
         f = xproc.NonBlockingFDReader(fobj.fileno(), timeout=0.1)
     sep = os.linesep.encode(enc, enc_errors)
     last_was_blank = False
-    while file_size is None or read_size < file_size:
+    while True:
```
scopatz: I think some of these conditions might need to be implemented in the body for safety

scopatz: pinginf this again @virusbb001

virusbb001: Thanks for reviewing and sorry for the delayed response.
I think you are right but I don't know what is the good condition for safety...
Does checking `endnow` meet safety?

scopatz: Is there a reason the original check doesn't apply anymore is what I was asking. I think it could be restored.

virusbb001: The original check ( 344671ba ) caused #3182 because `stat.S_ISREG(os.stat("/proc/cpuinfo").st_mode)` returns 0 and the file size of `/proc/cpuinfo` is 0.
But the previous of original check( 7b7a9a30 ) works fine, so I think it can be restored. I will try it.

scopatz:
Thanks @virusbb001!

PR #3210, opened 2019-07-13
Avoid stale data when determining active tasks

webmeister:
Two of the three callers of get_next_task already called _clear_dead_jobs
to update the task state before searching for active tasks.

xonsh.prompt.job._current_job did not call _clear_dead_jobs, which lead to
tasks still being displayed in the title, even though they had already
finished.

Moving the call to _clear_dead_jobs into get_next_task ensures that it
always returns correct data.

### DIFF
--- a/news/get_next_task.rst
+++ b/news/get_next_task.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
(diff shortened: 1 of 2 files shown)


scopatz:
thanks!

PR #3217, opened 2019-07-13
Cleaned up proxy loading and unloading.

dotsdl:
Also removed `DeprecationWarningProxy`, as it is no longer needed.


scopatz:
thanks!

PR #3219, opened 2019-07-13
Fix behavior of enter key in VI_MODE

### LINKED ISSUE #3218: In PTK shell, `Enter` should execute line in command mode
We override the Enter key to handle multi-line blocks without requiring users to "shift enter" -- this should have an extra filter on it to check if user is in VI Mode -- if they are, then `enter` should execute the current block.

gforsyth:
Closes #3218 

h/t to @ivanov for noticing the inconsistent behavior

When not in `insertmode`, whether this is in VI mode or Emacs mode, a carriage return should execute the current block, regardless of cursor position.  This makes that happen.

### DIFF
--- a/news/vi_mode_enter.rst
+++ b/news/vi_mode_enter.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks!

PR #3220, opened 2019-07-14
Try to open man page as help for subprocess commands

### LINKED ISSUE #3204: [Feature request] Superhelp `?` should open man pages for subprocess commands
Currently, trying to run superhelp on a subprocess command just fails:

```
gil@badcatredux ~ $ ls?                                                                       
xonsh: subprocess mode: command not found: ls?
    ls:  Alias
    ss:  Command (/usr/bin/ss)
    gs:  Alias
    sh:  Command (/usr/bin/sh)
    as:  Command (/usr/bin/as)
```

but it would be great if it first checked if there's a `man` page available and then display that if there is.

webmeister:
This makes for example 'sh?' equivalent to 'man sh', unless there actually
exists a 'sh?' executable.

Resolves #3204.

### DIFF
--- a/news/subprocess-help.rst
+++ b/news/subprocess-help.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Open man page when requesting help for subprocess commands, e.g. using ``sh?``
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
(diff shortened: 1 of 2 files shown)


scopatz:
This is awesome! Thanks so much!

PR #3222, opened 2019-07-14
Allow disabling individual items in gitstatus prompt

webmeister:
Setting $XONSH_GITSTATUS_* to an empty string lead to the number of files
still being displayed, without any way to see to which category those
belonged. It might even display "11", which could be "1 changed file" and
"1 staged file" or "11 staged files".

Changing the logic to hide the entire entry if the corresponding
$XONSH_GITSTATUS_* setting is empty allows for an easy way to disable some
of the categories.

### DIFF
--- a/xonsh/prompt/gitstatus.py
+++ b/xonsh/prompt/gitstatus.py
@@ -187,20 +187,23 @@ def gitstatus_prompt():
     if s.operations:
         ret += _get_def("OPERATION") + "|" + "|".join(s.operations)
     ret += "|"
-    if s.staged > 0:
-        ret += _get_def("STAGED") + str(s.staged) + "{NO_COLOR}"
-    if s.conflicts > 0:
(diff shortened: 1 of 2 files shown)


xonsh/prompt/gitstatus.py
```diff
@@ -187,20 +187,24 @@ def gitstatus_prompt():
     if s.operations:
         ret += _get_def("OPERATION") + "|" + "|".join(s.operations)
     ret += "|"
-    if s.staged > 0:
+    if s.staged > 0 and _get_def("STAGED"):
```
scopatz: It would be nice if get_def was only called once for each of these

scopatz:
thanks for putting this in!

xonsh/prompt/gitstatus.py
webmeister: That bothered me a little too. But how to do that nicely without assignment expressions?

webmeister:
How about this solution?

scopatz:
This is a lot better, thanks!

Issue #1798, opened 2016-10-04
Web Xonfig


AstraLuma:
Create a web app for xonfig.

anki-code:
This looks solved in `xonfig web` and should be closed.

PR #3226, opened 2019-07-14
WIP: Add initial webconfig from fish

Emiller88:
Start on #1798 during scipy19.

### DIFF
--- a/web_config/webconfig.py
+++ b/web_config/webconfig.py
@@ -119,6 +119,9 @@ class XonshConfigHTTPRequestHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):
         json.dump(output, self.wfile)
 
 
+where = os.path.dirname(sys.argv[0])
+os.chdir(where)
+
 PORT = 8000
 while PORT <= 9000:
     try:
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks for starting on this @Emiller88! Has there been any more progress?

Emiller88:
@scopatz no not yet, I've been busy with other things, if you'd like to take
over feel free!

scopatz:
OK thanks! I think this is open for folks to help out with then!

web_config/webconfig.py
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/python
```
jb-leger: The `/usr/bin/env python3` sheband should be used to allow user changing python exec from its env.

web_config/webconfig.py
```diff
@@ -0,0 +1,158 @@
...
+
+try:  # Python2
```
jb-leger: xonsh doesn't not support python-legacy (a.k.a. python2), you can assume python3 is used.

web_config/webconfig.py
```diff
@@ -0,0 +1,158 @@
...
+    import SimpleHTTPServer
+except ImportError:  # Python3
+    import http.server as SimpleHTTPServer
+try:  # Python2
```
jb-leger: same comment about python2/3.

web_config/webconfig.py
```diff
@@ -0,0 +1,158 @@
...
+    import SocketServer
+except ImportError:  # Python3
+    import socketserver as SocketServer
+import webbrowser
+import subprocess
+import re, json, socket, os, sys, cgi, select
```
jb-leger: this import should be splitted in many imports.

jb-leger:
Maybe we have a licensing issue with this PR. This code is adapted from fish [webconfig](https://github.com/fish-shell/fish-shell/blob/master/share/tools/web_config/webconfig.py), and I didn't find that the webconfig is under a different license than the code code of fish, *i.e.* GPLv2.

It could be not possible to merge this code in xonsh for this point, I think we have to consider this point.

scopatz:
As per licensing, I think it depends. Everything that is here now could be written to be sufficently different

Emiller88:
@jb-leger Thanks for the feedback! 

For reference here is [fish's webconfig](https://github.com/fish-shell/fish-shell/tree/master/share/tools/web_config)

PR #3215, opened 2019-07-13
Create a pre-commit hook for black

seanfarley:
After sitting next to @gforsyth and seeing him have broken tests due to not running `black`, he mentioned [pre-commit](https://pre-commit.com) and I was _hooked_[1]. So, I put it into a PR and sent it up.

[1] @scopatz and @gforsyth have been making terrible puns all week.

PR #3230, opened 2019-07-19
precommit

scopatz:
fixes for appveypor for #3215


gforsyth:
Hey @scopatz -- I pushed up a commit to fix travis (updating the req files path).  We might also think about adding something like:

```
repos:
-   repo: https://github.com/ambv/black
    rev: stable
    hooks:
    - id: black
      language_version: python3.5
```

so that `black` doesn't inadvertently reformat something in a non-py35 compliant way.
Or we can drop 3.5 :grin:

scopatz:
You are correct as always, @gforsyth!

scopatz:
Thanks @seanfarley @gforsyth!

PR #3238, opened 2019-07-29
remove landscape badge

chengxuncc:


laloch:
Hi @chengxuncc, could you please provide a description of why we want to remove the badge and why we don't want to run Landscape checks?

chengxuncc:
#3412

laloch:
Oh, I see. Thanks for removing it!
No news item needed.

Issue #3247, opened 2019-08-02
Hardcoded path for command-not-found


vlaci:
NixOS has its own `command-not-found` implementation which lives at a different path then on Ubuntu. Unfortunately its path is hardcoded in https://github.com/xonsh/xonsh/blob/975c07ef15991fb29f86a52a0e6577d24a55e24a/xonsh/tools.py#L835

### Suggestions
Either
* Make the path of `command-not-found` configurable in an environment variable.
* Add some heuristics using `xoreutils.which` or something like that to look up the command.

I am happy to open a puil request if we can agree what the solution should be.

laloch:
Hi @vlaci, thanks for reporting.
I would say that we should rather search for the executable on the contents of the `$PATH` environment variable using `shutil.which("command-not-found")`, which returns either the path to the executable or `None` if the executable cannot be found.
A PR would be great!

PR #3249, opened 2019-08-02
command-not-found: look up in PATH instead of using hard-coded location

vlaci:
Fixes: https://github.com/xonsh/xonsh/issues/3247

Motivation: command-not-found is available on a different location under
NixOS.

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -830,14 +830,21 @@ def command_not_found(cmd):
     """Uses the debian/ubuntu command-not-found utility to suggest packages for a
     command that cannot currently be found.
     """
+    import shutil
+
     if not ON_LINUX:
         return ""
-    elif not os.path.isfile("/usr/lib/command-not-found"):
+
+    cnf = shutil.which("command-not-found")
+    if cnf is None:
         # utility is not on PATH
         return ""
-    c = "/usr/lib/command-not-found {0}; exit 0"
+    c = "{0} {1}; exit 0"
     s = subprocess.check_output(
-        c.format(cmd), universal_newlines=True, stderr=subprocess.STDOUT, shell=True
+        c.format(cnf, cmd),
+        universal_newlines=True,
+        stderr=subprocess.STDOUT,
+        shell=True,
     )
     s = "\n".join(s.rstrip().splitlines()).strip()
     return s
--- a/news/command-not-found.rst
+++ b/news/command-not-found.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* command-not-found: now works on non-Debian bansed distributions
+
+**Security:**
+
+* <news item>


xonsh/tools.py
```diff
@@ -830,14 +830,22 @@ def command_not_found(cmd):
...
     """
+    # lazy import
+    import shutil
+
     if not ON_LINUX:
         return ""
-    elif not os.path.isfile("/usr/lib/command-not-found"):
+
+    cnf = shutil.which("command-not-found")
+    if not cnf:
```
laloch: This should probably better be `if cnf is None` since the `which` function can only return a path or `None`.

laloch:
Could you please be so kind and provide a news entry for the change? See [Developer's guide](https://xon.sh/devguide.html) for details.
Thank you!

xonsh/tools.py
vlaci: Done.

laloch: I can't see any changes. I guess you forgot to force-push the commit...

vlaci: Sorry for that.

laloch: Thank you!

vlaci: eh, dirty commit. Again, sorry for that. Should not commit anything on a Friday afternoon...

xonsh/tools.py
```diff
@@ -830,14 +830,22 @@ def command_not_found(cmd):
     """Uses the debian/ubuntu command-not-found utility to suggest packages for a
     command that cannot currently be found.
     """
+    # lazy import
```
laloch: I'm sorry for nagging you, but could you please remove this comment. It's pretty obvious what the code does.

vlaci: Oh, sure. I commented because the only other place where shutil was used it was commented like that.

xonsh/tools.py
laloch: Yeah, same here :wink:

xonsh/tools.py
laloch: Does it? I'm twice sorry for nagging you then.

laloch:
Thanks @vlaci for the fix and for bearing with me!

scopatz:
Hey all, sorry to be the bearer of bad news here, but I think that on ubuntu at least `/usr/lib` (where `command-not-found` lives) is not usually on the $PATH. This is why a hard coded location was used. 

Additionally, we should probably use the [CommandsCache](https://xon.sh/api/commands_cache.html) for finding this.  Would it be too much to ask for a follow up PR that updates this to be something like:

```
cnf = builtins.__xonsh__.commands_cache.lazyget("command-not-found", ("/usr/lib/command-not-found",))[0]
```

scopatz:
Thanks for working on this @vlaci!

laloch:
Sorry for the confusion. I completely missed the fact that the path is `/usr/lib`.

>Additionally, we should probably use the CommandsCache for finding this.

@scopatz, I did specifically propose `shutil.which`, because it looks for the binary on OS $PATH, not on Xonsh $PATH or aliases. Your approach may, however, make more sense.

vlaci:
If you'd like I can create a PR to fix the issue on Ubuntu some time this week.

scopatz:
@vlaci - please! That'd be awesome.  I am going to wait to put out the next micro release until this is in.

scopatz:
@laloch - no worries! I just think using the command cache is more perfomant because everthing is already in memory

PR #3250, opened 2019-08-02
Add man-page completion for bash and zsh

NilsIrl:
Not CHANGELOG worthy

### DIFF
--- a/docs/comparison.rst
+++ b/docs/comparison.rst
@@ -51,8 +51,8 @@ table lists features and capabilities that various tools may or may not share.
       - ✓
       - ✓
     * - Man-page completion
-      -
-      -
+      - ✓
+      - ✓
       -
       - ✓
       -


laloch:
Does bash (bash-completion) really support the feature? I don't think so.

NilsIrl:
> Does bash (bash-completion) really support the feature? I don't think so.

image

laloch:
Yes, but I think that the feature is about **options** completion, based on the contents of the corresponding man page.

gforsyth:
Thanks for putting this in @NilsIrl !  
Bash completion definitely offers the completion of the names of individual man-pages -- the feature mentioned on the website refers to completing `-args` options by parsing the man pages which (I believe) is offered only by `xonsh` and `fish`.

Maybe we can figure out a language tweak to clarify that point?

NilsIrl:
Maybe "command completion from man-pages".

PR #3251, opened 2019-08-03
vox activate: use absolute path in $PATH

cjrh:
# Overview

Currently, when I use `vox activate venv/`, a _relative path_ is prepended to `$PATH`, and this means that if I change away from the current directory (which would typically be the project root), the path no longer resolves correctly.

This PR proposes to insert the _absolute path_ into `$PATH`.

# xonfig

```
$ xonfig
| xonsh            | 0.9.9                                                 |
| Git SHA          | 5aac8711                                              |
| Commit Date      | Jul 19 17:43:33 2019                                  |
| Python           | 3.7.4                                                 |
| pygments         | 2.3.1                                                 |
| on cygwin        | <xonsh.lazyasd.LazyBool object at 0x000001C411D0B388> |
| on msys2         | <xonsh.lazyasd.LazyBool object at 0x000001C411D0B3C8> |
```

# Details

I have a project with a `venv/` directory in the project root.  This virtualenv was created with

```
$ py -3.8 -m venv venv
```

To activate the virtual environment in xonsh, I type:

```
$ vox activate venv/
```

This results in the following `$PATH`:

```
$ $PATH
EnvPath(
['venv\\Scripts',
 'C:\\ProgramData\\DockerDesktop\\version-bin',
 'C:\\Program Files\\Docker\\Docker\\Resources\\bin',
 'G:\\Programs\\Python37\\Scripts\\',
 'G:\\Programs\\Python37\\',
 'C:\\Program Files (x86)\\Microsoft SDKs\\Azure\\CLI2\\wbin',
 'C:\\Program Files\\Docker\\Docker\\resources\\bin',
 'C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v10.0\\bin',
 'C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v10.0\\libnvvp',
 <...snip...>
]
```

At this stage, `$which pip` and `$which python` resolve correctly to the ones inside the venv. However, if I change into a subdirectory below the root, they now resolve to my globally installed Python **even though the virtualenv is still "activated".**

After applying the change in this PR, the value of `$PATH` after `$ vox activate venv/` becomes:

```
$ $PATH
EnvPath(
['G:\\Documents\\repos\\numba-shootout\\venv\\Scripts',      # <----- absolute path
 'C:\\ProgramData\\DockerDesktop\\version-bin',
 'C:\\Program Files\\Docker\\Docker\\Resources\\bin',
 'G:\\Programs\\Python37\\Scripts\\',
 'G:\\Programs\\Python37\\',
 'C:\\Program Files (x86)\\Microsoft SDKs\\Azure\\CLI2\\wbin',
 'C:\\Program Files\\Docker\\Docker\\resources\\bin',
 'C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v10.0\\bin',
 'C:\\Program Files\\NVIDIA GPU Computing Toolkit\\CUDA\\v10.0\\libnvvp',
```

Now, `$ which pip` and `$ which python` continue to resolve to the correct copies inside the virtualenv even when changing in subdirectories.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,6 +2,7 @@
 branch = true
 source =
   xonsh/
+  xontrib/
 omit =
   */__amalgam__.py
   xonsh/lazyasd.py
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -202,5 +202,6 @@ Authors are sorted by number of commits.
 * Daniel Smith
 * goodboy
 * Atsushi Morimoto
+* Caleb Hattingh
 
 
--- a/tests/test_vox.py
+++ b/tests/test_vox.py
@@ -114,12 +114,20 @@ def test_activate_non_vox_venv(xonsh_builtins, tmpdir):
     assert env["PATH"][0] == vxv.bin
     assert os.path.isabs(vxv.env)
     assert env["VIRTUAL_ENV"] == vxv.env
-    assert last_event == ("activate", venv_dirname, str(pathlib.Path(tmpdir) / 'venv'))
+    assert last_event == (
+        "activate",
+        venv_dirname,
+        str(pathlib.Path(str(tmpdir)) / 'venv')
+    )
 
     vox.deactivate()
     assert not env["PATH"]
     assert "VIRTUAL_ENV" not in env
-    assert last_event == ("deactivate", tmpdir.join(venv_dirname), str(pathlib.Path(tmpdir) / 'venv'))
+    assert last_event == (
+        "deactivate",
+        tmpdir.join(venv_dirname),
+        str(pathlib.Path(str(tmpdir)) / 'venv')
+    )
 
 
 @skip_if_on_msys
--- a/tests/test_vox.py
+++ b/tests/test_vox.py
@@ -1,9 +1,10 @@
 """Vox tests"""
 
-import builtins
 import stat
 import os
+import subprocess as sp
 import pytest
+import sys
 from xontrib.voxapi import Vox
 
 from tools import skip_if_on_conda, skip_if_on_msys
@@ -78,6 +79,47 @@ def test_activate(xonsh_builtins, tmpdir):
     vox.deactivate()
     assert "VIRTUAL_ENV" not in xonsh_builtins.__xonsh__.env
     assert last_event == ("deactivate", "spam")
+
+
+@skip_if_on_msys
+@skip_if_on_conda
+def test_activate_non_vox_venv(xonsh_builtins, tmpdir):
+    """
+    Create a virtual environment using Python's built-in venv module
+    (not in VIRTUALENV_HOME) and verify that vox can activate it correctly.
+    """
+    xonsh_builtins.__xonsh__.env.setdefault("PATH", [])
+
+    last_event = None
+
+    @xonsh_builtins.events.vox_on_activate
+    def activate(name, **_):
+        nonlocal last_event
+        last_event = "activate", name
+
+    @xonsh_builtins.events.vox_on_deactivate
+    def deactivate(name, **_):
+        nonlocal last_event
+        last_event = "deactivate", name
+
+    with tmpdir.as_cwd():
+        venv_dirname = 'venv'
+        sp.run([sys.executable, '-m', 'venv', venv_dirname])
+        vox = Vox()
+        vox.activate(venv_dirname)
+        vxv = vox[venv_dirname]
+
+    env = xonsh_builtins.__xonsh__.env
+    assert os.path.isabs(vxv.bin)
+    assert env["PATH"][0] == vxv.bin
+    assert os.path.isabs(vxv.env)
+    assert env["VIRTUAL_ENV"] == vxv.env
+    assert last_event == ("activate", venv_dirname)
+
+    vox.deactivate()
+    assert not env["PATH"]
+    assert "VIRTUAL_ENV" not in env
+    assert last_event == ("deactivate", tmpdir.join(venv_dirname))
 
 
 @skip_if_on_msys
--- a/tests/test_vox.py
+++ b/tests/test_vox.py
@@ -93,14 +93,14 @@ def test_activate_non_vox_venv(xonsh_builtins, tmpdir):
     last_event = None
 
     @xonsh_builtins.events.vox_on_activate
-    def activate(name, **_):
+    def activate(name, path, **_):
         nonlocal last_event
-        last_event = "activate", name, _['path']
+        last_event = "activate", name, path
 
     @xonsh_builtins.events.vox_on_deactivate
-    def deactivate(name, **_):
+    def deactivate(name, path, **_):
         nonlocal last_event
-        last_event = "deactivate", name, _['path']
+        last_event = "deactivate", name, path
 
     with tmpdir.as_cwd():
         venv_dirname = 'venv'
--- a/tests/test_vox.py
+++ b/tests/test_vox.py
@@ -95,12 +95,12 @@ def test_activate_non_vox_venv(xonsh_builtins, tmpdir):
     @xonsh_builtins.events.vox_on_activate
     def activate(name, **_):
         nonlocal last_event
-        last_event = "activate", name
+        last_event = "activate", name, _['path']
 
     @xonsh_builtins.events.vox_on_deactivate
     def deactivate(name, **_):
         nonlocal last_event
-        last_event = "deactivate", name
+        last_event = "deactivate", name, _['path']
 
     with tmpdir.as_cwd():
         venv_dirname = 'venv'
@@ -114,12 +114,12 @@ def test_activate_non_vox_venv(xonsh_builtins, tmpdir):
     assert env["PATH"][0] == vxv.bin
     assert os.path.isabs(vxv.env)
     assert env["VIRTUAL_ENV"] == vxv.env
-    assert last_event == ("activate", venv_dirname)
+    assert last_event == ("activate", venv_dirname, str(pathlib.Path(tmpdir) / 'venv'))
 
     vox.deactivate()
     assert not env["PATH"]
     assert "VIRTUAL_ENV" not in env
-    assert last_event == ("deactivate", tmpdir.join(venv_dirname))
+    assert last_event == ("deactivate", tmpdir.join(venv_dirname), str(pathlib.Path(tmpdir) / 'venv'))
 
 
 @skip_if_on_msys
--- a/news/vox-activate-absolute-path.rst
+++ b/news/vox-activate-absolute-path.rst
@@ -1,0 +1,3 @@
+**Changed:**
+
+* `vox activate` will now prepend the absolute path of the virtualenv `bin/` directory (or `Scripts/` on Windows) to `$PATH`; before this was a relative path.
--- a/news/vox-activate-absolute-path.rst
+++ b/news/vox-activate-absolute-path.rst
@@ -1,3 +1,3 @@
 **Changed:**
 
-* `vox activate` will now prepend the absolute path of the virtualenv `bin/` directory (or `Scripts/` on Windows) to `$PATH`; before this was a relative path.
+* ``vox activate`` will now prepend the absolute path of the virtualenv ``bin/`` directory (or ``Scripts/`` on Windows) to ``$PATH``; before this was a relative path.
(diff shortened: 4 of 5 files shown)


tests/test_vox.py
```diff
@@ -80,6 +81,47 @@ def deactivate(name, **_):
...
+    assert os.path.isabs(vxv.bin)
+    assert env["PATH"][0] == vxv.bin
+    assert os.path.isabs(vxv.env)
+    assert env["VIRTUAL_ENV"] == vxv.env
+    assert last_event == ("activate", venv_dirname)
+
+    vox.deactivate()
+    assert not env["PATH"]
+    assert "VIRTUAL_ENV" not in env
+    assert last_event == ("deactivate", tmpdir.join(venv_dirname))
```
cjrh: Not necessarily a problem, but note that the `deactivate` event now emits the full absolute path of the venv path _in the case where a virtual environment was activated from a directory name_.  This is due to the branch inside the `active()` method which I've left as-is.  I think this behaviour is fine (what else would make sense here?!), but it does mean that the values emitted from the `activate` and `deactivate` events are assymetric in this specific case.

One way to resolve the asymmetry would be to modify the `activate()` method to return the result of calling `active()`. This is what `deactivate` does.  So, the new `activate()` might look like this:

```
    def activate(self, name):
        """
        Activate a virtual environment.

        Parameters
        ----------
        name : str
            Virtual environment name or absolute path.
        """
        env = builtins.__xonsh__.env
        ve = self[name]
        if "VIRTUAL_ENV" in env:
            self.deactivate()

        type(self).oldvars = {"PATH": list(env["PATH"])}
        env["PATH"].insert(0, ve.bin)
        env["VIRTUAL_ENV"] = ve.env
        if "PYTHONHOME" in env:
            type(self).oldvars["PYTHONHOME"] = env.pop("PYTHONHOME")

        events.vox_on_activate.fire(name=self.active())

```

This change will mean that for "vox" venvs (that get put into `VIRTUALENV_HOME`), the two events will continue to emit the "name" of the venv, but for "python -m venv venv/" venvs, the two events will emit the full absolute path of the venv directory.

@astronouth7303 should I also make this change, or leave things as is?

.coveragerc
```diff
@@ -2,6 +2,7 @@
 branch = true
 source =
   xonsh/
+  xontrib/
```
cjrh: @astronouth7303 was `xontrib/` omitted from coverage intentionally, or was it just an oversight?

astronouth7303: tbh, It's been too long, I have no idea.

It probably has to do with the testability of a lot of the xontribs?

tests/test_vox.py
astronouth7303: Ummm... Let me see if I can process all this.

My thought is that yes, the activate/deactivate events should be consistent about the form they take. That is, paired events should use the same value, so that code using it doesn't have to attempt to normalize it or w/e.

My thought as to if they should sometimes be relative (in the case of vox-based venvs) or always full directory is a matter of what sort of abstraction we want vox to do? If we should sometimes present the relative path, my thought is that venv names should be `str` and full paths should be `Path`.

cjrh: A better option would be for the events to return a consistent structure, something like

```
>>> from collections import namedtuple
>>> ActivationResult = namedtuple('ActivationResult', 'vox_name path active')
>>> emit_after_activate = ActivationResult(vox_name='blah', path='C:/blah/blah/venv', active=True)
>>> emit_after_activate
ActivationResult(vox_name='blah', path='C:/blah/blah/venv', active=True)
>>> emit_after_deactivate = ActivationResult(vox_name='blah', path='C:/blah/blah/venv', active=False)
>>> emit_after_deactivate
ActivationResult(vox_name='blah', path='C:/blah/blah/venv', active=False)
```
The above results would be for a vox-managed venv. Then, for a non-vox-managed venv (the py -m venv venv type), we could emit the same structure, but with a `None` for the vox name:

```
>>> emit_after_activate = ActivationResult(vox_name=None, path='C:/blah/blah/venv', active=True)
>>> emit_after_activate
ActivationResult(vox_name=None, path='C:/blah/blah/venv', active=True)
>>> emit_after_deactivate = ActivationResult(vox_name=None, path='C:/blah/blah/venv', active=False)
>>> emit_after_deactivate
ActivationResult(vox_name=None, path='C:/blah/blah/venv', active=False)
```
This brings consistency back to both kinds of venvs, and will make it much easier for users to process event results.

In my examples above, I used a `str` for the path but you indicated wanted `Path`: I'm guessing you mean specifically `pathlib.Path`?

astronouth7303: I mean, the event kwargs are supposed to be that prop bag. So I guess the thing to do is to add a new kwarg `path` and have `name` be either the name as was passed in, the directory name, or some other semi-reasonable thing.

And yes, I mean `pathlib.Path` as returned by xonsh p-strings.

cjrh: Yes, good point, the `kwargs` is clearly the returned structure! Apologies for missing that.  Ok, so I propose keeping the current `name` kwarg, and having it mean "vox name", and adding a new `path` kwarg like you suggest, returning a `pathlib.Path` instance.

For non-vox venvs, when firing the `activate` and `deactivate` events the new `path` kwarg will be populated with an absolute path just like the other vox-managed-venv case, but for the existing `name` kwarg there are three options:

1. the `name` kwarg will have value `None` and the `path` will be the absolute path (same as the vox-managed venv case).
2. the `name` kwarg will also have the abspath of the venv, i.e., duplicate the value of `path`
3. the `name` kwarg contains whatever was passed into `activate`.

Of the three I much prefer option 1.  Option 2 is pretty bad and option 3 is not going to be useful to a receiver of the event. They would have to already know whether the `name` value in this situation is _not_ a vox-managed-venv to do anything sensible with it. Maybe useful for logging? Dunno.  So I prefer option 1 because it is explicit that that venv does not have a name. (I guess I should update the vox docs too in this PR?)

You're the maintainer though, so what would you prefer?

astronouth7303: Based on some experiments, 3 is what is actually happening currently. We should probably keep backwards compatibility, whether it's useful or not.

cjrh: Option 1 would indeed be backwards-incompatible, but I had assumed that with the xonsh version number being < 1.0, API changes were still allowed.  Ok, so if no change in the deactivation event signature is allowed, we need a way of allowing the `deactivate()` method know what was originally passed into `activate()`.  Currently (i.e. without this PR), the way `deactivate()` obtains the `name` to emit comes from the `VIRTUAL_ENV` environment variable, here:

```
    def active(self):
        if "VIRTUAL_ENV" not in builtins.__xonsh__.env:
            return
        env_path = builtins.__xonsh__.env["VIRTUAL_ENV"]
        if env_path.startswith(self.venvdir):
            name = env_path[len(self.venvdir) :]
            if name[0] in "/\\":
                name = name[1:]
            return name
        else:
            return env_path    # <--- in our case, this is the value of VIRTUAL_ENV

    def deactivate(self):
        env = builtins.__xonsh__.env
        if "VIRTUAL_ENV" not in env:
            raise NoEnvironmentActive("No environment currently active.")

        env_name = self.active()    # <--- env_name comes from self.active()

        if hasattr(type(self), "oldvars"):
            for k, v in type(self).oldvars.items():
                env[k] = v
            del type(self).oldvars

        env.pop("VIRTUAL_ENV")

        print(env_name)
        events.vox_on_deactivate.fire(name=env_name)      # <--- env_name is emitted
        return env_name
```

After applying my PR, this no longer works because the `VIRTUAL_ENV` environment variable is set to the full path of the activated virtualenv. This causes the behaviour change.

A possible way forward here would be to store the original "as passed in" name on the instance, inside the `activate` method. Something like this:

```
    def activate(self, name):
        self.name = name  # <--- Save the passed-in name on the instance
        env = builtins.__xonsh__.env
        ve = self[name]
        if "VIRTUAL_ENV" in env:
            self.deactivate()

        type(self).oldvars = {"PATH": list(env["PATH"])}
        env["PATH"].insert(0, ve.bin)
        env["VIRTUAL_ENV"] = ve.env
        if "PYTHONHOME" in env:
            type(self).oldvars["PYTHONHOME"] = env.pop("PYTHONHOME")

        events.vox_on_activate.fire(name=name)

    def deactivate(self):
        """
        Deactivate the active virtual environment. Returns its name.
        """
        env = builtins.__xonsh__.env
        if "VIRTUAL_ENV" not in env:
            raise NoEnvironmentActive("No environment currently active.")

        env_name = self.active()

        if hasattr(type(self), "oldvars"):
            for k, v in type(self).oldvars.items():
                env[k] = v
            del type(self).oldvars

        env.pop("VIRTUAL_ENV")

        events.vox_on_deactivate.fire(name=self.name)    # <--- emit the saved name
        return env_name
```

ShouId I make this change, or is there something else you had in mind?

scopatz:
Thanks for putting this in @cjrh!

tests/test_vox.py
astronouth7303: Vox instances aren't persistent.

I don't recall any official backwards compatibility policy, but in general there's enough xonsh code floating around that we don't want to break things willy-nilly.

That said, I think previously the venv name was just a valid key to give to a Vox object? So using the full path/vox venv name might be ok? (Sorry for jerking you around like that)

cjrh: Good point -> I was unsure about persistence. 

Ok, so just to clarify, what other changes should I make to my PR, if any?

astronouth7303: I would suggest testing like:

```
    @xonsh_builtins.events.vox_on_activate
    def activate(name, path, **_):
        nonlocal last_event
        assert name in vox
        assert path.is_dir()
        last_event = "activate", path
```
(both the activate and deactivate handlers)

and update event generation to match. Technically, returning the full path of the venv will pass, but it would be nice if it tried to return a reduced string if it's a vox-managed venv.

scopatz:
@astronouth7303 - is there anything more to do for this?

astronouth7303:
@scopatz see my most recent comment.

You're more than welcome to override me on this, though.

scopatz:
> You're more than welcome to override me on this, though.

Ahh not trying to do that at all. I was just wondering about the status, and wanted to make sure you weren't waiting for me to merge. I think you should merge this whenever you feel it is ready @astronouth7303 :wink:

tests/test_vox.py
cjrh: Your example has both `name` and `path` in the activate event handler - are you now saying that we should proceed with adding the second `path` kwarg to the activate and deactivate event signatures?  If so, this wasn't clear to me from any of our previous dialogue.  The closest  I can see that might suggest that is:

> That said, I think previously the venv name was just a valid key to give to a Vox object? So using the full path/vox venv name might be ok? (Sorry for jerking you around like that)

which wasn't clear to me sorry!

If you're saying "do your Option 1", then it's clear and I can proceed with that?

astronouth7303: "`name` is a valid key to Vox" is actually closer to option 3, but not actually listed. But yes, I think we should add a new value `path` that contains the `Path` of the venv, and let name be formally defined as a valid key to `Vox`.

Also, I just looked at the event docs and I guess I left them horribly wrong!

tests/test_vox.py
```diff
@@ -95,12 +95,12 @@ def test_activate_non_vox_venv(xonsh_builtins, tmpdir):
     @xonsh_builtins.events.vox_on_activate
     def activate(name, **_):
         nonlocal last_event
-        last_event = "activate", name
+        last_event = "activate", name, _['path']
```
astronouth7303: ... you could just add it to the signature?

cjrh: Yes! I hadn't finished with my implementation yet and I still wanted to add the documentation for the new parameter.  Anyhow, that is now done so hopefully we're good to merge.

astronouth7303:
Thank you so much for your efforts! This looks good to me.

AUTHORS.rst
```diff
@@ -202,5 +202,6 @@ Authors are sorted by number of commits.
 * Daniel Smith
 * goodboy
 * Atsushi Morimoto
+* Caleb Hattingh
```
scopatz: In general, there is no need to edit this file, as it is autogenerated :wink:

scopatz:
Awesome! Thanks!

PR #3254, opened 2019-08-06
fix typo

rindeal:


laloch:
Thanks @rindeal!
No news entry needed.

PR #3259, opened 2019-08-06
fix XONSH_DATETIME_FORMAT VarDoc formatting

rindeal:


laloch:
Thanks @rindeal !
No news entry needed.

PR #3261, opened 2019-08-07
Update xontribs.json

michaelaye:
Recommend removal of click_tabcomplete because it's not working for a year now. It just decreases the new user experience creating error messages after launching xonsh if this is loaded.

### DIFF
--- a/xonsh/xontribs.json
+++ b/xonsh/xontribs.json
@@ -38,11 +38,6 @@
   "url": "https://github.com/ErickTucto/xontrib-base16-shell",
   "description": ["Change base16 shell themes"]
  },
- {"name": "click_tabcomplete",
-  "package": "xonsh-click-tabcomplete",
-  "url": "https://github.com/Granitosaurus/xonsh-click-tabcomplete",
-  "description": ["Adds tabcomplete functionality to click based python applications inside of xonsh."]
- },
  {"name": "coreutils",
   "package": "xonsh",
   "url": "http://xon.sh",


michaelaye:
Related unfixed issue is this: https://github.com/Granitosaurus/xonsh-click-tabcomplete/issues/1

astronouth7303:
We've had this problem before... I think `autoxsh`? Did we ever resolve that?

gforsyth:
> Did we ever resolve that?

No -- part of the 1.0 roadmap is deciding on what kind of governance protocols we have for this sort of thing.  That said, we can always re-add a xontrib if the author updates it, but we shouldn't encourage people to install broken things.

Other options are for us to take the broken xontrib, fix it, then add it to the bundled xontribs, but I would only consider that if there were a substantial userbase (no idea how to measure that).

astronouth7303:
We can't measure that with our current tools. We'd need server infra, with xonsh phoning home.

gforsyth:
> We can't measure that with our current tools. We'd need server infra, with xonsh phoning home.

Yeah, I think we can probably go with implicit scream tests instead.  If something is broken and the maintainer is unresponsive, we should take it out.

scopatz:
I think we should merge this, but I agree with @gforsyth that we need some governance rules here.

gforsyth:
Thanks for putting this in @michaelaye and for kicking off an overdue discussion!

michaelaye:
if one would like to be clever, one could check on successful run travis tests to automatically create xontribs.json? ;)

PR #3262, opened 2019-08-07
command-not-found: fix lookup of command-not-found in Ubuntu

vlaci:
The last change in #3249 has introduced a regression as
command-not-found does not reside on the $PATH on Ubuntu systems.

Additionally `CommandsCache` is introduced instead of `shutil.which` to
cache the result of $PATH lookup.

---
Question does a repeated failing lookup in `CommandsCache` hurt on Ubuntu?

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -830,15 +830,16 @@ def command_not_found(cmd):
     """Uses the debian/ubuntu command-not-found utility to suggest packages for a
     command that cannot currently be found.
     """
-    import shutil
-
     if not ON_LINUX:
         return ""
 
-    cnf = shutil.which("command-not-found")
-    if cnf is None:
-        # utility is not on PATH
+    cnf = builtins.__xonsh__.commands_cache.lazyget(
+        "command-not-found", ("/usr/lib/command-not-found",)
+    )[0]
+
+    if not os.path.isfile(cnf):
         return ""
+
     c = "{0} {1}; exit 0"
     s = subprocess.check_output(
         c.format(cnf, cmd),


vlaci:
I don't know if I should update the news file in this PR.

scopatz:
Thanks! I don't think you need to modify the news entry

xonsh/tools.py
```diff
@@ -830,12 +830,13 @@ def command_not_found(cmd):
...
     """
-    import shutil
-
     if not ON_LINUX:
         return ""
 
-    cnf = shutil.which("command-not-found")
+    cnf = builtins.__xonsh__.commands_cache.lazyget(
+        "command-not-found", ("/usr/lib/command-not-found",)
+    )[0]
```
scopatz: I think this needs to be a little more complex, sorry.  We need to check that the binary actually exists. If it does not, than `cnf` needs to be set to None so that the following block works. This is espcially true when `/usr/lib/command-not-found` is being used.

scopatz:
Thanks for putting this in!

xonsh/tools.py
vlaci: Bah, I have tested two variants when command-not-found is in the path and under /usr/lib but totally missed when it is not available. Good catch!

vlaci: Done. I have put the original `isfile()` check back in. There is still no check if the file is executable but I think it is a weird enough corner case to not care about it (as the original code did not care about it either)

scopatz: haha no worries!

PR #3264, opened 2019-08-07
remove dependency on 'random` library

rindeal:
I wonder what the use case was for a 'random' shell type feature,
nevertheless removing it will reduce start-up time and memory usage.
Apart from this obscure feature, there are no other uses of `import random`
in the main app.

This patch also includes a change in generation of choices for the `--shell-type`
argument. Prior to this patch the `--help` output looked like this:

```
  --shell-type {b,best,d,dumb,ptk,ptk1,ptk2,prompt-toolkit,prompt_toolkit,prompt-toolkit1,prompt-toolkit2,rand,random,rl,readline}
```

After this patch it looks like this:

```
  --shell-type {best,dumb,prompt_toolkit,prompt_toolkit1,prompt_toolkit2,readline}
```

I've changed it since it would still contain 'random' option otherwise,
but I think it also improves the readability of the `--help` output.

And for completeness sake, this feature was introduced in a9c606b8252c3b8ad50ac1735e32528940c9ed5e.

### DIFF
--- a/news/no-random.rst
+++ b/news/no-random.rst
@@ -1,0 +1,25 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* ``random`` shell type is deprecated and now falls back to ``best``.
+  This affects the parameter ``random`` of ``xonsh``'s ``--shell-type`` option
+  and the value ``random`` of ``SHELL_TYPE`` enviroment variable.
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1103,7 +1103,6 @@ def DEFAULT_DOCS():
             "Which shell is used. Currently two base shell types are supported:\n\n"
             "    - ``readline`` that is backed by Python's readline module\n"
             "    - ``prompt_toolkit`` that uses external library of the same name\n"
-            "    - ``random`` selects a random shell from the above on startup\n"
             "    - ``best`` selects the most feature-rich shell available on the\n"
             "       user's system\n\n"
             "To use the ``prompt_toolkit`` shell you need to have the "
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -211,10 +211,9 @@ def parser():
     p.add_argument(
         "--shell-type",
         help="What kind of shell should be used. "
-        "Possible options: readline, prompt_toolkit, random. "
         "Warning! If set this overrides $SHELL_TYPE variable.",
         dest="shell_type",
-        choices=tuple(Shell.shell_type_aliases.keys()),
+        choices=sorted(set(Shell.shell_type_aliases.values())),
         default=None,
     )
     p.add_argument(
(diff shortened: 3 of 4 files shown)


scopatz:
For what it is worth, I use "random" I'll the time.  I'll explain more tomorrow.

scopatz:
Hi @rindeal - thanks for opening this issue.  

The purpose of the "random" shell type is mostly for development and testing. This is used to make sure that xonsh devs can ensure that all of the user-facing shell types are actually being used and aren't suffering from any weird issues or bit-rot. I use random so that I am testing out both readline and PTK on a daily basis, and I can't just avoid using one of them. 

Your concern that `import random` adds to unneeded startup time is totally valid.  However, this should mostly be mitigated by our use of [amalgamate](https://github.com/xonsh/amalgamate).  This generates lazy imports for normal xonsh use by placing a line that looks like:

```python
random = _LazyModule.load('random', 'random')
```

into the `__amalgam__.py` file.  That said, I hear you that this is a single use only import.  If you want to move the import into the function where & when it is used, that would be totally fine with me!

As for the shell type, I think we could do a better job with displaying the abbreviation. But since this is the only place where the abbreviations are currently documented, I think they should stay. Or if you'd like to write a better help message that includes a little table that maps the abreviations to their expanded values, that'd be awesome!

Thanks for understanding

rindeal:
> The purpose of the "random" shell type is mostly for development and testing. This is used to make sure that xonsh devs can ensure that all of the user-facing shell types are actually being used and aren't suffering from any weird issues or bit-rot. I use random so that I am testing out both readline and PTK on a daily basis, and I can't just avoid using one of them.

This is an unusual way though. The industry standard is unit testing with as much coverage as possible, since it's a much more solid way of testing, preventing bugs like for example your false feeling that "all of the user-facing shell types are actually being used", while in fact only 50% of them are at best, because xonsh has 4 shell types as of now: readline, jupyter, ptk1 and ptk2, but _random_ uses just _readline_ and _"currently available ptk version if any with readline fallback"_.

Anyway, since the number of users is not greater than the number of fingers on one hand of a typical human being, any such option should have been hidden from users and only exposed in dev docs at best, but more likely kept just as an Easter egg.

> Your concern that import random adds to unneeded startup time is totally valid. However, this should mostly be mitigated by our use of amalgamate. This generates lazy imports for normal xonsh use by placing a line that looks like:
> 
> random = _LazyModule.load('random', 'random')
> 
> into the __amalgam__.py file. That said, I hear you that this is a single use only import. If you want to move the import into the function where & when it is used, that would be totally fine with me!

Yes, the `_LazyModule` cls makes the performance penalty argument moot and the only issue that stays is the code clarity.

All in all, I suggest to deprecate or remove `random` shell type at least from all user-facing places.

anki-code:
@rindeal @scopatz In [xxh project](https://github.com/xxh/xxh) I omitted the `random` library importing by [simple reimplementing](https://github.com/xxh/xxh/blob/master/xxh_xxh/xxh.py#L378-L379) `randint` function as datatime microseconds. In xonsh case it would enogh too.

scopatz:
> @rindeal @scopatz In xxh project I omitted the random library importing by simple reimplementing randint function as datatime microseconds. In xonsh case it would enogh too. You can save random option but reduce random dependency.

Cool idea! That would be fine.  I like the `random` shell feature though, so do want to keep it

anki-code:
Please close this. The proper solution in PR #3867.

PR #3274, opened 2019-08-18
Correct completer tutorial

con-f-use:
Corrects the code in the "lou carcolh" example for custom xonsh <kbd>TAB</kbd>-completers. Previously it would not correct on "lou", despite what the docstring said. Also it would throw an `IndexError` if empty.

Also corrects the description of `endidx`, which is not actually the index of the last character in prefix, but the length (i.e. "index of last character plus one").

Btw., I have written this decorator to debug completers more readily:

<details>
<summary>Completer debug decorator</summary>

    import tempfile, time
    from functools import wraps
    from pprint import pformat
    from pathlib import Path as P

    def complete_watcher(
        func=None, dirname=tempfile.gettempdir(),
        filename='xonsh_complete_watcher.txt',
        context=False
    ):
        if func is None:
            return functools.partial(complete_watcher,
                dirname=dirname, filename=filename, context=context )

        tmpfl = P(dirname) / filename
        tmpfl.write_text("\n** Opened on {} **\n".format(time.ctime()))

        def vertical_number(line, prefix=''):
            rows = len(str(len(line)))
            nums = ["{:<{r}d}".format(i,r=rows) for i in range(len(line))]
            return ("\n" + prefix).join([''.join(x) for x in zip(*nums)])

        @functools.wraps(func)
        def completer_(prefix, line, begidx, endidx, ctx, *args, **kwargs):
            compl = func(prefix, line, begidx, endidx, ctx, *args, **kwargs)
            complstr = pformat(compl)
            ctime = time.ctime()
            vernum = vertical_number(line+' ', ' '*6)
            markers = ' '*begidx + 'b' + ' '*(endidx-begidx-1) + 'e'
            ctxstr = pformat(ctx) + '\n' if context is True else ''
            tmpfl.write_text(
                "\n===========\n"
                "{ctxstr}"
                "{ctime}\n"
                "prefix: {prefix}\n"
                "line: {line}\n"
                "indx: {vernum}\n"
                "      {markers}\n"
                "------------\n"
                "{complstr}\n"
                "===========\n".format(**locals()) )
            return compl

        return completer_

    @completer_watcher
    def unbeliever_completer(prefix, line, begidx, endidx, ctx):
        # ...

use it with:

    tail -f /tmp/xonsh_completer_watcher.txt

</details>

### DIFF
--- a/docs/tutorial_completers.rst
+++ b/docs/tutorial_completers.rst
@@ -55,7 +55,7 @@ Completers are implemented as Python functions that take five arguments:
 * ``prefix``: the string to be matched (the last whitespace-separated token in the current line)
 * ``line``: a string representing the entire current line
 * ``begidx``: the index at which ``prefix`` starts in ``line``
-* ``endidx``: the index at which ``prefix`` ends in ``line``
+* ``endidx``: the length of the ``prefix`` in ``line``
 * ``ctx``: the current Python environment, as a dictionary mapping names to values
 
 This function should return a Python set of possible completions for ``prefix``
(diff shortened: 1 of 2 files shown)


con-f-use:
No news item warrented, IMHO. I can write one nonetheless, if you want.

scopatz:
Thanks @con-f-use! Sorry for the slow merge

con-f-use:
All good, I know your were on vacation. Hope you had a nice one.

scopatz:
Yep! It was great!  Thanks

PR #3285, opened 2019-08-26
mark git as unthreadable

### LINKED ISSUE #3263: git may need to be marked unthreadable (for "git commit")
I believe this has been an issue since 0.9.9 (I may have missed 0.9.8, so I'm not sure), but running "git commit" results in:

BlockingIOError: [Errno 35] write could not complete without blocking

## xonfig
```
| xonsh            | 0.9.9                |
| Git SHA          | 5aac8711             |
| Commit Date      | Jul 19 17:43:33 2019 |
| Python           | 3.7.4                |
| pygments         | 2.4.2                |
| on darwin        | True                 |
```

## Expected Behavior
Ideally, the git comment editor should appear.

## Current Behavior
Things blow up.  My instance of iTerm becomes sullen and refuses to talk to me.  And I *think* I saw my Mac giving my dog an angry look, but that could just be my imagination.

### Traceback (if applicable)

```
Exception in thread Thread-114:
  File "/usr/local/Cellar/python/3.7.4/Frameworks/Python.framework/Versions/3.7/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.7/site-packages/xonsh/__amalgam__.py", line 14206, in run
    i = self._read_write(procout, stdout, sys.__stdout__)
  File "/usr/local/lib/python3.7/site-packages/xonsh/__amalgam__.py", line 14250, in _read_write
    stdbuf.flush()
```

## Steps to Reproduce
$ mkdir tmp
$ cd tmp
$ git init .
$ touch foo
$ git add foo
$ git commit

fanosta:
Git may spawn an editor when used for committing or interactively rebasing.
Therefore it must not be run in a background thread, as this would cause issues with many editors such as neovim or nano,

### DIFF
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -422,6 +422,21 @@ def predict_hg(args):
         return not ns.interactive
 
 
+@lazyobject
+def GIT_PREDICTOR_PARSER():
+    p = argparse.ArgumentParser("git", add_help=False)
+    p.add_argument("command")
+    return p
+
+
+def predict_git(args):
+    """For some subcommands git may spawn an editor; in this case we do not want to
+    mark it as threadable
+    """
+    ns, _ = GIT_PREDICTOR_PARSER.parse_known_args(args)
+    return ns.command not in {"commit", "merge", "pull", "rebase"}
+
+
 def predict_env(args):
     """Predict if env is launching a threadable command or not.
     The launched command is extracted from env args, and the predictor of
@@ -458,7 +473,7 @@ def default_threadable_predictors():
         "ex": predict_false,
         "fish": predict_shell,
         "gawk": predict_true,
-        "git": predict_true,
+        "git": predict_git,
         "gvim": predict_help_ver,
         "hg": predict_hg,
         "htop": predict_help_ver,


fanosta:
this fixes #3263

gforsyth:
Hey @fanosta -- thanks for putting this in and sorry for the delay in getting to it! 
We've gone back and forth on marking `git` as unthreadable -- there appears to be an underlying issue causing this depending on the editor used.  

Unthreadable commands don't allow for captured output, so it might not be the right thing to mark `git` as unthreadable.

fanosta:
I was not aware of that limitation. Not being able to capture git output is a huge drawback but so is not being able to create commits.

The most common git commands that can spawn an editor would probably be
`commit`, `rebase`, `merge` and `pull`. Maybe we should mark only those as non-threadable. Additionally we could should the `GIT_MERGE_AUTOEDIT` environment variable. 

Using git a very common task i do in the shell. Not being able to commit properly nearly led me to abandoning xonsh altogether.

astronouth7303:
"Capturing git output" is kind of a big deal, considering the amount of git automation exists.

gforsyth:
> Using git a very common task i do in the shell. Not being able to commit properly nearly led me to abandoning xonsh altogether.

Which editor do you use?  If you use some flavor of `vim` that should work.
Only marking certain subcommands as unthreadable might be a good way forward -- someone will need to run tests to make sure that it's even possible with the current method.  I _think_ it should be, but I'm not certain.

`GIT_MERGE_AUTOEDIT` should work out of the box

fanosta:
I'm using neovim. Plain vim works fine though. nano on the other hand does not.

Marking certain subcommands as unthreadable is certainly possible:
The `args` parameter passed to the prediction function is simply the list of arguments.

gforsyth:
I thought we had `neovim` marked unthreadable already, but I don't see if there.  @fanosta can you try adding `__xonsh__.commands_cache.threadable_predictors['neovim'] = lambda x: False` to your `xonshrc`?

fanosta:
Neovim ist already marked as unthreadable. (The command to launch neovim is `nvim`)

Also it wouldn't make any difference because the shell is not involved in the creation of "grandchild" processes. The shell can only use the arguments of the process it has created for it's internal processing.

fanosta:
I've updated the pull request to only predict unthreadable for the subcommands commit merge pull and rebase

Previously you've mentioned:
> Unthreadable commands don't allow for captured output, so it might not be the right thing to mark `git` as unthreadable.

I am not able to reproduce that. I.e. The following works for me
```
__xonsh__.commands_cache.threadable_predictors['git'] = lambda args: False
x = $(git log)
```
Altough it is noticably slower when viewing a large log.


From browsing through the code I could see that when an unthreadable command is captured, first python waits for the process to finish and after thats it gathers the output.

Whereas otherwise it can gather the output while the process is running, which obviously has tremendous performance benefits.

qwenger:
Hi,

I have the same issue. I wanted to mention that:

1) using `__xonsh__.commands_cache.threadable_predictors['git'] = predict_false` works for me (`predict_false` from `xonsh.commands_cache`, probably the lambda works as well - did not test)
2) your blacklist does not seem sufficient, e.g. `git config -e` also spawns an editor. There may be other examples.

laloch:
Hi folks, +1 from me for not making non-interactive tools uncapturable/unthreadable. Interactive users can always run such tools uncaptured, whereas making them uncaptured by default breaks automation for everyone.

anki-code:
Closed as inactive.

PR #3307, opened 2019-09-09
Non-zero exit code when script path not found

### LINKED ISSUE #3292: xonsh should return a non-zero exit code when a script cannot be found
Currently xonsh returns a zero exit code when the script cannot be found:
```
$ xonsh xx.xsh && echo "OK"
xonsh: xx.xsh: No such file or directory.
OK
```
That seems like a bug. Compared to bash, which returns a non-zero exit code:
```
$ bash xx.xsh && echo "OK"
bash: xx.xsh: No such file or directory
```

gforsyth:
Fixes #3292 

Not sure if we want the `on_exit` event to still fire for this -- I've put it in for now.

### DIFF
--- a/news/fnf_exit_1.rst
+++ b/news/fnf_exit_1.rst
@@ -1,0 +1,31 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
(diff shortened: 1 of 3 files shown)


melund:
Also to me... I will merge it then.

certik:
Thanks for fixing it!

PR #3310, opened 2019-09-10
Add as_iterable to itertools

CJ-Wright:

### DIFF
--- a/xonsh/lib/itertools.py
+++ b/xonsh/lib/itertools.py
@@ -1,0 +1,35 @@
+def as_iterable(iterable_or_scalar):
+    """Utility for converting an object to an iterable.
+   Parameters
+   ----------
+   iterable_or_scalar : anything
+   Returns
+   -------
+   l : iterable
+       If `obj` was None, return the empty tuple.
+       If `obj` was not iterable returns a 1-tuple containing `obj`.
+       Otherwise return `obj`
+   Notes
+   -----
+   Although both string types and dictionaries are iterable in Python, we are treating them as not iterable in this
+   method.  Thus, as_iterable(dict()) returns (dict, ) and as_iterable(string) returns (string, )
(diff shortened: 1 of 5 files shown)


astronouth7303:
Why are dictionaries not counted?

CJ-Wright:
I think the reason is that dicts have a pair of values, one of which are missed if it is treated as an iterable.

CJ-Wright:
I'm happy to flip it the other way if needed. I mostly use the string usecase.

astronouth7303:
I mean, same. You very often want to treat strings/bytes as atomic scalars, not as collections.

I would rather see it dropped on the basis of simplicity and avoiding weird extra behaviors. These kinds of things usually come up when handling value-or-sequence arguments, and tbh passing a `dict` is just nonsensical in that context? Or it means something extra.

CJ-Wright:
ready for review!

xonsh/lib/itertools.py
```diff
@@ -0,0 +1,33 @@
...
+   iterable_or_scalar : anything
+   Returns
+   -------
+   l : iterable
+       If `obj` was None, return the empty tuple.
+       If `obj` was not iterable returns a 1-tuple containing `obj`.
+       Otherwise return `obj`
+   Notes
+   -----
+   Although both string types and dictionaries are iterable in Python, we are treating them as not iterable in this
```
gforsyth: Seems like the `Notes` here don't match the implementation anymore, otherwise, this looks good to me. Sorry for the delay in getting to it!

CJ-Wright:
@astronouth7303 @gforsyth poke

astronouth7303:
Alright, with two reviews, I'm going to pull the trigger.

gforsyth:
Thanks @CJ-wright!

PR #3313, opened 2019-09-16
Fix gray empty bar when updating on keypress

micimize:
This fixes [an issue I encountered](https://github.com/xonsh/xonsh/issues/1048#issuecomment-531497811) when using `$XONSH_UPDATE_PROMPT_ON_KEYPRESS` where the bottom bar is rendered as a grey block.

Root cause was essentially passing `"bottom_toolbar": lambda: None`. I only fixed `ptk2` because I'm not sure if it was an issue elsewhere.

The second gray bar is `tmux`, confirmed it appears outside of it though:
image

### DIFF
--- a/news/bottom_bar_keypress_issue.rst
+++ b/news/bottom_bar_keypress_issue.rst
@@ -1,0 +1,4 @@
+**Fixed:**
+
+* gray empty bottom bar when using $XONSH_UPDATE_PROMPT_ON_KEYPRESS
+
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks @micimize! Can you please add a news entry as per https://xon.sh/devguide.html#changelog

scopatz:
Travis passed, but didn't report back for some reason. Thanks for this!

PR #3323, opened 2019-09-29
add cwd args on_postcommand event

vaaaaanquish:
Hi, I want to save `open file history` by command. 
But `events.on_postcommand` doesn't have absolute path to where command executed. So we can't save path to file now.

example:
```
@events.on_postcommand
def open_file_history(cmd, rtn, **kw):
    if rtn == 0:
        cmd = cmd.split(' ')
        if cmd[0] in ['vim', 'v'] and len(cmd)==2:
            with open($OPEN_FILE_HIST_PATH, 'a') as dh:
                print(expanduser(cmd[1]), file=dh)    # cmd[1] is not necessarily absolute path
```

I changed on_postcommand args. Thanks:)

### DIFF
--- a/news/dev_on_postcommand.rst
+++ b/news/dev_on_postcommand.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* events.on_postcommand args now has cwd where command executed
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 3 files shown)


astronouth7303:
For the sorts of commands you're doing, the cwd of the command should be the same as the cwd of the event handler.

scopatz:
I think this pretty harmless to add.  I agree with @astronouth7303 that you can get this from other means though, like `os.cwd()` or `$PWD`.  Do either of these work well enough for you @vaaaaanquish?

scopatz:
Again, I don't mind having this, but I am wondering if it is needed :wink:

vaaaaanquish:
Thanks. 

> Do either of these work well enough for you
Yes, I do. I can restore the command's result from this function.

I thought that the opinion of @astronouth7303 was correct. I think it ’s better to support in a different way. I will investigate a little more.

scopatz:
Thanks @vaaaaanquish! Just let us know!

astronouth7303:
Note that if the command changes the cwd, things get weird.

Issue #3324, opened 2019-09-30
Add FAQ entry on "How to add xonsh into the context menu for Windows?"


nedsociety:
## xonfig

<details>

```
$ xonfig
+------------------+----------------------+
| xonsh            | 0.9.11               |
| Git SHA          | dd16d59d             |
| Commit Date      | Aug 25 13:21:16 2019 |
| Python           | 3.7.4                |
| PLY              | 3.11                 |
| have readline    | False                |
| prompt toolkit   | 2.0.9                |
| shell type       | prompt_toolkit2      |
| pygments         | None                 |
| on posix         | False                |
| on linux         | False                |
| on darwin        | False                |
| on windows       | True                 |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+
```

</details>

## Summary

Hi, I just wrote a script that registers xonsh into the context menu. It just works like "Open Powershell window here" menu.

![image](https://user-images.githubusercontent.com/11943013/65858473-fec35f00-e3a0-11e9-8b8d-621234cc75ca.png)

I'm using xonsh as an alternative shell for Windows so this is entirely written for personal needs. I scripted this to make it easier to setup for others, but I have no idea how much Windows users use xonsh like I do. I wonder if it worth documented somewhere like FAQ entry?

Code and usage are here: https://gist.github.com/nedsociety/91041691d0ac18bc8fd9e937ad21b055

scopatz:
Yes, please! This would be great to have in our documentation.  Please open a pull request!

melund:
I like it. Maybe this could even be a [xontrib](https://xon.sh/tutorial_xontrib.html). That would be a cool windows feature.

An other option for quickly opening xonsh: 
I am using the new [Windows terminal](https://www.microsoft.com/da-dk/p/windows-terminal-preview/9n0dx20hk701?SilentAuth=1&f=255&MSPPError=-2147217396), with xonsh configured as the default shell. Then launching xonsh from any folder is as easy as: Alt+D followed by "wt <return>"

![test-encodingFFMPEG](https://user-images.githubusercontent.com/1038978/65946426-65468c80-e436-11e9-89ac-1df489d88764.gif)

anki-code:
Hello @nedsociety!

The best way is to create a simple but searchable Github repository:
[...]

PR #3325, opened 2019-10-01
Add FAQ for adding context menu

nedsociety:
This adds a FAQ entry per #3324 .

Note that I don't mind if xonsh team forks the gist and change URLs it refers for various reasons:

1. The script uses a hardcoded invocation on xonsh (i.e. `__import__('xonsh.main').main.main()`) which might be changed later.
2. Hosting a third party gist that is meant to be automatically downloaded and run might be a liability, although it's merely a doc 🤔


scopatz:
Thanks!

PR #3326, opened 2019-10-01
Assume xsh scripts are in UTF-8

melund:
This PR changes the way xonsh reads script files so they always assumed to be utf-8, even on
Windows where the default encoding can vary. 

This allows for writing unicode characters in script files on Windows (most importantly `xonshrc`). 

Read this [PEP for more info](https://discuss.python.org/t/pep-597-use-utf-8-for-default-text-file-encoding/1819).

### DIFF
--- a/news/script_unicode.rst
+++ b/news/script_unicode.rst
@@ -1,0 +1,25 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks a million @melund!

PR #3329, opened 2019-10-01
Context vars fix

### LINKED ISSUE #3328: source breaks contextvars context
`script.xsh`:

```
source lib.xsh

print("Var", var.get())
```

`lib.xsh`:

```
from contextvars import ContextVar

var = ContextVar('var', default='spam')
var.set('foo')
```

Output:

```
🐚 xonsh script.xsh                                                 
Var spam
```

(Expected: `Var foo`)

Version: 0.9.11

scopatz:
related to #3328

### DIFF
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -368,20 +368,25 @@ print(check_output(["echo", "hello"]).decode("utf8"))
     #
     (
         """
-with open("sourced-file.xsh", "w") as f:
-    f.write('''
+import sys
+
+if sys.version_tuple[:2] >= (3, 7):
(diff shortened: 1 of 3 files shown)


tests/test_integrations.py
```diff
@@ -363,6 +363,34 @@ def _echo(args):
...
         0,
     ),
+    #
+    # test contextvars
+    #
+    (
+        """
+import sys
+
+if sys.version_tuple[:2] >= (3, 7):
```
melund: should maybe be `sys.version_info`

scopatz: haha it shy!

PR #3331, opened 2019-10-02
jedi error fix

scopatz:
closes #3319

PR #3335, opened 2019-10-02
Refactor jedi and fix pre-commit hook

seanfarley:
This is just a followup to #3331 to fix both cases of jedi loops. Also, because I couldn't commit since I didn't have python 3.5, I needed to relax the version to just `3.x`. Sorry for putting two things into one pull request.


scopatz:
Thanks! This is covered by the previous news

seanfarley:
Thanks!

PR #3336, opened 2019-10-02
Fix ANSI color default scheme

the-allanc:
The ANSI color code definitions for the "default" scheme preventing any of the modifiers (bold, underline, blink) from taking effect.

The `0;` code currently resets any of the modifiers specified before it - meaning that the current Xonsh code for "bold cyan" (`1;0;36m`) produces the same output as `36m` (cyan) rather than `1;36m` (bold cyan).

This change should fix it for the default color scheme.

### DIFF
--- a/news/fix-ansi-colors.rst
+++ b/news/fix-ansi-colors.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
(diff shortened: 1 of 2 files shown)


the-allanc:
Screenshot from 2019-10-02 21-02-39
Screenshot from 2019-10-02 21-18-51

Attaching the screenshot showing the differences between the codes, as well as showing Xonsh's current behaviour where the conversion from ANSI code to Xonsh color and back again produces a different code.

scopatz:
Thanks for making this consistent @the-allanc!

PR #3333, opened 2019-10-02
Rebuild CI

astronouth7303:
We've been evolving our CI since Travis was the only public CI option. It shows.

Time to wipe the slate clean, methinks.

This PR blows away all existing CI in favor of Cirrus CI.

I have been using Cirrus with [PursuedPyBear](https://github.com/ppb/) and [gqlmod](https://github.com/gqlmod/), [the xonsh containers](https://github.com/xonsh/container) are being built by Cirrus CI, and duckinator has been using it with their projects ([bork](https://github.com/duckinator/bork), [emanate](https://github.com/duckinator/emanate), [parts.horse](https://github.com/duckinator/parts.horse), [keress](https://github.com/duckinator/keress)). Several of these projects are performing deployment tasks as well as testing, linting, etc.

Why Cirrus?
* It has support for Linux, Windows, Mac, and FreeBSD under one service
* It is container based when available (Linux, Windows)
* It has caching, artifacts, container building, and other advanced features
* Their support has been unparalleled in my experience. Extremely responsive, extremely helpful, and improvements have been made in response to my questions and comments.

I think the diff of this PR speaks for the simplification.

PR #3337, opened 2019-10-03
Add msys to Cirrus CI

astronouth7303:
Adds an msys task to Cirrus CI.

Depends on #3333.


astronouth7303:
Ok, given that the xonsh test suite is not compatible with windows containers (something something UNC paths), I'm closing this.

Issue #2365, opened 2017-04-27
coreutils cat broken


brianmay:
```
$cat
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 13061, in run
    r = self.f(self.args, sp_stdin, sp_stdout, sp_stderr, spec)
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 12896, in proxy_four
    return f(args, stdin, stdout, stderr)
  File "/usr/local/lib/python3.6/site-packages/xonsh/xoreutils/cat.py", line 58, in cat
    o = _cat_single_file(opts, i, stdin, stdout, stderr, line_count)
  File "/usr/local/lib/python3.6/site-packages/xonsh/xoreutils/cat.py", line 21, in _cat_single_file
    _r = r = f.readline()
AttributeError: 'NoneType' object has no attribute 'readline'
```

Lets try:

```
$cat -
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 13061, in run
    r = self.f(self.args, sp_stdin, sp_stdout, sp_stderr, spec)
  File "/usr/local/lib/python3.6/site-packages/xonsh/__amalgam__.py", line 12896, in proxy_four
    return f(args, stdin, stdout, stderr)
  File "/usr/local/lib/python3.6/site-packages/xonsh/xoreutils/cat.py", line 58, in cat
    o = _cat_single_file(opts, i, stdin, stdout, stderr, line_count)
  File "/usr/local/lib/python3.6/site-packages/xonsh/xoreutils/cat.py", line 21, in _cat_single_file
    _r = r = f.readline()
AttributeError: 'NoneType' object has no attribute 'readline'
```

laerus:
I feel like we should remove xoreutils from core xonsh, they are poorly tested and a small burden for maintaining xonsh, they fit better in a xontrib anyway. i don't think others shells provide such builtins (`pwd` being an exception). @xonsh/core-developers feelings?

gforsyth:
Aren't they already a xontrib?  Or do you mean that they should live in a separate repo?

laerus:
yeah, a separate repo would be a good idea.

[...]

PR #3344, opened 2019-10-07
Cat stdin

melund:
This fixes running coreutils `cat` with no input or `cat -`. 

Fixes: #2365

@scopatz you better take a look if the fix makes sense. I am not really sure I know what I am doing here. 

It also fixes an Attribute error when pressing ctrl-C on windows. A followup to the [fix](#3327) by @cjrh the other day.

### DIFF
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -1487,7 +1487,7 @@ class ProcProxyThread(threading.Thread):
             safe_fdclose(handle)
         if self.poll() is not None:
             self._restore_sigint(frame=frame)
-        if on_main_thread():
+        if on_main_thread() and not ON_WINDOWS:
             signal.pthread_kill(threading.get_ident(), signal.SIGINT)
 
     def _restore_sigint(self, frame=None):
--- a/xonsh/xoreutils/cat.py
+++ b/xonsh/xoreutils/cat.py
@@ -1,5 +1,6 @@
 """Implements a cat command for xonsh."""
 import os
+import sys
 import time
 import builtins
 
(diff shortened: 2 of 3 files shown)


xonsh/xoreutils/cat.py
```diff
@@ -10,7 +11,7 @@
 def _cat_line(
     f, sep, last_was_blank, line_count, opts, out, enc, enc_errors, read_size
 ):
-    _r = r = f.readline(size=80)
+    _r = r = f.readline(80)
```
melund: Seemed like `stdin.readline()` didn't support keyword arguments.

xonsh/xoreutils/cat.py
```diff
@@ -44,7 +45,7 @@ def _cat_single_file(opts, fname, stdin, out, err, line_count=1):
     read_size = 0
     file_size = fobj = None
     if fname == "-":
-        f = stdin
+        f = stdin or sys.stdin
```
melund: In my case stdin was always `None` when calling `cat` with no arguments. This fixes  #2365, but maybe the error is somewhere else.

xonsh/proc.py
```diff
@@ -1487,7 +1487,7 @@ def _signal_int(self, signum, frame):
             safe_fdclose(handle)
         if self.poll() is not None:
             self._restore_sigint(frame=frame)
-        if on_main_thread():
+        if on_main_thread() and not ON_WINDOWS:
```
cjrh: 👍

scopatz:
Thanks @melund! This works for me!

PR #3346, opened 2019-10-07
Fix superhelp

melund:
This fixes a regression with superhelp and `which -v` which would print some of Python builtin docstrings instead of the user defined doctrings. 

image


scopatz:
Thanks a million @melund! I am glad this is finally fixed!

PR #3347, opened 2019-10-07
Remove built-in support for win unicode console.

melund:
The new windows terminal now gives full support for a unicode on windows. So there is really no reason to support the hack [win unicode console](https://github.com/Drekin/win-unicode-console) provided. 

This PR removes it and updates the docs.


scopatz:
Thanks so much @melund!

PR #3357, opened 2019-10-15
⚙️ Portable rootless xonsh build on AppImage

anki-code:
[AppImage](https://appimage.org/) is a format for distributing portable software on Linux without needing superuser permissions to install the application. It tries also to allow Linux distribution-agnostic binary software deployment for application developers, also called Upstream packaging. 

AppImage allows xonsh to be run on any AppImage supported Linux distributive without installation and root access.

[asciicast](https://asciinema.org/a/274543)

## Try it now

You can download and try [prebuilded xonsh.AppImage](https://github.com/anki-code/appimage-xonsh):
```
wget https://github.com/anki-code/appimage-xonsh/raw/master/xonsh-x86_64.AppImage && chmod +x xonsh-x86_64.AppImage
./xonsh-x86_64.AppImage
./xonsh-x86_64.AppImage --help
```

## Build xonsh.AppImage

Let's start:

1. Run docker with `fuse`:
```
sudo docker run --rm -it --privileged --device /dev/fuse -v `pwd`:/appimage ubuntu:16.04 bash
```

2. Install libraries for future building:

```
apt update -y && apt upgrade -y
apt install -y fuse wget mc \
	build-essential python-dev python-setuptools python-pip python-smbus \
	libncursesw5-dev libgdbm-dev libc6-dev  \
	zlib1g-dev libsqlite3-dev tk-dev \
	libssl-dev openssl \
	libffi-dev autoconf libfuse-dev
```

Not all required. Feel free to clean.

3. Getting the build scripts based on [linuxdeploy plugin python](https://github.com/niess/linuxdeploy-plugin-python):

```
mkdir -p /appimage && cd /appimage
wget https://github.com/anki-code/linuxdeploy-plugin-python/archive/entrypoint.zip -O linuxdeploy-plugin-python-entrypoint.zip
unzip linuxdeploy-plugin-python-entrypoint.zip
```

Here we see downloading of `entrypoint` branch from `anki-code/linuxdeploy-plugin-python` because the origin `niess/linuxdeploy-plugin-python` repository has not yet accepted the [pull request with xonsh](https://github.com/niess/linuxdeploy-plugin-python/pull/11).

4. Building:

```
cd linuxdeploy-plugin-python-entrypoint/appimage
./build-python.sh xonsh
ls -l
```

As result you'll find executable file `xonsh-x86_64.AppImage` that runs xonsh and can take command line arguments like xonsh:

```
# ./xonsh-x86_64.AppImage -c "echo @(1+1)"
2
```
[...]

Issue #3275, opened 2019-08-18
Autoloaded xontribs


AstraLuma:
Thinking about other ways/contexts to ship xonsh, and there are cases where you (the packager) want to inject loaded-at-init code into the distribution.

* zipapp: My zipapp distribution has a xontrib to fix `xpip` and to manage the user's site/venv
* CI: CI contexts need additional envvars registered
* GitHub Actions: Additional envvars, such as parsing `$GITHUB_EVENT_PATH` (a path to a JSON) and `$INPUT_*`.

I would argue that xonsh is ripe for automation like above, due to its ability to strongly manipulate both data and the system. Being able to inject autoloaded xontrib-like modules would be immensely helpful.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

astronouth7303:
This might be related to #2151

astronouth7303:
This is especially true since no rc files are loaded for scripts, so it's not obvious where to stick things like "parse the environment into a more usable form".

scopatz:
Yeah, this would be great! 

If the xontrib exists, we could add a `--xontrib` option to the CLI to load xontribs on the command line.

For conda, we add the following to the `xonshrc` file, 

```
# >>> conda initialize >>>
# !! Contents within this block are managed by 'conda init' !!
import sys as _sys
from types import ModuleType as _ModuleType
_mod = _ModuleType("xontrib.conda",
                   "Autogenerated from $(/home/scopatz/miniconda/bin/conda shell.xonsh hook)")
__xonsh__.execer.exec($("/home/scopatz/miniconda/bin/conda" "shell.xonsh" "hook"),
                      glbs=_mod.__dict__,
                      filename="$(/home/scopatz/miniconda/bin/conda shell.xonsh hook)")
_sys.modules["xontrib.conda"] = _mod
del _sys, _mod, _ModuleType
# <<< conda initialize <<<
```

This is effectively creating a fake xontrib

anki-code:
It will be cool to have a xontrib installation pattern that excludes the editing of the `.xonshrc` file i.e.:
```python
xpip install xontrib-prompt-bar[autoloaded] xontrib-sh[autoloaded]  # etc
```

Issue #3205, opened 2019-07-09
[report] xonsh as a zipapp


AstraLuma:
Ok, so I've been using xonsh as a zipapp as my shell for a few weeks now. I wanted to walk through how I did it and my experiences with it.

Previous discussions on this topic were #2459 

## How

https://gitlab.com/doyouevenclick/xonsh

1. [Build script](https://gitlab.com/doyouevenclick/xonsh/blob/master/buildxonsh.py) in CI
2. Upload the resulting artifact to my [salt](https://www.saltstack.com/) master
3. Use salt to install it on all my machines

Some of the highlights I did:
* The shebang uses the specific system python instead of `env` to maximize reliability in the face of all the python environments (venvs, pyenv, conda, etc etc etc). This is accomplished via distro-specific builds (although several of them could probably be shared).
* `-I` (isolate) is used to minimize the amount of whatever from entering the xonsh python environment (system packages still do, and this actually uncovered a bug in debian's pysocks)
* The use of system and zips means that there's not an installable location, so a specific user site is created and added.
* Stripped out `__amalgam__.py` because I'm a developer and I don't need that kind of obscurity.

## Changes to xonsh

Only one core change was needed to allow full functionality, which I mention below.

Two things I added:

* A new [launcher script](https://gitlab.com/doyouevenclick/xonsh/blob/master/main.py) which:
  1. Adds the xonsh user site (`~/.local/xonsh`) to the path via the `site` module
  2. (previously) monkey-patched `xonsh.xontrib` with the fix mentioned below
  3. Replaces `sys.executable` with `~/.local/xonsh/bin/python` so that Python code running in xonsh uses an executable that will inherit (most) of xonsh's environment
* A new xontrib [`xonshzip`](https://gitlab.com/doyouevenclick/xonsh/blob/master/xonshzip.py) (which is pretty mandatory) which:
  1. Sets `xpip` correctly from the user site bin
  2. Adds aliases to manage the user site

## Problems

First: #3169 / #3170, basically the `xontrib` command expected to find `xontribs.json` on the file system (based on `__file__`). This was an easy fix, which got PR'd and released.

Second: #3171, the xsh importer fails on zips. It is easy to identify the problem but the fix involves a bunch of fresh engineering (made harder by the fact that Python's version, `zipimport`, is C and not really extensible/copyable). `xonshzip` works around this by copying the bundled xsh-based xontribs into the user site on creation.

## Performance

[...]

PR #3362, opened 2019-10-18
xonsh.AppImage with support for portable python and pip

anki-code:
Hello again! This is continue of #3357

Changes:

1. **The links was fixed to origin repo** because @niess has merged all [required fixes](https://github.com/niess/linuxdeploy-plugin-python/pull/11) to origin master of `niess/linuxdeploy-plugin-python`.

2. **The building process rewritten** from text to two examples: `Dockerfile` and `build.sh` to easy automation.

3. **Added python and pip support** and appropriate section with description. As result it allows install and using xontribs in portable mode!<br>

Also I've tried to build xonsh AppImage on 3.8 and it works! Review from @niess [was requested](https://github.com/niess/linuxdeploy-plugin-python/issues/14).

@scopatz is there a chance to make process of building AppImage continuous? Current size of AppImage around 26MB and I can't commit it to github but the link for "Trying now" follows to my repo with the file. It would be great if it will follow to fresh build.


niess:
Hello @anki-code and @scopatz . Concerning the automation of builds, currently any successful push to the master branch of linux-deploy-python triggers an upload of all the images to a [continuous release](https://github.com/niess/linuxdeploy-plugin-python/releases/tag/continuous), including xonsh. For this I am using a "simple" script provided by @probonopd for releasing AppImages. See for example [L5-6](https://github.com/niess/linuxdeploy-plugin-python/blob/master/.travis/after-success.sh#L5).

Yet it would make more sense indeed if the trigger would be a push to the xonsh repo itself, e.g. by including the AppImage build in your .travis.yml file.

Also the size of the AppImage could be further reduced by removing unused Python packages, etc. If Xonsh uses only a subset of the Python Standard Library?

P.S: @anki-code , I just fixed the travis build, for now ...

anki-code:
@niess thank you for fast feedback! I've changed the links to your build.
@scopatz we're ready to merge! :)

laloch:
HI @anki-code, could you please pick a different name for the news file? This one conflicts with #3373 and does not make any sense in the given context. See [Dev's Guide](https://xon.sh/devguide.html#changelog) for details.

astronouth7303:
What does the python environment look like under AppImage? By default, it looks like it's primarily about filesystem isolation similar to flatpak.

Is that filesystem inside the "sandbox" writable? By whom? If it is not writable, how are user-installed packages handled? (Does `xpip` work?)

We have traditionally not handled ecosystem packaging, outside of PyPI and Conda. What additional burden does this place on developers? I don't see additions to the rever/release scripts.

I do have UX concerns about shipping a shell with these sandboxy methods. Namely, I would expect an unaware shell to show the fileystem and system context from inside the sandboxy/container/etc, not the host system. I consider this severely limiting and not great for first-choice installation methods.

niess:
@astronouth7303 The AppImage is read only. Its not a container. There is no sandboxing though it seems to be possible e.g. using firejail, yet I never tried it. Basically you get pre-compiled binaries: Python and most of its dependencies (app specific system libs, Python modules, ...) that should run on most Linux provided that the image is properly built. Building the image is the tricky part for the moment I would say. Everything is packaged (compressed) in a single file (squashFS archive) that auto-mounts with fuse at runtime. For extra Python packages one can `pip install --user`. When fuse is not available one can also extract the image using the `--appimage-extract` option.

Personally I use an AppImage for distributing a plug and play relocatable Python environment including some custom packages. I find it very convenient and efficient for that purpose. You download a single file and it works out of the box without any installation. I also use the AppImage distribution of neovim and never had troubles with it. Yet I don't know if AppImage is relevant for xonsh?

niess:
@astronouth7303 To make things clear, the image mounts to a temporary path at runtime, e.g. under /tmp. The system locations (/usr etc.) are not overridden/hidden. The tricky part when building the image is to make sure that at runtime your executable locates its specific dependencies (e.g. shared libs, Python modules) inside the temporary mount point and does not use the system ones instead if available. For the Python AppImage this is done by overriding the RPATH of all binaries with relative locations using [patchelf](https://github.com/NixOS/patchelf) and the $ORIGIN tag.

anki-code:
> What does the python environment look like under AppImage? By default, it looks like it's primarily about filesystem isolation similar to flatpak.
> Is that filesystem inside the "sandbox" writable? By whom? 

Inside `xonsh.AppImage` installed portable python 3.7 and pip packages you were builded while creating AppImage. It's minimal packages required by xonsh. When you run `xonsh.AppImage` it creates read only SquashFS in /tmp/.mount-xonsh-[random]/ and then run xonsh. Next you can use host mashine tools and also add path to SquashFS to PATH and use tools from `xonsh.AppImage`. For example you can add python from AppImage to PATH and if you run `python` you'll get python 3.7 from xonsh.

>If it is not writable, how are user-installed packages handled? (Does `xpip` work?)
I'll test `xpip`. I've tested `pip`. You can add `$PIP_TARGET="/any/writable/path"` and install packages like `pip install <package>` in xonsh and it will ve in `/any/writable/path`.

> We have traditionally not handled ecosystem packaging, outside of PyPI and Conda. What additional burden does this place on developers? 

No burden in short. Just stay portable and rootless like most of pip packages.

>I don't see additions to the rever/release scripts.

What kind of addition you mean?

> I do have UX concerns about shipping a shell with these sandboxy methods. Namely, I would expect an unaware shell to show the fileystem and system context from inside the sandboxy/container/etc, not the host system. I consider this severely limiting and not great for first-choice installation methods.

AppImage — is not a docker or env or sandbox or something isolated container. The power of xonsh.AppImage is in that it could be executed without root access and in any place and you see real host environment. For example I'm using `xonsh.AppImage` on all my servers without installation. I just have an alias['xonssh']="scp xonsh server:~/ && ssh server -t ~/xonsh" and thats it! It's like an ssh with xonsh. I plan to open source this code. With `xonsh.AppImage` you can run xonsh everywhere without install and use it like ordinary shell.  Also I trying to run xonsh scripts on servers thru `xonsh.AppImage` without installing xonsh. I'm very happy to avoiding `bash`.

You absolutely right that it's not really about distribution or shipping. It's kernel for stand alone portable apps based on xonsh.

astronouth7303:
Ok, because AppImage does no sandboxing of any kind (no chroot, no pid/network namespacing), some of the concerns above do not apply.

Before I would call the AppImage ready to use, my requirement is that `xpip` works out of the box, and that xonsh's environment is not intrinsically entangled with the system python. (eg, I'm not a fan of just aliasing `xpip` to `pip --user` because then modules will be installed for both xonsh and the system python, and separating them again is considerable effort for the user.)

This sounds like similar goals and similar problems as my [zipapp work](https://github.com/xonsh/xonsh/issues/3205). They way I solved that was adding stuff for homedir-based xonsh environment additions. (This solution does not fully meet my requirement above because of the difficulty of injecting the code into xonsh, see #3275/#2151.)

In general, because of xonsh's role as a shell and being in python, how xonsh's python environment interacts with the system python/active virtual environments/conda/pyenv/etc etc etc is a significant concern. Different distribution methods have different trade-offs and handle this differently. I want to make sure that xonsh doesn't break from this.

anki-code:
I suggest to just add the info about building AppImage for xonsh in documentation and add continuous building to `xonsh/xonsh` or make `xonsh/appimage` repo.

If you suggested to think about glogal distribution xonsh.AppImage around the world with many additional supportings — for me it's not the part of this PR.

I spent around several weeks to make it works for me and I beleave that current AppImage is minimal vaiable PoC and it's useful and probably your own zipapp too. If somewone will want to add xpip and have enouth time it's great! He will have entrypoint.

If we kill this PR on "distribution problem over world" without any continuous building process... So, I just going to close it.

anki-code:
This PR has changes in documentation. Current version in master is outdated. Please, just merge it.

anki-code:
> It's like an ssh with xonsh — xonSSH :)

Here it is https://github.com/xonssh/xxh

anki-code:
> This PR has changes in documentation. Current version in master is outdated. Please, just merge it

What we are waiting?

melund:
@astronouth7303 Can you take the decision and merge this. I think this sounds great, but I will not be the user of this.

scopatz:
I am in favor of this PR because it is just documentation and we can proceed with incremental changes. But I'll still leave it to @astronouth7303 - who knows the most about this

astronouth7303:
I'm not in principle against more package formats, mostly about managing our complexity and support load.

That said, if the rest of xore is ok with `xpip` being broken under appimage for the time being, then I won't block this.

anki-code:
@astronouth7303 

> That said, if the rest of xore is ok
What exactly you mean? `xontrib` and `xonfig` are working.

`xonsh.AppImage` is like a portable binary. It creates read only sqashfs directory with xonsh. `xpip` is not working and it is not should working because there is no installation just running "the binary".

anki-code:
What we are waiting? :)

astronouth7303:
(xore is the group of xonsh core developers, because we like our x's)

Assuming you'll work on the concerns I outlined above (namely about `xpip`), and this doesn't place any additional burden on xore, I'm ok with merging this in. (Especially since AppImage stuff already exists in the repo.)

anki-code:
Great! Let's do it )

anki-code:
What we are waiting? :)

anki-code:
Could it be merged before Christmas? 🎅

melund:
Sorry about that @anki-code. I guess all core devs have been very occupied the last months. I thought this went in a long time ago.

anki-code:
@astronouth7303 

> This sounds like similar goals and similar problems as my zipapp work. 

With [xxh 0.4.0](https://github.com/xxh/xxh) there is an ability to create any protable shell entrypoint. If you are confident with your zipapp you can try to create `xxh-shell-xonsh-zippapp` entrypoint like [xxh-shell-xonsh-appimage](https://github.com/xxh/xxh-shell-xonsh-appimage) entrypoint.

PR #3373, opened 2019-10-25
fix(signals): Fix SIGSTP not an attribute on Windows

BYK:
I kept getting `AttributeError: module 'signal' has no attribute 'SIGTSTP'` exceptions when I `^C`d out of a process and turns out there are a few places where we forgot to protect `signal.SIGSTP` on Windows.

This patch fixes those remaining places.

Follow up to #3344.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,10 +4,6 @@ Xonsh Change Log
 
 .. current developments
 
-**Fixed:**
-
-* Fix Ctrl-C event causing Atribute error on Windows. (for reals this time)
-
 v0.9.13
 ====================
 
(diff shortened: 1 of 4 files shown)


BYK:
No idea how to satisfy `lint/xonsh — No news item` :(

laloch:
Hi @BYK, revert the CHANGELOG.rst change, copy news/TEMPLATE.rst to news/sigstp.rst and fill the news entry there.

scopatz:
Thanks @BYK! Also, we have devdocs for the news here: https://xon.sh/devguide.html#changelog

BYK:
@gforsyth - Let's just say I had some busy times lately and couldn't use xonsh as my daily driver. Since that has changed, I'm hoping to be more present :)

PR #3408, opened 2019-12-19
Predict several REPLs as unthreadable

SirNickolas:
[Elvish](https://elv.sh) is another nice scripting language and shell.
[GHCi](https://wiki.haskell.org/GHC/GHCi) is Haskell’s REPL. (`ghci <filename>` enters interactive mode after loading the module, thus `predict_help_ver` is used.)
[Julia](https://julialang.org) is a language designed for numerical computations and data analysis. It has a REPL too.


SirNickolas:
I’m not sure, does that mean I must add a news entry? `CONTRIBUTING` says, “Pull requests will _often_ have CHANGELOG entries associated with,” so I thought it’s not required for such minor changes.

gforsyth:
Thanks for putting this in @SirNickolas !  I'm going to merge it -- the test failures are due to a previous news entry that was incorrectly formatted and I'll fix that separately.

PR #3411, opened 2019-12-21
Fix appimage news entry

gforsyth:
Restoring missing entries to the news entry from the appimage PR since it's breaking CI on all subsequent PRs.

We really need to move the news checker to a separate place.


gforsyth:
This was covered in #3419

PR #3413, opened 2019-12-26
Ls color - Colorize file names in subproc mode command lines.

bobhy:
New feature -- colorize file names used as arguments in command lines in subprocess mode.
Uses colors defined in $LS_COLORS (and updates when it changes).
Depends on prompt toolkit 2 and pygments.


bobhy:
OK, the feature is ready for your review, passing all tests except the xonsh.xonsh which I see is failing for everyone.

I had lots of difficulties with flake8 and black, and took a somewhat heavy hammer to flattening them.  Let's talk about that, as well as the feature itself.

gforsyth:
Hey @bobhy -- thanks for putting this in!  And sorry for the delay in getting to it.

I agree the hammer is somewhat heavy -- would you be willing to split the PR up and submit the changes to `black` configuration separately?

bobhy:
Can do, but won't get to it till next week.

Bob

bobhy:
@gforsyth, I've started looking at the black changes and am literally embarrassed to submit the change to `setup.cfg` and 'pyproject.toml` -- too much repeating of elements of the flake-8 config. But I am a noob with these tools and perhaps someone can suggest the smart way to handle this.
There's a year old standing issue with pytest-flake8 [pytest-flake8/22](https://github.com/tholo/pytest-flake8/issues/22) about the plugin not reading flake8 config, and no indication an improvement is pending.  The naive solution is to copy the whole thing in `setup.cfg`. If I ran flake-8 as a separate step in `circleci/config.yml`, I think I could do away with the dependency on pytest-flake8 plugin.  ??

Re black, ci seems to be assuming that the project was already formatted with black, but how do you all developers do this on each checkin without a config file `pyproject.toml` in the project?    I'm using vscode and found it needed this file to configure black as my code formatter.  

Anyway, I regard this redundancy as unnecessarily unmaintainable and need guidance!

bobhy:
OK, this and the related #3431 are split and ready for review.

scopatz:
Thanks @bobhy!

bobhy:
de nada.

PR #3426, opened 2020-01-21
Update pun classification

mikegraham:
fish is a name for a three classes of chordates.


astronouth7303:
Unclear what the pun is? Like, yes fish are, well, fish.

mikegraham:
It's the same basic pun as bash -- word ending in sh.

It's at least as punny as plumbum.

anki-code:
This PR looks old and should be closed.

PR #3428, opened 2020-01-28
3427 fix vc prompt color

### LINKED ISSUE #3427: [color] in .gitconfig should be stripped in the prompt branch functions
In clean install with no `.xonshrc`, when I change into a git repo, my prompt looks like:

prompt

Is there some setting I'm missing somewhere?

## xonfig

<details>

```
| Python           | 3.7.6                |
| prompt toolkit   | 3.0.0                |
| pygments         | 2.5.2                |
```

</details>

## Expected Behavior
I think something like this (without the ansi escape codes):
```
mcrowe@mike-XPS-15-9560 ~/.dotfiles/bash_it [master] $
```

drmikecrowe:
Fixes #3427

### DIFF
--- a/xonsh/prompt/vc.py
+++ b/xonsh/prompt/vc.py
@@ -8,6 +8,7 @@ import queue
 import builtins
 import threading
 import subprocess
+import re
 
 import xonsh.tools as xt
 
@@ -50,6 +51,8 @@ def get_git_branch():
     t.join(timeout=timeout)
     try:
         branch = q.get_nowait()
+        remove_ansi = re.compile(r'(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]')
+        branch = remove_ansi.sub('', branch or "")
     except queue.Empty:
         branch = None
     return branch
--- a/xonsh/prompt/vc.py
+++ b/xonsh/prompt/vc.py
@@ -58,7 +58,9 @@ def get_git_branch():
     t.join(timeout=timeout)
     try:
         branch = q.get_nowait()
-        branch = RE_REMOVE_ANSI.sub("", branch or "")
+        # branch = RE_REMOVE_ANSI.sub("", branch or "")
+        if branch:
+            branch = RE_REMOVE_ANSI.sub("", branch)
     except queue.Empty:
         branch = None
     return branch
--- a/xonsh/prompt/vc.py
+++ b/xonsh/prompt/vc.py
@@ -11,6 +11,13 @@ import subprocess
 import re
 
 import xonsh.tools as xt
+from xonsh.lazyasd import LazyObject
+
+RE_REMOVE_ANSI = LazyObject(
+    lambda: re.compile(r"(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]"),
+    globals(),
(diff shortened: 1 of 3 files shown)


xonsh/prompt/vc.py
```diff
@@ -50,6 +51,8 @@ def get_git_branch():
     t.join(timeout=timeout)
     try:
         branch = q.get_nowait()
+        remove_ansi = re.compile(r'(?:\x1B[@-_]|[\x80-\x9F])[0-?]*[ -/]*[@-~]')
```
gforsyth: This will compile the regex every time we hit a VC dir which we probably want to avoid.  

You can see an example of how to make it lazily evaluated here: https://github.com/xonsh/xonsh/blob/master/xonsh/built_ins.py#L179

drmikecrowe: Done

gforsyth:
Thanks for the updates @drmikecrowe -- Can you pull in teh changes from master to get past the spurious CI failures?  I think there are also some issues with branch names that this might mangle a little bit.

drmikecrowe:
@gforsyth I think that fixes it? Do the tests always fail on PR's (since they aren't `master`)?

gforsyth:
Thanks for updating @drmikecrowe !  Looks like the remaining failure is a legitimate one.  Our current state on many of the PRs notwithstanding, tests should pass on a PR before it is merged to master.

Looks like some of the ANSI character substitution may be having unexpected side-effects:
```
=================================== FAILURES ===================================
___________________________ test_vc_get_branch[git] ____________________________

test_repo = {'dir': '/tmp/tmptuphwz69', 'name': 'git'}
xonsh_builtins = <module 'builtins' (built-in)>

    def test_vc_get_branch(test_repo, xonsh_builtins):
        xonsh_builtins.__xonsh__.env = Env(VC_BRANCH_TIMEOUT=2)
        # get corresponding function from vc module
        fun = "get_{}_branch".format(test_repo["name"])
        obs = getattr(vc, fun)()
        if obs is not None:
>           assert obs == VC_BRANCH[test_repo["name"]]
E           AssertionError: assert '' == 'master'
E             + master

/home/circleci/project/tests/test_prompt.py:243: AssertionError
```

You can run the tests locally (and limit the scope of those run for speed) by running `pytest tests/test_prompt.py`

drmikecrowe:
@gforsyth I get those errors on master as well.  Do you see those?

drmikecrowe:
@gforsyth OK, I finally have this sorted.  A couple of things really jump out at me:

* For new developers, the amalgamation is a bit confusing as to how it impacts development.  I'm still not sure what the best way is handle this
* I think I got crossed up when I tried to edit my files and run the tests with conflicts with the install

gforsyth:
Awesome! I'll try to take a look soon (on a tiny keyboard right now). 

Much agreed about the confusion stemming from amalgamation. We have done some less developer friendly things in the name of startup time -- there's a discussion to be had about whether the particularities of amalgamation are worth it or if our efforts are better focused elsewhere.

In the interim, if you have any thoughts on how to improve developer documentation, please do let us know!

gforsyth:
Thanks for putting this in, @drmikecrowe!

PR #3431, opened 2020-02-02
black-fix - split from ls_color for separate review

bobhy:
Split out from ls_color for independent review.


bobhy:
Motivation for this PR is that I, as a noob using VS code as my IDE, kept getting CI failures on a feature PR I was submitting which was already passing pytest -q.  Failures were coming from pytest --flake8 (which I did not know to run locally) and from black "would reformat", which I had no expectation even mattered.

This fix a) lets me configure black as the code formatter for python in vscode, so all my edits will automagically be formatted; and b) copies the flake8 exceptions into a separate section in `setup.cfg` so flake-8 tool will report same issues when run interactively from IDE as pytest --flake8 would complain about eventually.  

It is a thousand pities pytest-flake8 and flake-8 tool itself require redundant configuration, but I can't  figure out a way to work around that -- suggestions solicited.

And I'd suggest a few more words in the developer's guide to the effect that black is now mandatory and to run pytest --flake8 -- that have saved me time, for sure.

bobhy:
Please look at this CI failure, restricted to build_36.  Maybe it's environmental?  I didn't change anything anywhere near `subprocess.py` (this time).

```
        # Optimization: If we are not worried about timeouts, we haven't
        # started communicating, and we have one or zero pipes, using select()
        # or threads is unnecessary.
        if (timeout is None and not self._communication_started and
            [self.stdin, self.stdout, self.stderr].count(None) >= 2):
            stdout = None
            stderr = None
            if self.stdin:
                self._stdin_write(input)
            elif self.stdout:
>               stdout = self.stdout.read()
E               Failed: Timeout >10.0s

../miniconda/envs/py36-xonsh-test/lib/python3.6/subprocess.py:830: Failed
----------------------------- Captured stdout call -----------------------------
```

bobhy:
Hmm, now ci failure is spreading from build_36 to build_37... I don't understand!

gforsyth:
Hey @bobhy -- I'll try to take a look at this soon.  Just an FYI that those timeout errors happen from time-to-time and are unrelated to your PR.  I restarted the circle CI checks and we'll see what happens.

I don't think I understood initially what you were claiming re: `flake8` configuration and I agree that the duplication is perhaps inelegant.  I'll try to think about this a bit more.

Thanks for putting this in!  And for the extra work splitting it off the other PR!

astronouth7303:
@gforsyth

gforsyth:
Thanks for the ping, @astronouth7303 !  

Hey @bobhy -- on further reflection, sometimes we just need to keep things simple and I think, given that other projects have hit this issue, that the simple solution is to duplicate the config.  

If we come up with something better we can always change it.  Thanks for putting this in and for splitting things up!

bobhy:
De nada.  Glad to make the fix.

PR #3432, opened 2020-02-03
autovox: Fix algorithm to cd correctly

astronouth7303:
autovox does not currently handle cd and manual activation correctly.

To reproduce:

1. `vox activate myenv`
2. `cd mydir` (Any directory without an associated venv)

What should happen: The venv is still `myenv`

What does happen: venv is deactivated.

This happens because it compares the current venv with the venv associated with the new dir. What should happen is that the venv associated with the old dir should be compared to the venv of the new dir.

### DIFF
--- a/news/autovox-cd.rst
+++ b/news/autovox-cd.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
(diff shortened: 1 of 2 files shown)


gforsyth:
Thanks @astronouth7303 !

PR #3447, opened 2020-02-17
Tutorial: note about $UPDATE_OS_ENVIRON

anki-code:
Hi!
The description of `${...}` is looks like that it is the synonym of `os.environ`. But it's not exectly true.
In this PR added the note about $UPDATE_OS_ENVIRON to the tutorial. It is great to know about special flags in the beginning of the tutorial and save time to understanding why:
```
# $TEST='test'
# 'TEST' in ${...}
True
# 'TEST' in os.environ
False
```

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -136,6 +136,11 @@ variable in Python.  The same is true for deleting them too.
 
 Very nice.
 
+.. note::
+
+   To update ``os.environ`` when the xonsh environment changes set 
+   `$UPDATE_OS_ENVIRON <envvars.html#update-os-environ>`_ to ``True``.
+
 The Environment Itself ``${...}``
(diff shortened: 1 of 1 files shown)


melund:
Great addition @anki-code. It can be a confusing for new users if they use os.subprocess in their python/xonsh scripts.

Just a minor thing. Can you make the link like this instead, so you use sphinx's reference system :
```
:ref:`$UPDATE_OS_ENVIRON <update_os_environ>`
```

anki-code:
@melund
> Can you make the link like this instead, so you use sphinx's reference system 

Done. Could you please do a little thing too? Take a short look at [xxh](https://github.com/xonssh/xxh) and may be put the star to it? 🙏 😃

melund:
Thanks @anki-code

PR #3451, opened 2020-02-23
Add useful tricks with Path object to the tutorial

anki-code:
Hi!

After reading the tutorial about ``g`...` `` and `p"..."` there is a question: how to globbing certain path because `` g`{mypath}/**` `` is not working (and gp/gf/gpf too). I've added info about one elegant way how to do it with Path object. And a bit more.

Thanks!


astronouth7303:
Thanks! Looks good!

@xonsh/xore do we want to add f-globs as an option?

PR #3465, opened 2020-03-01
Exitcode

### LINKED ISSUE #3309: Xonsh returns zero exit code when a command fails in a script
I haven't had time to triage it yet, but here is what I executed:
```
$ xonsh ci/build.xsh && echo "OK"
...
Interactive Fortran. Experimental prototype, not ready for end users.
  * Use Ctrl-D to exit
  * Use Enter to submit
Try: function f(); f = 42; end function
>>> Input:
fxunction f(); f = 42; end function
input:1:11 syntax error: token type 'identifier' is unexpected here
fxunction f(); f = 42; end function
          ^
terminate called after throwing an instance of 'LFortran::ParserError'
  what():  syntax error
Aborted (core dumped)
OK
```
The C++ code exited with an uncaught exception which generates a non-zero exit code as can be verified by:
```
$ ./lfortran < ../../../src/bin/example_input.txt  && echo "OK"
Interactive Fortran. Experimental prototype, not ready for end users.
  * Use Ctrl-D to exit
  * Use Enter to submit
Try: function f(); f = 42; end function
>>> Input:
fxunction f(); f = 42; end function
input:1:11 syntax error: token type 'identifier' is unexpected here
fxunction f(); f = 42; end function
          ^
terminate called after throwing an instance of 'LFortran::ParserError'
  what():  syntax error
Aborted (core dumped)
$
```
And yet Xonsh returns zero exit code and the "OK" gets printed.

This is a serious issue as I use Xonsh on a CI and I need it to fail when an exception occurs in my C++ program.

Here is the `ci/build.xsh` (as you can see I enabled the equivalent of `set -e`):
```
#!/usr/bin/env xonsh

$RAISE_SUBPROC_ERROR = True
trace on

$lfortran_version=$(cat version).strip()
cd lfortran-$lfortran_version

cd test-bld
if $WIN == "1":
    ./src/lfortran/tests/Release/test_llvm -s
else:
    ./src/lfortran/tests/test_llvm -s
ctest --output-on-failure
src/bin/lfortran < ../src/bin/example_input.txt
cd ..
```

scopatz:
This should close #3309.

The `os.abort()` in the test is the only reliable way I have found to produce a negative exit code with a compiler.


scopatz:
@xonsh/xore, this is ready for review!

PR #3466, opened 2020-03-02
trace subproc: fix pipeline case

anki-code:
Hi! It's a simple fix. Please merge after tests or restart the tests if it fails by other reasons.


scopatz:
Thanks!

PR #3467, opened 2020-03-02
Create xonshpytest.yml

gforsyth:
Trying out github actions for CI matrix


gforsyth:
Hey @xonsh/xore -- any thoughts on this approach?  I was mostly playing with github actions, but the matrix over versions + operating systems is nice.  Only thing it wouldn't cover is doc builds (still on travis for the moment)

melund:
I think it looks great. Nice and simple. I don't care much which we use. But going for something simple and well integrated like gh actions is a good idea. So +1 from me

scopatz:
Yep! This is awesome! Thanks a ton @gforsyth! 

My only comment here is that we should probably reinstate the OSX builds on travis until the GH actions work on OSX too.

gforsyth:
> My only comment here is that we should probably reinstate the OSX builds on travis until the GH actions work on OSX too.

I'm bummed because they _were_ working, but I kept getting timeout failures.  I'll re-enable OSX on travis for the moment and then we can track down flaky tests afterwards

scopatz:
The circle error is just because we are no longer using it :wink:

gforsyth:
I just deleted all of our unused webhooks (codecov, circle, azure) so they should stop showing up in PRs from now on

Issue #3380, opened 2019-11-01
xonsh incorrectly reorders the test for the pytest


andry81:
mixed (`.py`+`.xsh`) pytests runs in unpredictable or chaotical order.

The main issue:
https://github.com/pytest-dev/pytest/issues/6114

The issue which might be connected too:
https://github.com/pytest-dev/pytest/issues/6113

## xonfig

<details>

```
+------------------+-----------------------------------------------+
| xonsh            | 0.9.12                                        |
| Git SHA          | 50155c66                                      |
| Commit Date      | Oct 3 15:53:08 2019                           |
| Python           | 3.7.3                                         |
| PLY              | 3.11                                          |
| have readline    | False                                         |
| prompt toolkit   | 2.0.10                                        |
| shell type       | prompt_toolkit2                               |
| pygments         | None                                          |
| on posix         | False                                         |
| on linux         | False                                         |
| on darwin        | False                                         |
| on windows       | True                                          |
| on cygwin        | <xonsh.lazyasd.LazyBool object at 0x027659B0> |
| on msys2         | <xonsh.lazyasd.LazyBool object at 0x02765870> |
| is superuser     | False                                         |
| default encoding | utf-8                                         |
| xonsh encoding   | utf-8                                         |
| encoding errors  | surrogateescape                               |
+------------------+-----------------------------------------------+
```

</details>

## Expected Behavior
correct order independent to the test file extension

## Current Behavior
random tests order

## Steps to Reproduce
see the above link

scopatz:
Is there anything that we should do here?

andry81:
I don't see how the file is related to the issue is changed: https://github.com/pytest-dev/pytest/issues/6114#issuecomment-548786834
Should I retest it while a change is not take a place?

scopatz:
I think that @melund wrote most of https://github.com/xonsh/xonsh/blob/master/xonsh/pytest_plugin.py so maybe he knows

melund:
I can't remember any of it. I will try to one investigate why we choose to do this is the first place. Otherwise, I will definitely remove that code.

melund:
I think the bad attempt at reordering test was done because we had a bug once where certain test could not run after each other. I belive that has since been fixed.

andry81:
I just commented related line of code: https://github.com/xonsh/xonsh/blob/bb201d77bb0ad7e439739ae008a19dae65de6652/xonsh/pytest_plugin.py#L17
[...]

PR #3468, opened 2020-03-02
Remove test reordering in pytest plugin

melund:
The xonsh pytest plugin no longer messes up the test order for pytest. 
The original attempt to run xsh test first was a workaround 
for bug where certain test would not run after each other. This workaround
is no longer needed. 

fixes: #3380


scopatz:
Thanks @melund! It looks like there are a lot of test failures now :cry:

melund:
Ahh. So we still have problems with the test suite and the fact some test can't run after each other. I was hoping it was fixed. 

Anyway, I reimplement the ordering in more predictable way, so we can still close #3380

scopatz:
This looks great to me! Thanks!

andry81:
I don't see how this changes the reorder. On mine side nothing has changed. I just rewrited `pytest_plugin.py` and saw the same order.

melund:
@andry81 Maybe you reiterate what the problem was. 

You are right that it still reorders the tests by moving the `test_*.xsh` files first. That is unfortunately necessary because the fixtures used in the normal xonsh `test_*.py` files doesn't clean up after them self. Thus, the `test_*.xsh` files can't run after some of those tests. 

I thought the problem was a random order of the tests. I could see how the old implementation could cause a random order. Now it should only move the `xsh` test first.

andry81:
> That is unfortunately necessary because the fixtures used in the normal xonsh test_\*.py files doesn't clean up after them self. Thus, the test_\*.xsh files can't run after some of those tests.

Why do you need to change the test order to fix tests? May be the problem was in a badly designed tests?

> I thought the problem was a random order of the tests. I could see how the old implementation could cause a random order. Now it should only move the xsh test first.

That behaviour was before the change.

melund:
> Why do you need to change the tests order to fix tests? May be the problem was in a badly designed tests?

Sure. The fixtures which emulate a xonsh environment in pytest is a bit extensive. I don't know why it doesn't clean up nicely. I suspect it is pytest problem, but I really don't know. I hope this change fixes the most important things for you.

andry81:
> I hope this change fixes the most important things for you.

But is does not fix the initial problem. And actually I not much care, I just don't relie on pytest collection phase, instead I run each test in a separate pytest process to avoid any problems with plugins.

melund:
@andry81 Sorry, but what was the initial problem. I can see that a randomization of the tests would be a problem, but I can't see that moving certain test first is a problem?

andry81:
> @andry81 Sorry, but what was the initial problem. I can see that a randomization of the tests would be a problem, but I can't see that moving certain test first is a problem?

Under the `random tests order` I've meant the dependency on something else than on a file name or on a test case name. What the kind of randomization is not actually matter because externally it is observable as a "random" order instead of predefined order by the pytest.

melund:
Ok. But what did break for you? Currently, the order is still kept within the group of all `test_*.py` files, and the `test_*.xsh` files

andry81:
> Ok. But what did break for you? Currently, the order is still kept within the group of all `test_*.py` files, and the `test_*.xsh` files

the pytest order without plugin:
```
01_test.py -> test_01_foo()
02_test.xsh -> test_02_foo()
```
with plugin:
```
02_test.xsh -> test_02_foo()
01_test.py -> test_01_foo()
```

melund:
Sorry @andry81 I am bit slow here. But I thought it only ran the `test_*.xsh` if the plug-in was installed. So how did you see the different ordering without the plug-in. 

Also, what does the fact that `xsh` tests run first break?

andry81:
> Sorry @andry81 I am bit slow here. But I thought it only ran the `test_*.xsh` if the plug-in was installed. So how did you see the different ordering without the plug-in.

If you create .py and .xsh tests in the same directory, then pytest will run both.

> Also, what does the fact that xsh tests run first break?

Because it should not in first place. Pytest has it's own order for that by name of a test case function. Plugin changes the default pytest order.

andry81:
> But I thought it only ran the `test_*.xsh` if the plug-in was installed.

And I think you should not do like that, because it may break name conventions has configured somewhere else except of your plugin code:
http://doc.pytest.org/en/latest/example/pythoncollection.html#changing-naming-conventions

PR #3473, opened 2020-03-11
Fix behavior of cd -P and dot-dot paths

Gobbel2000:
Calling abspath() before realpath() suppresses some information about symbolic links in some cases. To illustrate this case here is what is expected to happen (and what happens in the standard /usr/bin/cd) with `$PROMPT = "{cwd} >>> "`:

```
/ >>> ls -l lib
lrwxrwxrwx 1 root root 7 13. Nov 17:23 lib -> usr/lib
/ >>> cd lib
/lib >>> cd -P ..
/usr >>> 
```
But this is what the current version brings:
```
/ >>> ls -l lib
lrwxrwxrwx 1 root root 7 13. Nov 17:23 lib -> usr/lib
/ >>> cd lib
/lib >>> cd -P ..
/ >>> 
```
This change implements the first behavior instead.

### DIFF
--- a/news/cd_symlinks.rst
+++ b/news/cd_symlinks.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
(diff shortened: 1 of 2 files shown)


scopatz:
Awesome! Thanks for the fix @Gobbel2000!

PR #3479, opened 2020-03-15
feat: add HISTCONTROL(erasedups) support to sqlite backend

jnoortheen:
fixes part of !3478

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1004,8 +1004,10 @@ def DEFAULT_DOCS():
             "that determine what commands are saved to the history list. By "
             "default all commands are saved. The option ``ignoredups`` will not "
             "save the command if it matches the previous command. The option "
-            "'ignoreerr' will cause any commands that fail (i.e. return non-zero "
-            "exit status) to not be added to the history list.",
+            "``ignoreerr`` will cause any commands that fail (i.e. return non-zero "
+            "exit status) to not be added to the history list. The option "
+            "``erasedups`` will remove all previous commands that matches and updates the frequency. "
+            "Note: ``erasedups`` is supported only in sqlite backend).",
             store_as_str=True,
         ),
         "IGNOREEOF": VarDocs("Prevents Ctrl-D from exiting the shell."),
--- a/xonsh/history/sqlite.py
+++ b/xonsh/history/sqlite.py
@@ -12,6 +12,10 @@ import time
 from xonsh.history.base import History
 import xonsh.tools as xt
 
+XH_SQLITE_CACHE = threading.local()
+XH_SQLITE_TABLE_NAME = "xonsh_history"
+XH_SQLITE_CREATED_SQL_TBL = "CREATED_SQL_TABLE"
+
 
 def _xh_sqlite_get_file_name():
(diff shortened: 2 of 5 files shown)


jnoortheen:
the circleci check fails, because of the different version of black. It is not pinned in requirements file. I recommend using `poetry` for managing the dependencies.

xonsh/history/sqlite.py
```diff
@@ -12,6 +12,10 @@
 from xonsh.history.base import History
 import xonsh.tools as xt
 
+CACHE = threading.local()
+TABLE_NAME = "xonsh_history"
+CREATED_SQL_TBL = "CREATED_SQL_TABLE"
```
scopatz: Please use more unique names here, since this sub-packages is using amalgamate.  For example, something like `XH_SQLITE_CACHE`

scopatz: Please use more unique names here, since this sub-packages is using amalgamate.  For example, something like `XH_SQLITE_CACHE`

jnoortheen: I have updated variables.

jnoortheen:
is there any way to retry the failure?

gforsyth:
> is there any way to retry the failure?

Hey @jnoortheen -- I've restarted it twice and it's still failing on 3.8

jnoortheen:
what could be the reason for failure? 
the tests for 'tests/test_main.py', 'tests/test_ptk_highlight.py',  be retried automatically? we can update the pytest to do that.

xonsh/environ.py
```diff
@@ -1004,8 +1004,9 @@ def DEFAULT_DOCS():
             "that determine what commands are saved to the history list. By "
             "default all commands are saved. The option ``ignoredups`` will not "
             "save the command if it matches the previous command. The option "
-            "'ignoreerr' will cause any commands that fail (i.e. return non-zero "
-            "exit status) to not be added to the history list.",
+            "``ignoreerr`` will cause any commands that fail (i.e. return non-zero "
+            "exit status) to not be added to the history list. The option "
+            "``erasedups`` will remove all previous commands that matches and updates the frequency.",
```
scopatz: Erasedups seems like it was only implemented for the SQLite backend.  Can you please add that as a comment in the docs here

xonsh/history/sqlite.py
```diff
@@ -12,6 +12,10 @@
 from xonsh.history.base import History
 import xonsh.tools as xt
 
+XH_SQLITE_CACHE = threading.local()
+XH_SQLITE_TABLE_NAME = "xonsh_history"
+XH_SQLITE_CREATED_SQL_TBL = "CREATED_SQL_TABLE"
```
scopatz: It seems like these last two would be better as local variables in the function

scopatz: Actually nm, I see what is happening here

scopatz:
Thanks @jnoortheen! The code looks good here.  Just the one docs change from me

xonsh/environ.py
jnoortheen: I have added a note section.

scopatz:
Thanks!

PR #3484, opened 2020-03-20
Add fish-like command abbreviations

### LINKED ISSUE #3471: fish-shell like abbr-expansion support
## xonfig
I am currently moving from fish to xonsh as my main linux-shell. I miss the fish's feature of [abbr](https://fishshell.com/docs/current/cmds/abbr.html) command. It will be cool to have this on xonsh. It will be aliases that are more explicit.

laloch:
Closes #3471.
Expands abbreviations from `abbrevs` dictionary on <kbd>Space</kbd> key press. I did not ~~(and will not)~~ implement expanding abbreviations on <kbd>Return</kbd>, ~~because I don't like it - it is too dangerous in my opinion.~~

To test it:
```
$ xontrib load abbrevs
$ abbrevs["gst"] = "git status"
$ gst # hit <space> and "gst" should get epanded into "git status"
```

### DIFF
--- a/xontrib/abbrevs.py
+++ b/xontrib/abbrevs.py
@@ -1,4 +1,8 @@
-"""Command abbreviations."""
+"""
+Command abbreviations.
+
+This expands input words from `abbrevs` disctionary as you type.
+"""
 
 import builtins
 from prompt_toolkit.filters import IsMultiline
--- a/xontrib/abbrevs.py
+++ b/xontrib/abbrevs.py
@@ -1,0 +1,33 @@
+"""Command abbreviations."""
+
+import builtins
+from xonsh.platform import ptk_shell_type
+
+__all__ = ()
+
+builtins.__xonsh__.ctx["abbrevs"] = dict()
+
+
+def expand_abbrev(buffer):
+    word = buffer.document.get_word_before_cursor()
+    if "abbrevs" not in builtins.__xonsh__.ctx.keys():
+        return
+    abbrevs = builtins.__xonsh__.ctx["abbrevs"]
+    if word in abbrevs.keys():
+        buffer.delete_before_cursor(count=len(word))
+        buffer.insert_text(abbrevs[word])
+
+
+@events.on_ptk_create
+def custom_keybindings(bindings, **kw):
+
+    if ptk_shell_type() == "prompt_toolkit2":
+        handler = bindings.add
+    else:
+        handler = bindings.registry.add_binding
+
+    @handler(" ")
+    def space(event):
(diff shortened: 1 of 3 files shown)


astronouth7303:
Can we make this an external project? Is there sufficient demand that we package it as part of xonsh itself?

laloch:
@astronouth7303, good question. I made it this way because of https://github.com/xonsh/xonsh/issues/3471#issuecomment-597225056.
Regarding the demand/usefulness, I would recommend this over normal aliases, but that's just my opinion.

astronouth7303:
Not a huge deal, just kinda don't think that xonsh should become a kitchen sink of features. (This is why xontribs exist, after-all.)

laloch:
Not a big deal for me either. I'll gladly make it an external xontrib. Let's wait for what others are going to say.

gforsyth:
I'm fine with this being a bundles xontrib or an external xontrib.  Probably a slight preference towards bundled, since @laloch is already part of xore (and having it internal makes it easier for the rest of us to continue maintenance, which is admittedly also a downside, but the world is confusing)

melund:
I am also in favor of a bundled xontrib. Sounds like a really useful functionality...

laloch:
I changed my mind and added expansion on <kbd>Return</kbd> key press. It's not more dangerous than regular aliases after all.

laloch:
This is now ready to land.

xontrib/abbrevs.py
```diff
@@ -0,0 +1,63 @@
...
+import builtins
+from prompt_toolkit.filters import IsMultiline
+from prompt_toolkit.keys import Keys
+from xonsh.platform import ptk_shell_type
+from xonsh.tools import check_for_partial_string
+
+__all__ = ()
+
+if "abbrevs" not in builtins.__xonsh__.ctx.keys():
+    builtins.__xonsh__.ctx["abbrevs"] = dict()
```
scopatz: This shouldn't be in the execution context, because it might overlap with variable names.  Rather it should be similar to the `events` or `aliases` dict which lives on the session (`builtins.__xonsh__.abbrevs`), but is linked to the `builtins.abbrevs` variable with a `DynamicAccessProxy`,  seen here https://github.com/xonsh/xonsh/blob/master/xonsh/built_ins.py#L1426

scopatz:

xontrib/abbrevs.py
laloch: Done. Thanks for the review.

scopatz:
Awesome! Thanks!

PR #3490, opened 2020-03-27
abbrevs: only trigger completion if the word is preceded by space

laloch:
The abbrevs expansion may be especially annoying if triggered after dash or equality sign. Let's reduce it to space-delimited word completion.
No news entry needed - alters unreleased code.


melund:
Looks good

PR #3494, opened 2020-04-01
Do not handle ampersand as an argument in subprocess mode

### LINKED ISSUE #3417: Impossible to escape trailing ampersand in subprocess mode
## xonfig
```
| Python           | 3.6.10               |
| prompt toolkit   | 3.0.2                |
| pygments         | 2.5.2                |
| distro           | linuxmint            |
```

## The situation
Assuming I'm in a directory with a file named `&` (this actually happened to me).
I am trying to examine it using `file &` or `file "&"` or `file '&'`

## Expected Behavior
By escaping the `&` through quotes, the single argument `&` should be passed to `file`.
The `bash` shell does it like this and outputs: `&: ASCII text`

## Current Behavior
`file` is started as a job with no argument
`[1]+ running: file & & (8218)`

Note that this is also very inconvenient if you want to `grep -r '&'` or something similar.

## Possible fixes
Although I just installed xonsh and are highly unfamiliar with the code, I took a quick look and came to the conclusion that it may be a bad idea not to treat `&` like `|`.
The command line input `file foo | less` will be represented as command sequence `(['file', 'foo'], '|', ['less'])` while the command line input `file bar '&'` will be represented as `(['file', 'bar', '&'])`, making the ampersand indistinguishable from a regular argument.

laloch:
This removes ambiguity of ampersand string literal in subprocess mode. Allows for passing a single quoted ampersand as the last subprocess argument.

Fixes #3417.


scopatz:
Thanks @laloch! Can you please add some parser tests as well as some docs about how to pass it in? Thanks!

laloch:
This should not really change anything functionally. It just fixes a bug. Previously it was not possible to `rm "&"`. Now it is.

scopatz:
yeah, but we should add a test so that the bug doesn't come back :wink:

laloch:
Yeah, right, but this should not require any documentation changes.

laloch:
What the heck is wrong with the AppVeyor? It [does not seem](https://ci.appveyor.com/project/AnthonyScopatz/xonsh-1m7j9/builds/32011992/job/sdj86r2hca2jc0gv#L732) to run "my" version of the parser.

laloch:
Travis broken due to #3506.

scopatz:
@laloch - can you merge from master to get the sphinx docs fix, please?

laloch:
I can see what the problem with AppVeyor is - there's no `echo` builtin.

scopatz:
yeah, I thought for some of the tests though we added an echo alias, but maybe not!

laloch:
Now this was embarrassing. It is, however, as green as it gets now.

melund:
So now this can get into the release :)

PR #3497, opened 2020-04-01
Unthread xdg open

### LINKED ISSUE #3496: xdg-open doesn't go to background by itself
## Current Behavior
When using `xdg-open doc.pdf`, okular (my default pdf reader) opens as expected. Terminal is blocked until okular is closed. Unless ` &` is used explicitly, of course.

## Expected Behavior
In bash, `xdg-open doc.pdf` *automatically* goes to the background after starting, even without using `&` explicitly. This is the behaviour I'm used to and expected. xdg-open is not aliased or anything in bash.

## Workaround
So I have `aliases["o"]="xdg-open"` normally, and this can be replaced by a function that calls xdg-open with `&` such that it is sent to background as expected. 

I filed this issue anyway because I don't really know why bash and xonsh differ in how they treat xdg-open (as I believe in both cases it resolves directly to the shell script `/usr/bin/xdg-open`), so I thought it could be a bug. 

## xonfig

```
$ xonfig
| Python           | 3.7.6                |
| pygments         | 2.4.2                |
| distro           | fedora               |
```

laloch:
Fixes #3496.

### DIFF
--- a/news/unthread-xdg-open.rst
+++ b/news/unthread-xdg-open.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* ``xdg-open`` now runs unthreaded.
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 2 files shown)


scopatz:
Looks good to me! Thanks!

PR #3500, opened 2020-04-04
xontrib-ssh-agent moved to github

dyuri:
I've migrated my repositories from bitbucket to github (due to the end of mercurial support on bitbucket).


laloch:
Thanks @dyuri for keeping us updated!

Issue #3506, opened 2020-04-07
Sphinx 3.0.0 broke the docs


gforsyth:
I'll grab a traceback after I reinstall sphinx 3 -- quick workaround is to downgrade to sphinx==2.4.4

gforsyth:
```
Exception occurred:
  File "/opt/miniconda3/lib/python3.7/site-packages/sphinx/util/docstrings.py", line 54, in prepare_docstring
    lines = s.expandtabs(tabsize).splitlines()
AttributeError: 'getset_descriptor' object has no attribute 'expandtabs'
The full traceback has been saved in /tmp/sphinx-err-k0ticwcz.log, if you want to report the issue to the developers.
```

gforsyth:
And here's the full traceback:
```
Traceback (most recent call last):
  File "/opt/miniconda3/lib/python3.7/site-packages/sphinx/cmd/build.py", line 280, in build_main
    app.build(args.force_all, filenames)
  File "/opt/miniconda3/lib/python3.7/site-packages/sphinx/application.py", line 348, in build
    self.builder.build_update()
  File "/opt/miniconda3/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 299, in build_update
    len(to_build))
  File "/opt/miniconda3/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 311, in build
    updated_docnames = set(self.read())
  File "/opt/miniconda3/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 418, in read
    self._read_serial(docnames)
  File "/opt/miniconda3/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 439, in _read_serial
    self.read_doc(docname)
  File "/opt/miniconda3/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 479, in read_doc
    doctree = read_doc(self.app, self.env, self.env.doc2path(docname))
  File "/opt/miniconda3/lib/python3.7/site-packages/sphinx/io.py", line 221, in read_doc
    pub.publish()
  File "/opt/miniconda3/lib/python3.7/site-packages/docutils/core.py", line 217, in publish
...skipping...
    self._read_serial(docnames)
  File "/opt/miniconda3/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 439, in _read_serial
    self.read_doc(docname)
  File "/opt/miniconda3/lib/python3.7/site-packages/sphinx/builders/__init__.py", line 479, in read_doc
[...]

PR #3507, opened 2020-04-07
Pin `sphinx` so the docs don't break

gforsyth:
Workaround for #3506 

No news entry

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -1,6 +1,6 @@
 cloud_sptheme
 numpydoc
-Sphinx
+Sphinx==2.4.4
 prompt_toolkit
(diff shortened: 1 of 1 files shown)


laloch:
Is this a temporary solution because of ENOTIME?

laloch:
I wonder, wasn't the Azure failure supposed to get fixed by #3505?

gforsyth:
Azure is starting to be a real pain.  Maybe we should push on #3467 and drop all of these scattered CI services.  I'll rebase that PR and rip out everything else and see what happens.

laloch:
Thanks, @gforsyth!

PR #3513, opened 2020-04-08
Make autovox work with Python 3.5

laloch:
We still can't use f-strings because of Python 3.5 support.

### DIFF
--- a/news/autovox-fstrings.rst
+++ b/news/autovox-fstrings.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks!

astronouth7303:
Wait, people are using 3.5?

laloch:
I don't think so, but [xonsh-in-docker.py](../blob/master/xonsh-in-docker.py) uses it by default 😉 

If we know no one uses it nowadays (we even consider the idea funny), why would we want to keep supporting it?

gforsyth:
I vote we stop supporting it

PR #3516, opened 2020-04-10
New default for THREAD_SUBPROC on Win

### LINKED ISSUE #3498: Output missing from history in Windows Terminal
In the latest Windows Terminal preview on windows 10, patches of output from xonsh are missing if you scroll up through history.

This does not occur if xonsh is run in the cmd.exe console host, and it does not occur if I run the same commands with the cmd shell inside Windows Terminal.

## xonfig

I've tried xonsh 0.9.15 and 0.9.9, python 3.7 and python 3.8, prompt_toolkit2 and 3, all seem to exhibit the problem.

<details>

```
$ xonfig
| xonsh            | 0.9.15               |
| Git SHA          | feeb6315             |
| Commit Date      | Mar 19 10:50:33 2020 |
| Python           | 3.7.7                |
| have readline    | False                |
| on posix         | False                |
| on windows       | True                 |
```

</details>

## Expected Behavior

History is complete

## Current Behavior

History has blank gaps in it (number of lines seems to be correct, they're just empty)

## Steps to Reproduce

This seems to occur in various commands, but as an example, change to a large directory e.g. `C:\windows\system32` and run `ls -la`

melund:
This changes the default value for $THREAD_SUBPROC on Windows.

Fixes #3498

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -779,7 +779,7 @@ def DEFAULT_VALUES():
         "SUGGEST_COMMANDS": True,
         "SUGGEST_MAX_NUM": 5,
         "SUGGEST_THRESHOLD": 3,
-        "THREAD_SUBPROCS": True,
+        "THREAD_SUBPROCS": False if ON_WINDOWS else True,
         "TITLE": DEFAULT_TITLE,
         "UPDATE_COMPLETIONS_ON_KEYPRESS": False,
         "UPDATE_OS_ENVIRON": False,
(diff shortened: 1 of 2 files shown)


xonsh/environ.py
```diff
@@ -1205,7 +1205,11 @@ def DEFAULT_DOCS():
             "* Stopping the thread with ``Ctrl+Z`` yields to job control.\n"
             "* Threadable commands are run with ``Popen`` and threadable \n"
             "  alias are run with ``ProcProxy``.\n\n"
-            "The desired effect is often up to the command, user, or use case."
+            "The desired effect is often up to the command, user, or use case.\n\n"
+            ".. note: This is ``False`` by default on Windows as as this feature has\n"
```
laloch, line 1209: Here's a typo in "as as".

melund:
Hmm. Our RST check prevent us from having real references in the changelog (for example: 
```
 :ref:`$THREAD_SUBPROCS=False <thread_subprocs>`)
```
Gives `Failed: THREAD_SUBPROC-win.rst:8: single grave accents are not valid rst`

xonsh/environ.py
melund: Thanks. I will fix it :)

laloch: Now there's a double space instead of double "as", but I don't think it really matters ;)

scopatz:
It seems like there are some legitimate windows tests that need to be moved

melund:
@scopatz, I don't think we should change this. We loose too much core functionality. So I will close the PR . 

But I would like too try a change so I un-captured sub-procs runs unthreadable. But when I tried that xonsh didn't start and I ran out is time.

PR #3536, opened 2020-04-25
Flake8

bobhy:
Superficial but extensive code cleanup to reduce ad-hoc linting exceptions, thanks to covid-19 downtime.
Touched many files, fixed 2 actual bugs `xonsh/pyghooks.py` at 240 and `xonsh\style_tools.py` at 68.
Will continue to research errors in `xonsh/tokenize.py`, it looks very wrong.

Please especially review use of `from builtins import events, aliases, XonshError` to resolve F821 for those (globals, DynamicProxy).

No one asked for this PR, and it will consume a lot of review time now with no functional gain.  But hopefully it's justified in future time saved by all contributors.


bobhy:
Eh?  Tests don't seem to have run `pytest --flake8`??

If they had, file `xonsh/tokenize.py` should have failed.  There are several F821 (undefined name) errors starting at line 885, and I don't see how they are supposed to get initialized.

gforsyth:
Hey @bobhy -- this is probably long overdue, but it's also a beast to review.

Is it possible to break this up into smaller PRs?  
As an example, it seems like we aren't running `black` or `flake8` on the tests directory, so one PR might add those tests in and also fix the resulting formatting and flake issues.

Also, `ply` is a bundled dependency and we don't want to touch the formatting in that directory at all -- it should remain as it is until we pull in updates from upstream.

Thanks!

bobhy:
tests, yes, and I'll look at ply, didn't intend to change it.  and a couple of rules caused actual code changes, let's do those separately, too...

bobhy:
Breaking this PR down into multiple smaller ones, starting with, #3550.

PR #3538, opened 2020-04-26
New xontrib-output-search

anki-code:
**[xontrib-output-search](https://github.com/anki-code/xontrib-output-search)** is to get identifiers, names, paths, URLs and words from the previous command output and use them for the next command.

* <b>Save time</b>. Forget about using mouse, touchpad or trackball to get any words from output to the next command.
* <b>Secure</b>. The xontrib-output-search is not writing any output on the hard disk. Only latest not empty output stored in the memory. It works the same way as xonsh shell and the security level is the same.
* <b>Universal</b>. Forget about searching autocomplete plugins for every app you use and get the identifiers from the output.

Screenshot_20200426_141924

Details in the [README](https://github.com/anki-code/xontrib-output-search).


gforsyth:
Cool!  Thanks @anki-code !

Issue #3539, opened 2020-04-26
/usr/bin/cut: the delimiter must be a single character with cut -d';'


sucrecacao:
try to run 
```
cat ~/.zsh_history  | cut -d';' -f2-
```
and got the error:
`/usr/bin/cut: the delimiter must be a single character`

## xonfig

<details>

```
$ xonfig
```

</details>

## Expected Behavior
Should show the trimmed command from my zsh history, like if I run it in zsh or bash
## Current Behavior
show the above error

## reproduce 
enter
```
cat ~/.zsh_history  | cut -d';' -f2-
``` 
in a xonsh terminal

sucrecacao:
I found out the solution while writing the issue. 
zsh/bash allow `-d';'` without space and xonsh need a space. I was used to skip the space.

sucrecacao:
`cut --delimiter=';'` won't work neither, I guess I understand why, but it's still a little bit disappointing ( first time using xonsh )

anki-code:
You expect that `';'` works the same way as in bash/zsh but no. The working example:
```bash
cat ~/.zsh_history  | cut -d@(';') -f2-
cat ~/.zsh_history  | cut --delimiter=@(';') -f2-
```
To understand the difference you can use `$XONSH_TRACE_SUBPROC=True`:
```bash
> $XONSH_TRACE_SUBPROC=True

> cat ~/.zsh_history  | cut --delimiter=';' -f1 o> /dev/null e> /dev/null  
TRACE SUBPROC: (['cat', '/home/user/.zsh_history'], '|', ['cut', "--delimiter=';'", '-f1', '>', '/dev/null'])

> cat ~/.zsh_history  | cut --delimiter=@(';') -f1 > /dev/null
TRACE SUBPROC: (['cat', '/home/user/.zsh_history'], '|', ['cut', '--delimiter=;', '-f1', '>', '/dev/null'])
```

anki-code:
@scopatz @gforsyth Meanwhile I can't find the explanation of this difference in the documentation.

Why xonsh can't do the same as bash? 

The xonsh works like bash in this examples:
```
xonsh$ $XONSH_TRACE_SUBPROC=True
xonsh$ echo 'qwe'                                                                                                                                                                                                                           
TRACE SUBPROC: (['echo', 'qwe'],)
```
```
bash$ set -x
bash$ echo 'qwe'
+ echo qwe
```

But:
```
[...]

PR #3540, opened 2020-04-27
Add $XONSH_TRACE_SUBPROC to the tutorial

anki-code:
To give users understanding how xonsh works [like in 3539](https://github.com/xonsh/xonsh/issues/3539#issuecomment-619737972) I've added `$XONSH_TRACE_SUBPROC` examples to the docs.


gforsyth:
The `no_close_fds` test is super flaky on OSX via github actions -- I'm inclined to mark it as an `xfail`, but in the interim, I'm merging things in that don't touch any integrations

PR #3550, opened 2020-04-29
Flake8 1

bobhy:
Retracting #3536 in favor of smaller, more digestible bites.
This PR fixes flake8 issues just in tests/ ; it's a fair number of files, but small changes.
Also changes to `setup.cfg` to allow `flake8` cli to produce results comparable to `pytest --flake8` (at least for tests/...)


gforsyth:
Hey @bobhy -- thanks for putting this in.  Can you merge/rebase from master so we can get a clean CI run?  Thanks!

tests/test_builtins.py
```diff
@@ -233,7 +230,7 @@ def test_convert_macro_arg_exec(kind):
 
 
 @pytest.mark.parametrize("kind", [type, "t", "type"])
-def test_convert_macro_arg_eval(kind):
+def test_convert_macro_arg_eval_type(kind):
```
gforsyth, line 233:
```suggestion
def test_convert_macro_arg_type(kind):
```

bobhy:
Not sure where to add `tests` to flake checks in CI!  I *removed* this directory from the list of exceptions that pytest gets from `setup.cfg` so local run of pytest indeed checks them.  Is there some other list of directories that CI uses?  Not seeing any...

gforsyth:
> Not sure where to add `tests` to flake checks in CI!

Ha, you know, we might've accidentally nuked that at some point.  We can add a separate github action for that.

bobhy:
If you would, I have no idea how to do it.  But please allow this action to
depend on the existing flake8/pytest-flake8 config in `setup.cfg`, so we
get consistent enforcement in IDE and CI.

Issue #3551, opened 2020-04-29
Bad test case causes incorrect CI failure


bobhy:
## Expected Behavior
Unit tests from main branch should run in CI, unless contributor introduces an error.

## Current Behavior
```
=================================== FAILURES ===================================
______________________ test_convert_macro_arg_eval[None] _______________________

kind = None

    @pytest.mark.parametrize("kind", [eval, None, "v", "eval"])
    def test_convert_macro_arg_eval(kind):
        # literals
        raw_arg = "42"
        arg = convert_macro_arg(raw_arg, kind, {}, None)
>       assert arg == 42
E       AssertionError: assert '42' == 42
```
`convert_macro_arg( , None)` is expected to return its argument unchanged.  This test case should be dropped.

PR #3552, opened 2020-04-29
fix test case for convert_macro_arg

bobhy:
Proposed fix for #3551.

### DIFF
--- a/news/is_3551.rst
+++ b/news/is_3551.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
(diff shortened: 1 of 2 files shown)


gforsyth:
Hey @bobhy -- thanks for putting this in.  Looks like this slipped by (for a long time) because there's a second `test_convert_macro_arg_eval` on Line 233.  Can you also change that test name to `test_convert_macro_arg_type` so we can be sure that both tests run?

Thanks!

bobhy:
Diddit.  The test failure in macOS is a timeout invoking subprocess.  I *bet* that's not a code issue.

PR #3563, opened 2020-05-07
Fix json history

### LINKED ISSUE #3562: AttributeError exception in JSON history backend
## xonfig

<details>

```
$ xonfig
| Git SHA          | 292d292d            |
| Commit Date      | May 1 13:35:50 2020 |
| Python           | 3.8.2               |
| distro           | arch                |
```

</details>

## Current Behavior
If `$HISTCONTROL` is set to either `ignoreerr`, `ignoredups` or both, JSON history backend may crash with the backtrace below. Though the backtrace was triggered by Powerline xontrib, it happens whenever something attempts to reach the end of the history (e.g. `__xonsh__.history.rtns[-1]`) of an already flushed history buffer. This is due to the fact that not all history entries are always flushed to the history file, but the change of the history length, caused by skipping some entries, is not reflected by the internal state (it simply thinks that the history is longer than it really is).

### Traceback (if applicable)

<details>

```
  File "/home/str/code/xonsh/xonsh/prompt/base.py", line 47, in __call__
    prompt = self._format_prompt(template=template)
  File "/home/str/code/xonsh/xonsh/prompt/base.py", line 55, in _format_prompt
    template = template() if callable(template) else template
  File "/usr/lib/python3.8/site-packages/xontrib/powerline.xsh", line 170, in prompt
    r = s()
  File "/usr/lib/python3.8/site-packages/xontrib/powerline.xsh", line 105, in rtns
    if __xonsh__.history.rtns and __xonsh__.history.rtns[-1] != 0:
  File "/home/str/code/xonsh/xonsh/history/json.py", line 273, in __getitem__
    rtn = lj["cmds"][key].get(self.field, self.default)
AttributeError: 'list' object has no attribute 'get'
```

</details>

## Steps to Reproduce

```
$ __xonsh__.history.buffersize = 1
$ $XONSH_SHOW_TRACEBACK=True
$ $XONSH_DEBUG=True
$ foobarfoobar # returns error
$ __xonsh__.history.rtns[-1]
  File "/home/str/code/xonsh/xonsh/base_shell.py", line 362, in default
  File "/home/str/code/xonsh/xonsh/codecache.py", line 67, in run_compiled_code
    func(code, glb, loc)
  File "/home/str/code/xonsh/xonsh/history/json.py", line 274, in __getitem__
    rtn = lj["cmds"][key].get(self.field, self.default)
AttributeError: 'list' object has no attribute 'get'
```

laloch:
Due to $HISTCONTROL setting some history entries may not be always flushed to the file and the on-disk history may actually be shorter than it was in the in-memory buffer. This patch fixes the indexing error occuring under such condition.

This also fixes race condition between `JSONHistory` and `JSONHistoryFlusher` leading to on-disk history truncation.

Closes #3562

### DIFF
--- a/xonsh/history/json.py
+++ b/xonsh/history/json.py
@@ -430,7 +430,7 @@ class JsonHistory(History):
         hf = JsonHistoryFlusher(
             self.filename, tuple(self.buffer), self._queue, self._cond, at_exit=at_exit
         )
-        self.buffer.clear()
+        self.buffer = []
         return hf
 
     def items(self, newest_first=False):
--- a/xonsh/history/json.py
+++ b/xonsh/history/json.py
@@ -222,7 +222,9 @@ class JsonHistoryGC(threading.Thread):
 class JsonHistoryFlusher(threading.Thread):
     """Flush shell history to disk periodically."""
 
-    def __init__(self, filename, buffer, queue, cond, at_exit=False, *args, **kwargs):
+    def __init__(
+        self, filename, buffer, queue, cond, at_exit=False, skip=None, *args, **kwargs
+    ):
         """Thread for flushing history."""
         super(JsonHistoryFlusher, self).__init__(*args, **kwargs)
         self.filename = filename
@@ -231,6 +233,7 @@ class JsonHistoryFlusher(threading.Thread):
         queue.append(self)
         self.cond = cond
         self.at_exit = at_exit
+        self.skip = skip
         if at_exit:
             self.dump()
             queue.popleft()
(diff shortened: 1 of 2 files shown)


xonsh/history/json.py
```diff
@@ -373,10 +381,19 @@ def flush(self, at_exit=False):
...
-            self.filename, tuple(self.buffer), self._queue, self._cond, at_exit=at_exit
+            self.filename,
+            tuple(self.buffer),
+            self._queue,
+            self._cond,
+            at_exit=at_exit,
+            skip=skip,
         )
-        self.buffer.clear()
+        self.buffer = []
```
scopatz, line 396: Why make a new list here?

scopatz:
I think this might have been addressed by the fixes @bobhy provided

xonsh/history/json.py
laloch: See https://github.com/xonsh/xonsh/issues/3562#issuecomment-625258720.

laloch:
Nope, #3600 is not related. This is not related to history garbage collection at all.

xonsh/history/json.py
laloch: Passing a list as an argument does not deepcopy the list, so you can't just pass a list to a background thread for processing and clear it immediately. If you do, there's nothing left for processing on the background thread - just an empty list.

scopatz:
Cool, can you maybe merge from master to get the new CI please?

laloch:
Rebased on master.

scopatz:
Thanks @laloch!

PR #3565, opened 2020-05-08
Fix column offset of unary operator AST node

### LINKED ISSUE #3564: chaining other commands after `nix-store` causes "NameError: name 'nix' is not defined"
## xonfig

<details>

```
$ xonfig                                                                                
| xonsh            | 0.9.17          |
| Python           | 3.7.7           |
| prompt toolkit   | 3.0.4           |
| pygments         | 2.5.2           |
| distro           | unknown         |
```

</details>

## nix-info -m

<details>

 - system: `"x86_64-linux"`
 - host os: `Linux 5.4.35, NixOS, 20.09pre223023.fce7562cf46 (Nightingale)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.3.4`
 - channels(root): `"nixos-20.09pre223023.fce7562cf46, nixos-unstable-20.09pre218613.ae6bdcc5358"`
 - channels(das-g): `""`
 - nixpkgs: `/nix/var/nix/profiles/per-user/root/channels/nixos`

</details>

## Expected Behavior
If `nix-store` is in `$PATH`, it can be used as part of longer command lines, too, with other commands chained after it.

## Current Behavior

If the command line continues after `nix-store` and its arguments, xonsh won't run `nix-store` and instead reports `NameError: name 'nix' is not defined`.

### Traceback (if applicable)

<details>

```
  File "/nix/store/y22pwfk39i9r22ga10c3ifbp345gdl0p-xonsh-0.9.17/lib/python3.7/site-packages/xonsh/base_shell.py", line 362, in default
  File "/nix/store/y22pwfk39i9r22ga10c3ifbp345gdl0p-xonsh-0.9.17/lib/python3.7/site-packages/xonsh/codecache.py", line 67, in run_compiled_code
    func(code, glb, loc)
NameError: name 'nix' is not defined
```

</details>

## Steps to Reproduce

1. make sure `nix-store` is in `$PATH`
    ```xonsh
    nix-store --version
    ```
    You should get some output like
    ```
    nix-store (Nix) 2.3.4
    ```
2. call the following command line
    ```xonsh
    nix-store --version or true
    ```

laloch:
Fixes #3564

### DIFF
--- a/news/fix-unary-position.rst
+++ b/news/fix-unary-position.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Logical subprocess operators now work after long arguments (e.g. ``--version``).
+
+**Security:**
+
+* <news item>
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -415,6 +415,49 @@ else:
         "Var foo\n",
         0,
     ),
+    #
+    # test logical subprocess operators
+    #
+    (
+        """
+def _echo(args):
+    print(' '.join(args))
+aliases['echo'] = _echo
+
+echo --version and echo a
+echo --version && echo a
+echo --version or echo a
+echo --version || echo a
+echo -+version and echo a
+echo -+version && echo a
+echo -+version or echo a
+echo -+version || echo a
+echo -~version and echo a
+echo -~version && echo a
+echo -~version or echo a
+echo -~version || echo a
+""",
+        """--version
+a
+--version
+a
+--version
+--version
+-+version
+a
+-+version
+a
+-+version
+-+version
+-~version
+a
+-~version
+a
+-~version
+-~version
+""",
+        0,
+    ),
 ]
 
 
--- a/xonsh/parsers/base.py
+++ b/xonsh/parsers/base.py
@@ -2153,12 +2153,13 @@ class BaseParser(object):
         p[0] = p[1]
 
     def p_factor_unary(self, p):
-        """factor : PLUS factor
-                  | MINUS factor
-                  | TILDE factor
+        """factor : plus_tok factor
+                  | minus_tok factor
+                  | tilde_tok factor
         """
-        op = self._factor_ops[p[1]]()
-        p[0] = ast.UnaryOp(op=op, operand=p[2], lineno=self.lineno, col_offset=self.col)
+        p1 = p[1]
+        op = self._factor_ops[p1.value]()
+        p[0] = ast.UnaryOp(op=op, operand=p[2], lineno=self.lineno, col_offset=p1.lexpos)
 
     def p_power_atom(self, p):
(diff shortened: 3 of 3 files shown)


news/fix-unary-position.rst
```diff
@@ -0,0 +1,23 @@
...
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Logical sobprocess operators now work after long arguments (e.g. --version).
```
das-g, line 19:
```suggestion
* Logical subprocess operators now work after long arguments (e.g. `--version`).
```

laloch: Thanks! New system - no proofreader :wink:

tests/test_aliases.py
```diff
@@ -121,6 +121,18 @@ def test_recursive_callable_partial_none(xonsh_execer, xonsh_builtins):
...
+        "echo --version or echo 'there'",
+        "echo --version || echo 'there'",
+        "echo -+version and echo 'there'",
+        "echo -+version && echo 'there'",
+        "echo -+version or echo 'there'",
+        "echo -+version || echo 'there'",
+        "echo -~version and echo 'there'",
+        "echo -~version && echo 'there'",
+        "echo -~version or echo 'there'",
+        "echo -~version || echo 'there'",
```
das-g, line 135: These tests look sensible, but they **do pass even without the fix** (i.e., without the change from 04ecb2661f82612a32dd5a105b48da4aa2116e35), so it seems they test something else than what was fixed here.

laloch: Damn it! Ho can it be? These should have triggered the NameError...
Thanks for pointing that out.

das-g: Maybe its different within an alias than when the command line is issued explicitly? Just guessing here, I don't yet know the innards of xonsh.

laloch: Should be OK now. Thanks for the review.

laloch:
The `p_factor_unary` is not really that important. It's just a part of the Python grammar constructing `UnaryOp` operator AST node. The real evil was happening in https://github.com/xonsh/xonsh/blob/14771a7e51441144148399ab286c81c0d5edbb3d/xonsh/ast.py#L396
When our `CtxAwareTransformer` decides, that the current input cannot be successfully evaluated as a Python code in the current scope, it calls `try_subproc_toks` that tries to find some sense in the input and to replace it with `![subproc args]` uncaptured subprocess construct. It begins by finding the first and the last character index (column) of the current part of the syntax tree within the current input line. The `maxcol` value, that should be the position of the last input character of the current AST node, is derived from the highest value in all of `col_offset` attributes of the current syntax tree. The issue you discovered was caused by a bug in `UnaryOp`'s `col_offset` value, that was pointing behind the text of its argument. It confused `try_subproc_toks` into thinking that the input was not meant to be a subprocess invocation and the whole story ended in a futile attempt to evaluate `(foo) - (bar) - (-version)` as Python code, which obviously led to `NameError`, since there was no `foo` symbol defined.

gforsyth:
And thanks @das-g for looking it over and testing it out!

Issue #3569, opened 2020-05-10
Can't use ulimit -n


RemiTorracinta:
Currently, I'm unable set my open file limit when a xonsh shell. I'm on MacOS 10.15.4 .

`ulimit -n` works fine on other shells, such as zsh, but as you can see in the example below nothing seems to happen for me with xonsh.

```
 ~  $ ulimit -a
core file size          (blocks, -c) 0
data seg size           (kbytes, -d) unlimited
file size               (blocks, -f) unlimited
max locked memory       (kbytes, -l) unlimited
max memory size         (kbytes, -m) unlimited
open files                      (-n) 256
pipe size            (512 bytes, -p) 1
stack size              (kbytes, -s) 8192
cpu time               (seconds, -t) unlimited
max user processes              (-u) 5568
virtual memory          (kbytes, -v) unlimited
 ~  $ ulimit -n 1000
 ~  $ _.rtn
0
 ~  $ ulimit -a
core file size          (blocks, -c) 0
data seg size           (kbytes, -d) unlimited
file size               (blocks, -f) unlimited
max locked memory       (kbytes, -l) unlimited
max memory size         (kbytes, -m) unlimited
open files                      (-n) 256
pipe size            (512 bytes, -p) 1
stack size              (kbytes, -s) 8192
cpu time               (seconds, -t) unlimited
max user processes              (-u) 5568
virtual memory          (kbytes, -v) unlimited
 ~  $
```

I don't know much about ulimit but it's my understanding that it should set resource limits for the current shell process and its children. So maybe something about the xonsh implementation of subprocess mode doesn't make this work.

laloch:
Hi @RemiTorracinta, unfortunately, Xonsh does not support the `ulimit` command (I'll consider implementing it). However, since Xonsh, essentially, is Python, you can use the Pythonic way of setting your limits:
```python
$ import resource
$ nofile = resource.getrlimit(resource.RLIMIT_NOFILE)
$ resource.setrlimit(resource.RLIMIT_NOFILE, (1000, nofile[1]))
```

RemiTorracinta:
Oh! I didn't realize ulimit is a shell built-in. The pythonic approach is working fine, thank you!

PR #3571, opened 2020-05-12
Add ulimit command to xoreutils

laloch:
It was asked for in #3569.

```
$ ulimit --help
Usage: ulimit [-h] [-SH] [-a] [-cdefilmnqrstuv] [LIMIT]

Set or get shell resource limits.

Provides control over the resources available to the shell and processes it
creates, on systems that allow such control.

Options:
-h, --help
                    show this help message and exit
-S, --soft
                    use the 'soft' resource limit for the following arguments
-H, --hard
                    use the 'hard' resource limit for the following arguments (default)
-a, --all
                    show all current limits
-c, --core-size
                    the maximum size of core files created.
                    By setting this limit to zero, core dumps can be disabled.
-d, --data-size
                    the maximum size of a process' data segment
-e, --nice
                    the maximum scheduling priority (`nice')
-f, --file-size
                    the maximum size of files created by the shell and its children
-i, --signals-pending
                    the maximum number of pending signals
-l, --lock-size
                    the maximum size a process may lock into memory
-m, --resident-set-size
                    the maximum resident set size.
-n, --file-descriptor-count
                    the maximum number of open file descriptors
-q, --queue-size
                    the number of bytes that can be allocated for POSIX message queues
-r, --rt-prio
                    the maximum real-time scheduling priority
-s, --stack-size
                    the maximum stack size
-t, --cpu-time
                    the maximum amount of CPU time in seconds
-u, --process-count
                    the maximum number of processes available to a single user
-v, --virtual-memory-size
                    the maximum amount of virtual memory available to the shell

Not all options are available on all platforms.

If LIMIT is given, it is the new value of the specified resource; the special
LIMIT values `soft', `hard', and `unlimited' stand for the current soft limit,
the current hard limit, and no limit, respectively. Otherwise, the current
value of the specified resource is printed. If no option is given, then -f is
assumed.
```

### DIFF
--- a/news/ulimit.rst
+++ b/news/ulimit.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* ``ulimit`` command.
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -21,6 +21,7 @@ from xonsh.platform import (
     ON_NETBSD,
     ON_OPENBSD,
     ON_DRAGONFLY,
+    ON_POSIX,
 )
 from xonsh.tools import (
     XonshError,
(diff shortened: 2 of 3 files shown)


laloch:
This is now ready to land. Not exactly the most beautiful Python code in the world, but it works and it's mostly compatible with bash, except that we support more argument combinations such as printing and setting soft and hard limits at once (`ulimit -SaHa`) or showing and setting limits in a single run (`ulimit -Sn -n 1000 -n`).

laloch:
yep, there is. I needed/wanted to preserve the order of the arguments and to make them repeatable, e.g. `ulimit -Hn -Sn`.

Issue #3573, opened 2020-05-12
Little question: How to return from a xonsh script prematurely


halloleo:
Say I have a xsh script but under one condition I want to exit the script early. E.g.:

```
# myscript.xsh

if $DONOTHING == 1:
   exit
# more code
# more code
# more code
```

`exit` on its own does not work. `return` doesn't ether. Do I have to import `sys` and write `sys.exit()` or is there a shorter way to do this?

***

BTW, thanks to everybody for xonsh! It's still my favourite shell (despite dabbling very briefly into [elvish](https://elv.sh/)).

anki-code:
Try `raise SystemExit` to avoid importing.

laloch:
This is a duplicate of #3241.

halloleo:
`raise SystemExit` is cool! Thanks! :-)

Issue #3241, opened 2019-07-30
How to exit a script?


chengxuncc:
I don't know how to exit a script, I had searched tutorial and tested this:
```
exit
echo should exit
```
Not working, still echo `should exit`.

jb-leger:
You can use the `exit` from `sys` module.

```
import sys
sys.exit(13)
echo should exit
```

Works as expected with the expected return code status.

chengxuncc:
Thanks, now I get it, `.xsh` script is a superset of python script.

anki-code:
> You can use the exit from sys module.
To avoid importing `sys`/`os` modules you can call `raise SystemExit` ([with or without return code](https://docs.python.org/3/library/exceptions.html#SystemExit)).

PR #3574, opened 2020-05-12
exit to docs

anki-code:
#3573 #3241

### DIFF
--- a/docs/bash_to_xsh.rst
+++ b/docs/bash_to_xsh.rst
@@ -68,6 +68,9 @@ line is ``#!/usr/bin/env xonsh``.
     * - ``$@``
       - ``$ARGS``
       - List of all command line argument and parameter strings.
+    * - ``exit``
+      - ``raise SystemExit`` or ``sys.exit()``
+      - Exiting from the current script.
 
 To understand how xonsh executes the subprocess commands try
 to set :ref:`$XONSH_TRACE_SUBPROC <xonsh_trace_subproc>` to ``True``:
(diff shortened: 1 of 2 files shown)


docs/bash_to_xsh.rst
```diff
@@ -68,6 +68,9 @@ line is ``#!/usr/bin/env xonsh``.
     * - ``$@``
       - ``$ARGS``
       - List of all command line argument and parameter strings.
+    * - ``exit``
+      - ``raise SystemExit`` or ``sys.exit()``
```
astronouth7303, line 72: While `raise SystemExit` does do it, I've literally never seen this in python code. I would not encourage its use.

news/exit.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Added two ways of exit to the "Bash to xsh" article.
```
astronouth7303, line 7: FYI, we generally don't require news items for docs, but thank you!

docs/bash_to_xsh.rst
anki-code: 304k in GitHub search - https://github.com/search?q=%22raise+SystemExit%22&type=Code

astronouth7303: That doesn't mean it's a good idea or idiomatic or something we should actively encourage.

anki-code: The best way is to have just `exit()` function by default without any imports. Ok, I remove SystemExit.

news/exit.rst
anki-code: Thank you for info!

PR #3575, opened 2020-05-13
get_hg_branch aware of topics

paugier:
Mercurial [topics](https://www.mercurial-scm.org/doc/evolution/tutorials/topic-tutorial.html) are not taken into account by xonsh.

- [x] changelog not yet updated (see https://xon.sh/devguide.html#changelog)
- [x] Check compatibility with the Python version targeted by xonsh master. Which version is it? Currently, the changes are compatible with Python >= 3.6.

This function works fine in my .xonshrc. I didn't try to run the xonsh tests with these modifications. I hope the CI will do that for me :slightly_smiling_face:

### DIFF
--- a/xonsh/prompt/vc.py
+++ b/xonsh/prompt/vc.py
@@ -9,6 +9,7 @@ import builtins
 import threading
 import subprocess
 import re
+import pathlib
 
 import xonsh.tools as xt
 from xonsh.lazyasd import LazyObject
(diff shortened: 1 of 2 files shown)


paugier:
Can this small PR be considered?

xonsh/prompt/vc.py
```diff
@@ -9,6 +9,7 @@
 import threading
 import subprocess
 import re
+from pathlib import Path
```
scopatz, line 12: Please import `pathlib` and then use `pathlib.Path` below so that this gets the benefit of amalgamation

xonsh/prompt/vc.py
```diff
@@ -101,27 +102,26 @@ def get_hg_branch(root=None):
...
-        branch = branch.strip(os.linesep)
+    # add activated bookmark and topic
+    for filename in ["bookmarks.current", "topic"]:
+        feature_branch_path = Path(root) / ".hg" / filename
+        if feature_branch_path.exists():
+            with open(feature_branch_path) as file:
+                feature_branch = file.read().strip()
+            if feature_branch:
+                if branch:
+                    branch = f"{branch}, {feature_branch}"
```
scopatz, line 121: @astronouth7303 @gforsyth - are we allowing f-strings yet?  Sorry I have been out of it for a while

fhgd, line 121: Since a topic belongs to a branch, it is possible to use a slash instead of a comma like
````python
branch = f"{branch}/{feature_branch}"
````
This is the notation from heptapod, e.g. . https://foss.heptapod.net/mercurial/hg-git where you can select the different (topic-) branches.

paugier: When I checked, I saw that xonsh master supports Python >= 3.6, so I guessed that f-strings were allowed.

xonsh/prompt/vc.py
paugier: Reading https://github.com/xonsh/amalgamate#imports, it seems that I could also just change the `from ... import` statement as `import pathlib.Path as Path` to get the advantage of the lazy import. Can I do that? Like that the rest of the code is not modified by this "amalgamate.py" particularity and stays as if it were written with the standard `from ... import` shown in https://docs.python.org/3/library/pathlib.html#basic-use.

xonsh/prompt/vc.py
paugier: Yes! But we should keep the comma notation for the bookmarks...

astronouth7303: We really should be. The blocker would be if any CI is still running 3.5.

xonsh/prompt/vc.py
scopatz: Yep! You could use that style of import. But to get the import advantage, you'd have to change other modules to use exactly the same import line.  Currently a lot of them just use `import pathlib`.  Your choice, but my advice would be to conform to the imports for this PR and then if you wanted to do a PR that changed all of the pathlib imports, you could put that in separately.

scopatz: Really either way though!

paugier: I understand. Let's go for `import pathlib` then!

scopatz:
Cool thanks!

PR #3588, opened 2020-05-22
Python 3.9 support

laloch:
**Preliminary.** Python is still in 3.9.0b1 by now.
No prerelease conda packages to run the tests on.
Currently blocked by python/cpython#20312.

There's currently a lot of unexpected `ast.Constant` occurences in the AST.
~~`apply-trailers` still asserts in some cases.~~

The tests do, however, at least run for me in docker, which is an improvement compared to the master branch state.
The tests are limited to v3.8 on ubuntu-latest - there's no point in stressing the CI.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -7,8 +7,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macOS-latest]
-        python-version: [3.6, 3.7, 3.8]
+        os: [ubuntu-latest]
+        python-version: [3.8]
     name: Python ${{ matrix.python-version }} ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


anki-code:
Many issues are related to lack of python 3.9 support and users ask how to force xonsh using python 3.8. I put here [the solution with pipx that tested](https://github.com/xonsh/xonsh/issues/4050#issuecomment-757005891):

```bash
# 1. Install Python 3.8 (Ubuntu example)
apt install python3.8

# 2. Install pipx and xonsh with binding to python 3.8
pip install pipx
pipx install --python python3.8 xonsh

# 3. First run xonsh just for testing
pipx run xonsh 

# 4. If you want to run short command `xonsh` add `/home/$USER/.local/bin` to PATH
```

anki-code:
Merged Py39 support - #4101

gforsyth:
Thanks for the effort here @laloch -- as we have preliminary 3.9 support added in #4101 I'm going to close this out.

PR #3561, opened 2020-05-07
Run (pure) flake8 in CI; deprecate config for pytest-flake8

bobhy:
Adds `flake8` back into CI.  Runs `flake8` itself rather than the (dimly-recalled)  `pytest --flake8`, which has a different syntax for configuring lint exceptions.  Now a single list of lint exceptions can apply in developer's IDE and to CI.  (DRY)

PR #3591, opened 2020-05-24
Flake8 4

bobhy:
Fourth of 3 planned lint cleanups in xonsh.  Like the previous ones, this one is dependent on #3561.
One more cleanup PR is pending, which will fix one last exception (F401) and collapse these four as well.


bobhy:
Closing pending refactoring per #3592

Issue #3599, opened 2020-05-29
history gc not working for JSON files backend


bobhy:
Discovered by code inspection in the course of working on #3586.

The fix is easy.  But the bug has been present for a couple of years now and reintroducing "correct" behavior will clip somebody's history, behavior indistinguishable from a new bug.

Do we need to add a new flag or warning to go along with this?  Let's discuss...

astronouth7303:
Deleting data unexpectedly isn't great.

Do we have a config on how old the history is?

I almost feel like we need to enshrine the "bug" as the default behavior? I don't like that option, but it's either that (and allow the default history to grow unbounded), or the next release will delete data.

bobhy:
Yes, there is $XONSH_HISTORY_SIZE which lets you configure history garbage
collection.  This has been documented behavior from the beginning of the
history feature.

I don't think it's a good idea to let history grow without bound.  Large
histories cause problems in Xonsh (#3586) and other PTK applications
and other shells have ways of enforcing a limit, so we're we're not being
arbitrary.

One way to go would be to issue a warning every time you start Xonsh if we
see that your existing history is "well" over the configured limit (which
defaults to 8128 commands.  Where did that number came from?) until you
either manually trim your history or change the configured limit.  So we
would not automatically and silently delete history.

Another way (more accomodating to senior users) would be to add some new
config flag to confirm you want to apply the configured or defaulted
limit.  E.g, add a third item in $XONSH_HISTORY_SIZE  or create a signal
file in the history directory?
I'd rather not do this, it complicates the elegant simplicity that is Xonsh
config today.

Your (all) thoughts?

laloch:
I think `$XONSH_HISTORY_SIZE` should be honored for JSON backend.
However:
Do I understand it right, that #3586 is in fact a PTK issue? Why do we want to work it around in Xonsh? Does @jonathanslenders even know it exists?

bobhy:
Yes, I do think it is an internal PTK issue.  See https://github.com/xonsh/xonsh/issues/3586#issuecomment-637512579 for more disussion.

laloch:
What about changing the default from 8128 (?) commands to something like 2 years?

bobhy:
[...]

PR #3600, opened 2020-05-30
fix history gc for json backend

bobhy:
Addressing issue #3599.

As mentioned there, need feedback whether to add a warning or an extra option before this fix deletes a lot of someone's old history.

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -8,7 +8,14 @@ import pytest
 
 from xonsh.lazyjson import LazyJSON
 from xonsh.history.dummy import DummyHistory
-from xonsh.history.json import JsonHistory
+from xonsh.history.json import (
+    JsonHistory,
+    _xhj_gc_commands_to_rmfiles,
+    _xhj_gc_files_to_rmfiles,
+    _xhj_gc_seconds_to_rmfiles,
+    _xhj_gc_bytes_to_rmfiles,
+)
+
 from xonsh.history.main import history_main, _xh_parse_args, construct_history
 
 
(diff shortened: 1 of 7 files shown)


xonsh/history/json.py
```diff
@@ -131,10 +131,14 @@ def files(self, only_unlocked=False):
...
         """
-        # pylint: disable=no-member
-        env = getattr(builtins, "__xonsh__.env", None)
-        if env is None:
+
+        x = getattr(builtins, "__xonsh__", None)
+        if x:
+            env = getattr(x, "env", None)
+
+        if not (x and env):
```
BYK, line 139:
```suggestion
        env = getattr(getattr(builtins, "__xonsh__", None), "env", None)
        if env is None:
```

Even if you don't go with this suggestion, I'd avoid the implicit `None` checks and use a more descriptive variable name rather than `x`.

BYK: Or I guess we can even do:
```suggestion
        try:
            buildins.__xonsh__.env
        except AttributeError:
        	return []
```

That said we don't even use `env` here, why do we check for that??

bobhy:
Ready for review, I think despite one out of 3 checks failing in CI on Windows.

Long note in news/fix-history-gc.rst with details of new warning, new `history gc --force` command and the one-line fix for the original problem..

tests/test_history.py
```diff
@@ -308,3 +315,170 @@ def test_construct_history_class(xonsh_builtins):
...
+    # time zone is UTC, all days have 24 h and 0 sec, no leap year or leap sec.
+
+    retval = []
+    for i in range(int((gen_count + 1) / 2)):
+        retval.append(
+            (
+                # first day in sec + #days * 24hr + #hr * 60min + # sec * 60sec + sec= sec to date.
+                HF_FIRST_DAY + (((((i * 24) + 9) * 60) + 0) * 60) + 0,  # mod dt,
+                100,
+                f"/home/bob/.argle/xonsh-{2*i:05n}.json",
```
BYK, line 344: Could this be the reason for Windows failures?

xonsh/history/json.py
```diff
@@ -118,31 +136,46 @@ def run(self):
...
         fs = _xhj_get_history_files(sort=False)
         files = []
         for f in fs:
             try:
-                if os.path.getsize(f) == 0:
+                cur_file_size = os.path.getsize(f)
+                if cur_file_size == 0:
                     # collect empty files (for gc)
-                    files.append((time.time(), 0, f))
+                    files.append(cur_file_size, (time.time(), 0, f))
```
BYK, line 178: I'm very confused between this line and the rest of the comments regarding the contents of this `files` array. How does this line work?

tests/test_history.py
bobhy: Don't think so, these tests don't try to open the history files, just manipulate the list.

xonsh/history/json.py
bobhy: Thanks for the close review!  First off, this line is a bug.  `cur_file_size` should be 4th element of the tuple (changed design, missed this leftover).

Second, I'm also confused by the special case treatment of zero-length history files.  It may have been a problem once upon a time...  But here, I'm pretty sure the timestamp should be the modification date of the file (which it should have even with zero length), otherwise it might not get garbage collected if user configures 's' age limit.  I'll make both changes and maybe get lucky in CI.

bobhy:
Merged and rebased on recent master, ready for review.

bobhy:
Pleased hold up, do *not* merge yet.
Discovered a functional problem where `history gc` hangs forever, and a build problem due to incorrect commit squashing that includes an unrelated PR.  I'll get this sorted and let you know when it's OK to review.

laloch:
You can turn the PR into a draft ;)

bobhy:
OK, this is now ready for review.  I want to stop thinking about it till voices *outside* my head start making suggestions.

BYK:
This is beyond me at this point @scopatz so passing to you :)

xonsh/history/json.py
```diff
@@ -15,53 +15,70 @@
...
+    hsize_removed (int):  units of history to be removed
+    rm_files ((mod_ts, num_commands, path, fsize)[]): list of files to remove.
+    """
     n = 0
     ncmds = 0
-    for ts, fcmds, f in files[::-1]:
-        if fcmds == 0:
-            # we need to make sure that 'empty' history files don't hang around
-            rmfiles.append((ts, fcmds, f))
+    for _, fcmds, f, _ in files[::-1]:
```
scopatz, line 33: should be `reversed(files)`, rather than `files[::-1]` for speed

xonsh/history/json.py
```diff
@@ -15,53 +15,70 @@
...
+
         if ncmds + fcmds > hsize:
             break
         ncmds += fcmds
         n += 1
-    rmfiles += files[:-n]
-    return rmfiles
+
+    cmds_removed = 0
+    for _, fcmds, f, _ in files[:-n]:
```
scopatz, line 42: maybe assign `files[:-n]` to a variable rather than creating the list twice, like `rm_files = files[:-n]`

xonsh/history/json.py
```diff
@@ -15,53 +15,70 @@
...
 
 
 def _xhj_gc_bytes_to_rmfiles(hsize, files):
     """Return the history files to remove to get under the byte limit."""
-    rmfiles = []
     n = 0
     nbytes = 0
-    for _, _, f in files[::-1]:
-        fsize = os.stat(f).st_size
+    for _, _, f, fsize in files[::-1]:
```
scopatz, line 72: reversed

xonsh/history/json.py
```diff
@@ -15,53 +15,70 @@
...
         if nbytes + fsize > hsize:
             break
         nbytes += fsize
         n += 1
-    rmfiles = files[:-n]
-    return rmfiles
+    bytes_removed = 0
+    for _, _, f, fsize in files[:-n]:
```
scopatz, line 78: variable

xonsh/history/json.py
bobhy: Every little bit helps, i guess.  `reversed()` is 10x faster than `[::-1]`, though this code path is on a background thread not contributing to user response time...

bobhy:
Better?

scopatz:
Thanks @bobhy!

Issue #3586, opened 2020-05-22
AssertionError when first typing in a fresh shell


qwenger:
This is a follow-up on #2339.

Since the upgrade to PTK3, I sometimes (like 1/5 times) get the traceback reproduced below when typing in a freshly opened terminal.

Pressing ENTER reprints the same traceback indefinitely.

I'm on ArchLinux, using xfce4-terminal; but it also happens in linux console (CTRL+ALT+Fx).

It can happen even when waiting for several seconds before typing, and happens after a couple characters are entered.

## xonfig

<details>

```
+------------------+---------------------+
| xonsh            | 0.9.18              |
| Git SHA          | 26475146            |
| Commit Date      | May 5 06:42:28 2020 |
| Python           | 3.8.2               |
| PLY              | 3.11                |
| have readline    | True                |
| prompt toolkit   | 3.0.5               |
| shell type       | prompt_toolkit      |
| pygments         | 2.6.1               |
| on posix         | True                |
| on linux         | True                |
| distro           | arch                |
| on darwin        | False               |
| on windows       | False               |
| on cygwin        | False               |
| on msys2         | False               |
| is superuser     | False               |
| default encoding | utf-8               |
| xonsh encoding   | utf-8               |
| encoding errors  | surrogateescape     |
+------------------+---------------------+
```
</details>

### Traceback (if applicable)

<details>

```
Unhandled exception in event loop:
  File "/usr/lib/python3.8/asyncio/events.py", line 81, in _run
    self._context.run(self._callback, *self._args)
  File "/usr/lib/python3.8/site-packages/prompt_toolkit/eventloop/utils.py", line 72, in schedule
    func()
  File "/usr/lib/python3.8/site-packages/prompt_toolkit/application/application.py", line 456, in redraw
    self._redraw()
  File "/usr/lib/python3.8/site-packages/prompt_toolkit/application/application.py", line 523, in _redraw
    self.context.run(run_in_context)
  File "/usr/lib/python3.8/site-packages/prompt_toolkit/application/application.py", line 509, in run_in_context
    self.renderer.render(self, self.layout)
  File "/usr/lib/python3.8/site-packages/prompt_toolkit/renderer.py", line 598, in render
    layout.container.write_to_screen(
  File "/usr/lib/python3.8/site-packages/prompt_toolkit/layout/containers.py", line 381, in write_to_screen
    sizes = self._divide_heights(write_position)
[...]

PR #3601, opened 2020-05-30
Limit history loaded into ptk.

bobhy:
Mitigation for #3586. Not a fix!  I think a real fix will require introduction of `asyncio` in Xonsh.


bobhy:
Retracting, I think #3600 is a better approach.

PR #3606, opened 2020-06-05
Add pdb xontrib

laloch:
Once loaded, the xontrib runs pdb debugger on reception of SIGUSR1 signal.

### DIFF
--- a/news/xontrib-pdb.rst
+++ b/news/xontrib-pdb.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* ``pdb`` xontrib, that runs pdb debugger on reception of SIGUSR1 signal.
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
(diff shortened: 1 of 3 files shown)


laloch:
The Windows CI failure seems to be consistent, but I don't thing it could be related.

astronouth7303:
Looks good. I just have to hit enter to get the debugger prompt?

laloch:
Nope, you need to `kill -s USR1 $PIDOF_XONSH` and the target process drops to the debugger prompt instantly.

laloch:
Oh, I see why you asked! If the shell is just sitting at the prompt, not running a script, you really need to press enter in order to get to the debugger if your prompt is readline. If your prompt is PTK, it drops you to the debugger instantly, but it does not echo the output afterwards.

scopatz:
This is super awesome!

scopatz:
I believe that the test failure is unrelated

PR #3624, opened 2020-06-17
xontrib: don't stop loading xontribs on error

laloch:
`xontrib load` action can load multiple xontribs at once. If, however, one of the xontribs fails to load, because, for instance, of a missing dependency, the command exits and the rest of the modules is not attempted to load at all.
This patch modifies the `load` action behavior to only print exception on error, but to continue loading remaining xontribs.

### DIFF
--- a/xonsh/xontribs.py
+++ b/xonsh/xontribs.py
@@ -8,7 +8,14 @@ import functools
 import importlib
 import importlib.util
 
-from xonsh.tools import print_color, unthreadable
+from enum import IntEnum
+from xonsh.tools import print_color, print_exception, unthreadable
+
+
+class ExitCode(IntEnum):
+    OK = 0
+    NOT_FOUND = 1
+    INIT_FAILED = 2
 
 
 def find_xontrib(name):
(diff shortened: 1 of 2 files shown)


xonsh/xontribs.py
```diff
@@ -109,18 +109,25 @@ def xontrib_metadata():
 def xontribs_load(names, verbose=False):
     """Load xontribs from a list of names"""
     ctx = builtins.__xonsh__.ctx
+    res = 0
```
astronouth7303, line 112: Why 0/1/2?

Can we name this something more descriptive? Or better, document what this function returns and why?

xonsh/xontribs.py
```diff
@@ -109,18 +109,25 @@ def xontrib_metadata():
...
     for name in names:
         if verbose:
             print("loading xontrib {0!r}".format(name))
-        update_context(name, ctx=ctx)
+        try:
+            update_context(name, ctx=ctx)
+        except Exception:
+            res = 2
+            print_exception("Failed to load xontrib {}.".format(name))
     if update_context.bad_imports:
```
astronouth7303, line 121: I know not your code, but should this be `hasattr(update_context, 'bad_imports')` or something similar?

laloch: Yes, it should. I didn't notice.

xonsh/xontribs.py
laloch: Is it better now?

xonsh/xontribs.py
laloch: Done.

xonsh/xontribs.py
astronouth7303: Oh! That's why more than I was expecting (I was thinking, like, docstring), but yes, that works.

PR #3627, opened 2020-06-18
Set argv[0] of a subprocess to the command as inserted by the user

### LINKED ISSUE #3617: Surprising `mpirun -np 1 which python`
I'm trying to understand how to use xonsh. This is particularly disturbing and looks like a bug!

```
pierre@kth ~ $ python --version                                                 
Python 3.8.3
pierre@kth ~ $ mpirun -np 1 python --version                                    
Python 2.7.18rc1
pierre@kth ~ $ which python                                                     
/home/pierre/.pyenv/shims/python
pierre@kth ~ $ mpirun -np 1 which python                                        
/usr/bin/python
pierre@kth ~ $ echo $PATH                                                       
/home/pierre/miniconda3/condabin/app:/home/pierre/opt/julia-1.0.2/bin/:/home/pierre/.pyenv/shims:/home/pierre/.pyenv/bin:/home/pierre/.local/bin/:/home/pierre/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
pierre@kth ~ $ mpirun -np 1 echo $PATH                                          
/home/pierre/miniconda3/condabin/app:/home/pierre/opt/julia-1.0.2/bin/:/home/pierre/.pyenv/shims:/home/pierre/.pyenv/bin:/home/pierre/.local/bin/:/home/pierre/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
```

## xonfig

(xonsh from master)

<details>

```
$ xonfig
| xonsh            | 0.9.18          |
| Python           | 3.8.3           |
| prompt toolkit   | 3.0.5           |
| pygments         | 2.6.1           |
| distro           | ubuntu          |
```

</details>

laloch:
Currently Xonsh passes a resolved path of an executable as argv[0] argument. This changes the argv[0] to contain the command as inserted by the user. The new behavior is consistent with bash and other shells. Some programs behave differently in dependence on the argv[0] value.
Fixes #3617.

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -206,8 +206,9 @@ def _un_shebang(x):
 
 def get_script_subproc_command(fname, args):
     """Given the name of a script outside the path, returns a list representing
-    an appropriate subprocess command to execute the script.  Raises
-    PermissionError if the script is not executable.
+    an appropriate subprocess command to execute the script or None if
+    the argument is not readable or not a script. Raises PermissionError
+    if the script is not executable.
     """
     # make sure file is executable
     if not os.access(fname, os.X_OK):
@@ -217,10 +218,10 @@ def get_script_subproc_command(fname, args):
         # execute permissions but not read/write permissions. This enables
         # things with the SUID set to be run. Needs to come before _is_binary()
         # is called, because that function tries to read the file.
-        return [fname] + args
+        return None
     elif _is_binary(fname):
         # if the file is a binary, we should call it directly
-        return [fname] + args
+        return None
     if ON_WINDOWS:
         # Windows can execute various filetypes directly
         # as given in PATHEXT
@@ -730,7 +731,9 @@ class SubprocSpec:
         if self.binary_loc is None:
             return
         try:
-            self.cmd = get_script_subproc_command(self.binary_loc, self.cmd[1:])
+            scriptcmd = get_script_subproc_command(self.binary_loc, self.cmd[1:])
+            if scriptcmd is not None:
+                self.cmd = scriptcmd
         except PermissionError:
             e = "xonsh: subprocess mode: permission denied: {0}"
             raise XonshError(e.format(self.cmd[0]))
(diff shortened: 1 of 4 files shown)


xonsh/built_ins.py
```diff
@@ -730,7 +731,9 @@ def resolve_executable_commands(self):
         if self.binary_loc is None:
             return
         try:
-            self.cmd = get_script_subproc_command(self.binary_loc, self.cmd[1:])
+            scriptcmd = get_script_subproc_command(self.binary_loc, self.cmd[1:])
```
scopatz, line 734: Maybe the right thing to do is to pass in `alias` here and then in `get_script_subproc_command()`, rather than returning None, return `[alias] + args`

scopatz:
Thanks @laloch! Can you please merge from master. Adding a test here would be great too. Thanks!

xonsh/built_ins.py
laloch: Nope, I don't think so. The point is, that there's a difference between running `ls` and `/bin/ls`. It's not related to `alias`.

laloch:
Rebased on master.

laloch:
I don't really know how to test this. I would need some external binary which would print its argv[0].

scopatz:
You can add something to do this in `tests/bin`. Something like `tests/bin/printname.xsh` is a good example

laloch:
Yeah, the problem is that the "something" needs to do an equivalent of
```c
#include <stdio.h>

int main(int argc, char **argv) {
  printf("[");
  for (int i = 0; i < argc; ++i) {
    printf("'%s', ", argv[i]);
  }
  printf("]\n");
}
```
On Linux I could do `cat /proc/@(os.getpid())/cmdline` in a XSH script, but I don't know any cross-platform method of doing this except for compiling the above C code.

laloch:
Test added. Unfortunately it only runs on Linux.

scopatz:
> Test added. Unfortunately it only runs on Linux.

That's great! Thanks!

PR #3634, opened 2020-06-24
Fix TypeError triggered by unexpected return from callable alias

### LINKED ISSUE #3598: Callable alias hung if returning list.
## xonfig

<details>

```
ftian@ftian-desktop:~$ xonsh
ftian@ftian-desktop ~ $ def _foo(args): 
.......................     return [1, 2], [3, 4] 
.......................                                                                         
ftian@ftian-desktop ~ $ aliases['foo'] = _foo                                                   
ftian@ftian-desktop ~ $ foo                                                                     
Exception in thread Thread-138:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.8/dist-packages/xonsh/__amalgam__.py", line 15040, in run
    self.returncode = parse_proxy_return(r, sp_stdout, sp_stderr)
  File "/usr/local/lib/python3.8/dist-packages/xonsh/__amalgam__.py", line 14785, in parse_proxy_return
    stdout.write(r[0])
TypeError: write() argument must be str, not list



```

</details>

## Expected Behavior

Error.

## Current Behavior

xonsh hung, ^C won't kill it either.

### Traceback (if applicable)

<details>

```
traceback
```

</details>

laloch:
Callable aliases returning tuple may raise `TypeError` exception effectively hanging the shell, if the tuple contains values of unexpected type.

Fixes #3598.

### DIFF
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -1203,12 +1203,12 @@ def parse_proxy_return(r, stdout, stderr):
     elif isinstance(r, cabc.Sequence):
         rlen = len(r)
         if rlen > 0 and r[0] is not None:
-            stdout.write(r[0])
+            stdout.write(str(r[0]))
             stdout.flush()
         if rlen > 1 and r[1] is not None:
-            stderr.write(r[1])
+            stderr.write(str(r[1]))
             stderr.flush()
-        if rlen > 2 and r[2] is not None:
+        if rlen > 2 and isinstance(r[2], int):
             cmd_result = r[2]
     elif r is not None:
         # for the random object...
--- a/news/fix-alias-return.rst
+++ b/news/fix-alias-return.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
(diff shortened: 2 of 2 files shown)


xonsh/proc.py
```diff
@@ -1203,12 +1203,18 @@ def parse_proxy_return(r, stdout, stderr):
     elif isinstance(r, cabc.Sequence):
         rlen = len(r)
         if rlen > 0 and r[0] is not None:
-            stdout.write(r[0])
+            if isinstance(r[0], str):
+                stdout.write(r[0])
+            else:
+                stdout.write(str(r[0]))
```
scopatz, line 1209: I think this can just be `stdout.write(str(r[0]))`, because `str()` of a str object is is basically free. Same below

scopatz:

xonsh/proc.py
laloch: I always prefer explicit over implicit, so yes, this is intentional. But OK, I can change it.

laloch: Done.

PR #3594, opened 2020-05-27
Fix Jedi warnings (plus some bits and pieces)

laloch:
Jedi v0.16.0 changed the completer API and started emitting deprecation warnings for the old API calls (see davidhalter/jedi@5f6a25fb5). This detects Jedi version and switches to the new API on jedi >= 0.16.0.

Plus:
- remove some strange completer registration steps.
Please correct and enlighten me if those were important.
- remove references to `xonsh.builtins` module.
- run black on xontrib/jedi.xsh.

PR #3643, opened 2020-07-23
Improve completers and Jedi xontrib

daniel-shimon:
Hi!

This PR contains a few changes to the completers ecosystem:

## Enable more advanced completers

These are two proposed API changes for completers.

### Enable display and description texts with RichCompletion

A completer can return a `RichCompletion` object that can create better completion displays:

```python
return set(
    RichCompleter(value, prefix_len, display, description)
    for c in completions
)
```

### Enable access to ptk's multiline document (*from a completer call*)

```python
ptk_completer = xonsh.completers.tools.get_ptk_completer()
if ptk_completer is not None:
    full_text = ptk_completer.current_document.text
    row = ptk_completer.current_document.cursor_position_row + 1
```

## Improvements to the Jedi xontrib

* Use new Jedi API (inspired by #3594)
* Replace the existing python completer
* Create rich completions with extra info (inspired by IPython)
* Use entire multiline document if available (so the completion is more accurate)
* Complete xonsh special tokens
* Be aware of _ (last result)
* Only show dunder attrs when prefix ends with '_' (also like in IPython)

## Other small fixes to other completers

* base completer: Don't complete unnecessarily
* completer: Fix crash in 'completer' completer
* aliases: Fix typo in 'source' alias

### DIFF
--- a/xonsh/completers/base.py
+++ b/xonsh/completers/base.py
@@ -11,6 +11,10 @@ def complete_base(prefix, line, start, end, ctx):
     and paths.  If we are completing the first argument, complete based on
     valid commands and python names.
     """
+    if line.strip() and prefix != line:
+        # don't do unnecessary completions
+        return set()
+
     # get and unpack python completions
     python_comps = complete_python(prefix, line, start, end, ctx)
     if isinstance(python_comps, cabc.Sequence):
--- a/xonsh/completers/completer.py
+++ b/xonsh/completers/completer.py
@@ -8,7 +8,14 @@ def complete_completer(prefix, line, start, end, ctx):
     args = line.split(" ")
     if len(args) == 0 or args[0] != "completer":
         return None
+
+    if end < len(line) and line[end] != " ":
+        # completing in a middle of a word
+        # (e.g. "completer some<TAB>thing")
+        return None
+
     curix = args.index(prefix)
+
     compnames = set(builtins.__xonsh__.completers.keys())
     if curix == 1:
         possible = {"list", "help", "add", "remove"}
--- a/xonsh/completers/tools.py
+++ b/xonsh/completers/tools.py
@@ -33,6 +33,41 @@ def justify(s, max_length, left_pad=0):
     return "\n".join(txt)
 
 
+class RichCompletion(str):
+    """A rich completion that completers can return instead of a string
+
+    Parameters
+    ----------
+    value : str
+        The completion's actual value.
+    prefix_len : int
+        Length of the prefix to be replaced in the completion.
+        If None, the default prefix len will be used.
+    display : str
+        Text to display in completion option list.
+        If None, ``value`` will be used.
+    description : str
+        Extra text to display when the completion is selected.
+    """
+
+    def __new__(cls, value, prefix_len=None, display=None, description=""):
+        completion = super().__new__(cls, value)
+
+        completion.prefix_len = prefix_len
+        completion.display = display or value
+        completion.description = description
+
+        return completion
+
+    def __repr__(self):
+        return "RichCompletion({}, prefix_len={}, display={}, description={})".format(
+            repr(str(self)),
+            self.prefix_len,
+            repr(self.display),
+            repr(self.description),
+        )
+
+
 def get_ptk_completer():
     """Get the current PromptToolkitCompleter
 
--- a/xonsh/completers/tools.py
+++ b/xonsh/completers/tools.py
@@ -31,3 +31,21 @@ def justify(s, max_length, left_pad=0):
     """
     txt = textwrap.wrap(s, width=max_length, subsequent_indent=" " * left_pad)
     return "\n".join(txt)
+
+
+def get_ptk_completer():
+    """Get the current PromptToolkitCompleter
+
+    This is usefull for completers that want to use
+    PromptToolkitCompleter.current_document (the current multiline document).
+
+    Call this function lazily since in '.xonshrc' the shell doesn't exist.
+
+    Returns
+    -------
+    The PromptToolkitCompleter if running with ptk, else returns None
+    """
+    if __xonsh__.shell is None or __xonsh__.shell.shell_type != "prompt_toolkit":
+        return None
+
+    return __xonsh__.shell.shell.pt_completer
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -21,6 +21,7 @@ import threading
 import subprocess
 import collections
 
+import xonsh.completers.tools as xct
 from xonsh.lazyasd import LazyObject, lazyobject
 from xonsh.base_shell import BaseShell
 from xonsh.ansi_colors import (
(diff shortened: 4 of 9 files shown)


daniel-shimon:
Now with ptk and Jedi it looks awesome!

image

Notice how:

* `center` is completed but `center()` is displayed in the list
* The function's signature appears under the list

xonsh/completers/base.py
```diff
@@ -11,6 +11,10 @@ def complete_base(prefix, line, start, end, ctx):
     and paths.  If we are completing the first argument, complete based on
     valid commands and python names.
     """
+    if line.strip() != "" and prefix != line:
```
scopatz, line 14: This would be better as `if line.strip() and prefix != line`

xonsh/completers/completer.py
```diff
@@ -8,7 +8,13 @@ def complete_completer(prefix, line, start, end, ctx):
     args = line.split(" ")
     if len(args) == 0 or args[0] != "completer":
         return None
-    curix = args.index(prefix)
+    try:
+        curix = args.index(prefix)
+    except ValueError:
```
scopatz, line 13: Rather than using index() &  Value error, it is probably best do something like:

```
curix = next(i for i, x in enumerate(args) if x == prefix, None)
if curix is None:
    return None
```

scopatz: I am pretty sure that this is more performant than try-except

xonsh/completers/tools.py
```diff
@@ -31,3 +31,53 @@ def justify(s, max_length, left_pad=0):
     """
     txt = textwrap.wrap(s, width=max_length, subsequent_indent=" " * left_pad)
     return "\n".join(txt)
+
+
+class RichCompletion(str):
```
scopatz, line 36: We probably shouldn't subclass string directly. Maybe `collections.UserString` is what you are looking for?

xonsh/completers/tools.py
```diff
@@ -31,3 +31,53 @@ def justify(s, max_length, left_pad=0):
...
+    """A rich completion that completers can return instead of a string
+
+    Parameters
+    ----------
+    value: str
+        The completion's actual value.
+    prefix_len : int
+        Length of the prefix to be replaced in the completion.
+        If None, the default prefix len will be used.
+    display: str
```
scopatz, line 46: space needed before colon

xonsh/completers/tools.py
```diff
@@ -31,3 +31,53 @@ def justify(s, max_length, left_pad=0):
...
+        Text to display in completion option list.
+        If None, `value` will be used.
```
scopatz, line 48: Double backticks for rst

xonsh/completers/tools.py
```diff
@@ -31,3 +31,53 @@ def justify(s, max_length, left_pad=0):
...
+    -------
+    The PromptToolkitCompleter if running with ptk, else returns None
+    """
+    if __xonsh__.shell is None:
+        return None
+
+    shell = __xonsh__.shell.shell
+
+    # avoid importing the class in case ptk isn't installed
+    if shell.__class__.__name__ == "PromptToolkitShell":
```
scopatz, line 81: I think it is more robust to check `__xonsh__.shell.shell_type` than the class name

xonsh/readline_shell.py
```diff
@@ -29,6 +29,7 @@
     ansi_color_style,
 )
 from xonsh.prompt.base import multiline_prompt
+from xonsh.completers.tools import RichCompletion
```
scopatz, line 32: I think this need to be `import xonsh.completers.tools` to get the automatic lazy imports

xonsh/readline_shell.py
```diff
@@ -281,6 +282,27 @@ def inserter():
...
+    """Render the completions according to the required prefix_len.
+
+    Readline will replace the current prefix with the chosen rendered completion.
+    """
+    chopped = prefix[:-prefix_len] if prefix_len else prefix
+
+    rendered_completions = []
+    for comp in completions:
+        if isinstance(comp, RichCompletion) and comp.prefix_len is not None:
+            if comp.prefix_len == 0:
```
scopatz, line 295: Should be `if comp.prefix_len:` for speed (and switch conditions)

xontrib/jedi.xsh
```diff
@@ -1,51 +1,138 @@
...
 
-@lazybool
-def HAS_JEDI():
-    """``True`` if `jedi` is available, else ``False``."""
-    spec = importlib.util.find_spec('jedi')
-    return (spec is not None)
+try:
+    import jedi
+except ImportError:
+    raise SystemExit
```
scopatz, line 19: I don't think xontribs should kill xonsh...

scopatz: At least not on import :wink:

scopatz:
Very cool! Thanks for this @daniel-shimon.  There are some comments inline to clean up

xonsh/completers/tools.py
daniel-shimon: Hi!
I'm not sure that would be better, since we *do* want the object to behave like a regular string and be a subclass of it.
For example, at `xonsh/ptk_shell/completer.py:72` the function `os.path.commonprefix` is being called on the completions, which demands them to be `str`s.
Also, according to the [python docs](https://docs.python.org/3/library/collections.html#userstring-objects):

> The need for this class has been partially supplanted by the ability to subclass directly from str; however, this class can be easier to work with because the underlying string is accessible as an attribute

... Which we don't really need

xontrib/jedi.xsh
daniel-shimon: It won't, `SystemExit` only exits the script (from #3241)
Is this bad practice?

xonsh/readline_shell.py
daniel-shimon: What (/where) are lazy imports in xonsh?

scopatz: They are generated as part of amalgamate

xonsh/completers/base.py
daniel-shimon: done

xonsh/completers/completer.py
daniel-shimon: Did it even more performantly 🐳 
What do you think?

xonsh/completers/tools.py
daniel-shimon: done

xonsh/readline_shell.py
daniel-shimon: done

daniel-shimon: done

xonsh/readline_shell.py
daniel-shimon: I can't find __amalgamate__.py anywhere, where is it supposed to be?

xonsh/completers/tools.py
daniel-shimon: done

xonsh/completers/tools.py
daniel-shimon: done

xonsh/completers/tools.py
scopatz: Yeah, maybe this is OK now

xonsh/readline_shell.py
scopatz: Yeah, I think there are issues with `pip install` not generating it right now. It should be in the install location.

scopatz: But the fix here is to say `import xonsh.completers.tools` or `import xonsh.completers.tools as xct` and then use this name below

xontrib/jedi.xsh
scopatz: Yeah, but xontribs are not scripts, exactly.  They can be loaded on REPL startup as well.  This is not what we want, I don't think.  I think there should be a different escape hatch, like a function which returns None on load failure, for example

daniel-shimon: OK so apparently `SystemExit` really exits the xontribs process (or the xonsh process when the xontribs command is in xonshrc).

But if some other exception is raised (say `ModuleNotFoundError`) it's printed to the screen, xontribs catches it and recognizes the xontrib failed to load.
So I left it as simply importing jedi in `jedi.xsh`, and I think the result is acceptable:

image

xonsh/readline_shell.py
daniel-shimon: done

xontrib/jedi.xsh
scopatz: Yep! I think that is good enough!

daniel-shimon:
@scopatz thanks a lot for the effort of reviewing my PR! :clap:
Can it be merged?

scopatz:
Thanks @daniel-shimon!

PR #3649, opened 2020-07-29
Remove distutils fallback.

### LINKED ISSUE #3638: Remove distutils install path
The current implementation of xonsh uses a bi-modal install routine that uses setuptools if available and falls back to distutils if setuptools is not available. Use of distutils is discouraged and distutils is likely to be deprecated and removed from stdlib. The recommended practice now is to require setuptools, either implicitly (failing when it's not available) or explicitly (using PEP 518 build requirements).

Dropping the distutils fallback from xonsh would simplify the install routine, prepare it for future compatibility, and enable features like declarative config.

Would you consider a PR to remove the distutils fallback? Are there known use-cases relying on it that could not be adapted to use Setuptools?

jaraco:
Fixes #3638.

Also remove Python 2 compatibility imports.

Also declare build-time dependency on setuptools.

### DIFF
--- a/news/goodbye-distutils.rst
+++ b/news/goodbye-distutils.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* Xonsh now relies exclusively on Setuptools for install.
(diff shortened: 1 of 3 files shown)


scopatz:
Love the branch name!

scopatz:
Thanks!

PR #3486, opened 2020-03-23
Fix error message for `vox rm`

sfermigier:
The error message for `vox rm` wrongly assumed that `vox deactivate` needs an argument.

Example of the error in action:

```
(jnov) fermigier@mbp-stefane-abilian ~/projects/jnov master $ vox rm jnov
The "jnov" environment is currently active. In order to remove it, deactivate it first with "vox deactivate jnov".

(jnov) fermigier@mbp-stefane-abilian ~/projects/jnov master $ vox deactivate jnov
usage: vox [-h] {new,create,activate,workon,enter,deactivate,exit,list,ls,remove,rm,delete,del,help} ...
vox: error: unrecognized arguments: jnov
(jnov) fermigier@mbp-stefane-abilian ~/projects/jnov master $ vox deactivate
Deactivated "jnov".
```

PR #3652, opened 2020-08-01
vox rm

scopatz:
added news to #3486


scopatz:
Self-merging because I only added news

PR #3653, opened 2020-08-01
docs: More examples of what SubprocSpec can do

### LINKED ISSUE #3648: How to detect if an alias' output is beign redirected?
Hi, I've been using xonsh for about a week now, and I'm writing a xontrib that acts like [lsd](https://github.com/Peltoche/lsd), however I would like to detect when my alias' output is being piped to another process (in order to fall back to the system's ls command).

Python's sys.stdout.isatty doesn't seem to work, as it always returns False, and the "stdout" argument that xonsh passes to the alias is a buffer, so I can't determine this way if the output is interactive.

Is there any way to do this? I couldn't find anything about this in the documentation.

Here is a basic example:
```
$ def _echo(args): 
.     if not sys.stdout.isatty(): 
.         print("Not a tty") 
.         @(["/usr/bin/env", "echo"] + args) 
.         return 
.     print("A tty") 
.     print(args) 
.
$ aliases['echo'] = _echo
$ echo a b
Not a tty
a b
$ echo a b | cat                                                                                                                                              
Not a tty
a b
```
In the above example, the code always goes inside the if.

Thanks in advance !

cafehaine:
Closes #3648 

The tone of the new example is a bit less "serious", so let me know if I need to change that!

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1268,12 +1268,11 @@ may have one of the following signatures:
         """
         import xonsh.proc
         if spec.captured in xonsh.proc.STDOUT_CAPTURE_KINDS:
-            end = ''
+            print("I'm being captured!")
+        elif not spec.last_in_pipeline:
+            print("Going through a pipe!")
         else:
-            end = '\n'
-        # Now we'll get a newline if the user is at the terminal, and no
-        # newline if we are captured
-        print('Hi Mom!', end=end)
+            print("Hello terminal!")
         return 0
 
     def mycmd6(args, stdin=None, stdout=None, stderr=None, spec=None, stack=None):


docs/tutorial.rst
```diff
@@ -1268,12 +1268,11 @@ may have one of the following signatures:
...
-            end = ''
+            print("I'm being captured!")
+        elif not spec.last_in_pipeline:
+            print("Going through a pipe!")
         else:
-            end = '\n'
-        # Now we'll get a newline if the user is at the terminal, and no
-        # newline if we are captured
-        print('Hi Mom!', end=end)
+                print("Hello terminal!")
```
scopatz, line 1275: The indentation here doesn't seem to match up. Are you using TABs?

cafehaine: Looks like I added 4 spaces for no reason here, fixed it

scopatz: Awesome, thanks!

scopatz:
Doc-only fix. Merging

PR #3654, opened 2020-08-03
Remove some of Python 3.5 compat.

Carreau:
Remove most of Python 3.5 compatibility code.

    There  now should not be any remaining  3,5 conditinal code.
    We now directly use inspect's getouterframes, so remove it from
    inspector

    The 3.5 parser get merged into 3.6 and a number of global vaiable
    referring to 3.5/3.6 are now gone.

    The only remaining mention is appveyor test, which are anyway failing
    but not being seen as so by GitHub, so I'll leave them alone.


xonsh/tokenize.py
```diff
@@ -113,18 +110,16 @@
     "DOLLARNAME",
     "IOREDIRECT",
 ]
-HAS_ASYNC = (3, 5, 0) <= PYTHON_VERSION_INFO < (3, 7, 0)
+HAS_ASYNC = PYTHON_VERSION_INFO < (3, 7, 0)
```
gforsyth, line 113: I think we can remove this check (and the conditionals below it) entirely.  All supported version of Python should be able to handle having `ASYNC` and `AWAIT` tokens now.

xonsh/tools.py
```diff
@@ -1208,9 +1190,7 @@ def to_logfile_opt(x):
     the filepath if it is a writable file or None if the filepath is not
     valid, informing the user on stderr about the invalid choice.
     """
-    superclass = pathlib.PurePath
-    if PYTHON_VERSION_INFO >= (3, 6, 0):
-        superclass = os.PathLike
+    superclass = os.PathLike
```
gforsyth, line 1193: Maybe just kill this line and do `isinstance(x, os.PathLike)`, below?

laloch:
Could you please also adjust the default Python version in [xonsh-in-docker.py](https://github.com/xonsh/xonsh/blob/master/xonsh-in-docker.py#L12)?

daniel-shimon:
Hi! Awesome PR

How about removing `requirements/test-35.txt`?
I see `appvayor_install.cmd` uses it but I don't really know what that is 😅

Carreau:
> How about removing `requirements/test-35.txt`?
> I see `appvayor_install.cmd` uses it but I don't really know what that is 😅

I'm not sure either, appveyor says it succeeds, even though the last command has non-zero exit...

xonsh/tokenize.py
Carreau: I _think_ that this is because ASYNC/AWAIT are _soft_ keyword below 3.7 and that will make some technically 3.7 compatible code broken. But we can _try_

Carreau: Nop. Does not work.

Carreau:
Allgood, except appveyor as daniel-shimon pointed out; though I'm afraid of touching it.

daniel-shimon:
Hey @scopatz, you were the last person to touch that file, can you help us out?

daniel-shimon:
BTW @carreau, can you squash these commits together with the first one's message? They're all basically `Remove some of Python 3.5 compat. `

Carreau:
DOne, updated Top level comment.

gforsyth:
Thanks for taking this on @Carreau !

Now, who wants to change _every single_ `.format` to an f-string?

laloch:
> Now, who wants to change _every single_ `.format` to an f-string?

I wouldn't joke about it. Someone could really do it :wink:.

Carreau:
> I wouldn't joke about it. Someone could really do it 😉.

Too late... See #3657

PR #3656, opened 2020-08-05
Call repr() when pretty() fails in _pprint_displayhook

daniel-shimon:
When evaluating mock objects, `pretty` returns the object's `.xonsh_display` attr which is itself a mock, and this fails later on:

image

We can handle it by calling `repr()` when it fails.

### DIFF
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -251,9 +251,11 @@ def _pprint_displayhook(value):
         builtins._ = value
         return
     env = builtins.__xonsh__.env
+    printed_val = None
     if env.get("PRETTY_PRINT_RESULTS"):
         printed_val = pretty(value)
-    else:
+    if not isinstance(printed_val, str):
+        # pretty may fail (i.e for unittest.mock.Mock)
         printed_val = repr(value)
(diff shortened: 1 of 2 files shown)


xonsh/main.py
```diff
@@ -251,9 +251,11 @@ def _pprint_displayhook(value):
         builtins._ = value
         return
     env = builtins.__xonsh__.env
+    printed_val = None
     if env.get("PRETTY_PRINT_RESULTS"):
         printed_val = pretty(value)
-    else:
+    if not isinstance(printed_val, str):
```
scopatz, line 257: Should be `if printed_val is None` for speed

scopatz:
Thanks for putting this in @daniel-shimon!

xonsh/main.py
daniel-shimon: But that's the point - when this issue happens, `pretty(value)` returns `value.xonsh_display()` which is itself a mock object.
So the way to make sure pretty succeeded is to check if it returned a `str` like it should.

scopatz: Oh I see.

PR #3657, opened 2020-08-05
Upgrade to py36+ only syntax.

Carreau:
As requested by @gforsyth  in https://github.com/xonsh/xonsh/pull/3654#issuecomment-669227292

all possible changes to fstrings, also drop inheritance from object when
unnecessay, and remove argumetns to super, also set literals, no unicode
comments...

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """Aliases for the xonsh shell."""
 import os
 import re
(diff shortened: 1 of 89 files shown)


laloch:
> As requested by @gforsyth in [#3654 (comment)](https://github.com/xonsh/xonsh/pull/3654#issuecomment-669227292)

I thought that was a joke :roll_eyes:
Personally, I prefer useful `git blame` over cosmetic changes.

Carreau:
> I thought that was a joke 🙄
> Personally, I prefer useful `git blame` over cosmetic changes.

I hear you, it was mostly doing it for fun, it's just a master of running `pyupgrade --py36-plus` and I expect test failure as this likely updated some tests that should't be updated.

So I'm also -1 on merging.

There might be some things to cherry-pick though.

Carreau:
(like also we probably don't want to touch ply)

gforsyth:
I agree doing it everywhere at once is a bit silly, but I don't view the changes as purely cosmetic.  f-strings are straight up faster

scopatz:
This is amazing work! Thanks @Carreau!

Carreau:
let's close this and let me open an issue with also suggesting to use `pyupgrade --py36-plus` in conjunction with black, potentially via https://github.com/akaihola/darker to avoid doing large swath of changes.

PR #3377, opened 2019-10-27
Environment variable registration, deregistration (addresses #3227)

dotsdl:
This PR addresses #3227.

Added `xonsh.environ.Env.register` and `xonsh.environ.Env.deregister` methods for registration, deregistration of new environment variables with types, default values, and docstrings. Currently registration/deregistration is not exposed to the user, so the addition of these gives the ability to add new variables with behavior like e.g. `PATH`.

PR #3658, opened 2020-08-06
Fix regression with LC_MESSAGES missing in locale [No news entry needed.]

melund:
This fixes a regression introduced in #3377, 

The fact that LC_MESSAGES could be missing was handled before. 
https://github.com/xonsh/xonsh/pull/3377/files#diff-b08c376dd29024ccd6f9aef90f2ea96fL836-L838

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -990,7 +990,9 @@ def DEFAULT_VARS():
             always_false,
             locale_convert("LC_MESSAGES"),
             ensure_string,
-            locale.setlocale(locale.LC_MESSAGES) if hasattr(locale, "LC_MESSAGES") else "",
+            locale.setlocale(locale.LC_MESSAGES)
+            if hasattr(locale, "LC_MESSAGES")
+            else "",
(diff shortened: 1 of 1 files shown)


melund:
Errors seems to be the annoying: ``AttributeError: module 'builtins' has no attribute '__xonsh__'``

laloch:
The CI failures are not related.
Thanks @melund!

bobhy:
This has the effect of defining environment variable LC_MESSAGES on platforms where `locale.LC_MESSAGES` doesn't exist, albeit to an empty string.  I actually prefer the way it was done (by me!) in https://github.com/bobhy/xonsh/blob/fd9e3f2ca921789bd63229a2d77994799a8a6d75/xonsh/environ.py#L1670-L1677.

melund:
Yes. I agree. Sorry I didn't look at your PR more closely. If you want your solution, create a new PR and I will merge it.

bobhy:
Thanks, new PR #3671.

PR #3660, opened 2020-08-06
Improve handling of f-strings

### LINKED ISSUE #2822: Environment variables inside f-strings

### LINKED ISSUE #3454: Python f-strings work strange in xonsh
Hi! Thank you for your great work!
 <table style="width:100%">
  <tr>
    <th>Python 3.7.6</th>
    <th>xonsh/0.9.13</th>
  </tr>
  <tr>
    <td>
<pre>
>>> test=123
>>> f"{{ {test} }}"
'{ 123 }'
>>> f"{{ ' {test} ' }}"
"{ ' 123 ' }"
>>> f"{{  {test} ' }}"  
"{  123 ' }"
</pre>
</td>
    <td>
<pre>
$ test=123
$ f"{{ {test} }}"
'{123}'
$ f"{{ ' {test} ' }}"
' {test} '
$ f"{{  {test} ' }}"
$ # no any output
</pre>
</td>
  </tr>
</table> 

Why the difference is so dramatic? Thanks!

**For community:**

### LINKED ISSUE #3572: How to use env variables in f-strings?
The line

    key='HOME'; print(f"{key} = {${key}")

gives me the error `ValueError: unexpected '{' in field name` which sort of makes sense, but renders f-strings for environment variables useless.

In f-strings, is there a way to access the environment array through a less ambiguous name maybe?

This issue might be related to #2822 and #2870.

### xonfig

The problem occurs in multiple xonsh versions:

1)

```
| Python           | 3.7.4                |
| shell type       | dumb                 |
| distro           | unknown              |
```

2)

```
| Git SHA          | 26475146            |
| Commit Date      | May 5 06:42:28 2020 |
| Python           | 3.7.7               |
| prompt toolkit   | None                |
| shell type       | dumb                |
```

### LINKED ISSUE #3611: Xonsh can't handle empty dict {} in fstring.
Looks related to  #3572 and  #3454

Use case:
 - Looping over a dict, where a subkey may not exist.
 --  If it exists, insert the subkey value into the fstring output, otherwise None (or something else).

```
f"{someDict.get('someKey', {}).get('subkey', 'UltimateFailure')}"
ValueError: unexpected '{' in field name
```

## xonfig
<details>

```
$ xonfig
| Git SHA          | 26475146            |
| Commit Date      | May 5 06:42:28 2020 |
| Python           | 3.8.3               |
| have readline    | False               |
| on posix         | False               |
| on windows       | True                |
```

</details>

## Expected Behavior
```
testdict = dict()
testdict['key'] = {'subkey': 'value'}

f"{testdict.get('key', {}).get('subkey')}"
'value'

f"{testdict.get('key', {}).get('non_existent')}"
'None'
```

## Current Behavior
```
f"{testdict2.get('key', {}).get('subkey')}"
ValueError: unexpected '{' in field name
```

## Steps to Reproduce

See above.

## Workaround
```
f"{testdict2.get('key', dict()).get('subkey')}"
```

### LINKED ISSUE #3655: Env var dict lookup in f-string not working
Looks related to #3572, #3454 and possibly #3611 but different enough that I think it warrants a separate bug report.

It's possible to use env vars in f-strings, dict lookups in f-strings but using dict lookups of an env var doesn't seem to work.

E.g.:

```
$  $foo = 'foo'
$  f"{$foo}"  # Env var, works
'foo'
$  bar = {'bar': 3}
$  f"{bar['bar']}"  # Dict lookup, works
'3'
$  $bar = bar
$  f"{$bar['bar']}"  # Dict lookup of env var, doesn't work
$  # NOTE: it shows nothing, not even an empty string
```

Also using `f"{$bar.get('bar')}"` as a work-around doesn't work (though `f"{$bar.get}"` *does* show some output).

## xonfig

<details>

```
$ xonfig
| Git SHA          | 26475146            |
| Commit Date      | May 5 06:42:28 2020 |
| Python           | 3.6.9               |
| prompt toolkit   | 2.0.10              |
| distro           | ubuntu              |
```

</details>

## Expected Behavior

```
$  $bar = {'bar': 3}
$  f"{$bar['bar']}"
'3'
$  f"{$bar.get('bar')}"
'3'
$  baz = f"{$bar['bar']}"
$  type(baz)
str
```

## Current Behavior

```
$  $bar = {'bar': 3}
$  f"{$bar['bar']}"
$  f"{$bar.get('bar')}"
$  baz = f"{$bar['bar']}"
$  type(baz)
...
NameError: name 'baz' is not defined
```

## For community

laloch:
This replaces the current method of scanning f-strings for special Xonsh syntax using `string.Formatter`. The current code does only work in the simplest cases, because the syntax of `__format__` fields is just not the same as the syntax of f-string expressions plus the current code is simply buggy.
The replacement code tries to parse the f-string using `ast.parse` and only steps in if the parser raises `SyntaxError`. It then attempts to locate the offending expression and replaces it by a wrapper call which in turn calls `evalx` with the failing expression as an argument. It can already handle complex strings:
```
$ a=18
$ f"""foo {$TERM!s:{a}} {f"hot {${'LA'+'NG'}:{a}} dog"} bar"""
'foo xterm-256color     hot en_US.UTF-8        dog bar'
```
It can even handle captured subprocess expansion within expressions:
```
$ echo f"""\"{${'TERM'}!s:10.{$(printf 5)}}\""""
"xterm     "
```

Closes #2822
Closes #3454
Closes #3572
Closes #3611
Closes #3655

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -1061,13 +1061,8 @@ def list_of_list_of_strs_outer_product(x):
     return rtn
 
 
-fstring_fields = {}
-
-
-def eval_fstring_field(field_id):
-    """Looks up its argument in the fstring_fields and evaluates the result
-    in Xonsh context."""
-    field = fstring_fields.pop(field_id)
+def eval_fstring_field(field):
+    """Evaluates the argument in Xonsh context."""
     res = __xonsh__.execer.eval(
         field[0].strip(), glbs=globals(), locs=builtins.__xonsh__.ctx, filename=field[1]
     )
@@ -1417,8 +1412,6 @@ class XonshSession:
         self.ensure_list_of_strs = ensure_list_of_strs
         self.list_of_strs_or_callables = list_of_strs_or_callables
         self.list_of_list_of_strs_outer_product = list_of_list_of_strs_outer_product
-
-        self.fstring_fields = fstring_fields
         self.eval_fstring_field = eval_fstring_field
 
         self.completers = xonsh.completers.init.default_completers()
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -1061,6 +1061,19 @@ def list_of_list_of_strs_outer_product(x):
     return rtn
 
 
+fstring_fields = {}
+
+
+def eval_fstring_field(field_id):
+    """Looks up its argument in the fstring_fields and evaluates the result
+    in Xonsh context."""
+    field = fstring_fields.pop(field_id)
+    res = __xonsh__.execer.eval(
+        field[0].strip(), glbs=globals(), locs=builtins.__xonsh__.ctx, filename=field[1]
+    )
+    return res
+
+
 @lazyobject
 def MACRO_FLAG_KINDS():
     return {
@@ -1403,8 +1416,10 @@ class XonshSession:
         self.all_jobs = {}
         self.ensure_list_of_strs = ensure_list_of_strs
         self.list_of_strs_or_callables = list_of_strs_or_callables
-
         self.list_of_list_of_strs_outer_product = list_of_list_of_strs_outer_product
+
+        self.fstring_fields = fstring_fields
+        self.eval_fstring_field = eval_fstring_field
 
         self.completers = xonsh.completers.init.default_completers()
         self.call_macro = call_macro
--- a/xonsh/codecache.py
+++ b/xonsh/codecache.py
@@ -65,6 +65,7 @@ def run_compiled_code(code, glb, loc, mode):
     else:
         func = eval
     func(code, glb, loc)
+    builtins.__xonsh__.fstring_fields.clear()
 
 
 def get_cache_filename(fname, code=True):
--- a/xonsh/codecache.py
+++ b/xonsh/codecache.py
@@ -65,7 +65,6 @@ def run_compiled_code(code, glb, loc, mode):
     else:
         func = eval
     func(code, glb, loc)
-    builtins.__xonsh__.fstring_fields.clear()
 
 
 def get_cache_filename(fname, code=True):
--- a/xonsh/parsers/base.py
+++ b/xonsh/parsers/base.py
@@ -4,13 +4,13 @@ import os
 import re
 import time
 import textwrap
+import builtins
 from threading import Thread
 from ast import parse as pyparse
 from collections.abc import Iterable, Sequence, Mapping
 
 from xonsh.ply.ply import yacc
 
-from xonsh.tools import FORMATTER
 from xonsh import ast
 from xonsh.ast import has_elts, xonsh_call, load_attribute_chain
 from xonsh.lexer import Lexer, LexToken
@@ -27,8 +27,8 @@ RE_STRINGPREFIX = LazyObject(
 
 
 @lazyobject
-def RE_FSTR_EVAL_CHARS():
-    return re.compile(".*?[!@$`]")
+def RE_FSTR_FIELD_WRAPPER():
+    return re.compile(r"__xonsh__\.eval_fstring_field\((\d+)\)")
 
 
 class Location(object):
@@ -211,47 +211,63 @@ def hasglobstar(x):
         return False
 
 
-def _wrap_fstr_field(field, spec, conv):
-    rtn = "{" + field
-    if conv:
-        rtn += "!" + conv
-    if spec:
-        rtn += ":" + spec
-    rtn += "}"
-    return rtn
-
-
 def eval_fstr_fields(fstring, prefix, filename=None):
     """Takes an fstring (and its prefix, ie f") that may contain
-    xonsh expressions as its field values and
-    substitues them for a xonsh eval() call as needed. Roughly,
-    for example, this will take f"{$HOME}" and transform it to
-    be f"{__xonsh__.execer.eval(r'$HOME')}".
+    xonsh expressions as its field values and substitues them for
+    a call to __xonsh__.eval_fstring_field as needed.
     """
-    last = fstring[-1]
-    q, r = ("'", r"\'") if last == '"' else ('"', r"\"")
     prelen = len(prefix)
-    postlen = len(fstring) - len(fstring.rstrip(last))
-    template = fstring[prelen:-postlen]
-    repl = prefix
-    for literal, field, spec, conv in FORMATTER.parse(template):
-        repl += literal
-        if field is None:
-            continue
-        elif RE_FSTR_EVAL_CHARS.match(field) is None:
-            # just a normal python field, simply reconstruct.
-            repl += _wrap_fstr_field(field, spec, conv)
+    quote = fstring[prelen]
+    if fstring[prelen + 1] == quote:
+        quote *= 3
+    template = fstring[prelen + len(quote) : -len(quote)]
+    while True:
+        repl = prefix + quote + template + quote
+        try:
+            res = pyparse(repl)
+            break
+        except SyntaxError as e:
+            # The e.text attribute is expected to contain the failing
+            # expression, e.g. "($HOME)" for f"{$HOME}" string.
+            if e.text is None or e.text[0] != "(":
+                raise
+            error_expr = e.text[1:-1]
+            epos = template.find(error_expr)
+            if epos < 0:
+                raise
+        # We can olny get here in the case of handled SyntaxError.
+        # Patch the last error and start over.
+        xonsh_field = (error_expr, filename)
+        field_id = id(xonsh_field)
+        builtins.__xonsh__.fstring_fields[field_id] = xonsh_field
+        eval_field = f"__xonsh__.eval_fstring_field({field_id})"
+        template = template[:epos] + eval_field + template[epos + len(error_expr) :]
+
+    # If we accidentally patched some false-positives within strings,
+    # revert the false matches and reparse.
+    reparse = False
+    for node in ast.walk(res.body[0].value):
+        if isinstance(node, ast.Constant) and isinstance(node.value, str):
+            value = node.value
+        elif isinstance(node, ast.Str):
+            value = node.s
         else:
-            # the field has a special xonsh character, so we must eval it
-            eval_field = "__xonsh__.execer.eval(r" + q
-            eval_field += field.lstrip().replace(q, r)
-            eval_field += q + ", glbs=globals(), locs=locals()"
-            if filename is not None:
-                eval_field += ", filename=" + q + filename + q
-            eval_field += ")"
-            repl += _wrap_fstr_field(eval_field, spec, conv)
-    repl += last * postlen
-    return repl
+            continue
+
+        match = RE_FSTR_FIELD_WRAPPER.search(value)
+        if match is None:
+            continue
+
+        try:
+            field = builtins.__xonsh__.fstring_fields.pop(int(match.group(1)))
+        except KeyError:
+            continue
+        repl = repl.replace(match.group(0), field[0], 1)
+        reparse = True
+
+    if reparse:
+        res = pyparse(repl)
+    return res
 
 
 class YaccLoader(Thread):
@@ -2219,7 +2235,7 @@ class BaseParser(object):
                     func=leader,
                     lineno=leader.lineno,
                     col_offset=leader.col_offset,
-                    **trailer
+                    **trailer,
                 )
             elif isinstance(trailer, (ast.Tuple, tuple)):
                 # call macro functions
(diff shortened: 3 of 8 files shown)


laloch:
@xonsh/xore, what are your opinions? I realize, that the code is scary, but it works much better than the current implementation and it doesn't require full-fledged f-strings parser.

xonsh/parsers/base.py
```diff
@@ -2453,8 +2470,13 @@ def p_string_literal(self, p):
...
             )
         elif "f" in prefix:
-            s = eval_fstr_fields(p1.value, prefix, filename=self.lexer.fname)
-            s = pyparse(s).body[0].value
+            try:
+                s = pyparse(p1.value)
+            except SyntaxError:
+                s = None
+            if s is None:
+                s = eval_fstr_fields(p1.value, prefix, filename=self.lexer.fname)
```
laloch, line 2478: This can raise an exception. @scopatz, @gforsyth, how can I emit syntax error from within the parser code?

scopatz:
FWIW, I think this is thr right approach.

xonsh/parsers/base.py
scopatz: you can call `self._parse_error(msg, self.currloc(lineno=p1.lineno, column=p1.lexpos))` for some message `msg`

laloch: Thanks for your help @scopatz, but no, that doesn't do the trick. It's strange, but it doesn't do anything in fact.

laloch:
This is now ready to land. The parser cannot emit syntax errors encountered while parsing f-strings, but that's probably due to `BaseParser._parse_error(...)` not doing its thing and should be handled separately.
I hope you guys will do your best to break it :wink:

xonsh/parsers/base.py
scopatz: hmmm that really should raise a SyntaxError.  Maybe it is being caught somewhere?

laloch: Yes, most probably yes. It seems to be broken for quite a long time. I don't remember ever getting a syntax error for `f"{}"`.

scopatz:
@gforsyth - can you try to bang on this a bit please?

xonsh/codecache.py
```diff
@@ -65,6 +65,7 @@ def run_compiled_code(code, glb, loc, mode):
     else:
         func = eval
     func(code, glb, loc)
+    builtins.__xonsh__.fstring_fields.clear()
```
laloch, line 68: This is wrong! I probably need to stick the dictionary to the compiled block and let it get garbage-collected as the code block goes away.
```
$ Timer(1, lambda: print(f"{$HOME}")).start()
$ Exception in thread Thread-47:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 1254, in run
    self.function(*self.args, **self.kwargs)
  File "<xonsh-code>", line 1, in <lambda>
  File "/home/str/code/xonsh/xonsh/built_ins.py", line 1070, in eval_fstring_field
    field = fstring_fields.pop(field_id)
KeyError: 139889979616832
```

xonsh/built_ins.py
```diff
@@ -1061,6 +1061,19 @@ def list_of_list_of_strs_outer_product(x):
     return rtn
 
 
+fstring_fields = {}
+
+
+def eval_fstring_field(field_id):
+    """Looks up its argument in the fstring_fields and evaluates the result
+    in Xonsh context."""
+    field = fstring_fields.pop(field_id)
```
laloch, line 1070: This is also wrong. Fails for `[f"{$HOME}" for i in range(2)]`.

laloch:
This needs to be backported to Python < 3.8 once again.

laloch:
This is ready for review again. Sorry for the mess. It turned out to be a lot more fun than I anticipated.

xonsh/parsers/fstring_adaptor.py
```diff
@@ -0,0 +1,182 @@
+import re
```
scopatz, line 1: Please add a module level docstring as well as a hook in the `docs/` dir as well.

xonsh/parsers/fstring_adaptor.py
```diff
@@ -0,0 +1,182 @@
...
+from xonsh.lazyasd import lazyobject
+from ast import parse as pyparse
+from xonsh import ast
+from xonsh.platform import PYTHON_VERSION_INFO
```
scopatz, line 5: Please reorder these imports to be in the following groups, as per PEP8:

1. standard library imports
2. xonsh imports

xonsh/parsers/fstring_adaptor.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+if PYTHON_VERSION_INFO > (3, 8):
+
+    @lazyobject
+    def RE_FSTR_SELF_DOC_FIELD_WRAPPER():
+        return re.compile(r"(__xonsh__\.eval_fstring_field\((\d+)\)\s*)=")
+
+
+class FStringAdaptor:
```
scopatz, line 23: Please add a docstring

xonsh/parsers/fstring_adaptor.py
```diff
@@ -0,0 +1,182 @@
...
+                        ast.NameConstant(
+                            value=None, lineno=lineno, col_offset=col_offset
+                        )
+                    )
+                field_node = ast.Tuple(
+                    elts=elts, ctx=ast.Load(), lineno=lineno, col_offset=col_offset,
+                )
+                node.args[0] = field_node
+
+    def run(self):
```
scopatz, line 175: Please add a docstring

xonsh/parsers/fstring_adaptor.py
```diff
@@ -0,0 +1,182 @@
...
+                value = node.value
+            elif isinstance(node, ast.Str):
+                value = node.s
+            else:
+                continue
+
+            match = RE_FSTR_FIELD_WRAPPER.search(value)
+            if match is None:
+                continue
+            try:
```
scopatz, line 81: please rewrite this to avoid the try-except. Something like:

```py
field = self.fields.pop(int(match.group(1)), None)
if field is None:
    continue
```

xonsh/parsers/fstring_adaptor.py
```diff
@@ -0,0 +1,182 @@
...
+            elif isinstance(node, ast.Str):
+                value = node.s
+            else:
+                continue
+
+            match = RE_FSTR_SELF_DOC_FIELD_WRAPPER.search(value)
+            if match is None:
+                continue
+            field_id = int(match.group(2))
+            try:
```
scopatz, line 106: Same, please avoid try-except:

```py
field = self.fields.get(field_id, None)
if field is None:
    continue
```

xonsh/parsers/fstring_adaptor.py
```diff
@@ -0,0 +1,182 @@
...
+                and node.func.attr == "eval_fstring_field"
+                and len(node.args) > 0
+            ):
+                continue
+
+            if PYTHON_VERSION_INFO > (3, 8):
+                if isinstance(node.args[0], ast.Constant) and isinstance(
+                    node.args[0].value, int
+                ):
+                    try:
```
scopatz, line 132: Same as above

xonsh/parsers/fstring_adaptor.py
```diff
@@ -0,0 +1,182 @@
...
+                                value=field[1], lineno=lineno, col_offset=col_offset
+                            ),
+                        ],
+                        ctx=ast.Load(),
+                        lineno=lineno,
+                        col_offset=col_offset,
+                    )
+                    node.args[0] = field_node
+            elif isinstance(node.args[0], ast.Num):
+                try:
```
scopatz, line 153: Same here

scopatz:
Thanks @laloch - this looks great! Just a couple of minor issues

xonsh/parsers/fstring_adaptor.py
laloch: I don't see where this fits within `docs/`.

xonsh/parsers/fstring_adaptor.py
laloch: Done.

xonsh/parsers/fstring_adaptor.py
laloch: Done.

xonsh/parsers/fstring_adaptor.py
laloch: Done.

xonsh/parsers/fstring_adaptor.py
laloch: Done.

xonsh/parsers/fstring_adaptor.py
laloch: Done.

xonsh/parsers/fstring_adaptor.py
laloch: Done.

xonsh/parsers/fstring_adaptor.py
laloch: Done.

laloch:
The try/except blocks were there to highlight very unlikely cases, but OK, I can live without them. 
Thanks for the review @scopatz!

xonsh/parsers/fstring_adaptor.py
scopatz: It's OK! I'll add after merge

scopatz:
This is awesome! Thanks so much @laloch!

xonsh/parsers/fstring_adaptor.py
scopatz: Oh weird, I guess we don't have parser docs linked in sphinx yet

Issue #3661, opened 2020-08-06
Xonsh invoked with wrong python (on Windows)


bobhy:
`py.exe` launcher for windows seems to choose the "wrong" interpreter when invoked with `py -3`.

In my case, the "wrong" interpreter is the system-wide one lacking Pygments, prompt_toolkit xonsh prerequisites.  The "right" interpreter is in a currently active virtual environment which does have these installed.  

For example:
<details>

```
(.venv) PS C:\Users\bobhy\src> xonsh
C:\Users\bobhy\AppData\Local\Programs\Python\Python38\lib\site-packages\xonsh\__amalgam__.py:16266: UserWarning: prompt_toolkit is not available, using readline instead.
  warnings.warn(
Skipping setup. Because no `readline` implementation available.
            Please install a backend (`readline`, `prompt-toolkit`, etc) to use
            `xonsh` interactively.
            See https://github.com/xonsh/xonsh/issues/1170
```

</details>

## Expected Behavior

Invoking `xonsh` from CLI command line (whether `cmd.exe` or `powershell` on Windows) should run with the python interpreter found first on the PATH (which should correspond to an activated virtual environment, if `venv` has done its job).

## Current Behavior

This seems to be a bug in [py.exe](https://docs.python.org/3/using/windows.html#launcher) when there are multiple python interpreters in PATH.

```
(.venv) PS C:\Users\bobhy\src> echo $env:PATH
C:\Users\bobhy\src\xonsh\.venv\scripts;C:\WINDOWS\system32;C:\WINDOWS;C:\WINDOWS\System32\Wbem;C:\WINDOWS\System32\WindowsPowerShell\v1.0\;C:\WINDOWS\System32\OpenSSH\;C:\Program Files\Intel\WiFi\bin\;C:\Program Files\Common Files\Intel\WirelessCommon\;C:\Program Files\Git\cmd;C:\Users\bobhy\bin;C:\Users\bobhy\AppData\Local\Programs\Python\Launcher\;C:\Users\bobhy\AppData\Local\Microsoft\WindowsApps;C:\Users\bobhy\AppData\Local\Programs\Microsoft VS Code\bin;C:\dev\nvm;C:\Program Files\nodejs;C:\Users\bobhy\AppData\Local\Programs\Azure Data Studio\bin
(.venv) PS C:\Users\bobhy\src> py -3 -c 'import sys; print(sys.executable, sys.version)'
[...]

Issue #3637, opened 2020-06-30
Attempting to install xonsh on Windows with pipx fails


jaraco:
## Expected Behavior

Using `pipx install xonsh` should install a working `xonsh` command.

## Current Behavior

Using `pipx install xonsh` results in an error, `No module named xonsh`.

It seems the executable wrappers (`.bat` files, yuck) are unaware of the Python virtual environment created by pipx.

```
PS C:\WINDOWS\system32> pipx install xonsh
  installed package xonsh 0.9.18, Python 3.8.3
  These apps are now globally available
    - xon.sh
    - xonsh-cat.bat
    - xonsh.bat
done!
←[?25h
PS C:\WINDOWS\system32> xonsh
C:\Python38\python.exe: No module named xonsh
PS C:\WINDOWS\system32> cat ~/.local/bin/xonsh.bat
@echo off
call :s_which py.exe
if not "%_path%" == "" (
  py -3 -m xonsh %*
) else (
  python -m xonsh %*
)

goto :eof

:s_which
  setlocal
  endlocal & set _path=%~$PATH:1
  goto :eof
```

This issue does not seem to occur with other programs installed by pipx, which end up with `.exe` wrappers.

## Steps to Reproduce

Using a late Windows 10, install Python 3.8 to C:\Python38 (default for Chocolatey). Then use `py -m pip install pipx` to install pipx system-wide, then use the pipx command to install xonsh.

melund:
I recall that we don't generate "exe" wrappers because they were slow to start up. Or used to be at least. You can see here:

https://github.com/xonsh/xonsh/blob/master/setup.py#L392-L401

The "exe" wrapper comes back if `'console_scripts': ['xonsh = xonsh.main:main'],` is added back to "entry_points".  

Maybe you could make the current batch scripts work with pipx? Or maybe we could add the entry point back for Windows only. I suspect the batch script wrapper on windows are probably also slow. So maybe the difference is marginal here. Also this will  probably not affect xonsh installed using conda since it creates its own exe wrappers.

jaraco:
> You can see [here](https://github.com/xonsh/xonsh/blob/master/setup.py#L392-L401)

I'm pretty sure that comment is stale.

[...]

PR #3662, opened 2020-08-07
 drop dependence on py.exe on Windows

bobhy:
Fix for #3661.


scopatz:
CC @jaraco @melund - can you take a look at this please?

melund:
I always only use mini-conda, so haven't run into the problem. I think it sounds reasonable to drop the py.exe wrappers: 

It was orignally suppose to solve https://github.com/xonsh/xonsh/issues/1395 , caused by having both 2.7 and 3 installed with 2.7 beeing the default python installation. 

@BYK  what do you think. You were the original author of this (https://github.com/xonsh/xonsh/pull/1578)

melund:
I really think it is OK to drop py.exe. But maybe we better let pip/setuptools generate the entry point. The reason we have a "xonsh.bat" file in the first place was that the entry points generated by setuptools were very slow to start. 

I think that is no longer an issue. See @jaraco  comments here: https://github.com/xonsh/xonsh/issues/3637. 

 So, @bobhy, could you try to remove the script from here:
https://github.com/xonsh/xonsh/blob/master/setup.py#L326-L328

And then add a console_script entry point here: (even though the warnings says not to).
https://github.com/xonsh/xonsh/blob/master/setup.py#L378-L387

bobhy:
This is coming along.  Apparently, console script startup is not slower than existing .bat files on Windows, here's a comparison:
```
(.venv) PS C:\Users\bobhy\src\xonsh> (measure-command -ex {python -c 'pass'}).milliseconds
74
(.venv) PS C:\Users\bobhy\src\xonsh> (measure-command -ex {python -c 'pass'}).milliseconds
65
(.venv) PS C:\Users\bobhy\src\xonsh> (measure-command -ex {python -c 'pass'}).milliseconds
67
(.venv) PS C:\Users\bobhy\src\xonsh> (measure-command -ex {xonsh-bat.bat -c 'pass'}).milliseconds
309
(.venv) PS C:\Users\bobhy\src\xonsh> (measure-command -ex {xonsh-bat.bat -c 'pass'}).milliseconds
839
(.venv) PS C:\Users\bobhy\src\xonsh> (measure-command -ex {xonsh-bat.bat -c 'pass'}).milliseconds
849
(.venv) PS C:\Users\bobhy\src\xonsh> (measure-command -ex {xonsh-bat.bat -c 'pass'}).milliseconds
847
(.venv) PS C:\Users\bobhy\src\xonsh> (measure-command -ex {xonsh.exe -c 'pass'}).milliseconds
96
(.venv) PS C:\Users\bobhy\src\xonsh> (measure-command -ex {xonsh.exe -c 'pass'}).milliseconds
78
(.venv) PS C:\Users\bobhy\src\xonsh> (measure-command -ex {xonsh.exe -c 'pass'}).milliseconds
32
```
I'd appreciate someone running a check on linux, too.

Next checkin will eliminate the .bat files, so save a copy before trying this yourself.

melund:
Looks promising. I am sure it will be the same on Linux. Looking forward to this 👍

bobhy:
Hmmm.  There seems to be a solid 100ms penalty for setuptuils on top of a 200ms via old script (which itself is too slow).  The following is with XONSH_DEBUG unset but running in Windows Subsystem for Linux, so probably not the most performant linux environment.

```
(.venv) bobhy@Jessie-pc:~/src/xonsh$ cat .venv/bin/xonsh
#!/home/bobhy/src/xonsh/.venv/bin/python
# EASY-INSTALL-ENTRY-SCRIPT: 'xonsh==0.9.18','console_scripts','xonsh'
__requires__ = 'xonsh==0.9.18'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
        load_entry_point('xonsh==0.9.18', 'console_scripts', 'xonsh')()
    )
(.venv) bobhy@Jessie-pc:~/src/xonsh$ cat .venv/bin/xonsh-scr
#!/usr/bin/env -S python3 -u

from xonsh.main import main
main()
(.venv) bobhy@Jessie-pc:~/src/xonsh$ time xonsh -c 'pass'

real    0m0.291s
user    0m0.244s
sys     0m0.066s
(.venv) bobhy@Jessie-pc:~/src/xonsh$ time xonsh -c 'pass'

real    0m0.289s
user    0m0.243s
sys     0m0.066s
(.venv) bobhy@Jessie-pc:~/src/xonsh$ time xonsh -c 'pass'

real    0m0.377s
user    0m0.358s
sys     0m0.044s
(.venv) bobhy@Jessie-pc:~/src/xonsh$ time xonsh-scr -c 'pass'

real    0m0.188s
user    0m0.159s
sys     0m0.040s
(.venv) bobhy@Jessie-pc:~/src/xonsh$ time xonsh-scr -c 'pass'

real    0m0.190s
user    0m0.158s
sys     0m0.042s
(.venv) bobhy@Jessie-pc:~/src/xonsh$ time xonsh-scr -c 'pass'

real    0m0.189s
user    0m0.157s
sys     0m0.042s
(.venv) bobhy@Jessie-pc:~/src/xonsh$
```

@scopatz, would you take a look at startup times and see whether we're suffering a regression with the old startup script?

melund:
Otherwise, just make the change for windows.

bobhy:
@jaraco, I cannot seem to get rid of the ~ 100ms startup penalty for using the console scripts, even with latest and greatest setuptools and distutils.  Can you spot what I'm doing wrong here? If not, I think I should revert to the hand-crafted startup scripts.

xonsh/environ.py
```diff
@@ -1675,6 +1667,14 @@ def DEFAULT_VARS():
             "Logging to a file is not enabled by default.",
         ),
     }
+    if hasattr(locale, "LC_MESSAGES"):
+        dv["LC_MESSAGES"] = Var(
```
scopatz, line 1671: @melund - is this correct?

scopatz:
This actually fails for me on linux for some reason:

```sh
$ /home/scopatz/.local/bin/xonsh
  File "/home/scopatz/.local/bin/xonsh", line 33, in <module>
    sys.exit(load_entry_point('xonsh==0.9.18', 'console_scripts', 'xonsh')())
  File "/home/scopatz/.local/bin/xonsh", line 25, in importlib_load_entry_point
    return next(matches).load()
StopIteration
```

And the console script looks like:

```py
#!/home/scopatz/miniconda/bin/python3
# EASY-INSTALL-ENTRY-SCRIPT: 'xonsh==0.9.18','console_scripts','xonsh'
import re
import sys

# for compatibility with easy_install; see #2198
__requires__ = 'xonsh==0.9.18'

try:
    from importlib.metadata import distribution
except ImportError:
    try:
        from importlib_metadata import distribution
    except ImportError:
        from pkg_resources import load_entry_point


def importlib_load_entry_point(spec, group, name):
    dist_name, _, _ = spec.partition('==')
    matches = (
        entry_point
        for entry_point in distribution(dist_name).entry_points
        if entry_point.group == group and entry_point.name == name
    )
    return next(matches).load()


globals().setdefault('load_entry_point', importlib_load_entry_point)


if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(load_entry_point('xonsh==0.9.18', 'console_scripts', 'xonsh')())
```

xonsh/environ.py
melund: It is not related to this PR. Instead it belongs to the discussion here https://github.com/xonsh/xonsh/pull/3658#issuecomment-670959034

melund: So it should probably be in its own PR

jaraco:
> I cannot seem to get rid of the ~ 100ms startup penalty

If you see `import pkg_resources` in the startup script, that means you're getting the slow, legacy version of an entry point script, probably because xonsh was installed with `setup.py install`, `easy-install`, or `pip install -e`. The first two of those are deprecated and should be replaced with `pip install`. The `pip install -e` still depends on easy_install, so will generate the slower entry points, but that should be an acceptable tradeoff during development. Production environments should always use `pip install` and should get the fast wrapper scripts.

`easy_install` is deprecated and users should expect to get degraded behavior if using it.

Are there any conditions where installation with `pip install` aren't fast?

> This actually fails for me on linux for some reason:

Looks like miniconda is involved. The script is a little cringey, but I understand what it's trying to do. The StopIteration error means that importlib.metadata is failing to find any metadata for a package called `xonsh`. Can you find out why that is or maybe suggest how I can set up a miniconda (silent, non-interactive install on Ubuntu would be nice) and how you've installed xonsh with that?

scopatz:
Yeah, I was doing a setup.py install.  I can try pip and report back

scopatz:
OK - I can confirm that this is working with `pip install`, and has good timings of the order:

```
real 0.11
user 0.09
sys 0.02
```

Which is pretty great, actually

bobhy:
Interesting that you have to use `pip install .` to get the right startup script generated.  

But mine looks very different from the one cited above:
```
(.venv) bobhy@bobhy-pc:~/src/xonsh/.venv/bin$ more xonsh
#!/home/bobhy/src/xonsh/.venv/bin/python3
# -*- coding: utf-8 -*-
import re
import sys
from xonsh.main import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

Is it what you expect to see?  It does not have any of the "entrypoint" definition that I read about.  But it is giving me about the same performance has the hand-coded script, FWIW.

tests/test_integrations.py
```diff
@@ -61,7 +61,7 @@ def run_xonsh(cmd, stdin=sp.PIPE, stdout=sp.PIPE, stderr=sp.STDOUT, single_comma
     env["XONSH_SHOW_TRACEBACK"] = "1"
     env["RAISE_SUBPROC_ERROR"] = "0"
     env["PROMPT"] = ""
-    xonsh = "xonsh.bat" if ON_WINDOWS else "xon.sh"
+    xonsh = "xonsh.exe" if ON_WINDOWS else "xonsh"
```
melund, line 64: I think you could just have "xonsh" in both cases. It will get the "exe" version on windows

bobhy: Yes, that's right.  Added to latest push, along with simplifying PATH for integration tests that run xonsh in subproc. Did you know that path to site-packages in venv in Windows does not have 'pythonxx/' where Linux does?  I don't think xonsh should try to track things in such detail, we'll forever be a step behind.

scopatz:
> Interesting that you have to use pip install . to get the right startup script generated.

Yeah that is pretty interesting and not what I expected. I thought that pip just ran setup.py, but I guess I am wrong.

BTW that script that you generated is almost exactly what I see on linux

melund:
@bobhy I still see the `xonsh.bat` files beeing generated together with the exe files. Do you see that as well?

scopatz:
Thanks @bobhy! This is loooking really good to me. Can you also please find the places in the docs / readmes / etc where we tell people not to pip install and change the text. Also, we might need to change the CI scripts too

bobhy:
I think that's all!  CI scripts were already OK.  PR ready for final review now.

CONTRIBUTING.rst
```diff
@@ -20,8 +20,11 @@ Making Your First Change
 ========================
 
 First, install xonsh from source and open a xonsh shell in your favorite
-terminal application. See installation instructions for details.
+terminal application. See installation instructions for details, but it
+is recommended to do an 'editable' install via `pip'
 
+        $ pip install -e .
```
melund, line 26: Previously new parser tables were not generated when doing an editable install. 

So for development on xonsh it is better to do a full install after changing stuff.

We can change this later.

laloch: I've been using editable installs (`pip install --no-use-pep517 --user -e .`) since forever and it always re-generates the parser tables for me.

laloch:
To whomever is going to merge this: Could you please do squash-and-merge? The commit history doesn't seem to be very helpful.

melund:
No. I don't like squash merges. It lies about the real effort made and the currency on GitHub is unfortunately commits. I consider squash merging open source contributions is rude and bad bad practice.

However, I also often make a lot back and forth dummy commits. I usually do an interactive rebase to get commits that.tell a story. @bobhy you are well come to clean up your commit history if you like. Otherwise I vote for normal merge.

CONTRIBUTING.rst
melund: Ahh. Yes. Of course. But you still need to rerun it to recreate the tables.

laloch:
> No. I don't like squash merges. It lies about the real effort made

I don't think commits like 6d5c2229ce8e3ee449c990a94d7aa30912090c1e or 2f439fef2101753b0eb67401a76e47aa9695149c do tell anything about the real effort. I can only see that the commit messages are not related to the diffs at all.

>  and the currency on GitHub is unfortunately commits.

I don't care about currencies. I do care about sanity of the master branch.

> I consider squash merging open source contributions is rude and bad bad practice.

I consider misleading commit messages and merge commits on topic branches to be bad bad bad practice. That's one bad more than squashing :wink:
Anyhow, my intention actually was not to be rude and not to bother @bobhy with manual cleanup.

> However, I also often make a lot back and forth dummy commits. I usually do an interactive rebase to get commits that.tell a story.

You see? You do, in fact, care too. I do interactive rebases all the time for the very same reason.

> @bobhy you are well come to clean up your commit history if you like. Otherwise I vote for normal merge.

@bobhy, I meant no offence. You are free to either do the cleanup manually, vote for the squash-merge, or leave the PR as is.

scopatz:
Hey @laloch - we normally don't squash. I don't feel like any offense was meant or taken here.

Let's just proceed with the merge as is.

scopatz:
Thanks a ton @bobhy!

melund:
I didn't mean that you were rude @laloch. Sorry if you took it that way. And you are right. I do care about commit history telling a story. 

I just want to ensure we don't demotivate people willing to contribute. I once [read this blog ](https://thanpol.as/javascript/Dont-Squash-its-rude-and-a-lie) on the issue which I kind of agree with.

Thanks for the good work @bobhy

Issue #3664, opened 2020-08-07
consider pyupgrade --py36-plus


Carreau:
Prompted by #3657 , which was (partially) a joke, 
this is a discussion to see if [pyupgrade](https://github.com/asottile/pyupgrade/) could be used in conjunction with black to automatically to keep the codebase using the latest python constructs. 

this for example can update all strings that support it to f-strings, update some of the conditional version_info, and a number of other things.

I think this can be used with [darker](https://github.com/akaihola/darker), to only apply to some subset of the code that was modified, though I want to add a feature to darker to not only affect non-commited changes but to give it a commit-ish, and that all-changes with respect to that commit should be considered.

PR #3665, opened 2020-08-07
Apply pyupgrade --py36-plus to all xoreutils.

Carreau:
See #3664 for some of the changes this does; mostly converting to
f-strings when possible.


scopatz:
Thanks @Carreau!

PR #3667, opened 2020-08-08
Fix link to photo in CONTRIBUTING.rst

dannysepler:


scopatz:
Thanks @dannysepler - I think this will break it for the actual docs though.

dannysepler:
ah, the link here? https://xon.sh/devguide.html oops thanks! i'll play around, and reopen if i can figure something out

scopatz:
Yep that's the link! Thanks!

PR #3668, opened 2020-08-09
callable default skips validation

scopatz:
Minor quality of life improvement for env var registration


scopatz:
@dotsdl - what is the need for the `is_callable_default`? I think I might have missed something here

scopatz:
Is it just to make it more explicit?

scopatz:
Oh I see, we pass in the env

daniel-shimon:
This is great!
Could you add a test to make sure the validation runs correctly with `pytest.raises`?
Once in `getitem` with a default and once in `setitem`

Another good test might be that the default function only runs once after multiple `getitem`s

scopatz:
Thanks for the review @daniel-shimon! Shouldn't those test you suggested be covered by other tests in the suite?  This really just affects the logic of `register()`

scopatz:
I have resolved the conflicts with master and I think that this is ready for review @xonsh/xore!

daniel-shimon:
> Thanks for the review @daniel-shimon! Shouldn't those test you suggested be covered by other tests in the suite?  This really just affects the logic of `register()`

I'm not sure about other tests, but these changes do affect the scenarios I suggested to test

anki-code:
hi! Maybe there is a real life example of using callable default? It will be great to look at it.

jnoortheen:
@gforsyth would you be kind to check this one too?

gforsyth:
Thanks for the ping @jnoortheen !

Thanks for putting this in @scopatz and thanks @daniel-shimon for the review!

PR #3671, opened 2020-08-10
Fix regression with LC_MESSAGES, don't add to environment unconditionally [no news article needed].

bobhy:
Improved fix for #3658, does not add `LC_MESSAGES` to environment if it's not defined in `locale`. Preapproved by @melund.


melund:
Thanks a lot @bobhy

Issue #3635, opened 2020-06-26
CI failure - AttributeError: module 'builtins' has no attribute '__xonsh__'


bobhy:
CI test failure that doesn't happen locally and doesn't seem to be related to any (recent) code change.  
Exact point of failure fluctuates from run to run and (is this just my paranoia?) time of day or week.

So far, I have always been able to get a clean run by manually re-running the jobs on the same code 
 (in my staging repo on GitHub there's a `re-run all jobs` option, though not on the PR to `xonsh/xonsh`)

Does anyone have insight as to what's actually going on? 

A superficial review of the failing  code (traceback, below) shows a sequence of statements successfully referring to `builtins.__xonsh__`, then boom! one fails, and it is not the last in the sequence, either.  This colud be a race condition between tests running in parallel, but I do not believe the Xonsh CI action invokes `pytest` that way.  

### Traceback (if applicable)

<details>

```
==================================== ERRORS ====================================
____________ ERROR at setup of test_premain_with_file_argument[-i] _____________

monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7fbe619419a0>
xonsh_events = <xonsh.events.EventManager object at 0x7fbe5ec582e0>

    @pytest.fixture
    def xonsh_builtins(monkeypatch, xonsh_events):
        """Mock out most of the builtins xonsh attributes."""
        old_builtins = set(dir(builtins))
        execer = getattr(getattr(builtins, "__xonsh__", None), "execer", None)
        session = XonshSession(execer=execer, ctx={})
        ensure_attached_session(monkeypatch, session)
        builtins.__xonsh__.env = DummyEnv()
        if ON_WINDOWS:
            builtins.__xonsh__.env["PATHEXT"] = [".EXE", ".BAT", ".CMD"]
        builtins.__xonsh__.shell = DummyShell()
        builtins.__xonsh__.help = lambda x: x
        builtins.__xonsh__.glob = glob.glob
        builtins.__xonsh__.exit = False
        builtins.__xonsh__.superhelp = lambda x: x
        builtins.__xonsh__.pathsearch = pathsearch
        builtins.__xonsh__.globsearch = globsearch
        builtins.__xonsh__.regexsearch = regexsearch
>       builtins.__xonsh__.regexpath = lambda x: []
E       AttributeError: module 'builtins' has no attribute '__xonsh__'

tests/conftest.py:122: AttributeError
```

</details>

## Steps to Reproduce
[...]

PR #3673, opened 2020-08-10
Fix test suite

melund:
Turned out that a few legitimate versions of the infamous 
```
AttributeError: module 'builtins' has no attribute '__xonsh__'.
```
had sneaked into the test suite. This fixes it.

### DIFF
--- a/tests/test_lib/test_os.xsh
+++ b/tests/test_lib/test_os.xsh
@@ -24,6 +24,8 @@ def test_indir():
 
 
 def test_rmtree():
+    if ON_WINDOWS:
+        pytest.skip("On Windows")
     with tempfile.TemporaryDirectory() as tmpdir:
         with indir(tmpdir):
             mkdir rmtree_test
--- a/news/test_suite.rst
+++ b/news/test_suite.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* The xonsh test suite has been cleaned up. So no more failing test. Hopefully.
+
+**Security:**
+
+* <news item>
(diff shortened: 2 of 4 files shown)


tests/test_lib/test_os.xsh
```diff
@@ -24,6 +24,8 @@ def test_indir():
 
 
 def test_rmtree():
+    if ON_WINDOWS:
+        pytest.skip("On Windows")
```
laloch, line 28: I think we have `@skip_if_on_windows` decorator for this.

melund: This is xsh file. So I am not sure that works.

melund: This paradigme was used elsewhere in `test_xxx.xsh` files.

https://github.com/xonsh/xonsh/pull/3673/files#diff-1e0c5474373de0df12e386bdcc371662R12-R13

melund:
Awesome. Tests suite passes now. 

Hopefully we will see less random failures in the future and with @scopatz splitting the test in individual yaml files we can now restart individually. 

Let get this in so we can test some of the waiting PRs

laloch:
Thanks @melund! The CI drives me crazy.

melund:
There may still be some issues with test timeout times. Especially, on MacOS CI which seems to slow.

laloch:
Yeah, I have already seen the 240s timeout to expire several times. This was, however, failing consistently.

bobhy:
Is this a fix for #3635 ?
Is the general rule that all tests which reference xonsh session attributes (__xonsh__) must use `xonsh_builtins`?

What is going on under the covers that causes this error (formerly all too common)?

melund:
Maybe. I am not sure. We sometimes saw the error popping up randomly for no reason. It is not a fix for that. We think that is related to some issue in pytest which doesn't mock xonsh_builtins correctly. 

But there were also some reproducible versions of the errors which had sneaked it. It was cases where tests were missing the `xonsh_builtins` fixture. Probably because the underlying code had changed. We had just ignored it because it looked like the other error.

Issue #3101, opened 2019-04-28
Xonsh broken on Python 3.8


carmenbianca:
## xonfig

Irrelevant because it doesn't run.

## Expected Behavior

Xonsh works correctly when you run it using a Python 3.8 interpreter.

## Current Behavior

Xonsh crashes upon starting.

### Traceback (if applicable)

```
$ python3 scripts/xonsh                                                                                                                                              
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/home/carmenbianca/Projektoj/xonsh/xonsh/proc.py", line 1709, in wait
    r = self.f(self.args, stdin, stdout, stderr, spec, spec.stack)
  File "/home/carmenbianca/Projektoj/xonsh/xonsh/proc.py", line 1204, in proxy_two
    return f(args, stdin)
  File "/home/carmenbianca/Projektoj/xonsh/xonsh/xontribs.py", line 178, in xontribs_main
    return _MAIN_XONTRIB_ACTIONS[ns.action](ns)
  File "/home/carmenbianca/Projektoj/xonsh/xonsh/xontribs.py", line 98, in _load
    xontribs_load(ns.names, verbose=ns.verbose)
  File "/home/carmenbianca/Projektoj/xonsh/xonsh/xontribs.py", line 90, in xontribs_load
    update_context(name, ctx=ctx)
  File "/home/carmenbianca/Projektoj/xonsh/xonsh/xontribs.py", line 69, in update_context
    modctx = xontrib_context(name)
  File "/home/carmenbianca/Projektoj/xonsh/xonsh/xontribs.py", line 33, in xontrib_context
    m = importlib.import_module(spec.name)
  File "/home/carmenbianca/.pyenv/versions/3.8-dev/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 773, in exec_module
  File "/home/carmenbianca/Projektoj/xonsh/xonsh/imphooks.py", line 127, in get_code
[...]

PR #3675, opened 2020-08-11
Add Python 3.8 assignment expressions ("walrus"/PEP 572)

laloch:
~~WIP; missing lexer/parser tests.~~

Suggestions welcome...

Relates to #3101

### DIFF
--- a/news/walrus.rst
+++ b/news/walrus.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Support for Python 3.8 PEP 572 assignment expressions (walrus operator).
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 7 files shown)


daniel-shimon:
Nice!

I wanted to do this myself but didn't really have the time

How about making the changes to xonsh/parsers/base.py in xonsh/parsers/v38.py instead?
For instance:

```python

def p_test_or_star_expr(self, p):
        """test_or_star_expr : test
        """test_or_star_expr : namedexpr_test
                             | star_expr
        """
        super().p_test_or_star_expr(p)

```

For all the functions you're changing

laloch:
Hi @daniel-shimon, the version you have commented on is now gone. But in general, I wanted to put as much as possible into the common base and only override in v38.py what's really specific for v3.8. For instance `namedexpr_test` can simply be `namedexpr_test : test` in older grammars and only becomes `namedexpr_test : test [COLONEQUAL test]` in v3.8. This reduces the amount of code that needs to be overriden.

laloch:
But yes, what you say @daniel-shimon, probably applies to `p_testlist_comp_*`. OK, let's change it once again...

scopatz:
This is awesome! Thanks @laloch!

PR #3677, opened 2020-08-11
Create ExecAlias for aliases that contain IO operators

daniel-shimon:
Currently after setting
```python
aliases['test'] = 'ls | wc'
```
Running `test` will result in `ls | wc` being printed to the screen.

The reason is `Aliases.__setitem__` parses this into a list instead of into an `ExecAlias` object.
The fix is to add the IO operators (`>`, `<`, and `|`) to the regex that decides whether to create an `ExecAlias` or not.

### DIFF
--- a/tests/test_aliases.py
+++ b/tests/test_aliases.py
@@ -128,3 +128,20 @@ def test_subprocess_logical_operators(xonsh_execer, xonsh_builtins, alias):
     ales = make_aliases()
     ales["echocat"] = alias
     assert isinstance(ales["echocat"], ExecAlias)
+
+
+@pytest.mark.parametrize(
+    "alias",
+    [
+        "echo 'hi' | grep h",
+        "echo 'hi' > file",
+        "cat < file",
+        "COMMAND1 e>o < input.txt | COMMAND2 > output.txt e>> errors.txt",
+        "echo 'h|i' | grep h",
+        "echo 'h|i << x > 3' | grep x",
+    ],
+)
+def test_subprocess_io_operators(xonsh_execer, xonsh_builtins, alias):
+    ales = make_aliases()
+    ales["echocat"] = alias
+    assert isinstance(ales["echocat"], ExecAlias)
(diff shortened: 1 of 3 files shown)


gforsyth:
Hey @daniel-shimon -- first, wow, sorry that this slipped through the cracks.  Very much our bad on that.

This looks like a reasonable solution to me and has that rare quality of being a PR that is submitted before (by several months!) an issue that it will close (#3891).

I've been away for a bit from xonsh and don't remember all of the knock-on effects of some of the aliasing stuff -- @laloch do you have a moment to take a quick look at this to double-check me?

laloch:
Nope, not good. It, unfortunately, fails for simple `echo 'h|i' | grep 'h'`.

laloch:
No, wait! It doesn't! I probably did something wrong the first time.

tests/test_aliases.py
```diff
@@ -128,3 +128,18 @@ def test_subprocess_logical_operators(xonsh_execer, xonsh_builtins, alias):
...
+
+@pytest.mark.parametrize(
+    "alias",
+    [
+        "echo 'hi' | grep h",
+        "echo 'hi' > file",
+        "cat < file",
+        "COMMAND1 e>o < input.txt | COMMAND2 > output.txt e>> errors.txt",
+    ],
+)
```
laloch, line 141: Could you please add some tests for IO redirection characters within command arguments (e.g. `echo 'h|i' | grep h`)?

laloch (changes requested):
Hi @daniel-shimon, this looks great! Could you please just add more test cases for some potentially dangerous inputs?

daniel-shimon:
Hi guys, it's all good, I've forgotten about this PR myself 😁

@laloch - if I understand correctly, ExecAlias behaves like normal xonsh (which would also work for simple commands without IO redirections), while creating a list alias is an optimization.
So how can it be dangerous to create an ExecAlias instead of a list alias?

tests/test_aliases.py
daniel-shimon: Done

laloch (approved):

laloch:
Thanks a lot @daniel-shimon! Sorry for the delay.

Closes #3891

PR #3679, opened 2020-08-12
Only return string keys while iterating over __xonsh__.env

laloch:
Since #3377, the environment contains compiled regular expression patterns. Those `re.Pattern` objects are being returned as environment variable keys while iterating over `__xonsh__.env`. That unfortunately breaks the existing code, that expects the keys to be strings. For instance
```
  File "/home/str/code/xonsh/xonsh/completer.py", line 36, in complete
  File "/home/str/code/xonsh/xonsh/completers/path.py", line 335, in complete_path
    paths.update(filter(filtfunc, _env(prefix)))
  File "/home/str/code/xonsh/xonsh/completers/path.py", line 103, in _env
    return {
  File "/home/str/code/xonsh/xonsh/completers/path.py", line 104, in <setcomp>
    "$" + k for k in builtins.__xonsh__.env if get_filter_function()(k, key)
  File "/home/str/code/xonsh/xonsh/completers/tools.py", line 7, in _filter_normal
    return s.startswith(x)

Exception 're.Pattern' object has no attribute 'startswith'
```
This fixes the issue.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1971,7 +1971,9 @@ class Env(cabc.MutableMapping):
             return default
 
     def __iter__(self):
-        yield from (set(self._d) | set(self._vars))
+        for key in set(self._d) | set(self._vars):
+            if isinstance(key, str):
+                yield key
 
     def __contains__(self, item):
         return item in self._d or item in self._vars
--- a/news/env-yield-str.rst
+++ b/news/env-yield-str.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Iterating over ``${...}`` or ``__xonsh__.env`` yields only string
+  values again.
+
+**Security:**
+
+* <news item>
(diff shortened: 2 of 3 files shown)


astronouth7303:
The container automation hit this.

https://github.com/xonsh/container/blob/3853876b033c8c39ab1c66f263ba5bbd279255dc/action.xsh#L7-L11

xonsh/environ.py
```diff
@@ -1971,7 +1971,9 @@ def get(self, key, default=None):
             return default
 
     def __iter__(self):
-        yield from (set(self._d) | set(self._vars))
+        for key in set(self._d) | set(self._vars):
+            if isinstance(key, str):
+                yield key
```
scopatz, line 1976: This should maybe return the pattern itself (a string) if this is a regex.  As it stands, I think this will break some of the docs

laloch: How can it break the docs? Returning a string representation of the regex does not make any sense, because you can not address such a key by string.

scopatz: We generate docs for the regular expressions keys.  See `docs/conf.py` down at the bottom.  Now these keys will be skipped I believe

laloch: OK, I see. `conf.make_envvars()` iterates over `xonsh.environ.DEFAULT_VARS`, so it's not affected by this change.
If you think that it have sense to return the patterns, I can open a PR for cbd0001 that allows access to the RE variables.

scopatz: I think maybe we could have a separate iterator which returns all the keys.  I am fine with the default `__iter__` method returning just strings.

PR #3682, opened 2020-08-12
Fix vended prompt_toolkit packages not beeing installed with pip

melund:
This fixes the problem where vended prompt-toolkit packages where not install when using `pip install .`

It uses setuptools.find_packages() to locate the prompt-toolkit packages. It works for me. @laloch can also test this?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ try:
 except ImportError:
     pass
 
-from setuptools import setup
+from setuptools import setup, find_packages
 from setuptools.command.sdist import sdist
 from setuptools.command.install import install
 from setuptools.command.develop import develop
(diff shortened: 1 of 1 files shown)


laloch:
Yes, I can confirm, that `pip install git+https://github.com/xonsh/xonsh.git@refs/pull/3682/head#egg=xonsh` now works.
Thanks!

laloch:
The CI runs twice in parallel once again. It seems that it triggers on both `push` and `pull_request` events simultaneously for PRs from local branches.

laloch:
No news entry needed.
Thanks a lot @melund!

melund:
> The CI runs twice in parallel once again. It seems that it triggers on both push and pull_request events simultaneously for PRs from local branches.

Maybe this can help:
https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662

melund:
@laloch  fix here: https://github.com/xonsh/xonsh/pull/3683

PR #3685, opened 2020-08-13
Allow for setting cursor position within expanded abbrev

laloch:
`abbrevs` expansion now allows for setting cursor to a specific position within the expanded abbrev. For instance
```
abbrevs["eswap"] = "with ${...}.swap(<edit>):\n    "
```
 expands `eswap` as you type to environment context manager ``swap()`` syntax and places the cursor at the position of the `<edit>` mark removing the mark itself in the process.


scopatz:
Thanks!

PR #3688, opened 2020-08-13
Add $HOSTNAME and $HOSTTYPE environment variables

laloch:
Adds `$HOSTNAME` and `$HOSTTYPE` variables that the user might expect to exist.

`$HOSTNAME` defaults to `platform.uname().node`.
`$HOSTTYPE` defaults to `platform.uname().machine`.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -12,6 +12,7 @@ import contextlib
 import collections
 import collections.abc as cabc
 import subprocess
+import platform
 
 from xonsh import __version__ as XONSH_VERSION
 from xonsh.lazyasd import LazyObject, lazyobject
@@ -86,7 +87,6 @@ from xonsh.ansi_colors import (
     ansi_style_by_name,
 )
 import xonsh.prompt.base as prompt
-
 
 events.doc(
     "on_envvar_new",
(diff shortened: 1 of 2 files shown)


laloch:
OK, there's no `os.uname()` on Windows :roll_eyes:

xonsh/environ.py
```diff
@@ -87,6 +87,7 @@
 )
 import xonsh.prompt.base as prompt
 
+from platform import uname
```
scopatz, line 90: Please `import platform` for amalgamation

xonsh/environ.py
```diff
@@ -942,6 +943,20 @@ def DEFAULT_VARS():
             "Note: ``erasedups`` is supported only in sqlite backend).",
             doc_store_as_str=True,
         ),
+        "HOSTNAME": Var(
+            is_string,
+            ensure_string,
+            ensure_string,
+            uname().node,
```
scopatz, line 950: This should be something like: `default_value(lambda env: uname.node)` so it is not called when the deafults are created.

xonsh/environ.py
```diff
@@ -942,6 +943,20 @@ def DEFAULT_VARS():
...
+            "Automatically set to the name of the current host.",
+        ),
+        "HOSTTYPE": Var(
+            is_string,
+            ensure_string,
+            ensure_string,
+            uname().machine,
```
scopatz, line 957: Similarly, this should be `default_value(lambda env: uname.machine)`

xonsh/environ.py
laloch: Oops, sorry!

xonsh/environ.py
laloch: I actually wanted it to be called only once.

scopatz:

xonsh/environ.py
laloch: Done.

xonsh/environ.py
laloch: Done.

laloch: Done.

xonsh/environ.py
laloch: Done.

scopatz:
Thanks!

PR #3690, opened 2020-08-14
ANSI escape code support in $PROMPT/$RIGHT_PROMPT

dyuri:
Adds ANSI escape code support in $PROMPT/$RIGHT_PROMPT, so we can use 3rd party prompt generators like `powerline` or `starship`.


xonsh/ptk_shell/shell.py
```diff
@@ -215,6 +216,8 @@ def prompt_tokens(self):
             p = self.prompt_formatter(p)
         except Exception:  # pylint: disable=broad-except
             print_exception()
+        if "\x1b" in p:
+            return ANSI(p)
```
scopatz, line 220: Should this check happen on each individual token? I think if xonsh coloring and ANSI coloring is mixed, it won't display correctly.

scopatz:
Thanks for putting this in @dyuri

xonsh/ptk_shell/shell.py
dyuri: Yep, currently it's either ANSI or *xonsh* coloring, but you're right, it might be possible to mix them if this is done after *xonsh* color tokenization, I'll give it a try.

scopatz: Awesome, thanks!

dyuri: Ok, it was a bit more complicated than I imagined, but got that finally, mixed formatting works, i.e.:
```
$PROMPT=lambda: $(starship prompt) + "\n{RED}{user}{GREEN}> "
```

xonsh/ptk_shell/tokenize_ansi.py
```diff
@@ -0,0 +1,28 @@
+from prompt_toolkit import ANSI
```
scopatz, line 1: This probably doesn't need to be in its own module since it is only used by shell

scopatz:
Thanks @dyuri - Can you move this function over to shell and maybe add a couple of tests for it now that it is broken out into a separate function? Thanks!

dyuri:
`tokenize_ansi` moved into `shell` + added some unit tests to verify ANSI tokenization

scopatz:
Thanks a ton @dyuri!

Issue #3033, opened 2019-02-28
Add support for clearing or selectively purging session history


jaraco:
## Current Behavior

Currently, it's not straightforward to clear the session history, such as after accidentally leaking sensitive information onto a command.

In [this discussion](https://gitter.im/xonsh/xonsh?at=5c77e94f5c5c1f016d4f8f5f) Anthony and I discuss the considerations and possible solutions.

## Expected Behavior

There should be a straightforward way to clear the history for a session. Bonus if the technique also allows to clear the last 'n' commands or otherwise purge a slice of commands from the history.

Anthony suggests a `history clear` command as a high-level interface. I'd also like to see a low-level interface for more direct manipulation.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

scopatz:
Thanks for opening this! Pull requests welcome

anki-code:
I'm going to close this issue because it has unfocused description:
* We already have `history clear` command now.
* We have [xontrib-history-encrypt](https://github.com/anki-code/xontrib-history-encrypt/) xontrib to encrypt the history to prevent leaking sensitive data (with `dummy` option that disables history)
* In case of using sqlite history there is simple way to dive into history or delete some rows - https://github.com/xonsh/xonsh/issues/3401#issuecomment-835272083

jaraco:
Sounds good. Thanks for the consideration. Would it also work to `del __xonsh__.history[-5:]`? Looks like that doesn't work yet. That would be the next iteration on this feature I'd like to see:

```
$ del __xonsh__.history[-5:]
AttributeError: __delitem__
```

PR #3691, opened 2020-08-14
History clear

Itay4:
Fixes https://github.com/xonsh/xonsh/issues/3033

### DIFF
--- a/xonsh/history/main.py
+++ b/xonsh/history/main.py
@@ -224,7 +224,7 @@ def _XH_HISTORY_SESSIONS():
     }
 
 
-_XH_MAIN_ACTIONS = {"show", "id", "file", "info", "diff", "gc", "flush"}
+_XH_MAIN_ACTIONS = {"show", "id", "file", "info", "diff", "gc", "flush", "clear"}
 
 
 @functools.lru_cache()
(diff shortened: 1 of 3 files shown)


laloch:
Thank's for opening this @Itay4.
It has, however, several drawbacks:
- It doesn't work with SQLite history backend at all.
- It doesn't work with JSON backend, if the history buffer has already been flushed to disk.
- It doesn't pass any parameters to the new history object initializer.

xonsh/history/main.py
```diff
@@ -417,5 +420,8 @@ def history_main(
         hf = hist.flush()
         if isinstance(hf, threading.Thread):
             hf.join()
+    elif ns.action == "clear":
+        builtins.__xonsh__.history = builtins.__xonsh__.history.__class__()
+        print("History cleared")
```
scopatz, line 425: please add `file=sys.stderr`

xonsh/history/main.py
```diff
@@ -417,5 +420,8 @@ def history_main(
         hf = hist.flush()
         if isinstance(hf, threading.Thread):
             hf.join()
+    elif ns.action == "clear":
+        builtins.__xonsh__.history = builtins.__xonsh__.history.__class__()
```
scopatz, line 424: Doesn't this just get a new history object?

scopatz:
Thanks for putting this in @Itay4! I agree with @laloch that this should be fleshed out a little more

PR #3693, opened 2020-08-15
Knight vs Snail should work on both Sphinx docs and CONTRIBUTING.rst

dannysepler:
It feels a tad hacky, but getting the image to render in and out of the symlink is tricky to do! Totally fine to prefer this pull request not be merged 👍

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -3,7 +3,7 @@
 =================
 Developer's Guide
 =================
-.. an image URL works for CONTRIBUTING.rst and the symlinked docs/devguide.html
+.. an image URL works for CONTRIBUTING.rst and the symlinked docs/devguide.rst
 .. image:: https://raw.githubusercontent.com/xonsh/xonsh/master/docs/_static/knight-vs-snail.jpg
    :width: 80 %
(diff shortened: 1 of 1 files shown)


scopatz:
Huh, will that actually copy the file into the sphinx `build/html/_static` directory?

dannysepler:
it becomes an external image reference in both cases (pointing to knight-vs-snail on the master branch)

Screen Shot 2020-08-15 at 11 06 03 AM

That's why it feels a tad hacky and totally fine if not desired!

scopatz:
Yeah, that would break locally hosted versions of the docs then.  Maybe it would be better, to have a third file which both include? Or have conf.py do rewrite this part of the file?

dannysepler:
Are locally hosted docs what "cd docs/; make html" does? If so that's what my screenshot came from. Happy to look into the conf.py option as well!

scopatz:
Right, yeah that is what `make html` does. And yeah, if you are online the image will still show up. My concern is more about having all of the assets available and displayed when offline

dannysepler:
eh, i'll reopen if i can figure somethin out

PR #3696, opened 2020-08-16
Env.rawkeys()

scopatz:
fixes issue in https://github.com/xonsh/xonsh/pull/3679#discussion_r470211364


laloch:
The test failures seem to be legit this time.

scopatz:
This is ready for review again! Thanks @laloch

scopatz:
Oh the test failure from the previous run was that in Python <=3.6, compiled regeular expressions were some private, internal match object.  They finally gave it a real API in 3.7

PR #3700, opened 2020-08-18
locked issue

### LINKED ISSUE #3699: KeyError `locked` in latest build of xonsh
I recently updated to the latest version of xonsh and also installed a bespoke branch of ptk2. When I did, I started getting the error on startup:

```
~ $ Exception in thread Thread-10:
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/Users/jaraco/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/history/__amalgam__.py", line 372, in run
    files = self.files(only_unlocked=True)
  File "/Users/jaraco/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/history/__amalgam__.py", line 426, in files
    if lj["locked"] and lj["ts"][0] < boot:
  File "/Users/jaraco/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/__amalgam__.py", line 783, in __getitem__
    rtn = self._getitem_mapping(key)
  File "/Users/jaraco/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/__amalgam__.py", line 767, in _getitem_mapping
    offset = self.offsets[key]
KeyError: 'locked'
```

I then re-installed the latest stable version of prompt_toolkit, but the error continues.

The prompt works after the error is emitted.

I have not tried troubleshooting further.

Given that others haven't reported this issue, I suspect it's somewhat localized to my environment (and history).

## xonfig

<details>

```
~ $ xonfig
| xonsh            | 0.9.19              |
| Git SHA          | 74450cb5            |
| Commit Date      | Aug 9 13:27:55 2020 |
| Python           | 3.8.5               |
| prompt toolkit   | 3.0.6               |
```

</details>

## Expected Behavior

Prompt should startup without error.

scopatz:
Closes #3699

### DIFF
--- a/news/unlocked.rst
+++ b/news/unlocked.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Addressed robustness issue with ``"locked"`` history key not
+  being present at startup.
+
+**Security:**
+
+* <news item>
--- a/xonsh/history/json.py
+++ b/xonsh/history/json.py
@@ -202,12 +202,7 @@ class JsonHistoryGC(threading.Thread):
                 # info: file size, closing timestamp, number of commands, filename
                 ts = lj.get("ts", (0.0, None))
                 files.append(
-                    (
-                        ts[1] or ts[0],
-                        len(lj.sizes["cmds"]) - 1,
-                        f,
-                        cur_file_size,
-                    ),
+                    (ts[1] or ts[0], len(lj.sizes["cmds"]) - 1, f, cur_file_size,),
                 )
                 lj.close()
                 if xonsh_debug:
--- a/xonsh/history/json.py
+++ b/xonsh/history/json.py
@@ -200,9 +200,10 @@ class JsonHistoryGC(threading.Thread):
                 if only_unlocked and lj.get("locked", False):
                     continue
                 # info: file size, closing timestamp, number of commands, filename
+                ts = lj.get("ts", (0.0, None))
                 files.append(
                     (
-                        lj["ts"][1] or lj["ts"][0],
+                        ts[1] or ts[0],
                         len(lj.sizes["cmds"]) - 1,
                         f,
                         cur_file_size,
--- a/xonsh/history/json.py
+++ b/xonsh/history/json.py
@@ -188,7 +188,7 @@ class JsonHistoryGC(threading.Thread):
                     files.append((os.path.getmtime(f), 0, f, cur_file_size))
                     continue
                 lj = xlj.LazyJSON(f, reopen=False)
-                if lj["locked"] and lj["ts"][0] < boot:
+                if lj.get("locked", False) and lj["ts"][0] < boot:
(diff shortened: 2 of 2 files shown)


scopatz:
@xonsh/xore - this is ready for review

jaraco:
Looking good, but after this change, my shell instead emits:

```
~ $ Exception in thread Thread-10:
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/Users/jaraco/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/history/__amalgam__.py", line 372, in run
    files = self.files(only_unlocked=True)
  File "/Users/jaraco/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/history/__amalgam__.py", line 440, in files
    lj["ts"][1] or lj["ts"][0],
  File "/Users/jaraco/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/__amalgam__.py", line 783, in __getitem__
    rtn = self._getitem_mapping(key)
  File "/Users/jaraco/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/__amalgam__.py", line 767, in _getitem_mapping
    offset = self.offsets[key]
KeyError: 'ts'
```

scopatz:
Ok thanks @jaraco - here is another fix for that issue

jaraco:
Yep - that did the trick. Thanks!

scopatz:
Awesome! This is ready for review again @xonsh/xore

laloch:
I wonder how those attributes could ever be missing. They should always be passed to the history object initializer and stored immediately in the history file.

scopatz:
Some sort of weird partial write maybe?

laloch:
I can imagine, that this could happen due to concurrency between several Xonsh instances running (or rather starting) in parallel. The reproducibility of this particular error does, however, suggest, that @jaraco has one or more corrupted history files. I think we should rather handle the `KeyError` exception and ignore the affected file(s) altogether.

scopatz:
Yeah, this is a minimal fix to get things working.  Happy to take suggestions on what to do in these cases

jaraco:
After adopting this latest version, I was prompted about a very large history (65k) being over the gc limit (8k), so I ran the gc command to clean things up. I hope that limits the issues I encounter, and might have been a factor in the conditions leading to the corrupted history. I would prefer when there's corrupted history to retain as much as can be reasonably retained.

laloch:
> I was prompted about a very large history (65k) being over the gc limit (8k)

Yeah, that's because the history garbage collection didn't work for quite a some time, so the old history didn't get deleted.

> I would prefer when there's corrupted history to retain as much as can be reasonably retained.

The JSON history is saved to per-session files, so if one o them gets corrupted, you only lose a history from single session.

BYK (approved):
This LGTM until we figure out the root cause.

laloch:
Thanks @scopatz!

scopatz:
Thanks All!

Issue #3703, opened 2020-08-19
settitle() corrupts prompt in Windows Terminal and cmd.exe


tc21:
On Windows, when using Windows Terminal, sometimes when I run a program, the prompt gets corrupted with parts of some ANSI control codes.

![demo2](https://user-images.githubusercontent.com/14813183/90605770-a6ad6b00-e1b3-11ea-9db5-e4b2337247be.png)

I've figured out that this started happening because https://github.com/xonsh/xonsh/pull/3377 gave `$ANSICON` a default value, which broke this code in BaseShell.settitle:

```
    if ON_WINDOWS and "ANSICON" not in env:
#                     ^^^^^^^^^^^^^^^^^^^^ this is broken
            kernel32.SetConsoleTitleW(t)
        else:
            with open(1, "wb", closefd=False) as f:
                # prevent xonsh from answering interactive questions
                # on the next command by writing the title
                f.write("\x1b]0;{0}\x07".format(t).encode())
                f.flush()
```

Remarks: 
- A similar thing happens in cmd.exe and powershell.exe.
- I was able to consistently reproduce it by running a git for Windows exe, but I couldn't figure out the specific reasons and circumstances that caused it to happen. if I changed `"\x1b]0;{0}\x07"` to `"\x1b]0;{0}\x07\x1b["` it would fix it in Windows Terminal (but either not help or break it elsewhere), for example.
- `$ANSICON` isn't the only environment variable that got a default value: a quick compare got me this list: `{'XONSH_SOURCE', 'OLDPWD', 'VIRTUAL_ENV', 'TERM', 'XONSH_INTERACTIVE', 'XONSH_GITSTATUS_*', 'ANSICON'}`

## xonfig

<details>

```
+------------------+-------------------------------------------------------+
| xonsh            | 0.9.19                                                |
| Python           | 3.8.2                                                 |
| PLY              | 3.11                                                  |
| have readline    | False                                                 |
| prompt toolkit   | 2.0.10                                                |
| shell type       | prompt_toolkit                                        |
| pygments         | 2.4.2                                                 |
| on posix         | False                                                 |
| on linux         | False                                                 |
| on darwin        | False                                                 |
| on windows       | True                                                  |
| on cygwin        | <xonsh.lazyasd.LazyBool object at 0x000001DF0EC23D60> |
| on msys2         | <xonsh.lazyasd.LazyBool object at 0x000001DF0EC23D00> |
| is superuser     | False                                                 |
[...]

PR #3704, opened 2020-08-19
Restore correct behavior to DEFAULT_VARS that should not have defaults

tc21:
This pull request addresses #3703, and probably some other related issues related to #3377. It restores pre-#3377 behavior to `environ.Env.__getitem__`, `__setitem__`, `__iter__`, and `__contains__` by removing the default values from the env vars `$XONSH_SOURCE`, `$OLDPWD`, `$VIRTUAL_ENV`, `$TERM`, `$XONSH_INTERACTIVE`, `$XONSH_GITSTATUS_*`, and `$ANSICON`.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1721,7 +1721,9 @@ class Env(cabc.MutableMapping):
         self._no_value = object()
         self._orig_env = None
         self._vars = {k: v for k, v in DEFAULT_VARS.items()}
-        self._defaults = {k: v.default for k, v in self._vars.items() if v.default is not None}
+        self._defaults = {
+            k: v.default for k, v in self._vars.items() if v.default is not None
+        }
 
         if len(args) == 0 and len(kwargs) == 0:
             args = (os_environ,)
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -682,7 +682,7 @@ def DEFAULT_VARS():
             is_string,
             ensure_string,
             ensure_string,
-            "",
+            None,
             "This is used on Windows to set the title, if available.",
             doc_configurable=False,
         ),
@@ -1086,7 +1086,7 @@ def DEFAULT_VARS():
             is_string,
             ensure_string,
             ensure_string,
-            ".",
+            None,
             "Used to represent a previous present working directory.",
             doc_configurable=False,
         ),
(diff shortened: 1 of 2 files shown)


xonsh/environ.py
```diff
@@ -682,7 +682,7 @@ def DEFAULT_VARS():
             is_string,
             ensure_string,
             ensure_string,
-            "",
+            None,
```
scopatz, line 685: I think these need to be strings

xonsh/environ.py
```diff
@@ -1848,13 +1851,6 @@ def get_detyper(self, key, default=None):
 
         return detyper
 
-    def get_default(self, key, default=None):
```
scopatz, line 1851: Please keep this API

xonsh/environ.py
```diff
@@ -1721,6 +1721,9 @@ def __init__(self, *args, **kwargs):
         self._no_value = object()
         self._orig_env = None
         self._vars = {k: v for k, v in DEFAULT_VARS.items()}
+        self._defaults = {
+            k: v.default for k, v in self._vars.items() if v.default is not None
+        }
```
scopatz, line 1726: Unfortunately, None is a valid default.  So I think you would nee to use a different sentinal, like `xonsh.tools.DefaultNotGiven`. Also, we have moved everything into a single `_vars` (finally) so we would like to keep it there

laloch:
Maybe something similar to 8e54fb0 would do the trick?

tc21:
re: 5b2eb08, the `is not DefaultNotGiven` check would need to happen in `__delitem__` as well, at which point it might be better to make `item in self._vars and self.get_default(item) is not DefaultNotGiven` into `self.has_default(item)`.

bobhy:
I think #3748 covers these issues. 
It takes a different approach: allows defaults to be registered only for environment variables that Xonsh itself defines.
Any envvars that other programs define which Xonsh registers (for documentation) have registered default set to DefaultNotGiven and that means that envvar could raise KeyError if not actually defined.

That applies to `$OLDPWD`, `$VIRTUAL_ENV`, `$TERM`, and `$ANSICON` (and fixes #3703).

But it doesn't cover `$XONSH_SOURCE`, `$XONSH_INTERACTIVE` or `$XONSH_GITSTATUS_*`.  These are xonsh-defined.  Now that they're registered with specific defaults, code that depends on `$XONSH_SOURCE in env` or `env.get($XONSH_SOURCE, "random_default")` is incorrect and will have to be revisited.

Perhaps this PR can be closed?

anki-code:
Hello @tc21! Could you please check the bobhy message above?

PR #3705, opened 2020-08-21
THREAD_SUBPROCS=None for xonshrc files

scopatz:
This is related to the issue at https://github.com/regro/rever/issues/217.  I independently verified that even xonsh itself was having trouble starting up if any threaded subproc was used in rc files.  Super weird and hard to diagnose. The issue seems to have first arrived in the environment refactor (a8d4a57f01a25b1ce77f0a3adec1c0b11f49ad1a), but seems to be a side effect of that commit rather than directly caused by it.

### DIFF
--- a/news/nothreadrc.rst
+++ b/news/nothreadrc.rst
@@ -1,0 +1,28 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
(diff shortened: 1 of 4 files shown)


scopatz:
@xonsh/xore - this is ready for review!

Issue #3702, opened 2020-08-18
Xonsh 0.9.19 has a considerably higher startup time on Mac OS


samueldg:
I use Xonsh as my default shell, and yesterday after updating via `brew` (new release: https://github.com/Homebrew/homebrew-core/pull/59385), I noticed it took more time than usual before a new terminal tab with the Xonsh shell became active.

I tested on another computer where I had not upgraded yet from `0.9.18`, and running Xonsh commands takes less than 0.5 second.

After the upgrade:

```sh
$ xonsh --version
xonsh/0.9.19
$ time xonsh -c '1+1'
2
        3.31 real         3.17 user         0.18 sys

# It could be related to my specific config,
# but I'm still seeing this behaviour with `--no-rc`:
$ time xonsh --no-rc -c '1+1'
2
        3.33 real         3.17 user         0.19 sys
```

## xonfig

<details>

```
+------------------+-----------------+
| xonsh            | 0.9.19          |
| Python           | 3.8.5           |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | 3.0.4           |
| shell type       | prompt_toolkit  |
| pygments         | 2.6.1           |
| on posix         | True            |
| on linux         | False           |
| on darwin        | True            |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
+------------------+-----------------+
```

</details>

In case that helps, here's my xonshrc: https://github.com/samueldg/dotfiles/blob/master/xonshrc

## Expected Behavior

I expect Xonsh to be somewhat faster than this, in the <1 second range I'm used to.

## Current Behavior

Any command (except `xonsh --help`) takes over 3 seconds.

## Steps to Reproduce

1. Find someone with xonsh 0.9.18 installed via `brew`
2. Ask them to run `time xonsh -c '1+1'` and note the results
3. Get them to `brew upgrade`
4. Repeat step 2
5. Compare the results

Note that I tried to reproduce in Docker for a totally self-contained environment, and I'm getting different results:

```
$ docker run --rm -it python:3.8-slim bash
root@4b9d2d2c6bd8:/# pip install xonsh
[...]
Successfully installed xonsh-0.9.19
root@4b9d2d2c6bd8:/# time xonsh -c '1+1'
2

real	0m0.214s
user	0m0.164s
[...]

PR #3710, opened 2020-08-23
vox: Fix virtual environment removal

laloch:
This is a follow-up to #3702.
No news item needed.

Virtual environment was impossible to delete after activation/deactivation cycle.


scopatz:
Thanks!

PR #3712, opened 2020-08-24
Avoid attributeError crash in xonfig wizard

bobhy:
Fix for #3711

Looks like changes in #3704 didn't ripple outward to xonfig.py. It was using old field names.

Also https://github.com/xonsh/xonsh/blob/528d2f913ae40c89420fad02c665f3916aae44a0/xonsh/environ.py#L1867-L1877 would return unconverted `DefaultNotGiven` in `.doc_default` if called with environment variable not defined in `env._vars`.

Issue #3711, opened 2020-08-24
AttributeError: 'Env' object has no attribute '_docs' when running `xonfig wizard`


bobhy:
Running master as of today, attempting to document `xonfig wizard` for Getting Started docs.

## Repro:
``` bash
$ xonfig wizard
```
### Expect
It works.

### Traceback 

<details>

```
Traceback (most recent call last):
  File "c:\users\bobhy\src\xonsh\xonsh\proc.py", line 1439, in run
    r = self.f(self.args, sp_stdin, sp_stdout, sp_stderr, spec, spec.stack)
  File "c:\users\bobhy\src\xonsh\xonsh\proc.py", line 1232, in proxy_two
    return f(args, stdin)
  File "c:\users\bobhy\src\xonsh\xonsh\aliases.py", line 703, in xonfig
    return xonfig_main(args)
  File "c:\users\bobhy\src\xonsh\xonsh\xonfig.py", line 689, in xonfig_main
    return _XONFIG_MAIN_ACTIONS[ns.action](ns)
  File "c:\users\bobhy\src\xonsh\xonsh\xonfig.py", line 450, in _wizard
    w = make_xonfig_wizard(
  File "c:\users\bobhy\src\xonsh\xonsh\xonfig.py", line 410, in make_xonfig_wizard
    wiz.YesNo(question=WIZARD_ENV_QUESTION, yes=make_env_wiz(), no=wiz.Pass()),
  File "c:\users\bobhy\src\xonsh\xonsh\xonfig.py", line 343, in make_env_wiz
    w = _make_flat_wiz(make_envvar, sorted(builtins.__xonsh__.env._docs.keys()))
AttributeError: 'Env' object has no attribute '_docs'
```

</details>

## xonfig

<details>

```
$ xonfig
+------------------+-------------------------------------------------------+
| xonsh            | 0.9.19                                                |
| Git SHA          | 528d2f91                                              |
| Commit Date      | Aug 24 04:25:32 2020                                  |
| Python           | 3.8.5                                                 |
| PLY              | 3.11                                                  |
| have readline    | False                                                 |
| prompt toolkit   | 3.0.6                                                 |
| shell type       | prompt_toolkit                                        |
| pygments         | 2.6.1                                                 |
| on posix         | False                                                 |
| on linux         | False                                                 |
| on darwin        | False                                                 |
| on windows       | True                                                  |
[...]

PR #3715, opened 2020-08-25
Avoid attributeError crash in xonfig wizard, V2.

bobhy:
*Real* fix for #3711.
Replaces PR #3712, which somehow got random other changes mixed into it.

--------------------------------------------

Looks like changes in #3704 didn't ripple outward to xonfig.py. It was using old field names.

Also https://github.com/xonsh/xonsh/blob/528d2f913ae40c89420fad02c665f3916aae44a0/xonsh/environ.py#L1867-L1877 would return unconverted `DefaultNotGiven` in `.doc_default` if called with environment variable not defined in `env._vars`.


xonsh/environ.py
```diff
@@ -1865,12 +1865,10 @@ def get_default(self, key, default=None):
...
 
     def get_docs(self, key, default=None):
-        """Gets the documentation for the environment variable."""
+        """Gets the documentation for the environment variable. May be None"""
         vd = self._vars.get(key, None)
         if vd is None:
-            if default is None:
-                default = Var()
-            return default
+            return None
```
scopatz, line 1871: Shouldn't this return `default`?

scopatz:
Thanks for putting this in! I think there is one small issue

xonsh/environ.py
bobhy: `get_docs()` needs to ensure returned *obj*.doc_default is `str`, need to fix up default.
But the latest commit is a cleaner way to handle this, and lets me to remove the extra code where invoked in xonsh/xonfig.py (and where it should have been in the 2 other invocations I missed on the first go-round).

scopatz:
Thanks @bobhy!

PR #3719, opened 2020-08-26
Add support for Python 3.8's starred expressions in return statement

laloch:
Working towards Python 3.8 full compatibility.
This enables `return`s and `yield`s with starred expressions:
```python
def foo():
    mylist = [2, 3]
    return 1, *mylist
```

### DIFF
--- a/news/return-star-expr.rst
+++ b/news/return-star-expr.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* Support for starred expressions within return statement
+  (``return x, *my_list``).
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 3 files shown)


laloch:
Unfortunately, Black v20.8b0 insists on reformatting docstrings :(

scopatz:
Yeah let's ignore all of the black errors until it can get installed properly everywhere.

This looks great @laloch! I really love all of the work that you are doing on the parser! It is great to see.

laloch:
> Yeah let's ignore all of the black errors until it can get installed properly everywhere.

The problem is, that we most probably do not want the docstrings to get reformatted. At least not in the parser files.

> This looks great @laloch! I really love all of the work that you are doing on the parser! It is great to see.

Thanks @scopatz. I think the last missing feature for full Python 3.8 syntax (not AST) compatibility are positional-only arguments. I'm afraid that they're, going to be pretty messy.

scopatz:
> The problem is, that we most probably do not want the docstrings to get reformatted. At least not in the parser files.

Right, we'll have to find a way to turn that off.

PR #3720, opened 2020-08-26
Add support for PEP 570 positional-only parameters

laloch:
Adds Python 3.8 PEP 570 positional-only parameters:
```python
def foo(x, /, y):
    pass

min = lambda x, y, /: y if y < x else x
min(1, x=1)
TypeError: <lambda>() got some positional-only arguments passed as keyword arguments: 'x'
```


laloch:
This is now ready for review.

scopatz:
Awesome thanks!

PR #3723, opened 2020-08-27
Raise error if non-default argument follows default argument

### LINKED ISSUE #3721: Xonsh allows for non-default arguments to follow default argument
## Expected Behavior
```
$ def foo(x=1, y):
      print(x, y)
  File "<stdin>", line 1
SyntaxError: non-default argument follows default argument
```
## Current Behavior
```
$ def foo(x=1, y):
      print(x, y)
$ foo(2)
2 1
```

## xonfig

<details>

```
| xonsh            | 0.9.20               |
| Git SHA          | 17e56563             |
| Commit Date      | Aug 26 21:17:07 2020 |
| prompt toolkit   | 3.0.4                |
```

</details>

## For community

laloch:
Prevents parser from accepting argument lists where non-default parameter follows default parameter (`def foo(x=1, y):`)

Closes #3721

### DIFF
--- a/news/non-default-follows-default.rst
+++ b/news/non-default-follows-default.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Non-default parameters can not follow defaults anymore.
+
+**Security:**
+
+* <news item>
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -3138,10 +3138,36 @@ def test_syntax_error_bar_kwonlyargs():
     with pytest.raises(SyntaxError):
         PARSER.parse("def spam(*):\n   pass\n", mode="exec")
 
+
+@skip_if_pre_3_8
 def test_syntax_error_bar_posonlyargs():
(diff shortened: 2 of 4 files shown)


laloch:
The tests are failing, because our parser is unable to raise an error from production rules. Until the error handler is fixed this at least prevents the parser from generating invalid code.

scopatz:
hmm that kinda sucks.  Thanks for working on this though @laloch

laloch:
See #3724.

laloch:
This is now good to go.

scopatz:
Yep! Looks great!

laloch:
Thanks @scopatz. I would like the following to produce correct error message, but I don't really know how to deduce the intended mode it was meant to run in. Those are the charms of context-sensitive parsing...
```
$ lambda x=1, y: x
xonsh: subprocess mode: command not found: lambda
Did you mean the following?
    samba:  Command (/usr/bin/samba)
```

scopatz:
Ugg yeah, that should parse as Python

scopatz:
Or wait! It shouldn't because `y` comes after `x=1`. Yeah that is tricky

laloch:
Yep, it should emit
```
$ lambda x=1, y: x
  File "<stdin>", line 1
SyntaxError: non-default argument follows default argument
```
Maybe I'll manage to come with some heuristics to pick the correct message...

PR #3728, opened 2020-08-28
Propagate parser errors from production rules

### LINKED ISSUE #3724: Parser doesn't emit syntax errors raised from production rules
Our parser currently doesn't report `SyntaxError` exceptions raised inside of parser production rules. This is due to the fact that Yacc catches such an exception and attempts recovery by injecting `error` token in place of the current token and continues parsing the input as if the `error` token was a part of the input text. Since we don't handle `error` token at all, the parser ultimately unwinds the stack to a state where none of the rules can be matched anymore and is supposed to call `p_error`, that is a catchall handler for unhandled  errors. But there's the catch - in case of production-rule-risen syntax error the parser does NOT call the catchall handler. See the docs at https://ply.readthedocs.io/en/latest/ply.html#signalling-an-error-from-a-production. Most importantly the documentation says:
> One important aspect of manually setting an error is that the p_error() function will NOT be called in this case. If you need to issue an error message, make sure you do it in the production that raises SyntaxError.

We need to re-implement our error handling logic to be able to report such errors. There is, however, another catch - since Xonsh fist attempts to parse the input in Python mode and if that fails it resorts to parsing in shell mode, commands like `echo foo` are in fact expected to raise an error in the first (Python mode) pass and we definitely do not want to first report a syntax error and then successfully execute the `echo` command.

## xonfig

<details>

```
| xonsh            | 0.9.20               |
| Git SHA          | 4be525ef             |
| Commit Date      | Aug 27 13:25:38 2020 |
| prompt toolkit   | 3.0.4                |
```

</details>

## Expected Behavior
```
$ f"{}"
SyntaxError: f-string: empty expression not allowed
```

## Current Behavior
```
$ f"{}"
$
```

## For community

laloch:
Fixes #3724

### DIFF
--- a/news/parser-emit-error.rst
+++ b/news/parser-emit-error.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed parser not emmiting errors in some cases.
+
+**Security:**
+
+* <news item>
--- a/xonsh/parsers/base.py
+++ b/xonsh/parsers/base.py
@@ -461,6 +461,7 @@ class BaseParser(object):
 
         # Keeps track of the last token given to yacc (the lookahead token)
         self._last_yielded_token = None
+        self._error = None
 
     def reset(self):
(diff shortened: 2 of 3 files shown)


scopatz:
This is a really nice change. Thanks a ton!

PR #3733, opened 2020-08-29
Unpin sphinx from 2.4.4

dannysepler:
According to #3506 , 3.0.0 broke the docs, but 3.0.1 fixed the docs again. Pinning to 3.1, just to be safe. I clicked around locally and things seem to work.


scopatz:
Thanks @dannysepler! No news item needed here

requirements/docs.txt
```diff
@@ -1,6 +1,6 @@
 cloud_sptheme
 numpydoc
-Sphinx==2.4.4
+Sphinx>=3.1
```
scopatz, line 3: should the be `>=3.1,<3`?

AstraLuma: pretty sure it's an AND?

dannysepler: right! it would fail on that i think?

scopatz: Ahh OK, this is fine then

PR #3737, opened 2020-08-30
Add --pytest argument to xonsh-in-docker script

laloch:
Automates pytest installation into the docker container.

No news entry needed.


scopatz:
This seems like it actually does need a news entry

laloch:
OK, let's have a news entry then :wink:

scopatz:
Thanks @laloch!

PR #3740, opened 2020-08-31
Fix video playback on xon.sh landing page

laloch:
This fixes browsers complaining about
> Refused to display 'https://www.youtube.com/watch?v=uaje5I22kgE&feature=youtu.be' in a frame because it set 'X-Frame-Options' to 'sameorigin'."

We need to expand the shortened youtu.be link ourselves in order to be able to play it within an iframe.

No news item needed.


laloch:
Tested to work in Chrome 85 and Firefox 80.

scopatz:
Thanks @laloch!

Issue #3739, opened 2020-08-31
Xonsh environment returns wrong value for undefined variables and also missing some OS environment variables.


bobhy:
Related to #3703 and PR #3377, but with more feeling.

## Steps to Reproduce / Current Behavior

```
# Case 1: not defined in OS environ, but Xonsh has a known default
$ os.environ['ANSICON']
KeyError: 'ANSICON'
$ ls $ANSICON
ls: cannot access '': No such file or directory
# As expected (but only because ANSICON has defined default of ``).

# Case 2: not defined in OS environ or Xonsh
$ ls $NO_SUCH_VARIABLE
ls: cannot access '$NO_SUCH_VARIABLE': No such file or directory
# Expected empty string.

# Case 3: defined in Xonsh environ (and synced to OS environ)
$ ls $SHELL_TYPE
ls: cannot access 'prompt_toolkit': No such file or directory
# As expected

# Case 4: defined in OS environment, but not in Xonsh environment
$ os.environ["WINDIR"]
'C:\\WINDOWS'
$ ls $WINDIR
ls: cannot access '$WINDIR': No such file or directory
# expected 'C:\WINDOWS'.

# Case 5: not defined in OS environment but not a Xonsh configuration parameter
$ "ANSICON" in os.environ
False
$ "ANSICON" in __xonsh__.env
True
# Expected 2nd to be False if first is False (issue #3703)
```

## Expected Behavior

3 bugs here:
1. OS environment variables that don't happen to match a key in `xonsh.environ.DEFAULT_VARS` are not present in `__xonsh__.env` and `$<env_var>`.  (Case 4 but also Case 2)
2. Reference to an undefined (xonsh) environment variable should evaluate to empty string at least in subproc mode. (Case 2 but also Case 4 and Case 1).
3. Xonsh environment returns a value to python code when the OS environment value is not defined. (case 5, issue of #3703).  This might be acceptable for variables that xonsh defines and consumes, but is clearly a lie for environment variables that other programs are supposed to define for xonsh.

We can address case 4 by simply ensuring *all* the OS environment variables are loaded into Xonsh at startup.

We might be able to fix the rest of the cases (and resolve #3703)  if we implement an is-defined function so code that cares about it can tell whether it was really defined.  And we'd return the defined default for references to undefined environment variables that had them and hard code empty string for other undefined variables..  For example:

``` 
$ __xonsh__.env["ANSICON"]
''
$ "ANSICON" in __xonsh__.env
False

[...]

PR #3742, opened 2020-09-01
Environment fixes -for unregistered, undefined env vars, return empty string, yet don't consider them "in" environ either.

bobhy:
Proposed fix for #3739 and #3703 
Draft status, pending one pesky problem..

### DIFF
--- a/tests/test_environ.py
+++ b/tests/test_environ.py
@@ -2,10 +2,9 @@
 """Tests the xonsh environment."""
 from __future__ import unicode_literals, print_function
 import os
-import itertools
 import re
 from tempfile import TemporaryDirectory
-from xonsh.tools import always_true
+from xonsh.tools import always_true, DefaultNotGiven
 
 import pytest
 
@@ -13,7 +12,6 @@ from xonsh.commands_cache import CommandsCache
 from xonsh.environ import (
     Env,
     locate_binary,
-    DEFAULT_VARS,
     default_env,
     make_args_env,
     LsColors,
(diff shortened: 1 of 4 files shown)


laloch:
This would make us behave like Bash featuring implicit variable declarations, which, in my opinion, is plain wrong. One of the Xonsh's goals is to get rid of Bash's mistakes and misconceptions. This is one of them. I only know of two other programming languages with implicit variable declarations - Perl and PHP, that both are iconic examples of horrible design.

bobhy:
Fair comment! However, Xonsh is already doing this for the environment variables that have a registered default.  It returns the default (often an empty string) if the variable is not defined in the actual environment.  Entirely correct that I'm extending that to return empty string for undefined and unregistered variables.

What really bugs me, and motivated this PR, it considers the variable "in" the environment as well.    You might as well have initialized the environment to the defaults!

Unlike Bash, this PR offers a "natural" way to tell whether the variable is actually in the environment regardless of the returned value.  It won't yield the name if you iterate over keys and returns False for `in __xonsh__.env`.  I think this is a reasonable control for that, but if we don't come to consensus on it, I think we should reconsider the original sin.

Maybe it would be better to not return the default for `__xonsh__.env[]` or `__xonsh__.env.get()`?  The registered defaults would still be available via `__xonsh__.env.get_defaults()`, but code would have to know this non-dict method to access them.

bobhy:
I'm retracting this approach and will do the fix a simpler way.

PR #3745, opened 2020-09-04
Add optional variable initializers preceding command

laloch:
This adds support for bash-like environment variable assignments preceding subprocess command.
```
$ $FOO="foo" $BAR=2 xonsh -c r"echo $FOO$BAR"
foo2
$ # Also works in pipelines...
$ $FOO="foo" $BAR=2 xonsh -c r"echo -n $FOO$BAR" | $FOO="nofoo" xonsh -c r"cat; echo $FOO"
foo2nofoo
```


anki-code:
Great feature! Please add the example to [Bash to Xonsh Translation Guide](https://xon.sh/bash_to_xsh.html).

scopatz:
Wow! This is incredible! Thanks so much @laloch!

PR #3747, opened 2020-09-07
xontrib-argcomplete

### LINKED ISSUE #1409: Add argcomplete support?
Does anyone have experience with the argcomplete package by @kislyuk?  https://github.com/kislyuk/argcomplete

It adds argparse completion to bash for packages that support it. It seems that conda does for example.   It seems to be good candidate to add to the set of completers in xonsh. It is targeted bash but since it it is a Python package it must be possible to use it directly.

anki-code:
Added [xontrib-argcomplete](https://github.com/anki-code/xontrib-argcomplete) to support [kislyuk/argcomplete](https://github.com/kislyuk/argcomplete) - tab completion for argparse.

Closes  #1409

### DIFF
--- a/news/argcomplete.rst
+++ b/news/argcomplete.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Added xontrib-argcomplete to support kislyuk/argcomplete - tab completion for argparse.
+
+**Changed:**
+
+* <news item>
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks!

PR #3748, opened 2020-09-08
Environment fixes

bobhy:
Proposed fix for #3739 and #3703
 
Allows you to specify default of `DefaultNotGiven when registering an environment variable, which means raise KeyError rather than return the default if that variable not actually defined in environment.  Used that for "ANSICON", which fixes the immediate problem, and for "TERM" and "VIRTUAL_ENV", reverting some recent changes to match).  The point is that these environment variables (and maybe OLDPWD?) are conventionally expected to be defined outside of Xonsh, so whether they exist is significant.  

No change in this patch to evaluation of non-registered environment variables when they aren't defined in environment, that's a topic for another day.


news/environment_fixes.rst
```diff
@@ -0,0 +1,30 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* xonsh/environ.py: new rule: if default for "registered" environment variable is set to ``DefaultNotGiven``_
```
scopatz, line 7: I think that the trailing underscore here should be removed in the in-line code

tests/test_environ.py
```diff
@@ -357,7 +355,7 @@ def test_register_custom_var_generic():
     assert env["MY_SPECIAL_VAR"] == 32
 
     env["MY_SPECIAL_VAR"] = True
-    assert env["MY_SPECIAL_VAR"] == True
+    assert env["MY_SPECIAL_VAR"] == True  # NOQA E712
```
scopatz, line 358: probably should either be `assert env["MY_SPECIAL_VAR"] is True` or `assert env["MY_SPECIAL_VAR"]`

tests/tools.py
```diff
@@ -109,7 +109,7 @@ class DummyEnv(MutableMapping):
     DEFAULTS = {
         "XONSH_DEBUG": 1,
         "XONSH_COLOR_STYLE": "default",
-        "VIRTUAL_ENV": "",
+        # "VIRTUAL_ENV": "",
```
scopatz, line 112: Please remove dead code

scopatz:
Thanks @bobhy - just a couple of small updates

news/environment_fixes.rst
bobhy: ok

tests/test_environ.py
bobhy: ok

tests/tools.py
bobhy: ok

bobhy: ok

scopatz:
It looks like there are conflicts here now, since I merged the other branch

bobhy:
This is ready for review, and I'd like to get it resolved, one way or the other.

scopatz:
Thanks!

Issue #3751, opened 2020-09-11
PROMPT: {cwd_dir} should append slash if it isn't empty and doesn't end with a slash


planet36:
## xonfig

<details>

```
$ xonfig info
+------------------+-----------------+
| xonsh            | 0.9.21          |
| Python           | 3.8.5           |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | 3.0.7           |
| shell type       | prompt_toolkit  |
| pygments         | 2.6.1           |
| on posix         | True            |
| on linux         | True            |
| distro           | arch            |
| on darwin        | False           |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
+------------------+-----------------+
```
</details>

My goal is to use both `{cwd_dir}` and `{cwd_base}` in my prompt, but make `{cwd_base}` a more intense color.

## Expected Behavior

I expect that concatenating `{cwd_dir}` and `{cwd_base}` would be equal to `{cwd}`.

| cwd      | cwd_dir | cwd_base |
|----------|---------|----------|
| ~        |         | ~        |
| ~/code   | ~/      | code     |
| /        | /       |          |
| /usr     | /       | usr      |
| /usr/bin | /usr/   | bin      |

## Current Behavior

Here is what the prompt elements show for various paths.

| cwd      | cwd_dir | cwd_base |
|----------|---------|----------|
| ~        |         | ~        |
| ~/code   | ~       | code     |
| /        | /       |          |
| /usr     | /       | usr      |
| /usr/bin | /usr    | bin      |

scopatz:
Ahh yeah, this is fair.  PRs welcome! Thanks for reporting @planet36!

jnoortheen:
Closing this as the PR is merged already

PR #3754, opened 2020-09-11
[PROMPT]: Concatenate slash to cwd_dir if it doesn't end with a slash.

Eadaen:
Implements change to $PROMPT_FIELDS["cwd_dir"] functionality as proposed in #3751. Now cwd_dir consistently ends with a slash, to ease usability.

Previous behaviour(cwd -> cwd_dir): "/foo/bar" -> "/foo"; "~/foo" -> "~".
New behaviour: "/foo/bar" -> "/foo/"; "~/foo" -> "~/".
Behaviour if cwd is one of "~", "/" or "/foo" is unchanged

The code is cross-platform, though has only been tested on Unix-like paths.

Let me know if I should add tests for the function added, and where they should be added if so. tests/test_prompt.py?


xonsh/prompt/cwd.py
```diff
@@ -89,3 +89,12 @@ def _dynamically_collapsed_pwd():
                 0 : int(target_width) - len(truncature_char)
             ] + truncature_char
     return full
+
+
+def _add_dir_slash(cwd_dir):
```
laloch, line 94: I think this is usually implemented using `os.path.join(cwd_dir, "")`. No extra function needed.

laloch (changes requested):
Hi @Eadaen, thanks for the PR. Could you please check my comment above?

xonsh/prompt/cwd.py
Eadaen: Thanks for the suggestion. That makes things simpler.

laloch (approved):

laloch:
Thanks @Eadaen!

PR #3758, opened 2020-09-11
feat: use threads to offload prompt sections

### LINKED ISSUE #2803: Async prompt?
## xonfig

```
| xonsh            | 0.7.7           |
```

## Question

I've looked around xonsh sourcecode, prompt_toolkit source, and the docs. But I am unable to find a way to update the prompt/screen while a command is running.

Example, if a command I run suddenly uses over 80% cpu, I would like a tiny message in the button right corner to give me the cpu percent. Is this possible?

I tried a couple of different things with prompt_toolkit, like using threads, but I could not update any text in the terminal until the process was done executing.

Based on https://github.com/xonsh/xonsh/blob/b69fb9ab4bf0def1b702db74e016bf43a8478606/xonsh/codecache.py#L68, I don't think this is possible? There is no async, it blocks.. Or am I wrong?

jnoortheen:
resolves #2803


anki-code:
It looks like cool feature! Could you please give an example the simple prompt with async sections loading? 

May be added it to the [xonsh tutorial](https://xon.sh/tutorial.html#customizing-the-prompt)?

jnoortheen:
this is after setting `$ENABLE_ASYNC_PROMPT = True`

render1599900379555

anki-code:
I've got an error:
```
git clone -b feat-threaded-prompt https://github.com/jnoortheen/xonsh
cd xonsh && pip install -U .
xonsh --no-rc
$XONSH_SHOW_TRACEBACK = True
$ENABLE_ASYNC_PROMPT=True
  File "/home/pc/.local/lib/python3.8/site-packages/xonsh/ptk_shell/shell.py", line 316, in _get_prompt_tokens
    self.prompt_updators[prompt_name] = self.prompt_formatter(
TypeError: __call__() got an unexpected keyword argument 'threaded'
  File "/home/pc/.local/lib/python3.8/site-packages/xonsh/ptk_shell/shell.py", line 316, in _get_prompt_tokens
    self.prompt_updators[prompt_name] = self.prompt_formatter(
TypeError: __call__() got an unexpected keyword argument 'threaded'
```

jnoortheen:
Could you check with a temporary rc file? I think some variables are not loaded once the prompt is started.

anki-code:
How I can check temporary rc files?

jnoortheen:
> How I can check temporary rc files?

you can use rc file option to start xonsh with a rc file. I dont remember the exact name. `xonsh --help` would do that.

anki-code:
Also I could you please explain how to make the prompt asynchronous?

For example I have:
```
pc@pc:~ 14:04:34> $PROMPT='{user}@{hostname}:{cwd} {localtime}> '
pc@pc:~ 14:04:35> 
```
Is there a way to update time without pressing Enter? Or could you please give an example of asynchronous custom field for the prompt in my example above?

jnoortheen:
> Also could you please explain how to make the prompt asynchronous?
> 
> For example I have:
> 
> ```shell
> pc@pc:~ 14:04:34> $PROMPT='{user}@{hostname}:{cwd} {localtime}> '
> pc@pc:~ 14:04:35> $ENABLE_ASYNC_PROMPT = True
> pc@pc:~ 14:04:35> # time is static
> ```
> 
> Is there a way to update time without pressing Enter? Or could you please give an example of asynchronous custom field for the prompt in my example above?

the environment variable in enough. 

```py

# save this to a file named rc.xsh
import time
$ENABLE_ASYNC_PROMPT = True

def _in_future():
    time.sleep(5)
    return "slept"

$PROMPT_FIELDS['in_future'] = _in_future
$PROMPT = '{user}@{hostname}:{cwd} {localtime} {in_future}> '

# run this with
# xonsh --rc rc.xsh
```

anki-code:
Many thanks for the example! It works!

Two questions about the example:
1. If I press Enter 10 times the function will be executed 10 times or previous 9 calls will be canceled?
2. I press Enter every second and after some time I'll get the result earlier than 5 second. Why?

jnoortheen:
1. Yes it will be cancelling all futures that were started and still running. this happens when new prompt is created. 
2. need more testing I think. I can't think of any reason. maybe changing the example to return the current time and comparing what is getting returned might explain it. 
3. in_future is a prompt field. This is how it works

 i. it parses the prompt tokens
 ii. when a token is a callback it is offloaded to a future.
 iii. to know its place where the result from future need to be updated, it returns the prompt_field name in present
 iv. in future the callback replaces the prompt_field_name with the value and calls `invalidate` this will update the prompt.

for info, please check the `formatter.py` for its working. I tried to make minimal number of changes and keep what is already there.

anki-code:
@jnoortheen While I'm viewing your gif above I see how slow `git status` working and I remembered faster implementation of gitstatus - https://github.com/romkatv/gitstatus. May be it will be interesting to you and you can add the support of it in xonsh.

jnoortheen:
> @jnoortheen While I'm viewing your gif above I see how slow `git status` working and I remembered faster implementation of gitstatus - https://github.com/romkatv/gitstatus. May be it will be interesting to you and you can add the support of it in xonsh.

I already checked it. There was an elvish plugin and one can copy the logic from it. But it will need more effort. Most of the time, this is enough.

anki-code:
@jnoortheen to avoid `TypeError: __call__() got an unexpected keyword argument 'threaded'` error may be to add try-except with message "$ENABLE_ASYNC_PROMPT=True can be done only inside your xonshrc file."

jnoortheen:
> @jnoortheen to avoid `TypeError: __call__() got an unexpected keyword argument 'threaded'` error may be to add try-except with message "$ENABLE_ASYNC_PROMPT=True can be done only inside your xonshrc file."

that is already taken care of. now the default formatter is threadedFormatter without affecting sync functionality

xonsh/ptk_shell/formatter.py
```diff
@@ -0,0 +1,45 @@
...
+from xonsh.prompt.base import PromptFormatter, DEFAULT_PROMPT
+
+
+class ThreadedPromptFormatter(PromptFormatter):
+    def __call__(self, template=DEFAULT_PROMPT, fields=None, threaded=False):
+        """Formats a xonsh prompt template string."""
+        self.template = template  # current template
+        self.threaded = threaded
+        self.futures = {}
+        if threaded and (not hasattr(self, "thread_pool")):
```
BYK, line 14: Wouldn't it be cheaper to set this on `__init__` rather than checking with `hasattr` for each call?

xonsh/ptk_shell/formatter.py
```diff
@@ -0,0 +1,45 @@
...
+            self.thread_pool = ThreadPoolExecutor(
+                max_workers=builtins.__xonsh__.env["ASYNC_PROMPT_THREAD_WORKERS"]
+            )
+
+        prompt = super().__call__(template, fields)
+
+        if threaded:
```
BYK, line 21: So the class name says `Threaded` but it also supports non-threaded? Shouldn't that be out of scope for this class?

xonsh/ptk_shell/shell.py
```diff
@@ -195,7 +203,21 @@ def singleline(
                 except (AttributeError, TypeError, ValueError):
                     print_exception()
 
+        if env["ENABLE_ASYNC_PROMPT"]:
+            # once the prompt is done, update it in background as each future is completed
+            updators = [
```
BYK, line 208: How about `updaters`?

```suggestion
            updaters = [
```

xonsh/ptk_shell/shell.py
```diff
@@ -195,7 +203,21 @@ def singleline(
...
+                threading.Thread(target=self.update_prompt, args=(pt_name, futures))
```
BYK, line 209: Do we really need more threads for handling each future? Aren't they already async? Also I don't see a locking mechanism down in `self.update_prompt` so this code seems to be prone to race conditions.

xonsh/ptk_shell/shell.py
```diff
@@ -240,50 +262,90 @@ def cmdloop(self, intro=None):
...
+            return kwargs.pop("default")
+
         try:
-            p = self.prompt_formatter(p)
+            template = p
+            # free-up resources
+            # prev_updator = self.prompt_updators.get(prompt_name)
+            # if prev_updator:
+            #     for fut in prev_updator:
+            #         fut.cancel()
```
BYK, line 314: Clean up unused code? :)

BYK (commented):
I love this feature but I have some concerns around these threads. Maybe I don't get the code fully but that would indicate a need for more comments :)

xonsh/ptk_shell/formatter.py
jnoortheen: the idea is to support both threaded and synchronous calls. So if we put this on init, it might be occupying unnecessary memory for sync prompts. I will change the class name of the formatter to be like PTKFormatter or something.

xonsh/ptk_shell/formatter.py
jnoortheen: yes the same case as said above.

xonsh/ptk_shell/shell.py
jnoortheen: I did try that. But we needed a thread to get result from those futures and update the prompt. There is no way to know that the prompt processing is done for those futures, (i.e. they can start calling prompt.invalidate)

xonsh/ptk_shell/shell.py
jnoortheen: Sure will remove

xonsh/ptk_shell/shell.py
jnoortheen: also I dont see any need for locking conditions. when the next set of futures come in and previous futures still running, it will get cancelled.

xonsh/ptk_shell/formatter.py
```diff
@@ -0,0 +1,62 @@
...
+            future = self.thread_pool.submit(self._run_func, func, field)
+
+            res = "{%s}" % field
+            if idx is None:
+                self.futures = res
+            else:
+                self.futures[future] = idx
+
+            # todo: the old result should be shown with faded colour.
+            return self.thread_results[field] if field in self.thread_results else res
```
jnoortheen, line 54: @BYK do you know how to style the token faded ? even when it  is surrounded by other colors ?

xonsh/ptk_shell/shell.py
BYK: My concern is this creating a new thread for each updater without any locking. Wouldn't that cause a race condition if multiple updaters try to update the prompt around the same time?

xonsh/ptk_shell/formatter.py
```diff
@@ -5,12 +5,22 @@
 from xonsh.prompt.base import PromptFormatter, DEFAULT_PROMPT
 
 
-class ThreadedPromptFormatter(PromptFormatter):
+class PTKPromptFormatter(PromptFormatter):
```
BYK, line 8: This is quite a bit clearer but I still think we may benefit from splitting these into two distinct classes: `AsyncPTKPromptFormatter` and `SyncPTKPromptFormatter` and remove a bunch of branching in the methods.

I guess the current way enables changing `$ENABLE_ASYNC_PROMPT` on the fly without a restart but not sure how common that would be, especially given the increased complexity here and the overhad of a bunch of if checks below per `prompt_formatter` call.

xonsh/ptk_shell/shell.py
jnoortheen: calling invalidate is thread safe. The commented code, was there to do that, but I commented as it will not be needed, since the future gets cleared at every call, the threads listening on them will die off. Here we dont need locking, only to clear off the old updaters to free up resources.

xonsh/ptk_shell/formatter.py
jnoortheen: for experimenting some try to set the variables within the prompt. Like @anki-code did

jnoortheen: I understand your point. I was concious to make the if checks to terminate on self.threaded (first check) and there are two places where it checks. 
regarding the additional methods, it is unavoidable, maybe we can move the methods out of classes.

laloch:
I don't like too much the automatic cancellation of the running futures. One can easily get into the situation that a future will never finish, only uselessly wasting resources.

jnoortheen:
> 't like too much the automatic cancellation of the running futures. One can easily get into the situation that a future will never finish, only uselessly wasting resou

there are two sets of threads. 
one inside Formatter class, another group inside shell class. 

Formatter: 
 - the pool is initiated with certain number of workers. and cancelled on new __call__ 
(thought it was already there. now added in the commit)

Shell Updator:
 - there will be maximum three threads at most for each prompt/rpromt/bottom_toolbar
 - it will get stopped once the futures cancelled.

laloch:
@jnoortheen, never mind. I didn't notice, that the timer gets only re-armed once the respective future has finished.

xonsh/ptk_shell/formatter.py
```diff
@@ -0,0 +1,68 @@
+import builtins
```
scopatz, line 1: Please add docstrings to this module as well as documentation hooks in the `docs/` dir

xonsh/ptk_shell/formatter.py
```diff
@@ -0,0 +1,68 @@
...
+            return tokens
+        return "".join(tokens)
+
+    def _no_cache_field_value(self, field, field_value, idx: int = None):
+        func = functools.partial(super()._no_cache_field_value, field, field_value, idx)
+
+        if self.threaded and callable(field_value):
+            future = self.thread_pool.submit(self._run_func, func, field)
+
+            res = "{%s}" % field
```
scopatz, line 53: please use f-strings

xonsh/ptk_shell/shell.py
```diff
@@ -240,50 +262,99 @@ def cmdloop(self, intro=None):
...
                     break
 
-    def prompt_tokens(self):
-        """Returns a list of (token, str) tuples for the current prompt."""
-        p = builtins.__xonsh__.env.get("PROMPT")
+    def update_prompt(self, prompt_name: str, toks_futures: Tuple[List[str], dict]):
+        """Listen on futures and update the prompt as each one is completed.
+
+        :param prompt_name : what prompt to update. One of ['message', 'rprompt', 'bottom_toolbar']
+        :param toks_futures : a tuple containing tokens and futures for that prompt.
```
scopatz, line 269: Please follow the numpydoc style, thanks

xonsh/ptk_shell/shell.py
```diff
@@ -240,50 +262,99 @@ def cmdloop(self, intro=None):
...
         try:
-            p = self.prompt_formatter(p)
+            template = p
+            if env["ENABLE_ASYNC_PROMPT"]:
+                self.prompt_updators[prompt_name] = self.prompt_formatter(
+                    template=template, threaded=True
+                )
+                p = "".join(self.prompt_updators[prompt_name][0])
+            else:
+                p = self.prompt_formatter(template=template)
```
scopatz, line 330: please try to limit the try-block to a single line.  Feel free to make a helper function if needed

xonsh/ptk_shell/formatter.py
jnoortheen: the alternative f-string will be `f"{{{field}}}"`

since it needs to interpolate. having three curly braces would be difficult to understand what is happening. I will rewrite with string concat. OK?

xonsh/ptk_shell/formatter.py
jnoortheen: done.

xonsh/ptk_shell/shell.py
jnoortheen: done.

docs/api/ptk_shell/formatter.rst
```diff
@@ -0,0 +1,10 @@
+.. _xonsh_ptk2_shell:
+
+******************************************************
+Prompt Toolkit 2+ Shell (``xonsh.ptk_shell.shell``)
```
scopatz, line 4: The title here should include that it is for the formatter

xonsh/ptk_shell/formatter.py
scopatz: String concat is fine too

xonsh/ptk_shell/formatter.py
scopatz: It would have to be the inner-most color code, which might be hard.

jnoortheen: perhaps, can we wrap these fields into a class? like this

```
Section:
  content = "text"
  style = "color"
  faded_style = "color"
```

later we can select the styles based on the place. What do you think?

xonsh/ptk_shell/formatter.py
jnoortheen: done

docs/api/ptk_shell/formatter.rst
jnoortheen: done.

scopatz:
@BYK or @laloch - do one of you who have looked at this more closely want to be in charge of final review & merge here?

xonsh/ptk_shell/formatter.py
jnoortheen: I have removed any color and made it to be GRAY color.

xonsh/ptk_shell/formatter.py
jnoortheen: any updates ?

scopatz:
pinging this @xonsh/xore

jnoortheen:
does this need any update ? please let me know.

xonsh/ptk_shell/formatter.py
jnoortheen: I have split the code into a separate module. Now it will be more clear I think.

scopatz:
Thanks @jnoortheen - I'll merge since it didn't get traction from other @xonsh/xore devs

PR #3760, opened 2020-09-12
Docs for keyboard shortcuts

dannysepler:
Figured it might be nice to follow up the shift-selection ticket with a doc page on all the various keyboard shortcuts available to xonsh.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -440,5 +440,5 @@ builtins.__xonsh__.commands_cache = CommandsCache()
 def setup(app):
     from xonsh.pyghooks import XonshConsoleLexer
 
-    app.add_lexer("xonshcon", XonshConsoleLexer())
+    app.add_lexer("xonshcon", XonshConsoleLexer)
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -440,5 +440,5 @@ def make_events():
 def setup(app):
     from xonsh.pyghooks import XonshConsoleLexer
 
-    app.add_lexer("xonshcon", XonshConsoleLexer())
+    app.add_lexer("xonshcon", XonshConsoleLexer)
```
dannysepler, line 443: sphinx said this is supposed to pass a class rather than an instance

scopatz: Yeah, for some reason the classes don't work

scopatz: You might need to leave this line as it was originally

scopatz:
Thanks for putting this in @dannysepler!

scopatz:
Thanks again!

Issue #3756, opened 2020-09-11
use via cygwin


wendellwt:
xonsh does not operate on windows via cygwin.  It installs via pip, and on startup the prompt shows up.  However, typical Unix commands either get permission denied or hang the terminal, as in:
```
$ ls
xonsh: subprocess mode: permission denied: ls
$ /usr/bin/ls
a_file  notes.txt
<terminal hangs>
```
Q: if one were to try to diagnose this in the source code, where would one start to look?  What environment variables/settings are useful for debugging xonsh itself?

wendellwt:
Well, it turns out that there were two issues:
1) in `get_script_subproc_command()` in `built_ins.py`, the cygwin version of `os.access()` doesn't check the PATH.  This is my workaround:
```
    # make sure file is executable
    if not os.access(fname, os.X_OK):
        # explicitly look at all PATH entries for cmd
        wt_path = os.getenv("PATH").split(':')
        wt_fpath = list(map(lambda p: p + os.sep + fname,wt_path))
        if not any(list(map(lambda p: os.access(p,os.X_OK), wt_fpath))):
            raise PermissionError

```
2) in `iterraw()` in `proc.py`, at line 2036, the `proc.wait()` never returned; removing it allowed the command to complete:
```
        self.stream_stderr(safe_readlines(stderr))
        # wt: proc.wait()   # <<<<<<<<<<<<<<< HELP
        self._endtime()

```
I have no idea if removing the proc.wait() causes other problems, but it allows the cygwin version to work.

scopatz:
> Q: if one were to try to diagnose this in the source code, where would one start to look? What environment variables/settings are useful for debugging xonsh itself?

You can set `XONSH_DEBUG` to get more info: https://xon.sh/envvars.html#xonsh-debug

scopatz:
You probably do need to wait for the process to end.  But maybe we could pass on that if in cygwin.  @wendellwt - would you be willing to put in a couple of PRs for these?

PR #3762, opened 2020-09-13
cygwin updates

wendellwt:
cygwin needs to check full path for permissions;
cygwin doesn't need to do `proc.wait()` inside `iterraw()`

fix to issue #3756

### DIFF
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -2033,8 +2033,7 @@ class CommandPipeline:
         # read from process now that it is over
         yield from safe_readlines(stdout)
         self.stream_stderr(safe_readlines(stderr))
-        if not ON_CYGWIN:
-            proc.wait()
+        proc.wait()
         self._endtime()
         yield from safe_readlines(stdout)
         self.stream_stderr(safe_readlines(stderr))
--- a/xonsh/proc.py
+++ b/xonsh/proc.py
@@ -2033,7 +2033,8 @@ class CommandPipeline:
         # read from process now that it is over
         yield from safe_readlines(stdout)
         self.stream_stderr(safe_readlines(stderr))
-        proc.wait()
+        if not ON_CYGWIN:
+            proc.wait()
         self._endtime()
         yield from safe_readlines(stdout)
         self.stream_stderr(safe_readlines(stderr))
--- a/news/cygwin-updates.rst
+++ b/news/cygwin-updates.rst
@@ -16,8 +16,7 @@
 
 **Fixed:**
 
-* cygwin needs full path to find exe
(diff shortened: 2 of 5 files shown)


xonsh/proc.py
```diff
@@ -2033,7 +2033,8 @@ def iterraw(self):
         # read from process now that it is over
         yield from safe_readlines(stdout)
         self.stream_stderr(safe_readlines(stderr))
-        proc.wait()
+        if not ON_CYGWIN:
+            proc.wait()
```
scopatz, line 2037: This doesn't seem correct. This seems like it would have processes return before they are finished running.

scopatz:
Thanks @wendellwt! This could use a news entry. 

Also, @xonsh/xore - didn't we used to have cygwin CI?

xonsh/proc.py
wendellwt: Well, I wasn't sure about it, but that was where simple shell commands were hanging.  The command (e.g. `ls`) would output as expected, then pause forever in that `proc.wait()`.  I'm not saying it it correct, or that I understand it, but cygwin liked it that way.  Is there some other flag/semaphore that would be a better choice?

scopatz: Can you try these commands with [`$THREAD_SUBPROCS=False`](https://xon.sh/envvars.html#thread-subprocs), and see if wait() still hangs?

wendellwt: That works fine!  With that set in the rc file the `proc.wait()` can be left as it was.  Thank you.

scopatz: Yeah, we may want to change the default for cygwin to False as a better fix than this

wendellwt: Yes, that seems like a more appropriate fix; I just don't see where the default is set.

scopatz: It is in environ.py

wendellwt: Ok, corrected in environ.py.  Btw, it passes all checks but that pesky "lint/xonsh" one, which doesn't indicate why the test fails.  How can I correct that?

laloch: Hi @wendellwt, that pesky test says literally "No news item" and, as @scopatz already pointed out, there's indeed no news entry. Could you please add one? See [Development Guide](https://xon.sh/devguide.html#changelog) for details.

scopatz: Yeah you need to add a news item, see https://xon.sh/devguide.html#changelog

wendellwt: Oh, _that_ news item.  Done.  Sorry about that.

scopatz:
Thanks @wendellwt!

Issue #3698, opened 2020-08-17
`NotImplementedError` trying to start xonsh kernel in Jupyter notebook 


bobhy:
Brand new virtual environemt containing latest-and-greatest of everything.  I'm trying this for the first time in order to add it to getting-started docs.

## xonfig

<details>

```
+------------------+----------------------+
| xonsh            | 0.9.19.dev87         |
| Git SHA          | 1c9fffed             |
| Commit Date      | Aug 16 16:31:17 2020 |
| Python           | 3.8.5                |
| PLY              | 3.11                 |
| have readline    | False                |
| prompt toolkit   | 3.0.6                |
| shell type       | prompt_toolkit       |
| pygments         | 2.6.1                |
| on posix         | False                |
| on linux         | False                |
| on darwin        | False                |
| on windows       | True                 |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+
```

</details>

## pip list

<details>

```
Package           Version
----------------- -------
argon2-cffi       20.1.0
attrs             19.3.0
backcall          0.2.0
bleach            3.1.5
cffi              1.14.2
colorama          0.4.3
decorator         4.4.2
defusedxml        0.6.0
entrypoints       0.3
ipykernel         5.3.4
ipython           7.17.0
ipython-genutils  0.2.0
jedi              0.17.2
Jinja2            2.11.2
jsonschema        3.2.0
jupyter-client    6.1.6
jupyter-core      4.6.3
MarkupSafe        1.1.1
mistune           0.8.4
nbconvert         5.6.1
nbformat          5.0.7
notebook          6.1.3
packaging         20.4
pandocfilters     1.4.2
parso             0.7.1
pickleshare       0.7.5
pip               20.1.1
prometheus-client 0.8.0
prompt-toolkit    3.0.6
ptk               1.3.7
pycparser         2.20
Pygments          2.6.1
pyparsing         2.4.7
pyrsistent        0.16.0
python-dateutil   2.8.1
pywin32           228
pywinpty          0.5.7
pyzmq             19.0.2
Send2Trash        1.5.0
setuptools        47.1.0
six               1.15.0
terminado         0.8.3
testpath          0.4.4
tornado           6.0.4
traitlets         4.3.3
wcwidth           0.2.5
webencodings      0.5.1
wheel             0.35.1
[...]

Issue #3452, opened 2020-02-24
How to use Juypter Kernel?


EdwardJB:
Hi,

I've searched the repo and found lots of references to a Xonsh Jupyter Kernel, but can't find anything on actually how to run it.

I just want a browser-based notebook using Xonsh. I've managed to launch a base Jupyter interface from my Xonsh shell but obviously it only lets me create traditional Python notebooks.

What am I missing?

laloch:
Hi @EdwardJB,
the Jupyter kernel should get installed automatically by the Xonsh setup script. It seems to work for me:
```$ xontrib load vox
$ vox create jupyter
$ vox enter jupyter
(jupyter) $ pip install jupyter
[...]
(jupyter) $ pip install "xonsh[linux]"
[...]
(jupyter) $ jupyter kernelspec list
Available kernels:
  xonsh      /home/laloch/.local/share/jupyter/kernels/xonsh
  python3    /home/laloch/.virtualenvs/jupyter/share/jupyter/kernels/python3
```
I can now run `jupyter notebook` and select "xonsh" from the list of available kernels.

EdwardJB:
Thanks for the reply @laloch . I've run those command (they worked) but I still only get the Python kernel.

```ed@eb /mnt/c/Users/ed $ jupyter kernelspec list
Available kernels:
  python3    /mnt/c/Users/ed/.local/share/jupyter/kernels/python3
```

Btw, I'm not using `vox` if that matters; I have `xonsh` installed under my system Python3.

laloch:
The platform seems to be WSL, right?
How did you install Jupyter?
How did you install Xonsh?
When issued from your system Python3 REPL, what does the following command do?
```
>>> from jupyter_client.kernelspec import KernelSpecManager
```

EdwardJB:
> The platform seems to be WSL, right?
> How did you install Jupyter?
> How did you install Xonsh?
> ...

I got busy with work, so sorry talking so long to respond.

Yes, it's WSL, but I traced the problem back to the `Dockerfile` I used to create the WSL environment originally. Here it is (with the steps you recommended added):

```
FROM python:3.8

RUN DEBIAN_FRONTEND=noninteractive apt-get update  -y

### Install shell
RUN python -m pip install jupyter
RUN python -m pip install "xonsh[linux]"
RUN jupyter kernelspec list
```

```
Deploying '<unknown> Dockerfile: shell/Dockerfile'...
Building image...
Preparing build context archive...
[...]

PR #3765, opened 2020-09-13
Move jup kernel spec from setup to xonfig; add doc.

bobhy:
This PR moves definition of the Xonsh kernel spec from an automatic step in `setup.py` to a manually invoked new subcommand, `xonfig kernel`.  Adds info to `xonfig info` to display current status of kernel.  Adds documentation for the new procedure, addresses #3452.  

Now it's easy to install the xonsh kernel, but #3698 is still a blocker (maybe just on Windows?),

`pip` appears to modify `sys.path`, preventing `setup.py` from detecting jupyter in the current virtual environment when installing via `pip install ...`; install from a wheel doesn't even invoke `setup.py`. The only reliable install that defines the kernel spec is `python setup.py install`.


xonsh/xonfig.py
```diff
@@ -672,6 +754,7 @@ def _xonfig_create_parser():
     "styles": _styles,
     "colors": _colors,
     "tutorial": _tutorial,
+    "kernel": _kernel,
```
scopatz, line 757: Maybe this subcommand etc should be called `jupyter` to distinguish it from other kernels

bobhy: I was trying to distinguish this option from the other xonfig options and trying to keep it short.  How about option "jupyter_kernel"? I just noticed xonfig has completion, it's not too much to type.  (Though I hate introducing the mental confusion between snake case and kebab case...)

scopatz: I would be ok with `jupyter-kernel`. The important bit is that jupyter appears in the name

bobhy: ok

xonsh/xonfig.py
```diff
@@ -662,31 +723,48 @@ def _xonfig_create_parser():
         "style", nargs="?", default=None, help="style to preview, default: <current>"
     )
     subp.add_parser("tutorial", help="Launch tutorial in browser.")
+    kern = subp.add_parser("jupyter_kernel", help="Generate xonsh kernel for jupyter.")
```
scopatz, line 726: Please call this `jupyter-kernel`. We generally use dashes for CLIs

scopatz:
Can you add `xonfig jupyter-kernel` to the test suite too please? Thanks!

bobhy:
@scopatz How do I monkeypatch a module when the package (`jupyter_client`) is not present in the environment?  I've tried various games with `sys.modules` and `monkeypatch`, but would appreciate a hint. I could drop a test stub of the package somewhere under tests/ and monkeypatch `sys.path`, but maybe there is a more self contained way?

scopatz:
Hmm you probably shouldn't be monkey-patching.  What are you needing to do exactly?

bobhy:
https://github.com/bobhy/xonsh/blob/e24fe29d50313c84329923cf75a2a0c3934c0df8/xonsh/xonfig.py#L640-L648

I figured I'd create a dummy package with a skeleton  class `KernelSpecManager` in some out-of-the-way subfolder and monkeypatch `sys.path` to find it (pytest says there's specific support for path that deals with caching). I don't see any other tests doing this, so thought I'd ask first.  Once the package is in path, I can monkeypatch methods on the class for the test cases as usual.

xonsh/xonfig.py
bobhy: ... Now he tells me...

scopatz:
Oh, I was thinking we should just run this command as part of the setup. After the `pip install`

bobhy:
Ready for review now.  The win/py3.6 failure is in the conda environment, not the tests.

xonsh/xonfig.py
```diff
@@ -15,6 +15,8 @@
 import contextlib
 import collections
 
+from tempfile import TemporaryDirectory
```
scopatz, line 18: This should just be `import tempfile`, like the imports above, so that we get the amalgamate benefit

xonsh/xonfig.py
```diff
@@ -523,6 +527,16 @@ def _info(ns):
             ("encoding errors", env.get("XONSH_ENCODING_ERRORS")),
         ]
     )
+    jup_ksm = jup_kernel = None
+    try:
+        from jupyter_client.kernelspec import KernelSpecManager
+
+        jup_ksm = KernelSpecManager()
+        jup_kernel = jup_ksm.find_kernel_specs()[XONSH_JUPYTER_KERNEL]
+    except (ImportError, KeyError):
```
scopatz, line 536: Can you separate this out into a couple of try-except blocks, please, with preferably only one line each? Also, you shouldn't need to use a try-KeyError.  Just use `.get()` or similar

xonsh/xonfig.py
```diff
@@ -623,6 +637,53 @@ def _web(args):
     subprocess.run([sys.executable, "-m", "xonsh.webconfig"] + args.orig_args[1:])
 
 
+def _kernel(args):
```
scopatz, line 640: Can you call this `_jupyter_kernel` please?

xonsh/xonfig.py
```diff
@@ -623,6 +637,53 @@ def _web(args):
...
+    """Make xonsh available as a Jupyter kernel."""
+    try:
+        from jupyter_client.kernelspec import KernelSpecManager
+    except ImportError:
+        raise ImportError("Jupyter not found in current Python environment")
```
scopatz, line 645: Probably should do a raise-from here: https://www.python.org/dev/peps/pep-3134/#id37

xonsh/xonfig.py
bobhy: Ummm, is there a plan to support a kernel for something other than jupyter in the future?  What if jupyter decides to rebrand their project again? why should xonsh be so excruciatingly explicit about the jupyter lernel?

xonsh/xonfig.py
bobhy: using `.get()` means there's only one anticipated exception, so I've simplified.  But if something other than `ImportError` is raised (e.g by `ksm = jupyter_client.kernelspec.KernelSpecManager()`) I'll let `info` simply display ``None``.  The user will get a traceback as soon has s/he tries to actually run the kernel, doesn't need a messy traceback obscuring the info output.

xonsh/xonfig.py
scopatz: Well, so (un)fortunately, because of the way that the amalgamate code works, it squishes all of the source files together to make a single large source file. This means that almost everything shares a single namespace in the xonsh code base. So it is less about other xonfig things being called kernel, and more that we might want to denote other functionality in xonsh as being related to some other kind of kernel (the python executable kernel, the OS kernel, etc).  Amalgamate sort of forces us to be very precise in our naming of everything so that namespace collisions don't happen.

scopatz: I know it is annoying, but it is good to be robust :rofl:

bobhy: OK, I changed the name to `_jupyter_kernel`.  

But what about `_web`, `_wizard` and the other global functions in `xonfig`?  If global uniqueness is really necessary, they could all be renamed `_xonfig_web`, `_xonfig_wizard`, etc, and then the new one should be called `_xonfig_kernel`.  I'm not going there, but I will agree `amalgamate.py` is annoying to live with.

And, see #3790; an additional annoyance is that it doesn't stop the show when it detects name conflicts.

scopatz:
This looks good & ready @bobhy. Is it OK to merge from your end?

bobhy:
Yep, I'm good.

scopatz:
Thanks for the work here!

PR #3768, opened 2020-09-14
New event on_run_subproc

### LINKED ISSUE #3767: Is there an event to change command or execute another command?
Hi! xonsh is amazing! Thanks!

I've read about `on_precommand` and `on_transform_command` events and wondering is there a way to change command or execute another command?

###  For community

anki-code:
Hello! xonsh is great and I want to make it a bit greater!

I've added `on_run_subproc` event to try a concept in [xontrib-onepath](https://github.com/anki-code/xontrib-onepath) where the path without previous leading commands do an action.

Example:
Screenshot_20200915_001941

Closes #3767


anki-code:
Closes #3767

xonsh/built_ins.py
```diff
@@ -949,6 +949,10 @@ def run_subproc(cmds, captured=False, envs=None):
     if builtins.__xonsh__.env.get("XONSH_TRACE_SUBPROC"):
         print("TRACE SUBPROC: %s" % str(cmds), file=sys.stderr)
 
+    precmds = events.on_subproc_precommand.fire(cmds=cmds)
+    if precmds:
+        cmds = precmds[0]
```
scopatz, line 954: Shouldn't this be `cmds =  precmds`?

xonsh/built_ins.py
```diff
@@ -949,6 +949,10 @@ def run_subproc(cmds, captured=False, envs=None):
     if builtins.__xonsh__.env.get("XONSH_TRACE_SUBPROC"):
         print("TRACE SUBPROC: %s" % str(cmds), file=sys.stderr)
 
+    precmds = events.on_subproc_precommand.fire(cmds=cmds)
```
scopatz, line 952: Maybe this should be called `change_subproc_cmds`?

news/subproc_precommand.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* Added event ``on_subproc_precommand`` that used in xontrib-onepath.
```
scopatz, line 3: This should be more desciptive

scopatz:
Thanks for putting this in @anki-code! I think a few things need to be added here:

1. Some docs (see how the other events are registered)
2. A test, probably in `test_integrations.py`

xonsh/built_ins.py
scopatz: Or maybe "on_run_subproc()"

xonsh/built_ins.py
anki-code: `precmds` will be `[(['ls',])]` after firing. So we need `[0]`.

xonsh/built_ins.py
anki-code: Please tell me the name you want and I change :)

news/subproc_precommand.rst
anki-code: Updated

xonsh/built_ins.py
scopatz: Why is it returning cmds wrapped in list?

xonsh/built_ins.py
scopatz: `on_run_subproc()` is best I think

xonsh/built_ins.py
anki-code: Because
https://github.com/anki-code/xonsh/blob/patch-2/xonsh/events.py#L178-L186

scopatz: Ahh right, then why is this only taking the first event's response? It seems like it should figure out what to do if multiple handlers are registered to this event

anki-code: Good question and I have no answer :) I suggest to use the first event and someone who will face with issues change this logic along his use case.

anki-code:
> Some docs (see how the other events are registered)

Added.

> A test, probably in test_integrations.py

I'm trying but have no success. Need help.

Example:
```python
# -*- coding: utf-8 -*-
from __future__ import unicode_literals, print_function
import os
import itertools
import re
from tempfile import TemporaryDirectory
from xonsh.tools import always_true
import pytest

def test_event_on_run_subproc(xonsh_builtins):
    @xonsh_builtins.events.on_run_subproc
    def handler(cmds, **kwargs):
        return (['echo','ok'],)
    output = xonsh_builtins.__xonsh__.subproc_captured_object(['echo', '123'])
    assert output == 'ok'
```
But:
```pytest
==================================================== FAILURES ====================================================
___________________________________________ test_event_on_run_subproc ____________________________________________

xonsh_builtins = <module 'builtins' (built-in)>

    def test_event_on_run_subproc(xonsh_builtins):
        @xonsh_builtins.events.on_run_subproc
        def handler(cmds, **kwargs):
            return (['echo','ok'],)
        output = xonsh_builtins.__xonsh__.subproc_captured_object(['echo', 'nok'])
>       assert output == 'ok'
E       AssertionError: assert 'no-ok\n' == 'ok'
E         - ok
E         + nok

============================================ short test summary info =============================================
FAILED test_qqq.py::test_event_on_run_subproc - AssertionError: assert 'nok\n' == 'ok'
=============================================== 1 failed in 0.17s ================================================
```

anki-code:
@xonsh/xore please advice how to run subproc function with test event? (code example above)

anki-code:
> @xonsh/xore please advice 
Hmmm, looks like github ignores "@xonsh/xore" mention because I'm not the part of the team :(

xonsh/built_ins.py
scopatz: Hmm we should probably figure this out

scopatz:
@anki-code - have a look at what [`test_integrations.py::test_script`](https://github.com/xonsh/xonsh/blob/master/tests/test_integrations.py#L462) does. Basically, it runs a full script in a new xonsh subprocess. This way you can cleanly register a new event with out affecting the xonsh instance that the tests are running with. You probably just need to add another case or two here.

anki-code:
Why I have this errors?

<details>

Screenshot_20200922_212935


</details>

scopatz:
try running from the root level of the repo, ie `pytest tests/test_integrations.py`

anki-code:
I'm going to use `on_transform_command` without any additional changes. Thanks!

PR #3774, opened 2020-09-17
TAB key automatically selects first completion when starting completion.

bobhy:
An unsolicited "enhancement" to ptk completion.

TAB character now automatically selects first completion when it starts completion, rather than simply displaying the completion options.  This more closely mimics behavior of `TAB menu-complete` in readline.  I'd argue this is more in the nature of a bug fix than enabling new behavior, so  I didn't implement a configuration option, simply changed the behavior.

### DIFF
--- a/news/ptk-complete-select-first.rst
+++ b/news/ptk-complete-select-first.rst
@@ -1,0 +1,25 @@
+**Added:**
+
+* <news item>
+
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks @bobhy!

PR #3777, opened 2020-09-18
Use commands_cache when finding suggestions

dannysepler:
When you put a not-recognized command, it takes a second to respond the error. The vast majority of this time seems to be finding command suggestions. If we use the commands_cache instead, it speeds up ~6x when there's no suggestions.

BEFORE:
```
$ timeit "asdfasdfasdf"
...
xonsh: subprocess mode: command not found: asdfasdfasdf
1 loops, best of 3: 544 ms per loop
```

AFTER:
```
$ timeit "asdfasdfasdf"
...
xonsh: subprocess mode: command not found: asdfasdfasdf
10 loops, best of 3: 90.4 ms per loop
```

On the other hand, there's basically no change in performance when it does find a suggestion.

### DIFF
--- a/news/cache-executables-in.rst
+++ b/news/cache-executables-in.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Added cache on finding executable files, to speedup command-not-found suggestions
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/news/cache-executables-in.rst
+++ b/news/cache-executables-in.rst
@@ -1,6 +1,6 @@
 **Added:**
 
-* Added cache on finding executable files, to speedup command-not-found suggestions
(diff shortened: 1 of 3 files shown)


scopatz:
Hey @dannysepler - thanks for putting this in. I definitely don't think the failure should take 2 sec to print the message.

However, I don't think that this is a safe fix. Folks can modify their environment between two commands (say by installing a pakage) and you would expect that after the install the new cmd would be available.  

I think a better fix would be for the alternatives message to use the lazy APIs on the CommandsCache.  The CommandsCache itself uses the `executable_in()` function.  Sorry to make this more difficult.

dannysepler:
not at all! these test failures were making me second guess myself anyway. thanks for giving me that direction!

scopatz:
This looks much better! What does the timing look like with this now?

dannysepler:
thanks, @scopatz! i've been updating the timing as i update the code. I'm not sure what was going on with the 1.77 seconds in my first version of this diff. my laptop now says it's closer to 500 ms -> 90 ms. It's still a significant speedup, but not quite the insanity I saw before :)

scopatz:
Thanks so much for this!

PR #3778, opened 2020-09-19
Is 3775 - ptk shell not using full height completion menu at bottom of screen

### LINKED ISSUE #3775: Support multiline prompt in ptk completer
Hello! xonsh is the best shell! Thanks!

By default I have only 2 lines when [xontrib-prompt-bar](https://github.com/anki-code/xontrib-prompt-bar) at the end of console window:

Screenshot_20200917_230746

Here is no support for multiline prompt:
https://github.com/xonsh/xonsh/blob/40352638bf6b01d308538b4a32e995bb68913a6c/xonsh/ptk_shell/completer.py#L118-L123

As result:
```
Count of lines in completer = ($COMPLETIONS_MENU_ROWS - LINES_IN_PROMPT - 1) = (5 - 2 - 1) = 2 
```

Workaround: increase `$COMPLETIONS_MENU_ROWS`.

Thanks!

## For community

bobhy:
Fixes #3775.
Not proud of this fix, it relies on empirical constants of dubious justification.  But it works even if you change the default `$COMPLETIONS_MENU_ROWS`.  However, it *always* leaves that number of rows below the prompt, even if there are only a few completions -- the prior code did better in this case.

It would be nice if PTK provided a documented interface for sizing the completions window, so we could do better.

### DIFF
--- a/news/is_3775.rst
+++ b/news/is_3775.rst
@@ -1,0 +1,25 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Issue #3775.  Prompt toolkit shell now displays the full number of completion rows (COMPLETIONS_MENU_ROWS)
+  even if current prompt is at the bottom of the screen.
+  Note that it will only display this many rows if it has enough actual completions to need so many.
+
+**Security:**
+
+* <news item>
--- a/xonsh/ptk_shell/completer.py
+++ b/xonsh/ptk_shell/completer.py
@@ -112,12 +112,12 @@ class PromptToolkitCompleter(Completer):
         render = app.renderer
         window = app.layout.container.children[0].content.children[1].content
 
-        if window and window.render_info and render._last_screen:
-            height_needed = window.render_info.content_height + builtins.__xonsh__.env.get("COMPLETIONS_MENU_ROWS")
-
-            last_h = max(render._last_screen.height, render._min_available_height)
-            if last_h < height_needed:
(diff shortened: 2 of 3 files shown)


anki-code:
@bobhy many thanks for this!

Could you please test this with [xontrib-prompt-bar](https://github.com/anki-code/xontrib-prompt-bar)?

It works perfect in `single` mode (5 lines):
Screenshot_20200919_095438

But I have 4 lines instead of 5 lines in `multi`:
Screenshot_20200919_095524

Maybe it's because prompt bar has 2 lines.

bobhy:
Latest version does handle different height prompts (at least 1 through 4). It works fine when cursor is already at bottom of screen or nowhere near the bottom, but it does not work when cursor is just a few lines *above* the bottom.  

In that case, I'm seeing that the rendering information lags by one screen and I cannot find the window that PTK is actually using from the context of the completer.  There must be a way to definitively tell PTK to scroll the window up the right number of lines, but I haven't found it.  I wish PTK exposed a documented interface to let the app configure the completion menu control.

And I'm done for now.  Please review this change and merge it if you think it's a worthwhile improvement.

anki-code:
@bobhy got it! I see what you described but it is not so annoying. It works for me in current state. Thanks!

scopatz:
This seems like a reasonable change to me.  Any thoughts from anyone else on @xonsh/xore?

bobhy:
Retracting this PR in favor of #3874.  Although that one requires PTK fixes, it's the only approach that will work reliably.  I may be able to refactor the PTK changes into xonsh prompt-toolkit shell, so we don't have to depend on PTK taking the fixes.

scopatz:
I think PTK had a release in the past couple of days

PR #3780, opened 2020-09-19
Minor performance improvements to the prompt

dannysepler:
There is a fraction of a delay on the prompt when you press enter. I imagine this is quite a tough case to crack, so this is just a start!

1. Showing the branch name seems to re-roll the `git rev-parse --abbrev-ref HEAD` function. This probably isn't great if there's lots of local branches for a user. I swapped it.
2. Locate binaries even more lazily. To be honest, I can't really track that much of a performance increase from this one, so down to revert it.
3. More efficient ways of finding whether a dirty working directory. Borrowed this from [this github conversation](https://github.com/sindresorhus/pure/issues/115).

**Results**:

Hard to evaluate this since the terminal speed is different each time ya run it. But for me, when in the xonsh repo, it seems to switch from an average of `0.025 < x < 0.030 seconds`, to `0.021 < x < 0.027` seconds. But it really varies widely, it would be cool if I had a better way of verifying!

### DIFF
--- a/xonsh/prompt/vc.py
+++ b/xonsh/prompt/vc.py
@@ -24,26 +24,15 @@ RE_REMOVE_ANSI = LazyObject(
 def _get_git_branch(q):
     denv = builtins.__xonsh__.env.detype()
     try:
-        branches = xt.decode_bytes(
-            subprocess.check_output(
-                ["git", "branch"], env=denv, stderr=subprocess.DEVNULL
-            )
-        ).splitlines()
+        cmd = ["git", "rev-parse", "--abbrev-ref", "HEAD"]
+        branch = xt.decode_bytes(
+            subprocess.check_output(cmd, env=denv, stderr=subprocess.DEVNULL)
+        )
+        branch = branch.splitlines()[0] or None
     except (subprocess.CalledProcessError, OSError, FileNotFoundError):
         q.put(None)
     else:
-        for branch in branches:
-            if not branch.startswith("* "):
-                continue
-            elif branch.endswith(")"):
-                branch = branch.split()[-1][:-1]
-            else:
-                branch = branch.split()[-1]
-
-            q.put(branch)
-            break
-        else:
-            q.put(None)
+        q.put(branch)
 
 
 def get_git_branch():
@@ -59,7 +48,6 @@ def get_git_branch():
     t.join(timeout=timeout)
     try:
         branch = q.get_nowait()
-        # branch = RE_REMOVE_ANSI.sub("", branch or "")
         if branch:
             branch = RE_REMOVE_ANSI.sub("", branch)
     except queue.Empty:
@@ -157,16 +145,17 @@ def current_branch():
     """
     branch = None
     cmds = builtins.__xonsh__.commands_cache
-    # check for binary only once
-    if cmds.is_empty():
-        has_git = bool(cmds.locate_binary("git", ignore_alias=True))
-        has_hg = bool(cmds.locate_binary("hg", ignore_alias=True))
-    else:
-        has_git = bool(cmds.lazy_locate_binary("git", ignore_alias=True))
-        has_hg = bool(cmds.lazy_locate_binary("hg", ignore_alias=True))
-    if has_git:
+
+    # This allows us to locate binaries after git only if necessary
+    def has(vc):
+        if cmds.is_empty():
+            return bool(cmds.locate_binary(vc, ignore_alias=True))
+        else:
+            return bool(cmds.lazy_locate_binary(vc, ignore_alias=True))
+
+    if has("git"):
         branch = get_git_branch()
-    if not branch and has_hg:
+    if not branch and has("hg"):
         branch = get_hg_branch()
     if isinstance(branch, subprocess.TimeoutExpired):
         branch = "<branch-timeout>"
@@ -175,19 +164,27 @@ def current_branch():
 
 
 def _git_dirty_working_directory(q, include_untracked):
-    status = None
     denv = builtins.__xonsh__.env.detype()
     try:
-        cmd = ["git", "status", "--porcelain"]
+        # Borrowing this conversation
+        # https://github.com/sindresorhus/pure/issues/115
         if include_untracked:
-            cmd.append("--untracked-files=normal")
+            cmd = [
+                "git",
+                "status",
+                "--porcelain",
+                "--quiet",
+                "--untracked-files=normal",
+            ]
         else:
-            cmd.append("--untracked-files=no")
-        status = subprocess.check_output(cmd, stderr=subprocess.DEVNULL, env=denv)
+            cmd = ["git", "diff", "--no-ext-diff", "--quiet"]
+        child = subprocess.Popen(cmd, stderr=subprocess.DEVNULL, env=denv)
+        child.communicate()
+        dwd = bool(child.returncode)
     except (subprocess.CalledProcessError, OSError, FileNotFoundError):
         q.put(None)
-    if status is not None:
-        return q.put(bool(status))
+    else:
+        q.put(dwd)
 
 
 def git_dirty_working_directory(include_untracked=False):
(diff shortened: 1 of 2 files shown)


dannysepler:
I imagine the best solution re git would be for the terminal to have some knowledge of things like branch switches as they happen, rather than evaluating with each new line. Is this possible / worth pursuing?

xonsh/base_shell.py
```diff
@@ -503,12 +503,14 @@ def settitle(self):
...
-        # Shells running in emacs sets TERM to "dumb" or "eterm-color".
-        # Do not set title for these to avoid garbled prompt.
+        # Shells running in emacs sets TERM to "dumb", "eterm-color"
+        # or xterm-256color (for OS X). Do not set title for these
+        # to avoid garbled prompt.
         if (term is None and not ON_WINDOWS) or term in [
             "dumb",
             "eterm-color",
             "linux",
+            "xterm-256color",
```
dyuri, line 513: I don't think that the title shouldn't be set if `$TERM` is `xterm-256color`, it's very commonly used.

dannysepler: thanks for the heads up! out of curiosity, do you know what `Shells running in emacs sets TERM to ...` means above? Is that EMACS_MODE? maybe I can check for that instead?

(also a really naive question: what _is_ the title? why do we have to evaluate these seemingly unused fields with every prompt?)

dyuri: I'm not sure but that might be some kind of sub-shell in emacs. (I'm a vim user, the `:terminal` in my vim does not break when using title setting escape sequences.)

This *title* (set via `$TITLE`) is the terminal window title, which can be set by ANSI escape sequence - if the terminal emulator supports it. Most xterm compatible terminal emulators does (those that identify themselves as `xterm-256color`should I think), this escape sequence is usually part of the $PROMPT, but currently xonsh - due to `prompt_toolkit` - doesn't support setting it this way. I just submitted a PR to support it (and other ANSI OSC escape sequences): #3784

dyuri: Some other examples: https://github.com/xonsh/xonsh/pull/3781#issuecomment-695809527

dannysepler: thanks so much for explaining! i'm going to revert this part then

xonsh/prompt/vc.py
```diff
@@ -24,26 +24,15 @@
 def _get_git_branch(q):
     denv = builtins.__xonsh__.env.detype()
     try:
-        branches = xt.decode_bytes(
-            subprocess.check_output(
-                ["git", "branch"], env=denv, stderr=subprocess.DEVNULL
-            )
-        ).splitlines()
+        cmd = ["git", "rev-parse", "--abbrev-ref", "HEAD"]
```
scopatz, line 27: Nice change, thanks!

xonsh/prompt/vc.py
```diff
@@ -157,16 +145,17 @@ def current_branch():
...
-        has_git = bool(cmds.lazy_locate_binary("git", ignore_alias=True))
-        has_hg = bool(cmds.lazy_locate_binary("hg", ignore_alias=True))
-    if has_git:
+
+    # This allows us to locate binaries after git only if necessary
+    def has(vc):
+        if cmds.is_empty():
+            return bool(cmds.locate_binary(vc, ignore_alias=True))
+        else:
+            return bool(cmds.lazy_locate_binary(vc, ignore_alias=True))
```
scopatz, line 154: This should not be defined in the closure. Please move this to the top-level, and perhaps add a leading underscore to the name

xonsh/prompt/vc.py
```diff
@@ -175,19 +164,27 @@ def current_branch():
...
+                "--porcelain",
+                "--quiet",
+                "--untracked-files=normal",
+            ]
         else:
-            cmd.append("--untracked-files=no")
-        status = subprocess.check_output(cmd, stderr=subprocess.DEVNULL, env=denv)
+            cmd = ["git", "diff", "--no-ext-diff", "--quiet"]
+        child = subprocess.Popen(cmd, stderr=subprocess.DEVNULL, env=denv)
+        child.communicate()
```
scopatz, line 182: please just do a subprocess.run() or check_output(), rather than making a Popen object and using communicate

scopatz:
Thanks for these @dannysepler!

xonsh/prompt/vc.py
```diff
@@ -149,24 +137,25 @@ def _first_branch_timeout_message():
     )
 
 
+def _has(binary):
```
scopatz, line 140: Please name this something more distinct because of amalgamation. Perhaps, `_vc_has`?

xonsh/prompt/vc.py
```diff
@@ -175,19 +164,26 @@ def current_branch():
...
+                "--porcelain",
+                "--quiet",
+                "--untracked-files=normal",
+            ]
         else:
-            cmd.append("--untracked-files=no")
-        status = subprocess.check_output(cmd, stderr=subprocess.DEVNULL, env=denv)
+            cmd = ["git", "diff", "--no-ext-diff", "--quiet"]
+        child = subprocess.run(cmd, stderr=subprocess.DEVNULL, env=denv)
+        dwd = bool(child.returncode)
```
scopatz, line 182: Returncodes don't follow normal boolean rules.  0 is True and non-zero is False

dannysepler: HA derp thank you so much!

dannysepler: actually i'm sorry, i'm not convinced this is wrong.

https://git-scm.com/docs/git-diff

quiet means it implies --exit-code. Here are the results of --exit-code
- 1 means differences. Therefore, we want dwd to be True
- 0 means no differences. Therefore, we want dwd to be False

dannysepler: i chose --quiet because it's supposed to be a bit faster. but is this unintuitive to read? should i leave more comments or handle things differently?

scopatz: More comments would be helpful. In xonsh, returncodes are always interpreted as above, so this would need some clarification. Thanks!

scopatz:
Is this more or less ready @dannysepler?

dannysepler:
yep! thanks for your patience!

scopatz:
Great! Thanks for putting this in!

xonsh/prompt/vc.py
```diff
@@ -175,19 +164,31 @@ def current_branch():
...
-        cmd = ["git", "status", "--porcelain"]
+        # Borrowed from this conversation
+        # https://github.com/sindresorhus/pure/issues/115
         if include_untracked:
-            cmd.append("--untracked-files=normal")
+            cmd = [
+                "git",
+                "status",
+                "--porcelain",
+                "--quiet",
```
nh2, line 176: I don't think this is correct, nor is it it correct in current `master`:

`git status` does not have a `--quiet` option; it does not exist in the man page and errors with

```
error: unknown option `quiet'
```

I also don't think that just calling `bool()` below on the return code is a good idea; it would be better to check the exact exit codes that are expected. That would probably have revealed this problem.

xonsh/prompt/vc.py
```diff
@@ -175,19 +164,31 @@ def current_branch():
...
+                "--untracked-files=normal",
+            ]
         else:
-            cmd.append("--untracked-files=no")
-        status = subprocess.check_output(cmd, stderr=subprocess.DEVNULL, env=denv)
+            unindexed = ["git", "diff", "--no-ext-diff", "--quiet"]
+            indexed = unindexed + ["--cached", "HEAD"]
```
nh2, line 181: This gets broken (also on current `master`) if there exists a file called `HEAD` in the repo.

Solution: Append another arg `--`: https://stackoverflow.com/questions/2657935/checking-for-a-dirty-index-or-untracked-files-with-git/2659808#comment76137052_2659808

xonsh/prompt/vc.py
dannysepler: ahh nice catches! you're right, using "bool" hid the error. thank you! i'll try to diff in a fix soon, unless you'd prefer to take this on

PR #3781, opened 2020-09-19
ANSI terminal title setter escape sequence support (fixes #1403)

dyuri:
Fixes #1403 - removes the ANSI terminal title setter escape sequence from `$PROMPT` and sets `$TITLE` to it.

Issue #1403, opened 2016-07-08
Escape sequences don't work on $PROMPT variable (using prompt-toolkit)


santagada:
if I use echo to print a escape sequence that changes term title:

``` bash
$ echo '\033]0;sadf@saf: asdf\007'

```

it works as expected, changes the term title and doesn't print anything on the screen. On the other hand if I do that on the $PROMPT variable:

``` bash
$ $PROMPT = '\033]0;sadf@saf: asdf\007'
^[]0;sadf@saf: as
```

My prompt becomes corrupted and the title doesn't change. It really looks like \033 is being converted to the ^[ set of chars and not being passed as is to the terminal... but it can be something else

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

scopatz:
Thanks for bringing this up, @santagada. what is the output of xonfig? Importantly, are you using readline or prompt toolkit?

santagada:
```
$  xonfig
+------------------+----------------+
| xonsh            | 0.4.3          |
| Python           | 3.5.2          |
| PLY              | 3.7            |
| have readline    | True           |
| prompt toolkit   | 1.0.3          |
| shell type       | prompt_toolkit |
| pygments         | 2.1.3          |
| on posix         | True           |
| on linux         | False          |
| on darwin        | True           |
| on windows       | False          |
| on cygwin        | False          |
| is superuser     | False          |
| default encoding | utf-8          |
| git SHA          | None           |
+------------------+----------------+
```

and I'm using prompt-toolkit

santagada:
using readline works... and colors even look better

scopatz:
Ahh yeah, @santagada. The ANSI escape sequences will currently only work with readline.  However, I guess it wouldn't be too hard to add support for ANSI colors to prompt-toolkit either. If anyone is interested, the relevant change would need to happen in `xonsh.pyghooks.partial_color_tokenize()` https://github.com/xonsh/xonsh/blob/master/xonsh/pyghooks.py#L197

I'd be happy to help guide someone through this if they are interested.  I think that most of the tools are there or could be built relatively easily for the common cases.

melund:
So this would also be useful for windows, when the next release comes out. 
I finally got my computer upgraded to windows 10 insider preview. And lo and behold;  ConHost (the windows console) has added support for ANSI escape sequences \o/ 

So now I can run the readline shell on windows with colors. 

[...]

PR #3784, opened 2020-09-20
ANSI OSC escape sequence support (including title setting)

### LINKED ISSUE #374: iTerm2 shell integration with xonsh

dyuri:
Improved version of #3781, with generic ANSI OSC escape sequence support.

Fixes #374 and #1403.

### DIFF
--- a/xonsh/ptk_shell/shell.py
+++ b/xonsh/ptk_shell/shell.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 """The prompt_toolkit based xonsh shell."""
 import os
+import re
 import sys
 import builtins
 from types import MethodType
@@ -37,6 +38,7 @@ from prompt_toolkit.styles.pygments import (
 )
 
 
+ANSI_OSC_PATTERN = re.compile("\x1b].*?\007")
 Token = _TokenType()
 
 events.transmogrify("on_ptk_create", "LoadEvent")
@@ -74,6 +76,25 @@ def tokenize_ansi(tokens):
         else:
             ansi_tokens.append((style, text))
     return ansi_tokens
+
+
+def remove_ansi_osc(prompt):
+    """Removes the ANSI OSC escape codes - ``prompt_toolkit`` does not support them.
+    Some terminal emulators - like iTerm2 - uses them for various things.
+
+    See: https://www.iterm2.com/documentation-escape-codes.html
+    """
+
+    osc_tokens = ANSI_OSC_PATTERN.findall(prompt)
+    prompt = re.sub(ANSI_OSC_PATTERN, "", prompt)
+
+    return prompt, osc_tokens
+
+
+def write_ansi_osc(ansi_osc_code):
+    with open(1, "wb", closefd=False) as output:
+        output.write(ansi_osc_code.encode())
+        output.flush()
 
 
 class PromptToolkitShell(BaseShell):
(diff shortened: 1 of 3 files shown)


anki-code:
I think to add the iTerm2 mention (or another big names) is cool for the god of buzzwording  :)

xonsh/ptk_shell/shell.py
```diff
@@ -76,6 +78,25 @@ def tokenize_ansi(tokens):
...
 
+def remove_ansi_osc(prompt):
+    """Removes the ANSI OSC escape codes - ``prompt_toolkit`` does not support them.
+    Some terminal emulators - like iTerm2 - uses them for various things.
+
+    See: https://www.iterm2.com/documentation-escape-codes.html
+    """
+
+    osc_tokens = ANSI_OSC_PATTERN.findall(prompt)
+    prompt = re.sub(ANSI_OSC_PATTERN, "", prompt)
```
scopatz, line 89: This should be `ANSI_OSC_PATTERN.sub("", prompt)`

xonsh/ptk_shell/shell.py
```diff
@@ -76,6 +78,25 @@ def tokenize_ansi(tokens):
...
+
+    return prompt, osc_tokens
+
+
+def write_ansi_osc(ansi_osc_code):
+    with open(1, "wb", closefd=False) as output:
+        output.write(ansi_osc_code.encode())
+        output.flush()
```
scopatz, line 97: Can you do this with `print(..., file=1, flush=True)`?

scopatz: or with `print(..., file=sys.__stdout__, flush=True)`

xonsh/ptk_shell/shell.py
dyuri: Fixed.

xonsh/ptk_shell/shell.py
dyuri: Fixed, thanks for the review!

xonsh/ptk_shell/shell.py
```diff
@@ -76,6 +78,23 @@ def tokenize_ansi(tokens):
...
+    See: https://www.iterm2.com/documentation-escape-codes.html
+    """
+
+    osc_tokens = ANSI_OSC_PATTERN.findall(prompt)
+    prompt = ANSI_OSC_PATTERN.sub("", prompt)
+
+    return prompt, osc_tokens
+
+
+def write_ansi_osc(ansi_osc_code):
```
scopatz, line 94: Please add a docstring to public functions

dyuri: Since it's only a single `print` now, I removed the whole function.

scopatz: Cool, even better!

scopatz:
Thanks!

PR #3785, opened 2020-09-20
Commands to easily manage deletion of history from the current session.

Eadaen:
Solves #3033.

Provides `history clear` which purges past commands from the current session; `history off`, which stops any history from being saved from the entire current session; and `history on`, which re-enables history saving in the current session. 

Works for both JSON and SQLite backends, and it works on both the history buffer and on history already saved to disk.

Builds off of #3691. Thank you @Itay4 for starting out.

### DIFF
--- a/xonsh/history/base.py
+++ b/xonsh/history/base.py
@@ -157,9 +157,12 @@ class History:
 
     @staticmethod
     def remember_history_check(f):
-        def new_f(self, at_exit=False):
+        """Allows the decorated function to run only if
+           self.remember_history is True.
+        """
+        def new_f(self, *args, **kwargs):
             if self.remember_history:
-                return f(self, at_exit=at_exit)
+                return f(self, *args, **kwargs)
             else:
                 return
 
@@ -176,4 +179,8 @@ class History:
 
     def wipe_memory(self):
         """Reinitialises History object with blank commands."""
+        pass
+
+    def remake_file(self):
+        """Makes new file if required, after old one gets deleted."""
         pass
--- a/xonsh/history/base.py
+++ b/xonsh/history/base.py
@@ -170,4 +170,9 @@ class History:
 
     def clear(self):
         """Wipes the current session from both the disk and memory."""
-        pass
+        self.buffer = None
+        self.inps = None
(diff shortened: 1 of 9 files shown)


xonsh/history/json.py
```diff
@@ -502,3 +507,22 @@ def run_gc(self, size=None, blocking=True, force=False):
         if blocking:
             while self.gc.is_alive():  # while waiting for gc.
                 time.sleep(0.1)  # don't monopolize the thread (or Python GIL?)
+
+    def clear(self):
+        """Clears the current session's history from both memory and disk."""
+
+        def wipe_memory(hist):
```
scopatz, line 514: This probably doesn't need to be a function

scopatz:
Thanks for putting this in @Eadaen

xonsh/history/base.py
```diff
@@ -153,3 +154,21 @@ def run_gc(self, size=None, blocking=True):
             If set blocking, then wait until gc action finished.
         """
         pass
+
+    @staticmethod
+    def remember_history_check(f):
```
scopatz, line 159: Why does this decorator need to exist? It seems easier/faster to put the logic in the actual functions where it is needed. All this is doing is a simple if-condition.

Eadaen: It doesn't need to exist. In my opinion it's cleaner/more readable to use the decorator, but I'm happy to move the check into the individual functions if that's what you prefer.

scopatz: Yeah, please move them. I am pretty sure that this implementation is slower.

scopatz:
Thanks @Eadaen - would you mind running black?

Eadaen:
Ignore the last commit. I didn't realise you'd already run Black on json.py in the master branch.

scopatz:
Looks like there is a conflict here now. Sorry for the back and forth @Eadaen

Eadaen:
I think that should fix the conflict. I wasn't sure the best way to revert the problem commit.

No problem @scopatz. Thanks for making sure everything is correct.

scopatz:
Thanks!

PR #3773, opened 2020-09-17
The order of sections in docs changed and "Try it" section removed

anki-code:
Hello!

1. The order of sections in docs changed.

2. "Try it" section removed.

3. Added text to readme.

Why removed? Because:

<details>

Screenshot_20200917_212832

Screenshot_20200917_215823

Screenshot_20200917_212854

</details>

PR #3791, opened 2020-09-23
The order of sections in docs changed 2

anki-code:
Rebuilded #3773 from master


anki-code:
@scopatz here is from master and it fails. Any thoughts?

anki-code:
Aha, it looks like Windows tests fail because ⭐ UTF symbol was in the README.

scopatz:
Thanks!

PR #3792, opened 2020-09-23
Fix xontribs link

anki-code:
Sorry for the noise around this micro PRs but I fixed the link to xontribs list :)


scopatz:
Thanks! It is OK, the more PRs the better!

PR #3793, opened 2020-09-23
Change the behavior of NO_COLOR, add RESET

cafehaine:
This commit adds a new color token "RESET", and makes "NO_COLOR" a
normal color. This means that you could make some text bold while
keeping the default color using `{BOLD_NO_COLOR}`, but it also means
that the current code relying on "NO_COLOR" resetting all attributes
will break.

This commit *does not* fix the current issue (maybe not reported?) that
the prompt toolkit backend doesn't behave the same as the readline
backend:
```python
import xonsh.tools
xonsh.tools.print_color("{BOLD_GREEN}green {RED}red")
```

I know that I haven't added the news entry, but I wanted to get some more feedback before doing it.


scopatz:
Thanks for putting this in @cafehaine! Is there a way to make NO_COLOR an alias of reset and then choose a different name for the "non-color"?  I think that would prevent this from being a breaking change.

cafehaine:
I thought about it, and the best solution IMO is the following:
- allow using both RESET and NO_COLOR to reset everything
- use DEFAULT to only reset the color
- add a deprecation warning for NO_COLOR?

scopatz:
Yep that sounds good to me!

cafehaine:
I think the only thing missing now is the news item, let me know if you want anything changed!

anki-code:
Please add the changes to the documentation (i.e. [Tutorial](https://github.com/cafehaine/xonsh/blob/master/docs/tutorial.rst)).

xonsh/color_tools.py
```diff
@@ -479,3 +480,16 @@ def make_palette(strings):
...
     return palette
+
+
+def warn_deprecated_no_color():
+    """Show a warning once if NO_COLOR was used instead of RESET."""
+    global _NO_COLOR_WARNING_SHOWN
+    if not _NO_COLOR_WARNING_SHOWN:
+        import warnings
+
+        warnings.warn(
```
scopatz, line 491: This should probably just be a print to stderr

cafehaine: Fixed in ab97137

scopatz:
Thanks @cafehaine - where is NO_COLOR being aliased to RESET? I couldn't find that

cafehaine:
> Thanks @cafehaine - where is NO_COLOR being aliased to RESET? I couldn't find that

It's directly in the color regex:
https://github.com/xonsh/xonsh/pull/3793/files#diff-7e06196d05e933cc5a6298190d5223cfL90-R92

scopatz:
Ahh perfect thanks! I read that but it did not fully register that was enough

cafehaine:
Thanks!

anki-code:
@cafehaine the message about "NO_COLOR is deprecated and should be replaced with RESET" is annoying on starting.

Could you please add environment variable to switch this annoying message off?

cafehaine:
Well, the real solution would be to remove the remaining NO_COLORS left in xonsh.

I think it's better to annoy the users today, rather than breaking it without any explanations in the future

Hiding the warnings is the wrong way of tackling this issue IMO

anki-code:
@cafehaine 
The warning is good. I want to suggest to show the stacktrace of the warning in debug mode and add the text about debug mode to warning text. In this case xonsh user will has a way to found what xonsh/xontrib/xonshrc raises the warning.

cafehaine:
Oh yeah, that's a good idea, not really sure how to implement that though

anki-code:
May be like here - https://github.com/anki-code/xontrib-pipeliner/blob/master/xontrib/pipeliner.xsh#L29-L30

PR #3797, opened 2020-09-24
Added the exit function to the globals to avoid importing from sys every time

### LINKED ISSUE #3794: Feature request: exit 1 should work
Currently the commands "exit 0" or "exit 1" don't do anything.
It would be nice if it would do the same as "sys.exit(0/1)" to make it easier and faster to convert a bash script to xonsh.

anki-code:
Hello! Just call `exit(code)` to exit from the xonsh script. It's so simple and beauty!

Closes #3794 #3241 #3573

### DIFF
--- a/docs/bash_to_xsh.rst
+++ b/docs/bash_to_xsh.rst
@@ -75,7 +75,7 @@ line is ``#!/usr/bin/env xonsh``.
       - ``$COMPLETIONS_DISPLAY = 'readline'``
       - Display completions will emulate the behavior of readline.
     * - ``exit``
-      - ``sys.exit()``
+      - ``exit()``
       - Exiting from the current script.
 
 To understand how xonsh executes the subprocess commands try
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1749,6 +1749,8 @@ following:
 
 That's All, Folks
 ======================
+To exit from xonsh script just call the ``exit(code)`` function.
+
 To leave xonsh, hit ``Ctrl-D``, type ``EOF``, type ``quit``, or type ``exit``.
 On Windows, you can also type ``Ctrl-Z``.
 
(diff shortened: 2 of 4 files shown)


scopatz:
Hey @anki-code - I think this is too magical for standard xonsh, and is something that could easily go in a user's xonshrc or in a xontrib if needed. 

Also, the right way to add a new variable to the builtins is with the XonshSession class.

scopatz:
Thanks for putting this in though!

anki-code:
> I think this is too magical 

Why? I think that is obvious to have just `exit(code)` function without imports. It's like `aliases` variable. All another shells have just `exit` function and it's simple and understandable.

> and is something that could easily go in a user's xonshrc or in a xontrib if needed.

No. In common case we write xonsh script and want to have elegant way to exit without `sys.exit(1)` or `from sys import exit` - it's too much for simple thing like exit.

> Also, the right way to add a new variable to the builtins is with the XonshSession class.

Are you against with appearing `exit` in non-interactive mode globally? Or I can do it in XonshSession class you merge the PR?

anki-code:
I've reviewed builtins and my commit to `main.py` looks better than XonshSession because we need `exit(code)` only for non interactive mode. In case I add it to the `ctx` in XonshSession it will appearas in interactive mode also and `exit` command will return `<function exit>`.

If you approve moving from `exit` to `exit(code)` in interactive and not interactive I'll think deeper.

scopatz:
Huh, this is weirder than I thought.  You are right that exit should be available as a function.  I am not sure why it doesn't appear in builtins already...

anki-code:
I reviewed the code and understand how it works more deeply :)

We have the `exit` alias that just stop the prompt in IM (interactive mode) and in do nothing in NIM (not interactive mode) - it's a bug. So we need to think how to fix it.

I suggest to improve the xonsh alias and use `exit` and `exit code` in IM and MIM both. @scopatz are you agree?

anki-code:
Oh, I've found that exiting not working from alias:
```python
import sys
def exx():
        sys.exit()
aliases['exx'] = exx
exx
print(123)
```
Returns `123`...

anki-code:
Closed to open new PR :)

PR #3805, opened 2020-09-28
Fix logic in working directory check

dannysepler:
Oops! The "||" did not do what I thought it did.

(It ran the first command, and if that command gave a 1, then ran the second command and returned the result of the second command.) Instead, I'll just keep the two checks in their own commands.


scopatz:
Thanks @dannysepler!

PR #3808, opened 2020-09-29
Added get_path to environment

anki-code:
Hi!

This PR adds `get_path` function to env.

Before:
```python
penv = p"{${...}.get('ENV','/default')}"   # overloaded syntax

if penv.exists():
    ...
```

or
```python
from pathlib import Path                    # overloaded imports
penv = Path(${...}.get('ENV','/default'))   # overloaded calls

if penv.exists():
    ...
```

After:
```python
penv = ${...}.get_path('ENV','/default')   # the clean view

if penv.exists():
    ...
```

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -13,8 +13,8 @@ import collections
 import collections.abc as cabc
 import subprocess
 import platform
+import pathlib
 
-from pathlib import Path
 from xonsh import __version__ as XONSH_VERSION
 from xonsh.lazyasd import LazyObject, lazyobject
 from xonsh.codecache import run_script_with_cache
@@ -2004,8 +2004,11 @@ class Env(cabc.MutableMapping):
             return default
 
     def get_path(self, *args, **kwargs):
+        """The environment will look up default values from its own defaults if a
+        default is not given here and then Path-object will return.
+        """
         v = self.get(*args, **kwargs)
-        return Path(v) if v is not None else None
+        return pathlib.Path(v) if v else None
 
     def rawkeys(self):
         """An iterator that returns all environment keys in their original form.
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -14,6 +14,7 @@ import collections.abc as cabc
 import subprocess
 import platform
 
+from pathlib import Path
 from xonsh import __version__ as XONSH_VERSION
 from xonsh.lazyasd import LazyObject, lazyobject
 from xonsh.codecache import run_script_with_cache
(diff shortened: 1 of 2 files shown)


xonsh/environ.py
```diff
@@ -14,6 +14,7 @@
 import subprocess
 import platform
 
+from pathlib import Path
```
scopatz, line 17: This should just be `import pathlib` and grouped with the standard library imports

xonsh/environ.py
```diff
@@ -2002,6 +2003,10 @@ def get(self, key, default=None):
         else:
             return default
 
+    def get_path(self, *args, **kwargs):
```
scopatz, line 2006: This needs a docstring

scopatz:
Hi @anki-code - thanks for putting this in. I am a little unclear on the semantics here.  How do you know an environment value would be convertable to a path in general? Wouldn't it be better for environment variables to be registered with Path to/from convertors?

anki-code:
> Wouldn't it be better for specific environment variables to be registered with Path to/from convertors?

Don't understand. Give an example please.

xonsh/environ.py
anki-code: Fixed

xonsh/environ.py
anki-code: Fixed

anki-code:
>  I am a little unclear on the semantics here. How do you know an environment value would be convertable to a path in general?

It's not needed to know. Because the main use case is to get the path from variable and if it's exists (here is the main check) then use it. For example I have docker-compose file that have 3 environment variables that contains paths. The paths in the environment variables is just requirement and if the path exist use it (example in the first message of the thread).

> Wouldn't it be better for specific environment variables to be registered with Path to/from convertors?

Don't understand. Give an example please.

scopatz:
Yeah I guess my point is that in xonsh, environment variables can be of any type (such as dicts) which don't naturally convert straight into Paths. 

> Give an example please.

Like in the default env vars: https://github.com/xonsh/xonsh/blob/master/xonsh/environ.py#L687 or in the down stream projects https://github.com/regro/rever/blob/master/rever/environ.xsh#L268

anki-code:
> Yeah I guess my point is that in xonsh, environment variables can be of any type (such as dicts) which don't naturally convert straight into Paths.

Right. In this case there be an exception and it's absolutely ok (we expect path but get dict/list).

> Like in the default env vars

Thanks! It's interesting! But I see not the same behavior:

```
$ $QWE = '/tmp'
$ ${...}.register(name='QWE', type='path')
$ $QWE
'/tmp'
```
No `path` type here. Why?

I understand that by current logic I should register before use BUT in case with docker I already have environment variables before registration.

anki-code:
Closed this because I should patch and use register.

anki-code:
Closed this because I should patch and use `register` function.

What I should patch:
1. rename `path` type to `env_path`
2. create `path` type as `pathlib.Path`
3. fix getting environment variable when it was set before register

PR #3812, opened 2020-09-30
feat: integrate flake8-docstring

jnoortheen:
most of the rules are disabled since it will take longer to fix them all

maybe in future version we can fix them one by one


scopatz:
Thanks @jnoortheen! I am fine with this, but we should note that ply does use docstrings to describe the parser.

PR #3813, opened 2020-09-30
Improved getting typed registered variables and path type changed

anki-code:
Hi! xonsh is great!

In this PR:

1. The `path` type in `${...}.register` was renamed to `env_path` as it should be and added new `path` type instead that represent `pathlib.Path`. Now you can register typed environment variables that will be converted to `Path` objects.

2. Fixed getting typed registered environment variable when it was initialized before registration:

    ```python
    $QWE='/tmp'
    ${...}.register(name='QWE', type='path')
    $QWE
    # Before fix: return string '/tmp'
    # After fix: return PosixPath('/tmp') and immediately do $QWE.exists()
    ```

3. Tests added.


anki-code:
@scopatz I'm done, please merge ;)

xonsh/tools.py
```diff
@@ -1171,11 +1171,22 @@ def ensure_string(x):
...
+    """This tests if something is a path."""
+    return isinstance(x, pathlib.Path)
+
+
 def is_env_path(x):
     """This tests if something is an environment path, ie a list of strings."""
     return isinstance(x, EnvPath)
 
 
+def str_to_path(x):
```
scopatz, line 1184: Can you please add some tests to test_tools.py for the functions you added here.

scopatz:
Thanks! I think this just needs a few more tests

xonsh/tools.py
anki-code: Done )

anki-code:
@scopatz a few more tests done )

scopatz:
Thanks a ton @anki-code!

PR #3814, opened 2020-09-30
Added exit to builtins in not interactive mode

### LINKED ISSUE #3794: Feature request: exit 1 should work

anki-code:
Hello again! I reviewed the code and understand how it works more deeply :)

We have the `exit` alias that just stop the prompt in IM (interactive mode) and do nothing in NIM (not interactive mode) - it's a bug. Current `exit` works as alias but `sys.exit` not working from inside an alias.

Next 2 hours I traced the builtins and main scripts and found that it's very complex to pass interactive/not interactive mode flag thru the call stack to the builtins. 

@scopatz this PR is most simple way to achieve the goal: avoid breaking the `exit` in interactive mode and have an `exit()` by default in not interactive. I did my best.

~Also I deleted the line which do nothing because the `XONSH_INTERACTIVE` variable is set in `premain` function already.~

Closes #3794 #3241 #3573. Please merge :)


xonsh/main.py
```diff
@@ -441,10 +441,13 @@ def func_sig_ttin_ttou(n, f):
...
     history = builtins.__xonsh__.history
     exit_code = 0
+
+    if shell and not env["XONSH_INTERACTIVE"]:
+        shell.ctx.update({"exit": sys.exit})
+
     try:
         if args.mode == XonshMode.interactive:
             # enter the shell
-            env["XONSH_INTERACTIVE"] = True
```
scopatz, line 447: I think this line should probably stay because it is possible to call `main_xonsh()` without calling `premain()`, namely in the tests.

scopatz:
Thanks for putting this in @anki-code.  And sorry for the slow reply, I was out on vacation

anki-code:
@scopatz the line added with clarification comment. Done to merge.

scopatz:
Thanks!

PR #3816, opened 2020-09-30
Improvements #3813 and a bit of docs

anki-code:
* Improved logic in #3813
* Added a couple of words about registration to the Tutorial


anki-code:
@scopatz I'm done :)

xonsh/environ.py
```diff
@@ -1985,10 +1985,11 @@ def __getitem__(self, key):
         ):
             self._detyped = None
 
-        validator = self.get_validator(key)
-        converter = self.get_converter(key)
-        if not validator(val):
-            val = converter(val)
+        if val:
```
scopatz, line 1988: You can validate against None, empty lists, empty strings etc

anki-code: Yeah, this should be the None check. Thanks! 
Pushed.

anki-code:
@scopatz you can merge #3819 instead of this PR because it has this changes.

anki-code:
@scopatz it looks very clear!

Example:
https://github.com/anki-code/metabase-sql-wrapper/blob/xonsh_future/run.xsh#L27-L43

Please merge :)

anki-code:
I want to change the logic and still working on this. Don't merge.

anki-code:
@scopatz I've changed the logic. The solution to change the variable on getter is not so good.

Current registration logic looks more clear:

When you do 
```
${...}.register(re.compile(r"\w*_DIR$"), 'path')
```
the register function check all variables that matched with regexp and then try to convert it to `path` type. If it was successfully the changes will be commited. If at least one variable hasn't converted successfully (raises `TypeError`) no any changes will be in variable values. In other words `register` shows the error message and nothing will be changed.

If we see that some variable fails on type converting (e.g. we want to register `*_DIR` as `path` but we have boolean `AUTOJUMP_PRINT_DIR=False`) we should just register this variable by name before registering regexp.

This logic looks more solid and clean.

anki-code:
@scopatz 
Finally it was long way. Now I understand why you're not hurry and it's really right. I taken interesting part of xonsh and happy to understand how it works.

At the end of environ story I implement the best logic ever:

1. When you do `${...}.register(re.compile(r"\w*_DIR$"), 'path')` all `*_DIR` variables with convertable type will be converted to `Path`. If there are some not convertable variables it will leave unchanged.

2. If variable that has type that differ than registered (e.g. `$AUTOJUMP_PRINT_DIR` has `False` value) it could be changed to any value without any errors (only warning in debug mode). 

3. But if we create new variable OR existent variable has `None` value OR existent variable has type that equal to registered and we try to set unregistered type (`$NEW_DIR=1`) there will be a `TypeError` like it works now.

I added `test_before_register_var_path` to cover checking of all this logic.

As result we can register regexps with less fear of blocking or corrupt the environment.

anki-code:
`macos` tests is unstable... bump again

anki-code:
@scopatz it looks very clear ([example](https://github.com/anki-code/metabase-sql-wrapper/blob/xonsh_future/run.xsh#L27-L43)). Done to merge.

tests/test_environ.py
```diff
@@ -422,23 +416,65 @@ def test_register_custom_var_str(val, converted):
     assert env["MY_SPECIAL_VAR"] == converted
 
 
+def test_register_var_without_setting_val():
+    env = Env()
+    env.register("MY_SPECIAL_VAR", type="str")
+    # If variable registered but not initialized it should be None.
+    assert env["MY_SPECIAL_VAR"] == None
```
scopatz, line 423: should be `is None`

anki-code:
Done to merge.

scopatz:
I'd rather keep the env logic as simple as possible. I am not sure I understand why this is needed.

anki-code:
@scopatz Can we solve it today? Because:
1. Let's start from that `register` code is not so good and someone merged it without any deep diving. I did my best to understanding this and create minimal changes to achieve more clean algorithm.
2. The `${...}` call is broken in master and the code in master less clear then in this PR.
3. I gave you real life example above (link)
4. I wrote complete pack of tests.
5. The changes in fact minimal - converting the type when you call `register` or set the new value and NO changes in other code.

It's my bad that previous PR breaks the `${...}` but I did my best to improve the logic. Please review this closer and you see that I fixed the bug in logic (not my) when environment variable initialized before registering and I made the logic more clear and no additional clarifications or documentation is not needed.

anki-code:
Ready to merge.

docs/tutorial.rst
```diff
@@ -210,7 +210,6 @@ this type is imposed based on the variable name. The current rules are pretty
 simple:
 
 * ``\w*PATH``: any variable whose name ends in PATH is a list of strings.
-* ``\w*DIRS``: any variable whose name ends in DIRS is a list of strings.
```
scopatz, line 213: Why remove this from the docs

scopatz: DIRS is very similar to PATH

tests/test_environ.py
scopatz, line 423: Well, it should be the default value not, None

scopatz: And if the default is not specified it should be `DefaultNotGiven`, not sure how `None` gets in here

xonsh/environ.py
```diff
@@ -1985,20 +1985,39 @@ def __getitem__(self, key):
...
     def __setitem__(self, key, val):
         validator = self.get_validator(key)
         converter = self.get_converter(key)
         detyper = self.get_detyper(key)
+        old_value = self._d[key] if key in self._d else self._no_value
         if not validator(val):
-            val = converter(val)
+            old_value_validate = validator(old_value)
+            try:
+                val = converter(val)
```
scopatz, line 1998: I pretty strongly feel that any errors raised by the conveters should be passed up to the user.

tests/test_environ.py
scopatz: Ohhh I see the confusion. I think the `default` keyword argument of register should be `DefaultNotGiven`, not `None`

xonsh/environ.py
```diff
@@ -2138,6 +2157,28 @@ def register(
                     f"Default value for {name} does not match type specified by validate"
                 )
 
+        # Convert existing variables
+        name_type = builtins.type(name)
+        converted_values = {}
+        for k, val in self._d.items():
```
scopatz, line 2163: I don't think registering a single env variable should affect any other environment variables.

docs/tutorial.rst
anki-code: Because there is no DIRS variable in the code. I'm glad that you ask about it because may be it was in the code and deleted now in issues reason? If no I can add DIRS regex like a PATH.

xonsh/environ.py
anki-code: It's not affect any other variables.

xonsh/environ.py
anki-code: Let's discuss about the real life example. There is `$AUTOJUMP_PRINT_DIR` from `autojump` package.  This variable has `False` value.
I want to register `*_DIR` as `path` variables. Should I get notifications about error of converting `$AUTOJUMP_PRINT_DIR` to `path` every day and every time I run the shell? 
My answer is no. I suggest convert variables which we can convert and leave unconverted others.
If I'll face with `$AUTOJUMP_PRINT_DIR` should be `path` but I see `bool` I'll register it if I want.

tests/test_environ.py
anki-code: Sorry, but I would not change `None` to `DefaultNotGiven` in this PR because it's a new requirement. This change could be made in another PR.

anki-code:
I commented the review notes.

xonsh/environ.py
anki-code: Don't forget that we can register the variable any time. For example you have 10 xontribs with different `*_DIR` variables with different types. Along your suggestion we should show i.e. 5 error messages to user about wrong convertation. Right? It's very annoying for user because it don't want to resolve this. In the current logic we just convert what we can and leave other values in original value - it is non breaking logic.

docs/tutorial.rst
scopatz: Oh really? That seems like a bug potentially

tests/test_environ.py
scopatz: OK, I think that PR needs to go in first then.

xonsh/environ.py
scopatz: Yeah, I guess what I am trying to say is that registering should be fast and we should not loop through the env every time a variable is registered.

docs/tutorial.rst
anki-code: `\w*DIRS` added with test.

xonsh/environ.py
anki-code: Loop for single variables excluded.

tests/test_environ.py
anki-code: No. Because `bool(DefaultNotGiven) == True` instead of `bool(None) == False`. We should use `None` value.
If we will use `DefaultNotGiven` we can't do this:
```
${...}.register('VAR', 'path')
if $VAR and $VAR.exists():   # in case of DefaultNotGiven object we will have an exception
   ...
```

anki-code:
I've added changes and commented the review notes again.

scopatz:
@anki-code - maybe we could break this into several smaller PRs to address each concern individually. Starting with the fix for #3813 would be great.

anki-code:
I don't see any big concerns around. I see that you don't pretty diving into the logic and suggest the changes that breaks or makes work harder then it should be. For example we don't need change None to NotFound - it breaks the `if $VAR` logic. Or we don't need to annoying the user by warnings. If he want he'll switch the debug mode on.

It's my thoughts and instead of them I open to discuss.

scopatz:
Actually, the more I think about this, the more I think what you want is a `revalidate()` or `check()` method on the environment itself. This would loop through the environment variables and make sure that the environment is consistent.

anki-code:
To diving in environ register we should know three aspects:

1. At start we have environment as string values and we convert it to objects or keep strings (i.e. `PATH=/tmp`)

2. Then xontribs or xonshrc are loaded and set typed variables (i.e. `CRAZYPATH=False`)

3. Finally we register some variables that could be names or regexps (i.e. `register('*PATH', 'path')`).

The problem of this situation is how to be agile around trying to keep variable type. I agree that propably we need more functions and more checks. But at this time this PR closes 90% of cases.

In the example above `CRAZYPATH` will stay `False` and `PATH` will be `Path('/tmp')`. The family is happy. The app that use `CRAZYPATH` as boolean will not breaked. The app that use `PATH` as `Path` will not breaked.

anki-code:
> This would loop through the environment variables and make sure that the environment is consistent.

What "consistent" means for the environment with variables of different types?

For example we `register('*_DIR', 'path')` what does it mean? If it means that ALL variables should be `path` it's wrong(!) because any xontrib/app/script could set CRAZY_DIR with any object BEFORE we call `register`.

Current realised logic in this PR is sounds in short:
1. Change what you can
2. Leave unchanged what you can't (and keep it changeble)
3. Prevent creating new wrong type variables

What is wrong in this logic?

scopatz:
Basically, I would like this split out into multiple methods so that people could choose when to rerun the validators on items already in the environment. These operations shouldn't be mixed with getting/setting/registering variables.

scopatz:
For example, imagine a situation where you are going to register many variables.  Revalidating the environment should wait until all variables are registered and set.  It should not happen each time a new variable is registered or set. A separate method would enable this.

anki-code:
> you are going to register many variables

How many? Maximum I register 4 :) For what case it is actual? Maybe it's overengineering?

scopatz:
Rever registers 40 or so variables at once: https://github.com/regro/rever/blob/master/rever/environ.xsh#L140

PR #3817, opened 2020-10-01
Added printx to builtins

anki-code:
Hi! xonsh is better and better after every PR!

The second thing (after builtin exit) I really miss while writing xonsh scripts is a simplest way to print colored messages. Colors make terminal alive and this is the good way to give developers easiest way to do print with colors without imports/xonshrc/xontribs.

Before this PR:
```python
from xonsh.tools import print_color   # imports... I love imports... but when I'm doing it every day...
print_color('{RED}Color!')      # long names... I love long names... but my keys are erased already...
```

After this PR:
```python
printx('{RED}Color!')    # any time, everywhere, just when the thought got into the head
```

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -1464,6 +1464,7 @@ class XonshSession:
             "execx": "__xonsh__.builtins.execx",
             "compilex": "__xonsh__.builtins.compilex",
             "events": "__xonsh__.builtins.events",
+            "print_color": "__xonsh__.builtins.print_color",
             "printx": "__xonsh__.builtins.printx",
         }
         for refname, objname in proxy_mapping.items():
@@ -1487,6 +1488,8 @@ class XonshSession:
             "compilex",
             "default_aliases",
             "events",
+            "print_color",
+            "printx",
         ]
 
         for name in names:
@@ -1503,7 +1506,7 @@ class _BuiltIns:
         self.execx = None if execer is None else execer.exec
         self.compilex = None if execer is None else execer.compile
         self.events = events
-        self.printx = print_color
+        self.print_color = self.printx = print_color
 
 
 class DynamicAccessProxy:
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -44,6 +44,7 @@ from xonsh.tools import (
     globpath,
     XonshError,
     XonshCalledProcessError,
+    print_color,
 )
 from xonsh.lazyimps import pty, termios, fcntl
 from xonsh.commands_cache import CommandsCache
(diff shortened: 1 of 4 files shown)


anki-code:
@scopatz I'm done :)

xonsh/built_ins.py
```diff
@@ -1468,6 +1469,8 @@ def link_builtins(self, execer=None):
             proxy = DynamicAccessProxy(refname, objname)
             setattr(builtins, refname, proxy)
 
+        builtins.printx = print_color
```
scopatz, line 1472: I think this needs to be a dynamic proxy, like everything else. Also, I would prefer the name to remain the same. That is it should be `print_color` in builtins too.

anki-code: With this long name this PR loses the power.
We already have `evalx`, `execx` instead of `eval_xonsh`, `exec_xonsh`. Why we should stay here on `print_color`. 
I can suggest an alternative - `printc` instead of `printx`. If it's also not a variant I'm going to close this PR.

scopatz: Actually that is a good point that those should probably have longer names too.  People can always add shorter names in their xonshrc if they want. I am personally tending in the direction that explicit is better than implicit

scopatz: But no matter what the name it does need to be a dynamic proxy

anki-code: I understand you Pythonish-way point. But xonsh more than Python. It makes a great try to replace shell scripting. From this point the way to reinvent bycicles i.e. creating xonshrc/xontrib/common_lib to add exit/print_color every time is not looks so useful for ordinary administrator/devops (I tested this on friends!). When you just want to write some commands and print beautiful colored text and exit you don't want to import something or thinking about what path you should register to init rc/xontrib/my_awesome_buildins_lib.

anki-code: > But no matter what the name it does need to be a dynamic proxy

Thank you for your help! I've moved `printx` to proxy and also added example to the "Bash to xonsh" doc. Take a look at the doc! I'm happy to write the docs where worst things becomes beautiful! Let's add printx! It's small thing that makes a great mood while scripting.

anki-code:
Ready to merge.

anki-code:
Globally we can think about it as "entrance threshold". We should create minimal builtins that allows new users come in with minimal efforts. The understanding what should be imported, or why we should create additional files - it's the additional effort. You ask user to make additional effort while he was not get the valuable result. It's wrong. 

In case of `printx` we should deliver easiest way to print colored message and when user will got the result (red string) it will screm "Aha! It's valuable! It works!" and will use the xonsh and some day he will want to understand how to use additionals.

anki-code:
Ready to merge.

scopatz:
I guess I am OK with both names, if you want to add both, but I think the docs should tend to use `print_color`.  

My issue with the `printx` name is that there is no "xonsh" way of printing, unlike evalx, etx, which function on xonsh code.  `print_color` really is just for printing colors, and xonsh happens to use it.  I understand that you think of this as xonsh's way of printing, but it doesn't really fit my brain :stuck_out_tongue_winking_eye:

scopatz:
The print function also needs to be in the unlink builtins too

anki-code:
> The print function also needs to be in the unlink builtins too

* Added `print_color` to builtins
* Added unlinking (thanks!)
* I mention only `print_color` in the doc

Ready to merge.

scopatz:
Thanks @anki-code!

PR #3819, opened 2020-10-01
Four changes around registered environment path types

anki-code:
In this PR:
1. Fix and docs from #3816
2. Added ``_DIRS``,``_FILES``,``_FILE``, ``_DIR`` to registered env_path and path types.
3. ``$EXAMPLE_PATH.paths`` is now function that returns list of Path objects by default. And the ``type`` argument could be set to ``str`` to get list of strings - it's the same as ``list($EXAMPLE_PATH)``.
4. ``str_to_path`` function returns the value without changes if it's not a ``str`` type to save environment variables that detected as path but not a path (e.g. ``$I_LIKE_PATH=True`` stay as ``True`` and ``$REAL_DIR='/tmp'`` will be `Path('/tmp')`).
5. Tests fixed, docs added, news painted

This allows:

1. Add `_FILE` postfix and the registering of env variable as path variable is not needed. 

2. And on Linux I see improvement for detection (e.g. `GTK_RC_FILES`, `XDG_CONFIG_DIRS`, `XDG_RUNTIME_DIR`).

3. `$ENV_PATH.paths()` allows iterate Path objects it's more useful:

    Before:
    ```python
    for p in $ENV_PATH:
        if pf'{p}'.exists(): ...     # syntax and string modifiers overhead
    ```
    or
    ```python
    from pathlib import Path     # import overhead
    for p in $ENV_PATH:
        if Path(p).exists(): ... # better but import
    ```

    After:
    ```python
    for p in $ENV_PATH.paths():      # pretty sane
        if p.exists(): ...
    ```

As result we've got great interface between path environment variables and xonsh.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -222,12 +222,12 @@ PATH examples:
 .. code-block:: xonshcon
 
     >>> $PATH
-    ['/home/snail/.local/bin', '/home/snail/sandbox/bin',
+    ['/home/snail/.local/bin', '/home/anki/sandbox/bin',
     '/home/snail/miniconda3/bin', '/usr/local/bin', '/usr/local/sbin',
     '/usr/bin', '/usr/sbin', '/bin', '/sbin', '.']
     >>> $PATH.paths()
     [PosixPath('/home/snail/.local/bin'),
-     PosixPath('/home/snail/sandbox/bin'),
+     PosixPath('/home/anki/sandbox/bin'),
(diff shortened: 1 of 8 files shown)


anki-code:
@scopatz I'm done ;)

anki-code:
I've faced with some not trivial logic around set and get item from environment. So I close this PR to thinking. Please merge #3819 instead.

PR #3822, opened 2020-10-01
Syntax highlighting for xsh on Github

anki-code:
Details - https://github.com/github/linguist#using-gitattributes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,7 +12,7 @@
 *.rst text
 *.sh text
 *.txt text
-*.xsh text
+*.xsh text linguist-language=Python
 *.yaml text
 *.yml text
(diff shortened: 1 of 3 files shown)


scopatz:
I think this is a reasonable kludge, but it would be best for someone to add xonsh to linguist https://github.com/github/linguist/blob/master/CONTRIBUTING.md

scopatz:
Could you add some docs for this to the developer docs? That way we'd have something to point people to

anki-code:
@scopatz I've added the section about it to wishlist. 
And because you have an access to change this PR I suggest you to remove the old sections from wishlist because I don't know exactly what is the active now.

PR #3824, opened 2020-10-02
test: regression to improve test suit

jnoortheen:
I am looking to add tests for the prompt rendering. 

When I try something like this 
```py
def test_xonsh_shell(capsys):
    # todo: make sure that shell executes and prompt is getting loaded as expected
    import xonsh
    from xonsh.built_ins import load_builtins, unload_builtins
    load_builtins()
    import xonsh.main
    # args = premain(['exit'])
    # main_xonsh(args)

    with pytest.raises(SystemExit):
        xonsh.main.main(['exit'])

    capture = capsys.readouterr()
    print(capture.out, capture.err)
```

it results in error saying builtins AttributeError. How should I test them ?

also I am curious why the tests are split and run separately. Any help and insight is appreciated.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -24,7 +24,7 @@ from xonsh.jobs import tasks
 from xonsh.events import events
 from xonsh.platform import ON_WINDOWS
 
-from tools import DummyShell, sp, DummyCommandsCache, DummyEnv, DummyHistory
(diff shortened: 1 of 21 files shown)


scopatz:
@jnoortheen - can you post the traceback for the attribute error? That might be this weird pytest issue

Issue #3825, opened 2020-10-02
github workflow refactoring


jnoortheen:
## xonfig

<details>

```
+------------------+----------------------+
| xonsh            | 0.9.22               |
| Git SHA          | 00a42e01             |
| Commit Date      | Sep 18 19:57:59 2020 |
| Python           | 3.8.5                |
| PLY              | 3.11                 |
| have readline    | True                 |
| prompt toolkit   | 3.0.3                |
| shell type       | prompt_toolkit       |
| pygments         | 2.7.1                |
| on posix         | True                 |
| on linux         | True                 |
| distro           | manjaro              |
| on darwin        | False                |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
| on jupyter       | False                |
| jupyter kernel   | None                 |
+------------------+----------------------+

```

</details>

## Current Behavior

I am curious about knowing the current setup of the github workflow. Any of the core members please clarify. I am ready to start working on them 

1. having multiple workflow files
  - there are 3 platforms * 3 py versions 
  - issue: upon failure it sends multiple emails. it is annoying. 
  - single file with matrix would suffice. I want to know why there are multiple files and why it needs generation.
2. conda on ci and requirements file splitup
  - the dependencies may diverge and we will never know. the local testing will not be the same as ci
  - can't we use tox with requirements/poetry/pipenv 
  - single set of dependency declaration will surely help. I suggest we use poetry+tox+tox-gh-actions to run the same setup on local and ci server
3. black versions 
  - in the precommit it uses pre-built hook which uses old version
  - in the requirements file, no version is mentioned
  - in the ci it uses pre-built action
  - suggesion: can't we just use a single version all across the systems. it is a python package. it pre-commit we can configure to run local package. the same can be used in ci as well. so a single version across systems and no confusions.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

jnoortheen:
@scopatz

gforsyth:
Hi @jnoortheen -- improvements to the CI are always welcome.  I'll try to answer a few of your questions

> * issue: upon failure it sends multiple emails. it is annoying.
You can turn off those emails.

> * single file with matrix would suffice. I want to know why there are multiple files and why it needs generation.
[...]

PR #3826, opened 2020-10-02
feat: add mypy to project

jnoortheen:
to have a better code quality.


ci/environment-3.6.yml
```diff
@@ -12,6 +12,7 @@ dependencies:
   - psutil
   - matplotlib
   - flake8
+  - mypy=0.782
```
scopatz, line 15: Is there a reason to pin this so tightly?

scopatz: I would prefer no pin

requirements/tests.txt
```diff
@@ -8,5 +8,6 @@ prompt-toolkit>=2.0
 pygments>=2.2
 codecov
 coverage
-black
+black==19.10b0 --pre
```
scopatz, line 11: Same, can this be loosened?

ci/environment-3.6.yml
jnoortheen: Yes. Mypy versions break the checks often

requirements/tests.txt
jnoortheen: same reason here. If you check the CI, you could see the version change breaks the formatting. This is put to match the CI's version.

scent.py
```diff
@@ -0,0 +1,50 @@
...
+def python(*_):
+    group = int(time.time())  # unique per run
+
+    for count, (command, title) in enumerate((
+            (('dmypy', 'run', "--", "xonsh"), "Lint"),
+            (('flake8', '--count'), "Lint"),
+            (('pytest', 'tests/test_main.py'), "Test main"),
+            (('pytest', 'tests/test_ptk_highlight.py'), "Test ptk highlight"),
+            (('pytest', '--ignore', 'tests/test_main.py', 'tests/test_ptk_highlight.py'), "Test Rest"),
+    ), start=1):
```
scopatz, line 34: Rather than running pytest, should it just run `run-tests.xsh`?

jnoortheen: it is a continuous test runner. Running all would take time. I used to hide/show each commands.

xonsh/ansi_colors.py
```diff
@@ -3,6 +3,7 @@
 import sys
 import warnings
 import builtins
+import typing as tp
```
scopatz, line 6: It is probably more consistent to just `import typing` here

xonsh/completers/bash_completion.py
```diff
@@ -16,6 +16,8 @@
 
 __version__ = "0.2.6"
 
+import typing as tp
```
scopatz, line 19: This should be grouped with other standard library imports

xonsh/ansi_colors.py
jnoortheen: it was taking lot of space. the Union and other complex notations will become long too. Do you want to replace yet?

xonsh/fs.py
```diff
@@ -2,97 +2,4 @@
...
-        return fsencode, fsdecode
-
-    fsencode, fsdecode = _fscodec()
-    del _fscodec
-
-    def open(file, *pargs, **kwargs):
-        if isinstance(file, PathLike):
-            file = fspath(file)
-        return io.open(file, *pargs, **kwargs)
+from os import PathLike, fspath, fsencode, fsdecode  # noqa
```
scopatz, line 5: maybe just remove this whole file at this point, if it is truly not needed anymore

xonsh/completers/bash_completion.py
jnoortheen: OK. I will update. Maybe using isort will do this automatically. Will accept adding that too to this PR?

xonsh/fs.py
jnoortheen: Ok

xonsh/ptk_shell/formatter.py
```diff
@@ -20,6 +20,7 @@ def __call__(
         fields=None,
         threaded=False,
         prompt_name: str = None,
+        **_
```
scopatz, line 23: What it is purpose of this?

jnoortheen: to match the signature - mypy was giving error before.

xonsh/tokenize.py
```diff
@@ -330,10 +330,17 @@ def maybe(*choices):
...
-_redir_check = {"{}>".format(i) for i in _redir_names}.union(_redir_check)
-_redir_check = {"{}>>".format(i) for i in _redir_names}.union(_redir_check)
-_redir_check = frozenset(_redir_check)
+
+
+def _get_redir_check():
+    _map = set(_redir_map)
+    _check1 = {"{}>".format(i) for i in _redir_names}.union(_map)
+    _check2 = {"{}>>".format(i) for i in _redir_names}.union(_check1)
+    return frozenset(_check2)
```
scopatz, line 339: Why is this function needed?

xonsh/tools.py
```diff
@@ -719,7 +721,7 @@ def dec(obj):
 # Copyright (c) Python Software Foundation. All rights reserved.
 class _RedirectStream:
 
-    _stream = None
+    _stream: str
```
scopatz, line 724: I think this needs to be set to None

scopatz:
Very cool @jnoortheen! Thanks for putting this in.

Do you have any timing statistics on how this affects startup time?

xonsh/ptk_shell/formatter.py
scopatz: Which signature? Of the same method on the parent?

xonsh/fs.py
scopatz: Thanks!

xonsh/completers/bash_completion.py
scopatz: I kind of dislike isort, so I would prefer if it wasn't added

xonsh/ptk_shell/formatter.py
jnoortheen: Yes. It has to have same signature.

xonsh/tools.py
jnoortheen: OK .

jnoortheen:
As of now no. The importing of typing module will add some time. But that will be negligible. Runtime dont get affected by annotations.

xonsh/ansi_colors.py
scopatz: Well, yeah, so the issue is that the same abbreviation needs to be used everywhere and `tp` will enter the globals of the amalgamated module.  I am OK with `import typing as tp`, but mostly because this is something that then gets used globally in annotations elsewhere. Is this is standard way to abbreviate typing?

requirements/tests.txt
scopatz: Ahh yeah, I guess I was hoping for a <=19.10b0

scent.py
scopatz: I am not sure I understand. Can you explain a bit more please?

ci/environment-3.6.yml
scopatz: Yeah, so shouldn't that be part of the test suite?

xonsh/tokenize.py
jnoortheen: to replace the imperative black that  used the same variable name. Also this is very outdated code taken from older python version. the mypy was complaining about it.

scopatz:
Yeah, I agree in principle. I have heard of type annotations adding significant time though in other projects.  If you could prove that, it would be best. Thanks!

jnoortheen:
https://stackoverflow.com/questions/41692473/does-python-type-hint-annotations-cause-some-run-time-effects 

it was in the intial phase of the PEP. Now the parser ignores the annotation part.

jnoortheen:
since Python 3.7 the startup time impact is also fixed
https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations

ci/environment-3.6.yml
jnoortheen: it is very much like a linter than a test suit. as you see here, it breaks between versions. 
https://github.com/pypa/pip/pull/5792

so it is better to pin them. It is a not library requirement that we specifiy in requirements.txt . so it is safe to pin.

requirements/tests.txt
jnoortheen: nope that would break as well. If anyone is to have a black version of 18. something. 

also the semantic versioning is not followed by black either. so it will be hard to pin relatively.  this is a development dependency, I was having failing CI jobs, because I have accidentally had latest version installed with the requirements file.

scent.py
jnoortheen: https://pypi.org/project/sniffer/ 

it is the config file for this tool. 

it runs the commands on terminal as the code changes. so it is easier to refactor the large number of errors from mypy. It is a developer tool. Not going to end-up in the test-suit or the build tools .

xonsh/ansi_colors.py
jnoortheen: I have seen it in some places in same vein as `numpy as np` . but definitely not the standard way. it is to reduce the visual Noise. If we keep it the same in all places, then I think we can add a flake8 rule.

ci/environment-3.6.yml
scopatz: Yeah, I just don't want the maintenance burden of having to update the pins

scent.py
scopatz: I guess my question is why are you running pytest directly rather than through the run-tests script? I would like to keep the list of test files that need to be run in a separate process in a single location

xonsh/completers/bash_completion.py
jnoortheen: now it has black support though. Anyways I will update this manually.

xonsh/ptk_shell/formatter.py
scopatz: According to mypy? That isn't generally true in Python

scopatz: Sorry I am pretty new to mypy because I have heard about how often it breaks :wink:

xonsh/tokenize.py
scopatz: Is there a way to keep it imperative by changing variable names?

ci/environment-3.6.yml
jnoortheen: if we use poetry, manging dependencies will not be a pain. Maybe I will open another PR if you want for that.

scopatz:
Ahh yeah I meant the xonsh startup time specifically

scent.py
jnoortheen: run-tests also has flake8 and mypy 

it is not really a make file, where we selectively run each task.

xonsh/ptk_shell/formatter.py
jnoortheen: Yes it adds some kind of rigidity to the dynamic ways. But I think they help bigger projects.

xonsh/tokenize.py
jnoortheen: yes we can do it. Will it also end up in global namespace ?

jnoortheen:
> Ahh yeah I meant the xonsh startup time specifically

a quick running gave these output

in master branch
image


in typing branch
image


this is with my rc though

scent.py
scopatz: Ahh I see ok.

xonsh/ansi_colors.py
scopatz: Yeah, it needs to be the same in all places, so a flake8 rule would be helpful.

xonsh/completers/bash_completion.py
scopatz: Cool thanks

xonsh/tokenize.py
scopatz: Yes, unless deleted, because tokenize is not part of the [excluded list](https://github.com/xonsh/xonsh/blob/b5a7439dca1ba9f4d587889daea2b90e1b128d49/xonsh/__init__.py#L4-L6)

xonsh/tools.py
jnoortheen: done.

xonsh/tokenize.py
jnoortheen: so can we replace this with the latest one from the Python's stdlib ?

xonsh/fs.py
jnoortheen: done.

xonsh/completers/bash_completion.py
jnoortheen: done

xonsh/ansi_colors.py
jnoortheen: Ok

xonsh/ptk_shell/formatter.py
jnoortheen: so closing it then. I am collapsing as I finish each. Please feel free to check and raise any queries.

xonsh/tokenize.py
jnoortheen: I have replaced the func with statements.

jnoortheen:
@scopatz I have done the changes. Let me know if all these are clear.

ci/environment-3.6.yml
scopatz: Is there anything we can do here to turn this into some sort of valid range?

scent.py
```diff
@@ -0,0 +1,50 @@
...
+
+        print("")
+        print("$ %s" % ' '.join(command))
```
scopatz, line 37: A format string is preferred here. They are faster

xonsh/proc.py
```diff
@@ -2356,8 +2356,6 @@ def returncode(self):
             return 1
         return self.proc.returncode
 
-    rtn = returncode
```
scopatz, line 2359: Please don't remove this API

ci/environment-3.6.yml
jnoortheen: it is a development dependency. do we really need to make it flexible ? we can make range like < 0.8 or something. 
but it doesn't follow semantic versioning.

scent.py
jnoortheen: done

xonsh/proc.py
jnoortheen: there is already a property with the same name. that was error from mypy

requirements/tests.txt
scopatz: My concern is that I don't know enough about mypy to update this number, if that makes sense. Maybe there is someone else on @xonsh/xore or you can reassure me about maintaining it?

jnoortheen: it is already part of CI where it was mentioning the version. Also here we have a conversation on using poetry for this project. 
https://github.com/xonsh/xonsh/issues/3825

then it will be very much easier. Also I assure you that I will help you maintaining the versions.

xonsh/proc.py
jnoortheen: ?

gforsyth:
Question: if this is merged in, is the assumption that moving forward all contributions should also contain type annotations?

jnoortheen:
it doesn't necessarily have to annotate. It needs to pass the mypy errors. like flake it is a linter. you can always disable complex cases.

jnoortheen:
type checking helps developers not restricts them
https://sunscrapers.com/blog/python-best-practices-static-typing-in-python-with-mypy/

xonsh/proc.py
scopatz: Oh you are right, thanks!

scopatz:
In my timing testing, this does seem to be fairly negligible in terms of headless startup time:

```sh
# master
base scopatz@artemis ~ $ timeit! xonsh -c "pass"
The slowest run took 6.54 times longer than the fastest. This could mean that an intermediate result is being cached.
100 loops, best of 3: 8 ms per loop
base scopatz@artemis ~ $ timeit! xonsh -c "pass"
100 loops, best of 3: 7.98 ms per loop

# 40d3e41c74896a19e6beddad1f36523fbf7ba78b
base scopatz@artemis ~ $ timeit! xonsh -c "pass"
100 loops, best of 3: 8.35 ms per loop
base scopatz@artemis ~ $ timeit! xonsh -c "pass"
100 loops, best of 3: 8.49 ms per loop
base scopatz@artemis ~ $ timeit! xonsh -c "pass"
100 loops, best of 3: 7.72 ms per loop
```

I am ok with this going in.  How does the rest of @xonsh/xore feel? @gforsyth?

gforsyth:
I think that it already found a redundant api def is a point in its favor and @jnoortheen has addressed all of my concerns.

So long as @jnoortheen is prepared to be patient with the rest of us as we come up to speed I'm ok with this going in :grin:

scopatz:
@jnoortheen let us know when this is ready for another look!

jnoortheen:
It is ready to be merged. I have rebased on master recently. Other than the pinning versions for black and mypy, other things are resolved I think. I also have answered why we need to pin them. Please let me know if you need to clarify.

xonsh/dirstack.py
```diff
@@ -1,19 +1,20 @@
...
+import os
 import subprocess
+import typing as tp
 
-from xonsh.lazyasd import lazyobject
-from xonsh.tools import get_sep
 from xonsh.events import events
+from xonsh.lazyasd import lazyobject
 from xonsh.platform import ON_WINDOWS
+from xonsh.tools import get_sep
```
scopatz, line 13: Please don't reorder imports in the future

scopatz:
Thanks a ton @jnoortheen!

scopatz:
I know this was a pretty huge effort!

xonsh/dirstack.py
jnoortheen: Ok noted.

scopatz: Thanks!

PR #3828, opened 2020-10-02
Path complete given path literal

deeuu:
Hey,

Thanks for all your great work - this is my first PR to `xonsh`, so let's see how I do ;-)

This is a PR to enable path completion of path literals, which without can be very frustrating
```
path = p"some/super/duper/long/path/to/comple<TAB> # :-(
```

I've added some minor modifications such that the `p` prefix is removed from the string passed to `ast.literal_eval` in `xonsh.completers._path_from_partial_string`, and also removed from the start of the path used by `xonsh.completers.complete_path`.

Please note that I don't have a deep understanding of these functions, but have implemented a few tests.

Let me know what you think.

Many thanks

### DIFF
--- a/xonsh/completers/path.py
+++ b/xonsh/completers/path.py
@@ -32,15 +32,19 @@ def cd_in_command(line):
             break
     return have_cd
 
+def _pstring_quote_length(s):
+    """Purposely excludes 'fp' and 'pf' strings"""
+    for pre in ('p', 'pr', 'rp'):
+        for q in ('"', "'"):
+            if s.startswith(f"{pre}{q}"):
+                return len(pre)
+    return 0
 
 def _path_from_partial_string(inp, pos=None):
     if pos is None:
         pos = len(inp)
     partial = inp[:pos]
     startix, endix, quote = xt.check_for_partial_string(partial)
-    # if pathlib literal, strip leading `p`
-    if quote in ('p"', "p'"):
-        startix += 1
     _post = ""
     if startix is None:
         return None
@@ -54,6 +58,13 @@ def _path_from_partial_string(inp, pos=None):
             else:
                 return None
         string = partial[startix:endix]
+
+    # If 'pr'/'rp', treat as raw string, otherwise strip leading 'p'
+    pstring_len = _pstring_quote_length(quote)
+    if pstring_len == 2:
+        string = f"r{string[2:]}"
+    elif pstring_len == 1:
+        string = string.lstrip('p') # explicit is better
 
     end = xt.RE_STRING_START.sub("", quote)
     _string = string
@@ -283,8 +294,11 @@ def complete_path(prefix, line, start, end, ctx, cdpath=True, filtfunc=None):
     lprefix = len(prefix)
     if p is not None:
         lprefix = len(p[0])
+        # Compensate for 'p' if p-string variant
+        if _pstring_quote_length(p[2]) > 0:
+            lprefix += 1
         prefix = p[1]
-        path_str_start = p[2].lstrip("p")
+        path_str_start = p[2]
         path_str_end = p[3]
         if len(line) >= end + 1 and line[end] == path_str_end:
             append_end = False
--- a/xonsh/completers/path.py
+++ b/xonsh/completers/path.py
@@ -32,13 +32,14 @@ def cd_in_command(line):
             break
     return have_cd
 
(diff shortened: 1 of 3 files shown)


xonsh/completers/path.py
```diff
@@ -280,7 +284,7 @@ def complete_path(prefix, line, start, end, ctx, cdpath=True, filtfunc=None):
     if p is not None:
         lprefix = len(p[0])
         prefix = p[1]
-        path_str_start = p[2]
+        path_str_start = p[2].lstrip("p")
```
scopatz, line 287: I think this might do the wrong thing for `fp""` strings

scopatz:
Thanks @deeuu! This is an awesome feature to have!  I have the one question about how this interacts with p-strings when combined with other prefixes, like f-strings.  

Any other thoughts on this @gforsyth?

gforsyth:
> Any other thoughts on this @gforsyth?

This looks good!  Thanks for putting this in @deeuu -- I have hit this particular issue myself so I'll enjoy having this in. As @scopatz mentioned, the only additional thing to cover here is that it is possible to have `pf"` or `fp"` (and the equivalent with single quotes).

I don't think that will complicate the prefix-stripping logic too much -- I think you'll want to check `if p in quote` and then adjust your strip index by the right number of characters.  Let me know if you hit any bumps!

xonsh/completers/path.py
deeuu: Nothing changes for the case when the quote is `pf"` or `fp"`, as `_path_from_partial_string` returns `None` for these cases. More than happy to add a check before stripping though.

deeuu:
Thanks both. Path completion will simply not work for `fp""` or `pf""` strings (see my response above).

Do we want to take things a step further and evaluate expresssions inside pf-strings, e.g.

```
a = "very"
b = "nice"
path = pf"~/{a}/{b}<TAB>
```
Happy to give this a stab if you wouldn't mind show me how to parse/evaluate pf-strings in xonsh!

scopatz:
> Path completion will simply not work for

Yeah, agreed. we don't want to start doing evaluation like that during completion because it could have too many side effects.  I think the concern here is just that `pf` strings were not being skipped properly in the logic.  Hope that helps clear it up!

laloch:
Hi folks, what's the status? I'd like the feature to get merged.

deeuu:
Hey, 

Hoping to finish this tomorrow, completion tests are currently failing on Windows so I need to investigate (I'm not on Windows but have created a VM for xonsh dev  :smile: )

deeuu:
I've made some improvements and path completion now works with path strings (`p""`) and raw path strings (`pr""`).

Unfortunately, the path completion test for this [is failing on Windows](https://github.com/xonsh/xonsh/blob/685ab1ba30b73e6bf2056b49a74f89adf8e0bb57/tests/test_path_completers.py#L55), but I can't reproduce on Windows 10 64-bit. Any ideas?

I suspect it's to an issue with backslashes in the temp-file name, but thought I'd gotten around that by [using a raw path string in the prefix](https://github.com/xonsh/xonsh/blob/685ab1ba30b73e6bf2056b49a74f89adf8e0bb57/tests/test_path_completers.py#L55) (at least that worked on my Windows VM!).

scopatz:
Very cool! Thanks for plugging away at this @dyuri!

scopatz:
FWIW, these windows errors seem like legitimate errors

deeuu:
After further digging by abusing the Windows CI, it looks like [`_iglobpath`](https://github.com/xonsh/xonsh/blob/master/xonsh/tools.py#L2460) is returning an empty set after globbing a case expanded path such as `C:\Users\RUNNER~1\AppData\Local\Temp\tmpcuo54ym8*` -> `C:\[Uu][Ss]...`.

@scopatz No need for a review (sorry for the notificatiion), but any ideas?

deeuu:
Okay, so Windows file system is case-insensitive (see [this discussion](https://bugs.python.org/issue26655) on `pathlib` `glob`), so running the test with `$CASE_SENSITIVE_COMPLETIONS=True` resolves the issue.

gforsyth:
Thanks for sticking with this through some thorny testing, @deeuu !

gforsyth:
And congrats on your first PR, and welcome!

PR #3845, opened 2020-10-06
Extend the history command with a "transfer" subcommand

cryzed:
Extend the history command with a "transfer" subcommand to transfer data between history backends

Extended history backend interface's `all_items()` method with a `full_item: bool` keyword-argument, that would yield a tuple of sessionid and the full "cmd"-dict.

Sample usage:

- Transfers the dummy entry into a new JSON history session: `xonsh history transfer --source dummy json`
- Transfers all JSON history entries into the dummy: `xonsh history transfer --source json dummy`
- Transfers from the configured default source (`$XONSH_HISTORY_BACKEND` or "json") to the target "sqlite": `xonsh history transfer sqlite`

`--source-filename` and `--target-filename` are available to potentially override the default value passed into the `filename` keyword-argument for the History implementation, in case the user has specified non-standard log locations.

There are no test for this currently, nor have I adjusted any of the existing tests for the `full_items()` change to the History interface yet. I would like to hear your thoughts on this first and maybe tips on what tests to touch. When and if everything looks ok, I will also add the news entry.

Original discussion here: https://gitter.im/xonsh/xonsh?at=5f79f6446e0eb8446965538b

### DIFF
--- a/xonsh/history/dummy.py
+++ b/xonsh/history/dummy.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 """Implements the xonsh history backend."""
 import collections
+import uuid
+
 from xonsh.history.base import History
 
 
--- a/xonsh/history/dummy.py
+++ b/xonsh/history/dummy.py
@@ -11,10 +11,15 @@ class DummyHistory(History):
         pass
 
     def items(self, newest_first=False):
-        yield {"inp": "dummy in action", "ts": 1464652800, "ind": 0}
+        yield {"inp": "dummy in action", "ts": 1464652800, "ind": 0, "rtn": 0}
 
-    def all_items(self, newest_first=False):
-        return self.items(newest_first=newest_first)
+    def all_items(self, newest_first=False, full_item=False):
+        if full_item:
+            sessionid = self.sessionid or uuid.uuid4()
+            for item in self.items(newest_first=newest_first):
+                yield sessionid, item
+        else:
+            yield from self.items(newest_first=newest_first)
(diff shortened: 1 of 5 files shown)


xonsh/history/dummy.py
```diff
@@ -11,10 +11,15 @@ def append(self, cmd):
...
 
     def items(self, newest_first=False):
-        yield {"inp": "dummy in action", "ts": 1464652800, "ind": 0}
+        yield {"inp": "dummy in action", "ts": 1464652800, "ind": 0, "rtn": 0}
 
-    def all_items(self, newest_first=False):
-        return self.items(newest_first=newest_first)
+    def all_items(self, newest_first=False, full_item=False):
+        if full_item:
+            sessionid = self.sessionid or uuid.uuid4()
```
gforsyth, line 18: `uuid` isn't imported, which is causing the `flake` error

gforsyth:
Thanks for putting this in @cryzed !  This looks pretty cool to me.  We'll want to have some tests to make sure it's working as expected.

I can also envision (not for this PR, just forward looking) having a markdown or an org-mode "backend" that would be super handy for writing out documentation / notes with example commands + their outputs

daniel-shimon:
Hi @cryzed !
Could you add some tests for this and resolve the comments so we can put this in?

jnoortheen:
Thanks for creating this. The history-alias now use functions from `cli_utils.py` . I've create another PR implementing this https://github.com/xonsh/xonsh/pull/4604

PR #3848, opened 2020-10-07
Tutorial: added short example

anki-code:
Hi! xonsh is the best of the best of the best! Thanks!

I read articles about xonsh and almost every article has comments from people about xonsh subproc determination - i.e. how fun that `ls -l` could becomes `ls - l` Python-mode variables.

I added the short example at the "determination" section of the tutorial to give the information about strict syntax.


anki-code:
@gforsyth @laloch @melund please help Anthony and merge this simple PR without his attention.

anki-code:
@scopatz if you prepare release and all core is sleeping please merge this ))

docs/tutorial.rst
```diff
@@ -315,10 +315,16 @@ The determination between the two modes is done well ahead of any execution.
 You do not need to worry about partially executed commands - that is
 impossible.
 
-If you absolutely want to run a subprocess command, you can always
-force xonsh to do so with the syntax that we will see in the following
-sections.
+If you still have concerns about the determination process and want to have strong confidence
+when running a subprocess command in your xonsh scripts, you can always use the elegant subprocess
+mode syntax that we will see in the following sections. Short example now:
```
scopatz, line 320: The english here should be cleaned up a little

anki-code: www.grammarly.com told me that
* Grammar - Correct
* Punctuation - Correct
No thoughts what you mean :)
Please fix, you have edit rights.

scopatz: Yeah, the syntax is all correct. It is that the word choice and idioms feel a little off to me. Sorry, but I can barely keep up with all of the other issues.  I have to let other people help.

anki-code: @gforsyth Gil, please help with idioms ;)

anki-code:
Fresh example where @scopatz explains that there is a way to set strict syntax - https://news.ycombinator.com/item?id=24744653

docs/tutorial.rst
scopatz: How about something like:

```
If you would like to explicitly run a subprocess command, you can always
use the formal xonsh subprocess syntax that we will see in the following sections. For example:
```

gforsyth, line 320: Maybe
```suggestion
If you know that you want to run a command in subprocess mode, you can force subprocess execution by using the following syntax:
```

anki-code: Thanks! I've chosen the scopatz version. It has more strict text.

anki-code: done

scopatz:
Thanks @anki-code!

PR #3850, opened 2020-10-07
removes validation on getting environment variables. 

scopatz:
Should only be used for setting

No news item needed, because no release has happened. 

CC @xonsh/xore

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1968,7 +1968,6 @@ class Env(cabc.MutableMapping):
     #
 
     def __getitem__(self, key):
-        # remove this block on next release
         if key is Ellipsis:
             return self
         elif key in self._d:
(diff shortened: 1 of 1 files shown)


scopatz:
This is ready for review!

anki-code:
After this PR the great case will not work:
```
$MY_DIR = '/tmp'
 ${...}.register('MY_DIR', type='path')
    if $MY_DIR.exists(): # here will be string variable instead of path and there will be an exception
        pass
```
I suggest to add at least this code to this PR 

https://github.com/xonsh/xonsh/blob/5363f4ffe06dd8d831fea6a28967f41feab7cd82/xonsh/environ.py#L2161-L2174

scopatz:
@anki-code - I think we should have that discussion in a separate issue / PR. As I have said elsewhere I think what you are proposing should be its own API and not be hidden or happen automatically in other operations. This is because it is likely to have unforeseen consequences by users.

anki-code:
Okay

scopatz:
CC @xonsh/xore - this is ready for review and is blocking the next release

gforsyth:
Thanks, @scopatz !

PR #3854, opened 2020-10-07
Prevent loging git when building prompt

mariusvniekerk:
Set GIT_OPTIONAL_LOCKS=0 to prevent locking a git repo when updating the prompt.

### DIFF
--- a/news/prevent-git-locking-when-getting-status.rst
+++ b/news/prevent-git-locking-when-getting-status.rst
@@ -6,9 +6,8 @@
 
 * When retrieving the git status or other fields for building the prompt xonsh will run
   the git commands with ``$GIT_OPTIONAL_LOCKS=0``.  For details on what this entails see
-  the git documentation for GIT_OPTIONAL_LOCKS_.
-
-.. _GIT_OPTIONAL_LOCKS:   https://git-scm.com/docs/git#Documentation/git.txt-codeGITOPTIONALLOCKScode
+  the git documentation for
+  `GIT_OPTIONAL_LOCKS <https://git-scm.com/docs/git#Documentation/git.txt-codeGITOPTIONALLOCKScode/>`_.
 
 **Deprecated:**
 
--- a/news/prevent-git-locking-when-getting-status.rst
+++ b/news/prevent-git-locking-when-getting-status.rst
@@ -1,0 +1,27 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* When retrieving the git status or other fields for building the prompt xonsh will run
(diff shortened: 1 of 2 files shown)


news/prevent-git-locking-when-getting-status.rst
```diff
@@ -0,0 +1,27 @@
...
+
+* <news item>
+
+**Changed:**
+
+* When retrieving the git status or other fields for building the prompt xonsh will run
+  the git commands with ``$GIT_OPTIONAL_LOCKS=0``.  For details on what this entails see
+  the git documentation for GIT_OPTIONAL_LOCKS_.
+
+.. _GIT_OPTIONAL_LOCKS:   https://git-scm.com/docs/git#Documentation/git.txt-codeGITOPTIONALLOCKScode
```
scopatz, line 11: probably should make this a direct link

xonsh/prompt/vc.py
```diff
@@ -20,9 +20,13 @@
     "RE_REMOVE_ANSI",
 )
 
+# when running git status commands we do not want to acquire locks running command like git status
+_GIT_OPTIONAL_LOCKS = {"GIT_OPTIONAL_LOCKS": "0"}
+
 
 def _get_git_branch(q):
     denv = builtins.__xonsh__.env.detype()
+    denv.update(_GIT_OPTIONAL_LOCKS)
```
scopatz, line 29: You shouldn't modify the detyped env directly. This object can be shared accidientally.  It would be better to wrap the `_get_git_branch()` call in an `with builtins.__xonsh__.env.swap(GIT_OPTIONAL_LOCKS=0):` context

xonsh/prompt/vc.py
```diff
@@ -20,9 +20,14 @@
     "RE_REMOVE_ANSI",
 )
 
+# when running git status commands we do not want to acquire locks running command like git status
+_GIT_OPTIONAL_LOCKS = {"GIT_OPTIONAL_LOCKS": "0"}
```
scopatz, line 24: Can you remove this as a global variable? Many people will never touch this if they are not using `curr_branch`

xonsh/prompt/vc.py
```diff
@@ -163,9 +168,11 @@ def current_branch():
     return branch or None
 
 
-def _get_exit_code(cmd):
+def _get_exit_code(cmd, additional_env=None):
```
scopatz, line 171: It seems like you can just update the denv and not need to add an argument to this function

scopatz: That is, all call sites of `_get_exit_code` seem to pass in the same env here

scopatz:
Thanks!

PR #3855, opened 2020-10-07
Add an optional message to the deprecation warning

cafehaine:
This commit improves the current deprecation warning for NO_COLOR by allowing to show the traceback if XONSH_SHOW_TRACEBACK is set to True.
See https://github.com/xonsh/xonsh/pull/3793#issuecomment-703290727 for reasoning.

It also adds a new xonsh tool function: print_warning.

It is similar to print_exception, however it requires a message, as there is no exception.

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -911,6 +911,51 @@ def suggest_commands(cmd, env, aliases):
     return rtn
 
 
+def print_warning(msg):
+    """Print warnings with/without traceback."""
+    env = getattr(builtins.__xonsh__, "env", None)
+    # flags indicating whether the traceback options have been manually set
+    if env is None:
+        env = os_environ
+        manually_set_trace = "XONSH_SHOW_TRACEBACK" in env
+        manually_set_logfile = "XONSH_TRACEBACK_LOGFILE" in env
+    else:
+        manually_set_trace = env.is_manually_set("XONSH_SHOW_TRACEBACK")
+        manually_set_logfile = env.is_manually_set("XONSH_TRACEBACK_LOGFILE")
+    if (not manually_set_trace) and (not manually_set_logfile):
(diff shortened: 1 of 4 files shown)


xonsh/tools.py
```diff
@@ -911,6 +911,51 @@ def suggest_commands(cmd, env, aliases):
     return rtn
 
 
+def print_warning(msg):
```
scopatz, line 914: There seems to be a lot of overlap in the logic between this and `print_exception()`. Could you maybe add some private functions or similar to reduce the code duplication?

scopatz:
A test or two for print_warning would be great too! Thanks @cafehaine!

xonsh/tools.py
cafehaine: Fixed in cdb2514

cafehaine:
> A test or two for print_warning would be great too! Thanks @cafehaine!

I'm not sure how I could implement those tests. The existing `print_exception` does not have test, and it seems quite complicated to create tests for my `_get_env_var` function as it uses the os environment as a fallback

scopatz:
>  The existing print_exception does not have any test

Yeah that is fair.

xonsh/tools.py
```diff
@@ -911,25 +911,66 @@ def suggest_commands(cmd, env, aliases):
     return rtn
 
 
-def print_exception(msg=None):
-    """Print exceptions with/without traceback."""
+def _get_env_var(name, default=None):
```
scopatz, line 914: maybe include something in the function name here about being manually set to make it more clear

scopatz: ie `_get_manual_env_var`

cafehaine: Fixed in a5b0993

scopatz:
Thanks!

PR #3867, opened 2020-10-09
Remove import random

anki-code:
Closes #3264

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -127,9 +127,12 @@ def _expandpath(path):
     return expand_path(path, expand_user=expand_user)
 
 
-def random_choice(lst):
-    """Returns random element from the list"""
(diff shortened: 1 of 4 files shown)


xonsh/tools.py
```diff
@@ -127,6 +127,11 @@ def _expandpath(path):
     return expand_path(path, expand_user=expand_user)
 
 
+def random_choice(lst):
+    """Returns random element from the list"""
+    return lst[datetime.datetime.now().microsecond % len(lst)]
```
scopatz, line 132: What if `len(lst)` is greater than the size of microsecond?  It seems like there should be some kind of warning about this.

anki-code: done

scopatz:
Thanks!

PR #3870, opened 2020-10-10
bashisms set -ex

### LINKED ISSUE #3294: Support "set -e" and "set -x"
I use `set -ex` in all my Bash scripts. It's easy to remember and short to type. In Xonsh the equivalent is:
```
$RAISE_SUBPROC_ERROR = True
trace on
```
Which is long and hard to remember for me. I think it would be great if Xonsh could support `set -e` as a shortcut for `$RAISE_SUBPROC_ERROR = True` and `set -x` as a shortcut for `trace on`, and also support combining them as `set -ex`.

anki-code:
Closes #3294

### DIFF
--- a/xontrib/bashisms.py
+++ b/xontrib/bashisms.py
@@ -105,6 +105,12 @@ def _set(args):
         __xonsh__.env["XONSH_TRACE_SUBPROC"] = True
     elif arg == "+x":
         __xonsh__.env["XONSH_TRACE_SUBPROC"] = False
+    elif arg in ["-ex", "-xe"]:
+        __xonsh__.env["RAISE_SUBPROC_ERROR"] = True
+        __xonsh__.env["XONSH_TRACE_SUBPROC"] = True
+    elif arg in ["+ex", "+xe"]:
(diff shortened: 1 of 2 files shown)


xontrib/bashisms.py
```diff
@@ -105,6 +105,12 @@ def _set(args):
         __xonsh__.env["XONSH_TRACE_SUBPROC"] = True
     elif arg == "+x":
         __xonsh__.env["XONSH_TRACE_SUBPROC"] = False
+    elif arg in ["-ex", "-xe"]:
+        __xonsh__.env["RAISE_SUBPROC_ERROR"] = True
+        __xonsh__.env["XONSH_TRACE_SUBPROC"] = True
+    elif arg in ["+ex", "+xe"]:
+        __xonsh__.env["RAISE_SUBPROC_ERROR"] = False
+        __xonsh__.env["XONSH_TRACE_SUBPROC"] = False
```
scopatz, line 113: At some point it will make more sense to implement this with argparse, or a similar command line parsing interface

scopatz: I think that time is now :wink:

PR #3872, opened 2020-10-10
fix: async prompt workers update cache at the end

jnoortheen:
this causes the getter to return every one in two times.
run 1-> the prompt is built the cache is cleared , but the async worker
updates the cache
run 2 -> the value is get from cache and no async worker created for
that field. cache is cleared after the prompt is built


scopatz:
Thanks!

Issue #3234, opened 2019-07-22
v1.0 Tracking Issue


scopatz:
@xonsh/xore, all, This is the great xonsh v1.0 tracking issue. The following items (in no particular order) are considered blockers for v1.0. Please feel free to add to this as needed. This list was developed at the very successful SciPy 2019 sprints, but of course, not everyone was there.

* [x] Job Control
* [ ] Write a manual
* [x] Make a landing page for the website, #3670 https://github.com/xonsh/landing-page-dev
* [ ] Reduce amount CLI wizard displays
* [x] Web xonfig application (like fish), #3226
* [ ] Develop governance documentation, including a community site for xontirbs
* [x] Environment registration / deregistration API, #3227 
* [x] Strip out `history replay` (done as part of #3377 )
* [ ] Remove threadable subproc notion

Topics pushed off to 2.0 include:

* Async execution interface for subprocesses. 
* Moving Popen capabilties to another library (slug)

As always, help for any of these points is extremely welcome!

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

gforsyth:
I propose we also strip out `history replay`.  Votes in favor / against?

scopatz:
I am in favor of removing `history replay`

scopatz:
or at least moving it to a xontrib or something

astronouth7303:
I should really revisit slug. I still feel like having an actual posix-model process library (with virtual process support) would help a lot of the issues. Remove foreground/background weirdness, let us do proper posix session foreground/background process groups, etc.

scopatz:
Yeah, I agree. I think slug is closer to a 2.0 issue at this point though.  I think it is important for us to get a "stable" v1.0 out for users, even though there are a lot of improvements we would all want to see in v1.0

scopatz:
@dotsdl - where is that branch that you were working on for environment variable registration?

scopatz:
Are we in a "good enough" spot for job control do we think?

gforsyth:
> Are we in a "good enough" spot for job control do we think?

I think we need to figure out if any xonsh users use job control at all frequently.  Maybe we can ping some folks on some of the previous job control issues and see if things are working?

gforsyth:
although, on further reflection, maybe we just say we're good enough for now.

scopatz:
[...]

PR #3873, opened 2020-10-11
chore (ci): improve ci workflows and deps management

### LINKED ISSUE #3825: github workflow refactoring

jnoortheen:
1. use black from local version in both pre-commit and ci checks
~~2. use poetry to manage dependencies~~
3. use conda to install python version and use pip to install python dependencies
4. cache path set as per current os - https://github.com/actions/cache/blob/master/examples.md#windows-1

resolves #3825


pyproject.toml
```diff
@@ -34,3 +34,55 @@ exclude = '''
      )/
 )
 '''
+
+# these four lines are needed by poetry to work
+[tool.poetry]
+name = "xonsh"
+version = "0.0.0" # check the setup.py file
```
scopatz, line 41: Does this need to be updated by rever?

pyproject.toml
```diff
@@ -34,3 +34,55 @@ exclude = '''
...
+description = ""
+authors = []
+
+[tool.poetry.dependencies]
+python = ">=3.6.1,<4.0"
+
+[tool.poetry.dev-dependencies]
+# tests deps
+pytest = "^6.1.1"
```
scopatz, line 50: Do most of these things need pinnings? It seems to me like they don't

pyproject.toml
jnoortheen: no need. This is just a dummy value.

pyproject.toml
jnoortheen: it is the convention to have these pinned. everytime `poetry update` it updates the lockfile to be exactly the same. You could check the projects using poetry. Most of the time it is important to have a correct set of working versions.

scopatz: Sure it just seems to be a little odd to have them pinned so tightly without knowing that the patch numbers are actually required.  Also, I'd like to avoid caret and tilde pins since conda does not understand them. (Also I personally find them harder to read.)  For example, this seems like it should be `>=6` or `>=6,<7`

scopatz: I guess maybe I don't fully understand what poetry --export does, either

jnoortheen: It would be better if you could give it a try. Just going through its quickstart. The best part about using it having the same environment in all the systems.

jnoortheen: > I guess maybe I don't fully understand what poetry --export does, either

it returns the dependencies in the lock file in requirements.txt format. I have used it in the rever.xsh 
to retain the original functionality.

AstraLuma:
Why poetry vs pipenv or pip tools?

xonsh uses setuptools for its packaging, which is redundant with poetry.

jnoortheen:
issues I had when I was developing 
1. wrong black version installed that made ci tests fail
2. when adding new test/lint dependency I had to edit tests.txt as well as environment.yml

what using poetry would give for the project
1. version and dependency management. same across local and ci environments
2. virtualenv management. so `poetry run flake8` will work from pre-commit or git-cola or any other gui tool 
3. adding new dependencies to the project will be easier.

what it is not for / not my intention
1. replace setup.py (I don't think it is possible. since it has a built step and poetry/flit dont support build step)
2. not specifying the extra dependencies - ptk, pygments etc., maybe this can be changed. we can always read from toml file and pass it to the setup.py . but that is not for now; until those extra dependencies become complex or large in number

pyproject.toml
scopatz: Ok yeah revert was the big concern

jnoortheen:
@scopatz does this need any change? do you want to abandon this or continue?

laloch:
This may sound stupid, but I don't like the notion of solving a problem with dependencies by introducing a dependency.
@scopatz, what about a feature freeze? The project does not seem to me to be stabilizing towards v1.0; quite the contrary. We are playing with colors while the whole thing feels to be more and more fragile.

laloch:
...and btw, the `poetry.lock` file is just crazy. Who is going to maintain that?

jnoortheen:
> This may sound stupid, but I don't like the notion of solving a problem with dependencies by introducing a dependency.
> @scopatz, what about a feature freeze? The project does not seem to me to be stabilizing towards v1.0; quite the contrary. We are playing with colors while the whole thing feels to be more and more fragile.

If you are talking about poetry attaining 1.0, then it is already past that. now it is version 1.1 and it is stable. I think you haven't tried it in a long time. I think you are one of those old school folks, avoiding all shiny things (colors ...) :smile: 

I just want to remove the hassle of updating requirements.txt and environment.yml .

laloch:
I'm talking about Xonsh v1.0 and yes, I am one of **those** old school folks :wink:

jnoortheen:
@laloch Oh I see, you meant xonsh. I wonder about that too. I think that is the problem with the test suit and code quality. It is one such move to improve that. We can add more linters and tools to project without hassle. 

> I'm talking about Xonsh v1.0 and yes, I am one of **those** old school folks
it is left to poetry cli. It is a generated file, to keep the environments have same versions installed.

gforsyth (changes requested):
I agree that we can improve our CI, and I think unifying pre-commit with CI makes a bunch of sense.
Also, as per my comment re: black as a separate CI job, we should probably at least move the flake checks to before the rest of the test suite (or to a separate precommit job).

I think the justifications for using poetry here don't really resonate with me.  Admittedly, python packaging is a bit of a mess, and xonsh is slightly messier, but I think this just trades one set of annoyances for another.

If we need to bump a dependency version (or add a new one), we have to change it in the `pyproject.toml` and `setup.py`.  

If this were getting rid of all the reqs files, and the `setup.cfg`, and the `setup.py`, that would be a little more compelling to me.  

There's also an implicit (or maybe explicit) command here that all those who would develop xonsh or build docs for it need to use `poetry` and that seems a bit overly prescriptive to me.

run-tests.xsh
```diff
@@ -33,3 +33,6 @@ python -m flake8
 echo "---------- Running mypy ----------"
 mypy --version
 mypy xonsh
+
+echo "---------- Check Black formatter -----------"
+black --check xonsh xontrib
```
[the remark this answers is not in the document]
jnoortheen: Yes I understand. I may have overlooked it a bit. I will create separate job for black and move lint to it as well.

rever.xsh
```diff
@@ -28,14 +28,10 @@ $TAG_TARGET = 'master'
...
-with open('requirements/docs.txt') as f:
-    conda_deps += f.read().split()
-for delimiter in '=<>':
-    conda_deps = {d.lower().split(delimiter)[0] for d in conda_deps}
-conda_deps.discard('prompt-toolkit')
-conda_deps |= {'prompt_toolkit', 'pip', 'psutil', 'numpy', 'matplotlib'}
+
+requirements = $(poetry export -f requirements.txt --without-hashes --dev --extras docs)
+
+conda_deps = {req.lower().split("==")[0] for req in requirements.splitlines()}
```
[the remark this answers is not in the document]
jnoortheen: is there anyway to use pip after activating the conda environment with rever ? 
is this used to create xonsh docker image?

jnoortheen:
> I agree that we can improve our CI, and I think unifying pre-commit with CI makes a bunch of sense.
> Also, as per my comment re: black as a separate CI job, we should probably at least move the flake checks to before the rest of the test suite (or to a separate precommit job).
> 
> I think the justifications for using poetry here don't really resonate with me. Admittedly, python packaging is a bit of a mess, and xonsh is slightly messier, but I think this just trades one set of annoyances for another.
> 
> If we need to bump a dependency version (or add a new one), we have to change it in the `pyproject.toml` and `setup.py`.
> 
> If this were getting rid of all the reqs files, and the `setup.cfg`, and the `setup.py`, that would be a little more compelling to me.
> 
> There's also an implicit (or maybe explicit) command here that all those who would develop xonsh or build docs for it need to use `poetry` and that seems a bit overly prescriptive to me.


Yes I understand. I will change this PR to keep requirements.txt as it is. 

Can environment.yml atleast be removed ? to prevent segmentation. I will try to make requirements.txt as the only source of truth

ci/environment.yml
```diff
@@ -0,0 +1,5 @@
+name: xonsh-test
+channels:
+  - conda-forge
+  - defaults
+dependencies:
```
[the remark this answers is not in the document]
jnoortheen: I'd probably remove that too. It is not used.

gforsyth:
> Yes I understand. I will change this PR to keep requirements.txt as it is.
> 
> Can environment.yml atleast be removed ? to prevent segmentation. I will try to make requirements.txt as the only source of truth

Thanks for all the work here @jnoortheen!  Yes, I think we can go ahead and remove the `environment.yml` file

scopatz:
FWIW - here is my take on the dev experience: 

* Deps can come from anywhere (conda, poetry, etc)
* development install should be done with `pip install --no-deps .`

> @scopatz, what about a feature freeze? The project does not seem to me to be stabilizing towards v1.0; quite the contrary. We are playing with colors while the whole thing feels to be more and more fragile.

@laloch - yeah, I understand that it seems that way. I think what is going on (to some degree) is an interest and willingness to make needed/wanted changes before a 1.0 declaration.  The color stuff (which were all 3rd party contributions) fixed some long standing issues.  `NO_COLOR` was a terrible name, because the underscore made it really annoying to parse full color names. The custom styles thing has been a desired feature for years now, so I am glad someone got a chance to tackle it.

Personally, I have come fully around to your opinion that the job control / thread subproc issues need to be dealt with before v1.0, and so that is what I have been working on. I would really love to remove the notion of threadable subprocs and threadable predictors entirely. It has proven just too difficult to do correctly. I think I know how to do this now (basically with [a PTY server](https://github.com/infodox/python-pty-shells)) on Linux. 

I also think that everytime a new version of Python comes out (and esp. with v3.9) there is a sense that things don't work perfectly. I believe this is unavoidable, mostly because Python does crazy things like change their AST in the rc phase of releases. This makes it super hard to actually target the next version before it comes out. Thanks for all of your help with the parser! It is a huge help, truly, because it lets me focus on the subproc issues.

In any event, I am totally open to discussing more about 1.0 and stabilization. Maybe we could move further conversation over to #3234?

jnoortheen:
@scopatz for now I don't want to disrupt the dev experience. So I kept requirements files. It could be sometime before the community agrees on a single format. Let me know if this PR needs any rework.

.github/workflows/pytest.tmpl
```diff
@@ -18,15 +18,26 @@ jobs:
     name: Python ${{ matrix.python-version }} ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/conda_pkgs_dir
+            ~/.cache/pypoetry
```
gforsyth, line 25: Is this still required?

.github/workflows/pytest.tmpl
```diff
@@ -18,15 +18,26 @@ jobs:
...
+          activate-environment: xonsh-test
           update-conda: true
-          python-version: ${{ matrix.python-version }}
+          python-version: ${{ matrix.python-version }}  # this itself makes sure that Python version is installed
           condarc-file: ci/condarc.yml
+      - uses: Gr1N/setup-poetry@v4
       - shell: bash -l {0}
         run: |
+          pip --version
+          pip install -r requirements/tests.txt
```
gforsyth, line 41: Nit, but always better to do an explicit 
```suggestion
          python -m pip install -r requirements/tests.txt
```
to avoid cross-env installs

.github/workflows/pytest.tmpl
```diff
@@ -18,15 +18,26 @@ jobs:
...
         uses: conda-incubator/setup-miniconda@v1
         with:
-          activate-environment: ${{matrix.python-version}}-xonsh-test
-          environment-file: ci/environment-${{matrix.python-version}}.yml
+          activate-environment: xonsh-test
           update-conda: true
-          python-version: ${{ matrix.python-version }}
+          python-version: ${{ matrix.python-version }}  # this itself makes sure that Python version is installed
           condarc-file: ci/condarc.yml
+      - uses: Gr1N/setup-poetry@v4
```
gforsyth, line 37: Do we still need this?

.github/workflows/pytest.tmpl
```diff
@@ -18,15 +18,26 @@ jobs:
     name: Python ${{ matrix.python-version }} ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/conda_pkgs_dir
+            ~/.cache/pypoetry
+            ~/miniconda*/envs/
+          key: ${{ runner.os }}-${{ matrix.python-version }}-env-${{ hashFiles('ci/environment.yml') }}-${{ hashFiles('poetry.lock') }}
```
gforsyth, line 27: I think this key needs to updated if `environment.yml` and `poetry.lock` files aren't present

gforsyth (changes requested):
The caching stuff here looks really nice @jnoortheen -- and i like the qa carveout for lint, black, mypy checks, that's a nice solution.

Since we aren't going with poetry for the moment, there is some stuff to pull out of the github action template file.

Thanks!

.github/workflows/pytest.tmpl
jnoortheen: nope. removed.

.github/workflows/pytest.tmpl
jnoortheen: nope. removed.

jnoortheen:
> @scopatz for now I don't want to disrupt the dev experience. So I kept requirements files. It could be sometime before the community agrees on a single format. Let me know if this PR needs any rework.

@scopatz

gforsyth (approved):
This is great @jnoortheen !  Thanks for all the hard work (and much iterative editing) to put this together!

Issue #3843, opened 2020-10-06
Completion dialog sometimes not shown when filenames with space exist (prompt_toolkit)


ju-kreber:
## xonfig

<details>

```
$ xonfig
+------------------+-----------------+
| xonsh            | 0.9.22          |
| Python           | 3.7.7           |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | 3.0.5           |
| shell type       | prompt_toolkit  |
| pygments         | 2.6.1           |
| on posix         | True            |
| on linux         | True            |
| distro           | unknown         |
| on darwin        | False           |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
+------------------+-----------------+

```

</details>

## Expected Behavior

I'm in a directory with multiple files where at least one of them contains a space.
I start entering a command like `file ` and press tab.
A dialog should open, listing all the files in the directory (quoted) for completion.

## Current Behavior

_Sometimes_ no dialog with completion suggestions is shown.
Only the first completion is chosen and pressing tab iterates through all completions (possibly some fallback behavior?).

_Edit:_ Include a recording of the issue (with `ls` instead of `file`):
![xonsh_completion_bug](https://user-images.githubusercontent.com/8841729/95678438-e0f02300-0bcc-11eb-8961-36cc6b3f08e7.gif)

## Steps to Reproduce

When exactly this happens is really weird, but I found out a way to reproduce this issue (at least it works relatively consistently for me).
Create said directory with a bunch of files where at least one them them contains spaces.
Typing `file <tab>` now shows the undesired behavior.
**To fix it** simply complete something where suggestions do not contain spaces, like `file /<tab>` (usually) of simply `fi<tab>`. The dialogue is shown as expected and also the original `file <tab>` lists the quoted files as expected now.
**To break it** again just press enter a few times. This really did it for me. It doesn't matter if you execute a command or enter a completely empty line, after a few new lines, `file <tab>` behaves unexpectedly again.

I really have no idea what causes this and hope it's not limited to my case only. I tried to have a look at the internal workings of prompt_toolkit, but I'm too unfamiliar with it to figure something out, it seems :/

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

scopatz:
[...]

PR #3815, opened 2020-09-30
Space for menu

bobhy:
A bunch of layout changes in prompt toolkit shell.
* Ensure completion menu displays a minimum of $MENU_COMPLETIONS_ROWS, even at bottom of screen (superseding doomed hack of #3778)
* Allow bottom toolbar to be cleared if set to something visible earlier in the session (#3810)
* Display right prompt at *top* of prompt, not at bottom of screen (#3810)

Fixes currently require changes in prompt-toolkit: see pending PR from prompt-toolkit/python-prompt-toolkit#1259.
However, this PR won't crash if run with published PTK.

@jnoortheen, please give this a careful look, My changes overlapped with yours, but I think I reconciled them all happily.
@anki-code, would also appreciate you giving this a try.

PR #3874, opened 2020-10-11
Completion menu rows - clean up completion menu display

bobhy:
Supercedes #3815, but doesn't include all the fixes yet.  When done, intended to address #3775 and #3384 
Also requires an as-yet-unreleased PR, prompt-toolkit/python-prompt-toolkit#1271.

@anki-code, @ju-kreber, this should address the height-of-the-completion-menu issue as is.  You'll need to install the above version of prompt-toolkit, but can then set $COMPLETION_MENU_ROWS to a reasonable positive number and get no more than that number of completion rows wherever the prompt is in the screen.  There are some other display fixes that should look better, as well..  But no changes yet for anything to do with completions when file name contains spaces, I wanted to get this out of the way first.

And if it's working for you, we can jointly lobby @jonathanslenders  to accept the ptk PR.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -847,7 +847,7 @@ def DEFAULT_VARS():
             int,
             str,
             5,
-            "Number of rows to reserve for tab-completions menu if "
+            "Maximum number of rows to display for tab-completions menu if "
             "``$COMPLETIONS_DISPLAY`` is ``single`` or ``multi``. This only affects the "
             "prompt-toolkit shell.",
         ),
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -851,6 +851,14 @@ def DEFAULT_VARS():
             "``$COMPLETIONS_DISPLAY`` is ``single`` or ``multi``. This only affects the "
             "prompt-toolkit shell.",
         ),
+        "RESERVE_SPACE_FOR_MENU": Var(
+            is_int,
+            int,
+            str,
+            DefaultNotGiven,
+            "Additional space to reserve at bottom of screen for tab-completions menu. "
+            "Only affects prompt-toolkit shell.  Deprecated.  See ``$COMPLETIONS_MENU_ROWS``.",
+        ),
         "COMPLETION_QUERY_LIMIT": Var(
             is_int,
             int,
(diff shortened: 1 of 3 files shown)


scopatz:
Thanks for putting this in @bobhy!

bobhy:
Fixed backward compatibility: setting $RESERVE_SPACE_FOR_MENU` is used if explicitly set (overrides `$COMPLETION_MENU_ROWS`).  Resulting value is fed to `PromptSession.prompt()` parameter `completion_menu_rows` if it has it, otherwise to parameter `reserve_space_for_menu`.

bobhy:
This PR could now be merged, it will work with the released version of prompt-toolkit (though it will not work any *better* with that version than the master branch does).  It really only adds value if you side-load the unreleased prompt-toolkit/python-prompt-toolkit#1271.  I'm continuing to work on a xonsh-only version of this, but that involves pulling in pieces of PTK into Xonsh so we can tweak its behavior.  Slow going!  Alternatively, we could put the changes in the vended version of PTK, but that would constitute a fork, and I'd recommend against it.

scopatz:
I would love it if some other member of @xore could find some time to review this.

xonsh/environ.py
```diff
@@ -847,10 +847,18 @@ def DEFAULT_VARS():
...
             "prompt-toolkit shell.",
         ),
+        "RESERVE_SPACE_FOR_MENU": Var(
+            is_int,
+            int,
+            str,
+            DefaultNotGiven,
+            "Additional space to reserve at bottom of screen for tab-completions menu. "
+            "Only affects prompt-toolkit shell.  Deprecated.  See ``$COMPLETIONS_MENU_ROWS``.",
+        ),
```
gforsyth, line 861: I'm deeply confused -- this adds a new env-var that is already deprecated?

laloch: @gforsyth, I'm going to stick that quote on my fridge :))

ju-kreber:
With the prompt-toolkit fix, this works for me and successfully fixes issue #3843.

jnoortheen:
maybe late to the show. I was annoyed with this behaviour as well. https://github.com/xonsh/xonsh/pull/4477 is simple but seems to  fix the issue for me. @bobhy Can you test the same?

jnoortheen:
I am closing this to clean-up as the said issues were closed by #4477 .

PR #3881, opened 2020-10-13
Document variable escaping in bash_to_xonsh

willsALMANJ:
This method was already documented in the tutorial, but it took me a while to find it there. I thought it would be a good addition to the bash to xonsh guide as well.


scopatz:
Thanks!

Issue #3875, opened 2020-10-11
TypeError: 'Call' object is not subscriptable


jaraco:
After copying the `.xonshrc` from my local macOS workstation to a Raspberry Pi Buster with Python 3.9 installed from source, I get the reported error when starting xonsh.

What can I do to troubleshoot?

## xonfig

<details>

```
$ xonfig
+------------------+-----------------+
| xonsh            | 0.9.23          |
| Python           | 3.9.0           |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | 3.0.7           |
| shell type       | prompt_toolkit  |
| pygments         | None            |
| on posix         | True            |
| on linux         | True            |
| distro           | unknown         |
| on darwin        | False           |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
| on jupyter       | False           |
| jupyter kernel   | None            |
+------------------+-----------------+
```

</details>

## Expected Behavior
`xonsh` should start up without an error or provide an error that's indicative of the issue.

## Current Behavior

I tried bisecting the error and found that if I remove this code and everything after it, the error goes away:

```
def get_repo_url():
	raw = $(git remote get-url --push origin).rstrip()
	return raw.replace('https://github.com/', '')
```

If I remove only that function, the error changes slightly to:

```
TypeError: 'Name' object is not subscriptable
```

The few lines that precede the `get_repo_url` definition look like this:

```
aliases.update(gclip='pbcopy', pclip='pbpaste')

# workaround for https://bugs.python.org/issue22490
${...}.pop('__PYVENV_LAUNCHER__', None)
```

Removing those lines has no effect on the error.

I suspect I've encountered this before, but I don't recall how I fixed it.

### Traceback (if applicable)

<details>

```
jaraco@Katsushiro ~ $ env XONSH_DEBUG=1 XONSH_SHOW_TRACEBACK=1 .local/bin/xonsh
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/home/jaraco/.local/pipx/venvs/xonsh/lib/python3.9/site-packages/xonsh/environ.py", line 2240, in xonsh_script_run_control
    run_script_with_cache(filename, execer, ctx)
[...]

PR #3883, opened 2020-10-13
get_repo_url() test

scopatz:
Test for issue in #3875


scopatz:
Just a test right now, no news item needed

scopatz:
Ready for review @xonsh/xore

BYK (approved):

laloch:
I don't think, that this actually tests for the bug/regression reported in #3875. @jaraco reports a failure even with this function definition removed.

scopatz:
Hmm yeah.  It might not.  Seems like it is a good test to have though anyway

laloch:
OK, let's have it then :wink:
Thanks.

PR #3884, opened 2020-10-14
Fix f-string parsing in Python 3.9

### LINKED ISSUE #3865: f-string with env vars raises syntxc error in 0.9.23 under Python 3.9
f-strings with env vars in expressions give an error. When you type into the repl

    f"hallo {$PATH}" 

you get an error:

_SyntaxError: <xonsh-code>:1:0: f-string: invalid syntax_

### xonfig

<details>

```
| xonsh            | 0.9.23          |
| Python           | 3.9.0           |

```

</details>

laloch:
The f-strings parser is very sensitive to changes in the textual representation of the exceptions raised by the Python parser. There seems to be a bug introduced in Python 3.9 causing the exception text to be terminated by an extra newline character.

Fixes #3865

### DIFF
--- a/news/fix-v39-fstrings.rst
+++ b/news/fix-v39-fstrings.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed f-strings parsing in Python 3.9
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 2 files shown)


scopatz:
Thanks @laloch!

tirkarthi:
@laloch Is the issue related to xonsh or something that happened in Python 3.9 that caused this? If you have a reproducer that explains the changes in Python 3.9 with respect to this it would be helpful to know if this is an upstream Python 3.9 bug. 

Thanks.

laloch:
@tirkarthi yeah, it's an upstream change that made `text` field of `SyntaxError` exception object newline terminated. It does not make much sense, but I'm not going to file a bug report just bacause of that.

tirkarthi:
Thanks @laloch

PR #3886, opened 2020-10-14
Rever: pin AppImage python version

anki-code:
Pin AppImage python until full support of python 3.9.

Update Rever before merge: https://github.com/regro/rever/pull/235


scopatz:
Thanks!

PR #3893, opened 2020-10-17
docs: fix typos

omjadas:
Fix typos on index page. Don't think this really deserves a news item, but let me know if you want me to do one up.


scopatz:
Thanks! You are right, no news item needed!

Issue #3864, opened 2020-10-09
Sourcing ~/.bash_aliases throws off highlighting of ls and cd


raghavmallampalli:
## xonfig
NOTE: I'm on Ubuntu 18.04 on WSL2, Windows 10. Maybe run a [WSL check](https://github.com/Microsoft/WSL/issues/423#issuecomment-221627364) while outputting xonfig?
<details>
```
$ xonfig
+------------------+-----------------+
| xonsh            | 0.9.22          |
| Python           | 3.6.9           |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | 3.0.7           |
| shell type       | prompt_toolkit  |
| pygments         | 2.7.1           |
| on posix         | True            |
| on linux         | True            |
| distro           | ubuntu          |
| on darwin        | False           |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
+------------------+-----------------+
```

</details>

## Expected Behavior

Sourcing `~/.bash_aliases` file should convert all aliases and make them available without affecting colors of the shell.

## Current Behavior

![image](https://user-images.githubusercontent.com/49435792/95557836-ff221b80-0a32-11eb-8184-df726c475db8.png)
`cd` is similarly affected.
## Steps to Reproduce

Install xonsh using pip3
Run xonfig web that results in the following `.xonshrc`:
```sh
# XONSH WEBCONFIG START
$PROMPT = '[{localtime}] {YELLOW}{env_name} {BOLD_BLUE}{user}@{hostname} {BOLD_GREEN}{cwd} {gitstatus}{NO_COLOR}\n> '
$XONSH_COLOR_STYLE = 'fruity'
xontrib load abbrevs apt_tabcomplete bashisms
# XONSH WEBCONFIG END
```
source the following `.bash_aliases` file:
```sh
alias maxvol="pactl set-sink-volume @DEFAULT_SINK@ 150%"
alias update="sudo apt update && sudo apt full-upgrade -y && sudo apt autoremove -y"
alias download="wget --random-wait -r -p --no-parent -e robots=off -U mozilla"
alias server="ifconfig | grep inet && python3 -m http.server"
alias la='ls -la --color'
[...]

PR #3894, opened 2020-10-17
xonfig xontribs

anki-code:
_Originally posted by @scopatz in https://github.com/xonsh/xonsh/issues/3864#issuecomment-710591574_:
> We really need to have `xonfig` print out what xontribs have been loaded.

### DIFF
--- a/xonsh/readline_shell.py
+++ b/xonsh/readline_shell.py
@@ -29,12 +29,7 @@ from xonsh.ansi_colors import (
     ansi_color_style,
 )
 from xonsh.prompt.base import multiline_prompt
-from xonsh.tools import (
-    print_exception,
-    to_bool,
-    columnize,
-    carriage_return,
-)
+from xonsh.tools import print_exception, to_bool, columnize, carriage_return
 from xonsh.platform import (
     ON_WINDOWS,
     ON_CYGWIN,
--- a/xonsh/xonfig.py
+++ b/xonsh/xonfig.py
@@ -478,7 +478,7 @@ def _xonfig_format_human(data):
     wcol1 = wcol2 = 0
     for key, val in data:
         wcol1 = max(wcol1, len(key))
-        if type(val) == list:
+        if isinstance(val, list):
             for subval in val:
                 wcol2 = max(wcol2, len(str(subval)))
         else:
@@ -487,7 +487,7 @@ def _xonfig_format_human(data):
     row = "| {key!s:<{wcol1}} | {val!s:<{wcol2}} |\n"
     s = hr
     for key, val in data:
-        if type(val) == list and val:
+        if isinstance(val, list) and val:
             for i, subval in enumerate(val):
                 s += row.format(
                     key=f"{key} {i+1}", wcol1=wcol1, val=subval, wcol2=wcol2
--- a/xonsh/xonfig.py
+++ b/xonsh/xonfig.py
@@ -548,7 +548,7 @@ def _info(ns):
     data.extend([("on jupyter", jup_ksm is not None), ("jupyter kernel", jup_kernel)])
 
     xontribs = [s.split(".")[1] for s in sys.modules if s.startswith("xontrib.")]
-    data.extend([("xontrib", xontribs)])
+    data.extend([("xontrib", xontribs if xontribs else "")])
 
     formatter = _xonfig_format_json if ns.json else _xonfig_format_human
     s = formatter(data)
--- a/xonsh/xonfig.py
+++ b/xonsh/xonfig.py
@@ -537,6 +537,10 @@ def _info(ns):
         pass
     data.extend([("on jupyter", jup_ksm is not None), ("jupyter kernel", jup_kernel)])
 
+    if "_xontribs" in builtins.__xonsh__.ctx:
+        xontribs = ", ".join(builtins.__xonsh__.ctx["_xontribs"])
+        data.extend([("xontribs", xontribs)])
+
     formatter = _xonfig_format_json if ns.json else _xonfig_format_human
     s = formatter(data)
     return s
--- a/xonsh/xonfig.py
+++ b/xonsh/xonfig.py
@@ -488,10 +488,13 @@ def _xonfig_format_human(data):
     s = hr
     for key, val in data:
         if type(val) == list:
-            for i, subval in enumerate(val):
-                s += row.format(
-                    key=f"{key} {i+1}", wcol1=wcol1, val=subval, wcol2=wcol2
-                )
+            if val == []:
+                s += row.format(key=key, wcol1=wcol1, val=val, wcol2=wcol2)
+            else:
+                for i, subval in enumerate(val):
+                    s += row.format(
+                        key=f"{key} {i+1}", wcol1=wcol1, val=subval, wcol2=wcol2
+                    )
         else:
             s += row.format(key=key, wcol1=wcol1, val=val, wcol2=wcol2)
     s += hr
@@ -548,7 +551,7 @@ def _info(ns):
     data.extend([("on jupyter", jup_ksm is not None), ("jupyter kernel", jup_kernel)])
 
     xontribs = [s.split(".")[1] for s in sys.modules if s.startswith("xontrib.")]
-    data.extend([("xontrib", xontribs if xontribs else "")])
+    data.extend([("xontrib", xontribs)])
 
     formatter = _xonfig_format_json if ns.json else _xonfig_format_human
     s = formatter(data)
--- a/xonsh/xonfig.py
+++ b/xonsh/xonfig.py
@@ -478,12 +478,22 @@ def _xonfig_format_human(data):
     wcol1 = wcol2 = 0
     for key, val in data:
         wcol1 = max(wcol1, len(key))
-        wcol2 = max(wcol2, len(str(val)))
+        if type(val) == list:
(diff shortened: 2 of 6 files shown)


gforsyth (changes requested):
Thanks for taking this on, @anki-code ! 

This isn't working on my end -- I don't think that `xontribs` necessarily add the `_xontribs` key to `__xonsh__.ctx` on load.  I think the more robust check is to follow the logic in `xonsh.xontribs._list` which checks if a given `spec.name in sys.modules`.

Also, I think the `xontrib` section of `xonfig` should still be displayed even if no xontribs are loaded (we want to disambiguate a user report where the version of xonsh doesn't _include_ the xontrib info vs. one where the user has no xontribs loaded)

anki-code:
@gforsyth logic changes.

Example with xontribs:

<details>

```
$ xonfig
+------------------+--------------------------------------------+
| xonsh            | 0.9.24.dev337                              |
| Git SHA          | dd87a57c                                   |
| Commit Date      | Oct 17 18:31:44 2020                       |
| Python           | 3.8.2                                      |
| PLY              | 3.11                                       |
| have readline    | True                                       |
| prompt toolkit   | 3.0.7                                      |
| shell type       | prompt_toolkit                             |
| pygments         | 2.6.1                                      |
| on posix         | True                                       |
| on linux         | True                                       |
| distro           | ubuntu                                     |
| on darwin        | False                                      |
| on windows       | False                                      |
| on cygwin        | False                                      |
| on msys2         | False                                      |
| is superuser     | False                                      |
| default encoding | utf-8                                      |
| xonsh encoding   | utf-8                                      |
| encoding errors  | surrogateescape                            |
| on jupyter       | True                                       |
| jupyter kernel   | /opt/miniconda/share/jupyter/kernels/xonsh |
| xontrib 1        | prompt_bar                                 |
| xontrib 2        | output_search                              |
| xontrib 3        | whole_word_jumping                         |
| xontrib 4        | pipeliner_parallel                         |
| xontrib 5        | pipeliner                                  |
| xontrib 6        | argcomplete                                |
| xontrib 7        | onepath                                    |
| xontrib 8        | sh                                         |
| xontrib 9        | back2dir                                   |
| xontrib 10       | autojump                                   |
| xontrib 11       | conda                                      |
+------------------+--------------------------------------------+
```
</details>

Example without xontribs:
<details>

```
$ xonfig
+------------------+--------------------------------------------+
| xonsh            | 0.9.24.dev337                              |
| Git SHA          | dd87a57c                                   |
| Commit Date      | Oct 17 18:31:44 2020                       |
| Python           | 3.8.2                                      |
| PLY              | 3.11                                       |
| have readline    | True                                       |
| prompt toolkit   | 3.0.7                                      |
| shell type       | prompt_toolkit                             |
| pygments         | 2.6.1                                      |
| on posix         | True                                       |
| on linux         | True                                       |
| distro           | ubuntu                                     |
| on darwin        | False                                      |
| on windows       | False                                      |
| on cygwin        | False                                      |
| on msys2         | False                                      |
| is superuser     | False                                      |
| default encoding | utf-8                                      |
| xonsh encoding   | utf-8                                      |
| encoding errors  | surrogateescape                            |
| on jupyter       | True                                       |
| jupyter kernel   | /opt/miniconda/share/jupyter/kernels/xonsh |
| xontrib          |                                            |
+------------------+--------------------------------------------+
```

</details>

xonsh/readline_shell.py
```diff
@@ -29,12 +29,7 @@
     ansi_color_style,
 )
 from xonsh.prompt.base import multiline_prompt
-from xonsh.tools import (
-    print_exception,
-    to_bool,
-    columnize,
-    carriage_return,
-)
+from xonsh.tools import print_exception, to_bool, columnize, carriage_return
```
laloch, line 32: Is this unrelated change intentional?

anki-code: No, it's mr. Black.

xonsh/xonfig.py
```diff
@@ -537,6 +550,9 @@ def _info(ns):
         pass
     data.extend([("on jupyter", jup_ksm is not None), ("jupyter kernel", jup_kernel)])
 
+    xontribs = [s.split(".")[1] for s in sys.modules if s.startswith("xontrib.")]
```
scopatz, line 553: I think there as API for getting the xontrib list. That would be good to use here

xonsh/xonfig.py
```diff
@@ -478,12 +478,22 @@ def _xonfig_format_human(data):
...
+            for subval in val:
+                wcol2 = max(wcol2, len(str(subval)))
+        else:
+            wcol2 = max(wcol2, len(str(val)))
     hr = "+" + ("-" * (wcol1 + 2)) + "+" + ("-" * (wcol2 + 2)) + "+\n"
     row = "| {key!s:<{wcol1}} | {val!s:<{wcol2}} |\n"
     s = hr
     for key, val in data:
-        s += row.format(key=key, wcol1=wcol1, val=val, wcol2=wcol2)
+        if type(val) == list:
```
gforsyth:
```suggestion
        if isinstance(val, list):
```

xonsh/xonfig.py
```diff
@@ -478,12 +478,22 @@ def _xonfig_format_human(data):
     wcol1 = wcol2 = 0
     for key, val in data:
         wcol1 = max(wcol1, len(key))
-        wcol2 = max(wcol2, len(str(val)))
+        if type(val) == list:
```
gforsyth, line 481:
```suggestion
        if isinstance(val, list):
```

anki-code: Changed. Thanks!

xonsh/xonfig.py
anki-code: Changed. Thanks!

xonsh/xonfig.py
anki-code: Added + test.

xonsh/xontribs.py
```diff
@@ -132,6 +132,11 @@ def xontribs_load(names, verbose=False):
     return res
 
 
+def xontribs_loaded():
+    """Returns list of loaded xontribs."""
+    return [s.split(".")[1] for s in sys.modules if s.startswith("xontrib.")]
```
gforsyth, line 137: I would rather this not iterate over the entirety of `sys.modules`

maybe something like

```
meta = xontrib_metadata()
for md in meta["xontribs"]:
    [spec.name in sys.modules for spec in find_xontrib(md["name"])
```

anki-code: My key idea is to show all xontribs. Not only that described in xontrib file.

anki-code: I also tried to use `importlib` and `pkgutil` for example:
```
import pkgutil
for importer, modname, ispkg in pkgutil.iter_modules(importlib.util.find_spec("xontrib").submodule_search_locati
ons):
     print("Found submodule %s (is a package: %s)" % (modname, ispkg))
```
But it is not working for xontribs with dash in the name. I.e. `readable-traceback`.

scopatz: Sure, but I think maybe xontribs that want to be found should add themselves to `xontribs_metadata`.  Maybe we should add a registration function for that.

anki-code: > xontribs that want to be found should add themselves

This is somewhat contrary to the goal:

> we want to disambiguate a user report where the version of xonsh doesn't include the xontrib info

If user played with some xontribs from github or try to write his own xontrib and foreget about it and then he breaked the xonsh it should run `xonfig` and remeber about his experiments.

scopatz: Yeah, but I think we want to be fast and stable before being fully complete. This is a piece of debugging code, so it needs to work even when other parts of xonsh/python are falling down. Does that make sense?

scopatz: Also, I think that once modules are loaded once, they are added to the metadata, so that `xontrib list` will show them

anki-code: We just need a simplest way to get the list of loaded xontribs. If `importlib` can't help because of missing the xontribs with dash in the name and `xontribs` module can't help because of using only list of known xontribs, I see the one way is to using `sys.modules` or similar. If you know the more optimal way to avoid `sys.modules` please point it out.

anki-code: @gforsyth there is a way to read files in `xontrib` directory instead of iterating `sys.modules`:
```
[x.name.split('.')[0] for x in Path(importlib.util.find_spec("xontrib").submodule_search_locations[0]).glob('*')
]
```
What do you think?

scopatz: Yeah, I think this is better. If you look at what [list does below](https://github.com/xonsh/xonsh/pull/3894/files#diff-0108dd9b8af7697497623c2f97c08b67536c0d7e4d593b5ce3e9656928854663R156-R161), it basically does a 

```
spec = find_xontrib(name)
loaded = spec.name in sys.modules
```

I would actually recommend splitting out the whole first part of the `_list` function into its own top level function for getting the `data` list.  Maybe `xontrib_data()`?  This function could then be reused in `xonfig`

anki-code: Done

gforsyth (approved):
Thanks for all your hard work on this, @anki-code !  This looks good to me.

PR #3901, opened 2020-10-18
Jedi xontrib: xsh to py for pyc

anki-code:


tests/xontribs/test_jedi.py
```diff
@@ -40,11 +40,7 @@ def jedi_xontrib(monkeypatch, source_path, jedi_mock, completer_mock):
 
 
 def test_completer_added(jedi_xontrib, completer_mock):
-    assert completer_mock.call_args_list == [
-        call(["remove", "python_mode"]),
-        call(["add", "jedi_python", "complete_jedi", "<python"]),
-        call(["remove", "python"]),
-    ]
+    assert 'jedi_python' in __xonsh__.completers
```
scopatz, line 43: This test seems to be a less strong check now

xontrib/jedi.py
```diff
@@ -138,7 +126,10 @@ def create_completion(comp):
 xonsh.completers.base.complete_python = complete_jedi
 
 # Jedi ignores leading '@(' and friends
-completer remove python_mode
+if "python_mode" in __xonsh__.completers:
+    del __xonsh__.completers["python_mode"]
```
scopatz, line 130: This should go through the API. Same below

anki-code:
Thanks for comments. I'm not using jedi now and just tried to make a quick fix. If it required more tests and another approach I think the best way is to leave it for someone else.

scopatz:
Sounds good! Thanks for trying the quick fix!

Issue #3882, opened 2020-10-13
Exception at start in uptime code on Haiku


extrowerk:
xonsh prints some errors at startup:
```

Welcome to the Haiku shell.

~> uname -a
Haiku shredder 1 hrev54608 Sep 30 2020 06:33:40 x86_64 x86_64 Haiku
~> uptime
 19:33:59  up   1:00,  0 users
~> xonsh
Exception in thread Thread-2:
Traceback (most recent call last):
  File "/packages/python3-3.7.7-2/.self/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/packages/python3-3.7.7-2/.self/lib/python3.7/vendor-packages/xonsh/history/__amalgam__.py", line 379, in run
    files = self.files(only_unlocked=True)
  File "/packages/python3-3.7.7-2/.self/lib/python3.7/vendor-packages/xonsh/history/__amalgam__.py", line 422, in files
    boot = uptime.boottime()
  File "/packages/python3-3.7.7-2/.self/lib/python3.7/vendor-packages/xonsh/xoreutils/uptime.py", line 280, in boottime
    up = uptime()
  File "/packages/python3-3.7.7-2/.self/lib/python3.7/vendor-packages/xonsh/xoreutils/uptime.py", line 265, in uptime
    or _uptime_linux()
  File "/packages/python3-3.7.7-2/.self/lib/python3.7/vendor-packages/xonsh/xoreutils/uptime.py", line 56, in _uptime_linux
    if xp.LIBC.sysinfo(buf) < 0:
AttributeError: 'NoneType' object has no attribute 'sysinfo'


                     Welcome to the xonsh shell (0.9.23)                      

         ~ Piggy glanced nervously into hell and cradled the xonsh ~          

------------------------------------------------------------------------------
xonfig tutorial    ->    Launch the tutorial in the browser
xonfig web         ->    Run the configuration tool in the browser and claim your shell 
(Note: Run the configuration tool or create a ~/.xonshrc file to suppress the welcome screen)

user@shredder ~ #      
```

## xonfig

<details>

```
user@shredder ~ # xonfig                                                      
+------------------+-----------------+
| xonsh            | 0.9.23          |
| Python           | 3.7.7           |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | None            |
| shell type       | prompt_toolkit  |
| pygments         | 2.4.2           |
| on posix         | True            |
| on linux         | False           |
[...]

PR #3904, opened 2020-10-20
Add uname support

Hierosme:
Hello,

Here the ``uname`` command line support with every tests, it take the actual ``xonsh-cat`` as base exemple .

As i know ``xonsh-cat`` don't provide a functional auto completion support, then ``uname`` too.

Regards

### DIFF
--- a/news/add-uname-command.rst
+++ b/news/add-uname-command.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Add command line ``uname``
+
+**Changed:**
+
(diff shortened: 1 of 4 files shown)


Hierosme:
After a share on #3882 issue, it look the uname command shoul dbe base on ``platform`` module and not ``os`` ...

I close the pull request and resend a new one with the migration to  ``platform`` module.

scopatz:
Ok thanks!

PR #3906, opened 2020-10-20
Add uname command line

Hierosme:
Here the support of ``uname`` normaly cross platform GNU/Linux, Haiku

### DIFF
--- a/tests/xoreutils/test_uname.py
+++ b/tests/xoreutils/test_uname.py
@@ -1,0 +1,220 @@
+import io
+import os
+import sys
+import pytest
+
+from xonsh.xoreutils import uname
+from xonsh.platform import DEFAULT_ENCODING
+
+
+@pytest.fixture
+def uname_env_fixture(xonsh_builtins):
+    with xonsh_builtins.__xonsh__.env.swap(
+            XONSH_ENCODING=DEFAULT_ENCODING, XONSH_ENCODING_ERRORS="surrogateescape"
+    ):
+        yield xonsh_builtins
+
+
+class TestUname:
+    def setup(self):
+        if hasattr(self, 'stdout'):
+            self.stdout.flush()
+        if hasattr(self, 'stdin'):
+            self.stdin.flush()
+        if hasattr(self, 'stderr'):
+            self.stderr.flush()
+        self.stdin = io.StringIO()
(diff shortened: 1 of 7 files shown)


xonsh/xoreutils/uname.py
```diff
@@ -0,0 +1,161 @@
+import platform
```
scopatz, line 1: Please add a docstring to the module

xonsh/xoreutils/uname.py
```diff
@@ -0,0 +1,161 @@
...
+import sys
+from xonsh.xoreutils.util import arg_handler
+
+__version__ = '0.1'
```
scopatz, line 5: I don't think this makes sense to version separately from xonsh.

xonsh/xoreutils/uname.py
```diff
@@ -0,0 +1,161 @@
...
+__name__ = 'uname'
```
scopatz, line 6: Please don't overwrite name

scopatz:
Thanks for putting this in @Hierosme!

xonsh/xoreutils/uname.py
Hierosme: that is because the original uname have --version what is it suppose to return ?

xonsh/xoreutils/uname.py
Hierosme: Will be done

tests/xoreutils/test_uname.py
```diff
@@ -0,0 +1,240 @@
...
+            sys.platform
+        ), DEFAULT_ENCODING)
+
+    # def test_uname_unknown(self, uname_env_fixture):
+    #     uname.uname(['--hello'], self.stdin, self.stdout, self.stderr)
+    #     self.stdout.flush()
+    #
+    #     assert self.stdout_buf.getvalue() == bytes('{0}\n'.format(
+    #         ""
+    #     ), DEFAULT_ENCODING)
```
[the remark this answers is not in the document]
Hierosme: Ok

xonsh/xoreutils/uname.py
Hierosme: My bad

PR #3908, opened 2020-10-20
Proc dep warnings

scopatz:
Some deprecation help for downstream code


scopatz:
Ready for review @xonsh/xore

xonsh/proc.py
```diff
@@ -0,0 +1,30 @@
+"""xonsh.proc is deprecated"""
+from xonsh.lazyasd import lazyobject
+from xonsh.tools import print_warning
+
+
+_WARNINGS_PRINTED = {}
+
+
+def _print_proc_warning(msg):
+    global _WARNINGS_PRINTED
```
daniel-shimon, line 10: Not needed (no assignment)

xonsh/proc.py
```diff
@@ -0,0 +1,30 @@
...
+    if msg not in _WARNINGS_PRINTED:
+        print_warning(msg)
+        _WARNINGS_PRINTED[msg] = True
```
daniel-shimon, line 13: Use a set instead

daniel-shimon (changes requested):

xonsh/proc.py
scopatz: That is technically correct, but I think this is good defensive coding / future-proofing.

scopatz:
Thanks for the review @daniel-shimon! I think this is ready for review again @xonsh/xore (and is blocking a release)

gforsyth (approved):
looks good -- sorry for the delay @scopatz, I am buried in notifications

Issue #3895, opened 2020-10-17
Windows 10 WSL2: fast from root, slow from user


anki-code:
Hi! xonsh is the best! Thanks!

I've made a try to run xonsh on Windows 10 WSL2 Ubuntu 20 and found that it works fast from root but slow from user.

Asciinema - https://asciinema.org/a/bFZb6nmH7bZSWq57VkQCfB0LQ

Screenshots:

<details>

Run from user:

![image](https://user-images.githubusercontent.com/1708680/96334426-7da53b80-1079-11eb-82b1-92fae97e1a49.png)

![image](https://user-images.githubusercontent.com/1708680/96334447-96adec80-1079-11eb-8af8-b2e42e8af5a3.png)

Run from root:

![image](https://user-images.githubusercontent.com/1708680/96334463-aaf1e980-1079-11eb-9d43-6f62443dd5c6.png)

</details>

In the Linux and WSL1 xonsh works perfect for me.

May be it's WSL2 problem. Any thought how to debug/fix it are very welcome.

Thanks!

scopatz:
Thanks for reporting @anki-code! Can you start xonsh with the `--timings` flag, ie `xonsh --timings` for both WSL1 & WSL2 so that we can start tracking down where the slowdown is coming from? Thanks!

anki-code:
Timing wasn't help. But I dive into the code and found the cause.

In WSL2 user has very rich PATH (by default!) with many directories with huge amount of files i.e. `/mnt/c/Windows/system32` with many subdirectories.

But root user has empty PATH (by default).

As result the call stack looks like:

1. PTK do `prompt_formatter` with `{env_name}{BOLD_GREEN}{user}@{hostname}{BOLD_BLUE} {cwd}{branch_color}{curr_branch: {}}{RESET} {BOLD_BLUE}{prompt_end}{RESET}`
2. call `branch_color`
3. call `__xonsh__.commands_cache`
4. call `locate_binary`
5. call `__xonsh__.commands_cache.all_commands` that get ALL 18,222 (!) files from PATH directories and it takes around 5-10 seconds.

scopatz:
Oh wow! That is bad... Maybe on WSL2 you can set the PATH to be something reasonable in the xonshrc

anki-code:
The solution - https://github.com/microsoft/WSL/issues/1493#issuecomment-580643734

scopatz:
[...]

PR #3911, opened 2020-10-20
Added warning about huge amount of files in PATH directories

anki-code:
Related to #3895

_Originally posted by @scopatz in https://github.com/xonsh/xonsh/issues/3895#issuecomment-713080742_
> It would be nice to detect if this was happening and issue a warning with the fix then.

I think we shouldn't support the solution and I added only warning.

### DIFF
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -101,15 +101,14 @@ class CommandsCache(cabc.Mapping):
         if cache_valid:
             return self._cmds_cache
         allcmds = {}
-        cnt = 0
         for path in reversed(path_immut):
             # iterate backwards so that entries at the front of PATH overwrite
             # entries at the back.
             for cmd in executables_in(path):
-                cnt += 1
                 key = cmd.upper() if ON_WINDOWS else cmd
                 allcmds[key] = (os.path.join(path, cmd), alss.get(key, None))
 
+        cnt = len(allcmds)
         if cnt > builtins.__xonsh__.env.get("XONSH_WARNING_PATH_FILES", 10000):
             print(
                 f"Warning! Found {cnt:,} executable files in the PATH directories!\n",
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -110,7 +110,7 @@ class CommandsCache(cabc.Mapping):
                 key = cmd.upper() if ON_WINDOWS else cmd
                 allcmds[key] = (os.path.join(path, cmd), alss.get(key, None))
 
-        if cnt > builtins.__xonsh__.env.get("XONSH_WARNING_PATH_FILES"):
+        if cnt > builtins.__xonsh__.env.get("XONSH_WARNING_PATH_FILES", 10000):
             print(
                 f"Warning! Found {cnt:,} executable files in the PATH directories!\n",
                 file=sys.stderr,
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -108,7 +108,7 @@ class CommandsCache(cabc.Mapping):
                 key = cmd.upper() if ON_WINDOWS else cmd
                 allcmds[key] = (os.path.join(path, cmd), alss.get(key, None))
 
-        warn_cnt = builtins.__xonsh__.env.get("XONSH_WARNING_PATH_FILES")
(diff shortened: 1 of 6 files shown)


xonsh/environ.py
```diff
@@ -1725,6 +1725,13 @@ def DEFAULT_VARS():
             "or None / the empty string if traceback logging is not desired. "
             "Logging to a file is not enabled by default.",
         ),
+        "XONSH_WARNING_PATH_FILES": Var(
+            is_int,
+            int,
+            str,
+            10000,
+            "Minimal number of files in the PATH directories to show warning.",
```
scopatz, line 1733: "Number of files on the PATH above which a warning is shown"

xonsh/commands_cache.py
```diff
@@ -100,12 +101,21 @@ def all_commands(self):
...
+        cnt = 0
         for path in reversed(path_immut):
             # iterate backwards so that entries at the front of PATH overwrite
             # entries at the back.
             for cmd in executables_in(path):
+                cnt += 1
                 key = cmd.upper() if ON_WINDOWS else cmd
                 allcmds[key] = (os.path.join(path, cmd), alss.get(key, None))
+
+        if cnt > builtins.__xonsh__.env.get("XONSH_WARNING_PATH_FILES", 10000):
```
scopatz, line 113: You don't need to have the `10000`, since the environment knows about its default values

xonsh/commands_cache.py
```diff
@@ -100,12 +101,21 @@ def all_commands(self):
         if cache_valid:
             return self._cmds_cache
         allcmds = {}
+        cnt = 0
         for path in reversed(path_immut):
             # iterate backwards so that entries at the front of PATH overwrite
             # entries at the back.
             for cmd in executables_in(path):
+                cnt += 1
```
scopatz, line 109: Rather than having a counter, you can just check the len(allcmds) after the loop.

xonsh/commands_cache.py
anki-code: But tests fails because it uses this function without environ.

scopatz: Then you should add the environment variable to the tests :wink:

anki-code: A-ha ;)
But:
```
sphinx-build -b html -d _build/doctrees   . _build/html

Running Sphinx v3.2.1

/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/cloud_sptheme/__init__.py:79: DeprecationWarning: get_theme_dir() is deprecated, and will be removed in version 2.0

  DeprecationWarning, stacklevel=1)

Exception occurred:

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/xonsh/commands_cache.py", line 112, in all_commands

    if cnt > builtins.__xonsh__.env.get("XONSH_WARNING_PATH_FILES"):

TypeError: '>' not supported between instances of 'int' and 'NoneType'

The full traceback has been saved in /tmp/sphinx-err-8ufyf6qk.log, if you want to report the issue to the developers.

Please also report this if it was a user error, so that a better error message can be provided next time.

A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!

Makefile:39: recipe for target 'html' failed
```
I think it's not good idea to put this env variable to the Makefile.

scopatz: Agreed, you should put them here if needed: https://github.com/xonsh/xonsh/blob/master/tests/tools.py#L109

anki-code: I've reverted all additions in tests because I think minor variable shouldn't be everywhere.

scopatz: OK, you should still remove the default here though

anki-code:
It's ready.

xonsh/environ.py
```diff
@@ -1725,6 +1725,13 @@ def DEFAULT_VARS():
             "or None / the empty string if traceback logging is not desired. "
             "Logging to a file is not enabled by default.",
         ),
+        "XONSH_WARNING_PATH_FILES": Var(
```
scopatz, line 1728: Maybe a name like COMMANDS_CACHE_SIZE_WARNING would be more descriptive

anki-code: renamed

scopatz:
Thanks @anki-code!

PR #3913, opened 2020-10-20
Reading the directory from PATH and the symlink to this directory from PATH only once while create CommandCache

### LINKED ISSUE #3915: CommandsCache: reading the directory from PATH and the symlink to this directory from PATH (double work)
This is the second part of #3912

While I worked on #3911 I found that reading of the files in the PATH directories was done many times. 

I found that there is reading the directory from PATH and reading the symlink to this directory from PATH again. It makes xonsh start on Linux slower.

For example details by directories on Ubuntu:
```python
all=0
for p in $PATH:
    print(p)
    a=len([l for l in fp'{p}'.glob('*')])
    all += a
    print(a)
print(all)
```
```bash
/opt/miniconda/bin
353
/opt/miniconda/condabin
1
/home/pc/.local/bin
2
/usr/local/sbin
0
/usr/local/bin
0
/usr/sbin
420
/usr/bin
2298
/sbin
420
/bin
2298
/usr/games
0
/usr/local/games
0
/snap/bin
3

5795
```
And we can see that:
```bash
ls -la /bin
#/bin -> usr/bin

ls -la /sbin
#/sbin -> /usr/sbin
```
So xonsh scan the same directory twice.

As result instead of scan 5795-2298-420=3077 files once it scans 5795 files 3 times = 17,385 in sum.

## For community

anki-code:
Closes #3915

### DIFF
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -70,16 +70,7 @@ class CommandsCache(cabc.Mapping):
     @staticmethod
     def remove_dups(p):
         ret = list()
-        for e in p:
-            if e not in ret:
-                ret.append(e)
-        return ret
-
-    @staticmethod
-    def remove_dups_path(p):
-        ret = list()
-        for e in p:
-            e = os.path.realpath(e)
+        for e in map(os.path.realpath, p):
             if e not in ret:
                 ret.append(e)
         return ret
@@ -87,10 +78,7 @@ class CommandsCache(cabc.Mapping):
     @property
     def all_commands(self):
         paths = builtins.__xonsh__.env.get("PATH", [])
-        if ON_WINDOWS:
-            paths = CommandsCache.remove_dups(paths)
-        else:
-            paths = CommandsCache.remove_dups_path(paths)
+        paths = CommandsCache.remove_dups(paths)
         path_immut = tuple(x for x in paths if os.path.isdir(x))
         # did PATH change?
         path_hash = hash(path_immut)
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -79,8 +79,7 @@ class CommandsCache(cabc.Mapping):
     def remove_dups_path(p):
         ret = list()
         for e in p:
-            rpe = os.path.realpath(e)
-            e = e if e.lower() == rpe.lower() else rpe
+            e = os.path.realpath(e)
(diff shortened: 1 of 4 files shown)


scopatz:
I don't know know that this is a good idea. A lot of downstream environment management tools (like venv, conda, etc) assume that they can add and remove PATH elements. In general, it is assumed that PATH is pretty dumb. Thoughts @xonsh/xore?

scopatz:
But thanks for the PR @anki-code!

scopatz:
Maybe there is a way to make this optional and off by default?

anki-code:
@scopatz please read the changes. I think you read only the title of the PR )

scopatz:
Oh right, thanks.  Yeah this is on a different object than I thought

xonsh/commands_cache.py
```diff
@@ -75,10 +75,20 @@ def remove_dups(p):
                 ret.append(e)
         return ret
 
+    @staticmethod
+    def remove_dups_path(p):
+        ret = list()
+        for e in p:
+            rpe = os.path.realpath(e)
+            e = e if e.lower() == rpe.lower() else rpe
```
scopatz, line 83: I don't think lowering paths is supported cross platform

scopatz: Maybe this should be a modification to `remove_dups()`. Is remove_dups used elsewhere?

anki-code: It's only for checking and this line is needed exactly to cross platform - windows support.

scopatz: The call site isn't guarded by an `if ON_WINDOWS` so this would break on Linux and other case-sensitive OSes

anki-code: Hmm... but test fails...

anki-code: OK, I avoided this.

scopatz:
@anki-code - I think that this could be a one line change to the existing function. Here is how I would write it

```py
    @staticmethod
    def remove_dups(p):
        ret = list()
        for e in map(os.path.realpath, p):
            if e not in ret:
                ret.append(e)
        return ret
```

scopatz:
This looks good to me now! Let me know when you think it is ready!

anki-code:
I think it's ready )

PR #3922, opened 2020-10-23
feat: add settings to set custom function to format prompt

jnoortheen:
Now the users can format the prompt as they like.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1194,6 +1194,16 @@ def DEFAULT_VARS():
             "NOTE: ``$UPDATE_PROMPT_ON_KEYPRESS`` must be set to ``True`` for this "
             "variable to take effect.",
         ),
+        "PROMPT_TOKENS_FORMATTER": Var(
+            validate=callable,
+            convert=None,
+            detype=None,
+            default=prompt.prompt_tokens_formatter_default,
+            doc="Final processor that receives all tokens in the prompt template. "
+            "It gives option to format the prompt with different prefix based on other tokens values. "
+            "Highly useful for implementing something like powerline theme.",
+            doc_default="``xonsh.prompt.base.prompt_tokens_formatter_default``",
+        ),
         "PROMPT_TOOLKIT_COLOR_DEPTH": Var(
             always_false,
             ptk2_color_depth_setter,
--- a/xonsh/prompt/base.py
+++ b/xonsh/prompt/base.py
@@ -7,6 +7,7 @@ import os
 import re
 import socket
 import sys
+import typing as tp
 
 import xonsh.lazyasd as xl
 import xonsh.tools as xt
@@ -29,6 +30,54 @@ def DEFAULT_PROMPT():
     return default_prompt()
 
 
+class _ParsedToken(tp.NamedTuple):
+    """It can either be a literal value alone or a field and its resultant value"""
+
+    value: str
+    field: tp.Optional[str] = None
+
+
+class ParsedTokens(tp.NamedTuple):
+    tokens: tp.List[_ParsedToken]
+    template: tp.Union[str, tp.Callable]
+
+    def process(self) -> str:
+        """Wrapper that gets formatter-function from environment and returns final prompt."""
+        processor = builtins.__xonsh__.env.get(  # type: ignore
+            "PROMPT_TOKENS_FORMATTER", prompt_tokens_formatter_default
+        )
+        return processor(self)
+
+    def update(
+        self,
+        idx: int,
+        val: tp.Optional[str],
+        spec: tp.Optional[str],
+        conv: tp.Optional[str],
+    ) -> None:
+        """Update tokens list in-place"""
+        if idx < len(self.tokens):
+            tok = self.tokens[idx]
+            self.tokens[idx] = _ParsedToken(_format_value(val, spec, conv), tok.field)
+
+
+def prompt_tokens_formatter_default(container: ParsedTokens) -> str:
+    """
+        Join the tokens
+
+    Parameters
+    ----------
+    container: ParsedTokens
+        parsed tokens holder
+
+    Returns
+    -------
+    str
+        process the tokens and finally return the prompt string
+    """
+    return "".join([tok.value for tok in container.tokens])
+
+
 class PromptFormatter:
     """Class that holds all the related prompt formatting methods,
     uses the ``PROMPT_FIELDS`` envvar (no color formatting).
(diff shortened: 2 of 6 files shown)


jnoortheen:
here is a script to implement powerline like prompt

```py
import os
from typing import NamedTuple, List
import builtins
from xonsh.prompt.base import ParsedTokens, _ParsedToken

GRAY = '#273746'
BLUE = '#004C99'
PROMPT_FIELD_COLORS = {
    "cwd": ("WHITE", "CYAN"),
    "gitstatus": ("WHITE", "BLACK"),
    "ret_code": ("WHITE", "RED"),
    "full_env_name": ("white", "green"),
    "hostname": ("white", BLUE),
    "localtime": ("#DAF7A6", "black")
}
POWERLINE_MODE = 'flame'

modes = {
    "powerline": "",
    "round": "",
    "down": "",
    "up": "",
    "flame": "",
    "squares": "",
    "ruiny": "",
    "lego": "",
}


class Section(NamedTuple):
    line: str
    fg: str = ''
    bg: str = ''


def prompt_builder(tokens: List[Section], right=False):
    size = len(tokens)
    prompt = []
    SEP, THIN, PL_RSEP, _ = modes[POWERLINE_MODE]
    for i, sec in enumerate(tokens):
        last = (i == size - 1)
        first = (i == 0)

        if right:
            prompt.append('{%s}%s{BACKGROUND_%s}{%s}%s' % (sec.bg, PL_RSEP, sec.bg, sec.fg, sec.line))
        else:
            if first:
                if sec.bg:
                    prompt.append('{BACKGROUND_%s}' % sec.bg)
            if sec.fg:
                prompt.append('{%s}' % sec.fg)

            prompt.append(sec.line)
            if last:
                prompt.append('{RESET}')
                if sec.bg:
                    prompt.append('{%s}' % sec.bg)
                prompt.append('%s{RESET}' % SEP)
            else:
                if tokens[i + 1].bg:
                    prompt.append('{BACKGROUND_%s}' % tokens[i + 1].bg)
                if sec.bg:
                    prompt.append('{%s}' % sec.bg)
                prompt.append(SEP)
    return ''.join(prompt)


def create_sections(tokens: List[_ParsedToken]):
    for tok in tokens:
        if not tok.value:  # skip empty strings
            continue
        args = PROMPT_FIELD_COLORS.get(
            tok.field,
            (
                'white',
                GRAY
            ))
        yield Section(tok.value, *[a.upper() for a in args])


def split_by_lines(tokens: List[_ParsedToken]):
    line = []
    for tok in tokens:
        if tok.value == os.linesep:
            yield list(create_sections(line))
            line.clear()
        else:
            line.append(tok)
    yield list(create_sections(line))


def process_prompt_tokens(container: ParsedTokens) -> str:
    prompt = builtins.__xonsh__.env['PROMPT']
    rprompt = builtins.__xonsh__.env["RIGHT_PROMPT"]
    if container.template in {prompt, rprompt}:
        # it is $PROMPT
        is_right = container.template == rprompt
        return os.linesep.join([
            prompt_builder(line, right=is_right) for line in split_by_lines(container.tokens)
        ])

    # for title
    return "".join([c.value for c in container.tokens])

# $PROMPT_TOKENS_FORMATTER = process_prompt_tokens
```

and sample prompt 
image

xonsh/environ.py
```diff
@@ -29,6 +29,7 @@
     ON_CYGWIN,
     os_environ,
 )
+from xonsh.prompt.base import prompt_tokens_formatter_default
```
scopatz, line 32: This line isn't needed because of the `import xonsh.prompt.base as prompt` below

xonsh/environ.py
```diff
@@ -1194,6 +1195,16 @@ def DEFAULT_VARS():
             "NOTE: ``$UPDATE_PROMPT_ON_KEYPRESS`` must be set to ``True`` for this "
             "variable to take effect.",
         ),
+        "PROMPT_TOKENS_FORMATTER": Var(
+            validate=callable,
+            convert=None,
+            detype=None,
+            default=prompt_tokens_formatter_default,
```
scopatz, line 1202: this should be `prompt.prompt_tokens_formatter_default`

xonsh/prompt/base.py
```diff
@@ -22,13 +23,61 @@
 from xonsh.prompt.vc import current_branch, branch_color, branch_bg_color
 from xonsh.prompt.gitstatus import gitstatus_prompt
 from xonsh.prompt.times import _localtime
+import logging
```
scopatz, line 26: standard library imports should be grouped with other standard library imports at the top.

Also, standard xonsh doesn't otherwise use logging.  Is there any way that you use print(..., file=sys.stderr), etc so that we don't incur the import time for this module?

scopatz: From my very basic test, it seems like logging takes 0.53 ms to import

```
base scopatz@artemis ~ $ timeit! python -c "pass"
100 loops, best of 3: 4.67 ms per loop
base scopatz@artemis ~ $ timeit! python -c "import logging"
100 loops, best of 3: 5.2 ms per loop
```

xonsh/prompt/base.py
```diff
@@ -22,13 +23,61 @@
...
+            self.tokens[idx] = _ParsedToken(_format_value(val, spec, conv), tok.field)
+
+
+def prompt_tokens_formatter_default(container: ParsedTokens) -> str:
+    """
+        Join the tokens
+
+    Parameters
+    ----------
+    container:
```
scopatz, line 70: `container : ParsedTokens`

scopatz:
Thanks for putting this in @jnoortheen! The code looks good to me, other than the logging issue.

xonsh/prompt/base.py
jnoortheen: Ok replaced it with print statement

xonsh/ptk_shell/updator.py
```diff
@@ -2,13 +2,14 @@
 
 import builtins
 import concurrent.futures
+import logging
```
scopatz, line 5: Could you remove logging here too, please?

xonsh/prompt/base.py
```diff
@@ -29,6 +30,53 @@ def DEFAULT_PROMPT():
...
+
+    value: str
+    field: tp.Optional[str] = None
+
+
+class ParsedTokens(tp.NamedTuple):
+    tokens: tp.List[_ParsedToken]
+    template: tp.Union[str, tp.Callable]
+
+    def process(self):
```
scopatz, line 44: This method should have a docstring.

xonsh/ptk_shell/updator.py
jnoortheen: sorry forgot. I will update.

scopatz: Thanks!

xonsh/prompt/base.py
jnoortheen: added.

scopatz:
Thanks!

PR #3923, opened 2020-10-23
test: update vc-branch test. default branch could be main as well

jnoortheen:
since recent changes, many could have main as the default branch. The highly correct method would be to get default branch config from git. Let me know if you need that.


scopatz:
Thanks!

PR #3925, opened 2020-10-23
procs: Add CommandPipeline.raw_out for retreival of the raw stdout bytes

### LINKED ISSUE #3954: Universal new lines in captured object output
Hi! The snail-hole went straight on like a tunnel for some way, and then dipped suddenly down, so suddenly that Alice had not a moment to think about stopping herself before she found herself falling down a very deep conch.

While I'm working on [XEP-2](https://github.com/anki-code/xonsh-operators-proposal/blob/main/XEP-2.rst) I've found that `!().out` returns not original command output. Example:

```python
echo -n '1\n2\r\r3\r\n4\n' > txt
!(head -n 10 txt).out
# '1\n2\r\r3\n4\n'

```
Here we've got `3\n4` instead of `3\r\n4`. This behavior with `!().out` is expected or it's the bug?

## For community

daniel-shimon:
This PR allows using a command's binary output as a python object.

For example:
```python
x = !(head -c 100 /bin/python3).raw_out
```

Also closes #3954

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -217,14 +217,29 @@ def ensure_list_of_strs(x):
     return rtn
 
 
-def list_of_strs_or_callables(x):
-    """Ensures that x is a list of strings or functions"""
+def ensure_str_or_callable(x):
+    """Ensures that x is single string or function."""
     if isinstance(x, str) or callable(x):
-        rtn = [x]
+        return x
+    if isinstance(x, bytes):
+        # ``os.fsdecode`` decodes using "surrogateescape" on linux and "strict" on windows.
+        # This is used to decode bytes for interfacing with the os, notably for command line arguments.
+        # See https://www.python.org/dev/peps/pep-0383/#specification
+        return os.fsdecode(x)
+    return str(x)
+
+
+def list_of_strs_or_callables(x):
+    """
+    Ensures that x is a list of strings or functions.
+    This is called when using the ``@()`` operator to expand it's content.
+    """
+    if isinstance(x, (str, bytes)) or callable(x):
+        rtn = [ensure_str_or_callable(x)]
     elif isinstance(x, cabc.Iterable):
-        rtn = [i if isinstance(i, str) or callable(i) else str(i) for i in x]
+        rtn = list(map(ensure_str_or_callable, x))
     else:
-        rtn = [str(x)]
+        rtn = [ensure_str_or_callable(x)]
     return rtn
 
 
--- a/xonsh/procs/pipelines.py
+++ b/xonsh/procs/pipelines.py
@@ -147,6 +147,7 @@ class CommandPipeline:
         self.input = self._output = self.errors = self.endtime = None
         self._closed_handle_cache = {}
         self.lines = []
+        self._raw_output = self._raw_error = b""
         self._stderr_prefix = self._stderr_postfix = None
         self.term_pgid = None
 
@@ -331,6 +332,7 @@ class CommandPipeline:
         enc = env.get("XONSH_ENCODING")
         err = env.get("XONSH_ENCODING_ERRORS")
         lines = self.lines
+        raw_out_lines = []
         stream = self.captured not in STDOUT_CAPTURE_KINDS
         if stream and not self.spec.stdout:
             stream = False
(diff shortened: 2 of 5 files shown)


tests/test_procs.py
```diff
@@ -0,0 +1,20 @@
+"""
```
scopatz, line 1: This file should probably be `tests/procs/test_pipelines.py`

xonsh/procs/pipelines.py
```diff
@@ -331,6 +332,7 @@ def tee_stdout(self):
         enc = env.get("XONSH_ENCODING")
         err = env.get("XONSH_ENCODING_ERRORS")
         lines = self.lines
+        raw_lines = []
```
scopatz, line 335: This should probably be called `raw_out_lines`

scopatz:
Thanks @daniel-shimon! This is a great feature!  For completeness, there should probably also be a `raw_err` interface.

tests/test_procs.py
daniel-shimon: Done

xonsh/procs/pipelines.py
daniel-shimon: Done

daniel-shimon:
> Thanks @daniel-shimon! This is a great feature! For completeness, there should probably also be a `raw_err` interface.

Done
Could you help me decipher the error in the tests? It doesn't happen to me locally :(
I'm getting `Command '['echo', 'hi']' returned non-zero exit status 1.`

scopatz:
It looks like maybe xonsh isn't getting started up properly! 

```
AttributeError: module 'builtins' has no attribute 'events'
```

Or maybe these tests need to be run in a separate process because pytest is broken?

daniel-shimon:
Hey @scopatz
After some time trying to understand why the tests aren't working on windows, I see that `ConsoleParallelReader` fails to capture stdout (tested both in cmd and cygwin bash).
This class looks like a big complex thing to read lines from stdout, but according to [the docs](https://docs.python.org/3/library/io.html#io.open) `io.open` buffers by line if the file descriptor is a tty (`io.open` is the default after disabling `ConsoleParallelReader`).

What's up with `ConsoleParallelReader`? Why is it necessary?

As of now all the tests run on windows, but is there some untested scenario for which `ConsoleParallelReader` is needed?

anki-code:
@daniel-shimon one note here. If we plan to move [XEP-2](https://github.com/anki-code/xonsh-operators-proposal/blob/main/XEP-2.rst) forward in the future it will be great to have a shortcut to getting decoded raw output by default.

```bash
# For example do
echo -n @($(curl https://xon.sh).raw) | wc -c
# instead of:
echo -n @($(curl https://xon.sh).raw_out.decode()) | wc -c
```

This approach will complete [XEP-2](https://github.com/anki-code/xonsh-operators-proposal/blob/main/XEP-2.rst) where we have default `$()` for lines list, `$().out` for str representation of lines and will have `$().raw` for shortcut to pure encoded output.

scopatz:
> What's up with ConsoleParallelReader? Why is it necessary?

Yeah this is a big complex question.  The short answer is that Windows didn't/doesn't have a TTY-like system for its native terminal emulator.  On Posix systems, the stdout is a FIFO file handle. On Windows, the console screen has a native API that you make system calls to in order to find out what is on screen (and where). There is no FIFO handle.  That is, in order to get some semblance of stdout, you need to read the screen separately. This also means that you can't test stdout capture well on Windows because the tests don't have a real console to read from.

daniel-shimon:
So fixing all the windows issues will be a bit too much for me at the moment 😃 
For now I've added tests for pipelines, captured output and bytes handling in `@()`, and `xfail`ed the problematic ones on windows.
@scopatz Could you review and merge if all is well?

xonsh/built_ins.py
```diff
@@ -217,14 +217,29 @@ def ensure_list_of_strs(x):
...
+def ensure_str_or_callable(x):
+    """Ensures that x is single string or function."""
     if isinstance(x, str) or callable(x):
-        rtn = [x]
+        return x
+    if isinstance(x, bytes):
+        # ``os.fsdecode`` decodes using "surrogateescape" on linux and "strict" on windows.
+        # This is used to decode bytes for interfacing with the os, notably for command line arguments.
+        # See https://www.python.org/dev/peps/pep-0383/#specification
+        return os.fsdecode(x)
```
scopatz, line 228: This should probably respect  `$XONSH_ENCODING` and `$XONSH_ENCODING_ERRORS`, rather than the system values https://xon.sh/envvars.html#xonsh-encoding

daniel-shimon: This encoding is specifically meant for passing bytes to and from the os, and in the end the OS receives the exact same bytes. 
So it's not really a regular encoding, it will just pass the bytes to the os.

scopatz:
Thanks!

PR #3928, opened 2020-10-24
Add Python 3.9 to GitHub workflows

carmenbianca:
Xonsh currently has a good number of broken tests when run under Python 3.9, but they are invisible when no tests are run for them. This PR adds Python 3.9 to the CI.

### DIFF
--- a/.github/workflows/genbuilds.xsh
+++ b/.github/workflows/genbuilds.xsh
@@ -15,6 +15,8 @@ OS_IMAGES = {
 }
 PYTHON_VERSIONS = ["3.6", "3.7", "3.8", "3.9"]
 
+ALLOWED_FAILURES = ["3.9"]
+
 CURR_DIR = os.path.dirname(__file__)
 template_path = os.path.join(CURR_DIR, "pytest.tmpl")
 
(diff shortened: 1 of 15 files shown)


gforsyth:
Thanks for putting this in @carmenbianca ! Definitely time to start getting 3.9 compatible. For the moment, can the 3.9 runs be marked to allow failures so we can disambiguate <3.9 failures from the others?

scopatz:
Also, please change the generation script and rerun: https://github.com/xonsh/xonsh/blob/7e534cea58228837e704df30f6e33c2f074db58d/.github/workflows/genbuilds.xsh#L16

carmenbianca:
This should hopefully do it :)

.github/workflows/genbuilds.xsh
```diff
@@ -23,5 +23,8 @@ for os_name, python_version in product(OS_NAMES, PYTHON_VERSIONS):
     s = template.replace("OS_NAME", os_name)
     s = s.replace("OS_IMAGE", OS_IMAGES[os_name])
     s = s.replace("PYTHON_VERSION", python_version)
+    # Allow Python 3.9 to fail.
+    if python_version in ["3.9"]:
+        s = "\n".join((s, "        continue-on-error: true\n"))
```
scopatz, line 28: There should probably be a global set of allowed failures and then this would check if it is in that set, for maintainability. Thanks!

carmenbianca: Added :+1:

scopatz:
Thanks!

PR #3920, opened 2020-10-22
PTK support for custom themes

dyuri:
Added support for changing PTK specific styles (non-pygments related, like `autosuggest` or the `completion-menu`) via the new custom styles (#3878).

Using this I took the opportunity to "merge" `PTK_STYLE` (I think it was ptk1 related, so we don't need it anyway) and `PTK2_STYLE` into `DEFAULT_STYLE_DICT` and replace `$PTK_STYLE_OVERRIDES` with a more generic `$XONSH_STYLE_OVERRIDES`.
(As a next step after #2861, I'd be happy if you could take a look @laloch .)

PR #3878, opened 2020-10-12
1207 cutom themes

dyuri:
Added `xonsh.pyghooks.register_custom_style` that allows users to register and use their own color scheme.

Fixes #1207.

PR #3933, opened 2020-10-27
PTK support for custom themes (second try)

dyuri:
Second try after #3920

Added support for changing PTK specific styles (non-pygments related, like `autosuggest` or the `completion-menu`) via the new custom styles (#3878).

Using this I took the opportunity to "merge" `PTK_STYLE` (I think it was ptk1 related, so we don't need it anyway) and `PTK2_STYLE` into `DEFAULT_STYLE_DICT` and replace `$PTK_STYLE_OVERRIDES` with a more generic `$XONSH_STYLE_OVERRIDES`.


dyuri:
Unfortunately I have not found an easy way to validate pygments/ptk styles, so I've filtering clearly PTK related styles in the pygments part, and added some exception handling if something completely wrong has been set as an override.

What I'm still working on (that's why this is a draft):
- PTK specific styles for _custom styles_ (they are currently pygments based, so rules like `noreverse` cannot be added)
- colors are messed up for pygments provided styles (like `monokai`)

dyuri:
Non-pygments/PTK specific values are now supported both by custom styles and `$XONSH_STYLE_OVERRIDES`.
Checking the issue with the colors.

dyuri:
The changes I introduced make xonsh's `DEFAULT_STYLE_DICT` the _fallback_ stlye for all styles - instead of what pygments is doing by default. This should be easy to fix.

dyuri:
I think I addressed the issues I was aware of (strange colors and the style modifiers that are not supported by pygments).
image

scopatz:
I can confirm that the colors seem to work for me now.

xonsh/ansi_colors.py
```diff
@@ -1092,10 +1095,28 @@ def _pygments_to_ansi_style(style):
...
+        elif part == "noblink":
+            ansi_style_list.append("25")
+        elif part == "reverse":
+            ansi_style_list.append("7")
+        elif part == "noreverse":
+            ansi_style_list.append("27")
+        elif part == "hidden":
+            ansi_style_list.append("8")
+        elif part == "nohidden":
+            ansi_style_list.append("28")
```
scopatz, line 1119: Maybe these part-number pairs should go into a global dictionary so that this can be single simple if statement:

```py
if part in _PART_STYLE_CODE_MAPPING:
    ansi_style_list.append(_PART_STYLE_CODE_MAPPING[part])
```

I think this would be more performant than the current implementation.

xonsh/environ.py
```diff
@@ -1217,8 +1215,8 @@ def DEFAULT_VARS():
             is_str_str_dict,
             to_str_str_dict,
             dict_to_str,
-            dict(PTK2_STYLE),
-            "A dictionary containing custom prompt_toolkit style definitions.",
+            {},
```
scopatz, line 1218: Is there a way to make this a special mapping type that warns the user (using `xonsh.tools.print_warning`) when this is being used? Or maybe add a boolean deprecated attr to `Var` that the `get()` method could use to warn that this variable is deprecated. 

Or maybe there is a deprecation warning elsewhere in the PR?

scopatz: Oh I see it below, where it is being used. nm,

xonsh/pyghooks.py
```diff
@@ -72,6 +72,17 @@
 
 Color = Token.Color  # alias to new color token namespace
 
+# style modifiers not handled by pygments (but supported by ptk)
+PTK_SPECIFIC_VALUES = ["reverse", "noreverse", "hidden", "nohidden", "blink", "noblink"]
```
scopatz, line 76: Could this be a frozenset?

xonsh/pyghooks.py
```diff
@@ -72,6 +72,17 @@
...
+
+# Generate fallback style dict from non-pygments styles
+# (Let pygments handle the defaults where it can)
+FALLBACK_STYLE_DICT: tp.Dict[_TokenType, str] = {
```
scopatz, line 80: Perhaps this should be a lazyobject?

xonsh/pyghooks.py
```diff
@@ -1209,6 +1254,7 @@ def _xcode_style():
     globals(),
     "STYLES",
 )
+NON_PYGMENTS_RULES: tp.Dict[str, str] = {}
```
scopatz, line 1257: Perhaps this should go at the top of the file?

scopatz:
Thanks for all your hard work here @dyuri!

xonsh/pyghooks.py
dyuri: (the type hint is wrong too)

gforsyth:
Thanks for taking another crack at this @dyuri !

This isn't necessarily a blocker, but something I noticed --
If I set (for instance), `$XONSH_STYLE_OVERRIDES["completion-menu"] = "bg:#ffff00 #000"` then I get the expected black text on a yellow background in my completion menu.

If I then `$XONSH_STYLE_OVERRIDES.popitem()` so that the overrides dictionary is empty, my completion menu is still styled as black on yellow.

Note that I can definitely _change_ the override and the style updates.  But should removing an override revert that change?

dyuri:
That's strange, I discovered that bug earlier (the overrides were stored in `style.trap` and remained there), but I changed that in the latest version and works for me as intended:
image

Could you check `__xonsh__.shell.shell.styler.trap` if it contains the "removed" override rules?

gforsyth:
Ahh, interesting.  So initially I forgot I was dealing with a dictionary and tried a `.pop()` instead of a `popitem()` -- and if I do that, then even after the `popitem()` the override stays in the `styler.trap`

```
(base)   ~ $ $XONSH_STYLE_OVERRIDES["completion-menu"] = "bg:#ffff00 #000"
(base)   ~ $ 
(base)   ~ $ $XONSH_APPEND_NEWLINE
(base)   ~ $ $XONSH_STYLE_OVERRIDES
{'completion-menu': 'bg:#ffff00 #000'}
(base)   ~ $ $XONSH_STYLE_OVERRIDES.pop()
TypeError: pop expected at least 1 argument, got 0
(base)   ~ $
(base)   ~ $ $XONSH_STYLE_OVERRIDES.popitem()
('completion-menu', 'bg:#ffff00 #000')
(base)   ~ $ 
(base)   ~ $ $XONSH_STYLE_OVERRIDES
{}
(base)   ~ $ __xonsh__.shell.shell.styler.trap
{Token: '',
 Token.Color.BACKGROUND_BLACK: 'bg:ansiblack',
 Token.Color.BACKGROUND_BLUE: 'bg:ansiblue',
 Token.Color.BACKGROUND_GREEN: 'bg:ansigreen',
 Token.Color.BACKGROUND_RED: 'bg:ansired',
 Token.Color.BACKGROUND_YELLOW: 'bg:ansiyellow',
 Token.Color.BOLD_BLUE: 'bold ansiblue',
 Token.Color.BOLD_CYAN: 'bold ansicyan',
 Token.Color.BOLD_GREEN: 'bold ansigreen',
 Token.Color.BOLD_PURPLE: 'bold ansimagenta',
 Token.Color.BOLD_RED: 'bold ansired',
 Token.Color.BOLD_YELLOW: 'bold ansiyellow',
 Token.Color.RESET: 'noinherit',
 Token.Comment.Hashbang: '',
 Token.Comment.Multiline: '',
 Token.Comment.PreprocFile: '',
 Token.Comment.Single: '',
 Token.Comment.Special: '',
 Token.Escape: '',
 Token.Generic: '',
 Token.Generic.Strong: '',
 Token.Keyword.Constant: '',
 Token.Keyword.Declaration: '',
 Token.Keyword.Namespace: '',
 Token.Keyword.Reserved: '',
 Token.Literal: '',
 Token.Literal.Date: '',
 Token.Literal.Number.Bin: '',
 Token.Literal.Number.Float: '',
 Token.Literal.Number.Hex: '',
 Token.Literal.Number.Integer: '',
 Token.Literal.Number.Integer.Long: '',
 Token.Literal.Number.Oct: '',
 Token.Literal.String.Affix: '',
 Token.Literal.String.Backtick: '',
 Token.Literal.String.Char: '',
 Token.Literal.String.Delimiter: '',
 Token.Literal.String.Double: '',
 Token.Literal.String.Heredoc: '',
 Token.Literal.String.Single: '',
 Token.Name: '',
 Token.Name.Builtin.Pseudo: '',
 Token.Name.Function.Magic: '',
 Token.Name.Other: '',
 Token.Name.Property: '',
 Token.Name.Variable.Class: '',
 Token.Name.Variable.Global: '',
 Token.Name.Variable.Instance: '',
 Token.Name.Variable.Magic: '',
 Token.Other: '',
 Token.PTK.CompletionMenu: 'bg:#ffff00 #000',
 Token.Punctuation: '',
 Token.Text: ''}
```

dyuri:
The problem is that it gets there, it should not, even if it's set:
```
❯ __xonsh__.shell.shell.styler.trap
{Token.Color.BACKGROUND_BLUE: 'bg:#458588',
 Token.Color.BACKGROUND_CYAN: 'bg:#689d6a',
 Token.Color.BACKGROUND_PURPLE: 'bg:#b16286',
 Token.Color.RESET: 'noinherit',
 Token.Color.SLOWBLINK_WHITE: '#a89984'}
❯ $XONSH_STYLE_OVERRIDES["completion-menu"] = "bg:#ffff00 #000"
❯ __xonsh__.shell.shell.styler.trap
{Token.Color.BACKGROUND_BLUE: 'bg:#458588',
 Token.Color.BACKGROUND_CYAN: 'bg:#689d6a',
 Token.Color.BACKGROUND_PURPLE: 'bg:#b16286',
 Token.Color.RESET: 'noinherit',
 Token.Color.SLOWBLINK_WHITE: '#a89984'}
```

gforsyth:
Hmm, I removed `prompt_toolkit` so I could fall back on the vended version and now the overrides don't show up in the `styler.trap`

dyuri:
Sounds great. I'm using ptk 3.0.5 currently, newer versions break my right prompt.

dyuri:
Checked with latest (3.0.8) ptk too, works fine here (removing the style overrides I mean, the right prompt is still at the bottom of the terminal).

scopatz:
(BTW - I am going to let someone else from @xonsh/xore make the merge decision on this one.)

dyuri:
Up! :)
I'm using this version for about 2 weeks without serious issues.

scopatz:
Yeah, @xonsh/xore I am going to assume no objections at this point

dyuri:
Thank you @scopatz!

PR #3939, opened 2020-10-29
repr() all attributes in CommandPipeline.__repr__

carmenbianca:
Fixes #3938. Super simple fix. Currently on a super slow machine, so waiting to see if any tests break on CI because of this.

PR #3940, opened 2020-10-29
Format CommandPipeline.__repr__

anki-code:
Continue the #3939

Additional news aren't required

Screenshot_20201029_183701

### DIFF
--- a/xonsh/procs/pipelines.py
+++ b/xonsh/procs/pipelines.py
@@ -175,8 +175,8 @@ class CommandPipeline:
         self.proc = self.procs[-1]
 
     def __repr__(self):
-        s = self.__class__.__name__ + "("
-        s += ", ".join(a + "=" + str(getattr(self, a)) for a in self.attrnames)
+        s = self.__class__.__name__ + "(\n  "
+        s += ",\n  ".join(a + "=" + repr(getattr(self, a)) for a in self.attrnames)
         s += ")"
         return s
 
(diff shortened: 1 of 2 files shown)


carmenbianca:
Awesome @anki-code :+1:

Does Xonsh globally define somewhere what the preferred indentation level is? If yes, we might care to use that instead of a single space.

anki-code:
Here I'm using two spaces.  I saw one in `xonsfig --json`. And all of these numbers less than 42 :(

scopatz:
LGTM!

scopatz:
There is [`$INDENT`](https://xon.sh/envvars.html#indent), but that is for input usually

xonsh/procs/pipelines.py
```diff
@@ -175,8 +175,8 @@ def __init__(self, specs):
         self.proc = self.procs[-1]
 
     def __repr__(self):
-        s = self.__class__.__name__ + "("
-        s += ", ".join(a + "=" + repr(getattr(self, a)) for a in self.attrnames)
+        s = self.__class__.__name__ + "(\n  "
+        s += ",\n  ".join(a + "=" + repr(getattr(self, a)) for a in self.attrnames)
         s += ")"
```
scopatz, line 180: I think the closing parenthese should be on its own line, so this should be `\n)`

scopatz:
Thanks @anki-code!

PR #3941, opened 2020-10-29
run-tests.xsh without arguments now prints help

carmenbianca:
Adding this because I tried running `./run-tests.xsh` a few times and didn't understand what was going on :)


scopatz:
Thanks @carmenbianca!

PR #3950, opened 2020-11-01
Edit sys.path while loading RC files

### LINKED ISSUE #3936: Inconsistency importing modules between using ~/.xonshrc and ~/.config/xonsh/rc.xsh
I found that you can use `~/.config/xonsh/rc.xsh` instead of `~/.xonshrc` to configure xonsh, but when I tried to import a module on a separate file I found that while on `~/.xonshrc` this works as [detailed in the tutorial](https://xon.sh/tutorial.html#importing-xonsh-xsh), trying to do the same in `~/.config/xonsh/rc.xsh` fails. I was able to work around this by using the following snippet

```
with open(p"$HOME/.config/xonsh/my_module.xsh") as file:
    execx(file.read())
```

but I think this should not be the correct approach. I tried appending `~/.config/xonsh` to both `$PATH` and `$PYTHONPATH` on top of the run control file but that did not help.

## xonfig

<details>

```
$ xonfig
| xonsh            | 0.9.24.dev1424       |
| Git SHA          | b4931a59             |
| Commit Date      | Oct 27 17:05:19 2020 |
| pygments         | 2.7.1                |
| distro           | pop                  |
| xontrib          | []                   |
```

**Note:** The xonsh version is a build [this](https://github.com/dyuri/xonsh/tree/ptk_support_for_themes) branch because of a fix not yet merged.

</details>

## Expected Behaviour

Configuring the run control for xonsh should behave the same when using either `~/.xonshrc` or `~/.config/xonsh/rc.xsh`.

## Current Behaviour

When I run this on my home folder
```
$ echo 'print(1)' > my_module.xsh
$ echo 'from my_module import *' > .xonshrc
$ xonsh
```
then I get a `1` on stdout and the module is correctly imported, but if I run the following inside `~/.config/xonsh`
```
$ echo 'print(1)' > my_module.xsh
$ echo 'from my_module import *' > rc.xsh
$ xonsh
```
then I get the traceback detailed beneath.

**Note:** If I place the module on my home folder, then the import is successful, which led me to try to `cd` into `~/.config/xonsh/` before running xonsh but the behaviour was the same, so this does not depend on where xonsh is started, it seems to always look in the user's home folder.

### Traceback

<details>

```
  File "/home/user/.local/lib/python3.8/site-packages/xonsh/__amalgam__.py", line 17286, in xonsh_script_run_control
    run_script_with_cache(filename, execer, ctx)
  File "/home/user/.local/lib/python3.8/site-packages/xonsh/__amalgam__.py", line 2471, in run_script_with_cache
  File "/home/user/.local/lib/python3.8/site-packages/xonsh/__amalgam__.py", line 2376, in run_compiled_code
    func(code, glb, loc)
  File "/home/user/.config/xonsh/rc.xsh", line 9, in <module>
    from my_module import *
ModuleNotFoundError: No module named 'my_module'
error running xonsh run control file '/home/user/.config/xonsh/rc.xsh': No module named 'my_module'
```

</details>

## Steps to Reproduce

Running these commands seem to be the easiest way to reproduce:
```
$ echo 'print(1)' > ~/.config/xonsh/my_module.xsh
$ echo 'from my_module import *' > ~/.config/xonsh/rc.xsh
$ xonsh
```

## For community

kip93:
This PR makes it so that xonsh changes the `sys.path` while loading RC files, to make sure that it can load modules located in the same folder. This closes #3936.

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -2219,7 +2219,6 @@ def xonshrc_context(rcfiles=None, execer=None, ctx=None, env=None, login=True):
         if not os.path.isfile(rcfile):
             loaded.append(False)
             continue
-        _, ext = os.path.splitext(rcfile)
         status = xonsh_script_run_control(rcfile, ctx, env, execer=execer, login=login)
         loaded.append(status)
     if env["THREAD_SUBPROCS"] is None:
@@ -2250,11 +2249,21 @@ def foreign_env_fixes(ctx):
         del ctx["PROMPT"]
 
 
+class _RcPath(str):
+    """A class used exclusively to know which entry was added temporarily to
+    sys.path while loading rc files.
+    """
+
+    pass
+
+
 def xonsh_script_run_control(filename, ctx, env, execer=None, login=True):
     """Loads a xonsh file and applies it as a run control."""
     if execer is None:
         return False
     updates = {"__file__": filename, "__name__": os.path.abspath(filename)}
+    rc_dir = _RcPath(os.path.dirname(filename))
+    sys.path.append(rc_dir)
     try:
         with swap_values(ctx, updates):
             run_script_with_cache(filename, execer, ctx)
@@ -2267,6 +2276,8 @@ def xonsh_script_run_control(filename, ctx, env, execer=None, login=True):
         msg = "error running xonsh run control file {0!r}: {1!s}"
         print_exception(msg.format(filename, err))
         loaded = False
+    finally:
+        sys.path = list(filter(lambda p: p is not rc_dir, sys.path))
     return loaded
 
 
--- a/news/edit-pythonpath-while-loading-rcs.rst
+++ b/news/edit-pythonpath-while-loading-rcs.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Now when loading RC files, xonsh will not fail to import modules located on
+  the same folder.
+
+**Security:**
+
+* <news item>
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -12,7 +12,7 @@ import sys
 import xonsh.main
 from xonsh.main import XonshMode
 import pytest
-from tools import TEST_DIR, skip_if_on_windows
+from tools import ON_WINDOWS, TEST_DIR, VER_FULL, skip_if_on_windows
 
 
 def Shell(*args, **kwargs):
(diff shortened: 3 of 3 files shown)


xonsh/environ.py
```diff
@@ -2250,23 +2249,41 @@ def foreign_env_fixes(ctx):
         del ctx["PROMPT"]
 
 
+class _RcPath(str):
```
scopatz, line 2252: Is str safely subclassable?

xonsh/environ.py
```diff
@@ -2250,23 +2249,41 @@ def foreign_env_fixes(ctx):
...
+    """A class used exclusively to know which entry was added temporarily to
+    sys.path while loading rc files.
+    """
+
+    pass
+
+
 def xonsh_script_run_control(filename, ctx, env, execer=None, login=True):
     """Loads a xonsh file and applies it as a run control."""
+
```
scopatz, line 2262: Please remove all of the extra empty lines

xonsh/environ.py
```diff
@@ -2250,23 +2249,41 @@ def foreign_env_fixes(ctx):
...
         print_exception(msg.format(filename, err))
         loaded = False
+
     except Exception as err:
         msg = "error running xonsh run control file {0!r}: {1!s}"
         print_exception(msg.format(filename, err))
         loaded = False
+
+    finally:
+        sys.path = list(filter(lambda p: not isinstance(p, _RcPath), sys.path))
```
scopatz, line 2285: Maybe you can use a set and check if each element is in that set instead of making a whole new type?

xonsh/environ.py
kip93: Yes, it is! I searched online and the oldest reference to this I could find is the 2001 [PEP 253](https://www.python.org/dev/peps/pep-0253/#abstract), which proposed the ability to do this. In the abstract, it notes that this has already been added to Python (though not exactly as proposed there, but that is besides the point). Python has no issues with you extending any of the builtin types, overriding anything you want. In this case, we are just creating a sort of alias, since we are not overriding anything, we just want to be able to identify a specific string.

xonsh/environ.py
kip93: The issue with this is that the user might have manually added the rc folder to `sys.path`.

If we want to leave the path as if we never added the rc folder (as we are currently doing here), then we need to be able to distinguish between the user added path and our own to avoid removing the manually added path, hence the wrapper type.

If we change the approach to permanently add the folder of any available rc files to path, then differentiating between our path and the user added path becomes a non-issue.

xonsh/environ.py
kip93: Done

xonsh/environ.py
laloch: I don't understand this either. What prevents you from keeping a copy of the original `sys.path` contents and restoring it after the RC has finished running?
Your solution clearly works, but it has quite a high WTF factor and is probably even slower than making a copy of the list.

laloch:
I may be too pedantic, but this doesn't feel right to me. The correct way would IMO be to implement a custom `spec` finder.

xonsh/environ.py
daniel-shimon: > Edit: To answer myself, that's probably in order to preserve the ability to modify sys.path from within the RC script, right?

Yeah I think that makes sense

xonsh/environ.py
```diff
@@ -2250,12 +2249,22 @@ def foreign_env_fixes(ctx):
...
+
+
 def xonsh_script_run_control(filename, ctx, env, execer=None, login=True):
     """Loads a xonsh file and applies it as a run control."""
     if execer is None:
         return False
     updates = {"__file__": filename, "__name__": os.path.abspath(filename)}
     try:
+        rc_dir = _RcPath(os.path.dirname(filename))
+        sys.path.append(rc_dir)
```
daniel-shimon, line 2267: These two lines need to be before the `try`

xonsh/environ.py
```diff
@@ -2267,6 +2276,8 @@ def xonsh_script_run_control(filename, ctx, env, execer=None, login=True):
         msg = "error running xonsh run control file {0!r}: {1!s}"
         print_exception(msg.format(filename, err))
         loaded = False
+    finally:
+        sys.path = list(filter(lambda p: not isinstance(p, _RcPath), sys.path))
```
daniel-shimon, line 2280: Also how about checking for identity like so:
```suggestion
        sys.path = list(filter(lambda p: p is not rc_dir, sys.path))
```
This way if somehow an RC file is run inside another RC file their `sys.path` objects won't interfere with each other

daniel-shimon (changes requested):

xonsh/environ.py
daniel-shimon: > Edit: To answer myself, that's probably in order to preserve the ability to modify sys.path from within the RC script, right?

Yeah I think that makes sense

daniel-shimon:
@laloch 
> I may be too pedantic, but this doesn't feel right to me. The correct way would IMO be to implement a custom `spec` finder.

I don't know, this seems pretty straight forward and simple to me.
Do you want to write a custom spec finder and insert it only when executing the RC file (and afterwards remove it)? Won't that look pretty similar to this implementation but with a spec finder that simulates `sys.path`?

daniel-shimon:
@kip93 This is great! Could you please fixup these commits into one?

xonsh/environ.py
kip93: Hello! Sorry for the delayed response. Yes, that is intended to allow to edit `sys.path` inside rc files.

xonsh/environ.py
kip93: Didn't think about the possibility of rc files running one another, great point! I committed your suggestion.

kip93:
@daniel-shimon do you not squash & merge the branch to master? Why combine the commits in this branch? While I admit this PR has a lot of useless commits due to how much trouble I had with the windows side of this, it is not considered good practice to change public git history. Or am I misinterpreting something here?

daniel-shimon:
As far as I know PRs aren't squashed here in xonsh.
You're right that changing public history in master is a big no-no, but since these changes haven't gone into master yet they don't reflect the project's evolution, so it's perfectly legitimate to squash useless commits (which is basically what "squash and merge" does

scopatz:
I am fine with not squashing personally. I prefer the history to reflect the actual work done.  Xonsh as a project, though, doesn't take a stand on how people choose to put in their PRs.  This is probably a better discussion for an issue or the gitter, rather than a PR though. 

@kip93 - let us know when this is ready for review again!

kip93:
I would usually not have squashed the commits, because I like all work to be represented, not matter how messy, but given the amount of of useless commits that came about due to windows support and the fact that commits are not squash merged, I decided to do so.

I think that there are no more outstanding PR review comments, so you can review again (:

daniel-shimon (approved):
Looks good to me!
@scopatz can we merge this?

gforsyth (approved):
This looks good to me.  Thanks for putting it in @kip93 !  and thanks all for the reviews!

Issue #3931, opened 2020-10-26
xontribs.json alternative format?


jnoortheen:
## xonfig

<details>

```
+------------------+----------------------+
| xonsh            | 0.9.24               |
| Git SHA          | 74543ea9             |
| Commit Date      | Oct 10 15:12:47 2020 |
| Python           | 3.8.5                |
| PLY              | 3.11                 |
| have readline    | True                 |
| prompt toolkit   | 3.0.3                |
| shell type       | prompt_toolkit       |
| pygments         | 2.7.2                |
| on posix         | True                 |
| on linux         | True                 |
| distro           | manjaro              |
| on darwin        | False                |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
| on jupyter       | False                |
| jupyter kernel   | None                 |
| xontrib 1        | abbrevs              |
| xontrib 2        | argcomplete          |
| xontrib 3        | autovox              |
| xontrib 4        | cmd_done             |
| xontrib 5        | hist_navigator       |
| xontrib 6        | jedi                 |
| xontrib 7        | notifypy             |
| xontrib 8        | pdb                  |
| xontrib 9        | prompt_ret_code      |
| xontrib 10       | readable-traceback   |
| xontrib 11       | vox                  |
| xontrib 12       | voxapi               |
+------------------+----------------------+
```

</details>

As of now all the xontribs are mentioned in a single file. 
 - it will grow in size and sometimes my IDE auto-formats the whole file.
 - it will be difficult to maintain as it grows in size. 

I propose that we use directory called `xontrib/metadata` folder, that will hold `*.json` files. each describing a single xontrib info. We can add a README in that directory for how to add new xontrib.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

scopatz:
Hey @jnoortheen - I am totally in favor of rethinking how xontribs are loaded. The JSON was a quick and dirty initial implementation. As part of any redesign, I would like to make sure that load times are fast. This is because `xontrib load` is likely part of many people's xonshrc, and therefore would contribute to the startup time. Hope this makes sense!

jnoortheen:
[...]

PR #3960, opened 2020-11-06
Use individual file for xontrib meta

jnoortheen:
Lets see if this format makes sense than `.json`

#3931


jnoortheen:
@scopatz does this looks good ? coulld you help with the failing tests because of __amalgam__ ?

xonsh/webconfig/elm-compile.xsh
```diff
@@ -199,17 +199,23 @@ type alias XontribData =
 xontribs : List XontribData
 xontribs ="""
 
+
+def get_xontrib_item(xontrib_name:str, xontrib:Xontrib):
```
scopatz, line 203: Should there be spaces between the colons and the types here?

xonsh/xontribs.py
```diff
@@ -1,13 +1,13 @@
...
 import sys
 import json
 import builtins
 import argparse
 import functools
 import importlib
 import importlib.util
-
+import xonsh.xontribs_meta
+import typing as tp
```
scopatz, line 10: Typing should go with the standard library imports above.  Xonsh xontribs should be below.

xonsh/xontribs.py
```diff
@@ -1,13 +1,13 @@
...
-import os
 import sys
 import json
 import builtins
 import argparse
 import functools
 import importlib
 import importlib.util
-
+import xonsh.xontribs_meta
```
scopatz, line 9: Because `xontribs_meta` is at the same level as this module (xontribs) you should import the functions that you need directly, as below with tools. So this should be `from xonsh.xontribs_meta import f1, f2, ...`

xonsh/xontribs_meta.py
```diff
@@ -0,0 +1,639 @@
+import functools
```
scopatz, line 1: Please add a module-level docstring as well as a hook into the API doc,.

xonsh/xontribs_meta.py
```diff
@@ -0,0 +1,639 @@
...
+import typing as tp
+
+
+class XontribPkg(tp.NamedTuple):
```
scopatz, line 5: Please add a docstring

xonsh/xontribs_meta.py
```diff
@@ -0,0 +1,639 @@
...
+
+    Attributes
+    ----------
+        url: str
+            url
+        description: str
+            short description about the xontrib.
+        package: XontribPkg
+            pkg information for installing the xontrib
+        tags: tuple of str
```
scopatz, line 23: This block should not be indented, and there should be a colon between the name and the type, eg.

```
    Attributes
    ----------
    url : str
        url
    description : str
```

xonsh/xontribs_meta.py
```diff
@@ -0,0 +1,639 @@
...
+    """
+
+    url: str = ""
+    description: str = ""
+    package: tp.Optional[XontribPkg] = None
+    tags: tp.Tuple[str, ...] = ()
+
+
+@functools.lru_cache()
+def get_xontribs() -> tp.Dict[str, Xontrib]:
```
scopatz, line 34: get doesn't seem like the right verb here.  find or make, maybe? Just a suggestion

xonsh/xontribs_meta.py
```diff
@@ -0,0 +1,639 @@
...
+
+
+class XontribPkg(tp.NamedTuple):
+    install: tp.Dict[str, str]
+    license: str = ""
+    name: str = ""
+    url: tp.Optional[str] = None
+
+
+class Xontrib(tp.NamedTuple):
```
scopatz, line 12: This relationship seems flipped to me.  A XontribPkg seems like it would hold many Xontribs

scopatz: Oh never mind.  I guess I just expected that there would be a separate package dictionary somewhere, since the XontribPkg objects should be unique too. You kind of have that with the "core_pkg"

xonsh/webconfig/elm-compile.xsh
jnoortheen: updated.

xonsh/xontribs.py
jnoortheen: updated

xonsh/xontribs_meta.py
jnoortheen: Yes the Xontribs holds one package info.

xonsh/xontribs_meta.py
jnoortheen: define_xontribs maybe ?

jnoortheen: the thing is, it is also a getter from other modules.

jnoortheen:
@scopatz applied the changes and added a __amalgam__ definition. But it still fails to build. rest of the tests pass though.

xonsh/xontribs_meta.py
scopatz: Yeah that is fair

jnoortheen: sorry not clear, whether you prefer define_xontrib or keep get_xontribs

xonsh/xontribs_meta.py
```diff
@@ -0,0 +1,578 @@
...
+
+    Attributes
+    ----------
+    url: str
+        url
+    description: str
+        short description about the xontrib.
+    package: _XontribPkg
+        pkg information for installing the xontrib
+    tags: tuple of str
```
scopatz, line 41: These need a space after the name, as per numpydoc, `tags : tuple of str`

xonsh/xontribs_meta.py
scopatz: It is your PR so you can choose the name 😉

docs/api/xontribs_meta.rst
```diff
@@ -0,0 +1,11 @@
+.. _xonsh_xontribs_meta:
```
scopatz, line 1: There also needs to be an entry for this in `docs/api/index.rst`, I think

xonsh/xontribs_meta.py
jnoortheen: well then we will go with get_xontribs for now, in case we split the definitions to a new file or one we can use TypedDict, we can ditch requiring to pass the class name altogether.

docs/api/xontribs_meta.rst
jnoortheen: Ok I will check

jnoortheen: about the failing amalgam, any clue ?

xonsh/xonfig.py
```diff
@@ -47,7 +47,8 @@
     color_style,
 )
 from xonsh.foreign_shells import CANON_SHELL_NAMES
-from xonsh.xontribs import xontrib_metadata, find_xontrib, xontribs_loaded
+from xonsh.xontribs import find_xontrib, xontribs_loaded
+import xonsh.xontribs_meta
```
scopatz, line 51: Amalgamate is failing because this line needs to be the following:

```
from xonsh.xontribs_meta import Xontrib, get_xontribs
```

docs/api/xontribs_meta.rst
scopatz: Yep! See other inline comment

xonsh/xonfig.py
jnoortheen: Thanks updated.

docs/api/xontribs_meta.rst
jnoortheen: I have updated the `index.rst` but I am not sure if that is the correct way.

xonsh/xontribs_meta.py
jnoortheen: removed the types. The attributes are already annotated.

scopatz:
Great! It seems like it works.  Can you also please update the xontrib tutorial page: https://xon.sh/tutorial_xontrib.html Thanks @jnoortheen!

jnoortheen:
> Great! It seems like it works. Can you also please update the xontrib tutorial page: https://xon.sh/tutorial_xontrib.html Thanks @jnoortheen!

updated. I hope it is informative.

scopatz:
Thanks @jnoortheen!

PR #3973, opened 2020-11-11
chore: add pip cache for github actions

jnoortheen:
this will further speedup the workflow.
- also the macos and windows should now have proper cache with conda


.github/workflows/genbuilds.py
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env xonsh
```
scopatz, line 1: Why translate from xonsh?

jnoortheen: wanted to autoformat and lint.

scopatz: This is not blocking, but I would prefer that we keep as much in xonsh as possible.

.github/workflows/genbuilds.py
```diff
@@ -0,0 +1,49 @@
...
+import os
+import jinja2
+from pathlib import Path
+
+CURR_DIR = Path(__file__).parent
+environment = jinja2.Environment(
+    loader=jinja2.FileSystemLoader(CURR_DIR),
+    trim_blocks=True,
+    variable_start_string="<%",
+    variable_end_string="%>",
```
scopatz, line 16: Why not use standard Jinja2 here?

jnoortheen: the usual ${{ }} brace is used by github actions

.github/workflows/genbuilds.py
jnoortheen: Ok do you want to rename to .xsh ? here is not any benefit using it though.

scopatz: No, its fine as is.

.github/workflows/genbuilds.py
scopatz: Doesn't Jinja use just `{{}}`  I thought there was a way to escape this for literal `{{` and `}}`

scopatz: Yeah, I think either strings or raw can be used: https://stackoverflow.com/a/55464045

so `{% raw %}{{{% endraw %}` or `{{ "{{" }}`

jnoortheen: raw is better than quoting. wouldn't be confusing what is getting rendered by the script?

scopatz: I don't think so.  I think my issue is that I do so much jinja2 in my life that I am never going to remember alternative openers and closers 🤣 .

jnoortheen: Oh that :)

i was considerate about curly braces since there are more of Github than what is replaced by context.

jnoortheen: mako for jinja :)

jnoortheen: there are 13 `${{ ` . If you still prefer not to use custom tags,then I will change it.

scopatz: Yeah, I think for (my) sustainability I would prefer if this was in standard Jinja

jnoortheen: updated

gforsyth:
Thanks for putting this in @jnoortheen !

PR #3974, opened 2020-11-11
chore: move from travis to ga

jnoortheen:

### DIFF
--- a/.github/workflows/docs-linux-3.8.yml
+++ b/.github/workflows/docs-linux-3.8.yml
@@ -1,0 +1,51 @@
+name: docs linux 3.8
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ ubuntu-latest ]
+        python-version: [ 3.8 ]
+    name: Python ${{ matrix.python-version }} ${{ matrix.os }}
+    steps:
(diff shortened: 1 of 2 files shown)


scopatz:
CC @gforsyth

gforsyth:
Thanks for putting this in @jnoortheen -- `doctr` is made specifically to work with Travis (it's name is short for DOCs on TRavis).
There are, I think, three options here:
1. Roll our own docs build + push in github actions
2. Update / wait for an update to `doctr` to work with github actions
3. Hope that 1000 hours of travis build time lets us build docs for a long time and then ask them for more.

I've love to update `doctr` for GA, because the whole purpose of it was to avoid option 1 (although at the time it was everyone rolling their own doc build+push on travis).  Realistically, I don't know when I'll have time for this and I imagine asmeurer is also pretty slammed

jnoortheen:
now in new PR the travis jobs are queued and takes long time to start. So I just wanted to move it to the GA fray as well.

jnoortheen:
deploying to GitHub pages is as simple as pushing the code right ? is that anything done more than that?

gforsyth:
> deploying to GitHub pages is as simple as pushing the code right ? is that anything done more than that?

Currently what our (very slow) travis job does is build the docs for current master and pushes it to a `dev/` subdirectory (see https://xon.sh/dev/contents.html) -- and when there is a release, it pushes that new release to the root of the gh-pages branch (which is what populates https://xon.sh/contents.html)

Issue #3988, opened 2020-11-20
ulimit builtin is inconsistent with bash


WouldYouKindly:
Hi!

`ulimit -n` in bash shows the correct number of max open files, while in xonsh it shows `unlimited`. 

Mac OS Catalina 10.15.7

## xonfig
<details>

```
$ xonfig
+------------------+-----------------+
| xonsh            | 0.9.24          |
| Python           | 3.7.7           |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | 3.0.8           |
| shell type       | prompt_toolkit  |
| pygments         | 2.3.1           |
| on posix         | True            |
| on linux         | False           |
| on darwin        | True            |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
| on jupyter       | False           |
| jupyter kernel   | None            |
+------------------+-----------------+
```

</details>

## Expected Behavior
```
$ bash
bash-3.2$ ulimit -n
256
bash-3.2$ ulimit -Sn -Hn
open files                      (-n) 256
open files                      (-n) 256
```

## Current Behavior

```
$ ulimit -n
unlimited
$ ulimit -Sn -Hn
open files                      (-n) 256
open files                      (-n) unlimited
```

### Traceback (if applicable)

<details>

```
traceback
```

</details>

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

laloch:
Hi @WouldYouKindly, thanks for reporting.
The `ulimit` builtin only displays what `resource.getrlimit` reports. So this may be a bug in the Python standard library. Can you please verify my assumption, that the incorrect value comes from `resource.getrlimit`?
```python
import resource; print(resource.getrlimit(resource.RLIMIT_NOFILE))
```

laloch:
You can also verify the correctness of the values using
```
$ launchctl limit maxfiles
```

dyuri:
Or it might be bash's "bug":
```
bash > ulimit -Hn -Sn
open files                      (-n) 1024
open files                      (-n) 1024
bash > ulimit -Hn
524288
bash > ulimit -Sn
1024
```

While in xonsh:
```
xonsh > ulimit -Hn -Sn
open files                      (-n) 524288
open files                      (-n) 1024
xonsh > ulimit -Hn
[...]

PR #3990, opened 2020-11-22
ulimit: Change default to access soft limits

laloch:
It makes more sense for `ulimit` builtin to manipulate "soft" limits by default, because those are the limits intended to be user-adjustable. It is also what bash does. Originally, I made `ulimit` to default to "hard" limits, because I was for some reason convinced that it's what bash does.

Relates to #3988


gforsyth:
Thanks for putting this in @laloch !

PR #3992, opened 2020-11-25
Fix the import in the branch colour override example.

colons:
i assume this module got renamed at some point, but the docs weren't updated to reflect that


scopatz:
Nice catch! Thanks!

Issue #3956, opened 2020-11-04
In prompt-toolkit Shift-Right (and other Shift + Cursor keys) does nothing in macOS Terminal 


halloleo:
When using prompt-toolkit `Shift-Right` is suppose to extend the selection in emacs-mode. At least it does so in xonsh on Linux (xterm) and of course in Emacs.

However in macOS default Terminal app  `Shift-Right` doesn't do anything.

Is this rather a prompt-toolkit problem? Maybe there is a custom fix possible in the .xonshrc file?

Here the xonfig:

<details>

```
+------------------+-----------------+
| xonsh            | 0.9.24          |
| Python           | 3.8.6           |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | None            |
| shell type       | prompt_toolkit  |
| pygments         | None            |
| on posix         | True            |
| on linux         | False           |
| on darwin        | True            |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
| on jupyter       | False           |
| jupyter kernel   | None            |
+------------------+-----------------+
```

</details>

scopatz:
This might be a xonsh problem. It probably could be fixed in a xonshrc, but the interface should be the same across platform.  Pull requests welcome!

halloleo:
I just found out this seems to be a problem with the vendored prompt-toolkit. When I install the current version prompt-toolkit 3.0.8 into my xonsh environment, everything works as expected. :-)

scopatz:
Ahh OK.. PRs to update the vended version welcome!

halloleo:
Well, teh issue is bettered when using PTK 3.0.8. How would I go about "vendoring" this into the xonsh project? Just copying the `site-packages/prompt_toolkit` files to `xonsh/vended_ptk/prompt_toolkit`?

gforsyth:
Hey @halloleo -- yes -- although if you are going to do that, can I suggest adding a directive to the `Makefile` that is similar to the one we use to (infrequently) update `ply`?  That way we can have a way of reliably repeating the process.

halloleo:
Ok, I had a look at the Makefile.

So, I don't _copy_ the prompt_toolkit-files into the xonsh tree, but rather _issue the following git command_ instead:

[...]

PR #3995, opened 2020-11-25
Vendore deps

jnoortheen:
- now it will be easy to upgrade vendored dependencies


scopatz:
Thanks for this massive effort @jnoortheen! I thought we wanted to get rid of vending anything but PLY and instead use the `xonsh[full]` install procedure. Maybe I missed a discussion though?

jnoortheen:
> Thanks for this massive effort @jnoortheen! I thought we wanted to get rid of vending anything but PLY and instead use the `xonsh[full]` install procedure. Maybe I missed a discussion though?

Yes that is good. Is there any timeline/version planned to drop vended_ptk ?

scopatz:
I think whenever someone wants to put in a PR and update the docs :wink:

jnoortheen:
OK then I will close this PR and will open one to remove vended_ptk from the xonsh source. 

One other thing, I want to vendor my lib https://github.com/jnoortheen/arger within xonsh to make redundant alias functions/CLIs to be more pleasant to write and type check with mypy . Also this will enable us to generate autocompletions from the parsers itself, instead of updating manually. what do you think ? I made it for this purpose without depending on 3rd party libs.

gforsyth:
I think I missed a few conversations -- we recently started including PTK in `xonsh` itself -- I don't have strong feelings one way or the other, but if we're now going to remove it (or not!) we should probably stick to that decision for a bit.  

Also pinging @halloleo since they were also looking at updating the vended version of `xonsh`

jnoortheen:
Yes I am also fine with either cases. This is just to make the vendoring painless. If we are going to keep it vendored, then it should be easy to upgrade to new versions.

halloleo:
I probably don't understand the `git` mechanics correctly, but I though in the `Makefile` is so far _no_ entry for PTK, which (as far as I can see) means PTK is vendored by copying in, not automaticlly via a `git pull`.

Anyway, do these efforts here make vendoring PTK via  #3956 obsolete?

jnoortheen:
The idea is to have an easy way to upgrade to vendored dependencies and pin the versions. If you change the deps.txt to 3.0.3 and rerun the command, there will not be any change at all.

The Makefile has ply vendoring not for ptk.

halloleo:
Thanks @jnoortheen. Ok, so the vendoring of PTK is done through the new `task.xsh` file, while the vendoring of PLY is done through the Makefile, right? Plus PLY is included via git's `subtree` command, but PTK is included via `pip`. 

Looks a bit like a mess to me, but maybe this is the first step to moving everything over to `task.xsh`? I'm not an expert in git, so I don't know whether subtree-ing is better or worse than including pip'ed (i.e. copied) files...

jnoortheen:
> Thanks @jnoortheen. Ok, so the vendoring of PTK is done through the new `task.xsh` file, while the vendoring of PLY is done through the Makefile, right? Plus PLY is included via git's `subtree` command, but PTK is included via `pip`.
> 
> Looks a bit like a mess to me, but maybe this is the first step to moving everything over to `task.xsh`? I'm not an expert in git, so I don't know whether subtree-ing is better or worse than including pip'ed (i.e. copied) files...

Yes you got it right. 

PLY is another matter. Its pypi package doesn't have all the files. Also I dont think it will receive any future release. Last one was in [2018](https://pypi.org/project/ply/) . The author has created a new library [sly](https://github.com/dabeaz/sly) though.

jnoortheen:
@gforsyth can we decide on this? I favour removing the vendored ptk. But was there any reason to add it in the first place?

gforsyth:
Also, though, I think if we have a reason to vend other dependencies in the future (although I would like to avoid that) that the `task.xsk` laid out here seems like a nice way to do that.  

`ply` as it exists today is intended to be a vended dependency.

jnoortheen:
@gforsyth  Ok then I shall proceed closing this and create a PR to remove the `vended_ptk`?

gforsyth:
> @gforsyth Ok then I shall proceed closing this and create a PR to remove the `vended_ptk`?

Yeah, let's do that and if someone screams we can reconsider.

PR #3998, opened 2020-11-28
chore: bring back codcoverage report

jnoortheen:
Also fix `run-tests.xsh test` accepting pytest arguments

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -22,4 +22,4 @@ exclude_lines =
 skip_covered = true
 skip_empty = true
 show_missing = true
-sort = Miss
+sort = Cover
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,7 +4,22 @@ source =
   xonsh/
   xontrib/
 omit =
+  .venv/*
   */__amalgam__.py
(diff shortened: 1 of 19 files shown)


jnoortheen:
For some reason the coverage makes the tests slower inside CI machine. In my system it doesn't had such significant effect.

jnoortheen:
> For some reason the coverage makes the tests slower inside CI machine. In my system it doesn't had such significant effect. Maybe I will try without conda and directly use the system's interpreter. Let's see if that brings any speedup. Here is the capture of current timings
> image

seems like there is not much room to improve with setup-python

jnoortheen:
@gforsyth @scopatz any blockers ?

gforsyth:
I was trying to figure out who removed code coverage in the first place so I could check why.

Turns out it was me 4 years ago 😆 -- anyway, then it was a bunch of noise because it never worked properly and no one wanted to get it fixed up.  Looks like you have, @jnoortheen so I'm happy to merge this in.

Thanks for getting it working!

jnoortheen:
the anti-climax moment :)

Issue #3977, opened 2020-11-16
%autoreload like functionality for rapid prototyping


naefl:
Thanks for the fantastic work.

It's very likely that I'm just not good enough in searching issues here/missing the correct keywords so apologies if that's the case in advance.

The one feature that keeps me from switching to xonsh from ipython is the availability of a %autoreload like functionality when using it as a REPL when developing, that recursively imports the modules and its dependencies.

Is there such a function available/in planning?

Thanks!

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

gforsyth:
Hey @naefl -- thanks for raising this.  We don't currently have an `autoreload`, although it should be possible to grab it from `IPython` and package it up as a `xontrib`.

In the interim, you can do

```python
import importlib
importlib.reload(module_name)
```

for _some_ of that kind of functionality (although not "auto")

scopatz:
PRs welcome! I think this would make a nice built-in xontrib

naefl:
Thanks for the pointers @scopatz / @gforsyth I'll have a look at it

gforsyth:
Awesome! Let us know if you have any questions (and please feel free to open "half-finished" PRs if you need pointers)

PR #4002, opened 2020-12-02
feat: add xontrib-commands

jnoortheen:
not exactly the same. but a small start
#3977 

Let me know if you want to add this to xonsh itself. I wanted to use nice rich.Table . I know you would be against using 3rd party libs. So here it is .


gforsyth (approved):
Sweet!  I really like that `Command` decorator!

PR #4010, opened 2020-12-10
fix: make sure that tokens are strings before showing in prompt

jnoortheen:
- happens when the prompt field returns non-strings
- not including news item since it is a small change no significant change and no reported issues


gforsyth (approved):

gforsyth:
thanks for putting this in @jnoortheen !

Issue #4008, opened 2020-12-09
rich's console.print not working in interactive mode


jnoortheen:
rich - https://github.com/willmcgugan/rich

## xonfig

<details>

```
+------------------+----------------------+
| xonsh            | 0.9.24               |
| Git SHA          | 74543ea9             |
| Commit Date      | Oct 10 15:12:47 2020 |
| Python           | 3.8.6                |
| PLY              | 3.11                 |
| have readline    | True                 |
| prompt toolkit   | 3.0.8                |
| shell type       | prompt_toolkit       |
| pygments         | 2.7.3                |
| on posix         | True                 |
| on linux         | True                 |
| distro           | manjaro              |
| on darwin        | False                |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
| on jupyter       | False                |
| jupyter kernel   | None                 |
| xontrib 1        | abbrevs              |
| xontrib 2        | argcomplete          |
| xontrib 3        | autovox              |
| xontrib 4        | back2dir             |
| xontrib 5        | broot                |
| xontrib 6        | cmd_done             |
| xontrib 7        | commands             |
| xontrib 8        | hist_navigator       |
| xontrib 9        | jedi                 |
| xontrib 10       | kitty                |
| xontrib 11       | notifypy             |
| xontrib 12       | pdb                  |
| xontrib 13       | prompt_ret_code      |
| xontrib 14       | vox                  |
| xontrib 15       | voxapi               |
+------------------+----------------------+
```

</details>

## Expected Behavior
expected to print the colored text 
![image](https://user-images.githubusercontent.com/6702219/101597999-7e28d580-3a1d-11eb-9fd6-e9782a4aa97f.png)

## Current Behavior

Running this interactively 
```
$from rich import print
$print("[italic red]Hello[/italic red] World!", locals())
```
raises error

### Traceback (if applicable)

<details>

```
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/home/noor/.config/xonsh/xsh-src/xonsh/base_shell.py", line 362, in default
    run_compiled_code(code, self.ctx, None, "single")
  File "/home/noor/.config/xonsh/xsh-src/xonsh/codecache.py", line 67, in run_compiled_code
[...]

PR #4013, opened 2020-12-12
feat: some python libraries depend on isatty to check for terminal ca…

jnoortheen:
…pabilities

related to #4008

### DIFF
--- a/news/delegate-isatty.rst
+++ b/news/delegate-isatty.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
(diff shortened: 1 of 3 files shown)


jnoortheen:
image

the same seems to be the case for alias functions as well. That too has a custom std wrapper. There are two things, handle and default while default seems to be returning correct isatty result.

jnoortheen:
I don't know the implementation detail. So please review this more carefully eventhough it is a small change.

jnoortheen:
@gforsyth could you check this?

gforsyth:
Thanks @jnoortheen !  I think this is a reasonable change, but if other things break, we should revisit

jnoortheen:
Yes sure

PR #4017, opened 2020-12-13
Completion context

daniel-shimon:
# TL;DR

When (tab) completing, parse the command-line for a context object that all completers can use.

# Why?

Currently, each completer receives the current line and needs to parse it by itself.
There are a few drawbacks to this approach:

* Some do a worst job than others - for example, the completer for the `completer` command used to crash if the cursor was inside a word.
* It doesn't handle non-trivial cases very well - for example, `pip | grep --` will try to complete for the whole line, and then `complete_skipper` will try to re-complete only with the second command. This has its own drawbacks:
  * It isn't robust - try `pip; grep --`, `pip $(echo --`, `man @($(echo)`, etc.
  * Makes redundant completions for the complete line.
* The completers don't always handle string literals correctly - try `python "--"`, `"bash" --`.
* The completers can't access the entire multiline command - try completing: 
```bash
man \
pip --
```
* Overhead - almost every completer runs `.split()` and other similar parsing logic on the input.
* Consistency - `completer_path` can handle partial strings and append to an existing string. This logic can be used throughout the completion system.

# How?

This PR creates a PLY parser that creates a `CompletionContext` object.
It's tolerant to partially typed commands like `ls $(echo`, `ls $(echo))`, etc.

For each completer, using the completion context is **opt-in**.
Instead of:
```python
def complete_completer(prefix, line, start, end, ctx):
    """
    Completion for "completer"
    """
    args = line.split(" ")
    if len(args) == 0 or args[0] != "completer":
        return None
```
The completer does:
```python
@contextual_completer
def complete_completer(context: CompletionContext):
    """
    Completion for "completer"
    """
    if not isinstance(context, CommandContext) or len(context.args) == 0 or context.args[0].value != "completer":
        return None
```

# Design

See `xonsh/parsers/completion_context.py` for the `CompletionContext` objects, and `tests/test_completion_context.py` for more examples.

```python
parse("whic", cursor_index=4) == CommandContext(args=(), arg_index=0, prefix="whic")
parse("pip install --", cursor_index=14) == CommandContext(args=(CommandArg(value="pip"), CommandArg(value="install")), arg_index=2, prefix="--")
```

# Todo

- [X] Create API for contextual completers
- [X] Allow contextual completers in all shells (ptk, readline, jupyter)
- [X] Enable a tolerant lexer
- [X] Parse simple  commands
- [X] Parse subcommands (`$()`, `![]`, etc)
- [x] Parse partial strings (namely triple-quotes strings)
- [x] Parse multiple commands (`|`, `and/&&`, `or/||`, `;`, `\n`)
- [x] Parse python context (`@()`)
- [x] Handle appending to a closed quoted string (`ls "/usr/"`)
- [x] Make built-in completers contextual
  - [x] Unify `complete_python` and `complete_python_mode`
  - [x] `complete_base`
  - [x] `complete_completer`
  - [x] `complete_skipper`
  - [x] `complete_pip`
  - [x] `complete_cd`
  - [x] `complete_rmdir`
  - [x] `complete_xonfig`
  - [x] `complete_xontrib`
  - [x] `complete_from_bash`
  - [x] `complete_from_man`
  - [x] `complete_import`

# Ideas for follow-up PRs

* Create unified completion code for ordinary programs
* Register command completers in a dictionary for faster completions
* Extract partial string logic from `complete_path` to the contextual completion system
  * Escaping the completion if needed (`cat 'one file'`)
  * Evaluating an escaped prefix
* Parse expression in f-string
* Make `complete_path` contextual (this is a bit complex)


daniel-shimon:
For anyone reviewing this, I recommend going per-commit. most of them are small and simple

xonsh/completers/tools.py
```diff
@@ -1,6 +1,9 @@
 """Xonsh completer tools."""
 import builtins
 import textwrap
+from typing import Callable, Optional, Set, Union
```
jnoortheen: the convention is to use `import typing as tp`

gforsyth:
Hey @daniel-shimon !  I see you've checked off all the boxes up at the top here.  Let me know when you're ready for an in-depth review (I've been keeping an eye on the changes and things look good!)

daniel-shimon:
Hi @gforsyth! That's great!
I'll make some finishing touches and ping you when I'm done

daniel-shimon:
Hi @gforsyth !

This PR is done! 🔥 💥 
If you go by commits, note that some over-complicated parenthesis parsing rules get simplified a bit later on (at eea2b868 and beyond)
Feel free to ask questions here/on gitter/by email anytime!

.gitignore
```diff
@@ -10,6 +10,8 @@ parser_table.py
 parser_test_table.py
 xonsh/lexer_table.py
 xonsh/parser_table.py
+xonsh/completion_parser_table.py
+xonsh/parsers/completion_parser_table.py
 tests/lexer_table.py
```
jnoortheen, line 15: is it possible to fix the generated parser table to single location ? I too get this issue when I try to install from source. I have to copy the generated parser_table to the root directory.

gforsyth: Yes, but we should handle that in a separate PR.  It's probably something we missed in the `setup.py` file

gforsyth:
Thanks for the huge lift here @daniel-shimon !  I'm still going through review because, you know, this is huge, but one thing I've run across in usage that will need fixing.

```
(base)   ~/tmp 🐚  tree
.
├── ok
│  └── few
│     └── dirs
│        └── here
└── space out
```

because `space out` has a space in it, all of the path completer options get wrapped in quotes, but then when tabbing deeper in the `ok` directory, the prefix len is off by 1 and takes over the space between `cd` and `'ok'`

e.g. `cd 'ok/'<TAB>` -> `cd'ok/few/'`

daniel-shimon:
@gforsyth Great catch!
The `path` completer returns a correct prefix len, I've just failed to preserve it in this scenario where the cursor is after a closing quote (fix in 25ff18b):
`cd "ok/"<TAB`

I've also noticed a bug with the bash completer and added a fix

xonsh/completers/init.py
```diff
@@ -21,10 +20,10 @@ def default_completers():
     """Creates a copy of the default completers."""
     return collections.OrderedDict(
         [
-            ("python_mode", complete_python_mode),
             ("base", complete_base),
             ("completer", complete_completer),
             ("skip", complete_skipper),
+            ("end_proc_tokens", complete_end_proc_tokens),
```
gforsyth, line 26: Should this be a separate completer? Or should it be part of the general completion framework?

gforsyth: And does the completer ordering have any impact on how this executes?

xonsh/completers/pip.py
```diff
@@ -24,33 +26,41 @@ def ALL_COMMANDS():
...
-    if (line_len > 1 and line.endswith(" ")) or line_len > 2:
-        # "pip show " -> no complete (note space)
-        return
-    if prefix not in ALL_COMMANDS:
-        suggestions = [c for c in ALL_COMMANDS if c.startswith(prefix)]
+            return None
+        packages = (
+            line.split(maxsplit=1)[0] for line in enc_items.decode("utf-8").splitlines()
+        )
+        return {package for package in packages if filter_func(package, prefix)}
```
gforsyth, line 58: `pip show ` also shows me an entry for `----------` that should be caught by the `filter_func` but isn't for some reason.

Also, `pip show<tab>` isn't inserting a space, which I would expect

gforsyth (commented):
Hey @daniel-shimon -- this is looking good, I have a few small questions and a couple of small tweaks.
Also, not a blocker for this PR, but we'll want to add/update the docs on writing custom completers

gforsyth:
Two small nits I've run across (at least one of them may be pre-existing)

The pre-existing one (I think): `~` expands but isn't removed from the line so we're left with garbage
```
$ ls tmp/
test
$ pwd
/Users/gil
$ vim ~/tmp/t<TAB> --> vim ~/tmp/t/Users/gil/tmp/test
```

New issue:

The path completer shows the full path for files, but strips off the parent folder for nested directories
```
$ ls -F tmp
mock/  ok/    test  test2
$ vim tmp/<TAB>
```

displays this list of options (sorry, can't screenshot on work laptop)
```
mock/
ok/
tmp/test
tmp/test2
```

e.g. `tmp` isn't stripped from the two non folder entries.

Weirdly, if there's a directory with a space in there, it goes away:
```
$ ls -F tmp
mock/      ok/        space out/ test       test2
$ vim tmp/<TAB>
```

-->

```
mock/
ok/
space out/
test
test2
```

xonsh/completers/tools.py
daniel-shimon: This file is heavily typed, so I rather keep it like this for readability

xonsh/completers/pip.py
```diff
@@ -5,16 +5,18 @@
...
 def PIP_RE():
-    return re.compile(r"\bx?pip(?:\d|\.)*\b")
+    return re.compile(r"\bx?pip(?:\d|\.)*$")
 
 
 @xl.lazyobject
-def PIP_LIST_RE():
-    return re.compile(r"\bx?pip(?:\d|\.)*\b (?:uninstall|show)")
+def PIP_LIST_COMMANDS():
+    return {"uninstall", "show"}
```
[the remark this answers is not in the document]
daniel-shimon: Done

xonsh/completers/init.py
daniel-shimon: It needs to be a regular completer in the system to see the token.
Because `complete_base` will now work for stuff like `ls &&<TAB>`, we need to put it as the first completer, so I moved it now to be the first.

xonsh/completers/pip.py
daniel-shimon: These two also exist before this PR, but I've added a fix for both (and tests!)

daniel-shimon:
> Two small nits I've run across (at least one of them may be pre-existing)

@gforsyth I've answered all of your code comments, but I can't reproduce both these issues

gforsyth:
Hey @daniel-shimon -- hmm, yeah, I'm still seeing them on this branch and not on `main`.

another thing to check is whether python modules show the full path or just the completion bit (e.g. if common prefixes are being removed)

pr/4017
```
$ import pathlib
$ pathlib.<TAB>
pathlib.attrgetter(
pathlib.EBADF
...
```

main
```
$ import pathlib
$ pathlib.<TAB>
attrgetter(
EBADF
...
```

daniel-shimon:
@gforsyth OK I think I found the issue and fixed it - see a999d98
Can you see whether this maybe fixes the other two issues?

gforsyth:
Hey @daniel-shimon -- getting back to this now.
The previous issues do seem fixed (yay!) -- the `~` expansion is still a problem but it's also an issue on `main` so we can fix that later.

Latest thing I've just run into some (corner case) issues with the `end_proc_tokens` -- I have a directory named `org/` 

If I `cd o<TAB>` it completes to `cd org/` as expected.  If I `cd or<TAB>` then the `end_proc_tokens` reads it as an `or` statement and adds a space to the end.

This also happens if you have a directory named e.g. `andiron`

rsteube:
Does this handle raw strings correctly? Just realized xonsh has these and partial completion of these fails in my script due to shlex.
E.g.:
```sh
command r'partial<TAB>
# command rpartial
```

daniel-shimon:
> Does this handle raw strings correctly?
@rsteube Yes
I'll be glad if you could checkout this branch and test it for yourself!

@gforsyth
> the ~ expansion is still a problem but it's also an issue on main so we can fix that later.
I'll look into it :) 

> If I cd or<TAB> then the end_proc_tokens reads it as an or statement and adds a space to the end.
Yeah I guess the correct fix will be to split this into two completers:
1. Completes non-arguments like `;` and `&&`, and will be the first completer (before `base_completer` can complete a command)
2. Completes `and` and `or`, and will be the last completer so other completers will have a chance (e.g. `complete_path`)
I'll get it done 👍

gforsyth:
> 2\. Completes `and` and `or`, and will be the last completer so other completers will have a chance (e.g. `complete_path`)
>     I'll get it done 👍

Awesome!  Before you go too far down the rabbit hole -- the current completion system allows for displaying completions from multiple completers, since there are ambiguous situations (like the `or` example above).  I think that for `cd or<TAB>` that offering both `or ` and `org/` as options is reasonable -- are we losing the combinatorial approach here?

daniel-shimon:
@gforsyth I'm not familiar with a way to allow multiple completions in the current system.
Anyway I think splitting it up is reasonable and added the changes needed.
Can you see that this works for you now?

> are we losing the combinatorial approach here?

We might. Maybe specifically in `complete_path` we should include completions for `or` and `and`?

gforsyth:
Hey @daniel-shimon -- sorry I dropped off last week.  I think this is overall a huge improvement over the existing completion infrastructure (not to mention having solid test coverage) so I'm going to merge it in and then we can fix and improve as we go.

Thanks so much for this massive undertaking!

gforsyth (approved):

PR #4032, opened 2020-12-25
refactor: convert prompt_ret_code to be a py file

jnoortheen:
to take advantage of flake/coverage

not including a news item as it is just a refactoring of internal code.


gforsyth:
@jnoortheen -- the code here looks fine, I'm just not sure that coverage / flake is a reason to rewrite something.  Especially when the file in question is written in the language that we're working on.  

Any thoughts from @xonsh/xore ?

jnoortheen:
not only that, the parsing time is also a concern.

gforsyth:
> the code should be maintainable

If xonsh code isn't maintainable we should all just hang up our hats and go home.

> the parsing time is also a concern

if you have numbers on this I'm interested

jnoortheen:
I said maintainable in the sense that tooling support like flake8, mypy, black, IDE etc.,
xonsh is great as a bashish python. Just a hitch editing/formatting and such. 

> if you have numbers on this I'm interested

sometime back I measred the xontrib load time for a .xsh file and .py file. Maybe I will test and update here if got time.

jnoortheen:
here is the timing test result

pure Python code
image

here is the xonsh one
image

only the first import time matters here. As you could see it is taking longer for xonsh files. 

I tried to cythonize ply package into .so. But that doesn't bring any speedup as well. I did the cython/nuitka compilings when chekcing the setup.py changes PR. There wasn't any difference between the compiled `.so` and plain `.py` versions.

gforsyth (approved):
thanks for putting this in @jnoortheen !

PR #4041, opened 2020-12-31
Add a better test suite for "vc.py", and fix "include_untracked" logic

dannysepler:
@nh2 had some good notes on this PR https://github.com/xonsh/xonsh/pull/3780 -- ultimately I think I don't know the plumbing commands well enough to use them yet, so I reverted back to `git status --porcelain`. I would have caught my syntax error if we had more testing, so I broke vc tests out into `test_vc.py` and added a few tests to it.

Also, "include_untracked" wasn't reachable at all (which is why we didn't notice it was broken). So, I added a `VC_GIT_INCLUDE_UNTRACKED` environment variable so that people can use it if they want.


tests/prompt/test_vc.py
```diff
@@ -0,0 +1,130 @@
...
+from xonsh.prompt import vc
+from tools import DummyEnv
+
+
+# Xonsh interaction with version control systems.
+VC_BRANCH = {"git": {"master", "main"}, "hg": {"default"}}
+
+
+@pytest.fixture(params=VC_BRANCH.keys())
+def test_repo(request, tmpdir_factory):
```
dannysepler, line 20: two notes here:

1. Removed the "module" scope, because if a certain test adds a file, I didn't want that to affect other tests
2. Using pytest's `tmpdir` functions here and elsewhere instead of tempfile since (I think?) they should better cleanup things after the test finishes

tests/prompt/test_base.py
```diff
@@ -186,95 +180,3 @@ def test_promptformatter_clears_cache(formatter):
     formatter(template, fields)
 
     assert spam.call_count == 2
-
-
-# Xonsh interaction with version control systems.
```
AsafFisher, line 191: maybe make the moment of this huge block a separate commit?

tests/prompt/test_vc.py
```diff
@@ -0,0 +1,130 @@
+import os
```
dannysepler, line 1: okay i juggled the commits around and this way should preserve git history. github isn't really showing it, but you can see the git history if you click "View file" and then "Blame", like [here](https://github.com/xonsh/xonsh/blame/52338336cc369be3904bd2927ef129f0dad61410/tests/prompt/test_vc.py)

gforsyth:
Hey @dannysepler -- thanks for putting this in!  This looks good to me, but conflicts with the environment grouping that I just merged in.  Can you pull in the latest changes from master?  Should just be a minor reshuffle of how you add the new untracked env-var.

xonsh/environ.py
```diff
@@ -1432,6 +1432,13 @@ def DEFAULT_VARS():
             "branch computations. This is a timeout per subprocess call, so the "
             "total time to compute will be larger than this in many cases.",
         ),
+        "VC_GIT_INCLUDE_UNTRACKED": Var(
+            is_bool,
+            to_bool,
+            bool_to_str,
+            False,
```
jnoortheen, line 1439: can we have this default as True here? to retain the current behaviour

dannysepler: hm good point. i thought it was `False` by default? just looking at the original `git_dirty_working_directory` args, btw

gforsyth: Yeah, the existing default is to not include untracked files

gforsyth (approved):
Woohoo!  This looks great!   Thanks @dannysepler !

Issue #4048, opened 2021-01-05
Uptime function on FreeBSD broken


yggdr:
## xonfig

<details>

```
$ xonfig
+------------------+-----------------+
| xonsh            | 0.9.24          |
| Python           | 3.8.6           |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | None            |
| shell type       | prompt_toolkit  |
| pygments         | None            |
| on posix         | True            |
| on linux         | False           |
| on darwin        | False           |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
| on jupyter       | False           |
| jupyter kernel   | None            |
+------------------+-----------------+
```

</details>

## Expected Behavior
Xonsh starts without error messages.

## Current Behavior
Xonsh starts with an error caused by the `xoreutils.uptime.uptime` function.

### Traceback (if applicable)

<details>

```
❯ env XONSH_DEBUG=1 xonsh
Exception in thread Thread-2:
Traceback (most recent call last):
  File "/home/yggdrasil/.asdf/installs/python/3.8.6/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/home/yggdrasil/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/history/json.py", line 137, in run
    files = self.files(only_unlocked=True)
  File "/usr/home/yggdrasil/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/history/json.py", line 180, in files
    boot = uptime.boottime()
  File "/usr/home/yggdrasil/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/xoreutils/uptime.py", line 280, in boottime
    up = uptime()
  File "/usr/home/yggdrasil/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/xoreutils/uptime.py", line 265, in uptime
    or _uptime_linux()
  File "/usr/home/yggdrasil/.local/pipx/venvs/xonsh/lib/python3.8/site-packages/xonsh/xoreutils/uptime.py", line 56, in _uptime_linux
    if xp.LIBC.sysinfo(buf) < 0:
AttributeError: 'NoneType' object has no attribute 'sysinfo'
[...]

PR #4049, opened 2021-01-05
Fix libc detection and uptime functionality on FreeBSD

yggdr:
This is a possible fix for the libc detection bug on Freebsd, as discussed in #4048.

- Fix libc detection on FreeBSD
- sysctlbyname needs bytes, not unicode strs
- Add news item for FreeBSD libc fix

### DIFF
--- a/news/fix_freebsd_libc_uptime.rst
+++ b/news/fix_freebsd_libc_uptime.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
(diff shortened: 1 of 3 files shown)


gforsyth:
@jnoortheen -- can you look at this and figure out why codecov is failing?  If it's going to mark entire PRs as failures for some minuscule change in test coverage I'm going to rip it out.

jnoortheen:
I think it is related to codecov config. If it is 0% I think it is giving error. I am checking.

gforsyth:
Thanks for putting this in @yggdr!

It doesn't break on Linux or OSX and I'm assuming since you put it up that it's working with BSD, so I'm going to merge.
If you have any thoughts / experience with BSD CI and want to move that effort along, I'm happy to help if I can.

PR #4051, opened 2021-01-06
Update codecov.yml

jnoortheen:


jnoortheen:
@gforsyth I think this will do for codecov fail. It is not related to pytest-coverage but with codecov service.

gforsyth:
Awesome, thanks @jnoortheen !

PR #4052, opened 2021-01-09
Update tutorial_ptk.rst

farisachugthai:
Using `_` to indicate italics only works in Markdown. Use `*` to modify text in restructuredText


gforsyth (approved):
Thanks for the fix @farisachugthai !

Issue #2418, opened 2017-06-30
HISTCONTROL ignorespace


nh2:
Many shells [have a feature](
https://unix.stackexchange.com/questions/115917/why-is-bash-not-storing-commands-that-start-with-spaces) that you can set `HISTCONTROL` to `ignorespace` to prevent commands starting with a space from going into history.

That is so that you can conveniently set an environment variable containing secret information that shouldn't go into history without having to resort to something like `getpass()`, or so that you can run a command that you don't want in the history to prevent yourself from accidentally running it again later, such as ` rm -rf *`.

nh2:
I took a quick look at what would be needed for this, and the logic change is trivial, like:

```diff
diff --git a/xonsh/history/json.py b/xonsh/history/json.py
index 9c11babe..d6eca11d 100644
--- a/xonsh/history/json.py
+++ b/xonsh/history/json.py
@@ -201,6 +201,9 @@ class JsonHistoryFlusher(threading.Thread):
             if 'ignoreerr' in opts and cmd['rtn'] != 0:
                 # Skipping failed cmd
                 continue
+            if 'ignorespace' in opts and cmd['inp'].startswith(' '):
+                # Skipping cmd starting with space
+                continue
             cmds.append(cmd)
             last_inp = cmd['inp']
         with open(self.filename, 'r', newline='\n') as f:
diff --git a/xonsh/history/sqlite.py b/xonsh/history/sqlite.py
index 0446699e..4648d4da 100644
--- a/xonsh/history/sqlite.py
+++ b/xonsh/history/sqlite.py
@@ -202,6 +202,12 @@ class SqliteHistory(History):
         if 'ignoreerr' in opts and cmd['rtn'] != 0:
             # Skipping failed cmd
             return
+        print(type(cmd))
+        print(cmd)
+        raise 'here'
+        if 'ignorespace' in opts and cmd['inp'].startswith(' '):
+            # Skipping cmd starting with space
+            return
```

But the problem is that `cmd['inp']` for some reason already has leading whitespace stripped.

It seems that `xonsh/ptk/shell.py`'s `line = self.singleline(auto_suggest=auto_suggest)` already returns it stripped.

I haven't looked deeper yet.

hugo-aa:
Hi @nh2, thanks for your inputs here. I think this is a nice feature.

> the problem is that cmd['inp'] for some reason already has leading whitespace stripped.

[...]

PR #4062, opened 2021-01-21
Add $HISTCONTROL=ignorespace support

aneeshdurg:

### DIFF
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -20,7 +20,11 @@ from xonsh.history.main import history_main, _xh_parse_args, construct_history
 
 
 CMDS = ["ls", "cat hello kitty", "abc", "def", "touch me", "grep from me"]
-
+IGNORE_OPTS = ",".join([
+    "ignoredups",
+    "ignoreerr",
+    "ignorespace"
+])
 
 @pytest.fixture
 def hist():
@@ -94,12 +98,13 @@ def test_hist_flush_with_hist_control(hist, xonsh_builtins):
     """Verify explicit flushing of the history works."""
     hf = hist.flush()
     assert hf is None
-    xonsh_builtins.__xonsh__.env["HISTCONTROL"] = "ignoredups,ignoreerr"
+    xonsh_builtins.__xonsh__.env["HISTCONTROL"] = IGNORE_OPTS
     hist.append({"inp": "ls foo1", "rtn": 0})
     hist.append({"inp": "ls foo1", "rtn": 1})
     hist.append({"inp": "ls foo1", "rtn": 0})
     hist.append({"inp": "ls foo2", "rtn": 2})
     hist.append({"inp": "ls foo3", "rtn": 0})
+    hist.append({"inp": " ls secret", "rtn": 0})
     hf = hist.flush()
     assert hf is not None
     while hf.is_alive():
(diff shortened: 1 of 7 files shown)


xonsh/history/json.py
```diff
@@ -422,8 +425,19 @@ def append(self, cmd):
...
+        skipped_by_ignore_space = "ignorespace" in opts and cmd.get("spc")
+        if not skipped_by_ignore_space:
+            # Skipping command starting with space
+            self.buffer.append(cmd)
+            self._len += 1  # must come before flushing
+
+        try:
+            del cmd["spc"]
+        except KeyError:
+            pass
```
gforsyth, line 439: Any reason we need to try to boot this key?

xonsh/history/json.py
```diff
@@ -422,8 +425,19 @@ def append(self, cmd):
...
             return
-        self.buffer.append(cmd)
-        self._len += 1  # must come before flushing
+
+        opts = builtins.__xonsh__.env.get("HISTCONTROL")
+        skipped_by_ignore_space = "ignorespace" in opts and cmd.get("spc")
+        if not skipped_by_ignore_space:
+            # Skipping command starting with space
+            self.buffer.append(cmd)
+            self._len += 1  # must come before flushing
```
gforsyth, line 434: I think for consistency with the `remember_history` check and also the sqlite backend implementation, this should do an

```
if skipped_by_ignore_space:
    return
```

xonsh/base_shell.py
```diff
@@ -342,9 +343,14 @@ def precmd(self, line):
...
 
-    def default(self, line):
+    def default(self, line, raw_line=None):
         """Implements code execution."""
         line = line if line.endswith("\n") else line + "\n"
+        if not self.need_more_lines:  # this is the first line
+            if not raw_line:
+                self.src_starts_with_space = False
+            else:
+                self.src_starts_with_space = raw_line[0].isspace()
```
gforsyth, line 353: Two things
1. Does this need to be here?  Can we stick in in the `finally` block before we call to `self._append_history`?
2. It looks like this is set up to only work if the command is a single-line command -- any reason this wouldn't be wanted for a multi-line command?

gforsyth (commented):
Hey @aneeshdurg -- thanks for putting this in!  I can remember this being requested some time ago.

Overall this looks good to me, I have a few questions re: how it works and a few suggested tweaks.

xonsh/base_shell.py
aneeshdurg: I should probably write a test for the multi line case, but my understanding was that this does work for multiline comands. We only want to set this flag when we're processing the first line, since for subsequent lines, whitespace prefixes have no meaning. If this is the first line, we compute the state of this flag, otherwise we preserve the flag's current state.

gforsyth: Ahh, right.  Sorry, I was thinking of the prompt_toolkit backend, where the multiline input is single string, but for readline you'll need to check it the first time.

aneeshdurg: I looked into moving it into the `finally`, but I think that won't work for multi-line commands since the first line would need more input and return early.

xonsh/history/json.py
aneeshdurg: I felt it was weird to leave that key around in the history file that's written out.

anki-code:
> I can remember this being requested some time ago.

Here it is: #2418

tests/test_history.py
```diff
@@ -94,20 +95,21 @@ def test_hist_flush_with_hist_control(hist, xonsh_builtins):
...
+    hist.append({"inp": "ls secret", "rtn": 0, "spc": True})
     hf = hist.flush()
     assert hf is not None
     while hf.is_alive():
         pass
     assert len(hist.buffer) == 0
     with LazyJSON(hist.filename) as lj:
         cmds = list(lj["cmds"])
-        assert len(cmds) == 2
+        assert len(cmds) == 2, [x["inp"] for x in cmds]
```
gforsyth, line 112: leftover debug statement?

gforsyth (commented):
Hey @aneeshdurg -- this looks good to me!  There's one leftover assert followup that can be removed but otherwise this can go in.

gforsyth (approved):
Thanks so much for putting this in @aneeshdurg and congrats on your first contribution!

PR #4064, opened 2021-01-23
rework old pytest plugin

elbehery95:
Current pytest plugin has many limitations

1. Plugin will always collect ``test_*.xsh`` while pytest allows an ini configuration that canges the test collection prefix
2. Plugin does not support parametrization nor loading and using existing fixtures
3. Plugin is **always** activated (in some cases users using xonsh migh want to build their own collection plugin that can conflict with the base plugin shipped with xonsh)

I have re-worked the existing plugin to overcome these limitations. The new plugin now accepts ``--xonsh`` such that using this flag will allow pytest to load ``.xsh`` files. Typical use case would be having ``--xonsh`` hardcoded in user's ``pytest.ini``

I have not seen any tests for the existing pytest plugin. Not sure if such tests need to be added at the current support or not. Let me know if these are required.

Example for `pytest.ini`:

```ini
[pytest]
python_files = test_*.py test_*.xsh
```

Example for test file
```py
import pytest


@pytest.fixture(scope="function", autouse=True)
def set_env():
    with ${...}.swap(TESTVAR='foo'):
        yield



@pytest.mark.parametrize('a', [1, 2, 3, 4])
def test_echo(a):
    echo @(a)
    echo $TESTVAR
```

Examples for running this test:

```shell
$ pytest --collect-only -q
no tests collected in 0.01s

$ pytest --collect-only -q --xonsh
test_hello.xsh::test_echo[1]
test_hello.xsh::test_echo[2]
test_hello.xsh::test_echo[3]
test_hello.xsh::test_echo[4]

4 tests collected in 0.05s

$ pytest --xonsh --capture=tee-sys

=================================== test session starts ====================================
platform win32 -- Python 3.7.0, pytest-6.2.1, py-1.10.0, pluggy-0.13.1
rootdir: example, configfile: pytest.ini
plugins: cov-2.11.1, timeout-1.4.2, xonsh-0.9.24
collected 4 items

test_hello.xsh 1
foo
.2
foo
.3

foo
.4
foo
.                                                                   [100%]

==================================== 4 passed in 0.60s =====================================
```

Let me know your thoughts about this :)

### DIFF
--- a/news/pytest_plugin.rst
+++ b/news/pytest_plugin.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* More generic pytest plugin that allows utilizing fixtures, parametrization and more of pytest core features
+
+**Changed:**
+
+* Existing pytest plugin
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/xonsh/pytest_plugin.py
+++ b/xonsh/pytest_plugin.py
@@ -1,78 +1,42 @@
 # -*- coding: utf-8 -*-
 """Pytest plugin for testing xsh files."""
-import sys
-import importlib
-from traceback import format_list, extract_tb
-
-import pytest
+import _pytest.pathlib
 
 from xonsh.main import setup
 
 
+class XshPytestPlugin:
+    def __init__(self):
+        setup()
+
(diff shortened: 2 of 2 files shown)


elbehery95:
I have not seen any tests for the existing pytest plugin. Not sure if such tests need to be added at the current support or not. Let me know if these are required

gforsyth:
Hey @elbehery95 -- sorry I didn't have a chance to look at this yet.  Are you still interested in putting this contribution in?

PR #4025, opened 2020-12-19
Using ujson (if installed) in LazyJSON to loading json history 15% faster

anki-code:
Closes #4000

Test:
```python
pip uninstall ujson
XONSH_DEBUG=1 xonsh
# [history.json] Enumerated 2,852 history files for 0.7310s
# [history.json] Enumerated 2,852 history files for 0.7258s
# [history.json] Enumerated 2,852 history files for 0.6990s
# [history.json] Enumerated 2,852 history files for 0.7035s

pip install ujson
XONSH_DEBUG=1 xonsh
# [history.ujson] Enumerated 2,852 history files for 0.5832s
# [history.ujson] Enumerated 2,852 history files for 0.6084s
# [history.ujson] Enumerated 2,852 history files for 0.6103s
# [history.ujson] Enumerated 2,852 history files for 0.5903s
```

Issue #4000, opened 2020-11-30
Using fastest json library to increase the speed of json history


anki-code:
Hi! I am a snailger, And I ride, and I ride, I ride through the ocean backsides, I see the conchs come out of the sky.

What about using fastest json library to increase the speed of json history?

For example:

* orjson - https://github.com/ijl/orjson - the fastest Python library for JSON.
* ujson - https://pypi.org/project/ujson/ - UltraJSON is an ultra fast JSON encoder and decoder written in pure C with bindings for Python 3.6+.
* And the standard json as fallback.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

anki-code:
Load time on 2785 history json files:
* json - 0.62s
* ujson - 0.52s
* orjson - 0.50s

PR #4073, opened 2021-01-29
Added ujson to appimage

anki-code:
Added `ujson` to AppImage to support #4025 #4000
News is not needed.


gforsyth:
Thanks @anki-code !

PR #4075, opened 2021-01-31
Fix readme and rever test call for release

gforsyth:
The rever command hasn't been tested with the new parameters to
`run-tests.xsh` (spoiler alert, it breaks).

Also converted the `raw` block to regular RST since PyPI doesn't allow
`raw` directives in the `long_description` on upload to PyPI

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,34 +1,44 @@
 xonsh
 =====
 
-.. raw:: html
+.. class:: center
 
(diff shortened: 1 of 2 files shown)


gforsyth:
@xonsh/xore I could use a quick check on this

gforsyth:
Merging as this is blocking a release

Issue #3763, opened 2020-09-13
On Windows, xonsh can't find executables that are in my PATH


beetleb:
## xonfig

<details>

```
+------------------+-------------------------------------------------------+
| xonsh            | 0.9.21                                                |
| Python           | 3.8.5                                                 |
| PLY              | 3.11                                                  |
| have readline    | False                                                 |
| prompt toolkit   | 3.0.7                                                 |
| shell type       | prompt_toolkit                                        |
| pygments         | 2.6.1                                                 |
| on posix         | False                                                 |
| on linux         | False                                                 |
| on darwin        | False                                                 |
| on windows       | True                                                  |
| on cygwin        | <xonsh.lazyasd.LazyBool object at 0x0000022DBFC6DDF0> |
| on msys2         | <xonsh.lazyasd.LazyBool object at 0x0000022DBFC7D070> |
| is superuser     | False                                                 |
| default encoding | utf-8                                                 |
| xonsh encoding   | utf-8                                                 |
| encoding errors  | surrogateescape                                       |
+------------------+-------------------------------------------------------+
```

</details>

## Expected Behavior
It should execute the executable in the path.

## Current Behavior

It says it can't find it:

```
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "C:\Users\me\AppData\Roaming\Python\Python38\site-packages\xonsh\built_ins.py", line 561, in _run_binary
    p = self.cls(self.cmd, bufsize=bufsize, **kwargs)
  File "C:\Users\me\AppData\Roaming\Python\Python38\site-packages\xonsh\proc.py", line 584, in __init__
    self.proc = proc = subprocess.Popen(
  File "c:\program files\python38\lib\subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "c:\program files\python38\lib\subprocess.py", line 1307, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
FileNotFoundError: [WinError 2] The system cannot find the file specified

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\me\AppData\Roaming\Python\Python38\site-packages\xonsh\proc.py", line 1889, in __init__
    proc = spec.run(pipeline_group=pipeline_group)
[...]

PR #4077, opened 2021-02-01
Use full paths when launching processes on Windows

auneri:
The change is related to the [subprocess PATH semantics and portability](https://bugs.python.org/issue8557) and addresses issue #3763.

### DIFF
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -822,12 +822,12 @@ def test_exec_function_scope(cmd):
 
 
 @skip_if_on_unix
-def test_run_currentfolder():
+def test_run_currentfolder(monkeypatch):
     """Ensure we can run an executable in the current folder
     when file is not on path
     """
     batfile = Path(__file__).parent / "bin" / "hello_world.bat"
-    os.chdir(batfile.parent)
+    monkeypatch.chdir(batfile.parent)
     cmd = batfile.name
     out, _, _ = run_xonsh(cmd, stdout=sp.PIPE, stderr=sp.PIPE, path=os.environ["PATH"])
     assert out.strip() == "hello world"
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -5,6 +5,7 @@ import os
 import shutil
 import tempfile
 import subprocess as sp
+from pathlib import Path
 
 import pytest
 
@@ -13,6 +14,7 @@ from xonsh.lib.os import indir
 
 from tools import (
     skip_if_on_msys,
+    skip_if_on_unix,
     skip_if_on_windows,
     skip_if_on_darwin,
     ON_WINDOWS,
@@ -46,9 +48,13 @@ def run_xonsh(
     stderr=sp.STDOUT,
     single_command=False,
     interactive=False,
+    path=None,
 ):
     env = dict(os.environ)
-    env["PATH"] = PATH
+    if path is None:
+        env["PATH"] = PATH
+    else:
+        env["PATH"] = path
     env["XONSH_DEBUG"] = "0"  # was "1"
     env["XONSH_SHOW_TRACEBACK"] = "1"
     env["RAISE_SUBPROC_ERROR"] = "0"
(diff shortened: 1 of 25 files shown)


melund:
I wonder why I was never able to reproduce the issue for my self. But enough people have reported it... I think the fix looks good. 

Especially if it also smooth out some of the difference  between *nix and windows induced by popen. 

Can you also add a news item so the changelog gets generated correctly on next release.

auneri:
Thank you, added the news item. A few questions though:
1. Would it be better to apply the change for all systems (not just for Windows) for consistency?
2. Is it safe to modify `self.cmd`? I could avoid it if not.
3. Is this proper use of xonsh's `which` command or is there a better way?
4. Not sure why some tests are failing...

auneri:
See my questions above; is there anything else I can provide?

melund:
Sorry. I haven't had time. We definitely shouldn't change the behavior on other than Windows. I think it is OK to use which like you do. 

However, I took a quick look at the tests. Some seem related to this fix. So maybe there are some side effects to this fix. I guess we need to get some one else to take a look at this as well.

auneri:
Anything I can do to help move this forward?
From what I can tell, the failing test is trying to execute `cat tttt | wc` on Windows, which is failing because `cat` is not a valid command.

anki-code:
@auneri try to merge main branch into your branch.

auneri:
Thanks for the suggestion, hope I did it correctly (merged from `xonsh:main`). The Windows errors seem to be unchanged from before.

anki-code:
@auneri Looks like the tests that fail are related to your changes. Click to the "Details" link and fix it.

> which is failing because cat is not a valid command.

These tests were working before your changes, right? This means that the error is unrelated to `cat`.

auneri:
Thanks. I will dig a bit further, but FYI, I am not able to replicate the failures on my local Windows machine. The tests pass regardless of these changes.

auneri:
I was trying to use `xonsh.xoreutils.which` before, but switching to the standard `shutil.which` fixed the tests.

Note also that I specify the path via `builtins.__xonsh__.env.get("PATH")`, omitting that would require `$UPDATE_OS_ENVIRON=True` for paths to resolve properly.

auneri:
Let me know if there is anything I can do on my end. Would love to have this in the next release.

anki-code:
Hi @auneri!

1. We have `resolve_binary_loc` function near your changes which uses `commands_cache.locate_binary`. Could you please check this function and may be it what you need and could be reused?

2. It will be good to have at least one small test for this new logic.

Thanks!

auneri:
1. Alright, turns out the full binary path was already made available via `resolve_binary_loc`; latest change uses that.
2. I will need some guidance on testing. Here is the logic that is corrected via this change (#3763):
```python
python -c "import sys; print(sys.executable)"  # outputs C:\Miniconda3\envs\test\python.exe
$PATH.insert(0, r'C:\Miniconda3\envs\py37')
# before fix
python -c "import sys; print(sys.executable)"  # outputs C:\Miniconda3\envs\test\python.exe
# after fix
python -c "import sys; print(sys.executable)"  # outputs C:\Miniconda3\envs\py37\python.exe
```
Setting `$UPDATE_OS_ENVIRON = True` also corrects this behavior, but only for executables besides `python.exe`. The only way to fix launching correct `python.exe` is with this change.

anki-code:
Ordinary I go to `xonsh/tests` directory and find the appropriate file with tests for module I changed. Then I can run i.e.:
```python
pytest tests/procs/test_specs.py
```
After this I add the new tests or I review the tests around to understand more testing patterns to write new test.

auneri:
A new test in [test_specs.py](https://github.com/xonsh/xonsh/blob/main/tests/procs/test_specs.py) would be appropriate, but the only existing test there is not designed to run on Windows. Looking around a bit more, I was not able to identify a test that is similar to my case, with most tests skipping on Windows.

At this time, my day work is catching up with me and I'm afraid I will not be able to spend more time on this... I would greatly appreciate if someone picks up the mantle.

daniel-shimon:
Hey @anki-code, would you like to pick this up and add a test?

anki-code:
Sorry I'm not the windows user to properly test this )

melund:
I have testet this, and I certainly think we shoud merge it. @daniel-shimon /@anki-code Any objections?

daniel-shimon:
@melund it L9274 good but we need a test in the repo :)
Shouldn't be too complicated, just mock the PATH to a local directory and make sure it calls a correct binary inside

melund:
Do we have some related tests anywhere? I.e. where should we place it?

daniel-shimon:
Probably in the `proc`or `specs` test files

melund:
> Probably in the `proc` or `specs` test files

I ended up with adding them to `test_integrations`. I tried to follow the approach in `test_spec` but I kept getting deadlocks when using `run_subproc` directly. I probably haven't understod how it works directly.

Anyway, I think this should be just as fine. 

Now, I just need to figure out why tests are failing again :-/

daniel-shimon:
You should probably rebase in main first :)

tests/test_integrations.py
```diff
@@ -813,3 +819,36 @@ def test_loading_correctly(monkeypatch, interactive):
...
     assert rtn == 0
+
+
+@skip_if_on_unix
+def test_run_currentfolder():
+    """Ensure we can run an executable in the current folder
+    when file is not on path
+    """
+    batfile = Path(__file__).parent / "bin" / "hello_world.bat"
+    os.chdir(batfile.parent)
```
daniel-shimon, line 830: Use the monkeypatch fixture with `monkeypatch.chdir`

daniel-shimon (changes requested):

tests/test_integrations.py
daniel-shimon: It's probably the reason other tests are failing

melund: Yes. Let me just try that. Thanks

melund:
Thanks, @daniel-shimon. Tests pass now. You better merge it.

daniel-shimon (approved):
Great! Thanks @auneri @melund @anki-code for pushing this!

PR #4089, opened 2021-02-07
Fix URL for xontrib-autojump

tejasvi:
The original url is broken. This one seems to be the one actively maintained.


gforsyth:
Thanks for the fix @tejasvi!  Travis is having some issues right now but they don't seem related.

Issue #2346, opened 2017-03-28
xonsh builtin - nohup & disown


mitnk:
Currently, when I want to run something in background with `nohup` like:
```
$ nohup py myserver.py -p 8080 &
```
After this, the server should still be running after `exit`. But we have:
```
$ exit
xonsh: there is an unfinished job
-----
[1]+ running: nohup py myserver.py .... (35438)
-----
Type "exit" or press "ctrl-d" again to force quit.
```

A "force quit" would kill the `myserver` process.

The server stay normal if I forcely close the terminal Tab, but it will be great if xonsh recognizes `nohup` "natively".

astronouth7303:
What we really need is a disown.

mitnk:
I never use disown. Seems similar things: http://unix.stackexchange.com/questions/3886/difference-between-nohup-disown-and

astronouth7303:
`nohup` does a bunch of piping/signal stuff and is more about POSIX. `disown` does job control stuff and is more about the shell. Together (`nohup mybackground thing & disown`), they allow you to spin off processes.

mitnk:
When using `nohup something &`, an extra `disown` would not be needed I guess.

astronouth7303:
Strictly it would, since `nohup something` would still be a job.

mitnk:
Practically, it's not :) At least for bash, nohup made the process (or process group) detached from the controlling terminal, and though it remains in jobs, there's not effects from the shell's exit, since SIGHUP is suppressed by nohup. The link I put above explains all of these details.

astronouth7303:
Oh, bash no longer blocks exit when there's uncompleted jobs. It used to.

zsh does warn but exits anyway.

Both of them list `nohup` processes as jobs.

anddam:
Worth mentioning #1246, I guess.

adqm:
I don't know anything about xonsh after the big subprocess rewrite.  In older versions, this was pretty close to trivial (see [this patch](https://notabug.org/adqm/tako/commit/31fd57b331cd228c5949a7731ee66e363bbcf2ed)), so maybe there is a similarly simple patch here (though I tried just using that patch, and it failed because of weird Thread-handling thing).

ion201:
[...]

Issue #1246, opened 2016-06-16
feature request: disown support


OneOfOne:
Is there a way to disown a process or start a background process that wouldn't be killed when you exit the shell?

scopatz:
Thanks for bringing this up @OneOfOne! I don't believe that there is (though I may be wrong).  A PR would be very welcome here.

PR #4090, opened 2021-02-07
add disown command

ion201:
Relevant issues: #2346 and #1246

This pr adds a "disown" command to the shell. The behavior is modeled after zsh's "disown" builtin command, with the main differentiating factor being that zsh's `disown [job] &|` syntax is super esoteric so I decided to replace it with the options `-c` or `--cont`

Example usage:
```sh
nate@spicy-tuna ~/ > nemo &
[1]+ running: nemo & & (107246)
nate@spicy-tuna ~/ > disown
Removed job 1 (running)
```
```sh
nate@spicy-tuna ~/ > $THREAD_SUBPROCS=False
nate@spicy-tuna ~/ > nemo
^Znate@spicy-tuna ~/ > jobs
[1]+ stopped: nemo (107323)
nate@spicy-tuna ~/ > disown -c 1
Removed job 1 (running)
```
```sh
nate@spicy-tuna ~/ > $THREAD_SUBPROCS=False
nate@spicy-tuna ~/ > nemo
^Znate@spicy-tuna ~/ > jobs
[1]+ stopped: nemo (107408)
nate@spicy-tuna ~/ > disown
warning: job is suspended, use 'kill -CONT -107408' to resume
Removed job 1 (stopped)
```

This was my first look at this code base so feedback welcome if I missed anything!

### DIFF
--- a/docs/aliases.rst
+++ b/docs/aliases.rst
@@ -51,6 +51,22 @@ Resume execution of the currently active job in the background, or, if a
 single number is given as an argument, resume that job in the background.
 
 
+``disown``
+====================
+The behavior of this command matches the behavior of zsh's disown
+command which is as follows:
+
+Remove the specified jobs from the job table; the shell will no longer
+report their status, and will not complain if you try to exit an
+interactive shell with them running or stopped. If no job is specified,
+disown the current job.
+If the jobs are currently stopped and the $AUTO_CONTINUE option is set
+($AUTO_CONTINUE = True), a warning is printed containing information about
+how to make them running after they have been disowned. If one of the
+latter two forms is used, the jobs will automatically be made running,
+independent of the setting of the $AUTO_CONTINUE option.
+
+
 ``EOF``, ``exit``, and ``quit``
 ===================================
 The commands ``EOF``, ``exit``, and ``quit`` all alias the same action, which is to
--- a/docs/aliases.rst
+++ b/docs/aliases.rst
@@ -52,7 +52,7 @@ single number is given as an argument, resume that job in the background.
 
 
 ``disown``
-====================
+===================
 The behavior of this command matches the behavior of zsh's disown
 command which is as follows:
 
--- a/docs/aliases.rst
+++ b/docs/aliases.rst
@@ -52,7 +52,7 @@ single number is given as an argument, resume that job in the background.
 
 
 ``disown``
-===================
+==========
 The behavior of this command matches the behavior of zsh's disown
 command which is as follows:
 
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1072,7 +1072,7 @@ class GeneralSetting(Xettings):
         False,
         "If ``True``, automatically resume stopped jobs when they are disowned. "
         "When stopped jobs are disowned and this option is ``False``, a warning "
-        "will print information about how to continue the stopped process."
+        "will print information about how to continue the stopped process.",
     )
 
 
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -774,6 +774,13 @@ class Xettings:
 class GeneralSetting(Xettings):
     """General"""
 
+    AUTO_CONTINUE = Var.with_default(
+        False,
+        "If ``True``, automatically resume stopped jobs when they are disowned. "
+        "When stopped jobs are disowned and this option is ``False``, a warning "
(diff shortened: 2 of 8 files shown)


jnoortheen (approved):

xonsh/environ.py
```diff
@@ -1068,6 +1068,12 @@ class GeneralSetting(Xettings):
     )
     STAR_PATH = Var.no_default("env_path", pattern=re.compile(r"\w*PATH$"))
     STAR_DIRS = Var.no_default("env_path", pattern=re.compile(r"\w*DIRS$"))
+    AUTO_CONTINUE = Var.with_default(
+        False,
+        "If ``True``, automatically resume stopped jobs when they are disowned. "
+        "When stopped jobs are disowned and this option is ``False``, a warning "
+        "will print information about how to continue the stopped process.",
+    )
```
gforsyth, line 1076: Can you move this up to the top of this class?  We try to keep this alphabetically sorted (dunno what happens with `STAR_PATH` and `STAR_DIRS` but that can be handled in a separate PR

xonsh/jobs.py
```diff
@@ -443,3 +445,67 @@ def bg(args, stdin=None):
...
+    parser = argparse.ArgumentParser("disown", description=disown.__doc__)
+    parser.add_argument(
+        "job_ids",
+        type=int,
+        nargs="*",
+        help="Jobs to act on or none to use the current job",
+    )
+    parser.add_argument(
+        "-c",
+        "--cont",
```
gforsyth, line 476: I'm all for ignoring outdated syntax and this seems like a good way to handle this.
My general take on `-` vs `--` args is that the `-` arg should be the abbreviation and `--` should be the verbose option.  Should this be `--continue` instead?

xonsh/procs/posix.py
```diff
@@ -52,6 +52,9 @@ class PopenThread(threading.Thread):
 
     def __init__(self, *args, stdin=None, stdout=None, stderr=None, **kwargs):
         super().__init__()
+
+        self.daemon = True
+
```
gforsyth, line 57: I haven't dived deep into the proc handling in a bit -- was the absence of this causing a breakage somewhere?

xonsh/procs/specs.py
```diff
@@ -441,12 +441,10 @@ def run(self, *, pipeline_group=None):
         self._pre_run_event_fire(event_name)
         kwargs = {n: getattr(self, n) for n in self.kwnames}
         self.prep_env(kwargs)
-        self.prep_preexec_fn(kwargs, pipeline_group=pipeline_group)
         if callable(self.alias):
-            if "preexec_fn" in kwargs:
-                kwargs.pop("preexec_fn")
             p = self.cls(self.alias, self.cmd, **kwargs)
         else:
+            self.prep_preexec_fn(kwargs, pipeline_group=pipeline_group)
```
gforsyth, line 447: nice!

gforsyth (changes requested):
Thanks so much for putting this in @ion201 !  This looks great and has been a long-standing request.

I have a few small nits, nothing major.  I'll also express a slight preference for f-strings over calls to `.format` but that isn't a blocker.

docs/aliases.rst
```diff
@@ -51,6 +51,22 @@ Resume execution of the currently active job in the background, or, if a
 single number is given as an argument, resume that job in the background.
 
 
+``disown``
+====================
```
gforsyth, line 55: Can we trim the trailing `=` here?

xonsh/procs/posix.py
ion201: This option is required to support `disown` and `nohup` because otherwise the shell exit will block waiting for subprocess completion.

Outside of that case, I believe this is the proper behavior since it will cause the shell to orphan frozen processes (ie, subprocess is stuck in a syscall) instead of waiting for them.

ion201:
Agree with all of the suggested changes (including f-strings). Thanks @gforsyth for your comments

docs/aliases.rst
```diff
@@ -52,7 +52,7 @@ single number is given as an argument, resume that job in the background.
 
 
 ``disown``
-====================
+===================
```
gforsyth, line 55:
```suggestion
==========
```

gforsyth (approved):
Awesome!! Thank you for putting this in @ion201 !  And welcome to `xonsh`!

PR #4098, opened 2021-02-09
Fix garbage collection of JSON history specified in seconds.

bhawkins:
I was getting the warning "History garbage collection would discard more history than it would keep" every time I opened a shell. I use JSON history with XONSH_HISTORY_SIZE specified in seconds. As of #3600, the history filter was returning the timestamp of the oldest history file, but this value was being interpreted as the amount of history that would be removed. (That value should be oldest_kept - oldest_removed.) Thus whenever the gc was supposed to trigger, instead it issued a warning and did nothing.

### DIFF
--- a/news/fix-json-hist-gc-seconds.rst
+++ b/news/fix-json-hist-gc-seconds.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed JSON history garbage collection for XONSH_HISTORY_SIZE in seconds.
+
+**Security:**
+
+* <news item>
--- a/tests/test_history.py
+++ b/tests/test_history.py
@@ -378,9 +378,16 @@ HISTORY_FILES_LIST = history_files_list(100)
 
 # TS of newest history file
 SEC_FROM_LATEST = time.time() - (HISTORY_FILES_LIST[-1][0])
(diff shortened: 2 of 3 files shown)


bhawkins:
Sorry, I had to rebase out some unintended author info.

Tests are failing because I changed the interface of the rmfiles function. I could change the tests, but I'm not sure if I'm on the right track with this patch.

gforsyth:
Hey @bhawkins -- thanks for putting this in and sorry for the delay in getting to it.  I don't use the time-based GC in history, so forgive the perhaps silly questions.

Am I correct that the end goal here is to be able to set something like `$XONSH_HISTORY_SIZE="2mo"` or something and then all history files older than 2 months would be deleted?

bhawkins:
Yes, that's right. As in the [docs](https://xon.sh/envvars.html?highlight=xonsh_history_size#xonsh-history-size) and [history tutorial](https://xon.sh/tutorial_hist.html#gc-action) the idea is that history times out after the specified duration.

gforsyth:
Hey @bhawkins -- sorry for the delay.  I've looked at this and done a bit of digging and I think that your change is correct and the tests should be updated.  There's even a comment about "weird" behavior from the `gc_seconds` function

bhawkins:
Thanks for having a look. I found rewriting the tests was a bit trickier than I expected, but I think I've got it.

gforsyth (approved):
Thanks for all your work on this @bhawkins !  And sorry for the delay in reviewing!

gforsyth:
Oh, and I see this is your first contribution!  Welcome!

PR #4101, opened 2021-02-11
Py39 support

jnoortheen:
this is an easy way to solve all py39 subscriptor related issues https://github.com/xonsh/xonsh/issues?q=is%3Aissue+is%3Aopen+label%3Apy39

Note:
`container[obj.attr] = Value` is still not parsed correctly. It may require some more time investment. But atleast now py39 is usable. (there is a test commented out, anyone implementing that may use it to test the case)

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -420,13 +420,14 @@ def test_subscription_special_syntaxes(arr_container):
 
 
 # todo: enable this test
-# def test_subscription_special_syntaxes_2(arr_container):
-#     # aliases
-#     d = {}
-#     eval_code("d[arr.__name__]=True", arr=arr_container, d=d)
-#     assert d == {"Arr": True}
-#     # extslice
-#     assert eval_code('arr[:, "2"]') == 2
+@pytest.mark.xfail
+def test_subscription_special_syntaxes_2(arr_container):
+    # aliases
+    d = {}
+    eval_code("d[arr.__name__]=True", arr=arr_container, d=d)
+    assert d == {"Arr": True}
+    # extslice
+    assert eval_code('arr[:, "2"]') == 2
 
 
 def test_str_idx():
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -2,8 +2,10 @@
 """Tests the xonsh parser."""
 import ast
 import builtins
+import logging
 import textwrap
 import itertools
+import traceback
 
 import pytest
 
@@ -62,7 +64,7 @@ def check_xonsh(xenv, inp, run=True, mode="exec"):
 
 
 def eval_code(inp, mode="eval", **loc_vars):
-    obs = PARSER.parse(inp)
+    obs = PARSER.parse(inp, debug_level=1)
     bytecode = compile(obs, "<test-xonsh-ast>", mode)
     return eval(bytecode, loc_vars)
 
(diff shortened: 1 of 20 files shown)


jnoortheen:
@gforsyth could you restart the tests?

gforsyth:
@jnoortheen -- restarted the test but it still failed

jnoortheen:
I am not sure why it is failing. Do you have any suggestions?

gforsyth:
> I am not sure why it is failing. Do you have any suggestions?

The test in question was added in #3950 -- the discussion in the linked issue #3936 indicates that this _shouldn't_ be a problem on windows starting with Python 3.9 but clearly we're having an issue here.

@kip93 -- do you have any thoughts on this?

Obviously the quick "fix" is to just always skip this test on Windows, but it would be good to know why it isn't working as expected.

jnoortheen:
Ok lets wait for @kip93 . Otherwise we can skip that test for now.

kip93:
Hello!

I just rechecked and the build did pass on my PR, which included running against 3.9, and there were no changes to the corresponding files since then that could affect the result. Maybe the test failure was being silently ignored before these changes and I did not notice? I see that you removed a line that says `ALLOWED_FAILURES = ["3.9"]`, so this looks like a mistake on my side for not knowing how the tests were configured.

The reason I said that it should work on 3.9 was because I had read that they changed how Python uses the Windows registry along with `sys.path`, but reading up on it again I think it is only if run with the `-E` flag (see the [Python 3.9 changelog](https://docs.python.org/3/whatsnew/3.9.html), look for the word `registry` or `sys.path`), otherwise the behaviour is the same as the older versions. This would explain why the test is failing.

In any case, I do not see too much issue with changing the test to remove the Python version check and always ignore Windows, since this problem is much less likely to arise in Windows, because we do not need to support the XDG spec (using `~/.config/` to store configurations).

gforsyth:
Ok, great. Thanks for the info @kip93 ! 

@jnoortheen , let's proceed with skipping that test on windows

jnoortheen:
Ok done.

cafehaine:
I can confirm this fixes loading my xontrib `xlsd` when using python3.9, tested on Arch Linux with python 3.9.1

tests/test_parser.py
```diff
@@ -43,7 +45,7 @@ def check_stmts(inp, run=True, mode="exec", debug_level=0):
 
 
 def check_xonsh_ast(xenv, inp, run=True, mode="eval", debug_level=0, return_obs=False):
-    __tracebackhide__ = True
+    # __tracebackhide__ = True
```
gforsyth, line 48:
```suggestion
```

tests/test_parser.py
```diff
@@ -390,6 +398,38 @@ def test_if_else_expr_expr():
...
+
+
+# todo: enable this test
+# def test_subscription_special_syntaxes_2(arr_container):
+#     # aliases
+#     d = {}
+#     eval_code("d[arr.__name__]=True", arr=arr_container, d=d)
+#     assert d == {"Arr": True}
+#     # extslice
+#     assert eval_code('arr[:, "2"]') == 2
```
gforsyth, line 430: Maybe mark this as an xfail instead of commenting out?

gforsyth (commented):
Two small nits, neither blocking.

@jnoortheen I'm inclined to merge this as is if you're comfortable with it.  It looks reasonable to me and gets us to a more functional place than we currently are.

tests/test_parser.py
jnoortheen: I wanted to create another PR where one could work on making this test pass. I think it is better to mark it instead of commenting. I will update it.

gforsyth:
Awesome!  Thanks for taking this on @jnoortheen !

PR #4103, opened 2021-02-15
completer: support quotes in begidx search

### LINKED ISSUE #4046: completion: begidx splits quoted values
The completion does a [simple rfind](https://github.com/xonsh/xonsh/blob/master/xonsh/ptk_shell/completer.py#L43) to set begidx. But this ignores quotation which is [needed in xonsh](https://xon.sh/tutorial_subproc_strings.html#no-escape) for values containing space.

## Expected Behavior
```sh
example multiparts --at "third with space@"<TAB>
# begidx=24
```

## Current Behavior
```sh
example multiparts --at "third with space@"<TAB>
# begidx=36
```

## Steps to Reproduce
Invoke completion on a quoted value containing space.

## For community

rsteube:
Support quoted values in begidx search (parse using shlex in non-posix mode and do manual rfind for space)

fix #4046

### DIFF
--- a/xonsh/ptk_shell/completer.py
+++ b/xonsh/ptk_shell/completer.py
@@ -106,9 +106,7 @@ class PromptToolkitCompleter(Completer):
             return 0
 
         try:
-            words_non_posix = split(
-                line[0:endidx], posix=False
-            )
+            words_non_posix = split(line[0:endidx], posix=False)
             if line[0:endidx].endswith(" "):
                 words_non_posix.append("")
         except ValueError:
--- a/xonsh/ptk_shell/completer.py
+++ b/xonsh/ptk_shell/completer.py
@@ -102,29 +102,29 @@ class PromptToolkitCompleter(Completer):
                 yield Completion(comp, -plen, display=disp)
 
     def __find_begidx(self, line, endidx):
-        if len(line[0:endidx]) == 0:
+        if endidx < 1:
             return 0
 
-        wordsNonPosix = []
         try:
-            wordsNonPosix = split(
-                line[0:endidx] + "_", posix=False
-            )  # ensure last word is empty when ends with space
-            wordsNonPosix[-1] = wordsNonPosix[-1][:-1]
+            words_non_posix = split(
+                line[0:endidx], posix=False
+            )
+            if line[0:endidx].endswith(" "):
+                words_non_posix.append("")
         except ValueError:
             try:
-                wordsNonPosix = split(
+                words_non_posix = split(
                     line[0:endidx] + '"', posix=False
                 )  # try with missing `"` quote
-                wordsNonPosix[-1] = wordsNonPosix[-1][:-1]
+                words_non_posix[-1] = words_non_posix[-1][:-1]
             except ValueError:
-                wordsNonPosix = split(
+                words_non_posix = split(
                     line[0:endidx] + "'", posix=False
                 )  # try with missing `'` quote
-                wordsNonPosix[-1] = wordsNonPosix[-1][:-1]
+                words_non_posix[-1] = words_non_posix[-1][:-1]
 
         begidx = endidx
-        for word in reversed(wordsNonPosix):
+        for word in reversed(words_non_posix):
             begidx = begidx - len(word)
             if line[begidx - 1] == " ":
                 break
--- a/xonsh/ptk_shell/completer.py
+++ b/xonsh/ptk_shell/completer.py
@@ -7,6 +7,8 @@ from prompt_toolkit.completion import Completer, Completion
 from prompt_toolkit.auto_suggest import AutoSuggestFromHistory
 from prompt_toolkit.application.current import get_app
 
+from shlex import split
+
 from xonsh.completers.tools import RichCompletion
 
 
@@ -40,7 +42,7 @@ class PromptToolkitCompleter(Completer):
         line_ex = builtins.aliases.expand_alias(line)
 
         endidx = document.cursor_position_col
-        begidx = line[:endidx].rfind(" ") + 1 if line[:endidx].rfind(" ") >= 0 else 0
+        begidx = self.__find_begidx(line, endidx)
         prefix = line[begidx:endidx]
         expand_offset = len(line_ex) - len(line)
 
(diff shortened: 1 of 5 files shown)


rsteube:
Seems there are already some tests for completions. Should i add some more? If yes, where?

xonsh/ptk_shell/completer.py
```diff
@@ -99,6 +101,35 @@ def get_completions(self, document, complete_event):
                 disp = comp[pre:].strip("'\"")
                 yield Completion(comp, -plen, display=disp)
 
+    def __find_begidx(self, line, endidx):
+        if len(line[0:endidx]) == 0:
```
gforsyth, line 105: I think an equivalent check here is that `endidx <=1`

xonsh/ptk_shell/completer.py
```diff
@@ -99,6 +101,35 @@ def get_completions(self, document, complete_event):
...
+            return 0
+
+        wordsNonPosix = []
```
gforsyth, line 108: Python style guidance here is to use `snake_case` for variable names, not `camelCase`.
Additionally, since this variable is populated by the output of `split`, you don't need to define it here

xonsh/ptk_shell/completer.py
```diff
@@ -99,6 +101,35 @@ def get_completions(self, document, complete_event):
...
+        try:
+            wordsNonPosix = split(
+                line[0:endidx] + "_", posix=False
+            )  # ensure last word is empty when ends with space
+            wordsNonPosix[-1] = wordsNonPosix[-1][:-1]
```
gforsyth, line 113: This seems a little bit circuitous -- what's the end goal here?
Make sure that the last element in `words_non_posix` is an empty string if `line[0:endidx] == 0`?
It's better to be explicit if that's what you're after -- something like

```
if line[0:endidx].endswith(" "):
    words_non_posix.append("")
```

xonsh/ptk_shell/completer.py
```diff
@@ -99,6 +101,35 @@ def get_completions(self, document, complete_event):
...
+            try:
+                wordsNonPosix = split(
+                    line[0:endidx] + '"', posix=False
+                )  # try with missing `"` quote
+                wordsNonPosix[-1] = wordsNonPosix[-1][:-1]
+            except ValueError:
+                wordsNonPosix = split(
+                    line[0:endidx] + "'", posix=False
+                )  # try with missing `'` quote
+                wordsNonPosix[-1] = wordsNonPosix[-1][:-1]
```
gforsyth, line 124: A good rule here with try/catch is to only put the lines likely to raise the error inside of the `try`, then put the followup lines outside of the try/catch block

gforsyth (changes requested):
Hey @rsteube -- thanks for putting this in!  I've added a few comments around some ways we might want to tweak the code a bit.  Happy to guide you if it raises more questions than it answers.  I think we can simplify the underlying checks a bit (which will make it more maintainable in the future).

In re: your question about tests -- I think a great addition here would be to come up with a test that fails on `master`, currently, but that passes with your changes here.  That way, if someone changes something in the future they don't inadvertently revert these changes.

xonsh/ptk_shell/completer.py
rsteube: removed and renamed

xonsh/ptk_shell/completer.py
rsteube: i think you mean `endidx < 1`

xonsh/ptk_shell/completer.py
rsteube: Yeah, been a while that i have written this and i think it was just because i initially had some issues with split and also needed an additional (empty) argument passed to carapace for the completion to work correctly.
Just had another look at it and while it shouldn't be needed anymore here it does make the loop below a bit easier.
Ah, might work to move the check downward, let's see..

xonsh/ptk_shell/completer.py
rsteube: Probably would have to add a pseudo character in the first try block for this as its meant to remove the temporary quotation character which is only added in the other blocks.

xonsh/ptk_shell/completer.py
gforsyth: python indexing (with positive integers) is `[inclusive: exclusive)`

rsteube: yeah, if fell for that as well - indexing starts at 0 so `[0:1]` gives one character

gforsyth:
Hey @rsteube -- I've been thinking about this a little more, does this accurately describe what we want to do here?

1. If there is either an unmatched `'` or `"`, then `begidx` should be the character after the unmatched quotation mark
2. Otherwise it should be the character after the rightmost space

rsteube:
@gforsyth, yes exactly. As `split` needs a well quoted line these are just temporarily added.

gforsyth:
> _sorry about the mixups, this is easy to get wrong_

All good, thanks for laying it out cleanly!

Given that, I'm wondering if something like this would be a good solution (only because the nested try/catch statements are hard to parse):

```python
fragment = line[:endidx +1]  # maybe no +1?

if fragment.count("'") % 2 == 1:
    # unmatched single quote
    search_char = "'"
elif fragment.count('"') %2 == 1:
    # unmatched double quote
    search_char = '"'
else:
    search_char = " "

begidx = fragment.rfind(search_char)
```

rsteube:
Yeah sth. like that should work as well but i assume there are quite some caveats like `''"` and `\"` so i felt it was better to leave the parsing to a tool that knows what it is doing.
I don't mind trying that but i expect a couple of issues on the way.

gforsyth:
> quite some caveats like `'"'` and `\"` so i felt it was better to leave the parsing to a tool that knows what it is doing.

Yeah, I totally get this.  I'm sure there are some corner cases, but I think there's a general quality of performance with the first-order check for quote pairing that will be easy to maintain and solve most problems.  We can tackle thornier issues as they arise.  So if you're game and can add a few tests to make sure we don't revert this, that would be awesome!

rsteube:
@gforsyth Can do. I need some help with the tests though as i am a bit blind at the moment and can't find the place where these should be added.

gforsyth:
Hey @rsteube -- tests can go in here: https://github.com/xonsh/xonsh/blob/master/tests/test_ptk_completer.py

And for ease of testing, since we don't actually need any information from the class instance, I'd make `__find_begidx` a function instead of a method.

daniel-shimon:
Hi @rsteube, nice PR!
1. Please have a look at https://github.com/xonsh/xonsh/issues/4046#issuecomment-788120873 and give me feedback :)
2. This function also needs to be used in `xonsh/jupyter_kernel.py` and `xonsh/readline_shell.py` (maybe move it to `xonsh/prompt/base.py`?)

rsteube:
@daniel-shimon 2: yes, seems like a good idea  - i can see there is also some quote handling in [readline_shell](https://github.com/xonsh/xonsh/blob/master/xonsh/readline_shell.py#L435) (not sure what it does exactly though)

rsteube:
closing since #4017 is merged

PR #4106, opened 2021-02-19
Feature/speed up startup

jnoortheen:
re-use commands-cache results by pickling between runs

ToDos:

- [x] add tests

### DIFF
--- a/tests/test_commands_cache.py
+++ b/tests/test_commands_cache.py
@@ -1,5 +1,7 @@
 import os
 import builtins
+import time
+from unittest.mock import MagicMock
 
 import pytest
 
@@ -10,6 +12,7 @@ from xonsh.commands_cache import (
     predict_true,
     predict_false,
 )
+from xonsh import commands_cache
 from tools import skip_if_on_windows
 
 
@@ -24,6 +27,32 @@ def test_predict_threadable_unknown_command(xonsh_builtins):
     cc = CommandsCache()
     result = cc.predict_threadable(["command_should_not_found"])
     assert isinstance(result, bool)
+
+
+def test_commands_cached_between_runs(xonsh_builtins, tmp_path, monkeypatch):
+    xonsh_builtins.__xonsh__.env["XONSH_DATA_DIR"] = tmp_path
+    xonsh_builtins.__xonsh__.env["COMMANDS_CACHE_SAVE_INTERMEDIATE"] = True
+    xonsh_builtins.__xonsh__.env["PATH"] = [tmp_path]
+    cc = commands_cache.CommandsCache()
+
+    exec_mock = MagicMock(return_value=["bin1", "bin2"])
+    monkeypatch.setattr(commands_cache, "executables_in", exec_mock)
+
+    # wait for thread to end
+    cnt = 0  # timeout waiting for thread
+    while True:
+        if cc.all_commands or cnt > 10:
+            break
+        cnt += 1
+        time.sleep(0.1)
+    assert [b.lower() for b in cc.all_commands.keys()] == ["bin1", "bin2"]
+
+    files = tmp_path.glob("*.pickle")
+    assert len(list(files)) == 1
+
+    # cleanup dir
+    for file in files:
+        os.remove(file)
 
 
 TRUE_SHELL_ARGS = [
--- a/tests/test_commands_cache.py
+++ b/tests/test_commands_cache.py
@@ -1,5 +1,6 @@
 import os
 import builtins
+import pickle
 import time
 from unittest.mock import MagicMock
 
@@ -29,14 +30,22 @@ def test_predict_threadable_unknown_command(xonsh_builtins):
     assert isinstance(result, bool)
 
 
-def test_commands_cached_between_runs(xonsh_builtins, tmp_path, monkeypatch):
+@pytest.fixture
+def commands_cache_tmp(xonsh_builtins, tmp_path, monkeypatch):
     xonsh_builtins.__xonsh__.env["XONSH_DATA_DIR"] = tmp_path
     xonsh_builtins.__xonsh__.env["COMMANDS_CACHE_SAVE_INTERMEDIATE"] = True
     xonsh_builtins.__xonsh__.env["PATH"] = [tmp_path]
-    cc = commands_cache.CommandsCache()
-
     exec_mock = MagicMock(return_value=["bin1", "bin2"])
     monkeypatch.setattr(commands_cache, "executables_in", exec_mock)
+    return commands_cache.CommandsCache()
+
+
+def test_commands_cached_between_runs(commands_cache_tmp, tmp_path):
+    # 1. no pickle file
+    # 2. return empty result first and create a thread to populate result
+    # 3. once the result is available then next call to cc.all_commands returns
+
+    cc = commands_cache_tmp
 
     # wait for thread to end
     cnt = 0  # timeout waiting for thread
@@ -53,6 +62,24 @@ def test_commands_cached_between_runs(xonsh_builtins, tmp_path, monkeypatch):
     # cleanup dir
     for file in files:
         os.remove(file)
+
+
+def test_commands_cache_uses_pickle_file(commands_cache_tmp, tmp_path):
+    cc = commands_cache_tmp
+    file = tmp_path / CommandsCache.CACHE_FILE
+    bins = {
+        "bin1": (
+            "/some-path/bin1",
+            None,
+        ),
+        "bin2": (
+            "/some-path/bin2",
+            None,
+        ),
+    }
+
+    file.write_bytes(pickle.dumps(bins))
+    assert cc.all_commands == bins
 
 
 TRUE_SHELL_ARGS = [
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -30,6 +30,8 @@ class CommandsCache(cabc.Mapping):
     the command has an alias.
     """
 
+    CACHE_FILE = "commands-cache.pickle"
+
     def __init__(self):
         self._cmds_cache = {}
         self._path_checksum = None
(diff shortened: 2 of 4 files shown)


anki-code:
I see around 50ms speed up on clear xonsh boot with this PR (210ms vs 260ms) on Ubuntu 20. Nice!

xonsh/commands_cache.py
```diff
@@ -76,59 +89,97 @@ def remove_dups(p):
...
+        if has_path_changed and has_any_path_updated:
+            if not has_alias_changed:
                 for cmd, alias in alss.items():
                     key = cmd.upper() if ON_WINDOWS else cmd
                     if key in self._cmds_cache:
                         self._cmds_cache[key] = (self._cmds_cache[key][0], alias)
                     else:
                         self._cmds_cache[key] = (cmd, True)
+                # save the changes to cache as well
+                self.set_cmds_cache(self._cmds_cache)
```
daniel-shimon, line 127: Maybe run this in a separate thread as well?

daniel-shimon (changes requested):

daniel-shimon:
Hi! This seems like a really good idea!
I've gone through the code and apart from two small things I wrote about it looks good!
Are there any other changes you want to make?

jnoortheen:
> Hi! This seems like a really good idea!
> I've gone through the code and apart from two small things I wrote about it looks good!
> Are there any other changes you want to make?
> 
> Also, can you add a test that the caching works correctly?

I like to write functional tests rather than unit tests. But here I think unitests will be simpler and quick to write. I will add it.

xonsh/commands_cache.py
```diff
@@ -76,59 +89,97 @@ def remove_dups(p):
...
+        yield max_mtime <= self._path_mtime
         self._path_mtime = max_mtime
 
-        if cache_valid_path and cache_valid_paths:
-            if not cache_valid_aliases:
+    @property
+    def all_commands(self):
+        paths = CommandsCache.remove_dups(builtins.__xonsh__.env.get("PATH", []))
+        path_immut = tuple(x for x in paths if os.path.isdir(x))
+        alss = getattr(builtins, "aliases", dict())
```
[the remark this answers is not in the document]
jnoortheen: this is inside a property. It is better to pass things, rather than depending on a global state. I know it is ugly to mix both. But here it is only for the property all_commands

xonsh/commands_cache.py
jnoortheen: these part of the code doesn't take much time. You could see that 99% of time is taken by the threaded function which is offloaded in a thread

jnoortheen:
> > Hi! This seems like a really good idea!
> > I've gone through the code and apart from two small things I wrote about it looks good!
> > Are there any other changes you want to make?
> > Also, can you add a test that the caching works correctly?
> 
> I like to write functional tests rather than unit tests. But here I think unitests will be simpler and quick to write. I will add it.

Well now the tests are done. I think it can be checked now.

daniel-shimon:
Hey @jnoortheen - I've looked at this some more and it seems that **all code in xonsh depends on `all_commands` updating the cache instantly**, so updating it in a background thread seems to break that.
Also, calling it multiple times will cause multiple threads to be created unnecessarily.
Another issue is that even if the pickled value is updated correctly, you still create a thread.

I have an alternative suggestion:
In the same pickle file, save `self._path_checksum`, `self._alias_checksum` and `self._path_mtime`.
In `all_commands`, load these attributes from the pickle file before checking the new values in `_update_if_changed`.

This way the rest of `all_commands` stays the same and its behavior doesn't break.

tests/test_commands_cache.py
```diff
@@ -26,6 +29,32 @@ def test_predict_threadable_unknown_command(xonsh_builtins):
     assert isinstance(result, bool)
 
 
+def test_commands_cached_between_runs(xonsh_builtins, tmp_path, monkeypatch):
```
daniel-shimon, line 32: This test checks that a pickle file was created, but it doesn't check that it will be used in a new `CommandsCache()` instance

daniel-shimon (changes requested):

xonsh/environ.py
```diff
@@ -785,6 +785,10 @@ class GeneralSetting(Xettings):
         6000,
         "Number of files on the PATH above which a warning is shown.",
     )
+    COMMANDS_CACHE_SAVE_INTERMEDIATE = Var.with_default(
+        False,
+        "If enabled, the results are saved between runs and can reduce the startup time.",
```
anki-code, line 790: What is "results"?

anki-code (changes requested):

xonsh/environ.py
jnoortheen: the binary list fetched from the function.

tests/test_commands_cache.py
jnoortheen: > Hey @jnoortheen - I've looked at this some more and it seems that **all code in xonsh depends on `all_commands` updating the cache instantly**, so updating it in a background thread seems to break that.
> Also, calling it multiple times will cause multiple threads to be created unnecessarily.
> Another issue is that even if the pickled value is updated correctly, you still create a thread.
> 
> I have an alternative suggestion:
> In the same pickle file, save `self._path_checksum`, `self._alias_checksum` and `self._path_mtime`.
> In `all_commands`, load these attributes from the pickle file before checking the new values in `_update_if_changed`.
> 
> This way the rest of `all_commands` stays the same and its behavior doesn't break.

The very first time when this setting is enabled, it will return empty results. Usually from the call of xonsh-pygments plugin. This is just a quick implementation. It doesn't use locks. If there is more calls when a thread is running, it will create new ones. But for this case, it is just a quick better option. We can improve it in another PR.

atmost even if there is no change, it will create a thread at the start of the program to update the pickle file. Next time onwards the function will see that there was no change and will just return the all_commands stored in the variable.

daniel-shimon:
OK I'm fine with that 👍

Still saving `self._path_checksum`, `self._alias_checksum` and
`self._path_mtime`
in the pickle and loading it before `_update_if_changed` will save an
unneeded update on every load :)

anki-code:
I suggest to put the group of the cache env variables to the `class CacheSetting(Xettings):` i.e. `COMMANDS_CACHE_SIZE_WARNING`, `COMMANDS_CACHE_SAVE_INTERMEDIATE`, `XONSH_CACHE_SCRIPTS`, `XONSH_CACHE_EVERYTHING`.

xonsh/environ.py
anki-code: I think it will be cool to have more detailed description of what will be saved i.e. "the CommandsCache will be saved between runs".

jnoortheen: updated

jnoortheen:
> I suggest to put the group of the cache env variables to the distinct `class CacheSetting(Xettings):` i.e. `COMMANDS_CACHE_SIZE_WARNING`, `COMMANDS_CACHE_SAVE_INTERMEDIATE`, `XONSH_CACHE_SCRIPTS`, `XONSH_CACHE_EVERYTHING`. This will make https://xon.sh/envvars.html more structured.

I think the belong two+two group. Lets hear others opinion.

tests/test_commands_cache.py
jnoortheen: > This test checks that a pickle file was created, but it doesn't check that it will be used in a new CommandsCache() instance

added another test that loads result from pickle file if available and checks the results.

jnoortheen:
@gforsyth please take a look at the PR.

gforsyth:
I'll try to carve out some time to review this week.

@daniel-shimon -- are you available to take a look at this again?

daniel-shimon (commented):
I've gone through the code itself and it's fine by me, except:

1. This should probably be enabled by default, it's a great feature!
2. The pickle cache doesn't contain `_path_checksum`, `_alias_checksum` and `_path_mtime`, which creates two issues:

- If these parameters haven't changed since the cache was made we're running the update thread unnecessarily.
- If these parameters have changed, we're using an invalid cache.

jnoortheen:
> I've gone through the code itself and it's fine by me, except:
> 
> 1. This should probably be enabled by default, it's a great feature!
> 2. The pickle cache doesn't contain `_path_checksum`, `_alias_checksum` and `_path_mtime`, which creates two issues:
> 
> * If these parameters haven't changed since the cache was made we're running the update thread unnecessarily.
> * If these parameters have changed, we're using an invalid cache.

we can do 2 in another PR I suppose. Regarding 1, not so many would want threads created. In some places with very less number of binaries, this is a overkill.

gforsyth (approved):
Ok, this is in!  Thanks @jnoortheen for the work and thanks @daniel-shimon for the extra eyes!

I'm not opposed to a follow-up PR making this the default behavior if it seems stable.

scopatz:
FWIW - these fix along with the XSH one seems to have broken pygments_cache/pyghooks/xo.  I am now seeing:

```sh
$ xo rever.xsh
  File "/home/scopatz/miniforge/bin/xo", line 10, in <module>
    sys.exit(main())
  File "/home/scopatz/miniforge/lib/python3.8/site-packages/xo.py", line 1119, in main
    main_display.main(line, col)
  File "/home/scopatz/miniforge/lib/python3.8/site-packages/xo.py", line 776, in main
    self.walker.goto(line, col)
  File "/home/scopatz/miniforge/lib/python3.8/site-packages/xo.py", line 530, in goto
    self._ensure_read_in(lineno)
  File "/home/scopatz/miniforge/lib/python3.8/site-packages/xo.py", line 480, in _ensure_read_in
    next_line = self.read_next_line()
  File "/home/scopatz/miniforge/lib/python3.8/site-packages/xo.py", line 457, in read_next_line
    self._ensure_lexer()
  File "/home/scopatz/miniforge/lib/python3.8/site-packages/xo.py", line 411, in _ensure_lexer
    lexer = pygments_cache.get_lexer_for_filename(self.name)
  File "/home/scopatz/miniforge/lib/python3.8/site-packages/pygments_cache.py", line 316, in get_lexer_for_filename
    lexer = cls(**options)
  File "/home/scopatz/miniforge/lib/python3.8/site-packages/pygments/lexer.py", line 579, in __call__
    return type.__call__(cls, *args, **kwds)
  File "/home/scopatz/miniforge/lib/python3.8/site-packages/xonsh/pyghooks.py", line 1665, in __init__
    XSH.commands_cache = CommandsCache()
  File "/home/scopatz/miniforge/lib/python3.8/site-packages/xonsh/__amalgam__.py", line 8603, in __init__
    if "XONSH_DATA_DIR" in env and env.get("COMMANDS_CACHE_SAVE_INTERMEDIATE")
TypeError: argument of type 'NoneType' is not iterable
```

jnoortheen:
> FWIW - these fix along with the XSH one seems to have broken pygments_cache/pyghooks/xo. I am now seeing:
> 

@scopatz  Are you still getting the error? XSH.load() need to be called before xonsh lexer is loaded.

PR #4107, opened 2021-02-19
Remove vended ptk

jnoortheen:
@gforsyth here is the PR to remove PTK vendoring


gforsyth:
Looks like we have a nasty import in `xonsh/tools.py` that should be `from os import scandir` instead of `from xonsh.platform import scandir`

jnoortheen:
@gforsyth could you restart the build?

gforsyth (approved):
Thanks for taking this on @jnoortheen !  And for fixing up all the circular dependency stuff with `scandir`!

PR #4114, opened 2021-02-22
A bit docs around subprocess strings for sh-ers

anki-code:
Hi! 
A bit docs around subprocess strings for sh-ers.
No news needed.
Thanks!

Related to #1432


gforsyth:
Looks great!  Thanks for the updates @anki-code !

PR #4115, opened 2021-02-22
Perform path completion filtering based on expanded path

ALSchwalm:
This fixes issues with the cd and rmdir completers not working correctly due to `os.path.isdir` returning false for paths like '\~/example'. That causes the filtering in `complete_path` to always remove all path completions when the prefix has a '~', which in turn causes the cd and rmdir completers to return nothing. That seems to trigger the general path completer to run, which returns all paths with the given prefix, including files.

This PR moves the filtering step in `complete_path` to before the 'unexpanding' step, which seems to resolve the issue. I have not added a test for this case, as it's not clear to me if the infrastructure in tests/competers can handle this case.

### DIFF
--- a/news/fix-cd-rmdir-completions.rst
+++ b/news/fix-cd-rmdir-completions.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed a bug where "cd" and "rmdir" would return non-directory completions
+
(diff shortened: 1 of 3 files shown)


ALSchwalm:
Hmm, it also appears that this is insufficient in some cases. For example, in directories that contain no subdirs, we are again in the situation where the cd completer will return nothing (correctly), but the generic path completer will show all of the files. I'm not sure if that can be easily fixed, but I'll look in to it.

gforsyth (approved):
Thanks for putting this in @ALSchwalm and sorry for the delay in getting to it!

gforsyth:
Also, I see this is your first contribution, welcome to xonsh!

PR #4121, opened 2021-02-26
Strict permission for history file (json, sqlite)

anki-code:
History files (json, sqlite) now have ``600`` (``-rw-------``) permission by default.


gforsyth (approved):
Seems like a good idea to me!  Thanks, @anki-code !

PR #4139, opened 2021-03-04
feat: add xontrib-linuxbrew

eugenesvk:
Since Homebrew on Linux doesn't support xonsh and rejected a PR to it natively, made this xontrib to automatically update shell environment on Linux to include Homebrew's vars and paths


gforsyth:
Thanks for putting this in @eugenesvk !

Issue #4141, opened 2021-03-05
The first user experience after removing PTK


anki-code:
Hi!

I have a concern about bad user experience after removing PTK in #4107.

The `readline` shell is not so good and full featured like `prompt-toolkit`. When I do
```python
docker run --rm -it python:3.9-slim /bin/bash -c "apt update && apt install -y git && pip install -U git+https://github.com/xonsh/xonsh.git; xonsh"
# I should know about `pip install prompt-toolkit` to get all features
```
I get xonsh 0.9.27 with readline by default and it starts fast but I'm feeling bad without beautiful ptk suggestion and grey line suggestion of the command. The first impression experience was decreased by removing PTK. Also the users who will install PTK-tested prompt themes may face with ugly issues.

~We released the new landing where `pip install xonsh` is used and user who will run this command will get the `readline` shell. It's sad.~ (PR with the fix below)

What do you think about these consequences?

I personally for the adding ptk as the default dependency for xonsh package.

@jnoortheen @gforsyth 

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

gforsyth:
Hey @anki-code -- there's a longer discussion to be had here, for sure, but as a stopgap, do we want to update the front page to suggest `python -m pip install xonsh[full]`?

anki-code:
> Hey @anki-code -- there's a longer discussion to be had here, for sure, but as a stopgap, do we want to update the front page to suggest python -m pip install xonsh[full]?

Yep, PR is here #4142

And while I update the page I think: how it works in package managers (apt/brew/etc)? Are they also installing pure xonsh?

gforsyth:
I know that brew, conda-forge, and pacman install with prompt toolkit, I'll wager apt does the same thing.  General policy of most packagers is to grab all options

gforsyth:
Yep, I think we're all set for `apt`, too: https://packages.debian.org/sid/xonsh

anki-code:
> Yep, I think we're all set for apt, too: https://packages.debian.org/sid/xonsh

I seen the 0.9.25 and imagined may be after update to 0.9.27 the ptk dependency will disappeare. But if you have an assurance, that's cool :) Thanks!

anki-code:
I think we should have a words in the welcome message (without rc) like "You're using readline, install prompt-toolkit for the best experience." (and a bit colors refactoring #2401). This feature would close this issue...

gforsyth:
[...]

PR #4142, opened 2021-03-05
Update index.html: install xonsh[full]

anki-code:
#4141
No news


gforsyth:
Thanks @anki-code !  I'll cut a small release this weekend so this makes it to the main front page

PR #4143, opened 2021-03-05
Welcome message: fix color, add ptk warning

### LINKED ISSUE #2401: Poor choice of text colors on welcome screen
I just tried starting up xonsh on a whim, and got this:

foo

I think there might be some text there? That I might want to read? It's hard to say. I'm guessing the vibe you were going for was like... "welcoming"? but it landed on "eyestrain".

(Possibly this is my fault for using an almost-but-not-quite-white background for my terminal.)

### LINKED ISSUE #4141: The first user experience after removing PTK

anki-code:
Closes #4141
Closes #2401

* Added warning about shell type.
* A bit fix of colors.
* Removed two phrases about bash because xonsh is not sh-compatible in fact.

Appearance:

<details>

image

</details>

### DIFF
--- a/xonsh/xonfig.py
+++ b/xonsh/xonfig.py
@@ -877,7 +877,6 @@ WELCOME_MSG = [
 
 def print_welcome_screen():
     shell_type = builtins.__xonsh__.env.get('SHELL_TYPE')
-    print(shell_type)
     subst = dict(tagline=random.choice(list(TAGLINES)), version=XONSH_VERSION)
     for elem in WELCOME_MSG:
         if elem == '[SHELL_TYPE_WARNING]':
--- a/xonsh/xonfig.py
+++ b/xonsh/xonfig.py
@@ -827,7 +827,6 @@ def TAGLINES():
         "Piggy glanced nervously into hell and cradled the xonsh",
         "The xonsh is a symbol",
         "It is pronounced conch",
-        "The shell, bourne again",
         "Snailed it",
         "Starfish loves you",
         "Come snail away",
@@ -848,21 +847,20 @@ def TAGLINES():
         "Python-powered, cross-platform, Unix-gazing shell",
         "Tab completion in Alderaan places",
         "This fix was trickier than expected",
-        "The unholy cross of Bash/Python",
     ]
 
 
 # list of strings or tuples (string, align, fill)
 WELCOME_MSG = [
     "",
-    ("{{INTENSE_WHITE}}Welcome to the xonsh shell ({version}){{RESET}}", "^", " "),
+    ("Welcome to the xonsh shell {version}", "^", " "),
     "",
     ("{{INTENSE_RED}}~{{RESET}} {tagline} {{INTENSE_RED}}~{{RESET}}", "^", " "),
     "",
     ("{{INTENSE_BLACK}}", "<", "-"),
     "",
     (
-        "{{INTENSE_BLACK}}Create ~/.xonshrc file manually or use xonfig to suppress the welcome screen",
+        "{{INTENSE_BLACK}}Create ~/.xonshrc file manually or use xonfig to suppress the welcome message",
         "^",
         " ",
     ),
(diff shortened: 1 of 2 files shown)


xonsh/xonfig.py
```diff
@@ -848,37 +847,44 @@ def TAGLINES():
...
 
 
 def print_welcome_screen():
+    shell_type = builtins.__xonsh__.env.get("SHELL_TYPE")
     subst = dict(tagline=random.choice(list(TAGLINES)), version=XONSH_VERSION)
     for elem in WELCOME_MSG:
+        if elem == "[SHELL_TYPE_WARNING]":
+            if shell_type != "prompt_toolkit":
+                print_color(
+                    "\n{INTENSE_BLACK}Install {RESET}prompt-toolkit{INTENSE_BLACK} to get the best prompt experience.{RESET}"
```
jnoortheen, line 885: can this best-experience be replaced with featureful-prompt or something similar.

anki-code: Give an example please )

jnoortheen: Install other xonsh dependencies with 'xonsh[full]' to get fully featured prompt.

anki-code: This message will not be compatible for non-pip package managers.

jnoortheen: It is just an example. It could be "Install prompt-toolkit to use featureful prompt and pygments for syntax color support."

anki-code: Ok, I suggest to wait @gforsyth`s thoughts too )

gforsyth: How about:

```
You are currently using the readline backend.  For interactive tab-completion, on-the-fly syntax highlighting, and more, install prompt_toolkit by running xpip install -U xonsh[full]
```

gforsyth: And I _think_ that `pip` (and other pypi-based installs) is the only installation method that won't bring down `prompt_toolkit` by default

anki-code:
New text:

<details>

image

</details>

xonsh/xonfig.py
```diff
@@ -848,40 +847,48 @@ def TAGLINES():
...
 
 
 def print_welcome_screen():
+    shell_type = builtins.__xonsh__.env.get("SHELL_TYPE")
     subst = dict(tagline=random.choice(list(TAGLINES)), version=XONSH_VERSION)
     for elem in WELCOME_MSG:
+        if elem == "[SHELL_TYPE_WARNING]":
+            if shell_type != "prompt_toolkit":
+                print_color(
+                    f"\n{{INTENSE_BLACK}}You are currently using the {shell_type} backend. For interactive tab-completion, on-the-fly syntax highlighting, and more, install prompt_toolkit by running:\n\n  {{GREEN}}xpip{{RESET}} install -U 'xonsh[full]'"
```
gforsyth, line 885: `flake8` is complaining about line length here.  We can either break it up or `noqa` it

gforsyth (changes requested):

gforsyth (approved):
Great!  Thanks for the fixes, @anki-code !

PR #4150, opened 2021-03-07
docs: added quotes to pip install

anki-code:
To avoid errors in zsh/xonsh we need to add quotes

```python
zsh
pip install xonsh[full]
# zsh: no matches found: xonsh[full]

xonsh
pip install xonsh[full]
# SyntaxError: <xonsh-code>:1:4: ('code: install',)
```

No news


gforsyth (approved):
:+1:

PR #4153, opened 2021-03-07
bash-completion: Update to 0.2.7

### LINKED ISSUE #4104: Absolute paths to executables breaks bash completions
I came across this because I have a few aliases to make sure I'm running a given executable no matter what environment I'm in, e.g.

`aliases["git"] = "/usr/local/bin/git"`

## xonfig

<details>

```
$ xonfig
| xonsh            | 0.9.26.dev7         |
| Git SHA          | c4e78435            |
| Commit Date      | Feb 9 10:54:17 2021 |
| Python           | 3.8.5               |
| prompt toolkit   | 3.0.8               |
| pygments         | 2.7.3               |
| on jupyter       | False               |
| jupyter kernel   | None                |
| xontrib          | []                  |
```

</details>

## Expected Behavior
bash completions should work when we pass in an absolute path to an executable

```
$ /usr/local/bin/git st<TAB>
```

should show me git completions

## Current Behavior
```
$ /usr/local/bin/git st<TAB>
```

shows me path completions as a fallback

## What's going on
the bash completer function (and all the others, too) expect inputs in the form `(prefix, line, begidx, endidx, ctx)` -- only `prefix` and `line` are applicable here.

If I use the absolute path, then `line == "/usr/local/bin/git st"`, at which point the bash completer helper looks for an executable named `/usr/local/bin/git` which doesn't exist, i.e. it is parsing the path to the executable as the full executable.

We should do a `rstrip("/")[-1]` or similar to return the correct completions

## Steps to Reproduce
Try to complete a `git` command using the absolute path to your `git` executable

## For community

daniel-shimon:
Closes #4104 

@gforsyth

### DIFF
--- a/news/update-bash-completion.rst
+++ b/news/update-bash-completion.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Absolute paths to executables don't break bash completions anymore
+
+**Security:**
+
+* <news item>
--- a/xonsh/completers/bash_completion.py
+++ b/xonsh/completers/bash_completion.py
@@ -15,7 +15,7 @@ import functools
 import subprocess
 import typing as tp
 
-__version__ = "0.2.6"
+__version__ = "0.2.7"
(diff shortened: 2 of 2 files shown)


gforsyth (approved):
Thanks for the update and the fix upstream @daniel-shimon !

PR #4155, opened 2021-03-08
The prompt end character is conch

### LINKED ISSUE #4152: Reinventing the prompt end character for xonsh
Hi folks! 

Prerequisites:
1. While I'm watching [@scopatz streams](https://www.youtube.com/watch?v=dGTkiGvrYHg) I noticed that [xonsh looks like bash](https://youtu.be/EbX2fhfwOWc?t=832) and if we don't know that it's the xonsh we will never understand that it is xonsh. 
2. When I alternately use Bash and Xonsh I caught myself thinking that I am making an effort to figure out which shell I am currently using.
3. When I'm reading the bug reports in the issues or another code quotations the `$` character means bash as well as xonsh and I need to special effort to understand the meaning.
4. The bash-like prompt looks cozy but after a few rakes with the syntax users feel misled. The conch character will remind them that here is Python mindset.
5. We have double dollar in case of using env: `user@home ~ $ $VAR=1`. The conch character will dilute the repetition.
6. The conch character (`@`) in the prompt will complete the xonsh identity. We can remove or replace this character between user and host to more significant effect.
7. The `@` character looks both like conch and python (snake).
8. Using `$` limit the spreading the word about xonsh and growing the community because people think that it's bash.
9. By using `@` we can separate `$` for bash and `@` for xonsh.
10. The `@` character with `~` looks like "snail (snake, python) going to conch (shell)" - very very very metaphoric - `host@user ~ @`.

Overview:
```bash
# xonsh default 
user@host /path $

# bash default
user@host:/path$

# zsh has percent by default
user@host /path %  

# fish wants to look like fish
user@host /path>

# Starship cross-shell prompt (bold angle bracket)
❯

# proposed for xonsh
user@host /path @
```

**What if we replace `$` to `@` and make/replace a space between user and host in xonsh?** 

image

image

This looks like snail and conch:

image

[xontrib-prompt-bar](https://github.com/anki-code/xontrib-prompt-bar) appearance:

image

### 📄 [XEP-like description](https://github.com/xonsh/xonsh/issues/4152#issuecomment-794875983)

# 📗 VOTE HERE - https://github.com/xonsh/xonsh/issues/4152#issuecomment-794178842

## For community

anki-code:
Closes #4152

[x] Replacing $ to @
[ ] Updating tutorial
[ ] Searching more occurrences (please help)
[x] Updating xon.sh

*Yeah, new epic PR from anki-code =)*

### DIFF
--- a/docs/_templates_overwrite/index.html
+++ b/docs/_templates_overwrite/index.html
@@ -142,11 +142,11 @@
                     <div class="text-center">
 
                         <pre class="code">
-<span unselectable-text="$ "></span><b>cd</b> /home
+<span unselectable-text="@ "></span><b>cd</b> /home
 
-<span unselectable-text="$ "></span><b>cat</b> /etc/passwd | <b>grep</b> root
+<span unselectable-text="@ "></span><b>cat</b> /etc/passwd | <b>grep</b> root
 
-<span unselectable-text="$ "></span><b>ls</b> -la</pre>
+<span unselectable-text="@ "></span><b>ls</b> -la</pre>
 
                     </div>
                 </div>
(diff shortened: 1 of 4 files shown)


anki-code:
Converted to draft until getting the feedback.

PR #4156, opened 2021-03-09
fix: abbrevs with callback alters buffer text before expansion

### LINKED ISSUE #3642: abbrevs should only be expanded at the start of the command
## xonfig

<details>

```
| Git SHA          | 26475146            |
| Commit Date      | May 5 06:42:28 2020 |
| Python           | 3.8.3               |
| distro           | manjaro             |
```

</details>

as of abbrevs expands in all places. So it may clash will subcommands. For example, if we have 

~~`aliases['ps'] = 'procs'`~~
```
abbrevs['ps'] = 'procs'
``` 

and do type `docker ps` then it will get expanded to `docker procs`. 

So the default behaviour should be to expand only at the start of the command. special handling for sudo can be added. 
Also we could have special symbols in aliases to mark it as expandable in all places like `aliases['*ps*']` like the glob pattern.

jnoortheen:
fixes #3642 bug specified in the comment https://github.com/xonsh/xonsh/issues/3642#issuecomment-793789741

### DIFF
--- a/tests/xontribs/test_abbrevs.py
+++ b/tests/xontribs/test_abbrevs.py
@@ -27,11 +27,17 @@ def abbrevs_xontrib(monkeypatch, source_path):
     del sys.modules[spec.name]
 
 
+ps_special_expand = (
+    lambda buffer, word: "procs" if buffer.text.startswith(word) else word
+)
+
+
 @mark.parametrize(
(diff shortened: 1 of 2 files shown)


gforsyth (approved):
thanks @jnoortheen !

PR #4162, opened 2021-03-10
update imphooks encoding regex to match the newer version at PEP 263

eugenesvk:
Noticed the PEP 263 update while fixing a BOM bug, though not sure this has any effect

### DIFF
--- a/xonsh/imphooks.py
+++ b/xonsh/imphooks.py
@@ -23,7 +23,7 @@ from xonsh.platform import ON_WINDOWS
 def ENCODING_LINE():
     # this regex comes from PEP 263
     # https://www.python.org/dev/peps/pep-0263/#defining-the-encoding
-    return re.compile(b"^[ tv]*#.*?coding[:=][ t]*([-_.a-zA-Z0-9]+)")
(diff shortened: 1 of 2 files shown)


xonsh/imphooks.py
```diff
@@ -23,7 +23,7 @@
 def ENCODING_LINE():
     # this regex comes from PEP 263
     # https://www.python.org/dev/peps/pep-0263/#defining-the-encoding
-    return re.compile(b"^[ tv]*#.*?coding[:=][ t]*([-_.a-zA-Z0-9]+)")
+    return re.compile(b"^[ tf]*#.*?coding[:=][ t]*([-_.a-zA-Z0-9]+)")
```
gforsyth, line 26: huh... should this be 
```suggestion
    return re.compile(b"^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+)")
```
?

eugenesvk: You're correct, the tabs etc should be escaped for the regex to match them
I just assumed the original code was correct without the escaping, so didn't adjust it in this way:(

gforsyth (approved):
Thanks @eugenesvk !

PR #4166, opened 2021-03-11
docs: added robots.txt

anki-code:
Hi @gforsyth !

I've added `robots.txt` file to avoid crawling` / dev / `content.

But I'm not sure that this file will appear as `https://xon.sh/robots.txt` because I don't understand how the building of the site is working. Please check me.

No news file

### DIFF
--- a/docs/_templates_overwrite/robots.txt
+++ b/docs/_templates_overwrite/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /dev/


gforsyth:
we'll see how it goes -- I think this will now show up at xon.sh/dev/robots.txt and then will also appear at xon.sh/robots.txt after the next release

anki-code:
@gforsyth https://xon.sh/dev/robots.txt returns 404. I think the start point is here: 

https://github.com/xonsh/xonsh/blob/17961e3c241e5c820616d17dd608f95db9131ace/docs/conf.py#L219-L221

But I can't find any other references )

gforsyth:
https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_extra_path is, I think, what you want

PR #4168, opened 2021-03-11
feat: Added expanding callable environment variables in subprocess mode.

### LINKED ISSUE #4167: Expand callable environment variable in subprocess mode
Hi! Think that's nice:

```python
$DATE = lambda: $(date --iso).strip()
echo The date is $DATE
```
Current output:
```
The date is <function <lambda> at 0x7f807ecda940>
```
Suggested output:
```
The date is 2021-03-11
```

## For community

anki-code:
Closed #4167

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -135,6 +135,14 @@ variable in Python.  The same is true for deleting them too.
     Become the Lord of the Files
     >>> del $GOAL
 
+The environment variable may be callable and be executed in the subprocess as easy as possible:
+
+.. code-block:: xonshcon
+
+    >>> $WHO = lambda: $(whoami).strip()
+    >>> echo "$WHO is here!"
+    snail is here!
+
 Very nice.
 
 .. note::
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -139,9 +139,10 @@ The environment variable may be callable and be executed in the subprocess as ea
 
 .. code-block:: xonshcon
 
-    >>> $WHO = lambda: $(whoami).strip()
-    >>> echo "$WHO is here!"
-    snail is here!
+    >>> $XONSH_ENV_ALLOW_CALLABLE = True
+    >>> $ANSWER = lambda: 2+2
+    >>> echo "2 + 2 = $ANSWER"
+    2 + 2 = 4
 
 Very nice.
 
--- a/news/envvar_expand_callable.rst
+++ b/news/envvar_expand_callable.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Added expanding callable environment variables in subprocess mode. Now you can set variable as callable and it will be called to get the value at use.
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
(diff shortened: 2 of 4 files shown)


anki-code:
The meaningful changes here - https://github.com/xonsh/xonsh/pull/4168/files#diff-dc6388ab2415b7a836976e59822abb75c28f9c81b9a0308fef6d9eef51249c13R2338-R2355
Other changes is tests + black. Sorry.

anki-code:
I want to do this optional. Please give me some time.

anki-code:
@gforsyth I've made this feature optional and by default add the raising the exception to avoid getting `<function lambda>` string as an argument. This is more stable logic I suppose.

Example:
```python
$ENV = lambda: 123
echo $ENV
# Exception: The ENV variable is callable. If you want to use callables set $XONSH_ENV_ALLOW_CALLABLE = True

$XONSH_ENV_ALLOW_CALLABLE = True
echo $ENV
# 123
```

jnoortheen:
any reason for raising exception if callable is assigned to a $variable?

anki-code:
> any reason for raising exception if callable is assigned to a $variable?

The reason is to avoid unexpected getting `<function __main__.<lambda>>` string argument.

For example you have
```python
echo "$key=123"
# $key=123
```
after time you're doing import:
```python
from somelib import *  # this has `key` function
```
As result you'll have unexpected result without any error:
```python
echo "$key=123"
# <function <lambda> at 0x7f72713500d0>=123
```
In case of exception you will go and replace it to `@("$key=123")` explicitly and make your code more stable.

I suppose that there is many places in xonsh where callables converts to the text without appropriate logic and you can catch unexpected result. This case is one of this.

anki-code:
Once again:

> any reason for raising exception if callable is assigned to a $variable?

The reason is to avoid unexpected getting `<function __main__.<lambda>>` string argument.

I suppose that there is many places in xonsh (essentially around environment) where callables were converted to the text without appropriate logic and you can catch unexpected result. This case is one of this.

If you do
```python
$VAR = 'Review'
echo "$VAR is good"
# Review is good
```
Then $VAR becomes callable you'll get without any error:
```python
echo "$VAR is good"
# <function __main__.<lambda>> is good
```
In case of exception you will go and fix the error.

jnoortheen:
but what If I expected callable ? in cases like $PROMPT,$RIGHT_PROMPT I expect to see a callable. Then one manipulates it right ?

anki-code:
The PROMPT variable traditionally has the string. When you replace it to the callable and then use it to ASYNC_PROMPT it's a question of design of achievething the goal. Probably the better way is to put `{async_prompt}` string into the PROMPT and call `async_prompt` from PROMPT_FIELDS on other place.

Back to this PR. I think you have no `echo $PROMPT` cases to catch the exception.

anki-code:
I lost motivation while thinking about this logic. Sorry.

It's not good to have callables that can uncontrolled be converted to strings. This is another one question that should be redisigned, reinvented, rethought. Some day...

anki-code:
Basically I see three different approaches:
1. **Transparent way**. If any object is callable and we expect another type (str, list, class) we should call it to get the actual value.
    * Pros: transparent, expected, extensible
    * Cons: The moment of executing the callable may be uncontrolled and for example if the callable depends on time we can get unexpected result when callable was executed in the past but we expect that it executes now. 
2. **The wall way**. If any object is callable and we expect another type (str, list, class) we should raise an exception to force explicit implementing the call from the developer i.e. replace `echo $VAR` to `echo @($VAR())`.
    * Pros: explicit calling
    * Cons: we should do the check in every plase where we could have callable (everywhere)
3. **Deprecation**. If someone try to set callable to env var raise an exception.
    * Pros: clean logic that told that variable can have only static value
    * Cons: no callable variables in the env, someone go to cry

jnoortheen:
IMO 1 is the way to go. Python zen has proved to be better than Ruby philosophy (of lots of magik).

PR #4170, opened 2021-03-11
docs: copy robots.txt

anki-code:
I've read about simpler way then html_extra_path. Let's try it first


gforsyth (approved):
definitely worth a shot!

PR #4172, opened 2021-03-12
docs: fix wizard text

anki-code:
Reduced expectations of implementing fish.

No news file.

### DIFF
--- a/xonsh/xonfig.py
+++ b/xonsh/xonfig.py
@@ -81,8 +81,8 @@ WIZARD_FS = """
 
                       {{BOLD_WHITE}}Foreign Shell Setup{{RESET}}
                       {{YELLOW}}-------------------{{RESET}}
-The xonsh shell has the ability to interface with foreign shells:
-Bash or zsh. The other shells can be implemented as well.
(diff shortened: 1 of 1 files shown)


xonsh/xonfig.py
```diff
@@ -81,8 +81,8 @@
 
                       {{BOLD_WHITE}}Foreign Shell Setup{{RESET}}
                       {{YELLOW}}-------------------{{RESET}}
-The xonsh shell has the ability to interface with foreign shells such
-as Bash, or zsh (fish not yet implemented).
+The xonsh shell has the ability to interface with foreign shells:
+Bash or zsh. The other shells can be implemented as well.
```
gforsyth, line 85:
```suggestion
The xonsh shell has the ability to interface with Bash or zsh
via the foreign shell interface.
```

Maybe something like this?  We mention the foreign shell interface but make no promises about adding new extensions.

anki-code: Sounds good!

gforsyth:
Thanks @anki-code !

PR #4175, opened 2021-03-14
feat: Added ability to set XONSH_HISTORY_FILE

anki-code:


gforsyth:
Hey @anki-code -- I'm fine with making this configurable, but I think the default behavior, at least with the json backend, would be to continuously overwrite the history file if this were used.
Given that, we should either have some guards against that, or a warning in the env-var description

anki-code:
@gforsyth

1. I've added more logic to json backend 

2. I've tested:
```python
XONSH_HISTORY_FILE='/tmp/hist_json.json' XONSH_HISTORY_BACKEND='json' xonsh --no-rc
XONSH_HISTORY_FILE='/tmp/hist_sqlite.db' XONSH_HISTORY_BACKEND='sqlite' xonsh --no-rc
# and also https://github.com/anki-code/xontrib-history-encrypt/
```

3. All works as expected:
    * If the history file does not exist it will be created
    * After exiting or on `history flush` the history will be appended to the file (nothing is overwritten / lost)

Please merge this pr and then merge 4181

anki-code:
@gforsyth please merge.
Take a look at minor docs changes at 4169, 4172, 4174. 
Thanks!

tests/test_shell.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+"""Testing for ``xonsh.shell.Shell``"""
+import os
+
+from xonsh.environ import Env
+from xonsh.shell import Shell
+from xonsh.history.json import JsonHistory
+
+def test_shell_with_json_history(xonsh_builtins, xonsh_execer, tmpdir_factory, monkeypatch):
```
gforsyth, line 9: Is `monkeypatch` being used here?

xonsh/environ.py
```diff
@@ -1398,7 +1398,7 @@ class PromptSetting(Xettings):
     )
     XONSH_HISTORY_FILE = Var.with_default(
         None,
-        "Location of history file that was set by history backend.",
+        "Location of history file that could be set in RC file or it will be set by history backend.",
```
gforsyth, line 1401:
```suggestion
        "Location of history file set by history backend (default) or set by user in RC file.",
```

I think we also want `is_configurable=True`

gforsyth (changes requested):
Thanks for putting this in @anki-code !  

I've requested a few tweaks -- and I think it would also be good to have a similar test for the sqlite history backend.

tests/test_shell.py
anki-code: nope, removed

anki-code:
@gforsyth Added tests for sqlite and for dummy in not interactive mode.

gforsyth (approved):
Woo! Thanks @anki-code !

PR #3632, opened 2020-06-22
#3402 set the exit code for single commands

dyuri:
Attempt to fix #3402 .

```
$ touch file_exists
$ xonsh -c "rm file_exists"
$ _.rtn
0
$ xonsh -c "rm no_such_file"
/usr/bin/rm: cannot remove 'no_such_file': No such file or directory
$ _.rtn
1
$ xonsh -c "sys.exit(123)"
$ _.rtn
123
```

(But it still does not seem to fix the issue with Fugitive.)

PR #4178, opened 2021-03-15
fix: Not load history in not interactive

### LINKED ISSUE #4177: Disable history loading in case of running the script or the command
Hi!

We found in https://github.com/xonsh/xonsh/issues/2496#issuecomment-798959368 that xonsh loads the history in case of running script and this produces some side effects.

Example for json backend:
```python
bash
cd /tmp
echo "echo 123" > e.xsh
XONSH_HISTORY_BACKEND=json xonsh e.xsh
## loading thousands of json files history
## running the script
# 123

XONSH_HISTORY_BACKEND=json xonsh -c 'echo 123'
## loading thousands of json files history
## running the command
# 123
```
Probably this is the cause of slow running the scripts in some cases.

We should disable history loading in case of running the script.

## For community

anki-code:
Closed #4177

### DIFF
--- a/xonsh/shell.py
+++ b/xonsh/shell.py
@@ -14,6 +14,7 @@ from xonsh.platform import (
 )
 from xonsh.tools import XonshError, print_exception, simple_random_choice
 from xonsh.events import events
+from xonsh.history.dummy import DummyHistory
 import xonsh.history.main as xhm
 
 
@@ -208,7 +209,8 @@ class Shell(object):
             )
             env["XONSH_HISTORY_FILE"] = hist.filename
         else:
-            builtins.__xonsh__.history = hist = env["XONSH_HISTORY_FILE"] = None
+            builtins.__xonsh__.history = hist = DummyHistory()
+            env["XONSH_HISTORY_FILE"] = None
 
         shell_type = self.choose_shell_type(shell_type, env)
 
--- a/xonsh/shell.py
+++ b/xonsh/shell.py
@@ -209,8 +209,7 @@ class Shell(object):
             )
             env["XONSH_HISTORY_FILE"] = hist.filename
         else:
-            builtins.__xonsh__.history = hist = DummyHistory()
-            env["XONSH_HISTORY_FILE"] = None
+            builtins.__xonsh__.history = hist = env["XONSH_HISTORY_FILE"] = None
 
         shell_type = self.choose_shell_type(shell_type, env)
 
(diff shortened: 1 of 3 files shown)


xonsh/shell.py
```diff
@@ -200,11 +201,16 @@ def __init__(self, execer, ctx=None, shell_type=None, **kwargs):
...
+
         # build history backend before creating shell
-        builtins.__xonsh__.history = hist = xhm.construct_history(
-            env=env.detype(), ts=[time.time(), None], locked=True
-        )
-        env["XONSH_HISTORY_FILE"] = hist.filename
+        if env.get("XONSH_INTERACTIVE"):
+            builtins.__xonsh__.history = hist = xhm.construct_history(
+                env=env.detype(), ts=[time.time(), None], locked=True
+            )
```
jnoortheen, line 209: could you check with lazyobj?

anki-code: @jnoortheen nice thought but we have no guarantee that history will not be called in the xonsh code in the future. Any developer can add calling the history and we will catching side effects again and without any warnings. 

To add backwards comaptibility I replaced history to DummyHistory in case of not interactive mode.

anki-code:
The one test fails after setting history backend to None in not interactive:
```python

xonsh -c "sh -c 'exit 100'"
__xonsh__.history[-1].rtn
# 0 instead of 100
```
No errors. No warnings. Any thoughts?

anki-code:
The one test fails after setting history backend to None in not interactive:
```python

xonsh -c "sh -c 'exit 100'"
__xonsh__.history[-1].rtn
# 0 instead of 100
```
No errors. No warnings. Any thoughts?

anki-code:
Hi @dyuri! The test above added by #3632. Could you please take a look how None-history backend could affect on this?

anki-code:
@jnoortheen the case above demonstrates what side effects we will have in case of using LazyJson in history.

anki-code:
So if we want None-history we should add `else` clause here:

https://github.com/xonsh/xonsh/blob/f32c36b600eaf780b6f1787f54d3de085c04acc0/xonsh/main.py#L462-L466

Any thoughts?

anki-code:
After research I've found that xonsh uses history backend as storage for the `last_cmd_rtn` in many places so the solution with Dummy history is most native.

anki-code:
@gforsyth We have consensus here. Ready to merge :)

gforsyth (approved):
Looks good to me.  Thanks @anki-code

PR #4181, opened 2021-03-16
fix: fix gc check

anki-code:
This is the small fix for #4178 that bring back the previous logic in the check to properly load gc.
@gforsyth please merge asap.
No news file


gforsyth (approved):
thanks!

PR #4190, opened 2021-03-17
#4148 $TITLE can be set to a function

### LINKED ISSUE #4148: Title isn't behaving as expected (prompt)
## xonfig

<details>

```
$ xonfig
| xonsh            | 0.9.27          |
| Python           | 3.9.2           |
| prompt toolkit   | 3.0.14          |
| pygments         | 2.7.4           |
| distro           | unknown         |
| xontrib          | []              |
```

</details>

## Expected Behavior

I'd like to use your `_dynamically_collapsed_pwd` function with slight modification for my terminal title, however, it seems to behave a bit different vs. the prompt function, where I managed to make it work

```python
from xonsh.prompt.cwd import _dynamically_collapsed_pwd
def rprompt():
  cwd = _dynamically_collapsed_pwd().removeprefix('/mnt/')
  out = cwd + '|'
  return out
```
I'd expect this function to use your `cwd`, but without the `/mnt/` prefix

## Current Behavior
This works fine for a prompt:
`$RIGHT_PROMPT = rprompt`
Right prompt is using your function without the prefix

However, this fails for a title
`$TITLE = rprompt`
Title displays `<function rprompt a...>`, so the function isn't expanded

## Steps to Reproduce

Define the function above in xonsh and assign it to the `$TITLE` env var

## For community

dyuri:
According to the documentation $TITLE should work like $PROMPT but it was restricted to be `string` instead of `string_or_callable`.

Fixes #4148 .

### DIFF
--- a/news/fix_prompt_like_title.rst
+++ b/news/fix_prompt_like_title.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Functions can be used for $TITLE, the same way as for $PROMPT. (#4148)
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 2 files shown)


gforsyth (approved):
thanks @dyuri!

PR #4196, opened 2021-03-19
Improve `which` output

daniel-shimon:
```python
aliases['a'] = 'ls'
aliases['b'] = 'ls | grep b'  # ExecAlias
aliases['c'] = lambda args: print(args)

```

Before:
```python
which a b c
# ls
# b
# c

```

After:
```python
which a b c
# ls
# ls | grep b
# <function <lambda> at 0x7f4d42bdc1f0>

```

### DIFF
--- a/news/improve-which-output.rst
+++ b/news/improve-which-output.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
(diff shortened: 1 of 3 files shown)


gforsyth (approved):
Looks good, thanks @daniel-shimon !

PR #4201, opened 2021-03-24
feat: Added ability to get arguments list in ExecAlias

anki-code:
This PR adds `$args` and `$arg0`..`$arg<n>` environment variables to the `ExecAlias` context. Motivation: https://github.com/xonsh/xonsh/pull/4201#issuecomment-810754606

Exampls:
```python
aliases['my'] = "echo Hello @($args)"    # ExecAlias
my 'world!'
# Hello world!
```
```python
aliases['cl'] = "cd $arg0 && ls $arg0"
cl /tmp
```

## For community

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1230,14 +1230,14 @@ being executed.
 
 If the string is representing a block of xonsh code, the alias will be registered
 as an ``ExecAlias``, which is a callable alias. This block of code will then be
-executed whenever the alias is run. The arguments are available in ``$ALIAS_ARGS``
-and ``$ALIAS_ARG<n>`` environment variables.
+executed whenever the alias is run. The arguments are available in the list ``$args``
+or by the index in ``$arg<n>`` environment variables.
 
 .. code-block:: xonshcon
 
     >>> aliases['answer'] = 'echo @(21+21)'
-    >>> aliases['piu'] = 'pip install -U @($ALIAS_ARGS)'
-    >>> aliases['cdls'] = 'cd $ALIAS_ARG0 && ls'
+    >>> aliases['piu'] = 'pip install -U @($args)'
+    >>> aliases['cdls'] = 'cd $arg0 && ls'
 
 .. note::
 
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1206,19 +1206,17 @@ matching only occurs for the first element of a subprocess command.
 
 The keys of ``aliases`` are strings that act as commands in subprocess-mode.
 The values are lists of strings, where the first element is the command, and
-the rest are the arguments. You can also set the value to a string, in which
-one of two things will happen:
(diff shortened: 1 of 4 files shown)


anki-code:
@gforsyth please take a look ;)

gforsyth:
Hey @anki-code !  
Is this inherently different from passing through alias arguments using the `def _func(args):` callable aliases?  Or is it the same (ish) and providing a simpler syntax for handling that?

anki-code:
Yep, this is the simpler syntax. For example you want to create alias that do `cd` and `ls`:

You can't write one line using `$[]` because it has no `&&` support:
```python
aliases['cdls'] = $[cd @(args[0]) && ls @(args[0])]
# Syntax error around &&
```

You can write 4 lines:
```python
def _cd_and_ls(args):
    cd @(args[0])
    ls @(args[0])
aliases['cdls'] = _cd_and_ls 
```

Also you can write the syntax overloaded example:
```python
aliases['cdls'] = lambda args: execx(f"cd @({repr(args[0])}) && ls @({repr(args[0])})")
```

This PR allows to make ExecAlias completed by adding ability to use args and to write the sane alias:
```python
aliases['cdls'] = "cd $ALIAS_ARG0 && ls $ALIAS_ARG0"
```

Naming. Instead of `$ALIAS_ARGS` I thought about using `$ARGS` and `$ARG1`..`$ARGn` like in non-interactive xsh. But we can use aliases in the scripts and potentially we can want to use `$ARGS` of the script in the aliases. I prefer to make the name shorter but have no idea how to keep it sane. New shorter naming is welcome.

anki-code:
The tests looks broken in the main branch.

gforsyth:
Hey @anki-code -- thanks for the explanation!  I agree that this fits a model that is currently really awkward with `xonsh`, so I'm game to add it in.  
I'd like to sit and think a little bit -- I would love if we could re-use the `ARGn` identifiers, but you make a good point about potential collisions.

Either way we'd want to add a section to the docs about how to use this.

And yeah, not sure what's going on with the test suite.  That's a new wacky failure for me (or I've forgotten 😅 )

anki-code:
Eurika! What about `$args` in lowercase?

```python
aliases['cdls'] = "cd $arg0 && ls $arg0"
```

anki-code:
@gforsyth we're on the way to finish line here ;)

jnoortheen (approved):

jnoortheen:
> @jnoortheen -- can you take a look at this and see what you think?

It all looks good, with small change we get a shorter way to specify alias function. Though I prefer having only one way to do things,  it looks reasonable addition (more bashy ).

anki-code:
Hi @gforsyth! Let's go forward! :)

gforsyth:
Thanks @anki-code !

gforsyth:
And thanks for the review @jnoortheen !

PR #4202, opened 2021-03-25
Updated references to point to 'main' instead of 'master'

eugenesvk:
Couldn't update to `HEAD` since the link was pointing to the outdated `master.zip`, so changed that and also edited a few other outdated references

Didn't add a changelog entry as I think these edits are too minor to mention, but let me know if you think otherwise


gforsyth:
Awesome!  Thanks for catching all of those @eugenesvk !

gforsyth (approved):

PR #4211, opened 2021-03-31
Put history json to subdirectory

anki-code:
Hi!
To keep `XONSH_DATA_DIR` clean the history json files now will be in `XONSH_DATA_DIR/history_json` directory.
The backwards compatibility of reading the history json from `XONSH_DATA_DIR` is on board.

## For community

### DIFF
--- a/news/json_history_dir.rst
+++ b/news/json_history_dir.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* New json history will be in XONSH_DATA_DIR/history_json directory.
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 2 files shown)


anki-code:
@gforsyth my XONSH_DATA_DIR want to tell thanks for this PR :)

gforsyth:
Thanks for this @anki-code -- I'm good with creating a subdirectory (so the code and script cache directories are more easily visible).  I know "flat is better than nested" but perhaps we should create a generic `history/` subdirectory and stick both json and sqlite files in there?

anki-code:
Hi @gforsyth ! Thank you for the review!

I thought about it and found that:
* we already have `xonsh_code_cache` and `xonsh_script_cache` as the distinct directories instead of in `cache/`.
* we will not have big amount of history/cache backends to make subdirectories for them. 

So I prefer to do the one step for improvement the history json and the data directory already will look very clean.

gforsyth (approved):
Thanks for this, @anki-code !  

I've tested it locally and it loads history from the old and new locations just fine.

Issue #4213, opened 2021-03-31
Remove ./ and ../ from cd and path completion


anki-code:
Hi!

What about to remove `./` and `../` from the completion? 

![image](https://user-images.githubusercontent.com/1708680/113190898-90957c00-9265-11eb-9c28-dae6846bee84.png)

It's annoying when the directory has the one subdirectory and it's needed to see the list of `['../','./','subdir']` to choose the `subdir`.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

Baileymnm1999:
@anki-code Does this accomplish what you would like?

PR #4215, opened 2021-04-01
Added env variable to ignore ./ and ../ in filepath completion

Baileymnm1999:
## For community
**Added:**

* Added environment variable `$IGNORE_DOTS` to ignore ./ and ../ in file path completions.

This addresses issue  #4213 

### DIFF
--- a/news/ignore_dots.rst
+++ b/news/ignore_dots.rst
@@ -1,6 +1,9 @@
 **Added:**
 
-* Added environment variable to ignore ./ and ../ in file path completions.
+* Added environment variable `$COMPLETE_DOTS` to specify how current and previous directories should be tab completed in cd  ('./', '../'):
+    - ``always`` Always complete paths with ./ and ../
+    - ``never`` Never complete paths with ./ and ../
+    - ``matching`` Complete if path starts with . or ..
 
 **Changed:**
 
--- a/news/ignore_dots.rst
+++ b/news/ignore_dots.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Added environment variable to ignore ./ and ../ in file path completions.
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/xonsh/completers/path.py
+++ b/xonsh/completers/path.py
@@ -124,13 +124,16 @@ def _env(prefix):
 
 
 def _dots(prefix):
-    ignore_dots = bool(builtins.__xonsh__.env.get("IGNORE_DOTS"))
-    if ignore_dots:
+    complete_dots = builtins.__xonsh__.env.get("COMPLETE_DOTS").lower()
+    if complete_dots == 'never':
         return ()
     slash = xt.get_sep()
     if slash == "\\":
         slash = ""
-    if prefix in {"", "."}:
+    prefixes = {"."}
+    if complete_dots == 'always':
+        prefixes.add("")
+    if prefix in prefixes:
         return ("." + slash, ".." + slash)
     elif prefix == "..":
         return (".." + slash,)
--- a/xonsh/completers/path.py
+++ b/xonsh/completers/path.py
@@ -124,7 +124,7 @@ def _env(prefix):
 
 
 def _dots(prefix):
-    complete_dots = builtins.__xonsh__.env.get("COMPLETE_DOTS").lower()
+    complete_dots = builtins.__xonsh__.env.get("COMPLETE_DOTS", "matching").lower()
     if complete_dots == 'never':
         return ()
     slash = xt.get_sep()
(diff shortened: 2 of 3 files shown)


anki-code:
Awesome! Thanks! Please add the default value i.e. `env.get("IGNORE_DOTS", False)`. In some test cases it's good to avoid the value error.

gforsyth (changes requested):
Thanks for putting this in @Baileymnm1999 !

I am actually just going to paste out @anki-code 's review here, because he hit all the points I was going to raise 😅 

>Please add the default value i.e. env.get("IGNORE_DOTS", False). In some test cases it's good to avoid the value error.
> It will be cool to have the name of the variable in the news.
> And also Gil will ask you about adding the env variable to the white list. 

Regarding the last point -- it should be defined where @anki-code indicated and it should have a default value.  Usually I would have the default value reflect current behavior, but in this case I'm leaning towards dropping the dots by default as offering tab completion for them doesn't really offer any benefits in terms of saved keystrokes.

anki-code:
> I'm leaning towards dropping the dots by default as offering tab completion for them doesn't really offer any benefits in terms of saved keystrokes.

May be just drop them without env variable :)

gforsyth:
Hmm, that's a good point @anki-code.

How about we don't include any `.` results unless a user has already typed one?

e.g.:

in an empty directory

`cd <tab>` -> no completions

`cd .<tab>` -> `./` and `../`

@Baileymnm1999 -- would you be up for tackling this?

Baileymnm1999:
@gforsyth what do you think of changing the environment variable to `COMPLETE_DOTS = {always, never, matching}` or something similar with it defaulting to matching?

Baileymnm1999 (commented):
Here are the changes to address all the coments

Baileymnm1999:
I am not sure why qa checks are failing, pep8 linter doesn't show me anything

anki-code:
You can click on Details and get the error in tests:
```
tests/completers/test_dir_completers.py
# ...
    def _dots(prefix):
>       complete_dots = builtins.__xonsh__.env.get("COMPLETE_DOTS").lower()
E       AttributeError: 'NoneType' object has no attribute 'lower'

/usr/share/miniconda/envs/xonsh-test/lib/python3.7/site-packages/xonsh/completers/path.py:127: AttributeError
```
Then you can run the test locally:
```python
pytest tests/completers/test_dir_completers.py
```
The error is about you should set the default value for the env variable i.e. `get("COMPLETE_DOTS", "matching")`.

Baileymnm1999:
I think this change affects expected functionality in the unit tests. Should they be changed?

Baileymnm1999:
Sorry, for the difficulties and thanks for the patience. This is my first ever attempted pr on github.

anki-code:
Okay, we're very close to success. 

Check the tests details. After short look I see two errors:

1. The syntax in the news file is wrong for RST. I recommed to simplify the news: `Added environment variable COMPLETE_DOTS to specify how current and previous directories should be tab completed in cd.` (without back quotes)

2. The test `tests/completers/test_base_completer.py` should be fixed.

anki-code:
Hi @Baileymnm1999 ! Please pull the code from the main branch the fix for one test that failed before. And it will be great to continue the work on this feature! I see many thumbs up on the first message, it's great!

news/ignore_dots.rst
```diff
@@ -0,0 +1,26 @@
+**Added:**
+
+* Added environment variable `$COMPLETE_DOTS` to specify how current and previous directories should be tab completed in cd  ('./', '../'):
```
gforsyth, line 3:
```suggestion
* Added environment variable ``$COMPLETE_DOTS`` to specify how current and previous directories should be tab completed in cd  ('./', '../'):
```

rst is very picky about the number of backticks

xonsh/completers/path.py
```diff
@@ -124,10 +124,16 @@ def _env(prefix):
 
 
 def _dots(prefix):
+    complete_dots = builtins.__xonsh__.env.get("COMPLETE_DOTS", "matching").lower()
+    if complete_dots == 'never':
```
gforsyth, line 128:
```suggestion
    if complete_dots == "never":
```

We use `black` to autoformat all python files and it prefers double-quotes to single (unless they're nested)

xonsh/completers/path.py
```diff
@@ -124,10 +124,16 @@ def _env(prefix):
...
+        return ()
     slash = xt.get_sep()
     if slash == "\\":
         slash = ""
-    if prefix in {"", "."}:
+    prefixes = {"."}
+    if complete_dots == 'always':
```
gforsyth, line 134:
```suggestion
    if complete_dots == "always":
```

gforsyth (changes requested):
Hey @Baileymnm1999 -- this looks good and it's very close!  I think just a few small formatting fixes and a merge from the `main` branch to fix a flaky test and you should be good to go.

gforsyth:
> I think this change affects expected functionality in the unit tests. Should they be changed?

If the test expects the `.`, `..` to be in the results, it's ok to remove that from the expected results as that will reflect the new expected behavior.

PR #4217, opened 2021-04-02
Add xfail for windows + py3.8 for strange test failure

gforsyth:
Marking `test_main.py::test_rc_with_modules` as an `xfail` for Python 3.8 on Windows until we have time to dig in to the weirdness causing it to fail.

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -70,6 +70,7 @@ def test_premain_custom_rc(shell, tmpdir, monkeypatch):
     assert f.strpath in builtins.__xonsh__.env.get("XONSHRC")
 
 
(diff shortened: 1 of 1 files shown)


daniel-shimon:
Since the error is inside the fixture, the `xfail` doesn't seem to help.
Probably using an `xpass` will solve this :)

daniel-shimon (approved):
Great!

gforsyth:
Sigh.  I'll have to dig into this more later this week.

anki-code:
@gforsyth I've fixed this PR. Please merge #4223 - it's the same as this PR but with the fix.

PR #4218, opened 2021-04-03
feat: Ability to call the program by name from callable alias with the same name without the infinite loop error

### LINKED ISSUE #3159: Infinite alias call
I'm trying to override 'ls' command to display dotfiles in my $DOTFILES directory.
This code goes into an endless loop because _ls function calls ls command and it calls _ls function.

```python
def _ls():
    if $(pwd).rstrip(os.linesep) == $DOTFILES:
        ls -Ga
    else:
        ls -G
aliases['ls'] = _ls
```

## xonfig

<details>

```
| xonsh            | 0.8.10              |
| Git SHA          | 2cb42bdb            |
| Commit Date      | Feb 6 16:49:16 2019 |
| Python           | 3.6.4               |
| prompt toolkit   | 2.0.7               |
| pygments         | 2.3.1               |
```

</details>

## Current Behavior

When I input "ls" command, the terminal freezes.

anki-code:
Closes #3159

Now we have transparent way to wrap commands into callable aliases.

Example 1:
```python
def _echo(args):
    echo @(args)
aliases['echo'] = _echo
echo ok
# Before: freezed on infinite loop
# After: ok
```

Example 2 - elegant ExecAlias (it will be awesome with #4201):
```python
aliases['echo'] = "echo @('ok')"  # ExecAlias
echo
# Before: freezed on infinite loop
# After: ok
```

Example 3 - show error for the user without freezing
```python
aliases['first'] = "second @(1)"  # ExecAlias just for demo
aliases['second'] = "first @(1)"  # ExecAlias just for demo
first
# Before: freezed on infinite loop
# After: Exception: Infinite loop of calls for "first" alias.
```

## For community

### DIFF
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -128,7 +128,6 @@ echo
     # testing alias stack: callable alias (ExecAlias) + no binary location + infinite loop
     (
         """
-# skip_if_on_windows        
 aliases['first'] = "second @(1)"
 aliases['second'] = "first @(1)"
 first
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -102,40 +102,6 @@ f
         "hello\n",
         0,
     ),
-    # testing alias stack: lambda function
-    (
-        """
-# skip_if_on_windows
-def _echo():
-    echo hello
-
-aliases['echo'] = _echo
-echo
-""",
-        "hello\n",
-        0,
-    ),
-    # testing alias stack: ExecAlias
-    (
-        """
-# skip_if_on_windows        
-aliases['echo'] = "echo @('hello')"
-echo
-""",
-        "hello\n",
-        0,
-    ),
-    # testing alias stack: callable alias (ExecAlias) + no binary location + infinite loop
-    (
-        """
-# skip_if_on_windows        
-aliases['first'] = "second @(1)"
-aliases['second'] = "first @(1)"
-first
-""",
-        lambda out: 'Recursive calls to "first" alias.' in out,
-        0,
-    ),
     # test redirecting a function alias to a file
     (
         """
@@ -506,19 +472,56 @@ a
     ),
 ]
 
+UNIX_TESTS = [
+    # testing alias stack: lambda function
+    (
+        """
+def _echo():
+    echo hello
+
+aliases['echo'] = _echo
+echo
+""",
+        "hello\n",
+        0,
+    ),
+    # testing alias stack: ExecAlias
+    (
+        """
+aliases['echo'] = "echo @('hello')"
+echo
+""",
+        "hello\n",
+        0,
+    ),
+    # testing alias stack: callable alias (ExecAlias) + no binary location + infinite loop
+    (
+        """
+aliases['first'] = "second @(1)"
+aliases['second'] = "first @(1)"
+first
+""",
+        lambda out: 'Recursive calls to "first" alias.' in out,
+        0,
+    ),
+]
 
 @skip_if_no_xonsh
 @pytest.mark.parametrize("case", ALL_PLATFORMS)
 def test_script(case):
     script, exp_out, exp_rtn = case
-    if ON_WINDOWS and 'skip_if_on_windows' in script:
-        return
     out, err, rtn = run_xonsh(script)
     if callable(exp_out):
         assert exp_out(out)
     else:
         assert exp_out == out
     assert exp_rtn == rtn
+
+@skip_if_no_xonsh
+@skip_if_on_windows
+@pytest.mark.parametrize("case", UNIX_TESTS)
+def test_unix_tests(case):
+    test_script(case)
 
 
 ALL_PLATFORMS_STDERR = [
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -133,7 +133,7 @@ aliases['first'] = "second @(1)"
 aliases['second'] = "first @(1)"
 first
 """,
-        lambda out: 'Infinite loop of calls for "first" alias.' in out,
+        lambda out: 'Recursive calls to "first" alias.' in out,
         0,
     ),
     # test redirecting a function alias to a file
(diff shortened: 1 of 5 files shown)


gforsyth:
Thanks for tackling this, @anki-code !  

I think the general solution you have here is correct.  The usage of a "private" environment variable to handle it seems a little messy to me, though.  Do you think you can implement this with an attribute in the `spec`?

anki-code:
Hi @gforsyth ! Thank you for the review!

I'm using environment variable because we have different types of aliases i.e. `ExecAlias` that can execute subcommands in the distinct process that has less control from parent.

gforsyth:
Hey @anki-code -- sorry for the delay.  Anytime we mess with the proxies I need to take a closer look and I'm a bit slammed for time.  I'm going to try to get to this later this week

anki-code:
Hi @gforsyth! It's kind reminder :)

gforsyth:
Hey @scopatz -- dunno if you're around but if you can take a quick peek at this, I'd appreciate it.

anki-code:
I've added master and broken tests looks unrelated to this issue. @jnoortheen it's new tests from 4106 please take look.

anki-code:
@gforsyth I think @scopatz is far far in a galaxy :)
We can wait but it will be good to have this nice fix in the next release.

xonsh/procs/proxies.py
```diff
@@ -500,10 +500,16 @@ def run(self):
...
+            if self.env["ALIAS_NAME"]:
+                alias_stack += ":" + self.env["ALIAS_NAME"]
+
             with STDOUT_DISPATCHER.register(sp_stdout), STDERR_DISPATCHER.register(
                 sp_stderr
             ), xt.redirect_stdout(STDOUT_DISPATCHER), xt.redirect_stderr(
                 STDERR_DISPATCHER
+            ), builtins.__xonsh__.env.swap(
+                ALIAS_STACK=alias_stack
             ):
```
jnoortheen, line 513: can this be more static? instead of relying on env variable as suggested by gforsyth.

anki-code: The short answer is no because we have ExecAlias. As you can see we use the env variable temporarily only during the executing the command.

anki-code:
@gforsyth does this pr have a chance to be released? :)

xonsh/procs/specs.py
```diff
@@ -589,17 +592,29 @@ def redirect_trailing(self):
...
 
     def resolve_binary_loc(self):
         """Sets the binary location"""
         alias = self.alias
         if alias is None:
-            binary_loc = xenv.locate_binary(self.cmd[0])
+            cmd0 = self.cmd[0]
+            binary_loc = xenv.locate_binary(cmd0)
+            if binary_loc == cmd0 and cmd0 in self.alias_stack:
+                raise Exception(f'Infinite loop of calls for "{cmd0}" alias.')
```
daniel-shimon, line 617:
```suggestion
                raise Exception(f'Recursive calls to "{cmd0}" alias.')
```

We don't know it's infinite, only that it's recursive :)

tests/test_integrations.py
```diff
@@ -477,8 +511,13 @@ def _echo(args):
 @pytest.mark.parametrize("case", ALL_PLATFORMS)
 def test_script(case):
     script, exp_out, exp_rtn = case
+    if ON_WINDOWS and 'skip_if_on_windows' in script:
```
daniel-shimon, line 514: Use `pytest.param` with `skip_if_on_windows` instead

xonsh/procs/proxies.py
```diff
@@ -500,10 +500,16 @@ def run(self):
             sp_stderr = sys.stderr
         # run the function itself
         try:
+            alias_stack = builtins.__xonsh__.env.get("ALIAS_STACK", "")
+            if self.env["ALIAS_NAME"]:
```
daniel-shimon, line 504:
```suggestion
            if self.get("ALIAS_NAME"):
```

xonsh/procs/specs.py
```diff
@@ -442,6 +444,7 @@ def run(self, *, pipeline_group=None):
         kwargs = {n: getattr(self, n) for n in self.kwnames}
         self.prep_env(kwargs)
         if callable(self.alias):
+            kwargs["env"]["ALIAS_NAME"] = self.alias_name
```
daniel-shimon, line 447: I'm fine with setting env vars for this, as long as we use names that are obviously internal like `__XONSH_ALIAS_NAME` and `__XONSH_ALIAS_STACK`

xonsh/procs/proxies.py
```diff
@@ -500,10 +500,16 @@ def run(self):
...
+                alias_stack += ":" + self.env["ALIAS_NAME"]
+
             with STDOUT_DISPATCHER.register(sp_stdout), STDERR_DISPATCHER.register(
                 sp_stderr
             ), xt.redirect_stdout(STDOUT_DISPATCHER), xt.redirect_stderr(
                 STDERR_DISPATCHER
+            ), builtins.__xonsh__.env.swap(
+                ALIAS_STACK=alias_stack
```
daniel-shimon, line 512: Don't we need to add `ALIAS_NAME=""` here as well?

daniel-shimon (changes requested):
This is great! A few minor things and I think we'll be ok

tests/test_integrations.py
anki-code: Please take a look to the `test_integrations.py`. It has a different structure than other test files.

xonsh/procs/proxies.py
anki-code: Nope because we need only stack. If we will have an alias in further calls it will be just checked and added to stack.

anki-code:
@daniel-shimon, all comments are done.

daniel-shimon:
@anki-code FYI you can wrap parameterized values with `pytest.param`:

```
@pytest.mark.parametrize("case", 
    "will always run", 
    pytest.param("will run on unix", marks=skip_if_on_windows), 
) 
```

daniel-shimon (approved):
Anyway I think this PR can be merged :)

anki-code:
@gforsyth we're ready to merge :)

PR #4223, opened 2021-04-06
Fix 4217

anki-code:
Fix #4217
No news


gforsyth:
Thanks @anki-code !

gforsyth:
Also, TIL that you can autoclose PRs via # references

PR #4230, opened 2021-04-11
Do not repeat commands when searching them in history (using up arrow)

jmoranos:
If you execute the same command N times, it appears also N times when using up arrow to search previous commands. It is more convenient for commands not to be repeated during this type of search (as, for example, in fish shell).
I have used a python set to solve this issue, although there may be a better solution.

## For community

### DIFF
--- a/xonsh/ptk_shell/history.py
+++ b/xonsh/ptk_shell/history.py
@@ -25,8 +25,12 @@ class PromptToolkitHistory(prompt_toolkit.history.History):
         hist = builtins.__xonsh__.history
         if hist is None:
             return
+        # Do not repeat commands when searching them in history (using up arrow)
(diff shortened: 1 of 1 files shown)


anki-code:
Hi!

> If you execute the same command N times, it appears also N times when using up arrow to search previous commands.

There is an option to ignore duplicates using [$HISTCONTROL](https://xon.sh/envvars.html#histcontrol):
```
$HISTCONTROL='ignoredups,erasedups'
```

jmoranos:
Thank you very much. It works perfectly.
Sorry for the inconvenience.

anki-code:
@jmoranos don't worry! Thank you for your good intention! Feel free to ask questions and make improvements!

PR #4233, opened 2021-04-17
DOC: Autoreformat most of the docs.

Carreau:
this uses https://github.com/Carreau/velin to try to autoreformat all
docstrings in xonsh/*.py

In particular in Parameters section the space before the colon is
necessary for numpydoc to properly find the name and the type.

When no type is there numpydoc says that colon should be removed.

This also try to find if there are typo in parameter names and fix them
when possible.

As a sideeffect of parsing with numpydoc and reconstructing the
docstring from the parsed data some whitespace information is lost; but
that has the advantage of making the whitespace rules uniform
everywhere.

## For community

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -661,7 +661,6 @@ def xexec(args, stdin=None):
     If '-a' is supplied, the shell passes name as the zeroth argument
     to the executed command.
 
-
     Notes
     -----
     This command **is not** the same as the Python builtin function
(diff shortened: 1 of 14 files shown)


jnoortheen:
This looks good. can this be included as part of tooling? like flake8 checks

Carreau:
> This looks good. can this be included as part of tooling? like flake8 checks

Probably once it's a bit more stable, but yes you can already and you could also set it as a pre commit.

gforsyth (approved):
Lovely, thanks @Carreau !

PR #4257, opened 2021-05-06
Fix commands cache testing

jnoortheen:
Also it has some other works 

1. update mypy and black. Not including news since it is mostly internal
2. fix ply `output_dir` different from default `tab_module` (you could see in the screenshot, how it generates parser tables under directory `xonsh/parsers` instead of `xonsh/` before this fix.)

image

## For community

### DIFF
--- a/requirements/docs.txt
+++ b/requirements/docs.txt
@@ -8,7 +8,5 @@ pyzmq
 matplotlib
 doctr
 tornado
-black
-pre-commit
 runthis-sphinxext
 livereload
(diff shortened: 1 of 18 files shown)


anki-code:
```python
cd /tmp
git clone https://github.com/jnoortheen/xonsh -b fix-commands-cache-testing
cd xonsh
for i in range(0,30):
      pytest tests/test_commands_cache.py
```
This works good for me. Many thanks @jnoortheen !
@gforsyth let's merge to fix test failing.

gforsyth:
thanks @jnoortheen

PR #4278, opened 2021-05-11
Fix annotated assignment

### LINKED ISSUE #3959: xonsh does not support type hinting
I'm writing a plugin for Xonsh in pure Python using the Xonsh library. I've added [type hints](https://docs.python.org/3/library/typing.html) so that I can use `mypy` and other static checkers on it.

However, it seems Xonsh does not support this. Type hints not only improve correctness, but also serve as a form of documentation. Since Xonsh is intended to be a superset of Python, it would be nice if Xonsh adds support for type hints.

**Example:**

This works:
```shell
python3 -c "x: int = 5" 
```

This doesn't:
```shell
xonsh -c "x: int = 5" 
```

daniel-shimon:
Closes #3959.

Enables the following:
```python
x: int = 42
x: int
```

## For community

### DIFF
--- a/news/annotated-assignment.rst
+++ b/news/annotated-assignment.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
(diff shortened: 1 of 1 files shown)


gforsyth (approved):
lgtm, thanks @daniel-shimon !

Issue #4096, opened 2021-02-08
startup files and XONSH_INTERACTIVE


bestlem:
I am trying to use xonsh as my shell.

Interactive use works well and as expected.

But xonsh does not seem to read any startup files if it is launched non interactively. This differs from POSIX shells and fish where the .profile .zshenv and other config files are read in all cases.

I want to be able to control my configuration when I run say a compile process from my editor or from cron and then test on $XONSH_INTERACTIVE to control what is done.

How do I read startup files when run such that $XONSH_INTERACTIVE == False

I know there there is a --rc flag but not all tools all  extra parameters to be passed in other than the command to be run.

gforsyth:
Hi @bestlem -- there's a `-i` / `--interactive` flag to force running in interactive mode.  Does that cover your request?

bestlem:
No sorry I want to run xonsh in non-interactive mode but load the startup files e.g. ~/.config/rc.xsh - I can then test in the file to see if interactive or not

gforsyth:
Got it.  We removed static config files a while back, but sounds like we might want to bring those back.

bestlem:
I think we are still at cross purposes here or else we are using different names for startup files. The files I am talking about are still there and used by xonsh.

The initialisation files e.g .xonshrc ~/.config/rc.xsh i.e. the files listed in $XONSHRC are still picked up if you don't provide a command or file on the command line. xonsh will then read the file and set $XONSH_INTERACTIVE from other shells I would expect these to be independent actions

Although looking at history this decision was made long ago for 0.6.0 The changes start at 1f464fb1939b6b1b68d10fe02735e80507900b21 

I disagree with the comment - but that does need to be checked. I tried ```zsh -c "echo $PATH" ``` and ```fish -c "echo $PATH" ``` and they loaded their config files when I put an echo "In shell" in their config files.

Interactive is not the same as as reading startup files.

Comment is
Don't load xonshrc is xonsh is running a script

As per discussions with @asmeurer -- if xonsh is being used to launch a script,
e.g.

```
xonsh run_my_cool_deploy.xsh
```

then it should not load the `xonshrc` file since this should be reserved for
interactive functionality and we don't want side effects in the script.

Basically, if you need a xontrib to be loaded for your script, load it in your
script

bestlem:
Looking at other issues especially #3176 Interactive and possibly login shells are not used much in xonsh and have some rough edges.

anki-code:
Also similar issue #3579

[...]

PR #4279, opened 2021-05-11
Test RC loading in different startup configurations

chronitis:
This is a followup to @anki-code in https://github.com/xonsh/xonsh/pull/4256 and https://github.com/xonsh/xonsh/issues/4096#issuecomment-791991841

This adds a parameterised test with all the combinations of startup flags from #4096, and a few more, and checks which RC files are loaded. It should enforce the current behaviour, although having expanded many cases, I think some are probably arguable (eg, an explicit RC file `xonsh --rc foo.xsh bar.xsh` is not loaded unless you also specify interactive mode, which is probably not obvious).

### DIFF
--- a/news/test_startup_order.rst
+++ b/news/test_startup_order.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Test consistent RC loading behaviour in a variety of startup scenarios
+
+**Security:**
+
+* <news item>
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -141,6 +141,9 @@ def test_rcdir_empty(shell, tmpdir, monkeypatch, capsys):
 # the parameterisation is a list of xonsh args, followed by the list
 # of RC files (see function body) expected to be loaded in order
 # note that a tty is not faked, so this will default to non-interactive
+# (skipped on windows because XONSHRC="/path/to/f1.xsh:/path/to/f2.xsh"
+# doesn't appear to work, while it does on other platforms)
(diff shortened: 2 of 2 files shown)


daniel-shimon:
This parametrization is great! Thanks @chronitis!
What about testing the XONSHRC directories?

chronitis:
> What about testing the XONSHRC directories?

@daniel-shimon can you expand on what you mean here? There are other tests already in `test_main` covering `XONSHRC`, and this one does test that 1) multiple files can be specified in `XONSHRC` and 2) they will be loaded in the given order. Testing that the expected filesystem paths (`/etc/xonshrc, XDG_CONFIG_HOME/xonsh/rc.xsh, HOME/.xonshrc`) are considered if XONSHRC isn't set could be done elsewhere, but I don't think can easily be combined here without a lot more filesystem mocking and cleanup than is currently done.

daniel-shimon:
I mean testing that XONSHRC_DIR is used at all.
You can set it to some a tmp directory to make sure it is considered (like you are doing with the --rc flag)

chronitis:
That is already done. The body of the test sets up a `XONSHRC_DIR` and adds some files, and the tests where it should be loaded check for `D0` and `D1` printed by those files.

daniel-shimon:
I see! Great!

bestlem:
> This is a followup to @anki-code in #4256 and [#4096 (comment)](https://github.com/xonsh/xonsh/issues/4096#issuecomment-791991841)
> 
> This adds a parameterised test with all the combinations of startup flags from #4096, and a few more, and checks which RC files are loaded. It should enforce the current behaviour, although having expanded many cases, I think some are probably arguable (eg, an explicit RC file `xonsh --rc foo.xsh bar.xsh` is not loaded unless you also specify interactive mode, which is probably not obvious).

Not only not obvious but it is wrong. How in the script do you test so that all the code that is a if statement that checks for interactive is NOT run. ie I don't need any form of completion or access to ptk, man etc

PR #4283, opened 2021-05-15
procs: Add $XONSH_CAPTURE_ALWAYS variable for opt-in interactive capt…

### LINKED ISSUE #3311: Cannot use emacsclient -t as $EDITOR
I try using `emacsclient -t` as an editor to be called from e.g. git. But this does not work. The screen shows the emacs frame, but key presses just get printed to the screen. Nothing seems to reach the emacsclient. Escapes are also just printed. ^C does exit it, but the shell often becomes unusable afterwards.

## xonfig
```
$ xonfig
| xonsh            | 0.9.10          |
| Python           | 3.7.4           |
| prompt toolkit   | 2.0.9           |
| pygments         | 2.4.2           |
| distro           | manjaro         |
```

## Expected Behavior

emacsclient should receive keystrokes.

## Current Behavior

emacsclient does not receive any keystrokes. Input gets printed directly onto the screen.

I tried mlterm and kitty, both have the issue. I tried it in tmux and without, does not matter. Happens also with readline shell. It does not happen in bash. It is also fine if I call the emacsclient myself, not indirectly via some tool. It used to work with git at least, not with others like yaourt. After the latest system update, git also does not work. I am not sure unfortunately if git and/or xonsh were part of that upgrade.

### Traceback (if applicable)

<details>

```
traceback
```

</details>

## Steps to Reproduce

At least on my machine, as far as I can tell, this is enough:
```
$ $EDITOR = 'emacsclient -t -a=""'
$ git commit
```

### LINKED ISSUE #3584: job control: ctrl-z in foreground process does not return to shell if xonsh is not direct parent
## xonfig

<details>

```
$ xonfig
| xonsh            | 0.9.14          |
| Python           | 3.8.2           |
| prompt toolkit   | 3.0.5           |
| pygments         | 2.6.1           |

```

</details>

## Expected Behavior

On `ctrl-z`, control should return to shell.

## Current Behavior

When ctrl-z is pressed in a process that's not a direct child of xons, control does not return to shell. Only way to escape from here is `pkill -CONT vim` from another shell.

## Steps to Reproduce

Can be reproduced with any curses like program if ran from a child of xonsh. Examples:

- `bash -c "vim"`
- `bash -c "mc"`
- `sops ./my-test-file` (https://github.com/mozilla/sops)

Any of the above, then `Ctrl-Z` hangs the shell

### LINKED ISSUE #3672: Neovim when called through $EDITOR is broken
When opening `neovim` through a program that subprocess `$EDITOR` the editor is unusable in xonsh; e.g. `git commit`. 
Here's a screenshot:
image
In the example above I called command `kubectl edit` which opens `$EDITOR` with temporary file. This works absolutely fine with other editors such as `vim` or calling `nvim` in `bash` but something between `xonsh` and `neovim` is not working.

## xonfig

<details>

```
| xonsh            | 0.9.19.dev8          |
| Git SHA          | 0fe76b53             |
| Commit Date      | Aug 10 06:28:52 2020 |
```

</details>

## Expected Behavior

Should open `neovim` and edit file the same way opening `nvim` process would

## Steps to Reproduce

1. Install `neovim` (tested with v0.4.4)  
2. set is as EDITOR: `$EDITOR='nvim'`  
3. call edit command like `git commit`

### LINKED ISSUE #4024: Ctrl+z seems not working
I tried to suspend the `gitk` command using the shortcut `Ctrl+z` and it is not working as shown in the screenshot.

## xonfig

<details>

```
| xonsh            | 0.9.24               |
| Git SHA          | 0f6e1812             |
| Commit Date      | Dec 13 19:19:32 2020 |
| Python           | 3.8.6                |
| prompt toolkit   | 3.0.8                |
| pygments         | 2.7.3                |
| xontrib 1        | abbrevs              |
| xontrib 2        | argcomplete          |
| xontrib 3        | avox_poetry          |
| xontrib 4        | back2dir             |
| xontrib 5        | broot                |
| xontrib 6        | cmd_done             |
| xontrib 7        | commands             |
| xontrib 8        | fzf-widgets          |
| xontrib 9        | hist_navigator       |
| xontrib 10       | jedi                 |
| xontrib 11       | kitty                |
| xontrib 12       | notifypy             |
| xontrib 13       | powerline3           |
| xontrib 14       | prompt_ret_code      |
| xontrib 15       | vox                  |
| xontrib 16       | voxapi               |

```

</details>

## Expected Behavior
it should be suspended to background

## Current Behavior

image

the control wasn't returned to shell.

## For community

### LINKED ISSUE #4249: Why xonsh need commands_cache.threadable_predictors in case when we want to just run the command?
_Originally posted by @Techcable in https://github.com/xonsh/xonsh/issues/4243#issuecomment-830621609_

> Is there a more general solution to the issues with "thread able" subprocess.
> I don't really get why xonsh has these "thread able" problems but other shells don't.

I'm also wondering why?

When we want just run the command without any threading/capturing why xonsh does not just run the command?

## For community

### LINKED ISSUE #4304: Problem with interfering with OSC sequences
See: https://github.com/muesli/duf/issues/129, you can reproduce using `duf`.

### LINKED ISSUE #4318: ^Z is ignored on linux
## xonfig

xonsh 0.9.27
Fedora Linux version 33
xonsh installed via pip3

I have removed all entries from .xonshrc except setting the pronpt
$PROMPT = '{GREEN}⇨ {RESET}'

<details>
When running an application in the foreground ^Z is ignored.  I.e., the running application is NOT suspended.  <br>The only command xonsh accepts is ^C which kills the running application.
<br><br>
Example of running a pdf viewer evince.  I twant to move the application to the background so I may enter additional commands:
<br><br>
Enter the command:<br>

```
⇨ evince file.pdf
^Z
```
^Z echos as shown above but nothing else happens.<br><br>
Try ^C to kill the application
```
⇨ evince file.pdf
^C
⇨
```
^C worked<br><br>

There appears no way to suspend a running application

</details>

## Expected Behavior

<details>
^Z should suspend the running application
</details>

## Current Behavior

<details>
Nothing happens
</details>

### Traceback (if applicable)

<details>

</details>

## Steps to Reproduce

1.  Enter any long running CLI command
2. type ^Z
3. Note that the ^Z is ignored - nothing happens

## For community

daniel-shimon:
…uring

Closes #4249
Closes #4304
Closes #4024
Closes #3672
Closes #3311
Closes #4318
Closes #3584
Closes #3672

Explanation from one of the issues:

> All these issues are indeed because xonsh tries to capture the output of interactive processes. 
> Turning `THREAD_SUBPROCS` off kinda fixes it but it isn't good either, since it means `$(...)` doesn't always work.
> We've decided to add a new `XONSH_CAPTURE_INTERACTIVE` variable that will give us better control and will be False by default

## For community

### DIFF
--- a/xonsh/procs/specs.py
+++ b/xonsh/procs/specs.py
@@ -734,6 +734,7 @@ def _update_last_spec(last):
         cmds_cache = XSH.commands_cache
         thable = (
             env.get("THREAD_SUBPROCS")
+            and (captured != "hiddenobject" or env.get("XONSH_CAPTURE_ALWAYS"))
             and cmds_cache.predict_threadable(last.args)
             and cmds_cache.predict_threadable(last.cmd)
         )
--- a/news/optional-interractive-capture.rst
+++ b/news/optional-interractive-capture.rst
@@ -1,0 +1,26 @@
+**Added:**
+
+* New ``$XONSH_CAPTURE_ALWAYS`` variable for opt-in interactive capturing.
+  Since this capturing breaks background jobs and some interactive programs (like ``git`` invoking an editor),
+  This behavior is now opt-in using this variable.
+  See https://github.com/xonsh/xonsh/pull/4283 and linked issues.
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/tests/procs/test_specs.py
+++ b/tests/procs/test_specs.py
@@ -16,6 +16,16 @@ from .tools import skip_if_on_windows
 def test_cmds_to_specs_thread_subproc(xession):
     env = xession.env
     cmds = [["pwd"]]
+
+    # XONSH_CAPTURE_ALWAYS=False should disable interactive threaded subprocs
+    env["XONSH_CAPTURE_ALWAYS"] = False
+    env["THREAD_SUBPROCS"] = True
+    specs = cmds_to_specs(cmds, captured="hiddenobject")
+    assert specs[0].cls is Popen
+
+    # Now for the other situations
+    env["XONSH_CAPTURE_ALWAYS"] = True
+
     # First check that threadable subprocs become threadable
     env["THREAD_SUBPROCS"] = True
     specs = cmds_to_specs(cmds, captured="hiddenobject")
--- a/tests/procs/test_specs.py
+++ b/tests/procs/test_specs.py
@@ -1,11 +1,11 @@
 """Tests the xonsh.procs.specs"""
+import itertools
 from subprocess import Popen
 
 import pytest
 
-from xonsh.procs.specs import cmds_to_specs
-
-from xonsh.environ import Env
+from xonsh.procs.specs import cmds_to_specs, run_subproc
+from xonsh.built_ins import XSH
 from xonsh.procs.posix import PopenThread
 from xonsh.procs.proxies import ProcProxy, ProcProxyThread
 
@@ -35,3 +35,46 @@ def test_cmds_to_specs_thread_subproc(xession):
     env["THREAD_SUBPROCS"] = False
     specs = cmds_to_specs(cmds, captured="hiddenobject")
     assert specs[0].cls is ProcProxy
+
+
+@skip_if_on_windows
+@pytest.mark.parametrize(
+    "thread_subprocs, capture_always", list(itertools.product((True, False), repeat=2))
+)
+def test_capture_always(capfd, thread_subprocs, capture_always):
+    env = XSH.env
+    exp = "HELLO"
+    cmds = [["echo", "-n", exp]]
+
+    env["THREAD_SUBPROCS"] = thread_subprocs
+    env["XONSH_CAPTURE_ALWAYS"] = capture_always
+
+    hidden = run_subproc(cmds, "hiddenobject")  # ![]
+    # Check that interactive subprocs are always printed
+    assert exp in capfd.readouterr().out
+
+    if capture_always and thread_subprocs:
+        # Check that the interactive output was captured
+        assert hidden.out == exp
+    else:
+        # without THREAD_SUBPROCS capturing in ![] isn't possible
+        assert not hidden.out
+
+    # Explicitly captured commands are always captured
+    hidden = run_subproc(cmds, "object")  # !()
+    hidden.end()
+    if thread_subprocs:
+        assert not capfd.readouterr().out
+        assert hidden.out == exp
+    else:
+        # for some reason THREAD_SUBPROCS=False fails to capture in `!()` but still succeeds in `$()`
+        assert exp in capfd.readouterr().out
+        assert not hidden.out
+
+    output = run_subproc(cmds, "stdout")  # $()
+    assert not capfd.readouterr().out
+    assert output == exp
+
+    # Explicitly non-captured commands are never captured (/always printed)
+    run_subproc(cmds, captured=False)  # $[]
+    assert exp in capfd.readouterr().out
--- a/tests/test_lib/test_os.xsh
+++ b/tests/test_lib/test_os.xsh
@@ -12,15 +12,15 @@ def test_indir():
     if ON_WINDOWS:
         pytest.skip("On Windows")
     with tempfile.TemporaryDirectory() as tmpdir:
-        assert ![pwd].output.strip() != tmpdir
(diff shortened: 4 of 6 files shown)


anki-code:
Nice start. We have this on the current master:
```python
$XONSH_TRACE_SUBPROC=True

echo 1
# TRACE SUBPROC: (['echo', '1'],), captured=hiddenobject

_ = $(echo 1)
# TRACE SUBPROC: (['echo', '1'],), captured=stdout

_ = !(echo 1)
# TRACE SUBPROC: (['echo', '1'],), captured=object

![echo 1]
# TRACE SUBPROC: (['echo', '1'],), captured=hiddenobject

$[echo 1]
# TRACE SUBPROC: (['echo', '1'],), captured=False
```

And we want to switch first case to `captured=False` without any effect on other cases.

By the proposed changes:

> if captured == "hiddenobject" and not builtins.__xonsh__.env.get("XONSH_CAPTURE_ALWAYS"):

we will change also `![]` behavior that we should not change.

anki-code:
Also I think this setting should be in `xonfig` to analysis the future issues around freezing the terminal in the captured mode.

daniel-shimon:
When running commands in xonsh they are actually transformed into `![]`. It's the same thing

anki-code:
Yep, and I think we should have the way to switch this transformation to `$[]`.

daniel-shimon:
Why? That way you can't know the return value and other info.
As I see it, since `![]` denotes an "uncaptured subprocesses", xonsh shouldn't capture its stdout unless explicitly told to via ``XONSH_CAPTURE_ALWAYS``.
`![]` and `$[]` should behave identically (except their return values), a la the docs:
```
Uncaptured subprocesses are denoted with the $[] and ![] operators.
They are the same as $() captured subprocesses in almost every way.
The only difference is that the subprocess’s stdout passes directly through xonsh and to the screen.
The return value of $[] is always None.
```

daniel-shimon:
Another current issue is that when `THREAD_SUBPROCS=False`, `!()` fails to capture stdout while `$()` succeeds (without running the process in another thread).
As I see it, this doesn't make any sense and they should behave identically as well.

anki-code:
> Why? That way you can't know the return value and other info.

Yep, maybe it should (have to) be `![]`. I just have no success with this operator:
```python
cd /tmp
wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
chmod +x Miniconda3-latest-Linux-x86_64.sh
![./Miniconda3-latest-Linux-x86_64.sh ]
# press Enter and got freezed shell
```
It looks like we should have another operator that can run uncaptured command but can catch return code etc.

anki-code:
Subproc has stdout/stderr and return code. Sometimes we want to catch both. Sometimes only return code. Sometimes nothing. Is the current operators/run_subproc functionality enough?

daniel-shimon:
> It looks like we should have another operator that can run uncaptured command but can catch return code etc.

That's  exactly what XONSH_CAPTURE_ALWAYS is for. No need for a different operator.
This whole 'printing and also capturing at the same time' mechanism is unexpected and causes most of these problems.

anki-code:
I've tested the branch and it works for me!

xonsh/procs/specs.py
```diff
@@ -857,8 +857,15 @@ def run_subproc(cmds, captured=False, envs=None):
     if builtins.__xonsh__.env.get("XONSH_TRACE_SUBPROC"):
         print(f"TRACE SUBPROC: {cmds}, captured={captured}", file=sys.stderr)
```
anki-code, line 858: Please put the actual `captured` value here.

anki-code:
> No need for a different operator.

We lose the use case like:
```python
r = ![ls]  # the tool that produce some output
if 'file' in r.out:  # searching in the output
   print('found file')
```
In this case `![ls]` returns colored output but `r.out` has the output without colors. The case looks useful.

daniel-shimon:
> r = ![ls]  # the tool that produce some output

So this is the case to turn on XONSH_CAPTURE_OUTPUT
But in reality if you want the output of a command you need to run `$()` or `!()`

> In this case ![ls] returns colored output but r.out has the output without colors. The case looks useful.

I wonder why that is.. this is really strange! (just to be clear it also happens on main)

anki-code (approved):

daniel-shimon:
Thanks @anki-code !
@gforsyth I would also appreciate your review since this changes default behavior

gforsyth (approved):
Nice!  I think this is a good step forward and all the functionality is still available but with a slightly friendlier default.

Thanks @daniel-shimon for this!  And thanks @anki-code for your input!

PR #4286, opened 2021-05-19
Handle cancelled futures in AsyncPrompt

cryzed:
Prevent cancelled future errors for async prompt fields from bubbling up (and destroying the prompt's formatting). This happens for example when I keep holding down my enter key: after a while I will get tracebacks about uncaught `concurrent.futures.CancelledError`. This fixes this issue and prevents them from bubbling up.

### DIFF
--- a/news/asyncprompt-handle-cancelled-future.rst
+++ b/news/asyncprompt-handle-cancelled-future.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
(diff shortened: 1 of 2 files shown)


anki-code:
Looks good!

daniel-shimon:
Shouldn't we report this somehow?
Can you elaborate on when this happens?

cryzed:
@daniel-shimon I didn't report it, because I already fixed it, I also added a matching entry to the changelog.

It happens when `$ENABLE_ASYNC_PROMPT = True`, and you actually customize your `$PROMPT` to evaluate long-running functions. Once you have that set up, just keep holding your Enter key -- the default key repeat interval of your operating system should be quick enough to run into this issue; if it isn't, choose longer-running functions as part of your `$PROMPT`.

daniel-shimon (approved):
Great! Thanks!

Issue #4289, opened 2021-05-20
XSH singleton doesn't work with amalgamated (==production) xonsh


daniel-shimon:
@jnoortheen 

```sh
# Ensure amalgamate.py is installed
which amalgamate.py
# Install xonsh in production mode (without '-e')
pip install '.[full]'
# Try running xonsh
xonsh
```

I'm getting:
```
AttributeError: 'XonshSession' object has no attribute 'completers'
Failed to load xontrib jedi.
Traceback (most recent call last):
  File "/home/daniel/testxonsh/venv/lib/python3.8/site-packages/xonsh/__amalgam__.py", line 22082, in main
    args = premain(argv)
  File "/home/daniel/testxonsh/venv/lib/python3.8/site-packages/xonsh/__amalgam__.py", line 22022, in premain
    env = start_services(shell_kwargs, args, pre_env=pre_env)
  File "/home/daniel/testxonsh/venv/lib/python3.8/site-packages/xonsh/__amalgam__.py", line 21970, in start_services
    XSH.shell = Shell(execer=execer, **shell_kwargs)
  File "/home/daniel/testxonsh/venv/lib/python3.8/site-packages/xonsh/__amalgam__.py", line 16365, in __init__
    XSH.history = hist = xhm.construct_history(
  File "/home/daniel/testxonsh/venv/lib/python3.8/site-packages/xonsh/history/__amalgam__.py", line 1241, in construct_history
    backend = env.get("XONSH_HISTORY_BACKEND")
AttributeError: 'NoneType' object has no attribute 'get'
Xonsh encountered an issue during launch
```

jnoortheen:
`XSH.load()` is not called. I am not sure why this is the case. Does the same run after removing all generated __amalgam__.py modules?

daniel-shimon:
> `XSH.load()` is not called. I am not sure why this is the case. Does the same run after removing all generated **amalgam**.py modules?

No. Installing without amalgamation (`pip install -e`) works fine.
The issue seems to be that code from submodules imports and uses the non-amalgamated XSH, while amalgamated code uses the amalgamated XSH (I printed in a few places and indeed XSH in `xonsh/history/__amalgam__.py` and in `xonsh/__amalgam__.py` are two different objects).
Seems to me we need to disable amalgamating `built_ins.py` altogether

PR #4290, opened 2021-05-20
tests: Run tests with the amalgamated xonsh

daniel-shimon:
Our tests currently don't run with the amalgamated version of xonsh.
This creates bugs we can miss that are created by the amalgamation (see #4289).

Since the production version of xonsh is amalgamated, ensure the tests run with it amalgamated as well.

## For community


daniel-shimon:
This should currently fail until #4289 is fixed.
No news needed

daniel-shimon:
I think amalgamation is important (I do see a big improvement to load time after amalgamating), but I have some issues with it:
1. First of all I think currently `pip install xonsh` doesn't actually amalgamate for some reason - @gforsyth do you know whats up with this?
2. Also I see that some modules are imported twice (like what happened with `built_ins.py`) - we can probably improve this somehow (I was thinking about ensuring all imports are done correctly in the CI)

gforsyth:
>     1. First of all I think currently `pip install xonsh` doesn't actually amalgamate for some reason - @gforsyth do you know whats up with this?

I don't.  Is this happening only when pulling down from pypi or also when installing with `pip` locally?



>     2\. Also I see that some modules are imported twice (like what happened with `built_ins.py`) - we can probably improve this somehow (I was thinking about ensuring all imports are done correctly in the CI)

Yeah, this is something that's historically been done by eye and it's obviously prone to errors

gforsyth:
> In my machine, I don't see any speedup from amalgamating. What are your thoughts ?

I'd love to remove it, but it consistently improves startup times across several machines I've tested it on by ~100ms

jnoortheen:
>  it consistently improves startup times across several machines I've tested it on by ~100ms

are they running in HDD ?

gforsyth:
> are they running in HDD

do you mean on spinning disks vs. solid state?  All solid state

PR #4299, opened 2021-05-29
Update completer cli

daniel-shimon:
1. Print which completers are non exclusive in `completer list`:
image

2. When adding a new completer via `completer add`, add it before the **exclusive** completers.
This means it won't interfere with the non-exclusive ones. (see more detailed comment in code)

## For community


jnoortheen (approved):

daniel-shimon:
@jnoortheen thanks for the review!

PR #4267, opened 2021-05-08
Argparser auto-completion and colored help

jnoortheen:
## TODO:
- [x] update completer to handle options and nargs
- [x] add coloring 

## sub PRs:

I think it is better to split this PR
- [x] support callable objects - https://github.com/xonsh/xonsh/pull/4379
- [x] argparser colored help and auto-completions - https://github.com/xonsh/xonsh/pull/4391
- [x] dirstack - https://github.com/xonsh/xonsh/pull/4446
- [x] xonfig - https://github.com/xonsh/xonsh/pull/4436/
- [x] xontrib - https://github.com/xonsh/xonsh/pull/4441
- [ ] vox - https://github.com/xonsh/xonsh/pull/4437
- [x] run-tests.xsh - https://github.com/xonsh/xonsh/pull/4434

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #4300, opened 2021-05-31
feat: add new environment variable to customize tracing

jnoortheen:
## For community

### DIFF
--- a/news/sub-proc-tracer.rst
+++ b/news/sub-proc-tracer.rst
@@ -1,7 +1,7 @@
 **Added:**
 
-* new environment variable ``$XONSH_SUBPROC_TRACER``
-  to provide custom function to format ``$XONSH_TRACE_SUBPROC`` output
+* new environment variable ``$XONSH_TRACE_SUBPROC_FUNC``
+  to handle ``$XONSH_TRACE_SUBPROC`` output
 
 **Changed:**
 
--- a/news/sub-proc-tracer.rst
+++ b/news/sub-proc-tracer.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* new environment variable ``$XONSH_SUBPROC_TRACER``
+  to provide custom function to format ``$XONSH_TRACE_SUBPROC`` output
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/run-tests.xsh
+++ b/run-tests.xsh
@@ -19,8 +19,6 @@ def colored_tracer(cmds, **_):
     cmd = " ".join(itertools.chain.from_iterable(cmds))
     print_color(f"{{GREEN}}$ {{BLUE}}{cmd}{{RESET}}", file=sys.stderr)
 
-$XONSH_SUBPROC_TRACER = colored_tracer
-
 
 def _replace_args(args: List[str], num: int) -> List[str]:
     return [
@@ -66,6 +64,7 @@ def test(
 def qa():
     """QA checks"""
     $XONSH_NO_AMALGAMATE = True
+    $XONSH_SUBPROC_TRACER = colored_tracer
 
     black --check xonsh xontrib tests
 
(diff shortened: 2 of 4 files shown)


daniel-shimon:
Hey, why do this? It just creates unnecessary outputs for tests that run subprocs without any benefits

jnoortheen:
> Hey, why do this? It just creates unnecessary outputs for tests that run subprocs without any benefits

It is to use xonsh as a build/release tool where one would want to see what command is running. Like fabric/invoke. If you see the output of qa , you can see how it can be useful .

jnoortheen:
Running pytest in subprocess is needed since Xonsh variables would get passed to integration tests. In my environment, i had failing tests because xonshrc will get passed and integration tests would fail

run-tests.xsh
```diff
@@ -26,26 +37,23 @@ def test(ns: argparse.Namespace):
...
+    #     echo "Tests need to run from the amalgamated xonsh! install with `pip install .` (without `-e`)"
+    #     exit(1)
 
+    cmd = ["pytest"] + _replace_args(args, 0)
     if ns.report_coverage:
-        ![pytest @(_replace_args(args, 0)) --cov --cov-report=xml --cov-report=term]
-    else:
-        ![pytest @(_replace_args(args, 0))]
+        cmd += "--cov --cov-report=xml --cov-report=term".split()
+    subprocess.check_call(cmd, stderr=sys.stderr, stdout=sys.stdout)
```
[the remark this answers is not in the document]
jnoortheen: Using xonsh here would cause this environment variables to get passed onto integration tests and they will fail. 
One other thing, it uses pytest-timeout on windows which causes slowdown on windows. We can use the subprocess timeout here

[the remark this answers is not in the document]
jnoortheen: seems like removing pytest-timeout doesn't reduce the total time on windows. reverted the change

jnoortheen:
we get this beatified output 
image

daniel-shimon (commented):
Is this still a draft?

jnoortheen:
> Is this still a draft?

it will conflict with https://github.com/xonsh/xonsh/pull/4267 . So just putting this to draft until it is cleared.

daniel-shimon (dismissed):
Awesome!

news/sub-proc-tracer.rst
```diff
@@ -0,0 +1,24 @@
+**Added:**
+
+* new environment variable ``$XONSH_SUBPROC_TRACER``
```
gforsyth, line 3: I'd suggest that the variable name be something like `XONSH_TRACE_SUBPROC_RUNNER` -- it's a bit awkward and I'm open to suggestions, but since it modifies / extends the behavior of a different env-var, it's better if they're clustered alphabetically

jnoortheen: I have used `XONSH_TRACE_SUBPROC_FUNC` since `RUNNER` sounds like handling the subprocess call.

xonsh/environ.py
```diff
@@ -1076,6 +1076,19 @@ class GeneralSetting(Xettings):
...
+            "A callback function used to format the trace output shown when $XONSH_TRACE_SUBPROC=True."
+        ),
+        doc_default="""\
+By default it just prints ``cmds`` like below.
+
+.. code-block:: python
+def tracer(cmds: list, captured: Union[bool, str]):
+    print(f"TRACE SUBPROC: {cmds}, captured={captured}", file=sys.stderr)
+""",
+    )
```
gforsyth, line 1091: Style nit -- since all the other `doc_defaults` here use implicit string concatenation, can we do the same here?

gforsyth (changes requested):
Thanks for putting this in @jnoortheen !  Two small nits and then this is good to go

xonsh/environ.py
jnoortheen: I want to keep this way. It is cleaner and easier to edit. I remember breaking some long lines in this file that resulted in the  implicit string concats here, I didn't wanted to mess up the whitespaces.

gforsyth: Ok, that's fair.  Let's try, as changes are made to the descriptions (if they are) to push folks to also convert to the triple-quoted blocks.

PR #4316, opened 2021-06-11
completers: bash: Fix completing an empty prefix

daniel-shimon:
No need for news since this bug was introduced after the last version

## For community

### DIFF
--- a/tests/completers/test_bash_completer.py
+++ b/tests/completers/test_bash_completer.py
@@ -164,3 +164,13 @@ def test_quote_handling(command_context, completions, lprefix):
         isinstance(comp, RichCompletion) and not comp.append_closing_quote
         for comp in bash_completions
     )  # make sure the completer handles the closing quote by itself
+
+
+@skip_if_on_darwin
+@skip_if_on_windows
+def test_bash_completer_empty_prefix():
+    context = CompletionContext(
+        CommandContext(args=(CommandArg("git"),), arg_index=1, prefix="")
+    )
+    bash_completions, bash_lprefix = complete_from_bash(context)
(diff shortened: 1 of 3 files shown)


tests/completers/test_bash_completer.py
```diff
@@ -164,3 +164,13 @@ def test_quote_handling(command_context, completions, lprefix):
...
+
+
+@skip_if_on_darwin
+@skip_if_on_windows
+def test_bash_completer_empty_prefix():
+    context = CompletionContext(
+        CommandContext(args=(CommandArg("apt"),), arg_index=1, prefix="")
+    )
+    bash_completions, bash_lprefix = complete_from_bash(context)
+    assert {"install", "show"}.issubset(bash_completions)
```
jnoortheen, line 176: can't the subprocess call be mocked? without apt (like my system, it will fail). Ultimately the bash-completion invokes subprocess right. You can refer test_gitstatus.py.

daniel-shimon: It won't really help since we're testing the integration between xonsh and bash.. If we mock bash we won't really be testing anything.
Maybe can you find some other program that exists broadly instead of apt?

jnoortheen: Yes that will do. one of coreutils.

daniel-shimon: Changed to git

daniel-shimon: Changed to git

jnoortheen (approved):

jnoortheen:
I think the mac test is a flaky one.

daniel-shimon:
@jnoortheen thanks for the review!

PR #4308, opened 2021-06-07
Fix empty command error message

yaxollum:
Fixes #2329
Fixes #4287 

Prints a helpful error message (`xonsh: subprocess mode: command is empty`) when an empty command is run (for example using `@$(echo)` or `$['']`).

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #4319, opened 2021-06-12
feat: make completer accept generators

### LINKED ISSUE #3968: autocompletions support for generators
[PromptToolkit](https://python-prompt-toolkit.readthedocs.io/en/master/pages/asking_for_input.html#a-custom-completer) lets you return generators from autocompleters. But xonsh requires one to return `set`. It beats the purpose of having the env variable `COMPLETION_QUERY_LIMIT` . All the functions will be running and that is not efficient. 

## xonfig

<details>

```
| xonsh            | 0.9.24               |
| Git SHA          | 74543ea9             |
| Commit Date      | Oct 10 15:12:47 2020 |
| Python           | 3.8.6                |
| pygments         | 2.7.2                |
| xontrib 1        | abbrevs              |
| xontrib 2        | argcomplete          |
| xontrib 3        | autovox              |
| xontrib 4        | back2dir             |
| xontrib 5        | cmd_done             |
| xontrib 6        | hist_navigator       |
| xontrib 7        | jedi                 |
| xontrib 8        | kitty                |
| xontrib 9        | notifypy             |
| xontrib 10       | pdb                  |
| xontrib 11       | prompt_ret_code      |
| xontrib 12       | vox                  |
| xontrib 13       | voxapi               |

```

</details>

## Expected Behavior

xonsh should enable returning autocompleters with yield support as well. 

## Current Behavior

1. not able to use generators as completers
2. not able to style the autocompletion menu. eventhough the shell supports it.

I think it is this way to support the compatibility with readline. At this point, does xonsh need to support readline? as it now ships with `prompt_toolkit`.

## For community

jnoortheen:
fixes #3968

## For community

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -78,13 +78,55 @@ class Completer(object):
             (prefix, line, begidx, endidx, ctx),
         )
 
-    def complete_from_context(self, completion_context, old_completer_args=None):
-        trace = XSH.env.get("XONSH_TRACE_COMPLETIONS")
-        if trace:
-            print("\nTRACE COMPLETIONS: Getting completions with context:")
-            sys.displayhook(completion_context)
-        lprefix = 0
-        completions = set()
+    @staticmethod
+    def _format_completion(
+        completion,
+        completion_context,
+        completing_contextual_command: bool,
+        lprefix: int,
+        custom_lprefix: bool,
+    ) -> tp.Tuple[Completion, int]:
+        if (
+            completing_contextual_command
+            and completion_context.command.is_after_closing_quote
+        ):
+            """
+            The cursor is appending to a closed string literal, i.e. cursor at the end of ``ls "/usr/"``.
+            1. The closing quote will be appended to all completions.
+                I.e the completion ``/usr/bin`` will turn into ``/usr/bin"``
+                To prevent this behavior, a completer can return a ``RichCompletion`` with ``append_closing_quote=False``.
+            2. If not specified, lprefix will cover the closing prefix.
+                I.e for ``ls "/usr/"``, the default lprefix will be 6 to include the closing quote.
+                To prevent this behavior, a completer can return a different lprefix or specify it inside ``RichCompletion``.
+            """
+            closing_quote = completion_context.command.closing_quote
+            if not custom_lprefix:
+                lprefix += len(closing_quote)
+            if closing_quote:
+                if isinstance(completion, RichCompletion):
+                    if completion.append_closing_quote:
+                        completion = completion.replace(
+                            value=completion.value + closing_quote
+                        )
+                else:
+                    completion = completion + closing_quote
+
+        completion = list(apply_lprefix([completion], lprefix))[0]
+
+        if (
+            isinstance(completion, RichCompletion)
+            and completion.append_space
+            and not completion.value.endswith(" ")
+        ):
+            # append spaces AFTER appending closing quote
+            completion = completion.replace(value=completion.value + " ")
+
+        return completion, lprefix
+
+    @staticmethod
+    def generate_completions(
+        completion_context, old_completer_args, trace: bool
+    ) -> tp.Iterator[tp.Tuple[Completion, int]]:
         for name, func in XSH.completers.items():
             try:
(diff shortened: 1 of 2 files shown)


jnoortheen:
seems like the test from #4308 causes failures on main branch.

https://github.com/xonsh/xonsh/runs/2810366168?check_suite_focus=true

daniel-shimon:
Yeah.. Can you fix it in a separate pr so we can merge it immediately?

jnoortheen:
> Yeah.. Can you fix it in a separate pr so we can merge it immediately?

created a separate one. can you check?

xonsh/completer.py
```diff
@@ -124,60 +166,52 @@ def complete_from_context(self, completion_context, old_completer_args=None):
...
-                            )
-                        return completion
-                    return completion + closing_quote
-
-                res = map(append_closing_quote, res)
-
-            completions.update(apply_lprefix(res, lprefix))
+                sys.displayhook(items)
+
+            if not items:  # empty completion
```
daniel-shimon, line 190: This check needs to be before tracing so that we won't trace an empty result

xonsh/completer.py
```diff
@@ -124,60 +166,52 @@ def complete_from_context(self, completion_context, old_completer_args=None):
...
+
+        for comp in self.generate_completions(
+            completion_context,
+            old_completer_args,
+            trace,
+        ):
+            completion, lprefix = comp
+            completions.add(completion)
+            if len(completions) >= query_limit:
+                break
```
daniel-shimon, line 214: Add a `TRACE COMPLETIONS: Stopped after $COMPLETION_QUERY_LIMIT completions` message if `trace`

xonsh/completer.py
```diff
@@ -124,60 +166,52 @@ def complete_from_context(self, completion_context, old_completer_args=None):
...
+        query_limit = XSH.env.get("COMPLETION_QUERY_LIMIT", 100)
+
+        for comp in self.generate_completions(
+            completion_context,
+            old_completer_args,
+            trace,
+        ):
+            completion, lprefix = comp
+            completions.add(completion)
+            if len(completions) >= query_limit:
```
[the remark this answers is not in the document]
jnoortheen: 100 comes from the default value of the environment.

xonsh/completer.py
jnoortheen: But sometimes one might want to see that their functions returned empty results right?

xonsh/completer.py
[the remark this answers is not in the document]
daniel-shimon: In that case you don't need to write the default value here

xonsh/completer.py
daniel-shimon: Not really. You can see which completers are registered with `completer list`
The tracing helps you to understand what really happened in runtime so these prints are just getting in the way imo

xonsh/completer.py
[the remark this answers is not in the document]
jnoortheen: During tests it may not have the variable filled right. It needs some numerical value to be safe.

xonsh/completer.py
jnoortheen: Having it registered and the function returning empty results are different I think. Why do you wanr skip the comment on empty set?

xonsh/completer.py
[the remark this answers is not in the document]
daniel-shimon: So either mock the value while testing or enable the value to be None for limitless completions
Anyway I think it's good to enable limitless completions

xonsh/completer.py
daniel-shimon: Because all completers are run until a non exclusive one returns something.
So it's clear and known what completers will run, and there's no need to print their empty results
I think it just clutters the trace

xonsh/completer.py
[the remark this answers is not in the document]
jnoortheen: Ok that is viable.

jnoortheen:
@daniel-shimon done suggested changes.

daniel-shimon (approved):
Great! Thanks!

PR #4321, opened 2021-06-13
refactor: remove non-source files from ply vendored

jnoortheen:
Also updated the Makefile to add only source folder from repo.

## For community


daniel-shimon:
I'm missing something, with the makefile after your changes won't the non-source files be inserted again?

jnoortheen:
> I'm missing something, with the makefile after your changes won't the non-source files be inserted again?

it copies only part from the `branch/ply`  directory into `xonsh/ply/ply`

jnoortheen:
as of now, using that command would copy from master branch of ply. But 4.0 versions don't have caching and will require some changes to xonsh parser. (like doing the caching ourselves)

gforsyth:
I'm not opposed to this cleanup, but it feels a little bit like a footgun for now since the latest version of `ply` isn't compatible with our current setup

jnoortheen:
These extra files are making source packages bigger. 

Regarding the newer versions, perhaps we can fork ply under xonsh? There won't be any features added to ply in future. To the fork, we can try to add back caching and some optimization if possible.

jnoortheen:
Also it would reduce the source package's size https://pypi.org/project/xonsh/#files.

gforsyth:
The repo is going to stay bloated unless we decide to do something rash and rewrite history, which I'm not going to condone.

I agree reducing the sdist size is a good thing -- I don't see a reason to update PLY at this point unless there's some heinous bug.
I don't think anyone is volunteering to rewrite the parser anytime soon.  So maybe just strip out the update command from the Makefile?

jnoortheen:
Do you want to revert the changes to the Makefile ?

gforsyth:
> Do you want to revert the changes to the Makefile ?

Just remove it entirely, I'd say (the `ply` directive), since if we upgrade now it will break

jnoortheen:
> Just remove it entirely, I'd say (the `ply` directive), since if we upgrade now it will break

Having that doesn't mean ply should be upgraded. Anyways, I will remove that command.

jnoortheen:
@gforsyth I have removed the command.

gforsyth:
Thanks @jnoortheen !

PR #4322, opened 2021-06-13
Fix atdollar_no_output test

yaxollum:
This test was failing on Windows because of an incorrect indentation. The `echo` command was not actually getting aliased to the `_echo` function.

## For community


daniel-shimon:
Thanks!

PR #4324, opened 2021-06-14
chore: update mypy to 0.902

jnoortheen:
not including news item as this is development only

## For community


daniel-shimon:
Umm.. What's types-ujson?

jnoortheen:
> Umm.. What's types-ujson?

in recent release they dropped packaging some default stubs. now we need to explicitly install them. ujson is used somewhere in the code

daniel-shimon:
Thanks!

PR #4328, opened 2021-06-16
Move coloring stat calls to another thread

### LINKED ISSUE #2578: xonsh becomes *painfully* slow in case of some network problems
I'm mounting samba shares in my laptop and if the server becomes unreachable and I try to unmount them from xonsh it's really confusing and hard because **while I'm typing** commands that contain the mount path xonsh stalls for half a minute or even more. e.g. If you type **umount -l** and then try to paste the misbehaving path (/mnt/server_share) xonsh becomes completely unresponsive (to the point of not even echoing characters).

If you need more information/debugging I'll be happy to help. 

Nick

yaxollum:
Fixes #2578 

Moves the `stat` calls required for coloring paths to another thread, and sets a timeout. This prevents xonsh from becoming unresponsive when coloring is enabled and the path to an unreachable network share is entered.

## For community

### DIFF
--- a/xonsh/pyghooks.py
+++ b/xonsh/pyghooks.py
@@ -9,6 +9,8 @@ from collections import ChainMap
 from collections.abc import MutableMapping
 from keyword import iskeyword
 import typing as tp
+import multiprocessing
+from multiprocessing.pool import ThreadPool
 
 from xonsh.built_ins import XSH
 from xonsh.lazyimps import os_listxattr
@@ -1628,9 +1630,20 @@ def subproc_arg_callback(_, match):
     yieldVal = Text
     try:
         path = os.path.expanduser(text)
-        path_stat = os.lstat(path)  # lstat() will raise FNF if not a real file
-        yieldVal, _ = color_file(path, path_stat)
+
+        color_file_with_lstat = lambda p: color_file(
+            p, os.lstat(p)
+        )  # lstat() will raise FNF if not a real file
+
+        pool = ThreadPool(processes=1)
+        color_async = pool.apply_async(color_file_with_lstat, (path,))
+
+        # the lstat call in the lambda and the stat call in color_file()
+        # may cause a timeout (issue #2578)
+        yieldVal, _ = color_async.get(timeout=0.5)
     except (OSError):
+        pass
+    except (multiprocessing.TimeoutError):
         pass
 
     yield (match.start(), yieldVal, text)
--- a/news/coloring-stat-calls.rst
+++ b/news/coloring-stat-calls.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed unresponsive prompt when coloring is enabled and the path to an unreachable network share is entered
+
+**Security:**
+
+* <news item>


xonsh/pyghooks.py
```diff
@@ -1628,10 +1630,21 @@ def subproc_arg_callback(_, match):
...
         path = os.path.expanduser(text)
-        path_stat = os.lstat(path)  # lstat() will raise FNF if not a real file
-        yieldVal, _ = color_file(path, path_stat)
+
+        color_file_with_lstat = lambda p: color_file(
+            p, os.lstat(p)
+        )  # lstat() will raise FNF if not a real file
+
+        pool = ThreadPool(processes=1)
+        color_async = pool.apply_async(color_file_with_lstat, (path,))
```
jnoortheen, line 1639: this will slowdown in other use-cases right?

yaxollum: @jnoortheen Yes it will.

I wrote a script to benchmark the effects of using `ThreadPool`: [benchmark_threading_lstat.py](https://gist.github.com/yaxollum/3edbbb3e41f70f7996bba800e019cd1c)

The script tests three different methods:
1. Creating a new `ThreadPool` every time the function is called (this is what my PR currently does)
2. Creating only a single `ThreadPool` and using it for every function call
3. Not using any threads at all

Here are the results which I got (time per `lstat` call):
```
With thread new pool:    0.6194 ms
With thread single pool: 0.0770 ms
Without thread:          0.0013 ms
```
As you can see, using a single `ThreadPool` improves the performance by around 8x compared to creating a new `ThreadPool` each time, but that is still around 60x slower than not using any threads.

I'm not sure whether this slowdown is acceptable since although there is a big performance difference ratio, the magnitudes are all under a millisecond.

yaxollum: Also, I just realized that using a single `ThreadPool` wouldn't work in the case of `lstat` calls. When an `lstat` call times out, the thread is still running (it doesn't get killed), which would keep the `ThreadPool` occupied. Subsequent calls to valid paths would be queued and also time out.

jnoortheen: you can hold reference to the previous future and cancel if it was pending after new request comes in.
Also making this optional will help users who work with slow fs.

yaxollum: > you can hold reference to the previous future and cancel if it was pending after new request comes in.

What if the previous future is already executing? As far as I know, [futures that have already started executing cannot be canceled](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.cancel):

_"Attempt to cancel the call. **If the call is currently being executed or finished running and cannot be cancelled** then the method will return False, otherwise the call will be cancelled and the method will return True."_

> Also making this optional will help users who work with slow fs.

Do you mean making an environment variable like `$THREAD_COLORING_STAT_CALLS`? Should it be enabled or disabled by default?

jnoortheen: > What if the previous future is already executing?

can use [killable threads](http://tomerfiliba.com/recipes/Thread2/). But I need opinion on this from others as well. We can also give a warning that stat calls are slow and can suggest to turn-off coloring.

> Should it be enabled or disabled by default?
I suggest to make the default debhaviour to not thread any call.

yaxollum: > can use killable threads.

Could another option be to just let the thread continue to run in the background (without killing it)?

jnoortheen: That will exhaust resources if stuck.

daniel-shimon:
Hi @yaxollum !
I think another approach can be to use a single global `multiprocessing.pool.Pool`, and when we get a timeout error we call `pool.terminate()` and replace the global pool with a new one.

This way normally there's almost no overhead, and only bad paths create new process overheads

jnoortheen:
Hi @yaxollum , are you still interested in fixing this? 

I think the following would be more robust. 
1. update $COLOR_FILE to accept three values 
   i. 0 -> disable coloring -- no stat or mount checks
   ii. 1 -> disable coloring for non-mount paths 
              (you can check if a path is mounted or not by using `os.path.ismount()`.  Also caching the result for paths names is needed to avoid repeated calls to the same base path) -- the default value
  iii. 2 -> enable coloring for both cases -- no check for ismount and always result in lstat call. The decision is left to user, we can show the warning in the environment variable documentation about enabling it.

No threads involved.

PR #4334, opened 2021-06-21
Fix output capturing for !() and @$() when $THREAD_SUBPROCS is set to False

### LINKED ISSUE #4056: @$(cmd) substitution doesn't work properly with $THREAD_SUBPROCS off, but de-sugaring it does
The [tutorial](https://xon.sh/tutorial.html#command-substitution-with) says that `@$(cmd)` is syntax sugar for `@([i.strip() for i in $(cmd).split()])`, but if `$THREAD_SUBPROCS` is set to `False`, then they behave differently, with only the latter having the correct behavior. `@$(cmd)` fails to capture the output of `cmd`, and `cmd`'s output is written directly to standard out instead of being passed as arguments, which since `$THREAD_SUBPROCS == False` is not particularly surprising, but de-sugaring the input to `@([i.strip() for i in $(cmd).split()])` correctly captures the output of the substituted command just fine, and correctly substitutes it in as arguments.

## xonfig

<details>

```
$ xonfig
| xonsh            | 0.9.24          |
| Python           | 3.9.1           |
| prompt toolkit   | 3.0.8           |
| pygments         | 2.7.3           |

```

</details>

## Expected Behavior

Where `args` is a simple program that prints the arguments as a list, for clarity's sake.
```python
# args.py
import sys
print(sys.argv)
```
Then:
```python
>>> args @$(echo 1 2 3)
['/home/qyriad/.local/bin/args', '1', '2', '3']
>>> args @([i.strip() for i in $(echo 1 2 3).split()])
['/home/qyriad/.local/bin/args', '1', '2', '3']
```

## Current Behavior

```python
>>> args @$(echo 1 2 3)
1 2 3
['/home/qyriad/.local/bin/args']
>>> args @([i.strip() for i in $(echo 1 2 3).split()])
['/home/qyriad/.local/bin/args', '1', '2', '3']
```

## Steps to Reproduce

An example without that `args` script:
```
$ xonsh --no-rc
>>> $THREAD_SUBPROCS = False
>>> $XONSH_SHOW_TRACEBACK = True
>>> aliases['ls'] = 'ls --color=auto'
>>> @([i.strip() for i in $(which ls).split()])
# -Snip directory listing-

>>> @$(which ls)
  File "/usr/lib/python3.9/site-packages/xonsh/__amalgam__.py", line 17275, in default
  File "/usr/lib/python3.9/site-packages/xonsh/__amalgam__.py", line 2377, in run_compiled_code
    func(code, glb, loc)
  File "/usr/lib/python3.9/site-packages/xonsh/__amalgam__.py", line 23342, in subproc_captured_hiddenobject
    return run_subproc(cmds, captured="hiddenobject", envs=envs)
  File "/usr/lib/python3.9/site-packages/xonsh/__amalgam__.py", line 23263, in run_subproc
    specs = cmds_to_specs(cmds, captured=captured, envs=envs)
  File "/usr/lib/python3.9/site-packages/xonsh/__amalgam__.py", line 23216, in cmds_to_specs
    spec = SubprocSpec.build(cmd, captured=captured, env=env)
  File "/usr/lib/python3.9/site-packages/xonsh/__amalgam__.py", line 22978, in build
    spec.resolve_alias()
  File "/usr/lib/python3.9/site-packages/xonsh/__amalgam__.py", line 23009, in resolve_alias
    cmd0 = self.cmd[0]
```

## For community

yaxollum:
Fixes #4056

!() and @$() both use `xonsh.procs.specs.run_subproc(cmds, captured="object", envs=envs)` to capture the command results. This PR allows them to work properly when $THREAD_SUBPROCS is set to False.

Currently, when $THREAD_SUBPROCS is set to False:
- `!(echo hi).out` is an empty string (it should be `'hi\n'`)
- Running the command `@$(echo ls)` results in the error message `xonsh: subprocess mode: command is empty`

## For community

### DIFF
--- a/news/fix-capture-object-nonthread.rst
+++ b/news/fix-capture-object-nonthread.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed output capturing for !() and @$() when $THREAD_SUBPROCS is set to False
+
+**Security:**
+
+* <news item>
--- a/xonsh/procs/pipelines.py
+++ b/xonsh/procs/pipelines.py
@@ -249,6 +249,10 @@ class CommandPipeline:
                 proc.wait()
                 self._endtime()
                 if self.captured == "object":
+                    if stdout:
+                        b = stdout.read()
+                        s = self._decode_uninew(b, universal_newlines=True)
+                        self.lines = s.splitlines(keepends=True)
                     self.end(tee_output=False)
                 elif self.captured == "hiddenobject" and stdout:
                     b = stdout.read()
--- a/xonsh/procs/specs.py
+++ b/xonsh/procs/specs.py
@@ -742,7 +742,7 @@ def _update_last_spec(last):
         elif not thable:
             # foreground processes should use Popen
             last.threadable = False
-            if captured == "object" or captured == "hiddenobject":
+            if captured == "hiddenobject":
                 # CommandPipeline objects should not pipe stdout, stderr
                 return
     # cannot used PTY pipes for aliases, for some dark reason,


yaxollum:
Hi, I'm going to close this PR since I've found that the changes which I've made to `!()` have some issues (for example, stderr is not captured). 

To fix `@$()` when `THREAD_SUBPROCS` is `False`, I think a better approach would be to use `captured="stdout"` instead of `captured="object"` for that operator. I will make another PR for that.

PR #4336, opened 2021-06-22
Stop $() from silently capturing stderr

### LINKED ISSUE #3786: Capturing subprocess stdout without stderr
Suppose I try to `ls` in a directory without proper permissions in Bash:

```bash
> result=$(ls)
ls: cannot open directory '.': Permission denied
```

The command fails, and stderr is printed directly to the console. However, in xonsh, the same code silently fails:

```
> result=$(ls)
> result
''
```

Is there a way to capture only stdout, so that errors show up immediately? I did see `!()`, but using that means we lose out on colored output in the error messages from a lot of programs.

## For community

yaxollum:
Fixes #3786 

Currently, doing something like:
```xonsh
result=$(ls /root)
```
silently fails. This is because `$()` captures both stdout and stderr, but does not do anything with the stderr.

This PR makes `$()` only capture stdout, so that the stderr is printed to the console. This way, running `result=$(ls /root)` prints the error message:
```
ls: cannot open directory '/root': Permission denied
```

## For community

### DIFF
--- a/news/capture-stdout-without-stderr.rst
+++ b/news/capture-stdout-without-stderr.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* $() no longer silently captures stderr
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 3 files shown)


daniel-shimon:
Hi! Great idea!
Can you add a test for this?

gforsyth (approved):
Thanks a ton @yaxollum!

PR #4338, opened 2021-06-23
feat: add completions for vox

zscholl:
This adds completion for vox. The completer is only registered when the vox xontrib is loaded.

Here it is in action:
xonsh_vox_completions

## For community

### DIFF
--- a/news/vox-completions.rst
+++ b/news/vox-completions.rst
@@ -1,0 +1,3 @@
+**Added:**
+
+* Completions for vox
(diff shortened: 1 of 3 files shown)


jnoortheen:
Actually I have added this with venv name completion as well here https://github.com/xonsh/xonsh/pull/4267

zscholl:
Ahh yeah @jnoortheen that looks like the same functionally so we can close this PR in favor of yours 🙏

jnoortheen:
@zscholl if possible, please take a moment to test my PR.

Issue #4305, opened 2021-06-04
Next release


anki-code:
Hi! Three months have passed since previous release. We have a bunch of new features in https://github.com/xonsh/xonsh/tree/main/news. So let's release them!

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

daniel-shimon:
Hi! Before we release a version:
1. I'd like to wait at least for the completers tutorial in #4285
I'm working on it and will finish it in the coming days
2. Make sure amalgamation works in pip (as discussed in #4290). It seems to work now when `pip install .` from the source code, but `pip install xonsh` with the previous version doesn't do it

PR #4355, opened 2021-07-03
Amalgamated release

daniel-shimon:
Ensure the release version is always amalgamated.

Also update some other release stuff

Promote #4305


xonsh/__init__.py
```diff
@@ -28,6 +26,8 @@
         _sys.modules["xonsh.platform"] = __amalgam__
         pretty = __amalgam__
         _sys.modules["xonsh.pretty"] = __amalgam__
+        xontribs_meta = __amalgam__
+        _sys.modules["xonsh.xontribs_meta"] = __amalgam__
         codecache = __amalgam__
```
jnoortheen, line 31: the init files change every time the amalgamate runs right? I don't think we need to commit this change

daniel-shimon: AFAIK they don't change every time, currently I think they've changed because of the new `xonsh.completers.environment` package

jnoortheen: then looks alright then

jnoortheen (approved):

daniel-shimon:
@jnoortheen thanks!

PR #4379, opened 2021-07-19
feat: support var-args signature for aliases

jnoortheen:
it fixes following scenarios

problem-signature

even though they are the same objects, different signature returned when referenced from aliases dict.

## For community

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,4 @@
 import builtins
-import glob
 import os
 import sys
 import types
@@ -8,6 +7,7 @@ from unittest.mock import MagicMock
 
 import pytest
 
+from xonsh.aliases import Aliases
 from xonsh.built_ins import XonshSession, XSH
 from xonsh.execer import Execer
 from xonsh.jobs import tasks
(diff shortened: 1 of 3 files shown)


daniel-shimon (changes requested):
Looks good! Can you add tests for this scenario?

jnoortheen:
@daniel-shimon try checking now.

daniel-shimon:
Hey @jnoortheen, listen it's a bit difficult when you do these non-related changes like you did here by removing the `xonsh_execer` fixture from other tests.

It is just inserted into some non related commit without any explanation or reasoning and makes the CR process longer.
For instance now I need to go to those fixture and tests and make sure that it makes sense and doesn't break anything else instead of just approving and merging the PR.

I would very much prefer you do these changes in separate PRs so we can have faster and easier CR processes in the project.

jnoortheen:
@daniel-shimon it was just I first tried to add the test and some fixtures in the test_aliases.py. Later moved them to pipelines test. I've created a separate PR.

daniel-shimon (approved):
Great! Thanks!

PR #4384, opened 2021-07-21
test: speedup test suit by reusing costly vars

jnoortheen:
- Not including a news item as it doesn't affect the users

## For community

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,13 +27,11 @@ def source_path():
 
 
 @pytest.fixture
-def xonsh_execer(monkeypatch, session_vars):
+def xonsh_execer(monkeypatch):
     """Initiate the Execer with a mocked nop `load_builtins`"""
-    execer = session_vars["execer"]
-    XSH.load(execer=execer, ctx={})
+    execer = Execer(unload=False)
     monkeypatch.setattr(XSH, "execer", execer)
     yield execer
-    XSH.unload()
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,11 +27,13 @@ def source_path():
 
 
 @pytest.fixture
-def xonsh_execer(monkeypatch):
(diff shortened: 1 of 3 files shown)


daniel-shimon:
Isn't it bad practice to reuse the same `Execer` instance in all the tests that require mocked builtins?
Can't this possibly leak between tests?

gforsyth:
> Isn't it bad practice to reuse the same `Execer` instance in all the tests that require mocked builtins?
> Can't this possibly leak between tests?

This was also my concern.
Test speedups are very welcome but there's already enough magic in pytest that this makes me uncomfortable.

jnoortheen:
> Isn't it bad practice to reuse the same `Execer` instance in all the tests that require mocked builtins?

I am not sure about that part. Execer init calls XSH.load and parser init every time, making it slow. The ctx attribute is the one that can leak between tests if reused. it needs more investigation though. as of now I will revert changes to xonsh_execer fixture. Regarding the xonsh_builtins fixture, I think it patches most of the mutable variables for each function run with monkeypatch. I can say that the situation is same as main branch. 

> Can't this possibly leak between tests?
the mutable variables are `Env's` dict, `CommandsCache`,  `Execer` ctx all of these are patched for each test run xonsh_builtins fixture. 
Also CommandsCache creates a thread each time it is called. Execer() constructs parser each time, which is not necessary. the parser code is not going to change.

jnoortheen:
Actually the test suit runtime is doubled after my changes in PR for removing usage of builtins. I want to improve the coverage situation further and need to run the suit manytimes. it is just one of the low-hanging fruits to have runtime as before.

daniel-shimon:
@jnoortheen by how much does this speed up the tests?

jnoortheen:
around 50%

these are sample run time for py3.6
OS   | before | after
linux | 1m46s | 56s
macos | ~2m | ~1m
windows | ~8m | ~3m

daniel-shimon (approved):
Awesome!

PR #4389, opened 2021-07-25
feat: add xontrib-homebrew

eugenesvk:
replaces xontrib-linuxbrew by expanding OS support to Homebrew on macOS

## For community

### DIFF
--- a/news/xontrib-homebrew.rst
+++ b/news/xontrib-homebrew.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Replaced `xontrib-linuxbrew <https://github.com/eugenesvk/xontrib-linuxbrew>`_ with `xontrib-homebrew <https://github.com/eugenesvk/xontrib-homebrew>`_, which also supports Homebrew on macOS
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/xonsh/xontribs_meta.py
+++ b/xonsh/xontribs_meta.py
@@ -383,6 +383,16 @@ def define_xontribs():
                 url="https://github.com/con-f-use/xontrib-histcpy",
             ),
         ),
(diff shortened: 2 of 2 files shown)


news/xontrib-homebrew.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Replaced `xontrib-linuxbrew <https://github.com/eugenesvk/xontrib-linuxbrew>`_ with `xontrib-homebrew <https://github.com/eugenesvk/xontrib-homebrew>`_, which also supports Homebrew on macOS
+
```
jnoortheen, line 8: in .rst double ` is used for highlight.

eugenesvk: There is `_` at the end to make it a url

\`Link text \<link URL\>\`_

jnoortheen: Yes True, then  test_news.py  needs update.

image

gforsyth: Hey @eugenesvk -- for the moment, can you remove the first link?  The RST checker in our CI isn't super advanced and this is tripping it up.  We should fix that, but it shouldn't block this.

gforsyth:
Hey @eugenesvk -- scratch my previous comment -- @jnoortheen added a proper RST linter to our news check so if you can merge/rebase from `main` everything should work

gforsyth (approved):
Actually, I just realized that it doesn't matter,  since the concern would be that this would fail on `main`, but it won't now.

Thanks for the update!

news/xontrib-homebrew.rst
eugenesvk: ...and here I've almost added a temp fix of removing proper links first (so that the next `single_grave_reg` wouldn't trip over them)
```python
rst_quote_reg = re.compile(r"[^`]`[^`]+`[_]") # RST quote
l = re.sub(rst_quote_reg, '', l) # remove legitimate RST `quotes`_ to avoid errors later
```
But obviously @jnoortheen's solution of using a proper linter is the proper way to go!
Thanks for the fix!

PR #4395, opened 2021-07-27
Fix WSL lazybools

gforsyth:
`lazybool`s that rely on other `lazybool`s need to explicitly call
`bool()` on those values or else a non-bool value gets returned when the
parent's `__bool__` method is called.

No news item since this is fixing up a recent change.

### DIFF
--- a/xonsh/platform.py
+++ b/xonsh/platform.py
@@ -92,12 +92,12 @@ def ON_WSL():
 
 @lazybool
 def ON_WSL1():
-    return ON_WSL and not ON_WSL2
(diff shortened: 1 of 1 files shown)


jnoortheen (approved):

daniel-shimon:
Wait this is a bit wierd
I see bool is implemented for LazyBool

gforsyth:
> Wait this is a bit wierd

the nested `lazybool` stuff has caused problems before and it would be good to fix, but for now, this fixes things (and doesn't hurt any of the laziness)

PR #4396, opened 2021-07-27
Improve import completions

daniel-shimon:
Previously the 'import' completions only completed loaded modules (from `sys.modules`).
Now it uses ipython's logic which involves finding modules in the `sys.path` and finds modules that weren't imported yet!
Also completing `import A,B` and `from A import a,b` statements has been added.

## For community


gforsyth (approved):
Sweet -- this looks good to me!

PR #4401, opened 2021-07-28
Auto suggest word alias

### LINKED ISSUE #3639: Feature request: Autosuggesion accept word
I'm a happy new xonsh user and beside being able to run python commands I do like the autosuggest a lot. However, very often I want to accept only part of the suggestion, so it would be very nice if it is possible to be able to accept only the first word/part. In fish that's possible with `Alt+→`.

Settings:
- keybinding: I'd like `→` for finishing the word and `Alt+→` or `End` to finish the whole line.
- word separators: whitespace and '/' would be a good start I'd say.

I had a look at the code to see if I could quickly implement it, but couldn't really figure out what can be done in xonsh, and what needs to be changed in prompt_toolkit.

### LINKED ISSUE #4330: fish-like partial suggestion completion
Xonsh already supports the fish-like suggestion completion, which is great. I.e., on a greyed suggestion, typing the right arrow completes with the full suggestion.

One thing I miss though is the fish partial suggestion completion. In fish, if I am right, when a suggestion is provided, typing "alt + right_arrow" completes the suggestion only until the next separator. Great to use parts of a suggestion only.

Any way a partial suggestion completion with alt + right_arrow is either already available, or could be added? :) .

daniel-shimon:
Closes #4330 
Closes #3639

This adds a `CTRL-Right` shortcut to add a single auto-suggestion word (like in fish).
This is already implemented in ptk as in `ALT-f` which isn't very ergonomic or expected, so we add a comfortable alias!

## For community

### DIFF
--- a/docs/keyboard_shortcuts.rst
+++ b/docs/keyboard_shortcuts.rst
@@ -31,4 +31,6 @@ Xonsh comes pre-baked with a few keyboard shortcuts. The following is only avail
       - Cut highlighted section
     * - Ctrl-V *[Beta]*
       - Paste clipboard contents
+    * - Ctrl-Right
+      - Complete a single auto-suggestion word
 
--- a/news/auto-suggest-word-alias.rst
+++ b/news/auto-suggest-word-alias.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Add ``CTRL-Right`` key binding to complete a single auto-suggestion word.
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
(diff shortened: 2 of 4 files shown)


jnoortheen:
Any reason using Ctrl+Right ? these issues mention Alt+Right

daniel-shimon:
@jnoortheen yeah all our other gestures are with ctrl (moving words and highlighting words), and also the original fish gesture is actually with ctrl and not alt as mentioned in the issue

jnoortheen (approved):

daniel-shimon:
@jnoortheen thanks!

PR #4406, opened 2021-08-01
Change !() to also capture background subprocesses

### LINKED ISSUE #4392: !() doesn't capture information from subprocesses initiated in the background
## Steps to Reproduce

```
proc = !(ls &)
[1]+ running: ls & & (22822)
print(proc)
>>> None
```

## Expected behaviour
Something similar to this:
```
proc = !(ls)
proc
>>> CommandPipeline(
  stdin=<_io.BytesIO object at 0x7f333de561d0>,
  stdout=<_io.BytesIO object at 0x7f333de56180>,
  stderr=<_io.BytesIO object at 0x7f333de569f0>,
  pid=17977,
  returncode=0,
  args=['ls'],
  ...)
```

I can't see why a background process should return None. If I want to access said process' pid, then I have to do this:
```
import xonsh
xonsh.jobs.get_task(xonsh.jobs.tasks[-1])['pids']
```
Which is also a pretty slow expression (to write and to run).

## For community

yaxollum:
Closes #4392 

This PR makes `!()` to also capture information for background subprocesses. Currently, `!()` returns `None` for background subprocesses.

In the final `if` structure of the `run_subproc()` function, the `captured == "object"` branch does not call `command.end()` before returning (unlike all the other branches), so it can be used for background subprocesses.

https://github.com/xonsh/xonsh/blob/f5acbe30d40322ecab40d4e46d0c5f4417e59aaa/xonsh/procs/specs.py#L898-L909

## For community

### DIFF
--- a/news/capture-object-background.rst
+++ b/news/capture-object-background.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Changed !() to also capture background subprocesses
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/tests/procs/test_specs.py
+++ b/tests/procs/test_specs.py
@@ -102,6 +102,23 @@ def test_capture_always(capfd, thread_subprocs, capture_always):
     assert exp in capfd.readouterr().out
 
 
+@skip_if_on_windows
+@pytest.mark.parametrize(
+    "captured, exp_is_none",
+    [
+        ("object", False),
+        ("stdout", True),
+        ("hiddenobject", True),
+        (False, True),
+    ],
+)
+def test_run_subproc_background(captured, exp_is_none):
+
+    cmds = (["echo", "hello"], "&")
+    return_val = run_subproc(cmds, captured)
+    assert (return_val is None) == exp_is_none
+
+
 @pytest.mark.parametrize("thread_subprocs", [False, True])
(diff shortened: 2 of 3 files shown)


daniel-shimon:
Hi! Can you simplify the logic to:
```
if capture == "object":
elif background:
elif capture == ...
```
This is getting a bit tangled

daniel-shimon (approved):
Nice @yaxollum ! Thanks!

polvoazul:
Thanks for the fast response to my issue! I am truly impressed. :)

polvoazul:
Just installed from github main and tested it! It works fine, but i noticed the behavior is not implemented with ![]. I don't need this now but it probably should work similarly, for consistency, right?

daniel-shimon:
Not really because `![]` waits for the process to end, unlike `!()`

polvoazul:
Wait, from the docs i thought the only difference between them was the captured stdout/stderr. Why would ![sleep 10 &] wait for the process to end?

daniel-shimon:
Running commands by default uses `![]`, meaning running `sleep 10 and echo hi` actually turns into `![sleep 10] and ![echo hi]`

polvoazul:
Fair enough, but what about ![command **&**] ? Shouldn't the ampersand be respected?

Think of this use case: I want to run a process in the background, streaming its stdout/stderr directly to the terminal.

Is there any way of doing that? (obviously i can use python's subprocess.run, but i mean using xonsh literals)

daniel-shimon:
Yeah on second thought it makes sense.
`!()` and `![]` differ by capturing the output, but both return objects that can be queried, so both can contain a background process

Issue #4394, opened 2021-07-27
`xonsh -c` incorrectly sets $XONSHRC and $XONSHRC_DIR to empty lists, breaking any further interactive xonsh subprocesses, as of 0.10


Qyriad:
In xonsh 0.9.27, `xonsh -c sh` would set `$XONSHRC` to the relevant RC files. As of xonsh 0.10, `xonsh -c sh` sets $XONSHRC, but to nothing (which xonsh interprets as an empty list), meaning if any subprocess of `xonsh -c` then itself tries to start an interactive shell, it will fail to read *any* RC files at all.

Though probably not the only one, I have discovered at least one functional and practical consequence of this if `xonsh` is your default shell and you use Gnome under GDM. GDM will use `$SHELL -c` to start the session, setting `$XONSHRC=[]` for the Wayland compositor process, and thus any `xonsh` instances spawned by any subprocesses of the compositor will act as if there is no RC file at all, prompting the user to *create* a xonshrc file despite there already being one!

## xonfig (with RC)

<details>

```
$ xonfig
+------------------+--------------------+
| xonsh            | 0.10.0             |
| Python           | 3.9.6              |
| PLY              | 3.11               |
| have readline    | True               |
| prompt toolkit   | 3.0.8              |
| shell type       | prompt_toolkit     |
| history backend  | json               |
| pygments         | 2.9.0              |
| on posix         | True               |
| on linux         | True               |
| distro           | arch               |
| on wsl           | False              |
| on darwin        | False              |
| on windows       | False              |
| on cygwin        | False              |
| on msys2         | False              |
| is superuser     | False              |
| default encoding | utf-8              |
| xonsh encoding   | utf-8              |
| encoding errors  | surrogateescape    |
| on jupyter       | False              |
| jupyter kernel   | None               |
| xontrib 1        | output_search      |
| xontrib 2        | whole_word_jumping |
+------------------+--------------------+
```

</details>

## Expected Behavior
For `xonsh -c "env | rg XONSHRC"` to either print *nothing* (the variable being unset), or to print the proper RC files as it did until 0.10 (and still does when passed `--login`. That is, `xonsh -c` should either *not* set `$XONSHRC` at all, or should set it to the RC files as it did until 0.10.

## Current Behavior
`xonsh -c "env | rg XONSHRC"` prints `XONSHRC=` — `xonsh -c` sets `$XONSHRC`, but not to any value (making it empty).

## Steps to Reproduce
```sh
[...]

PR #4408, opened 2021-08-03
Don't modify XONSHRC and XONSHRC_DIR during startup

chronitis:
This is a possible solution to #4394, as discussed in https://github.com/xonsh/xonsh/issues/4394#issuecomment-890979982.

This PR:

* Ceases to modify `XONSHRC` and `XONSHRC_DIR` during startup, which (as per the issue above) causes problems if xonsh processes are nested. These variables will be left as-is even if their contents have been ignored (due to command line flags, for example).
* Drops the environment variable `LOADED_RC_FILES`, which contained a list of whether RC files loading was successful, but not the corresponding paths, making the output hard to interpret.
* Adds the list of loaded RC files to the output of `xonfig`, as a replacement for inspecting `XONSHRC`, `XONSHRC_DIR` and `LOADED_RC_FILES` within the session.

## For community


gforsyth:
@chronitis -- so it seems like the only way to set `XONSHRC_DIR` to something other than the default now is to use a tool like `env` or similar to set it before starting `xonsh`.
I'm not necessarily opposed to this, and the recursive nature of setting that variable in the `xonshrc` is definitely an issue otherwise.

chronitis:
Yes, that's right - it either uses the default value, or an environment variable.

As currently implemented, it is too late to modify either `XONSHRC` or `XONSHRC_DIR` in another RC file, since the list of files to be loaded has already been determined before it starts sourcing them. In principle, you could change the implementation to check after each file is sourced if those options have changed, but I'm not sure that's necessarily better - if you change those variables half way through loading multiple RC files, deciding how to correct handle the changes is probably non-trivial or arbitrary, and probably risks infinite loops.

 I didn't add a CLI option when I originally defined it, reasoning that if you wanted to specify RC files on the command line, it was more likely to be just a single file to specify. It would probably be fairly easy to either add `--rc-dir DIR` or special-case `--rc DIR` to do the same thing, if that would be useful.

gforsyth:
> It would probably be fairly easy to either add `--rc-dir DIR` or special-case `--rc DIR` to do the same thing, if that would be useful.

I think we should have one of these -- I'm slightly in favor of allowing a directory to be passed to `--rc` but if it proves problematic I'm fine with adding a specific `--rc-dir` switch

chronitis:
Yes, I can do that - but do you want it part of the same PR? It seems slightly different from the original scope of handling how environment variables affect nested xonsh processes.

gforsyth (approved):

gforsyth:
Good point @chronitis !  
Thanks for this fix!

PR #4417, opened 2021-08-12
fix: ignore bad file-description error during bash completion generation

### LINKED ISSUE #4414: bash-completer bad-file descriptor
image

## xonfig

<details>

```
$ xonfig
```

</details>

### Traceback (if applicable)

<details>

```
disown xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
  File "/usr/lib/python3.9/subprocess.py", line 507, in run
    stdout, stderr = process.communicate(input, timeout=timeout)
  File "/usr/lib/python3.9/subprocess.py", line 1134, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "/usr/lib/python3.9/subprocess.py", line 1999, in _communicate
    data = os.read(key.fd, 32768)

  File "/src/shell/xsh/xonsh/completer.py", line 135, in generate_completions
    out = func(completion_context)
  File "/src/shell/xsh/xonsh/completers/tools.py", line 156, in _completer
    return func(context.command)
  File "/src/shell/xsh/xonsh/completers/bash.py", line 36, in complete_from_bash
    comps, lprefix = bash_completions(
  File "/src/shell/xsh/xonsh/completers/bash_completion.py", line 387, in bash_completions
    out = subprocess.check_output(
  File "/usr/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
  File "/usr/lib/python3.9/subprocess.py", line 1019, in __exit__
    self.stdout.close()
```

</details>

## Steps to Reproduce

This happens occasionally when trying to complete a command with tab keypress. 

## For community

jnoortheen:
fixes #4414

## For community

### DIFF
--- a/xonsh/completers/bash_completion.py
+++ b/xonsh/completers/bash_completion.py
@@ -394,7 +394,7 @@ def bash_completions(
             raise ValueError
     except (
         subprocess.CalledProcessError,
-        FileNotFoundError,
+        OSError,
         ValueError,
     ):
         return set(), 0


daniel-shimon:
Can you try to further understand why this is happening at all?
Just having the completer fail some of the time doesnt seem like a very
helpful fix..

jnoortheen:
the error comes from subprocess module. If other exceptions are ignored, I think this is similar

daniel-shimon:
I'm not so sure.. `FileNotFoundError` means bash cannot be found. Maybe it isn't installed or isn't in the path and that is exeptably ignored.
But this might be due to something we're doing wrong when calling bash, and since I can't reproduce this can you try to understand why this is happening?

jnoortheen:
Yes I want a proper fix too. But the problem is that, it is not reproducible. I have seen somewhere that python garbage collector is the reason for such errors. I will try to check it more deeper later. Now you can hold.

jnoortheen:
Since the source of error can't be identified, I will close this for now.

PR #4419, opened 2021-08-16
fix: ptk completer renders new line as ^J

jnoortheen:
## For community

### DIFF
--- a/xonsh/ptk_shell/completer.py
+++ b/xonsh/ptk_shell/completer.py
@@ -98,13 +98,17 @@ class PromptToolkitCompleter(Completer):
         for comp in completions:
             # do not display quote
             if isinstance(comp, RichCompletion):
-                desc = comp.description or ""
+                desc = (
+                    comp.description.replace(os.linesep, " ")
+                    if comp.description
+                    else None
+                )
                 yield Completion(
                     comp,
                     -comp.prefix_len if comp.prefix_len is not None else -plen,
                     display=comp.display or comp[pre:].strip("'\""),
                     # ptk doesn't render newlines. This can be removed once it is supported.
-                    display_meta=desc.replace(os.linesep, " "),
+                    display_meta=desc,
                     style=comp.style or "",
                 )
             elif isinstance(comp, Completion):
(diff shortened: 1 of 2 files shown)


xonsh/ptk_shell/completer.py
```diff
@@ -98,11 +98,17 @@ def get_completions(self, document, complete_event):
...
+                    comp.description.replace(os.linesep, " ")
+                    if comp.description
+                    else None
+                )
                 yield Completion(
                     comp,
                     -comp.prefix_len if comp.prefix_len is not None else -plen,
                     display=comp.display or comp[pre:].strip("'\""),
-                    display_meta=comp.description or None,
+                    # ptk doesn't render newlines. This can be removed once it is supported.
```
daniel-shimon, line 110: Shouldn't this comment be above the `replace` line?

xonsh/ptk_shell/completer.py
```diff
@@ -98,11 +98,17 @@ def get_completions(self, document, complete_event):
         for comp in completions:
             # do not display quote
             if isinstance(comp, RichCompletion):
+                desc = (
+                    comp.description.replace(os.linesep, " ")
```
daniel-shimon, line 102: Maybe `. ` (dot newline) so the sentences appear split?

jnoortheen: putting dot would mean certain meaning. then we may want to title the new start ...

xonsh/ptk_shell/completer.py
jnoortheen: updated.

xonsh/ptk_shell/completer.py
jnoortheen: we can add some special characters like the return character. is that supported in most of the terminals? or we can have that in a setting ?

daniel-shimon: I don't think this is a big deal
Just decide on something that looks nice

jnoortheen: space is fine for now

daniel-shimon (approved):
Great!

PR #4420, opened 2021-08-16
Lab/py control file

jnoortheen:
## For community

### DIFF
--- a/docs/xonshrc.py
+++ b/docs/xonshrc.py
@@ -1,0 +1,23 @@
+from xonsh.built_ins import XSH
+
+env = XSH.env
+# adjust some paths
+env["PATH"].append("/home/scopatz/sandbox/bin")
+env["LD_LIBRARY_PATH"] = ["/home/scopatz/.local/lib", "/home/scopatz/miniconda3/lib"]
+
+# alias to quit AwesomeWM from the terminal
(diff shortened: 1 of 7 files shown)


gforsyth:
Thanks for putting this in @jnoortheen -- I'd like to see a test for this.
Also, in re: your news item, is it only 3.8+ syntax supported?  This seems misleading -- I can't imagine it would work if the underlying environment is 3.7

jnoortheen:
>  is it only 3.8+ syntax supported?

xonsh parser is 3.8 only. some 3.9/3.10 syntaxes (match statement ...) will fail. I wanted to mention that.

gforsyth:
Ahh, so to me, 3.8+ implies support for 3.9 and (soon) 3.10

jnoortheen:
> Ahh, so to me, 3.8+ implies support for 3.9 and (soon) 3.10

I will update that part to mention just the python part. Also we can mention that it will get parsed faster than xonsh control files.

gforsyth:
This looks good @jnoortheen -- could you also add a short note to https://github.com/xonsh/xonsh/blob/main/docs/xonshrc.rst about the new option?

gforsyth (approved):
Great, thanks for putting this together @jnoortheen !

PR #4422, opened 2021-08-17
Support `--rc DIR`

chronitis:
This is a followup to https://github.com/xonsh/xonsh/pull/4408#issuecomment-899629748, adding support for `--rc DIR` (which loads, in sorted order, `DIR/*.xsh` during startup).

This also completely drops `--config-path`, which still existed as an option but was no longer actually used anywhere (but was the only other user of `path_argument()`, which I redefined to accept files or directories).

## For community

### DIFF
--- a/news/rcdir-cli.rst
+++ b/news/rcdir-cli.rst
@@ -1,0 +1,27 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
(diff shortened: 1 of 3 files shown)


gforsyth (approved):
Thanks @chronitis !  This looks (and works) great!

PR #4427, opened 2021-08-19
Restore _repr_pretty_ for ${...} Env class

gforsyth:
Restoring pretty repr for the `Env` class, raised in #4424

Now when you type just `${...}` you see your environment printed

## For community

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -2057,6 +2057,15 @@ class Env(cabc.MutableMapping):
     def __repr__(self):
         return "{0}.{1}(...)".format(self.__class__.__module__, self.__class__.__name__)
 
+    def _repr_pretty_(self, p, cycle):
+        name = "{0}.{1}".format(self.__class__.__module__, self.__class__.__name__)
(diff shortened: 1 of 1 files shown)


xonsh/environ.py
```diff
@@ -2057,6 +2057,15 @@ def __str__(self):
     def __repr__(self):
         return "{0}.{1}(...)".format(self.__class__.__module__, self.__class__.__name__)
 
+    def _repr_pretty_(self, p, cycle):
+        name = "{0}.{1}".format(self.__class__.__module__, self.__class__.__name__)
+        with p.group(0, name + "(", ")"):
```
daniel-shimon, line 2062: Maybe provide a single space indentation with `p.group(1, ...)`? Makes it clearer IMO

gforsyth: I was just restoring the original code, but I agree, that looks nicer.
Also changed the `name` string to an f-string

daniel-shimon (approved):
Great! Thanks!

PR #4429, opened 2021-08-20
Lab/lazy shell attrs

jnoortheen:
## For community

### DIFF
--- a/xonsh/ptk_shell/shell.py
+++ b/xonsh/ptk_shell/shell.py
@@ -3,6 +3,7 @@
 import os
 import re
 import sys
+import time
 from functools import wraps
 from types import MethodType
 
@@ -22,7 +23,6 @@ from xonsh.ptk_shell.key_bindings import load_xonsh_bindings
 
 from prompt_toolkit import ANSI
 from prompt_toolkit.auto_suggest import AutoSuggestFromHistory
-from prompt_toolkit.lexers import PygmentsLexer
 from prompt_toolkit.enums import EditingMode
 from prompt_toolkit.key_binding.bindings.emacs import (
     load_emacs_shift_selection_bindings,
@@ -205,7 +205,7 @@ class PromptToolkitShell(BaseShell):
             ptk_args.setdefault("clipboard", PyperclipClipboard())
         self.prompter: PromptSession = PromptSession(**ptk_args)
 
-        self.prompt_formatter = PTKPromptFormatter(self.prompter)
+        self.prompt_formatter = PTKPromptFormatter(self)
         self.pt_completer = PromptToolkitCompleter(self.completer, self.ctx, self)
         ptk_bindings = self.prompter.app.key_bindings
         self.key_bindings = load_xonsh_bindings(ptk_bindings)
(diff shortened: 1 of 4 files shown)


xonsh/ptk_shell/shell.py
```diff
@@ -298,58 +355,12 @@ def singleline(
...
-                    print_warning(
-                        f"Error applying style override!\n{style_exception}\n"
-                    )
-
-            prompt_args["style"] = style
-            events.on_timingprobe.fire(name="on_post_prompt_style")
-
         if env["ENABLE_ASYNC_PROMPT"]:
             # once the prompt is done, update it in background as each future is completed
             prompt_args["pre_run"] = self.prompt_formatter.start_update
```
daniel-shimon, line 360: Seems to me calling `add_attr` from the `pre_fun` will run it lazily. Can you try that to see if it works?

Doing it in another thread isn't really a good solution because it isn't io-bound and will do more harm than good (https://stackoverflow.com/questions/62154869/multi-threading-and-single-threading-performance-issues-in-cpu-bound-task)
Also running in it a thread isn't really _lazy_, but _parallel_ (and won't really be parallel because of the GIL)

xonsh/ptk_shell/updator.py
```diff
@@ -157,37 +157,53 @@ def submit_section(
 class PromptUpdator:
     """Handle updating multiple AsyncPrompt instances prompt/rprompt/bottom_toolbar"""
 
-    def __init__(self, session: PromptSession):
+    def __init__(self, shell):
+        from xonsh.ptk_shell.shell import PromptToolkitShell
```
daniel-shimon, line 161: Does it really change the uptime if we import this normally? If not pls move this to the top of the file

xonsh/ptk_shell/shell.py
```diff
@@ -205,7 +204,7 @@ def __init__(self, **kwargs):
             ptk_args.setdefault("clipboard", PyperclipClipboard())
         self.prompter: PromptSession = PromptSession(**ptk_args)
 
-        self.prompt_formatter = PTKPromptFormatter(self.prompter)
+        self.prompt_formatter = PTKPromptFormatter(self)
```
daniel-shimon, line 207: What do these changes have to do with the lazy loading?

xonsh/ptk_shell/updator.py
```diff
@@ -157,37 +157,53 @@ def submit_section(
...
         self.executor = Executor()
+        self.futures = {}
+        self.attrs_loaded = None
 
     def add(self, prompt_name: tp.Optional[str]) -> tp.Optional[AsyncPrompt]:
         # clear out old futures from the same prompt
         if prompt_name is None:
             return None
 
-        if prompt_name in self.prompts:
```
daniel-shimon, line 170: Why remove this check?

xonsh/ptk_shell/updator.py
```diff
@@ -157,37 +157,53 @@ def submit_section(
...
+            self.attrs_loaded = self.executor.thread_pool.submit(self.add_attrs)
 
-        for th in threads:
-            th.start()
+        prompts = list(self.prompts)  # removal safe
+        for pt_name in prompts:
+            if pt_name not in self.prompts:
+                continue
+            prompt = self.prompts[pt_name]
+            future = self.executor.thread_pool.submit(
```
daniel-shimon, line 196: Is this thread pool created in xonsh? Can we use it safely?

xonsh/ptk_shell/updator.py
```diff
@@ -157,37 +157,53 @@ def submit_section(
...
+                prompt.start_update, self.on_complete
+            )
+            self.futures[pt_name] = future
```
daniel-shimon, line 199: Why save this future? Isn't `start_update` a one-time function?
Seems like we don't care about its return value, and anyway canceling a threadpool submittion doesn't do anything AFAIK

daniel-shimon (changes requested):

xonsh/ptk_shell/shell.py
jnoortheen: it is like running it later. Since it is an interactive prompt, it is going to wait for user input at later point. This is just choosing to run certain things later. I know it is not parallel. 

> Seems to me calling add_attr from the pre_fun will run it lazily. Can you try that to see if it works?

add_attr ?

xonsh/ptk_shell/updator.py
jnoortheen: this is to prevent circular import

xonsh/ptk_shell/shell.py
jnoortheen: it needs to call a method on shell class

xonsh/ptk_shell/updator.py
jnoortheen: this threadpool is created for async-prompt related functions. What do you mean by safe?

xonsh/ptk_shell/updator.py
jnoortheen: this is to cancel it later if not started already.

xonsh/ptk_shell/shell.py
daniel-shimon: Instead of submitting `self.add_attrs` to the threadpool in `start`, set `ptk_kwargs['pre_run']` to a function which calls `self.add_attrs` and `self.prompt_formatter.start_update`.
I think this will call it lazily as needed (not sure though)

jnoortheen: `pre_run` is called in the same thread. The idea here is to offload the style related call to another thread. Because it is not essential for the prompt to start. Once the user starts typing we will be seeing the color. 

Also this asynchronous prompt sets style per session. Otherwise it is set per prompt.

xonsh/ptk_shell/updator.py
jnoortheen: the stop method already checks for this. I just removed the redundant check

daniel-shimon: > I see it checks for `self.futures` but not `self.prompts`

jnoortheen: there are two checks in them 
image

daniel-shimon (changes requested):
This creates an unpleasant user experience since the prompt initially loads without colors and it looks bad:
Animation

Maybe we can return the previous style in `singleline` until the new one loads?

daniel-shimon:
Also I don't think there's a real reason to create the PygmentsLexer object each time right?

jnoortheen:
> Also I don't think there's a real reason to create the PygmentsLexer object each time right?

This would be because the style is set in interactive shell. This behaviour is not changed when async-prompt is disabled.


> This creates an unpleasant user experience since the prompt initially loads without colors and it looks bad:

That would be only the first time. The attrs is already set for next times. The color change you see is for branch attribute. which is shown in grey to show that it is still old value. It is updated later by threads. That is the primary reason having async-prompts https://github.com/xonsh/xonsh/pull/3758

daniel-shimon (approved):
Thanks! The async prompt is really cool!

PR #4437, opened 2021-08-26
argparserAlias - vox

jnoortheen:
## For community

### DIFF
--- a/tests/test_vox.py
+++ b/tests/test_vox.py
@@ -287,14 +287,20 @@ def test_autovox(xession, tmpdir, load_vox):
 
 
 @pytest.fixture
-def venvs(tmpdir):
+def venv_home(tmpdir):
+    """Path where VENVs are created"""
+    return tmpdir
+
+
+@pytest.fixture
+def venvs(venv_home):
     """create bin paths in the tmpdir"""
     from xonsh.dirstack import pushd, popd
 
-    pushd([str(tmpdir)])
+    pushd([str(venv_home)])
     paths = []
     for idx in range(2):
-        bin_path = tmpdir / f"venv{idx}" / "bin"
+        bin_path = venv_home / f"venv{idx}" / "bin"
         paths.append(bin_path)
 
         (bin_path / "python").write("", ensure=True)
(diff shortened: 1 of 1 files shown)


daniel-shimon:
I'm not familiar enough with vox to test this out maybe @gforsyth or @anki-code can give us a hand?

anki-code:
Looks working for me:

image

xontrib/vox.py
```diff
@@ -1,48 +1,41 @@
 """Python virtual environment manager for xonsh."""
 
-import sys
-import textwrap
 import xontrib.voxapi as voxapi
-import xonsh.lazyasd as lazyasd
+from xonsh.built_ins import XSH
+from xonsh.cli_utils import Annotated, Arg, ArgParserAlias
```
gforsyth, line 5: In #4454 you are using a different import scheme for the `cli_utils` -- it should be consistent across all usage to avoid double-counting in amalgamation

jnoortheen: Yes I tried to keep 'import ... as' but it raised error sometimes during amalgamation. I will try changing it again.

jnoortheen: I think those times are when the module level is same as `xonsh.*` . That makes sense since all those definitions will end up in single module. I think we can keep this style in other module levels.

xontrib/vox.py
```diff
@@ -75,181 +68,169 @@ def parser():
...
-                    file=sys.stderr,
+                self.parser.error(
+                    f'The "{name}" environment is currently active. '
+                    'In order to remove it, deactivate it first with "vox deactivate".\n',
                 )
-                return
             except KeyError:
-                print('"%s" environment doesn\'t exist.\n' % name, file=sys.stderr)
-                return
+                self.parser.error('"%s" environment doesn\'t exist.\n' % name)
```
gforsyth, line 230: f-string?

xontrib/vox.py
```diff
@@ -75,181 +68,169 @@ def parser():
...
         try:
-            self.vox.activate(args.name)
+            self.vox.activate(name)
         except KeyError:
-            print(
+            self.parser.error(
                 'This environment doesn\'t exist. Create it with "vox new %s".\n'
-                % args.name,
-                file=sys.stderr,
+                % name,
```
gforsyth, line 157: f-string?

gforsyth (commented):
style nits but nothing blocking.  I don't use `vox` or `virtualenv` -- @anki-code do you already use vox?  I'll defer to your review

xontrib/vox.py
jnoortheen: done

PR #4440, opened 2021-08-27
feat: use argparserAlias for xexec

jnoortheen:
## For community

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -639,7 +639,7 @@ def source_cmd(args, stdin=None):
         return source_foreign(args, stdin=stdin)
 
 
-def xexec(
+def xexec_fn(
     command: Annotated[tp.List[str], Arg(nargs=argparse.REMAINDER)],
     login: Annotated[bool, Arg("-l", "--login", action="store_true")] = False,
     clean: Annotated[bool, Arg("-c", "--clean", action="store_true")] = False,
@@ -702,7 +702,7 @@ def xexec(
         )
 
 
-xexec_alias = ArgParserAlias(func=xexec, has_args=True, prog="xexec")
+xexec = ArgParserAlias(func=xexec_fn, has_args=True, prog="xexec")
 
 
 def xonfig(args, stdin=None):
@@ -783,8 +783,8 @@ def make_default_aliases():
         "EOF": xonsh_exit,
         "exit": xonsh_exit,
         "quit": xonsh_exit,
-        "exec": xexec_alias,
-        "xexec": xexec_alias,
+        "exec": xexec,
+        "xexec": xexec,
         "source": source_alias,
         "source-zsh": ["source-foreign", "zsh", "--sourcer=source"],
         "source-bash": ["source-foreign", "bash", "--sourcer=source"],
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -677,7 +677,6 @@ def xexec_fn(
     please use the xexec command directly or dive into subprocess mode
     explicitly with ![exec command]. For more details, please see
     http://xon.sh/faq.html#exec.
-    # todo: doc below params are ignored from generated CLI. make sure to show in epilog
     """
     if len(command) == 0:
         return (None, "xonsh: exec: no command specified\n", 1)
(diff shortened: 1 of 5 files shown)


xonsh/aliases.py
```diff
@@ -665,44 +678,32 @@ def xexec(args, stdin=None):
...
         return (
             None,
-            "xonsh: exec: file not found: {}: {}"
-            "\n".format(e.args[1], args.command[0]),
+            "xonsh: exec: file not found: {}: {}" "\n".format(e.args[1], command[0]),
             1,
         )
 
 
+xexec = ArgParserAlias(func=xexec_fn, has_args=True, prog="xexec")
```
daniel-shimon, line 704: Why not call the function `xexec` and overwrite the variable so that `prog` is deduced?

xonsh/cli_utils.py
```diff
@@ -85,7 +85,21 @@ def _get_param_doc(doc: str, param: str) -> tp.Iterator[str]:
...
+            and set(lin.strip()) == {"-"}
+            and (idx != 0)
+            and (not lines[idx - 1].startswith("Parameters"))
+        ):
+            return os.linesep.join(lines[idx - 1 :])
+
+    return ""
+
+
+def get_doc(func: tp.Union[tp.Callable, str], parameter: str = None, epilog=False):
```
daniel-shimon, line 102: `epilog` -> `epilogue`

xonsh/cli_utils.py
```diff
@@ -85,7 +85,21 @@ def _get_param_doc(doc: str, param: str) -> tp.Iterator[str]:
         yield lin
 
 
-def get_doc(func: tp.Union[tp.Callable, str], parameter: str = None):
+def _get_epilog_doc(doc: str) -> str:
+    lines = doc.splitlines()
+    for idx, lin in enumerate(lines):
+        if (
+            lin.startswith("-")
```
daniel-shimon, line 92: `lin.lstrip`?

xonsh/cli_utils.py
```diff
@@ -85,7 +85,21 @@ def _get_param_doc(doc: str, param: str) -> tp.Iterator[str]:
         yield lin
 
 
-def get_doc(func: tp.Union[tp.Callable, str], parameter: str = None):
+def _get_epilog_doc(doc: str) -> str:
+    lines = doc.splitlines()
+    for idx, lin in enumerate(lines):
```
daniel-shimon, line 90: Isn't there a better way to parse pandoc text?

daniel-shimon:
Why not show the notes section like it appears I nthe documentation, after the parameters?

xonsh/aliases.py
jnoortheen: the xexec_fn can be used during tests and other pure Python scripts. Also explicitly we are defining the object.

xonsh/cli_utils.py
jnoortheen: this is just short

xonsh/cli_utils.py
jnoortheen: since we use cleandoc, any padding will get removed.

xonsh/cli_utils.py
jnoortheen: this doesn't need results from other two functions and independent. Let me know if you have any suggestion

jnoortheen:
> Why not show the notes section like it appears I nthe documentation, after the parameters?

After the cli_utils change, it shows the notes section.

xonsh/cli_utils.py
```diff
@@ -105,10 +121,13 @@ def get_doc(func: tp.Union[tp.Callable, str], parameter: str = None):
 
     doc = inspect.getdoc(func) or ""
```
daniel-shimon, line 122: So call `cleandoc` here before passing the docs to `_get_epilog_doc` and friends

jnoortheen: it already does https://docs.python.org/3/library/inspect.html#inspect.getdoc

daniel-shimon:
So why are you running it again letr in the function?

jnoortheen:
> So why are you running it again letr in the function?

the param doc will have some indendation. it is to strip such extra noise.

daniel-shimon:
> the param doc will have some indendation. it is to strip such extra noise.

I understand 
So can you add a comment explaining this?

jnoortheen:
> I understand
> So can you add a comment explaining this?


added

daniel-shimon (approved):
Thanks!

PR #4441, opened 2021-08-28
feat: use ArgparserAlias for xontrib

jnoortheen:
## For community


xonsh/xontribs.py
```diff
@@ -78,8 +79,38 @@ def update_context(name, ctx=None):
     return ctx.update(modctx)
 
 
-def xontribs_load(names, verbose=False):
-    """Load xontribs from a list of names"""
+def xontrib_names_completer(**_):
```
daniel-shimon, line 82: Can you add a test for this?

xonsh/xontribs.py
```diff
@@ -118,11 +144,11 @@ def xontrib_installed(names: tp.Set[str]):
     return installed_xontribs
 
 
-def xontrib_data(ns):
+def xontrib_data(names=()):
     """Collects and returns the data about xontribs."""
     meta = get_xontribs()
     data = {}
-    names: tp.Set[str] = set() if not ns else set(ns.names)
+    names: tp.Set[str] = set(names or ())
```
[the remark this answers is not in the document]
jnoortheen: sometimes, the value from ns would be None. this is to avoid.

xonsh/xontribs.py
jnoortheen: added

daniel-shimon (approved):
Cool!

PR #4442, opened 2021-08-28
Ap/disown

jnoortheen:
## For community

### DIFF
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -11,6 +11,7 @@ import collections
 import typing as tp
 
 from xonsh.built_ins import XSH
+from xonsh.cli_utils import Annotated, Arg, ArgParserAlias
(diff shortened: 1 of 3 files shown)


xonsh/jobs.py
```diff
@@ -447,53 +453,50 @@ def bg(args, stdin=None):
         return res
 
 
-def disown(args, stdin=None):
-    """
-    xonsh command: disown
+# todo: add tests for completer
```
daniel-shimon, line 456: Let me know when this is done and I'll merge

daniel-shimon (changes requested):

xonsh/jobs.py
jnoortheen: added tests

daniel-shimon (approved):
Great! Thanks!

PR #4446, opened 2021-08-28
Ap/dirstack

jnoortheen:
## For community

### DIFF
--- a/news/ap-dirstack.rst
+++ b/news/ap-dirstack.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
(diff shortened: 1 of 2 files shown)


daniel-shimon (commented):
Rename the functions to leading underscores so that `prog` will be deduced.
E.g. `dirs_fn` -> `_dirs`

jnoortheen:
> Rename the functions to leading underscores so that `prog` will be deduced.
> E.g. `dirs_fn` -> `_dirs`

these can be used as Python interface, so better not be making them as private. Setting them explicitly is good sometimes.

daniel-shimon (approved):
Thanks!

PR #4447, opened 2021-08-29
Add test for readline shell

jnoortheen:
internal change: no need for news item.

## For community

### DIFF
--- a/tests/test_readline_shell.py
+++ b/tests/test_readline_shell.py
@@ -28,7 +28,7 @@ def test_render_completions(prefix, completion, prefix_len, readline_completion)
 @pytest.mark.parametrize(
     "line, exp",
     [
-        [repr("hello"), None],
+        [repr("hello"), repr("hello")],
         ["2 * 3", "6"],
     ],
 )
@@ -41,4 +41,4 @@ def test_rl_prompt_cmdloop(line, exp, readline_shell, capsys):
     # xonsh, doesn't write all its output to shell.stdout
     # so capture sys.stdout
     out, err = capsys.readouterr()
-    assert out.strip() == exp or line
+    assert out.strip() == exp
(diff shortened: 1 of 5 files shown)


tests/test_readline_shell.py
```diff
@@ -23,3 +23,22 @@ def test_render_completions(prefix, completion, prefix_len, readline_completion)
...
+def test_rl_prompt_cmdloop(line, exp, readline_shell, capsys):
+    shell = readline_shell
+    shell.use_rawinput = False
+    shell.stdin.write(f"{line}\nexit\n")  # note: terminate with '\n'
+    shell.stdin.seek(0)
+    shell.cmdloop()
+    # xonsh, doesn't write all its output to shell.stdout
+    # so capture sys.stdout
+    out, err = capsys.readouterr()
+    assert out.strip() == exp or line
```
daniel-shimon, line 44: What's `line` doing here? This can never fail

tests/test_readline_shell.py
```diff
@@ -23,3 +23,22 @@ def test_render_completions(prefix, completion, prefix_len, readline_completion)
     assert _render_completions({completion}, prefix, prefix_len) == [
         readline_completion
     ]
+
+
+@pytest.mark.parametrize(
+    "line, exp",
+    [
+        [repr("hello"), None],
```
daniel-shimon, line 31: Why won't this show hello?

daniel-shimon (changes requested):

tests/test_readline_shell.py
jnoortheen: it shows hello. instead of repeating, i just use the line again

tests/test_readline_shell.py
jnoortheen: when exp is none this gets used.

tests/test_readline_shell.py
jnoortheen: It is for it to be concise, anyway I just refactored it.

jnoortheen:
@daniel-shimon now you can take a look at this. Now the tests are passing.

daniel-shimon (approved):
Yeah!

PR #4448, opened 2021-08-29
test: improve ptk_shell tests

jnoortheen:
## For community

### DIFF
--- a/tests/test_ptk_shell.py
+++ b/tests/test_ptk_shell.py
@@ -139,4 +139,4 @@ def test_ptk_prompt(line, exp, ptk_shell, capsys):
     shell.cmdloop()
     out, err = capsys.readouterr()
     # todo: check rendered output using https://pyte.readthedocs.io/
-    assert out.strip() == exp or line
+    assert out.strip() == (exp or line)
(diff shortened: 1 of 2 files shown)


tests/test_ptk_shell.py
```diff
@@ -126,10 +126,17 @@ def test_remove_ansi_osc(raw_prompt, prompt, osc_tokens):
...
     inp, out, shell = ptk_shell
-    text = "hello"
-    inp.send_text(f"{text}\n")  # note: terminate with '\n'
-    result = shell.singleline()
+    inp.send_text(f"{line}\nexit\n")  # note: terminate with '\n'
+    shell.cmdloop()
+    out, err = capsys.readouterr()
     # todo: check rendered output using https://pyte.readthedocs.io/
-    assert result == text
+    assert out.strip() == exp or line
```
daniel-shimon, line 142: Again, what's `line` doing here?

jnoortheen: when exp is not given it is expected to be same as input. 
like when giving strings

daniel-shimon: So you need to put parentheses around the `or` expression

jnoortheen: Done!

gforsyth (approved):
This is very slick @jnoortheen -- I desperately wanted something like this back in 2015, glad to see it coming together now!

PR #4449, opened 2021-08-30
chore: combine coverage reports from all platforms supported

jnoortheen:
also upgrade codecov-action version

* internal change: no news item

## For community

### DIFF
--- a/.github/workflows/pytest-windows-3.9.yml
+++ b/.github/workflows/pytest-windows-3.9.yml
@@ -1,4 +1,4 @@
-name: pytest windows 3.9
+name: pytest cov windows 3.9
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


.github/workflows/pytest-windows-3.9.yml
```diff
@@ -55,6 +55,11 @@ jobs:
         run: |
           python -m pip --version
           python -m pip install -r requirements/tests.txt
-          python -m pip install . --no-deps
+          python -m pip install -e . --no-deps
       - name: Run tests
-        run: python -m xonsh run-tests.xsh test -- --timeout=240
+        run: python -m xonsh run-tests.xsh test --report-coverage --no-amalgam -- --timeout=240
```
daniel-shimon, line 60: Why `--no-amalgam`? These tests need to run with the amalgamated version to prevent bugs like #4289

daniel-shimon (changes requested):

.github/workflows/pytest-windows-3.9.yml
jnoortheen: these measure coverage and need actual code.

jnoortheen: There are other tests running with amalgam

daniel-shimon: So now only 3.9 tests run without amalgamation?

jnoortheen: Yes. now only linux+3.9 is used for getting coverage

daniel-shimon (approved):
Cool! Thanks

PR #4461, opened 2021-09-07
Fix: EnvPath.add() not maintain uniqueness with Path objects on replace

### LINKED ISSUE #4366: EnvPath.add() doesn't maintain uniqueness with Path objects
## xonfig

<details>

```
| prompt toolkit   | None                 |
| xontrib 4        | linuxbrew            |
```

</details>

## Expected Behavior
`EnvPath.add(thing)` (e.g. `$PATH.add(thing)`) should not allow multiple copies of `thing` to enter the list, compared as strings.

## Current Behavior (and reproduction steps)
`EnvPath.add("str")` and `EnvPath.add(p"str")` do not detect each other's entries:

```
egnor@ostrich ~ $ $PATH
EnvPath(
['/usr/local/sbin',
 '/usr/local/bin',
 '/usr/sbin',
 '/usr/bin',
 '/sbin',
 '/bin',
 '/usr/games',
 '/usr/local/games',
 '/snap/bin']
)
egnor@ostrich ~ $ $PATH.add("/home/egnor/bin")
egnor@ostrich ~ $ $PATH.add(p"~/bin")
egnor@ostrich ~ $ $PATH
EnvPath(
['/usr/local/sbin',
 '/usr/local/bin',
 '/usr/sbin',
 '/usr/bin',
 '/sbin',
 '/bin',
 '/usr/games',
 '/usr/local/games',
 '/snap/bin',
 '/home/egnor/bin',
 '/home/egnor/bin']
)
```

This is because `EnvPath.add(data)` will put its data (of any type, `Path` or `str`) into the object directly, and `Path("foo") != "foo"`.

My first instinct was to have `EnvPath.add(data)` include `data = str(data)`, but then I realized that `EnvPath.__init__` was explicitly written to allow a mixed bag of `Path` and `str` objects, so I wasn't sure what was desired. `EnvPath` is documented as being a "list of strings" so I would expect it to stringify everything on entry, but maybe there's some concerns with that?

Note that `$PATH.add(p"~/bin")` seems like a very logical and idiomatic thing to put in one's `.xonshrc`, but due to this bug will end up with repeated copies when subshells occur.

## For community

meramsey:
Fixes #4366

Tested via docker python.
image

Original credits for the filter out duplicates to source: https://stackoverflow.com/a/25251306/1621381  
Seems to work great when testing.

## For community

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -928,9 +928,9 @@ def test_env_path_to_str(inp, exp):
     "left, right, exp",
     [
         (
-                EnvPath(["/home/wakka"]),
-                ["/home/jawaka"],
-                EnvPath(["/home/wakka", "/home/jawaka"]),
+            EnvPath(["/home/wakka"]),
+            ["/home/jawaka"],
+            EnvPath(["/home/wakka", "/home/jawaka"]),
         ),
         (["a"], EnvPath(["b"]), EnvPath(["a", "b"])),
         (EnvPath(["c"]), EnvPath(["d"]), EnvPath(["c", "d"])),
@@ -946,24 +946,27 @@ def test_env_path_add_replace_no_dupes_front_replace_existing():
     # Test replaces without dupes when added to front when adding existing entry
     path = EnvPath(["/home/wakka", "/home/wakka/bin"])
     path.add("/home/wakka/bin", front=True, replace=True)
-    exp = ["/home/wakka/bin", "/home/wakka"]
-    assert exp == path
+    assert path == ["/home/wakka/bin", "/home/wakka"]
 
 
 def test_env_path_add_replace_no_dupes_front_replace_multiple():
     # Test replaces without dupes when added to front when multiple existing occurrences
     path = EnvPath(["/home/wakka", "/home/wakka/bin", "/home/wakka/bin"])
     path.add("/home/wakka/bin", front=True, replace=True)
-    exp = ["/home/wakka/bin", "/home/wakka"]
-    assert exp == path
+    assert path == ["/home/wakka/bin", "/home/wakka"]
 
 
 def test_env_path_add_replace_no_dupes_back_replace_multiple():
     # Test replaces without dupes when not added to front
     path = EnvPath(["/home/wakka", "/home/wakka/bin", "/home/wakka/bin"])
     path.add("/home/wakka/bin", front=False, replace=True)
-    exp = ["/home/wakka", "/home/wakka/bin"]
-    assert exp == path
+    assert path == ["/home/wakka", "/home/wakka/bin"]
+
+
+def test_env_path_add_pathlib():
+    path = EnvPath(["/home/wakka", "/home/wakka/bin", "/home/wakka/bin"])
+    path.add(pathlib.Path("/home/wakka/bin"), front=False, replace=True)
+    assert path == ["/home/wakka", "/home/wakka/bin"]
 
 
 # helper
@@ -996,12 +999,12 @@ def test_env_path_getitem(inp, exp, xession, env):
     "inp, exp",
     [
         (
-                os.pathsep.join(["xonsh_dir", "../", ".", "~/"]),
-                ["xonsh_dir", "../", ".", "~/"],
+            os.pathsep.join(["xonsh_dir", "../", ".", "~/"]),
+            ["xonsh_dir", "../", ".", "~/"],
         ),
         (
-                "/home/wakka" + os.pathsep + "/home/jakka" + os.pathsep + "~/",
-                ["/home/wakka", "/home/jakka", "~/"],
+            "/home/wakka" + os.pathsep + "/home/jakka" + os.pathsep + "~/",
+            ["/home/wakka", "/home/jakka", "~/"],
         ),
     ],
 )
@@ -1023,8 +1026,8 @@ def test_env_path_multipath(inp, exp, xession, env):
         (pathlib.Path("~/"), ["~"]),
         (pathlib.Path("."), ["."]),
         (
-                ["/home/wakka", pathlib.Path("/home/jakka"), "~/"],
-                ["/home/wakka", "/home/jakka".replace("/", os.sep), "~/"],
+            ["/home/wakka", pathlib.Path("/home/jakka"), "~/"],
+            ["/home/wakka", "/home/jakka".replace("/", os.sep), "~/"],
         ),
         (["/home/wakka", pathlib.Path("../"), "../"], ["/home/wakka", "..", "../"]),
         (["/home/wakka", pathlib.Path("~/"), "~/"], ["/home/wakka", "~", "~/"]),
@@ -1052,8 +1055,8 @@ def mkpath(*paths):
     "inp, exp",
     [
         (
-                [mkpath("home", "wakka"), mkpath("home", "jakka"), mkpath("home", "yakka")],
-                [mkpath("home", "wakka"), mkpath("home", "jakka")],
+            [mkpath("home", "wakka"), mkpath("home", "jakka"), mkpath("home", "yakka")],
+            [mkpath("home", "wakka"), mkpath("home", "jakka")],
         )
     ],
 )
@@ -1066,8 +1069,8 @@ def test_env_path_slice_get_all_except_last_element(inp, exp):
     "inp, exp",
     [
         (
-                [mkpath("home", "wakka"), mkpath("home", "jakka"), mkpath("home", "yakka")],
-                [mkpath("home", "jakka"), mkpath("home", "yakka")],
+            [mkpath("home", "wakka"), mkpath("home", "jakka"), mkpath("home", "yakka")],
+            [mkpath("home", "jakka"), mkpath("home", "yakka")],
         )
     ],
 )
@@ -1080,14 +1083,14 @@ def test_env_path_slice_get_all_except_first_element(inp, exp):
     "inp, exp_a, exp_b",
     [
         (
-                [
-                    mkpath("home", "wakka"),
-                    mkpath("home", "jakka"),
-                    mkpath("home", "yakka"),
-                    mkpath("home", "takka"),
-                ],
-                [mkpath("home", "wakka"), mkpath("home", "yakka")],
-                [mkpath("home", "jakka"), mkpath("home", "takka")],
+            [
+                mkpath("home", "wakka"),
+                mkpath("home", "jakka"),
+                mkpath("home", "yakka"),
+                mkpath("home", "takka"),
+            ],
+            [mkpath("home", "wakka"), mkpath("home", "yakka")],
+            [mkpath("home", "jakka"), mkpath("home", "takka")],
         )
     ],
 )
@@ -1102,14 +1105,14 @@ def test_env_path_slice_path_with_step(inp, exp_a, exp_b):
     "inp, exp",
     [
         (
-                [
-                    mkpath("home", "wakka"),
-                    mkpath("home", "xakka"),
-                    mkpath("other", "zakka"),
-                    mkpath("another", "akka"),
-                    mkpath("home", "bakka"),
-                ],
-                [mkpath("other", "zakka"), mkpath("another", "akka")],
+            [
+                mkpath("home", "wakka"),
+                mkpath("home", "xakka"),
+                mkpath("other", "zakka"),
+                mkpath("another", "akka"),
+                mkpath("home", "bakka"),
+            ],
+            [mkpath("other", "zakka"), mkpath("another", "akka")],
         )
     ],
 )
@@ -1256,8 +1259,8 @@ def test_ensure_slice(inp, exp):
     [
         ((range(50), slice(25, 40)), list(i for i in range(25, 40))),
         (
-                ([1, 2, 3, 4, 5, 6, 7, 8, 9, 10], [slice(1, 4), slice(6, None)]),
-                [2, 3, 4, 7, 8, 9, 10],
+            ([1, 2, 3, 4, 5, 6, 7, 8, 9, 10], [slice(1, 4), slice(6, None)]),
+            [2, 3, 4, 7, 8, 9, 10],
         ),
         (([1, 2, 3, 4, 5], [slice(-2, None), slice(-5, -3)]), [4, 5, 1, 2]),
     ],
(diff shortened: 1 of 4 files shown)


daniel-shimon:
Hi @meramsey!
Glad to have you contributing to the project!
Could you add some tests to make sure this works as expected?
I'd start looking in `tests/test_environ.py` for examples

meramsey:
> Hi @meramsey!
> Glad to have you contributing to the project!
> Could you add some tests to make sure this works as expected?
> I'd start looking in `tests/test_environ.py` for examples

Yeah ran out of time last night, but will look into that file and add some appropriate tests when I get a sec.

At first glance though should I be doing the tests in  `tests/test_tools.py` instead since that is where function i modified exists? or did you had another thought for adding/referencing them in the `tests/test_environ.py`?

Thanks for the fast reply and review

daniel-shimon:
> At first glance though should I be doing the tests in tests/test_tools.py instead since that is where function i modified exists? or did you had another thought for adding/referencing them in the tests/test_environ.py?

I've seen `test_environ.py ` already has some tests for environment path variables, but if `test_tools` has tests for its logic then that can be a good place too!
I don't think it's too important as long as the tests you're adding are next to pre-existing tests about the same piece of code.

meramsey:
> > At first glance though should I be doing the tests in tests/test_tools.py instead since that is where function i modified exists? or did you had another thought for adding/referencing them in the tests/test_environ.py?
> 
> I've seen `test_environ.py ` already has some tests for environment path variables, but if `test_tools` has tests for its logic then that can be a good place too!
> I don't think it's too important as long as the tests you're adding are next to pre-existing tests about the same piece of code.

I'm having trouble finding any examples of stuff that are actually using EnvPath or EnvPath.add to reference.  Not written much in this style of so having a hard time finding a good reference of anything that is dealing with these directly.

The closest thing that looks related doesn't instantiate anything it just merges two parameterized lists. Not sure if this test here has much value unless im misunderstanding how it works?
https://github.com/xonsh/xonsh/blob/main/tests/test_tools.py#L927-L942

I was thinking I could do something like the below to instantiate an environment with paths from input but no idea how to call that function and it update an object's properties.
```
@pytest.mark.parametrize("env", [TOOLS_ENV])
@pytest.mark.parametrize(
    "inp, exp",
    [
        (["/home/wakka"], ["/home/wakka/bin"], ["/home/wakka/bin"]),
        (["/home/wakka"], ["/home/wakka/bin"]),
    ],
)
def test_env_path_add_replace_no_dupes(inp, exp, env):
    EnvPath.add(env, inp, front=True, replace=True)
    assert env == exp
```

When testing the function separately it works fine from what i can tell for all the cases expected
```
self = ["/home/wakka", "/home/wakka/bin"]
expected = ["/home/wakka/bin", "/home/wakka"]


def add(data, front=False, replace=False):
    global self
    """Add a value to this EnvPath,

        path.add(data, front=bool, replace=bool) -> ensures that path contains data, with position determined by kwargs

        Parameters
        ----------
        data : string or bytes or pathlib.Path
            value to be added
        front : bool
            whether the value should be added to the front, will be
            ignored if the data already exists in this EnvPath and
            replace is False
            Default : False
        replace : bool
            If True, the value will be removed and added to the
            start or end(depending on the value of front)
            Default : False

        Returns
        -------
        None

        """
    if data not in self:
        self.insert(0 if front else len(self), data)
    elif replace:
        # https://stackoverflow.com/a/25251306/1621381
        self = list(filter(lambda x: x != data, self))
        self.insert(0 if front else len(self), data)

# Test replaces without dupes when added to front when adding existing entry
add("/home/wakka/bin", front=True, replace=True)
assert self == expected

# Test replaces without dupes when added to front when multiple existing occurrences
self = ["/home/wakka", "/home/wakka/bin", "/home/wakka/bin", "/home/wakka/bin"]
add("/home/wakka/bin", front=True, replace=True)
print(self)
print(expected)
assert self == expected

# Test replaces without dupes when not added to front
self = ["/home/wakka", "/home/wakka/bin", "/home/wakka/bin", "/home/wakka/bin"]
expected = ["/home/wakka", "/home/wakka/bin"]
add("/home/wakka/bin", front=False, replace=True)
assert self == expected
```


Class that this method is under if that helps
https://github.com/xonsh/xonsh/blob/main/xonsh/tools.py#L166-L287

It looks like no tests failed with the new code but I agree it would be great to have the functionality properly tested to prevent regressions in future just not sure how to write them in this case for that section of code. If you have any pointers that would be helpful. I'm used to mocking objects/classes and stuff in other languages, but not done this in python before.

daniel-shimon:
> I'm having trouble finding any examples of stuff that are actually using EnvPath or EnvPath.add to reference

You can add new tests that create a `EnvPath` object :)

> Test replaces without dupes when added to front when adding existing entry
> Test replaces without dupes when added to front when multiple existing occurrences
> Test replaces without dupes when not added to front

These cases seem to cover it, you can add a test for each (doesn't have to be using parametrization).
Just make sure to test them with a `Path` object!

meramsey:
> > I'm having trouble finding any examples of stuff that are actually using EnvPath or EnvPath.add to reference
> 
> You can add new tests that create a `EnvPath` object :)
> 
> > Test replaces without dupes when added to front when adding existing entry
> > Test replaces without dupes when added to front when multiple existing occurrences
> > Test replaces without dupes when not added to front
> 
> These cases seem to cover it, you can add a test for each (doesn't have to be using parametrization).
> Just make sure to test them with a `Path` object!

I added some tests let me know what you think.

tests/test_tools.py
```diff
@@ -942,6 +942,30 @@ def test_env_path_add(left, right, exp):
     assert exp == obs
 
 
+def test_env_path_add_replace_no_dupes_front_replace_existing():
+    # Test replaces without dupes when added to front when adding existing entry
+    path = EnvPath(["/home/wakka", "/home/wakka/bin"])
+    path.add("/home/wakka/bin", front=True, replace=True)
+    exp = ["/home/wakka/bin", "/home/wakka"]
+    assert exp == path
```
daniel-shimon, line 950: Inlining `exp` will make the tests clearer:
```python
assert path == ["/home/wakka/bin", "/home/wakka"]
```

tests/test_tools.py
```diff
@@ -972,12 +996,12 @@ def test_env_path_getitem(inp, exp, xession, env):
...
-            os.pathsep.join(["xonsh_dir", "../", ".", "~/"]),
-            ["xonsh_dir", "../", ".", "~/"],
+                os.pathsep.join(["xonsh_dir", "../", ".", "~/"]),
+                ["xonsh_dir", "../", ".", "~/"],
         ),
         (
-            "/home/wakka" + os.pathsep + "/home/jakka" + os.pathsep + "~/",
-            ["/home/wakka", "/home/jakka", "~/"],
+                "/home/wakka" + os.pathsep + "/home/jakka" + os.pathsep + "~/",
+                ["/home/wakka", "/home/jakka", "~/"],
```
daniel-shimon, line 1004: I see you've done a lot of reformatting - we use `black` for our linting.
Re-run `black` on these files so the linting will go back to normal

tests/test_tools.py
```diff
@@ -942,6 +942,30 @@ def test_env_path_add(left, right, exp):
     assert exp == obs
 
 
+def test_env_path_add_replace_no_dupes_front_replace_existing():
+    # Test replaces without dupes when added to front when adding existing entry
+    path = EnvPath(["/home/wakka", "/home/wakka/bin"])
+    path.add("/home/wakka/bin", front=True, replace=True)
```
daniel-shimon, line 948: Add a test with `pathlib.Path("/home/wakka/bin")` to make sure that works correctly (which is the original issue)

daniel-shimon (changes requested):
Looks good! A few fixes and we're done :)

tests/test_tools.py
```diff
@@ -946,24 +946,27 @@ def test_env_path_add_replace_no_dupes_front_replace_existing():
...
     path.add("/home/wakka/bin", front=False, replace=True)
-    exp = ["/home/wakka", "/home/wakka/bin"]
-    assert exp == path
+    assert path == ["/home/wakka", "/home/wakka/bin"]
+
+
+def test_env_path_add_pathlib():
+    path = EnvPath(["/home/wakka", "/home/wakka/bin", "/home/wakka/bin"])
+    path.add(pathlib.Path("/home/wakka/bin"), front=False, replace=True)
+    assert path == ["/home/wakka", "/home/wakka/bin"]
```
meramsey, line 969: "Add a test with pathlib.Path("/home/wakka/bin") to make sure that works correctly (which is the original issue)"
Added test here as requested.

xonsh/tools.py
```diff
@@ -280,6 +280,7 @@ def add(self, data, front=False, replace=False):
         None
 
         """
+        data = str(expand_path(data))
```
meramsey, line 283: Added this to account for the pathlib stuff so its always expanded and converted to a string before the comparison

meramsey:
I reran black twice and it says the remaining whitespace changes were due to black best i can tell. 
```
@ black tests/test_tools.py
All done! ✨ 🍰 ✨
1 file left unchanged.
 v3.9.7 (.venv) ~/PycharmProjects/xonsh on  fixduplicatepaths [ 2●] at 08:31:21
 
@ git add tests/test_tools.py
 v3.9.7 (.venv) ~/PycharmProjects/xonsh on  fixduplicatepaths [ 2●] at 08:31:34
 
@ git commit -m 'Add test for EnvPath.add() function with pathlib path'
black....................................................................Passed
Trim Trailing Whitespace.................................................Passed
Check for case conflicts.................................................Passed
Check for merge conflicts................................................Passed
Check Yaml...........................................(no files to check)Skipped
Check Toml...........................................(no files to check)Skipped
Debug Statements (Python)................................................Passed
Check for added large files..............................................Passed
[fixduplicatepaths deccfeeb] Add test for EnvPath.add() function with pathlib path
 2 files changed, 83 insertions(+), 79 deletions(-)
 v3.9.7 (.venv) ~/PycharmProjects/xonsh on  fixduplicatepaths [ ⇡1]  took 5s  at 08:31:44
 
@ git push
Enumerating objects: 11, done.
Counting objects: 100% (11/11), done.
Delta compression using up to 16 threads
Compressing objects: 100% (6/6), done.
Writing objects: 100% (6/6), 1.67 KiB | 1.67 MiB/s, done.
Total 6 (delta 5), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (5/5), completed with 5 local objects.
To github.com:meramsey/xonsh.git
 v3.9.7 (.venv) ~/PycharmProjects/xonsh on  fixduplicatepaths at 08:31:56
 
@ black tests/test_tools.py --check
All done! ✨ 🍰 ✨
1 file would be left unchanged.
 v3.9.7 (.venv) ~/PycharmProjects/xonsh on  fixduplicatepaths at 08:40:38
 
@

```

 Thanks also for the reminder about pathlib stuff i totally spaced on that part.

Should be all set i think now. :)

gforsyth:
Hey @meramsey -- thanks for putting this in!  The test failures are because of how Windows uses different path separators than *nix:

```
    def test_env_path_add_pathlib():
        path = EnvPath(["/home/wakka", "/home/wakka/bin", "/home/wakka/bin"])
        path.add(pathlib.Path("/home/wakka/bin"), front=False, replace=True)
>       assert path == ["/home/wakka", "/home/wakka/bin"]
E       AssertionError: assert ['/home/wakka...\\wakka\\bin'] == ['/home/wakka...me/wakka/bin']
E         Left contains 2 more items, first extra item: '/home/wakka/bin'
E         Use -v to get the full diff
```

I would suggest either using `os.sep` to get the appropriate OS-specific separator character, or build up those "correct" paths using Pathlib itself

meramsey:
> Hey @meramsey -- thanks for putting this in! The test failures are because of how Windows uses different path separators than *nix:
> 
> ```
>     def test_env_path_add_pathlib():
>         path = EnvPath(["/home/wakka", "/home/wakka/bin", "/home/wakka/bin"])
>         path.add(pathlib.Path("/home/wakka/bin"), front=False, replace=True)
> >       assert path == ["/home/wakka", "/home/wakka/bin"]
> E       AssertionError: assert ['/home/wakka...\\wakka\\bin'] == ['/home/wakka...me/wakka/bin']
> E         Left contains 2 more items, first extra item: '/home/wakka/bin'
> E         Use -v to get the full diff
> ```
> 
> I would suggest either using `os.sep` to get the appropriate OS-specific separator character, or build up those "correct" paths using Pathlib itself

Is this for all tests of just that one?  Cause it looks like other existing tests passed with some of the setup so was unsure.

> Hey @meramsey -- thanks for putting this in! The test failures are because of how Windows uses different path separators than *nix:
> 
> ```
>     def test_env_path_add_pathlib():
>         path = EnvPath(["/home/wakka", "/home/wakka/bin", "/home/wakka/bin"])
>         path.add(pathlib.Path("/home/wakka/bin"), front=False, replace=True)
> >       assert path == ["/home/wakka", "/home/wakka/bin"]
> E       AssertionError: assert ['/home/wakka...\\wakka\\bin'] == ['/home/wakka...me/wakka/bin']
> E         Left contains 2 more items, first extra item: '/home/wakka/bin'
> E         Use -v to get the full diff
> ```
> 
> I would suggest either using `os.sep` to get the appropriate OS-specific separator character, or build up those "correct" paths using Pathlib itself

Refactored them all to use os independent paths and ran tests successfully locally in windows vm to confirm prior.  Should now work. :crossed_fingers:

meramsey:
@daniel-shimon did you have a chance to review this? I see all tests pass now. I did also put in a few minor fixes to xontribs meta since as well.

daniel-shimon (approved):
Hi @meramsey the fixes are good!
Thanks for putting this in!

xonsh/xontribs_meta.py
```diff
@@ -675,7 +675,7 @@ def define_xontribs():
         ),
         "z": Xontrib(
             url="https://github.com/AstraLuma/xontrib-z",
-            description="Tracks your most used directories, based on 'frecency'.",
+            description="Tracks your most used directories, based on 'frequency'.",
```
gforsyth, line 678:
```suggestion
            description="Tracks your most used directories, based on 'frecency'.",
```
Looks like a typo but it isn't -- it's a portmanteau of "frequency" and "recency"

gforsyth (changes requested):
Code and tests look good @meramsey !  Just one change in the text that needs to be reverted and then this is good to go.

gforsyth (approved):
Looks great!  Thanks @meramsey and welcome to xonsh!

Just waiting on CI to finish and this is good to go

PR #4464, opened 2021-09-08
Lab/lexer improv

jnoortheen:
I don't see any time speedup on my machine. But these are some obvious improvements to lexing.

## For community

### DIFF
--- a/xonsh/lexer.py
+++ b/xonsh/lexer.py
@@ -33,7 +33,7 @@ from xonsh.tokenize import (
     GREATER,
     LESS,
     RIGHTSHIFT,
-    tokenize,
+    _tokenize,
     TokenError,
     HAS_WALRUS,
 )
@@ -370,11 +370,12 @@ def get_tokens(s, tolerant):
     Given a string containing xonsh code, generates a stream of relevant PLY
     tokens using ``handle_token``.
     """
+    input_cp = io.BytesIO(s.encode("utf-8"))
     state = {
         "indents": [0],
         "last": None,
         "pymode": [(True, "", "", (0, 0))],
-        "stream": tokenize(io.BytesIO(s.encode("utf-8")).readline, tolerant),
+        "stream": _tokenize(input_cp.readline, "utf-8", tolerant),
         "tolerant": tolerant,
     }
     while True:
--- a/xonsh/tokenize.py
+++ b/xonsh/tokenize.py
@@ -13,7 +13,7 @@ Original file credits:
                   'Skip Montanaro, Raymond Hettinger, Trent Nelson, '
                   'Michael Foord')
 """
-
+import functools
 import re
 import io
 import sys
@@ -375,6 +375,7 @@ PseudoToken = Whitespace + group(
 )
 
 
+@functools.lru_cache(maxsize=None)
 def _compile(expr):
     return re.compile(expr, re.UNICODE)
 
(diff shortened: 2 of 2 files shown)


xonsh/lexer.py
```diff
@@ -33,12 +32,27 @@
...
+            self.type,
+            self.value,
+            self.lineno,
+            self.lexpos,
+        )
+
+    def __repr__(self):
+        return str(self)
+
+
```
gforsyth, line 55: I would prefer we use the `LexToken` from `ply`, if only to keep it obvious which of the parts are part of `ply` and which are custom to `xonsh`

xonsh/lexer.py
```diff
@@ -370,11 +384,13 @@ def get_tokens(s, tolerant):
     Given a string containing xonsh code, generates a stream of relevant PLY
     tokens using ``handle_token``.
     """
+    input_cp = io.BytesIO(s.encode("utf-8"))
+    del s
```
gforsyth, line 388: Is there any reason to do this beyond a _very_ tiny memory savings?

jnoortheen: Yes that is the reasin. I was testing with some huge file. Do you want to remove ?

xonsh/lexer.py
jnoortheen: the idea is to not import lex. That will be saving some import cost.

xonsh/lexer.py
gforsyth: It should get cleaned up shortly afterwards by garbage collection, no?

xonsh/lexer.py
gforsyth: If there's a measurable difference then we can make the change.

xonsh/lexer.py
jnoortheen: I wasn't sure about that. It yields the dict and might hold reference to the locals. anyway, this is small size normally I will remove it. Let me know, if you want me to remove the LexToken class?

xonsh/lexer.py
jnoortheen: In my machine I get these numbers 

Screenshot_20210909_202910

gforsyth: I'm not sure that's a completely accurate representation of what the difference might be -- especially since we're importing other parts of `ply` elsewhere in the codebase, so those imports will be cached.
The difference should be measurable in startup time

jnoortheen: with the 
image

, it prints `Import cost ply.lex:  0.0027875900268554688`

gforsyth: Hey @jnoortheen -- sorry I wasn't clearer -- I'm interested in `xonsh` general startup time, so running `time -p xonsh --no-rc -c "echo HI"` a number of times on `main` vs. this branch and determining if there's any noticeable improvement.

jnoortheen: That will not be same all the time. That is the actual cost.

gforsyth: That is the cost assuming that we haven't imported any other part of `ply`, which I doubt.  

I don't trust benchmarks that exist in a vacuum.

jnoortheen: as you can see the numbers are small and doesn't make much difference. 
xexec

importing `ply.yacc` will have minimum effect on importing `ply.lex` after. Since they are separate modules. The interpreter has to parse the file and built the tree right ?

gforsyth: Based on those 4 measurements, it has no appreciable effect at all, so I'm not sure about the utility of breaking out one class from a vended subpackage

jnoortheen: OK understandable. I will revert this.

PR #4476, opened 2021-09-18
feat: improve pip completer

jnoortheen:
## For community

### DIFF
--- a/tests/completers/test_pip_completer.py
+++ b/tests/completers/test_pip_completer.py
@@ -1,9 +1,7 @@
-import os
-
 import pytest
 
-from xonsh.completers.pip import PIP_RE
 from tests.tools import ON_WINDOWS
+from xonsh.completers.pip import PIP_RE
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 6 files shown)


jnoortheen:
@daniel-shimon can you check the failing tests? they are related to the `alias` changes. Seems like the threaded part is not waited.

tests/completers/test_pip_completer.py
```diff
@@ -1,16 +1,22 @@
+import os
```
gforsyth, line 1: Is this being used?

gforsyth: And if it isn't, what happened to the flake8 check?

gforsyth (commented):
Thanks for this @jnoortheen -- overall looks good, one nit

tests/completers/test_pip_completer.py
jnoortheen: Imports inside tests are not properly checked . I will check it.

jnoortheen: image

tests are excluded from checks. I will try to enable flake8 for them in another PR. I will remove os import in this.

gforsyth:
Thanks @jnoortheen !

PR #4477, opened 2021-09-18
fix: not enough space for completions

jnoortheen:
fixes https://github.com/xonsh/xonsh/issues/3775

I think it was set to 0 because there is a `reserve_space` method in `PTKCompleter` class. Anyhow it doesn't seem to be working. This change fixes the issue with completions not having enough space

## For community

### DIFF
--- a/news/fix-reserve-menu.rst
+++ b/news/fix-reserve-menu.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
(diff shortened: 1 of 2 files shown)


gforsyth (changes requested):
Thanks for putting this in @jnoortheen!

Two things:
1. please add a news entry for this
2. I've tested this locally and it does create space for competions, but rather oddly, it creates `n-1` lines of completions where `n=$COMPLETIONS_MENU_ROWS`.

Bit of a hack, but I suggest we just add a `+1` in the argument to `reserve_space_for_menu`

jnoortheen:
@gforsyth done the changes

gforsyth (approved):
Thanks @jnoortheen !

PR #4486, opened 2021-09-26
Ap/aliases completer

jnoortheen:
This adds generic attribute check. So no need for instance checks https://github.com/xonsh/xonsh/pull/4267#discussion_r676066853

## For community


xonsh/cli_utils.py
```diff
@@ -371,6 +373,149 @@ def dispatch(parser: ap.ArgumentParser, args=None, lenient=False, **ns):
...
+
+    def __init__(self, parser: ap.ArgumentParser, command, **kwargs):
+        args = tuple(c.value for c in command.args[: command.arg_index])
+
+        self.parser, self.remaining_args = self.get_parser(parser, args[1:])
+
+        self.command = command
+        kwargs["command"] = command
+        self.kwargs = kwargs
+        """will be sent to completer function"""
```
gforsyth, line 387: should this line be here or did it get cut from somewhere else?

gforsyth (commented):
Looks good @jnoortheen -- just one comment on a line that looks like it belongs elsewhere

xonsh/cli_utils.py
jnoortheen: the comment is correct. The kwargs will be used to call the completer functions

gforsyth: Ok, can you make it a comment and not a string then, please?

jnoortheen: done.

gforsyth (approved):

gforsyth:
Thanks for this, @jnoortheen and sorry for the delay in getting it in!

PR #4493, opened 2021-09-29
feat: hide select branch names in gitstatus

t184256:
I'm not particularly fond of `main` or `master` being constantly displayed when I'm working on the main branch, but I want working on a feature branch to be immediately noticeable. This is a proposal  to hide a configurable list of branch names. Pretty opinionated, I know, and I don't particularly insist on merging it, but I felt like I should share it since I've hacked it together.

## For community

### DIFF
--- a/news/gitstatus-hide-branch.rst
+++ b/news/gitstatus-hide-branch.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Branch names listed in ``XONSH_GITSTATUS_BRANCHES_HIDDEN`` will be hidden from gitstatus output.
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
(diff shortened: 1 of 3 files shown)


gforsyth:
Thanks for putting this in @t184256 !  I definitely understand your point about displaying certain information too much.
I'll think on this.   I don't have an issue with this particular change, but it's also easy to imagine that there's a follow-on set of changes to further tweak how `gitstatus` is displayed and we end up with an (even bigger than already) explosion of very specific configuration flags.
Might be that we want to have a more general system for fine-tuned configuration of those fields.

jnoortheen:
Closing this as this will be a more personal choice and one can customize the output of the fields if https://github.com/xonsh/xonsh/pull/4697 is merged .

For this branch hiding, we can do something like

```py
def branch_name_updator(fld, ctx):
    info = ctx.pick(gs_porcelain).value
    branch = info["branch"]
    if branch not in SOME_LIST:
        fld.value = branch

env["PROMPT_FIELDS"]["gs_branch"].updator = branch_name_updator
```

PR #4497, opened 2021-10-02
perf: cache suggested commands

P403n1x87:
Calls to `suggest_commands` would result in repeated calls to the Levenshtein distance routine, which whilst optimally implemented is still expensive. The change proposes to cache the result instead.

Profiling data before:

test_casting_before

and after

test_casting_after

This was obtained by running the `test_casting` from the test suite. With caching the expensive calls to `levenshtein` disappear.

## For community


gforsyth:
Changes look good to me @P403n1x87 !  Could you run `black` to autoformat and also add a news entry?  Thanks!

P403n1x87:
> Changes look good to me @P403n1x87 ! Could you run `black` to autoformat and also add a news entry? Thanks!

Done.

gforsyth:
Thanks @P403n1x87 !

PR #4498, opened 2021-10-02
perf(readers): use mutable buffers

P403n1x87:
This change uses the mutable `bytearray` to append bytes from a queue for better performance.

Before:

readline_before

After:

readline_after

The profiling data shows a reduction in the `readline` own time due to the fact that the byte string is no longer copied over at each concatenation.

## For community


xonsh/procs/pipelines.py
```diff
@@ -425,7 +424,7 @@ def _decode_uninew(self, b, universal_newlines=None):
         """Decode bytes into a str and apply universal newlines as needed."""
         if not b:
             return ""
-        if isinstance(b, bytes):
+        if isinstance(b, (bytes, bytearray)):
```
P403n1x87, line 427: Not sure whether `bytes` here is still needed

xonsh/procs/readers.py
```diff
@@ -54,31 +54,27 @@ def read_queue(self):
 
     def read(self, size=-1):
         """Reads bytes from the file."""
-        i = 0
-        buf = b""
-        while size < 0 or i != size:
+        buf = bytearray()
+        while size < 0 or len(buf) != size:
```
P403n1x87, line 58: I wonder if this shouldn't be `len(buf) <= size` instead and have the method return a truncated value.

xonsh/procs/readers.py
```diff
@@ -54,31 +54,27 @@ def read_queue(self):
...
         return buf
 
     def readline(self, size=-1):
         """Reads a line, or a partial line from the file descriptor."""
-        i = 0
         nl = b"\n"
-        buf = b""
-        while size < 0 or i != size:
+        buf = bytearray()
+        while size < 0 or len(buf) != size:
```
P403n1x87, line 70: I wonder if this shouldn't be `len(buf) <= size` instead and have the method return a truncated value.

P403n1x87 (commented):
I fixed the cause of the test failures and left a couple of questions, which might all be silly.

xonsh/procs/pipelines.py
```diff
@@ -1,20 +1,19 @@
 """Command pipeline tools."""
+import io
```
P403n1x87, line 2: I'll revert these formatting changes

gforsyth:
Thanks for putting this in @P403n1x87 !  I'll test it out locally and compare performance but the profiling you posted looks promising!

Could you add a news entry to the PR?  https://xon.sh/devguide.html#changelog

P403n1x87:
> Could you add a news entry to the PR? https://xon.sh/devguide.html#changelog

Done.

Issue #4450, opened 2021-08-30
`curr_branch` gets nothing on xonsh


glyh:
## xonfig

<details>

```
$ xonfig
+------------------+-----------------+
| xonsh            | 0.10.0          |
| Python           | 3.9.6           |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | 3.0.19          |
| shell type       | prompt_toolkit  |
| history backend  | json            |
| pygments         | 2.10.0          |
| on posix         | True            |
| on linux         | True            |
| distro           | arch            |
| on wsl           | False           |
| on darwin        | False           |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
| on jupyter       | False           |
| jupyter kernel   | None            |
| xontrib 1        | coreutils       |
| xontrib 2        | onepath         |
| xontrib 3        | z               |
+------------------+-----------------+
```

</details>

## Expected Behavior

The `curr_branch` shows branch in prompt

## Current Behavior

Shows nothing. 

## Steps to Reproduce

Here's my xonshrc:

<details>

```python
from re import finditer
from os import getcwd, path, chdir
from xonsh.tools import EnvPath

xontrib load coreutils onepath z
$VI_MODE = True
$PROMPT = '{BLUE}{cwd} {curr_branch}\n{PURPLE}❯ '

$PATH += ['~/.local/bin',
          '~/.yarn/bin',
	  '~/usr/lib/jvm/default/bin']
$CLASSPATH = ['.', '/usr/bin/share/java/antlr-complete.jar']
$VISUAL = 'nvim'
$EDITOR = 'nvim'
$PAGER = 'page'

for m in finditer(r"export ([A-Z_]+)='([^']*)'", $(luarocks path)):
    env_var = m.group(1)
    if env_var == 'PATH':
        ${env_var} = EnvPath(m.group(2))
    else:
        ${env_var} = m.group(2)

def edit_neovim_config():
    wd = getcwd()
    chdir(path.expanduser('~/.config/nvim'))
    nvim init.lua
    chdir(wd)

aliases['e'] = 'nvim'
aliases['en'] = 'nvim -u NONE'
aliases['er'] = 'sudoedit'
aliases['ec'] = edit_neovim_config
aliases['g'] = 'git'
aliases['l'] = aliases['ls']
[...]

PR #4506, opened 2021-10-10
use git branch --show-current to get current branch

sousajf1:
this probably fixes https://github.com/xonsh/xonsh/issues/4450

### DIFF
--- a/news/git-prompt-fix-curr-branch.rst
+++ b/news/git-prompt-fix-curr-branch.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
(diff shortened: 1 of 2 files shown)


sousajf1:
@anki-code can you please review?

jnoortheen:
@sousajf1 I suggest using the same approach as [here](https://github.com/oh-my-fish/oh-my-fish/blob/master/lib/git/git_branch_name.fish)

jnoortheen:
Thanks @sousajf1 for trying to tackle. I have made a PR with similar approach to what fish does here - https://github.com/xonsh/xonsh/pull/4607

PR #4520, opened 2021-10-15
adding convention for short boolean flags

jnoortheen:
Not adding news item. This is just an internal refactoring. This will reduce the boilerplate code required to make an argument a flag

## For community

### DIFF
--- a/tests/test_cli_utils.py
+++ b/tests/test_cli_utils.py
@@ -1,10 +1,9 @@
 """Test module xonsh/cli_utils.py"""
-import pytest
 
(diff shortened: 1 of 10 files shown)


gforsyth (approved):
I'm not sure I'm a fan of the mismatched signatures now, where only non-booleans will have their type hints in the function signature, but I recognize the utility of avoiding all the boilerplate. 

We may have to revisit this if it becomes too confusing for other contributors.

gforsyth:
Thanks @jnoortheen !

PR #4521, opened 2021-10-16
Xompletions

jnoortheen:
new completions will be added under `xompletions` package in future PRs.

## For community


gforsyth:
I'm inclined to defer to @daniel-shimon if he's around.

Any concerns around splitting the completion machinery from the completion definitions themselves?

jnoortheen:
@gforsyth can we check this ? the reason is so that large number of completions will not slow the shell startup. Also we can put many completions in python itself and don't have to depend on bash or external shells. e.g. bash is not uptodate on OSX and all brew installs use zsh completion files

xompletions/django-admin.py
```diff
@@ -0,0 +1,22 @@
...
+
+    # https://github.com/django/django/blob/main/extras/django_bash_completion
+    return bash_complete_completer(ctx, DJANGO_AUTO_COMPLETE="1")
+
+
+if __name__ == "__main__":
+    # todo:
+    #   1. add test script.
+    #   2. add python completer and wrap this as `python manage.py`
+    pass
```
gforsyth, line 22: This seems unnecessary -- if it was copied from somewhere, we should provide attribution

xonsh/completers/tools.py
```diff
@@ -201,3 +203,35 @@ def apply_lprefix(comps, lprefix):
                 yield comp
         else:
             yield RichCompletion(comp, prefix_len=lprefix)
+
+
+def bash_complete_completer(ctx: CommandContext, **env: str):
```
gforsyth, line 208: this needs a different name to better distinguish from the _other_ bash completer.

gforsyth, line 208: should this be decorated as a `contextual` completer?

gforsyth (changes requested):

xompletions/django-admin.py
jnoortheen: that is a reminder for myself. I will remove them

xonsh/completers/tools.py
jnoortheen: I should not be decorated. it is just a helper function and not intended to be called directly as a completer.

gforsyth (approved):
Thanks for this, @jnoortheen !  And yeah, I can definitely see the benefit if we add an env-var to point at a directory to load extra custom `*.xsh` or `*.py` completions from

PR #4531, opened 2021-11-02
Fix: don't display welcome if $XONSHRC_DIR member is found

agoose77:
Currently, even if there is an `$XONSHRC_DIR` in use, xonsh shows the welcome banner. I prefer to use an `rc.d` approach for configuring my system, so I'd rather not create an empty `.xonshrc` just to disable the banner!

I don't know whether we also want to check that these directories are non-empty. I think probably not - simply creating this special directory is enough to suggest that the user knows what they're doing.

### DIFF
--- a/news/fix-xonshrc-dir-welcome.rst
+++ b/news/fix-xonshrc-dir-welcome.rst
@@ -1,0 +1,14 @@
+**Added:**
+
+**Changed:**
+
+**Deprecated:**
+
+**Removed:**
+
+**Fixed:**
+
+*  Do not show welcome message when any `$XONSHRC_DIR` directory entry exists.
+
+**Security:**
+
--- a/news/fix-xonshrc-dir-welcome.rst
+++ b/news/fix-xonshrc-dir-welcome.rst
@@ -8,7 +8,7 @@
 
 **Fixed:**
 
-*  Do not show welcome message when any `$XONSHRC_DIR` directory entry exists.
+*  Do not show welcome message when any ``$XONSHRC_DIR`` directory entry exists.
 
 **Security:**
 
--- a/news/fix-xonshrc-dir-welcome.rst
+++ b/news/fix-xonshrc-dir-welcome.rst
@@ -1,14 +1,23 @@
 **Added:**
 
+* <news item>
+
 **Changed:**
+
+* <news item>
 
 **Deprecated:**
 
+* <news item>
+
 **Removed:**
+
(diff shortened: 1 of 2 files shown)


news/fix-xonshrc-dir-welcome.rst
```diff
@@ -0,0 +1,14 @@
...
+
+**Changed:**
+
+**Deprecated:**
+
+**Removed:**
+
+**Fixed:**
+
+*  Do not show welcome message when any `$XONSHRC_DIR` directory entry exists.
```
gforsyth, line 11:
```suggestion
*  Do not show welcome message when any ``$XONSHRC_DIR`` directory entry exists.
```

rst wants double backticks

xonsh/main.py
```diff
@@ -466,6 +466,8 @@ def func_sig_ttin_ttou(n, f):
             ignore_sigtstp()
             if env["XONSH_INTERACTIVE"] and not any(
                 os.path.isfile(i) for i in env["XONSHRC"]
+            ) and not any(
+                os.path.isdir(i) for i in env["XONSHRC_DIR"]
```
gforsyth, line 470: Can you run `black` on this?  It's complaining about the formatting

gforsyth (changes requested):
Thanks for putting this in @agoose77 !  Makes sense to me, just two small nits to fix

news/fix-xonshrc-dir-welcome.rst
```diff
@@ -0,0 +1,14 @@
...
+**Deprecated:**
+
+**Removed:**
+
+**Fixed:**
+
+*  Do not show welcome message when any ``$XONSHRC_DIR`` directory entry exists.
+
+**Security:**
+
```
gforsyth, line 14:
```suggestion
**Added:**

* <news item>

**Changed:**

* <news item>

**Deprecated:**

* <news item>

**Removed:**

* <news item>

**Fixed:**

* Do not show welcome message when any ``$XONSHRC_DIR`` directory entry exists.

**Security:**

* <news item>
```

Sorry, the news checker is very annoying (but it does help with releases)

agoose77: Oh does one need to leave those placeholders intact? Thanks for the heads up! I entirely understand, it makes us do the work up front!

gforsyth (approved):
thanks for putting this in!

PR #4532, opened 2021-11-02
Fix: support Conda environments when running on non-conda interpreter

agoose77:
Xonsh currently doesn't display the virtual environment if the user is not running xonsh under Conda:
https://github.com/xonsh/xonsh/blob/7331d8aee50e8939f8fe4d5b7133ed3907f204f4/xonsh/prompt/env.py#L15

I have xonsh running under pipx, which uses a virtualenv based on the system Python. I still use Conda to provision other virtualenvs (mainly for the ability to set the Python version; I used to use pyenv). As observed in my case, the environment that xonsh is running in should be independent from activated environments. This PR decouples the platform from the environment.

This PR changes the logic here _slightly_. `$CONDA_DEFAULT_ENV` returns the _name_ of the environment, so we don't need to invoke `os.path.basename`. I also use the `bool("") is False` idiom as I assume that only strings are ever returned here.

## For community

### DIFF
--- a/news/fix-conda-env-system.rst
+++ b/news/fix-conda-env-system.rst
@@ -8,7 +8,7 @@
 
 **Fixed:**
 
-* `$CONDA_DEFAULT_ENV` is now respected when xonsh is run outside of conda.
+* ``$CONDA_DEFAULT_ENV`` is now respected when xonsh is run outside of conda.
 
 **Security:**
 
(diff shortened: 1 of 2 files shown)


jnoortheen (approved):

jnoortheen:
Thanks @agoose77 . I hope the news items don't get in the way next time, rest of the part needs to be left as it is.

PR #4539, opened 2021-11-10
Refactor: minimise globals in codebase

agoose77:
## TLDR

This draft PR intends to drastically minimise the number of global references to the Xonsh machinery. Right now, there are several problems:
1. XSH is a god object
2. XSH is a singleton
3. Most tools reference XSH instead of the actual component (e.g. `XSH.execer.parser.lexer`)

#4280 was a good decision to make the global state explicit (for testing), and this PR takes that to the next stage. In particular, this PR has two goals:
1. Replace the mutable XSH singleton with a stack mechanism.
2. Replace references to `xonsh.session.XSH` with explicit dependency injection.

## Details
On the face of it, there are no pressing benefits for moving towards a stack mechanism. In particular, I'm not sure how often users are going to want to have multiple Xonsh sessions floating around. However, really this gives us a robust way of modifying the Xonsh session in a reversible way - e.g. for testing, just push a test session, and pop it when done. This is much safer than mutating the session in place. It might *also* be useful for _clients_ of Xonsh, rather than the Xonsh shell itself. The main design choice here is to move from mutating (`XSH.load`) into pushing onto the stack, making it clear what's going on. 

With the possibility of multiple sessions in Xonsh, there would also be the need for xontribs to re-register themselves with each session. Rather than executing at module import, as they do at present, xontribs should declare a registrar function that accepts the session similar to IPython e.g.
```python3
def load_xontrib(session):
   pass
```
This is a non-goal for this PR.

Given how Xonsh has grown organically over time, there are many places where we should be just passing in the necessary dependencies rather than using `XSH`, e.g. `CommandsCache`. We should replace those lookups with the appropriate values e.g. `cache_file` for `CommandsCache`: In my view, many of the free functions that refer to `__xonsh__` / `XSH` should instead be bound methods which refer to the context directly. Not only do these functions prevent us from shrinking the large API surface of XSH, they are also highly stateful, requiring lots of monkeypatching in the test suite to circumvent this. It is not easy when given a free function to know whether it is stateful, and what its side effects are.

This PR is a move-fast-and-break-things approach. I am familiarising myself with the code-base as I go, and my direction may change as things progress!

## For community

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,7 +27,7 @@ from xonsh.environ import Env, Var, Xettings
 
 if tp.TYPE_CHECKING:
     from xonsh.environ import VarKeyType
-from xonsh.built_ins import XSH
+import xonsh.session as xsh
 from xonsh.xontribs_meta import get_xontribs
 from xonsh.commands_cache import CommandsCache
 
@@ -37,7 +37,7 @@ import rst_helpers
 spec = importlib.util.find_spec("prompt_toolkit")
 if spec is not None:
     # hacky runaround to import PTK-specific events
-    XSH.env = Env()
+    xsh.XSH.env = Env()
     from xonsh.ptk_shell.shell import events
 else:
     from xonsh.events import events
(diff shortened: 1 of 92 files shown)


agoose77:
This approach is too aggressive - it will probably be easier to leave the XSH infrastructure intact, and tackle it last.

PR #4535, opened 2021-11-03
Refactor: reduce API surface of lexer

agoose77:
This is a small PR to clean up the lex/parse logic of xonsh. Right now, we have a lot of global state through `XSH.execer.parser.lexer`, and a large API surface. This PR will reduce the size of public API, whilst also simplifying / cleaning up the code. 

I'm not very familiar with the xonsh codebase, so the direction might change as I learn more. 

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #4540, opened 2021-11-11
Refactor: make session loading explicit

agoose77:
## TLDR

This PR moves the `XSH.load()` logic out of the execer, making it clearer where the global session is being directly modified. It also removes the execer creation from the import hooks (instead we pass it in).

This is part of a wider objective (#4535) to make the code-base more testable, and easier to reason about.  My long term goal is to significantly reduce the number of `XSH` references.

## For community


xonsh/built_ins.py
```diff
@@ -518,23 +513,68 @@ def xonsh_builtins(execer=None):
...
+        self.list_of_list_of_strs_outer_product = list_of_list_of_strs_outer_product
+        self.eval_fstring_field = eval_fstring_field
+
+        # Session attributes
+        self.exit = None
+        self.stdout_uncaptured = None
+        self.stderr_uncaptured = None
+        self._py_exit = None
+        self._py_quit = None
+        self.execer = None
```
jnoortheen, line 553: redefining the same execer twice

xonsh/commands_cache.py
```diff
@@ -67,10 +67,11 @@ def iter_commands(self):
             yield cmd, (path, is_alias)
 
     def __len__(self):
+        self.update_cache()
         return len(self.all_commands)
```
jnoortheen, line 71: will result in calling update_cache twice.

jnoortheen (approved):

jnoortheen:
Thanks @agoose77 . I have merged with few changes and fixes.

PR #4542, opened 2021-11-11
Refactor: split parsing logic from Execer

agoose77:
Currently the Execer contains parsing logic that handles the contextual resolution of Xonsh syntax. This PR separates this logic into a new `Parser` class that serves as the main parser interface.

The only backwards incompatible change is that the `xonsh.parser.Parser` object is now a different class with a different API. This can be changed if necessary.

## For community

### DIFF
--- a/xonsh/lexer.py
+++ b/xonsh/lexer.py
@@ -8,7 +8,6 @@ import re
 
 # 'keyword' interferes with ast.keyword
 import keyword as kwmod
-import typing as tp
 
 from xonsh.ply.ply.lex import LexToken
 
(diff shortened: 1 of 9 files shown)


xonsh/lexer.py
```diff
@@ -408,10 +407,39 @@ def _new_token(type, value, pos):
     return o
 
 
+def _get_token_types():
+    kwlist = kwmod.kwlist[:]
+    if PYTHON_VERSION_INFO >= (3, 9, 0) and PYTHON_VERSION_INFO < (3, 10):
+        kwlist.remove("__peg_parser__")
```
jnoortheen, line 413: This can be simply `version == 3.9` ?

agoose77: Yes, simply `sys.version_info[:2] == (3, 9). Note that I didn't write this code - it was just moved. I don't know why we have our own variable for this.

xonsh/parsers/base.py
```diff
@@ -238,20 +237,6 @@ def raise_parse_error(
...
-        super().__init__(*args, **kwargs)
-        self.daemon = True
-        self.parser = parser
-        self.yacc_kwargs = yacc_kwargs
-        self.start()
-
-    def run(self):
-        self.parser.parser = yacc.yacc(**self.yacc_kwargs)
-
-
```
jnoortheen, line 254: Does this mean, the parser is loaded eagerly? I have all my config in pure Python and parser is not needed for the startup making it faster with threaded version

jnoortheen:
@agoose77 are you still working on this?

agoose77:
> @agoose77 are you still working on this?

Hi! I'm not _right_ now as I'm not really able to spare many development cycles at the moment. I  hope to pick it up in the future. However if you need / want to work on any of my PRs, you're more than welcome! Thanks.

jnoortheen:
> I hope to pick it up in the future. 

There are file conflicts. It needs to be rebased on latest changes. Also the way it loads the parser need to be threaded too. I am closing it for now. Please feel free to re-open or create new one when it is ready.

PR #3521, opened 2020-04-16
Add ptk3

bobhy:
Support for prompt-toolkit 3.0 #3391
Comes at the cost(?) of dropping support for ptk1.
Now there's only 1 flavor of prompt_toolkit shell, so some simplifications in $SHELL_TYPE.
Updates to docs to track these.

These unit test failures were in master when I started and are still there:
```
=============================================== short test summary info ================================================
FAILED tests/test_environ.py::test_events_on_envvar_called_in_right_order - AssertionError: assert ['change'] == ['new']
FAILED tests/test_man.py::test_man_completion - AttributeError: module 'builtins' has no attribute '__xonsh__'
ERROR tests/test_main.py::test_force_interactive_custom_rc_with_script - KeyError: 'PYTEST_CURRENT_TEST'
ERROR tests/test_man.py::test_man_completion - AttributeError: __xonsh__
```

PR #4543, opened 2021-11-11
Refactor: move shells into xonsh.shell

agoose77:
The pattern `this_XXX` and `that_XXX` indicates that these modules should be placed in a parent module.

This PR also drops `xonsh.ptk2`, which was a shim around `ptk_shell`, and we might as well break things now.

See https://github.com/xonsh/xonsh/pull/3521 for the PR that added this shim.

## For community


agoose77:
I don't have time to figure out some of the quirks / design interactions with amalgam, so closing this for now.

PR #4547, opened 2021-11-16
Feat: add `Aliases.register`

agoose77:
This PR adds `Aliases.register` to register functional aliases. In these cases, we already have a name from the function itself, and this decorator simplifies user code (slightly).

Usage:
```python3
>>> @aliases.register
>>> def _some_cmd(args):
>>>     print(args)

>>> some_cmd 1 2
['1', '2']
```

## For community

### DIFF
--- a/news/feat-add-alias-decorator.rst
+++ b/news/feat-add-alias-decorator.rst
@@ -1,0 +1,17 @@
+**Added:**
+* `Aliases.register` to register an alias function.
+
+**Changed:**
+
+
+**Deprecated:**
+
+
+**Removed:**
+
+
+**Fixed:**
+
+
+**Security:**
+
--- a/news/feat-add-alias-decorator.rst
+++ b/news/feat-add-alias-decorator.rst
@@ -4,15 +4,20 @@
 
 **Changed:**
 
+* <news item>
 
 **Deprecated:**
 
+* <news item>
 
 **Removed:**
 
+* <news item>
 
 **Fixed:**
 
+* <news item>
 
 **Security:**
 
+* <news item>
--- a/news/feat-add-alias-decorator.rst
+++ b/news/feat-add-alias-decorator.rst
@@ -1,5 +1,6 @@
 **Added:**
-* `Aliases.register` to register an alias function.
+
+* ``Aliases.register`` to register an alias function.
 
 **Changed:**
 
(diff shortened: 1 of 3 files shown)


tests/test_aliases.py
```diff
@@ -192,3 +192,21 @@ def test_exec_alias_args(xonsh_builtins):
...
+    assert aliases['debug'] is debug
+
+
+def test_private_decorator(xonsh_builtins):
+    aliases = Aliases()
+    @aliases.register
+    def _debug():
+        return "DEBUG"
+
+    assert aliases['debug'] is _debug
```
gforsyth, line 212: This seems like a good spot to use `parametrize`

news/feat-add-alias-decorator.rst
```diff
@@ -0,0 +1,17 @@
+**Added:**
+* `Aliases.register` to register an alias function.
```
gforsyth, line 2:
```suggestion
**Added:**

* ``Aliases.register`` to register an alias function.
```

double backticks in RST and the news aggregator thing is picky about whitespace

gforsyth (changes requested):
This seems handy!  Thanks @agoose77 !  Two small changes and this is good to go

tests/test_aliases.py
agoose77: I considered using `parametrize` but I'm not sure whether it will be that useful here?

To test the `__name__` behaviour one still needs to declare the functions outside of the decorator, and the behaviour is so simple that I though it sensible just to hard code the tests. Would you prefer the two cases to be a single `test_decorator` case?

xonsh/aliases.py
```diff
@@ -63,6 +63,17 @@ def __init__(self, *args, **kwargs):
         self._raw = {}
         self.update(*args, **kwargs)
 
+    def register(self, func):
+        """Decorator to register the given function by name"""
+        name = func.__name__
+
+        # Strip leading underscore
+        if name.startswith("_"):
+            name = name[1:]
```
jnoortheen, line 72: A way to pass the alias name and do this by default will be great. Also replacing snake-case to dash-case will be good for cli usage

jnoortheen (approved):

jnoortheen:
Thanks @agoose77 for putting this. I had some some internal functions to do something like this. Now others can benefit from it.

PR #4564, opened 2021-11-24
feat: auto-completion support for source-foreign

jnoortheen:
## For community

### DIFF
--- a/news/foreign-funcs-completions.rst
+++ b/news/foreign-funcs-completions.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* auto-completion support for command : ``source-foreign``
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/news/foreign-funcs-completions.rst
+++ b/news/foreign-funcs-completions.rst
@@ -1,6 +1,6 @@
 **Added:**
 
-* auto-completion support for command : ``source-foreign``
+* auto-completion support for commands : ``source-foreign``, ``source-bash``, ``source-zsh``, ``source-cmd``
 
 **Changed:**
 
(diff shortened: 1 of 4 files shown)


jnoortheen:
@omjadas I haven't used foreign functions. Could you check I am not missing anything? also the todo item. I am not sure how `sourcer` is set as None and used to set prevcmd

omjadas:
> @omjadas I haven't used foreign functions. Could you check I am not missing anything? also the todo item. I am not sure how `sourcer` is set as None and used to set prevcmd

`sourcer` should either be made required, or `DEFAULT_SOURCERS` should be used to determine a default (potentially with `source` as a fallback)

jnoortheen:
> 
> `sourcer` should either be made required, or `DEFAULT_SOURCERS` should be used to determine a default (potentially with `source` as a fallback)

@omjadas I have set it to return error if sourcer not specified. 

Can you please check the working of these commands ? I have added auto-completion support to these commands.

omjadas:
> > `sourcer` should either be made required, or `DEFAULT_SOURCERS` should be used to determine a default (potentially with `source` as a fallback)
> 
> @omjadas I have set it to return error if sourcer not specified.
> 
> Can you please check the working of these commands ? I have added auto-completion support to these commands.

I have tested `source-foreign`, `source-zsh`, and `source-bash`. The actual foreign shell functionality is working as I would expect. One thing that might be a bit confusing with this change though is how the `interactive` and `safe` command line flags now work. Since they both default to `True` when I specify them from the CLI it actually sets them to False. I would probably recommend inverting the flags to be `--no-interactive` (or maybe `--non-interactive`) and `--no-safe` (or maybe `--unsafe`) and defaulting them to `False`, just so it makes a bit more sense when they are specified.

jnoortheen:
> One thing that might be a bit confusing with this change though is how the `interactive` and `safe` command line flags now work. Since they both default to `True` when I specify them from the CLI it actually sets them to `False` (previously you had to do `--flag false`). I would probably recommend inverting the flags to be `--no-interactive`/`--non-interactive` and `--no-safe`/`--unsafe` and defaulting them to `False`, just so it makes a bit more sense when they are specified.

Thanks for the check. I have changed the flag names.

Issue #4560, opened 2021-11-20
Requiring >=python3.8 versions in CI


jnoortheen:
Most of the popular libraries have adapted to support latest three versions of Python. Xonsh can also follow the suit and can drop versions lower than py3.8 ? 

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

gforsyth:
I'm not opposed to dropping lower versions but we should follow a deprecation schedule. NEP29 is one I've used on other projects and drops 3.7 quite soon.
https://numpy.org/neps/nep-0029-deprecation_policy.html

jnoortheen:
That looks good. Can we put that as part of the readme? As of version 0.11 we can drop py3.6?

gforsyth:
Sounds good to me. We can release a 0.11.1 that formally drops support

PR #4568, opened 2021-11-27
adopt NEP-0029

jnoortheen:
## For community


jnoortheen:
closes #4560

jnoortheen:
@gforsyth could you review this PR?

gforsyth (changes requested):
Nice!  I think we should also bump the `python_requires` in `setup.py`

jnoortheen:
> Nice! I think we should also bump the `python_requires` in `setup.py`

@gforsyth updated it

gforsyth (approved):
to the future!

PR #4580, opened 2021-12-02
[RFC] pip: remove pyinotify

seanfarley:
This dependency isn't used anywhere in the codebase and is linux-only,
making development on a mac immediately error out.

I'm unsure how to test this (or if it is worth testing) so I'm making this PR as a RFC.

### DIFF
--- a/requirements/development.txt
+++ b/requirements/development.txt
@@ -1,4 +1,3 @@
 -r tests.txt
 sniffer
-pyinotify
 jinja2 >= 2


seanfarley:
If this is something that y'all want, I can add the news items and fix other feedback in a followup.

requirements/development.txt
```diff
@@ -1,4 +1,3 @@
 -r tests.txt
 sniffer
-pyinotify
```
jnoortheen, line 3: using environment marker like this `pyinotify; sys_platform == 'linux'` will solve the issue.

jnoortheen: This is used alongside sniffer for watching tasks defined in `scent.py`

seanfarley: Then why not use something cross-platform like watchman or some other?

jnoortheen: It was a requirement from the library sniffer. If there is a better alternative, we can opt for that.

seanfarley: So, sniffer is a requirement of xonsh not pyinotify. Sniffer needs to install one of {pyinotify, MacFSEvents, pywin32}. Forcing that dependency into xonsh is a bad code smell.

Also, sniffer is no longer maintained so maybe something based on watchdog or watchman would be better.

jnoortheen: It is not a requirement of xonsh. If you don't use sniffer, it will not be getting in the way. It is part of the tooling like pre-commit. 

Yes I searched for some other task-runners with watch support, but Pythonic ones didn't work for me (like watchmedo, doit ...)

seanfarley: > It is not a requirement of xonsh. If you don't use sniffer, it will not be getting in the way.

This is false, as I mentioned in my PR summary because pip-installing will fail leading to much confusion. If it's really optional, then maybe another requirements file.

>  It is part of the tooling like pre-commit.

Last I checked, the only pre-commit hook was for `black`.

I'm really opposed to having phantom dependencies that are only required for certain devs that use specific OSes which aren't actually tested to make sure everything works on other OSes. Plus, sniffer isn't mentioned anywhere in the docs nor commit hooks. Yet, it breaks the dev experience on macos with no indication as to *why*.

jnoortheen: I understand the issue with naming of the file. Mainly requirements/tests.txt is ever needed to run the whole test suit. `requirements/development.txt` is for tools. 
 As noted above using environment markers will fix the issue for the devs on other OSs. Perhaps we can improve the tooling since many PRs fail on qa (black, flake8, mypy) tests. Installing those hooks will prevent that.

seanfarley: Sure, we could add the dependencies of `sniffer` (with env markers) though it's really something that `sniffer` should be doing. Perhaps something like `sniffer[extras]` will suffice, though I haven't tested. I'm not sure what the dev workflow is supposed to be wrt `dmypy` tooling (i.e. what `scent.py` is running) because it has many errors when I run it. `flake8`, `black`, and `mypy` all run clean.

jnoortheen: Sniffer has some extras defined but they will not work here 
image

It has been a long time since I used sniffer. I have moved filewatcher from Pycharm to run these mundane checks. I am not opposed to deleting sniffer completely (I was the one added it in the first place ), but it may be useful to people working in CLI primarily.

seanfarley: Then I would vote for removing it but I'll have to check back tomorrow and see.

jnoortheen: Ok lets wait for others opinion. In the meantime, I will create a PR for adding the environment markers, would you be available to review it?

jnoortheen: Oh Boy I accidentally committed on the main branch. I think the protected branch settings need to be updated

seanfarley:
Outdated now so I'll close it

PR #4606, opened 2021-12-19
fix: xonfig web is not upto-date

jnoortheen:
it should load data from backend

I want to remove elm completely and use ~~lxml builder~~ `xml.etree.ElementTree` for generating html (all in python). What do you think @gforsyth ?

## Todos

- [x] setting color, prompt from web
- [ ] ~~managing meta data with json. (we can reuse webconfig)~~

## Next PR

- [ ] prompt show left side as well
- [ ] page to update alias/abbrev/env-variable
- [ ] show key-bindings

## For community


jnoortheen:
These changes are good enough. But elm ecosystem doesn't seem to move much. I was searching for an alert package like https://ned.im/noty/ but there was no binding available for elm. I think the best option forward would be 

1. keep everything in Python  -- with small amount of JS where interactivity is needed
2. Typescript  -- if we want a heavy front-end app 
3. https://www.transcrypt.org/  -- It seems good. We can replace the small js part in the option 1 with this.

jnoortheen:
@xonsh/xore any suggestions?

jnoortheen:
@gforsyth It is functional now. It updates the variables in that particular session. This PR is getting big. I will create a follow up PR that can persist the updates to rc file.

gforsyth:
Hey @jnoortheen -- no objections from me on what system we use for the web config.  Whatever is easiest to maintain is probably best.  @anki-code -- do you have time to take a look at this?  I trust your web sensibilities over mine.

jnoortheen:
@gforsyth I chose pure Python and stdlib only. It doesn't rely on any client code. I hope it will make it easier to maintain along the rest of the code base.

gforsyth (approved):
Thanks for putting this in @jnoortheen -- I agree that having this up-to-date and easily edited is a good change.  FWIW for something like a browser-based config, I'm fine with us pulling external dependencies -- I'm also totally good with what you've done here, but if it becomes a burden in the future I have no qualms about using flask or fastapi or whatever the latest ubiquitous python web framework is.

Thanks for pushing on this!

jnoortheen:
The stdlib is good mostly. Let's see how it evolves.

PR #4611, opened 2021-12-23
fix: append space at the end of fish completions

jnoortheen:
Not including news item, because fish-completer is not released yet.

## For community


jnoortheen:
@anki-code could you review this? this fixes some path completions issues that I faced with fish completions.

anki-code (approved):

PR #4622, opened 2022-01-04
ast: do not descend into any comprehensions

### LINKED ISSUE #4529: Valid Python fails when invoked inside a function
This simple code should return the same list as it is provided with:
```python3
def _fail(x):
    return [i for i in x if i is not None and i < 10]
```
However, it fails when invoked:

```
$ _fail([1,2,3])
xonsh: subprocess mode: command not found: i
    z:   Alias
    X:   Command (X)
    w:   Command (w)
    vi:  Command (vi)
    [:   Command ([)

i: command not found
xonsh: subprocess mode: command not found: i
    z:   Alias
    X:   Command (X)
    w:   Command (w)
    vi:  Command (vi)
    [:   Command ([)

i: command not found
xonsh: subprocess mode: command not found: i
    z:   Alias
    X:   Command (X)
    w:   Command (w)
    vi:  Command (vi)
    [:   Command ([)

i: command not found
[]
```

This does *not* happen if one uses a lambda
```python3
_fail_lambda = lambda: [i for i in x if i is not None and i < 10]
```
Which produces the correct result:

```
$ _fail_lambda([1,2,3])  
[1, 2, 3]
```

## For community

yaxollum:
Fixes #4529

# Cause of the issue
The example code in #4529:
```xonsh
def f(x):
    return [i for i in x if i is not None and i < 10]

print(f([1,2,3]))
```
currently fails because of a bug in the AST `CtxAwareTransformer`. When the transformer visits the boolean operation in the list comprehension, it sees that `i` is not in scope (because list comprehensions aren't currently handled by the transformer), so it parses the entire expression as a subprocess command:
https://github.com/xonsh/xonsh/blob/35d44068d399398ec8bb3979d7f26a35a2914ad9/xonsh/ast.py#L513-L522
# Solution
This PR solves the problem by not allowing `CtxAwareTransformer` to descend into any comprehensions (list comprehensions, set comprehensions, dictionary comprehensions, and generators). This seems to be the expected behavior, because the `isdescendable` function says that UnaryOp and BoolOp are the only expressions that should be visited:
https://github.com/xonsh/xonsh/blob/35d44068d399398ec8bb3979d7f26a35a2914ad9/xonsh/ast.py#L301-L305

## For community

### DIFF
--- a/news/fix-list-comp-boolop.rst
+++ b/news/fix-list-comp-boolop.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed list comprehension in return statement incorrectly being parsed as a subprocess command.
+
+**Security:**
+
+* <news item>
--- a/tests/test_ast.py
+++ b/tests/test_ast.py
@@ -104,6 +104,9 @@ for root, dirs, files in os.walk(path):
 """,
         """lambda x: x + 1
 """,
+        """def f(x):
+    return [i for i in x if i is not None and i < 10]
+    """,
     ],
 )
 def test_unmodified(inp):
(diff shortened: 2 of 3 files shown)


gforsyth (approved):
Nice work @yaxollum !  Thanks a ton for the fix!

PR #4634, opened 2022-01-10
rm: threaded subprocs

jnoortheen:
subprocess handling without threads, the aim is to solve various thread related issues like suspending/resuming and not using threads for capturing outputs.

## Todos:

- [ ] backgrounding
- [x] piping
- [ ] remove predictable threading

## Related

#2159

## For community


Techcable:
Wait is this the new l default? I get this is probably less buggy then the old threaded subprocesses, but plase make it behind a config flag in case it ever breaks. 

I'm fine if it's on by default I just want a fallback

jnoortheen:
The idea was to remove unthreading/threading behaviour without any disruptions. It will not be merged until then. It needs to be feature compatible, and also this is just at the start.

jnoortheen:
related discussion https://github.com/xonsh/xonsh/discussions/4710

jnoortheen:
I am not working on it currently. I rarely face errors related to this as I dont use cmd capturing. I will probably work on it iin future.

jnoortheen:
found an earlier attempt here https://github.com/xonsh/xonsh/compare/main...nothread

PR #4635, opened 2022-01-11
Completers path

jnoortheen:
## For community

### DIFF
--- a/xonsh/completers/commands.py
+++ b/xonsh/completers/commands.py
@@ -1,8 +1,10 @@
 import functools
-import importlib.util
+import importlib
+import importlib.util as im_util
 import os
 import re
 import typing as tp
+from collections import OrderedDict
 
 import xonsh.tools as xt
 import xonsh.platform as xp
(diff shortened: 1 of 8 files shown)


jnoortheen:
This has two different commits, it is better to rebase-merge than squash to keep the commit msg

xonsh/completers/commands.py
```diff
@@ -105,27 +108,87 @@ def complete_end_proc_keywords(command_context: CommandContext):
...
+        """
         # list of pre-defined patterns. More can be added using the public method ``.wrap``
         self._patterns: tp.Dict[str, str] = {
             r"\bx?pip(?:\d|\.)*(exe)?$": "pip",
         }
         self._compiled: tp.Dict[str, tp.Pattern] = {}
         self.contextual = True
         self.base = base
+        # unique but maintain order
+        self._paths = OrderedDict([(pth, None) for pth in extra_paths])
```
gforsyth, line 131: This can be a regular `dict`, I think (and still be ordered)

gforsyth:
> This has two different commits

I see 5 commits -- might need a bit of squashing on your end

gforsyth (changes requested):
one nit, otherwise this looks good

xonsh/completers/commands.py
jnoortheen: Yes I considered but it doesn't feel right putting it though. It will be explicit that we follow the path order for preferring

jnoortheen:
> I see 5 commits -- might need a bit of squashing on your end

I have updated them

gforsyth (approved):

PR #4639, opened 2022-01-13
fix #4637

### LINKED ISSUE #4637: __file__ should be absolute to be consistent with python >=3.4
If a `.xsh` file is imported,  the resulting module may has a relative `__file__`, which is inconsistent with pythons behavior [since python 3.4](https://docs.python.org/3.4/whatsnew/3.4.html#other-language-changes):

> Module \_\_file\_\_ attributes (and related values) should now always contain absolute paths by default, with the sole exception of \_\_main\_\_.__file\_\_ when a script has been executed directly using a relative path. (Contributed by Brett Cannon in issue 18416.)

See also: [issue 18416](https://bugs.python.org/issue18416).

## Expected Behavior

```sh
# ensure cwd is in path so we can import local modules
import sys
sys.path.insert(0,'')

echo pass > xsh_module.xsh
echo pass > py_module.py

import xsh_module
print(xsh_module.__file__)

import py_module
print(py_module.__file__)

```
Expected output: two absolute paths 

## Current Behavior
Actual output: a relative path ("./xsh_module.xsh"), and an absolute path 

## For community

dev2718:
Fix #4637.

## For community

### DIFF
--- a/news/fix-__file__-attribute.rst
+++ b/news/fix-__file__-attribute.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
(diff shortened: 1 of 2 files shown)


gforsyth (approved):
Thanks a ton @dev2718 !

PR #4640, opened 2022-01-13
fix #4638

### LINKED ISSUE #4638: Empty .xsh files should be importable to be consistent with python
Empty or comments only `.xsh` files cannot be imported, which is in contrast to pythons behavior.

## Expected Behavior

```sh
# ensure cwd is in path so we can import local modules
import sys
sys.path.insert(0,'')

echo "#empty" > xsh_module.xsh
echo "#empty" > py_module.py

import xsh_module
import py_module

```
Expected: no errors.

## Current Behavior

`ImportError: cannot load module 'xsh_module' when get_code() returns None`

## For community

dev2718:
Fix #4638

## For community

### DIFF
--- a/news/fix-empty-modules.rst
+++ b/news/fix-empty-modules.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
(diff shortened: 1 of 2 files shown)


gforsyth (approved):
looks good, thanks @dev2718 !

PR #4658, opened 2022-01-25
chore: fix generated token for docs push

gforsyth:
we need to create the token so that it lives in the appropriate target repository (in this case, xonsh-docs)


jnoortheen:
Thanks @gforsyth

jnoortheen:
It seems to need further inspection https://github.com/xonsh/xonsh/runs/4946211832?check_suite_focus=true

PR #4659, opened 2022-01-26
add default SIGPIPE signal handler (#4655)

### LINKED ISSUE #4655: BrokenPipeError: [Errno 32] Broken pipe
This is from the article "[100 Languages Speedrun: Episode 66: Xonsh ](https://dev.to/taw/100-languages-speedrun-episode-66-xonsh-5ade)":

```python
 echo @("""
  for i in range(1, 1000001):
    print(i)
""") > /tmp/print

python /tmp/print | head -n 5

#1
#2
#3
#4
#5
#Traceback (most recent call last):
#  File "/tmp/print", line 3, in <module>
#    print(i)
#BrokenPipeError: [Errno 32] Broken pipe

```

## For community

dev2718:
Add a default signal handler for SIGPIPE that tells the kernel to not propagate SIGPIPE to xonsh.

This will fix #4655 if run with xonsh, i.e.
```
echo @("""
for i in range(1, 1000001):
  print(i)
""") > /tmp/tst.py

xonsh /tmp/tst.py | head -n 5
```

will work, whereas the following will not, as this is a problem with python itself and will also occur in other shells:

```
echo @("""
for i in range(1, 1000001):
  print(i)
""") > /tmp/tst.py

python /tmp/tst.py | head -n 5
```

## For community

### DIFF
--- a/news/fix_sigpipe.rst
+++ b/news/fix_sigpipe.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Added a default signal handler for SIGPIPE that tells the kernel to not propagate SIGPIPE to xonsh.
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Piping long output to other commands that prematurely close the pipe now works as expected. (#4655)
+
+**Security:**
+
+* <news item>
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -589,6 +589,24 @@ xexec python3 -c 'import os; print(os.environ["SHLVL"])' # == 4
 """,
         0,
     ),
+    # test piping long output to command that prematurely closes the end of its pipe. (#4655)
+    (
+        """
+echo @(\"\"\"
+for i in range(1, 1000001):
+  print(i)
+\"\"\") > /tmp/tst.py
+
+xonsh /tmp/tst.py | head -n 5 # raises BrokenPipeError if run with python
+""",
+        """1
+2
+3
+4
+5
+""",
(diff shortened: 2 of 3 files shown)


jnoortheen:
I haven't tested this yet, here is the code from stdlib doc https://docs.python.org/3/library/subprocess.html#replacing-older-functions-with-the-subprocess-module

This specifies closing the pipe after creating second process.

dev2718:
Setting the signal handler like I have done may lead to [problems](https://docs.python.org/3/library/signal.html#note-on-sigpipe) in other areas, so we need another solution ...

jnoortheen:
The recommended way to handle brokenpipe seems to be 
```py
import time
import errno


for i in range(1, 1000001):
  # time.sleep(1)
  try:
    print(1, flush=True)
  except IOError as ex:
    if ex.errno == errno.EPIPE:
      break
    else:
      raise ex
```

I think we should do the same, and let programs handle the pipe errors.

dev2718:
I agree, it is probably best to let developers decide how to handle SIGPIPE.

Actually, bash shows the same behavior for the bash equivalent of this example and will also terminate, but will not report any errors.
This is due to bash's [DONT_REPORT_SIGPIPE](https://github.com/bminor/bash/blob/f3a35a2d601a55f337f8ca02a541f8c033682247/config-top.h#L51) which defaults to true.

So one possibility is to:
- At least only report something like "Broken Pipe" instead of the long intimidating stack trace in case xonsh dies due to SIGPIE, 
- And maybe an option `XONSH_REPORT_SIGPIPE` which defaults to False to mirror bash

What do you think?

jnoortheen:
I like the second option, we can use the variable to enable the traceback if one wants it

PR #4671, opened 2022-02-18
use recommended pytest entry point in docs and comments

obestwalter:
Since pytest 3.0 (2016) - the recommended commandline invocation is pytest (without the dot).

py.test is still available for backwards compatibility but might be removed at some point in the future and the new entry point is already well established.

I didn't add a changelog enry as this is only a teeny tiny correction/improvement of the docs and comments and not really worth mentioning.

## For community

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -204,7 +204,7 @@ Running the Tests - Basic
 
 Run all the tests using pytest::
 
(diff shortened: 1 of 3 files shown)


jnoortheen (approved):

jnoortheen:
Thanks @obestwalter !

PR #4672, opened 2022-02-18
Feat: support for match statement for python >= 3.10 

dev2718:
This PR implements the parsing required for `PEP 634 -- Structural Pattern Matching: Specification` yielding full support for pythons match statement for python >= 3.10.
Furthermore, it introduces a list of additional (completely optional and noninterfering) xonsh-specific patterns.

`match` and `case` are implemented as "soft" keywords (according to the spec), such that e.g. `match = 1` is still valid syntax.

## Xonsh-extensions to pythons match statement (not part of this PR anymore)

These are completely optional. In case they are not used, exactly the same abstract syntax tree is produced as it would be by vanilla python.

For more examples, see [test_match.py](https://github.com/dev2718/xonsh/blob/feat_match/tests/test_match.py).

If the community/the maintainers approve of these additions I will update the tutorial accordingly.

### Safe transformer patterns

Passes the match subject to a function and matches against a subpattern (Syntax: `subpattern -> function`). Exceptions deriving from `Exception` are swallowed and result in match failure.

This is useful to all at once parse, validate and convert (nested) data and can be used for all sorts of custom matching behaviors.

For example:
```python
import json
json_str = ' {"pi":"3.141"} '
match json_str:
    case {"pi": pi -> float} -> json.loads:
        # the string is correct json with a "pi" field containing a value that can be casted to float.
        # pi contains the parsed float
        print(pi)
```

#### Without subpattern
In its simplest form, we can `re.fullmatch` the match subject cast as string against a regex pattern specified via a backticks-string literal.

Example:
```python
match "test":
        case `.est`:
            pass
```

#### With subpattern

Additionally, we can extract the specified regex groups as a list, and match it against a subpattern via the syntax `subpattern -> regex-literal`.

Example:
```python
def reverse(x):
    match x:
        case [head, tail] -> `(.)(.*)`:
            return reverse(tail) + head
        case ``:
            return ""
```

#### Custom search-functions
By prepending a function with `@` we can perform custom matching that need not be related with regular expressions.
For example, we could extract urls from html using something like this:

```python
match html_str:
    case @find_urls`https.*` as secure_urls:
        pass
```

For a full example, see [test_match.py](https://github.com/dev2718/xonsh/blob/feat_match/tests/test_match.py).

### Expression patterns
This allows to match (by equality) against xonsh-specific constructs like `$...` or `$(...)` and arbitrary python expressions via `@(...)` as either stand-alone patterns, or as keys in mapping patterns.

Example:
```python
$x = 1
$y = 4

result = []
for x in [1, 2, "3\\n", 4, {5 : "e"}]:
    match x:
        case $x:
            result.append("a")
        case @(1+1):
            result.append("b")
        case $(echo 3):
            result.append("c")
        case ${"y"}:
            result.append("d")
        case {@(4+1) : _}:
            result.append("e")
        case _:
            raise f"no match for {x}"
```

### Predicate patterns
Use any function that takes a single value as matching-criterion via the syntax `? func` or `? not func`.
Exceptions deriving from `Exception` are swallowed, so that for example `? int` can be used to check whether a value can be converted to an integer (and is non-zero).

Example:
```python
match string:
        case ?str.isupper:
                # uppercase string
                pass
        case ? not str.isupper:
                # lowercase string or not a string
                pass
        case ?(lambda x: x > 0):
                # numeric value greater than zero
                pass
```


jnoortheen:
AFAIK All Xonsh specific syntax sugars are to make interacting with subprocess easier. I think we should stick to this. I can see that these additions will be bery much helpful, but will be adding more complexity. Can't we just use plain old function calls while matching ? it will be very readable

jnoortheen:
Ok. Lets wait for others opinion

jnoortheen:
pinging @xonsh/xore

agoose77:
Hi @dev2718, thanks for this PR! I can see `match` being very useful to Xonsh applications.

I am personally strongly against adding any additional syntactic-sugar here without there being a basis / need for it. Every feature, even when ignored by the user, adds more code to the core and test suite that the developers need to maintain. 

More importantly though, it means we are more likely to encounter a case where Python introduces new syntax / rules that are incompatible with Xonsh. As an aside, the return annotation implementation is a little confusing to me - I would expect the transformation to somehow apply to the match result rather than the case.

gforsyth:
Hey @dev2718 -- thanks for putting this in!  I'm definitely on board for `match` being available in `xonsh` -- I'll echo the points made by @jnoortheen and @agoose77 that we want to avoid introducing new syntax without a specific cause.  It's very hard to change syntax once it's being used by a large group of people and the case of syntax collision with Python itself (due to new features added in the future) is very real.

dev2718:
> I am personally strongly against adding any additional syntactic-sugar here without there being a basis / need for it. Every feature, even when ignored by the user, adds more code to the core and test suite that the developers need to maintain.

On the issue of need and basis, I do think that 
- parsing a data source (e.g. from json)
- validating that it conforms to some format
- extracting information, oftentimes using regex
- converting the validated data to another form

are all tasks pretty commonly performed in shell scripts.

Just as a (toy) example, I think

```python
match ' {"information":"value=123,name=TEST"} ':
    case {"information": [value -> float, ?str.isupper as name] -> `value=(\d+),name=(\w+)`} -> json.loads:
        print(value, name)
    case "something different":
        print("something different")
```
is more descriptive and less complex than the equivalent
```python
data = ' {"information":"value=123,name=TEST"} '
try:
    parsed = json.loads(data)
    match parsed:
        case {"information": information_str}:
            import re
            if m := re.fullmatch(r"value=(\d+),name=(\w+)", information_str):
                value_str, name_str = m.groups()
                value = float(value_str)
                if name.isupper():
                    print(value, name)
except ValueError:
    pass
# this cannot be another case but needs to be a full match statement
match data:
    case "something different":
        print("something different")
```


> More importantly though, it means we are more likely to encounter a case where Python introduces new syntax / rules that are incompatible with Xonsh. 

I don't think there could be semantic incompatibility with future versions of python since the implementation only relies on attribute value patterns working as specified, and I don't think these will be removed from python.
Syntax collisions on the other hand are of course possible.

> As an aside, the return annotation implementation is a little confusing to me - I would expect the transformation to somehow apply to the match result rather than the case.

The transformation does apply to the match result (in the subpattern).
If the feature is somehow counter intuitive or does not work as expected, that would be good criticism.

dev2718:
> Hey @dev2718 -- thanks for putting this in! I'm definitely on board for `match` being available in `xonsh` -- I'll echo the points made by @jnoortheen and @agoose77 that we want to avoid introducing new syntax without a specific cause. It's very hard to change syntax once it's being used by a large group of people and the case of syntax collision with Python itself (due to new features added in the future) is very real.

I can get where you're coming from.
What about putting the additions behind a flag like $XONSH_EXPERIMENTAL_SYNTAX?
In this way it could be removed/altered should problems arise in the future, while setting expectations for the users that do want to use it.

gforsyth:
> I can get where you're coming from.
> What about putting the additions behind a flag like $XONSH_EXPERIMENTAL_SYNTAX?
> In this way it could be removed/altered should problems arise in the future, while setting expectations for the users that do want to use it.

I'm not opposed to feature flags, especially if we're actually trying things out experimentally -- although I've found in the past that anything that is available _will_ be used and users get annoyed when things they use go away.  

I'm going to suggest that given the size of this PR (and again, _thank you for this awesome work_) that we split it into the implementation of the `match` statement in this PR and then open up separate PRs for the extensions.

dev2718:
> I'm not opposed to feature flags, especially if we're actually trying things out experimentally -- although I've found in the past that anything that is available _will_ be used and users get annoyed when things they use go away.
> 
> I'm going to suggest that given the size of this PR (and again, _thank you for this awesome work_) that we split it into the implementation of the `match` statement in this PR and then open up separate PRs for the extensions.

I have removed the extensions for now, but I hope that they will be considered as an experimental feature in the future.

.github/workflows/qa-linux-3.10.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: [ "3.9" ]
+        python-version: [ "3.10" ]
```
jnoortheen, line 18: these are generated file and you have to update the line [here](https://github.com/xonsh/xonsh/blob/81c8fe6bb92abe53277fbfffe6461630440c4abd/.github/workflows/genbuilds.py#L42) to generate them with py3.10

xonsh/parsers/v310.py
```diff
@@ -1,7 +1,12 @@
+# type: ignore
+# TODO: remove line above once mypy understands the match statement
+
 """Handles changes since PY310
 
 handle
 - import-alias requiring lineno
+- match statement
+- xonsh additions to pythons match statement
```
jnoortheen, line 9: this line is no longer relevant in this PR.

jnoortheen (approved):

jnoortheen:
@dev2718 it looks great. Just the small nit about the docstring.

jnoortheen:
Neat Thanks @dev2718

gforsyth:
woot!  Thanks @dev2718 !

PR #4673, opened 2022-02-19
make env_name aware of prompt in pyvenv.cfg

### LINKED ISSUE #4670: make promp.env.env_name venv --prompt aware (set in pyvenv.cfg)
## xonfig

<details>

```
> xonfig
<xonsh-code>:1:0 - xonfig
<xonsh-code>:1:0 + ![xonfig]
| xonsh            | 0.11.0               |
| Git SHA          | adfa60ea             |
| Commit Date      | Feb 11 14:53:00 2022 |
| Python           | 3.9.7                |
| prompt toolkit   | None                 |
| pygments         | 2.11.2               |
| distro           | unknown              |
| on wsl           | False                |
| xontrib 1        | coreutils            |
| xontrib 2        | vox                  |
| xontrib 3        | voxapi               |
| RC file 1        | /home/ob/.xonshrc    |
```

</details>

## Expected Behavior

When activating a venv via `vox activate` the name should be set to the value of `prompt` key in `pyvenv.cfg` if present (file and key) - see https://docs.python.org/3/library/venv.html.

## Current Behavior

The prompt is always set to name of venv directory independent of prompt settings.

## Steps to Reproduce

```shell
$ python -m venv --prompt "MY SPECIAL PROMPT" .venv
$ cat .venv/pyvenv.cfg | grep prompt
prompt = 'MY SPECIAL PROMPT'
vox activate ./.venv
```

new prompt looks like this:
`[17:58:10] (.venv)  ob@ob1 ~/oss/xonsh dropthedot|✓`

but should look like this:
`[17:58:10] (MY SPECIAL PROMPT)  ob@ob1 ~/oss/xonsh dropthedot|✓`

## Also add `--prompt` to `vox new`?
If this is done, maybe a good idea would be to also add `--prompt` to `vox new` to make this possible on creation (also in the interest of symmetry), but it seems like the common pattern for many xonsh users is to have virtualenvs in `~/.virtualenv` with the venv folder name being the same as the project, therefore automatically having a meaningful prompt, so this does not seem to be an urgently wanted feature for most.

I want to prepare a PR for the "passive" part of the functionality, but could also have a stab at adding `--prompt`  to `vox new` at a later date. My main itch would be to respect `prompt` from already existing venvs first.

## For community

obestwalter:
prompt.env.env_name is now aware of the "prompt" key in ``pyvenv.cfg`` - search order from first to last is: ``$VIRTUAL_ENV_PROMPT``, ``$VIRTUAL_ENV`` (prompt from ``pyvenv.cfg`` or basename of venv path if not set), ``$CONDA_DEFAULT_ENV``.

fixes #4670

## For community

### DIFF
--- a/xonsh/prompt/env.py
+++ b/xonsh/prompt/env.py
@@ -2,7 +2,6 @@
 import functools
 import re
 from pathlib import Path
-from typing import Optional
 
 from xonsh.built_ins import XSH
 
--- a/xonsh/prompt/env.py
+++ b/xonsh/prompt/env.py
@@ -1,39 +1,55 @@
 """Prompt formatter for virtualenv and others"""
 
-import os
+import re
+from pathlib import Path
 
 from xonsh.built_ins import XSH
 
 
-def find_env_name():
-    """Finds the current environment name from $VIRTUAL_ENV or
-    $CONDA_DEFAULT_ENV if that is set.
-    """
-    env_path = XSH.env.get("VIRTUAL_ENV", "")
-    if env_path:
-        env_name = os.path.basename(env_path)
-    else:
-        env_name = XSH.env.get("CONDA_DEFAULT_ENV", "")
-    return env_name
-
-
 def env_name():
-    """Returns the current env_name if it non-empty, surrounded by the
-    ``{env_prefix}`` and ``{env_postfix}`` fields.
-    """
-    env_name = find_env_name()
-    if XSH.env.get("VIRTUAL_ENV_DISABLE_PROMPT") or not env_name:
-        # env name prompt printing disabled, or no environment; just return
-        return
+    """Build env_name based on different sources. Respect order of precedence.
 
-    venv_prompt = XSH.env.get("VIRTUAL_ENV_PROMPT")
-    if venv_prompt is not None:
-        return venv_prompt
-    else:
-        pf = XSH.shell.prompt_formatter
-        pre = pf._get_field_value("env_prefix")
-        post = pf._get_field_value("env_postfix")
-        return pre + env_name + post
+    Name from VIRTUAL_ENV_PROMPT will be used as-is.
+    Names from other sources are surrounded with ``{env_prefix}`` and
+    ``{env_postfix}`` fields.
+    """
+    if XSH.env.get("VIRTUAL_ENV_DISABLE_PROMPT"):
+        return
+    virtual_env_prompt = XSH.env.get("VIRTUAL_ENV_PROMPT")
+    if virtual_env_prompt:
+        return virtual_env_prompt
+    virtual_env = XSH.env.get("VIRTUAL_ENV")
+    if virtual_env:
+        venv_path = Path(virtual_env)
+        pyvenv_cfg_prompt = prompt_from_pyvenv_cfg(venv_path)
+        if pyvenv_cfg_prompt:
+            return surround_env_name(pyvenv_cfg_prompt)
+        return surround_env_name(venv_path.name)
+    from_conda = XSH.env.get("CONDA_DEFAULT_ENV")
+    if from_conda:
+        return surround_env_name(from_conda)
+    return
+
+
+def prompt_from_pyvenv_cfg(venv_path):
+    """Grab the prompt from the venv configuration, if it exists.
+
+    Tries to be resilient to subtle changes in whitespace and quoting in the
+    configuration file format as it adheres to no clear standard.
+    """
+    assert isinstance(venv_path, Path), venv_path
+    pyvenv_cfg = venv_path / "pyvenv.cfg"
+    if pyvenv_cfg.is_file():
+        match = re.search(r"prompt\s*=\s*(.*)", pyvenv_cfg.read_text())
+        if match:
+            return match.group(1).strip().lstrip("'\"").rstrip("'\"")
+
+
+def surround_env_name(name):
+    pf = XSH.shell.prompt_formatter
+    pre = pf._get_field_value("env_prefix")
+    post = pf._get_field_value("env_postfix")
+    return f"{pre}{name}{post}"
 
 
 def vte_new_tab_cwd():
--- a/xonsh/prompt/env.py
+++ b/xonsh/prompt/env.py
@@ -7,7 +7,7 @@ from typing import Optional
 from xonsh.built_ins import XSH
 
 
-def env_name() -> Optional[str]:
+def env_name() -> str:
     """Build env_name based on different sources. Respect order of precedence.
 
     Name from VIRTUAL_ENV_PROMPT will be used as-is.
@@ -15,7 +15,7 @@ def env_name() -> Optional[str]:
     ``{env_postfix}`` fields.
     """
     if XSH.env.get("VIRTUAL_ENV_DISABLE_PROMPT"):
-        return
+        return ""
     virtual_env_prompt = XSH.env.get("VIRTUAL_ENV_PROMPT")
     if virtual_env_prompt:
         return virtual_env_prompt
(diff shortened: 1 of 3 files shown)


obestwalter:
I don't understand why black is failing. I followed the developer guide and when I run `black --check xonsh/ xontrib/` locally all is good.

my devenv:

```
(XONSH DEV) [0] < ~/oss/xonsh > 16:00:49 $
pip list
Package               Version Location
--------------------- ------- ------------------
amalgamate            0.1.3
attrs                 21.4.0
black                 22.1.0
cfgv                  3.3.1
click                 8.0.4
coverage              6.3.1
distlib               0.3.4
distro                1.7.0
docutils              0.18.1
filelock              3.6.0
flake8                4.0.1
flake8-bugbear        22.1.11
flake8-docstrings     1.6.0
identify              2.4.10
iniconfig             1.1.1
isort                 5.10.1
mccabe                0.6.1
mypy                  0.931
mypy-extensions       0.4.3
nodeenv               1.6.0
packaging             21.3
pathspec              0.9.0
pip                   21.2.3
platformdirs          2.5.0
pluggy                1.0.0
pre-commit            2.17.0
prompt-toolkit        3.0.28
py                    1.11.0
pycodestyle           2.8.0
pydocstyle            6.1.1
pyflakes              2.4.0
Pygments              2.11.2
pyparsing             3.0.7
pyperclip             1.8.2
pyte                  0.8.0
pytest                7.0.1
pytest-cov            3.0.0
pytest-mock           3.7.0
pytest-rerunfailures  10.2
pytest-subprocess     1.4.1
pytest-timeout        2.1.0
PyYAML                6.0
restructuredtext-lint 1.3.2
setuptools            57.4.0
six                   1.16.0
snowballstemmer       2.2.0
toml                  0.10.2
tomli                 2.0.1
types-ujson           4.2.1
typing_extensions     4.1.1
virtualenv            20.13.1
wcwidth               0.2.5
wheel                 0.36.2
xonsh                 0.11.0  /home/ob/oss/xonsh
```

obestwalter:
oh ... looks like the developer guide needs an update - tests/ is also "blacked" now :)

xonsh/prompt/env.py
```diff
@@ -1,39 +1,55 @@
...
+    """Grab the prompt from the venv configuration, if it exists.
+
+    Tries to be resilient to subtle changes in whitespace and quoting in the
+    configuration file format as it adheres to no clear standard.
     """
-    env_name = find_env_name()
-    if XSH.env.get("VIRTUAL_ENV_DISABLE_PROMPT") or not env_name:
-        # env name prompt printing disabled, or no environment; just return
-        return
+    assert isinstance(venv_path, Path), venv_path
```
jnoortheen, line 40: can the assert statement be skipped?

xonsh/prompt/env.py
```diff
@@ -1,39 +1,55 @@
...
+    pyvenv_cfg = venv_path / "pyvenv.cfg"
+    if pyvenv_cfg.is_file():
+        match = re.search(r"prompt\s*=\s*(.*)", pyvenv_cfg.read_text())
+        if match:
```
jnoortheen, line 44: can we cache the compiled regex with lru_cache ? this function is invoked for each new prompt and needs to be as fast as possible

jnoortheen:
> oh ... looks like the developer guide needs an update - tests/ is also "blacked" now :)

the guide should be `xonsh run-tests.xsh qa` to run all the checks.

xonsh/prompt/env.py
```diff
@@ -1,39 +1,55 @@
...
+    if XSH.env.get("VIRTUAL_ENV_DISABLE_PROMPT"):
+        return
+    virtual_env_prompt = XSH.env.get("VIRTUAL_ENV_PROMPT")
+    if virtual_env_prompt:
+        return virtual_env_prompt
+    virtual_env = XSH.env.get("VIRTUAL_ENV")
+    if virtual_env:
+        venv_path = Path(virtual_env)
+        pyvenv_cfg_prompt = prompt_from_pyvenv_cfg(venv_path)
+        if pyvenv_cfg_prompt:
```
jnoortheen, line 25: we can use caching here as well. with path as the key using lru_cache with size as 3-5

xonsh/prompt/env.py
obestwalter: I wasn't sure if there is some magic caching going on, that I am not aware of, that's why I didn't do anything yet. If this is the case thoug, the file access is also not so great there.

I have another idea. This might feel kinda dirty, but it would solve all potential speed problems in one fell swoop: when this function is called, this means that VIRTUAL_ENV_PROMPT was not set (as it has highest precedence). So this could be used as the cache.

xonsh/prompt/env.py
obestwalter: See my comment above. This code path wouldn't even be hit anymore then.

xonsh/prompt/env.py
jnoortheen: we can't be sure that VIRTUAL_ENV is not updated externally, either by running a source command or such. so we should not be making assumptions on that. Here we can compile the regex part and match, and cache the result from opening and reading a file.

xonsh/prompt/env.py
obestwalter: You mean remove the assert? I usually like to be a bit more paranoid, especially in highly dynamic code without type hinting. It also helps IDEs to figure out, what is going on. But if this against the grain of this project, I can remove it. If it is again about speed, this problem would als be solved, if the prompt is stored in VIRTUAL_ENV_PROMPT after first access.

jnoortheen: well that can be an if statement. We use mypy, and try to add type hint wherever possible.

xonsh/prompt/env.py
obestwalter: ah, ok - good point. What do you think about caching the whole function result based on the path that was passed in?

jnoortheen:
Please configure pre-commit. It can help avoid the style issues.

xonsh/prompt/env.py
obestwalter: ah ... I think this is what you meant, I said the same thing above :D - will have a go at this now. Thanks!

xonsh/prompt/env.py
jnoortheen: Yes we can cache the whole function as well as the re.compiled pattern to match it. Since it will save some when a new path is hit

xonsh/prompt/env.py
obestwalter: The assert has lost its usefulness in the new version anyway, so it's gone. But good to know that you use type hints. In the parts of the code I was until now.

obestwalter:
My local qa check failed now, because isort wanted to resort a whole bunch of files now that I didn't even touch. I pushed without those changes now, as I want to see if this also fails on CI now.

obestwalter:
Well, this got ugly fast :D - I am not really with the caching solution, as this now introduces the possibility that the user will get a stale prompt, if they fiddle with the config or recreate the venv with different prompt settings, but I only added an xfailing reproducer for the problem now and still think that this is at least less bad as completely ignoring prompt setting in cfg.

obestwalter:
As env.py has quite a lot already anyway, I added the missing type hints now. I will leave this hanging around until tomorrow now and see if anything else needs to be done, before it can be merged. Thanks for your help @jnoortheen!

xonsh/prompt/env.py
jnoortheen: @obestwalter still the regex is not compiled, please create a function that caches the regex part `r"prompt\s*=\s*(.*)"`. rest of them looks fine.

obestwalter:
Is this good to go, or do I need anything else before this can be merged?

jnoortheen (approved):

jnoortheen:
Thanks @obestwalter !

PR #4683, opened 2022-02-26
feat: colored tracebacks during interactive use

dev2718:
Since pygments includes `PythonTracebackLexer`, we can use it to make tracebacks prettier during interactive use.

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -42,8 +42,10 @@ import warnings
 # dependencies
 from xonsh import __version__
 from xonsh.lazyasd import LazyDict, LazyObject, lazyobject
+from xonsh.lazyimps import pygments
 from xonsh.platform import (
     DEFAULT_ENCODING,
+    HAS_PYGMENTS,
     ON_LINUX,
     ON_WINDOWS,
     expanduser,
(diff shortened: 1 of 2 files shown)


xonsh/tools.py
```diff
@@ -1032,7 +1034,28 @@ def print_exception(msg=None, exc_info=None):
...
+        )
+
+        # color the traceback if available
+        _, interactive = _get_manual_env_var("XONSH_INTERACTIVE", 0)
+        _, color_results = _get_manual_env_var("COLOR_RESULTS", 0)
+        _, shell_type = _get_manual_env_var("SHELL_TYPE")
+        if (
+            interactive
+            and color_results
+            and shell_type == "prompt_toolkit"
```
jnoortheen, line 1049: does shell_type need to be ptk?

dev2718: Thank you, I have now removed this restriction. The original problem was that I did not import enough, and it only worked "by chance" on prompt_toolkit since it loaded the required submodules.

jnoortheen:
Looks good! Thanks @dev2718

jnoortheen (approved):

PR #4689, opened 2022-03-03
Add goat counter to track hits on the docs

gforsyth:
See goatcounter.com for more information.

Goat Counter strives to be privacy preserving and complies with GDPR.  Also doesn't use cookies for tracking so no annoying cookie announcement banner -- this is really just to track hits to the xonsh documentation to maybe start to get a sense of the size of the userbase.  Who knows?

Anyone in @xonsh/xore who wants access just ping me.


jnoortheen:
OK we can create an issue to track that ?

PR #4691, opened 2022-03-03
fix(maybe): use generated token to push docs

gforsyth:
Got the token to generate correctly, but now it's complaining about
scoping -- might be that it needs to be generated for xonsh/xonsh to
push to xonsh/xonsh-docs

## For community


jnoortheen:
Until the docs generation fixed, we can temporarily commit to main branch. It may require some iterations before being correct

jnoortheen (approved):

PR #4694, opened 2022-03-04
docs: add note about Vim and :read command

Flimm:
This helps xonsh users who use Vim work around a bug. Users who have set `xonsh` to the default shell see this error when running a command like `:read !date` in Vim:

> SyntaxError
> shell returned 1
>
> E485: Cannot read file /tmp/vsrBORD/5

This documentation adds a workaround.

### DIFF
--- a/docs/editors.rst
+++ b/docs/editors.rst
@@ -95,3 +95,14 @@ There is `xonsh syntax file for vim`_. To install run:
     git clone --depth 1 https://github.com/linkinpark342/xonsh-vim ~/.vim
 
 .. _xonsh syntax file for vim: https://github.com/linkinpark342/xonsh-vim
+
+In Vim, it is possible to read the ouput of an external command using
+``:read``, prefacing the shell command with ``!``. For example, running
+``:read !date`` inserts the current date into the text file being edited. You
+may find that this breaks once xonsh is set as the default shell. As a
+workaround, you may wish to add this line to your ``.vimrc`` configuration file
+to use Bash for this purpose instead:
+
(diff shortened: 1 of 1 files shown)


docs/editors.rst
```diff
@@ -95,3 +95,14 @@ There is `xonsh syntax file for vim`_. To install run:
...
+In Vim, it is possible to read the ouput of an external command using
+``:read``, prefacing the shell command with ``!``. For example, running
+``:read !date`` inserts the current date into the text file being edited. You
+may find that this breaks once xonsh is set as the default shell. As a
+workaround, you may wish to add this line to your ``.vimrc`` configuration file
+to use Bash for this purpose instead:
+
+.. code-block::
+
+    set shell=bash
```
jnoortheen, line 108: instead of specifically mentioning `bash` , you can say some other POSIX compliant shells like bash,zsh...

dev2718:
I think the better solution would be to fix the root cause.
I did some tracing, (with neovim) and found that (neo)vim invokes xonsh like this (for :read !ls -l):
```bash
/path/to/xonsh -c "(ls -l) > /tmp/nvimNWysmi/5"
``` 
This raises a SyntaxError, since apparently the path the output is directed to cannot begin with '/' when xonsh is invoked with the `-c` option.

jnoortheen:
> I think the better solution would be to fix the root cause

Yes that will be better

jnoortheen:
closing infavour of https://github.com/xonsh/xonsh/pull/4699

Issue #4343, opened 2021-06-29
Vim system fails when $SHELL is pointing to `xonsh`


igormorgado:
## Current behavior

When running vim's commands `system` and `systemlist`  with `$SHELL = /usr/bin/xonsh`  (or to local pip installed xonsh. Vim fails with message:

```
E484: Can't open file /tmp/vlsyynp/4
```

The path is randomly generated to receive the vim output. But for some reason isn't created correctly. 

## Expected behavior

Of course, the expected output is receive the output of `ls` command as a string, or as a vim list (when calling `systemlist`)

## Steps to reproduce:

1. Start `xonsh`
2. Set `$SHELL = PATH_TO_XONSH`
3. Start `vim`
4.1. type `:echom system ('ls')`; or
4.2. type `:echom systemlist ('ls')`.

It works fine if $SHELL is let alone pointing to `/bin/bash`,  `/usr/bin/fish` or `/usr/bin/zsh`, hence I think that is somehow related to `xonsh`.

## xonfig
<details>

```
[127]@ xonfig info
+------------------+----------------------+
| xonsh            | 0.9.27               |
| Git SHA          | 71fe9014             |
| Commit Date      | Jan 29 08:58:58 2021 |
| Python           | 3.9.2                |
| PLY              | 3.11                 |
| have readline    | True                 |
| prompt toolkit   | 3.0.16               |
| shell type       | prompt_toolkit       |
| pygments         | 2.7.1                |
| on posix         | True                 |
| on linux         | True                 |
| distro           | debian               |
| on darwin        | False                |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
| on jupyter       | True                 |
| jupyter kernel   | None                 |
| xontrib 1        | apt_tabcomplete      |
| xontrib 2        | autovox              |
| xontrib 3        | cmd_done             |
| xontrib 4        | gitinfo              |
| xontrib 5        | jedi                 |
| xontrib 6        | kitty                |
| xontrib 7        | mpl                  |
| xontrib 8        | pdb                  |
| xontrib 9        | prompt_bar           |
| xontrib 10       | prompt_ret_code      |
| xontrib 11       | prompt_vi_mode       |
| xontrib 12       | ssh_agent            |
| xontrib 13       | vox                  |
| xontrib 14       | voxapi               |
[...]

PR #4699, opened 2022-03-06
fix 4343 by always trying greedy=True in _parse_ctx_free

dev2718:
This change makes some statements that were syntax errors previously like
```bash
(ls) > /x
```
valid syntax, which solves #4343 and replaces #4694.

## For community


jnoortheen (approved):

jnoortheen:
Looks good!

PR #4715, opened 2022-03-14
Fix regex globbing for paths that contain special regex characters

### LINKED ISSUE #4690: Regex globbing broken for path including active regex characters
Paths including active regex characters such as `test+a/model` are not matched by regex globbing:

```xonsh
mkdir -p test/model
mkdir -p test+a/model
mkdir -p test*1/model
for d in `.*/model`:
    print(d)
```

Gives the following output

```output
test/model
```

## For community

yaxollum:
Fixes #4690

This PR fixes the issue that regex globbing for paths like `test+a/model` and `test*1/model` is broken. Currently, these paths do not get matched by the pattern `.*/model`.

## Cause of the issue
The cause of the issue is that `reglob()` currently joins `base` (part of the file path containing problematic special characters) with `parts[i]` (part of the intended regular expression), and _compiles the whole thing as a regular expression_. This causes special regex characters in file paths to be incorrectly interpreted as part of the search pattern:
https://github.com/xonsh/xonsh/blob/78f91952273f0a81dc226abf91ae16e07c125103/xonsh/built_ins.py#L95-L100

## Solution
This PR fixes the issue by not joining `base` into the regular expression, and instead matching `parts[i]` on the last component of the current path. `regex.fullmatch(p)` gets changed into `regex.fullmatch(f)`

## For community

### DIFF
--- a/news/fix-globbing-path-containing-regex.rst
+++ b/news/fix-globbing-path-containing-regex.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed regex globbing for file paths that contain special regex characters (e.g. "test*1/model")
+
+**Security:**
+
+* <news item>
--- a/tests/test_builtins.py
+++ b/tests/test_builtins.py
@@ -1,8 +1,10 @@
 """Tests the xonsh builtins."""
 import os
 import re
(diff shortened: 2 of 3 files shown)


yaxollum:
It looks like I'll have to fix the code for Windows

jnoortheen:
This may help you test windows testing https://github.com/xonsh/xonsh/blob/main/CONTRIBUTING.rst#cross-platform-testing

jnoortheen (approved):

jnoortheen:
Looks good! Thanks @yaxollum

PR #4718, opened 2022-03-15
chore: update `mypy`

gforsyth:
Github stopped allowing `git+ssh` installs and we were due to move to `0.940`


jnoortheen (approved):

jnoortheen:
Thanks @gforsyth

Issue #4721, opened 2022-03-18
More bugs: os.fork and providing opening parenthesis on non-callables


GanerCodes:
https://user-images.githubusercontent.com/34655998/159010125-1a538a27-5ece-4d23-a110-b7f5eca6894b.mp4

jnoortheen:
It would help to have the environment info . Also text format is preferred than videos where we can reproduce it by simple copy paste.

GanerCodes:
text is just most non-method things followed by `(` for the callable example, and import os; os.fork()  
  
environment is:
```
~ λ ${...}
xonsh.__amalgam__.Env(
 {'ALIAS_COMPLETIONS_OPTIONS_BY_DEFAULT': False,
  'ASYNC_INVALIDATE_INTERVAL': 0.05,
  'ASYNC_PROMPT_THREAD_WORKERS': None,
  'AUTO_CD': False,
  'AUTO_CONTINUE': False,
  'AUTO_PUSHD': False,
  'AUTO_SUGGEST': True,
  'AUTO_SUGGEST_IN_COMPLETIONS': False,
  'BASH_COMPLETIONS': EnvPath(
['/usr/share/bash-completion/bash_completion']
),
  'BOTTOM_TOOLBAR': '',
  'CASE_SENSITIVE_COMPLETIONS': False,
  'CDPATH': (),
  'COLORFGBG': '15;0',
  'COLORTERM': 'truecolor',
  'COLOR_INPUT': True,
  'COLOR_RESULTS': True,
  'COMMANDS_CACHE_SAVE_INTERMEDIATE': True,
  'COMMANDS_CACHE_SIZE_WARNING': 6000,
  'COMPLETE_DOTS': 'matching',
  'COMPLETIONS_BRACKETS': False,
  'COMPLETIONS_CONFIRM': False,
  'COMPLETIONS_DISPLAY': 'single',
  'COMPLETIONS_MENU_ROWS': 5,
  'COMPLETION_IN_THREAD': True,
  'COMPLETION_MODE': 'default',
  'COMPLETION_QUERY_LIMIT': 100,
  'CONFIG_PROTECT': '/usr/share/gnupg/qualified.txt /usr/share/config',
  'CONFIG_PROTECT_MASK': '/etc/sandbox.d /etc/fonts/fonts.conf /etc/gentoo-release /etc/gconf /etc/terminfo /etc/dconf /etc/ca-certificates.conf /etc/revdep-rebuild',
[...]

PR #4722, opened 2022-03-18
Fix annoying console spam

GanerCodes:
fixes stuff like `import os; os(` causing console spam when it gets completions


xonsh/completers/python.py
```diff
@@ -273,7 +273,7 @@ def python_signature_complete(prefix, line, end, ctx, filter_func):
         return set()
     try:
         sig = inspect.signature(val)
-    except ValueError:
+    except (ValueError, TypeError):
```
jnoortheen, line 276: I am not sure this is the correct way. For deprecation warning, one should [filter](https://stackoverflow.com/questions/879173/how-to-ignore-deprecation-warnings-in-python) it in their config
(.xonshrc)

GanerCodes: It's a bug in the autocomplete that will spam the console mid-typing

jnoortheen:
is this the issue you were trying to solve https://github.com/xonsh/xonsh/issues/4721 ?

GanerCodes:
> is this the issue you were trying to solve #4721 ?

it solves one of them yes. The os.fork thing is another thing I should of probs made it's own issue

jnoortheen (approved):

jnoortheen:
Thanks @GanerCodes !

PR #4732, opened 2022-03-24
Add xontrib that brings many useful shortcuts to xonsh.

AsafFisher:
Example:
https://user-images.githubusercontent.com/6796683/159818602-00460771-7965-4745-b12e-233c191afa5a.mp4

### DIFF
--- a/xontrib/oh-my-xonsh.xsh
+++ b/xontrib/oh-my-xonsh.xsh
@@ -1,0 +1,263 @@
+#!/usr/bin/env xonsh
+"""A shameless copy of oh-my-zsh."""
+from xontrib.abbrevs import abbrevs
+
+from packaging import version as ver
+
+# Git version checking
+git_version = $(git version 2>/dev/null).split()[2]
+
+#
+# Functions
+#
+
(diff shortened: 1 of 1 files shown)


jnoortheen:
@AsafFisher this is a case where the names will clash with what one has defined or the binary names. Like other shells it is better installed as an extension. Please create a xontrib for this, it will be helpful to others. 0lease check the template to fastrack the work - https://github.com/xonsh/xontrib-cookiecutter

AsafFisher:
Puttying it in the xontrib folder is not enough?

jnoortheen:
Yes they can be a separate Python packages installable along xonsh with pip. These kind of opinionated names are better kept as a separate package. If some has name clash they can simply update or fork

AsafFisher:
How to I add 
https://github.com/AsafFisher/xontrib-oh-my-xonsh
to the default xontrib list?

jnoortheen:
You can add it to https://github.com/xonsh/xonsh/blob/main/xonsh/xontribs_meta.py and create a PR

PR #4743, opened 2022-03-31
chore: update black to fix CI broken with new click release

gforsyth:
xref: https://github.com/psf/black/issues/2964

TLDR: Click removed an internal module that `black` was relying on and it broke stuff.  Updating `black` to unbreak CI.


jnoortheen:
Thanks @gforsyth

PR #4747, opened 2022-04-09
Pr1 xontrib django

jnoortheen:
## For community


jnoortheen:
It is better to have django-completions separately since it may not be beneficial to ML/sysadmin users

jnoortheen:
@gforsyth can you do a release after merging this? We need to release the four xontribs pulled out of xonsh after that. I can automate releasing those.

anki-code (approved):

PR #4750, opened 2022-04-11
Fix tab-completion trailing space for cmd arguments that use equals sign

### LINKED ISSUE #4749: Trailing space incorrectly appended when tab-completing command arguments that use equals sign
Related to #4746

When completing commands like `dd status=progress`, when I type `dd sta` and hit `TAB`, I get `dd status= ` (with a trailing space). 

There should not be that trailing space because the `dd` command accepts `status=<something>` as a single argument that does not contain spaces.

## xonfig

<details>

```
+------------------+-------------------------+
| xonsh            | 0.11.0                  |
| Git SHA          | 18d655e1                |
| Commit Date      | Apr 10 04:16:56 2022    |
| Python           | 3.9.10                  |
| PLY              | 3.11                    |
| have readline    | True                    |
| prompt toolkit   | 3.0.28                  |
| shell type       | prompt_toolkit          |
| history backend  | sqlite                  |
| pygments         | 2.9.0                   |
| on posix         | True                    |
| on linux         | True                    |
| distro           | fedora                  |
| on wsl           | False                   |
| on darwin        | False                   |
| on windows       | False                   |
| on cygwin        | False                   |
| on msys2         | False                   |
| is superuser     | False                   |
| default encoding | utf-8                   |
| xonsh encoding   | utf-8                   |
| encoding errors  | surrogateescape         |
| xontrib          | []                      |
| RC file 1        | /home/yaxollum/.xonshrc |
+------------------+-------------------------+
```

</details>

## For community

yaxollum:
Fixes #4749 

With this PR, `dd sta` gets completed to `dd status=`, and `date --da` gets completed to `date --date=` (without trailing spaces).

## For community


xonsh/completers/path.py
```diff
@@ -185,7 +185,7 @@ def _quote_paths(paths, start, end, append_end=True, cdpath=False):
         expanded = expand_path(s)
         if os.path.isdir(expanded) or (cdpath and _is_directory_in_cdpath(expanded)):
             _tail = slash
-        elif end == "":
+        elif end == "" and not s.endswith("="):
             _tail = space
```
jnoortheen, line 189: does the completion get generated from path-completer or the bash-completer ?

yaxollum: @jnoortheen it gets generated from the bash-completer:
https://github.com/xonsh/xonsh/blob/bff0794b68f0d9266d8d920f10170c982927ccf6/xonsh/completers/bash_completion.py#L425-L426

jnoortheen: the module bash-completer already has a quote-paths function https://github.com/xonsh/xonsh/blob/09a12e4e85f331d0d281aafacd2d7f69558d7014/xonsh/completers/bash_completion.py#L167  . Is that redundant ? 
also a small question, will this affect other completions ?

yaxollum: > will this affect other completions ?

Good catch! It seems like file paths ending with an equal sign will now be completed without a trailing space (I've added a failing test to verify this)

Perhaps a better approach would be to restore `_quote_paths` to the original and only apply the equal sign fix to `_bash_quote_paths`.

This way, regular paths ending with an equal sign would be correctly completed _with_ a trailing space, and bash command arguments that use an equal sign would be correctly completed _without_ a trailing space.

yaxollum: also, should `tests/test_path_completers.py` be moved to `tests/completers/test_path_completers.py`? (this isn't really related to this PR though)

jnoortheen: yes moving that file will be good.

jnoortheen: > Perhaps a better approach would be to restore _quote_paths to the original and only apply the equal sign fix to _bash_quote_paths.

yes that will be good.

xonsh/completers/bash.py
```diff
@@ -41,7 +40,7 @@ def complete_from_bash(context: CommandContext):
         env=env,
         paths=paths,
         command=command,
-        quote_paths=_quote_paths,
+        quote_paths=_bash_quote_paths,
```
jnoortheen, line 43: you don't have to pass this. it is the default one

jnoortheen (approved):

jnoortheen:
Thanks @yaxollum !

yaxollum:
@jnoortheen thanks for all your help reviewing this!

PR #4758, opened 2022-04-15
fix: guard modal cursor import for prompt toolkit

gforsyth:
We shouldn't fail to start if the available PTK version doesn't support
modal cursor stuff.
(We should probably also make modal cursor stuff optional, but that can
happen later)

## For community

### DIFF
--- a/news/guard_cursor_shape.rst
+++ b/news/guard_cursor_shape.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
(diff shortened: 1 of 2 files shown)


jnoortheen (approved):

jnoortheen:
Looks good @gforsyth !

gforsyth:
woot!  Thanks @jnoortheen -- new release is out!

jnoortheen:
can we change the changelog process we use? something like here https://pdm.fming.dev/dev/contributing/#news-fragments

it will be good to have links to the PRs where we can get more info about the change. 

I think we can get the PR number automatically from commit if issue is not linked in the file name. 
What do you think?

gforsyth:
I'm up for any and all changes to the release process. The more of it that's automated and on GitHub actions, the better

PR #4771, opened 2022-04-21
FIX: pick the value of the field not the field it self

tacaswell:


tacaswell:
This fix is via https://gitter.im/xonsh/xonsh?at=626110b3ddcba117a27ea520

This is the patch I have in my .xonshrc :

```python
def _get_field_value(self, field, **_):
    import sys
    import xonsh.tools as xt
    try:
        return self.fields.pick_val(field)
    except Exception:  # noqa
        print("prompt: error: on field {!r}" "".format(field), file=sys.stderr)
        xt.print_exception()
        value = f"{{BACKGROUND_RED}}{{ERROR:{field}}}{{RESET}}"
    return
import xonsh.prompt.base

xonsh.prompt.base.PromptFormatter._get_field_value = _get_field_value

```

jnoortheen:
Thanks for the report. This function does actually need to pick the object which would be converted to string later when formatted. Closing it in favour of #4772

tacaswell:
Sounds good!

PR #4798, opened 2022-05-06
Returncode enhancements

dev2718:
This PR proposes these changes:

* Added a special '$?' environment variable to access the return code of the last issued command. (Only set during interactive use).
* New prompt-customization fields: ``last_return_code_if_nonzero``, ``last_return_code``.
* The default prompt (on unix-systems) now includes a red [\<errorcode\>] field in case a command failed. This offers a definitive visual cue that something went wrong (commands may produce messy output...). Inspired by the fish shell. See the video for a demonstration.

https://user-images.githubusercontent.com/25764660/167212732-f009a6b1-ff47-4e59-83b7-06db42f58c4b.mp4

## For community

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1554,6 +1554,8 @@ By default, the following variables are available for use:
   * ``localtime``: The current, local time as given by ``time.localtime()``.
     This is formatted with the time format string found in ``time_format``.
   * ``time_format``: A time format string, defaulting to ``"%H:%M:%S"``.
+  * ``last_return_code``: The return code of the last issued command. 
+  * ``last_return_code_if_nonzero``: The return code of the last issued command if it is non-zero, otherwise ``None``. This is useful for only printing the code in case of errors.
 
 .. note:: See the section below on ``PROMPT_FIELDS`` for more information on changing.
 
--- a/news/feat_returncode_enhancements.rst
+++ b/news/feat_returncode_enhancements.rst
@@ -1,0 +1,25 @@
+**Added:**
+
+* Added a special '$?' environment variable to access the return code of the last issued command. (Only set during interactive use).
+* New prompt-customization fields: 'last_return_code_if_nonzero', 'last_return_code'.
+
+**Changed:**
+
+* '$?' is now legal syntax.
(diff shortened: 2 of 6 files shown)


gforsyth (approved):
Thanks for putting this in @dev2718!

bestlem:
How do you get the [<errorcode>] field - that is not normal Prompt syntax. 
You say the default prompt now does this - I have no idea what the default prompt is as I have replaced it long ago (with the xontrib [jnoortheen](https://github.com/jnoortheen) mentions)

In the xontrib there is the field {ret_code_color} which sets the color if the last return code was not zero. How do we do that now?

dev2718:
You can add

`"{RED}{last_return_code_if_nonzero:[{BOLD_INTENSE_RED}{}{RED}] }{RESET}"`

to your `$PROMPT`.

This produces [<code>] in red if the return code was non-zero and nothing otherwise.

bestlem:
I like how things get more complex. The old xontrib was much easier to understand.

Please document this.  Even better provide that as a field.

Also I note the PR says"The default prompt (on unix-systems) now includes a red [<errorcode>] field in case a command failed." why not on Windows as the xontrib did?

Also following from @gforsyth's comment "Few comments -- I'm not totally sold on adding more syntax but I really like the prompt fields" when should code be 
1) in xonsh itself 
2) an xontrib that is maintained comes with xonsh 
3) an xontrib separate from xonsh?

This PR seems to suggest that those xontribs in point 2 should be replaced.

jnoortheen:
having separate field makes it easier for customization. The suggested example can simply be `{last_return_code_if_nonzero:{RED}[{}]}` with brackets being the same color. 

For example, I have a custom field here, https://github.com/jnoortheen/xontrib-powerline3/blob/715d42d625db76dfceb6b3d59ec99668831a1df3/xontrib_powerline3/fields.py#L169 with this PR , things can be reused.

bestlem:
OK that makes sense to me - the lack of documentation is the main issue and use of the simpler example helps.

I don't use Windows but not implementing this on Windows looks like the other issue.

jnoortheen:
This PR works on Windows as well. Contributions are always welcome. I'd be happy to help in PR process

PR #4802, opened 2022-05-09
We can't be sure there's a python binary

stefanor:
Debian doesn't currently ship a `/usr/bin/python` binary, unless a user specifically installs `python-is-python3`.

However, we can safely assume that a `python3` binary exists.

### DIFF
--- a/tests/test_virtualenv_activator.py
+++ b/tests/test_virtualenv_activator.py
@@ -23,13 +23,19 @@ def test_xonsh_activator(tmp_path):
 
     # Sanity
     original_python = check_output(
-        [sys.executable, "-m", "xonsh", "-c", "which python3"]
+        [
+            sys.executable,
+            "-m",
+            "xonsh",
+            "-c",
+            "import shutil; shutil.which('python') or shutil.which('python3')",
+        ]
     ).decode()
     assert Path(original_python).parent != bin_path
 
     # Activate
     venv_python = check_output(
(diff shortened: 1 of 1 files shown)


tests/test_virtualenv_activator.py
```diff
@@ -23,13 +23,13 @@ def test_xonsh_activator(tmp_path):
 
     # Sanity
     original_python = check_output(
-        [sys.executable, "-m", "xonsh", "-c", "which python"]
+        [sys.executable, "-m", "xonsh", "-c", "which python3"]
     ).decode()
```
jnoortheen, line 27: I think this can be improved in general using shutil.which

tests/test_virtualenv_activator.py
```diff
@@ -23,13 +23,13 @@ def test_xonsh_activator(tmp_path):
...
     assert Path(original_python).parent != bin_path
 
     # Activate
     venv_python = check_output(
-        [sys.executable, "-m", "xonsh", "-c", f"source {activate_path}; which python"]
+        [sys.executable, "-m", "xonsh", "-c", f"source {activate_path}; which python3"]
     ).decode()
```
jnoortheen, line 33: here the venv is supposed to have python symlink created

jnoortheen (approved):

jnoortheen:
Thanks @stefanor

Issue #3582, opened 2020-05-21
What is the best way to get shell type in xontrib? Wrong SHELL_TYPE in xontrib. 


anki-code:
Hello! xonsh is super cool every time I try to do something new! Thank you!

1. Create xontrib with:
```python
print(__xonsh__.env.get('SHELL_TYPE'))
# or __xonsh__.shell.shell_type
```
And add the `xontrib load ...` to `~/.xonshrc`.

2. Run xonsh with setting readline shell:
```
bash$ xonsh --shell-type readline
```
3. You will get `best` in the output instead of `readline`.

4. After this run in xonsh:
```
$ print(__xonsh__.env.get('SHELL_TYPE'))
readline
```
You've got the right shell type.

What is the best way to get shell type in xontrib?

Thanks!

laloch:
Hi @anki-code, the question is not how to get the shell type in xontrib, but rather how to load a xontrib that needs to know the shell type :wink:
Since the shell type is decided **after** your RC files are processed according the final `$SHELL_TYPE` value (that's by design), you can't really determine the shell type from the RC files themselves. You can, however, install an `on_post_init` handler and load your xontrib from there.

.xonshrc:
```python
...
echo pre: $SHELL_TYPE

@events.on_post_init
def _rc_post_init():
    echo post: $SHELL_TYPE
    #xontrib load ...
```

anki-code:
It works great! Thanks for explanation!

PR #4817, opened 2022-05-23
xontrib load/unload

jnoortheen:
- [x] implement unloading

## For community


docs/tutorial_xontrib.rst
```diff
@@ -24,15 +24,50 @@ took inspiration from for xonsh:
...
 
-
 Structure
-==========
-Xontribs are modules written in either xonsh (``*.xsh``) or Python (``*.py``).
+================
+Xontribs are modules with some special functions written
+in either xonsh (``*.xsh``) or Python (``*.py``)  that has a couple of special functions to load and unload it.
+
+It is inspired from
```
yaxollum, line 39: @jnoortheen Did you accidentally delete the rest of this sentence?

jnoortheen: I think that is the case, there should be ipython,sphinx ... in the mix

jnoortheen: Ok they are above, I should be removing the hanging sentence

anki-code:
How to create autoloadable xontrib like mentioned in #3275?

jnoortheen:
@anki-code it is not implemented yet. I am planning to do so after this PR. basically we will autoload xontribs with entrypoints, there will be option to control this behaviour .

anki-code:
Cool! Two points around this JFYI:
* probably there is some management of loading priority is needed
* there is `events.on_post_init` event ([example](https://github.com/xonsh/xonsh/issues/3582#issuecomment-632688394)) and probably it will be cool to have something like `__rc_xontrib_post_init` too

anki-code (approved):

jnoortheen:
- Yes I thought about it too. We first implement a naive version and improve upon it, besides the xontribs can utilize events themselves. xontribs should be loaded at `on_pre_rc ` so the rc file can override any environment variables it sets.
- yes it would have a specific event for it to track the time as well

jnoortheen:
Thanks @yaxollum and @anki-code for the reviews

Issue #2976, opened 2018-12-30
Git completion extra characters when completing a remote branch for deletion


jaraco:
Tab completion on a git command duplicates characters. Consider this when attempting to delete a remote branch called `bugfix/tests-require-late-pip` and I type:

```
$ git push origin :bugfix/tests
```

Then hit tab and I see:

```
$ git push origin :bugfix/testsbugfix/tests-require-late-pip
```

## xonfig
```
+------------------+----------------------+
| xonsh            | 0.8.7                |
| Git SHA          | f6a5ac5b             |
| Commit Date      | Dec 27 10:47:21 2018 |
| Python           | 3.7.2                |
| PLY              | 3.11                 |
| have readline    | True                 |
| prompt toolkit   | 2.0.7                |
| shell type       | prompt_toolkit2      |
| pygments         | None                 |
| on posix         | True                 |
| on linux         | False                |
| on darwin        | True                 |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+
```
## Expected Behavior

Only the additional characters needed to complete the branch name should be added to the command-line.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

scopatz:
Thanks for reporting @jaraco! I have noticed this too and it needs a fix for sure

anki-code:
Hi @jaraco! Please try to repeat in xonsh 0.11 and show the result.

jaraco:
It's working great. Problem solved! Whom can I thank for the fix?

gforsyth:
I'll bet that came along when @daniel-shimon rewrote a good chunk of the completion framework.

jaraco:
Oh, dear, I seem to have been mistaken. Maybe I didn't read the repro carefully enough when I attempted to test it before, but I just ran and found the issue still exists:

```
distutils main $ git push origin :debt/debt/collect-msvc9compiler
error: unable to delete 'debt/debt/collect-msvc9compiler': remote ref does not exist
error: failed to push some refs to 'https://github.com/pypa/distutils'
distutils main $ xonfig
+------------------+------------------------+
| xonsh            | 0.11.0                 |
| Git SHA          | 337cf25a               |
| Commit Date      | Nov 17 15:37:41 2021   |
[...]

PR #4826, opened 2022-05-30
Fixing bash completion bug when prefix started with '>', '<', or ':'

jbw3:
This fixes a bug in bash completion when the argument being completed was prefixed with certain characters such as when piping output to a file.

For example, `ls >/dev/nul<TAB>` would complete to `ls >/dev/nu/dev/null` instead of `ls >/dev/null`.

I noticed that issue #2976 was related, so I fixed it as well.

## For community

### DIFF
--- a/xonsh/completers/bash_completion.py
+++ b/xonsh/completers/bash_completion.py
@@ -432,6 +432,10 @@ def bash_completions(
     # to be incorrectly calculated, so it needs to be fixed here
     if "=" in prefix and "=" not in commprefix:
         strip_len = prefix.index("=") + 1
+    # Fix case where remote git branch is being deleted
+    # (e.g. 'git push origin :dev-branch')
+    elif ":" in prefix and ":" not in commprefix:
+        strip_len = prefix.index(":") + 1
 
     return out, max(len(prefix) - strip_len, 0)
 
--- a/xonsh/completers/bash_completion.py
+++ b/xonsh/completers/bash_completion.py
@@ -432,6 +432,16 @@ def bash_completions(
     # to be incorrectly calculated, so it needs to be fixed here
     if "=" in prefix and "=" not in commprefix:
         strip_len = prefix.index("=") + 1
+    # Fix cases where stdout/stderr is being redirected (e.g. 'ls >out.txt')
+    elif ">" in prefix and ">" not in commprefix:
+        strip_len = prefix.rindex(">") + 1
+    # Fix cases where stdin is being redirected (e.g. 'cat <in.txt')
+    elif "<" in prefix and "<" not in commprefix:
+        strip_len = prefix.rindex("<") + 1
+    # Fix case where remote git branch is being deleted
+    # (e.g. 'git push origin :dev-branch')
+    elif ":" in prefix and ":" not in commprefix:
+        strip_len = prefix.index(":") + 1
 
     return out, max(len(prefix) - strip_len, 0)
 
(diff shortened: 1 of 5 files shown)


jbw3:
Looks like one of the new tests I added is failing. I was trying to test `git push origin :mai<TAB>` to make sure it completed to `git push origin :main`, but it doesn't look like it worked on the CI machine. If anyone has an idea on how to write a better test for this, I'd appreciate it.

jnoortheen:
> Looks like one of the new tests I added is failing. I was trying to test `git push origin :mai<TAB>` to make sure it completed to `git push origin :main`, but it doesn't look like it worked on the CI machine. If anyone has an idea on how to write a better test for this, I'd appreciate it.

you might want to mock the output from bash_completer subprocess call. You can use [this approach ](https://github.com/xonsh/xonsh/blob/347bf201de3687b8d2a919270be4e6bcb2872613/tests/xontribs/test_fish_completer.py#L5) to fake the output so that it will run on systems without proper bash setup

xonsh/completers/bash_completion.py
```diff
@@ -432,6 +432,16 @@ def bash_completions(
     # to be incorrectly calculated, so it needs to be fixed here
     if "=" in prefix and "=" not in commprefix:
         strip_len = prefix.index("=") + 1
+    # Fix cases where stdout/stderr is being redirected (e.g. 'ls >out.txt')
```
jnoortheen, line 435: instead of doing this by hand, have you tried using the xonsh parser or perhaps update completion parser ?

will this approach work when the arg has > inbetween ?

jbw3:
> > Looks like one of the new tests I added is failing. I was trying to test `git push origin :mai<TAB>` to make sure it completed to `git push origin :main`, but it doesn't look like it worked on the CI machine. If anyone has an idea on how to write a better test for this, I'd appreciate it.
> 
> you might want to mock the output from bash_completer subprocess call. You can use [this approach ](https://github.com/xonsh/xonsh/blob/347bf201de3687b8d2a919270be4e6bcb2872613/tests/xontribs/test_fish_completer.py#L5) to fake the output so that it will run on systems without proper bash setup

Thanks! I will do that.

xonsh/completers/bash_completion.py
jbw3: My current logic should work when '>' is in the middle, but I like your idea of using the parser better. I will look into that.

jbw3:
@jnoortheen I made changes to the update completion parser to handle the IO redirect tokens. I didn't add any code to the update parser to handle `:`. I wasn't sure if I should since `:` is really not a separate argument token.

jnoortheen:
Yes ":" and the such symbols should not be tokenized. I am not familiar with how bash-completer works, but this ":" should be sent as part of the args to the completer and it should be handled by the underlying bash process.

jbw3:
> Yes ":" and the such symbols should not be tokenized. I am not familiar with how bash-completer works, but this ":" should be sent as part of the args to the completer and it should be handled by the underlying bash process.

@jnoortheen Sounds good. Everything should be working now.

jnoortheen:
this seems a little hacky solution. I usually add space around operators and I would like these kind of nuances to be handled by the underlying bash process instead of doing it on xonsh side.

jbw3:
I believe bash is doing the correct thing here. The problem seems to come from how something like `>file` is parsed. Bash correctly splits this into 2 tokens (`>` and `file`) and correctly does the completion. However, xonsh was previously just treating it as one token (`>file`), so it did not correctly replace what it got from bash. Note that xonsh runs commands like `ls >out.txt` correctly, so there seems to be a difference in how xonsh parses a command to run it, and how it parses commands for completion. All that to say that I think this is really a problem with xonsh. However, if there is a better way to solve this problem, I'm open to suggestions.

jnoortheen (approved):

jnoortheen:
@jbw3 thanks for tackling this!

PR #4827, opened 2022-06-04
Pr5 xontrib entrypoints

jnoortheen:
- [x] auto-load xontribs
- [x] `xontrib list` discover

## For community


anki-code:
Great PR! Please add the example of autoloadable xontrib code to [xontrib-template](https://github.com/xonsh/xontrib-template).

jnoortheen:
> Great PR! Please add the example of autoloadable xontrib code to [xontrib-template](https://github.com/xonsh/xontrib-template).

Sure I will add later. Have you checked this https://github.com/xonsh/xontrib-template/pull/13 ?

anki-code:
Done!

gforsyth:
Sorry for the delay in getting to this @jnoortheen !  This looks good to me, I'm going to play with it locally for a bit, but then I'll merge it in

jnoortheen:
@gforsyth this seems like the special case mentioned here https://docs.python.org/3/library/importlib.html#importlib.util.find_spec 

Could you use something like https://github.com/Qix-/better-exceptions ? so that we can see what module is causing the issue

jnoortheen:
> I should note that the `conda` xontrib is still _working_, it just breaks the functionality of `list`

could you point to how to reproduce this?

jnoortheen:
Ok I will try.
Here `_sys.modules["xontrib.conda"] = _mod` the module should have a __spec__ attribute. Otherwise we can skip such modules without actual __spec__ attribute

gforsyth:
Yeah, it only has a `__spec__` and that is `None`

jnoortheen:
@gforsyth , I have pushed the fix. In the meantime, is there any need for the snippet to load the conda xontrib this way? what about `xontrib load conda` ?

jnoortheen:
Ok. Please make a point release after this. I'd like to update the xontrib-template to use auto-loading.

gforsyth:
Fixes look good.  What do you think about adding a column to the `xontrib list` output that indicates whether a given xontrib has been auto- vs. manually loaded?

jnoortheen:
that will be helpful. I will add it.

jnoortheen:
@anki-code you can check this https://github.com/jnoortheen/xontrib-avox-poetry for example as of now

xonsh/xontribs.py
```diff
@@ -236,16 +324,48 @@ def _list(
...
+            s += "{PURPLE}" + name + "{RESET}  " + " " * (nname - len(name))
             if d["loaded"]:
-                s += "{GREEN}loaded{RESET}"
+                s += "{GREEN}loaded{RESET}" + " " * 4
             else:
                 s += "{RED}not-loaded{RESET}"
+            s += "  "
+            if d["auto"]:
+                s += "{GREEN}auto{RESET}"
+            else:
```
gforsyth, line 335:
```suggestion
            elif d["loaded"]:
```

I think this might be a bit better visually -- so we only display `auto` or `manual` for a given `xontrib` when it's loaded

jnoortheen: you mean having only two columns and we print auto-loaded and manually-loaded ?

gforsyth: No, I mean we should leave the row blank if the xontrib isn't loaded at all, e.g.

```
abbrevs             loaded      manual
autovox             not-loaded  
bashisms            not-loaded  
conda               loaded      manual
coreutils           not-loaded  
direnv              loaded      manual
fish_completer      not-loaded  
free_cwd            not-loaded  
goenv               not-loaded  
langenv_common      not-loaded  
nodenv              not-loaded  
pdb                 not-loaded  
prompt_ret_code     not-loaded  
prompt_starship     not-loaded  
pyenv               not-loaded  
rbenv               not-loaded  
term_integration    not-loaded  
vox                 not-loaded  
voxapi              not-loaded  
whole_word_jumping  not-loaded  
xog                 not-loaded 
```

jnoortheen: I have moved the code to run when it is loaded. so that it will work when auto-loading is disabled

xonsh/environ.py
```diff
@@ -1197,6 +1198,17 @@ def tracer(cmds: list, captured: Union[bool, str]):
         "or None / the empty string if traceback logging is not desired. "
         "Logging to a file is not enabled by default.",
     )
+    XONTRIBS_AUTOLOAD_DISABLED = Var(
```
gforsyth, line 1201: I think it's probably clearer to name this `XONTRIBS_AUTOLOAD` or `XONTRIBS_AUTOLOAD_ENABLED` and default it to `True`

jnoortheen: Yes I thought about that naming, when setting this to a list of names, it has to make sense.

gforsyth: Ahh, that's a good point.  Ok, all set!

gforsyth (approved):

jnoortheen:
It is a different name under poetry 
https://github.com/jnoortheen/xontrib-avox-poetry/blob/db39939e2a8a6b5b5495fa7d5ee0861022a75233/pyproject.toml#L46

jnoortheen:
Yes sure I will add it next.

jnoortheen:
It is possible, with another package for setting that entrypoint. I won't recommend it though, The extensions should stick to one style. The user can stop auto-loading it with the given environment variable any time

jnoortheen:
The environment variable accepts a list of names as well. That way it can be not loaded. Also things like completers, shortcuts should exit when it is not interactive.

PR #4835, opened 2022-06-11
feat: f-glob strings

### LINKED ISSUE #4545: Feat: add f-glob strings
In Xonsh we have augmented string literals with `p`-strings (that return a Path object) and glob strings g`...` (that return a list of results). It would seem that we want glob strings to behave similarly to normal strings, and so there is a case to be made for supporting `f`-strings in this context.

An example:
```xonsh
>>> ls
file-a  file-b
>>> prefix = "file"
>>> echo fg`{prefix}*`
file-a file-b
```

## For community

yaxollum:
Closes #4545 

## For community

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1054,6 +1054,23 @@ mode or subprocess mode) by using the ``g````:
     5
 
 
+Formatted Glob Literals
+-----------------------
+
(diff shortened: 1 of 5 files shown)


jnoortheen (approved):

jnoortheen:
Looks good!

PR #4836, opened 2022-06-12
Fixed the bug in the "timeit" function

amukher3:
## For community


news/branch.rst
```diff
@@ -0,0 +1,25 @@
...
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <for _i in _it:
```
jnoortheen, line 16: the content added here will end up in the release notes. Please make it more descriptive to tell what is exactly changed/added

amukher3: Ok. Sure.

xonsh/timings.py
```diff
@@ -172,8 +172,8 @@ def timeit(self, number=timeit.default_number):
 def inner(_it, _timer):
     #setup
     _t0 = _timer()
-    for _i in _it:
-        {stmt}
+    for _i in range(1, {stmt}):
+        pass
     _t1 = _timer()
```
jnoortheen, line 177: what does it solve?

amukher3: the "timeit" command from the shell was showing syntax error, I think it seems to work fine after the change

jnoortheen: Ok it will be good to add a test for that. Please check the folder tests for similarly written test cases.

amukher3: I see. Ok.

PR #4862, opened 2022-07-01
move ptk-shell out

jnoortheen:
This will be moved to https://github.com/xonsh/xontrib-ptk-shell. 

## Pros:

- no need for version compatibility checks inside xonsh code
- makes xonsh modular to give interface for integration with other line editing libs

## For community


gforsyth:
I'll write more later (sorry, work is very busy right now), but I'm generally in favor of making things more modular and decoupling the back ends. 

One thing I'm not sure about is whether it's worth the overhead of putting the decoupled package in a separate repo -- it tends to make CI much harder and also trickier for new contributors.

jnoortheen:
> tends to make CI much harder

If you mean the release process, it would be automated. Splitting the packages helps reducing the CI run time as well.

> also trickier for new contributors

Yes I also thought about that. We can add more docs on these core projects to balance this. 

It is just that, ptk is memory hungry (almost like a full blown GUI app with PyQt; 25+MB). But it is featureful. Splitting this, it can serve as an example for integration with other rich shell libraries.

jnoortheen:
> So I suggest: xontrib-shell-ptk

This is just following `xontrib-jupyter-shell` xontrib name.

jnoortheen:
Yes sure lets wait for others opinion. Having separate package makes easier for it to be installed with package managers and as I have mentioned reduced testing time. The shell backend doesn't need to wait for parser tests. It should pass the abstractions like any other packages. This way we can find any shortcomings in the abstraction as well.

gforsyth:
> Having separate package makes easier for it to be installed with package managers and as I have mentioned reduced testing time.

I think I'm still not being clear -- I think we can have the prompt toolkit backend as a separate package, with a separate `pyproject.toml`, separate pypi record, etc, but have it in this repository

gforsyth:
> Do you mean a mono-repo with workspaces like?

Yeah.  And mono-repos can _very_ quickly get out of control but it does make for simpler upstream / downstream testing

jnoortheen:
> mono-repos can very quickly get out of control 

Yes I see a lot of redundant github-actions, flake8, mypy configs etc.,

jnoortheen:
regarding the testing, we can test against latest main branch at scheduled interval.

jnoortheen:
I am closing this PR. I will move it as a package in the same repo as xonsh later.

gforsyth:
There is a distinction between modules being decoupled and modules living in a separate repository.  

A separate repository enforces the separation, at the cost of a huge amount of extra overhead for coordinating CI and publishing.



> to just execute commands

I think you'll find that the "just" there is quite complicated, but a no-dependency prompt interface is certainly welcome.
Again, it doesn't need to live in a separate repository

gforsyth:
> in the future I would prefer to have an ability to use another prompt backend.

There's the `readline` backend which can be used independently of the prompt_toolkit backend.

Prompt toolkit itself is an optional dependency (albeit one we strongly encourage using).

When you say "remove from core", if the intention is to separate something into a module with isolated dependencies, then that is definitely OK.  If it means removing it from `xonsh/xonsh` then that is a non-starter for me.

PR #4901, opened 2022-07-30
Fix job control for callable aliases

### LINKED ISSUE #4277: Using captured stdout breaks the piping from the callable alias
Hi!

The captured subprocs breaks the piping from the callable alias.

### Case 1: piping into capturable command

This works as expected:
```python
xonsh --no-rc

def _callme(args):
    print('one\ntwo\nthree')

aliases['callme'] = _callme
callme | grep t
# two
# three

```

But the same code with added the one line:

```python
xonsh --no-rc

def _callme(args):
    $(echo 1)
    print('one\ntwo\nthree')

aliases['callme'] = _callme
callme | grep t
# No response, terminal freezed, Ctrl+C Ctrl+D Ctrl+Z have no effect
```

Workaround is using `!()`:
```python
def _callme(args, stdin=None, stdout=None):
    out = !(echo -n "one\ntwo\nthree")
    for line in out:
        stdout.write(line + '\n')

aliases['callme'] = _callme
callme | grep t
# two
# three
```

### Case 2: piping into uncapturable command

We want to `less` many lines and this is working:
```python
def _lines(cmd, stdin=None, stdout=None, stderr=None):
    for line in [str(i) for i in range(1,100)]:
        stdout.write(line+'\n')

aliases['lines'] = _lines
lines | less
# Less allows to scroll lines
```
But the same code with added the one line with `$()`:
```python
def _lines(cmd, stdin=None, stdout=None, stderr=None):
    $(echo 0)  # Just added this line without using it
    for line in [str(i) for i in range(1,100)]:
        stdout.write(line+'\n')

aliases['lines'] = _lines
lines | less
# Freezed
```
And the same code with added the one line with `![]`:
```python
def _lines(cmd, stdin=None, stdout=None, stderr=None):
    ![echo 0]  # Just added this line without using it
    for line in [str(i) for i in range(1,100)]:
        stdout.write(line+'\n')

aliases['lines'] = _lines
lines | less
# Printing first part of lines and less is not working.
```
And the same code with added the one line with `$[]`:
```python
def _lines(cmd, stdin=None, stdout=None, stderr=None):
    $[echo 0]  # Just added this line without using it
    for line in [str(i) for i in range(1,100)]:
        stdout.write(line+'\n')

aliases['lines'] = _lines
lines | less
# Print 0
# Freezed
```

### Case 3: Using uncaptured subprocess breaks the piping from the callable alias with the error

See #4281

## For community

### LINKED ISSUE #4281: Using uncaptured subprocess breaks the piping from the callable alias
Hi! 

Using uncaptured subprocess breaks the piping from the callable alias with the error:

```python
def _qwe(args, stdin):
    $[echo 1]
    $[echo 2]
    $[echo 3]
    return 0
    
aliases['qwe'] = _qwe

echo pipe | qwe
# Traceback (most recent call last):
#   File "/opt/miniconda3/lib/python3.8/site-packages/xonsh/procs/proxies.py", line 514, in run
#     r = self.f(self.args, sp_stdin, sp_stdout, sp_stderr, spec, spec.stack)
#   File "/opt/miniconda3/lib/python3.8/site-packages/xonsh/procs/proxies.py", line 300, in proxy_two
#     return f(args, stdin)
#   File "<xonsh-code>", line 2, in _qwe
#   File "/opt/miniconda3/lib/python3.8/site-packages/xonsh/built_ins.py", line 208, in subproc_uncaptured
#     return xonsh.procs.specs.run_subproc(cmds, captured=False, envs=envs)
#   File "/opt/miniconda3/lib/python3.8/site-packages/xonsh/procs/specs.py", line 901, in run_subproc
#     command.end()
#   File "/opt/miniconda3/lib/python3.8/site-packages/xonsh/procs/pipelines.py", line 452, in end
#     self._end(tee_output=tee_output)
#   File "/opt/miniconda3/lib/python3.8/site-packages/xonsh/procs/pipelines.py", line 460, in _end
#     for _ in self.tee_stdout():
#   File "/opt/miniconda3/lib/python3.8/site-packages/xonsh/procs/pipelines.py", line 362, in tee_stdout
#     for line in self.iterraw():
#   File "/opt/miniconda3/lib/python3.8/site-packages/xonsh/procs/pipelines.py", line 247, in iterraw
#     task = xj.wait_for_active_job()
#   File "/opt/miniconda3/lib/python3.8/site-packages/xonsh/jobs.py", line 210, in wait_for_active_job
#     _, wcode = os.waitpid(obj.pid, os.WUNTRACED)
# TypeError: an integer is required (got type NoneType)

```

## For community

yaxollum:
Fixes #4277
Fixes #4281

## Problem
Piping a callable alias containing a subprocess call into another command is currently broken. This is because the job control inside the callable alias gets confused by the main thread's job control. For example, the following example is given in #4277:
```xonsh
def _callme(args):
    $(echo 1)
    print('one\ntwo\nthree')

aliases['callme'] = _callme
callme | grep t
```
After the callable alias `callme` spawns the `echo 1` command, it calls `wait_for_active_job`:
https://github.com/xonsh/xonsh/blob/1e20ee04c83389500a6328262af83982e35e865e/xonsh/jobs.py#L200-L230
The first active job is `echo 1`. `wait_for_active_job` correctly waits for `echo 1` to complete, and then it starts waiting for the *next* active job, which it determines is the `grep t` command. This causes a deadlock because the `grep` command is still waiting for more input from `callme`, but `callme` is stuck inside the `wait_for_active_job` function waiting for `grep` to finish.

## Solution
This PR solves this problem by separating callable alias' job control from the main thread's job control. It makes the job control variables `jobs` and `tasks` thread-local. This way, callable aliases can manage their own jobs without getting affected by what is happening on the main thread. In the `callme | grep t` example, the `wait_for_active_job` call inside `callme` would wait for `echo 1` to complete, and since there would be no other jobs on the thread, it would return. This would avoid the deadlock.

An issue with making the job control thread-local is that threaded commands such as `jobs`, `bg`, and `disown` stop working because they need access to the main thread's job control. This is solved using the `use_main_jobs` context manager, which temporarily replaces a thread's job control information with that of the main thread.

Apart from the callable alias job control issue, there are two other minor issues:

1. (Addressed in this PR) In case 2 of #4277, the `less` command doesn't work because xonsh uses `give_terminal_to` inside the callable alias to give the terminal to the `echo` command. This takes the terminal away from the `less` command, causing it to stop. [This issue is addressed in this PR by making `echo` run in the same process group as xonsh.](https://github.com/xonsh/xonsh/commit/6ef2084448b83d974f17578d828336827724efcb)
2. (Not addressed in this PR) When `$[]` is used inside a callable alias, it does not properly redirect its output to the callable alias' output. [This is demonstrated in the integration test that I have marked as `xfail`](https://github.com/xonsh/xonsh/blob/52b4bacec276117e544acabd906612834dabea58/tests/test_integrations.py#L624-L642). I'll also make a separate issue for this.

## For community

### DIFF
--- a/news/fix-callable-alias-subproc.rst
+++ b/news/fix-callable-alias-subproc.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* A callable alias containing subprocess commands no longer freezes when piped to another command
+* ``less`` no longer stops when a callable alias containing subprocess commands is piped into it
+
+**Security:**
+
+* <news item>
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -588,6 +588,58 @@ xexec python3 -c 'import os; print(os.environ["SHLVL"])' # == 4
 """,
         0,
     ),
+    # test $() inside piped callable alias
+    (
+        r"""
+def _callme(args):
+    result = $(python -c 'print("tree");print("car")')
+    print(result[::-1])
+    print('one\ntwo\nthree')
+
+aliases['callme'] = _callme
+callme | grep t
+""",
+        """eert
+two
+three
+""",
+        0,
+    ),
+    # test ![] inside piped callable alias
+    (
+        r"""
+def _callme(args):
+    python -c 'print("tree");print("car")'
+    print('one\ntwo\nthree')
+
+aliases['callme'] = _callme
+callme | grep t
+""",
+        """tree
+two
+three
+""",
+        0,
+    ),
+    # test $[] inside piped callable alias
+    pytest.param(
+        (
+            r"""
+def _callme(args):
+    $[python -c 'print("tree");print("car")']
+    print('one\ntwo\nthree')
+
+aliases['callme'] = _callme
+callme | grep t
+""",
+            """tree
+two
+three
+""",
+            0,
+        ),
+        marks=pytest.mark.xfail(reason="$[] does not send stdout through the pipe"),
+    ),
 ]
 
 if not ON_WINDOWS:
--- a/tests/test_jobs.py
+++ b/tests/test_jobs.py
@@ -31,6 +31,6 @@ def test_disown_completion(
     }
     all_jobs = {1: job, 2: job}
 
-    monkeypatch.setattr(xsh_with_aliases, "all_jobs", all_jobs)
-    monkeypatch.setattr(jobs, "tasks", [2, 1])
+    monkeypatch.setattr(jobs._jobs_thread_local, "jobs", all_jobs, raising=False)
+    monkeypatch.setattr(jobs._jobs_thread_local, "tasks", [2, 1], raising=False)
     assert check_completer(args, prefix=prefix) == exp
--- a/tests/test_pipelines.py
+++ b/tests/test_pipelines.py
@@ -13,9 +13,9 @@ from xonsh.pytest.tools import skip_if_on_unix, skip_if_on_windows
 
 @pytest.fixture(autouse=True)
 def patched_events(monkeypatch, xonsh_events, xonsh_session):
-    from xonsh.jobs import tasks
+    from xonsh.jobs import get_tasks
 
-    tasks.clear()
+    get_tasks().clear()
     # needed for ci tests
     monkeypatch.setitem(
         xonsh_session.env, "RAISE_SUBPROC_ERROR", False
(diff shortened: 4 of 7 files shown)


jnoortheen:
What do you think about calling the aliases using multiprocessing module? That way it will have true STDIN/OUT, also other shells fork the parent in these scenarios

yaxollum:
Yeah, I think using multiprocessing would be a better solution for most callable aliases, but some aliases might want to share data with the main thread, so should we still support threaded aliases? For example,
```xonsh
count=0

def _foo(cmd, stdin=None, stdout=None, stderr=None):
    global count
    count+=1
    print(count)

aliases['foo'] = _foo
foo # 1
foo # 2
foo # 3
```

jnoortheen:
Using globals seems anti-pattern of shell scripting. Anyways, we will conitue with this until a robust pattern is established. Thanks for tackling this @yaxollum

jnoortheen (approved):

PR #4917, opened 2022-08-09
We can't be sure there's a python binary

stefanor:
Debian doesn't currently ship a /usr/bin/python binary, unless a user
specifically installs python-is-python3.

See also: #4802

## For community


jnoortheen:
Looks good! Thanks @stefanor @anki-code

jnoortheen (approved):

PR #4879, opened 2022-07-14
fix(xpip): never add `sudo`, append `--user` instead

gforsyth:
Fixes #4878 
Fixes #4369 

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #4922, opened 2022-08-18
Fix: xpip return None instead of command

anki-code:
Fix #4879 because `basecmd.extend` returns `None` ;)

## For community

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -815,7 +815,7 @@ def detect_xpip_alias():
             # XXX: Does windows have an installation mode that requires UAC?
             return basecmd
         elif not os.access(os.path.dirname(sys.executable), os.W_OK):
-            return basecmd + ["--user"]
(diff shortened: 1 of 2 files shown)


jnoortheen:
pre-commit.ci autofix

xonsh/aliases.py
```diff
@@ -815,7 +815,7 @@ def detect_xpip_alias():
             # XXX: Does windows have an installation mode that requires UAC?
             return basecmd
         elif not os.access(os.path.dirname(sys.executable), os.W_OK):
-            return basecmd.extend(["--user"])
+            return sys.executable + " -m pip @(['install', '--user'] + $args[1:] if $args and $args[0] == 'install' else $args)"
```
jnoortheen: it is better to return a list instead of string, it would be much faster and doesn't involve parser.

jnoortheen (approved):

jnoortheen:
Thanks @anki-code !

PR #4927, opened 2022-08-24
commands_cache: fix `_update_cmds_cache()`

yotamolenik:
Prevent adding the same path to `allcmds` when `key` already exists.
This could result is wrong command resolution.

## For community

### DIFF
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -197,7 +197,8 @@ class CommandsCache(cabc.Mapping):
 
         for cmd, path in self._get_all_cmds(paths).items():
             key = cmd.upper() if ON_WINDOWS else cmd
-            allcmds[key] = (path, aliases.get(key, None))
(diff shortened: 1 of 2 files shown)


xonsh/commands_cache.py
```diff
@@ -197,7 +197,8 @@ def _update_cmds_cache(
 
         for cmd, path in self._get_all_cmds(paths).items():
             key = cmd.upper() if ON_WINDOWS else cmd
-            allcmds[key] = (path, aliases.get(key, None))
+            if key not in allcmds:
+                allcmds[key] = (path, aliases.get(key, None))
```
jnoortheen, line 201: this means any override will not work. If Python is overridded by virtualenv, it will not have the updated path.

yotamolenik:
after further testing, I now understand i still have a bug but it's not in the _update_cmds_cache() function. I will try to find a good example and a better solution this week

PR #4936, opened 2022-09-11
ci: add minimum GitHub token permissions for workflow

boahc077:
### Description
This PR adds minimum token permissions for the GITHUB_TOKEN in GitHub Actions workflows using https://github.com/step-security/secure-workflows.

The GitHub Actions workflow has a GITHUB_TOKEN with write access to multiple scopes.
Here is an example of the permissions in one of the workflow runs:
https://github.com/xonsh/xonsh/runs/8196887287?check_suite_focus=true#step:1:19

After this change, the scopes will be reduced to the minimum needed for the following workflow:

- docs.yml

The following workflows already have the least privileged token permission set:

- check-news-item.yml
- publish.yml
- test.yml

### Motivation and Context

- This is a security best practice, so if the GITHUB_TOKEN is compromised due to a vulnerability or compromised Action, the damage will be reduced.
- GitHub recommends defining minimum GITHUB_TOKEN permissions.
  https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token
- The Open Source Security Foundation (OpenSSF) [Scorecards](https://github.com/ossf/scorecard) also treats not setting token permissions as a high-risk issue. This change will help increase the Scorecard score for this repository.

Signed-off-by: Ashish Kurmi <akurmi@stepsecurity.io>

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -11,6 +11,9 @@ on:
       - main
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   build:
     if: github.repository_owner == 'xonsh'


jnoortheen:
@gforsyth will this affect [deploying the docs](https://github.com/xonsh/xonsh/blob/c0072d9c4f823c84b594b92680c9f1a5af612e6d/.github/workflows/docs.yml#L37) to doc repo - `xonsh/xonsh-docs`  ?

gforsyth:
I'm not sure but it looks like it might. Also, the tokens in use here are generated per-workflow run so I don't think the tokens being compromised is as much of a concern

boahc077:
@jnoortheen @gforsyth the security concern here is that the default GitHub access token provided by GitHub has all the write permissions as shown [here](https://github.com/xonsh/xonsh/runs/8196887287?check_suite_focus=true#step:1:19). This is a risk from a supply chain security perspective as a compromised action in the CI/CD can misuse this token and overwrite source code. The Open Source Security Foundation (OpenSSF) flags this as a high risk. The other three workflow files in the repository are already operating with the least privileged GitHub token so it it makes sense to update _docs.yml_ with the least privileged GitHub token as well.

IMO, this change should not have an impact on deploying the doc as that action accepts a different PAT as per the workflow definition: https://github.com/xonsh/xonsh/blob/c0072d9c4f823c84b594b92680c9f1a5af612e6d/.github/workflows/docs.yml#L41

gforsyth:
Thanks for the fix and the explanation, @boahc077 !

gforsyth (approved):

Issue #3279, opened 2019-08-21
Question: is it possible to exclude certain commands from being added to history


manfredlotz:
## xonfig

<details>

```
$ xonfig
```

</details>

### Traceback (if applicable)

<details>

```
traceback
```

</details>

scopatz:
There isn't a good method for doing this right now, other than setting the history size to zero.  We are open to options here

Meptl:
One option is a custom history backend. It's hacky though and would break if the current backend moves around. I haven't used this code since last year so I'm not sure if it still works. It prevents placing the command to disk, but I believe the command would still be available when moving up and down history (pressing the up and down characters):
```
from xonsh.history.json import JsonHistory

_filtered = ('ls', 'cat', 'rg', 'ag', 'grep', 'echo', 'print', 'man',
             'alsamixer', 'dmesg', 'light', 'reboot', 'shutdown',
             # This also filters git stash
             'git commit', 'git st')

class FilteredJsonHistory(JsonHistory):
    _filtered = _filtered

    def flush(self, at_exit=False):
        """Don't flush filtered commands to disk."""
        if len(self.buffer) == 0:
            return
        self.buffer = list(filter(lambda cmd: not cmd['inp'].lstrip().startswith(self._filtered), self.buffer))
        super().flush(at_exit)

$XONSH_HISTORY_BACKEND = FilteredJsonHistory
```

manfredlotz:
First, thanks for providing an idea. I didn't post anything because on the one hand I'm a Python beginner, and on the other hand I don't know the xonsh code at all.

Are there events in history processing where some code could be attached to?

A bit related: it just occured to me that other shells (for instance {z,fi,ba}sh offer to not include a command into history when it will be preceded by at least one blank. But I don't think it would be a solution to artificially prepend the commands in a filtered list with a blank.

Meptl:
> I don't know the xonsh code at all.
Me neither! Fortunately there was https://xon.sh/tutorial_history_backend.html available.

> Are there events in history processing where some code could be attached to?
I'll leave this for others to answer. My answer is "No" just following JsonHistory's flush method and the list of events in https://xon.sh/events.html 
Can also grab the latest events using `grep -RA1 events.doc .` on the source code.

manfredlotz:
Thanks for the backend link. Yes, I think you are right that there is no possibility to work with events. 

[...]

PR #4953, opened 2022-10-02
Add XONSH_HISTORY_IGNORE_REGEX support

bcr:
This is a fix for #3279 using the strategy proposed by @anki-code with some slight modifications.

I will add unit tests and proposed changelog text, I just want to make sure this is generally the right direction. Works on my machine with the JSON implementation 😆 

## For community

### DIFF
--- a/tests/test_history_dummy.py
+++ b/tests/test_history_dummy.py
@@ -16,19 +16,12 @@ def test_ignore_regex_invalid(xession, capsys):
     xession.env["XONSH_HISTORY_BACKEND"] = "dummy"
     xession.env["XONSH_HISTORY_IGNORE_REGEX"] = "**"
     history = construct_history()
-    assert history.history_ignore_regex is None
     captured = capsys.readouterr()
     assert (
         "XONSH_HISTORY_IGNORE_REGEX is not a valid regular expression and will be ignored"
         in captured.err
     )
-
-
-def test_ignore_regex_valid(xession):
-    xession.env["XONSH_HISTORY_BACKEND"] = "dummy"
-    xession.env["XONSH_HISTORY_IGNORE_REGEX"] = "ls"
-    history = construct_history()
-    assert history.history_ignore_regex is not None
+    assert not history.is_ignored({"inp": "history"})
 
 
 def test_is_ignore(xession):
@@ -38,3 +31,11 @@ def test_is_ignore(xession):
     assert history.is_ignored({"inp": "cat foo.txt"})
     assert not history.is_ignored({"inp": "history"})
     assert history.is_ignored({"inp": "ls bar"})
+
+
+def test_is_ignore_no_regex(xession):
+    xession.env["XONSH_HISTORY_BACKEND"] = "dummy"
+    history = construct_history()
+    assert not history.is_ignored({"inp": "cat foo.txt"})
+    assert not history.is_ignored({"inp": "history"})
+    assert not history.is_ignored({"inp": "ls bar"})
--- a/tests/test_history_dummy.py
+++ b/tests/test_history_dummy.py
@@ -10,3 +10,19 @@ from xonsh.history.main import construct_history
 def test_construct_history_str(xession, backend):
     xession.env["XONSH_HISTORY_BACKEND"] = backend
     assert isinstance(construct_history(), DummyHistory)
+
+
+@pytest.mark.parametrize("backend", ["dummy", DummyHistory, DummyHistory()])
+def test_ignore_regex_invalid(xession, backend):
+    xession.env["XONSH_HISTORY_BACKEND"] = backend
+    xession.env["XONSH_IGNORE_REGEX"] = "**"
+    history = construct_history()
+    assert history.history_ignore_regex is None
+
+
+@pytest.mark.parametrize("backend", ["dummy", DummyHistory, DummyHistory()])
+def test_ignore_regex_valid(xession, backend):
+    xession.env["XONSH_HISTORY_BACKEND"] = backend
+    xession.env["XONSH_IGNORE_REGEX"] = "ls"
+    history = construct_history()
+    assert history.history_ignore_regex is not None
--- a/tests/test_history_dummy.py
+++ b/tests/test_history_dummy.py
@@ -29,3 +29,12 @@ def test_ignore_regex_valid(xession):
     xession.env["XONSH_HISTORY_IGNORE_REGEX"] = "ls"
     history = construct_history()
     assert history.history_ignore_regex is not None
+
+
+def test_is_ignore(xession):
+    xession.env["XONSH_HISTORY_BACKEND"] = "dummy"
+    xession.env["XONSH_HISTORY_IGNORE_REGEX"] = "(ls|cat)"
+    history = construct_history()
+    assert history.is_ignored({"inp": "cat foo.txt"})
+    assert not history.is_ignored({"inp": "history"})
+    assert history.is_ignored({"inp": "ls bar"})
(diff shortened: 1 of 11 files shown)


xonsh/history/json.py
```diff
@@ -445,6 +446,12 @@ def __init__(
             if save_cwd is not None
             else XSH.env.get("XONSH_HISTORY_SAVE_CWD", True)
         )
+        ignore_regex = XSH.env.get("XONSH_HISTORY_IGNORE_REGEX")
```
jnoortheen, line 449: this better be moved to the historybackend base class.

xonsh/history/base.py
```diff
@@ -73,6 +73,7 @@ def __init__(self, sessionid=None, **kwargs):
         self.hist_size = None
         self.hist_units = None
         self.remember_history = True
+        self.should_append = lambda cmd: True
```
jnoortheen, line 76: this can be a regular method. It will be more explicit and easier to extend if needed.

jnoortheen: I also suggest naming the method something like, `is_ignored(cmd) -> bool`

xonsh/history/json.py
bcr: The discussion in #3279 was explicit about not putting it in the backend. I personally don't have any stake in this, I'll put it wherever you want, but if you and @anki-code don't agree, I'm not sure what to do about that.

xonsh/history/base.py
```diff
@@ -164,3 +165,20 @@ def clear(self):
...
+        Call this in your append method.
+
+        Parameters
+        ----------
+        cmd: The prospective item to append.
+
+        Returns
+        -------
+        True if the item should be appended, False if not.
+        """
```
jnoortheen, line 180: please check the [numpy-docstring](https://numpydoc.readthedocs.io/en/latest/format.html#parameters) format for parameters. The description should appear in the next line with indent

xonsh/history/json.py
```diff
@@ -446,12 +446,7 @@ def __init__(
             if save_cwd is not None
             else XSH.env.get("XONSH_HISTORY_SAVE_CWD", True)
         )
-        ignore_regex = XSH.env.get("XONSH_HISTORY_IGNORE_REGEX")
-        self.is_ignored = (
-            (lambda cmd: re.match(ignore_regex, cmd["inp"]) is not None)
-            if ignore_regex is not None
-            else self.is_ignored
-        )
+        self.history_ignore_regex = XSH.env.get("XONSH_HISTORY_IGNORE_REGEX")
```
jnoortheen, line 448: I think it is better to move this to the base class as well. We don't need to repeat unless there is any specialization

xonsh/environ.py
```diff
@@ -1580,6 +1580,7 @@ class PromptHistorySetting(Xettings):
         "Save current working directory to the history.",
         doc_default="True",
     )
+    # !!! XONSH_HISTORY_IGNORE_REGEX goes here?
```
jnoortheen, line 1583: yes you can add it here

xonsh/history/json.py
```diff
@@ -467,7 +468,7 @@ def append(self, cmd):
         hf : JsonHistoryFlusher or None
             The thread that was spawned to flush history
         """
-        if not self.remember_history:
+        if not self.remember_history or self.is_ignored(cmd):
             return
```
jnoortheen, line 472: it is better to put first not expression inside a bracket

xonsh/history/json.py
bcr: `History` is not currently aware of `XSH`. Is it OK to introduce this? I'm cool with it if you are. @anki-code I'm defending your honor, you got an opinion?

Do you want an intermediate class from which both the JSON and SQLite derived, and we leave History ignorant of `XSH`?

tests/test_history_dummy.py
```diff
@@ -10,3 +10,19 @@
 def test_construct_history_str(xession, backend):
     xession.env["XONSH_HISTORY_BACKEND"] = backend
     assert isinstance(construct_history(), DummyHistory)
+
+
+@pytest.mark.parametrize("backend", ["dummy", DummyHistory, DummyHistory()])
+def test_ignore_regex_invalid(xession, backend):
+    xession.env["XONSH_HISTORY_BACKEND"] = backend
+    xession.env["XONSH_IGNORE_REGEX"] = "**"
```
bcr, line 18: I can't figure out how to inject `XONSH_IGNORE_REGEX` -- I don't think that `XSH.env` in the `History` class can see it.

jnoortheen: setting them here is fine. It should be reflected during test run. You can see similar usage by other tests

xonsh/history/dummy.py
```diff
@@ -7,6 +7,9 @@
 class DummyHistory(History):
     """A dummy implement of history backend."""
 
+    def __init__(self, **meta):
+        super().__init__(meta)
```
jnoortheen, line 11: this should pass the '**kwargs'?

xonsh/history/base.py
```diff
@@ -163,3 +169,35 @@ def clear(self):
...
         """
         pass
+
+    def validate_ignore_regex(self):
+        """Validates self.history_ignore_regex to make sure it's a valid
+        regex. If invalid, sets it to None and outputs a warning."""
+        if self.history_ignore_regex is not None:
+            try:
+                re.compile(self.history_ignore_regex)
+            except re.error:
```
jnoortheen, line 179: I think we can store the compile result lazily. Like a `cached_property`

tests/test_history_dummy.py
bcr: In this case empirically it did not work. I'll hunt around and also make sure I didn't spell something wrong.

xonsh/history/dummy.py
bcr: I cribbed it from `JsonHistory.__init__`. I did forget the `**`, I'll fix it.

xonsh/history/base.py
bcr: This is just meant to validate the expression, but the point is taken that I can compile it once and use `match` on the compiled expression.

tests/test_history_dummy.py
bcr: `XONSH_HISTORY_IGNORE_REGEX` != `XONSH_IGNORE_REGEX` 🤦

bcr:
I don't have any further changes that I'm thinking about, I think we've covered all of the comments so far, there are unit tests and docs in place, is this in good enough shape? Happy to hear any change suggestions. @jnoortheen @anki-code

jnoortheen:
@bcr this looks good. One small nit though, still the regex is not compiled. It would be faster when compiled and matched. The easier way would be to add a `cached_property` like

```py
# this is just a psuedo-code, feel free to use what you think the best and concise

import functools

class BaseHistory():
  @functools.cached_property
  def ignore_regex(self):
      regex = XSH.env.get("XONSH_HISTORY_IGNORE_REGEX")
      if regex:
        return re.compile(regex)

  def is_ignored(self, cmd):
    return self.ignore_regex and self.ignore_regex.match(cmd)
  ...
```

bcr:
Let me take a whack at this. @anki-code had it as a string, so I was honoring that. Understood that having it pre-compiled will be faster.

bcr:
@jnoortheen it is done. May our unit tests be mighty and save us from folly.

xonsh/history/base.py
```diff
@@ -163,3 +169,36 @@ def clear(self):
         memory.
         """
         pass
+
+    @functools.cached_property
+    def ignore_regex(self):
+        self.history_ignore_regex = None
```
jnoortheen, line 175: you don't have to store it in an attribute. that will be handled by the decorator

jnoortheen (approved):

Issue #2455, opened 2017-07-24
Add a missing command event


AstraLuma:
Add an event that fires when a command is not found:

```
on_command_not_found(name: str) -> None
```

In addition to factoring out the default handler (spelling-based) into a xoretrib, xontribs can be written to handle things like checking the package manager.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

laerus:
+1

scopatz:
Great idea!

PR #4959, opened 2022-10-04
feat: add on_command_not_found event

bcr:
#2455 requests the addition of an `on_command_not_found` event which is in this PR.
## For community


bcr:
I have some issues I'm working through:

[ ] The help for the event doesn't show up until `specs` is loaded
[ ] I can't find my event in the HTML docs if I do a `make html` from the docs directory. I think I don't have something hooked up right.
[ ] A testing strategy to check that the hook fires

jnoortheen:
@bcr you can check the `xontribs/autovox.py` or any other built-in events for implementation/testing example. The events tutorial will also be helpful.

bcr:
In this particular case I need to have an instantiated `SubprocSpec` so I can try and `run` to make sure the event fires. I did track down a lot of other event handlers, I just didn't find a `test_specs` yet or a framework to get into the guts of `SubprocSpec`. The events model in general seems pretty straightforward.

bcr:
To fix my first problem (help for the event), I need to move:

```python
XSH.builtins.events.doc(
    "on_command_not_found",
    """
on_command_not_found(command: list[str]) -> None

Fired if a command is not found
""",
)
```
outside of specs.py somewhere so that it's available before specs gets loaded. Not sure where the least offensive place for that is.

gforsyth:
> Not sure where the least offensive place for that is.

Hey @bcr -- thanks for the work here!  I would suggest sticking it in either `xonsh/shell.py` or `xonsh/main.py` -- both have pre- and post- execution events so I think either is a reasonable choice.

jnoortheen:
Also this should be available only during interactive sessions.

bcr:
I think I've addressed the open concerns. For some reason the CI Test for Python 3.10 macOS-latest is jammed. @gforsyth @jnoortheen

gforsyth (approved):
Looks good!  Thanks @bcr !

crazyminecuber:
I have not managed to get this event working correctly. Could you provide a minimal example that just prints something when you type a command that does not exist?

bcr:
When you tried the examples in the unit tests did they work? I am afk right now, but I can provide a .xonshrc example later.

gforsyth:
@crazyminecuber -- are you running off of `main`?  

A short example:
```
(base) gil@stravaig ~ 🐚  @events.on_command_not_found
......................... def ruhroh(cmd):
.........................     print(f"There isn't a command named {cmd}")
.........................
(base) gil@stravaig ~ 🐚 
(base) gil@stravaig ~ 🐚  notacommand
There isn't a command named ['notacommand']
xonsh: subprocess mode: command not found: notacommand
```

crazyminecuber:
No I am on 13.3 but I now see that this was merged after that release. I just thought I had it because when I ran ``help(events.on_command_not_found)`` it did not return anything obvious indicating that the event does not exist. Thanks for the quick answer and everything is good then.

bcr:
> I just thought I had it because when I ran `help(events.on_command_not_found)` it did not return anything obvious indicating that the event does not exist.

This is an interesting point. You can do `help(events.xyzzy)` or whatever you want, and I think you'll see the same thing. This may be a deeper philosophical discussion about the dynamic nature of events...

PR #4970, opened 2022-10-18
Fix auto-suggestion completion stacking if no normal completion exists

### LINKED ISSUE #4668: Weird stacking auto-complete
It happens a lot and is pretty annoying, anyone get this/know a fix?

https://user-images.githubusercontent.com/34655998/153747557-fc8d07f3-78e4-4606-b8f6-415c26a5503d.mp4

Here are my custom environment variables
```
$DOTGLOB = True
$COMPLETIONS_CONFIRM = False
$COMPLETION_IN_THREAD = True
$COMPLETIONS_BRACKETS = False
$COMPLETIONS_DISPLAY = "single"
$CASE_SENSITIVE_COMPLETIONS = False
$AUTO_SUGGEST_IN_COMPLETIONS = True
$UPDATE_COMPLETIONS_ON_KEYPRESS = True
$COMMANDS_CACHE_SAVE_INTERMEDIATE = True
$XONSH_HISTORY_BACKEND = "sqlite"
$XONSH_HISTORY_FILE = "/home/ganer/.xonsh_history"
$XONSH_CACHE_EVERYTHING = True
$XONSH_CTRL_BKSP_DELETION = True
$UPDATE_PROMPT_ON_KEYPRESS = True
$XONSH_COLOR_STYLE = 'default'
$XONSH_STYLE_OVERRIDES['Token.Literal.Number'] = '#1111ff'
```

edit: also on a another note, typing backshashes before most characters cause a error to appear in the console

vgerak:
Fixes #4668 

It's possible I missed some edge cases, worked fine on 2 simple tests:

```sh
# run once to add in history (expected fail)
$ test-missing-command
# then test completion
test-missing- #<TAB>
```

```sh
$ ssh foobar@non-existent-host
$ ssh foobar@non- #<TAB>
```

## For community

### DIFF
--- a/xonsh/ptk_shell/completer.py
+++ b/xonsh/ptk_shell/completer.py
@@ -72,6 +72,7 @@ class PromptToolkitCompleter(Completer):
             if sug_comp is None:
                 pass
             elif len(completions) == 0:
+                plen = len(prefix)
                 completions = (sug_comp,)
             else:
                 completions = set(completions)
--- a/news/4668-fix-autocomplete-suggest.rst
+++ b/news/4668-fix-autocomplete-suggest.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* #4668 Fix ptk completion stacking when auto-suggest is on and no normal completions are generated.
+
+**Security:**
+
+* <news item>
--- a/tests/test_ptk_completer.py
+++ b/tests/test_ptk_completer.py
@@ -54,6 +54,35 @@ def test_rich_completion(completion, lprefix, ptk_completion, monkeypatch, xessi
         assert completions == [ptk_completion]
 
 
+@pytest.mark.parametrize(
+    "completions, document_text, ptk_completion",
+    [
+        (set(), "", "test_completion"),
+        (set(), "test_", "test_completion"),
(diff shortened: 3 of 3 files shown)


xonsh/ptk_shell/completer.py
```diff
@@ -114,6 +114,8 @@ def get_completions(self, document, complete_event):
                 yield comp
             else:
                 disp = comp[pre:].strip("'\"")
+                if not plen:
```
jnoortheen, line 117: does this plen is `0` or `None` here?

vgerak: This catches any False-y value. I think the only possibility is 0, since it ultimately comes from `xonsh.completer.complete_from_context()`, so we could compare with it directly.

Othewise, in case `plen==0` is a "valid" value, we could set it within the `elif len(completions) == 0:` clause in L74, which guarantees no "normal" completions were generated.

gforsyth: `0` is definitely a valid value -- if you do `cd <TAB>` then every directory at your location will be returned and the prefix length will be `0` for all of them

vgerak: amended to change `plen` only if no regular completions found

jnoortheen:
@vgerak can you add a test case for the issue described?

vgerak:
@jnoortheen 
There are no tests of auto-suggestions in the codebase on which I could build upon/extend.

I added a commit so we have something concrete to discuss on, but I'm not confident that I've mocked (or even tested) the correct things.

jnoortheen (approved):

jnoortheen:
Looks good! Thanks @vgerak

PR #4976, opened 2022-10-26
pylint fixes

marksmayo:
encodings, simplification of if/elses, and some other small fixes


jnoortheen:
It would be good to stick with just flake8. Most of pylint codes are now available with https://github.com/wemake-services/wemake-python-styleguide or its lite implementation. Thanks @marksmayo  for checking on the code quality.

PR #4978, opened 2022-10-27
Use interactive shell for generating completions

uosis:
Non-interactive shells may not load all the appropriate config scripts to provide completions. Since completion is an interactive feature, it seems appropriate to ask interactive shell for completions. 

This should also make `$BASH_COMPLETIONS` largely unnecessary, since interactive shell will load whatever it needs.

## For community

### DIFF
--- a/xonsh/completers/bash_completion.py
+++ b/xonsh/completers/bash_completion.py
@@ -386,7 +386,7 @@ def bash_completions(
         command = _bash_command(env=env)
     try:
         out = subprocess.check_output(
-            [command, "-c", script],
+            [command, "-i", "-c", script],
             text=True,
             stderr=subprocess.PIPE,
             env=env,


uosis:
> I don't think an interactive Bash shell is needed to run `complete`.

No, but interactive shell is needed to ensure all completions are loaded. [Non-interactive shells do not load `.bashrc`](https://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html#:~:text=Invoked%20as%20an%20interactive%20non%2Dlogin%20shell), which is where completions are normally defined.

> `$BASH_COMPLETIONS` provides extra completions for various commands by loading the `bash-completion` package (https://github.com/scop/bash-completion). I don't think this is loaded by default, since on my system, after I start an interactive bash shell, I have to run `source /usr/share/bash-completion/bash_completion` in order to get those completions.

Yes, that `source /usr/share/bash-completion/bash_completion` is normally part of `.bashrc` that is loaded by an interactive shell.

bestlem:
Surely completions only matter when you are typing into a shell. Thus they should only be loaded if and only the shell is interactive.

Thus the error being fixed is not an error. In fact shouldn't the first sentance be

Non-interactive shells should not load any of the appropriate config scripts to provide completions.

uosis:
> Surely completions only matter when you are typing into a shell. Thus they should only be loaded if and only the shell is interactive.
> 
> Thus the error being fixed is not an error. In fact shouldn't the first sentance be
> 
> Non-interactive shells should not load any of the appropriate config scripts to provide completions.

Correct. The problem is that xonsh is asking non-interactive bash to provide completions, which, as per "Non-interactive shells should not load any of the appropriate config scripts to provide completions", non-interactive bash cannot do. The fix is to start bash as interactive shell so it can provide completions.

doronz88:
Can you elaborate what does such rc script do to require the completions to work? Env variables for example are inherited from the xonsh shell

gforsyth:
Closing out old PRs, thanks!

PR #4954, opened 2022-10-02
cmd cache refactor - optimize cache usage

jnoortheen:
- [x] No need to cache aliases
- [x] lazy cache loading per path

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #4988, opened 2022-11-09
commands_cache: ignore mtime check on non-darwin os

yotamolenik:
darwin's stat uses nano seconds accuracy for the files mtime property, whereas
on linux it returns a unix timestamp (seconds), meaning we cannot rely on that
field when checking for dir modifications per each path for creating `CommandsCache`.

consider the following shell script:

```shell
cp /path/to/executable /bin/new_executable

new_executable
```

## For community

### DIFF
--- a/tests/test_commands_cache.py
+++ b/tests/test_commands_cache.py
@@ -1,5 +1,6 @@
 import os
 import pickle
+import time
 
 import pytest
 
@@ -173,3 +174,31 @@ class Test_is_only_functional_alias:
     def test_bash_and_is_alias_is_only_functional_alias(self, xession):
         xession.aliases["git"] = lambda args: os.chdir(args[0])
         assert xession.commands_cache.is_only_functional_alias("git") is False
+
+
+def test_update_cache(xession, tmp_path):
+    xession.env["ENABLE_COMMANDS_CACHE"] = False
+    basename = "PITA.EXE"
+    subdir1 = tmp_path / "subdir1"
+    subdir2 = tmp_path / "subdir2"
+    subdir1.mkdir()
+    subdir2.mkdir()
+    file1 = subdir1 / basename
+    file2 = subdir2 / basename
+    file1.touch()
+    file1.chmod(0o755)
+
+    cache = CommandsCache({"PATH": [subdir2, subdir1]})
+    cached = cache.update_cache()
+
+    assert file1.samefile(cached[basename][0])
+
+    # give the os enough time to update the mtime field of the parent directory
+    # (represented in seconds on Linux and Windows systems)
+    time.sleep(2)
+    file2.touch()
+    file2.chmod(0o755)
+
+    cached = cache.update_cache()
+
+    assert file2.samefile(cached[basename][0])
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -175,8 +175,10 @@ class CommandsCache(cabc.Mapping):
         updated = False
         for path in paths:
             modified_time = os.stat(path).st_mtime
-            if (path not in self._paths_cache) or (
-                self._paths_cache[path].mtime != modified_time
+            if (
+                (not self.env.get("ENABLE_COMMANDS_CACHE", True))
+                or (path not in self._paths_cache)
+                or (self._paths_cache[path].mtime != modified_time)
             ):
                 updated = True
                 self._paths_cache[path] = _Commands(
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -865,7 +865,7 @@ class GeneralSetting(Xettings):
     ENABLE_COMMANDS_CACHE = Var(
         default=True,
         doc="command names in a directory are cached when enabled.",
-        doc_default="It makes use of directory's modified time. "
+        doc_default="True"
         "On some platforms it may not be accurate enough"
         "(e.g. Windows, Linux save mtime in seconds). "
         "setting it to False would disable the caching mechanism "
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -111,7 +111,6 @@ cause a recursion until the limit.
 """,
 )
 
-
 events.doc(
     "on_envvar_change",
     """
@@ -123,7 +122,6 @@ cause a recursion until the limit.
 """,
 )
 
-
 events.doc(
     "on_pre_spec_run_ls",
     """
@@ -133,7 +131,6 @@ Fires right before a SubprocSpec.run() is called for the ls
 command.
 """,
 )
-
 
 events.doc(
     "on_lscolors_change",
@@ -863,6 +860,16 @@ class GeneralSetting(Xettings):
     COMMANDS_CACHE_SAVE_INTERMEDIATE = Var.with_default(
         False,
         "If enabled, the CommandsCache is saved between runs and can reduce the startup time.",
+    )
+
+    ENABLE_COMMANDS_CACHE = Var(
+        default=True,
+        doc="command names in a directory are cached when enabled.",
+        doc_default="It makes use of directory's modified time. "
+        "On some platforms it may not be accurate enough"
+        "(e.g. Windows, Linux save mtime in seconds). "
+        "setting it to False would disable the caching mechanism "
+        "and may slow down the shell",
     )
 
     HOSTNAME = Var.with_default(
--- a/news/bugfix_commands_cache.rst
+++ b/news/bugfix_commands_cache.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* commands_cache: add a configurable value to disable cache. useful for running scripts
+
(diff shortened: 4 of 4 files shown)


jnoortheen:
Thanks @yotamolenik for trying to fix this. I was working on `CommandsCache` and this fix was part of #4954 . Now it is ready, could you test that PR?

yotamolenik:
@jnoortheen the bug persists on linux, but I rebased my changes on top of your branch and tested the fix.
it works now on both linux and darwin 🥳

jnoortheen:
@yotamolenik Ok thats good. Lets wait until #4954 is merged.

xonsh/commands_cache.py
```diff
@@ -166,10 +166,13 @@ def _update_paths_cache(self, paths: tp.Sequence[str]) -> bool:
                 self.cache_file.unlink(missing_ok=True)
 
         updated = False
+        is_darwin = os.uname().sysname.lower() == "darwin"
```
jnoortheen, line 169: the xonsh.platform has already similar functions. Lets reuse that.

yotamolenik: done

jnoortheen:
Not using cache at all seems a bit too much. It may lead to shell being sluggish. Perhaps we can add a warning in the doc?

doronz88:
@jnoortheen what do you propose instead of not using cache on non-darwin OS? i think sluggish is better than a shell that relies on something with chances of failure.
the worst thing about this bug is that it may resolve in really bad errors. i encountered this bug when i had a python entry point both inside a venv and on global path.
it sometimes resolved the entry point to the system-one and sometimes to the venv one. only workarround is adding a `sleep 1` after the venv activation

jnoortheen:
I think we can use a setting value to control this behaviour while defaulting to being enabled. We can show the warning. 
Previously there was a bug which caused stale results. It was fixed with  #4954 .

jnoortheen:
The branch is now merged. You can start with `xonsh/environ.py`. I am on mobile, just pointing out the file name.

yotamolenik:
@jnoortheen Do you mean like this?

xonsh/environ.py
```diff
@@ -865,6 +865,15 @@ class GeneralSetting(Xettings):
         "If enabled, the CommandsCache is saved between runs and can reduce the startup time.",
     )
 
+    COMMANDS_CACHE_IGNORE_MTIME_CHECK = Var.with_default(
+        True,
+        "If enabled, ignore mtime check when checking if directories changed"
+        "will decrease performance, but ensures scripts will run as expected."
```
doronz88, line 871: reformat. it ensures executables' path will be resolved to correct one on platforms where the directories `mtime` field isn't represented accurate enough (nanoseconds on Darwin, but seconds on Linux).

doronz88 (changes requested):

xonsh/commands_cache.py
```diff
@@ -168,8 +169,10 @@ def _update_paths_cache(self, paths: tp.Sequence[str]) -> bool:
         updated = False
         for path in paths:
             modified_time = os.stat(path).st_mtime
-            if (path not in self._paths_cache) or (
-                self._paths_cache[path].mtime != modified_time
+            if (
+                GeneralSetting.COMMANDS_CACHE_IGNORE_MTIME_CHECK
+                or (path not in self._paths_cache)
```
jnoortheen, line 174: this should be , `self.env.get('COMMANDS_CACHE_IGNORE_MTIME_CHECK')` .

xonsh/environ.py
```diff
@@ -865,6 +865,14 @@ class GeneralSetting(Xettings):
         "If enabled, the CommandsCache is saved between runs and can reduce the startup time.",
     )
 
+    COMMANDS_CACHE_IGNORE_MTIME_CHECK = Var.with_default(
+        True,
+        "If enabled, ignore mtime check when checking if directories changed"
```
jnoortheen, line 870: - the default can be False. 
- Ultimately this variable disables commands_cache's cache system not just the mtime check. mtime-check is an implementation detail. I think this should be named `ENABLE_COMMANDS_CACHE(default=True)` . I suggest the description to be `command names in a directory are cached when enabled. It makes use of directory's modified time. On some platforms it may not be accurate enough (e.g. Windows, Linux save mtime in seconds). setting it to False would disable the caching mechanism and may slow down the shell`

yotamolenik: sounds good!
I also changed my test a little bit to run only if the cache is disabled.

tests/test_commands_cache.py
```diff
@@ -173,3 +173,31 @@ def test_non_exist(self, xession):
     def test_bash_and_is_alias_is_only_functional_alias(self, xession):
         xession.aliases["git"] = lambda args: os.chdir(args[0])
         assert xession.commands_cache.is_only_functional_alias("git") is False
+
+
+def test_update_cache(xession, tmp_path):
+    # run test only if the cache is disabled
+    if xession.env.get("ENABLE_COMMANDS_CACHE"):
```
jnoortheen, line 180: the test should set the variable to False and proceed. You can see the other tests in this file for similar tests.

xonsh/commands_cache.py
```diff
@@ -168,8 +168,10 @@ def _update_paths_cache(self, paths: tp.Sequence[str]) -> bool:
         updated = False
         for path in paths:
             modified_time = os.stat(path).st_mtime
-            if (path not in self._paths_cache) or (
-                self._paths_cache[path].mtime != modified_time
+            if (
+                self.env.get("ENABLE_COMMANDS_CACHE")
```
jnoortheen, line 172: it is best to pass the default value True here to `.get()`

tests/test_commands_cache.py
yotamolenik: done

xonsh/commands_cache.py
yotamolenik: Why is it best?🤔
If the value doesn't exist it should be a bug

jnoortheen: sometimes the variable might be None; during unit tests.

xonsh/commands_cache.py
```diff
@@ -168,8 +168,10 @@ def _update_paths_cache(self, paths: tp.Sequence[str]) -> bool:
         updated = False
         for path in paths:
             modified_time = os.stat(path).st_mtime
-            if (path not in self._paths_cache) or (
-                self._paths_cache[path].mtime != modified_time
+            if (
+                self.env.get("ENABLE_COMMANDS_CACHE", True)
```
doronz88, line 172: the condition is the exact opposite. currently if its enabled, it means to disable

doronz88 (changes requested):

doronz88 (approved):

tests/test_commands_cache.py
```diff
@@ -173,3 +174,32 @@ def test_non_exist(self, xession):
     def test_bash_and_is_alias_is_only_functional_alias(self, xession):
         xession.aliases["git"] = lambda args: os.chdir(args[0])
         assert xession.commands_cache.is_only_functional_alias("git") is False
+
+
+def test_update_cache(xession, tmp_path):
+    xession.env["ENABLE_COMMANDS_CACHE"] = False
```
doronz88, line 180: redundant

xonsh/commands_cache.py
```diff
@@ -168,8 +168,10 @@ def _update_paths_cache(self, paths: tp.Sequence[str]) -> bool:
         updated = False
         for path in paths:
             modified_time = os.stat(path).st_mtime
-            if (path not in self._paths_cache) or (
-                self._paths_cache[path].mtime != modified_time
+            if (
+                not self.env.get("ENABLE_COMMANDS_CACHE", True)
```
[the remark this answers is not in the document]
yotamolenik: done

yotamolenik:
I believe the code is ok now. 
About the linux failure it might be related to the os kernel and the scheduler for the time.sleep() function
About the windows failure I made a tiny mistake with the name of the exe - fixed it now and it should work

jnoortheen:
@yotamolenik Please make sure that the test passes on Linux as well.

yotamolenik:
It should always pass now, i did have another mistake

jnoortheen:
pre-commit.ci autofix

xonsh/environ.py
```diff
@@ -865,6 +862,16 @@ class GeneralSetting(Xettings):
...
 
+    ENABLE_COMMANDS_CACHE = Var(
+        default=True,
+        doc="command names in a directory are cached when enabled."
+        " It makes use of directory's modified time. "
+        "On some platforms it may not be accurate enough"
+        " (e.g. Windows, Linux save mtime in seconds). "
+        "setting it to False would disable the caching mechanism"
+        " and may slow down the shell",
+    )
```
jnoortheen, line 873: it is good to define `doc_default` as well. It will appear in the docs generated

yotamolenik: do you mean like this?

jnoortheen: I've updated it.

jnoortheen (approved):

jnoortheen:
Thanks @yotamolenik for the continued effort!

PR #5026, opened 2023-01-11
ci: Add 3.11 to CI and fix existing failures

gforsyth:
~temporary fix while we sort out what broke upstream~

`Token.Text` for `\n` is now `Token.Text.Whitespace` -- doesn't really matter on our end since `prompt_toolkit` does the right thing, but I've updated the expected tokens in our tests.

Also adding 3.11 to CI, just to make things interesting.

## For community

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,6 +35,7 @@ jobs:
           - "3.8"
           - "3.9"
           - "3.10"
+          - "3.11"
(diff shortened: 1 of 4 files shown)


gforsyth:
xref: https://github.com/simonw/sqlite-utils/issues/503 and thanks to @simonw for figuring out the Python 3.11 / sqlite / windows thing

PR #5028, opened 2023-01-13
xontrib load: add option to skip not installed xontribs

### LINKED ISSUE #5023: xontrib load: add option to skip not installed xontribs
It will be cool to have an ability to load xontribs without errors about not installed ones.

For example this useful in case when you run xonsh in virtual environment and xonsh tries to run ~/.xonshrc and because of the virtual env has no xontribs xonsh produces errors. Using skip missing xontribs option in xonshrc you can prevent errors. 

## Current Behavior

Show error and return code 1:

```xsh
xontrib load whole_word_jumping xontrib_not_installed
# The following xontribs are enabled but not installed:
#   ['xontrib_not_installed']
# Please make sure that they are installed correctly by checking https://xonsh.github.io/awesome-xontribs/
# Return code 1
```

## Expected Behavior
Add an option `-s` ("(S)ilent" or "(S)kip missing") and return zero code without errors:
```xsh
xontrib load -s whole_word_jumping xontrib_not_installed
# Return code 0
```

## Current workaround

[xontrib-rc-awesome](https://github.com/anki-code/xontrib-rc-awesome/blob/5105050cf2f7318eb8b011f1f7a5897574e6ca05/xontrib/rc_awesome.xsh#L68-L83) has this workaround:

```xsh
from xonsh.xontribs import get_xontribs 
_xontribs_installed = set(get_xontribs().keys()) 
_xontribs_to_load = ('back2dir', 'prompt_bar', 'pipeliner', 'sh',)
xontrib load @(_xontribs_installed.intersection(_xontribs_to_load))
```

## Start point

https://github.com/xonsh/xonsh/blob/3716f3da8636e41482358b61779e274aade8ecf7/xonsh/xontribs.py#L198

## For community

anki-code:
Closes #5023

## For community


news/xontrib_load_skip_missing.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* xontrib load: added option ``-s`` to skip not installed xontribs.
```
jnoortheen, line 3: skip essentially would mean silencing the warning and not skipping.

xonsh/xontribs.py
```diff
@@ -202,6 +202,7 @@ def xontribs_load(
     ] = (),
     verbose=False,
     full_module=False,
+    skip_missing=False,
```
jnoortheen, line 205: could you name this to silence or something similar?

jnoortheen: it doesn't signify what it is doing. By default it is going to skip loading xontribs not installed but will show warning that is not installed. So we have to make it clear that we are silencing those warnings shown and no other behaviour is changed.

gforsyth: how about `no_warn_missing` ?

gforsyth: that's a bit of an anti-pattern -- better to have the name be `warn_missing` and then check for `if not warn_missing` or similar

xonsh/xontribs.py
```diff
@@ -225,7 +228,8 @@ def xontribs_load(
         try:
             update_context(name, ctx=ctx, full_module=full_module)
         except XontribNotInstalled:
-            bad_imports.append(name)
+            if not no_warn_missing:
```
gforsyth, line 231: This is hard to understand when reading through -- it's much better if it's something like `if warn_missing:`

xonsh/xontribs.py
gforsyth: We can decide what the name of an `argparse` argument is, and what the name of the variable it targets is -- if the automatic argparse thing can't do that, we should just use straight argparse for this.

gforsyth: @anki-code, how about `suppress_warnings` or `silence_warnings`?

gforsyth (approved):
looks good to me!  @jnoortheen I'll leave this open if you want to take another look.

jnoortheen (approved):

jnoortheen:
LGTM, I am merging. 

Thanks! @anki-code

PR #5046, opened 2023-02-03
Add support for Fossil SCM branch names in vc.py.

luziferius:
The xonsh prompt now shows the currently active Fossil branch, while inside a Fossil checkout.

See also https://fossil-scm.org

The feature works similarly to how git and hg are currently handled: While inside a Fossil checkout, the currently active branch is shown in the prompt between the path and `$`.

Note: 

There are further possibilities for refactoring in [vc.py](https://github.com/xonsh/xonsh/blob/main/xonsh/prompt/vc.py). The old code used for git and hg uses manual thread management and passes data via queues to implement a command timeout. This can be replaced completely by using the `timeout` parameter of [`subprocess.check_output()`](https://docs.python.org/3/library/subprocess.html#subprocess.check_output). I guess the code once targeted Python < 3.3

I did not include the refactoring here, as that’s out of scope for this particular PR. If you wish, I can include that in follow-up commits or submit as a new PR.

Additional note:

The default value for `$VC_BRANCH_TIMEOUT` can be a little low. I run into it when switching branches with larger amount of differences

## For community

### DIFF
--- a/news/vc.rst
+++ b/news/vc.rst
@@ -1,0 +1,24 @@
+**Added:**
+
+* Display the current branch of Fossil VCS checkouts in the prompt,
+  similar to git and hg.
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/tests/prompt/test_vc.py
+++ b/tests/prompt/test_vc.py
@@ -1,5 +1,7 @@
 import os
 import subprocess as sp
+import textwrap
+from typing import Dict, List
 from pathlib import Path
 from unittest.mock import Mock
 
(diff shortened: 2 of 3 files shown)


jnoortheen:
Yes this module requires a cleanup. It can reuse the parts of `gitstatus.py`. Also the prompt functions need not initialize a repo.

luziferius:
> Also the prompt functions need not initialize a repo

I’m not sure what you mean by that. The tests do so, in order to test the functionality. And because the initial repository setup is a bit more complex, I had to modify the `repo` test fixture and [supporting data structures](https://github.com/xonsh/xonsh/blob/01085d8277cdaa4b2b8558f59ce80c60ac7dd5eb/tests/prompt/test_vc.py#L18) to support that.

The prompt functions themselves do not. Those run `fossil branch current` to determine the current branch name and `fossil changes` to determine if there are pending, uncommitted changes.

jnoortheen:
@luziferius The tests do not need to have full setup of fossil/hg. It might not be present in all the dev systems like git. The calls to the subprocess can be mocked with this https://pypi.org/project/pytest-subprocess/ as well.

luziferius:
On the decrease in the code coverage report: The CI VM misses the fossil executable, thus automatically skips the added test cases.

On Linux, installation is easy. On the Ubuntu VM, add the [`fossil`](https://packages.ubuntu.com/jammy/fossil) package to the dependencies.

On Windows, I’m unfortunately not sure how to easily get the latest binary. It is always present in the [binaries listing](https://fossil-scm.org/home/uvlist) or the respective JSON endpoint [here](https://fossil-scm.org/home/juvlist). It requires a bit of processing, though.
If that’s solved, installation is simply pulling the ZIP and extracting the binary into any directory in $PATH.

luziferius:
> @luziferius The tests do not need to have full setup of fossil/hg. It might not be present in all the dev systems like git. The calls to the subprocess can be mocked with this https://pypi.org/project/pytest-subprocess/ as well.

The tests for test_vc.py did it that way, so I simply extended that.

jnoortheen:
> The tests for test_vc.py did it that way. 

It is part of the old code. Please check the [test_gitstatus.py](https://github.com/xonsh/xonsh/blob/main/tests/prompt/test_gitstatus.py#L23) on how fake_process is used to abstract away system interaction and need for external applications to be present.

jnoortheen (approved):

jnoortheen:
Thanks @luziferius !

PR #5054, opened 2023-02-14
fix(environ): push PATH updates into os.environ on change

gforsyth:
This issue was raised in
https://github.com/xonsh/xonsh/discussions/5050.

If users (or `xontribs`) make use of `subprocess`, the `PATH` available
to `subprocess` is the value of `PATH` in `os.environ`, which is set at
the first import of `os.environ`.

This means (meant) that if a user updates their `PATH` in `xonshrc`,
those changes aren't visible to a bare `subprocess` call.
(This is what `UPDATE_OS_ENVIRON` is nominally for, but since changes to
`PATH` are mutating a mutable object, we don't hit any of the
`__setitem__` checks in the env object).

So, instead, I've added an explicit call to update the `os.environ`
value of `PATH` after each mutating call to our `EnvPath` class, so this
should also work for the various `_PATH` variables (e.g. `LDD_LIBRARY_PATH`).

## For community

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -8,6 +8,7 @@ import warnings
 from tempfile import TemporaryDirectory
 
 import pytest
+
 from xonsh import __version__
 from xonsh.lexer import Lexer
 from xonsh.platform import HAS_PYGMENTS, ON_WINDOWS, PYTHON_VERSION_INFO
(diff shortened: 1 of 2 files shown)


anki-code:
Closed with reference to https://github.com/xonsh/xonsh/pull/6171

PR #5065, opened 2023-02-20
Add Semantic Prompt support to multiline continuation lines

### LINKED ISSUE #5058: $MULTILINE_PROMPT doesn't filter out OSC commands
I'm adding "terminal integration" to xonsh and as such need to add various prompt commands to prompt for the terminal to recognize when input/output starts/ends etc. [doc](https://github.com/wez/wezterm/blob/main/docs/shell-integration.md)

It's working fine in other prompt variables, but then in `$MULTILINE_PROMPT` it breaks down, I see the commands as literal ugly strings instead of them being their regular invisible selves with just `.`

## xonfig

<details>

```
$ xonfig
| xonsh            | 0.13.4          |
| Python           | 3.10.10         |
| prompt toolkit   | 3.0.36          |
| pygments         | 2.14.0          |
| xontrib 1        | cd              |
| RC file          | []              |
```

</details>

## Expected Behavior

```xsh
$MULTILINE_PROMPT = "\001" + "\x1b]133;P;k=c\x07" + "\002" + "."
```

Signals that every new line in the multiline propmt is a "continuation prompt", but it should be invisible, the result should be just the same multiple `.` dots depending on your prompt width

## Current Behavior

Instead, I get this
```xsh
$MULTILINE_PROMPT = "\001" + "\x1b]133;P;k=c\x07" + "\002" + "."
~ $ if 2:
^A^[]133;P;k=c^G^B.^A^[]133;P;     echo 1
#                 ↑ this is our dot :)
```

while this works fine
```xsh
~ $ $PROMPT = "\001" + "\x1b]133;P;k=i\x07" + "\002" + "<"
<
```

It might be due to some issues in parsing in this command here? Maybe it's due to repetition of the `.`?:

https://github.com/xonsh/xonsh/blob/main/xonsh/prompt/base.py#L184

### Traceback (if applicable)

<details>

```
traceback
```

</details>

## Steps to Reproduce

- Start `xonsh --no-rc`
- Execute `$MULTILINE_PROMPT = "\001" + "\x1b]133;P;k=c\x07" + "\002" + "."`
- Type a multiline command

## For community

eugenesvk:
This PR allows marking each line of the multiline continuation prompt to allow the terminal to separate user input from the `...` continuation dots and do something useful with it (e.g., select&copy all the multiline input that you can safely paste elsewhere)
It does so by adding two environment variables that other [xonribs](https://github.com/jnoortheen/xontrib-term-integrations/pull/9) can then use to insert invisible markers that a given terminal supports
(this also fixes a bug where an empty `{BLUE}` dots would lead to a concatenation error)

Tested it live, but don't understand how to make a proper test for this, tried to copy the `test_ptk_multiline.py`'s https://github.com/xonsh/xonsh/blob/eae3fcc255d9c540b3fae34a35e659f5fd7c1e7d/tests/test_ptk_multiline.py#L69

by adding 
```py
prefix = '\x01\x1b]133;P;k=c\x07\x02'
xession.env
envx["MULTILINE_PROMPT_PRE"] = prefix # or even a random "string" doesn't show up in the output
```

but it doesn't seem to have any effect as it's not running through that `continuation_tokens` path it seems, so any tips on which function to invoke to call the changed code path are welcome

Closes #5058

## For community

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,8 @@ include/
 venv/
 .venv*/
 
+# Python project version files
+.python-version
 
 # Mac
 .DS_Store
--- a/news/multil.rst
+++ b/news/multil.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Add support for `Semantic Prompt <https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md>`_ for line continuations in multiline prompts via two environment variables: ``$MULTILINE_PROMPT_PRE`` (e.g., ``\x01\x1b]133;P;k=c\x07\x02``), and ``$MULTILINE_PROMPT_POS`` (e.g., ``\x01\x1b]133;B\x07\x02``) that are inserted before/after each continuation line 'dots' block to mark input
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -1386,6 +1386,30 @@ class PromptSetting(Xettings):
         "Prompt text for 2nd+ lines of input, may be str or function which "
         "returns a str.",
     )
+    MULTILINE_PROMPT_PRE = Var(
+        is_string_or_callable,
+        ensure_string,
+        ensure_string,
+        "",
+        "Indicator inserted before the line continuation marks set "
+        "in ``$MULTILINE_PROMPT``. Can be used to mark the start of "
+        "a semantic continuation prompt "
+        "(see `Semantic Prompts <https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md>`_ "
+        "or `WezTerm <https://wezfurlong.org/wezterm/shell-integration.html>`_ "
+        "for more details). May be str or function which returns a str.",
+    )
+    MULTILINE_PROMPT_POS = Var(
+        is_string_or_callable,
+        ensure_string,
+        ensure_string,
+        "",
+        "Indicator inserted after the line continuation marks set "
+        "in ``$MULTILINE_PROMPT``. Can be used to mark the end of "
+        "a semantic continuation prompt and the beginning of user input "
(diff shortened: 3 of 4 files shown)


eugenesvk:
You should test the new variables, not `MULTILINE_PROMPT`, that one should remain pristine `.`-only without any control chars

You can either install this xontrib that does set the vars for you https://github.com/eugenesvk/xontrib-term-integrations/tree/wez_multiline

Or set them manulally like so:
```xsh
$MULTILINE_PROMPT_PRE = '\x01\x1b]133;P;k=c\x07\x02'
$MULTILINE_PROMPT_POS = '\x01\x1b]133;B\x07\x02'
```

I just didn't want to convert structured data into a single unstructured `MULTILINE_PROMPT` mess to then try to split it into structured data again

eugenesvk:
I don't know whether iTerm2 supports it (and haven't implemented any support for iTerm2 in the xontrib because of this), so with the xontrib you should see no difference there, and with manually setting the vars you should get whatever iTerm2 does with these control sequences :)

But if you try the changes in WezTerm and set your triple-click to select semantic zones like this https://wezfurlong.org/wezterm/config/lua/keyassignment/SelectTextAtMouseCursor.html, then you'll be able to triple-click to select a command on a continuation line without selecting the `...` dots (default triple-click selects the whole line).
You still can't select the whole multiline input as that functionality isn't supported yet by WezTerm

eugenesvk:
> May be this:
> 
> ```python
> $MULTILINE_PROMPT_PRE = '\x01\x1b]133;P;k=c\x07\x02'
> $MULTILINE_PROMPT_POS = '\x01\x1b]133;B\x07\x02'
> ```
> 
> should be implemented by default? 

> Or there will be issues?

No idea, I don't know how other terminals react to this, so migh be safer to leave it as is and let the users install the terminal integration xontrib explicitly if they want these nicer defaults 

(also I'm not even 100% certain these will be the correct markers as that will depend on a specific terminal implementation. I mean, they make sense spec-wise (before `..` you tell the terminal that the continuation prompt starts and after `..` you tell the terminal that the prompt ends and user input begins), but who knows)

eugenesvk:
> What is the order of commands for iTerm2?

sorry, which commands? tried setting env vars both before and after loading the xontrib, no difference

eugenesvk:
(by the way, for some reason the website `make html` build is not reliable, after updating the environ.py description for the variables, I still get older version)

jnoortheen (approved):

jnoortheen:
Thanks @eugenesvk !

PR #5066, opened 2023-02-20
fix: Suppress subprocess traceback in case XONSH_SHOW_TRACEBACK=False and $RAISE_SUBPROC_ERROR=True

xmnlab:
Don't show the traceback for `RAISE_SUBPROC_ERROR` when `XONSH_SHOW_TRACEBACK` is false.

https://github.com/xonsh/xonsh/discussions/4708

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #5072, opened 2023-02-27
Improve error message for RAISE_SUBPROC_ERROR=True

xmnlab:
This PR aims to Improve the error message when RAISE_SUBPROC_ERROR is set to True

This is a follow-up for PR #5066 

## For community

### DIFF
--- a/news/pr-5072.rst
+++ b/news/pr-5072.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Improve the error message when ``$RAISE_SUBPROC_ERROR`` is set to True.
+
+**Security:**
+
+* <news item>
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -329,7 +329,7 @@ def test_rc_with_failing_module(shell, tmpdir, monkeypatch, capsys, xession):
 
     stdout, stderr = capsys.readouterr()
     assert len(stdout) == 0
-    assert "RuntimeError: Unexpected error" in stderr
+    assert "Unexpected error" in stderr
 
     # Check that the temporary rc's folder is not left behind on the path
     assert tmpdir.strpath not in sys.path
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -562,9 +562,8 @@ def main_xonsh(args):
             err_type, err, _ = exc_info
             if err_type is SystemExit:
                 raise err
-            else:
-                print_exception(None, exc_info)
-                exit_code = 1
+            print_exception(None, exc_info)
+            exit_code = 1
         events.on_exit.fire()
         postmain(args)
     return exit_code
--- a/xonsh/main.py
+++ b/xonsh/main.py
@@ -23,7 +23,12 @@ from xonsh.platform import HAS_PYGMENTS, ON_WINDOWS
 from xonsh.pretty import pretty
 from xonsh.shell import Shell
 from xonsh.timings import setup_timings
-from xonsh.tools import display_error_message, print_color, to_bool_or_int
+from xonsh.tools import (
+    display_error_message,
+    print_color,
+    print_exception,
+    to_bool_or_int,
+)
 from xonsh.xonfig import print_welcome_screen
 from xonsh.xontribs import auto_load_xontribs_from_entrypoints, xontribs_load
 
@@ -558,8 +563,7 @@ def main_xonsh(args):
             if err_type is SystemExit:
                 raise err
             else:
-                if XSH.env.get("XONSH_SHOW_TRACEBACK"):
-                    traceback.print_exception(*exc_info)
+                print_exception(None, exc_info)
                 exit_code = 1
         events.on_exit.fire()
         postmain(args)
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -1059,22 +1059,7 @@ def print_exception(msg=None, exc_info=None):
                 "$XONSH_TRACEBACK_LOGFILE = <filename>\n"
             )
 
-        traceback_str = "".join(
-            traceback.format_exception(*exc_info, limit=limit, chain=chain)
-        )
-
-        # color the traceback if available
-        _, interactive = _get_manual_env_var("XONSH_INTERACTIVE", 0)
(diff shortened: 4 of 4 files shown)


xmnlab:
for the example used in #5066, this is the result:

```xonsh
$ xonsh /tmp/test.xsh 
LINE 1
cp: cannot stat 'nonexisting_file.txt': No such file or directory
File "/tmp/test.xsh", line 5, in <module>
    cp nonexisting_file.txt other_name.txt
```

xmnlab:
> The code to make the output short may be looks good in [xonsh.tools.print_exception](https://github.com/xonsh/xonsh/blob/c63f75efd05f240c5f878a4ae778efb3fe9893b1/xonsh/tools.py#L999) and with using [`limit=1`](https://github.com/xonsh/xonsh/pull/5066#issuecomment-1447528538).

@anki-code I changed a bit the code using the parameter limit and the result is the same as mentioned here: https://github.com/xonsh/xonsh/pull/5072#issuecomment-1447083766 with a much better code. thanks!

question: do you want me to implement limit to [xonsh.tools.print_exception](https://github.com/xonsh/xonsh/blob/c63f75efd05f240c5f878a4ae778efb3fe9893b1/xonsh/tools.py#L999) ?

if yes, it would be ok if I remove the first line ("Traceback ...") and the last line ("subprocess.CalledProcessError:  ...") from the output?

also, I can work on that as a follow-up PR if that makes sense

xmnlab:
@anki-code could you check if you like this new approach?

xmnlab:
nice. glad to be in the right direction! I am working now  on the missing parts! thanks!

xmnlab:
@anki-code , I made this change in order to fix the tests: https://github.com/xonsh/xonsh/pull/5072/files#diff-41d180f6f7233d175c2dfe36c45c4ea80eed4754fa7ed4eed46ccde1b2a778c7R332

but I am not sure if that test should be changed. I mean, not sure if my changes correctly changed the output for that specific test. Or if I should if it in some way in the code.

do you have any idea about that?

xmnlab:
The output from my PR for that test is: 

```xonsh
File "/tmp/pytest-of-xmn/pytest-7/test_rc_with_failing_module0/rc.xsh", line 1, in <module>
    from my_failing_module import *
error running xonsh run control file '/tmp/pytest-of-xmn/pytest-7/test_rc_with_failing_module0/rc.xsh': Unexpected error

```

and for the current code is:

```xonsh
RuntimeError: Unexpected error
error running xonsh run control file '/tmp/pytest-of-xmn/pytest-8/test_rc_with_failing_module0/rc.xsh': Unexpected error

```

xmnlab:
> I like the output but the implementation looks like you ignore the whole code in print_exception and just add the logic for your case at the beggining of the function. For example in the function we see show_trace, color_results variables and probably it's good to use them.

oh sorry .. I thought it was just about improving the message for the case when `RAISE_SUBPROC_ERROR=True`. For example, my implementation (the error message) would just affect the case where show_trace is False.

could you provide more details about what you have in mind? what functions it would be impacted for this PR?

thanks for the review and all the guidance.

xmnlab:
> Can we use color_results with the short traceback?

Ok, I can give it a shot :) I will work on that later today

xmnlab:
@anki-code 

xonsh-output

I needed to remove the "strip" method because the file path was not being colored ... and it looked better with the "Traceback" first line, so I added that back as well.

let me know if you have more suggestion.

PS: I rebased the code .. and the pre-commit hooks for black, pyupgrade, and trim trailing whitespace is failing now (+ flake8 that was already failing)

xmnlab:
@anki-code thank you for your feedback and review!

I added your suggestions, but not sure if I managed to capture what you have in mind. let me know if I am in a wrong direction or if I could do it better in any way. 

thank you!

xmnlab:
mm it seems I would need to update one test there. I am working on that.

xmnlab:
btw, the `run-tests.xsh` is crashing sometimes with this error:

```
xonsh.tools.XonshError: xonsh: subprocess mode: command not found: 'xonsh'
```

for example, a test in the tests_scripts function is raising this error locally:

```
- xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
  - Traceback (most recent call last):
  -   File "/home/xmn/mambaforge/envs/xonsh/lib/python3.11/site-packages/xonsh/procs/specs.py", line 473, in _run_binary
  -     p = self.cls(cmd, bufsize=bufsize, **kwargs)
  -         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  -   File "/home/xmn/mambaforge/envs/xonsh/lib/python3.11/subprocess.py", line 1022, in __init__
  -     self._execute_child(args, executable, preexec_fn, close_fds,
  -   File "/home/xmn/mambaforge/envs/xonsh/lib/python3.11/subprocess.py", line 1899, in _execute_child
  -     raise child_exception_type(errno_num, err_msg, err_filename)
  - FileNotFoundError: [Errno 2] No such file or directory: 'xonsh'
  - 
  - During handling of the above exception, another exception occurred:
  - 
  - Traceback (most recent call last):
  -   File "/home/xmn/mambaforge/envs/xonsh/lib/python3.11/site-packages/xonsh/procs/pipelines.py", line 166, in __init__
  -     proc = spec.run(pipeline_group=pipeline_group)
  -            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  -   File "/home/xmn/mambaforge/envs/xonsh/lib/python3.11/site-packages/xonsh/procs/specs.py", line 455, in run
  -     p = self._run_binary(kwargs)
  -         ^^^^^^^^^^^^^^^^^^^^^^^^
  -   File "/home/xmn/mambaforge/envs/xonsh/lib/python3.11/site-packages/xonsh/procs/specs.py", line 492, in _run_binary
  -     raise xt.XonshError(e)
  - xonsh.tools.XonshError: xonsh: subprocess mode: command not found: 'xonsh'
  - Command 'xonsh' not found, but can be installed with:
  - sudo apt install xonsh
    5
    4
```

I am using a conda environment .. not sure if that is the problem (I am saying that because on CI it is not using conda env for the test)

any tips would be very appreciated

xmnlab:
thanks for the review and the guidance @anki-code !

jnoortheen:
pre-commit.ci autofix

jnoortheen:
Thanks @xmnlab !

PR #5075, opened 2023-03-01
Using `repr(cmd)` for `command not found`

anki-code:
The `command not found` error will show the ``repr(cmd)`` to uncover the cases when the command name has ``\n``, ``\t`` or not visible color codes and raises the error.

Before:

```xsh
# For example newbie don't know about @$() and trying to do something like this:

aliases['my'] = 'echo echo'

# BEFORE:

$(@(my))
# xonsh: subprocess mode: command not found: echo      <--- unclear output
#

# AFTER:

$(@(my))
# xonsh: subprocess mode: command not found: 'echo\n'   <-- pretty clear output
```

## For community

### DIFF
--- a/news/not_found_repr.rst
+++ b/news/not_found_repr.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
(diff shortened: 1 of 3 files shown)


jnoortheen (approved):

jnoortheen:
Thanks @anki-code !

PR #5077, opened 2023-03-02
Fix missing webconfig/js/xonsh_sticker.svg

anki-code:
Closes #5060

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #5079, opened 2023-03-02
Remove webconfig image

anki-code:
subj 

follow up #5077

## For community


pyproject.toml
```diff
@@ -61,7 +61,7 @@ xontrib = ["*.xsh"]
     "js/app.min.js",
     "js/bootstrap.min.css",
     "js/LICENSE-bootstrap",
-    "js/xonsh_sticker.svg",
```
jnoortheen, line 64: `.gif ` is not an optimal image format. can we use something like png?

xonsh/webconfig/index.html
```diff
@@ -11,7 +11,6 @@
 <body>
 <nav class="navbar navbar-expand-lg navbar-dark bg-dark">
     <a class="navbar-brand" href="#">
-        <img src="/js/xonsh_sticker.svg" width="30" height="30" class="d-inline-block align-top" alt="">
```
jnoortheen, line 14: I suggest adding back this image while optimizing svg file for its size. svg will look good on all sizes of displays.

PR #5097, opened 2023-03-17
Decorator to register function as an alias

anki-code:
We have pattern:
```xsh
def hello():                  # step 1
    echo world
aliases['hello'] = hello      # step 2
del hello                     # step 3

hello
# world
```
I want to simplify this. This PR adds decorator:
```xsh
@aliases.add                  # just step 1
def __hello():
    echo world

hello
# world
```
It looks pretty sane.

Note:
* The name of function should begin with `__` to avoid intersection between function name and alias.
* There will be `'__hello': None` in `globals()` after running this. 

## For community

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -99,6 +99,17 @@ class Aliases(cabc.MutableMapping):
 
         return wrapper
 
+    def add(self, func):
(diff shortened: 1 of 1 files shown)


jnoortheen:
I remember one such decorator in the code base already. Could you check @anki-code ?

jnoortheen:
Yes it strips functions with leading underscore. double underscore mangles the namespace and will make functions

PR #5128, opened 2023-04-26
News only about xontribs

anki-code:
subj

## For community


jnoortheen:
These are xontribs outside of this repo, so giving out links to the report would be more useful. I would suggest to make this as a post in the reddit page or some social page.

PR #5156, opened 2023-06-15
Py 312 pre-liminary support

jnoortheen:
## For community

### DIFF
--- a/tests/test_parser.py
+++ b/tests/test_parser.py
@@ -5,7 +5,7 @@ import textwrap
 
 import pytest
 
-from xonsh.ast import AST, Call, Pass, Str, With
+from xonsh.ast import AST, Call, Pass, With, is_const_str
 from xonsh.parser import Parser
 from xonsh.parsers.fstring_adaptor import FStringAdaptor
 from xonsh.pytest.tools import (
@@ -2795,7 +2795,7 @@ def test_macro_call_one_arg(check_xonsh_ast, s):
     assert isinstance(tree, AST)
     args = tree.body.args[1].elts
     assert len(args) == 1
-    assert args[0].s == s.strip()
+    assert args[0].value == s.strip()
 
 
 @pytest.mark.parametrize("s,t", itertools.product(MACRO_ARGS[::2], MACRO_ARGS[1::2]))
@@ -2805,8 +2805,8 @@ def test_macro_call_two_args(check_xonsh_ast, s, t):
     assert isinstance(tree, AST)
     args = tree.body.args[1].elts
     assert len(args) == 2
-    assert args[0].s == s.strip()
-    assert args[1].s == t.strip()
+    assert args[0].value == s.strip()
+    assert args[1].value == t.strip()
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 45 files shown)


jnoortheen:
If merged, lets keep the commits separate instead of squashing.

jnoortheen:
reported here https://github.com/python/cpython/issues/105881

jnoortheen:
It seems multiple slashes will no longer be resolved and be considered UNC paths from now on. So we will continue to skip the test from `py3.12` onwards

jnoortheen:
@anki-code would you be able to test it?

jnoortheen:
around 30m gets saved in tests with https://github.com/xonsh/xonsh/pull/5156/commits/3898d9e2e0ff3c368a7919316d31e56f290947d7
image

image

tests/test_parser.py
```diff
@@ -168,6 +168,8 @@ def test_string_literal_concat(first_prefix, second_prefix, check_ast):
 
 
 def test_f_env_var(check_xonsh_ast):
+    if VER_MAJOR_MINOR > (3, 11):
+        pytest.xfail("f-string with special syntax are not supported yet")
```
gforsyth, line 172: Can we open a tracking issue for this so it doesn't get missed?

tests/xoreutils/test_which.py
```diff
@@ -1,14 +1,15 @@
...
 from xonsh.tools import ON_WINDOWS
 from xonsh.xoreutils import _which
 
 
-class TestWhich:
+class TestWhich(TestCase):
     # Tests for the _whichgen function which is the only thing we
     # use from the _which.py module.
-    def setup(self):
+    def setUp(self):
```
gforsyth, line 12:
```suggestion
    @classmethod
    def setup_class(cls):
```

I think this is the standard way to handle class-wrapped setup and teardown in pytest (will require swapping out all the `self` below, though).  Not a blocker -- I am not a fan of test classes anyway.

xonsh/tokenize.py
```diff
@@ -83,9 +84,7 @@
 from xonsh.lazyasd import LazyObject
 from xonsh.platform import PYTHON_VERSION_INFO
 
-HAS_WALRUS = PYTHON_VERSION_INFO > (3, 8)
-if HAS_WALRUS:
-    from token import COLONEQUAL  # type:ignore
+HAS_WALRUS = True
```
gforsyth, line 87: I think we can remove this entirely, and then skip the checks on lines 128 and 235.

gforsyth (commented):
This looks like a good start to me @jnoortheen!

I'd like to remove the remaining `HAS_WALRUS` checks, but except for that none of my other comments are blocking.

tests/test_parser.py
jnoortheen: https://github.com/xonsh/xonsh/issues/5166 done

tests/xoreutils/test_which.py
jnoortheen: setup_class will run at class level only during the start of the test. It is a not straight equivalent to the setup of nosetest . For now we can use the unit-test and later refactor to a pytest if needed.

xonsh/tokenize.py
jnoortheen: done

gforsyth (approved):
This looks good to me, @jnoortheen !

I'll leave it for you to merge in case there's anything you'd like to tweak before it goes in.

PR #5160, opened 2023-06-26
Unset the default Semantic Prompt line continuation environment variables

eugenesvk:
Unset the default Semantic Prompt line continuation environment variables (``$MULTILINE_PROMPT_PRE`` and ``$MULTILINE_PROMPT_POS``) to allow differentiating between user setting an empty value (so nothing should be added) and not setting anything (so it's ok to set them to something useful)

## For community

### DIFF
--- a/news/multil.rst
+++ b/news/multil.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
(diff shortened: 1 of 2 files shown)


jnoortheen (approved):

jnoortheen:
now it is not doing anything different based on the user set value. But it can be useful in the future. Thanks @eugenesvk !

Issue #4929, opened 2022-08-28
Add `history delete` 


anki-code:
We need two things:
1. `history delete <pattern>` command that will delete certain command from history
2. ~The way to mark commands and skip it while flushing history on disk i.e. `echo 123 #skip-history`. In bash this work by [adding the space](https://stackoverflow.com/questions/6475524/how-do-i-prevent-commands-from-showing-up-in-bash-history) before command.~ (implemented, see [$XONSH_HISTORY_IGNORE_REGEX](https://xon.sh/envvars.html#xonsh-history-ignore-regex))

### Discussed in https://github.com/xonsh/xonsh/discussions/4928

<div type='discussions-op-text'>

<sup>Originally posted by **manujchandra** August 28, 2022</sup>
I recently processed some OTP codes URI on the terminal using python qr library. It contains sensitive information regarding token IDS.

After that, I issued the commands history flush and history clear. But there are two problems:

1) When I press the up arrow, it refills the previous command which contains the sensitive information and it keeps going back in history.
2) When I write qr it tries to autocompletes the full otp uri with sensitive codes (in gray color).

How do I delete the history in such a manner that there is no autocomplete and also no history when I press the up button. Because that data should not get flashed during screen sharing accidentally.

![Screenshot_2022-08-28_19-49-14](https://user-images.githubusercontent.com/15191493/187078662-449c6b3e-a542-4ac1-bf02-20c14531f2a2.png)

</div>

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #5172, opened 2023-07-12
Add `history delete` #4929

HackTheOxidation:
This MR adds the `delete` command to `history`, which allows the user to only delete commands from the history,
that matches a specific pattern (in the form of a regular expression).

This is done by extending `xonsh.history.base.History` and `xonsh.history.main.HistoryAlias` as well as provide default implementations for the `JsonHistory` and `SqliteHistory` backends.

## For community

### DIFF
--- a/xonsh/history/sqlite.py
+++ b/xonsh/history/sqlite.py
@@ -223,7 +223,7 @@ def xh_sqlite_wipe_session(sessionid=None, filename=None):
 
 def xh_sqlite_delete_input_matching(pattern, filename=None):
     """Deletes entries from the database where the input matches a pattern."""
-    sql = f"DELETE FROM xonsh_history WHERE inp LIKE '{pattern}' ESCAPE '\\';"
+    sql = f"DELETE FROM xonsh_history WHERE inp REGEXP '{pattern}' ESCAPE '\\';"
     with _xh_sqlite_get_conn(filename=filename) as conn:
         c = conn.cursor()
         _xh_sqlite_create_history_table(c)
--- a/xonsh/history/sqlite.py
+++ b/xonsh/history/sqlite.py
@@ -387,3 +387,6 @@ class SqliteHistory(History):
         self.cwds = []
 
         xh_sqlite_wipe_session(sessionid=self.sessionid, filename=self.filename)
+
+    def delete(self, pattern):
+        print('You called delete!')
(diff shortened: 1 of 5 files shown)


xonsh/history/sqlite.py
```diff
@@ -220,6 +221,15 @@ def xh_sqlite_wipe_session(sessionid=None, filename=None):
         c.execute(sql, (str(sessionid),))
 
 
+def xh_sqlite_delete_input_matching(pattern, filename=None):
+    """Deletes entries from the database where the input matches a pattern."""
+    sql = f"DELETE FROM xonsh_history WHERE inp LIKE '{pattern}' ESCAPE '\\';"
```
jnoortheen, line 226: Here the like pattern is different from regex pattern right?

HackTheOxidation: Yes, I should probably rename it to `sql_pattern` or something similar to make it clear?

jnoortheen: what I meant was that the json implementation and this is not the same. `LIKE` operator doesn't implement the regex

HackTheOxidation: Ah, I think I understand :sweat_smile: 

When reading the sqlite documentation for [REGEXP](https://www.sqlite.org/lang_expr.html),
it says that there is no `regexp()`-function defined by default.

Is there anything in the `xonsh` code base that defines this, or is it fine to just assume that it is defined? :thinking:

jnoortheen: The base function can iterate over the entries and delete those matching the pattern. The sub-classes can implement an option to delete an item at the given index

jnoortheen (approved):

xonsh/history/sqlite.py
```diff
@@ -220,6 +221,17 @@ def xh_sqlite_wipe_session(sessionid=None, filename=None):
...
 
+def xh_sqlite_delete_input_matching(pattern, filename=None):
+    """Deletes entries from the database where the input matches a pattern."""
+    with _xh_sqlite_get_conn(filename=filename) as conn:
+        c = conn.cursor()
+        _xh_sqlite_create_history_table(c)
+        for inp, *_ in _xh_sqlite_get_records(c):
+            if pattern.match(inp):
+                sql = f"DELETE FROM xonsh_history WHERE inp LIKE '{inp}'"
+                c.execute(sql)
```
jnoortheen, line 232: Instead of using like you can use = sign. If the input itself had '%' in it , it might not properly escape it.

HackTheOxidation: Great! Thanks. I've changed the code to use the '='-operator instead. :+1:

jnoortheen:
LGTM, Thanks @HackTheOxidation !

PR #5193, opened 2023-08-23
Mention 3.6 instead of 3.5 in docs

ShalokShalom:
The homepage mentions 3.6, so I update this section to be in line, hope this doesn't need a news item.

## For community


bestlem:
As the tests are run only on python 3.9 and up is it correct to suggest 3.6 is supported.

ShalokShalom:
So 3.9 is more appropriate?

PR #5232, opened 2023-11-20
add generic on_pre_spec_run and on_post_spec_run

### LINKED ISSUE #5231: General on_pre_spec_run that runs for all commands
I want something like https://xon.sh/events.html#event-categories but for all commands, like just an on_pre_spec_run event. 

> Ahh yeah, you want something like: https://xon.sh/events.html#event-categories but for all commands, like just an on_pre_spec_run event. That would need to be added

https://github.com/xonsh/xonsh/issues/3767#issuecomment-692264398

Joshix-1:
Closes  #5231

## For community


Joshix-1:
like this?

PR #5261, opened 2024-01-08
Actually fully disable system clipboard (#5155)

### LINKED ISSUE #5155: `$XONSH_COPY_ON_DELETE = False` not respected when `$VI_MODE = True`
When `$VI_MODE = True`, deleting any chunk of text (eg `<esc>dd` to delete the current line) will copy it to the clipboard. Even when explicitly adding `$XONSH_COPY_ON_DELETE = False` (which should be the default behaviour) this still occurs.

## minimum reproducible `xonfig`

```python
$VI_MODE = True
$XONSH_COPY_ON_DELETE = False
```
Version info:
```bash
~ $ xonsh --version
xonsh/0.13.4
```
```
iterm2: build 3.4.19
macos: 13.4 (22F66)
```

azertyfun:
Fixes #5155

In VI mode, disabling "delete" hotkeys isn't enough.

However not using the `PyperclipClipboard` seems like a better solution to the problem anyway (at least to me, maybe I missed something as I am far from a PTK specialist!).

The configuration variable has been renamed for consistency, though the old one still works.

## For community


xonsh/environ.py
```diff
@@ -1689,7 +1689,12 @@ class PTKSetting(PromptSetting):  # sub-classing -> sub-group
     )
     XONSH_COPY_ON_DELETE = Var.with_default(
         False,
-        "Whether to copy words/lines to clipboard on deletion (must be set in .xonshrc file)."
+        "Whether to copy words/lines to clipboard on deletion (must be set in .xonshrc file). (deprecated)"
+        "Only available under the prompt-toolkit shell.",
```
jnoortheen, line 1693: `.xonshrc` can be changed to something generic. The config file can live in multiple places.

xonsh/ptk_shell/shell.py
```diff
@@ -191,10 +148,14 @@ def __init__(self, **kwargs):
         self.push = self._push
 
         ptk_args.setdefault("history", self.history)
-        if not XSH.env.get("XONSH_COPY_ON_DELETE", False):
-            disable_copy_on_deletion()
-        if HAVE_SYS_CLIPBOARD:
-            ptk_args.setdefault("clipboard", PyperclipClipboard())
+        if HAVE_SYS_CLIPBOARD and (
+            XSH.env.get("XONSH_USE_SYSTEM_CLIPBOARD", False)
+            or XSH.env.get("XONSH_COPY_ON_DELETE", False)
```
jnoortheen, line 153: Here the copy_on_delete controls all copy actions. It should only override the on_delete functions

azertyfun: My understanding is that outside of VI mode, "copy on delete" is the only possible clipboard interaction anyway.  
So I was thinking that just deprecating the `XONSH_COPY_ON_DELETE` and making it equivalent to `XONS_USE_SYSTEM_CLIPBOARD` would be good enough since there would be no behavior change if `$VI_MODE == False`, and no-one using VI mode would have used this setting anyway (at least I can't think of a use-case for using the system clipboard for `p` but the in-memory clipboard for `dd`?). Hence #5155.

Of course I could leave the old implementation be, but it will be messier so I want to be sure that there is a real use-case we'd be supporting.

jnoortheen: With PTK, we can select and copy the text using shift and ctrl+c like text editors. This would be disabling that case. Also ctrl+x is also implemented in PTK

news/disable-system-clipboard.rst
```diff
@@ -0,0 +1,23 @@
...
+
+* <news item>
+
+**Changed:**
+
+* Disabled PTK copying/pasting from system clipboard in VI mode by default (can be re-enabled with ``$XONSH_USE_SYSTEM_CLIPBOARD``).
+
+**Deprecated:**
+
+* Deprecated ``$XONSH_COPY_ON_DELETE`` in favor of ``$XONSH_USE_SYSTEM_CLIPBOARD``.
```
jnoortheen, line 11: There should be a reason to deprecate the current behaviour. Also a warning msg need to be added for users.

jnoortheen:
I'd suggest adding a separate config variable to control system clipboard integration if there is not one already. This on_delete function controls behaviour of all kill-ring operations

jnoortheen (approved):

jnoortheen:
Thanks @azertyfun !

Issue #4469, opened 2021-09-11
Running the quoted command with the quoted arguments


jaraco:
The tutorial says:

> While all Python code is also xonsh, not all Bash code can be used in xonsh. That would defeat the purpose, and Python is better anyway! Still, xonsh is Bash-wards compatible in the ways that matter, such as for running commands, reading in the Bash environment, and utilizing Bash tab completion.

Today I discovered a case where "running commands" isn't Bash-wards compatible. In an [issue I reported to PyCharm](https://youtrack.jetbrains.com/issue/PY-50694), PyCharm attempts to load the environment from the user's shell, but it fails when the user's shell is xonsh. PyCharm attempts to run the command equivalent to:

```
$ /Users/jaraco/.local/bin/xonsh -l -i -c "'/Applications/PyCharm.app/Contents/bin/printenv' 'foo'"
'/Applications/PyCharm.app/Contents/bin/printenvfoo'
```

I presume they're putting printenv and foo in single quotes because either filename could potentially have spaces in the name.

I've advised them that the uncaptured subprocess syntax (`$[]`) may be necessary for xonsh and requested they infer the presence of xonsh to apply that syntax, but I wanted to report the issue here too in case there's something that I've missed or something that xonsh could do to support this fairly simple use-case that appears to violate the shell compatibility goal of xonsh.

Is there a Bash-compatible syntax that would work in this scenario?

anki-code:
In the nutshell the problem is that the command:
```python
'echo' '123'
```
returns string `'echo123'` instead of interpret the line as a command with two arguments and run `echo 123`.

@gforsyth @daniel-shimon @laloch what do you think about parse this kind of commands as a list of arguments?

jaraco:
I like the idea of parsing the syntax as a list of arguments. I’ve frequently stumbled using xonsh to invoke a command with spaces in the name. The trickiest edge case might be the simplest:

```
‘cat’
```

Should that invoke the command cat or print a simple string? I’d argue if the above proposal is accepted, it should invoke the command cat.

jaraco:
I'm just checking in on this again. It still affects me and upstream, PyCharm is awaiting a call on this issue before taking any special action to support xonsh.

anki-code:
@jaraco I think the best way is to add explicit xonsh support like you suggested.

gforsyth:
[...]

PR #5273, opened 2024-01-29
Mention VS Code supports xonsh as default shell

jaraco:
Ref #4469

Also, updated the text to use 'VS Code', which is the preferred short name.

## For community


jaraco:
I have tested compiling the file using rst2html.py and this change does not introduce any compilation errors or warnings.

PR #5295, opened 2024-03-07
feat(bashpath): add env-var to allow overriding path to bash

gforsyth:
I've run into issues with bash completion failing when using `nix` to handle environments.
The nix environment often installs its own `bash` which doesn't respect the `BASH_COMPLETIONS` env-var.

This allows an (off-by-default) option to override what is returned by `xonsh.platform.bash_command`.

The removal of the `lru_cache` should be negligible since `windows_bash_command` has its own `lru_cache` and the other options are doing an env lookup or returning a string.

## For community


gforsyth:
updated.

PR #5327, opened 2024-04-09
Flatten redirects before passing command to job control

yaxollum:
Fixes error pointed out by @jnoortheen in https://github.com/xonsh/xonsh/pull/5322#issuecomment-2045091900

## For community


jnoortheen:
Thanks for the quick fix !

PR #5330, opened 2024-04-10
New parameters "no_rc" and "norccmd" for foreign_shell_data

rautyrauty:
- fix source-foreign bug: avoid sources a personal initialization file

## About bug

When trying to run source-bash with some file, it also sources ~/.bashrc

## For community

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -506,6 +506,7 @@ def source_foreign_fn(
         seterrpostcmd=seterrpostcmd,
         show=show,
         dryrun=dryrun,
+        no_rc=True,
         files=files,
     )
     if fsenv is None:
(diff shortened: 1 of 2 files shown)


rautyrauty:
Should I change something?

gforsyth:
Thanks for putting this in @rautyrauty -- I don't think we need an additional keyword argument for what you are looking for -- if a foreign shell isn't started in `interactive` mode, then the default RC files won't be loaded.  I"m not sure if there's a use-case for starting a foreign shell in interactive mode _without_ the RC file.

That said, this will be a change from how things currently work and there are previous issues detailing why we err on the side of interactive mode.  We'll have to look over those again and make sure this won't break a bunch of existing users.

gforsyth:
Yeah, on thinking about it, I'm good with setting `interactive=False` to disable loading rc files in `source-foreign` if you'd like to put in that change.

rautyrauty:
Yes, please do it.

jnoortheen:
superseded by #5339 which will make non-interactive mode the default . Also, even now it is possible use `source-bash --non-interactive` to not load rc files.

PR #5339, opened 2024-04-16
refactor(aliases): source foreign shell funcs without interactive mode

### LINKED ISSUE #5530: DeprecationWarning for Constant.__init__(is_raw)
## Current Behavior

While troubleshooting https://github.com/xonsh/xonsh/issues/3189, I encountered a DeprecationWarning:

```
C:\Users\jaraco\draft\.VENV\Lib\site-packages\xonsh\ast.py:145: DeprecationWarning: Constant.__init__ got an unexpected keyword argument 'is_raw'. Support for arbitrary keyword arguments is deprecated and will be removed in Python 3.15.
```

## xonfig

<details>

```xsh
jaraco@bulletproof ~\draft @# xonfig
+-----------------------------+----------------------+
| xonsh                       | 0.17.0               |
| Git SHA                     | 5332d755             |
| Commit Date                 | Jun 18 08:47:50 2024 |
| Python                      | 3.13.0               |
| PLY                         | 3.11                 |
| have readline               | False                |
| prompt toolkit              | 3.0.47               |
| shell type                  | prompt_toolkit       |
| history backend             | json                 |
| pygments                    | 2.18.0               |
| on posix                    | False                |
| on linux                    | False                |
| on darwin                   | False                |
| on windows                  | True                 |
| on cygwin                   | False                |
| on msys2                    | False                |
| is superuser                | True                 |
| default encoding            | utf-8                |
| xonsh encoding              | utf-8                |
| encoding errors             | surrogateescape      |
| xontrib                     | []                   |
| RC file                     | []                   |
| UPDATE_OS_ENVIRON           | False                |
| XONSH_CAPTURE_ALWAYS        | False                |
| XONSH_SUBPROC_OUTPUT_FORMAT | stream_lines         |
| THREAD_SUBPROCS             | True                 |
| XONSH_CACHE_SCRIPTS         | True                 |
+-----------------------------+----------------------+
```

</details>

## For community

gforsyth:
Resolves #5330

Sourcing a foreign shell file shouldn't require interactive mode (which will cause `rc` files to also load/execute).

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -86,7 +86,8 @@ class Aliases(cabc.MutableMapping):
     @tp.overload
     def register(
         self, name: str, *, dash_case: bool = True
-    ) -> tp.Callable[[types.FunctionType], types.FunctionType]: ...
+    ) -> tp.Callable[[types.FunctionType], types.FunctionType]:
+        ...
 
     def register(self, func_or_name, name=None, dash_case=True):
         """Decorator to register the given function by name."""
@@ -406,7 +407,7 @@ def xonsh_reset(args, stdin=None):
 def source_foreign_fn(
     shell: str,
     files_or_code: Annotated[list[str], Arg(nargs="+")],
-    interactive=True,
+    interactive=False,
     login=False,
     envcmd=None,
     aliascmd=None,
@@ -433,7 +434,7 @@ def source_foreign_fn(
         Name or path to the foreign shell
     files_or_code
         file paths to source or code in the target language.
-    interactive : -n, --non-interactive
+    interactive : -i, --interactive
         whether the sourced shell should be interactive
     login : -l, --login
         whether the sourced shell should be login
@@ -549,6 +550,9 @@ def source_foreign_fn(
 
 source_foreign = ArgParserAlias(
     func=source_foreign_fn, has_args=True, prog="source-foreign"
+)
+source_foreign.parser.add_argument(
+    "-n", "--non-interactive", action="store_false", dest="interactive"
 )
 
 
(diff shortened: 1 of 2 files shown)


xonsh/aliases.py
```diff
@@ -406,7 +406,7 @@ def xonsh_reset(args, stdin=None):
 def source_foreign_fn(
     shell: str,
     files_or_code: Annotated[list[str], Arg(nargs="+")],
-    interactive=True,
+    interactive=False,
```
jnoortheen, line 409: if the default becomes False, the argument should be updated as `--interactive` from `--non-interactive` . It is specified in the docstring

gforsyth: Ahh, right.  @jnoortheen do you have a pattern with the argparse aliases to keep the argument `-n` around but have it be a no-op?  I realized there are a few downstream libraries that depend on the `-n` flag.

xonsh/aliases.py
```diff
@@ -550,6 +550,9 @@ def source_foreign_fn(
 source_foreign = ArgParserAlias(
     func=source_foreign_fn, has_args=True, prog="source-foreign"
 )
+source_foreign.parser.add_argument(
```
jnoortheen, line 553: Though it is correct, adding it here would cause the parser to be built during import time. Instead we can update the parameters.

gforsyth: Is there currently a way of specifying multiple parameter sets for a single variable?  Otherwise I can make it lazy

jnoortheen: Yes , the ArgParserAlias can be sub-classed. It will work with the other composed functions as well.

jnoortheen: here is the change  that would do it and work for source-bash/zsh 
https://github.com/xonsh/xonsh/commit/36b05d3d74afd09953dbfdadf378ae39b82f289e

jnoortheen: https://github.com/xonsh/xonsh/compare/main...jnoortheen:xonsh:update_source_fn_interactive

PR #5340, opened 2024-04-17
Fix jobs.py: list index out of range

anki-code:
Fixes https://github.com/xonsh/xonsh/issues/2544#issuecomment-2059957967

Closes #2544

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #5344, opened 2024-04-18
refactor(aliases): source foreign shell funcs without interactive mode

gforsyth:
supercedes #5343 because I forgot which remote I was on

### DIFF
--- a/news/source_foreign_notinteractive.rst
+++ b/news/source_foreign_notinteractive.rst
@@ -1,0 +1,25 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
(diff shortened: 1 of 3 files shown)


jnoortheen (approved):

gforsyth:
ugh, I don't have time right now to deal with the venv plugin stuff -- I'll try to take a look later

jnoortheen:
> ugh, I don't have time right now to deal with the venv plugin stuff -- I'll try to take a look later

It is not possible to fix on our end. It seems to be an upstream issue. The best option is to move it out to a virtualenv plugin .

jnoortheen:
PS https://github.com/xonsh/xonsh/pull/5340#issuecomment-2061175951

PR #5351, opened 2024-04-21
Update posix.py: blind fix

anki-code:
ignore


xonsh/procs/posix.py
```diff
@@ -463,7 +463,7 @@ def signal(self):
         """Process signal, or None."""
         s = getattr(self.proc, "signal", None)
         if s is None:
-            rtn = self.returncode
+            rtn = self.proc.returncode
```
jnoortheen, line 466: this is actually a propery setter. https://github.com/xonsh/xonsh/blob/aa857e8260478f81e8c427cce1e7486f44a4b048/xonsh/procs/posix.py#L457

Issue #4577, opened 2021-11-30
Function aliases in subprocesses can cause Xonsh to exit


mingmingrr:
Any aliases that are python functions has a seemingly random chance of causing Xonsh to exit, if they are used at the ends of pipes inside subprocesses.

## xonfig

<details>

```
$ xonfig
+------------------+-----------------+
| xonsh            | 0.11.0          |
| Python           | 3.9.6           |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | 3.0.19          |
| shell type       | readline        |
| history backend  | json            |
| pygments         | 2.10.0          |
| on posix         | True            |
| on linux         | True            |
| distro           | unknown         | (NixOs 22.04 unstable)
| on wsl           | False           |
| on darwin        | False           |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
| on jupyter       | False           |
| jupyter kernel   | None            |
| xontrib          | []              |
| RC file          | []              |
+------------------+-----------------+
```

</details>

## Expected Behavior

With `aliases['foo'] = lambda args: 0`, `$(echo | foo)` should produce `''`.

## Current Behavior

Around 1/4 of the time, `$(echo | foo)` triggers a TypeError and causes Xonsh to exit.

```xonsh
$(foo) # works fine
$(echo | foo | cat) # works fine
echo | foo # works fine
```

### Traceback (if applicable)

<details>

```
[bash] $ # shell type doesn't matter, ptk just produces an extra termios error 
[bash] $ XONSH_DEBUG=1 XONSH_SHOW_TRACEBACK=1 xonsh --no-rc --shell-type rl
mingmingrr@potato ~ master $  history files for 0.0160s.
mingmingrr@potato ~ master $ aliases['foo'] = lambda args: 0
mingmingrr@potato ~ master $ $(echo | foo)
''
mingmingrr@potato ~ master $ $(echo | foo)
''
mingmingrr@potato ~ master $ $(echo | foo)
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "/nix/store/7pxlzy2j9dnz6aagg6j6lyni12f1psm8-xonsh-0.11.0/lib/python3.9/site-packages/xonsh/__amalgam__.py", line 16776, in default
[...]

PR #5361, opened 2024-04-24
Read stop signals from the process and update the process state.

### LINKED ISSUE #4752: Alias doesn't set returncode
## xonfig

<details>

```
$ xonfig
| xonsh            | 0.10.1               |
| Git SHA          | e9b12c8b             |
| Commit Date      | Jul 24 22:47:37 2021 |
| Python           | 3.9.10               |
| on darwin        | True                 |
| xontrib 1        | autoxsh              |
| xontrib 2        | back2dir             |
| xontrib 3        | coreutils            |
| xontrib 4        | output_search        |
| xontrib 5        | pipeliner            |
| xontrib 6        | powerline3           |
```

</details>

## Expected Behavior
Simple shlex aliases set the return code and other properties correctly

```
$ aliases["falsy"] = "false"
$ !(falsy).returncode
1
```

More complicated aliases such as those with a redirect get handled through the Alias mechanism, but I would expect the returned CommandPipeline to have the same properties as if the alias text were executed directly.

## Current Behavior

Instead, Alias calls always give 0 status. For instance, adding a redirect forces the alias to be an Alias:

```
$ aliases['negate'] = "false e>/dev/null"
$ !(negate).returncode
0
```

I think that other parameters such as IO streams are not set either. For instance, `aliases['dog'] = xonsh.aliases.ExecAlias('cat')` does not set stdin and stdout correctly, while `aliases['dog'] = 'cat'` works fine.

## For community

anki-code:
Reading stop signals from the process and update the process state.

### The issue

Technically. In a couple of places that critical for processing signals we have `os.waitpid()`. The function behavior is pretty unobvious and one of things is processing return code after catching the signal. We had no good signal processing around this and this PR fixes this. See also `proc_untraced_waitpid` function description.

From user perspective. For example we have process that is waiting for user input from terminal e.g. `python -c "input()"` or `fzf`. If this process will be in captured pipeline e.g. `!(echo 1 | fzf | head)` it will be suspended by OS and the pipeline will be in the endless loop with future crashing and corrupting std at the end. This PR fixes this.

### The solution

Technically. The key function is `proc_untraced_waitpid` - it catches the stop signals and updates the process state.

From user perspective. First of all we expect that users will use captured object `!()` only for capturable processes. Because of it our goal here is to just make the behavior in this case stable.
In this PR we detect that process in the pipeline is suspended and we need to finish the command pipeline carefully:
* Show the message about suspended process.
* Keep suspended process in `jobs`. The same behavior we can see in bash. This is good because we don't know what process suspended and why. May be experienced user will want to continue it manually.
* Finish the CommandPipeline with returncode=None and suspended=True.

### Before

```xsh
!(fzf) # or !(python -c "input()")
# Hanging / Exceptions / OSError / No way to end the command.
# After exception:
$(echo 1)
# OSError / IO error
```

### After

```xsh
!(fzf) # or `!(ls | fzf | head)` or `!(python -c "input()")`
# Process ['fzf'] with pid 60000 suspended with signal 22 SIGTTOU and stay in `jobs`.
# This happends when process start waiting for input but there is no terminal attached in captured mode.
# CommandPipeline(returncode=None, suspended=True, ...)

$(echo 1)
# Success.
```
Closes #4752 #4577

### Notes

* There is pretty edge case situation when the process was terminated so fast that we can't catch pid alive and check signal ([src](https://github.com/xonsh/xonsh/blob/67d672783db6397bdec7ae44a9d9727b1e20a772/xonsh/jobs.py#L71-L80)). I leave it as is for now. 

### Mentions

#2159

## For community

### DIFF
--- a/xonsh/jobs.py
+++ b/xonsh/jobs.py
@@ -16,7 +16,7 @@ from xonsh.cli_utils import Annotated, Arg, ArgParserAlias
 from xonsh.completers.tools import RichCompletion
 from xonsh.lazyasd import LazyObject
 from xonsh.platform import FD_STDERR, LIBC, ON_CYGWIN, ON_DARWIN, ON_MSYS, ON_WINDOWS
-from xonsh.tools import on_main_thread, unthreadable
+from xonsh.tools import on_main_thread, unthreadable, get_signal_name
 
 # Track time stamp of last exit command, so that two consecutive attempts to
 # exit can kill all jobs and exit.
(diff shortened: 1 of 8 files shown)


xonsh/procs/posix.py
```diff
@@ -168,6 +168,26 @@ def run(self):
...
+                        signal.SIGTTOU,
+                        signal.SIGTTIN,
+                    ]:
+                        try:
+                            self.proc.send_signal(signal.SIGINT)
+                            self.proc.send_signal(signal.SIGCONT)
+                        except ProcessLookupError:
+                            pass
+                except ChildProcessError:
+                    pass
```
jnoortheen, line 189: it will be good to record this errors somewhere, like in the xonsh log if enabled.

jnoortheen:
I came across this lib https://github.com/oconnor663/duct.py/blob/master/gotchas.md . Can this be helpful?

jnoortheen (approved):

jnoortheen:
Yes this is great work. We can take pointers from this repo.

xonsh/jobs.py
```diff
@@ -39,18 +39,84 @@
...
+        return info
+
+    if proc is not None and getattr(proc, "pid", None) is None:
+        """
+        When the process stopped before os.waitpid it has no pid.
+        Note that in this case there is high probability
+        that we will have return code 0 instead of real return code.
+        """
+        if raise_child_process_error:
+            raise ChildProcessError("The process PID not found.")
```
gforsyth, line 78:
```suggestion
            raise ChildProcessError("Process Identifier (PID) not found.")
```

xonsh/procs/pipelines.py
```diff
@@ -513,6 +521,20 @@ def _endtime(self):
...
+    def _procs_suspended(self):
+        """Check procs and return suspended proc."""
+        for proc in self.procs:
+            info = xj.proc_untraced_waitpid(proc, hang=False)
+            if getattr(proc, "suspended", False):
+                proc = getattr(proc, "proc", proc)
+                procname = f"{getattr(proc, 'args', '')} with pid {proc.pid}".strip()
+                print(
+                    f"Process {procname} suspended with signal {info['signal_name']} and stay in `jobs`.\n"
+                    f"This happends when process start waiting for input but there is no terminal attached in captured mode.",
```
gforsyth, line 533:
```suggestion
                print(
                    f"Process {procname} was suspended with signal {info['signal_name']} and placed in `jobs`.\n"
                    f"This happens when a process starts waiting for input but there is no terminal attached in captured mode.",
```

xonsh/jobs.py
```diff
@@ -364,7 +422,7 @@ def format_job_string(num: int, format="dict") -> str:
         "cmd": " ".join(
             [" ".join(i) if isinstance(i, list) else i for i in job["cmds"]]
         ),
-        "pid": int(job["pids"][-1]) if job["pids"] else None,
+        "pids": job["pids"] if "pids" in job else None,
     }
 
     if format == "posix":
```
gforsyth, line 428: If someone passes `jobs --posix`, the code below still refers to the `pid` key but it should be `pids`

gforsyth (commented):
This looks good @anki-code -- I made a few suggestions for the error message and we should update the formatting for `jobs --posix`, but otherwise this can go in.

gforsyth (approved):
:shipit: 

Nice work @anki-code -- I think the helpful notes for the knight is an especially good write-up.  Really awesome stuff here.

PR #5370, opened 2024-04-25
feat: add --no-env flag to avoid inheriting parent-shell env

### LINKED ISSUE #4921: Add `--no-env` argument
xonsh [has `--no-rc`](https://github.com/xonsh/xonsh/blob/6fa83f830579d2bad39b1842911951f55a9da22c/xonsh/main.py#L180-L186) to avoid loading RC files. But if we want to run completely clean xonsh we need to avoid inheritance environment.

For example sometimes I faced with issues when I want to run xonsh from xonsh and `$XONSH_HISTORY_FILE` goes to the new xonsh instance and lock the file system. I believe that the similar logic could produce the issues with any other use cases. For example we had issues around multiple running `xonsh -c ...` because the history backend uses the same file (fixed in #4178).

I suggest to add `--no-env` as well as [`--no-rc`](https://github.com/xonsh/xonsh/blob/6fa83f830579d2bad39b1842911951f55a9da22c/xonsh/main.py#L180-L186) argument that will run xonsh with `default_env` only:

https://github.com/xonsh/xonsh/blob/6fa83f830579d2bad39b1842911951f55a9da22c/xonsh/environ.py#L2499-L2500

So to get clean xonsh instance we can run `xonsh --no-rc --no-env`.

## For community

anki-code:
### Motivation

Basically we use `xonsh --no-rc` to have pure xonsh. 
I noticed many times when parent environment can affect `xonsh --no-rc` e.g. `$XONSH_HISTORY_FILE`.
Current workaround is `env -i @$(which xonsh) --no-rc` but I think we need to have pure xonsh way.

Closes #4921

### Before

```xsh
xonsh --no-rc
$QWE = 1
xonsh --no-rc
$QWE
# 1
```

### After
```xsh
xonsh --no-rc
$QWE = 1
xonsh --no-rc --no-env
$QWE
# Unknown environment variable: $QWE

$XONSH_ENV_INHERITED  # Marker to have a way to catch this situation.
# False
```

## For community

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -611,7 +611,11 @@ class XonshSession:
         if ctx is not None:
             self.ctx = ctx
 
-        self.env = kwargs.pop("env") if "env" in kwargs else Env(default_env() if load_env else {'XONSH_NO_ENV': True})
+        self.env = (
+            kwargs.pop("env")
+            if "env" in kwargs
+            else Env(default_env() if load_env else {"XONSH_NO_ENV": True})
+        )
 
         self.exit = False
         self.stdout_uncaptured = None
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -602,7 +602,7 @@ class XonshSession:
             Xonsh execution object, may be None to start
         ctx : Mapping, optional
             Context to start xonsh session with.
-        load_env : bool
+        inherit_env : bool
             If ``True``: inherit environment variable values from ``os.environ``.
             If ``False``: use default values for environment variables and
             set ``$XONSH_ENV_INHERITED = False``.
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -603,7 +603,7 @@ class XonshSession:
         ctx : Mapping, optional
             Context to start xonsh session with.
         inherit_env : bool
-            If ``True``: inherit environment variable values from ``os.environ``.
+            If ``True``: inherit environment variables from ``os.environ``.
             If ``False``: use default values for environment variables and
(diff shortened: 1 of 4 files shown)


xonsh/main.py
```diff
@@ -191,6 +191,13 @@ def parser():
         action="store_true",
         default=False,
     )
+    p.add_argument(
+        "--no-env",
+        help="Do not load parent environment variables.",
+        dest="noenv",
+        action="store_true",
```
gforsyth, line 198: the double-negation gets confusing, probably better to do this:
```suggestion
        dest="env",
        action="store_false",
```

xonsh/built_ins.py
```diff
@@ -593,7 +593,7 @@ def _restore_python_exit(self):
         if self._py_quit is not None:
             builtins.quit = self._py_quit
 
-    def load(self, execer=None, ctx=None, **kwargs):
+    def load(self, execer=None, ctx=None, load_env=True, **kwargs):
```
gforsyth, line 596: can you add the new kwarg to the docstring?

xonsh/built_ins.py
```diff
@@ -611,7 +611,11 @@ def load(self, execer=None, ctx=None, **kwargs):
         if ctx is not None:
             self.ctx = ctx
 
-        self.env = kwargs.pop("env") if "env" in kwargs else Env(default_env())
+        self.env = (
+            kwargs.pop("env")
+            if "env" in kwargs
+            else Env(default_env() if load_env else {"XONSH_NO_ENV": True})
```
gforsyth, line 617: should this be `env` or `load_env`?

gforsyth (approved):
thanks, @anki-code !

Issue #5304, opened 2024-03-16
Case with endless loop with no way to interrupt


anki-code:
Sometimes I'm facing with loops without the way to interrupt it (MacOS):

```xsh
while 1:
    xonsh -c 'date; sleep 1'
# Ctrl-C
# KeyboardInterrupt
# Ctrl-C
# KeyboardInterrupt
# Ctrl-D
# ^D
# Ctrl-C
# KeyboardInterrupt
# ...
# No way to stop it.
```

We need to add the way to stop the execution.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

JamesParrott:
If this outer loop is run in xonsh, Even if Ctrl+D quits the inner sleep process, the outer while loop will then immediately start another one.

```
while 1:
    xonsh -c 'date; sleep 1'
```

Is this a request to pass an interrupt from a xonsh process, to its calling process? 
https://xon.sh/aliases.html#eof-exit-and-quit

JamesParrott:
I think this functionality might even already exist.  Or at the very least I can't reproduce the failure.  Admittedly it can cause the outer interactive xonsh hang, but I was able to quit the loop in MINGW64 Git Bash on Windows by simply spamming (not holding down) Ctrl+C:

```
@LAPTOP ~ @ while 1:
.........................   xonsh -c 'date; sleep 1'
.........................
Sun, Mar 17, 2024 12:01:35 PM
Sun, Mar 17, 2024 12:01:37 PM
Sun, Mar 17, 2024 12:01:39 PM

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\\.local\bin\xonsh.exe\__main__.py", line 4, in <module>
  File "C:\Users\\.local\pipx\venvs\xonsh\Lib\site-packages\xonsh\main.py", line 2, in <module>
    import argparse
  File "C:\Users\\AppData\Local\Programs\Python\Python311\Lib\argparse.py", line 94, in <module>

    from gettext import gettext as _, ngettext
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1138, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1078, in _find_spec
  File "<frozen importlib._bootstrap_external>", line 1504, in find_spec
[...]

PR #5399, opened 2024-05-07
Fixed processing exit signals and exit exception

### LINKED ISSUE #5381: SIGHUP signal not raises exit immediately
I'm not sure if this is an alacritty issue, or a xonsh issue, so I'm starting here.
When using alacritty in single-instance mode, xonsh processes remain if their parent window is closed by the window manager (e.g. Meta+Q for Sway). I've only been able to find this issue when using both alacritty in single-instance mode and xonsh. Alacritty single-instance with Zsh works as expected, alacritty individual instances works with xonsh, and xonsh with foot's single-instance mode works. The issue also does not occur if xonsh has a parent process other than alacritty (e.g. alacritty -> zsh -> xonsh)

## Expected Behavior

Xonsh processes should exit when parent window closes.

## Current Behavior

Xonsh process remains when parent window is closed.

## Steps to Reproduce

1. `export SHELL=$(which xonsh)` if xonsh isn't your default shell.
2. Start an alacritty instance with `alacritty --socket /tmp/alacritty.sock`
3. Create a second alacritty window with `alacritty msg --socket /tmp/alacritty.sock create-window`. This creates a new terminal attached to the same instance of alacritty. Notice now there are two xonsh processes attached to the same alacritty process.
4. (optional) run some command in the second terminal window to help identify it (e.g. man bash).
5. Close the second terminal window using the window manager instead of exiting xonsh directly. Notice there are still two xonsh processes attached to the alacritty process.

## xonfig

<details>

```
+------------------+----------------------------------+
| xonsh            | 0.16.0                           |
| Python           | 3.11.9                           |
| PLY              | 3.11                             |
| have readline    | True                             |
| prompt toolkit   | 3.0.40                           |
| shell type       | prompt_toolkit                   |
| history backend  | sqlite                           |
| pygments         | 2.17.2                           |
| on posix         | True                             |
| on linux         | True                             |
| distro           | unknown                          |
| on wsl           | False                            |
| on darwin        | False                            |
| on windows       | False                            |
| on cygwin        | False                            |
| on msys2         | False                            |
| is superuser     | False                            |
| default encoding | utf-8                            |
| xonsh encoding   | utf-8                            |
| encoding errors  | surrogateescape                  |
| xontrib 1        | vox                              |
| xontrib 2        | coreutils                        |
| xontrib 3        | voxapi                           |
| RC file 1        | /etc/xonsh/xonshrc               |
| RC file 2        | /home/wombat/.config/xonsh/rc.xsh |
+------------------+----------------------------------+
```

</details>

### System
Distro: Nixos
WM: Sway 1.9
Alacritty: 0.13.2

## For community

anki-code:
### Before

Case 1: Handler catches the exit signal and do not pass it forward. As result xonsh could be suspended by OS or crash. Also exit code is wrong. See repeatable examples with SIGHUP in https://github.com/xonsh/xonsh/issues/5381#issuecomment-2097961804.

Case 2: From bash/zsh as login shell run xonsh. Then send quit signal to xonsh. The terminal state will be broken: disabled SIGINT, mouse pointer produces mouse state codes.

### After

Case 1: Xonsh exit normally with right exit code. Fixed #5381 #5304 #5371. 

Case 2: After exiting with right exit code the state of the terminal is good.

## For community

### DIFF
--- a/news/sig_exit_fix.rst
+++ b/news/sig_exit_fix.rst
@@ -16,7 +16,7 @@
 
 **Fixed:**
 
-* Fixed processing exit signals and exceptions (e.g. SIGHUP in #5381) to provide careful exiting with right exit code.
+* Fixed processing exit signals and exceptions (e.g. SIGHUP in #5381) to provide careful exiting with right exit code and TTY cleaning.
 
 **Security:**
 
--- a/news/sig_exit_fix.rst
+++ b/news/sig_exit_fix.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Fixed processing exit signals (e.g. SIGHUP).
+
+**Security:**
+
+* <news item>
--- a/news/sig_exit_fix.rst
+++ b/news/sig_exit_fix.rst
@@ -16,7 +16,7 @@
 
 **Fixed:**
 
-* Fixed processing exit signals (e.g. SIGHUP).
+* Fixed processing exit signals and exceptions (e.g. SIGHUP in #5381) to provide careful exiting with right exit code.
 
 **Security:**
 
--- a/tests/procs/test_specs.py
+++ b/tests/procs/test_specs.py
@@ -141,6 +141,7 @@ def test_capture_always(
 
 
 @skip_if_on_windows
+@pytest.mark.flaky(reruns=3, reruns_delay=1)
 def test_interrupted_process_returncode(xonsh_session):
     xonsh_session.env["RAISE_SUBPROC_ERROR"] = False
     cmd = [["python", "-c", "import os, signal; os.kill(os.getpid(), signal.SIGINT)"]]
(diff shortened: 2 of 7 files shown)


FlyingWombat:
Tried this out, and while it is exiting when sent SIGHUP now, it doesn't always exit properly. What I noticed is, after xonsh exits, zsh no longer recognizes Ctrl+C.

When `kill -SIGHUP 1234` is run from xonsh itself, it exists correctly (zsh works fine after). But running `kill -SIGHUP 1234` from another terminal running e.g. zsh, goes like this:
```
# running zsh, launch xonsh
xonsh on  sig_exit_fix [?] via 🐍 v3.12.3 (.venv)
❯ xonsh

# send SIGHUP
(.venv) wombat@wombat-pc ~/builds/dev/xonsh sig_exit_fix @ 
%

xonsh on  sig_exit_fix [?] via 🐍 v3.12.3 (.venv) took 14s
❯ ^C^C
# zsh no longer recognizes Ctrl+C

```
If I do the same thing with bash, zsh is working properly afterwards.

FlyingWombat:
The new revision is working for me. Xonsh is now accepting exit signals and exiting properly.

gforsyth (approved):

Issue #5405, opened 2024-05-11
print_exception: rewrite deprecated code


anki-code:
We have this line:

https://github.com/xonsh/xonsh/blob/55b341d47753967c70a4dcf9ff30690877f66048/xonsh/tools.py#L1056-L1058

But https://docs.python.org/3/library/sys.html:

> These three variables are deprecated; 
> sys.last_type
> sys.last_value
> sys.last_traceback
> Use [sys.last_exc](https://docs.python.org/3/library/sys.html#sys.last_exc) instead. They hold the legacy representation of sys.last_exc, as returned from [exc_info()](https://docs.python.org/3/library/sys.html#sys.exc_info) above.

We also have related issue with this code: https://github.com/xonsh/xonsh/issues/5408

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

kulkarniniraj:
I have a fix ready but can't test it. `python xonsh-in-docker.py` gives me error that `xonsh:latest` cannot be pulled.

kulkarniniraj:
Thanks @anki-code. Turns out problem was with python 3.6. Once I passed 3.12 to xonsh-in-docker.py, it built docker image.

anki-code:
This was solved as I remember.

PR #5407, opened 2024-05-12
Print exception fix

kulkarniniraj:
Fix for issue #5405


kulkarniniraj:
@anki-code sure, I'll add a condition based on python version.

PR #5409, opened 2024-05-13
Show ``root`` and ``@#`` in prompt if user is superuser.

anki-code:
### Before

```xsh
sudo xonsh --no-rc
# <user>@host ~ #
```

### After

```xsh
sudo xonsh --no-rc
# root@host ~ @#
```

It's in sync with:

```xsh
docker run -it ubuntu bash
# root@ce01c2856819:/#
echo $USER
# 
# root@ce01c2856819:/#
```
## For community

### DIFF
--- a/xonsh/prompt/base.py
+++ b/xonsh/prompt/base.py
@@ -338,7 +338,8 @@ class PromptFields(tp.MutableMapping[str, "FieldType"]):
         self.update(
             dict(
                 user=xp.os_environ.get(
-                    "USERNAME" if xp.ON_WINDOWS else "USER", "root" if xt.is_superuser() else "<user>"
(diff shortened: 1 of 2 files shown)


xonsh/prompt/base.py
```diff
@@ -338,9 +338,10 @@ def load_initial(self):
         self.update(
             dict(
                 user=xp.os_environ.get(
-                    "USERNAME" if xp.ON_WINDOWS else "USER", "<user>"
+                    "USERNAME" if xp.ON_WINDOWS else "USER",
+                    "root" if xt.is_superuser() else "<user>",
```
gforsyth, line 342: Is this needed? 

What's the situation where the user is a superuser but the username isn't `root`?

gforsyth (approved):

PR #5418, opened 2024-05-15
History: Saving history in case of any type of exiting the shell.

### LINKED ISSUE #2657: Save history when the shell is exiting abnormally
When using the JSON backend, the history is not saved unless the shell is exited properly (for example, by ^D). When the `exec` alias is used, or a crash is induced (by for example adding invalid colors into `$BOTTOM_TOOLBAR`), the history file is not saved, and is left locked without commands being written into it.

My suspicion is that this is due to the fact that exit  handlers are not ran when exec is used. This doesn't happen with SQLite for obvious reasons.

## For community

anki-code:
### Before

```xsh
echo 1 
echo 2
# <<Crash>>
# No json history
```

### After

Closes #2657

```xsh
echo 1 
echo 2
# <<Crash>>
# In json history
# echo 1
# echo 2
```

## For community

### DIFF
--- a/news/history_save_unload.rst
+++ b/news/history_save_unload.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Saving history in case of any type of exiting the shell.
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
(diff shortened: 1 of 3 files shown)


jnoortheen (approved):

jnoortheen:
Thanks @anki-code !

PR #5424, opened 2024-05-16
Refreshing `procs/specs.py`

anki-code:
### Motivation

This PR just clean the code in `procs/specs.py` without changes the logic: group the code into functions. This is needed to the next feature PR #5443.

## For community

### DIFF
--- a/news/refresh_specs.rst
+++ b/news/refresh_specs.rst
@@ -4,7 +4,7 @@
 
 **Changed:**
 
-* Cleaning logic and functions around threading and capturing in ``procs/specs.py``.
+* Cleaning logic and functions around threading and capturing in ``procs/specs.py`` file.
 
 **Deprecated:**
 


news/thr_unthr.rst
```diff
@@ -0,0 +1,24 @@
+**Added:**
+
+* Added ``xth`` (x-threadable) and ``xunth`` (x-unthreadable) aliases to force command running in thread or not.
```
jnoortheen, line 3: It is better to remove threading/unthreading behaviour at all. The main changing factor when running shell is capturing/non-capturing . Threading should be left to the shell to decide. Is there any difference in behaviour when running threaded or unthreaded ?

jnoortheen: OK, now I understand the case where one wants capture explicitly. 

3. We can add it if needed later. As you have mentioned anything that is specified with the surrounding `$()`

jnoortheen (approved):

PR #5439, opened 2024-05-24
Mitigate exception on start up if ``dircolors`` binary is broken

anki-code:
### Motivation

If `dircolors` utility is wrong by any reason you can't start xonsh shell. This should not be stopper.

### Before

```xsh
cd /tmp
echo '#!/bin/qwe' > dircolors && chmod +x dircolors
echo '$PATH.append("/tmp")' > rc
xonsh --no-env --rc /tmp/rc
# OSError: [Errno 8] Exec format error: 'dircolors'
# Xonsh encountered an issue during launch
```

### After

It's working.

## For community

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -80,8 +80,8 @@ from xonsh.tools import (
     path_to_str,
     pathsep_to_upper_seq,
     print_color,
-    print_warning,
     print_exception,
+    print_warning,
     ptk2_color_depth_setter,
     seq_to_upper_pathsep,
     set_to_csv,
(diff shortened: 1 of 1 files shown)


news/fix_dircolors.rst
```diff
@@ -0,0 +1,23 @@
...
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Prevent exception on start up if ``dircolors`` util is wrong.
```
jnoortheen, line 19: It should show a error message in case of failure to read from dircolors. Otherwise the user should be given option to not to use dircolors at all.

jnoortheen: I think if there is an issue in the binary it should be shown to the user atleast. If this is a edge case, we can keep the current logic .

jnoortheen:
Yes this is better than crashing

jnoortheen (approved):

PR #5445, opened 2024-05-26
Revert partial_proxy

anki-code:
In #5366 we removed partial_proxy. But @doronz88 [found the case](https://xonsh.zulipchat.com/#narrow/stream/435525-xonsh-xontrib/topic/.E2.9C.94.20xontrib-z.20and.20maybe.20other.20break.20in.20master) where this functionality is good and needed. In xontrib-z there is a [handler](https://github.com/AstraLuma/xontrib-z/blob/a40abb4d8b00fcc54877647ba64cf94bcb55b0a5/xontrib/z.py#L225-L227) that requires partial_proxy. So I bring partial_proxy back because it looks useful.

## For community

### DIFF
--- a/tests/procs/test_specs.py
+++ b/tests/procs/test_specs.py
@@ -316,6 +316,7 @@ def test_partial_args_from_classmethod(xession):
         def alias(cls, args, stdin, stdout):
             print("ok", file=stdout)
             return 0
+
     xession.aliases["alias_with_partial_args"] = Class.alias
     out = run_subproc([["alias_with_partial_args"]], captured="stdout")
     assert out == "ok"


xonsh/procs/proxies.py
```diff
@@ -283,6 +285,69 @@ def parse_proxy_return(r, stdout, stderr):
     return cmd_result
 
 
+def proxy_zero(f, args, stdin, stdout, stderr, spec, stack):
```
jnoortheen, line 288: this dispatcher can be used where an alias with less number of args can be satisfied. https://github.com/xonsh/xonsh/blob/2d966fb1b56f2399b9753880744645b62133e982/xonsh/cli_utils.py#L369  .

jnoortheen (approved):

PR #5454, opened 2024-05-28
aliases.py: micro fix

anki-code:
It's after all merged features.

## For community


xonsh/aliases.py
```diff
@@ -169,6 +169,7 @@ def eval_alias(
         # only once.
         if (
             isinstance(value, cabc.Iterable)
+            and hasattr(value, 'len')
```
jnoortheen, line 172: do you mean the magic method __len__ ?

PR #5477, opened 2024-06-02
Perform proper case-insensitive matching in command_cache on Windows

### LINKED ISSUE #5469: Windows: case of commands is incorrectly upper-cased on Windows
## Current Behavior

As reported and investigated in https://github.com/rust-lang/rustup/issues/3848#issuecomment-2143100269, on Windows, it seems that xonsh will transform an executable name to a non-canonical uppercase form before invoking in a subprocess.

```
 ~ # import xonsh.environ as xenv
 ~ # xenv.locate_binary('./rustup-init')
'C:\\Users\\jaraco\\RUSTUP-INIT.EXE'
```

This causes problems when an executable is named `rustup-init` and that executable depends on recognizing its own name to dispatch its behavior but fails when `RUSTUP-INIT.EXE` is what appears in ARG0.

This undesirable upper casing is almost certainly happening in one or more of the Windows-specific `upper()` calls in [commands_cache](https://github.com/xonsh/xonsh/blob/main/xonsh/commands_cache.py).

This mismatch of case led to obscure and annoying problems while attempting to install Rust from xonsh using the recommended install technique.

## Expected Behavior

At the very least, xonsh should not alter the case of the characters typed by the user for the command. If they typed `rustup-init`, then those characters should appear in arg0 of the subprocess.

Even better would be to do what Powershell does and use case-insensitive matching to locate the command, but then use the case of the characters in the filename itself for arg0. That way, if the user types `./RUSTUP-INIT`, the arg0 will be `full\path\to\rustup-init.exe`. For example:

```
 ~ # powershell -c '.\\Print_ARG0.py'
C:\Users\jaraco\print_arg0.py
```

## xonfig

<details>

```xsh
 ~ # xonfig
| xonsh            | 0.16.0          |
| Python           | 3.12.2          |
| have readline    | False           |
| prompt toolkit   | 3.0.40          |
| pygments         | 2.18.0          |
| on posix         | False           |
| on windows       | True            |
| is superuser     | True            |
| xontrib 1        | vox             |
| xontrib 2        | voxapi          |
| xontrib 3        | jaraco.xonsh    |
| RC file          | []              |
```

</details>

## For community

### LINKED ISSUE #5476: Windows: Don't imply PATH=.
In 48dac6e9cdaec8f20d7054aa44213fed6803a462, @byk added a few lines to what is now [found](https://github.com/xonsh/xonsh/blob/6245e460a6841b1bc51382265e4bb9afa8c9ec58/xonsh/commands_cache.py#L281-L286) in the command_cache module:

https://github.com/xonsh/xonsh/blob/6f8f82646e95ab5e626de7cf9b0dbf9717e2b2dd/xonsh/environ.py#L770-L772

It's not even obvious from that commit why the behavior was even relevant to enabling sudo.

In my opinion, `xonsh` shouldn't be adopting platform-specific behaviors from other shells but should instead provide a unified "xonsh" experience across platforms. Additionally, the presence of this platform-specific behavior complicates the implementation and adds cognitive load to maintenance.

And while it's It's true that `cmd.exe` will allow execution of bare executables in the current working directory, that's just a cmd.exe behavior. Powershell, for example, does not have that behavior:

```
PS C:\Users\jaraco> Get-ChildItem -File -Filter "*.exe" | Select-Object -ExpandProperty Name
rustup-init.exe
PS C:\Users\jaraco> rustup-init
rustup-init : The term 'rustup-init' is not recognized as the name of a cmdlet, function, script file, or operable program. 
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ rustup-init
+ ~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (rustup-init:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException


Suggestion [3,General]: The command rustup-init was not found, but does exist in the current location. Windows PowerShell does not load commands from the current location by default. If you trust this command, instead type: ".\rustup-init". See "get-help about_Command_Precedence" for more details.
```

xonsh should drop this legacy divergent behavior and replace it with a unified expectation that matches that of xonsh on other platforms and (incidentally) Powershell.

jaraco:
Closes #5476
Closes #5469

- [x] Include a news file (https://xon.sh/devguide.html#changelog).
- [x] ~~Add the documentation for your feature into `/docs`.~~
- [x] Add the example of usage or before-after behavior.
- [x] Mention the issue that this PR is addressing e.g. `#1234`.

## For community

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,6 +15,9 @@ authors = [{ name = "Anthony Scopatz" }, { email = "scopatz@gmail.com" }]
 maintainers = [{ name = "Anthony Scopatz" }, { email = "scopatz@gmail.com" }]
 license = { text = "BSD 2-Clause License" }
 requires-python = ">=3.9"
+dependencies = [
+    "case-insensitive-dictionary; platform_system=='Windows'",
+]
 
 [tool.setuptools.dynamic]
 version = {attr = "xonsh.__version__"}
--- a/tests/test_commands_cache.py
+++ b/tests/test_commands_cache.py
@@ -203,20 +203,22 @@ def test_update_cache(xession, tmp_path):
     assert file2.samefile(cached[basename][0])
 
 
-def test_find_binary_retains_case(tmp_path):
-    faux_binary = tmp_path / "runme.exe"
-    faux_binary.touch()
-    faux_binary.chmod(0o755)
+@pytest.fixture
+def faux_binary(tmp_path):
+    binary = tmp_path / "runme.exe"
+    binary.touch()
+    binary.chmod(0o755)
+    return binary
+
+
+def test_find_binary_retains_case(faux_binary):
     cache = CommandsCache({"PATH": []})
     loc = cache.locate_binary(str(faux_binary))
     assert "runme.exe" in loc
 
 
-def test_exes_in_cwd_are_not_matched(tmp_path, monkeypatch):
-    monkeypatch.chdir(tmp_path)
-    faux_binary = tmp_path / "runme.exe"
-    faux_binary.touch()
-    faux_binary.chmod(0o755)
+def test_exes_in_cwd_are_not_matched(faux_binary, monkeypatch):
+    monkeypatch.chdir(faux_binary.parent)
     cache = CommandsCache({"PATH": []})
     assert cache.locate_binary("runme.exe") is None
 
--- a/tests/test_commands_cache.py
+++ b/tests/test_commands_cache.py
@@ -157,7 +157,6 @@ def test_commands_cache_predictor_default(args, xession, tmp_path):
 
 
 class Test_is_only_functional_alias:
-    @skip_if_on_windows
     def test_cd(self, xession):
         xession.aliases["cd"] = lambda args: os.chdir(args[0])
         xession.env["PATH"] = []
@@ -204,7 +203,6 @@ def test_update_cache(xession, tmp_path):
     assert file2.samefile(cached[basename][0])
 
 
-@skip_if_on_windows
 def test_nixos_coreutils(tmp_path):
     """On NixOS the core tools are the symlinks to one universal ``coreutils`` binary file."""
     path = tmp_path / "core"
--- a/tests/test_commands_cache.py
+++ b/tests/test_commands_cache.py
@@ -205,7 +205,12 @@ def test_update_cache(xession, tmp_path):
 
 @pytest.fixture
 def faux_binary(tmp_path):
-    binary = tmp_path / "runme.exe"
+    """
+    A fake binary in the temp path.
+
+    Uses mixed case so tests may make assertions about it.
+    """
+    binary = tmp_path / "RunMe.exe"
     binary.touch()
     binary.chmod(0o755)
     return binary
@@ -214,13 +219,13 @@ def faux_binary(tmp_path):
 def test_find_binary_retains_case(faux_binary):
     cache = CommandsCache({"PATH": []})
     loc = cache.locate_binary(str(faux_binary))
-    assert "runme.exe" in loc
+    assert faux_binary.name in loc
 
 
 def test_exes_in_cwd_are_not_matched(faux_binary, monkeypatch):
     monkeypatch.chdir(faux_binary.parent)
     cache = CommandsCache({"PATH": []})
-    assert cache.locate_binary("runme.exe") is None
+    assert cache.locate_binary(faux_binary.name) is None
 
 
 def test_nixos_coreutils(tmp_path):
--- a/xonsh/commands_cache.py
+++ b/xonsh/commands_cache.py
@@ -99,8 +99,16 @@ class CommandsCache(cabc.Mapping):
         return len(self._cmds_cache) == 0
 
     def get_possible_names(self, name):
-        """Generates the possible `PATHEXT` extension variants of a given executable
-        name on Windows as a list, conserving the ordering in `PATHEXT`."""
+        """Expand name to all possible variants based on `PATHEXT`.
+
+        PATHEXT is a Windows convention containing extensions to be
+        considered when searching for an executable file.
+
+        Conserves order of any extensions found and gives precedence
+        to the bare name.
+
+        Incidentally, also uppercase the name on Windows (why?).
+        """
         if ON_WINDOWS:
             name = name.upper()
         pathext = [""] + self.env.get("PATHEXT", [])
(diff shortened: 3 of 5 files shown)


jaraco:
Confirmed that 11d2e46f fixes #5469:

```
jaraco@bulletproof ~\draft @# ./rustup-init --version
rustup-init 1.27.1 (54dd3d00f 2024-04-24)
```

pyproject.toml
```diff
@@ -15,6 +15,9 @@ authors = [{ name = "Anthony Scopatz" }, { email = "scopatz@gmail.com" }]
 maintainers = [{ name = "Anthony Scopatz" }, { email = "scopatz@gmail.com" }]
 license = { text = "BSD 2-Clause License" }
 requires-python = ">=3.9"
+dependencies = [
+    "case-insensitive-dictionary; platform_system=='Windows'",
```
jaraco, line 19: Obviously adding the first dependency is a concern. Is it acceptable or would you want to vendor something instead? I originally was planning to use [jaraco.collections.FoldedCaseKeyedDict](https://jaracocollections.readthedocs.io/en/latest/#jaraco.collections.FoldedCaseKeyedDict), which also works, but it brings along a slew of transitive dependencies. For that, I could at least attest to the maintainability (especially given the Tidelift enrollment).

xonsh/commands_cache.py
```diff
@@ -18,6 +18,11 @@
 from xonsh.platform import ON_POSIX, ON_WINDOWS, pathbasename
 from xonsh.tools import executables_in
 
+if ON_WINDOWS:
```
jaraco, line 21: This behavior could be unified too if xonsh wanted to take a position on case-sensitivity (it could be case-sensitive on all platforms or case-insensitive on all platforms).

jaraco:
> @jaraco nice work! Is it possible to add test for every change?

Some of the changes rely on the existing tests (and retain the same expectation). Two changes that could benefit from tests:

- [ ] Demonstrate that executables in the current working directory aren't executed.
  - The reason I didn't do this one is because it's the same expectation as on Unix and it wasn't worth capturing that. But I see some tests of similar behavior, so it should be possible to include it.
- [ ] Demonstrate that an executable's ARG0 is lowercase when invoked a such.
  - I was thinking this test would be (possibly prohibitively) difficult until I realized we can assert that `find_binary` returns the name in the same case as the file.

jaraco:
@anki-code I'm ready for another review.

BYK (changes requested):

tests/test_commands_cache.py
```diff
@@ -204,7 +203,15 @@ def test_update_cache(xession, tmp_path):
     assert file2.samefile(cached[basename][0])
 
 
-@skip_if_on_windows
+def test_find_binary_retains_case(tmp_path):
```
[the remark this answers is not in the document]
jaraco: > It's not clear to me how we ensure the casing is retained. Was it all upper case earlier?

Yes. The test would xfail on Windows prior to the fix (at b37148141cd5176d42164ea8e1ab11549924fadd). That was the failure mode of #5469.

> Can we make the test file name MiXeD case so it's clear what we're preserving,

Yes, perhaps. In df1defb7661a436c3ae47c3b061fcfe7f1010ed3, I extracted a fixture to be used across a couple of tests. Making that fixture use a name with mixed case would more explicitly show that case is being retained, but it would also cause that concern to leak across tests, which is a little awkward. e.g.

```diff
diff --git a/tests/test_commands_cache.py b/tests/test_commands_cache.py
index cc72b33f..edbce701 100644
--- a/tests/test_commands_cache.py
+++ b/tests/test_commands_cache.py
@@ -205,7 +205,7 @@ def test_update_cache(xession, tmp_path):
 
 @pytest.fixture
 def faux_binary(tmp_path):
-    binary = tmp_path / "runme.exe"
+    binary = tmp_path / "RunMe.exe"
     binary.touch()
     binary.chmod(0o755)
     return binary
@@ -214,13 +214,13 @@ def faux_binary(tmp_path):
 def test_find_binary_retains_case(faux_binary):
     cache = CommandsCache({"PATH": []})
     loc = cache.locate_binary(str(faux_binary))
-    assert "runme.exe" in loc
+    assert "RunMe.exe" in loc
 
 
 def test_exes_in_cwd_are_not_matched(faux_binary, monkeypatch):
     monkeypatch.chdir(faux_binary.parent)
     cache = CommandsCache({"PATH": []})
-    assert cache.locate_binary("runme.exe") is None
+    assert cache.locate_binary("RunMe.exe") is None
 
 
 def test_nixos_coreutils(tmp_path):
```

Note that we're not testing anything about alternate casing being passed (e.g. `locate_binary('rUNmE.exe')`) - just that if a matching case of the name is passed that it's retained. We could strengthen this guarantee, but I'm not sure it's necessary, especially while Windows and Unixes have divergent behavior.

I think you bring up a good point, though, that if there were a regression to convert everything to lowercase but the filename had uppercase characters, the test as written wouldn't catch that. So having mixed case in the filename makes sense.

Done in 602d00eb.

xonsh/commands_cache.py
jaraco: I don't think it's obvious that the change should apply to macOS. While it's true that on Windows and Mac, the default file systems are case-insensitive, all operating systems offer file systems that can be either case sensitive or not.

Probably what we really want is to have the _lookup_ be case insensitive, but then additionally verify that the file exists by the name given. The real challenge arises when a file exists under two or more names varying only by case. In that scenario, the caching wouldn't work unless it mapped to all possible values (i.e. a CaseInsensitive MultiDict).

I was thinking that xonsh could be case sensitive or case insensitive on all operating systems, but if it's case insensitive on a case-sensitive filesystem, it will not be able to reach filenames that vary only by case, and if it's case sensitive on case-insensitive filesystems, users might be surprised when a file is only reachable by the exact case.

The latter option (being case-sensitive on all platforms) seems the most safe and tolerable, especially if all we're talking about is command matching.

Regardless, I don't want to try to solve that issue here. In this PR, I'm focused on the more narrow issue of name mangling that was happening in the command cache on Windows. If someone is interested in willing to explore the issue in more depth, I welcome them to file an issue.

jaraco:
Yes, it's ready for review. I addressed BYK's comments, although there's something weird about the GitHub UI that it's not allowing me to respond inline to the comments, I guess because they were responses to my original comments. If you scroll up, you can see my responses.

PR #5478, opened 2024-06-02
list_lines: list for single line output

anki-code:
### Motivation

This is fix for introduced before `$XONSH_SUBPROC_OUTPUT_FORMAT='list_lines'` (#5377) so news file is not needed.

We need to return the list for single line output to avoid unintended reading the string.

### Before
```xsh
$XONSH_SUBPROC_OUTPUT_FORMAT='list_lines'
mkdir -p /tmp/list_lines
cd /tmp/list_lines
touch 123

du $(ls)
# 0	123

for f in $(ls):
    print(f)
# 1
# 2
# 3

touch 321

for f in $(ls):
    print(f)
# 123
# 321
```

### After

```xsh
$XONSH_SUBPROC_OUTPUT_FORMAT='list_lines'
mkdir -p /tmp/list_lines
cd /tmp/list_lines
touch 123

du $(ls)
# 0	123  # no changes

for f in $(ls):
    print(f)
# 123  # now it works right

touch 321

for f in $(ls):
    print(f)
# 123   # no changes
# 321
```

## For community

### DIFF
--- a/tests/procs/test_specs.py
+++ b/tests/procs/test_specs.py
@@ -186,8 +186,8 @@ def test_specs_with_suspended_captured_process_pipeline(
 @pytest.mark.parametrize(
     "cmds, exp_stream_lines, exp_list_lines",
     [
-        ([["echo", "-n", "1"]], "1", "1"),
-        ([["echo", "-n", "1\n"]], "1", "1"),
+        ([["echo", "-n", "1"]], "1", ["1"]),
+        ([["echo", "-n", "1\n"]], "1", ["1"]),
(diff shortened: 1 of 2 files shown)


jaraco (approved):
This change looks good to me, especially because it makes the return symmetric.

jaraco:
That's weird. A test relating to pipelines that was previously passing in #5477 is now [failing](https://github.com/xonsh/xonsh/actions/runs/9340636970/job/25706426330?pr=5477) when the [only diff](https://github.com/xonsh/xonsh/pull/5477/commits/2cfd8ed5c5ea4e1df128ab9c129a36bd90f6b60d) is obviously unrelated. I suspect this change may be interacting with that PR or this change may have an intermittent failure.

PR #5099, opened 2023-03-19
Always load rc files except --no-rc is passed

cosineFish:
This change fixes issue #4855.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #5491, opened 2024-06-11
Do not load `~/.xonshrc` in not interactive mode

### LINKED ISSUE #5488: Unintended loading `xonshrc` file in script mode
My xonshrc is pretty silent and I suddenly noticed that xonshrc is loaded in script mode:

```xsh
cd /tmp
echo 'echo RC' >> ~/.xonshrc
echo '#!/usr/bin/env xonsh' > myscript
chmod +x myscript

./myscript
# RC

xonsh myscript.xsh
# RC
```

When I run any xonsh script I have no any expectation that my huge `.xonshrc` with features for prompt will be loaded.

Bash/zsh also don't load RC files when you run script:

```xsh
cd /tmp
echo 'echo RC' >> ~/.bashrc
echo 'echo 1' > script.sh
bash script.sh
# 1  # No RC

echo 'echo RC' >> ~/.zshrc
echo 'echo 1' > script.zsh
zsh script.zsh
# 1  # No RC
```

## For community

anki-code:
### Motivation

In #5099 was introduced the logic where all RC files are loaded in interactive and non-interactive modes. This logic is not working good for home based `~/.xonshrc` file.

First of all `~/.xonshrc` is the buffer of settings focused on interactive mode. Most tools with integration with xonsh (e.g. `conda`, `zoxide`, etc) offer to just do `echo "init_tool()" >> ~/.xonshrc` (`conda` has around 20 lines of init code) to start using the tool and users are doing this without any doubts.

But because of after 5099 `~/.xonshrc` is executed in non-interactive mode the adding code to it leads to unexpected and unintended side effects: 

* If you run a script with shebang (e.g. `#!/usr/bin/env xonsh` or [xonsh-awesome-cli-app](https://github.com/anki-code/xonsh-awesome-cli-app)) or just from `xonsh script.xsh` the code will be unexpected and unintended slower.

* If you're using xonsh-based tools (e.g. you install them using pip) and run them in environment that has no packages that initiated in `~/.xonshrc` you will see epic errors.

* Additional context: 
  * Bash and Zsh do not load `~/.bashrc` and `~/.zshrc` in non-interactive mode by the same reasons.
  * We have welcome message `Create ~/.xonshrc file manually or use xonfig to suppress the welcome message` and we don't want to make the process of creating this file complex.

All of this leads to bad unexpected and unintended experience. This PR is to fix this.

### Expectation

By doing this fix we assume that experienced user who wants to build good repeatable run control files will use another ways to create config files and this has [reflection in docs](https://github.com/xonsh/xonsh/blob/8860f2bd5273d5f3fc08ccf6be6af8163bfec0bd/docs/xonshrc.rst). In the nutshell if you want to create the RC files that affect every run of code you should use one or many of these ways:

* Cross-desktop group (XDG) compliant `~/.config/xonsh/rc.xsh` control file.
* The system-wide control file `/etc/xonsh/xonshrc` for Linux and OSX and in `%ALLUSERSPROFILE%\xonsh\xonshrc` on Windows. It controls options that are applied to all users of Xonsh on a given system.
* The home-based directory `~/.config/xonsh/rc.d/` and system `/etc/xonsh/rc.d/` can contain .xsh files. They will be executed at startup in order. This allows for drop-in configuration where your configuration can be split across scripts and common and local configurations more easily separated.

In your configs you need to check `$XONSH_INTERACTIVE` and `$XONSH_LOGIN` explicitly.

### Before

`~/.xonshrc` is used in non-interactive mode.

```xsh
echo "echo RC" >> ~/.xonshrc
cd /tmp
echo "echo Script" > script.xsh
xonsh script.xsh
# RC
# Script
```
```xsh
cd /tmp
echo 'echo RC' >> ~/.xonshrc
echo '#!/usr/bin/env xonsh' > myscript
chmod +x myscript

./myscript
# RC
```

### After

`~/.xonshrc` is not used in non-interactive mode. Use `-i` if you need it.

```xsh
echo "echo RC" >> ~/.xonshrc
cd /tmp
echo "echo Script" > script.xsh
xonsh script.xsh
# Script

xonsh -i script.xsh
# RC
# Script
```
```xsh
cd /tmp
echo 'echo RC' >> ~/.xonshrc
echo '#!/usr/bin/env xonsh' > myscript
chmod +x myscript

./myscript
```
Closes #5488 #4096 #5496

### Fun

I want to leave here some nice representation of how it works in bash/sh/zsh from [twitter post](https://twitter.com/paxx39/status/1742768007154479109):

image

[How it works in xonsh](https://github.com/xonsh/xonsh/blob/20081b5a5179afa82b587e0bd3795b72caa2a2a5/docs/xonshrc.rst):
* Non-interactive ->  `$XONSHRC` and `$XONSHRC_DIR`
* Interactive -> The same + `~/.xonshrc`

And all logic inside files ;)

## For community

### DIFF
--- a/news/xonshrc.rst
+++ b/news/xonshrc.rst
@@ -16,7 +16,7 @@
 
 **Fixed:**
 
-* <news item>
+* Windows: fixed path to RC file in ``xonfig web``.
 
 **Security:**
 


xonsh/main.py
```diff
@@ -308,6 +308,16 @@ def _get_rc_files(shell_kwargs: dict, args, env):
...
     rc = env.get("XONSHRC")
     rcd = env.get("XONSHRC_DIR")
+
+    if not env.get("XONSH_INTERACTIVE", False):
+        """
+        Home ``~/.xonshrc`` file has special meaning and history. The ecosystem around shells treat this kind of files
+        as the place where interactive tools and configs may be added. To avoid unintended and unexpected affection
+        of this file on non interactive behavior we remote this file in non interactive mode e.g. script with shebang.
+        """
+        home_xonshrc = os.path.expanduser("~/.xonshrc")
```
gforsyth, line 318: Does this work on Windows?  Maybe instead you can do
```suggestion
        home_xonshrc = str((Path("~") / ".xonshrc").expanduser())
```

gforsyth (commented):
Looks good overall, just one question about cross-platform RC file location

gforsyth:
Hey @bestlem -- I think you're on board with these changes, yes?

gforsyth (approved):
Thanks for this @anki-code !

I saw there may be another bug with when we drop into interactive mode, but I think we can handle that in a followup

PR #5506, opened 2024-06-15
xonshrc: ability to set suffix with mode

anki-code:
### Motivation

After https://github.com/xonsh/xonsh/pull/5099 https://github.com/xonsh/xonsh/pull/5491 I want to give a way to avoid parsing and executing unwanted RC files completely.

### Before

If you have `~/.config/xonsh/rc.d/myrc.xsh` it will be parsed and executed in interactive and non-interactive mode and we expect that user will have `$XONSH_INTERACTIVE` checking inside.

### After

The same as before + you can rename the file to `myrc_int.xsh` and it will be loaded only in interactive mode without any parsing and execution in non-interactive mode. This allows to have this kind of splitting in rc-directory *if you want*:
```xsh
# ~/.config/xonsh/rc.d/
myrc.xsh        # e.g. `$PATH.prepend("/my/lovely/path/bin")`
myrc_int.xsh    # e.g. `xontrib load my-lovely-interactive-completions`
myrc_noint.xsh  # e.g. `aliases |= {"my-lovely-sugar": "for-scripts"}`
```
This allows to load xonsh as faster as possible.

## For community

### DIFF
--- a/news/xonshrc.rst
+++ b/news/xonshrc.rst
@@ -5,7 +5,7 @@
 **Changed:**
 
 * xonshrc: the home based ``~/.xonshrc`` will not be executed in non-interactive mode (#5491).
-* xonshrc: you can add prefix to xonshrc files to set preferred mode e.g. ``myrc_noint.xsh`` or ``myrc_int.xsh`` to load in non-interactive or interactive mode respectively.
+* xonshrc: you can add prefix to xonshrc files to set preferred mode e.g. ``myrc_noint.xsh`` or ``myrc_int.xsh`` to load in non-interactive or interactive mode respectively (#5506).
 
 **Deprecated:**
 


gforsyth:
I am not sure that I like encoding this kind of granular execution context in a filename.

Isn't the entire premise of having multiple configuration file directories that we don't have to parse the file in `$HOME` if we aren't running in interactive mode?

gforsyth:
Personally, I don't like this kind of magic-filename scope creep unless there's a specific user request for it.

That said, you are a user -- is this something you want in xonsh?

jnoortheen:
@anki-code If I remember correctly, xonsh used to load the RC files only during interactive mode and not load during script runs. I think that behaviour is the one that matches other shells. If you are running a script from somewhere that environment is expected to have all variables setup. These extra behaviour makes shell convoluted.

jnoortheen:
Yes that is what I mean, it is recently introduced behaviour. 
Loading specific files can be done like what fish does 

image

jnoortheen:
Simply use plain Python. Also there is new PEG parser you can start using probably soon.

Issue #5517, opened 2024-06-19
Some tests in tests/test_integrations.py failed after the timeout expired on 0.17.0


rautyrauty:
## Current Behavior
[build2.log](https://github.com/user-attachments/files/15902917/build2.log)
FAILED tests/test_integrations.py::test_catching_system_exit - subprocess.TimeoutExpired
FAILED tests/test_integrations.py::test_catching_exit_signal - subprocess.TimeoutExpired
FAILED tests/test_integrations.py::test_suspended_captured_process_pipeline - subprocess.TimeoutExpired
FAILED tests/test_integrations.py::test_alias_stability - subprocess.TimeoutExpired
FAILED tests/test_integrations.py::test_alias_stability_exception - subprocess.TimeoutExpired
FAILED tests/test_integrations.py::test_xonshrc[-i-.*CONFIG_XONSH_RC_XSH.*HOME_XONSHRC.*CONFIG_XONSH_RCD.*] - subprocess.TimeoutExpired
FAILED tests/test_integrations.py::test_xonshrc[-i --rc rc.xsh-.*RCXSH.*] - subprocess.TimeoutExpired
= 7 failed, 5060 passed, 35 skipped, 17 xfailed, 1 xpassed, 3 warnings, 9 rerun in 192.45s (0:03:12) =

I Succeeded in fixing the test_catching_system_exit and test_catching_exit_signal by adding a
```python
@@ -105,6 +106,7 @@ def run_xonsh(
     if stdin_cmd:
         proc.stdin.write(stdin_cmd)
         proc.stdin.flush()
+        time.sleep(0.5)
 
     try:
         out, err = proc.communicate(input=input, timeout=timeout)
```

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

rautyrauty:
<details>
<summary>xonfig</summary>
<pre>
+-----------------------------+-----------------+
| xonsh                       | 0.17.0          |
| Python                      | 3.12.2          |
| PLY                         | 3.11            |
| have readline               | True            |
| prompt toolkit              | 3.0.43          |
| shell type                  | prompt_toolkit  |
| history backend             | json            |
| pygments                    | 2.18.0          |
| on posix                    | True            |
| on linux                    | True            |
| distro                      | altlinux        |
| on wsl                      | False           |
| on darwin                   | False           |
| on windows                  | False           |
| on cygwin                   | False           |
| on msys2                    | False           |
[...]

PR #5536, opened 2024-06-24
prompt: switching to prompt_toolkit in case of stdin with interactive mode

### LINKED ISSUE #5462: Support reading stdin in `readline` interactive move (`TERM=dumb xonsh -i` hangs)
## Current Behavior

```xsh
echo whoami | TERM=dumb xonsh -i --no-rc
# or the same
echo whoami | xonsh -st dumb -i  # dumb is readline in fact
# or the same
echo whoami | xonsh -st readline -i
```

right now, this opens an interactive terminal which hangs indefinitely and does not respond to input:
image

xonsh has to be manually killed by another process.

## Expected Behavior

xonsh behaves the same as if TERM were unset, or set to a "normal" value (e.g. `linux` or `xterm-256color`).

## Workaround
Set any other `TERM=linux` or `TERM=xterm` or etc.

## xonfig

<details>

```xsh
+-----------------------------+----------------------+
| xonsh                       | 0.16.0               |
| Git SHA                     | b5ebcd3f             |
| Commit Date                 | May 30 09:51:25 2024 |
| Python                      | 3.10.12              |
| PLY                         | 3.11                 |
| have readline               | True                 |
| prompt toolkit              | 3.0.45               |
| shell type                  | none                 |
| history backend             | json                 |
| pygments                    | 2.18.0               |
| on posix                    | True                 |
| on linux                    | True                 |
| distro                      | pop                  |
| on wsl                      | False                |
| on darwin                   | False                |
| on windows                  | False                |
| on cygwin                   | False                |
| on msys2                    | False                |
| is superuser                | False                |
| default encoding            | utf-8                |
| xonsh encoding              | utf-8                |
| encoding errors             | surrogateescape      |
| xontrib                     | []                   |
| RC file                     | []                   |
| UPDATE_OS_ENVIRON           | False                |
| XONSH_CAPTURE_ALWAYS        | False                |
| XONSH_SUBPROC_OUTPUT_FORMAT | stream_lines         |
| THREAD_SUBPROCS             | True                 |
| XONSH_CACHE_SCRIPTS         | True                 |
+-----------------------------+----------------------+
```

</details>

## more info

note that setting `--no-env` makes the problem go away, mostly:
```
$ echo whoami | TERM=dumb xonsh -i --no-rc --no-env
Warning: Input is not a terminal (fd=0).
jyn@pop-os ~/src/xonsh title @ whoami
jyn
/usr/bin/stty: 'standard input': Inappropriate ioctl for device
jyn@pop-os ~/src/xonsh title @ 
```

i think this is a red herring though; it just means that `TERM=dumb` gets ignored by xonsh itself. `echo env | env -i TERM=dumb $(which xonsh) -i --no-rc ` still hangs indefinitely.

originally i thought this was related to [`DumbShell`](https://github.com/xonsh/xonsh/blob/38295a1dd941451362d2e3d14dd29d94ba4f54ce/xonsh/dumb_shell.py#L7) behaving differently than ReadLine shell, but i tried this diff and it didn't fix the issue:
```diff
diff --git a/xonsh/shell.py b/xonsh/shell.py
index f54c5350..4e60b6ae 100644
--- a/xonsh/shell.py
+++ b/xonsh/shell.py
@@ -208,6 +208,4 @@ class Shell:
                 from xonsh.ptk_shell.shell import PromptToolkitShell as cls
-            elif backend == "readline":
+            elif backend == "readline" or backend == "dumb":
                 from xonsh.readline_shell import ReadlineShell as cls
-            elif backend == "dumb":
-                from xonsh.dumb_shell import DumbShell as cls
             else:
```

## For community

anki-code:
### Motivation

There is an edge case that we're using mostly in integration tests from the beginning of the world:
`echo 'echo 1' | xonsh -i` and it's not working with `TERM=dumb` because dumb is readline.
So in this case we need to force using prompt_toolkit to avoid hanging.

Closes #5462 #5517

### Before

These all produces hanging because reading stdin is not implemented in readline backend:
```xsh
echo whoami | TERM=dumb xonsh -i --no-rc   # dumb is readline in fact
# or the same
echo whoami | xonsh -st dumb -i  # dumb is readline in fact
# or the same
echo whoami | xonsh -st readline -i
```

### After

Switching to prompt_toolkit in these cases.

## For community

### DIFF
--- a/news/prompt_switch_to_ptk.rst
+++ b/news/prompt_switch_to_ptk.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* prompt: Switching to prompt_toolkit in edge case of sending stdin to interactive mode (#5462 #5517).
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/xonsh/lib/__init__.py
+++ b/xonsh/lib/__init__.py
@@ -1,0 +1,9 @@
+"""
+Originally posted by @scopatz in https://github.com/xonsh/xonsh/issues/2726#issuecomment-406447196 :
+
+    Recently @CJ-Wright has started up a ``xonsh.lib`` sub-package, which is usable from pure Python.
+    This is meant as a standard library for xonsh and downstream tools.
+    Currently there are some xonsh-ish wrappers around ``os`` and ``subprocess``.
+    We'd love to see more contributions to this effort! So if there is something
+    like ``sh()`` that you'd like to see, by all means please help us add it!
+"""
--- a/xonsh/shell.py
+++ b/xonsh/shell.py
@@ -207,7 +207,9 @@ class Shell:
             `echo 'echo 1' | xonsh -i` and it's not working with `TERM=dumb` (#5462 #5517).
             So in this case we need to force using prompt_toolkit.
             """
-            is_stdin_to_interactive = not sys.stdin.isatty() and XSH.env.get("XONSH_INTERACTIVE", False)
+            is_stdin_to_interactive = not sys.stdin.isatty() and XSH.env.get(
+                "XONSH_INTERACTIVE", False
+            )
 
             if backend == "none":
                 from xonsh.base_shell import BaseShell as cls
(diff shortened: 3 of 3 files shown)


jnoortheen (approved):

jnoortheen:
Thanks @anki-code !

Issue #5309, opened 2024-03-23
xonsh in WSL, slow startup due to PATH pollution


kesnar:
First of all, I have to say that I love the work you have done for xonsh! It is an awesome tool and hope that it gets the recognition it deserves in the future!

So, I am using xonsh in WSL in Windows 11. I have tried both in Debian WSL and Ubuntu WSL, with both the xonsh installed through apt and through pipx.
The problem I face is that there is a big startup time for xonsh of around 2-3 seconds. I searched a bit around and found out that the reason is that the PATH variable in WSL is too long (both in length and in terms of programs) because Windows append the PATH with the Windows PATH.

I was able to stop that from happening, and have normal startup times, by following this microsoft doc: https://learn.microsoft.com/en-us/windows/wsl/wsl-config#configure-per-distro-launch-settings-with-wslconf

Bellow in the Showcase section you can see the difference and steps to reproduce it.

So my question is basically if there is another way to handle this, instead of just removing the Windows PATH?
Also maybe this should also be documented somewhere in the docs?

## xonfig

<details>

```
kesnar@glotka ~ $ xonsh --timings
 Debug level: Off
|-------------------------|-----------|-----------|
|Event name               | Time (s)  | Delta (s) |
|-------------------------|-----------|-----------|
|start                    |   0.000   |   0.000   |
|pre_execer_init          |   0.004   |   0.004   |
|post_execer_init         |   0.028   |   0.024   |
|pre_xontribs_autoload    |   0.028   |   0.000   |
|post_xontribs_autoload   |   0.036   |   0.008   |
|on_pre_rc                |   0.036   |   0.000   |
|on_post_rc               |   0.087   |   0.051   |
|on_ptk_create            |   0.096   |   0.010   |
|on_post_init             |   0.097   |   0.000   |
|on_pre_cmdloop           |   0.097   |   0.000   |
|on_pre_prompt_format     |   0.097   |   0.000   |
|on_pre_prompt_tokenize   |   0.120   |   0.024   |
|on_post_prompt_tokenize  |   0.126   |   0.005   |
|on_pre_prompt_style      |   0.144   |   0.019   |
|on_post_prompt_style     |   0.146   |   0.001   |
|on_pre_prompt            |   0.146   |   0.000   |
|-------------------------|-----------|-----------|
kesnar@glotka ~ $ xonfig
+------------------+-----------------------+
| xonsh            | 0.13.4                |
| Python           | 3.11.2                |
[...]

PR #5539, opened 2024-06-25
Command cache: fix update cache logic

anki-code:
### Motivation

I noticed that when I type every letter in the prompt on the remote server the typing is lagging. I thought it's ssh but no.
After inspection the commands cache code I found that on every key press there is reading all 2000+ files in PATH because the code for cache update has funny issue.

### Quiz

```xsh
def f():
    print(1)
    yield False
    print(2)
    yield True
    print(3)

print(any(f()))
```
What will be in the output? Answer:

<details>

```
1
2
True
```

The execution of `print(3)` ignored because `any()` interrupts the execution of the function. 
If we call `print(list(f()))` the output will be `1 2 3 [False, True]`.

</details>

### Before

* Updating cache (read all files in PATH and list of aliases) two times after start.
* Updating cache (read all files in PATH and list of aliases) on every key press in prompt.

### After

* Update cache once at start.
* Update cache only when we have real changes of paths or aliases.

cc #4954 #3895 #5309

## For community

### DIFF
--- a/news/commands_cache_fix_update.rst
+++ b/news/commands_cache_fix_update.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* Commands Cache: Fixed cache update logic that lead to lagging during typing.
+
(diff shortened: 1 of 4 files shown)


jnoortheen (approved):

jnoortheen:
LGTM! Thanks @anki-code

PR #5543, opened 2024-06-26
Lazy xaliases

jnoortheen:
to reduce the number of imports at the start, we can use this method of lazy evaluation. It can be part of modularizing the codebase. 

## For community


jnoortheen:
@anki-code  I prefer `xonsh/aliases` package instead of a nested one.

jnoortheen:
will split the test improvements.

Issue #5541, opened 2024-06-26
Refactoring: commands cache


anki-code:
We need to refactor commands_cache:

* Do not create one `_cmds_cache` list based on aliases and list of files from PATHs because any change will force rebuilding the cache. Implement layers approach instead of this. Aliases - first layer, files - second layer. Now you can update first or second layer without rebuilding everything. The get function just overlap layers on the fly to extract single row.

* Use list of files from Windows/System32 as a permanent layer and do not re-read it. The System32 diretory is huge and persistent.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

eugenesvk:
By the way, is there a place in the current xonsh code where a list of strings is split into a character-by-character tree so that you can see for each character whether there are branches that start with this (and previous) chars so that if there aren't you can stop early instead of continuing to search with each new char? You've mentioned something in the other issue, but from my shallow overview command cache just uses a dictionary, so would have to check the full string every time?

eugenesvk:
timings per search might be the same, but the number of searches will differ a lot since with a dict you wouldn't know that you don't need to search for the `111111` key because there is not a single key starting with `1`, so you stopped searching 5 keystrokes ago

eugenesvk:
FYI You can use `CharTrie` from this pure Python trie library https://pypi.org/project/pygtrie/ library (archived by Google, but still maintained by someone else?) to test for every typed command whether it's a prefix of any existing file from PATH and if not stop checking on further keystrokes
It even has an example of typing-by-char and checking 

You don't have Cython extensions in xonsh, right? Otherwise could check some faster native trie libs, but don't think that would make any noticeable difference

Issue #5014, opened 2022-12-22
WSL2 Ubuntu Xonsh setup very slow


KoStard:
## xonfig

<details>

```
$ xonfig
+------------------+------------------------+
| xonsh            | 0.13.4                 |
| Python           | 3.10.6                 |
| PLY              | 3.11                   |
| have readline    | True                   |
| prompt toolkit   | 3.0.36                 |
| shell type       | prompt_toolkit         |
| history backend  | json                   |
| pygments         | 2.13.0                 |
| on posix         | True                   |
| on linux         | True                   |
| distro           | ubuntu                 |
| on wsl           | True                   |
| wsl version      | 2                      |
| on darwin        | False                  |
| on windows       | False                  |
| on cygwin        | False                  |
| on msys2         | False                  |
| is superuser     | False                  |
| default encoding | utf-8                  |
| xonsh encoding   | utf-8                  |
| encoding errors  | surrogateescape        |
| xontrib          | []                     |
| RC file 1        | /home/kostard/.xonshrc |
+------------------+------------------------+
```

</details>

## Expected Behavior
The startup, the updates when typing, when executing a command and other interactions should be fast (comparable to bash and other shells).

## Current Behavior
Everything was very slow in the beginning. Found this issue: https://github.com/xonsh/xonsh/issues/3895
I followed the steps there, and can confirm that my $PATH is very long and windows paths are included there, which I don't necessarily need so I tried the mentioned steps for removing it (although same is true when I use bash in WSL and it's fast, so can't fully agree that expecting the users to remove it is perfect).
Added the following:
```
$ cat /etc/wsl.conf
[Interop]
appendWindowsPath = False
```

It solved the problem when opening a terminal window in WSL. But for some reason, this didn't affect VSCode that is connected to the WSL. It was still very slow.

Added this line to the .xonshrc
```
[$PATH.remove(path) for path in $PATH.paths if path.startswith("/mnt/c/")]
```
This way at least the commands run quickly, but the startup is very slow.

## Steps to Reproduce
Open VSCode, connect to a WSL2 with Ubuntu, that is using xonsh as default shell. Try to open a terminal window and run anything.

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

danielcranford:
[...]

PR #5544, opened 2024-06-27
Introduce new resolver for executables to replace commands_cache usages and speed up everything

anki-code:
### Motivation

We have `commands_cache` (CC) - the list of executable files and aliases. To keep this list actual there is `cc.update_cache` and `_update_and_check_changes` functions that read the list of all files in `$PATH`. Every time the code uses CC it reads the list and attributes of files from `$PATH` (2000-3000 files).

It looks inefficient because in fact what we really need in most cases is to just `resolve_executable` by checking the resolved command name in `$PATH`. Instead of doing this simple action we're using CC with `cc.update_cache`. I think it's legacy from the previous contributors who just wanted to reuse what they see in the code without asking additional questions. It's time to replace CC usages in some key places.

### Before

* main: During xonsh start we have a couple of places that produce useless `cc.update_cache`.
* pygments: the hook starts every time you press a key and do `cc.update_cache`. 
  * It adds lagging to prompt and kills typing in case the disk is network device or another slow drive.
* specs: calls `cc.update_cache` two times per call subprocess command.
* Startup time on my mac: `xonsh --no-rc --timings` -> `on_pre_prompt` ~ 0.3 s - 10 runs.

### After

New `xonsh.specs.executables` module introduces a fast replacement for `xonsh.environ.locate_binary` (it
s `commands_cache.locate_binary` in fact). Instead of using commands_cache and reading all files in the `$PATH` many times we're checking only the name of the executable in `$PATH` using `xonsh.specs.executables`.

* main: CC will not run until the real need.
* pygments: do not use CC in pygments hook.
* specs: do not use CC in final stage when we just need to resolve executive path.
* source: `locate_binary` (it looks that it is not working in fact) replaced to new `locate_file` (and it's working now).
* Startup time on my mac: `xonsh --no-rc --timings` -> `on_pre_prompt` ~ 0.16 s - 10 runs.
* (We continue using CC for completion or commands suggestion.)

Closes:
* #5014 
* #3895
* #5309 
* #5541

### Testing

```xsh
xpip install -U --force-reinstall git+https://github.com/xonsh/xonsh@executables
# restart xonsh
# type and run commands
```

Because I leaved `xonsh.environ.locate_binary` and `commands_cache` unchanged I believe that most non-key functionality will work as before.

The changes in `specs` and `pygments` also have tests and I believe it will work.

The pain point to me is Windows. We need to test how it works on Windows. Help here please! cc @jaraco @Qyriad 

### Future impact on the code and maybe next steps

* Please do not use `commands_cache`, `xonsh.environ.locate_binary` for single operations around executables. If we have features around locating executables we need to enrich and use `xonsh.procs.executables`. 
* Someday we need to refactor `which`.
* Someday we need to replace `xonsh.environ.locate_binary` in other places to `xonsh.procs.executables`.

## For community

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,7 @@ xonsh
             cat /etc/passwd | grep root > ~/root.txt
 
             $PROMPT = '@ '
-            
+
 
       -  .. code-block:: python
 
@@ -55,7 +55,7 @@ xonsh
 
             aliases['e'] = 'echo @(2+2)'
             aliases['a'] = lambda args: print(args)
-            
+
 
 If you like xonsh, :star: the repo, `write a tweet`_ and stay tuned by watching releases.
 


jaraco:
By the way, `xpip` won't work on Windows ([error](https://gist.github.com/jaraco/1d572df5c32d59e0242b97c8b6c46400)) because you can't overwrite an executable that's in use.

jaraco:
I installed the branch on my Windows ARM host and it seems to be running great. I can't test on WSL2 (due to virtualization limitations) and performance has been adequate for me.

I can see that now the commands cache is much smaller:

```
 ~ @# len(__xonsh__.commands_cache.all_commands)
863
```

Compare that with ~5000 as reported earlier.

xonsh/procs/executables.py
```diff
@@ -0,0 +1,112 @@
...
+    Conserves order of any extensions found and gives precedence
+    to the bare name.
+    """
+    env = env if env is not None else XSH.env
+    env_pathext = env.get("PATHEXT", [])
+    if not env_pathext:
+        return [name]
+    upper = name.upper() == name
+    extensions = [""] + env_pathext
+    return [name + (ext.upper() if upper else ext.lower()) for ext in extensions]
```
jaraco, line 25: I don't care for the case manipulation here. It feels like extra complexity, though I guess I see why you might do it (to retain some sort of consistency). I'm a little sad to see how much more complicated it is than the previous implementation (which was two lines). What about this instead:

```suggestion
    env = env if env is not None else XSH.env
    extensions = [""] + env.get("PATHEXT", [])
    return [name + match_case(ext, name) for ext in extensions]
```

where

```python
def match_case(ext, name):
    return ext.upper() if name.upper() == name else ext.lower()
```

Or if you really need the efficiency of only checking `name.upper() == name` once:

```suggestion
    env = env if env is not None else XSH.env
    extensions = [""] + env.get("PATHEXT", [])
    matcher = str.upper if name.upper() == name else str.lower
    return [name + matcher(ext) for ext in extensions]
```

xonsh/procs/executables.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def clear_paths(paths):
+    """Remove duplicates and nonexistent directories from paths."""
+    cont = set()
+    for p in map(os.path.realpath, paths):
+        if p not in cont:
+            cont.add(p)
+            if os.path.isdir(p):
+                yield p
```
jaraco, line 35: This is the `unique_everseen` pattern:

```suggestion
    return filter(os.path.isdir, unique_everseen(map(os.path.realpath, paths)))
```

Unique_everseen is available in [more_itertools](https://pypi.org/project/more-itertools) or copy/pasted from stdlib docs into xonsh.lib.itertools.

xonsh/procs/executables.py
```diff
@@ -0,0 +1,112 @@
...
+        # exist even if isdir returns True. This may happen for instance
+        # if the path contains trailing spaces.
+        return False
+
+
+def is_executable_in_posix(filepath):
+    """Check the file is executable in POSIX."""
+    try:
+        if filepath.is_file() and os.access(filepath, os.X_OK):
+            return True
```
jaraco, line 69: You can avoid a literal and boolean check with:

```suggestion
        return filepath.is_file() and os.access(filepath, os.X_OK)
```

xonsh/procs/executables.py
```diff
@@ -0,0 +1,112 @@
...
+    try:
+        try:
+            if not filepath.is_file():
+                return False
+        except OSError:
+            return False
+
+        env = env if env is not None else XSH.env
+        env_pathext = env.get("PATHEXT", [])
+        if filepath.suffix.upper() in env_pathext:
```
jaraco, line 56: I don't think you can assume `upper` here. Windows users can (and do) put lowercase extensions in PATHEXT. Use a case-insensitive comparison instead.

jaraco: In fact, maybe XSH.env should expose elements of PATHEXT (and probably PATH and others) as case-insensitive strings.

xonsh/procs/executables.py
```diff
@@ -0,0 +1,112 @@
...
+    env = env if env is not None else XSH.env
+    env_path = env.get("PATH", [])
+    paths = tuple(reversed(tuple(clear_paths(env_path))))
+    possible_names = get_possible_names(name, env) if use_pathext else [name]
+
+    if check_executable:
+        if ON_WINDOWS:
+            is_executable = is_executable_in_windows
+        else:
+            is_executable = is_executable_in_posix
```
jaraco, line 101: Can this be done globally to avoid making the decision here?

xonsh/procs/executables.py
```diff
@@ -0,0 +1,112 @@
...
+
+    for path in paths:
+        for possible_name in possible_names:
```
jaraco, line 104:
```suggestion
    for path, possible_name in itertools.product(paths, possible_names):
```

jaraco (approved):
I've made a few (mostly nitpicky) suggestions.

Feel free to take them or leave them.

Overall, this seems like a solid improvement. Thanks!

jaraco:
> There is no changes in commands_cache logic. Could you please confirm that you see decreased amount of files and this is after installation this PR. If it's right it's unexpected and I want to understand the cause.

It's not in this PR. I just reinstalled `main` and the number is the same.

Qyriad:
Startup time is definitely improved:
`on_pre_prompt` before: 0.456
`on_pre_prompt` after: 0.133

Overall laggyness at the prompt seems about unchanged:

https://github.com/xonsh/xonsh/assets/1542224/3fdf8d31-18c2-4c60-8eaf-d3434a37032e

jnoortheen:
The logic of CC is that, it should only be checking the last modified time of a path before loading binaries from that path. After the initial time, the real work is checking the modified time of all the PATHS. It doesn't need to be checked for every keystroke. It can only be checked after a word (space) is entered.

jnoortheen:
@anki-code sorry, I haven't tried this yet. I will try to get back this week.

jnoortheen (approved):

eugenesvk:
> 2\. update_cache [scans ALL files](https://github.com/xonsh/xonsh/blob/6860388c887abb647e6082517faf3f8a5a8754b3/xonsh/tools.py#L829-L835) in all directories during checking for the update. Checking for the update don't rid us for scanning all possible files.

where are you seeing this? The update cache invokes that `executables_in` function that scans all files only does that if the mtime changed or the path isn't cached, otherwise it's not scanning every dir during checking for the update

https://github.com/xonsh/xonsh/blob/9e5fc887478bc30aafce6f21b9d6ab334c13b102/xonsh/commands_cache.py#L245
```py
        for path in paths:
            modified_time = os.path.getmtime(path)
            if (
                (not self.env.get("ENABLE_COMMANDS_CACHE", True))
                or (path not in self._paths_cache)
                or (self._paths_cache[path].mtime != modified_time)
            ):
                updated = True
                self._paths_cache[path] = _Commands(
                    modified_time, tuple(executables_in(path))
                )
```

> Startup time on my mac: xonsh --no-rc --timings -> on_pre_prompt ~ 0.3 s - 10 runs.

Well, there is a special `$COMMANDS_CACHE_SAVE_INTERMEDIATE` to 
> If enabled, the CommandsCache is saved between runs and can reduce the startup time.

It's disabled by default (why?), so yeah you'd get a delay listing all files in all path dirs with `--no-rc`

> to just `resolve_executable` by checking the resolved command name in `$PATH`. Instead of doing this simple action we're using CC with `cc.update_cache`.

This turned out to be not a simple action after all, but the one with typing lag since you need to check dozens of files in dozens of dirs (at least on Windows). Getting one mtimes for a dir seems to be faster than getting dozens of file_exist in that same dir

So I'm not sure this was a good idea to disable commands cache based on these faulty assumptions.

Maybe instead the `COMMANDS_CACHE_SAVE_INTERMEDIATE` should be enabled by default to save startup time?

Then you could reduce delays further by reducing the number of dirs for which you do `mtime` checks
  - having some dirs in "permacache", never do mtime (like those System32 dirs)
  - some in "cache per prompt", check mtime for each prompt only
  - some in "cache per session", check mtime once per xonsh session

This should quickly get down to no delays on every keystroke while maintaining syntax highlighting since you'd only check ~few mtimes and see that nothing changed and not do any more IO

(though the last 2 are lossy and might only be worth for syntax highlighting, but not for resolving)

eugenesvk:
> If mtime is working we should have the same time with or without System32 but no.

No, we shouldn't, for the reasons explained above. 
- startup: If you don't save a cache of 18,222 files from 54 directories, you're going to waste time recreating that cache on each startup
- typing: if you add 54 empty dirs to PATH, you're going to get higher lag since very dir will have to be checked `is_exist` and then `mtime`'d

PR #5579, opened 2024-07-05
spec modifier: rename to decorator alias

anki-code:
### Motivation

After using #5574, #5443 and [xontrib-spec-mod](https://github.com/anki-code/xontrib-spec-mod) I see that conch in the beginning of spec modifier looks awesome because this is a decorator alias in fact and in the future it can work to decorate any part of command pipeline. So we have to rename SpecModifierAlias to DecoratorAlias to be clear and on the one wave with Python.

### Before

`xthread`, `xunthread`, `SpecModifierAlias`.

### After

`@thread`, `@unthread`, `DecoratorAlias`.

## For community

### DIFF
--- a/xonsh/aliases.py
+++ b/xonsh/aliases.py
@@ -1064,11 +1064,11 @@ def make_default_aliases():
         "completer": xca.completer_alias,
         "xpip": detect_xpip_alias(),
         "xonsh-reset": xonsh_reset,
-        "xthread": SpecAttrDecoratorAlias(
+        "@thread": SpecAttrDecoratorAlias(
             {"threadable": True, "force_threadable": True},
             "Mark current command as threadable.",
         ),
-        "xunthread": SpecAttrDecoratorAlias(
+        "@unthread": SpecAttrDecoratorAlias(
             {"threadable": False, "force_threadable": False},
             "Mark current command as unthreadable.",
         ),


docs/tutorial.rst
```diff
@@ -1609,26 +1609,26 @@ best used in conjunction with the ``unthreadable`` decorator.  For example:
 Note that ``@()`` is required to pass the python list ``args`` to a subprocess
 command.
 
-Specification Modifier Aliases
-------------------------------
+Decorator Aliases
+-----------------
 
-Using ``SpecAttrModifierAlias`` and callable ``output_format`` you can
+Using ``DecoratorAlias`` and ``SpecAttrDecoratorAlias`` and callable ``output_format`` you can
```
jaraco, line 1615: This change introduces for the first time the `DecoratorAlias` but doesn't include it in the example. Should the documentation be expanded to also illustrate when and how the base class should be used, and maybe the relationship between the two?

xonsh/procs/specs.py
```diff
@@ -738,9 +736,9 @@ def resolve_alias(self):
                     # E.g. `alias == ['ls', '-la']`
                     self.alias = alias
 
-            if found_spec_modifiers:
-                for mod in found_spec_modifiers:
-                    self.add_spec_modifier(mod)
+            if decorators:
```
jaraco, line 739: Unrelated to this PR, this if block is unnecessary.

news/rename_xthread.rst
```diff
@@ -0,0 +1,24 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* Rename: ``SpecModifier``, ``xthread``, ``xunthread`` renamed to ``DecoratorAlias``, ``@thread``, ``@unthread`` to support idea that
+  spec modifier is like a Python decorator.
```
jaraco, line 8: I don't see where `@thread` or `@unthread` is presented. I see some tests that construct such names, but nothing in the running code.

I fumbled trying to use `@thread` but was unsuccessful:

```
 @ $(@thread echo 'foo')
/var/folders/f2/2plv6q2n7l932m2x004jlw340000gn/T/pip-run-s1tb1kdf/xonsh/parsers/ast.py:143: DeprecationWarning: Constant.__init__ got an unexpected keyword argument 'is_raw'. Support for arbitrary keyword arguments is deprecated and will be removed in Python 3.15.
  return Constant(value=s, kind="str", **kwargs)
xonsh: subprocess mode: command not found: '@thread'
Did you mean the following?
    xthread:  Alias
''
```

Maybe I'm not using it right, but it sure looks to me like `xthread` is still the presented name. What am I missing?

jaraco (approved):
I'm slightly uneasy about naming things based on their language construct, in this case a "decorator". It's almost always the case that this sort of naming is both overspecified (capturing language details) and lacking specificity (not addressing the core semantic concept of the object). Consider the basic example:

```python
string = 'foo'
```

In that expression, one cannot tell what "string" represents, only that it's probably meant to be used as a string. Much better would be to name the object based on what it is rather than how it is to be used:

```
username = 'foo'
# or
username : str = 'foo'
```

In the same way, `SpecModifier` was more direct about its purpose and semantic meaning than `Decorator`. Is this change aiming to expand the scope of a SpecModifier to something broader?

On the other hand, `xonsh` is not just built in Python, but inhabits the same spirit. Concepts of `xonsh` are modeled in Python. So to the extent that a user thinks of the essential purpose of these objects as "decorator aliases", the new naming is appropriate and informative.

The one thing I'd like to ask - is there a naming of these objects that would be just as useful to users if it were named something else (like `SpecModifier` or maybe something more general), or is `Decorator` the most precise descriptor for the construct we're aiming to present?

jaraco (approved):

PR #5586, opened 2024-07-09
xontrib: do not autoload xontribs in ``xonsh --no-rc`` mode

anki-code:
### Motivation

Basically we treat `xonsh --no-rc` mode as pure mode but I notices that xontribs autoload is working in this mode.
This PR is to swtich it off. 

### Notes

* I noticed that reading the xontribs list for autoloaders takes time. It will be cool to avoid this completely e.g. when there is no xontrib module in site-packages.
* It's not good to name env like `XONTRIBS_AUTOLOAD_DISABLED`. We need to have `$XONTRIBS_AUTOLOAD=True` instead of this.

## For community


jnoortheen (approved):

jnoortheen:
I am ok with the change.

PR #5588, opened 2024-07-10
builtins: add color to printx

anki-code:
### Motivation

To simplify colored printing. 

### Before

```xsh
printx(f"{{RED}}Error in this {file}{{RESET}}")  # looks overloaded
```

### After

```xsh
printx(f"Error in this {file}", color="RED")
printx(f"Error in this {file}", "RED")
```

## For community

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -57,8 +57,10 @@ from xonsh.platform import (
 
 
 @contextmanager
-def chdir(adir):
+def chdir(adir, mkdir=False):
     old_dir = os.getcwd()
+    if mkdir:
+        os.makedirs(adir, exist_ok=True)
     os.chdir(adir)
     try:
         yield
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -276,6 +276,11 @@ Running subprocess commands should work like in any other shell.
 
 This should feel very natural.
 
+.. note::
+
+    Access the last run subprocess command using ``__xonsh__.last``;
+    e.g. to get the return code, run ``__xonsh__.last.rtn``.
+
 
 Python-mode vs Subprocess-mode
 ================================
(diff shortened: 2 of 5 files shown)


xonsh/tools.py
```diff
@@ -2018,11 +2018,14 @@ def format_color(string, **kwargs):
         return ansi_partial_color_format(string, style=style)
 
 
-def print_color(string, **kwargs):
+def print_color(string, color=None, **kwargs):
```
jaraco, line 2021: What do you think about making it keyword-only? I'm unfamiliar with where `printx` is exposed or used, but given the name, I'd maybe have expected it to take similar arguments to `print` (i.e. `def print(*strs, **kwargs)`. If it's keyword-only, it'll be "one best way" to do it and won't be confused with `print()` semantics.

```suggestion
def print_color(string, *, color=None, **kwargs):
```

xonsh/tools.py
```diff
@@ -2018,11 +2018,14 @@ def format_color(string, **kwargs):
...
     """Prints a string that may contain colors. This dispatched to the shell
     method of the same name. Colors will be formatted if they have not already
     been.
     """
+    if color:
+        string = f"{{{color}}}{string}{{RESET}}"
```
jaraco, line 2027: Nitpick: If it were me, I'd probably write this logic in an `colorize(string, color)` function so that (a) you don't end up mutating the value of `string` conditionally, (b) you can colorize in a single expression instead of a conditional block, and (c) you can test the behavior in isolation.

```python
def colorize(string: str, color: str | None) -> str:
    """
    >>> colorize('input', 'RED')
    '{RED}input{RESET}'
    >>> colorize('input', None)
    'input'
    """
    return f"{{{color}}}{string}{{RESET}}" if color is not None else string
```

I'm not sure if doctests are enabled on the project. If not, you'd probably want to test the behavior in the test suite. I'm unsure if the expected format is `{{RED}}` or `{RED}`.

jaraco, line 2027: Maybe consider `color.upper()` so the input can be `color='red'`?

```suggestion
        string = f"{{{color.upper()}}}{string}{{RESET}}"
```

jaraco (dismissed):

Issue #5411, opened 2024-05-13
Add `@` as alternative to `__xonsh__`


anki-code:
## Concept

Add `@` as alternative to `__xonsh__`. E.g.:

```xsh
@.env                                    # instead of `__xonsh__.env` and forget about `${...}`
@.history                                # instead of `__xonsh__.history`
@.imp.json.loads('{}')                   # https://github.com/xonsh/xonsh/pull/5595
```

### Prerequisites

1. Any time any where people want to reduce keystrokes. Since we introduced xonsh inline importer `__xonsh__.imp` (e.g. `__xonsh__.imp.json.loads()`) it has become such a handy tool and people already [implemented completion](https://github.com/xonsh/xonsh/pull/5917) and it will be cool to have a shortcut for `__xonsh__`.

2. Xonsh has cryptic syntax like `${...}` or `_.rtn` and it will be cool to have more readable alternative.

### Solution

What alternative for the xonsh session object `__xonsh__` we can introduce that:

3. don't affect namespace;
4. be handful shortcut;
5. be extendable for the future;
6. has xonsh spirit and design?

To me there is just one answer that satisfy all requirements - having `@` as alternative to `__xonsh__`.

### Why it's the best solution

1. Reduce keystrokes. We have minimal way to access it : `@.imp.json.loads()`, `@.imp.pandas.read_csv()`. Note that `@` under left hand and dot under right hand - it's convenient to type.

2. Sane syntax. We can use `@.env` instead of `${...}`, `@.last.rtn` instead of `_.rtn`, `@.history` instead of `__xonsh__`.

3. Clean namespace. With `@` we have isolated namesapce that we can fill by useful functions and keep the global namespace clean.

4. Shortcut. Can you imagine more shorter/sane/clean shortcut for the xonsh seesion object? May be `X` or `__X__`? No. There is only one sane way with `@`.

5. Extendable. Yes, we can incapsulate `@.events`, `@.history`, `@.prompt`, `@.jobs`, `@.snail` any xonsh part and it will stay redable.

6. Spirit and design. Totally. You see `@` on xonsh [icon](https://github.com/anki-code/xonsh-logo). You see `@` as prompt end. So `@` associated with xonsh absolutely. You need just one time to understand that `@` is the session object and you can interact with it.
And this is not an invention. It has been around for almost 20 years [jQuery](https://api.jquery.com/jQuery/) uses `$.` as the way to call the main global jQ object.
[...]

PR #5595, opened 2024-07-11
builtin: added ``__xonsh__.imp`` to have an ability to import and use modules immediately in one line

anki-code:
### Motivation

In the shell experience we want to import modules and use the functionality as fast as possible without additional keystrokes.

For a long time I'm using this sugar from [xonsh-cheatsheet](https://github.com/anki-code/xonsh-cheatsheet):
```xsh
imp = type('ImpCl', (object,), {'__getattr__':lambda self, name: __import__(name) })()

j = imp.json.loads('{}')
d = imp.datetime.datetime.now()
```
And now it's time to bring this to xonsh core.

Moreover in the future we will use `@` as alternative for `__xonsh__` and the final solution will look pretty elegant and with awesome xonsh style. (https://github.com/xonsh/xonsh/issues/5411#issuecomment-2223362149)

### Before

```xsh
import json
j = json.loads('{}')

from xonsh.tools import chdir
with chdir('/tmp'):
    echo > new.txt

xonsh -c 'import json; json.loads('{}')'
```

### After

```xsh
j = __xonsh__.imp.json.loads('{}')
```

Future:
```xsh
j = @.imp.json.loads('{}')
d = @.imp.datetime.datetime.now()

with @.imp.xonsh.tools.chdir('/tmp/'):
    echo > new.txt

xonsh -c "@.imp.json.loads('{}')"
```

# Notes

* In the future we can enrich inline importer e.g. `loads, path = @.imp("json.loads", "pathlib.Path")`.
* We can implement this as [DecoratorAlias](https://github.com/xonsh/xonsh/pull/5443) e.g. `j = $(@.imp.json echo '{}')` to avoid writing any additional import lines.

## For community

### DIFF
--- a/xonsh/built_ins.py
+++ b/xonsh/built_ins.py
@@ -524,7 +524,7 @@ class InlineImporter:
     """Inline importer allows to import and use module attribute or function in one line."""
 
     def __getattr__(self, name):
-        if name.startswith('__'):
+        if name.startswith("__"):
             return getattr(super(), name)
         return __import__(name)
 


jaraco:
> Moreover in the future we will use `@` as alternative for `__xonsh__` and the final solution will look pretty elegant and with awesome xonsh style.

Have we also considered allowing for 🐚 as another alternative for `__xonsh__`? Surely it's nice to have an ASCII option, but if it's viable to use a unicode character, it might also be cool if a little weird to use an actual shell.

jaraco (approved):

Issue #5612, opened 2024-07-18
Slow key repeat on Windows


eugenesvk:
## Current Behavior
Was excited to see that v18 was supposed to reduce input lag, but it didn't seem to have helped. Am I missing some extra knobs that could help achieve that stated goal?

With `xonsh --no-rc --no-env`

https://github.com/user-attachments/assets/b75930c8-4e08-4710-aa99-37781f2ab7ee

(with actual config it's more noticably slower)

another empty shell for comparison

https://github.com/user-attachments/assets/6c161006-c82a-494b-98c2-bd03cefb0754

## Expected Behavior
Input lag getting smaller

## xonfig

<details>

```xsh
# Please paste the output of the `xonfig` command here.
+-----------------------------+-----------------+
| xonsh                       | 0.18.2          |
| Python                      | 3.12.0          |
| PLY                         | 3.11            |
| have readline               | False           |
| prompt toolkit              | 3.0.47          |
| shell type                  | prompt_toolkit  |
| history backend             | json            |
| pygments                    | 2.18.0          |
| on posix                    | False           |
| on linux                    | False           |
| on darwin                   | False           |
| on windows                  | True            |
| on cygwin                   | False           |
| on msys2                    | False           |
| is superuser                | False           |
| default encoding            | utf-8           |
| xonsh encoding              | utf-8           |
| encoding errors             | surrogateescape |
| xontrib                     | []              |
| RC file                     | []              |
| UPDATE_OS_ENVIRON           | False           |
| XONSH_CAPTURE_ALWAYS        | False           |
| XONSH_SUBPROC_OUTPUT_FORMAT | stream_lines    |
| THREAD_SUBPROCS             | True            |
| XONSH_CACHE_SCRIPTS         | True            |
+-----------------------------+-----------------+
```

</details>

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

eugenesvk:
On careful video comparison there is a visible improvement!! Thanks for your patches!

https://github.com/user-attachments/assets/d12d9eab-cdda-4d1d-ab79-00b0db33316a

Still slow on real configs, though :(

eugenesvk:
Your pointers don't resolve this valid issue of significant input lag on Windows, so not good reasons to close

1. What do you check the file name against that it's so slow? And why does the check continue even if there are no matches? Which function does the check? Maybe could see which part is slow. Because unfortunately it's not much faster practically

[...]

PR #5620, opened 2024-07-22
fix(cache): Cache clean paths for per-keystroke checks

eugenesvk:
Cache a list of cleaned paths to avoid wasting time cleaning it on every keystroke (which on Windows seems to be especially noticeable since the number of items in path is higher and isdir/isrealpath seems to be more expensive? and can be in some cases as expensive as actually checking whether the typed string exists as an executable)

(this is on top https://github.com/xonsh/xonsh/pull/5652 to avoid merging conflicts)

This is done only for the default XSH environment (when no env is passed)
And cache is set to dirty (and reset on the next request) on new prompt (before every `on_pre_prompt_format` event is fired)

**Assumption** here is that while you're typing you can't change the list of dirs in `$PATH` (NOT the list of actual files within those dirs, this PR is only about cleaning up the list of dirs in the env var, file check is done separately and this PR doesn't improve that part), so there is no risk of missing some color highlighting just because a new folder with the command you're typing is added while you're typing it

2. no docs as it's a perf improvement
3. Add the example of usage or before-after behavior: regular typing after is faster than before on Windows
4. partially addresses https://github.com/xonsh/xonsh/issues/5612 (the other part should come from caching the list of dirs/files https://github.com/xonsh/xonsh/issues/5541)

Sorry, dont get these tests: for example, the [Mac](https://github.com/xonsh/xonsh/actions/runs/10046102773/job/27764856296?pr=5620) one says 
`tests/aliases/test_source.py .F..                                        [  0%]`

but local `pytest tests/aliases/test_source.py` doesn't fail?

## For community

### DIFF
--- a/docs/platform-issues.rst
+++ b/docs/platform-issues.rst
@@ -309,5 +309,5 @@ Typing can be slow due to testing whether the typed text is an executable file
 - for each of 10+ file.pathext
 
 For smaller dirs (~few dozen files) it's faster to list the dir, so you can add such dirs to
-:ref:`$XONSH_WIN_PATH_DIRS_TO_LIST <xonsh_win_path_dirs_to_list>`
+:ref:`$XONSH_DIR_CACHE_TO_LIST <xonsh_dir_cache_to_list>`
 to reduce typing lag
--- a/docs/platform-issues.rst
+++ b/docs/platform-issues.rst
@@ -296,3 +296,18 @@ Although not recommended, to restore the behavior found in the
     >>> $PATH.append('.')
 
 Add that to ``~/.xonshrc`` to enable that as the default behavior.
+
+
+Reduce typing delay
+^^^^^^^^^^^^^^^^^^^
+
+Typing can be slow due to testing whether the typed text is an executable file
+(for color highlighting), which tests whether a file exists:
+
+- for each of the dozen of dirs in ``PATH``
+
+- for each of 10+ file.pathext
+
+For smaller dirs (~few dozen files) it's faster to list the dir, so you can add such dirs to
+:ref:`$XONSH_WIN_PATH_DIRS_TO_LIST <xonsh_win_path_dirs_to_list>`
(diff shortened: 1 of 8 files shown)


eugenesvk:
before every call to check executable:
cleaning path 0.094s
checking files 0.094s

After
cleaning path 0.0s
checking files 0.094s

from https://github.com/xonsh/xonsh/issues/5612#issuecomment-2241698846

eugenesvk:
> For example cache for clean PATH can be part of commands_cache object.

This creates a circular import, see the lastest reverted commit

> * Also it will be good to have `clean_path_cache=False` as argument by default and use `True` in pygments hook to directly say that it uses cache.

why should other users (e.g., xontribs) suffer? They've already lost access to commands cache

eugenesvk:
Added an arg to not use cache, but the issue was the default, which should be fast instead of requiring all the users to learn about this option and use it in their invokations.

eugenesvk:
> * I'm not happy with idea of is_dirty flag in PTK and readline

Why? What event do you want to hook cache reset to?

> * Having trivial global class for saving cache... it's not good to me.
What's your suggstion? The previous alternative can't be imported (also it's wrongly named, path cache isn't commands cache)

> * We have no tests for this. It will be cool to test pygments as well.

There is a test for the most important benefit - performance. What specifically do you want to test?

eugenesvk:
rebased on the latest main and also https://github.com/xonsh/xonsh/pull/5652 as otherwise they'd conflict so it's easier to use both (and anyway none are expected to be merged soon to maintan both separately)

anki-code:

eugenesvk:
The linked PR is again irrelevant here, it doesn't do anything for the path checks, it does a different async change that only hides the issue, so also should be reopened

anki-code:
Outdated

eugenesvk:

PR #5628, opened 2024-07-24
docs: Update bash to xonsh guide

anki-code:
subj

## For community

### DIFF
--- a/docs/bash_to_xsh.rst
+++ b/docs/bash_to_xsh.rst
@@ -1,6 +1,6 @@
 Bash to Xonsh Translation Guide
 ================================
-As you have probably figured out by now, xonsh is not POSIX compliant.
+Xonsh relies primarily on the Python syntax and is not a suitable replacement for `sh` and is thus considered a non-POSIX shell.
 This page provides xonsh equivalents for common patterns in Bash.
 
 .. list-table::


docs/bash_to_xsh.rst
```diff
@@ -1,11 +1,7 @@
 Bash to Xonsh Translation Guide
 ================================
-As you have probably figured out by now, xonsh is not ``sh``-lang compliant.
-If your muscles have memorized all of the Bash prestidigitations, this page
-will help you put a finger on how to do the equivalent task in xonsh.
-
-For shell scripts, the recommended file extension is ``xsh``, and shebang
-line is ``#!/usr/bin/env xonsh``.
+As you have probably figured out by now, xonsh is not POSIX compliant.
```
jaraco, line 3: I think the prior language was more correct. I recently [looked into the POSIX spec](https://youtrack.jetbrains.com/issue/IJPL-38237/Failed-to-load-environment-xonsh#focus=Comments-27-8610445.0-0) and the only thing it says about shell expectations is that `sh` must exist (not that all shells must be compatible with `sh`).

The way I'd phrase it:

```suggestion
Although xonsh aims to be largely compatible with Bash syntax, there are many cases where its Python-oriented syntax deviates from the customs of Bash, and as such, Xonsh is not a superset of `sh`.
```

docs/bash_to_xsh.rst
```diff
@@ -1,11 +1,7 @@
...
+This page will help you put a finger on how to do the equivalent task in xonsh.
```
jaraco, line 4: Without the prior mention of Bash, the phrase "equivalent task" is lacking a referent. How about:

```suggestion
This page provides xonsh equivalents for common patterns in Bash.
```

jaraco (dismissed):

docs/bash_to_xsh.rst
jaraco: > * xonsh is non-POSIX

I'm concerned about this statement. I don't know what it means. Isn't xonsh POSIX compliant in that it runs on POSIX systems? As far as I can tell, there's no spec that one could follow to be POSIX compliant and no hyperlink we could put here to help a reader understand what is the implications of being not POSIX compliant. Isn't "Xonsh is not a superset of `sh`" a more clear indication of the limitation?

```suggestion
Although xonsh takes inspiration from shells like Bash, it relies primarily on the Python syntax and is not a suitable replacement for `sh`.
```

jaraco:
```suggestion
Xonsh relies primarily on the Python syntax and is not a suitable replacement for `sh` and is thus considered a non-POSIX shell.
```

jaraco (approved):

Issue #5149, opened 2023-05-31
Refactoring: environment variables


anki-code:
Refactoring:

### First step

- [x] In completion search by substring instead of search by lprefix. Solved in #5388.
- [x] Fix setting variable using `xonsh -DVAR=VAL`. Solved in #5396.
- [x] Group env variables by xonsh components. First step is in #5501

### Second step

Xonsh has [many environment variables](https://xon.sh/envvars.html) and part of them are marked as `XONSH_` and the rest not. I suggest to add `XONSH_` prefix to all environment variables that are related to xonsh itself i.e. rename `RAISE_SUBPROC_ERROR` to `XONSH_RAISE_SUBPROC_ERROR` etc. Pros: Clear understanding of `env`. The way to remove xonsh env variables by running `env | grep -v XONSH_ | grep -v XONTRIB_`.

- [x] Implemet an approach to mark env variable as deprecated and provide backwards compatibility (e.g the state of `RAISE_SUBPROC_ERROR` should update `XONSH_SUBPROC_RAISE_ERROR` values). Done in - https://github.com/xonsh/xonsh/pull/5961

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

eugenesvk:
Could we also replace or at least add an alternative naming which is lowercase? It's rather unergonomic to type all these long names ALL UPPERCASE, especially in cases like `env | grep -v XONSH_ | grep -v XONTRIB_` where you can't use auto-complete

eugenesvk:
1. And it's a bad standard forcing needless pinky Shifts.
2. The problem isn't only with grep , simply typing these long XONSH_SUBPROC_CAPTURED_PRINT_STDERR is already an issue
3. Doesn't help with the $XONSH env vars

anki-code:
What you can do to help:
1. Grab the setting group e.g. [PTKSettings](https://github.com/xonsh/xonsh/blob/ea84fd5781810ecf338455c43f066c25f7c8f83c/xonsh/environ.py#L1714). Tell us about the group you will work.
[...]

PR #5642, opened 2024-08-03
XONTRIBS_AUTOLOAD_DISABLED --> XONTRIBS_AUTOLOAD

seemethere:
Switches XONTRIBS_AUTOLOAD_DISABLED to XONTRIBS_AUTOLOAD in order to make it explicit when you disable autoload.

It might also be worth it to make this a strtobool type of string in the future but going for simplicity first.

Is part of addressing https://github.com/xonsh/xonsh/issues/5149

## For community

### DIFF
--- a/news/change_xontribs_autoload_env_var.rst
+++ b/news/change_xontribs_autoload_env_var.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* xontrib: Changed XONTRIBS_AUTOLOAD_DISABLED=1 to XONTRIBS_AUTOLOAD=False
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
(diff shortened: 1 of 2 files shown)


seemethere:
> Hey @seemethere! Nice initiative! Please take a look how environment variables defined in environ.py. And you can use type e.g. `False` instead of `"False"`. It will be cool to [use IDE for development](https://github.com/anki-code/xonsh-developer-toolkit/?tab=readme-ov-file#pycharm). Thanks!

Howdy! I switched it over to utilize the XontribSettings.XONTRIB_AUTOLOAD directly from environ.py.

In the future it might also make sense to make it so that we could load environment variables directly from those variable declarations, i.e.

```python
# These two lines should basically return the equivalent values
# old way
XontribSettings.XONTRIB_AUTOLOAD.convert(
    env.get("XONTRIB_AUTOLOAD", XontribSettings.XONTRIB_AUTOLOAD.default)
)
# potential new way
XontribSettings.XONTRIB_AUTOLOAD.get()
```

seemethere:
Actually on a second pass through I realized that XONTRIBS_AUTOLOAD_DISABLED also took a list of xontribs to not autoload, is that still a behavior that we'd like to support here?

If so does it still even make sense to change this environment variable altogether?

PR #5652, opened 2024-08-08
fix(windows): Allow listing small Win dirs in PATH instead of testing for file.ext.exists() 

eugenesvk:
Checking for an executable on every keystroke (for syntax highlighting) is slow on Windows because for each of a dozen of dirs in `PATH` you need to check whether ~15 files exist (due to `PATHEXT`)

(on non-Win this will be used later to highlight typos in command names, even though not useful for a speed boost)

This PR allows to list smaller (user-defined in `XONSH_DIR_CACHE_TO_LIST`) dirs to be listed directly, which is faster than those file.exists checks

- pre: ~2.5ms per check of 16 files per dir, which for a few dozen dirs makes it visibly slow (see #5612 for a video)
- pos: ~0.1–2 ms per dir listing (depending on how large the dir is) and checking whether a file.ext is in the list

- [x] Include a news file
- [x] docs: added tips (and think autodoc generation would pick up the desription of the new environment variable?)
- [x] before-after: no use difference, just faster if you add the var documented in the tips
- [x] added a test that checks speed of executable lookup with smaller dirs added to the new `XONSH_DIR_CACHE_TO_LIST`
- [x] Addressing #5612

## For community

### DIFF
--- a/docs/platform-issues.rst
+++ b/docs/platform-issues.rst
@@ -309,5 +309,5 @@ Typing can be slow due to testing whether the typed text is an executable file
 - for each of 10+ file.pathext
 
 For smaller dirs (~few dozen files) it's faster to list the dir, so you can add such dirs to
-:ref:`$XONSH_WIN_PATH_DIRS_TO_LIST <xonsh_win_path_dirs_to_list>`
+:ref:`$XONSH_DIR_CACHE_TO_LIST <xonsh_dir_cache_to_list>`
 to reduce typing lag
--- a/docs/platform-issues.rst
+++ b/docs/platform-issues.rst
@@ -296,3 +296,18 @@ Although not recommended, to restore the behavior found in the
     >>> $PATH.append('.')
 
 Add that to ``~/.xonshrc`` to enable that as the default behavior.
+
+
+Reduce typing delay
+^^^^^^^^^^^^^^^^^^^
+
+Typing can be slow due to testing whether the typed text is an executable file
+(for color highlighting), which tests whether a file exists:
+
+- for each of the dozen of dirs in ``PATH``
+
+- for each of 10+ file.pathext
+
+For smaller dirs (~few dozen files) it's faster to list the dir, so you can add such dirs to
+:ref:`$XONSH_WIN_PATH_DIRS_TO_LIST <xonsh_win_path_dirs_to_list>`
+to reduce typing lag
--- a/news/listWinDir.rst
+++ b/news/listWinDir.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* (partially) Slow input lag due to testing whether the typed text is an executable file (for color highlighting) by scanning user-supplied (via XONSH_WIN_PATH_DIRS_TO_LIST, best for smaller dirs with a few dozen files) dirs in PATH instead of checking whether each defined executable extension exists
+
+**Security:**
+
+* <news item>
--- a/news/listWinDir.rst
+++ b/news/listWinDir.rst
@@ -16,7 +16,7 @@
 
 **Fixed:**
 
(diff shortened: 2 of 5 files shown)


eugenesvk:
> 1. I think we already have more simpler solution on the table:
>    * Just disable highlighting on Windows until we will have elegant solution.

this is no solution

>    * or Just use highlighting after pressing space instead of doing this every key press on Windows (I'm mostly for this or for next solution).

neither is this, real-time feedback is too valuable to give up! And it exists in other shells

>    * or Reuse commands_cache on Windows. I don't know but maybe historically it was made for this case and we need to reuse it in pygments hook for Windows again.

was it enabled before? Because xonsh has always had significant input lag even in earlier versions (though maybe it's because it was updating cache on each key stroke?)

> 2. I don't like tests with comparing timings. It's meaningless.

How it it meaningless to compare timings when the whole benefit is improved timings???
(I see that it failed in CI, but think found/fixed the reason)

> How to avoid this kind of situations. We need to talk approaches and solutions

There is currently not even a project issue opened for this huge performance issue on Windows (mine was also closed)! The other PR with noticeable  performance improvement was similarly rejected with no feasible alternative suggested.

This one is better, but also your 2 solutions are to just give up a piece of common/useful functionality

And then given the attention Windows gets since you're not using it, this might linger for another ever.

But at least now I can enjoy a less painful lag

eugenesvk:
> 1. I have a question and have no answer in previous threads: if we see that Windows subsystem works slow during checking the file name

This is about Windows, not Linux subsystem

> How it works in fish shell?

Fish doesn't work on Windows

> Is it known Windows issue?

Slow IO is a universal issue. And I've already answered you why this doesn't happen on nonWindows - because you don't do hundreds of IO checks on each keystroke! You don't have `pathext`

> What is the workaround?

Async? Cache? (I don't even see `pathext` reference in fish)

> Where is the research?

I've spend all my research credits finding the cause of the slowdown in xonsh and fixing that, wading into the unknown codebases of tools that don't work on Windows will take me much longer that the time to fix it

eugenesvk:
> 4\. Current implementation overloads the logic of resolving so much and has copy-pasting of 3 sections.

that's intentional to not touch anything in the current logic and make changes easier to review,  but consolidated one source of dupes into a separate function

eugenesvk:
> * Emotional comments about how great is the new env variable in the code and docs.

A single example for this to make any sense?

> * A person physically cannot type meaningful text at the speed with which you hold down one letter "q" to test the speed (in another thread). Therefore, I consider the pursuit of the output speed of the pressed key "q" not very representative.

Of course you can. You can't sustain this speed for a long time, but for a few chars you can type even faster since you can press 3 keys ~once in a roll like `wer`. Also the huge delay on Windows is visible without holding a key, that's just the easiest (and yes, representable) test to record

jnoortheen:
@eugenesvk I suggest closing #5652 , #5654  to use the existing mechanism of using Commands Cache once at the start of the shell initialization. There would be some compromises by enabling it back, but we can add env variables for configuring them. One issue I remember about CommandsCache was that, it is checing the mtime of every directory in the PATH at every keystroke. I think this behaviour can be controlled by an env variable to not do that by default on Windows.

eugenesvk:
The existing mechanism was disabled https://github.com/xonsh/xonsh/pull/5544 based on incorrect understanding of what was wrong, but even before that it was broken and caused noticeable lag. But it's less of an issue of which class name to use but rather what to do: if all the improvements implemented in these PRs are readded to the existing Commands cache, then you'd achieve the same great no-lagging performance

> using Commands Cache once at the start of the shell initialization

With proper caching of dirs that don't change you wouldn't even need any compromises, like listing all win paths with thousands of files on every shell init is a waste!

And I think mtime was never the issue, it was lagging due to checking dozens of extensions for a single file for every dir in path, so hundreds of io calls per key.

eugenesvk:
rebased on latest main

Spitfire1900:
This is in a pretty good state at this point. @anki-code and @eugenesvk I'd like to re-visit outstanding concerns blocking considering this for merge.

eugenesvk:
I've resonded to all the questions raised in various PRs, so what can I revisit?

Spitfire1900:
> I've resonded to all the questions raised in various PRs, so what can I revisit?

Err, I meant #5660

anki-code:

eugenesvk:
Same issue as with other PRs: the linked PR is not relevant to the changes fixed here, so should not close it

anki-code:
Outdated

eugenesvk:

PR #5654, opened 2024-08-08
fix(windows): Allow caching path dirs to achieve low input lag

eugenesvk:
This PR (with others) finally allows to have no input lag on Windows!!!

(it's on top of the other related PRs https://github.com/xonsh/xonsh/pull/5620 https://github.com/xonsh/xonsh/pull/5652 to avoid code conflicts and allow using all optimizations, and given that the expectation of merging either of the changes are the same)

It allows a user to provide a list of dirs from the PATH to cache the content of and thus avoid having to do 15+ file.exists checks in them on every keystroke.

The downside is that this cache persists per session, but for a lot of paths this is a well worthy tradeoff.

In principle, some future PR could have another config for more granular caching, e.g.:
  - cache some dirs per prompt so that you only get the lag for the first key (or not even that as some hooks may refresh the cache before a user starts typing?)
  - cache some dirs per duration (e.g., an hour?). Maybe this could be useful for some large dirs that add a noticeable delay when listing (so not great even for the first key), but yet need to be kept relatively "fresh"?
  - cache some dirs semi-permanently per OS update, but that's likely covered by this refactoring proposal https://github.com/xonsh/xonsh/issues/5541

By the way, even though the current method is precise in theory as it does the checks on every keystroke, it's imprecise in practice (though this might be some other bug, so opened a separate https://github.com/xonsh/xonsh/issues/5655):

There is one `qqqqqq.exe` executable in a dir from `PATH`, so when you type `qqqqqq` 6 times you should get a glimpse of green color

However, the current method can skip a key, so you get no color highlighting, but q×5 jumps to q×7 skipping q×6!

https://github.com/user-attachments/assets/7e2c5b92-596c-44a0-bcfe-1ac497e546e0

(not tested it thoroughly enough, it seems to work fine after you type slowly to find q×6 in path, then even holding the key matches on q×6 where it previously didn't match, so maybe it's not because it's slow but due to something else?)

If you add enough of dirs from PATHs to be cached `XONSH_WIN_DIR_SESSION_CACHE` (and yet others to be listed `XONSH_WIN_PATH_DIRS_TO_LIST` instead of  doing 15+ file.ext.exists()), then the new methods eliminates input lag completely!

https://github.com/user-attachments/assets/059a72d9-4d8e-4761-9d92-e05103355733

(the delay after the first `q` isn't lag, but platform key repeat delay)

Checklist
- [x] Include a news file 
- [x] doc: updated platform tips
- [x] before-after: no use difference, just faster if you add the var documented in the tips
- [x] addressing https://github.com/xonsh/xonsh/issues/5612

------------
ADDED
------------

> You closed a working solution with a reference to a broken one

PR #5654 vs PR #6183: Comparison

### Problem — the same

Slow syntax highlighting on Windows/WSL due to repeated `stat()` / `is_file()` calls in `locate_executable()` on every keystroke.

### Fundamental Architectural Difference: Synchronous vs Asynchronous

This is **the key** difference between the two PRs.

**#5654 — synchronous approach.** `locate_executable()` is still called **on the main thread** on every keystroke inside `_command_is_valid()`. The optimization only makes the call itself faster via caches:

```python
# pyghooks.py in #5654
def _command_is_valid(cmd):
    use_dir_session_cache = "XONSH_DIR_SESSION_CACHE" in XSH.env.keys()
    return (
        cmd in XSH.aliases
        or locate_executable(cmd, use_dir_session_cache=use_dir_session_cache)
    ) and not iskeyword(cmd)
```

**#6183 — asynchronous approach.** `locate_executable()` is moved to a **background thread** with debounce (10ms). The main thread is never blocked on I/O:

```python
# pyghooks.py in #6183
def _command_is_valid(cmd):
    if cmd in _cmd_valid_cache:
        return _cmd_valid_cache[cmd]      # O(1), no I/O
    if cmd in XSH.aliases:
        return True                        # O(1)
    # Expensive locate_executable → background thread
    _pending_cmds.add(cmd)
    _schedule_bg_validation()              # 10ms debounce + bg thread
    return False                           # Pessimistic default until validated
```

This means in #6183 **even with a cold cache** (first time, uncached directory) typing never lags. In #5654 — it lags until the cache warms up.

---

## Caching Strategies

| | #5654 | #6183 |
|---|---|---|
| **Strategies** | 3 | 1 + async |
| **PATH cleanup cache** | Yes — `PathCache` class with sha256 hash of PATH | No (unnecessary due to async) |
| **Dir listing cache** | `XONSH_DIR_CACHE_TO_LIST` — re-reads directory every time (accurate but slower) | No |
| **Session cache** | `XONSH_DIR_SESSION_CACHE` — cached for the entire session | `XONSH_COMMANDS_CACHE_READ_DIR_ONCE` — cached for the entire session |
| **Cache structure** | `dict[str, list[list[str]]]` + sha256 | `dict[str, frozenset[str]]` (simpler, O(1) lookup) |
| **Matching** | Exact path (`path in dir_to_cache`) | Prefix-based (`path.startswith(prefix)`) — more convenient |

---

## Configuration and Defaults

**#5654**: Fully opt-in, defaults to empty `set()`. The user must figure out and configure everything manually. Variable names are Windows-specific (`XONSH_WIN_DIR_SESSION_CACHE`).

**#6183**: Smart per-platform defaults:
- Windows → `['C:\Windows']` (automatically from `%WINDIR%`)
- WSL → auto-detects `/mnt/*/Windows`
- Linux/Mac → `[]`

Works out of the box. Cross-platform naming (`XONSH_COMMANDS_CACHE_*`). Includes a debug mode (`XONSH_COMMANDS_CACHE_DEBUG`).

---

## Invasiveness of Changes

**#5654** is more invasive:
- Changes signatures of `locate_executable()` and `locate_file()` (breaks API)
- Changes `clear_paths()` from `realpath` → `normpath` (global behavior change)
- Introduces `PathCache` class with `is_dirty` flag
- Modifies `ptk_shell/__init__.py` and `readline_shell.py` for cache invalidation
- 45 commits

**#6183** is less invasive:
- Does not change existing signatures — caching is transparent inside `locate_file_in_path_env()`
- Does not touch `clear_paths()`
- Uses module-level variables instead of a class
- Async logic is isolated in `pyghooks.py`
- 22 commits

---

## Where #6183 is Better

1. **Non-blocking input** — the fundamental advantage. Even with a cold cache, typing never lags. #5654 does not solve this: uncached directories will still cause lag.

2. **Works out of the box** — smart defaults vs fully manual configuration.

3. **Simpler** — fewer moving parts, no sha256 hashing, no class with dirty flag, no breaking API changes.

4. **Cross-platform** — naming and defaults account for Windows, WSL, and POSIX.

5. **Generation tokens** — correct handling of rapid typing (debounce + cancellation of stale threads).

## Where #5654 is Better

1. **No "flash of invalid command"** — in #6183 a command is briefly highlighted as invalid (red) until the background thread confirms it. In #5654 highlighting is always immediately correct (once the cache is warm).

2. **Intermediate `DIR_CACHE_TO_LIST` strategy** — for small directories, re-reads contents every time, which is more accurate than session cache but faster than per-file `stat()`. #6183 has no such middle ground.

3. **`clear_paths()` optimization** — replacing `realpath` with `normpath` removes unnecessary symlink resolution overhead. #6183 does not touch `clear_paths()`.

4. **No threading complexity** — synchronous code is easier to debug, no race conditions.

---

## Summary

Both PRs solve the same problem but at **different levels**:

- **#5654** optimizes **the call itself** to `locate_executable()` (makes it faster) but does not remove it from the main thread.
- **#6183** removes **the entire expensive call** from the main thread + caches results.

The #6183 approach is architecturally stronger because even a perfectly fast synchronous `locate_executable()` on network/WSL drives can take tens of milliseconds, while the async approach guarantees 0ms input lag. The trade-off is a brief "flash" of incorrect highlighting when typing a new command for the first time.

eugenesvk's claim that #6183 is "broken" is likely related to exactly this: they see commands briefly flash green only after the background thread validates them, and consider this a bug. From their perspective, synchronous correct highlighting is more important than zero-lag input.

## For community

### DIFF
--- a/docs/platform-issues.rst
+++ b/docs/platform-issues.rst
@@ -320,3 +320,14 @@ This will cache a list of files within these dirs per Xonsh session and thus avo
 on subsequent typing.
 However, if Xonsh or any other process changes the list of files in these dirs, you'll lose the
 accuracy of syntax highlighting since the cache will not be updated to reflect it
+Since paths are normalized before comparison, you might also want to do so in your config
+to avoid mismatches:
+
+.. code-block:: xonshcon
+
+    from xonsh.built_ins import XSH
+    import os
+    xonsh_win_path_dirs_to_list = [R"C:\my\path\to\list",]
+    XSH.env["XONSH_WIN_PATH_DIRS_TO_LIST"] = set(os.path.normpath(p) for p in xonsh_win_dir_session_cache)
+    xonsh_win_dir_session_cache = [R"C:\my\path\to\cache",]
+    XSH.env["XONSH_WIN_DIR_SESSION_CACHE"] = set(os.path.normpath(p) for p in xonsh_win_dir_session_cache)
--- a/docs/platform-issues.rst
+++ b/docs/platform-issues.rst
@@ -311,3 +311,12 @@ Typing can be slow due to testing whether the typed text is an executable file
 For smaller dirs (~few dozen files) it's faster to list the dir, so you can add such dirs to
 :ref:`$XONSH_DIR_CACHE_TO_LIST <xonsh_dir_cache_to_list>`
 to reduce typing lag
+
+And to the typing lag even more (at the cost of a small loss of precision of syntax highlighting)
+you can list all the rarely changing (like ``C:\Program Files\Python\``) or not very important dirs
+from your ``$PATH`` in
+:ref:`$XONSH_WIN_DIR_SESSION_CACHE <xonsh_win_dir_session_cache>`
+This will cache a list of files within these dirs per Xonsh session and thus avoid any IO checks
+on subsequent typing.
+However, if Xonsh or any other process changes the list of files in these dirs, you'll lose the
+accuracy of syntax highlighting since the cache will not be updated to reflect it
--- a/docs/platform-issues.rst
+++ b/docs/platform-issues.rst
@@ -309,5 +309,5 @@ Typing can be slow due to testing whether the typed text is an executable file
 - for each of 10+ file.pathext
 
 For smaller dirs (~few dozen files) it's faster to list the dir, so you can add such dirs to
-:ref:`$XONSH_WIN_PATH_DIRS_TO_LIST <xonsh_win_path_dirs_to_list>`
+:ref:`$XONSH_DIR_CACHE_TO_LIST <xonsh_dir_cache_to_list>`
 to reduce typing lag
--- a/docs/platform-issues.rst
+++ b/docs/platform-issues.rst
@@ -296,3 +296,18 @@ Although not recommended, to restore the behavior found in the
     >>> $PATH.append('.')
 
 Add that to ``~/.xonshrc`` to enable that as the default behavior.
+
+
+Reduce typing delay
+^^^^^^^^^^^^^^^^^^^
+
+Typing can be slow due to testing whether the typed text is an executable file
+(for color highlighting), which tests whether a file exists:
+
+- for each of the dozen of dirs in ``PATH``
+
+- for each of 10+ file.pathext
+
+For smaller dirs (~few dozen files) it's faster to list the dir, so you can add such dirs to
+:ref:`$XONSH_WIN_PATH_DIRS_TO_LIST <xonsh_win_path_dirs_to_list>`
+to reduce typing lag
--- a/news/cacheWinPathSession.rst
+++ b/news/cacheWinPathSession.rst
@@ -16,7 +16,7 @@
 
 **Fixed:**
 
-* (by a lot, though with a small loss of precision) Slow input lag on Windows due to testing whether the typed text is an executable file (for color highlighting) by caching a list of files in a user configured dirs (``$XONSH_WIN_DIR_SESSION_CACHE``) for the duration of the Xonsh session.
+* (by a lot, though with a small loss of precision) Slow input lag on Windows due to testing whether the typed text is an executable file (for color highlighting) by caching a list of files in a user configured dirs (``$XONSH_DIR_SESSION_CACHE``) for the duration of the Xonsh session (so any changes to executable files within this session will not be visible to the color highlighter).
 
 **Security:**
 
--- a/news/cacheWinPathSession.rst
+++ b/news/cacheWinPathSession.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* (by a lot, though with a small loss of precision) Slow input lag on Windows due to testing whether the typed text is an executable file (for color highlighting) by caching a list of files in a user configured dirs (``$XONSH_WIN_DIR_SESSION_CACHE``) for the duration of the Xonsh session.
+
+**Security:**
+
+* <news item>
--- a/news/constPath.rst
+++ b/news/constPath.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* (partially) Slow input lag due to testing whether the typed text is an executable file (for color highlighting) by not cleaning up PATH on every keystroke, but using a cached version instead.
+
+**Security:**
+
+* <news item>
--- a/news/listWinDir.rst
+++ b/news/listWinDir.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* (partially) Slow input lag due to testing whether the typed text is an executable file (for color highlighting) by scanning user-supplied (via XONSH_WIN_PATH_DIRS_TO_LIST, best for smaller dirs with a few dozen files) dirs in PATH instead of checking whether each defined executable extension exists
+
+**Security:**
+
+* <news item>
--- a/news/listWinDir.rst
+++ b/news/listWinDir.rst
@@ -16,7 +16,7 @@
 
 **Fixed:**
 
-* (partially) Slow input lag due to testing whether the typed text is an executable file (for color highlighting) by scanning user-supplied (via XONSH_WIN_PATH_DIRS_TO_LIST, best for smaller dirs with a few dozen files) dirs in PATH instead of checking whether each defined executable extension exists
+* (partially) Slow input lag due to testing whether the typed text is an executable file (for color highlighting) by scanning user-supplied (via XONSH_DIR_CACHE_TO_LIST, best for smaller dirs with a few dozen files) dirs in PATH instead of checking whether each defined executable extension exists
 
 **Security:**
 
--- a/tests/procs/test_executables.py
+++ b/tests/procs/test_executables.py
@@ -103,24 +103,21 @@ def test_locate_file_clean_path_cache_time(tmpdir, xession):
     t0 = ttime()
     _f = locate_executable("nothing")
     t1 = ttime()
-    _f = locate_executable("nothing")
-    t2 = ttime()
-    _f = locate_executable("nothing")
-    t3 = ttime()
     dur1 = (t1 - t0) / ns
-    dur2 = (t2 - t1) / ns
-    dur3 = (t3 - t2) / ns
+    t0 = ttime()
+    iters = 100
+    for _i in range(iters):
+        _f = locate_executable("nothing")
+    t1 = ttime()
+    dur2 = (t1 - t0) / iters / ns
     env = xession.env
     env_path = env.get("PATH", [])
     if env_path and dur1 > 0:
         print("$PATH length = ", len(env_path))
-        print(
-            f"t1 (no cache) = {dur1:.6f}\nt2 (   cache) = {dur2:.6f}\nt3 (   cache) = {dur3:.6f}"
-        )
+        print(f"t1 (no cache) = {dur1:.6f}\nt2 (   cache) = {dur2:.6f}")
         assert (
             dur2 < 0.90 * dur1
-        )  # 2nd run with cache should always be noticeable faster
-        assert dur3 < 0.90 * dur1  # as well as all the subsequent calls
+        )  # 2nd+ run with cache should always be noticeable faster
 
 
 def test_xonsh_dir_cache_to_list(tmpdir, xession):
--- a/tests/procs/test_executables.py
+++ b/tests/procs/test_executables.py
@@ -193,3 +193,50 @@ def test_xonsh_dir_cache_to_list(tmpdir, xession):
     )
     print(f"🕐{dur1:.6f}\t(file.ext)\n🕐{dur2:.6f}\t(list small dirs)\t")
     assert dur2 < 0.90 * dur1  # listing method should be noticeable faster
+
+
+def test_xonsh_win_dir_session_cache(tmpdir, xession):
+    # check that adding dirs to a XONSH_WIN_DIR_SESSION_CACHE var to cache a list of files in
+    # them per session instead of checking for the existence of every file.pathext
+    # is faster
+    if not ON_WINDOWS:
+        return
+
+    from os import walk
+
+    xonsh_win_dir_session_cache = []
+    env = xession.env
+    pathext = env["PATHEXT"]
+    env_path = env.get("PATH", [])
+    for path in env_path:  # cache all dirs in path
+        f = []
+        for _dirpath, _dirnames, filenames in walk(path):
+            f.extend(filenames)
+            break
+        xonsh_win_dir_session_cache += [path.rstrip(os.path.sep)]
+
+    from math import pow
+    from time import monotonic_ns as ttime
+
+    ns = pow(10, 9)  # nanosecond, which 'monotonic_ns' are measured in
+
+    env["XONSH_WIN_DIR_SESSION_CACHE"] = None
+    t0 = ttime()
+    for _i in range(100):
+        f = locate_executable("nothing")
+    t1 = ttime()
+    dur1 = (t1 - t0) / ns
+
+    env["XONSH_WIN_DIR_SESSION_CACHE"] = xonsh_win_dir_session_cache
+    f = locate_executable("nothing")  # to cache dirs
+    t0 = ttime()
+    for _i in range(100):
+        f = locate_executable("nothing")
+    t1 = ttime()
+    dur2 = (t1 - t0) / ns
+
+    print(
+        f"{len(pathext)} file.exists checks; {len(xonsh_win_dir_session_cache)} paths from ∑{len(env_path)}"
+    )
+    print(f"🕐{dur1:.6f}\t(file.ext)\n🕐{dur2:.6f}\t(cache dirs)\t")
+    assert dur2 < 0.90 * dur1  # caching dirs should be noticeable faster
--- a/tests/procs/test_executables.py
+++ b/tests/procs/test_executables.py
@@ -196,7 +196,7 @@ def test_xonsh_dir_session_cache(tmpdir, xession):
     # check that adding dirs to a XONSH_DIR_SESSION_CACHE var to cache a list of files in
     # them per session instead of checking for the existence of every file.pathext
     # is faster
-    if not ON_WINDOWS: # other OS use this for syntax checking benefit, not speed
+    if not ON_WINDOWS:  # other OS use this for syntax checking benefit, not speed
         return
 
     from os import walk
--- a/tests/procs/test_executables.py
+++ b/tests/procs/test_executables.py
@@ -112,7 +112,7 @@ def test_locate_file_clean_path_cache_time(tmpdir, xession):
     dur3 = (t3 - t2) / ns
     env = xession.env
     env_path = env.get("PATH", [])
-    if env_path:
+    if env_path and dur1 > 0:
         print("$PATH length = ", len(env_path))
         print(
             f"t1 (no cache) = {dur1:.6f}\nt2 (   cache) = {dur2:.6f}\nt3 (   cache) = {dur3:.6f}"
--- a/tests/procs/test_executables.py
+++ b/tests/procs/test_executables.py
@@ -91,3 +91,45 @@ def test_locate_file(tmpdir, xession):
     with xession.env.swap(PATH=[str(bindir1), str(bindir2), str(bindir3)]):
         f = locate_file("findme")
         assert str(f) == str(file)
+
+def test_xonsh_win_path_dirs_to_list(tmpdir, xession):
+    # check that adding smaller dirs to a XONSH_WIN_PATH_DIRS_TO_LIST var to list them
+    # instead of checking for the existence of every file.pathext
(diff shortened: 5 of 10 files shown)


eugenesvk:
**Question:** why does clear_paths resolve symlinks? If I have a symlinked dir in my `$PATH` and I add that item to be cached, it fails since searching for an executable invokes `clear_paths`, which iterates not over the paths users defined, but over symlinked-resolved paths

```py
def clear_paths(paths):
    """Remove duplicates and nonexistent directories from paths."""
    return filter(os.path.isdir, unique_everseen(map(os.path.realpath, paths)))
```

eugenesvk:
I've responded to all points in that note

eugenesvk:
rebased on the lastest main/other prs

anki-code:

eugenesvk:
You closed a working solution with a reference to a broken one

anki-code:
Just to the history. Comparison PR #5654 and PR #6183 made by an independent model on request:

```
please compare https://github.com/xonsh/xonsh/pull/5654 and https://github.com/xonsh/xonsh/pull/6183
and tell the difference, pros and cons of every approach
```
The result:

### Problem — the same

Slow syntax highlighting on Windows/WSL due to repeated `stat()` / `is_file()` calls in `locate_executable()` on every keystroke.

### Fundamental Architectural Difference: Synchronous vs Asynchronous

This is **the key** difference between the two PRs.

**#5654 — synchronous approach.** `locate_executable()` is still called **on the main thread** on every keystroke inside `_command_is_valid()`. The optimization only makes the call itself faster via caches:

```python
# pyghooks.py in #5654
def _command_is_valid(cmd):
    use_dir_session_cache = "XONSH_DIR_SESSION_CACHE" in XSH.env.keys()
    return (
        cmd in XSH.aliases
        or locate_executable(cmd, use_dir_session_cache=use_dir_session_cache)
    ) and not iskeyword(cmd)
```

**#6183 — asynchronous approach.** `locate_executable()` is moved to a **background thread** with debounce (10ms). The main thread is never blocked on I/O:

```python
# pyghooks.py in #6183
def _command_is_valid(cmd):
    if cmd in _cmd_valid_cache:
        return _cmd_valid_cache[cmd]      # O(1), no I/O
    if cmd in XSH.aliases:
        return True                        # O(1)
    # Expensive locate_executable → background thread
    _pending_cmds.add(cmd)
    _schedule_bg_validation()              # 10ms debounce + bg thread
    return False                           # Pessimistic default until validated
```

This means in #6183 **even with a cold cache** (first time, uncached directory) typing never lags. In #5654 — it lags until the cache warms up.

---

## Caching Strategies

| | #5654  | #6183  |
|---|---|---|
| **Strategies** | 3 | 1 + async |
| **PATH cleanup cache** | Yes — `PathCache` class with sha256 hash of PATH | No (unnecessary due to async) |
| **Dir listing cache** | `XONSH_DIR_CACHE_TO_LIST` — re-reads directory every time (accurate but slower) | No |
| **Session cache** | `XONSH_DIR_SESSION_CACHE` — cached for the entire session | `XONSH_COMMANDS_CACHE_READ_DIR_ONCE` — cached for the entire session |
| **Cache structure** | `dict[str, list[list[str]]]` + sha256 | `dict[str, frozenset[str]]` (simpler, O(1) lookup) |
| **Matching** | Exact path (`path in dir_to_cache`) | Prefix-based (`path.startswith(prefix)`) — more convenient |

---

## Configuration and Defaults

**#5654**: Fully opt-in, defaults to empty `set()`. The user must figure out and configure everything manually. Variable names are Windows-specific (`XONSH_WIN_DIR_SESSION_CACHE`).

**#6183**: Smart per-platform defaults:
- Windows → `['C:\Windows']` (automatically from `%WINDIR%`)
- WSL → auto-detects `/mnt/*/Windows`
- Linux/Mac → `[]`

Works out of the box. Cross-platform naming (`XONSH_COMMANDS_CACHE_*`). Includes a debug mode (`XONSH_COMMANDS_CACHE_DEBUG`).

---

## Invasiveness of Changes

**#5654** is more invasive:
- Changes signatures of `locate_executable()` and `locate_file()` (breaks API)
- Changes `clear_paths()` from `realpath` → `normpath` (global behavior change)
- Introduces `PathCache` class with `is_dirty` flag
- Modifies `ptk_shell/__init__.py` and `readline_shell.py` for cache invalidation
- 45 commits

**#6183** is less invasive:
- Does not change existing signatures — caching is transparent inside `locate_file_in_path_env()`
- Does not touch `clear_paths()`
- Uses module-level variables instead of a class
- Async logic is isolated in `pyghooks.py`
- 22 commits

---

## Where #6183 is Better

1. **Non-blocking input** — the fundamental advantage. Even with a cold cache, typing never lags. #5654 does not solve this: uncached directories will still cause lag.

2. **Works out of the box** — smart defaults vs fully manual configuration.

3. **Simpler** — fewer moving parts, no sha256 hashing, no class with dirty flag, no breaking API changes.

4. **Cross-platform** — naming and defaults account for Windows, WSL, and POSIX.

5. **Generation tokens** — correct handling of rapid typing (debounce + cancellation of stale threads).

## Where #5654 is Better

1. **No "flash of invalid command"** — in #6183 a command is briefly highlighted as invalid (red) until the background thread confirms it. In #5654 highlighting is always immediately correct (once the cache is warm).

2. **Intermediate `DIR_CACHE_TO_LIST` strategy** — for small directories, re-reads contents every time, which is more accurate than session cache but faster than per-file `stat()`. #6183 has no such middle ground.

3. **`clear_paths()` optimization** — replacing `realpath` with `normpath` removes unnecessary symlink resolution overhead. #6183 does not touch `clear_paths()`.

4. **No threading complexity** — synchronous code is easier to debug, no race conditions.

---

## Summary

Both PRs solve the same problem but at **different levels**:

- **#5654** optimizes **the call itself** to `locate_executable()` (makes it faster) but does not remove it from the main thread.
- **#6183** removes **the entire expensive call** from the main thread + caches results.

The #6183 approach is architecturally stronger because even a perfectly fast synchronous `locate_executable()` on network/WSL drives can take tens of milliseconds, while the async approach guarantees 0ms input lag. The trade-off is a brief "flash" of incorrect highlighting when typing a new command for the first time.

eugenesvk's claim that #6183 is "broken" is likely related to exactly this: they see commands flash green only after the background thread validates them, and consider this a bug. From their perspective, synchronous correct highlighting is more important than zero-lag input.

eugenesvk:
This is a bunch of garbage that you didn't spend time on proof-reading, so I'll not waste a lot of time on refuting. If you want to have a serious discussion, use your own intellience to filter it out before dumping onto others, just will point a couple of quick things

> Problem — Slow syntax highlighting

> Fundamental Architectural Difference: Synchronous vs Asynchronous

Async does **NOT** address the slow problem at all! It helps with a different one

  > -  Fully opt-in, defaults to empty set().

Thats is **YOUR OWN** personal requirement as stated in another comment, nothing prevents xonsh from adding some common folders to the cache sets!

>  Smart per-platform defaults:
  > Windows → ['C:\Windows'] (automatically from %WINDIR%)
  > Works out of the box. Cross-platform naming 

That's the same nonsense, there is nothing smart about it: not only does it not "work out of the box", but it just does not work at all: adding 1 dir out of 50 does ~nothing

anki-code:
> Async does NOT address the slow problem at all!

You talk about this, but when I ask you what you see as the reason, you avoid answering, referring me to your previous messages somewhere. This is not constructive.
We can try to fix the real reason before 0.23 by forming new PR.

> Fully opt-in, defaults to empty set().
> Thats is YOUR OWN personal requirement

This falls under the "configuration" section and does not indicate any weakness in your approach.

> Smart per-platform defaults:
> adding 1 dir out of 50 does ~nothing

'c:\windows' is a prefix (as described above). So any subdirectory will be cached. Not 1.

eugenesvk:
> but when I ask you what you see as the reason, you avoid answering, referring me to your previous messages somewhere

It wasn't "somewhere", I gave you a link (to a comment you've replied to, so might've even read) to avoid repeating myself for the 5th time that has this answer 
>it's not the reason for the slowdown, the issue is many checks per file per dir regardless of how many files there are in a dir.

It's also right **here** in this very PR, not somewhere:

> avoid having to do 15+ file.exists checks in them on every keystroke.

> This falls under the "configuration" section and does not indicate any weakness in your approach.

It does, it directly compares "user must figure out and configure everything manually" vs "Works out of the box". OOTB is a superior user experience precisely because it doesn't require spending time figuring it out!

> 'c:\windows' is a prefix (as described above). So any subdirectory will be cached. Not 1.

Ok, 4 out of 50 instead of 1. What does that change?

(By the way, this isn't a good solution either, how can you know than any subdir has the same caching requirements as its parent dir?)

anki-code:
> the issue is many checks per file per dir regardless of how many files there are in a dir.

And we're back to [comparing](https://github.com/xonsh/xonsh/pull/5654#issuecomment-4174902159) and contrasting approaches:
* Synchronous with (or with not) aggressive caching.
* or Asynchronous with deferred highlighting.

I believe, and the experience of other shells (fish is considered the most advanced), that asynchronous operation is the best option and 6183 is the right solution and direction.

The synchronous approach will always harbor problems. As soon as someone decides to add some additional functionality to the highlighting or wants to exclude something from aggressive caching, it will _immediately impact the overall input speed_. This approach may be useful, but it shouldn't be the primary one for the shell.

The asynchronous approach ensures zero input lag forever on any platform and with any implementation of the backlight logic itself.

If you want to add aggressive caching, there are two ways:
1. Testing something like this:
  ```xsh
  $XONSH_COMMANDS_CACHE_READ_DIR_ONCE = ['c:\\', 'd:\\']  # etc
  ```
  If it's not enough:
2. Suggest a PR that improves the <i>**current**</i> asynchronous solution in a way that meets your goals and we (or someone from xore) take a look how it works.


> (By the way, this isn't a good solution either, how can you know than any subdir has the same caching requirements as its parent dir?)

This is why we have minimal defaults. And users can specify what they want.

eugenesvk:
> And we're back to comparing and contrasting approaches:

Nope, how does 

> deferred highlighting

address the problem of 

> Slow syntax highlighting

???

Also,
  > Synchronous with (or with not) aggressive caching.
  > or Asynchronous with deferred highlighting.

Is a false dichotomy, caching is orthogonal. You can do
   what you do now:
  - give up on highlighting and do async
  Or you can do
  - fast highlighting via caching with async
  Or you can do what these PRs do:
  - fast highlighting via caching without async
  Or you can optionally
  - add async to it

> I believe, and the experience of other shells (fish is considered the most advanced)

"Where is the research?" What's fish experience like on Windows?

> the best option ... the right solution and direction

solution for what problem? How can you evaluate solutions and determine the best if you can't see that they don't solve the problem you write they do?

> And user can specify what he want

He can't, your cache only works as a list of prefixes!

> improves the current asynchronous 

The core improvement here is to make it **configurable** so that users relying on syntax highlighting being correct can continue to do so and decide for themselves whether the delay is worth it. (there are ways to make that decision less painful, e.g., by having the option to autoenable async if sync fails in Xms, but that's a separate improvement).


But also, no one has reviewed your vibe code, and judging by the fact that you can't even describe it properly because you use LLMs too much even for basic communication, I wouldn't trust the code, so I definitely don't plan to waste any time building on top of a broken foundation.

anki-code:
I’d like to make a small remark. **I don’t want to discuss it, just to note it here.** When someone says that typing speed is not important and syntax highlighting is very important, consider this situation:

------

Imagine you’re on an SSH server, the CPU is maxed out, and everything is running slowly. You urgently need to type a kill command to stop a frozen process. In that stressful moment, all you need is to type the command as quickly as possible. If, in those conditions, the shell starts lagging just to color the word "kill" in a nice green, that would be a completely ridiculous and stupid situation.

------

This example clearly shows what is important and what is secondary. I’m really glad that secondary features are finally being moved to asynchronous processing ([1](https://github.com/xonsh/xonsh/pull/5654#issuecomment-4174902159), [2](https://github.com/xonsh/xonsh/pull/5654#issuecomment-4175623892)).

anki-code:
> If you want to add aggressive caching, there are two ways:

I'm finish here with this thread.

anki-code:
> I wouldn't trust the code, so I definitely don't plan to waste any time building on top of a broken foundation.

To make such claims, you should test the code first. Your reports about delayed highlights are not an error or a bug — this is expected behavior for 0.23.0 and we will improve it in the future. 

Are there any real bugs in 0.23.0 I can fix? Just report them to issue tracker with the steps to reproduce and stack trace.

I'm finish this thread [here](https://github.com/xonsh/xonsh/pull/5654#issuecomment-4175623892):

> If you want to add aggressive caching, there are two ways:

eugenesvk:
> When someone says that typing speed is not important and syntax highlighting is very important, consider this situation:

That's someone you've just made up! The importance of typing speed is the whole reason I've added these PRs after years of Windows neglect by xonsh

> Imagine you’re on an SSH server

Why do I care about this hypothetical??? Imagine it's never happened in your life and you don't want to degrade 100% of your use cases for a fictional benefit?
Besides, this has the easier solution of temporarily disabling highlighting if becomes a huge blocker

> If, in those conditions, the shell starts lagging just to color the word kill in a nice green, that would be a completely ridiculous and stupid situation.

But the shell lagging and highlighting the word `klil` is not ridiculous anymore as it allows you to actually fix it instead of failing to notice the typo due to all the stress!

> This example clearly shows what is important and what is secondary

Nah, that's just fiction, which can just as easily be turned 180°: imagine you're in this real-life situation https://xkcd.com/1168/ and correctness is life-and-death. You'd surely be fine with small delays to typing so that your `tra` is highlighted as an error!

anki-code:
> I definitely don't plan to waste any time building on top of a broken foundation.

I forgot to mention: you don’t have to write the code. You can simply propose a solution in plain words, and we (or someone from xore) can see how well it fits the async approach. It’s called a PEP in the Python world :) See you

anki-code:
> you don't want to degrade 100% of your use cases

Yeah. Show 2-3 of your personal real life examples (videos) where async highlighting from 0.23.0 is not working to you personally.

The video should be:
1. Not synthetic examples with pressing one button.
2. Real typing of commands on the keyboard for real life use case.
3. Not only first start of the prompt. Show how it works on next prompts.
4. (If possible) Use `xonsh --no-rc --no-env` or `xonsh --no-rc` at least.
 
Also please show how [point 1 suggestion](https://github.com/xonsh/xonsh/pull/5654#issuecomment-4175623892) is working to you.

It will give more clear understanding of why you're so angry.

eugenesvk:
> Yeah. Show 2-3 of your personal real life examples (videos) where async highlighting from 0.23.0 is not working to you personally.

Show me at least **1** of your personal real life examples where

> "Imagine you’re on an SSH server, the CPU is maxed out, and everything is running slowly. You urgently need to type a kill...

> The video should be:

Well, any video would suffice.

> Not synthetic examples with pressing one button

<kbd>⌫</kbd> is 1 button! It exists in reality and is used in reality as 1 button. So it's not a hypothetical unlike your "examples". Now justify your demand of excluding 1 button

> Not only first start of the prompt.

And justify this as well. Do you only use xonsh in the middle of the prompt???

anki-code:
>  Do you only use xonsh in the middle of the prompt?

I mean not only the first line (not "middle" word) e.g.
```xsh
xonsh --no-rc --no-env
tar -czf src.tar.gz ~/Documents/git/myproject  # first line (populating cache maybe)
ls -al src.tar.gz  # second line
```

> Show me at least 1 of your personal real life examples where

I have no issues on Mac, Linux, and Windows 11 tests. Sometimes I see blink (expected in 0.23.0) on any system but I'm totally ok with this. If I don't pay special attention to it, I won't notice it at all.

If I connected slow drive (e.g. s3) and add it to PATH (it's not my common use case) I see highlighting lag and it's ok to me. Probably I will add s3 drive path to XONSH_COMMANDS_CACHE_READ_DIR_ONCE if I need this in real life.

> ⌫ is 1 button

Just to let anybody knows: just pressing `Ctrl+C` declined the current input without need to press and hold backspace. We need to add this to https://xon.sh/keyboard_shortcuts.html. Nice catch!

eugenesvk:
> I forgot to mention: you don’t have to write the code. You can simply propose a solution in plain words, and we (or someone from xore) can see how well it fits the async approach. 

That's already been done, with code, and none of you even managed to follow the discussion to its logical resolution (see my responses unaddressed). So why are you suggesting something that already failed?

> It’s called a PEP in the Python world :)

So where are the PEPs for the latest changes?

> I mean not only the first line (not middle) e.g.
> tar -czf src.tar.gz ~/Documents/git/myproject  # first line (populating cache maybe)
> ls -al src.tar.gz  # second line

Such a typical example of these discussions: these videos **already exist**!!! I've even written you a special text **tutorial** on how to **add a new line** to test it

> press **Enter**

https://github.com/xonsh/xonsh/issues/5655#issuecomment-4168841925

And how is this relevant?
> populating cache maybe)

when you only cache Windows folder? Come up with a real test use case that tests the **changes** that you've **added**!!! Async is architecturally incapable of improving the speed of anything, so explain what it is exactly you think you're testing when asking for more videos???
(or is this some different cache?)

anki-code:
The algorithm from https://github.com/xonsh/xonsh/issues/5655#issuecomment-4168841925 is not represent [real life use case](https://github.com/xonsh/xonsh/pull/5654#issuecomment-4176088097) (typing qqqqqq by holding the key is not the real life use case and it's [wrong test](https://github.com/xonsh/xonsh/issues/5655#issuecomment-4177019591)). Other videos from you are also synthetic and based on old version.

eugenesvk:
> typing qqqqqq by holding the key is not the real life use case). Other videos from you are also synthetic 

But we've already discussed this error of yours multiple times! If you don't address the counter-arguments you'll keep repeating the same mistake over and over and over again.

>> ⌫ is 1 button

> Just to let anybody knows: just pressing Ctrl+C declined the current input without need to press and hold backspace

So again you haven't addressed the core of it. Just to let you know: reality is not limited to "declined current input", it also includes use cases such as "decline part of current input with the most commonly used method of using ⌫"

Here is another simple case of me using ripgrep to find stuff without help from highlighter

https://github.com/user-attachments/assets/0acbd7af-70a8-4809-bcb6-0db912f8e81b

Or trying to correct a typo without highlighting help to signal there is no typo this time (the 2nd error is unrelated)

https://github.com/user-attachments/assets/3cf5365f-e261-4a9d-854b-8636709d9c23


> and based on old version.

The new version didn't improve anything outside of your confusion of async and speed, so also doesn't matter?

> I have no issues on Mac, Linux

This has also been repeated and addressed many times , how is it relevant??? Those OS don't do the same amount if IO there due to lack of PATHEXT and smaller paths, so of course lag is smaller and harder to notice?!

> , and Windows 11 tests.

Add better tests? You're easy to dismiss one of the most commonly used keys - Backspace - so maybe your tests are way too limited???

> Sometimes I see blink (expected in 0.23.0) on any system but I'm totally ok with this.

> If I don't pay special attention to it, I won't notice it at all.

So what should the users slightly higher standards that do notice do?

anki-code:
> The new version didn't improve anything outside of your confusion of async and speed, so also doesn't matter?

Are you testing on 0.23.0? It has not only async changes.

> listmediainf

How much time on 0.23.0 it takes to be green? How much dirs/files in $PATH? Is it network device or what? What about adding it to READ_ONCE

eugenesvk:
Yes, 0.23. What else does it have that would improve it this issue?
~50 dirs, not network

anki-code:
Outdated

eugenesvk:

Issue #4558, opened 2021-11-20
Launch .xsh without an extension and changing file associations (Windows)


eugenesvk:
I'd like to run `xonsh_script.xsh` files just like I can run other `executable.exe` on Windows, e.g. without an extension by simlpy invoking `xonsh_script`

There is a `$PATHEXT` environment variable that allows to consider some extensions as markers of an executable file to be auto-added to a file name passed without an extension (auto-completing it), and xonsh does do this check and auto-completion.
However, xonsh skips shebang parsing for such extensions, assuming Windows can handle these files on its own. But Windows would need to have additional file associations, and I don't want them as they mess up non-command-line operations like double-click on `xonsh_script.xsh`.

I'm not sure what a good way to achieve this separation between active file associations in a shell and no file associations on a desktop, but this [simple commit](https://github.com/eugenesvk/xonsh/commit/6b69ede18e7cf5f1c575508330ad7eca9d929f27) seems to achieve exactly what I want — I can now launch `xonsh_script` without an extension, it's being looked up  in `PATH` and launched, and I don't need to mess with desktop file associations.

(it also happens to fix the other issue https://github.com/xonsh/xonsh/issues/4555 by bypassing the new fix for when a hardcoded interpreter is found)

## xonfig

<details>

```
$ xonfig
+------------------+-----------------+
| xonsh            | 0.11.0          |
| Python           | 3.9.7           |
| PLY              | 3.11            |
| have readline    | False           |
| prompt toolkit   | 3.0.22          |
| shell type       | prompt_toolkit  |
| history backend  | json            |
| pygments         | 2.10.0          |
| on posix         | False           |
| on linux         | False           |
| on darwin        | False           |
| on windows       | True            |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
| on jupyter       | False           |
| jupyter kernel   | None            |
| xontrib          | []              |
| RC file          | []              |
+------------------+-----------------+
```

</details>

## Expected Behavior

After extending env var `$PATHEXT += ['.XSH']` in my `rc.xsh` config I should be able to launch `xonsh_script.xsh` via a no-extension command `xonsh_script` and the xonsh interpreter specified in a shebang `#!/usr/bin/env xonsh` should work

## Current Behavior

[...]

Issue #4555, opened 2021-11-20
v0.11.0 broke launching xonsh scripts from shell


eugenesvk:
I can't execute xonsh scripts anymore as I get an `OSError: [WinError 193] %1 is not a valid Win32 application` since version 0.11.0 due to this [PR](https://github.com/xonsh/xonsh/pull/4077)

If I change the [code here](https://github.com/auneri/xonsh/blob/03f1f43a0ecac01f2047e260b2a40a40a6e075ea/xonsh/procs/specs.py#L467) like so, I get the old behavior back

```py
#cmd = [self.binary_loc] + self.cmd[1:]
cmd = self.cmd
```

## xonfig

<details>

```
$ xonfig
+------------------+----------------------+
| xonsh            | 0.11.0               |
| Git SHA          | 337cf25a             |
| Commit Date      | Nov 17 15:37:41 2021 |
| Python           | 3.9.7                |
| PLY              | 3.11                 |
| have readline    | False                |
| prompt toolkit   | 3.0.22               |
| shell type       | prompt_toolkit       |
| history backend  | json                 |
| pygments         | 2.10.0               |
| on posix         | False                |
| on linux         | False                |
| on darwin        | False                |
| on windows       | True                 |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
| on jupyter       | False                |
| jupyter kernel   | None                 |
| xontrib          | []                   |
| RC file          | []                   |
+------------------+----------------------+
```

</details>

## Expected Behavior

Executing a xonsh script from the command line should work

## Current Behavior

Instead, I get an `OSError: [WinError 193] %1 is not a valid Win32 application`

This is likely due to the fact that I don't have `.xsh` as a registered file association, and I don't want to register it, I want to be able to double-click to open a file instead of execute it.
So the new changes that check `[self.binary_loc]` fails and prevents from launching xonsh scripts from the command line

### Traceback (if applicable)

<details>

```
Traceback (most recent call last):
  File "/path/to/xonsh\lib\site-packages\xonsh\procs\pipelines.py", line 160, in __init__
    proc = spec.run(pipeline_group=pipeline_group)
  File "/path/to/xonsh\lib\site-packages\xonsh\procs\specs.py", line 460, in run
    p = self._run_binary(kwargs)
[...]

PR #5659, opened 2024-08-09
fix(Windows): always run interpeted executable files via an interpreter

eugenesvk:
And avoid `OSError: [WinError 193] %1 is not a valid Win32 application` when xonsh skips parsing shebangs in `.xsh` and other files and redirects them to the OS

Given that the opened issues (see the last list item) have gotten no better resolution suggestions, I'm PRing the slightly updated fix I've been using locally 

- [x]  added news
- [x] feature docs: not a feature
- [x] before-after: same usage, just old use got broken and is now fixed
- [x] Addresses #4555 and #4558 

## For community

### DIFF
--- a/news/win_deinterp.rst
+++ b/news/win_deinterp.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* <news item>
+
+**Deprecated:**
+
+* <news item>
+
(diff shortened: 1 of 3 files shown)


anki-code:
Closed with reference to https://github.com/xonsh/xonsh/pull/6180

PR #5660, opened 2024-08-11
feat(windows): Allow perma-caching stable path dirs to achieve low input lag

eugenesvk:
This PR (with others) finally allows to have no input lag on Windows!!!

(It's on top of https://github.com/xonsh/xonsh/pull/5654 for the same reasons)

It allows caching executables in **large non-changing dirs** like `C:\Windows\System32` (user configurable) to avoid any IO during per-keystroke checks for color highlighting
(though the actual cost doesn't really depend on the size of the dir, it's the same 15+ `"file.ext".is_file()` for every `pathext`)

- Also allows caching (by mtime) current working dir
  - since (to avoid per-file IO) non-executables are also cached, pyghooks with highlight prefixes for any file, not only commands
  - this can be user-fixed by caching executables, but then it will negate the benefit of no per-file IO (at least once per every newly seen dir per session)

As a bonus it adds color highlighting for partial commands (for commands in this perma cache, session cache, and dirs listed every time) via a new `Token.Name.Cmdprefix` token scope (default is white underlines), which is enabled by the new dependency [pygtrie](https://pypi.org/project/pygtrie/) that creates a character-by-character visibility into the cache list, so for every typed string you know whether there is a command that starts with that string (as discussed in https://github.com/xonsh/xonsh/issues/5541), this could also be expanded into other command cache types
And since this cool syntax check is helpful on other platforms as well, I've removed the "Windows" limits from the caches

https://github.com/user-attachments/assets/b8138c98-614a-4b94-9696-e5e80f4b0a15

- [x] added news
- [x] doc: added to the platform issues
- [x] example: cache dirs just imrov perf
- [x] Addressing https://github.com/xonsh/xonsh/issues/5612 and also some ideas from https://github.com/xonsh/xonsh/issues/5541 and a suggested feature from https://github.com/xonsh/xonsh/discussions/5653
- [x] added cache and `Cmdprefix` color highlighting

**Questions** 
1. currently there are still 2 IO checks - if a name if found in permacache, it still checks whether a file exists (though twice)
Wouldn't it be ok to just assume it does since it's `perma` cache, so assumed to not change?
Added a new config var to allow skipping that

2. How to set the style for the new partial highlighting to only be underline without specifying the color? I've tried just setting `underline`, but then the color changes. And if I specify `underline ansiwhite` as currently, then I lose the light-gray color on hyphens
 
3. Is there an example from xonsh code I could copy to make permacache generation a non-blocking operation as it's potentially expensive to list System32 folders (though it should only happen once, so maybe not a big deal)

## For community


eugenesvk:
And I've already addressed that note

Spitfire1900:
The best test for this is to connect to a Netshare folder over a slow network. This is definitely faster yet is still has opportunities for improvement.

In the optimal case there should be no blocking item before accepting a keypress, all highlighting cases are run on a different thread and update the highlighting when available.

eugenesvk:
> This is definitely faster yet is still has opportunities for improvement.

Have you tried adding all PATHs to e.g. a session cache? Then it should have no IO per keystroke for highlighting; or maybe it bugs somewhere, or maybe xonsh does some other checks? Think the local folder check that can add lag only happens if you type `./`, but don't remember

> In the optimal case there should be no blocking item before accepting a keypress, all highlighting cases are run on a different thread and update the highlighting when available.

Sure, these PRs aren't that, they can only achive no lag at some loss of precision. There is an open issue from a few years back where the solution was to do all this highlighting stuff in its rightful place - off the main thread...

Spitfire1900:
> > This is definitely faster yet is still has opportunities for improvement.
> 
> Have you tried adding all PATHs to e.g. a session cache? Then it should have no IO per keystroke for highlighting; or maybe it bugs somewhere, or maybe xonsh does some other checks? Think the local folder check that can add lag only happens if you type `./`, but don't remember

Are these comma separated?
https://github.com/xonsh/xonsh/pull/5660/files#diff-545bb312bf20b24ad2654671db11758ff3504266bd7521e0b0ce43558603fac3

eugenesvk:
Those are regular python sets/lists, so each dir is a separate element

Spitfire1900:
@eugenesvk is the following valid configuration in `.xonshrc`?

```xonsh
$XONSH_DIR_SESSION_CACHE = $PATH[-1] # add everything but current directory to Session cache.
$XONSH_DIR_CACHE_SKIP_EXIST = True
```

eugenesvk:
But your current directory isn't part of $PATH, so -1 doesn't make sense.
Anyway, those are invalid, you'd need a list of actual dirs, these lists are static as they were meant to be used to handle static PATH var 

Interesting idea about changing the  logic for handling of the current dir, but note that it's not guaranteed to be faster, e.g. for large dirs listing would be slower than checking for 15.pathext. Though maybe if you cache it might still be worth it, especially if you don't expect to visit very large dirs often?

Spitfire1900:
See https://github.com/xonsh/xonsh/pull/5677#issuecomment-2538930507's attached video  for performance differences in a worst case scenario (corporate on-access antivirus scanner)

https://github.com/user-attachments/assets/7ca0e18f-17cf-46b2-a0b5-74e0deb14f27

Spitfire1900:
>But your current directory isn't part of $PATH, so -1 doesn't make sense (but should be valid, it should just expand to whatever your startup $PATH is less its last element and added as a list of dirs to cache).

Any empty string is the last entry in `$PATH`
```xonsh
@ $PATH[-1]
''
@ 
''`

**Comment by Spitfire1900** [comment]
@eugenesvk is there a good way to check the state of the cache while in an interactive xonsh session?

**Comment by eugenesvk** [comment]
well, yeah, you're not caching anything from your quoted config, all the cache vars are empty

`$PATH[-1]` maybe that's just a Windows issue with parsing the default path that it leaves the last one empty, but just print the full one`$PATH`, copy all paths from there and add then to one of the 3 cache dirs, you can temporarily add all of them to the permacache just for testing. Checking current dir isn't related to your $PATH var

Re. speed comparison: the best one would be
- do 2 videos of just a single prompt line with recording started one you press & hold 1 key (no need to show OSD, it doesn't help gauge the lag)
- put these videos one on top of the other 
- while recording, press play at the same time (e.g., one with a key, another with a mouse)
- then you'll se by the distance the spammed key travels how much faster one is

Similar to this: https://github.com/xonsh/xonsh/issues/5612 (though this adds 2 videos separately, but if you click play at the same time you'll see one being faster)

>  is there a good way to check the state of the cache while in an interactive xonsh session?

Nothing comprehensive, there is a singleton `PathCache` which you could inspect, or use some rudimentary summary `from xonsh.procs.executables import PathCache; pc = PathCache(None); pc.get_cache_info()` (though this one hasn't been pushed here yet)

and compare to `$XONSH_DIR_PERMA_CACHE; $XONSH_DIR_SESSION_CACHE; $XONSH_DIR_CACHE_TO_LIST` to see which dirs you wanted cache

**Comment by Spitfire1900** [comment]
What triggers whether un-cached gets cached?

```
@ PathCache(xonsh.environ).get_cache_info()
PATH    : ∑  62 dirty
          └  54 clean (unique & existing)
Cached  : ∑   0 of which:              (pc = PathCache(None))
          ├   0 permanently            (pc.usr_dir_list_perma   ← $XONSH_DIR_PERMA_CACHE  )
          ├   0 this session           (pc.usr_dir_list_session ← $XONSH_DIR_SESSION_CACHE)
          └   0 by dir mtime, list onΔ (pc.usr_dir_list_key     ← $XONSH_DIR_CACHE_TO_LIST)
Uncached: ∑  54 including:
  C:\Program Files\Git\usr\bin
  C:\Program Files\Git\mingw64\bin
  C:\Program Files\Python313\Scripts
  C:\Program Files\Python313
  C:\Program Files\Eclipse Adoptium\jdk-17.0.13.11-hotspot\bin
  ...
```

**Comment by eugenesvk** [comment]
> $XONSH_DIR_CACHE_TO_LIST = {''} # use the current directory as a list instead of checking for executablity.

You can now use `$XONSH_DIR_CWD_CACHE` and derivatives to add current working dirs to the cache using the same logic as XONSH_DIR_CACHE_TO_LIST, caching by mtime and listing on changes

**Comment by eugenesvk** [comment]
> What triggers whether un-cached gets cached?

the 3 environment variables listed in the printed output. You need to actually add dirs you want to cache to them in your config

**Comment by Spitfire1900** [comment]
Yeah, running into issues here...

```xonsh
🐚 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh                                                                                                                                    cacheWinPathPerma   xonsh-cacheWinPathPerma  24-12-14 10:19:27-05
@ from xonsh.procs.executables import PathCache; pc = PathCache(None); pc.get_cache_info()
PATH    : ∑  66 dirty
          └  55 clean (unique & existing)
Cached  : ∑   1 of which:              (pc = PathCache(None))
          ├   1 permanently            (pc.usr_dir_list_perma   ← $XONSH_DIR_PERMA_CACHE       )
          ├   0 this session           (pc.usr_dir_list_session ← $XONSH_DIR_SESSION_CACHE     )
          └   0 by dir mtime, list onΔ (pc.usr_dir_list_key     ← $XONSH_DIR_CACHE_TO_LIST     )
Uncached: ∑  54 including:
  File "<stdin>", line 1, in <module>
  File "C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py", line 334, in get_cache_info
    +("✓" if env.get("XONSH_DIR_CWD_CACHE_NON_EXE", False) else "✗")
TypeError: bad operand type for unary +: 'str'
Return 1


🐚 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh                                                                                                                                    cacheWinPathPerma   xonsh-cacheWinPathPerma  24-12-14 10:19:45-05
@ $XONSH_DIR_PERMA_CACHE
['C:\\Program Files\\Git\\usr\\bin',
 'C:\\Program Files\\Git\\mingw64\\bin',
 'C:\\TMP\\xonsh-cacheWinPathPerma\\Scripts',
 'C:\\TMP\\xonsh-cacheWinPathPerma\\Scripts',
 'C:\\Program Files\\Git\\usr\\bin',
 'C:\\Program Files\\Git\\mingw64\\bin',
 'C:\\Program Files\\Python313\\Scripts\\',
 'C:\\Program Files\\Python313\\',
...
@ s = pickle.dumps(pathlib.Path('C:/Users/XXXXXXX/.cache/xonsh/dir_perma_cache.pickle').read_bytes())
@ import pprint

  obj = pickle.loads(s)
  pprint.pprint(obj)

(b'\x80\x04\x95`\x01\x00\x00\x00\x00\x00\x00]\x94(}\x94\x8c\x01.\x94\x8c\x07py'
 b'gtrie\x94\x8c\x08CharTrie\x94\x93\x94)\x81\x94}\x94(\x8c\x05_root'
 b'\x94h\x03\x8c\x05_Node\x94\x93\x94)\x81\x94]\x94(K\x0b\x8c\x01c\x94\x8c\x01o'
 b'\x94\x8c\x01n\x94\x8c\x01f\x94\x8c\x01t\x94\x8c\x01e\x94\x8c\x01s\x94h\x11h'
 b'\x02\x8c\x01p\x94\x8c\x01y\x94\x8c\x0bconftest.py\x94J\xf5\xff\xff\xff'
 b'K\x08h\x13h\x12h\x11\x8c\x01u\x94h\x14h\x02h\x14h\x15\x8c\x08setup.py\x94J'
 b'\xf8\xff\xff\xffK\x12\x8c\x01x\x94h\x0eh\x0fh\x13\x8c\x01h\x94\x8c\x01-\x94'
 b'\x8c\x01i\x94h\x0fh\x1b\x8c\x01d\x94h\x0eh\r\x8c\x01k\x94h\x12\x8c\x01'
 b'r\x94h\x02h\x14h\x15\x8c\x12xonsh-in-docker.py\x94eb\x8c\x07_sorted'
 b'\x94\x89ubs\x8f\x94(\x8c\x04.MSC\x94\x8c\x04.BAT\x94\x8c\x04.PYW'
 b'\x94\x8c\x04.CMD\x94\x8c\x04.WSF\x94\x8c\x04.VBS\x94\x8c\x04.VBE'
 b'\x94\x8c\x04.JSE\x94\x8c\x04.WSH\x94\x8c\x04.CPL\x94\x8c\x04.COM'
 b'\x94\x8c\x04.EXE\x94\x8c\x03.JS\x94\x8c\x03.PY\x94\x90e.')

```

`.xonshrc`:
```xonsh
# https://github.com/xonsh/xonsh/pull/5660#issuecomment-2537131523
$XONSH_DIR_PERMA_CACHE  = [i for i in $PATH]
$XONSH_DIR_CACHE_SKIP_EXIST  = True
$XONSH_DIR_CWD_CACHE = True
```

Spitfire1900:
```diff
diff --git a/xonsh/procs/executables.py b/xonsh/procs/executables.py
index ffbe41d4..cc7c4696 100644
--- a/xonsh/procs/executables.py
+++ b/xonsh/procs/executables.py
@@ -324,14 +324,16 @@ class PathCache:  # Singleton
                 lbl += "S" if pn in self.usr_dir_list_session else " "
                 lbl += "M" if pn in self.usr_dir_list_key else " "
                 msg += f"\n {lbl} {p}"
+        check_or_x = "✓" if env.get("XONSH_DIR_CWD_CACHE", False) else "✗"
         msg += (
             "\n\n"
-            + ("✓" if env.get("XONSH_DIR_CWD_CACHE", False) else "✗")
+            + check_or_x
             + " current working dir cache ($XONSH_DIR_CWD_CACHE)\n"
         )
         if env.get("XONSH_DIR_CWD_CACHE", False):
-            msg += (
-                +("✓" if env.get("XONSH_DIR_CWD_CACHE_NON_EXE", False) else "✗")
+            check_or_x = "✓" if env.get("XONSH_DIR_CWD_CACHE_NON_EXE", False) else "✗"
+            msg += (""
+                + check_or_x
                 + " (cwd) cache non-executable ($XONSH_DIR_CWD_CACHE_NON_EXE)"
             )
         if len(self.cwd_too_long):
```

```xonsh
+1@ from xonsh.procs.executables import PathCache; pc = PathCache(None); pc.get_cache_info()
PATH    : ∑  66 dirty
          └  55 clean (unique & existing)
Cached  : ∑   1 of which:              (pc = PathCache(None))
          ├   1 permanently            (pc.usr_dir_list_perma   ← $XONSH_DIR_PERMA_CACHE       )
          ├   0 this session           (pc.usr_dir_list_session ← $XONSH_DIR_SESSION_CACHE     )
          └   0 by dir mtime, list onΔ (pc.usr_dir_list_key     ← $XONSH_DIR_CACHE_TO_LIST     )
Uncached: ∑  54 including:
  C:\Program Files\Git\usr\bin
  C:\Program Files\Git\mingw64\bin
  C:\TMP\xonsh-cacheWinPathPerma\Scripts
  C:\Program Files\Python313\Scripts
  C:\Program Files\Python313
  ...

✓ current working dir cache ($XONSH_DIR_CWD_CACHE)
✓ (cwd) cache non-executable ($XONSH_DIR_CWD_CACHE_NON_EXE)
```

Spitfire1900:
I wonder if non-existent dirs is impacting this?
```xonsh
+1@ _ = [print(f"{pathlib.Path(i).resolve()}, {pathlib.Path(i).is_dir()}") for i in $PATH if pathlib.Path(i).is_dir() is False]

C:\Program Files (x86)\Microsoft SQL Server\150\DTS\Binn, False
C:\Program Files (x86)\Microsoft SQL Server\140\DTS\Binn, False
C:\Program Files (x86)\Microsoft SQL Server\160\DTS\Binn, False
C:\Program Files (x86)\Meld, False
```

eugenesvk:
Fixed the wrong `+` concat

Also added experimetnal `$XONSH_DIR_CACHE_LIST_SAVE_ON_EXIT` so you could save the cache to all the listed dirs to a file so you wouldn't list them on every new xonsh session start if no dirs were modified

eugenesvk:
Also, try `from xonsh.procs.executables import PathCache; pc = PathCache(None); pc.get_cache_info(2)`, it hints that `pc.get_cache_db('p')` can get you the database for the permacache without having to unpickle yourself, otherwise it's a list of 2 caches


https://github.com/xonsh/xonsh/blob/fff344de7cbf03f0df173a8d1a546ef956eccda1/xonsh/procs/executables.py#L479-L481

eugenesvk:
I see there is indeed some bug with the environment being set too early before your `$XONSH_DIR_PERMA_CACHE` is read, for testing maybe move it to the beginning of your rc.xsh?

Spitfire1900:
> I see there is indeed some bug with the environment being set too early before your `$XONSH_DIR_PERMA_CACHE` is read

Thanks a lot @eugenesvk;  I'll wait on this before trying again. I should be able to `$XONSH_DIR_PERMA_CACHE = $PATH` in `.xonshrc` and it should handle non-existent folders and Network Drives.

Spitfire1900:
Performance still having a lot of issues, this is running a profile over typing the entire alphabet 3 times.

```
@ pstats.Stats('stats.profile').sort_stats(SortKey.CUMULATIVE).print_stats('executables.py:')
Sun Dec 15 16:07:13 2024    stats.profile

         12895477 function calls (12859914 primitive calls) in 76.906 seconds

   Ordered by: cumulative time
   List reduced from 1287 to 20 due to restriction <'executables.py:'>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        5    0.000    0.000   56.548   11.310 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:87(locate_executable)
        5    0.000    0.000   56.548   11.310 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:586(locate_file)
        5    0.001    0.000   56.548   11.310 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:716(locate_file_in_path_env)
        5    0.000    0.000   56.537   11.307 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:470(get_dir_cache_perma)
        5    0.000    0.000   56.537   11.307 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:475(update_cache)
        5    0.012    0.002   56.537   11.307 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:487(_update_paths_cache)
    12270    0.004    0.000   56.351    0.005 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:110(executables_in)
    12270    0.250    0.000   56.347    0.005 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:129(_executables_in_windows)
    65260    0.142    0.000   55.853    0.001 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:46(is_executable_in_windows)
   780200    0.343    0.000    1.183    0.000 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:58(<genexpr>)
       10    0.000    0.000    0.004    0.000 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:223(get_clean_path)
       10    0.000    0.000    0.004    0.000 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:211(get_clean_paths)
        2    0.000    0.000    0.004    0.002 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:704(hash_s_list)
        5    0.000    0.000    0.000    0.000 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:613(locate_relative_path)
        5    0.000    0.000    0.000    0.000 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:17(get_possible_names)
        5    0.000    0.000    0.000    0.000 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:386(__init__)
        5    0.000    0.000    0.000    0.000 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:409(set_usr_dir_list)
       10    0.000    0.000    0.000    0.000 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:440(cache_file)
        5    0.000    0.000    0.000    0.000 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:173(__new__)
        5    0.000    0.000    0.000    0.000 C:\TMP\xonsh-cacheWinPathPerma\src\xonsh\xonsh\procs\executables.py:83(__init__)

```

I only had `$XONSH_DIR_PERMA_CACHE  = [i for i in $PATH]` set during this run.

Spitfire1900:
As a reference with what I'm working with here the following script completes in 42 seconds on my workstation, a recently modern Windows laptop with an on-access virus scanner. On a spinning rust Linux box I have from 2016 it takes ~1 second.

```python
import os
import pathlib

from datetime import datetime

files = 0
iteration = 0
MAX_FILES = 360000
now = datetime.now()
while True:
    print(f'{iteration=}, {files=}')
    for i in pathlib.Path('C:/Program Files/Git/usr/bin/').iterdir():
        stat = os.stat(i)
        files += 1
        if files >= MAX_FILES:
            break
    if files >= MAX_FILES:
        finished = datetime.now()
        break
    else:
        iteration += 1
```

eugenesvk:
Yeah, there is already a simple performance test here, which is currently failing on Windows. Think there is some simple failing cache invalidation check that caused the caches to be rebuilt on every keystroke, but haven't investigated it yet

eugenesvk:
should be better now, no cache vs `$XONSH_DIR_PERMA_CACHE = list($PATH)`
(think it bugs with the EnvPath type as it has no len, but then it's not really meant to  be used like that)

https://github.com/user-attachments/assets/032a8a4d-fbc7-486c-92c9-1aafecdbafd0

Spitfire1900:
That did the trick. On first launch it hung for ~5-7 seconds as it built the cache (presumably) then responded well. Then I re-launched and the hang did not reproduce.

`.xonshrc`
```xonsh
$XONSH_DIR_PERMA_CACHE = list($PATH)
```

At least one of these folders is a network share

```python
@
total_exes = 0
for k, v in pc.get_cache_db('p').items():
    total_exes += len(v)
    print(f'{len(v)=}')
total_folders = len(pc.get_cache_db('p'))
print(f'{total_exes=}')
print(f'{total_folders=}')
...
len(v)=262
len(v)=54
len(v)=18
len(v)=3
len(v)=2
len(v)=35
len(v)=0
len(v)=0
len(v)=3
len(v)=2
len(v)=0
len(v)=0
len(v)=40
len(v)=687
len(v)=14
len(v)=9
len(v)=2
len(v)=9
len(v)=9
len(v)=1
len(v)=1
len(v)=6
len(v)=6
len(v)=6
len(v)=6
len(v)=6
len(v)=0
len(v)=108
len(v)=99
len(v)=15
len(v)=1
len(v)=687
len(v)=14
len(v)=9
len(v)=2
len(v)=9
len(v)=91
len(v)=0
len(v)=2
len(v)=2
len(v)=6
len(v)=16
len(v)=1
len(v)=38
len(v)=8
len(v)=1
len(v)=18
len(v)=58
len(v)=2
len(v)=1
len(v)=7
len(v)=8
len(v)=14
len(v)=1
len(v)=3
total_exes=2402
total_folders=55
```

Spitfire1900:
I'll be using this in lieu of https://github.com/xonsh/xonsh/pull/5677 for awhile.

What is the recommended way to refresh the PERM cache, manually delete the pickle file?

eugenesvk:
15 sec is a bit high, but with a slow/large network share seems ok (perma caches do check each file for the existence + executable suffix)?

- You can delete the pickle file directly or
- run this `pc` method with an extra flag to delete both perma and listed caches
```
    def reset(cls, delfiles: bool = False):
        """Clean PathCache to allow creating a new one with a different env
        delfiles: Also delete cached files
        """
```
- cahe also resets if your `pathext` var changes since then the definition of what a command is is different vs what's cached, though that should be optimized on pathext getting "smaller" by removing the extra .exts instead of removing all the cache... so shouldn't be relied on to track changes

But note that perma cache was meant to be... permanent :) Maybe in addition to non-changing dirs like `C:Windows/`you could add **frequently** changing **large** dirs where listing them on changes might be noticeably bad?

But otherwise have you tried moving non-perma dirs to the `XONSH_DIR_CACHE_TO_LIST` cache (with or without `XONSH_DIR_CACHE_LIST_NON_EXE` depending on how much you care about highlight precision)?

It can be pickled on xonsh exit to avoid startup costs with `XONSH_DIR_CACHE_LIST_SAVE_ON_EXIT`, so for rarely changed dirs or smaller dirs that should work?

Spitfire1900:
I'll see about leveraging other env vars but at the moment `$XONSH_DIR_PERMA_CACHE = list($PATH)` is the only thing I'm going to have set, then just delete the pickle file once a quarter or so. I may explore the other options once it bites me.

In practice I think that's going to be used most often.

Spitfire1900:
> > That did the trick. On first launch it hung for ~15 seconds
> 
> wonder, is there an easy way to do that work when the user isn't typing anything at the prompt instead of blocking the UI for so long?

Delegate auto-completion related tasks off of the main thread.

eugenesvk:
Btw, I've added a `help_me_choose` method to offer suggestions on which cache dirs from `$PATH` could be added.

`from xonsh.procs.executables import PathCache; pc = PathCache; pc.help_me_choose()`

eugenesvk:
and moved cache updates off the main thread, so hopefully no more huge delays on startup

Spitfire1900:
> and moved cache updates off the main thread, so hopefully no more huge delays on startup

This change worked wonders :) 

On the aside, it looks like $PATH entries on a network drive causes `pc.help_me_choose()` to hang when interrogating that entry
```xsh
14·ext   list_exe   list_all  # files  # execs   mtime     CacheIn    Cached?
...
 .   6    .   2    .   1           3 =     3  2024-01-18        A   ✓ P      C:\Program Files\Python312\Scripts
 .   6    .   6    .   1           8       2  2024-01-18      L     ✓ P      C:\Program Files\Python312
### next entry is a Network drive, hangs here.
```

eugenesvk:
It tries to check every path 10 times using various methods unless the 1st list results in many items, but not sure this has any benefit tbh, so might change all to a single list for all dirs. Though in theory this should just take a lot of time, not hang?

You could check by calling `def help_me_choose(cls, iters: int = 10):` with iters set to `1`? This should still do a few per-every-file calls to assess various methods, but hopefully should complete in a more reasonable time

Spitfire1900:
It ended up just taking a long time

```xsh
🐚 N:\NetShare\CmdUtils\bin     24-12-21 12:44:34-0
@ from datetime import datetime; print(datetime.now()); from xonsh.procs.executables import PathCache; pc = PathCache; pc.help_me_choose(); print(datetime.now())
2024-12-21 12:45:23.248089
...
14·ext   list_exe   list_all  # files  # execs   mtime     CacheIn    Cached?
 .   6    . 256    .   3         397     264  2023-08-22    S?      ✓ P      C:\Program Files\Git\usr\bin
 .   6    . 108    .   2         161      56  2023-08-22    S?      ✓ P      C:\Program Files\Git\mingw64\bin
 .   6    .   2    .   1           3 =     3  2024-11-21        A   ✓ P      C:\Program Files\Python313\Scripts
 .   6    .   6    .   1           8       2  2024-11-21      L     ✓ P      C:\Program Files\Python313
 .  19   12.0333    .   6          55      40  2020-06-30    S?      ✓ P     N:\NetShare\CmdUtils\bin # Netshare on PATH enforced by GPO
...
 .   6    .  12    .   1          17       2  2023-08-22      L?    ✓ P      C:\Program Files (x86)\Notepad++
 .  19   11.9524    .   4          55      40  2020-06-30    S?      ✓ P      .
2024-12-21 12:49:53.606617
```

eugenesvk:
Do you know how to detect `N:\NetShare\CmdUtils\bin` as a network drive?

Spitfire1900:
> and moved cache updates off the main thread, so hopefully no more huge delays on startup

This brings up a higher level discussion, should the entire completers system in Xonsh be moved off of the main thread and is that feasible?

eugenesvk:
That's the tricky part I mentioned ealier. There was an attempt for pyghooks syntax highlighting https://github.com/xonsh/xonsh/pull/4328/files following https://github.com/xonsh/xonsh/issues/2578, but that stumbled since threads aren't interruptable (though there was a sensible suggestion to use just 1 and if it's unavailable, do nothing?)

But otherwise haven't looked into it, not sure to what extent it's feasible and whether there are any bigger issues

would be nice if you never got ui-blocked on typing...

Spitfire1900:
> Do you know how to detect `N:\NetShare\CmdUtils\bin` as a network drive?

I used https://discourse.mcneel.com/t/system-io-is-missing-driveinfo-in-python2-rhino8/175458 as a reference.

```python
# pip install pythonnet
import clr
clr.AddReference('System.Management')  # Add reference for WMI access
from System.Management import ManagementObjectSearcher

def get_drive_type(type_number):
    types = {
        0: "Unknown",
        1: "No Root Directory",
        2: "Removable Disk",
        3: "Local Disk",
        4: "Network Drive",
        5: "Compact Disc",
        6: "RAM Disk"
    }
    return types.get(int(type_number), "Unknown")


searcher = ManagementObjectSearcher('SELECT * FROM Win32_LogicalDisk')

for disk in searcher.Get():
    print(f'{disk['DeviceID']=}, type: {get_drive_type(disk['DriveType'])=}')

```
Outputs:
```
disk['DeviceID']='C:', type: get_drive_type(disk['DriveType'])='Local Disk'
disk['DeviceID']='N:', type: get_drive_type(disk['DriveType'])='Network Drive'
```

eugenesvk:
I've meanwhile excluded dirs that take > 5 secs from doing the iterations multiple times similar to > 1000 items dirs

Spitfire1900:
A network drive test is only useful to present to the user that you're working on benching the entry and not stuck. Though I think it's generally better to just tell the user that some entries may take a long time to benchmark if they're network drives.

Spitfire1900:
So I really like this feature and this branch is my primary installation on my workstation, but on my personal machine I don't have a reason to adopt these features cause I don't have a performance ruining on-access virus scanner installed.

It was mentioned much earlier that there were serious concerns preventing this from being pulled into main, perhaps this is better left as a completers xontrib than in the main project?

eugenesvk:
What's you video of a held key like on a personal machine with/without cache? For me xonsh has always lagged, those hundreds of IO calls do accumulate on fast typing/deletes if it's all sub-second.

But even outside of perf the new prefixed highlight is great for catching typos early, so I'm using this on non-Win as well.

This mods the core locate executable/file functions, so no idea how to do this in a xontrib

eugenesvk:
> Linux box and I could not detect lag.

that's because only Windows does an order of magnitude more checks due to pathext... (note sure, but maybe also Wins have more dirs in paths?)

Yes, the underline, try to type `anit` and see it disapper

Spitfire1900:
With the change to make caching happen off of the main thread I've changed my configuration to use the session cache instead of the permanent cache, `$XONSH_DIR_SESSION_CACHE = list($PATH)`.

Spitfire1900:
Oh man, GPO is enforcing both `'C:\\windows\\system32'` and `'C:\\WINDOWS\\system32'` on my PATH, resulting in two cache entries for it. 

While generally Windows is not case sensitive, the fact that this is toggle-able at the directory level means that the caching code is handling this how it should, despite the double-entry.

https://learn.microsoft.com/en-us/windows/wsl/case-sensitivity

Spitfire1900:
> that's because only Windows does an order of magnitude more checks due to pathext... (note sure, but maybe also Wins have more dirs in paths?)

I have 4604 files in System32, of which 669 are considered executable, I wanted to say that was the reason, alas Linux has many more.

Cumulatively though on my Windows install amongst 56 folders I have 12,403 total executables

Linux: 
```
1·ext   list_exe   list_all  # files  # execs   mtime     CacheIn    Cached?
 .        .   2    .              15 =    15  2024-12-19        A    ✗       /home/kyle/.cargo/bin
 .        .   5    .   2          21 =    21  2024-11-07        A    ✗       /home/kyle/.local/bin
 .        .        .               0 =     0  2015-10-21        A    ✗       /usr/local/sbin
 .        .        .               0 =     0  2024-03-02        A    ✗       /usr/local/bin
 .        .  73    .   5         650 =   650  2024-12-17  P?    A    ✗       /usr/sbin
 .        . 341    .  30        3058 =  3058  2024-12-23  P?    A    ✗       /usr/bin
 .        .  70    .   5         650 =   650  2024-12-17  P?    A    ✗       /sbin
 .        . 318    .  27        3058 =  3058  2024-12-23  P?    A    ✗       /bin
 .        .   1    .              14 =    14  2024-11-05        A    ✗       /usr/games
 .        .        .               0 =     0  2015-10-21        A    ✗       /usr/local/games
 .        .   1    .               8       7  2024-12-23             ✗       /snap/bin
 .        .        .               1 =     1  2023-09-21        A    ✗       //home/kyle/.pyenv/bin
```

eugenesvk:
The number of files doesn't matter for the original lag, only 
- pathext * dirs in path

because that's what's being checked, each possible variant you type, the dirs aren't listed

And you have 1 pathext on nix

For caching it matters, but then only for the initial/update on change (and the huge Windows system 32 is recommended to be in permacache, so it doesn't even update

eugenesvk:
I thought there was some normalization/deduping going on as part of path cleanup routines, but I guess not

In general,  think case sensitivity should be enabled separately as it's too rare to have on by default

Spitfire1900:
I have a new xontrib created in the simliar vein as this, [xontrib-pygitstatus](https://pypi.org/project/xontrib-pygitstatus/) which eliminates calls to git to populate the PROMPT, using PyGit2 instead.

Spitfire1900:
This is in a pretty good state at this point and is my daily runner on Windows, I've not needed on Linux.

@anki-code may merging this in be re-visited?

anki-code:
Hello all! I'm doing significant effort to improve xonsh core in 6159. Next thing will be Windows.
Can anybody describe step by step with links and examples how to setup xonsh on Windows to reproduce lags/slow downs and glitches?

I installed xonsh on Windows 11 using python binaries and PyCharm and I see no lags during work. No significant slow downs. So how I can reproduce this?

@eugenesvk @Spitfire1900

Spitfire1900:
The best way to reproduce slow downs is to make the current working directory a folder on a network drive. Also put a bunch of bogus exes in a network drive and add that folder to PATH.

Spitfire1900:
The more files the better.

anki-code:
@Spitfire1900 Why it's needed to have network drive in PATH?

Spitfire1900:
xonsh iterates over all folders in PATH to find executables. If one of those folders is on a network drive the iteration operation is slow.

My org used to but no longer did this, where a network drive location was in the System path.

The observation is that typing feedback slows considerably especially on first character in a line when doing this.

anki-code:
@Spitfire1900 thanks, I know why. So we will search the fix.

Spitfire1900:
Originally the primary root problem for me was a very aggressive, corporate anti-malware program that  would block on all file opens until the virus scanner finished scanning the file making any Python program very slow to start, cause pip builds to fail because it would time out on temporary workspace deletion (the scanner wanted to scan the file before you could delete it 🤦‍♂️)and cause XONSH to be slow to type unless I disabled executable highlighting.

anki-code:
@Spitfire1900 I'm in implementing commands cache improvements. I hope we will have it near release.

Spitfire1900:
Sounds good. Similar issues was also the reason for why I made xontrib-pygitstatus. No process had to be forked off to get current git status, resulting in faster execution in that corporate environment.

anki-code:

eugenesvk:
This has the exact same issue as the other reopened PR: the reference PR is irrelevant and doesn't address the issue fixed here

Spitfire1900:
@eugenesvk is `XONSH_COMMANDS_CACHE_READ_DIR_ONCE`added with 2835fa6e55fcb7757ff60468fcb417e0cead645e insufficient for your use case?

eugenesvk:
insufficient, it simultaneously goes too far: only offers a speed improvement with a loss precision

but also doesn't go far enough: wastes time caching every session what doesn't change (at all or often) between sessions

So if your slow network drive takes 40sec to get cached, in every new shell session you'll not have accurate highlighting for 40sec, and it will be harder to notice because it's masked beyond async

anki-code:
Outdated

eugenesvk:

Spitfire1900:
@eugenesvk I'd argue having so many executables the startup impact is a concern or having valid executables added after startup that could benefit from short-lived caches that are not session long are out-of-scope.

eugenesvk:
So what's your argument given that any default Windows installation has a lot of executables?

Or why wouldn't you want a newly created user executable from showing up in your shell right away without a session restart?

Spitfire1900:
I use xonsh on Windows as my primary work environment, xonsh is way better at handling this case on the latest feature version than last year.

PR #5661, opened 2024-08-12
refactor: Disable double check is_file on Windows systems to speedup IO

anki-code:
### Motivation

Windows has slow IO operation on is_file check. So this PR is to reduce running is_file two times.

## For community

### DIFF
--- a/xonsh/procs/executables.py
+++ b/xonsh/procs/executables.py
@@ -122,7 +122,8 @@ def locate_relative_path(name, env=None, check_executable=False, use_pathext=Fal
             filepath = p.parent / possible_name
             try:
                 if not is_file(filepath) or (
-                    check_executable and not is_executable(filepath, check_file_exist=False)
+                    check_executable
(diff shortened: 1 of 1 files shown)


xonsh/procs/executables.py
```diff
@@ -57,10 +74,20 @@ def is_executable_in_windows(filepath, env=None):
...
+
+    Parameters
+    ----------
+    filepath : str
+        Path to file.
+    check_file_exist : bool
+        If ``False`` do not check that file exists. This made to consistency with ``is_executable_in_windows``.
+    """
     try:
-        return filepath.is_file() and os.access(filepath, os.X_OK)
```
jnoortheen, line 63: it will be enough to check the access? why do we need to check is_file and access?

jnoortheen: Ok it seems, the dir can have X_OK

jnoortheen (approved):

jnoortheen:
Thanks @anki-code ! lets see how the changelog gets generated

PR #5664, opened 2024-08-13
Replace zero to nothing in --timings table

eugenesvk:
Use contiguous table separators

Remove empty `0.000` as those just obscure changes

Before
```
|-------------------------|-----------|-----------|
|Event name               | Time (s)  | Delta (s) |
|-------------------------|-----------|-----------|
|on_post_prompt           |   0.000   |   0.000   |
|on_pre_prompt_format     |   0.000   |   0.000   |
|on_pre_prompt_tokenize   |   0.000   |   0.000   |
|on_post_prompt_tokenize  |   0.001   |   0.001   |
|on_pre_prompt_style      |   0.001   |   0.000   |
|on_post_prompt_style     |   0.003   |   0.002   |
|on_pre_prompt            |   0.003   |   0.000   |
|-------------------------|-----------|-----------|
```

After (github doesn't connect vertical box lines for some reason, but it's contiguous in the terminal)
```
┌─────────────────────────┬───────────┬───────────┐
│Event name               │ Time (s)  │ Delta (s) │
├─────────────────────────┼───────────┼───────────┤
│on_post_prompt           │           │           │
│on_pre_prompt_format     │           │           │
│on_pre_prompt_tokenize   │           │           │
│on_post_prompt_tokenize  │   0.001   │   0.001   │
│on_pre_prompt_style      │   0.001   │           │
│on_post_prompt_style     │   0.003   │   0.002   │
│on_pre_prompt            │   0.003   │           │
└─────────────────────────┴───────────┴───────────┘
```

An even better design would be to remove non-significant 0s altogether - like here it's immediately clear there is only a single operation >1s since there is only a single number in that column
```
┌─────────────────────────┬───────────┬───────────┐
│Event name               │ Time (s)  │ Delta (s) │
├─────────────────────────┼───────────┼───────────┤
│on_post_prompt           │           │           │
│on_pre_prompt_format     │           │           │
│on_pre_prompt_tokenize   │           │           │
│on_post_prompt_tokenize  │    .  1   │    .  1   │
│on_pre_prompt_style      │    .  1   │           │
│on_post_prompt_style     │    .  3   │    .  2   │
│on_pre_prompt            │   1.003   │           │
└─────────────────────────┴───────────┴───────────┘
```

1. Include a news file with your PR (https://xon.sh/devguide.html#changelog).
2. docs: no docs, a tiny formatting change
3. usage example: no difference
4. issue: without an issue

## For community

### DIFF
--- a/xonsh/timings.py
+++ b/xonsh/timings.py
@@ -338,9 +338,11 @@ def setup_timings(argv):
             prevtime = tstart = times[0][1]
             for name, ts in times:
                 print(
-                    entry_format.format(name, ts - tstart, ts - prevtime).replace(
-                        "0.000", "     "
-                    )
+                    entry_format.format(name, ts - tstart, ts - prevtime)
+                    .replace("0.000", "     ")
+                    .replace("0.00", " .  ")
+                    .replace("0.0", " . ")
+                    .replace("0.", " .")
                 )
                 prevtime = ts
             print(endline)
--- a/xonsh/timings.py
+++ b/xonsh/timings.py
@@ -325,16 +325,18 @@ def setup_timings(argv):
             times = list(_timings.items())
             times = sorted(times, key=lambda x: x[1])
             width = max(len(s) for s, _ in times) + 2
-            header_format = f"|{{:<{width}}}|{{:^11}}|{{:^11}}|"
-            entry_format = f"|{{:<{width}}}|{{:^11.3f}}|{{:^11.3f}}|"
-            sepline = "|{}|{}|{}|".format("-" * width, "-" * 11, "-" * 11)
+            begline = "┌{}┬{}┬{}┐".format("─" * width, "─" * 11, "─" * 11)
+            header_format = f"│{{:<{width}}}│{{:^11}}│{{:^11}}│"
+            entry_format = f"│{{:<{width}}}│{{:^11.3f}}│{{:^11.3f}}│"
+            sepline = "├{}┼{}┼{}┤".format("─" * width, "─" * 11, "─" * 11)
+            endline = "└{}┴{}┴{}┘".format("─" * width, "─" * 11, "─" * 11)
             # Print result table
             print(" Debug level: {}".format(os.getenv("XONSH_DEBUG", "Off")))
-            print(sepline)
+            print(begline)
             print(header_format.format("Event name", "Time (s)", "Delta (s)"))
             print(sepline)
             prevtime = tstart = times[0][1]
             for name, ts in times:
-                print(entry_format.format(name, ts - tstart, ts - prevtime))
+                print(entry_format.format(name, ts - tstart, ts - prevtime).replace('0.000','     '))
                 prevtime = ts
-            print(sepline)
+            print(endline)
--- a/xonsh/timings.py
+++ b/xonsh/timings.py
@@ -337,6 +337,10 @@ def setup_timings(argv):
             print(sepline)
             prevtime = tstart = times[0][1]
             for name, ts in times:
-                print(entry_format.format(name, ts - tstart, ts - prevtime).replace('0.000','     '))
+                print(
+                    entry_format.format(name, ts - tstart, ts - prevtime).replace(
+                        "0.000", "     "
+                    )
+                )
                 prevtime = ts
             print(endline)
--- a/news/timing_table.rst
+++ b/news/timing_table.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
+**Changed:**
+
+* ``--timings`` table format to use contiguous box drawing symbols and remove empty 0.000
+
+**Deprecated:**
+
+* <news item>
+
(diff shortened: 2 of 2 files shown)


eugenesvk:
What confuses you to see nothing if there is nothing?

eugenesvk:
it's literary a symbol for nothing: from a slightly more reliable source on the meaning of things than Quora, which has gems of absurdity like  "And, of course, nothing is itself not nothing!" in the top link (though it also has the correct "Zero is the mathematical description of nothing.")

"Zero (0) is a [special](https://simple.wiktionary.org/wiki/special) [number](https://simple.m.wikipedia.org/wiki/Number).[[1]](https://simple.m.wikipedia.org/wiki/Zero#cite_note-1) If there are zero things, then there is nothing at all.

eugenesvk:
You're just making up your own definition that nothing means no measurement. Look it up in a dictionary and cite a reference where nothing means this

This is just patenly false, if there is nothing on the table, theare are no apples there, or 0 apples, not that we didn't measure (we did, with our eyes)

Also, `None` is a pythonic construct, you're just confusing yourself more by using it, this table has nothing to do with `None`, it's not about Python, but a very basic measurement of time and useless values of `0.000` obscuring other numbers so it's harder to read them

eugenesvk:
> It's pretty common knowledge in physics 

It's a mistake, not knowledge, let alone common. Care to cite any specific physics definiton that says you can't say "nothing" of 0?

> and programming ... None, NULL, NaN

Which is an irrelevant domain. In a programming language zero can have a sign, in math it doesn't

> The white space in scientific articles means "no data" as well.

There is no data, it's all hidden from view, why don't you use `5e-16` instead of `0.000` if you want the precision of science?

> I feel here the "design thinking" - hide everything that has "no meaning for eyes". 

You'd benefit from a bit of that if you feel poorly readable noise of 0s is beautiful and clarity is ugly

> What I really want is to have a heatmap:

Why doesn't your table have `0.000` everywhere? But also, instead of the imaginary complexity of 40 numbers in 4 columns open your real terminal with this PR to see only **2** significant numbers and think how useful it'd be to add colors those

The startup column is bigger, but it also only has ~**3** significant numbers, and you could make `.700` stand out vs `.001 with a bit fewer 0s, but since you find the clarity of space ugly you'll need heat colors for that

```
┌─────────────────────────┬───────────┐
│Event name               │ Delta (s) │
├─────────────────────────┼───────────┤
│start                    │           │
│pre_execer_init          │    . 23   │
│post_execer_init         │    . 20   │
│pre_xonsh_session_load   │           │
│post_xonsh_session_load  │    .333   │
│on_pre_rc                │           │
│on_post_rc               │           │
│on_ptk_create            │    . 99   │
│on_post_init             │           │
│on_pre_cmdloop           │    .  2   │
│on_pre_prompt_format     │           │
│on_pre_prompt_tokenize   │    .111   │
│on_post_prompt_tokenize  │    .222   │
│on_pre_prompt_style      │    . 60   │
│on_post_prompt_style     │    .  4   │
│on_pre_prompt            │           │
└─────────────────────────┴───────────┘
```

jnoortheen:
Actually, 0 is not exactly 0 here. Instead, it can show more denomination (like 0.0001 or 1ms)

eugenesvk:
And why would you want even higher resolution? What question would that help answer?

jnoortheen:
to differentiate between all the zeroes. If there is a need for change to the table, then this would be more informative.

eugenesvk:
It's more noise, not information. For info you'd actually need to know what you'll use it for, and in this case nanoseconds will give you nothing.

And this PR is about reducing the number of 0s you need to differentiate between, while you're doing the opposite while claiming differentiation is an issue?

xonsh/timings.py
```diff
@@ -325,16 +325,24 @@ def timing_on_pre_prompt(**kw):
...
-            print(sepline)
+            print(begline)
             print(header_format.format("Event name", "Time (s)", "Delta (s)"))
             print(sepline)
             prevtime = tstart = times[0][1]
             for name, ts in times:
-                print(entry_format.format(name, ts - tstart, ts - prevtime))
+                print(
+                    entry_format.format(name, ts - tstart, ts - prevtime)
+                    .replace("0.000", "     ")
```
jnoortheen, line 342: this is obscuring information.

eugenesvk: What information is obscured? That there is nothing? Space highlights nothingness just fine!

jnoortheen:
I don't see other option than closing it as I want to cleanup the open PRs. We can revisit, if there is more user interest.

eugenesvk:
The other option is to leave the PR open to gauge user interest since closed prs/issues are even less visible to the users

eugenesvk:
But you've been doing most of the bikeshedding! So all you had to do for this simple improvement is to just merge it instead of wasting time, especially for a feature you don't even find useful!

PR #5665, opened 2024-08-13
test: split a test to include Windows

eugenesvk:
(moving non-Win commands like ls to a separate test)

was checking why some of the newer commits failed CI tests on non-Windows and to allow for testing on Windows split this failing test that was skipped due to some non-Win commands (`ls`) into two:
  - one with those commands, and
  - another without them (but now working on Win)

so non-Win should have the same tests, but Win gets an extra coverage

(internal test, so no news)

## For community

### DIFF
--- a/tests/shell/test_ptk_highlight.py
+++ b/tests/shell/test_ptk_highlight.py
@@ -46,16 +46,6 @@ def check_token(xsh):
 
 
 _cases = {
-    "ls": {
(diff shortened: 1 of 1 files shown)


jnoortheen (approved):

jnoortheen:
LGTM, Thanks @eugenesvk !

PR #5677, opened 2024-08-30
Disable executable lookup for workstations with ondemand antivirus scanners

Spitfire1900:
I run xonsh in an environment with on-demand scanners, which will pre-scan any file hit by the OS prior to allowing an I/O to be performed against it. This includes the many calls to `filepath.is_file()` on all files in the PATH env var and cwd.

This disables the lookup entirely using the NO_EXECUTABLE_TOKENS env var.

For reference, the mean length of time for `is_executable_in_windows()` on my workstation is 1.1928 ms
```
@ pstats.Stats('C:/TMP/xonsh-longrun.profile').sort_stats(SortKey.CUMULATIVE).print_stats('is_executable_in_windows')
Thu Aug 29 20:45:59 2024    C:/TMP/xonsh-longrun.profile

         7727767 function calls (7621195 primitive calls) in 80.831 seconds

   Ordered by: cumulative time
   List reduced from 4908 to 1 due to restriction <'is_executable_in_windows'>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     7743    0.014    0.000    9.236    0.001 C:\Users\XXXXXX\.local\pipx\venvs\xonsh\Lib\site-packages\xonsh\procs\executables.py:41(is_executable_in_windows)
```

## For community

### DIFF
--- a/xonsh/pyghooks.py
+++ b/xonsh/pyghooks.py
@@ -1596,7 +1596,10 @@ def color_file(file_path: str, path_stat: os.stat_result) -> tuple[_TokenType, s
 
 
 def _command_is_valid(cmd):
-    return (cmd in XSH.aliases or locate_executable(cmd)) and not iskeyword(cmd)
+    if XSH.env.get("NO_EXECUTABLE_TOKENS"):
+        return cmd in XSH.aliases and not iskeyword(cmd)
+    else:
+        return (cmd in XSH.aliases or locate_executable(cmd)) and not iskeyword(cmd)
 
 
 def _command_is_autocd(cmd):


Spitfire1900:
> **Warning!** No news item is found for this PR. If this is an user facing change/feature/fix, please add a news item by copying the format from `news/TEMPLATE.rst`.

This should be fixed with 6af400ef20dcdec4a9d53abada01fa440ff479f8 ✅

Spitfire1900:
I will acknowledge this is a hack to allow Xonsh to stay viable for me and not lag on every keypress. Disabling executable lookup causes some jank in the syntax highlighting. A better solution would be a cache that updates behind-the-scenes on a new prompt.

CHANGELOG.rst
yaxollum, line 1: Hi @Spitfire1900 , the `CHANGELOG.rst` file isn't meant to be manually edited (it gets automatically updated each release from the files in the `news/` folder. Could you please create a news file following [the instructions here](https://xon.sh/devguide.html#changelog)?

Spitfire1900:
> > Disabling executable lookup causes some jank in the syntax highlighting.
> 
> Can you show what is happening?

Here's an example. Python keywords are marked as commands as `git` is not being found since executable lookup is disabled. Normally `with`, `all`, and `12345` would have no syntax highlighting.

image

xonsh/pyghooks.py
```diff
@@ -1596,7 +1596,10 @@ def color_file(file_path: str, path_stat: os.stat_result) -> tuple[_TokenType, s
 
 
 def _command_is_valid(cmd):
-    return (cmd in XSH.aliases or locate_executable(cmd)) and not iskeyword(cmd)
+    if XSH.env.get("XONSH_PROMPT_HIGHLIGHT_EXECUTABLE"):
+        return cmd in XSH.aliases and not iskeyword(cmd)
+    else:
+        return (cmd in XSH.aliases or locate_executable(cmd)) and not iskeyword(cmd)
```
Spitfire1900, line 1602: The performance and recommendation of  `executables.locate_executable()` vs `environ.locate_binary()` needs to be reviewed, as while `environ.locate_binary()` is not recommended `environ.locate_binary()` per:
> NOT RECOMMENDED because ``commands_cache.locate_binary`` contains ``update_cache``
    with scanning all files in ``$PATH``.
    
it does not look like this is actually the case anymore.

eugenesvk:
FYI I've added a bunch of caches to significantly reduce the amount of wasted IO calls per keystroke (consolidated https://github.com/xonsh/xonsh/pull/5660 and other PRs it's based on linked there) so you could configure to permacache some unchanging dirs from PATH and also list smaller dirs (faster than doing dozens of calls for each possible binary extension)

Out of curiosity, is that enough to make the on-access slowdown manageable or still requires a more aggressive approach?

Spitfire1900:
In practice this is superseded by #5660

Spitfire1900:
This use case is no longer relevant on my side, I've finally been able to move from https://github.com/Spitfire1900/xonsh/tree/windows to the main branch as:

1. Our org no longer forces an SMB share to be on `$PATH`.
2. Our current on-access antivirus scanner has better performance than our previous one.

Issue #2975, opened 2018-12-30
Docu that en vars are set in Python mode


halloleo:
By chance I came acroos an old issue (#2521) and @scopatz [mentions there](https://github.com/xonsh/xonsh/issues/2521#issuecomment-343347671_):

> The `@(expr)` operator is only valid in subprocess mode.  I think the confusion here is that environment variables are set and manipulated in Python mode.

I sort of was aware of this, but I think it should be explained explicitly in the doco. - Is this explained in the doco (and i just missed it) or should I add a paragraph about this? 

Furthermore I think it could be worth it to explain, that subprocess mode always starts with a *command* like `command arg1 arg2...`, for example `ls /tmp`. (I hope I am right about this!)

scopatz:
Adding a paragraph about this would be great! Thanks ion advance @halloleo!

halloleo:
Another thing to document (just that I don't forget): 

* Best way to install `pip` is via `pipx`. 
* Then `pipx inject`  gnureadline/prompt_toolkit.

scopatz:
Thanks again!

PR #5714, opened 2024-10-29
Mention that environment variables are set in python mode

Supreeeme:
#2975 
## For community


Supreeeme:
Oh I see. I actually didn't catch this text the first time in the docs:
> You can set (and export) environment variables like you would set any other variable in Python.

I do agree that it's clear enough that environment variables are done in Python mode from this then. I just changed it then to add the examples you posted of combining environment variables. I think that will help ward off some bash brain 😛

PR #5743, opened 2024-12-02
feat(api): Add xonsh.api.jobs module

ading2210:
As suggested on https://github.com/xonsh/xonsh/discussions/5310#discussioncomment-10806014, I have created a new module (`xonsh.api.jobs`) which implements a utility to get the last job. 

There are two functions, `get_last_job()` and `get_last_pid()`. `get_last_job` gets the dict for the latest running job, and `get_last_pid` gets its PID. The functionality is roughly equivalent to `$!` in Bash. 

For example:
```xonsh
from xonsh.api.jobs import get_last_job

sleep 100 & 
job = get_last_job()
print(job)
# {'cmds': [['sleep', '100'], '&'], 'pids': [830943], 'status': 'running', 'obj': <Popen: returncode: None args: ['sleep', '100']>, 'bg': True, 'pipeline': , 'pgrp': 830943, 'started': 1733173797.6885946}
```

```xonsh
from xonsh.api.jobs import get_last_pid

sleep 100 & 
pid = get_last_pid()
print(pid)
# 831090
kill @(pid)
```

## For community


tests/api/test_api_jobs.xsh
```diff
@@ -0,0 +1,22 @@
+from xonsh.api.jobs import get_last_job, get_last_pid
+from xonsh.pytest.tools import ON_WINDOWS
+import pytest
+
+
+$RAISE_SUBPROC_ERROR = True
+$XONSH_SHOW_TRACEBACK = True
+
```
jnoortheen, line 8: kindly use Python for tests.

ading2210: I thought Xonsh for tests would be fine because the tests for all the other `xonsh.api` modules are using Xonsh too.

jnoortheen: That causes some unexpected behaviours and slows down the test suit.

anki-code:
Outdated

PR #5783, opened 2025-01-22
feat(environ): allow XONSH_*_DIR to be configurable

amadaluzia:
## What does this PR do?
This PR changes how the configuration is loaded so that `xonsh_{cache,config,data}_dir` will look for their respective environment variables. If any of them have any value, that will be loaded instead of only `XDG_CONFIG_HOME`.

## Why?

If you have used ZSH in the past then you may think of ZDOTDIR, however it isn't as useful with xonsh as it is already following XDG specification. Though this is also useful when you want to wrap your configuration alongside xonsh and other programs you may want to include. (See how I have done that [here.](https://codeberg.org/amadaluzia/alqueva/src/branch/master/overlays/derivations/xonsh/default.nix))

I can confirm that `--rc-file`, `--no-rc` and `--no-env` still work, though you may ask me for more in-depth testing.

## For community


amadaluzia:
For clarity, I haven't wrote down any documentation other than in the changelog so if you want me to add that then I shall.

amadaluzia:
@anki-code I have never interacted with pytest before, and I am out of ideas for what I could write in the test. Any hints would be appreciated.

amadaluzia:
Cheers! :)

PR #5788, opened 2025-02-03
Implement `history pull` for JSON history backend

jfmontanaro:
Hi, I'm back, like a bad penny. This PR adds support for the `history pull` command to the JSON history backend, and adds a `session-id` parameter to `history pull` so that you can pull history from just a specific session.

My main motivation for the `session-id` parameter is that I would like to at some point make a xontrib that does history syncing based on file-watching (with inotify etc), and with the JSON history backend this should be significantly more performant than having to list the whole history dir and check the mtime of each file.

There are a few questions that should probably be answered before moving forward with this:

### Consecutive duplicate commands are shown in output of `history pull`, but not loaded into prompter history

I noticed that the existing code for `history pull` in the SQLite backend could introduce consecutive duplicate commands like I fixed in my previous PR. I went ahead and fixed that, but I left the duplicate commands in the output of history pull when `--show-commands` is passed.

My thinking was that `history pull` operates on Xonsh history, and Xonsh history permits consecutive duplicate commands. So when we output the list of commands, we should show the duplicates because we did in fact read those from the source. But when we load them into the prompter we should remove the duplicates, because they don't make sense in that particular context.

### Code structure in `xhj_list_history_files`

For pulling with the JSON history backend, I opted to filter history files based on mtime, that is, we only consider files with an mtime later than our `last_pull_time`. This required adding a `modified_since` parameter to `_xhj_history_files`, which made it a little more complex than seemed ideal to me, so I ended up splitting out an inner function that does the listing logic for a single data directory, and then call that in a loop.

However, we are now potentially using the `mtime` of the history files for two separate purposes: 1) to filter, and 2) to sort. Both of these behaviors are optional. The obvious place to handle filtering would be in the inner function, `_xhj_get_data_dir_files`, but that function can't also do the sorting because the sorting needs to work across both data dirs. So the inner function needs to pass the mtimes back to the calling function. But this needs to be optional, so that we can avoid making an extra `stat` call on every file if it isn't necessary. 

To solve this I decided to just add an `include_mtime` parameter to the inner function, and if that is `False` it just returns `None`s instead of the mtimes. This feels like kind of an ugly hack, though, so I'm open to any suggestions of a better way to structure it.

Maybe I'm getting too into the weeds about this and nobody actually cares, but I figured I might as well mention it.

## For community

### DIFF
--- a/docs/tutorial_hist.rst
+++ b/docs/tutorial_hist.rst
@@ -262,6 +262,10 @@ Tries to pull the history from parallel sessions and add to the current session.
 For example if there are two parallel terminal windows the run of ``history pull``
 command from the second terminal window will get the commands from the first terminal.
 
+The optional `--session-id` allows you to specify that history should only be pulled
+from a specific other session. Most useful when using the JSON history backend, as
+the overhead of an unfiltered `pull` can be significantly higher.
+
 ``clear`` action
 ================
 Deletes the history from the current session up until this point. Later commands
--- a/news/json-history-pull.rst
+++ b/news/json-history-pull.rst
@@ -1,6 +1,6 @@
 **Added:**
 
-* Added `--session-id` paramter to `history pull` command.
+* Added and documented `--session-id` paramter for `history pull` command.
 * Implemented `history pull` for JSON history backend.
 
 **Changed:**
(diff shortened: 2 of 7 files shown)


jfmontanaro:
Ok, fixed now I think.

PR #5806, opened 2025-03-02
Fix ignored exception in tracer shutdown related to module deletion

ambv:
With the tracer on, modules at shutdown can be destroyed in an order that prevents the `trace()` function from working. This was currently the case with Python 3.13 where the function was still being called during the `xonsh.tracer` module's destruction.

This change adds a test to prevent this from happening in the future, and works around the issue by creating local references to necessary functions that were otherwise set to None at shutdown.

Without this fix, the added test was failing with:
```py
>       assert proc.stderr == ""
E       assert 'Exception ignored in: <function _removeHandlerRef at 0x103b0a2a0>\nTraceback (most recent call last):\n  File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/logging/__init__.py", line 882, in _removeHandlerRef\n  File "/Users/ambv/Python/xonsh/xonsh/tracer.py", line 88, in trace\nTypeError: \'NoneType\' object is not callable\n' == ''
E
E         + Exception ignored in: <function _removeHandlerRef at 0x103b0a2a0>
E         + Traceback (most recent call last):
E         +   File "/opt/homebrew/Cellar/python@3.13/3.13.2/Frameworks/Python.framework/Versions/3.13/lib/python3.13/logging/__init__.py", line 882, in _removeHandlerRef
E         +   File "/Users/ambv/Python/xonsh/xonsh/tracer.py", line 88, in trace
E         + TypeError: 'NoneType' object is not callable

tests/built_ins/test_tracer.py:31: AssertionError
```

## For community


ambv:
Haha, took a while to get Windows working. While this doesn't matter here in particular, I wanted to see how little of the environment do we actually need to test xonsh as a black box in a subprocess on Windows. This black box test is important since the failure mode here showed up at interpreter shutdown so it would have been impossible to detect with in-process unit tests.

It would have been easier to just copy `os.environ` entirely, but this can make local user's env var settings cause our tests to fail, if incompatible. In fact, there is one test in xonsh that is consistently failing for me since I have `$XONSH_HISTORY_BACKEND='sqlite'` set. But I'll get to that separately.

So for now, we're creating an empty environment save for the Windows-required env vars.

PR #5826, opened 2025-04-05
fix: don't skip history items when running full history pull after session-specific pull

jfmontanaro:
Previously, only a single timestamp was tracked for the most recent history pull. Any pull would reset the timestamp, regardless of which session it was pulling from. This meant that if you ran a pull with a specific session id, then ran a general pull with no session id, you would only get history items from after the session-specific pull, even ones in other sessions (that you didn't pull from in the first place).

Now we track a timestamp for each individually-pulled session id, plus a timestamp for the last general pull.

In theory there's no limit on how large the mapping of session ids to timestmaps can get, but in practice it's unlikely to be a problem since you will typically only be pulling from a few particular ids (i.e. whatever other sessions you currently have open) in a given session.

It looks like we are no longer using `news` to document changes, correct? I tried to follow the conventional-commits standard, let me know if I didn't do it right!

## For community


jfmontanaro:
I see that the tests are failing on Windows, I suspect it's just due to filesystem timing weirdness but I'm having trouble getting the env set up on my Windows machine to test. I have to leave for now but I'll update later when I have some more time to work on this.

jfmontanaro:
Ok, I think I've got the tests consistently passing now.

To explain all the extra calls to `time.sleep(0.032)` that I had to add - apparently on Windows, `time.time()` has only about ~16ms granularity, so if you call it twice in a row you're basically guaranteed to get the same number back. This was causing problems because the tests depend on timestamps for ordering commands, so it was changing the order in which commands came back, and also messing with the detection for "have we already pulled this command" checks.

In reality, it's unlikely that these timestamps will collide even on Windows, since we're talking about commands that are entered interactively, and even repeatedly alternating between up-arrow and Enter, or something like that, is going to take longer than 16ms.

PR #5838, opened 2025-05-04
feat: #5745 - send Ctrl+C event on Windows instead of forceful terminate

### LINKED ISSUE #5745: Pressing Ctrl-C in Windows cancels a subprocess
## Current Behavior

When in a Xonsh interactive session running a subprocess such as ssh, pressing Ctrl-C will cause the subprocess to cancel instead of forwarding the SIGINT to the subprocess.

```
localhost @ ssh remotehost
remotehost $ man grep
...
   Matcher Selection
       -E, --extended-regexp
              Interpret PATTERN as an extended regular expression (ERE, see below).

 Manual page grep(1) line 1 (press h for help or q to quit)
...
<Press Ctrl-C>
 Return 1
<Xonsh exits SSH session with exit code: 1>
```

## Expected Behavior

The SIGINT should be forwarded to the subprocess instead of canceling the subprocess.

## xonfig

<details>

```xsh
@ xonfig
+-----------------------------+---------------------------+
| xonsh                       | 0.19.0                    |
| Python                      | 3.13.1                    |
| PLY                         | 3.11                      |
| have readline               | False                     |
| prompt toolkit              | 3.0.48                    |
| shell type                  | prompt_toolkit            |
| history backend             | sqlite                    |
| pygments                    | 2.18.0                    |
| on posix                    | False                     |
| on linux                    | False                     |
| on darwin                   | False                     |
| on windows                  | True                      |
| on cygwin                   | False                     |
| on msys2                    | False                     |
| is superuser                | False                     |
| default encoding            | utf-8                     |
| xonsh encoding              | utf-8                     |
| encoding errors             | surrogateescape           |
| xontrib 1                   | base16_shell              |
| xontrib 2                   | prompt_bar                |
| xontrib 3                   | pygitstatus               |
| xontrib 4                   | vox                       |
| xontrib 5                   | voxapi                    |
| xontrib 6                   | commands                  |
| RC file 1                   | C:\Users\XXXXXX\.xonshrc |
| UPDATE_OS_ENVIRON           | False                     |
| XONSH_CAPTURE_ALWAYS        | False                     |
| XONSH_SUBPROC_OUTPUT_FORMAT | stream_lines              |
| THREAD_SUBPROCS             | True                      |
| XONSH_CACHE_SCRIPTS         | True                      |
+-----------------------------+---------------------------+
```

</details>

## For community

Spitfire1900:
Resolve #5745. The existing behavior [was introduced in 2015](https://github.com/xonsh/xonsh/commit/3aa37682f9441bb7c8bfb492325c88e0900ed327#diff-75c83180c6250bfa7c142ed214f8c71e7704872f3b0c63b167895cf0ae892989R106-R107) ~~and I suspect was because there was no equivalent to [signal.CTRL_C_EVENT](https://docs.python.org/3/library/signal.html#signal.CTRL_C_EVENT) at the time in Python 2.7, it was introduced in Python 3.2.~~ Xonsh only supported 3.4+ at the time and [signal.CTRL_C_EVENT](https://docs.python.org/3/library/signal.html#signal.CTRL_C_EVENT) was already available; it's unclear why the existing implementation choice was made by @adqm .

# Before behavior:
When pressing Ctrl+C on Windows a forceful terminate was sent using the taskkill command.
```
localhost@ ssh remote
remote$ hello <Press Ctrl+C>
Return 1
localhost@
```
# After behavior:
When pressing Ctrl+C on Windows a [signal.CTRL_C_EVENT](https://docs.python.org/3/library/signal.html#signal.CTRL_C_EVENT) will be sent instead.
```
localhost@ ssh remote
remote$ hello^C
remote$ 
```

## For community

### DIFF
--- a/xonsh/procs/jobs.py
+++ b/xonsh/procs/jobs.py
@@ -214,6 +214,9 @@ if ON_WINDOWS:
             stderr=subprocess.STDOUT,
         )
 
+    def _ctrl_c(job):
+        os.kill(job["obj"].pid, signal.CTRL_C_EVENT)
+
     _hup = _kill  # there is no equivalent of SIGHUP on Windows
 
     def ignore_sigtstp():
@@ -239,10 +242,7 @@ if ON_WINDOWS:
             except subprocess.TimeoutExpired:
                 pass
             except KeyboardInterrupt:
-                try:
-                    _kill(active_task)
-                except subprocess.CalledProcessError:
-                    pass  # ignore error if process closed before we got here
+                _ctrl_c(active_task)
         return wait_for_active_job(last_task=active_task)
 
 else:
--- a/xonsh/procs/jobs.py
+++ b/xonsh/procs/jobs.py
@@ -234,12 +234,8 @@ if ON_WINDOWS:
         proc = active_task["obj"]
         _continue(active_task)
         while proc.returncode is None:
-            try:
+            with contextlib.suppress(subprocess.TimeoutExpired, KeyboardInterrupt):
                 proc.wait(0.01)
-            except subprocess.TimeoutExpired:
-                pass
-            except KeyboardInterrupt:
-                pass  # No proxying of SIGINT required, Ctrl-C is sent along to the child process.
         return wait_for_active_job(last_task=active_task)
 
 else:
--- a/xonsh/procs/jobs.py
+++ b/xonsh/procs/jobs.py
@@ -239,7 +239,7 @@ if ON_WINDOWS:
             except subprocess.TimeoutExpired:
                 pass
             except KeyboardInterrupt:
-                pass # No proxying of SIGINT required, Ctrl-C is sent along to the child process.
+                pass  # No proxying of SIGINT required, Ctrl-C is sent along to the child process.
         return wait_for_active_job(last_task=active_task)
 
 else:
--- a/xonsh/procs/jobs.py
+++ b/xonsh/procs/jobs.py
@@ -214,9 +214,6 @@ if ON_WINDOWS:
             stderr=subprocess.STDOUT,
         )
 
-    def _ctrl_c(job):
-        os.kill(job["obj"].pid, signal.CTRL_C_EVENT)
-
     _hup = _kill  # there is no equivalent of SIGHUP on Windows
 
     def ignore_sigtstp():
@@ -242,7 +239,7 @@ if ON_WINDOWS:
             except subprocess.TimeoutExpired:
                 pass
             except KeyboardInterrupt:
-                _ctrl_c(active_task)
+                pass # No proxying of SIGINT required, Ctrl-C is sent along to the child process.
         return wait_for_active_job(last_task=active_task)
 
 else:


xonsh/procs/jobs.py
```diff
@@ -239,10 +242,7 @@ def wait_for_active_job(last_task=None, backgrounded=False, return_error=False):
             except subprocess.TimeoutExpired:
                 pass
             except KeyboardInterrupt:
-                try:
-                    _kill(active_task)
-                except subprocess.CalledProcessError:
-                    pass  # ignore error if process closed before we got here
+                _ctrl_c(active_task)
```
ambv, line 245: I actually don't think you need to send *anything*. Both the Python REPL and NeoVim (installed by winget) do their CTRL-C handling just fine on their own, tested both in Windows Terminal and in Command Prompt (cmd.exe).

When you do this `_ctrl_c()` here, processes get a duplicate CTRL-C in short succession. In case of Python 3.13 from the Windows Store it sends it into an infinite loop of KeyboardInterrupts.

Just saying `pass` here works fine for me.

Spitfire1900: Validated locally, thanks! 👍

adqm:
> it's unclear why the existing implementation choice was made by @adqm .

Not a whole lot of careful thought went into the specifics.  I am not a Windows user, and that behavior felt roughly right after a few small tests.  It wasn't a carefully-calculated decision by any stretch, just a quick attempt to get some rudimentary job control working for our Windows users.  I'm certainly not opposed to any changes here.

jaraco (dismissed):
I did some basic testing of this change in my Windows environment and it seems to be behaving just fine. I installed it thus:

```
pipx runpip xonsh install git+https://github.com/xonsh/xonsh refs/pull/5838/head
```

I wasn't able to discern any difference in behavior, but I don't have any good test methodology either. I tried using Ctrl+C to terminate a pipx run, but I couldn't tell if it was being forcibly killed.

Perhaps you could suggest a testing methodology (maybe a Python command to run that when terminated would reveal the difference)?

Spitfire1900:
The best testcase is pressing Ctrl+C when in a ssh session where Xonsh is the parent process. When in a an interactive ssh session you have have some data entries into your shell's line buffer that is incorrect. Pressing Ctrl+C should abort the buffer and start a new line,
e.g.
This should do the following 

```shell
$ my inorccetly entered data <Press Ctrl+C>
$ <new line buffer>
```

But instead Xonsh would kill the entire ssh session.

Spitfire1900:
Updated the PR Summary.

ambv:
@jaraco said:
> Perhaps you could suggest a testing methodology (maybe a Python command to run that when terminated would reveal the difference)?

Run a Python REPL in xonsh and press CTRL-C. It should show "KeyboardInterrupt" but not exit. Run neovim in xonsh and press CTRL-C. It should show you how to exit Vim but not exit.

ambv (approved):
The PR in its current state resolves my problems with CTRL-C. Thank you!

Spitfire1900:
Windows docs on this behavior, when Ctrl+C or Ctrl+BREAK is pressed all processes in the console window receive the signal.

gforsyth (approved):

gforsyth:
Thanks for putting this in @Spitfire1900 !  And Thanks to @ambv and @jaraco for testing and reviewing!

PR #5886, opened 2025-07-27
feat: Allow specifying special files as scripts

azagon63:
Allows specifying any existing non-directory path as a script to run. Primarily I intend for this to allow running scripts out of named pipes (FIFOs). Examples here use bash, but the principal is not tied to any shell, see the unit tests for how this can be done in python.
Before:
```console
$ xonsh <(echo "print('Hello')")
xonsh: /dev/fd/63: No such file or directory.
```
After:
```console
$ xonsh <(echo "print('Hello')")
Hello
```

This type of script specification is supported by many interpreters including python and bash:
```console
$ python3 <(echo "print('Hello')")
Hello
$ bash <(echo "echo Hello")
Hello
```

#### Reading the script from stdin

You could accomplish something similar using `echo "print('Hello')" | xonsh` or `xonsh < <(echo "print('Hello')")` however that prevents reading from stdin as part of your script.

#### Use -c and specify the script as an argument

You could also do `xonsh -c "$(echo "print('Hello')")"` but that leaks the contents of the script as an argument, is pretty noisy to look at in process lists, and imposes a maximum size on the script (see [this StackExchange answer](https://unix.stackexchange.com/a/120842) for a good explanation).

## For community

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -7,6 +7,7 @@ import os.path
 import sys
 from contextlib import contextmanager
 from pathlib import Path
+from tempfile import TemporaryFile
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


jnoortheen (approved):

jnoortheen:
Looks good! Thanks @azagon63

PR #5891, opened 2025-08-02
fix: allow numerical tokens in environment variable references

### LINKED ISSUE #5663: Unexpected numerical token
```xsh
echo $123
# Unexpected token: TokenInfo(type=55 (OP), string='$123', start=(1, 7), end=(1, 11), line='![echo $123]\n')

showcmd echo $123
# ['echo', "Unexpected token: TokenInfo(type=55 (OP), string='$123', start=(1, 15), end=(1, 19), line='![showcmd echo $123]\\n')"]

$(echo $123)
# "Unexpected token: TokenInfo(type=55 (OP), string='$123', start=(1, 7), end=(1, 11), line='$(echo $123)\\n')"
```
Here instead of returning error as result we need to do the same as:
```xsh
echo $qwe
# $qwe
echo $123
# $123
```

## For community

PodioSpaz:
Previously, environment variable references like $123 would cause "Unexpected token" errors because the tokenizer only recognized variables starting with alphabetic characters or underscores (using .isidentifier()).

This was inconsistent with alphabetic variables like $qwe, which would gracefully return the literal string "$qwe" when undefined.

The fix changes the tokenization condition from:
  token[1:].isidentifier()
to:
  len(token) > 1 and token[1:].replace('_', '').isalnum()

This allows $123, $abc123, $123abc to be tokenized as DOLLARNAME tokens, ensuring consistent behavior where undefined variables return their literal string representation instead of parser errors.

Fixes #5663

## For community

### DIFF
--- a/xonsh/environ.py
+++ b/xonsh/environ.py
@@ -2311,10 +2311,6 @@ class Env(cabc.MutableMapping):
         self._set_item(key, val)
 
     def _set_item(self, key, val, thread_local=False):
-        if not key.isidentifier():
-            raise Exception(
-                f"Invalid name {repr(key)}. A valid Python identifier is expected."
-            )
         validator = self.get_validator(key)
         converter = self.get_converter(key)
         detyper = self.get_detyper(key)


anki-code:
Looks nice. Thanks!
Before:
```xsh
$123=321
${'123'}
# KeyError: 'Unknown environment variable: $321'
echo $123
# Unexpected token: TokenInfo(type=55 (OP), string='$123', start=(1, 7), end=(1, 11), line='![echo $123]\n')
```
After:
```xsh
$123=321
${'123'}
# 321
echo $123
# 321
env | grep 123
# 123=321
```

anki-code:
Please add test to https://github.com/xonsh/xonsh/blob/main/tests/test_environ.py

I think we may have downstream issues with numerical keys for env but now I see no reasons why we can't support them. cc @jnoortheen @jaraco

jnoortheen:
I don't think other shells support using numbers as environment variables. What could be some reasons to have them?

anki-code:
@jnoortheen just because we can. The main "pros" I think is an ability to put numbers and identifiers to env. I don't know for what but may be it will be fun for some cases e.g.
```xsh
$100 = 'fistful of dollars'
$0xffffff = 'white'
$0s0me_h4Sh = 'I dont know'
$0_0 = 'what?'
$01010101010 = 'binary'
$1, $2, $3 = 'some', 'short', 'cut'
$127_0_0_1 = 'localhost'
```
Yeah...

If you see this is not good for parser we need another PR to fix https://github.com/xonsh/xonsh/issues/5663

anki-code:
Thanks @PodioSpaz for workin on this!

PR #5906, opened 2025-08-27
Update pyproject.toml: Add prompt-toolkit dependent for Windows platform.

wangzhaohan2910:
I just added the prompt-toolkit dependent to xonsh.

I didn't even clone it because I think I will just change one file. So I changed online.

If somewhere must be changed but I didn't do it, then help me to change it, thanks.

## For community


anki-code:
Hi @wangzhaohan2910! Please tell me what issue you address?
As I understand xonsh general way is to have pure installation when using `pip install xonsh` and full installation when using `pip install 'xonsh[full]'` so we don't need to install ptk by default.

wangzhaohan2910:
> Hi @wangzhaohan2910! Please tell me what issue you address? As I understand xonsh general way is to have pure installation when using `pip install xonsh` and full installation when using `pip install 'xonsh[full]'` so we don't need to install ptk by default.

Thanks! That's what I need.

Issue #1180, opened 2016-06-08
Full text of how to download missing program gets cut off


kirbyfan64:
By default on Ubuntu distros (I'm using Elementary OS), when you enter a missing program, the system will hint on how to download it:

```
ryan@DevPC-LX:~$ realpath
The program 'realpath' is currently not installed. You can install it by typing:
sudo apt-get install realpath
ryan@DevPC-LX:~$ 
```

However, this doesn't seem to work fully in Xonsh:

```
ryan@DevPC-LX ~ $ realpath
xonsh: subprocess mode: command not found: realpath
The program 'realpath' is currently not installed. You can install it by typing:
ryan@DevPC-LX ~ $ 
```

scopatz:
Thanks for reporting @kirbyfan64!

PaulReiber:
The command-not-found package is what implements command-not-found processing for bash and zsh for ubuntu.  More info on it here: https://wiki.ubuntu.com/CommandNotFoundMagic

Code tarball, if you're interested: https://launchpad.net/command-not-found/+download

Snippet:
`import sys, os, os.path, gdbm, posix, grp, string, subprocess
from util import gettext_wrapper as _
[...]
                    print >>sys.stderr, _("The program '%s' is currently not installed. ") % command,
                    print >>sys.stderr, _("You can install it by typing:")
                    print >>sys.stderr, cmd
`
The only difference I note is the final line is not getting sent thru gettext_wrapper.

Anthony, I wrote up a test that wrote 3 lines to stderr using print >>sys.stderr, "quoted strings 1 thru 3" and, as I expected, xonsh outputs all 3 lines just fine.

Any ideas?

scopatz:
I am not sure I understand the fix here. Can you please put in a PR?

PaulReiber:
I don't have a fix at this point - I was just sharing the investigation and observations I've done so far.  

This particular bug's an odd one, as such very compelling for me, to better understand things ( how could that possibly happen? why?)

What I've learned is, the difference is that the final line with the command isn't sent through the internationalization function _  but that other than that the recommended commandline is just printed to stderr, as were the lines above it.  The stream gets flushed when the program exits and its stderr stream closes, so that line should come thru... so where did it go?

Are there any other cases of oddness with processing of stderr of a subcommand?

scopatz:
> Are there any other cases of oddness with processing of stderr of a subcommand?

[...]

PR #5915, opened 2025-10-12
fix: remove redundant "command not found"

### LINKED ISSUE #4374: [minor, cosmetic] double error messages for command not found
## xonfig

<details>

```
```

</details>

## Expected Behavior
When running a (pipeline) command that's not found, a simple error message should be generated, e.g. bash does this
```
egnor@ostrich:~$ asfdafd
asfdafd: command not found
```

## Current Behavior
With xonsh I see two generally redundant error prints:
```
$ asfdafd
xonsh: subprocess mode: command not found: asfdafd
asfdafd: command not found
```

## Steps to Reproduce
1. Run a command that doesn't exist
2. Experience a teensy, tiny, itty bitty moment of minor cringe at the duplicate message

## For community

jleen:
Fixes #4374. Some Linux systems show a "command not found" message at the bottom of their /usr/lib/command-not-found output, and Linux systems don't. Previously, xonsh dropped the last line of command-not-found output, to suppress this message. But this led to useful information being dropped on systems that don't include the redundant line. That led to issue #1180. But the fix caused a regression, #4374.

I believe this new logic works with both cases, by simply filtering out any line with the redundant "command not found".

## For community

### DIFF
--- a/xonsh/tools.py
+++ b/xonsh/tools.py
@@ -861,7 +861,11 @@ def debian_command_not_found(cmd):
         stderr=subprocess.STDOUT,
         shell=True,
     )
-    lines = [line for line in s.rstrip().splitlines() if not line.endswith(": command not found")]
+    lines = [
+        line
+        for line in s.rstrip().splitlines()
+        if not line.endswith(": command not found")
+    ]
     s = "\n".join(lines).strip()
     return s
 


anki-code:
Thanks for diving into it.

> Some Linux systems show a "command not found" message at the bottom of their /usr/lib/command-not-found output, and Linux systems don't.

Is it possible do have podman/docker repeatable example?

jleen:
Yes indeed! This will repro in a stock `ubuntu:24.04` docker image. Happy to add more detail if it’s helpful, but the basic steps are just
- `docker pull ubuntu:24.04`
- `docker run -i -t ubuntu:24.04 /bin/bash`

And then from inside the container
- `apt update && apt install command-not-found python3 python3-pip python3-venv git`

and then I was able to clone the xonsh source and confirm the bug and the fix.

anki-code:
> Some Linux systems show a "command not found" message at the bottom of their /usr/lib/command-not-found output, and Linux systems don't.

I mean is it possible to see what Linux distro shows the message and what Linux distro don't in podman/docker.
There is no tests for this PR so we will have a list of commands to repeat the issue at least.

jleen:
> I mean is it possible to see what Linux distro shows the message and what Linux distro don't in podman/docker.
> There is no tests for this PR so we will have a list of commands to repeat the issue at least.

Ah yes, I see your point … I suppose the best we could do for an automated test would be to mock the command output. Probably that wouldn’t be the most valuable automation since the actual logic is pretty trivial and we’d be mocking most of the potentially interesting parts.

Anyway, in terms of manual validation, I poked around a bit more and you can actually repro both of the cases with the `ubuntu:24.04` docker image. After you’ve installed the relevant utility within Ubuntu …

`apt update && apt install -y command-not-found && apt update`

then you can just do `/usr/lib/command-not-found qwertyuiop` which will print the redundant "not found" message, versus `/usr/lib/command-not-found apr-get` which will not. So this should exercise both cases.

anki-code:
Thanks!

PR #5940, opened 2025-11-22
fix: filter escape sequences after decoding in captured stdout and stderr (#5939)

### LINKED ISSUE #5939: Cutting Unicode characters in captured stdout
## Current Behavior

The capturing subprocess `!(@thread ...)` fails to handle certain Unicode characters, and generates two low surrogates. I found this when I'm evaluating a student's submission, and his name causes the problem. It is initially triggered by a glob, even if the problem-causing character is not appearing in xonsh's command. The following is a reproducible example.

```xsh
@ proc = !(@thread echo "左涛/lab1")
@ proc.rtn
0
@ proc.output
'左\udce6\udcb6ab1'
@ proc.raw_out
b'\xe5\xb7\xa6\xe6\xb6\x9b/lab1\n'
@ proc.raw_out.decode("u8")
'左涛/lab1\n'
@ $(echo "左涛/lab1")
'左涛/lab1'
```

It seems `涛/l` becomes `\udce6\udcb6` in the output of `!()`, which are two consecutive low surrogates, while the raw output (in bytes) and explicitly decoded raw output (in str) is correct. Also, it seems the `/` is not important, and replacing it with certain string of ASCII non-characters just get ignored, and the `l` is consumed. If another non-problematic Chinese character follows instead, e.g. `左涛左/lab1`, xonsh will copy it into `proc.output` normally. And if a problematic character follows, e.g. `左涛涛lab1`, xonsh will get `左涛\udce6\udcb6ab1`, leaving the first problematic character unchanged.

By the way, `涛` encodes to UTF-8 as `b'\xe6\xb6\x9b'`, whose first two bytes seem to correspond with the lower 8 bits of the two low surrogates.

Traceback (if applicable):

<details>

```xsh
# Please paste the traceback here.
# No traceback as it is not throwing anything
```

</details>

## Expected Behavior

Handle these Unicode characters correctly and put them into `proc.output`.

## xonfig

<details>

The xonfig of the environment executing the reproducible example:

```xsh
# Please paste the output of the `xonfig` command here.
| xonsh                       | 0.19.9          |
| Python                      | 3.13.9          |
| prompt toolkit              | 3.0.52          |
| shell type                  | prompt_toolkit  |
| history backend             | json            |
| pygments                    | 2.19.2          |
| on linux                    | True            |
| distro                      | unknown         | (it's NixOS actually)
| on wsl                      | False           |
| on darwin                   | False           |
```

My system locale is:
```xsh
@ localectl
System Locale: LANG=en_HK.UTF-8
    VC Keymap: (unset)                
   X11 Layout: us
    X11 Model: pc104
  X11 Options: terminate:ctrl_alt_bksp
```

</details>

## For community

inmaldrerah:
Fixes #5939.

According to console_codes(4) (Linux manual):
> If the console is in UTF-8 mode, then the incoming bytes are first assembled into 16-bit Unicode codes. Otherwise each byte is transformed according to the current mapping table (which translates it to a Unicode value). See the Character Sets section below for discussion.
> 
> In the normal case, the Unicode value is converted to a font index, and this is stored in video memory, so that the corresponding glyph (as found in video ROM) appears on the screen. Note that the use of Unicode (and the design of the PC hardware) allows us to use 512 different glyphs simultaneously.
> 
> If the current Unicode value is a control character, or we are currently processing an escape sequence, the value will treated specially. Instead of being turned into a font index and rendered as a glyph, it may trigger cursor movement or other control functions. See the Linux Console Controls section below for discussion.

So the escape sequences should be stripped from the str, not the bytes. Therefore, this commit filters escape sequences after decoding the raw bytes.

## For community

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -22,7 +22,7 @@ jobs:
       UV_SYSTEM_PYTHON: 1
     name: Xonsh docs to gh-pages
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: Setup Python
         uses: actions/setup-python@v6
         with:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -22,7 +22,7 @@ jobs:
           - "3.12"
           - "3.13"
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: setup-python
         uses: actions/setup-python@v6
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -38,9 +38,10 @@ jobs:
           - "3.11"
           - "3.12"
           - "3.13"
+          - "3.14"
     name: Test Python ${{ matrix.python-version }} ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: Set up Python
         uses: actions/setup-python@v6
         with:


inmaldrerah:
It seems some checks are failing because of prior commits. I don't know how to work around the final variable issues. Maybe the other info messages can be resolved by adding `self: Self` for `None`-returning methods with no extra parameters, or `-> None` for `None`-returning functions taking no parameters.

anki-code:
Thanks for introducing this! Is it possible to have a test?

inmaldrerah:
Tests added. Also, it seems `@thread` is not working on windows from previous tests, and locally I have to use `@thread` to capture the output (even if tests pass online without it). I wonder if this is expected.

inmaldrerah:
Hmmm... It seems I cannot trigger the 3.10 tests.

anki-code:
Yes, waiting for release manager who will take a look.

anki-code:
Thanks @inmaldrerah!

Issue #5954, opened 2025-11-26
Tutorial command not working: pipe into alias


fpusan:
## Current Behavior

Running the `echo "hello" | @(lambda a, s=None: s.read().strip() + " world\n")` from the [documentation](https://xon.sh/tutorial.html#python-evaluation-with) gives an error, instead of the expected result.

Traceback (if applicable):

<details>

```xsh
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK = True
Exception in thread {'cls': 'ProcProxyThread', 'name': 'Thread-7', 'func': <function <lambda> at 0x7f24a9230ea0>, 'alias': '', 'pid': None}
AttributeError: 'NoneType' object has no attribute 'read'
```

</details>

## Expected Behavior
According to the example in the documentation, the command should echo `hello world`.

## xonfig

<details>

```xsh
+-----------------------------+-----------------+
| xonsh                       | 0.19.4          |
| Python                      | 3.13.5          |
| PLY                         | 3.11            |
| have readline               | True            |
| prompt toolkit              | 3.0.51          |
| shell type                  | prompt_toolkit  |
| history backend             | json            |
| pygments                    | 2.18.0          |
| on posix                    | True            |
| on linux                    | True            |
| distro                      | debian          |
| on wsl                      | False           |
| on darwin                   | False           |
| on windows                  | False           |
| on cygwin                   | False           |
| on msys2                    | False           |
| is superuser                | False           |
| default encoding            | utf-8           |
| xonsh encoding              | utf-8           |
| encoding errors             | surrogateescape |
| xontrib                     | []              |
| RC file                     | []              |
| UPDATE_OS_ENVIRON           | False           |
| XONSH_CAPTURE_ALWAYS        | False           |
| XONSH_SUBPROC_OUTPUT_FORMAT | stream_lines    |
| THREAD_SUBPROCS             | True            |
| XONSH_CACHE_SCRIPTS         | True            |
+-----------------------------+-----------------+
```

</details>

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

anki-code:
@fpusan thanks! Could you please fix it here - https://github.com/xonsh/xonsh/blob/e1dce1821fc57e1b80dfff59b31cf56552f03201/docs/tutorial.rst?plain=1#L620

The right code for current xonsh version:
```xsh
echo "hello" | @(lambda args, stdin=None: stdin.read().strip() + " world\n")
```

PR #5955, opened 2025-11-27
fix: Update tutorial example for piping into an alias

fpusan:
Update tutorial command for piping into an alias, to make it work in the current xonsh version, see https://github.com/xonsh/xonsh/issues/5954

## For community


anki-code:
Thanks @fpusan !

fpusan:
thanks to you guys for the great tool!

Issue #5802, opened 2025-02-26
Python builtins shadowing


anki-code:
I think we need to consider the issues around shadowing and find the way to solve it may be on parser side.

### Examples
* functions
  * `id` (#3508)
  * `zip`
* key words
  * `import` (#3612)

### Proposed solutions
* If user type function name and it has the subprocess tool with the same name - use subprocess mode. Example: `id`, `zip` will be called as subprocess commands.

### Workaround

Use uppercase to call variable name that doesn't exist in python and will be translated to subprocess mode:

```xsh
ID
# success
Id
# success
```

All builtins:

```
Here are the Python built-in functions (Python 3.x), in alphabetical order:

    abs
    aiter
    all
    anext
    any
    ascii
    bin
    breakpoint
    bytearray
    bytes
    callable
    chr
    classmethod
    compile
    complex
    delattr
    dict
    dir
    divmod
    enumerate
    eval
    exec
    filter
    float
    format
    frozenset
    getattr
    globals
    hasattr
    hash
    help
    hex
    id
    input
    int
    isinstance
    issubclass
    iter
    len
    list
    locals
    map
    max
    memoryview
    min
    next
    object
    oct
    open
    ord
    pow
    print
    property
    range
    repr
    reversed
    round
    set
    setattr
    slice
    sorted
    staticmethod
    str
    sum
    super
    tuple
    type
    vars
    zip
    import


Built-in functions with known CLI name overlaps:

    ascii — ascii: prints ASCII table (available via ascii(1) on some Linux distros)
    dict — dict: DICT protocol client (dictd package)
    dir — dir: directory listing (GNU coreutils); dir: Windows CMD command; PowerShell alias to Get-ChildItem
    format — format: disk formatting utility (Windows)
    hash — hash: shell builtin (bash/zsh/ksh)
    help — help: shell builtin (bash/zsh/fish); Windows CMD help; PowerShell help via Get-Help
    id — id: show user/group IDs (GNU coreutils)
    next — next: Next.js CLI (installed via npm)
    open — open: open files/URLs (macOS)
    set — set: shell builtin (POSIX shells) and Windows CMD
    sum — sum: checksum utility (historical, still present on many Unix systems)
    type — type: shell builtin (bash/zsh) describing command resolution; type: Windows CMD prints file contents
    zip — zip: create ZIP archives (Info-ZIP)

Also relevant (keyword, not a builtin function):

[...]

PR #5956, opened 2025-11-27
feat!: Added `@` as a global object that provides instant functionality

### LINKED ISSUE #5411: Add `@` as alternative to `__xonsh__`

shevtsova:
Resolves #5411 ([motivation](https://github.com/xonsh/xonsh/issues/5411#issue-2293423821)) and reduces `${...}` mentions in the tutorial.

```xsh
@
# <xonsh.built_ins.XonshSessionInterface>

@?
# Description

@.imp.json.loads('{}')
# {}

@.env  # as alternative to `{...}`

with @.env.swap(CONCH='snail'): $[echo $CONCH]
# snail

@.lastcmd.rtn
# 0

@aliases.register
@@.imp.xonsh.tools.unthreadable
def _mycmd(args, stdin=None):
    return 'ok'
mycmd
# ok
```

## For community

### DIFF
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -147,7 +147,7 @@ print(isinstance(@, type(__xonsh__.interface)))
     ),
     (
         """
-$CONCH=24        
+$CONCH=24
 with @.env.swap(CONCH=42):
     print(@.imp.json.loads('{"@":"~"}'), $CONCH)
 """,


anki-code:
Hey @shevtsova ! It's very nice thing, thanks! Could you please add this to the docs as well?

anki-code:
@jnoortheen LGTM, what do you think?

jnoortheen:
As one comment has pointed out, promoting this would make the shell more cryptic. I had be happy if this was kept as an option for part of the users who feel it is necessary to have single characters. Lets have some more discussion before pushing it. Also I don't prefer replacing all usage of __xonsh__ in the documentation with `@`; Instead just mention it as an option. Also note that the need to use __xonsh__ should be minimal.

anki-code:
@jnoortheen first of all let's read [motivation part](https://github.com/xonsh/xonsh/issues/5411#issue-2293423821) that has big amount of points.

> I don't prefer replacing all usage of xonsh in the documentation with @; Instead just mention it as an option. 

@shevtsova could you please keep both ways?

shevtsova:
@jnoortheen I've added both ways.

anki-code:
@jnoortheen I think it's ready to come in.

jnoortheen:
@anki-code I want some more discussion from people. @gforsyth what do you think ?

shevtsova:
Personally, after `__xonsh__.imp` was introduced, I found myself reaching for it in the shell whenever I needed a function, so having a one-character shortcut really helps me reduce keystrokes.

gforsyth:
I understand wanting a single-key shortcut for accessing the `__xonsh__` session object.  I have a few thoughts:

1. `__xonsh__` is supposed to be internals.  Now, that ship has probably already sailed, but this is making it a documented and exposed part of the API.  @jaraco noted that there were no docs on the `__xonsh__` object.  There probably _should_ be docs on it, but also it wasn't created as a "featureful" object.

It's ok to make it a public object, just recognize that there are consequences if we do that.

2. I'm concerned that `@` is overloaded. We currently use it as:
    - the default prompt symbol
    - forcing python-mode parsing via `@()`
    - explicit threading commands

and additionally _everything that Python itself uses it for_.

I share the thoughts that it's making the shell look more arcane, like bash.  Full disclosure, I've contributed my fair share of arcane behavior to xonsh.

3. The fact that adding this requires a parser rule is a mixed bag.  Imagine that we chose the letter `x` as the `__xonsh__` shortcut instead, there are pros and cons:

    - It can be implemented as `x = __xonsh__`, easy to override or change (pro, to me)
    - Possibility of collisions with existing aliases or executables (con)

Now, with `@`, users cannot add it themselves, because `@ = __xonsh__` isn't valid Python syntax. So there's a plus in that a user isn't going to have a pre-existing `@` object lying around. But now we have an extra parser rule.

Around ergonomics, I think `x` is more convenient (I'm not advocating for `x`, just as an example) since it removes the need for the `Shift + 2` (on a US keyboard).


I'm not convinced this is a good idea, specifically with regards to the `@` character. 
If y'all want to proceed with it, I think we probably need tests demonstrating that we can handle situations where someone looks up a function using `@` and then wants to use it as a decorator. I don't think someone _should_ do that, but someone will try, and if we want to add syntax, we need to make sure it plays well with others. 

So, some wacky options to consider:

```
@@.imp.importlib.contextmanager
def my_mgr():
    foo()
    yield
    bar()
```

```
A @ @.env["NUMPY_ARRAY_ENV_VAR"]
```

jnoortheen:
Thanks @gforsyth for the detailed comment. 

As we see, the `__xonsh__.imp` is being useful, and that was the main motive for repeatedly using `__xonsh__`, lets make `imp` available in the session like `aliases`. Other such usage could also be made available through descriptive names.

anki-code:
@gforsyth @jnoortheen I understand your point, but as with many similar cases I've encountered before in the xonsh world (I've done 337 PRs to xonsh, redesign the xonsh website and logos, contribute to the core), we need to make a decision that may not be perfect, but it will move us forward instead of stuck in workarounds. What do I mean:

----------

> Do `x = __xonsh__`

🔴 THIS IS STEP BACKWARDS. Because in every Github thread, every Zulip/Gitter chat, in every SSH session, in every repeateable example in the [xonsh cheatsheet](https://github.com/anki-code/xonsh-cheatsheet), in every single line of code I want to share with someone I need to do `x = __xonsh__` before. @jnoortheen @gforsyth No, I don't want to do this. 

🟡 It's totally ok to do `x = __xonsh__` in the python code in a xontrib or script but it's totally not working for instant scripting in the shell.

🟢 Having built in shortcut solving this: `@.env.swap(PGPASSWORD=$(op read 'op://pwd')): $[psql connect to database]`. @gforsyth @jnoortheen I'm doing this many times a day on different SSH sessions and in different environments, I'm answering on xonsh users question and share the examples. No, I don't want to do `x = __xonsh__` every time.


-----------

> Let's say that __xonsh__ now X or `__X__` or any other valid Python global variable
> Let's add `imp` as a global variable.

🔴 THIS IS STEP BACKWARDS. We already have an issue with shadowing of functions: `id`, `zip`, `dir`, [etc](https://github.com/xonsh/xonsh/issues/5802). Having global `X` is a step to the shadowing issue. Having global `imp` python object is a step to the shadowing issue.

🔴 In the future with shell development we will make shadowing, shadowing and shadowing? What's next? Add `prompt` as a command? Add `events` as a command? Oh, noooo, `events` is already `xonsh.built_ins.DynamicAccessProxy`, whaaaat? Why `jobs` is a callable alias but `events` is a `DynamicAccessProxy`? God dammit! IT IS NOT CLEAR.

🔴 Having `imp` is a solution with dropping context. From where `imp` comes from? What is this? Why it's global object? Call `imp` and have `xonsh.built_ins.InlineImporter` - what(??) it's not the command? Why `jobs` is a command but `events` and `imp` are not. It's frustrated new comers. IT IS NOT CLEAR.

🟡  There is `env` major GNU core util. Doing `env = __xonsh__.env` (like idea to have `imp`) is a bad pattern. But there is a way out again! Having `@.env` we have xonsh environment. IT IS CLEAR.

🟢 Having `@` and `@.imp` and `@.env` gives clear concept. There is xonsh session object (`@`, `__xonsh__`) and it's python object. It has components. You can call component from xonsh session object. IT IS CLEAR AND STEP FORWARD.

🟢 With `@` you can make a clear two contexts of usage: you can have `jobs` as a subprocess command and you can have `@.jobs` as a python object. Want to operate fast in the shell? Run `jobs kill 123`. Want to debug something or have more details that is not provided? Do `print([j.details() for j in @.jobs._all])`. IT IS CLEAR.

----------

> it's making the shell look more arcane
> @@.imp.importlib.contextmanager
> A @ @.env["NUMPY_ARRAY_ENV_VAR"]

🟢  In xonsh we forever had many worlds and the key thing is we have a way to implement what we want to see.
You showing an examples with double conch and it's ok but YOU can implement it differently with `from importlib import contextmanager` and `A @ $NUMPY_ARRAY_ENV_VAR` - it's totally on YOUR side.
Do import if you need it. Use `__xonsh__` if you need it. It's totally on USER side.
You can come up with a lot of terrible examples in any language.

🔴 Doing `x = __xonsh__`, `${...}` looks like cryptic overhead in the shell context as well.

--------

> It's an additional `@`

🟢  Yes, that is true and we need to understand that this is NOT an additional conch. From the very beginning, the author intended `@()` as design pattern in xonsh. Having different meanings behind `@` is a xonsh distinctive feature.
It's like a culture, it's like a window in Windows, it's like daemon in Linux, it's like matrix in machine learning, it's like atom in parser, it's like burger in McDonalds. The conch in xonsh is a thing that did additional useful work and the bridge between shell world and Python world. Did you understand the idea @jnoortheen @gforsyth ?

🐚  The entire story, before we changed the prompt_end, the shell tried to apologize for not being like bash/posix or has mixed syntax. Stop it. We want to build on the xonsh's strengths, not apologize for not being like the world of the past.

-------

🟢  Finally I'm totally go forward with:
* `@` is an alternative to `__xonsh__`, not replacement. 
* Improving `__xonsh__` objects with good public interface e.g. having `__xonsh__.<component>.<func>` e.g. `__xonsh__.prompt.switch_syntax_highlighting('SQL')`.
* Add more docs around `__xonsh__`.This PR has the first introduction.

----

🚀  Future:
* I believe that in the future we will improve `__xonsh__` and we will have all that belongs to xonsh session in this object plus builtin commands i.e. having global `events` as `DynamicAccessProxy` is a bad design. The events is a part of xonsh session and it's needed to be the part of `__xonsh__`.
* Our namespace will be clear: builtin commands, tools from PATH, xonsh session object that helps to patch prompt, access xonsh features (events, history, env) and provide helpful shell shortcuts (imp, last).
------

✅  I'm totally to merge this. If you have another solution in the mind that solves points above, please share.

shevtsova:
> If y'all want to proceed with it, I think we probably need tests demonstrating that we can handle situations where someone looks up a function using @ and then wants to use it as a decorator. I don't think someone should do that, but someone will try

@gforsyth It's good idea and I've added decorator support. Thanks! In fact the changes in parser is not so big.

>  we need to make sure it plays well with others.

I've tested some cases. Perhaps the example with the decorator was the only one.

jnoortheen:
`__xonsh__` was designed to be used as an internal object. 
One example that stood out was `[j.details() for j in @.jobs._all])` which would need one to know about the `jobs._all` attribute. It is again leaking internal interface. This would block any future changes in them. If the `jobs` cli doesn't expose all the data, perhaps it should be added with a flag like `--verbose`.

anki-code:
> which would need one to know about the jobs._all attribute. It is again leaking internal interface

@jnoortheen 

1. Are you ok if we will have XonshSessionInterface behind `@`? And we will treat it and expand carefully. I mean:
    
    ```xsh
    class XonshSessionInterface:  # What is the name for class is better?
        """Proper description of object for users
        env - ...
        imp - ...
        """
        def __init__():
            self.env=__xonsh__.env
            self.imp=__xonsh__.imp  # or directly InlineImporter()
            # ...
    
    @
    # Proper description of object for users
    #    env - ...
    #    imp - ...
    ```
    This looks awesome!

2. Can we remove `__xonsh__.env` from tutorial after this and use `@.env`?

I like it! @shevtsova please take a look.

gforsyth:
> 1. Are you ok if we will have XonshSessionInterface behind `@`? And we will treat it and expand carefully. I mean:

This is a better idea.  Isolates the actual internals from the exposed interface and allows us to be deliberate in what we expose.

gforsyth:
> @gforsyth It's good idea and I've added decorator support. Thanks! In fact the changes in parser is not so big.

nice!

shevtsova:
Okay, what about this? @jnoortheen @gforsyth @anki-code 

```xsh
@
# <xonsh.built_ins.XonshSessionInterface>

@?
# Type:         XonshSessionInterface
# String form:  <xonsh.built_ins.XonshSessionInterface object at 0x100857f90>
# File:         /home/she/coding/xonsh/xonsh/built_ins.py
# Docstring:
# Xonsh Session Interface
# 
# Attributes
# ----------
# env : xonsh.environ.Env
#     A xonsh environment e.g. `@.env.get('HOME', '/tmp')`.
# 
# imp : xonsh.built_ins.InlineImporter
#     The inline importer provides instant access to library
#     functions and attributes e.g. `@.imp.time.time()`.
# 
# lastcmd : xonsh.procs.pipelines.CommandPipeline
#     Last executed subprocess-mode command pipeline
#     e.g. `@.lastcmd.rtn` returns exit code.

@.<Tab>
# env imp lastcmd

@.imp.json.loads('{}')
# {}
@.env.get('HOME')
# '/home/she'

ls no
@.lastcmd.rtn
# 2

@aliases.register
@@.imp.xonsh.tools.unthreadable
def _mycmd(args, stdin=None):
    return 'ok'
mycmd
# ok
```

shevtsova:
@gforsyth I have doubts about the correct description. Please fix wording in the tutorial in case if it's needed. Thanks!

anki-code:
@gforsyth @jnoortheen do we need to put history and completers to `@` shortcut? We have this about in the docs.
If yes I prefer to replace it to `@.history` and `@.completers` in the docs. May be with mention `__xonsh__` shortly.

### History

https://github.com/xonsh/xonsh/blob/29e38b65549754dbdfab6f20d27fe7d0314af4fe/docs/tutorial_hist.rst?plain=1#L339

https://github.com/xonsh/xonsh/blob/29e38b65549754dbdfab6f20d27fe7d0314af4fe/docs/tutorial_hist.rst?plain=1#L74-L75

https://github.com/xonsh/xonsh/blob/29e38b65549754dbdfab6f20d27fe7d0314af4fe/docs/tutorial_hist.rst?plain=1#L367-L374

https://github.com/xonsh/xonsh/blob/29e38b65549754dbdfab6f20d27fe7d0314af4fe/docs/tutorial_hist.rst?plain=1#L419

### Completer

https://github.com/xonsh/xonsh/blob/29e38b65549754dbdfab6f20d27fe7d0314af4fe/docs/tutorial_completers.rst?plain=1#L31-L33

https://github.com/xonsh/xonsh/blob/29e38b65549754dbdfab6f20d27fe7d0314af4fe/docs/tutorial_completers.rst?plain=1#L208-L209

shevtsova:
I've added `XonshSession.lastcmd` to be in sync with `XonshSessionInterface.lastcmd`. Now `last` is deprecated.

anki-code:
@gforsyth @jnoortheen I think we're ready to move forward with this feature.

anki-code:
@jnoortheen @gforsyth This is a very bright feature and it would be fun to release it before Christmas. That's why I'm a bit persistent. Just to have time on finalizing the release.

docs/tutorial.rst
```diff
@@ -154,39 +171,38 @@ Very nice.
    :ref:`$UPDATE_OS_ENVIRON <update_os_environ>` to ``True``.
 
 
-The Environment Itself ``${...}``
----------------------------------
+The Environment Itself ``@.env``
+--------------------------------
 
-All environment variables live in the built-in ``${...}`` (aka ``__xonsh__.env``) mapping.
+All environment variables live in the built-in ``@.env`` mapping.
```
gforsyth:
```suggestion
All environment variables live in the built-in ``@.env`` mapping.

(``@.env`` is a replacement for the somewhat ungainly ``${...}`` but code making use of ``${...}`` will still work.)
```

tests/test_integrations.py
```diff
@@ -137,6 +137,33 @@ def check_run_xonsh(cmd, fmt, exp, exp_rtn=0):
...
+        """
+@aliases.register
+@@.imp.xonsh.tools.unthreadable
+def _mycmd(args, stdin=None):
+    return 'ok'
+mycmd
+""",
+        "ok",
+        0,
+    ),
```
gforsyth: Nice, this is a good test of very probable diabolical behavior

gforsyth (changes requested):
Thanks for putting this together @shevtsova! 

I have some small changes to documentation and a request for one of the tests, but overall this looks very clean.

gforsyth (approved):
Ok, I'm on board with this.  Thanks @shevtsova !  I'm going to leave this unmerged for a bit so @jnoortheen can also take a look.

jnoortheen:
Thanks to @shevtsova and everyone involved!

anki-code:
It's so elegant for the shell command now 🤩 
```xsh
for f in g`*.csv`: @.imp.pandas.read_csv(f)['recipient'].to_csv(f+'-rec.csv')
```

anki-code:
We had a discussion about do we need `@` and here I see one additional +1. The conch is a single entry point to discover the extendable shell functionality. Instead of observing huge list in globals() or discover documentation you can just run `@.<tab>` or `@?` or `help(@)` and discover:

image

We can cleanly separate:
* Put objects to `@.` (aliases, env).
* Put commands to aliases (xontrib, xonfig).

anki-code:
Just to the history

> `__xonsh__` is supposed to be internals. Now, that ship has probably already sailed, but this is making it a documented and exposed part of the API. jaraco noted that there were no docs on the `__xonsh__` object. There probably should be docs on it, but also it wasn't created as a "featureful" object.

I read this point many times and thinking about what we can do in 6160 e.g. introduce XS/XSH as public XonshSession and private XonshSessionHandler, etc, etc. But I see that XSH was introduced (that is the same as `__xonsh__`) as PUBLIC object to USE in new xontribs interface, in xontrib-template, in documentation. Everywhere we have mentions of XSH and people are using this uncontrolled and unmanageable way in their projects (conda, mamba, AI tools, xontribs, etc, etc). In fact we have ugly XSH as "the main xonsh interface/API/representation for the world" and this spreaded in the world. This produces the case "Python2 vs Python3" because some day we will want (it's already here!) to break this and refactor this. I'm angry about that. I'm sad about that. I had to express this, sorry.

gforsyth:
Hey @anki-code -- one of the curses of having your software used is that you are often stuck with decisions that have been made in the past.  Sometimes you might even find that a decision you made in the past is different from what you might do today.  That's just the way it goes.

It's ok to be frustrated and it's ok to be angry.  It's NOT ok to call _any_ contributors work "shit" or to insult or demean any users or contributors to the project.

anki-code:
@gforsyth sorry, edited

anki-code:
@gforsyth please give me full access to xontrib-template I need to prepare it to 0.23.0 release, thanks!
(it will be cool to me to have all acess to all xontribs because I'm the one active contributor now)

PR #5959, opened 2025-11-28
feat: Added `$XONSH_PSTR_CHECK_EXISTS` env variable to enable checking path in p-string before usage

### LINKED ISSUE #5958: p-string with checking that directory or file exists
We have p-strings:

```xsh
p'/qwe/qwe'
# Path('/qwe/qwe')
```

To implement [fail-fast system](https://en.wikipedia.org/wiki/Fail-fast_system) what about having env for this:

```xsh
$XONSH_PSTR_CHECK_EXISTS = True

p'/home'
# Path('/qwe/qwe')

p'/home/user/file'
# Exception

urgent_function(p'/we/dont/know/that/this/path/exists')
# Exception
```

## For community

anki-code:
Resolves #5958

```xsh
echo p'/qwe'
# /qwe

$XONSH_PSTR_CHECK_EXISTS = True
echo p'/qwe'
# FileNotFoundError: Path '/qwe' does not exist.
```

## For community

### DIFF
--- a/tests/test_builtins.py
+++ b/tests/test_builtins.py
@@ -19,11 +19,11 @@ from xonsh.built_ins import (
     in_macro_call,
     list_of_list_of_strs_outer_product,
     list_of_strs_or_callables,
+    path_literal,
     pathsearch,
     regexsearch,
     reglob,
     superhelper,
-    path_literal,
 )
 from xonsh.environ import Env
 from xonsh.pytest.tools import skip_if_on_windows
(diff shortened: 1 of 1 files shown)


jnoortheen:
I don't see this will be beneficial to fail-fast system. The path string should stay as a simple wrapper around Path object. Tracebacks are not good in some cases; it is better to do defensive programming and check first and then run things.

anki-code:
Ok, let's keep it for future. I found this useful for one task. If I will face with this more times I will back to this PR.

PR #5960, opened 2025-11-28
refactor: remove artefacts of previous outdated build tools

anki-code:
Now we're using https://github.com/xonsh/xonsh/tree/main/.github/workflows

## For community


anki-code:
@jnoortheen what we need to clean in addition, what do you think?

jnoortheen:
I think that is all. We still need to migrate the Authors.rst generation from rever.xsh

PR #5962, opened 2025-11-29
feat: allow on_command_not_found handlers to return replacement commands (#5952)

### LINKED ISSUE #5952: feat: Ability to replace command in on_command_not_found event
We have [on_command_not_found](https://xon.sh/events.html#on-command-not-found) event. What about add an ability to replace command

```xsh

@events.on_command_not_found
def _e(cmd, args):
    if cmd[0] == 'vim':
        return ['vi'] + args

vim file  # runs `vi file` if there is no vim
```

## For community

EclipseAditya:
## Description

Handlers for the `on_command_not_found` event can now return a replacement command to execute instead of raising an error.

Fixes #5952

## Changes

- Capture return values from `on_command_not_found.fire()`
- Accept `list[str]` or `tuple[str, ...]` as replacement commands  
- Return `None` to let the original error be raised (backward compatible)
- Invalid returns (strings, empty lists) are ignored
- If replacement command also fails, original error is shown
- Fixed pre-existing bug: duplicate `suggest_commands()` call

## Example

```python
@events.on_command_not_found
def fallback(cmd, **kwargs):
    if cmd[0] == 'vim':
        return ['vi'] + cmd[1:]  # Replace vim with vi
    return None  # Let error happen for other commands

vim file.txt  # runs `vi file.txt` if vim is not installed

### DIFF
--- a/xonsh/shell.py
+++ b/xonsh/shell.py
@@ -75,6 +75,17 @@ Returns:
 * ``None``: To let the error be raised normally
 
 Note: If the replacement command also fails, the original error is shown.
+
+Example:
+
+.. code-block:: python
+
+    @events.on_command_not_found
+    def _vim_to_vi(cmd, **kwargs):
+        '''If vim not found let's try to use vi.'''
+        if cmd[0] == 'vim':
+            return ['vi'] + cmd[1:]
+
 """,
 )
 


anki-code (changes requested):

anki-code (approved):

anki-code:
Thanks and welcome @EclipseAditya !

PR #5966, opened 2025-11-30
fix(completer): Use substring instead of startswith for base completer

anki-code:
## Motivation

I don't know why we're using startswith but substring search gives more sane result.

## Before
```xsh
aliases['dev-xonsh-build'] = 'echo 1'
aliases['dev-xonsh-deploy'] = 'echo 2'
depl<Tab>
# Not working
```

## After
```xsh
aliases['dev-xonsh-build'] = 'echo 1'
aliases['dev-xonsh-deploy'] = 'echo 2'
depl<Tab>
dev-xonsh-deploy
```
Also non trivial example of getting `gmd5sum`:
image

## For community

### DIFF
--- a/xonsh/completers/tools.py
+++ b/xonsh/completers/tools.py
@@ -31,7 +31,6 @@ def _filter_ignorecase(text, prefix):
     return _filter_with_func(text, prefix, check_func)
 
 
-
 def get_filter_function():
     """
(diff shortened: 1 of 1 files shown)


anki-code:
Solved in https://github.com/xonsh/xonsh/issues/6125

PR #5986, opened 2025-12-12
feat(source): Added `.xonsh` extension support

anki-code:
### Motivation

We have `.xsh` as xonsh script extension but it's pretty common extension/abbreviation ([google](https://www.google.com/search?q=xsh+file)):

* XSH is a powerfull command-line XML editing tool.
* XSH - is an extension of bash. It works as a bash library framework.
* XSH - A simple framework for shell configuration management.
* XSH files package together 3D models with the precomputed smoothed normal vectors necessary for our method.
* XSH - The X-Shooter pipeline
* XSH opener is a tool with which you will find a way to easily open the XSH file.

So it will be cool to have an explicit alternative. 

This PR is to add `*.xonsh` files support. Pros:

* Ability to set explicit extension that has no intersections from the world of tools.
* Easy to place and search xonsh integrations and implementations.

## For community

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -51,7 +51,7 @@
         "Tab completion as a service",
         "Now powered by vibes and unit tests",
         "Infrastructure as a conchline",
-        "Reproducibility is a lifestyle"        
+        "Reproducibility is a lifestyle"
         ];
     document.write(taglines[Math.floor(Math.random() * taglines.length)]);
     </script>


anki-code:
@jnoortheen @gforsyth done

gforsyth:
I don't know that I've ever seen an issue in 10+ years where someone had a collision with one of these other formats.

Currently, if you have two files with the same name but `py` and  `xsh` extensions and you try to `import`, the `xsh` file will "win".  With this PR, the `xonsh` file will now "win".

My concern is that we're adding additional surface to all the import hook machinery to guard against something that we don't specifically know is an issue.

anki-code:
Ok let's close for now. My main motivation was. I made a try to integrate xonsh with some tools. And I see that every tool has listing of extensions of supported shells e.g. .sh, .zsh, .fish, etc. The xsh in this list looks not explicit, hidden. What is xsh? I don't know. Having xonsh extension allows to uncover and present xonsh shell more directly. 
But... yeah we have additional overhead with adding new extension.

PR #5993, opened 2025-12-15
fix: allow spaces in auto-suggest completions

### LINKED ISSUE #4633: Allow spaces in AUTO_SUGGEST_IN_COMPLETIONS?
I'd like to allow spaces to be within the suggestion feature in completions however currently it strips out all after the first space. I tried messing around with `completer.py` and `auto_suggest` for a while but gave up because I couldn't get it working consistantly

Hammer2900:
### Description
When `$AUTO_SUGGEST_IN_COMPLETIONS` is enabled, the auto-suggestion inserted into the completion menu was previously truncated at the first space.
This PR fixes it by using the full suggestion text, allowing users to complete entire commands from history via the completion menu.

### Related Issue
Closes #4633

### DIFF
--- a/xonsh/shells/ptk_shell/completer.py
+++ b/xonsh/shells/ptk_shell/completer.py
@@ -130,7 +130,9 @@ class PromptToolkitCompleter(Completer):
         full_completion = prev + sug.text
         if len(full_completion) > 60:
             display_text = full_completion[:57] + "..."
-            return RichCompletion(full_completion, display=display_text, append_space=False)
+            return RichCompletion(
+                full_completion, display=display_text, append_space=False
+            )
         return full_completion
 
     def reserve_space(self):


anki-code:
Please add a test or repeatable example.

Hammer2900:
Added a unit test reproducing the issue and verifying the fix.

anki-code:
What about Gil's comment - https://github.com/xonsh/xonsh/issues/4633#issuecomment-1024239012 ?

Did you tried to use this with long commands? How it looks?

Hammer2900:
@anki-code I've implemented the truncation logic as suggested by Gil.

**Long commands truncation**: If the auto-suggestion is longer than 60 characters, it now returns a RichCompletion object. The full command is preserved as the value (inserted on selection), but the display text is truncated with ... to keep the menu clean.

**Unit tests**: Updated the test case to cover both normal and long suggestions, verifying that RichCompletion is used correctly for long strings.

anki-code:
Thanks @Hammer2900 !

PR #5999, opened 2025-12-16
fix(completer): completion strips trailing newlines from filenames (#5507)

### LINKED ISSUE #5507: Auto-completion of file names chokes on embedded newlines (I guess?)
I accidentally created a file name with an embedded newline. (I wanted to write a systemd mount service and didn't consider that the output of `systemd-escape` includes a newline)
I then wanted to get rid of the file and tried to auto-complete the file name, which fails.

## Current Behavior

```xsh
$ touch $(systemd-escape --suffix=mount --path /media/backup-local)
$ ls media[TAB]  # Expands to:
$ ls r'media-backup\x2dlocal.mount'
ls: cannot access 'media-backup\x2dlocal.mount': No such file or directory
$ ls -l
-rw-r--r--  1 root   root     112 16. Jun 16:15 'media-backup\x2dlocal.mount'$'\n'
```

## Expected Behavior
Auto-completion should complete to some string that matches the file on disk. For example

```xsh
$ rm media[TAB]  # Should expand to:
$ rm r'''media-backup\x2dlocal.mount
'''
$
```

## For community

Hammer2900:
Fixes #5507

### Description
This PR fixes a bug where the path completer was incorrectly stripping trailing whitespace (newlines `\n` and spaces ` `) from filenames returned by the filesystem. This caused auto-completion to generate invalid paths for files created with trailing control characters (e.g. `touch "name\n"`).

The issue was caused by `p.rstrip()` in `xonsh/completers/path.py`, which removed necessary characters from valid paths found via glob.

### Changes
- Removed `p.rstrip()` in `xonsh/completers/path.py` to preserve exact filenames.
- Added a regression test `test_complete_path_strip_check` in `tests/completers/test_path_completers.py` covering both trailing newlines and trailing spaces.

### Before/After Behavior

**Setup:**
```xsh
touch "badname\n"
```

**Before:**
```xsh
ls bad<TAB>
# Expands to: ls 'badname'
# Result: ls: cannot access 'badname': No such file or directory
```

**After:**
```xsh
ls bad<TAB>
# Expands to: ls 'badname\n' (or properly quoted equivalent)
# Result: File is found.
```


wajdiJomaa:
hey @anki-code, I was reviewing this pull request and I found some issues:
-  the completion of commands like `ls bad<tab>`  is relying on bash completion and not path completion, this issue also exists in normal bash shell, so the applied fix on path completion didn't fix it
-  also for path completion even after disabling strip the completion will not be displayed correctly the completion will not be "badname\n" instead it will jump new line and this is also false

Screenshot from 2026-02-07 21-41-01

this is the result of completion from path

It should be rare that file names contains \n so maybe in the case the user is forced to write "name\n" instead of relying on autocompletion, as I said bash also will fail to autocomplete

wajdiJomaa:
@anki-code I am not sure if this is AI generated but it could be 😬

anki-code:
Thanks for diving into it @wajdiJomaa! We want to be better than bash so do we need to merge this PR or you will introduce new fix?

wajdiJomaa:
@anki-code I don't think the current pr will fix the issue, I'll check what we can do

PR #6005, opened 2025-12-17
chore: TEST add support for '@' prompt in token patterns for docs

anki-code:
I will revert this if its will not work in docs

## For community


anki-code:
I will rever this if it will not work in docs

PR #6009, opened 2025-12-18
fix(environ): prevent tilde expansion inside environment variables

### LINKED ISSUE #5970: env: unexpected expanding string to path
```xsh
with @.env.swap(QWE="~"):
    echo $QWE
# /home/user
```
Expected `~` as an output.

It looks this is old bug. I can reproduce on xonsh 0.18.

Workaround is using `@()`: 
```xsh
with @.env.swap(QWE="~"):
    echo @($QWE)
# ~
```

## For community

Hammer2900:
Previously, environment variables were expanded before tilde expansion. This caused variables containing '~' to be incorrectly expanded to the user's home directory. This change swaps the order so tilde expansion happens first.

Closes #5970

## For community


anki-code:
After diving into this logic I see that we need to understand the motivation behind this logic.
Please find the initial PR/commit that introduced this:

https://github.com/xonsh/xonsh/blob/b4224b011702eaa6b173378f0ca2752ca8d95693/xonsh/tools.py#L147-L157

May be we need to introduce `XONSH_ENV_EXPANDUSER` that `False` by default and switch this logic off by default OR we need to implement this for certain env vars like `PATH` OR we need to detect the nev variables carefully.

In the current implementation I see the issue in autogenerated passwords e.g.:
```xsh
with @.env.swap(PGPASSWORD=random_password()):  # PGPASSWORD='~/@#Dqdq1)*!@#'
    echo $PGPASSWORD
# '/home/user/@#Dqdq1)*!@#'
```
So we need to fix the logic of applying expanduser.

Hammer2900:
> Thanks for the detailed explanation! You are right, simply swapping the order breaks workflows where users expect tilde expansion from variables (e.g. `ROOT=~`).
>
> I have updated the PR to introduce a new environment variable `XONSH_ENV_EXPANDUSER`.
>
> *   **Default (`False`)**: "Safe mode". Tilde expansion happens **before** variable expansion. This prevents generated passwords like `~secret` from being corrupted. This fixes the reported issue #5970.
> *   **Opt-in (`True`)**: "Legacy mode". Variable expansion happens **first**, so variables containing `~` will be expanded to the home directory.
>
> I also updated `xonsh/tools.py` to handle `pathlib.Path` and `bytes` inputs correctly, fixing the CI errors we saw earlier. Tests have been added to cover both modes.

anki-code:
Thanks for pushing this forward!

> Please find the initial PR/commit that introduced this:

We need this to understand where we need to save the initial logic. Please mention the PR or commit.

I suppose this logic was introduced to be in sync with bash in case of PATH or DIRS. We need to understand initial motivation of adding this.

anki-code:
Hi @Hammer2900, I've made a bit research and it looks most of previously implemented logic has different use cases so we can't just change the order. What I did is to highlight the area for improvement by making this optional in - https://github.com/xonsh/xonsh/pull/6240

So I'm going to close this PR. We need to form full picture research with suproc and python behavior and after this make breaking changes if it's needed.

Thanks for the effort @Hammer2900 !

Issue #3914, opened 2020-10-21
File path completer: 5 cases


bobhy:
Several issues with Xonsh path completer displaying either completions in the (prompt-toolkit) completion menu or displaying a single selected completion on the command line.  No issues (found so far) with completer selecting the wrong set of completions.

I can take a crack at fixing these; I'm pretty sure most can be addressed by having the path completer return `RichCompletion`s and not have to rely on the Xonsh completer to do context-free quote stripping. But I'd especially like to hear from someone who knows what the reference to `readline` is about in item 5.

## xonfig

<details>

```
$ xonfig
+------------------+----------------------+
| xonsh            | 0.9.24               |
| Git SHA          | 3afd5430             |
| Commit Date      | Oct 20 07:21:11 2020 |
| Python           | 3.8.6                |
| PLY              | 3.11                 |
| have readline    | False                |
| prompt toolkit   | 3.0.8                |
| shell type       | prompt_toolkit       |
| pygments         | 2.7.1                |
| on posix         | False                |
| on linux         | False                |
| on darwin        | False                |
| on windows       | True                 |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
| on jupyter       | False                |
| jupyter kernel   | None                 |
| xontrib 1        | coreutils            |
| xontrib 2        | readable-traceback   |
| xontrib 3        | vox                  |
| xontrib 4        | voxapi               |
+------------------+----------------------+
```

</details>

##  Issues

All based on directory with these (contrived) contents:
![image](https://user-images.githubusercontent.com/1971300/96674103-db0ce700-131c-11eb-81a8-7444584f9871.png)

### Case 1 (fix https://github.com/xonsh/xonsh/pull/5816)

If file name contains backslash, completion is displayed in the menu with leading `r'`. I expect the names in completion menu to not be quoted, as they are not when name contains a special character other than backslash.  (this case is non-Windows only, on Windows, file path cannot contain backslash at all.  Backslash as path (directory) separator on Windows has different issues, see next.)
 
![image](https://user-images.githubusercontent.com/1971300/96674661-1f4cb700-131e-11eb-877e-8adfe778c831.png)

1. type `ls test<TAB>`, then `<TAB>` again to select first completion.

### Case 2 (fix https://github.com/xonsh/xonsh/issues/3914#issuecomment-2184979794)
[...]

PR #6030, opened 2025-12-31
fix: prevent stripping r except leading ones of r-strings

### LINKED ISSUE #5967: fix(completer): Do not show part of the name in completer list
```xsh
aliases['reload'] = 'echo 1'
```

Image

Expected: `reload` as an option instead of `eload`.

## For community

### LINKED ISSUE #6003: completer: show the exact match on first position
Image

Expected `-i` on first position:

Image

## For community

inmaldrerah:
Trying to fix #5967 and fix #6003 (which I believe is the same as #6049), at least the obvious appearance of it.

Fixes the first case of #3914 correctly.

Now the leading `r` of non-r-strings will not get stripped.

E.g. `r` in `result` won't get removed, while the leading `r'` and trailing `'` in `r'r\r'` gets removed.

image 

## For community

### DIFF
--- a/docs/aliases.rst
+++ b/docs/aliases.rst
@@ -181,7 +181,7 @@ isolated environment (e.g. conda, mamba, homebrew).
 
 .. code-block:: console
 
-    @ which pip 
+    @ which pip
     /usr/bin/pip  # system pip
     @ which xpip
     /home/snail/.local/xonsh-env/bin/python -m pip  # current xonsh session pip
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -39,7 +39,7 @@ the main approaches and their advantages.
       - ●
       - ●
       - ◯
-      - 
+      -
     * - PyPi package
       -
       -
@@ -62,7 +62,7 @@ the main approaches and their advantages.
       -
       -
       - ◯
-      - 
+      -
       -
 
 Work in progress:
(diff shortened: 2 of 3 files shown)


anki-code:
Resolve comflicts and add tests please.

inmaldrerah:
I have included some tests, and some more fixes (of undocumented issues).

Now this PR also fixes these two problems: 

1. trailing quotes are no longer removed (and yes single/double quotes are valid characters in file names):

   image

   where the first completion actually means `/'''`, but the trailing `'`s are `.strip`ped. Now it works:

   image

   ~The `'quote'` test case in `test_base_completer.py` should be the test for this.~
   ~Oh, it seems it's in the wrong file. Reverted.~

2. This PR also fixes the leading `/` which should also be stripped (paths should be totally stripped with no `/` kept, according to the logic of the [code](https://github.com/xonsh/xonsh/blob/main/xonsh/shells/ptk_shell/completer.py#L91-L95)), because the prefix should be removed after stripping quotes (while it wasn't previously, at line [113](https://github.com/xonsh/xonsh/blob/main/xonsh/shells/ptk_shell/completer.py#L113) and [120](https://github.com/xonsh/xonsh/blob/main/xonsh/shells/ptk_shell/completer.py#L120)), since [the shared prefix length `pre`](https://github.com/xonsh/xonsh/blob/main/xonsh/shells/ptk_shell/completer.py#L96-L100) is calculated from [completions with removed quotes](https://github.com/xonsh/xonsh/blob/main/xonsh/shells/ptk_shell/completer.py#L89-L90).

   ~However, I don't know how to test this situation. Any suggestions on this?~

I have now included test cases for all issues fixed.

anki-code:
Thanks @inmaldrerah !

JFYI if you have time to continue improving xonsh completion I have some thoughts. Let me know.

PR #6035, opened 2026-01-04
fix: completer  show the exact match on first position

wajdiJomaa:
should not strip the "r"s from the completer options, full details #6003 

## For community

### DIFF
--- a/docs/aliases.rst
+++ b/docs/aliases.rst
@@ -123,7 +123,7 @@ Runs timing study on arguments. Similar to IPython's ``%timeit`` magic.
 
 ``showcmd``
 ============
-Displays how commands and arguments are evaluated.
+Displays how commands and arguments are evaluated. Use ``-e`` to expand aliases.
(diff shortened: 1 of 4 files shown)


inmaldrerah:
Oh, it seems we have done similar things. I've just made #6030 a few days before yours, and it tries to fix case 1 of #3914 as well.

wajdiJomaa:
@inmaldrerah I only noticed your pull request after I opened this one, the only difference that I am not trying to fix the first case of #3914, I am only reverting the changes in order to fix the completion issue for #6003, I am not sure should I close this pull request

inmaldrerah:
I don't know whether closing or not is better. Maybe just leave it here and let the maintainers decide which PR to merge.

anki-code:
We had a regress and the same change was applied as revert in #6052 so I'm going to close this.
In the future we will merge PRs for completer only with tests.

PR #5941, opened 2025-11-22
feat: Keep hash (`#`) if it's part of the string in subprocess command

anki-code:
## Motivation

Resolves #5937 with examples where hash is the part of argument:
```xsh
showcmd nix shell nixpkgs#mc
showcmd curl https://xon.sh/#faq-area
showcmd pip install git+https://github.com/user/repo.git#egg=package
showcmd npm install git+https://github.com/user/repo.git#commit
```

## Before
|  cmd | bash, fish    | xonsh   | 
|---|---|---|
|  `echo 1#2` |  `1#2` |   🔴 `1`  |  
|  `echo 1# 2` |  `1# 2`  |   🔴 `1` | 
|  `echo 1 #2` |   `1` |   🟢 `1` |
## After
|  cmd | bash, fish  |  xonsh   | 
|---|---|---|
|  `echo 1#2` |  `1#2` |   🟢 `1#2`  |  
|  `echo 1# 2` |  `1# 2`  |   🟢 `1# 2` | 
|  `echo 1 #2` |   `1` |   🟢 `1` |

In addition:
* It works nice also for `$(echo 1#2)`, `!(echo 1#2)`, `![echo 1#2]`, `$[echo 1#2]`.
* Fixes https://github.com/xonsh/xonsh/issues/5928

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #6042, opened 2026-01-10
fix(lexer): handle comment to keep `a#b` as an argument

anki-code:
Work in progress

### Before

After https://github.com/xonsh/xonsh/pull/5941 we need to fix this case:

```xsh
aliases['a'] = 'echo a#b c # d'
aliases['a']
# ['echo', 'a']
```

Workaround is to use list of arguments:
```xsh
aliases['a'] = ['echo', 'a#b', 'c']
aliases['a']
# ['echo', 'a#b', 'c']
```

### After 

```xsh
aliases['a'] = 'echo a#b c # d'
aliases['a']
# ['echo', 'a#b', 'c']
```

## For community

### DIFF
--- a/xonsh/parsers/lexer.py
+++ b/xonsh/parsers/lexer.py
@@ -238,6 +238,47 @@ def handle_ignore(state, token):
     yield from []
 
 
+def handle_comment(state, token):
+    """
+    Treat '#' as a comment only if it is preceded by whitespace.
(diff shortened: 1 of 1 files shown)


anki-code:
I have not enough time to complete this right now but I see that Copilot with GPT-5 is proposing pretty sane implementations to fix this.

PR #6056, opened 2026-01-16
feat(builtin): Added `@.history` as a shortcut to `__xonsh__.history`

anki-code:
We have many examples and code with `__xonsh__.history` so let's add it to the shortcut `@.history` for consistency.

## For community

### DIFF
--- a/tests/lib/test_jsonutils.py
+++ b/tests/lib/test_jsonutils.py
@@ -4,8 +4,8 @@ import json
 
 import pytest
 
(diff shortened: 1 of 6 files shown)


anki-code:
Ok, we have many good changes and fixes in main so we want to have a monthly release. The xore team is busy so I'm going to do self-merge and self-release.

This PR has clean logic and having `@.history` (xonsh session history) looks clear.

PR #5938, opened 2025-11-21
feat: Allow sourcing files with different extensions

slacksystem:
Resolves #5480 

1. Add `.xonshrc` to the list of acceptable extensions.
2. Add flag `-e --ignore-ext` that disable extension checking.
3. Add info about flag to the error message

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #6059, opened 2026-01-16
fix(mamba-installer): Rename xmamba to xmamba.xsh

anki-code:
### Motivation

To avoid `source -e` after it was introduced in #5938 

### Before

```xsh
source xmamba
xmamba --help
```

### After
```xsh
source xmamba.xsh
xmamba --help
```

## For community


anki-code:
Trivial fix, self-merge

Issue #6058, opened 2026-01-16
Edge case: redirect without left part follows to IndexError: list index out of range


anki-code:
Any redirect:
```xsh
> 1.txt
< 1.txt

>> 1.txt
2> 1.txt
1> 1.txt
e> 1.txt
```
Follows to:
```xsh
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/pc/.local/xonsh-env/lib/python3.12/site-packages/xonsh/built_ins.py", line 271, in subproc_captured_hiddenobject
    return xonsh.procs.specs.run_subproc(cmds, captured="hiddenobject", envs=envs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pc/.local/xonsh-env/lib/python3.12/site-packages/xonsh/procs/specs.py", line 1126, in run_subproc
    specs = cmds_to_specs(cmds, captured=captured, envs=envs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pc/.local/xonsh-env/lib/python3.12/site-packages/xonsh/procs/specs.py", line 1056, in cmds_to_specs
    spec = SubprocSpec.build(cmd, captured=captured, env=env)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pc/.local/xonsh-env/lib/python3.12/site-packages/xonsh/procs/specs.py", line 676, in build
    spec.resolve_alias()
  File "/Users/pc/.local/xonsh-env/lib/python3.12/site-packages/xonsh/procs/specs.py", line 735, in resolve_alias
    cmd0 = self.cmd[0]
           ~~~~~~~~^^^
IndexError: list index out of range
```

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

wajdiJomaa:
hey @anki-code, what is the correct behavior? what is the expected result?
I can fix the list out of range to "xonsh: subprocess mode: command is empty" 
but the commands above are valid bash commands so that is why I am asking

anki-code:
That's will be enough, PR is welcome!

wajdiJomaa:
[...]

PR #6060, opened 2026-01-17
fix: Raise an error in case of redirect without command

wajdiJomaa:
this a fix for the issue #6058, in case we have a redirect without left part I added an empty "" as the command this will return a message "xonsh: subprocess mode: command is empty" when running the empty command

## For community

### DIFF
--- a/tests/procs/test_specs.py
+++ b/tests/procs/test_specs.py
@@ -18,13 +18,13 @@ from xonsh.procs.specs import (
     get_script_subproc_command,
     run_subproc,
 )
-from xonsh.pytest.tools import skip_if_on_windows
+from xonsh.pytest.tools import ON_WINDOWS, VER_MAJOR_MINOR, skip_if_on_windows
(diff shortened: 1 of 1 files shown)


anki-code:
1. It will be cool to implement this explicitly i.e. if there is no command during resolve_redirects [raise XonshError](https://github.com/xonsh/xonsh/blob/03c2c3d1af8193718080b92dfb15dd8abd313d07/xonsh/procs/specs.py#L227).

2. Please add a test e.g. https://github.com/xonsh/xonsh/blob/03c2c3d1af8193718080b92dfb15dd8abd313d07/tests/procs/test_specs.py#L436-L437

wajdiJomaa:
@anki-code  I changed it to raise xonsherror inside resolve_redirect + added a test, sorry messed up a little bit with the commits

anki-code (approved):

anki-code:
Thanks @wajdiJomaa!

PR #6093, opened 2026-02-05
fix(prompt): Keep original key_bindings in prompt-toolkit shell to easy add new key bindings during session

anki-code:
### Motivation

Reduce complexity on managing key bindins during session.

### Before

We want to add `Ctrl+F` hotkey during session and our first thought is about doing same as binding approach in [`on_ptk_create` event](https://xon.sh/events.html#on-ptk-create).

```xsh
__xonsh__.shell.shell
# <xonsh.shells.ptk_shell.PromptToolkitShell>
__xonsh__.shell.shell.key_bindings
# <prompt_toolkit.key_binding.key_bindings._MergedKeyBindings>

@__xonsh__.shell.shell.key_bindings.add("c-f")
def _my_ctrl_f(event):
    pass
# AttributeError: '_MergedKeyBindings' object has no attribute 'add'
# Yes, because key_bindings are inside MergedKeyBindings
```

So we need to patch deeper and it's working:

```xsh
current_bindings = __xonsh__.shell.shell.key_bindings.registries[0]
@current_bindings.add("c-f")
def _my_ctrl_f(event):
    if not str(event.current_buffer.text):
        (buf := event.current_buffer).insert_text('echo F')
        buf.validate_and_handle()

# Ctrl+F
# echo F
# F
```

### After

Very close approach to using [`on_ptk_create` event](https://xon.sh/events.html#on-ptk-create):

```xsh
__xonsh__.shell.shell.key_bindings
# <prompt_toolkit.key_binding.key_bindings.KeyBindings>


@__xonsh__.shell.shell.key_bindings.add("c-f")
def _my_ctrl_f(event):
    if not str(event.current_buffer.text):
        (buf := event.current_buffer).insert_text('echo F')
        buf.validate_and_handle()

# Ctrl+F
# echo F
# F
```

### Implementation

I have feeling that we need here more eyes and mamanging merged key bindings more carefully and may be provide more clean API to add incude-exclude things like `load_emacs_shift_selection_bindings`. But we have no user issues around this and current PR looks the shortest way to rid of deep patching on user side.

## For community


anki-code:
I'm going to self-merge this to have it in fix release. In case of issues we can revert because it's a few lines change.

PR #6102, opened 2026-02-17
fix(completion): Fix os.path.commonprefix deprecation warning in python3.15

lambda-abstraction:
python/cpython#144436 deprecates `os.path.commonprefix` in 3.15, causing a `DeprecationWarning` each time when using completion

i copied the implementation of `commonprefix` (https://github.com/python/cpython/blob/957f9fe162398fceeaa9ddba8b40046b8a03176d/Lib/genericpath.py#L115-L129) and removed the unused `fspath`, as xonsh only uses `commonprefix` with `str | RichCompletion`, not `PathLike`

i first thought about putting the function into the `xonsh.tools` module, but one of the places where `commonprefix` is used in `xonsh/completers/bash_completion.py`, which is intended to be usable as a standalone script, and adding an `import xonsh.tools` increased the startup time by a significant amount, so i decided to create a new helper module

## For community

### DIFF
--- a/tests/lib/test_string.py
+++ b/tests/lib/test_string.py
@@ -1,8 +1,11 @@
+import pytest
+
 from xonsh.lib.string import commonprefix
 
 
-def test_commonprefix():
-    cases = [
+@pytest.mark.parametrize(
+    "strings, expected_prefix",
+    [
         ([], ""),
         (["ab", "cd"], ""),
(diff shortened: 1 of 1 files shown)


anki-code:
LGTM, please add a test

anki-code:
Thanks @lambda-abstraction !

PR #6119, opened 2026-03-05
fix(alias_return_command): Better error message in case of wrong return

anki-code:
subj

## For community


anki-code:
I'm going to do self merge because of fix release.

Issue #5631, opened 2024-07-25
Callable alias with `execx` can provoke `Bad file descriptor` exception in the internal or external code


anki-code:
### How it looks for a user

The exceptions below are not the root of the issue. It's just consequences of closing the descriptor somewhere in upstream or in parallel thread during execution of callable alias.

Where I can catch this exception:
* In case of using ExecAlias e.g. `aliases['a'] = "echo @(1)"`
* In case of using `execx` in callable alias e.g. `aliases['a'] = lambda: execx('echo 1')`
* I catch this during `conda` and `mamba` hooks usage.

How exception looks in ptk:
```xsh
xonsh --no-rc
aliases['t'] = 'echo 1 && echo 2'  # or just 'echo @(1)'
aliases['t']
# ExecAlias
t
t
t
# ...
```
```xsh
  File "/Users/pc/git/xonsh/xonsh/main.py", line 512, in main
    sys.exit(main_xonsh(args))
             ^^^^^^^^^^^^^^^^
  File "/Users/pc/git/xonsh/xonsh/main.py", line 558, in main_xonsh
    shell.shell.cmdloop()
  File "/Users/pc/git/xonsh/xonsh/shells/ptk_shell/__init__.py", line 415, in cmdloop
    line = self.singleline(auto_suggest=auto_suggest)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pc/git/xonsh/xonsh/shells/ptk_shell/__init__.py", line 383, in singleline
    line = self.prompter.prompt(**prompt_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pc/.local/mamba-envs/Users_pc_git_xonsh/lib/python3.12/site-packages/prompt_toolkit/shortcuts/prompt.py", line 1024, in prompt
    return self.app.run(
           ^^^^^^^^^^^^^
  File "/Users/pc/.local/mamba-envs/Users_pc_git_xonsh/lib/python3.12/site-packages/prompt_toolkit/application/application.py", line 973, in run
    return loop.run_until_complete(coro)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pc/.local/mamba-envs/Users_pc_git_xonsh/lib/python3.12/asyncio/base_events.py", line 674, in run_until_complete
    self.run_forever()
[...]

PR #6123, opened 2026-03-06
test: Test fixing the case when last proc in the pipe closed without waiting previous processes

anki-code:
Fixed https://github.com/xonsh/xonsh/issues/5631#issuecomment-4010217740:
```xsh
@aliases.register
def _a1(args, stdin, stdout):
    @.imp.time.sleep(0.05)
    print(1)

@aliases.register
def _a2(args, stdin, stdout):
    print(2)
            
a1 | a2  # a2 will be ended before a1

for i in range(100):
    a1 | a2
```

## For community

### DIFF
--- a/xonsh/procs/pipelines.py
+++ b/xonsh/procs/pipelines.py
@@ -307,7 +307,7 @@ class CommandPipeline:
         # GNU Parallel, which has a long startup time.
         first_read = True
 
-        while first_read or (proc.poll() is None and self._any_proc_running()):
+        while proc.poll() is None or first_read or self._any_proc_running():
             first_read = False
             if getattr(proc, "suspended", False) or self._procs_suspended() is not None:
                 self.suspended = True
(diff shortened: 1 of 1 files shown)


anki-code:
Ok, it's working but my test is not stable:

```xsh
test_pipe_last_proc_ending = [
    """
$XONSH_SHOW_TRACEBACK = True

@aliases.register
def _a1(args, stdin, stdout):
    @.imp.time.sleep(0.01)
    print(1)

@aliases.register
def _a2(args, stdin, stdout):
    print(2)
            
for i in range(10):
    a1 | a2

"""
]

@skip_if_on_windows
@pytest.mark.parametrize("test_code", test_pipe_last_proc_ending)
def test_last_proc_ending_before_others(test_code):
    """Test last proc ending before others (#6123)."""

    out, err, ret = run_xonsh(
        test_code, interactive=True, single_command=True, timeout=60
    )
    assert out == '2\n'*10
```
Sometimes ok sometimes `Bad file descriptor` so we need to back to fixing FD duplication in #5631.

PR #5961, opened 2025-11-28
refactor!: First introduction deprecation warning for env variables. AUTO_SUGGEST now XONSH_PROMPT_AUTO_SUGGEST.

anki-code:
Start working on #5149

In this PR:
* Rename AUTO_SUGGEST to XONSH_PROMPT_AUTO_SUGGEST. I'm not using PTK prefix here to make this variable usable for potential new prompts e.g. cmd2 or inquire.
* Added DeprecatedSettings with deprecated env variables
* Added sync between new and old variables for backwards compatibility.
* Added DeprecationWarning when you set the value to deprecated env variable.

Example:

```xsh
$AUTO_SUGGEST
# True
$XONSH_PROMPT_AUTO_SUGGEST
# True

$AUTO_SUGGEST = 0
# DeprecationWarning: env: Using deprecated env variable 'AUTO_SUGGEST'. Replace it to 'XONSH_PROMPT_AUTO_SUGGEST'.

$AUTO_SUGGEST
# False
$XONSH_PROMPT_AUTO_SUGGEST
# False

$XONSH_PROMPT_AUTO_SUGGEST = 1

$AUTO_SUGGEST
# True
$XONSH_PROMPT_AUTO_SUGGEST
# True
```

BEGIN_COMMIT_OVERRIDE
refactor!: `$AUTO_SUGGEST` is now deprecated in favour of `$XONSH_PROMPT_AUTO_SUGGEST`

END_COMMIT_OVERRIDE

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

Issue #4489, opened 2021-09-27
Fuzzy completions


jnoortheen:
PTK supports having [fuzzy completions](https://python-prompt-toolkit.readthedocs.io/en/master/pages/asking_for_input.html#fuzzy-completion). Xonsh should also support fuzzy search. It would require changes from all the current completers. As of now these individual functions send filtered completions with prefix. Now the top-level completer should be handling the filtering process.

### Traceback (if applicable)

<details>

```
traceback
```

</details>

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

gforsyth:
This would be nice to have.  I'll point out that we already have fuzzy matching in all path-like things

PR #6125, opened 2026-03-06
feat(Completer)!: Order completions by both prefix and substring matches and sort by substring position. Remove `$CASE_SENSITIVE_COMPLETIONS`.

### LINKED ISSUE #6082: Switch completers to search by substring instead of beginning of the line
I'm totally disappointed that after refactoring most of completers are working by the beginning of the line i.e.

```xsh
aliases |= {'aaasdfgh':'echo', 'qwe-asd':'echo', 'zxc-cvb':'echo', 'asdfgh':'echo'}
asd<tab>
# asdfgh  # no qwe-asd, aaasdfgh
```
Same with options e.g. using [fish completer](https://github.com/xonsh/xontrib-fish-completer) for `grep -i<tab>` we have `-iA`,`-ia`, etc, instead of just `-i` first.

How it should work:
1. Find all `*asd*` options by substring: `qwe-asd`, `aaasdfgh`, `asdfgh`.
2. Show the result by the position in the string - from the beginning: `asdfgh`, `aaasdfgh`, `qwe-asd`.

## For community

costajohnt:
## Summary

Fixes #6082

Switches completers from `startswith` to substring (`in`) matching, so that typing a partial string matches candidates containing it anywhere — not just at the beginning.

**Before:**
```xsh
aliases |= {'dev-xonsh-deploy':'echo'}
deploy<Tab>
# nothing
```

**After:**
```xsh
aliases |= {'dev-xonsh-deploy':'echo'}
deploy<Tab>
# dev-xonsh-deploy
```

## Changes

- **`xonsh/completers/tools.py`**: Switched `_filter_ignorecase` from `startswith` to `in` for substring matching. Removed `_filter_normal` and simplified `get_filter_function()` to always return the case-insensitive filter.
- **`xonsh/completers/commands.py`**: Replaced two hardcoded `startswith` calls with `get_filter_function()` for consistency
- **`xonsh/completer.py`**: Replaced the old `sortkey` (substring position + alphabetical) with a tiered ranking: case-sensitive prefix > case-insensitive prefix > case-sensitive substring > case-insensitive substring > no match. Within each tier, `_`-prefixed completions sort last, then by match position, then alphabetically.
- **`xonsh/completers/path.py`**: Removed `CASE_SENSITIVE_COMPLETIONS` usage. Path completion globs now skip `expand_case_matching` on Windows (NTFS is already case-insensitive), avoiding broken glob patterns from backslash path separators.
- **`xonsh/environ.py`**: Removed `$CASE_SENSITIVE_COMPLETIONS` env var and unused `ON_LINUX` import
- **`xonsh/shells/readline_shell.py`**: Simplified `_rebind_case_sensitive_completions` to always set `completion-ignore-case on`, since xonsh now handles case preference via sort-order tiers
- **`xonsh/tools.py`**: Fixed `expand_case_matching` to normalize backslashes to forward slashes on Windows before inserting bracket patterns (defense-in-depth)
- **`xonsh/built_ins.py`**: Removed `CASE_SENSITIVE_COMPLETIONS` from `globsearch`. Also fixed an unrelated bug in `DynamicAccessProxy.__setattr__` where `self.obj` was incorrectly passed as the attribute name to `super().__setattr__()`.

## Breaking change

`$CASE_SENSITIVE_COMPLETIONS` is removed. Completions are now always filtered case-insensitively. The tiered sort order already ranks exact-case matches above case-insensitive ones, making the env var unnecessary.

## Scope

This affects all completers that use `get_filter_function()`:
- Command completions (aliases, executables)
- Argument completions from subprocesses (e.g., fish completer)
- Import completions
- Python completions

Path completion uses its own matching logic and is **not** affected by the substring change, but it is affected by the `CASE_SENSITIVE_COMPLETIONS` removal.

## Test plan

- [x] All existing completer tests pass across Linux, macOS, and Windows
- [x] New tests verify substring matching for both case-sensitive and case-insensitive modes
- [x] New integration test verifies `grep` is found when typing `rep`
- [x] New test verifies tiered sort ordering (case-sensitive prefix < case-insensitive prefix < substring)
- [x] CI green on all 12 matrix jobs (Python 3.11-3.14 × ubuntu/macOS/Windows)


anki-code:
It works to me, thanks for taking this @costajohnt !

Please check the cases:

1. `decoder` expected as a first option

image

2. `-i` expected as first option after loading fish_completer. If it's fish completer issue feel free to leave it.

image

image

anki-code:
Better but now case sens:

image

image

And no grep completion:

image

costajohnt:
The case sensitivity comes from `$CASE_SENSITIVE_COMPLETIONS` defaulting to `True` on Linux — same as before with `startswith`. Would you want substring matching to always be case-insensitive regardless of that setting?

On `grep -i` completion: I tested locally and `rep` → `grep` works on this branch (it doesn't on main since `startswith` can't match it). The screenshot looks like argument completion after `grep -i`, which this PR doesn't touch — only command name matching changed.

Screenshot 2026-03-15 at 6 57 30 PM

anki-code:
> The case sensitivity comes from $CASE_SENSITIVE_COMPLETIONS defaulting to True on Linux — same as before with startswith. Would you want substring matching to always be case-insensitive regardless of that setting?

1. Let's make it False by default. I'm for searching case insensitive by default because search and autocompletion are created to search and find.
2. Let's rename `CASE_SENSITIVE_COMPLETIONS` to `XONSH_PROMPT_COMPLETION_CASE_SENSITIVE` like it was made for [deprecation AUTO_SUGGEST](https://github.com/xonsh/xonsh/pull/5961/changes) (rename env var, add `sync`, move old one to `DeprecatedSetting`).

> On grep -i completion

I see. It's the testing in docker issue. It works the same. Thanks!

jnoortheen:
I'd like the completions to be ranked based on like a basic fuzzy search 

1. case matching prefixes
2. case insensitive prefixes
3. case sensitive substrings
4. case insensitive substrings
5. (if at all simple and possible) an abbreviation style fuzzy matches

I wanted to implement this, sometime back - https://github.com/xonsh/xonsh/issues/4489

anki-code:
1. Let's remove CASE_SENSITIVE_COMPLETIONS completely as commented above.

2. I'm trying to test but:

image

image

Here I'm expecting that JSONDecoder will be the first because of "case matching prefixes" is first.

3. Let's use length and alphabetical order as well so `JSONDecoder` will be the first and `JSONDecodeError` will be the second i.e. `a`, `aa`, `ab`, `abc`, etc.

jnoortheen:
> Here I'm expecting that JSONDecoder will be the first because of "case matching prefixes" is first.

@anki-code kindly check the order of things. It still prefers the prefix match first. substring case sensitivity comes after.

anki-code:
@jnoortheen what about moving "case sensitive substrings" before "case insensitive prefixes" (OR we can apply it before if the string is case sensitive)? This allows to exclude unwanted functions. So if I need a class `SomeClassName` I just type `Name` and have it without having `get_some_name` functions in first positions.

So, if the string in lower case: case matching prefixes, case insensitive prefixes, case sensitive substrings, case insensitive substrings.
If the string case secnsitive (e.g. `Aaaa`): case matching prefixes, [case sensitive substrings], case insensitive prefixes, case insensitive substrings.

@jnoortheen what do you think? The logic: when you're using case sensitive completion you know what you want.

costajohnt:
@anki-code @jnoortheen Any update on the direction for this PR?

anki-code:
@costajohnt 
1. we told about removing both env variables: old one and new one.

2. I suggest to test the more complex logic:

If the string in lower case (e.g.`dec`) apply order:
1. case matching prefixes (decoder)
2. case insensitive prefixes (Decoder)
3. case sensitive substrings (MyDecoder)
4. case insensitive substrings (some_decoder)

If the string case secnsitive (e.g. `Dec`) apply order: 
1. case matching prefixes (Decoder)
2. case sensitive substrings (MyDecoder)  -- my assumption that if I'm using case sensitive I know what I want
3. case insensitive prefixes (decoder)
4. case insensitive substrings (some_decoder)

Every group uses substring position to sort results in group.

(If we finalize this PR it will go to 0.23.0)

jnoortheen:
@anki-code I am not in favor of that order. Typing capital letters are mostly two key strokes. How about just matching the `m-d` for `MyDe`.

anki-code:
@jnoortheen ok, let's keep it as is now and see how it will work.
@costajohnt just remove env variables.

anki-code:
@costajohnt I've found issues:

1. 
image

Put all things that started from `_` to the end.

2.
And the order:

image

Here we have all rows lowercase but the order has no sorting by "init" position.

anki-code:
Fix duplications:

The first _cd is `_cd`, second `_cd `(with space)

image

anki-code:
@costajohnt `codecs` shold be the second option and private _ shold be at the end of list. (@jnoortheen are you ok to put private to the end?)

image

jnoortheen:
@anki-code yes, unless the user startswith `_`; the completions should match the input.

jnoortheen:
@anki-code we can handle such special cases in another PR. Shell completions would not be having these characters .

anki-code:
@jnoortheen We have time to fix this until other PRs are in testing. I wanted to make a release at Apr 1 but I see that the changes that are waiting for review so good that I want all of them in 0.23 so I'm waiting for feedback from you and others.

Also I see that @costajohnt is using LLM for this PR so let's have meaningful PR. @costajohnt please fix https://github.com/xonsh/xonsh/pull/6125#issuecomment-4153377332

costajohnt:
Fix applied for https://github.com/xonsh/xonsh/pull/6125#issuecomment-4153377332

Screenshot 2026-03-30 at 9 49 11 PM

anki-code:
Thanks @costajohnt !

PR #6128, opened 2026-03-07
fix: Xonsh hangs when trying to autocomplete captured object with long-running process

### LINKED ISSUE #6127: Xonsh hangs when trying to autocomplete on long-running process capture
## Current Behavior

```xsh
p = !(caffeinate)
p. # Press TAB here
```

Xonsh is frozen and doesn't respond to `Ctrl + C`.

## Expected Behavior

Xonsh is not frozen and stays responsive.

## For community

vadym-shavalda:
Closes #6127

### DIFF
--- a/xonsh/procs/pipelines.py
+++ b/xonsh/procs/pipelines.py
@@ -9,7 +9,6 @@ import subprocess
 import sys
 import threading
 import time
-from contextlib import suppress
 
 import xonsh.lib.lazyasd as xl
 import xonsh.platform as xp


anki-code:
I think it will be breaking change for previous users. What about to treat this in completer.

vadym-shavalda:
@anki-code Yeah, figured as much. And it wouldn't work either way without updating a ton of code. :/
I really wanted to generalize this somehow, but the current patch is the best I could come up with without breaking backwards compatibility. Let me know what you think.

anki-code:
Does this code LLM-generated? 
I'm not fan of `suppress(Exception)`.

vadym-shavalda:
> Does this code LLM-generated?

No. I've used LLMs for discovery, though.

> I'm not fan of `suppress(Exception)`.

That's fine. I've replaced exceptions with default values for `getattr` and `try/except` for `Popen.poll`.

anki-code:
@vadym-shavalda thank you for your effort! I've reused part of it in https://github.com/xonsh/xonsh/pull/6229

The difference between blocking repr and blocking completer is in that in repr you can press Ctrl+C for exit but completer follows to complete hanging so let's keep repr blocking as it was implemented. May be some day we change it but not now.

vadym-shavalda:
@anki-code No probs, glad I could help.

PR #6132, opened 2026-03-10
fix: Fix tests that fail on NixOS

anki-code:
Closes #5569
cc https://github.com/NixOS/nixpkgs/pull/497851

I'm going to fix errors that I see here:

```xsh
docker run --rm -it nixos/nix bash
nix-shell -p python3
python3 -m venv venv
source venv/bin/activate
git clone https://github.com/xonsh/xonsh
cd xonsh
pip install '.[test]'

# to fix tests/test_man.py::test_man_completion[man-exp1] - FileNotFoundError: [Errno 2] No such file or directory: 'col'
nix profile add --extra-experimental-features nix-command --extra-experimental-features flakes 'nixpkgs#util-linux'

python -m pytest
```
Fixed:
```xsh
++ FAILED tests/built_ins/test_tracer.py::test_trace_in_script - assert 1 == 0
++ FAILED tests/platform/test_platform.py::test_githash_value_error - AssertionError: assert 'Mar 10 14:15:14 2026' is None
++ FAILED tests/shell/test_ptk_highlight.py::test_xonsh_lexer_no_win[ls-bin-0] - Failed: Token (Token.Name.Builtin, '/bin/ls') missing: [(Token.Text.Whitespace, ''), (Token.Operator, '/'), (Token.Name.Builtin, 'bin'), (Token.Operator, '/'), (Token.Name, 'ls'), (Token.Text, ' '), (Token.Op...
++ FAILED tests/test_builtins.py::test_repath_backslash - AssertionError: assert set() == {'~/\\w\\w.*'}
++ FAILED tests/test_integrations.py::test_loading_correctly[True] - AssertionError: assert 'AAA /xonsh-dev/xonsh/__init__.py BBB' in 'AAA /xonsh-dev/venv/lib/python3.12/site-packages/xonsh/__init__.py BBB'
++ FAILED tests/test_integrations.py::test_loading_correctly[False] - AssertionError: assert 'AAA /xonsh-dev/xonsh/__init__.py BBB' in 'AAA /xonsh-dev/venv/lib/python3.12/site-packages/xonsh/__init__.py BBB'
++ FAILED tests/test_man.py::test_man_completion[yes-exp0] - FileNotFoundError: [Errno 2] No such file or directory: 'col'
++ FAILED tests/test_man.py::test_man_completion[man-exp1] - FileNotFoundError: [Errno 2] No such file or directory: 'col'
```
Ignored (future PR with fix is welcome! I think it's related to bash completer installation on Nix):
```xsh
[...]

PR #6136, opened 2026-03-11
fix: Fix unstable tests (test_trace_in_script, test_callable_alias_no_bad_file_descriptor)

anki-code:
Fixed test_trace_in_script (cc #6132):
1. We do not need to use path in the output because test can run from different places.
2. Tracing output can be before the output or after.

Fixed test_callable_alias_no_bad_file_descriptor until fix https://github.com/xonsh/xonsh/issues/5631.

## For community


anki-code:
I'm going to do self-merge to have stable tests in main.

PR #6139, opened 2026-03-11
fix: Using dummy history by default in tests

anki-code:
We are using json history by default in tests. This raises silent exceptions because in some tests we're mocking open function or just overheat history backend during automated testing. Also this can slow down tests in case of running on real local environment with huge amount of json history files. In this PR we switch to dummy history by default as it should be.

```xsh
# cd xonsh
# python -m pytest
Exception in thread Thread-2:
  File "/Users/pc/micromamba/envs/xonsh-dev/lib/python3.11/threading.py", line 1045, in _bootstrap_inner
    files = self.files(only_unlocked=True)
.            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pc/Documents/git/xonsh/xonsh/history/json.py", line 279, in files
    lj = xlj.LazyJSON(f, reopen=False)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/pc/Documents/git/xonsh/xonsh/lib/lazyjson.py", line 210, in __init__
    self._load_index()
  File "/Users/pc/Documents/git/xonsh/xonsh/lib/lazyjson.py", line 239, in _load_index
    f.seek(9)
```
## For community

### DIFF
--- a/tests/api/test_os.xsh
+++ b/tests/api/test_os.xsh
@@ -4,23 +4,27 @@ import tempfile
 from xonsh.api.os import indir, rmtree
 
 import pytest
-
+from pathlib import Path
 from xonsh.pytest.tools import ON_WINDOWS
 
+def resolve_path(p):
+    """Path can be a symlink (e.g. on macOS) so we need to resolve it first."""
+    return str(Path(p).resolve())
 
 def test_indir():
     if ON_WINDOWS:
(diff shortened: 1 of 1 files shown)


anki-code:
I'm going to self merge to have stable tests in main.

PR #6143, opened 2026-03-13
feat: Add `@.aliases`

anki-code:
Aliases is a pyhon object so it becomes the part of xonsh session object organically.

cc https://github.com/xonsh/xonsh/pull/5956#issuecomment-4054322076

## For community


anki-code:
I'm thinking about more complex concept.

PR #6145, opened 2026-03-14
feat(ptk): make `tempfile_suffix` ".xsh" for `open_in_editor`

jsquaredosquared:
### Before
When using `C-x C-e` (open in editor keybinding), the tempfile would have the file extension `.txt`.

### After
When using `C-x C-e`, the tempfile would have the file extension `.xsh`. 
This is useful because then editors such as neovim can apply the correct syntax highlighting, lsp, etc (instead of manually running `:set ft=xonsh` each time).

## For community

### DIFF
--- a/xonsh/shells/ptk_shell/key_bindings.py
+++ b/xonsh/shells/ptk_shell/key_bindings.py
@@ -274,6 +274,7 @@ def load_xonsh_bindings(ptk_bindings: KeyBindingsBase) -> KeyBindingsBase:
     @handle(Keys.ControlX, Keys.ControlE, filter=~has_selection)
     def open_editor(event):
         """Open current buffer in editor"""
+        event.current_buffer.tempfile_suffix = ".xsh"
         event.current_buffer.open_in_editor(event.cli)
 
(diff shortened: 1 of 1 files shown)


anki-code:
Thanks @jsquaredosquared !

anki-code:
@jsquaredosquared if you can provide here the example of how to setup neovim config to highlight xsh it will be awesome!

jsquaredosquared:
Hi @anki-code, I use this [syntax highlighting plugin](https://github.com/meatballs/vim-xonsh) (works on vim and neovim) and [this language server](https://github.com/FoamScience/xonsh-language-server).

The way to set them up depends on which plugin manager you use (if any). For example, I use `lazy.nvim`, so to get syntax highlighting all I have to do is add the following to one of my lua plugins files (e.g., `~/.config/nvim/lua/plugins/xonsh.lua`:
```lua
return {
  { "meatballs/vim-xonsh", ft = "xonsh" },
}
```

Soon neovim will have a built-in plugin manager `vim.pack`, so it would be something like:
```lua
vim.pack.add({ src = "https://github.com/meatballs/vim-xonsh.git" })
```

PR #6149, opened 2026-03-18
feat(completer): underline matched substring in completion menu

### LINKED ISSUE #6141: Add underline in completion list
Add underline in completion list for options where substring is in the middle:

Image

## For community

costajohnt:
Highlight the matched portion of completion candidates with underline
styling when the match is a substring (not a prefix). This gives
visual feedback showing where the user's typed text appears within
each completion.

## Usage

When using substring completion matching, the matched portion in
the completion menu is rendered with an underline so you can see
exactly where your typed text appears in each candidate.

Only non-prefix matches are underlined — prefix matches at the
start of the displayed text don't need visual highlighting since
they're already obvious.

## Screenshots

Closes #6141

## For community

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,8 +58,8 @@ jobs:
       - name: Run tests
         if: ${{ !startsWith(matrix.python-version, env.DEFAULT_PYTHON_VERSION) }}
         run: |
-          python -m xonsh run-tests.xsh test -- --timeout=240
+          python -m xonsh run-tests.xsh test -- --timeout=600
       - name: Run tests with coverage
         if: ${{ startsWith(matrix.python-version, env.DEFAULT_PYTHON_VERSION) }}
         run: |
-          python -m xonsh run-tests.xsh test --report-coverage -- --timeout=240
+          python -m xonsh run-tests.xsh test --report-coverage -- --timeout=600
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -5,7 +5,7 @@
   "bump-patch-for-minor-pre-major": true,
   "include-v-in-tag": false,
   "release-search-depth": 1,
-  "commit-search-depth": 100,
+  "commit-search-depth": 1000,
   "packages": {
     ".": {
       "changelog-path": "CHANGELOG.md",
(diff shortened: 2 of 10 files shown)


anki-code:
I'm trying to test using [docker-xonsh-branch](https://github.com/anki-code/xonsh-developer-toolkit/blob/bc0b3f17e205faba0a464613919ebd5979b960c9/dev.xsh#L18) but have no result:

image

costajohnt:
Thanks for testing! The issue was that this PR was branched from main, which still uses startswith matching. Since all completions are prefix matches with startswith, the underline code correctly skips them (prefix matches don't need highlighting since the match is already at the start of the displayed text).

I've rebased this branch onto #6125 (the substring matching PR) so the two features work together. With substring matching active, typing `deploy` and pressing Tab returns completions like `dev-xonsh-deploy` where "deploy" appears in the middle, and the underline highlight makes that visible in the menu.

This PR depends on #6125 being merged first.

anki-code:
Ok, let's keep it as draft and I will test after merging 6125.

anki-code:
@costajohnt is it possible to continue here? Because of https://github.com/xonsh/xonsh/pull/6125 was merged.

costajohnt:
Sure @anki-code. Updated the PR and retested. Things are looking good to me. LMK what you think.

Screenshot 2026-04-01 at 12 33 59 PM
Screenshot 2026-04-01 at 12 23 56 PM
Screenshot 2026-04-01 at 12 22 06 PM

anki-code:
looks nice, 
1. we're not using classes in tests
2. fix import case:
    image

costajohnt:
dotted prefix got me again. fixed that and the classes in tests issue.

Screenshot 2026-04-01 at 2 49 40 PM

anki-code:
@jnoortheen do you like how it looks? What do you think?

jnoortheen:
@anki-code looks good!

anki-code:
@costajohnt two things:

1. Duplicated FormattedText construction

  The two branches (full-text match and visible-prefix fallback) build parts identically. Extract a helper:

  def _underline_span(text, start, end):
      parts = [("", text[:start]), ("underline", text[start:end])]
      if end < len(text):
          parts.append(("", text[end:]))
      return FormattedText(parts)

  2. Missing test cases

  - _highlight_match("", "abc", "abc", 3) — empty display_text
  - _highlight_match("Foo", "xFoo", "foo", 1) — case-insensitive match landing at display position 0 (should NOT underline)

anki-code:
Thanks @costajohnt!

PR #6159, opened 2026-03-21
fix!: Processes: Major improvements to core logic for threads, pipes, file descriptors, and process output management across all platforms.

anki-code:
This is perhaps the deepest improvement to the kernel since https://github.com/xonsh/xonsh/pull/3897. 

### Motivation

Callable alias is awesome thing! The output capturing that is implemented in it is amazing! But it requires more careful stdout/stderr management in case of using it with threads and pipes. We have issues that raises I/O errors and "Bad file descriptor" exceptions because pipe fds are not protected from multiple use/closing and race conditions. This PR is to significantly improve the situation.

### Changes

1. Fixed fd (pipe fd) leaking by introducing PipeChannel - single-owner pipe fd manager.
    * No I/O errors after running callable aliases in pipe.
    * No "Bad file descriptor" in any component after running callable aliases.
    * No I/O issues with using `execx()` (`ExecAlias`) in callable alias.

2. Fixed usage of STDOUT_DISPATCHER to prevent restoring std in parallel threads.
    * No unexpected stdout/stderr leaking during usage of callable alias.
    * Captured stdout/stderr are stable.

3. Fixed premature closing of descriptors if a process in a pipe terminated or was interrupted (including the last process, which is the main one according to the current logic).
    * No unexpected I/O errors after running pipes with prematurely closed processes.

4. Fixed signal management especially for captured object `!()`
    * Ability to cancel pipeline any time.
    * Careful stdout management in background process.

6. Massive Windows fix:
    * Reduced outdated code.
    * Fix fd leak (was made by additional Handler that was removed now).

5. Now the fd leaking integration stress test is working for Linux, Mac and Windows.

As result callable aliases are working much more stable: with stdout/stderr, in pipes, with whole environment.

### Demo test

In this example callable aliases `a` and `b` have unexpected finish time, child subprocesses with output and piping that produces race condition:

```xsh
@aliases.register
def _a():
    echo 1 && echo 2

@aliases.register
def _b():
    echo 3 && echo 4

for i in range(100):
    $(a | b)

# BEFORE:
# I/O errors
# Bad file descriptor
# Exceptions and errors
# Unstable output.
# AFTER:
# No any error, stable output.
```

### Issues

Fixed:
* Fixes https://github.com/xonsh/xonsh/issues/5631
* Fixes https://github.com/xonsh/xonsh/issues/5647
* Fixes https://github.com/xonsh/xonsh/issues/5648
* Fixes https://github.com/xonsh/xonsh/issues/5482
[...]

PR #6163, opened 2026-03-23
fix: Windows: fix unstable tests

anki-code:
I ran xonsh tests on Windows 11, Python 3.11 and fix all tests with errors.

Fixed:
* test_complete_command - we need to use temp dirs in the tests.
* test_complete_command - added windows support.
* test_imphook - windows `echo` has different behaviour than GNU echo.
* test_on_command_not_found_replacement - on windows `echo` is `cmd /c echo`.
* test_remove_hide_escape - windows have no support escape in `cmd /c echo`.

## For community

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -42,7 +42,7 @@ Basics
 =======================
 The xonsh language is based on Python, and the xonsh shell uses Python to
 interpret any input it receives. This makes simple things simple and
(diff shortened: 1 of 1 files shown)


anki-code:
@jnoortheen it's ready to review and merge. Now anybody can run tests on Windows without unexpected errors.

anki-code:
I'm going to do self merge to reuse it in #6159

PR #6166, opened 2026-03-24
feat: Added support for subprocess substitution in the middle of a string e.g. `echo prefix_$(whoami)_suffix` or `echo path/$(@lines echo '1\n2').suffix`

### LINKED ISSUE #5713: why `echo x=$(pwd)` is SyntaxError
`echo x=@(123)` works.
`echo x=@( $(pwd) )` works.

Any reason why `echo x=$(pwd)` not work? Does it cause any syntax ambiguity?
Also `echo x=()` not work. Maybe of the same reason?

anki-code:
### Motivation

Fixing the issue that mentioned by @ambv in [The xonsh cheat sheet](https://lukasz.langa.pl/8909455e-e824-4b1b-9bfd-8e9fdc156325/):
```xsh
echo this does$(pwd) not
# SyntaxError
```

Also closes #5713

### Before
```xsh
echo $(echo 1)
# 1

echo prefix$(echo 1)
# SyntaxError
echo prefix$(echo 1)suffix
# SyntaxError
echo $(echo 1)suffix
# SyntaxError
```

### After

```xsh
echo prefix$(echo 1)
# prefix1
echo prefix$(echo 1)suffix
# prefix1suffix
echo $(echo 1)suffix
# 1suffix
```

Moreover, using command decorator makes additional use case:

```xsh
showcmd path/$(@lines echo '1\n2\n3').suffix
# ['path/1.suffix', 'path/2.suffix', 'path/3.suffix']
```

## For community

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -2017,10 +2017,10 @@ written in pure Python.
     Convert a number or string to an integer, or return 0 if no arguments
     are given.  If x is a number, return x.__int__().  For floating point
     numbers, this truncates towards zero.
-    
+
     @ @.imp.json.loads??
     def loads(s, *, cls=None, object_hook=None, parse_float=None,
-        parse_int=None, parse_constant=None, object_pairs_hook=None, **kw):    
+        parse_int=None, parse_constant=None, object_pairs_hook=None, **kw):
 
     @ @.imp.json?.loads?
     <json help>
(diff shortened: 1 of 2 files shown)


anki-code:
Thanks @jnoortheen !

ambv:
Thanks, @anki-code. I'll update the cheatsheet! 😎

anki-code:
@ambv please replace `${...}` to `@.env` in the cheatsheet as well. We don't like this outdated syntax now ;)

ambv:
Will do! I'll provide an explanation as to why the old form is discouraged if you care to explain.

anki-code:
We strive to make the syntax clean, logical and short. The old one is so ugly so we have new concept with `@`-object that represents xonsh session. Now you can use `@.env`, `@.imp`, `@.history` and it looks and sounds much clean: xonsh session env/importer/history.

(I also recommend to put the link to the [big cheatsheet](https://github.com/anki-code/xonsh-cheatsheet) with many use cases and tricks.)

PR #6169, opened 2026-03-25
fix: Env: Fix empty path in PATH

anki-code:
During testing Windows I've found that $PATH has empty path because it has `/path;/path;` structure and the last part interpreted as empty path.

### Before

Windows: 
```xsh
$PATH
# EnvPath(['c:', 'd:', ''])
```

### After

Windows:
```xsh
$PATH
# EnvPath(['c:', 'd:'])
```

## For community


anki-code:
I'm going to self merge this becase we have tests coverage and I need this on Windows and the fix is small.

PR #6178, opened 2026-03-26
fix(Tests): Fix unwanted output and exceptions from tests

anki-code:
Fixed:
* There are eceptions during tests that produced because some tests do not close std properly.
* In resolve_redirects we had issue in logic with checking the empty command.

+1 to tests stability.

## For community


anki-code:
Self merge to have these fixes in main.

PR #6179, opened 2026-03-26
fix: Windows Completer: Fixed raw paths completion

### LINKED ISSUE #4557: Path autocompletion fails in raw strings (Windows)
I have two test paths:
`C:\App\x\a\inA`
`C:\App\x\b\inB`

When I <kbd>Tab</kbd> after `C:\App\x\`, I get both `a\` and `\b` suggested
When I <kbd>Tab</kbd> after `r"C:\App\x\`, my path changes to `\\` and I get no suggestions (empty box), the second <kbd>Tab</kbd> reverts path to `r"C:\App\x\`, but the suggestion box is still empty

## xonfig

<details>

```
$ xonfig
| xonsh            | 0.11.0               |
| Git SHA          | 337cf25a             |
| Commit Date      | Nov 17 15:37:41 2021 |
| Python           | 3.9.7                |
| have readline    | False                |
| prompt toolkit   | 3.0.22               |
| pygments         | 2.10.0               |
| on posix         | False                |
| on windows       | True                 |
| xontrib          | []                   |
| RC file          | []                   |
```

</details>

## Expected Behavior

Autocomplete within raw strings (needed for paths with spaces) works the same as without them

## Current Behavior

Autocomplete fails within raw strings

## Steps to Reproduce

Please see a simple test above

## For community

anki-code:
Fixes:
* Fixes #4557
* Fixes https://github.com/xonsh/xonsh/issues/983
* Fixes https://github.com/xonsh/xonsh/pull/6188
* Removed bash and man completers for windows (https://github.com/xonsh/xonsh/commit/59a915baf40883197fd66d8e9d78bafba2205e50) 

Raw strings can't end with \ before closing quote (e.g. r"path\" is a SyntaxError). Use / as the trailing directory separator instead.

## Before
```
cd r'c:\W<Tab>
cd \\  # bug
```

## After

```
cd r'c:\W<Tab>
cd r'c:\Windows\\'

```

## For community

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ ci:
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: 'v0.15.7'
+    rev: 'v0.15.8'
     hooks:
       - id: ruff
         args: [., --fix, --exit-non-zero-on-fix]
--- a/docs/contents.rst
+++ b/docs/contents.rst
@@ -60,7 +60,6 @@ Creator
     tutorial_events
     tutorial_completers
     tutorial_history_backend
-    Xonsh Community <https://xonsh.zulipchat.com/join/hbvue5rimpdkwkdjuiqfs7tv/>
 
 Xontributor
 -----------
@@ -72,7 +71,6 @@ Xontributor
     devguide
     api/index
     Xonsh on Github <https://github.com/xonsh/xonsh>
-    Xonsh Zulip Community <https://xonsh.zulipchat.com/join/hbvue5rimpdkwkdjuiqfs7tv/>
 
 
 Helpful Links
@@ -81,6 +79,7 @@ Helpful Links
 * :ref:`genindex`
 * :ref:`modindex`
 * :ref:`search`
+* `Xonsh Community <https://xonsh.zulipchat.com/join/hbvue5rimpdkwkdjuiqfs7tv/>`_
 * `Follow us on Mastodon <https://mastodon.online/@xonsh>`_
 * `Want a t-shirt? <https://www.hellotux.com/xonsh>`_
 
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -96,7 +96,7 @@ such as Miniconda or Micromamba, should be used.
 macOS / Linux / WSL
 -------------------
 
-Install independent Xonsh with Micromamba:
+Install Xonsh independently using Micromamba:
 
 .. code-block:: console
 
@@ -109,7 +109,7 @@ Learn more: `Mamba installer <install_mamba.html>`_.
 Windows
 -------
 
-Install independent Xonsh on Windows (no admin rights):
+Install Xonsh independently on Windows (without admin rights):
 
 .. code-block:: console
 
--- a/tests/aliases/test_xexec.py
+++ b/tests/aliases/test_xexec.py
@@ -93,3 +93,92 @@ def test_c_switch(monkeypatch):
     command = "sleep"
     xexec(["-c", command, "1"])
     assert called["env"] == {}
+
+
+def test_alias_stack_cleared(monkeypatch, xession):
+    """exec must not pass __ALIAS_STACK to the new process.
+
+    Regression test for #5216 / #5709: when a xonsh script does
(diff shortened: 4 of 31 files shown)


anki-code:
Tests passed.

@eugenesvk @Spitfire1900 could you please confirm it's working to you?
To install from branch:
```xsh
pip install -U --force-reinstall git+https://github.com/anki-code/xonsh@anki_win_fix_path
```

Spitfire1900:
Seems to respond very quickly 👍 🚀 


`uv tool install 'git+https://github.com/anki-code/xonsh@anki_win_fix_path[full]' --python=3.13 --with-requirements U:\xonsh-requirements.txt`


```
typing-extensions
xontrib-base16-shell
xontrib-commands
xontrib-jump-to-dir
xontrib-prompt-bar
xontrib-prompt-vi-mode
xontrib-pygitstatus
xontrib-sh
xontrib-vox
pip
```

eugenesvk:
This part seems to be fixed, but another one isn't: from
`R"D:\App\4dev\xonsh` pressing tab I get
`rR"D:\App\4dev\xonsh/"`

for some reason I get two `rR` instead of the original `R` being preserved (

anki-code:
@eugenesvk I've fixed in this branch, please try again

anki-code:
@eugenesvk please test this. I replaced `/` (previous solution) to `\\` to have more Windows-native string because not all apps support `/`.
So instead of
```
cd r'c:\w<tab>
cd r'c:\windows/'
```
now we have:
```
cd r'c:\w<tab>
cd r'c:\windows\\'<tab>
cd r'c:\windows\System32\\'<tab>
cd r'c:\windows\System32\bin\\'
# etc
```
Please test.

anki-code:
@eugenesvk I've found additional issue with completer - spawning `bash` and `man` completers can slow down completion as well as disabling prompt-toolkit arrow keys state (it looks like hanging but it is just disabled arrow keys).
It's fixed in this PR (https://github.com/xonsh/xonsh/pull/6179/commits/59a915baf40883197fd66d8e9d78bafba2205e50).
Please test it.

eugenesvk:
Seems to be working better, though the slashes in path should be left untouched, currently the autocompleter replaces everything with `\`

Also for some reason it doesn't work with `r".\`, turns into `r"..\` instead of enumerating paths in the current dir

anki-code:
> Also for some reason it doesn't work with r".\, turns into r"..\ instead of enumerating paths in the current dir

I can't reproduce this on main.

image

image


> Seems to be working better, though the slashes in path should be left untouched, currently the autocompleter replaces everything with \

Why we need to keep slashes?

eugenesvk:
> I can't reproduce this on main.

You're not reproducing `R".\`, you have `cat`. Strangely enough, `R"./` works

> Why we need to keep slashes?

To preserve user's intent expressed via typing and not pollute history with `\`? For example, I try not to use those since they require escaping in many places, so I can't easily copy&paste command from history if autocompleter replaces all my typed `/` with `\`
For the same reason, it's best to allow the separator for autocompletion be configurable, I'd switch that to `/` as well, will also remove the ugly `\\"` at the end

anki-code:
I have no issues with completion both sequences:
image

image

> To preserve user's intent

Makes sense. I will take a look.

anki-code:
@eugenesvk posix paths already implemented - https://xon.sh/envvars.html#force-posix-paths

After `$FORCE_POSIX_PATHS=1`:
image

anki-code:
So it looks we're finished here.

eugenesvk:
That forces one way, but instead it should preserve what users input. For example, I use `/`, but then if I follow some snippet/tutorial that use `\` and autocomplete, I won't be able to later find that command in my scrollback because my history was tainted by the `$FORCE_POSIX_PATHS`, so if I copy the command again and Ctrl+F search for it, it fails

The first issue was caused by `aliases['..'] = ['cd', '..']`, apparently autocompleter expands aliases first? But why does it only do that on `\` so `R"dif/` doesn't expand, but `R"dif\` expands to `diff` since I have an alias of `diff`

anki-code:
Please report the first one as a feature request. The second one as a completer issue.

PRs are welcome! LLM is working good with xonsh code.

PR #6184, opened 2026-03-27
fix(Performance): Reduce expensive operations around getting VCS branch on every prompt

anki-code:
I found this during work on 6183

### Before

On every prompt we run:
* `git symbolic-ref` - expensive in thread
*  `hg` - not found on 99% systems but it takes scanning all PATHs 
*  `fossil` - not found on 99% systems but it takes scanning all PATHs 

### After

After:
* git: check `.git` and parents (very fast) before running `git symbolic-ref`.
* Check `hg` and `fossil` once on start.

## For community


anki-code:
I'm going to merge this because it's much faster than previous approach.

PR #6190, opened 2026-03-27
fix(Mac): Fix `xonsh.platforms.sysctlbyname` returns bytes instead of string when return_str=True

### LINKED ISSUE #6189: fix: sysctlbyname returns bytes instead of string when return_str=true
## Description

When `return_str=True` (the default), the function returns `buf.value` which is 
a bytes object in Python 3, not a str. Users calling this function expecting a 
string will get bytes, causing TypeError in downstream code that expects str.

**Severity**: `medium`
**File**: `xonsh/platforms/macutils.py`

## Expected Behavior

The code should handle this case properly to avoid unexpected errors or degraded quality.

Manhhoangvp95:
### Problem
When `return_str=True` (the default), the function returns `buf.value` which is 
a bytes object in Python 3, not a str. Users calling this function expecting a 
string will get bytes, causing TypeError in downstream code that expects str.

**Severity**: `medium`
**File**: `xonsh/platforms/macutils.py`

### Solution
Change to: `return buf.value.decode() if return_str else buf.raw`

### Changes
- `xonsh/platforms/macutils.py` (modified)

## For community

Closes #6189

### DIFF
--- a/xonsh/platforms/macutils.py
+++ b/xonsh/platforms/macutils.py
@@ -18,6 +18,6 @@ def sysctlbyname(name, return_str=True):
     # Re-run, but provide the buffer
     LIBC.sysctlbyname(name, buf, byref(size), None, 0)
     if return_str:
-        return buf.value
+        return buf.value.decode()
     else:
         return buf.raw


anki-code:
Thanks @Manhhoangvp95 !

Issue #5253, opened 2023-12-20
Parser fail with numerical and logical AND


anki-code:
```xsh
xonfig
# xonsh 0.14.3

mkdir -p /tmp/abc /tmp/123 /tmp/a123

cd /tmp/abc && ls
# Result: empty
# Expected: empty


cd /tmp/a123 && ls
# Result: empty
# Expected: empty

# But:

cd /tmp/123 && ls
# Result: NameError: name 'cd' is not defined
# Expected: empty

echo / 123 && ls
# Result: NameError: name 'echo' is not defined
# Expected: `/ 123 ...`
```

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

JamesParrott:
I could recreate this in an alpine:edge container:

```
<user>@bf5102635826 / # mkdir -p /tmp/abc /tmp/123 /tmp/a123
<user>@bf5102635826 / # cd /tmp/abc && ls
<user>@bf5102635826 /tmp/abc # cd /tmp/a123 && ls
<user>@bf5102635826 /tmp/a123 # cd /tmp/123 && ls
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK = True
NameError: name 'cd' is not defined
<user>@bf5102635826 /tmp/a123 [1] # echo / 123 && ls
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK = True
NameError: name 'echo' is not defined
```

And with XONSH_SHOW_TRACEBACK = True:

```
<user>@bf5102635826 / # cd /tmp/123 && ls
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'cd' is not defined

<user>@bf5102635826 / [1] # echo / 123 && ls
xonsh: To log full traceback to a file set: $XONSH_TRACEBACK_LOGFILE = <filename>
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'echo' is not defined
```

JamesParrott:
I'm new to Xonsh.  But I believe the correct behavious would be to change the `&&` into an and.  
https://xon.sh/tutorial.html#logical-subprocess-and

These errors occur with `||` too.  And when replacing `&&` with `and`, the echo treats the and as an arg, but the cd raises a different error:

```
[...]

PR #6193, opened 2026-03-28
fix(Parser): Fix parsing path with num in case `cd /tmp/123 && ...`

anki-code:
Fixes https://github.com/xonsh/xonsh/issues/5253

### Before
```xsh
cd /tmp/123 && ls
# NameError: name 'cd' is not define
echo / 123 && ls
# NameError: name 'echo' is not define
```

### After
```xsh
cd /tmp/123 && ls
# <cd and ls result>
echo / 123 && ls
# / 123
# <ls result>
```

## For community


anki-code:
Self merge as trivial fix.

Issue #5839, opened 2025-05-24
Compiling xsh script never finish if wrong space put after "\" for linebreak and before a line with "\" at the end


MajoranaOedipus:
## Current Behavior
Sorry for a long title, the way to reproduce is simple:

Create a xsh file: 
```
# test.xsh
ls\ 
    -l\
    .
```

Please notice that at the first line `ls\ `, there is a wrongly put space after `\`.

And if the script is run via:
```shell
xonsh --no-rc --no-env test.xsh
```

The compilation never ends.

If it gets Ctrl+C, the trace back shows that the Xonsh is still running tokenizer:

<details>

```xsh
^CTraceback (most recent call last):
  File "/usr/bin/xonsh", line 8, in <module>
    sys.exit(main())
             ~~~~^^
  File "/usr/lib/python3.13/site-packages/xonsh/main.py", line 512, in main
    sys.exit(main_xonsh(args))
             ~~~~~~~~~~^^^^^^
  File "/usr/lib/python3.13/site-packages/xonsh/main.py", line 581, in main_xonsh
    exc_info = run_script_with_cache(
        args.file, shell.execer, glb=shell.ctx, loc=None, mode="exec"
    )
  File "/usr/lib/python3.13/site-packages/xonsh/codecache.py", line 176, in run_script_with_cache
    ccode = compile_code(filename, code, execer, glb, loc, mode)
  File "/usr/lib/python3.13/site-packages/xonsh/codecache.py", line 134, in compile_code
    ccode = execer.compile(code, glbs=glb, locs=loc, mode=mode, filename=filename)
  File "/usr/lib/python3.13/site-packages/xonsh/execer.py", line 131, in compile
    tree = self.parse(input, ctx, mode=mode, filename=filename, transform=transform)
  File "/usr/lib/python3.13/site-packages/xonsh/execer.py", line 87, in parse
    tree, input = self._parse_ctx_free(input, mode=mode, filename=filename)
                  ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/xonsh/execer.py", line 344, in _parse_ctx_free
    return _try_parse(input, greedy=False)
  File "/usr/lib/python3.13/site-packages/xonsh/execer.py", line 235, in _try_parse
    tree = self.parser.parse(
        input,
    ...<2 lines>...
        debug_level=(self.debug_level >= 2),
[...]

PR #6194, opened 2026-03-28
fix(Parser): Fix loop on `\ ` sequence

anki-code:
Fixes https://github.com/xonsh/xonsh/issues/5839

### Before

test.xsh with `\ `
```xsh
ls\ 
    -l\
    .
```

`xonsh test.xsh` - hanging

### After

`xonsh test.xsh` - Syntax error as expected (same as Python behavior).

## For community


anki-code:
Self merge trivial fix.

PR #6180, opened 2026-03-26
fix!: Windows Executor: Fixed executing the scripts and binary files

anki-code:
### Motivation

Fixes: 
* Fixes #4555
* Fixes https://github.com/xonsh/xonsh/issues/5109
* Fixes https://github.com/xonsh/xonsh/issues/4558
* Partially https://github.com/xonsh/xonsh/issues/4481

### Before

Running scripts just not working properly on Windows. Probably was broken in #4077.

### After

Update: The examples below are modified to represent https://github.com/xonsh/xonsh/pull/6208 as well.

```xsh
mkdir ~/tmp
mkdir ~/tmp/dir
mkdir ~/tmp/bin

cd ~/tmp/dir
echo 'print(0)' > fileunknown
echo 'print(1)' > filexsh.xsh
echo 'print(2)' > filepy.py
echo 'MsgBox 123' > filevbs.vbs

cd ~/tmp/bin
echo 'print(0)' > binfileunknown
echo 'print(1)' > binfilexsh.xsh
echo 'print(2)' > binfilepy.py
echo 'MsgBox 123' > binfilevbs.vbs


# Case 1. No PATH, no PATHEXT, call filename.

$PATH = []
$PATHEXT = ['.xsh', '.py']

cd ~/tmp/dir

# With path prefix
./fileunknown  # Error (unknown file)
./filexsh.xsh  # Run in xonsh (known ext)
./filepy.py    # Run using shebang (known ext)
./filevbs.vbs  # Error (not in PATHEXT)

# With path by name
./fileunknown  # Error (unknown file)
./filexsh  # Run in xonsh (known ext)
./filepy    # Run using shebang (known ext)
./filevbs  # Error (not in PATHEXT)

# No path prefix
fileunknown  # Error (not in PATH and not in PATHEXT)
filexsh.xsh  # Error (not in PATH)
filepy.py    # Error (not in PATH)
filevbs.vbs  # Error (not in PATH and not in PATHEXT)

# Names
fileunknown # Error (not in PATH and not in PATHEXT)
filexsh     # Error (not in PATH)
filepy      # Error (not in PATH)
filevbs     # Error (not in PATH and not in PATHEXT)


# Case 2. No PATH, with PATHEXT

$PATH = []
$PATHEXT = ['.VBS','.xsh', '.py']

# With path prefix
./fileunknown  # Error
./filexsh.xsh  # Run in xonsh (known ext)
./filepy.py    # Run using shebang (known ext)
[...]

PR #6208, opened 2026-03-31
chore: Revert unreleased PATHEXT implicit behavior

anki-code:
After thinking about https://github.com/xonsh/xonsh/pull/6180 I see that using .xsh and .py implicitly (like ps1 in PowerShell) is not a good idea. So I'm going to revert this and leave it for the future.

## For community


anki-code:
cc @eugenesvk, I'm not ready to release implicit support of .xsh and .py so do things explicitly e.g.
```xsh
# ~/.xonshrc
$PATHEXT.insert(0, '.XSH')
```
until we will have better idea.

eugenesvk:
Why not make it "explicit" and prepend it to pathexti instead of relying on internal logic, so that users can remove it in their configs? (just xsh, not all py)

anki-code:
After thinking about and trying different options, it turned out that this part is just as tricky as many parts of a command shell. It’s better to leave it as it is than to build a complex hidden logic system that we would later have to support and debug in edge cases. As soon as you add some implicit logic, it quickly grows into dozens of use cases.

After working on the shell for four years, I now understand much better the team who built it in the first place: edge cases and non-trivial logic in OS behavior, traditional shells legacy algorithms and compromises in integrations show up everywhere, and sometimes it’s really better to leave things unresolved and just provide a stable core.

Coming back to the PATHEXT logic. There are several factors that affect how this should be done, and they need careful and structured thinking:

* Security. PowerShell does not run .ps1 files by their name you have to type the extension.
* Reasonable design. On Linux, there is the +x flag and shebangs to mark executable files and how they should run. On Windows, PATHEXT gives access to all kinds of files without clear distinction.
* Recursive behavior. If every xonsh session adds its own extensions to PATHEXT, we would need logic to remove duplicates and to check when it is okay to add them.
* Explicit control. What if someone wants to disable running .xsh files by name, for a secure or restricted environment. We need to support that too.

As a result, instead of just relying on the current PATHEXT value or just doing `$PATHEXT.append('.xsh')` if it's really need, we end up with a lot of hidden logic. Personally, I’m not ready to think about and maintain all of this right now, especially since I’m not a Windows user.

For tasks like this, we need well-prepared PEP proposals that clearly describe the approach, the implementation, and cover the use cases of many different people. Anybody can propose this!

eugenesvk:
> PowerShell does not run .ps1 files by their name you have to type the extension

It does? That's what you've supposedly investiated in the other PR as we discussed that it? That's like the whole point of comparison?

Also, what's the complex logic of adding a value to an environment variable?

> If every xonsh session adds its own extensions to PATHEXT, we would need logic to remove duplicates and to check when it is okay to add them.

We don't need to remove anything, just a simple check if it exists
`$PATHEXT.insert(0,'.xsh') if '.xsh' not in $PATHEXT else None`

> Explicit control. What if someone wants to disable running .xsh files by name, for a secure or restricted environment. We need to support that too.

We do support that by requiring a user configuring `$PATHEXT.remove('.xsh') if '.xsh' in $PATHEXT else None` if needed

The issue is entirely about the **default**, not about eliminating an option!

> Reasonable design. On Linux, there is the +x flag and shebangs to mark executable files and how they should run. On Windows, PATHEXT gives access to all kinds of files without clear distinction

The [definition is identical](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7.6), could you specify what distinction is not clear?:
> The $Env:PATHEXT variable contains a list of file extensions that Windows considers to be executable files

> we need well-prepared PEP proposals that clearly describe the approach, the implementation, and cover the use cases of many different people

Is there a single example for any of the previous changes?

anki-code:
Sorry I have no time on discussing this. I made a huge effort to prepare [0.23](https://github.com/xonsh/xonsh/pull/6152) and gave attention to Windows too so I'm happy that some things now working on Windows and users have a path to do things. Not perfect but someday someone else will improve it. May be it will be me again :)

eugenesvk:
Someone else has already improved it years ago, you just rejected those improvements until this huge effort that had to be reverted. So what's your "path to do things" exactly? How is a PEP going to fix lack of time for a discussion?

anki-code:
Your sarcasm and dissatisfaction make it sound as if I owe you something, but I’m just a developer like you. Perhaps you’re still upset with me for not accepting your changes at some point. Let me recall.

I remember that your PRs contained a lot of agressive caching — for example, you were trying to achieve faster shell performance by storing many things in memory. At the same time, the implementation had weak or no test coverage and questionable logic. It worked for you personally, and your shell ran instantly — and maybe it still does. That’s fine.

I don’t feel that I acted incorrectly. With the current progress of LLM models, I can suggest that you use AI to discuss how your solutions and approaches align with the logic of a command-line shell, as well as their pros and cons. You might discover something new, and it could help you prepare more thoroughly developed proposals. You have good intentions and a solid approach, and I believe that a capable LLM can help you take a more systematic approach to designing and testing changes.

eugenesvk:
> Perhaps you’re still upset with me for not accepting your changes at some point. 

I was referencing a **different** PR about **this** specific **issue** that allowed using `.xsh` files on Windows without an error, not the caching one

> Let me recall.

You can just read it again instead of making wrong recollections

> I remember that your PRs contained a lot of unjustified caching — for example, you were trying to achieve faster shell performance by storing many things in memory.

- Where did you raise this concern? Seems like the first time you mention it https://github.com/xonsh/xonsh/pull/5652#issuecomment-2275465662
- That also makes no sense, you can't have "unjustified caching" if there is no caching! All those caching dirs were set by the **USER**, the defaults were empty, so if you didn't want to use memory for significant speedups, you could just... not do anything
- Also, have you actually measured memory use to make a claim about "many things"? How does it compare to the current caching?

> At the same time, the implementation had weak or no test coverage and

There were tests, so again, instead of making stuff up, read your own words!

Re. "weak", maybe, but 

That was your stance on "test coverage"
> I don't like tests with comparing timings. It's meaningless

And also no response on what you wanted to test for a performance improvement other than timing.

> questionable logic

Like what?

> LLM models ... You might discover something new, and it could help you prepare more thoroughly developed proposals.

I see how you've handled it in the last batch of vibe coding where issues are just being closed without any resolution and lengthy comments that are contradictory after you've "learned how PowerShell launches .ps1 without an extension and then learned how PowerShell requires an extension to launch scripts".
And major changes are landed willy nilly after a few days with no code review.

And there was nothing thorough in your proposal, for example, for caching you're still under the misconception that caching Windows dir is good because it's huge, so you've added it to the defaults. But I've pointed out to you personally several times that it's not the reason for the slowdown, the issue is many checks per file per dir regardless of how many files there are in a dir.
Ask your favorite LLM to read previous discussions if you forgot it yourself, maybe it can help you being thorough!

> use AI to discuss how your solutions and approaches align with the logic of a command-line shell, as well as their pros and cons.

So where is your thorough proposal after such a discussion about the "pros and cons" of a broken highlighting solution that can't keep up with typing?

> that a capable LLM can help you take a more systematic approach to designing and testing changes.

But you've shown neither!!! Again, why are you suggesting a broken process?

anki-code:
So much criticism.

I think the best way is to ask someone else review your PRs. I like it.

PR #6211, opened 2026-04-01
fix(readline): resolve prefix swallowing by enforcing prefix safety and boundary alignment

### LINKED ISSUE #6209: readline: completion is not working for partial
```xsh
xonsh --no-rc -st readline
# @.imp.jso<Tab> 
# @.imp.                             # prefix removed
# @.imp.json.de<Tab>
# @.imp.json.                     # prefix removed
```
## For community

Nesar21:
## Description
This PR fixes the prefix swallowing bug in the readline shell where completion on certain partials (like `@.imp.jso`) would incorrectly delete the prefix instead of completing it.

## Root Cause
Readline's backend replaces the current word with the "longest common prefix" of all returned completion candidates. Xonsh's completion engine uses substring matching (e.g., `"jso" in "_json"` is True), which returns valid substring matches alongside exact prefix matches. When Readline calculates the common prefix between an exact match (e.g., `@.imp.json`) and a substring match (e.g., `@.imp._json`), the common prefix becomes *shorter* than the typed word (e.g., `@.imp.`), causing Readline to pull the text backwards and "swallow" the prefix.

## The Fix
- Implemented a **Readline Prefix Safety Filter** in `completedefault` that enforces strict prefix-only matching at the Readline boundary.
- Improved **Boundary Alignment** to handle edge cases where Xonsh's `plen` and Readline's `readline_plen` disagree (padding vs. trimming).
- Safely preserved `RichCompletion` metadata by replacing string concatenation with `.replace()`, removing the flawed `_render_completions` utility entirely.

> Parts of this fix were developed with AI/LLM assistance. The logic has been manually verified and tested.

## Before vs After
**Before:** `@.imp.jso<Tab>` → `@.imp.` (prefix removed)
**After:** `@.imp.jso<Tab>` → `@.imp.json` (correct)

## Tests
Added regression tests in `tests/shell/test_readline_shell.py`:
- `test_completedefault_prefix_safety` — verifies substring matches are filtered out
- `test_boundary_alignment` — verifies plen padding/trimming preserves RichCompletion metadata

All 7 tests passing locally.

Closes #6209

Screenshot 2026-04-01 at 14 32 48
Screenshot 2026-04-01 at 12 49 58
Screenshot 2026-04-01 at 12 50 03

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,8 +58,8 @@ jobs:
       - name: Run tests
         if: ${{ !startsWith(matrix.python-version, env.DEFAULT_PYTHON_VERSION) }}
         run: |
-          python -m xonsh run-tests.xsh test -- --timeout=240
+          python -m xonsh run-tests.xsh test -- --timeout=600
       - name: Run tests with coverage
         if: ${{ startsWith(matrix.python-version, env.DEFAULT_PYTHON_VERSION) }}
         run: |
-          python -m xonsh run-tests.xsh test --report-coverage -- --timeout=240
+          python -m xonsh run-tests.xsh test --report-coverage -- --timeout=600
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,7 +27,7 @@ from xonsh import __version__ as XONSH_VERSION
 from xonsh.environ import Env, Var, Xettings
 
 if tp.TYPE_CHECKING:
-    from xonsh.environ import VarKeyType
+    pass
 import xonsh.main as xmain
 
 xmain.setup()
@@ -280,7 +280,7 @@ class VarDoc(tp.NamedTuple):
 
 
 class EnvVarGroup(tp.NamedTuple):
-    vars: tp.Dict["VarKeyType", VarDoc]  # sorted variables in the section
+    vars: tp.Dict[str, VarDoc]  # sorted variables in the section
     children: tp.Dict[Xettings, "EnvVarGroup"]  # child sections
 
 
--- a/docs/contents.rst
+++ b/docs/contents.rst
@@ -70,6 +70,7 @@ developer tools.
     python_virtual_environments
     xonsh_projects
     events_tutorial
+    env
     completers
     xontrib
     Create an extension using template <https://github.com/xonsh/xontrib-template>
--- a/docs/env.rst
+++ b/docs/env.rst
@@ -1,0 +1,124 @@
+.. _env:
+
+Environment
+===========
+
+Basics
+------
+
+Environment variables are written as ``$`` followed by a name (e.g. ``$HOME``,
+``$PWD``, ``$PATH``).  They can be set, deleted, and used just like regular
+Python variables.
+
+For a full introduction see the :ref:`Environment Variables <tutorial>` section
+of the tutorial.
+
+Registering Environment Variables
+---------------------------------
+
+You can manually register environment variables to define their type and documentation.
+This is particularly useful for extensions or complex configurations. The documentation provided
+will be shown during tab-completion.
+
+.. code-block:: xonsh
+
+    @.env.register('MY_VAR1', type='int', default=1, doc='Demo variable 1.')
+    @.env.register('MY_VAR2', type='int', default=2, doc='Demo variable 2.')
+
+Now, when you type ``$MY_<Tab>``, you will see the description.
+
+Available types: ``"bool"``, ``"str"``, ``"int"``, ``"float"``, ``"path"``,
+``"env_path"``, ``"abs_path"``.  To get the current list:
+
+.. code-block:: xonshcon
+
+    @ list(@.imp.xonsh.environ.ENSURERS.keys())
+    ['bool', 'str', 'path', 'env_path', 'abs_path', 'float', 'int', 'var_pattern']
+
+Getting Help on a Variable
+--------------------------
+
+Use the ``$VAR?`` syntax to see the description, default value, and other
+metadata of any environment variable:
+
+.. code-block:: xonshcon
+
+    @ $XONSH_COMMANDS_CACHE_READ_DIR_ONCE?
+    Name: $XONSH_COMMANDS_CACHE_READ_DIR_ONCE
+    Description: List of directory prefixes whose contents are cached on first
+    access and never re-read within the session.
+    Default: []
+    Configurable: True
+
+This works for both built-in and registered variables.
+
+Variable Patterns (``VarPattern``)
+----------------------------------
+
+Xonsh allows defining **pattern rules** that automatically apply type handling
+to environment variables whose names match a regex pattern.  This is powered by
+the ``VarPattern`` class.
+
+Built-in patterns
+^^^^^^^^^^^^^^^^^
+
+Xonsh ships with two default patterns:
+
+* ``$XONSH_ENV_PATTERN_PATH`` -- variables ending with ``PATH`` are treated as
+  ``env_path`` (e.g. ``$MYPATH``, ``$LD_LIBRARY_PATH``).
+* ``$XONSH_ENV_PATTERN_DIRS`` -- variables ending with ``DIRS`` are treated as
+  ``env_path`` (e.g. ``$XDG_DATA_DIRS``).
+
+Creating a pattern
+^^^^^^^^^^^^^^^^^^
+
+Set a ``VarPattern`` value directly:
+
+.. code-block:: xonsh
+
+    $XONSH_ENV_PATTERN_NUM = @.imp.xonsh.environ.VarPattern(r"\w*_NUM$", "int")
+
+Or register with a default value and documentation:
+
+.. code-block:: xonsh
+
+    @.env.register(
+        "XONSH_ENV_PATTERN_NUM",
+        type="var_pattern",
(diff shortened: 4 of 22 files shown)


anki-code:
Please check the pre-commit exception (below) and why https://github.com/xonsh/xonsh/pull/6125 is not working in readline:

```
@.imp.json.enc<tab>  # no substring completion
```

Nesar21:
Hi @anki-code, confirmed — `@.imp.json.enc<Tab>` is not completing with the current fix. The prefix safety filter is too aggressive and is breaking the substring completion behavior from #6125. I'll investigate a fix that preserves both behaviors and push an update.

Nesar21:
@anki-code — investigated further. This is a fundamental readline 
architectural constraint vs #6125 substring matching.

Readline's GCP logic replaces the typed word with the common prefix 
of ALL returned candidates. Since `decoder` and `JSONDecoder` share 
no common prefix (GCP = ""), passing both causes readline to delete 
the user's typed `dec`.

Three options:

1. Strict filter (current) — preserves typed prefix, only shows 
   prefix matches in readline. Stable but inconsistent with ptk.

2. No filter — shows all substring matches but readline deletes 
   typed prefix before displaying. Feature parity but poor UX.

3. GCP-aware smart filter (current implementation) — allows 
   substring matches when their GCP >= typed length. Falls back to 
   prefix-only when matches diverge too much. Balanced but complex.

Prompt_toolkit handles this natively via dropdown without touching 
the buffer. Readline cannot replicate this behavior at this layer.

If you have a specific approach in mind or know of a readline trick 
that handles this, happy to implement it. Otherwise let me know 
which option you prefer and I'll finalize accordingly.

anki-code:
## Review of PR #6211: Readline prefix swallowing fix

**Author:** Nesar21 | **Issue:** #6209

### Summary

Removes `_render_completions`, replacing it with inline boundary alignment + a GCP-aware substring safety filter in `completedefault`. The goal is to prevent readline from deleting the typed prefix when substring matches diverge.

---

### Bug: `completions = rtn_completions` makes the safety filter's zip a self-join

After boundary alignment:

```python
completions = rtn_completions       # <-- aliases them
# ...
for orig, rtn in zip(completions, rtn_completions, strict=False):
    #  orig IS rtn — same list, same elements
    safe_rtn.append(rtn)
    safe_orig.append(orig)

rtn_completions = safe_rtn
completions = safe_orig   # identical to safe_rtn
```

`completions` and `rtn_completions` point to the **same list**. The `orig`/`rtn` distinction is meaningless — `safe_orig` and `safe_rtn` contain the same elements.

This breaks the logic below:
```python
elif show_completions == 2:
    return completions   # supposed to be originals, now returns aligned
```

In the original code `completions` is the raw output from `self.completer.complete()`, and `rtn_completions` is the rendered version. This distinction is lost.

**Fix:** don't overwrite `completions`. Keep a separate reference to the originals.

---

### Bug: duck typing instead of isinstance

```python
is_rich = hasattr(c, "replace") and hasattr(c, "prefix_len")
```

**`str.replace()` exists on every string.** The `hasattr(c, "replace")` check is always `True` — it contributes nothing. The code works only because `prefix_len` is RichCompletion-specific. The rest of the codebase uses `isinstance`:

```python
# deleted _render_completions:
if isinstance(comp, xct.RichCompletion) and comp.prefix_len is not None:
```

**Fix:** `isinstance(c, RichCompletion)`.

---

### Commented-out debug logging in production code

```python
# with open("/tmp/xonsh_readline.log", "a") as f:
#     f.write(...)
```

Two blocks of this. Remove before merge.

---

### `strict=False` in zip masks potential bugs

```python
for orig, rtn in zip(completions, rtn_completions, strict=False):
```

These lists are always the same length (built from the same input). `strict=True` would catch regressions. The commit message says "appease ruff B905" — but ruff wants a **conscious choice**, and here the correct one is `True`.

---

### Test coverage regression

The old `test_render_completions` covered 10 cases including:
- Plain string completions (not RichCompletion)
- `RichCompletion` with `prefix_len=0` and `prefix_len=None`
- Empty prefix

The new `test_boundary_alignment` only tests `RichCompletion` with explicit `prefix_len`. No tests for:
- Plain string completions through `completedefault`
- `prefix_len=None` fallback to `plen`
- Empty prefix / empty line
- The `show_completions == 2` path (display-all case)

---

### Overlap check silently passes untrimmed value

```python
if c_plen > readline_plen:
    ...
    if len(c_str) >= offset and c_str.lower().startswith(overlap.lower()):
        new_val = c_str[offset:]
    else:
        new_val = c_str  # <-- no trim, readline gets wrong-length string
```

If the overlap check fails, the completion passes through untrimmed. Readline will replace `readline_plen` characters with this string that still has `c_plen` worth of prefix — producing garbled output. The old code trimmed unconditionally.

---

### Verdict

The approach is sound — readline's GCP problem with substring matches is real and the safety filter is a reasonable solution. But the implementation has a **logic bug** (aliased lists), **style issues** (duck typing, debug comments), and **test coverage regressions**. Needs a revision before merge.

Nesar21:
@anki-code — addressed all 5 issues from your review:
1. Fixed aliased lists — added orig_completions reference before boundary alignment
2. Replaced hasattr duck typing with isinstance(c, RichCompletion)
3. Removed commented-out debug logging blocks
4. Changed strict=False to strict=True in zip
5. Fixed untrimmed fallback — now trims unconditionally
6. Fixed show_completions==2 return path to return orig_completions

10 tests passing locally.

anki-code:
# Review: PR #6211 — readline prefix swallowing fix

**Author:** Nesar21 | **Issue:** #6209 | **CI:** green

## Summary

Removes `_render_completions`, adds inline boundary alignment + GCP-aware substring safety filter in `completedefault`. Fixes `@.imp.jso<Tab>` deleting the typed prefix.

## Issues

### 1. Tests don't cover boundary alignment

`test_boundary_alignment` uses `show_completions==2` which returns raw `orig_completions`, bypassing trim/pad code. The author confirms: `# path 2 returns raw original, not trimmed`.

### 2. Breaks substring completion (#6125)

Author confirmed `@.imp.json.enc<Tab>` no longer completes. Safety filter aggressively removes substring matches.

### 3. Overcomplicated inline code

Removed `_render_completions` (~15 lines) replaced with ~50 lines inline in already complex `completedefault`. Should be a separate testable function.

### 4. Dead comment

`# Bug 5 fix: trim unconditionally (see below)` — no "below" exists.

### 5. Unsafe fallback trim

```python
else:
    new_val = c_str[offset:]  # trim unconditionally
```

Trims even when overlap check fails — can produce garbage.

### 6. Silent behavior changes

- `.lower()` makes matching case-insensitive (original was case-sensitive)
- `show_completions == 2` returns `list` instead of `set`

### 7. Over-mocked tests

Original: simple unit tests of a pure function. New: mock half the module, fragile to refactoring.

## Verdict

Fixes the reported bug but regresses substring completion, key logic is untested, code needs simplification. Needs rework.

Nesar21:
@anki-code — I have fully resolved the other 6 items from your review locally (extracted pure functions, `strict=True` on zip, removed `.lower()`, fixed aliased lists, etc.).

Before I push the updated commits, I need your product decision on how you want to handle the #6125 substring completion inside the `readline` backend.

Because GNU Readline computes the Longest Common Prefix (LCP) inline, passing it divergent substrings (like `encoder` and `JSONEncoder` for the typed word `enc`) results in an LCP of `""`. Readline natively executes this and deletes the user's typed prompt. We cannot bypass this C-level behavior.

This leaves us with two UX options for the `readline` shell. Which do you prefer I implement?

---

**Option A: The Aggressive Filter (Current Stage 2 active)**
We strictly filter the matches back to prefix-only before handing them to Readline.
- **Pro:** The user's typed prompt is safely preserved.
- **Con:** It hides substring completion in Readline. Typing `enc<Tab>` exclusively completes `encoder` and hides `JSONEncoder`.

Screenshot 2026-04-03 at 10 16 13
Screenshot 2026-04-03 at 10 16 16

---

**Option B: No Filter (Stage 2 removed)**
We pass all the divergent substrings directly to Readline.
- **Pro:** A double-tab will successfully print all options (`JSONEncoder`, `detect_encoding`) to standard output so the user can see them.
- **Con:** The moment they press Tab, Readline's LCP forcefully deletes their typed `enc` from the buffer, forcing them to manually retype everything while looking at the printed list.

Screenshot 2026-04-03 at 10 10 56
Screenshot 2026-04-03 at 10 11 06

---

Let me know which tradeoff you want for the `readline` shell, and I will push the final code matching that decision.

anki-code:
> Before I push the updated commits, I need your product decision on how you want to handle the https://github.com/xonsh/xonsh/pull/6125 substring completion inside the readline backend.

Let's do nothing for this or only what is working.
The main goal is to just fix completion that is broken on read line.

Nesar21:
@anki-code — resolved per your direction. Stripped Stage 2 filter entirely. Stage 1 boundary alignment only.

@.imp.jso<Tab> → prefix preserved, completes correctly
enc<Tab> → native readline LCP behaviour, no filter applied
All 6 tests passing locally
ruff/mypy/pre-commit passing

anki-code:
@Nesar21 I've made a small fix that solves the situtation faster - https://github.com/xonsh/xonsh/pull/6230
Thanks you for your try!

PR #6216, opened 2026-04-02
fix(parser): Fix parsing macro block with comments

anki-code:
Fixed:
* Fixed https://github.com/xonsh/xonsh/issues/4207
* Fixed https://github.com/xonsh/xonsh/issues/4382

### Before

```xsh
# https://github.com/xonsh/xonsh/issues/4207
from xonsh.contexts import Block
class TestBlock(Block):
    def __exit__(self, *a, **kw):
        print(repr(self.macro_block))
        
with! TestBlock() as b:
    # first
    # first
    echo 1

# comment 1
# comment 2

if 1:
    pass

# https://github.com/xonsh/xonsh/issues/4382
from xonsh.contexts import Block
class TestBlock(Block):
    def __exit__(self, *a, **kw):
        print(repr(self.macro_block))

with! TestBlock() as a:
    # comment 1
    # comment 2
    echo hello
    # comment 3
    # comment 4
pass
```
Result:
```xsh
'    echo 1\n\n# comment 1\n# comment 2\n\n' 🔴 
'echo hello\n# comment 3\n# comment 4\n' 🔴 
```

### After

```xsh
'# first\n# first\necho 1\n' 🟢 
'# comment 1\n# comment 2\necho hello\n# comment 3\n# comment 4\n' 🟢 
```

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #6217, opened 2026-04-02
fix(parser): Fix single word gulping after block with indent

anki-code:
I've found this during testing #6216

```xsh
from xonsh.contexts import Block
class TestBlock(Block):
    def __exit__(self, *a, **kw):
        print(repr(self.macro_block))

with! TestBlock() as a:
    # comment 1
    # comment 2
    echo hello
    # comment 3
    # comment 4
pass<Cursor><Enter>
p  # Gulping part of the word. 
```

## For community


anki-code:
Fix is small and well covered by tests. Self merge.

Issue #5291, opened 2024-03-01
Importing logic is different from python


anki-code:
```xsh
mkdir -p /tmp/123
cd /tmp/123
mkdir -p pkg

echo "print('hello')" > pkg/importme.py
echo "import pkg.importme" > run.py

python run.py
# hello

xonsh run.py
# ModuleNotFoundError: No module named 'pkg'
```
Do we need to do the same as in python?

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #6219, opened 2026-04-02
fix(import): Fixed CPython sys_path_init requrement to have current directory in sys path for executed script

anki-code:
### Requirement

CPython — https://docs.python.org/3/library/sys_path_init.html:

>The first entry in the module search path is the directory that contains the input script, if there is one.

### Before

```xsh
mkdir -p /tmp/123
cd /tmp/123
mkdir -p pkg

echo "print('hello')" > pkg/importme.py
echo "import pkg.importme" > run.py

python run.py
# hello 🟢 

xonsh run.py
# ModuleNotFoundError: No module named 'pkg' 🔴 
```

### After

```xsh
python run.py
# hello 🟢 

xonsh run.py
# hello 🟢 
```

Fixed https://github.com/xonsh/xonsh/issues/5291

## For community

### DIFF
--- a/tests/test_integrations.py
+++ b/tests/test_integrations.py
@@ -862,6 +862,25 @@ def test_single_command_no_windows(cmd, fmt, exp):
 
 
 @skip_if_no_xonsh
+def test_script_local_import(tmp_path):
+    """xonsh script-file should add script dir to sys.path like CPython does."""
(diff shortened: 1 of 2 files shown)


anki-code:
It's strict requirement and trivial fix so merge now.

PR #6224, opened 2026-04-03
chore: Tests: Split integration tests into fast integration tests and slow stress tests

anki-code:
Changes:
* Split integration tests to integration+stress (fd leaking test is very helpful but it takes time).
* Rename integration to xintegration to put it at the end.

## For community


anki-code:
I need this for continue testing and changes are trivial so self merge.

Issue #5458, opened 2024-05-29
Warlus with env in python substitution: NamedExpr target must be a Name


anki-code:
Error:

```xsh
echo @($name := 'Alex')
# TypeError: NamedExpr target must be a Name
```

Expected:
```xsh
echo @($name := 'Alex')
# Alex
```

Workaround:
```xsh
echo @(name := 'Alex')
echo reuse @(name)
# Alex
# reuse Alex
```

Real life case:
```xsh
git checkout master
git checkout -b @($branch_name := input('Branch:'))
git push --set-upstream origin $branch_name
```

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

anki-code:
This is a Python limitation, not xonsh. Walrus (:=) requires target to be a plain Name:

(name := 'Alex')      # ✓ Name
(__xonsh__.env["name"] := 'Alex')  # ✗ Subscript → SyntaxError

$name compiles to __xonsh__.env["name"] (Subscript) — Python rejects it.

@(name := 'Alex') (without $) already works — assigns to a Python variable.

Supporting $name := would require an AST transform to rewrite NamedExpr with env-var targets into
env.__setitem__ + return value. Non-trivial parser change.

PR #6226, opened 2026-04-03
feat(parser): Added walrus operator support for env variables e.g. `echo @($TMP := '/tmp') && ls $TMP` 

anki-code:
Closes https://github.com/xonsh/xonsh/issues/5458

### Implementation

CPython has no support of complex variables from left in `$left := right` so we can't just add env variable that is object to parser. Instead of this we transform `$left := right` to `$left.set(right)` that returns env var value after transformation (for registered variables) or row value (for not registered vars).

### Before

```xsh
echo @($name := 'Alex')
# TypeError: NamedExpr target must be a Name
echo $name
# Error
```

### After

```xsh
echo @($name := 'Alex')
# Alex
echo $name
# Alex

echo @($PATHEXT := '/dir1:/dir2')  # PATHEXT has type 'path_env' and will be typed into list
# /DIR1 /DIR2
$PATHEXT
# ['/DIR1', '/DIR2'] 

if $QWE := False:
    echo 1
else:
    echo 2
# 2
```
Nice patterns:
```xsh
echo Who is @($name := 'Alex')?
echo Oh, this is $name
# Who is Alex?
# Oh, this is Alex

if $NAME := $(whoami):
    echo Play with $NAME
else:
   echo Who are you?
# Play with snail
```

## For community

### DIFF
--- a/tests/parsers/test_ast.py
+++ b/tests/parsers/test_ast.py
@@ -125,6 +125,25 @@ def test_whitespace_subproc(test_input, xonsh_execer_parse):
     assert xonsh_execer_parse(test_input)
 
 
+def test_paren_boolop_no_subshell(xonsh_execer):
+    """``(cd subdir && ls)`` must not create a subshell that duplicates ``cd``.
+
+    Regression: the non-greedy ``subproc_toks`` result for ``ls`` was falsely
+    rejected by the consistency check because ``maxcol`` captured the closing
+    ``)``. This caused a greedy fallback that wrapped the entire line into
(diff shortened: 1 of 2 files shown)


anki-code:
@jnoortheen what do you think about `$left.set(right)` implementation? Is it ok for now?

anki-code:
After additional review I see no any alternative how we can implement this. Current implementation is compact and clear. So I'm going to merge.

anki-code:
1 failed test just because of slow CI

Issue #3612, opened 2020-06-09
ImageMagick's import is syntax-highlit as an import statement


qwenger:
I have ImageMagick with its cli tool `import` installed. When I type in e.g.:

```
import -window root test.png
```

the command runs successfully (a screenshot of my screen is saved to test.png), but the syntax highlighting corresponds to python mode (the "import" has the same color as if I write e.g. "import os", not the same as e.g. "echo 1").

Neither `window` nor `root` nor `test` are defined python variables in my shell.

## xonfig

<details>

```
+------------------+---------------------+
| xonsh            | 0.9.18              |
| Git SHA          | 26475146            |
| Commit Date      | May 5 06:42:28 2020 |
| Python           | 3.8.3               |
| PLY              | 3.11                |
| have readline    | True                |
| prompt toolkit   | 3.0.5               |
| shell type       | prompt_toolkit      |
| pygments         | 2.6.1               |
| on posix         | True                |
| on linux         | True                |
| distro           | arch                |
| on darwin        | False               |
| on windows       | False               |
| on cygwin        | False               |
| on msys2         | False               |
| is superuser     | False               |
| default encoding | utf-8               |
| xonsh encoding   | utf-8               |
| encoding errors  | surrogateescape     |
+------------------+---------------------+
```

</details>

## Expected Behavior
The line should be syntax-highlit following subprocess mode, for example the "import" word should be colored the same as "echo".

## Current Behavior
The line seems to be syntax-highlit following python mode.

scopatz:
Yeah, this is a legitimate issue.  The highlighting goes through a totally separate code path from execution, though. PRs welcome!

qwenger:
Yes I see and that's certainly not an easy thing to change. Note that is affects also e.g. indented blocks, for instance

```
def test():
    echo 1
```

Here the `echo` is formatted as Python.

I think that the execer should somehow be the single source of truth regarding the decision on Python vs. subprocess mode, and the lexer/style should consult it on that matter.

I'll see whether I get some time for a PR. But that's not necessarily in a near future.

anki-code:
In xonsh >= 0.23.0 use `$XONSH_BUILTINS_TO_CMD=True` to enable experimental feature: if plane name is python builtin and has executatble or alias use it as cmd (so simple `id`, `zip`, `import` are working).

Issue #3508, opened 2020-04-07
Python builtins shadowing some programs


sawolford:
## Python builtins shadowing some programs

Some shell commands share the same name with Python builtin functions (e.g., "id"). I often run "id" to see what user I am currently in my shell. It also gives me group information at a glance, including current primary group.

I honestly think this is a symptom of a larger problem in that there is no good way that I'm aware of to force subprocess interpretation of a command. Even if I alias "id" to something that will then be interpreted in subprocess mode (e.g., aliases["id"] = "id -a"), Xonsh still interprets the line in Python mode before it even substitutes the alias.

Although I think the default mechanism of determining the command type by heuristic is fairly robust in practice, there are some edge cases that it can't handle. There ought to be a way to force one interpretation or the other (Python or subprocess mode) in a repeatable fashion if that's the intent of the user.

Alternatively, I would like to know how to handle this specific case without requiring unnecessary command-line switches to be passed (e.g., "id -a") or strange machinations (e.g., "echo $(id)").

# xonfig

```
+------------------+----------------------+
| xonsh            | 0.9.15               |
| Git SHA          | feeb6315             |
| Commit Date      | Mar 19 10:50:33 2020 |
| Python           | 3.7.7                |
| PLY              | 3.11                 |
| have readline    | True                 |
| prompt toolkit   | 2.0.9                |
| shell type       | prompt_toolkit2      |
| pygments         | 2.4.2                |
| on posix         | True                 |
| on linux         | False                |
| on darwin        | True                 |
| on windows       | False                |
| on cygwin        | False                |
| on msys2         | False                |
| is superuser     | False                |
| default encoding | utf-8                |
| xonsh encoding   | utf-8                |
| encoding errors  | surrogateescape      |
+------------------+----------------------+
```

## Expected Behavior

Xonsh should run the program and not return the type of the Python builtin of the same name (REPL-like behavior).

```
$ id
uid=501(username) gid=20(staff) groups=20(staff),<...>
```

## Current Behavior

```
$ id
<function id>
```

## Steps to Reproduce

Run "id" command on Unix-like operating system with no arguments.

laloch:
@sawolford, as you already discovered yourself, Python built-ins and anything that can be evaluated as a Python statement/expression in general take precedence over Xonsh shell mode. That's a feature. If you need to override the precedence, you can:
```
$ ![id]
$ @$(which id)
$ echo $(id)
$ echo @$(which id)
```

[...]

PR #6227, opened 2026-04-03
feat(parser): Reduce shadowing by using experimental `$XONSH_BUILTINS_TO_CMD=True`

anki-code:
### Motivation

To ability to reduce shadowing issues I'm introducing experimental `$XONSH_BUILTINS_TO_CMD` that enables the logic:
1. If the input is a single name and nothing else e.g. `id`
2. And the name represents Python builtin e.g. `id()` function
3. And there is executable or alias with this name
4. Then run this name as subprocess command.
5. Otherwise fallback to Python.

Related to:
* https://github.com/xonsh/xonsh/issues/5802
* https://github.com/xonsh/xonsh/issues/3508
* https://github.com/xonsh/xonsh/issues/3612

### Before

```xsh
zip
# zip builtin
id
# <function id>
...
# elepsis
```
Workarounds were:
```
Zip # upper case
@('zip') # substitution
```

### After

```xsh
$XONSH_BUILTINS_TO_CMD=True  # default False

zip
# Zip 3.0 (July 5th 2008). Usage:
# zip [-options] [-b path] [-t mmddyyyy] [-n suffixes] [zipfile list] [-xi list]

id 
# uid=321(pc) gid=12(staff) groups=32(staff),34(everyone)

id(@)  # still working in Python mode
# 4310051072

id = 1  # now it's not Python builtin it's user object
id
# 1

aliases['...'] = 'echo 1'
...
# 1
del aliases['...']
...
# elepsis
```

## For community

### DIFF
--- a/docs/callable_aliases.rst
+++ b/docs/callable_aliases.rst
@@ -1,0 +1,462 @@
+Callable Aliases
+================
+
+In xonsh, a Python function can be registered as a shell command (alias). The
+function declares which arguments it needs, and xonsh fills them automatically
+based on parameter names. When a function runs as an alias, xonsh redirects
+``sys.stdout`` and ``sys.stderr`` inside it, making it possible to capture all
+output that happens within the function — including ``print()`` calls and
+subprocess commands.
+
+
+Signature Parameters
+--------------------
+
+Callable aliases receive arguments matched by name. You can request any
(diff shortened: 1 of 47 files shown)


anki-code:
Merged as experimental feature (disabled by default).

PR #6230, opened 2026-04-03
fix(readline): Fixed the case when completion can remove prefix

anki-code:
Fixes https://github.com/xonsh/xonsh/issues/6209

Our main goal here is to just make it works because now it's broken completely.

## For community


anki-code:
Tested on Mac and Win

Issue #5831, opened 2025-04-13
&& and || operators incorrectly parsed without surrounding spaces


cyb3rmonk:
## Current Behavior

Fails: `xonsh --no-rc --no-env -c 'ls&&ls'`
Works: `xonsh --no-rc --no-env -c 'ls && ls'`

## Expected Behavior

I would expect && and || to work the same with or without spaces around them

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

Issue #5251, opened 2023-12-20
cd subdir && gives false errors


jun0:
## xonfig

<details>

```
$ xonfig
+------------------+-----------------+
| xonsh            | 0.14.3          |
| Python           | 3.11.6          |
| PLY              | 3.11            |
| have readline    | True            |
| prompt toolkit   | 3.0.41          |
| shell type       | prompt_toolkit  |
| history backend  | json            |
| pygments         | 2.17.2          |
| on posix         | True            |
| on linux         | False           |
| on darwin        | True            |
| on windows       | False           |
| on cygwin        | False           |
| on msys2         | False           |
| is superuser     | False           |
| default encoding | utf-8           |
| xonsh encoding   | utf-8           |
| encoding errors  | surrogateescape |
| xontrib          | []              |
| RC file          | []              |
+------------------+-----------------+
```

</details>

## Expected Behavior
`(cd subdir && ls)` should print out the contents of `subdir` assuming a directory by that name exists, without prefixing the names by `subdir/`.  If I understand the semantics of `()` correctly, the shell's cwd should remain unchanged.

## Current Behavior

xonsh says `cd: no such file or directory: subdir` and skips the `ls`.  But the shell's cwd is switched to `subdir` anyway.

### Traceback (if applicable)

N/A

## Steps to Reproduce

Start up xonsh and type 
```
mkdir subdir
touch subdir/file-a
(cd subdir && ls)
```

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

gnewson:
Hi, I'm new to the community but I'd be interested in working on this.

I've had a look through the docs and I can't find where @jun0 you get the semantics of `()` from.  Could you point me in that direction so I can start from the same understanding, thanks?

Further to this, if that understanding is correct, should `(cd subdir)` effectively enter the subdirectory but not switch to it?

gnewson:
Thanks, that give me more focus.  I was thrown a bit by the supporting text.

PR #6233, opened 2026-04-03
fix(parser): Fixed try_subproc_toks issue with wrong wrapping a command with parentheses

anki-code:
### Issue

The consistency check spline != f"![{line[mincol:maxcol].strip()}]" compares the subproc_toks result against
a raw slice of the source line, but maxcol is computed from the AST node (max_col(node) + 1) and can capture
characters beyond the token itself — in this case the closing ). This is not a bug in subproc_toks (which    
produced the correct result), but a bug in the validation of its result.

### Before

Two cases:
```xsh
cd /tmp
mkdir subdir
touch subdir/file-a

cd subdir && ls
# file-a  # as expected
cd ..

(cd subdir && ls)
# cd: no such file or directory: subdir  # the issue 🔴 
cd ..

(cd /tmp/subdir && ls)  # full path
# file-a  # as expected
```
```xsh
Works: xonsh --no-rc --no-env -c 'ls && ls'
Fails: xonsh --no-rc --no-env -c 'ls&&ls' 🔴 
```

### After

```xsh
(cd subdir && ls)
# As expected 🟢 
```
```xsh
xonsh --no-rc --no-env -c 'ls&&ls'
# As expected 🟢 
```

Fixes:
* Fixes https://github.com/xonsh/xonsh/issues/5251
* Fixes https://github.com/xonsh/xonsh/issues/5831

## For community

### DIFF
--- a/tests/parsers/test_ast.py
+++ b/tests/parsers/test_ast.py
@@ -125,6 +125,25 @@ def test_whitespace_subproc(test_input, xonsh_execer_parse):
     assert xonsh_execer_parse(test_input)
 
 
+def test_paren_boolop_no_subshell(xonsh_execer):
+    """``(cd subdir && ls)`` must not create a subshell that duplicates ``cd``.
+
+    Regression: the non-greedy ``subproc_toks`` result for ``ls`` was falsely
+    rejected by the consistency check because ``maxcol`` captured the closing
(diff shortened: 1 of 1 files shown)


anki-code:
It's simple change but we need to merge it to start testing with all changes as soon as possible.

Issue #3381, opened 2019-11-03
Regular expression globbing mangles the regex, breaking some patterns


inventhouse:
In the current implementation of [`reglob`](https://xon.sh/_modules/xonsh/built_ins.html#reglob),  the expression is split and matched segment-by-segment; this allows paths to be pruned early, but breaks otherwise valid searches:

```
# find .py files 0-or-more directories deep
$ r`(.*/)*\w+\.py`                                                                          
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK = True
re.error: missing ), unterminated subpattern at position 0
```

I originally filed this as a documentation issue (#3377), but realized that we could test a partial path against a full pattern and if we could determine whether the pattern consumes all of the string, continue to search that branch, but if it is actively rejected, prune.  That could get us similar efficiency without butchering the pattern.  (Although a lot of patterns, like the one above, could easily degenerate into full-searches anyway.)

I posted [this question on StackOverflow](https://stackoverflow.com/questions/58599729/programmatically-determine-where-pythons-re-match-stopped-matching) and dug into the implementation of the `re` module, but found that, as far as I can tell, `match` is [implemented in C](https://github.com/python/cpython/blob/d0e0f5bf0c07ca025f54df21fd1df55ee430d9fc/Modules/_sre.c#L589) and hides the specific match failure.

This issue is meant to track either getting access to the match error in the standard `re` module or finding a pure-python version and implement the algorithm above.

```
$ xonsh --version                                                                           
xonsh/0.9.13
```

inventhouse:
From #3372, workaround to do a full-depth search and filter by a complete pattern:

```
[ p for p in g`**/*` if re.match(r"(.*/)*\w+\.py", p) ]
```

Some efficiency can be regained by crafting the glob to "pre-filter" paths ahead of the regex

scopatz:
We are very open to improvements here.  PRs very welcome!

inventhouse:
I've pursued this far enough to find that we will need to significantly modify the `re` module to implement the algorithm above; I may get back to this at some point because I think some _very_ cool regex improvements and tools could fall out as a side-effect, but I have lots of pet projects 😄

astronouth7303:
Wouldn't it be simpler to just try windows paths with slashes and backslashes? afaik windows doesn't allow slashes in filenames and accepts them as directory separators anyways.

[...]

Issue #3372, opened 2019-10-25
Regular expression globbing documentation lacks important details


inventhouse:
I'm new to xonsh and very excited to have things like `re` at my fingertips at all times, but my first attempt to use re-globbing produced a strange and confusing error message:
```
# find .py files 0-or-more directories deep
$ r`(.*/)*\w+\.py`                                                                          
xonsh: For full traceback set: $XONSH_SHOW_TRACEBACK = True
re.error: missing ), unterminated subpattern at position 0
```

Clearly, I thought, my subpattern is balanced, and nothing in the [documentation] lead me to believe that it would be parsed any differently than standard `re`:

> Any string that uses backticks (`) instead of quotes (', ") is interpreted as a regular expression to match filenames against.

[documentation]: https://xon.sh/tutorial.html#regular-expression-globbing

I was able to find [the code], which showed me that the expression is split and matched segment-by-segment, which is not necessarily a bad way to do it, but does create very important limitations and gotchas that should be documented.  It would also be helpful if the standard error message thrown by `re` were augmented to include the pattern (or partial-pattern) and a bit of explanation.

[the code]: https://xon.sh/_modules/xonsh/built_ins.html#reglob

inventhouse:
You can, of course, easily do a full-depth search and filter by a complete pattern if you want:

`[ p for p in g`**/*` if re.match(r"(.*/)*\w+\.py", p) ]`

inventhouse:
For posterity:
```
$ xonsh --version                                                                           
xonsh/0.9.13
```

gforsyth:
Thanks for reporting, @inventhouse ! We'd welcome any clarifying documentation (or we can look into changing the implementation to better match expectations).

astronouth7303:
Ummm... that's a bug. That's a valid regular expression. You should not be getting that error.

inventhouse:
I don't mind that the implementation is more like "path-wise regular expression", as that can prune a great deal of the search space down in mosts cases; also the workaround to do a full search by a complete pattern is trivial, whereas to re-implement the more efficient "path-wise" search is not.

I just think it needs to be documented and perhaps a small improvement to the error messages; when I have a bit of time, I think this would be a good first contribution.

inventhouse:
[...]

PR #6234, opened 2026-04-03
fix!: glob: `$DOTGLOB` now control all forms of globbing uniformly (normal, regex) with special documentation page

anki-code:
### Motivation

`$DOTGLOB` should control all forms of globbing uniformly but having regex globs silently ignore it while normal globs respect it is inconsistent and surprising for users.

### 1. Unification

| Scenario | Before | After |
|---|---|---|
| `` g`*` `` | respects `$DOTGLOB` 🟢 | respects `$DOTGLOB` 🟢 |
| `` g`**/*` `` | respects `$DOTGLOB` 🟢 | respects `$DOTGLOB` 🟢 |
| `` `\.hidden` `` | **always matches** 🔴 | respects `$DOTGLOB` 🟢 |
| `` r`\..*` `` | **always matches** 🔴 | respects `$DOTGLOB` 🟢 |
| `$DOTGLOB = True;` `` r`\..*` `` | always matches 🔴 | matches 🟢 |
| `$DOTGLOB = False;` `` r`\..*` `` | **still matches** 🔴 | **no match** 🟢 |

### 2. Added detailed Globbing page to documentation

Fixes https://github.com/xonsh/xonsh/issues/3372

### 3. Improved error message

```xsh
@ r`(.*/)*\w+\.py`
# Regex glob error in segment '(.*' of pattern '(.*/)*\\w+\\.py': missing ), unterminated 
# subpattern at position 0. Regex globs are split by '/' and each segment is compiled 
# separately, so groups and backreferences cannot span across '/'. 
# See https://xon.sh/globbing.html
```
Fixes https://github.com/xonsh/xonsh/issues/3372

### 4. Regex search

The "Full-Path Regex via Custom Search" section in docs/globbing.rst provides a solution for patterns that span across /. Users can define a custom search function and use it with the ``funcpattern`` syntax to get full-path regex matching without segment splitting.   

Fixes https://github.com/xonsh/xonsh/issues/3381

## For community

### DIFF
--- a/docs/callable_aliases.rst
+++ b/docs/callable_aliases.rst
@@ -1,0 +1,462 @@
+Callable Aliases
+================
+
+In xonsh, a Python function can be registered as a shell command (alias). The
+function declares which arguments it needs, and xonsh fills them automatically
+based on parameter names. When a function runs as an alias, xonsh redirects
+``sys.stdout`` and ``sys.stderr`` inside it, making it possible to capture all
+output that happens within the function — including ``print()`` calls and
+subprocess commands.
+
+
+Signature Parameters
+--------------------
+
+Callable aliases receive arguments matched by name. You can request any
+combination of the following parameters in any order:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 25 75
+
+   * - Parameter
+     - Description
+   * - ``args``
+     - List of strings — command-line arguments passed to the alias.
+   * - ``stdin``
+     - File-like object for reading piped input, or ``None`` if no pipe.
+   * - ``stdout``
+     - File-like object for writing standard output.
+   * - ``stderr``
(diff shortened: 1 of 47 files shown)


anki-code:
@jnoortheen this change has no backwards compatibility so if user has DOTGLOB=False and use r`\..*` it will have empty list after release. BUT because of:
* the changes are all around dot-files (hidden files)
* the files just dissapear from list (no files better than having trash)
* we have no reports about difference between g and r and DOTGLOB setting

There is no big impact and we can release it just with "breaking change" in the list.

So please review as soon as possible because we're going to have week of testing and release 0.23.0.

anki-code:
Ok, in our plan there was a code freeze week with overall integration testing. So I'm going to merge this and then we will have time to fix something.

PR #6238, opened 2026-04-04
fix(completer): Add `@aliases.completer` decorator and fix FuncAlias support

anki-code:
Added:

```xsh
def _complete_hello(command, alias):
    return {'world', 'there', 'xonsh'}

@aliases.register
@aliases.completer(_complete_hello)
def _hello(args):
    echo @(args)
```

And fixed issue with FuncAlias completion.

## For community


anki-code:
Trivial merge.

Issue #4654, opened 2022-01-24
Python-mode strings are sometimes expanded when adjacent


agoose77:
Given the following string
```python
x = "$HOME"
```
Xonsh will expand the environment variable in the following cases:
```xonsh
$ echo @(x).@(x)
/home/angus./home/angus
```

```xonsh
$ echo @(x):@(x)
/home/angus:/home/angus
```

This looks like a special case:
```xonsh
$ echo @(x)0@(x)
$HOME0/home/angus
```

But in fact it is concatenating and _then_ expanding. As I do not have an env var `$HOME0` set, it does not expand.

It looks like this might be related to the outer-product logic mentioned [in the tutorial](https://xon.sh/tutorial.html#:~:text=When%20used%20inside%20of%20a%20subprocess%20argument%20and%20%3Cexpr%3E%20evaluates%20to%20a%20non%2Dstring%20iterable%2C%20%40()%20will%20expand%20to%20the%20outer%20product%20of%20all%20given%20values%3A), although I wouldn't expect it to apply here. I also do not think we should apply expansion rules in the product case vs in the non-product case.

anki-code:
Nice catch!
Anyone who want to work on this issue can start from reviewing #3832 that points to the tests and the logic.

PR #6240, opened 2026-04-04
feat: Add `$XONSH_SUBPROC_ARG_EXPANDUSER` to ability to switch expanding off

anki-code:
Related to https://github.com/xonsh/xonsh/issues/5970

We have a behavior that was implemented based on POISX shells logic:  expand `~` and `~username` to `/home/username` implicitly. Those shells, unlike xonsh, there are huge amount of implicit logic that we love like differences `QWE=~asd`, `QWE='~asd'`, `echo $QWE`, `echo "$QWE"`, etc.

I think "user expanding" is another feature that was made to please traditional command shells.
In this PR I introduced XONSH_SUBPROC_ARG_EXPANDUSER to make this shit visible and we will think what to do with this later.

------------------------------------

Yes, I think it was made to support things like `cmd VAR=~/dir`. But my current thoughts are about this should work explicit in the future so if you have direct intention to use expansion you can have an ability to write `cmd e"VAR=~/dir"` or something like this. 

The main thougt that I'm staying here: in xonsh we must have an ability to write things totally explicit. For example we need to have an ability to write:

```xsh
$XONSH_SUBPROC_ARG_EXPANDUSER = False
from xonsh.tools import expanduser as exp

cmd ~args
cmd @(exp("ARG=~user/VAR"))
cmd ~args
```
And this PR is the step to this world.

cc 
* https://github.com/xonsh/xonsh/issues/4654

## For community

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -1733,6 +1733,24 @@ def test_expand_path(expand_user, inp, expand_env_vars, exp_end, xession):
         assert path == "~" + exp_end
 
 
+@pytest.mark.parametrize(
+    "inp, exp_expanded, exp_literal",
+    [
+        ("~/docs", os.path.expanduser("~") + "/docs", "~/docs"),
+        ("~", os.path.expanduser("~"), "~"),
+        ("x=~/path", "x=" + os.path.expanduser("~") + "/path", "x=~/path"),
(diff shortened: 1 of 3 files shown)


anki-code:
Trivial merge. Tests are ok. CI so slow.

PR #6283, opened 2026-04-11
feat: Add `xxonsh` - launches exactly the same ``xonsh`` that was used to start the current session.

anki-code:
### Motivation

We already have `xcontext`, `xpip`, `xpython`. It's time to add `xxonsh`.

### Implementation

`xonsh.aliases.get_xxonsh_alias()` logic:
* if xonsh is entry point it returns entrypoint e.g. `~/.local/bin/xonsh`
* if xonsh is from source e.g. `__main__.py` it returns `[sys.executable, "-c", bootstrap]` where bootstrap has guarantee that you will run execatly the same source as used for current session.

### Making aliases with xxonsh

If you want to have `xtmux` for tmux launching or something like this you can add it easily:

```xsh
aliases['xtmux'] = ['tmux', 'new-session'] + @.imp.xonsh.aliases.get_xxonsh_alias()
```

## For community

### DIFF
--- a/docs/aliases.rst
+++ b/docs/aliases.rst
@@ -53,7 +53,7 @@ Xonsh-specific Aliases
 --------------------
 Displays how commands and arguments are evaluated. Use ``-e`` to expand aliases.
 
-.. code-block:: console
+.. code-block:: xonshcon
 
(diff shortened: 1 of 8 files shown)


anki-code:
Additive change.

PR #6288, opened 2026-04-11
feat(nix): Add Nix packaging

SamLukeYes:
image

Add a Nix Flake that builds xonsh from git source.

```shell
# build xonsh from the latest git source (should work after merging)
nix build 'github:xonsh/xonsh'

# build xonsh from the branch of this PR
nix build 'github:SamLukeYes/xonsh/nix'

# build xonsh from your local copy of source code
cd /path/to/git/repo/xonsh
nix build
```

The nix commands above requires experimental features `nix-command` and `flakes` to be enabled. To run xonsh directly after building, replace the `nix build` commands above with `nix run`.

## Possible next steps

- [ ] Set up garnix.yaml

Garnix CI should work with zero configuration after [setting it up as GitHub App](https://garnix.io/docs/getting-started/github-app/). However, it only builds for `x86_64-linux` platform by default, while this flake also includes `aarch64-linux`, `x86_64-darwin` and `aarch64-darwin`. To set up Garnix for those platforms, garnix.yaml needs to be added.

- [x] Update flake.lock regularly using GitHub Actions
  - Already done in this PR

flake.lock needs to be updated regularly to let nix use up-to-date dependencies to build xonsh package. There's a [GitHub Action](https://github.com/marketplace/actions/update-nix-flake-lock) to do this, and I've been trying it in my own repos (e.g. https://github.com/SamLukeYes/nixos/blob/5d4d3fa1cf47086779e2d9dc98acb883c1b43d56/.github/workflows/update-flake-lock.yml) for a few weeks.

- [ ] Add shell.nix and expose it in flake.nix

Nix shell could be an alternative way to set up the environment for development and testing. Maybe it can be done more easily and properly using `uv2nix`, after https://github.com/xonsh/xonsh/pull/6318

- [ ] NixOS integration

Add an example NixOS config, and document about how to set up and test Xonsh on NixOS.

## For community


anki-code:
What about to put this to separate repository under xonsh project `xonsh/nix` (or what name is right)? In this case we can grant you an access for future updates and editing.

SamLukeYes:
> What about to put this to separate repository under xonsh project `xonsh/nix` (or what name is right)? In this case we can grant you an access for future updates and editing.

My initial thought is to make this repo itself a flake, so that one can use `nix run 'github:xonsh/xonsh'` to run the latest xonsh with a single command. Also, CI that builds and cache nix derivations (like [Garnix CI](https://garnix.io)) can be directly integrated to this repo, so it can run on every push or even for pull requests.

Meanwhile, a separated flake that periodically updates to the latest xonsh commit is also possible. I can try it later if you think it is a better approach.

anki-code:
> to run the latest xonsh with a single command

This sounds awesome! Let's do it here.

Please add all info about using and running to https://xon.sh/dev/platforms.html - `docs/platforms.rst`

SamLukeYes:
> Please add all info about using and running to https://xon.sh/dev/platforms.html - docs/platforms.rst

Maybe docs/install.rst is a better place for this, as it should also work on MacOS, although I don't have the resource to test it. Should we also mention the xonsh package and NixOS modules in nixpkgs?

anki-code:
It depends.

In the Xonsh installation guide, we emphasize that the mamba installer creates an isolated environment from which Xonsh is run. This is the ideal scenario—an entirely independent environment that can be accessed via xpip and xcontext.

In the same guide, system packages are described as packages that are always lagging behind, tightly integrated into the system, and strongly discouraged for installation for these reasons.

Given these two extremes, we need to clarify how Nix installs Xonsh.

I’m not a Nix user yet, so it’s difficult for me to answer this. My understanding is that Xonsh would be installed in an isolated environment where xpip would not work.

Since in the Install Guide we want to provide both installation and usage instructions, I think it would make sense for you to add a separate install.rst section for Nix and describe the installation, usage, and update workflow there. I can adjust the section placement if needed.

anki-code:
As for the ability to install Xonsh from the main branch (and it would be great to add an example of installing from any branch), this would definitely be helpful for the Dev Guide, where there is already a section on Nix.

bestlem:
Whichever way this is done nix runs xonsh in an independent environment no need for mamba etc. The dependent packages are passed in using the extraPackages argument to the nix code that sets up xonsh.

Running via nix run is useful (although you can do this for a released version using the nixpkgs setup as well). 
```nix run "nixpkgs#xonsh"```

The nixpkgs use needs to be documented here as that will be the usual way nix users will install xonsh - most people just want the latest release and don't want the latest main and do not want to manage the xonsh entry in their local nix setup.

The advantage of this being in the xonsh repo is that it would not need to be edited (the commit and its hash)  each time the HEAD commit changed. The disadvantage is that if nixpklgs xonsh receipe changed (except for the commit) then the nix file in the xonsh repo should be changed.

As for testing on a Mac - the same git actions used to do a nix run test can do that on both Linux and macOS.

I did a `nix build` on my macOS 26.4 machine and it ran and passed all the tests that are used in nix.

**Comment by anki-code** [comment]
We're very welcome to have nix support in our .github/workflows and clear description with repeatable commands in docs/.

**Comment by anki-code** [comment]
Just for sync. We're going to have a release on Apr 15. If you want to have nix package merged to xonsh repo in 0.23.0 let us know.
For now I feel the lack of docs in install.rst and devguide.rst.

**Comment by bestlem** [comment]
One problem with nix is there are many ways to install programs so installation instructions  probably should not be here except to mention that xonsh is in nixpkgs and installed like any other package, and perhaps nix run as an example. 
Doing a nix build to build a test version might make sense as well.

**Comment by anki-code** [comment]
I still don't understand the nix workflow.

Ubuntu with mamba xonsh-env:
```xsh
# xonsh in ~/.local/xonsh-env and in $PATH
xonsh # run xonsh from any place
xpip install lolcat # install any package in xonsh environment
xbin-add lolcat  # make it visible
lolcat  # use it
```

(stupid thoughts start) If on Nix 1. xonsh run via `nix-shell` and totally isolated 2. it can't install packages via xpip, so how I can use it day by day? What I'm doing if I want to install package like `xpip`? How I run xonsh in other nix environments if next `nix-shell` will create a new one without xonsh? I don't understand how to use xonsh as core base shell that can be accesible from any place. (stupid thoughts stop)

bestlem:
There are several nix flows.
Your question also shows why giving the details in xonsh documentation rather than nix is an issue.

In general you need to learn and setup nix first.

I think you can use nix in the way you have shown but I don't use nix-shell so can't give details.

In general nix is declarative in that you define all the packages (e.g. xontribs and python packages) up from with the install instructions so xpip etc is never used.

I use flakes (which a lot of nix people say is wrong and a group in their fork are beginning to remove).

In the file I have something like
```

{
  pkgs,
  perSystem,
  ...
}:
let

  pythonPkgs =
    pypkgs:
    with pypkgs;
    [

      # General
      lxml

      # scientific
      numpy
      scipy
      pandas
      matplotlib

      # Language server
      python-lsp-server
      jedi

      # xonsh deps
      ply
      pygments
      prompt-toolkit
      setproctitle

      # OS process helpers
      plumbum


    ]
    ++ (with perSystem.nur.repos.xonsh-xontribs; [
      xontrib-direnv
      xontrib-readable-traceback
      xontrib-term-integrations

    ]);
in
{
  imports = [ ./config ]; # xonshrc
  home.packages = [
    (pkgs.xonsh.override { extraPackages = pypkgs: (pythonPkgs pypkgs); }) # Pass python packages to xonsh path
  ];

}

```

But this depends on flakes, a particular flake framework (blueprint) , a xontrib nur repo (that probably should be in xonsh docs)

I'll bet noone else uses xonsh and nix in the same way as this.

anki-code:
Ok, what we can do to start support nix-compatible repository? Just merge this config? Is it ready?

bestlem:
As the only users of this will be using the main branch or a local copy of the repo it is independent of any xonsh release.

The code seems to be OK in that I have used it to build a working xonsh and the author is the maintainer of the nixpkgs package so matches that build. It probably does need documentation  on what `nix build` produces in xonsh documentation. Ideally it should support `nix-build` as well.

I would like comments from others who know nix if I have got things wrong or not - especially @SamLukeYes

SamLukeYes:
> The disadvantage is that if nixpklgs xonsh receipe changed (except for the commit) then the nix file in the xonsh repo should be changed.

Come to think of it, what I expect is quite the opposite -- if we build from the latest source frequently, we should be more likely to notice and fix breakages of xonsh packaging ahead of time, so that we can be aware of the actions need to be done in nixpkgs on the next xonsh release.

> Is it ready?

Currently this PR contains a flake that provides xonsh packaged with multiple python versions as well as a github action that updates flake.lock every week. It's ok to merge as a first step, or you can wait for a few days so that I can add some docs (maybe some basic usage in dev guide) when I have time :)

anki-code:
>  so that I can add some docs

Go for it!

goodboy:
@SamLukeYes if u need any help testing this please do ping!

i've been waiting for this more serious support for a while now 🙏🏼

anki-code:
Hi @SamLukeYes! Thank you for continuing the effort!
Do you have  plans to improve [install.rst](https://xon.sh/dev/install.html#system-package)?

bestlem:
The edit for testing using nix looks good to me.

As for install.rst - nix does not fit in the System Packages part as written as all three recommendations not to use it are false for nix. It is simpler to leave nix out.

I think the instructions would be something like the following which gives four choices and I use a different way.

xonsh is in the nix package manager [nixpkgs](https://search.nixos.org/packages?channel=unstable&include_modular_service_options=1&include_nixos_options=1&query=xonsh&show=xonsh) See the entry for xonsh there for various installation instructions.

anki-code:
@SamLukeYes @bestlem could you please provide at least one way to install xonsh as core shell on nix? We have no goal to cover all use cases in install.rst but let's add at least one repeatable way and note about there are many others.
(As I understand `nix-shell -p xonsh` - is working just for testing)

If it's not the good way for Nix I'm ok to merge this without description.

anki-code:
I'm waiting the go-nogo comment here to merge or wait.

SamLukeYes:
> could you please provide at least one way to install xonsh as core shell on nix?

Nix package itself does not provide system integration. It can be done with NixOS modules on NixOS, and maybe other ways on other distros. Whether we should add this to Xonsh docs or not, I think it is out of the scope of this PR.

bestlem:
I would say go for it.

nix/unwrapped.nix
```diff
@@ -0,0 +1,146 @@
...
+    pip
+    pyte
+    pytest-mock
+    pytest-rerunfailures
+    pytest-subprocess
+    pytestCheckHook
+    requests
+
+    # required by test_xonsh_activator
+    virtualenv
```
anki-code, line 62: Just want to let you known that xonsh shell support was reintroduced in vitualenv (https://github.com/pypa/virtualenv/pull/3125)

anki-code:
@SamLukeYes thanks for pushing all forward and welcome!

anki-code:
@bestlem @goodboy thanks for support!

PR #6290, opened 2026-04-11
feat(main): acquire controlling terminal at startup (terminal handshake)

anki-code:
## Motivation

I tested xonsh 0.23.0 with Flatpak package and noticed that xonsh has no handshake with terminal at start. After this I used code analyzer to deep dive and implement basic logic of handshake.
I've tested this manually and with analyzer on Mac, Windows and Linux. 
This handshake is an additive change and it's the minimal standard to have it for the shell.

After this I found an issue that seen before (Bad file descriptor debugging) - it was also about lack of handshake - 99% CPU load and fix it as well. 
After this fix xonsh more organically integrated with OS and do not produce 99% CPU on signal flood in edge cases.

So below I provided the report from code analyzer.

------------------

## Summary

xonsh now performs the same TTY startup handshake that bash, zsh and other interactive shells do: it makes itself the foreground process group of its controlling terminal before `premain` runs. The implementation lives in `xonsh/main.py` as four small helpers (`_acquire_controlling_terminal`, `_release_controlling_terminal`, `_setup_controlling_terminal`, `_handle_sig_ttin_ttou`) and is called from both `main()` and `main_xonsh()`. A new section in `docs/launch.rst` explains the full rationale and the failure modes it fixes. Verified end-to-end inside a sandboxed environment.

## Problem

Historically xonsh relied on its parent shell to have already placed it in the foreground process group of whatever TTY it was attached to. That assumption holds when xonsh is launched from a well-behaved interactive shell, but it breaks in several real-world environments:

- **Bubblewrap and similar sandboxes** — `bwrap` does not reassign the TTY's foreground group to the sandboxed child.
- **Build systems, CI runners, IDE integrated terminals** and service managers like `systemd --user` that launch xonsh via `posix_spawn` without wiring up job control.
- **Nested containers** that reuse the outer PTY.

In those cases xonsh is a background process of its TTY. Every terminal operation — `tcgetattr` during prompt_toolkit init, `tcsetpgrp` when launching a pipeline, plain writes under `TOSTOP` — raises `SIGTTIN` or `SIGTTOU`. Two failure modes follow:

1. **asyncio wakeup pipe overflow** — Python routes signals to a non-blocking self-pipe so the main event loop can wake up. Under a startup signal storm the pipe overflows and xonsh crashes with `BlockingIOError: [Errno 11] Resource temporarily unavailable`.

2. **termios EINTR** — low-frequency signals (SIGCHLD, SIGWINCH) interrupt `tcsetattr` calls and surface as `termios.error: (4, 'Interrupted system call')`.

Users in affected environments worked around this with a third-party launcher wrapper that installed `SIG_IGN` for `SIGTTIN`/`SIGTTOU` before importing xonsh.

## Solution

Perform the handshake bash has always done, adapted for Python. `_setup_controlling_terminal` works like this:

1. **Gate on POSIX.** Windows returns immediately.
2. **Install a livelock-guarded Python no-op handler for `SIGTTIN`/`SIGTTOU` unconditionally** — matches the historical behavior that `main_xonsh` had before this change. The handler is `_handle_sig_ttin_ttou`: under normal operation it is a cheap counter bump, above a high threshold it escalates to `SIG_IGN`.
3. **Gate the handshake itself on stderr being a real TTY.** Script mode, piped input, redirected stderr and pytest runs stop after step 2.
4. **Check if we are already foreground via `tcgetpgrp == getpgrp && pgid > 0`.** The `pgid > 0` guard handles PID namespaces where the kernel returns `0` for outer pgids.
5. **Block `SIGTTOU`, `SIGTTIN`, `SIGTSTP`, `SIGCHLD`** in the calling thread with `pthread_sigmask`.
6. **Call `setpgid(0, 0)`** to become our own process group leader.
7. **Call `tcsetpgrp(tty_fd, getpgrp())`** to install our new group as the TTY's foreground.
8. **Remember the previous foreground group** and register an `atexit` handler to restore it on shutdown.

If the handshake fails, the step-2 Python handlers are replaced with `SIG_IGN` so the kernel discards `SIGTTIN`/`SIGTTOU` outright and they never reach Python's asyncio wakeup pipe.

The handshake is called from the top of `main()` (so premain / xontribs / xonshrc run with xonsh already foreground) and from `main_xonsh()` as an idempotent backup via the `_tty_setup_done` flag.

## Why before `premain`?

`premain` loads xontribs and runs the user's xonshrc files. Rc files contain `$(...)` / `!(...)` subprocess captures, including interactive tools like `fzf` that take over the TTY via `tcsetattr`. Running the handshake before rc ensures child processes fork from a foreground xonsh and inherit foreground pgrp on the very first instruction, so rc-launched interactive tools do not race with xonsh's pgrp management.

## Escape hatch

`XONSH_NO_FG_TAKEOVER=1` in the parent environment disables the handshake entirely. The step-1 Python handlers still get installed. Documented in `docs/launch.rst`.

## Edge cases handled

**PID namespaces.** Inside a PID namespace (Bubblewrap, Podman rootless, systemd-nspawn, kubectl exec, Chrome/Firefox renderer sandboxes), the Linux kernel returns `0` from `getpgrp()` / `tcgetpgrp()` for pgids inherited from outside the namespace ([`pid_namespaces(7)`](https://man7.org/linux/man-pages/man7/pid_namespaces.7.html)). The fast-path check includes `and my_pgid > 0` so `0 == 0` does not spuriously declare success, ensuring `setpgid(0, 0)` runs and installs a valid namespace-local pgid. Caught by real-world sandbox testing and has a dedicated regression test.

**Livelock guard on the Python no-op handler.** The historical `def func_sig_ttin_ttou(n, f): pass` can feed two independent retry loops into a 99% CPU spin: PEP 475 auto-retry on `os.read` (SIGTTIN path), and the xonsh PR #6192 application-level retry on `tcsetattr` (SIGTTOU path). `_handle_sig_ttin_ttou` counts deliveries and escalates to `SIG_IGN` above `_TTIN_TTOU_LIVELOCK_THRESHOLD` (= 1000), breaking the loop. Under normal operation the handler fires zero times.

**Non-TTY regression preserved.** Step 1 of `_setup_controlling_terminal` installs Python no-op handlers unconditionally, matching the pre-PR behavior for script mode. The handshake itself is gated on `isatty`, the baseline signal policy is not.

## Changes

- `xonsh/main.py`: four new helpers (`_acquire_controlling_terminal`, `_release_controlling_terminal`, `_setup_controlling_terminal`, `_handle_sig_ttin_ttou`), three module-level globals (`_fg_tty_state`, `_tty_setup_done`, `_ttin_ttou_counter`), one constant (`_TTIN_TTOU_LIVELOCK_THRESHOLD`), `_setup_controlling_terminal()` called from the top of `main()` and `main_xonsh()`, old inline `func_sig_ttin_ttou` replaced, `atexit` added to imports. All major design decisions carry inline comments pointing at the relevant man page.
- `tests/test_main_terminal_llm.py`: new file with 22 mock-based unit tests covering acquire/release/setup paths, PID namespace edge case, livelock guard, idempotency, non-TTY path, Windows no-op, `XONSH_NO_FG_TAKEOVER` gate, session leader gate. Uses a `FakeOS` helper and recording stubs so no real kernel state is mutated.
- `docs/launch.rst`: new final section "Controlling Terminal and Foreground Process Group" explaining the POSIX background, failure modes, signal policy, shutdown restore, no-op situations, and escape hatch.
- `docs/embedding.rst`: new stub page documenting how embedded xonsh users can invoke `_setup_controlling_terminal` when they bypass `main()` via `xonsh.main.setup`.

### Prior art

- xonsh PR #6192 — prior partial fix at the `ptk_shell.singleline` level.

### bash source code

Handshake approach and `pgrp_pipe` synchronization verified at https://cgit.git.savannah.gnu.org/cgit/bash.git/tree/jobs.c. Functions verified: `start_pipeline()`, `make_child()`, `stop_pipeline()`, `default_tty_job_signals()`, `ignore_tty_job_signals()`.

### POSIX.1-2017 specifications

- `tcsetpgrp(3)`: https://pubs.opengroup.org/onlinepubs/9699919799/functions/tcsetpgrp.html
- `tcgetpgrp(3)`: https://pubs.opengroup.org/onlinepubs/9699919799/functions/tcgetpgrp.html
- `setpgid(2)`: https://pubs.opengroup.org/onlinepubs/9699919799/functions/setpgid.html
- `pthread_sigmask(3)`: https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_sigmask.html
- `execve(2)`: https://pubs.opengroup.org/onlinepubs/9699919799/functions/execve.html
- General Terminal Interface: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap11.html

### Linux man pages

- `tcsetpgrp(3)`: https://man7.org/linux/man-pages/man3/tcsetpgrp.3.html
- `tcsetattr(3)`: https://man7.org/linux/man-pages/man3/tcsetattr.3.html
- `setpgid(2)`: https://man7.org/linux/man-pages/man2/setpgid.2.html
- `pthread_sigmask(3)`: https://man7.org/linux/man-pages/man3/pthread_sigmask.3.html
- `pid_namespaces(7)`: https://man7.org/linux/man-pages/man7/pid_namespaces.7.html
- `signal(7)`: https://man7.org/linux/man-pages/man7/signal.7.html
- `read(2)`: https://man7.org/linux/man-pages/man2/read.2.html

### Python-specific

- CPython `Modules/termios.c`: https://github.com/python/cpython/blob/main/Modules/termios.c
- PEP 475: https://peps.python.org/pep-0475/
- prompt_toolkit `input/vt100.py`: https://github.com/prompt-toolkit/python-prompt-toolkit/blob/main/src/prompt_toolkit/input/vt100.py

### DIFF
--- a/docs/contents.rst
+++ b/docs/contents.rst
@@ -77,6 +77,7 @@ developer tools.
     env
     completers
     xontrib
+    embedding
     api/index
     Create an extension using template <https://github.com/xonsh/xontrib-template>
 
--- a/docs/embedding.rst
+++ b/docs/embedding.rst
@@ -1,0 +1,140 @@
+.. _embedding:
+
+***************************
+Embedding Xonsh in Your App
+***************************
+
+.. warning::
+
+    This page is a **work-in-progress stub**. It currently documents
+    only the controlling-terminal handshake for embedded use cases.
+    A fuller guide — covering :func:`xonsh.main.setup`, custom
+    :class:`~xonsh.shell.Shell` subclasses, teardown, thread safety,
+    and integration patterns for LLM agents / TUI apps / REPLs —
+    will appear in a later release. If you are embedding xonsh and
+    need guidance that is not covered here, please open an issue or
+    discussion on GitHub.
+
+Xonsh exposes itself as a Python library in addition to the standalone
+``xonsh`` command. Third-party projects can create a xonsh session,
+load xontribs and rc files, and drive the execer / shell from their
+own code. The public entry point for this is :func:`xonsh.main.setup`,
+documented in :mod:`xonsh.main`.
+
+.. contents::
+    :local:
+    :depth: 1
+
+
+Controlling Terminal Handshake for Embedded Interactive Shells
+==============================================================
+
+When xonsh runs as the standalone ``xonsh`` command, its entry point
+(:func:`xonsh.main.main`) automatically performs a startup handshake
+that makes xonsh the *foreground process group* of its controlling
+terminal. This avoids a class of startup crashes and hangs in
+environments where the parent process does not arrange TTY ownership
+correctly — Flatpak / Bubblewrap sandboxes, build systems, nested
+containers, ``systemd --user`` services, some IDE terminals.
+
+See :ref:`launch-fg-takeover` for the full rationale, the failure
+modes it fixes (``BlockingIOError`` on asyncio wakeup pipe,
+``termios.error`` EINTR, and related), the three-step signal policy,
+and the ``XONSH_NO_FG_TAKEOVER`` escape hatch.
+
+When the handshake runs
+-----------------------
+
+The handshake is invoked automatically on **both** of these paths:
+
+* :func:`xonsh.main.main` — the default CLI entry point.
+* :func:`xonsh.main.main_xonsh` — the inner entry point used when
+  some code paths bypass ``main``.
+
+It is **not** invoked from :func:`xonsh.main.setup`. That means an
+embedded project that builds its own interactive shell by calling
+``setup()`` (or by instantiating :class:`~xonsh.shell.Shell`,
+:class:`~xonsh.execer.Execer` and :class:`~xonsh.built_ins.XonshSession`
+directly) will **not** get the handshake for free. If your embedded
+xonsh runs in one of the affected environments, you need to invoke
+it yourself.
+
+How to invoke it from embedded code
+-----------------------------------
+
+The helper currently lives in ``xonsh.main`` as a module-private
+function: ``_setup_controlling_terminal``. It is private by
+convention (leading underscore) but the implementation is stable and
+idempotent. A future release is expected to expose a public alias;
+until then, calling the private function directly is the supported
+path.
+
+Call it **before** you start your interactive shell loop — ideally as
+early in your program's startup as possible, so that any xonshrc or
+xontrib code your embedder runs already has foreground ownership:
+
+.. code-block:: python
+
+    # embedded_launcher.py
+    from xonsh.main import setup
+    from xonsh.main import _setup_controlling_terminal
+
+    # Acquire foreground of controlling TTY (idempotent, safe to
+    # call multiple times — only the first call does real work).
+    # No-op on Windows, in non-TTY contexts (pytest, piped input,
+    # redirected stderr), and when XONSH_NO_FG_TAKEOVER=1 is set.
+    _setup_controlling_terminal()
+
+    # Your existing xonsh setup stays unchanged.
+    setup(
+        shell_type="prompt_toolkit",
+        # ... your ctx, env, xontribs, aliases, etc.
+    )
+
+    # Your custom shell / REPL / agent loop starts here.
+
+What the helper does
+--------------------
+
+The call has three possible outcomes, all handled internally:
+
+1. **Fast path** — xonsh is already the foreground process group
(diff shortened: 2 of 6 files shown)


anki-code:
I've test it on many systems: Windows, WSL, Mac, Linux, with Flatpak, with AppImage and it works much better than before.

PR #6293, opened 2026-04-12
feat(history): Reimplement sqlite history backend `erasedups` and add json history backend support

### LINKED ISSUE #5919: `history clear` deletes entries from global history
## Current Behavior

With SQLite backend and erasedups, `history clear` deletes entries from global history, i.e if `ls` is used in the session and had been used previously, it will be deleted from the history database. 

Config: `xonsh --no-rc --no-env --shell-type prompt-toolkit -D XONSH_HISTORY_BACKEND=sqlite -D HISTCONTROL='erasedups'`

1. Open new xonsh session, and type a command that will be saved (e.g. `foobar`)
2. Close that session to save it to history.
3. Open a new xonsh session, and enter the same command as above and another
bogus command (e.g. `bazbar`).
4. Run `history clear`, and close the session.
5. Open a new xonsh session, and notice that `foobar` is no longer in history.

## Expected Behavior

`foobar` is still in history, but the new command `bazbar` is not in history.

`history clear` (by my interpretation) should effectively be as if you just started a new session -- as if nothing had ever happened. Repeated existing commands should still be untouched in history, and new commands should not be in history. Deleting existing entries goes against this by having a side-effect, and instead acting like `history delete`.

## xonfig

<details>

```xsh
| xonsh                       | 0.19.9          |
| Python                      | 3.13.7          |
| prompt toolkit              | 3.0.52          |
| shell type                  | prompt_toolkit  |
| history backend             | sqlite          |
| pygments                    | 2.19.2          |
| on linux                    | True            |
| distro                      | unknown         |
| on wsl                      | False           |
| on darwin                   | False           |
```

</details>

## For community

anki-code:
### Motivation

After reviewing the erasedups option for the SQLite HISTCONTROL, it turned out to have multiple issues, edge cases, and unintended side effects. For example, when combined with history clear, it may result in commands being permanently removed from history.

Moreover, the cleanup routine was triggered on every write to the database, and it was performed synchronously, which caused increasing slowdowns as the history database grew.

The behavior of `erasedups` becomes unpredictable when multiple sessions are running in parallel, especially when combined with `history clear`.

* Fixes https://github.com/xonsh/xonsh/issues/5919

### Implementation

* The `erasedups` option has been removed from HISTCONTROL.
* A dedicated command, `history erasedups`, has been introduced, allowing users to run deduplication explicitly when needed.
* Additionally, `history erasedups` now supports JSON history.

## For community

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,7 +18,7 @@ jobs:
           - "3.11"
           - "3.12"
           - "3.13"
-          - "3.14"          
+          - "3.14"
     steps:
       - uses: actions/checkout@v6
       - name: setup-python
--- a/appimage/xonsh.appdata.xml
+++ b/appimage/xonsh.appdata.xml
@@ -6,9 +6,9 @@
     <name>Xonsh</name>
     <summary>Xonsh shell on Python {{ python-fullversion }}</summary>
     <description>
-        <p>Xonsh shell, bundled with Python {{ python-fullversion }}. Xonsh (sounds like "consh") is a modern, 
-           full-featured and cross-platform Python-based shell. The language is a superset of Python 3 with 
-           seamless integration of shell functionality and commands. Xonsh is meant for the daily use of experts 
+        <p>Xonsh shell, bundled with Python {{ python-fullversion }}. Xonsh (sounds like "consh") is a modern,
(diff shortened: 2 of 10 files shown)


anki-code:
Tested on SQLite and Json history backends on my personal history with 10 000 lines.

PR #6294, opened 2026-04-12
fix(events): Do not strip leading whitespace from command before going thru pre-cmd events

### LINKED ISSUE #5920: on_precommand hook strips leading whitespace from command
## Current Behavior
The cmd argument passed to the `on_precommand` hook seems to have the leading spaces removed.

Running the command `  echo test` with the following hook:
```py
@events.on_precommand
def _test_precommand(cmd: str):
    print(cmd)
```
will print `echo test`, without any spaces at the start.

## Expected Behavior
The cmd argument should match what the user typed in, without stripping spaces.

This is relevant for tools like Atuin, as commands starting with a space are supposed not to be logged.
But since the hook strips leading whitespace, it's doesn't seem possible to ignore those at the moment.

## For community

anki-code:
### Motivation

* Fixes https://github.com/xonsh/xonsh/issues/5920

## For community

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,7 +18,7 @@ jobs:
           - "3.11"
           - "3.12"
           - "3.13"
-          - "3.14"          
+          - "3.14"
     steps:
       - uses: actions/checkout@v6
       - name: setup-python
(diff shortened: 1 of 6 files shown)


anki-code:
No current tests are broken BUT it's possible to have a regress on users side if someone so not process leading (lstrip) spaces BUT it's not the common case it's edge case SO I believe we can release it in this major version with attention in release notes.

PR #6298, opened 2026-04-13
fix: Ignore empty strings in completion fallback chain (Fixes #5810)

### LINKED ISSUE #5810: Refactoring: don't allow completer return empty values
To prevent issues like https://github.com/xonsh/xonsh/issues/5809 in general we need to treat the completers which returns the list of empty lines like return None. Start point is [here](https://github.com/xonsh/xonsh/blob/bb75a5a53ba6b5e4ed60ae88839283075e9f47dc/xonsh/completer.py#L312).

## For community

DeineBruder:
My solution was line 239, when getting the items of completion. Before adding a completion, I skip it if its empty or just consisting of spaces or /t.

cc https://github.com/xonsh/xonsh/issues/5810

## For community

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -238,6 +238,9 @@ class Completer:
 
             items = []
             for comp in res:
+                if not comp or not str(comp).strip():
+                    continue
+                
                 if (not is_filtered) and (not filter_func(comp, prefix)):
                     continue
                 comp = Completer._format_completion(
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -240,7 +240,7 @@ class Completer:
             for comp in res:
                 if not comp or not str(comp).strip():
                     continue
-                
+
                 if (not is_filtered) and (not filter_func(comp, prefix)):
                     continue
                 comp = Completer._format_completion(


anki-code:
Thanks for the proposal. Fixed in https://github.com/xonsh/xonsh/pull/6311

DeineBruder:
@anki-code, Could you please provide me with an issue I could work on, with hopes on getting merged ?

If there is any issue with my code I would be happy to correct it or change it.

anki-code:
@DeineBruder, at the moment we are using a fairly strong and well-tuned LLM for in-depth problem analysis. I’m quite satisfied with the results, and it’s difficult for me to point to a ticket that a human would clearly solve better from a coding perspective.

That said, I do believe that human judgment is still essential when it comes to product decisions and use cases — answering questions like “how should xonsh behave in a given situation?” requires thoughtful consideration and a deeper understanding of context. (LLMs can be helpful here as well, but they don’t fully replace that kind of reasoning.)

So, to be completely honest, many of the recent pull requests I’ve seen have been handled very effectively by our LLM from an implementation standpoint. If you’d like to contribute to xonsh and you actively use and care about the project, one of the most valuable ways to help would be to go through existing issues and propose well-thought-out tickets — for refactoring or broader improvements that can help move xonsh forward.

PR #6309, opened 2026-04-13
feat(completer): show per-item completion source in trace output

### LINKED ISSUE #6292: completer: Add the source of completion for item in debug mode
```xsh

$XONSH_TRACE_COMPLETIONS=True

dir<Tab>
# dir <- bash_completer 
# directory() <- function
# dir_setup <- fish completer
# dirmode <- aliases
```
The tracing should work like `$XONSH_COMMANDS_CACHE_DEBUG=1` is working - tracing rows are above the shell:

```xsh
$XONSH_COMMANDS_CACHE_DEBUG=1
# xonsh-commands-cache: not found `d` (0.0026 sec)
# xonsh-commands-cache: not found `di` (0.0018 sec)
# xonsh-commands-cache: get from cache `/opt/gnubin/dir` (0.0003 sec
dir
```

## For community

yPin9:
# Summary

This improves `XONSH_TRACE_COMPLETIONS` by showing per-item completion sources in trace output.

## What changed

* added `source` metadata to `RichCompletion`
* updated trace output to print `item <- source`
* preserved more specific sources when completions are wrapped by higher-level completers
* added finer-grained sources for command and Python completions where available

## Before

Trace mode reported which completer produced a batch of results, but did not show the source for each individual completion item.

## After

Trace mode now shows per-item sources, for example:

```text id="5vdj0i"
TRACE COMPLETIONS: Got 30 results from exclusive completer 'base':
# os.path <- module
# os.PathLike <- class
# os.posix_spawnp( <- function
# dirmode <- aliases
# lsusb <- command
```

## Notes

* this only affects trace/debug output
* normal completion behavior is unchanged
* when a completion does not provide a more specific source, trace output falls back to the completer name

Closes #6292

LLM assistance was used during development for patch planning, debugging, and drafting code changes. The final code was reviewed and edited manually.

## For community

### DIFF
--- a/xonsh/completer.py
+++ b/xonsh/completer.py
@@ -164,7 +164,6 @@ class Completer:
                     completion = completion + closing_quote
 
         completion = list(apply_lprefix([completion], lprefix))[0]
-
         if (
             isinstance(completion, RichCompletion)
             and completion.append_space
@@ -258,7 +257,10 @@ class Completer:
                     f"TRACE COMPLETIONS: Got {len(items)} results"
                     f" from {'' if is_exclusive_completer(func) else 'non-'}exclusive completer '{name}':"
                 )
-                sys.displayhook(items)
+                for completion, _ in items:
+                    shown = getattr(completion, "display", None) or str(completion)
+                    source = getattr(completion, "source", None) or name
+                    print(f"# {shown} <- {source}")
 
             if is_exclusive_completer(func):
                 # we got completions for an exclusive completer
--- a/xonsh/completers/base.py
+++ b/xonsh/completers/base.py
@@ -32,7 +32,6 @@ def _with_source(completions, source):
                 source=source,
             )
 
-
 @contextual_completer
 def complete_base(context: CompletionContext):
     """If the line is empty, complete based on valid commands, python names, and paths."""
--- a/xonsh/completers/base.py
+++ b/xonsh/completers/base.py
@@ -32,6 +32,7 @@ def _with_source(completions, source):
                 source=source,
             )
 
+
 @contextual_completer
 def complete_base(context: CompletionContext):
     """If the line is empty, complete based on valid commands, python names, and paths."""
--- a/xonsh/completers/base.py
+++ b/xonsh/completers/base.py
@@ -5,8 +5,32 @@ import collections.abc as cabc
 from xonsh.completers.commands import complete_command
 from xonsh.completers.path import contextual_complete_path
 from xonsh.completers.python import complete_python
-from xonsh.completers.tools import apply_lprefix, contextual_completer
+from xonsh.completers.tools import (
+    RichCompletion,
+    apply_lprefix,
+    contextual_completer,
+)
 from xonsh.parsers.completion_context import CompletionContext
+
+
+def _with_source(completions, source):
+    for comp in completions:
+        if isinstance(comp, RichCompletion):
+            yield RichCompletion(
+                str(comp),
+                prefix_len=comp.prefix_len,
+                display=comp.display,
+                description=comp.description,
+                style=comp.style,
+                append_closing_quote=comp.append_closing_quote,
+                append_space=comp.append_space,
+                source=getattr(comp, "source", None) or source,
+            )
+        else:
+            yield RichCompletion(
+                str(comp),
+                source=source,
+            )
 
 
 @contextual_completer
@@ -22,9 +46,12 @@ def complete_base(context: CompletionContext):
     python_comps = complete_python(context) or set()
     if isinstance(python_comps, cabc.Sequence):
         python_comps, python_comps_len = python_comps  # type: ignore
-        yield from apply_lprefix(python_comps, python_comps_len)
+        yield from _with_source(
+            apply_lprefix(python_comps, python_comps_len),
+            "python",
+        )
     else:
-        yield from python_comps
+        yield from _with_source(python_comps, "python")
 
     # add command completions
     yield from complete_command(context.command)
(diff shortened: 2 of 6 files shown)


anki-code:
Hey, I like how it looks but some thoughts.

The PR description says "normal completion behavior is unchanged", but `attr_complete` in `python.py` replaces `_safe_eval` validation with a bare `try: getattr()`. The original code ran the full expression through `_safe_eval` (xonsh eval with user ctx + builtins fallback) and skipped attributes where both attempts failed. The new code uses direct `getattr(val, opt)` which is more permissive and will surface completions that were previously filtered out. This needs to be either reverted to preserve existing behavior, or explicitly described as an intentional change with justification.

In `commands.py` the last `RichCompletion(os.path.join(base, i))` gained `append_space=True` which was not there before. This changes completion behavior for executables found in directories — a space will now be appended after them. This is unrelated to the trace feature and should be reverted or split into a separate commit.

The original code iterated `ctx` keys as strings: `{s for s in ctx if filt(s, prefix)}`. The PR accesses `ctx[s]` to classify the object, but `ctx` can be a `ChainMap` or custom mapping where `__iter__` yields keys not safely accessible via `__getitem__`. Use `ctx.get(s)` with a fallback instead.

yPin9:
Thanks for the review.

I pushed updates addressing all three points:

* restored the original`_safe_eval()` -based filtering in `attr_complete()` so normal completion filtering behavior is preserved
* removed the unrelated command completion behavior change from`commands.py`
* switched the context-based source lookup to safe `ctx.get(..., fallback)` access instead of direct indexing

anki-code:
Thanks for the iteration — all three review points are addressed correctly: the `_safe_eval`-based filtering in `attr_complete` is restored, the stray `append_space=True` on the last `RichCompletion` in `commands.py` is gone, and `ctx` access is now safe via `ctx.get(..., sentinel)` guarded with `hasattr`. CI is green across the whole Ubuntu/macOS/Windows × Python 3.11–3.14 matrix, backward compatibility via `source=None` is preserved, the priority rule for more specific sources when wrapping (`getattr(comp, "source", None) or source`) is a nice design choice, and the docstring on the new `source` parameter of `RichCompletion` clearly marks it as debug-only. Before merge, though, a few things would be worth closing out.

First and most important — there are no tests in this PR, and the project's `CLAUDE.md` explicitly requires tests to accompany every change; at minimum we'd want coverage for the trace output format (`# item <- source`), for `source` propagation through the `_with_source` wrapper layer, and for the object classification in `_python_source_from_obj` / `_attr_source` across class/module/function/mapping/sequence cases.

Second — there's a lot of noise in `xonsh/pyghooks.py` (+64/-16): all of those changes are pure reformatting — single quotes to double quotes and splitting long tuples — and none of it is related to the PR topic. It looks like a side effect of the branch not being rebased on current master (those lines landed with merge #6310 on April 14), and `pre-commit.ci` auto-fix reformatted them. Please rebase on the current master and revert the formatting, or move the reformat into a separate PR.

Third, performance — `_with_source` in `xonsh/completers/base.py` wraps every completion in a fresh `RichCompletion` even when `$XONSH_TRACE_COMPLETIONS=False`, but completion is a hot path (fires on every Tab/keypress). Given that `source` is, in your own words, "primarily intended for debugging," it would be more correct to either wrap only when trace is enabled, or to drop the wrapper entirely and let `source` be set only at the original producers rather than bolted on by a wrapper layer.

Fourth — `RichCompletion.replace(**kwargs)` already exists in `xonsh/completers/tools.py:113` and is designed for exactly this kind of transformation; the current `_with_source` and `_rich_with_source` manually enumerate every attribute (`prefix_len`, `display`, `description`, `style`, `append_closing_quote`, `append_space`) and will silently drop any new parameter added to `RichCompletion.__init__` later on. Replacing the manual construction with `comp.replace(source=getattr(comp, "source", None) or source)` is both shorter and more robust.

Fifth — `_python_source_from_obj` and `_attr_source` are near-duplicates, differing only in the `mapping`/`sequence` branches; they should be merged into one function with a flag.

Sixth — the semantics of source labels is inconsistent: some labels describe *where the completion came from* (`aliases`, `commands_cache`, `cwd`, `path`, `python`), while others describe *what the completion is* (`function`, `class`, `module`, `attribute`, `mapping`, `sequence`, `keyword`, `function_arg`). From the trace output alone it's not always obvious what a given label means; please either document the allowed values directly in the `RichCompletion.source` docstring, or align them to a single axis of meaning. Related nit: `function_arg` is inaccurate — `python_signature_complete` returns keyword-argument names shaped like `foo=`, so `kwarg` or `signature` would fit better.

Seventh, and important because it's not documented in the PR — the line `a = getattr(val, opt)` was removed in `attr_complete`, and all subsequent classification (`callable`, `isinstance(_, cabc.Sequence | cabc.Mapping)`, `inspect.isclass`, `inspect.ismodule`) now operates on `_val_`, the result of `_safe_eval(_expr, _ctx)`. In 99% of cases this is equivalent to the previous `getattr(val, opt)`, but when `_safe_eval` fails on its first attempt and falls back to `builtins.__dict__`, `_val_` is evaluated in the builtins context and may be a different object than `getattr(val, opt)`. This is a subtle behavioral change, mentioned in neither the PR description nor the commit message — please either restore `a = getattr(val, opt)` to keep the semantics unchanged, or explicitly explain in the description why `_val_` is preferable here.

Minor nits: in `_rich_with_source`, `**kwargs` is silently ignored when `value` is already a `RichCompletion`, which is an asymmetric and error-prone API; a blank line after `completion = list(apply_lprefix([completion], lprefix))[0]` in `xonsh/completer.py` was removed for no stated reason; and the three commits, including `[pre-commit.ci] auto fixes`, should be squashed at merge time.

Bottom line: the feature is close to merge and the output format matches what #6292 asked for; blocker — tests; important — drop the `pyghooks.py` noise via rebase and eliminate the wrapping overhead when trace is disabled; the rest (using `replace()`, merging the two classification helpers, documenting the source-label vocabulary, explaining the `_val_` switch) is nice-to-have.

anki-code:
Thank you for your try @yPin9! I'm going to close this with the reference to https://github.com/xonsh/xonsh/pull/6311

Issue #5809, opened 2025-03-17
Empty completion list for file in `aws s3 cp`


anki-code:
Typing on Debian:

```xsh
pip install awscli

$XONSH_TRACE_COMPLETIONS=True
aws s3 cp <press tab to insert filename> s3://path/
```

Result

<img width="139" alt="Image" src="https://github.com/user-attachments/assets/411ccc3e-492e-43c1-af1a-8315e97f5e3d" />

Trace (one result with empty space `' '`):
```xsh
TRACE COMPLETIONS: Getting completions with context:
TRACE COMPLETIONS: Got 1 results from exclusive completer 'bash':
CompletionContext(command=CommandContext(args=(CommandArg(value='aws', opening_quote='', closing_quote='', is_io_redir=False), CommandArg(value='s3', opening_quote='', closing_quote='', is_io_redir=False), CommandArg(value='cp', opening_quote='', closing_quote='', is_io_redir=False), CommandArg(value='s3://path/', opening_quote='', closing_quote='', is_io_redir=False)), arg_index=3, prefix='track', suffix='', opening_quote='', closing_quote='', is_after_closing_quote=False, subcmd_opening=''), python=PythonContext('aws s3 cp track s3://path/', 15, is_sub_expression=False))
[(RichCompletion(' ', prefix_len=5, append_closing_quote=False, append_space=True),
  5)]
```

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #6312, opened 2026-04-14
fix: filter out empty/whitespace-only completions

### LINKED ISSUE #5810: Refactoring: don't allow completer return empty values

Ai-chan-0411:
## Summary

Fixes #5810 and #5809.

Completers returning only empty or whitespace-only values (e.g. `{" "}`) now have those values filtered out before processing. If all returned values are empty/whitespace-only, the completer result is treated as `None` (no completions), preventing empty entries from appearing in the tab-completion menu.

## Changes

- **`xonsh/completer.py`**: Added filtering of empty/whitespace-only completion values in `generate_completions()` after the `res is None` check
- **`tests/test_completer.py`**: Added `test_empty_completions_filtered` covering whitespace-only strings, empty strings, `RichCompletion` with whitespace, and mixed valid/empty completions

## Test plan

- [x] All 13 existing `test_completer.py` tests pass
- [x] New `test_empty_completions_filtered` passes with 4 sub-cases
- [ ] Manual verification: `aws s3 cp <tab>` no longer shows empty completion entry

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/tests/test_completer.py
+++ b/tests/test_completer.py
@@ -44,6 +44,23 @@ def test_sanity(completer, completers_mock):
     )
 
 
+def test_empty_completions_filtered(completer, completers_mock):
+    """Completers returning only empty/whitespace values should be treated as
+    returning no completions (see #5809, #5810)."""
+    # whitespace-only completion:
+    completers_mock["a"] = lambda *a: {" "}
+    assert completer.complete("pre", "", 0, 0) == ((), 0)
+    # empty string completion:
+    completers_mock["a"] = lambda *a: {""}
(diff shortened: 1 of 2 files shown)


anki-code:
Already solved, thanks

PR #6442, opened 2026-05-17
fix(execer): parse multiline @(…) with && in phase-1 recovery

### LINKED ISSUE #6011: parser: SyntaxError: multiline text and logical and
```xsh
echo @("""1""") && echo 2
# 1
# 2

echo @("""1
""") && echo 2
#   File "<stdin>", line 3
#    && echo 2]
# SyntaxError: ('code: and',)
```

## For community

HopperAlex04:
Phase-1 subprocess wrapping broke when a logical line joined across physical lines contained a multiline triple-quoted @() argument and a shell combinator (e.g. echo @(1
…) && echo 2). splitlines() cut
inside the string, replace_logical_line mis-split wrapped source, and greedy wrapping put && inside a single ![] block.

Keep embedded newlines as a placeholder in the recovery buffer while subproc_toks runs, parse restored source each attempt, wrap at the first &&/|| break then the RHS on the next pass, and collapse continuation lines to one buffer row. Only treat an empty Interactive/Module as incomplete while the placeholder remains, so empty RC files and newline-only input still compile.

Fixes #6011

This is my first PR on anything so feel free to tear into it. The core of the fix is above. The core change is that the example syntax error presented in issue #6011 now will run as expected. Added a test for the fix in test_excer_llm.py, let me know if more is required.

Example from issue:
```
echo @("""1""") && echo 2    # works
echo @("""1
""") && echo 2               # SyntaxError: ('code: and',)
```

Will now output:

Important Notes: The usage of \x00 as _LOGICAL-NL could cause problems if someone actually uses the character, as the character gets replaced by a newline during restoration. 

'''
1

2
'''

## For community

### DIFF
--- a/tests/test_execer_llm.py
+++ b/tests/test_execer_llm.py
@@ -96,6 +96,26 @@ def test_andor_chain_eval_mode(line, xession):
     )
 
 
+# --- GH-6011: multiline @("""…""") with && --------------------------------
+
+
+def test_multiline_atlparen_triple_string_and_chain(xession):
+    """Phase-1 recovery must wrap both sides of ``&&`` on a logical line that
+    contains a multiline triple-quoted ``@()`` argument (issue #6011).
+    """
+    execer = xession.execer
+    ctx = {"__xonsh__": object()}
+    src = (
+        'echo @("""1\n'
+        '                                                             """) && echo 2\n'
+    )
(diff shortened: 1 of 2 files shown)


anki-code:
Hi, thanks for the effort! I've prepared the solution without \x00 here - https://github.com/xonsh/xonsh/pull/6465

PR #6451, opened 2026-05-20
test: Make inspector and xontribs smoke tests robust to shell selection

rautyrauty:
## Motivation

The xonsh pytest plugin calls `xonsh.main.setup()` in `pytest_configure`,
which assigns `XSH.shell` a real `Shell`. Which concrete implementation ends
up in `XSH.shell.shell` is resolved from the environment by
`Shell.choose_shell_type` / `construct_shell_cls` — `$TERM`, whether stdin is
a terminal, prompt_toolkit availability.

These two LLM-generated smoke tests take no session fixture, so
`format_color()` / `print_color()` dispatch into whatever shell that
resolution produced. Shell implementations disagree on the result:
`BaseShell.format_color` returns a plain `str`, while a tokenizing shell
(e.g. `PromptToolkitShell`) returns a pygments token list, and
shell-dispatched output is written at the file-descriptor level rather than
to `sys.stdout`.

The tests assumed the string-shell outcome, so they pass under some
environments and fail under others. This surfaced as a hard test failure
in a downstream distro build whose chroot resolves to a tokenizing shell:

\`\`\`
test_inspector_format_fields_str_smoke
  assert 'Title' in [(Token.Color.BOLD_RED, 'Title:'), (Token.Color.RESET, '  value\n')]
test_xontribs_list_human_format
  assert ('loaded' in '' or 'not-loaded' in '')   # capsys saw nothing
\`\`\`

## Changes

| Test | Before | After |
|------|--------|-------|
| `test_inspector_format_fields_str_smoke` | `assert "Title" in out` — fails when `out` is a token list | normalize `out` to text first, then assert |
| `test_xontribs_list_human_format` | `capsys` — empty when output goes through the shell | `capfd` — captures fd-level output |

Both make the tests pass regardless of which shell the environment resolves
to. Test-only change; no product behavior or public API affected, so no docs
update is needed.

### DIFF
--- a/tests/lib/test_inspectors_llm.py
+++ b/tests/lib/test_inspectors_llm.py
@@ -332,5 +332,8 @@ def test_inspector_noinfo_without_oname(capsys):
 def test_inspector_format_fields_str_smoke():
     insp = Inspector()
     out = insp._format_fields_str([("Title", "value")])
-    assert "Title" in out
-    assert "value" in out
+    # ``format_color`` yields a token list when pygments is installed and a
+    # plain string otherwise; normalise to text before asserting.
+    text = out if isinstance(out, str) else "".join(tok[1] for tok in out)
+    assert "Title" in text
+    assert "value" in text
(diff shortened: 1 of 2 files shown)


rautyrauty:
This is necessary to  avoid: [check-failed.log](https://github.com/user-attachments/files/28064481/check-failed.log)

anki-code:
Thanks @rautyrauty !

PR #6452, opened 2026-05-20
fix(xontribs): Invalidate import caches before discovering xontribs

rautyrauty:
## Motivation

A xontrib can be installed **after** the interpreter has started — via
`xpip` mid-session, or written into a tmp directory by a test.

`importlib`'s path finders cache directory listings keyed by the
directory mtime. On a filesystem with coarse mtime resolution, a freshly
written xontrib can fall into the **same mtime tick** as the cached
(stale) listing, so `FileFinder` never rescans and the xontrib stays
invisible to `find_spec()`. Xontrib discovery then silently misses it.

This surfaced as an arch/filesystem-dependent flake while building the
package downstream — the **same revision** failed on one builder while
passing on the others, in two different discovery paths:

- `tests/built_ins/test_xontribs.py::test_xontrib_unload` —
  `xontribs_load()` loaded nothing:
  ```
  assert "script" in xontribs_loaded()
  AssertionError: assert 'script' in []
  ```
- `tests/built_ins/test_xontribs.py::test_xontrib_info_xsh_file` —
  `xontrib info` reported the xontrib as missing:
  ```
  assert "xshmod.xsh" in out
  AssertionError: assert 'xshmod.xsh' in 'Xontrib xshmod is not installed.\n'
  ```

## Change

Call `importlib.invalidate_caches()` in the two xontrib discovery entry
points:

- `find_xontrib()` — used by load / reload / unload / context, and the
  `xontrib info` fallback lookup;
- `_get_installed_xontribs()` — used by `get_xontribs()`, behind
  `xontrib list` and `xontrib info`.

Discovery no longer depends on mtime granularity. This is also correct
product behaviour: xonsh explicitly supports installing xontribs
mid-session, and they were subject to the same stale cache.

## Tests

`test_find_xontrib_invalidates_import_caches` and
`test_get_xontribs_invalidates_import_caches` assert that each discovery
entry point invalidates importlib's path-finder caches. Both fail
without the change and pass with it.

(An earlier revision of this PR carried a test that poked
`FileFinder._path_mtime` directly to emulate the race; that proved
non-deterministic across import-system states and was replaced.)

One-line-per-function product change; no public API or behaviour change.

---

🤖 Reviewed against the repository `CLAUDE.md` instructions.

### DIFF
--- a/tests/built_ins/test_xontribs.py
+++ b/tests/built_ins/test_xontribs.py
@@ -1,5 +1,7 @@
 """xontrib tests, such as they are"""
 
+import importlib.util
+import os
 import sys
 
 import pytest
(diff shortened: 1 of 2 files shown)


rautyrauty:
This is necessary to avoid: [build-418753-x86_64.log](https://github.com/user-attachments/files/28067398/build-418753-x86_64.log)

anki-code:
Thanks @rautyrauty !

PR #6494, opened 2026-06-01
docs: clarify xontrib loading modes

### LINKED ISSUE #5125: `xontrib` tutorial should be tweaked to clarify more details, how to not do autoloading.. etc.
This took me a shamefully long time to understand, mostly due to the combination of the docs being organized somewhat *in a strange order* (in my humble opinion) as well as reading through a bunch of `awesome-rc-xxx` repos and the [template repo](https://github.com/xonsh/xontrib-cookiecutter) which also did not clarify what I hopefully now understand 😂 

Actually, reading other xontrib projects, the core `xonsh` source code and reversing engineering how other repos do things ended up being more helpful then the actual docs, which imo is not that ideal 😂 

So hopefully this issue serves to at least one of either, help others who were as confused about the details as I, and provide motivation to make the docs a bit more detailed (with potentially a more in depth example) to explain the multiple ways in which `xontrib`s can be written and registered for use.

As an example of a project I work on on which eventually offered something similar, check out the [complete example](https://pluggy.readthedocs.io/en/latest/#a-complete-example) section of the `pluggy` docs (the plugin sys for `pytest`).

To begin i'm going to summarize what i think i do and don't understand.. and here is the resource list i've been using:
- the tut: https://xon.sh/tutorial_xontrib.html#structure
- the template: https://github.com/xonsh/xontrib-template
- the source code: https://github.com/xonsh/xonsh/blob/main/xonsh/xontribs.py

In preface, i think my main critique is that there seems to be lots of "templating code / repos" for how to produce some `xontrib` but very little docs on the details of how to do such things manually 😉 

----
#### things i still don't grok:
- afaiu there are *2 main ways* to offer a `xontrib`:
  - through exposing it in some external package, as is most documented in the [structure docs](https://xon.sh/tutorial_xontrib.html#structure)
  - or, (presumably) through some other explicit loading (via `xontrib load -f` or the py api) from wtv namespaces are immediately available inside the `~/.xonshrc` environment?
    - why nothing in the docs mentions the `-f` flag is somewhat baffling, i had to read the source code to figure it out!
    - can you load such a `xontrib` without it being in an environment registered package? for eg. if you load some sibling module to your `rc.xsh` file?
    - what would be the best practice for file org if you are presumably hacking a `xontrib` this way to start, with the eventual intention of packaging it as per the `setuptools` methodology as is well documented?
- there seems to be a difference between `xontrib`s which are **auto-loaded** and those which are not but instead are expected to be **manually** loaded, normally via `xontrib load -f` (again, took me way to long to figure out the `-f` :joy:)
- there are 2 ways to offer `xontrib`s as 2 different file types: `.xsh` files and `.py` python modules but why would you use one over the other?
  - `.xsh` is some superset of python, so why ever offer a `.py`?
  - why would you put a `_load_xontrib_()` hook inside a `.xsh` module?
    - it seems as though you can access global vars like `aliases:  dict[str, Callable]` directly? is there some advantage to using it over accessing the runtime env that `.xsh` files seem to be run with?
  - if there are good reasons for using one file type for the other, they should probably be documented?
- if you **don't** register a `.xsh` file using the "special" `setuptools` params (as mentioned below), but the ext pkg which contains the `xontrib/` dir **is installed in the `xonsh` py env**, you can still load the `xontrib` using `xontrib load <name>` ?

---
### Premises (for things i hopefully *do* grok?):
assuming I actually understand how things work now i just want to assert what i think is true:

- depending on the file type, there are 2 different ways to register the `xontrib` using `setuptools`, **but this is only important for them to be autoloaded**:
  - for `.py` modules, they must be exposed by some pkg using something like,
  ```python
  entry_points={
      "xonsh.xontribs": [
            "xontrib_name = ns.path.to.xontrib_py_module",
        ],
  },
  ```
  - for `.xsh` modules/files, they (supposedly?) must be exposed using something like,
  ```python
      package_dir={
        'xontrib': 'xontrib',
    },
    package_data={
        'xontrib': ['*.xsh'],
    },
  ```
  yes, i understand you can also expose `.py` modules this way, but why would you? I guess just so that you can org all xontribs in one dir?
- if you want to offer a `xontrib` that is intentionally **not autoloaded** you only need to write what? nothing except for the file and add it to the python path such that you can load it with `xontrib load [-f]`?

### My critique:
- in the tut, there's a file-tree breakdown but no follow up to the comments beside each leaf node for how or why you'd use each potential `xontrib` entry and have it organized this way:
  - there should be small examples of the different ways in which you can offer `xontrib`s using both file types, 
  - auto-loading vs. no auto-loading, 
  - how (and presumably why) you'd use `_load_xontrib_()` hooks in (only?) `.py` modules.
    - yes i realize the obvious answer is that you're implementing the `xontrib` file as a `.py` mod and thus want to hook into the runtime loading, but this isn't really explicitly documented..
- there **definitely needs to be** documentation on the `xontrib load -f` flag for doing fully qualified module loading!
  - pretty sure the tut is borked without it for the example snippet after the file-tree section: `xontrib load myext mpl mypkg.show`
  
---
### in summary:
I'm more then happy to put up a PR to detail the tutorial with the examples that would address all of my own quiffs (presuming I have at least some of this figured out). I'm super interested to hear any explanation for the points I'm still unsure on; i think getting the details fully laid out in the docs will definitely help with more `xontrib`s being produced by new users in the future 🏄🏼

224236927:
### Description

Clarifies the xontrib loading and authoring tutorial around the points raised in #5125.

The existing page already describes xontrib discovery, loading, entry points, and authoring hooks, but it does not spell out when to use `xontrib load -f`, how manual loading differs from entry-point autoloading, or how to choose between `.xsh`, `.py`, and `_load_xontrib_`.

This patch keeps the existing tutorial structure and adds short explanations in the relevant sections.

Fixes #5125.

### Changes

- Document `xontrib load -f` / `--full` for fully qualified module paths.
- Add the Python API equivalent with `full_module=True`.
- Clarify that autoloading and manual loading are separate choices.
- Add guidance for choosing `.xsh` vs `.py` xontrib files.
- Clarify when `_load_xontrib_` is useful and what happens without it.

### Verification

```powershell
git diff --check
rg -n "xontrib load -f|--full|manual-only|autoloading|\\$PYTHONPATH|_load_xontrib_|full_module=True" docs\xontrib.rst
```

Results:

- `git diff --check` passed.
- The targeted `rg` check found the new loading/autoloading/hook guidance in `docs/xontrib.rst`.

Note: `python -m sphinx --version` failed because Sphinx is not installed in the current Windows Python environment, so a full Sphinx docs build was not run locally.

### DIFF
--- a/docs/xontrib.rst
+++ b/docs/xontrib.rst
@@ -82,6 +82,17 @@ after xonsh has started up.
 
     xontrib load myext mpl mypkg.show
 
+By default, each name is resolved as a xontrib name. Xonsh first checks
+installed ``xonsh.xontribs`` entry points, then modules in the ``xontrib``
+namespace package, and finally top-level modules available on ``$PYTHONPATH``.
+
+Pass ``-f`` (``--full``) when the arguments are already full module paths and
+should be imported exactly as written:
+
+.. code-block:: xonsh
+
+    xontrib load -f mypkg.show mypkg.subpkg.done
+
 Pass ``-s`` (``--suppress-warnings``) to load every xontrib that is installed and
 silently skip any name that isn't. Useful in a :doc:`xonsh RC <xonshrc>` that is shared across machines
 where only a subset of xontribs is installed.
@@ -92,6 +103,7 @@ The same can be done in Python as well
 
     from xonsh.xontribs import xontribs_load
     xontribs_load(['myext', 'mpl', 'mypkg.show'])
+    xontribs_load(['mypkg.show'], full_module=True)
 
 A xontrib can be unloaded from the current session using ``xontrib unload``
 
@@ -110,6 +122,12 @@ to mark themselves available for autoloading using the below format.
 
 Here the module should contain ``_load_xontrib_`` function as described above.
 
+Autoloading and manual loading are separate choices. Entry points in the
+``xonsh.xontribs`` group let xonsh discover and autoload the xontrib at startup,
+unless autoloading is disabled. A xontrib can still be intentionally manual-only:
+install the module so it is importable, but omit the ``xonsh.xontribs`` entry
+point and load it explicitly with ``xontrib load`` when needed.
+
 .. note::
 
     Please make sure that importing the xontrib module and calling ``_load_xontrib_`` is fast enough.
@@ -127,6 +145,11 @@ Structure
 ================
 Xontribs are modules with some special functions written
 in either xonsh (``*.xsh``) or Python (``*.py``).
+
+Use ``*.xsh`` when the extension is mostly xonsh syntax or shell-oriented
+setup. Use ``*.py`` when the extension is regular Python code, benefits from
+standard Python tooling, or needs normal package imports. Both file types can
+participate in the same loading model.
 
 Here is a template:
 
@@ -157,6 +180,12 @@ Here is a template:
 
 This _load_xontrib_() function is called after your extension is imported,
 and the currently active :py:class:`xonsh.built_ins.XonshSession` instance is passed as the argument.
+
+Use ``_load_xontrib_`` when loading should register aliases, environment
+variables, event handlers, completers, or other session state. Return values
+from ``_load_xontrib_`` are added to the current execution context. Without the
+hook, xonsh still supports simple modules by placing names from ``__all__`` (or
+public names when ``__all__`` is absent) into the context.
 
 .. note::
 


anki-code:
Let's start from splitting https://github.com/xonsh/xonsh/issues/5125 into a group of small issues with one question/task per issue and with repeatable examples or detailed description. After this we're welcome to review small PRs to fix/implement every fix.

PR #6495, opened 2026-06-01
fix(tools): preserve env var tildes in subprocess args

### LINKED ISSUE #5970: env: unexpected expanding string to path

Himanshuagrawal4:
## Summary
- Preserve tildes that come from environment variable expansion in subprocess arguments.
- Normalize `bytes` and `pathlib.Path` inputs before the reordered expansion path.
- Add unit and integration coverage for the reported `$QWE="~"` behavior.

## Root cause
`expand_path()` expanded environment variables before running expanduser, so `$QWE` with value `~` became a literal `~` and was immediately expanded to the home directory. Shell expansion order expands literal tildes before parameter expansion, so tildes introduced by env expansion should remain literal.

## Validation
- `python -m pytest tests\test_tools.py -q`
- `python -m pytest tests\xintegration\test_integrations.py::test_env_var_tilde_is_not_expanduser_in_subproc -q`
- `python -m ruff check xonsh\tools.py tests\test_tools.py tests\xintegration\test_integrations.py`
- `python -m ruff format --check xonsh\tools.py tests\test_tools.py tests\xintegration\test_integrations.py`
- `git diff --check`

Fixes #5970

### DIFF
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -1845,6 +1845,16 @@ def test_expand_path(expand_user, inp, expand_env_vars, exp_end, xession):
         assert path == "~" + exp_end
 
 
+def test_expand_path_preserves_tilde_from_envvar_value(xession):
+    tilde_path = os.path.join("~", "docs")
+    xession.env.update({"tilde": "~", "tilde_path": tilde_path})
+    xession.env["EXPAND_ENV_VARS"] = True
+
+    assert expand_path("$tilde") == "~"
+    assert expand_path("$tilde_path") == tilde_path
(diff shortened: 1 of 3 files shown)


anki-code:
I'm going to close this with the comment - https://github.com/xonsh/xonsh/issues/5970#issuecomment-4620977439

PR #6543, opened 2026-07-14
fix(tests): Drop the harness PYTHONPATH before completing pip

rautyrauty:
The root conftest prepends the source tree to PYTHONPATH so that subprocesses import xonsh from there. The pip completer hands the session env straight to its `pip` subprocess, so pip picks that up too and finds the build-time xonsh.egg-info sitting at the repo root, which shadows the really-installed xonsh distribution.

Whenever pip itself runs from a venv, its completion offers only the distributions local to that venv. The shadowed egg-info one is not, so `pip show <TAB>` has nothing left to offer and falls back to printing option flags, and test_completions fails. This is what any packaging build hits, where the suite runs inside a throwaway venv with xonsh as the only package installed into it.

No real shell exports PYTHONPATH like this, so unsetting it keeps the assertion honest rather than relaxing it.

## For community

### DIFF
--- a/tests/completers/test_pip_completer.py
+++ b/tests/completers/test_pip_completer.py
@@ -56,6 +56,15 @@ def pip_installed():
     ],
 )
 def test_completions(line, prefix, exp, check_completer, xsh_with_env):
+    # The root conftest prepends the source tree to PYTHONPATH, and the completer
+    # passes the session env straight to the `pip` subprocess. That puts the repo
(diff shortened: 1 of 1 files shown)


rautyrauty:
This is necessary to avoid: [build.builder0.log](https://github.com/user-attachments/files/30016867/build.builder0.log)

anki-code:
Thanks @rautyrauty !

PR #6547, opened 2026-07-19
fix(tools): accept a bare number for $DYNAMIC_CWD_WIDTH

chuenchen309:
### Problem

Setting `$DYNAMIC_CWD_WIDTH` to a bare number crashes, even though the string form works:

```xonsh
$DYNAMIC_CWD_WIDTH = 20     # TypeError: 'int' object is not subscriptable
$DYNAMIC_CWD_WIDTH = '20'   # fine -> (20.0, 'c')
```

`DYNAMIC_CWD_WIDTH` is documented as "Maximum length in number of characters ... For example, `"20"` is a twenty character width", so an unquoted `20` is the natural way to express it — but it raises a `TypeError` at assignment time instead of applying the width.

### Cause

`to_dynamic_cwd_tuple` (`xonsh/tools.py`) handled a `str` and a `(value, unit)` sequence, but a bare `int`/`float` fell through to `return (float(x[0]), x[1])`, which subscripts a scalar and raises.

### Fix

Add an `int`/`float` branch that returns `(float(x), unit)` with the default `"c"` unit. This mirrors the sibling `to_history_tuple`, which already defaults the unit for a bare number. Now `20`, `20.0`, and `"20"` all yield `(20.0, "c")`; the `"%"` and tuple forms are unchanged.

### Testing

Added `20` and `20.0` cases to the existing `test_to_dynamic_cwd_tuple` parametrize (they raise `TypeError` before this change, pass after). `tests/test_tools.py`: 576 passed. `ruff` clean.

---

*Disclosure: found, written and verified by an AI coding agent (Claude Code) on this account — it wrote the repro, the fix and this description. I review every change and am accountable for it. Happy to adjust anything.*


anki-code:
Thanks @chuenchen309 !

PR #6548, opened 2026-07-19
fix(tools): match stdlib for a trailing recursive glob in case-insensitive mode

chuenchen309:
### The bug

`_case_insensitive_iglob` walks a recursive `**` segment by collecting only directories. That is right when `**` is followed by another segment (`a/**/b.txt`), but a **trailing** `**` must also match files — stdlib `glob.glob('**', recursive=True)` returns files *and* directories.

Because `globsearch` runs case-insensitive globbing on non-Windows, a `g` `` `**` `` (or `` `some/dir/**` ``) therefore dropped every file under the tree and additionally emitted a spurious `.` / `''` self-entry:

```python
>>> from xonsh.tools import _case_insensitive_iglob as g
>>> sorted(g("**", recursive=True))
['.', 'Foo', 'Foo/sub']                       # files gone, spurious '.'
>>> sorted(glob.glob("**", recursive=True))    # stdlib
['Foo', 'Foo/bar.txt', 'Foo/sub', 'top.txt']
```

The module documents that it should mirror stdlib `glob`, and the case-sensitive path (`glob.iglob`) already does — case-sensitivity should only change *casing*, never *which* paths match.

### The fix

By the time the walk reaches a trailing `**`, the prefix is already resolved into real on-disk directories, so the subtree has no remaining case ambiguity — delegate a trailing `**` to stdlib `glob.iglob(..., recursive=True)`. Intermediate `**` keeps the directory-only walk. The empty current-directory self-match is also dropped so `**` / `**/` mirror stdlib for an implicit-cwd pattern.

`glob`'s `include_hidden` keyword is used for the dotfile case; it exists on all supported interpreters (`requires-python >= 3.11`).

### Verification

- Regression cases added to `tests/test_tools_iglob_llm.py`: `**`, `**/`, `dir/**` all fail before the fix (files dropped / spurious entries) and match stdlib after; case-insensitive resolution of the prefix is preserved (`foo/**` still finds `Foo/`).
- `pytest tests/test_tools_iglob_llm.py tests/test_tools.py tests/test_builtins.py` → 707 passed. `ruff check`/`ruff format --check` clean; `mypy xonsh/tools.py` unchanged from baseline.

---

This PR was authored by an AI coding agent (Claude Code) running on this account: the AI found the bug, ran the repro against stdlib glob, wrote the tests, and wrote this description. The human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff. If this isn't the kind of contribution you want, say so and I'll close it.


anki-code:
I need to think about this. It looks like fixing the regressed but the source of regression was determined wrong.

anki-code:
@chuenchen309 does that issue affects you in real life use case or it's just found from LLM?
If yes please provide an example.

chuenchen309:
Straight answer to the direct question: **it came from the agent, not from a real-life use case of mine.** As the description says, the AI found it, wrote the repro and the tests; there was no incident behind it and I'm not going to dress one up. Sorry for the slow reply.

So that you can judge whether it's worth anything, here is the divergence run against the helper as it stands on `main` today — a temp tree with `top.txt`, `Foo/bar.txt`, `Foo/sub/`:

```
'**'      xonsh  = ['.', 'Foo', 'Foo/sub']
          stdlib = ['Foo', 'Foo/bar.txt', 'Foo/sub', 'top.txt']
'**/'     xonsh  = ['', 'Foo/', 'Foo/sub/']
          stdlib = ['Foo/', 'Foo/sub/']
'Foo/**'  xonsh  = ['Foo', 'Foo/sub']
          stdlib = ['Foo/', 'Foo/bar.txt', 'Foo/sub']
```

The user-visible version is that on non-Windows, ``g`**` `` returns no files at all — only directories, plus a spurious `.`. That is reachable by anyone who types it, which is the argument for it mattering; it is not an argument that anyone has actually hit it.

On your other point — you're right that I didn't identify a regressing commit. I compared behaviour against stdlib because the docstring says this helper mirrors `glob`, and treated any divergence as the bug. If what you mean is that the real problem is the hand-rolled walker existing at all, and that the fix is to narrow or remove it rather than to keep patching individual cases, I think that's the better framing and this PR is the wrong shape for it. Tell me which way you'd rather go and I'll either rework it or close it — no attachment either way.

PR #6549, opened 2026-07-19
fix(tokenize): tokenize fR/FR/Rf triple-quoted raw f-strings on Python 3.11

chuenchen309:
### The bug

`xonsh/parsers/tokenize.py`'s `endpats` table was missing six triple-quoted keys — `fR'''`, `fR"""`, `FR'''`, `FR"""`, `Rf'''`, `Rf"""` — even though `triple_quoted` and the `StringPrefix` regex both list them. On **Python 3.11**, a legal raw f-string such as:

```python
Rf"""hi"""
```

reaches `endpats[token]` with a key that isn't in the table and raises `KeyError: 'Rf"""'`. The same happens for `fR"""..."""` and `FR"""..."""` (and the `'''` variants).

```
>>> from xonsh.parsers.lexer import Lexer
>>> Lexer().split('Rf"""hi"""')   # on 3.11
KeyError: 'Rf"""'
```

On Python 3.12+ the PEP 701 f-string tokenizer path is taken, so `endpats` is never consulted and the omission is masked — which is why CI on 3.12/3.13 stayed green while 3.11 (a supported, CI-tested interpreter) crashes.

### The fix

Add the six missing keys so `set(triple_quoted) - set(endpats)` is empty, matching the case-sibling prefixes (`rf`/`fr`/`Fr`/`rF`/`RF`) that already worked. These tables mirror CPython's `tokenize`, which generates every case permutation, so this is completing the transcription.

### Verification

- New parametrized regression test (all six prefixes × both triple-quote styles) in `tests/parsers/test_lexer.py`: fails with `KeyError` before the fix on 3.11, passes after.
- Python 3.11: `tests/parsers/test_lexer.py` → 115 passed. Python 3.12: `test_lexer.py` + fstring tests → 245 passed (the fix is a no-op there). `ruff check`/`ruff format` clean; `mypy` on the file unchanged from baseline.

---

This PR was authored by an AI coding agent (Claude Code) running on this account: the AI found the bug, reproduced the `KeyError` on Python 3.11, wrote the test, and wrote this description. The human account holder reviews every change and is accountable for it. The verification above is real and re-runnable from the diff. If this isn't the kind of contribution you want, say so and I'll close it.


anki-code:
Thanks @chuenchen309 !

Issue #5563, opened 2024-06-30
Refactoring: split `tools.py` into libs


anki-code:
`tools.py` has 2877 lines. We need to review the functions and:
* group functions by topics in `xonsh.lib`
* move functions to `xonsh.platforms` e.g. `hardcode_colors_for_win10`, `WIN10_COLOR_MAP`, `_win_bold_color_map`, etc

## For community
⬇️  **Please click the 👍 reaction instead of leaving a `+1` or 👍  comment**

PR #6556, opened 2026-07-28
refactor(platforms): move Windows color helpers

JJordan0C:
## Summary

- move the Windows 10 console color maps and `hardcode_colors_for_win10()` into `xonsh.platforms.win_colors`
- keep compatibility re-exports in `xonsh.tools` while updating the internal consumer to import from the focused module
- add coverage for the color mappings, bold-color conversion, color-depth setup, and compatibility exports

## Rationale

These helpers are specific to the Windows console host but were implemented in the general-purpose `xonsh.tools` module. Moving them into `xonsh.platforms` reduces that module's responsibilities while preserving downstream imports.

Refs #5563

## Tests

- `python -m ruff check xonsh/platforms/win_colors.py xonsh/tools.py xonsh/pyghooks.py tests/platforms/test_win_colors.py`
- `python -m pytest -q` — 6,821 passed, 901 skipped, 2 xfailed, 2 xpassed

### DIFF
--- a/tests/platforms/test_win_colors.py
+++ b/tests/platforms/test_win_colors.py
@@ -1,0 +1,31 @@
+from xonsh import tools
+from xonsh.platforms import win_colors
+
+
+def test_win10_color_map():
+    assert win_colors.WIN10_COLOR_MAP["ansired"] == "#c50f1f"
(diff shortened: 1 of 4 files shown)


anki-code:
Thanks @JJordan0C !

PR #6567, opened 2026-08-08
refactor(tools): move swap helpers to collections

Himanshuagrawal4:
## Summary

- move `swap` and `swap_values` from `xonsh.tools` to `xonsh.lib.collections`
- update internal consumers to import from the focused library module
- retain compatibility re-exports from `xonsh.tools` and add collection-helper coverage

Refs #5563

## Validation

- `ruff check` and `ruff format --check` on the modified Python files
- `pytest tests/lib/test_collections.py tests/test_tools.py` (`584 passed, 9 skipped`)

### DIFF
--- a/tests/lib/test_collections.py
+++ b/tests/lib/test_collections.py
@@ -1,4 +1,24 @@
-from xonsh.lib.collections import ChainDB
+from types import SimpleNamespace
+
+from xonsh.lib.collections import ChainDB, swap, swap_values
+
+
+def test_swap_restores_attribute():
+    namespace = SimpleNamespace(value="before")
+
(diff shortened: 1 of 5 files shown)


anki-code:
As I see `collections` - is a specific file for ChainDB functionality. Moving swap to this file is questionable. To me it's better to rename `collections.py` to `chaindb.py` or something similar. And keep `swap` into tools because it's small helper context manager.

Himanshuagrawal4:
That makes sense. The collections module is really about ChainDB, so moving the swap helper there would blur that boundary. I've closed this and left the helper where it is. Thanks for the direction.

Issue #6456, opened 2026-05-21
Intermittent: failing $()/@$() subprocess form does not raise CalledProcessError under load


rautyrauty:
### Summary

Under load, a failing captured (`$(...)`) or inject (`@$(...)`) subprocess
form intermittently does **not** raise `CalledProcessError`, even with
`$XONSH_SUBPROC_RAISE_ERROR` enabled. For a standalone `@$(...)` it then
surfaces as `XonshError: subprocess mode: command is empty`.

xonsh 0.23.7, Linux.

### Observed

Running the full `pytest` suite while building the package, the **same
revision** intermittently fails on loaded build hosts — seen on both
`aarch64` and `x86_64`. It passes when the test runs in isolation, so it
is timing-dependent:

```
    def test_inject_form_raises(xonsh_execer, raise_env):
        with pytest.raises(CalledProcessError):
>           xonsh_execer.exec("@$(ls /__nope__)\n")

xonsh/procs/specs.py:796: in build
    raise xt.XonshError("xonsh: subprocess mode: command is empty")
E   xonsh.tools.XonshError: xonsh: subprocess mode: command is empty
```

### Analysis

`_check_subproc_helper_raise` (`xonsh/built_ins.py`) decides whether to
raise:

```python
rtn = getattr(cp, "returncode", None)
if rtn is None or rtn == 0:
    return
```

This conflates "return code **undetermined**" (`None`) with "succeeded"
(`0`). `CommandPipeline.returncode` is a `blocking_property` that calls
`self.end()`, but `end()` is a no-op once `self.ended` is set (it already
ran in `_run_specs`), so it returns `self.proc.returncode` as-is — which
can still be `None` if the underlying proc has not been reaped yet.

When that happens the helper silently skips the raise; the inject helper
then returns an empty token list and the outer command is built empty.

### Suggested direction

The raise-check (or `CommandPipeline.returncode`) should not treat an
undetermined return code as success — it should obtain a definitive code
before deciding. Filing for maintainer input rather than a PR: reliable
reproduction needs a loaded machine and the fix touches the concurrent
pipeline code.

rautyrauty:
I couldn't reproduce the logs the second time:
[b20-x86_64.log](https://github.com/user-attachments/files/28104737/b20-x86_64.log)
[b20-aarch64.log](https://github.com/user-attachments/files/28104736/b20-aarch64.log)

anki-code:
[...]

PR #6584, opened 2026-08-24
refactor: move iterator helpers to library

Yuki9814:
## Summary

- move `get_portions` and `all_permutations` into `xonsh.lib.itertools`
- update the history caller to import from the focused library module
- keep explicit compatibility re-exports in `xonsh.tools`, including the existing keyword parameter names
- add generated tests in the policy-required module-mirroring `_llm.py` file

The public behavior stays the same: existing imports from `xonsh.tools` continue to resolve to the moved functions, while new internal code can use the focused library module directly.

Addresses #5563

## Validation

- iterator, compatibility, and tools tests: 582 passed
- history tests: 84 passed
- Ruff check and format checks pass
- targeted mypy check for all four changed files passes
- `git diff --check` passes

AI assistance was used for repository research, implementation, and test generation. I reviewed the final diff and validation results.

## For community
Please use the thumbs-up reaction instead of leaving a `+1` comment.


Yuki9814:
The single failing check matches the known intermittent `test_inject_form_raises` case from #6456: it expected `CalledProcessError` but received `XonshError: xonsh: subprocess mode: command is empty` under load.

The job finished with 7,719 passed and 1 failed. All other 12 matrix jobs passed, including Python 3.14 on macOS and Windows, and the failing test is outside this PR's changed files. I do not have repository permission to rerun the failed job; could a maintainer rerun it?

PR #6599, opened 2026-09-13
fix(ptk): handle Shift+Space and Shift+Backspace modifyOtherKeys reports

### LINKED ISSUE #6597: Shift+Space inserts `[27;2;32~` inside tmux
## Current Behavior

Inside tmux with `extended-keys on` (commonly enabled so applications can receive Shift+Enter), **Shift+Space inserts `[27;2;32~`** at the xonsh prompt and **Shift+Backspace inserts `[27;2;127~`**, instead of acting as Space and Backspace. This is easy to hit when typing `cmd | grep ...` quickly with Shift still held after the `|`.

It does not happen with the same shell outside tmux, or in tmux with `extended-keys off`.

### Cause

1. Before every prompt, xonsh enables xterm modifyOtherKeys level 1 by writing `\x1b[>4;1m` (`xonsh/shells/ptk_shell/__init__.py`), so the terminal can distinguish Shift+Enter from Enter.
2. However, only the Shift+Enter replies (`\x1b[27;2;13~` and `\x1b[13;2u`) are registered in prompt_toolkit's `ANSI_SEQUENCES` (`xonsh/shells/ptk_shell/key_bindings.py`). Any other modified-key report is unknown to the parser: it consumes the `ESC` and inserts the remainder as text.
3. tmux honors the request (`#{pane_key_mode}` becomes `Ext 1`). At level 1 it keeps only certain Ctrl/Meta combinations in legacy encoding (`input_key_mode1()` in tmux's `input-keys.c`), so Shift on keys that have no distinct shifted character, such as Space and Backspace, is reported as `CSI 27 ; 2 ; <codepoint> ~` (or `CSI <codepoint> ; 2 u` with `extended-keys-format csi-u`).

With a real terminal, the outer terminal also has to report modifiers to tmux (e.g. iTerm2 with tmux's `extkeys` terminal feature). For example, iTerm2 + tmux is affected, while Terminal.app + tmux and iTerm2 without tmux are not.

Side note: the comment above the request says "Mode 2", but the code sends level 1 (`\x1b[>4;1m`).

### Reproduction

No terminal emulator interaction is needed, because `send-keys` goes through the same tmux key encoder as a real key press:

```sh
tmux -L repro -f /dev/null new-session -d -s d -x 120 -y 20 'sleep 600' \; set -s extended-keys on
tmux -L repro new-window -d -t d:1 'xonsh --no-rc --no-env'
sleep 5  # wait for the prompt
tmux -L repro send-keys -t d:1 -l 'echo 1'; tmux -L repro send-keys -t d:1 S-Space
tmux -L repro send-keys -t d:1 -l '2x';     tmux -L repro send-keys -t d:1 S-BSpace
tmux -L repro send-keys -t d:1 -l '3'
tmux -L repro display -p -t d:1 '#{pane_key_mode}'
tmux -L repro capture-pane -p -t d:1
tmux -L repro kill-server
```

Result with `extended-keys on` (pane key mode `Ext 1`):

```
user@host ~ @ echo 1[27;2;32~2x[27;2;127~3
```

Result with `extended-keys off` (pane key mode `VT10x`):

```
user@host ~ @ echo 1 23
```

Traceback (if applicable): N/A, no exception is raised.

## Expected Behavior

Shift+Space behaves like Space and Shift+Backspace behaves like Backspace, as they do outside tmux. More generally, any modified-key report that xonsh enables via modifyOtherKeys should be parsed, not inserted as text.

Possible directions:

1. Parse extended key reports generically: map `CSI 27 ; <mod> ; <cp> ~` and `CSI <cp> ; <mod> u` for keys whose modifier carries no meaning of its own to the plain key (at minimum Space and Backspace), instead of registering only Shift+Enter.
2. Additionally, make the modifyOtherKeys request configurable via an environment variable, so users can opt out.

### Workaround

This xonshrc snippet uses the same single-character key slot technique as xonsh's `SHIFT_ENTER`, and re-feeds the plain key so it behaves identically in every editing mode. It works for me interactively:

```python
from prompt_toolkit.input import ansi_escape_sequences
from prompt_toolkit.key_binding.key_processor import KeyPress

# Private key slot -> (code point the terminal reports, plain byte the key stands for).
_SHIFT_NEUTRAL_KEYS = {
    "\x81": (32, " "),      # Shift+Space
    "\x82": (127, "\x7f"),  # Shift+Backspace
}

def _register_shift_neutral_key(bindings, slot, code, plain):
    ansi_escape_sequences.ANSI_SEQUENCES[f"\x1b[27;2;{code}~"] = slot  # xterm modifyOtherKeys
    ansi_escape_sequences.ANSI_SEQUENCES[f"\x1b[{code};2u"] = slot     # CSI u
    plain_key = ansi_escape_sequences.ANSI_SEQUENCES.get(plain, plain)

    @bindings.add(slot)
    def _feed_plain_key(event):
        event.app.key_processor.feed(KeyPress(plain_key, plain), first=True)

@events.on_ptk_create
def _normalize_shift_neutral_keys(bindings, **kw):
    for slot, (code, plain) in _SHIFT_NEUTRAL_KEYS.items():
        _register_shift_neutral_key(bindings, slot, code, plain)
```

## xonfig

Environment: macOS (Darwin 25.6.0), tmux 3.7c, iTerm2 3.7.1. The `xonfig` output below was captured with `xonsh --no-rc --no-env -c xonfig`, which is why shell type shows `none`; the interactive shell is `prompt_toolkit`.

<details>

```xsh
| xonsh                       | 0.24.2          |
| Python                      | 3.14.7          |
| prompt toolkit              | 3.0.53          |
| shell type                  | none            |
| history backend             | json            |
| pygments                    | 2.21.0          |
| on linux                    | False           |
| on darwin                   | True            |
```

</details>

## For community

anandghegde:
Before each prompt xonsh turns on modifyOtherKeys (`\x1b[>4;1m`) so that Shift+Enter can be told apart from Enter. tmux with `extended-keys on` then also reports Shift+Space as `CSI 27;2;32~` and Shift+Backspace as `CSI 27;2;127~` (or `CSI 32;2u` / `CSI 127;2u` with `extended-keys-format csi-u`). Only the Shift+Enter sequences were registered in prompt_toolkit's `ANSI_SEQUENCES`. For the other two, the parser dropped the `ESC` and inserted the rest of the sequence as text.

**Fix:** register both formats for Shift+Space and Shift+Backspace in `key_bindings.py`, right next to Shift+Enter. They use the same single-char key slot trick (`\x81`, `\x82`). Each slot's handler feeds the plain Space / Backspace key back into the key processor. The key then does whatever Space or Backspace does in the current mode: insert, delete, vi motions, search, autopair and so on. This follows the workaround @tsobczynski posted in the issue.

## Before / after

Using the script from the issue (tmux 3.7c, `pane_key_mode` = `Ext 1`). It types `echo 1`, S-Space, `2x`, S-BSpace, `3`, then Enter:

Before, `extended-keys-format xterm`:
```
@ echo 1[27;2;32~2x[27;2;127~3
```
Before, `extended-keys-format csi-u`:
```
@ echo 1[32;2u2x[127;2u3
```
After, both formats:
```
@ echo 1 23
1 23
@
```

## Tests

Added to `tests/shell/test_ptk_key_bindings_llm.py`. The tests pass raw bytes through prompt_toolkit's `Vt100Parser` and the key processor:
- both formats of Shift+Space / Shift+Backspace produce `echo 1 23`. Without the fix you get the literal sequences, as in the issue.
- in vi navigation mode Shift+Space moves the cursor the way Space does and does not insert anything.

`pytest tests/shell` passes (223 tests), and `pre-commit run` (ruff, ruff-format, mypy) passes on the changed files.

Shift+Escape is also reported as `CSI 27;2;27~` under the same settings. I left it out to keep this PR to the keys the issue covers, and can add it if you'd like.

Closes #6597


anandghegde:
The one red check here looks unrelated to this change.

`Test Python 3.13 windows-latest` fails on:

```
FAILED tests/xintegration/test_integrations.py::test_ampersand_argument[echo foo'&'-foo'&'\n]
  - subprocess.TimeoutExpired: Command '[...python.exe, '-m', 'xonsh', '--no-rc']' timed out after 20 seconds
= 1 failed, 7254 passed, 526 skipped, 2 xfailed, 2 xpassed in 631.85s =
```

That is an integration test spawning a xonsh subprocess and hitting the 20s timeout on a slow Windows runner. This PR only touches `modifyOtherKeys` handling for Shift+Space and Shift+Backspace in `ptk_shell/key_bindings.py`, which that test does not exercise — every other job, including the other Windows and Python 3.13 jobs, is green.

The branch is up to date with `main` (behind_by 0), so there is nothing to rebase onto, and I can't re-run the job myself. Could someone with write access re-run it? Happy to push a no-op commit instead if you'd rather retrigger that way.

anki-code:
Thanks @anandghegde !

PR #6604, opened 2026-09-15
fix(shell): do not write the terminal title escape when stdout is not a tty

doronz88:
## Problem

`BaseShell.settitle()` writes the terminal-title escape (`ESC ] 0 ; <title> BEL`) straight to fd 1 whenever `TERM` is set and `$TITLE` is not `None`. It runs before every subprocess command in interactive mode (`_shell_set_title` in `procs/specs.py`), so with a forced-interactive session whose stdout is a pipe or a file the escape lands in the captured output, ahead of each command's output:

```console
$ TERM=xterm-256color xonsh --no-rc -i -c 'echo hi' | od -c
0000000  033   ]   0   ;   e   c   h   o       |       u   s   e   r   @
0000020    h   o   s   t   :       ~       |       x   o   n   s   h  \a
0000040    h   i  \n
```

Anything parsing that output breaks — e.g. a GUI application that spawns `xonsh -l -i -c …` to run a command with the user's rc-file environment and then parses the command's JSON. Setting `$TITLE = None` or `""` does not help: an empty `ESC ] 0 ; BEL` is still emitted.

## Fix

Skip the write when fd 1 is not a tty. There is no terminal to title in that case, and the escape can only end up in a pipe or a file.

## After

```console
$ TERM=xterm-256color xonsh --no-rc -i -c 'echo hi' | od -c
0000000    h   i  \n
```

On a real terminal the title is still set (verified under `script(1)`: the `ESC ] 0 ;` sequence is present once).

## Tests

Two tests in `tests/shell/test_base_shell_llm.py`: the escape is written to fd 1 when `os.isatty(1)` is true, and nothing is written when it is false. `pytest tests/shell/` passes (230 tests); `ruff check` / `ruff format --check` clean on the touched files.

## For community


anki-code:
Thanks @doronz88 !

